Issue #1079, opened 2019-06-24
Implement firstSuccessOf


jdegoes:
Using `orElse`, add the following method:

```scala
trait ZIO[R, E, A] {
   def firstSuccessOf(rest: Iterable[ZIO[R, E, A]]): ZIO[R, E, A]
}

object ZIO {
   def firstSuccessOf[R, E, A](first: ZIO[R, E, A], rest: Iterable[ZIO[R, E, A]]): ZIO[R, E, A]
}
```

These methods will use `orElse` to reduce the non-empty iterable of effects into a single effect.

It may make sense to add `firstSuccessOfPar` as a synonym for `raceAll`, and deprecate `raceAll` as the preferred name.

peterlopen:
Should implementation in ZIO object look similar to this?
`rest.foldLeft[ZIO[R, E, A]](first)(_ orElse  _).refailWithTrace`

Adriani277:
Would like to pick this up if no has yet :)

PR #1085, opened 2019-06-25
Implement firstSuccessOf

Adriani277:
Ref #1079

### DIFF
--- a/core/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core/jvm/src/test/scala/zio/RTSSpec.scala
@@ -1217,7 +1217,7 @@ class RTSSpec(implicit ee: ExecutionEnv) extends TestRuntime {
 
   def testFirstSuccessOfFailures = {
     implicit val d
-    : Diffable[Left[Int, Nothing]] = Diffable.eitherLeftDiffable[Int] //    TODO: Dotty has ambiguous implicits
+      : Diffable[Left[Int, Nothing]] = Diffable.eitherLeftDiffable[Int] //    TODO: Dotty has ambiguous implicits
     unsafeRun(ZIO.firstSuccessOf(IO.fail(24).delay(10.millis), List(IO.fail(24))).either) must_=== Left(24)
   }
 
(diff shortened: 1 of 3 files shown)


core/jvm/src/test/scala/zio/RTSSpec.scala
```diff
@@ -1209,6 +1212,20 @@ class RTSSpec(implicit ee: ExecutionEnv) extends TestRuntime {
       b      <- effect.await
     } yield b) must_=== 42
 
+  def testFirstSuccessOfValues =
+    unsafeRun(IO.firstSuccessOf(IO.fail(42), List(IO.succeed(24))).either) must_=== Right(24)
+
+  def testFirstSuccessOfFailures = {
+    implicit val d
+      : Diffable[Left[Int, Nothing]] = Diffable.eitherLeftDiffable[Int] //    TODO: Dotty has ambiguous implicits
+    unsafeRun(ZIO.firstSuccessOf(IO.fail(24).delay(10.millis), List(IO.fail(24))).either) must_=== Left(24)
```
ghostdogpr: Maybe use a different value for the 2 failures, so that you check it returns the last one. Since both are 24 we can't be sure.

jdegoes: Due to jitter, sometimes the `10.millis` will be first. It's best to use `IO.never` here for the left value.

core/jvm/src/test/scala/zio/RTSSpec.scala
```diff
@@ -1209,6 +1212,20 @@ class RTSSpec(implicit ee: ExecutionEnv) extends TestRuntime {
...
+  }
+
+  def testFirstSuccessOfFailuresOneSuccess =
+    unsafeRun(ZIO.firstSuccessOf(IO.fail(42), List(IO.succeed(24).delay(1.millis))).either) must_=== Right(
+      24
```
jdegoes: Due to jitter, it's best to use `IO.never` for the left value.

core/jvm/src/test/scala/zio/RTSSpec.scala
Adriani277: Will make the changes

core/jvm/src/test/scala/zio/RTSSpec.scala
Adriani277: Will change, thanks

jdegoes: Ah, sorry, I misread the test. I think the delay is fine. 👍

jdegoes:
@Adriani277 Looks great. Thank you for your contribution, and....

giphy

...congratulations on your first ZIO pull request!

PR #1088, opened 2019-06-26
Fix the Dotty build, remove some workarounds for old Dotty versions, make the CI compile the tests with Dotty too

smarter:


smarter:
May I ask why you squash all commits together on merge ? Ideally, each commit should be meaningful on its own, squashing everything together makes the history much harder to understand.

iravid:
Not all contributors are unfortunately skilled with rebasing so they often end up with big branches of "WIP, WIP, fix error, fmt, ...", etc :-)

kaishh:
Also pull requests are generally quite small on their own. Typical dotty PR is the same size as largest ever ZIO PR by Wiem. ;)

Issue #1089, opened 2019-06-27
New combinator: catchAllCause


ghostdogpr:
Similar to `catchAll` but takes a `Cause[E]` instead of an `E` so that we can get execution traces easily.

softinio:
I'll take this related to other PRs I am doing. Thanks!

ghostdogpr:
@softinio as we agreed, I assigned this to one attendee at Scala Matsuri Hackathon. Thanks!

PR #1093, opened 2019-06-27
Implemented new combinator catchAllCause

yangzai:
Resolve https://github.com/zio/zio/issues/1089 .


yangzai:
Let me re submit this.

Issue #460, opened 2018-12-16
Promise's implementation should extends AtomicReference directly.


hepin1989:
Event the current encoding is great with `extends AnyVal` but without opaque type support which will still be a wrapper.

http://normanmaurer.me/blog/2013/10/28/Lesser-known-concurrent-classes-Part-1/

let's save memory by directly extends the AtomicReference or with an updater.

PR #1095, opened 2019-06-27
Implement 460

Jentsch:
With this PR Promise extends directly AtomicReference as proposed in #460. I skipped the alternative of using AtomicReferenceFieldUpdater, as I assume would not work with ScalaJS and I see no significant benefits.

I added two very simple benchmark cases to see the performance implications. The numbers below indicate that the new version is maybe a bit faster when wrapping is required, but the improvement is still in the error range. 

It would be nice to run then benchmark on a more suitable computer. A laptop with an open IDE and thermal throttling is maybe not most reliable way to measure performance.

I think a memory usage analyses isn't necessary, as it should be strictly lower than before.

Benchmark results:

| Benchmark                                  | Mode  | Cnt |       Score |        Error | Units |
|--------------------------------------------|-------|-----|------------:|-------------:|-------|
| Promise extends AnyVal (0dfdc99646701611908c56785941532a21450be8)           |       |     |             |              |       |
| IOPromiseBenchmark.listOfPromises          | thrpt | 25  |   2.703,992 |     ± 87,956 | ops/s |
| IOPromiseBenchmark.twoPromises             | thrpt | 25  | 681.622,489 | ± 25.155,348 | ops/s |
| Promise extends AtomicReference (4a6276bde1ca909796ba326a8f782e1996a2595a) |       |     |             |              |       |
| IOPromiseBenchmark.listOfPromises          | thrpt | 25  |   2.782,024 |     ± 32,184 | ops/s |
| IOPromiseBenchmark.twoPromises             | thrpt | 25  | 678.854,900 | ± 14.989,510 | ops/s |

### DIFF
--- a/benchmarks/src/main/scala/zio/IOPromiseBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IOPromiseBenchmark.scala
@@ -1,0 +1,39 @@
+package zio
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class IOPromiseBenchmark {
+
+  var length: Int = 500
+
+  @Benchmark
+  def listOfPromises(): Int = {
+    val task: UIO[Int] = for {
+      promises <- UIO.foreach(1 to length)(_ => Promise.make[Nothing, Int])
+      _        <- UIO.foreach_(promises)(_.succeed(1))
+      sum      <- UIO.foldLeft(promises)(0)((s, p) => p.await.map(_ + s))
+    } yield sum
+
+    IOBenchmarks.unsafeRun(task)
+  }
+
+  @Benchmark
+  def twoPromises(): Int = {
+    def task: UIO[Int] =
+      for {
+        one <- Promise.make[Nothing, Int]
+        two <- Promise.make[Nothing, Int]
+        _   <- one.succeed(1)
+        _   <- two.succeed(2)
+        o   <- one.await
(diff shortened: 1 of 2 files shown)


Jentsch:
I can't reproduce the test_211_jdk11_jvm. Is it a flaky test?
I'm not sure what mdoc does. I think it's an expected failure that Promise isn't a class anymore. Or not?

Jentsch:
Ok, somehow the failing test where executed again and now test_211_jdk11_jvm is green. Only mdoc is still red.

jdegoes:
@Jentsch What do you think about this? It's small enough I don't think we _need_ to merge it. If we do merge it, we'll have to update the interops modules in order to get tests passing because `mdoc` currently depends on them (something we need to change!).

Jentsch:
I personally have no attachment to this ticket. I just saw the #460 and thought that's something that I can contribute. The intent made sense to me, as ZIO has a focus on performance and the benchmarks indicate some performance improvements in some situations. On the downside this PR increase the complexity of Promise a bit.

I can't decide if this is good or bad trade off. Either way the effect is minor, so yes we don't need this to be merged. If no one wants this, I'm would suggest closing this PR and also #406.

Regrading the mdoc situation: wouldn't I best to split the documentation into a separate repository? Or maybe let every project have their own documentation? Right now the documentation in the zio/zio repository covers its own content, but also of the external interops repositories.

jdegoes:
@Jentsch 

> Regrading the mdoc situation: wouldn't I best to split the documentation into a separate repository? Or maybe let every project have their own documentation? Right now the documentation in the zio/zio repository covers its own content, but also of the external interops repositories.

I agree. I will open a ticket.

Jentsch:
This PR seems to be stale. I'm closing it.

PR #1099, opened 2019-06-28
Handle spurious wakeup in JVM OneShot

quelgar:
The spec for `Object#wait` allows early return without `notify` being called, a so-called _spurious wakeup_. To handle for this, `wait` should be called in a loop.

It so happens that when running on Hotspot, such spurious wakeups do not occur when running the test suite. If running on [OpenJ9](https://www.eclipse.org/openj9/) instead of Hotspot, then spurious wakeups do occur, leading to many failed tests.

Limiting `wait` to reliably block for at most X ms requires some tedious book keeping. I've split out the common case of "block forever" in its own method, as this doesn't require the book keeping overhead.

While I'm here: bizarrely, the OpenJ9 implementation of `Thread.sleep` does not sleep at all if the argument is the kind of ridiculously large value that would only ever appear in test code.

### DIFF
--- a/core/jvm/src/main/scala/zio/internal/OneShot.scala
+++ b/core/jvm/src/main/scala/zio/internal/OneShot.scala
@@ -46,17 +46,19 @@ private[zio] class OneShot[A] private (@volatile var value: A) {
   /**
    * Retrieves the value of the variable, blocking if necessary.
    *
-   * @param timeout The maximum amount of time the thread will be blocked.
+   * @param timeout The maximum amount of time the thread will be blocked, in milliseconds.
    * @throws Error if the timeout is reached without the value being set.
    */
   final def get(timeout: Long): A = {
-    var remaining = timeout
-    while (value == null && remaining > 0L) {
-      val start = System.currentTimeMillis()
+    var remainingNano = math.min(timeout, Long.MaxValue/1000000L)*1000000L
+    while (value == null && remainingNano > 0L) {
+      val waitMilli = remainingNano / 1000000L
+      val waitNano = (remainingNano % 1000000L).toInt
+      val start = System.nanoTime()
       this.synchronized {
-        if (value == null) this.wait(remaining)
+        if (value == null) this.wait(waitMilli, waitNano)
       }
-      remaining -= System.currentTimeMillis() - start
+      remainingNano -= System.nanoTime() - start
     }
 
     if (value == null) throw new Error("Timed out waiting for variable to be set")
--- a/core/jvm/src/main/scala/zio/internal/OneShot.scala
+++ b/core/jvm/src/main/scala/zio/internal/OneShot.scala
@@ -45,18 +45,39 @@ private[zio] class OneShot[A] private (@volatile var value: A) {
 
   /**
    * Retrieves the value of the variable, blocking if necessary.
+   *
+   * @param timeout The maximum amount of time the thread will be blocked.
+   * @throws Error if the timeout is reached without the value being set.
    */
-  final def get(timeout: Long = Long.MaxValue): A = {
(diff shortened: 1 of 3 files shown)


core/jvm/src/main/scala/zio/internal/OneShot.scala
```diff
@@ -45,18 +45,39 @@ private[zio] class OneShot[A] private (@volatile var value: A) {
...
+    var remaining = timeout
+    while (value == null && remaining > 0L) {
+      val start = System.currentTimeMillis()
       this.synchronized {
-        if (value == null) this.wait(timeout)
+        if (value == null) this.wait(remaining)
       }
-
-      if (value == null) throw new Error("Timed out waiting for variable to be set")
+      remaining -= System.currentTimeMillis() - start
```
hepin1989: nano

quelgar: ❓ I don't follow.

miriarte: @hepin1989 We shouldn't use System.nanotime as it can provide negative values: https://stackoverflow.com/questions/510462/is-system-nanotime-completely-useless/54566928#54566928

quelgar: Now that I look into it, we _should_ use `nanoTime`. Yes, it can return negative values, but that doesn't matter because all it can be used for is subtraction from another `nanoTime` reading. The difference is all we need to know.

I hadn't previously realised that `currentTimeMillis` is not ideal for measuring elapsed time, as it is a wall-clock time that can shift around if the system time is corrected.

Unfortunately the `wait` API doesn't have a variant taking a single nanoseconds long parameter 😠 .

core/jvm/src/main/scala/zio/internal/OneShot.scala
```diff
@@ -45,18 +45,41 @@ private[zio] class OneShot[A] private (@volatile var value: A) {
 
   /**
    * Retrieves the value of the variable, blocking if necessary.
+   *
+   * @param timeout The maximum amount of time the thread will be blocked, in milliseconds.
+   * @throws Error if the timeout is reached without the value being set.
    */
-  final def get(timeout: Long = Long.MaxValue): A = {
-    if (value == null) {
+  final def get(timeout: Long): A = {
```
quelgar: This is simpler if we use the `Lock` API, shame about the memory overhead:

```scala
  private val lock = new ReentrantLock()
  private val valueAvailable = lock.newCondition()

  final def get(timeout: Long): A = {
    var remainingNano = math.min(timeout, Long.MaxValue/1000000L)*1000000L
    while (value == null && remainingNano > 0L) {
      lock.lock()
      try {
        if (value == null) {
          remainingNano = valueAvailable.awaitNanos(remainingNano)
        }
      } finally {
        lock.unlock()
      }
    }

    if (value == null) throw new Error("Timed out waiting for variable to be set")

    value
  }
```

core/jvm/src/main/scala/zio/internal/OneShot.scala
```diff
@@ -45,18 +45,41 @@ private[zio] class OneShot[A] private (@volatile var value: A) {
...
+    var remainingNano = math.min(timeout, Long.MaxValue/1000000L)*1000000L
+    while (value == null && remainingNano > 0L) {
```
jdegoes: Can we factor `1000000L` into a constant (even a local one)? It is used many times.

quelgar:
✅ should be good to go.

jdegoes:
Looks great, @quelgar! Thanks for your contribution.

PR #1107, opened 2019-06-30
Add mapErrorCause method

kaishh:
add a `@see` for sibling defect-catching methods in scaladoc

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -357,6 +357,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self.foldCauseM(new ZIO.MapErrorFn(f), new ZIO.SucceedFn(f))
 
   /**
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -356,6 +356,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
 
+  /**
+   * Returns an effect with its full cause of failure mapped using the
+   * specified function. This can be used to transform errors while
+   * preserving the original structure of `Cause`.
+   *
+   * @see [[absorb]], [[sandbox]], [[catchAllCause]] - other functions for dealing with defects
+   */
+  final def mapErrorCause[E2](h: Cause[E] => Cause[E2]): ZIO[R, E2, A] =
+    self.foldCauseM(new ZIO.MapErrorCauseFn(h), new ZIO.SucceedFn(h))
```
jdegoes: Good name, seems useful! 👍

PR #1116, opened 2019-06-30
Fixed #1012 - added zoneId to Clock

### LINKED ISSUE #1012: Add `timezone` to DefaultRuntime.Environment
We are finding `zio` is a great way to mocking the non-deterministic aspects of our app.

One gap that caught us out though was the absence of a `timezone` field for us to override in any of the default modules, which resulted in different timestamps being generated in our CI environment when the unit tests ran.

To work around this, we've introduced a new `TimeZone` module, which uses `ZoneId.systemDefault` in `TimeZoneLive`, but in our tests we set this explicitly.

It seems a bit overkill to add a whole new module for this though, and `timezone` could instead fit alongside the other fields in `Clock`. Though perhaps then `Clock` needs to be broadened, and given a new name like `Time` in order to accommodate this.

spf3000:


core/shared/src/main/scala-2.12+/clock.scala
```diff
@@ -41,4 +42,10 @@ package object clock extends Clock.Service[Clock] {
    */
   final def sleep(duration: Duration): ZIO[Clock, Nothing, Unit] =
     ZIO.accessM(_.clock sleep duration)
+
+  /**
+   * Returns the java.time.ZoneId corresponding to the System timeZone
+   */
+  val timeZone: ZIO[Clock, Nothing, ZoneId] =
+    ZIO.accessM(_.clock.timeZone)
 }
```
scurvydoggo: Should we also consider introducing `currentDateTime`, as a high-level alternative to `currentTime`?

Something like:
```scala
    /** Get the current time, represented in the current timezone. */
    val currentDateTime: ZIO[Clock, Nothing, OffsetDateTime] = {
      for {
        millis <- currentTime(TimeUnit.MILLISECONDS)
        zone   <- timezone
      } yield OffsetDateTime.ofInstant(Instant.ofEpochMilli(millis), zone)
    }
```

I feel this is what the vast majority of people will be using `timezone` for, so it would be convenient to have it built-in.

spf3000:
@scurvydoggo I added another commit to address your previous comments.  I removed the zoneId method, as this felt quite repetitious now, seeing as you can get it from the OffsetZoneDateTime

testkit/jvm/src/main/scala/zio/testkit/TestClock.scala
```diff
@@ -38,13 +44,26 @@ case class TestClock(ref: Ref[TestClock.Data]) extends Clock.Service[Any] {
...
+
+  def offset(millis: Long): OffsetDateTime =
+    OffsetDateTime.ofInstant(Instant.ofEpochMilli(millis), ZoneId.of("Z"))
 
-  case class Data(nanoTime: Long, currentTimeMillis: Long, sleeps0: List[Duration])
+  case class Data(
+    nanoTime: Long,
+    currentTimeMillis: Long,
+    sleeps0: List[Duration],
+    currentDateTime: OffsetDateTime
```
jdegoes: There is duplicated information now across `currentTimeMillis` and `currentDateTime`. Instead, just store a `timezone`, you can default to UTC, and construct `currentDateTime` from `currentTimeMillis` and from `timezone`.

core/shared/src/main/scala-2.12+/clock.scala
fsvehla: That would be very useful IMO

core/shared/src/main/scala-2.12+/clock.scala
```diff
@@ -30,6 +31,12 @@ package object clock extends Clock.Service[Clock] {
   final def currentTime(unit: TimeUnit): ZIO[Clock, Nothing, Long] =
     ZIO.accessM(_.clock currentTime unit)
 
+  /**
+   * Get the current time, represented in the current timezone.
+   */
+  def currentDateTime: ZIO[Clock, Nothing, OffsetDateTime] =
+    ZIO.accessM(_.clock.currentDateTime)
+
```
scurvydoggo: @spf3000 I feel that the `timezone` should be accessible on its own, so that I don't have to generate a new timestamp in trying to get it.

I wouldn't really say the information is duplicated. Rather, `timezone` is a primitive from which you construct a higher-order value, `currentDateTime`. The important thing is that one depends on the other.

spf3000:
Please see change above with revisions in response to changes requested by @jdegoes. @scurvydoggo I tried to reinstate zoneId as per your suggestion - you can probably see if you look through the history.  The problem I ran into was the scalajs compilation always failed. I believe that this is because ZoneId has not been implemented for scalajs, so you get linking errors if you try to include it.  This may require some work on porting ZoneId into scalajs.  So I would suggest to raise a separate issue for it.

testkit/jvm/src/main/scala/zio/testkit/TestClock.scala
```diff
@@ -17,17 +17,22 @@
...
 import zio.clock.Clock
 import zio.testkit.TestClock.Data
+import java.time.{ Instant, OffsetDateTime }
 
 case class TestClock(ref: Ref[TestClock.Data]) extends Clock.Service[Any] {
 
   final def currentTime(unit: TimeUnit): UIO[Long] =
     ref.get.map(data => unit.convert(data.currentTimeMillis, TimeUnit.MILLISECONDS))
 
+  final def currentDateTime: UIO[OffsetDateTime] =
```
jdegoes: Can you add one test for this?

spf3000: Hey you’re welcome! So it doesn’t come up here because of the sequence of commit/ reset - force push I did, but there is actually not 1 but 2! additional tests in ClockSpec - if you look in the overall changes you should see them. Let me know if you feel a test is missing though. 👍

jdegoes:
@spf3000 Looks great, thank you for your work on this issue, and congratulations on your first ZIO pull request!

giphy

spf3000:
Thank you John!

PR #1117, opened 2019-06-30
Fix #1113 - Move Exit.Cause to top Level

### LINKED ISSUE #1113: Move Exit.Cause to top-level
`Cause` appears as a nested class inside the `Exit` companion object, but its importance and visibility has grown, and it makes sense to divorce it from the `Exit` type, by moving it to the top-level `zio` package.

We could introduce temporary `type Cause[+E] = zio.Cause[E]` and `val Cause = zio.Cause` synonyms inside `Exit` to maintain backward compatibility, but mark them `@deprecated`.

dorsev:
I would love some help with the check 'ci/circleci: test_211_jdk8_jvm'. not really sure why it failed.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -1,0 +1,378 @@
+/*
+ * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+sealed trait Cause[+E] extends Product with Serializable { self =>
+  import Cause._
+
+  final def &&[E1 >: E](that: Cause[E1]): Cause[E1] = Both(self, that)
+
+  final def ++[E1 >: E](that: Cause[E1]): Cause[E1] = Then(self, that)
+
+  final def defects: List[Throwable] =
+    self
+      .fold(List.empty[Throwable]) {
+        case (z, Die(v)) => v :: z
+      }
+      .reverse
+
+  final def died: Boolean =
+    self match {
+      case Die(_)            => true
+      case Then(left, right) => left.died || right.died
+      case Both(left, right) => left.died || right.died
+      case Traced(cause, _)  => cause.died
+      case _                 => false
+    }
+
+  final def failed: Boolean =
+    self match {
+      case Fail(_)           => true
+      case Then(left, right) => left.failed || right.failed
+      case Both(left, right) => left.failed || right.failed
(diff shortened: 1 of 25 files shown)


dorsev:
would love some help with the check 'ci/circleci: test_211_jdk8_jvm'. I am not really sure why it failed.

softinio:
Rebuilding to see if it helps resolve the issue via: 

https://circleci.com/workflow-run/ea0bba52-17cb-4eeb-9326-fa321a1192f8

ghostdogpr:
I re-ran it and it passes this time.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -0,0 +1,362 @@
+package zio
```
jdegoes: Don't forget the Apache copyright header, you can copy it from another file!

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -197,8 +197,8 @@ trait Fiber[+E, +A] { self =>
     UIO.effectTotal {
       val p: concurrent.Promise[A] = scala.concurrent.Promise[A]()
 
-      def failure(cause: Exit.Cause[E]): UIO[p.type] = UIO(p.failure(cause.squashWith(f)))
-      def success(value: A): UIO[p.type]             = UIO(p.success(value))
+      def failure(cause: zio.Cause[E]): UIO[p.type] = UIO(p.failure(cause.squashWith(f)))
```
jdegoes: Like @iravid mentioned, it would be nice to shorten these to `Cause` instead of `zio.Cause`, both here and elsewhere.

core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -233,370 +233,16 @@ object Exit extends Serializable {
...
-
-    final case class Fail[E](value: E) extends Cause[E] {
-      override final def equals(that: Any): Boolean = that match {
-        case fail: Fail[_]     => value == fail.value
-        case traced: Traced[_] => this == traced.cause
-        case _                 => false
-      }
-    }
+  @deprecated("use zio.Cause", "1.0.0")
+  type Cause[+E] = zio.Cause[E]
```
jdegoes: Also be sure to add `val Cause = zio.Cause`, with the same deprecation notice.

jdegoes:
Great work, just a few minor tweaks and will be good to go!

core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -193,10 +193,10 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
 
 object Exit extends Serializable {
 
-  final case class Success[A](value: A)        extends Exit[Nothing, A]
-  final case class Failure[E](cause: Cause[E]) extends Exit[E, Nothing]
+  final case class Success[A](value: A)            extends Exit[Nothing, A]
+  final case class Failure[E](cause: zio.Cause[E]) extends Exit[E, Nothing]
```
dorsev: @iravid I was not able to avoid mentioning the full package name here(since the object has a similar value with the same name).
Would love some help here.

dorsev:
@iravid @jdegoes Thank you for the review!

core/shared/src/main/scala/zio/Exit.scala
iravid: yeah that's probably unavoidable (unless you import and rename the top-level cause...)

iravid:
🎉🎉

jdegoes:
giphy

PR #1124, opened 2019-07-02
Alias `foreach` to `traverse` and `collectAll` to `sequence` (#1121)

### LINKED ISSUE #1121: Add traverse and sequence aliases
Try as I might, I keep reaching for these names instead of `foreach` and `collect`.

What do everyone think of re-adding these aliases?

A data point in favour of this is the equivalence with `Future.traverse`, `Future.sequence`.

These need to be added for:
- Exit
- ZIO
- ZManaged

dsebban:
Resolves #1121

### DIFF
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -206,6 +206,12 @@ object Exit extends Serializable {
         .map(_.reverse)
     }
 
+  /**
+   *  Alias for [[Exit.collectAll]]
+   */
+  final def sequence[E, A](exits: Iterable[Exit[E, A]]): Option[Exit[E, List[A]]] =
+    collectAll[E, A](exits)
+
   final def collectAllPar[E, A](exits: Iterable[Exit[E, A]]): Option[Exit[E, List[A]]] =
     exits.headOption.map { head =>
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -206,6 +206,12 @@ object Exit extends Serializable {
         .map(_.reverse)
     }
 
+  /**
+   *  Alias for `collectAll`.
```
iravid: Can you use `[[ZIO.collectAll]]` for the scaladoc references?

core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -206,6 +206,12 @@ object Exit extends Serializable {
         .map(_.reverse)
     }
 
+  /**
+   *  Alias for `[[ZIO.collectAll]]`.
```
iravid: Thanks! the backticks are unnecessary when you have the square brackets

dsebban: I generated the docs and checked the links work

iravid: Ah that's great then

Issue #1102, opened 2019-06-28
Add ZStream#partition


DamianReeves:
Add the following method to `ZStream`:

``` scala
def partition[B, C, R1 <: R, E1 >: E](p: A => ZIO[R1, E1, Either[B,C]]): ZManaged[R1, E1, (ZStream[R1, E1,B], ZStream[R1, E1, C])]
```

spf3000:
@DamianReeves wouldn't split accept a function of `p: A => Boolean` ? Similar to for example ZStream#filter:   `def filter(pred: A => Boolean): ZStream[R, E, A]`  - and then the signature you mentioned could be splitM similar to ZStream#filterM `  final def filterM[R1 <: R, E1 >: E](pred: A => ZIO[R1, E1, Boolean]): ZStream[R1, E1, A]`?

didierd:
What about the more general 

`def split[B, C, R1 <: R, E1 >: E](p: A => ZIO[R1, E1, Either[B,C]]): ZManaged[R1, E1, (ZStream[R1, E1,B], ZStream[R1, E1, C])]` ?

iravid:
@didierd I missed this comment - that's a good generalization! We should do that.

vasilmkd:
Is anyone working on this actively?

PR #1127, opened 2019-07-03
Introduce the ZStream#split operator - #1102

DamianReeves:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -725,6 +725,37 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+      _           <- self.fold[R, E, A, Unit].flatMap { fold =>
+        fold((), _ => true, folder)
+            .foldCauseM(
+              e => leftQueue.offer(Take.Fail(e)).fork.zipPar(rightQueue.offer(Take.Fail(e)).fork).unit.toManaged_,
+              _ => leftQueue.offer(Take.End).fork.zipPar(rightQueue.offer(Take.End).fork).unit.toManaged_
+            )
+            .unit
+            .fork
+      }
+      leftStream = ZStream.fromQueue[R,E,Take[E,A]](leftQueue).collectWhile {
```
iravid: replace `collectWhile` with `unTake` so that errors and ends are signalled

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -725,6 +725,37 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    for {
+      leftQueue   <- ZManaged.make(Queue.bounded[Take[E, A]](queueCapacity))(_.shutdown)
+      rightQueue  <- ZManaged.make(Queue.bounded[Take[E, A]](queueCapacity))(_.shutdown)
+      folder      = (_:Unit, item:A) => {
+        p(item).flatMap {
+          case true   => leftQueue.offer(Take.Value(item)).fork.unit
+          case false  => rightQueue.offer(Take.Value(item)).fork.unit
+        }
+      } 
+      _           <- self.fold[R, E, A, Unit].flatMap { fold =>
```
iravid: an equivalent (and simpler) way to do this is with `foreachManaged`:
```
self.foreachManaged { item => 
  // folder body here
}.foldCauseM(...).fork
```

DamianReeves:
> Left two comments inline. So did you test this? Is it misbehaving?

Added tests and updated not to fork the offers. This seems to work.

iravid:
Cool. So just please switch to using `foreachManaged`, run a format and we can get this in when the build passes!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -725,6 +725,33 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
         }
     }
 
+  def split(p : A => UIO[Boolean]) : ZManaged[R, E, (ZStream[R,E,A],ZStream[R,E,A])] = 
+    split(2)(p)
+
+  def split[B1,B2](queueCapacity : Int)(p : A => UIO[Boolean]) : ZManaged[R, E, (ZStream[R,E,A],ZStream[R,E,A])] = {
```
iravid: Instead of the overload, let's use a default parameter:
`def split[B1, B2](p: A => UIO[Boolean], queueCapacity: Int = 16): ZManaged[R, E, (ZStream[R,E,A],ZStream[R,E,A])]`

iravid: Also, please add scaladoc :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
DamianReeves: Do you mean replace `self.fold` with that i.e. starting at line **741**?

iravid: yup

jdegoes:
@DamianReeves No rush on this, please just re-open the pull request when you're ready to wrap it up! I will close for now just for hygiene.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: We should also call it `partition` IMO.

PR #1128, opened 2019-07-03
Remove sbt microsite as no longer in use

softinio:


ghostdogpr:
How about removing sbt-scoverage as well?

softinio:
> How about removing sbt-scoverage as well?

good idea

PR #1140, opened 2019-07-06
Introduce real companion object for Schedule - Fix #1111

### LINKED ISSUE #1111: Introduce real companion objects for Stream, Schedule, Sink, Managed, Queue
Currently the type aliases only have "virtual companion objects". We should introduce real companion objects that delegate to the methods in the most polymorphic object, but use restricted type parameter lists. This will provide the best possible user experience.

 - [x] Stream
 - [x] Schedule
 - [x] Sink
 - [ ] Managed
 - [x] Queue

dorsev:


core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -0,0 +1,154 @@
...
+  final def logInput[A](f: A => UIO[Unit]): Schedule[A, A] =
+    ZSchedule.logInput(f)
+
+  /**
+   * See [[ZSchedule.recurs]]
+   */
+  final def recurs(n: Int): Schedule[Any, Int] = ZSchedule.recurs(n)
+
+  /**
+   * See [[ZSchedule.delayed]]
```
ghostdogpr: Should be `spaced` here

dorsev: Thanks for the catch! fixed it

dorsev:
@ghostdogpr, regarding the failing tests I agree with you.
Would love any help debugging this since the tests are passing on my computer(and are passing CI occasionally)

dorsev:
Got a question while re-reading this code

Does it make sense that there are functions in ```ZSchedule``` that don't take an R parameter, and return a ```Schedule``` (like ```collectAll```)? 
Or do we expect that ```ZSchedule``` methods will return a ```ZSchedule``` and that all of them will be parameterized by an Environment type R?

dorsev:
OK, was able to reproduce the error locally.

dorsev:
Sorry for all the re-runs here, trying to figure out why my tests are failing.

Was able to reproduce it locally by running `++2.11.12 test` 3-4 times.
1. I changed all the code in RetrySpec and RepeatSpec to use ZSchedule(instead of Schedule) and all the tests were passing consistently I think. No idea why.

2.  It feels like the tests are 'hanging' or not finishing when this timeout happens(Increasing it to 120 seconds timeout did not help)

3. mdoc started failing and throwing an exception I do not know:
`Caused by: java.lang.NoSuchMethodError: zio.Runtime.unsafeRun(Lzio/ZIO;)Ljava/lang/Object;
	at zio.interop.reactiveStreams.QueueSubscriber$$anon$2.onSubscribe(QueueSubscriber.scala:71)
	at org.reactivestreams.example.unicast.RangePublisher.subscribe(RangePublisher.java:53)
	... 58 more`

4. when re-running the tests locally 10-20 times I found another inconsistent test that keeps on popping(at least on my local machine)
`get must fail if no value is set` in OneShotSpec.scala
for example
`[error]  14391574803 is not less than 10200000000 (OneShotSpec.scala:39)`

ghostdogpr:
@dorsev thanks for the contribution!

PR #1142, opened 2019-07-07
 Move Queue specific methods to its own companion object (#1111)

dsebban:


dsebban:
It seems there is no way to create a `ZQueue` from the companion object the only thing that was there was a way to create a simple `Queue`, I moved it to its own companion object. I am not sure if it makes sense, any feedback is welcome. Thanks

iravid:
@dsebban Can you rebase on master?

ghostdogpr:
@dsebban thanks for the contribution! I like how it separates the core queue implementation from the combinators.

PR #1144, opened 2019-07-07
Introduce real companion object for Sink (#1111)

### LINKED ISSUE #1111: Introduce real companion objects for Stream, Schedule, Sink, Managed, Queue

geymed:
Fix #1111 (for Sink)

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/Sink.scala
+++ b/streams/shared/src/main/scala/zio/stream/Sink.scala
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


geymed:
`mdoc` compilation fails, I suspect that because of this import: https://github.com/zio/interop-reactive-streams/blob/master/src/main/scala/zio/interop/reactiveStreams/SubscriberHelpers.scala#L4

iravid:
This is failing `mdoc` because the change is binary incompatible with the current version of reactive-streams, which is expected.

#1147 disables typechecking on that doc for now until we update interop-reactive-streams, so we can ignore that failure.

geymed:
@iravid thanks for reviewing this. Looks like all checks are passing now.

PR #1145, opened 2019-07-07
Fix CI release failure "SCM URL missing"

kaishh:
mirroring what other interop modules do. 🤷‍♀️ Resulting pom seems to have correct `<scm>` section


NeQuissimus:
I hate publishing artifacts so much :-)
@jdegoes needs to start a competitor startup to Sonatype already... :-P

softinio:
nice

PR #1147, opened 2019-07-08
Introduce Stream companion object  (#1111)

bobrowadam:


iravid:
ReactiveStreams docs are broken by this change, but it's expected. Should be fixed after we release ZIO and bump the RS dependency.

Issue #857, opened 2019-05-13
Improve error message when ZIO async callback is called more than 1 time


cecol:
I am trying to use `ZIO Queue` to consume messages from `google pubsub`.
The following is my code
```
import com.typesafe.scalalogging.Logger
import scalaz.zio.{App, IO, Task, ZIO, ZQueue}

import com.google.cloud.pubsub.v1.{AckReplyConsumer, MessageReceiver, Subscriber}
import com.google.pubsub.v1.{ProjectSubscriptionName, PubsubMessage}

object PubsubApiTest extends App {
  private[this] val log = Logger(this.getClass)

  override def run(args: List[String]): ZIO[PubsubApiTest.Environment, Nothing, Int] = {
    for {
      msgQueue <- ZQueue.bounded[(PubsubMessage, AckReplyConsumer)](100)
      subscriptionName: ProjectSubscriptionName = ProjectSubscriptionName.of("GCP Project", "Pubsub Subscription")
      _ <- IO.effectAsync[Throwable, Unit] { cb =>
        val sub = Subscriber
          .newBuilder(subscriptionName, new MessageReceiver() {
            override def receiveMessage(m: PubsubMessage, c: AckReplyConsumer): Unit = cb(msgQueue.offer((m, c)).map(_ => ()))
          })
          .build()
        sub.startAsync().awaitRunning() // run subscriber
      }.either
      _ <- msgQueue.take.flatMap { case (msg, consumer) =>
        Task {
          log.info(s"receive message id: {}", msg.getMessageId)
          consumer.ack()
        }.either
      }.forever
    } yield 1
  }
}
```

Above program can successfully receive the first message from pubsub, but fail at second message with following exception:
```
Exception in thread "main" Die(java.lang.ClassCastException: scala.runtime.BoxedUnit cannot be cast to scalaz.zio.ZIO)
Die(java.lang.ClassCastException: scala.runtime.BoxedUnit cannot be cast to scalaz.zio.ZIO)
scalaz.zio.FiberFailure: Fiber failed.
An unchecked error was produced.
java.lang.ClassCastException: scala.runtime.BoxedUnit cannot be cast to scalaz.zio.ZIO
	at scalaz.zio.internal.FiberContext.nextInstr(FiberContext.scala:412)
	at scalaz.zio.internal.FiberContext.evaluateNow(FiberContext.scala:164)
	at scalaz.zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:322)
[...]

PR #1153, opened 2019-07-09
clarify that callback must be called at most once for async effects

swachter:
Studying the source code I have the impression that the callback function that is provided by the register function for asynchronous effects must be called at most once. I added that information to scala doc.


ghostdogpr:
Nice 👍 it was also mentioned here https://github.com/zio/zio/issues/857

PR #1154, opened 2019-07-09
Add ZIO alias

ioleo:
We have all sorts of aliases for ZIO, but not for `ZIO[R, Nothing, A]`.
Following the `TaskR` pattern, I propose we add `UIOR` alias.


jdegoes:
@ioleo I really like this! 

We have several naming choices:

 - `UIOR`
 - `URIO`
 - `RUIO`

I would like to re-alias `TaskR` as `RIO`, after the Haskell library of the same name.

So I'd have a preference for the latter 2 aliases, either `URIO` or `RUIO`. 

Slight bias toward `URIO`, meaning "unexceptional `RIO`".

ioleo:
@jdegoes I like `RIO` and `URIO` pattern. I'll update the PR sometime this week.

ioleo:
@jdegoes Done. I expected more occurences of `TaskR`, but it turned out there were just a few :)

jdegoes:
@ioleo This looks good! However, we also have to keep `TaskR` for now (for at least several releases), but deprecate it. This means by companion object and type alias. Can you add them and deprecate them?

jdegoes:
@ioleo The code has changed slightly, can you rebase and we can get this merged?

ioleo:
@jdegoes Ok, finally, after retriggering the build several times all tests are green.

ioleo:
Retriggered build, finally passed <3

Issue #1027, opened 2019-06-16
Add throttling combinators to ZStream


iravid:
A token bucket implementation.

Variants:
- [x] `ZStream#throttleEnforce, ZStream#throttleEnforceM` - a version that drops elements that exceed the available bandwidth
- [x] `ZStream#throttleShape, ZStream#throttleShapeM` - a version that delays elements until sufficient bandwidth is available
- [ ] Accomodate bursts by allowing the token bucket to go into a (configurable) negative amount or by accumulating tokens over the available bandwidth up to a configurable amount

NeQuissimus:
It would be cool if there was an implementation of this that has some sort of `f: Token => UIO[Unit]`, which can be used to persist the tokens (think storing the token in a Kafka `KTable`)

iravid:
@NeQuissimus I think we could implement the token bucket as a `ZSink`, then its internal state can be persisted.

vasilmkd:
I would like to try my hand at this.

vasilmkd:
I would like to get some clarification on the requirements. Should there be a way to express how many tokens each stream element is worth (with the possibility of different elements requiring different number of tokens), or should I assume that they are all equal, e.g. one token? Thanks.

vasilmkd:
Also, would the ZSink's state extending Serializable make it persistable or am I looking for something else here?

vasilmkd:
I guess that each element can have a certain token cost by passing it through a costFn: A => Long or something equivalent.

iravid:
Hi @vasilmkd, thanks for looking into this. Indeed a cost function is required to compute the weight of each element. `A => Long` is ok, but `A => ZIO[R, E, Long]` is better :-)

Don't worry about making ZSink's state serializable. We can deal with that concern later.

vasilmkd:
Already on it.

vasilmkd:
Regarding the throttling combinator that accounts for burst.

In the currently merged `throttleEnforce` implementation, the bandwidth that can be achieved is at best the specified bandwidth in the arguments, meaning, the bucket at any time can contain at most `units` worth of tokens.

Could the burst combinator be implemented in such a way that would allow the bucket to accumulate `n` cycles worth of `units` tokens, such that in the case of delayed production of elements followed by a burst of elements all elements would go through if sufficient tokens are present?

Or is this proposal less user friendly than having a configurable negative threshold?

[...]

PR #1155, opened 2019-07-09
Add ZStream#throttle (#1027)

vasilmkd:
Token bucket throttling implementation.

I would have liked to use ZSchedule to control the bucket refilling, but I can't seem to figure out how to create a schedule which adds an initial delay to the refilling effect, otherwise, the bucket is initialized with the configurable initial capacity and immediately refilled without any delay. Any help is appreciated.

This is my first open-source contribution. Looking forward to the comments.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1082,13 +1082,12 @@ object ZSink extends ZSinkPlatformSpecific {
   final def throttle[R, E, A](units: Long, duration: Duration)(
     costFn: A => ZIO[R, E, Long]
   ): ZManaged[R with Clock, E, ZSink[R with Clock, E, Nothing, A, Option[A]]] = {
+    import ZSink.internal._
 
     def bucketSink(bucket: Ref[(Long, Long)]) = new ZSink[R with Clock, E, Nothing, A, Option[A]] {
-      import ZSink.internal._
-
       type State = (Ref[(Long, Long)], Option[A])
 
-      val initial = assertNonNegative(units) *> UIO.succeed(Step.more((bucket, None)))
+      val initial = UIO.succeed(Step.more((bucket, None)))
 
       def step(state: State, a: A) =
         for {
@@ -1113,6 +1112,7 @@ object ZSink extends ZSinkPlatformSpecific {
     }
 
     for {
+      _       <- assertNonNegative(units).toManaged_
       current <- clock.currentTime(TimeUnit.NANOSECONDS).toManaged_
       bucket  <- Ref.make((units, current)).toManaged_
     } yield bucketSink(bucket)
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1076,6 +1076,17 @@ object ZSink extends ZSinkPlatformSpecific {
   /**
    * Creates a sink which throttles input elements of type A according to the given bandwidth parameters
    * using the token bucket algorithm. Elements that do not meet the bandwidth constraints are dropped.
+   * The weight of each element is determined by the `costFn` function. Elements are mapped to
+   * `Option[A]`, and `None` denotes that a given element has been dropped.
+   */
+  final def throttleEnforce[A](units: Long, duration: Duration)(
+    costFn: A => Long
+  ): ZManaged[Clock, Nothing, ZSink[Clock, Nothing, Nothing, A, Option[A]]] =
+    throttleEnforceM[Any, Nothing, A](units, duration)(a => UIO.succeed(costFn(a)))
+
+  /**
+   * Creates a sink which throttles input elements of type A according to the given bandwidth parameters
+   * using the token bucket algorithm. Elements that do not meet the bandwidth constraints are dropped.
    * The weight of each element is determined by the `costFn` effectful function. Elements are mapped to
    * `Option[A]`, and `None` denotes that a given element has been dropped.
    */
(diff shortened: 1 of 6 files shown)


iravid:
Hey @vasilmkd, this is awesome! Thank you for this contribution, this is going to be super useful. I'll get to work on reviewing this tomorrow.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1063,4 +1094,50 @@ object ZSink extends ZSinkPlatformSpecific {
...
+
+    def refillFn(tokens: Long): Long = Math.min(bucketCapacity, tokens + refill)
+
+    def refillBucket(bucket: Ref[Long]) =
+      bucket.update(refillFn).unit.delay(refillInterval).repeat(ZSchedule.spaced(refillInterval))
+
+    for {
+      _      <- ZManaged.fromEffect(assertBucketConfiguration(bucketCapacity, initialTokens, refill))
+      bucket <- ZManaged.fromEffect(Ref.make(initialTokens))
+      sink   <- ZManaged.succeed(bucketSink(bucket))
```
iravid: I think you could just `yield bucketSink(bucket)` below; should be more efficient

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1063,4 +1094,50 @@ object ZSink extends ZSinkPlatformSpecific {
...
+      def extract(state: State) = UIO.succeed(state._2)
+    }
+
+    def refillFn(tokens: Long): Long = Math.min(bucketCapacity, tokens + refill)
+
+    def refillBucket(bucket: Ref[Long]) =
+      bucket.update(refillFn).unit.delay(refillInterval).repeat(ZSchedule.spaced(refillInterval))
+
+    for {
+      _      <- ZManaged.fromEffect(assertBucketConfiguration(bucketCapacity, initialTokens, refill))
```
iravid: The `ZManaged.fromEffect(x)` calls can be replaced by `x.toManaged_`

vasilmkd:
Awesome suggestions. I feel like I've overengineered the solution a bit. Thanks for the insightful comments.

streams/shared/src/main/scala/zio/stream/ZSink.scala

streams/shared/src/main/scala/zio/stream/ZSink.scala

vasilmkd:
I'm interested to know if it's possible to integrate the test environment under testkit/ (TestClock in particular for this issue). It would be helpful in order to write a deterministic throttling test. However, I think it should be done in a different PR, as it will probably involve some build file manipulation.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1063,4 +1072,49 @@ object ZSink extends ZSinkPlatformSpecific {
...
+
+  /**
+   * Creates a sink which throttles input elements of type A according to the given bandwidth parameters
+   * using the token bucket algorithm. The weight of each element is determined by the `costFn` effectful
+   * function. Stream elements are mapped to `Option[A]`, and `None` denotes that a given element has
+   * been throttled.
+   */
+  final def throttle[R, E, A](units: Long, duration: Duration)(
+    costFn: A => ZIO[R, E, Long]
+  ): ZManaged[R with Clock, E, ZSink[R with Clock, E, Nothing, A, Option[A]]] = {
```
iravid: So this doesn't need to be a `Managed` now right? The `Ref` and clock read can be done in `initial`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1063,4 +1072,49 @@ object ZSink extends ZSinkPlatformSpecific {
...
       def extractPure(state: State): Either[Nothing, B]        = Right(b)
     }
+
+  /**
+   * Creates a sink which throttles input elements of type A according to the given bandwidth parameters
+   * using the token bucket algorithm. The weight of each element is determined by the `costFn` effectful
+   * function. Stream elements are mapped to `Option[A]`, and `None` denotes that a given element has
+   * been throttled.
+   */
+  final def throttle[R, E, A](units: Long, duration: Duration)(
```
iravid: Let's call this `throttleEnforceM` and add a version `throttleEnforce` with `costFn: A => Long`

iravid:
@vasilmkd Would be great if you could create follow-up tickets for things that should be deferred :-)

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: I forgot to change this. I wanted to send some code in before leaving for a festival. Working on this PR right now.

vasilmkd: I have some trouble removing the `ZManaged` from the type signature. For each stream element, the `ZSink` gets reset (by calling `initial` state), which is why we need to keep a `Ref` as the state holder. Also, the ref initialization needs to happen exactly once, which means that we cannot make the ref in `initial`.

streams/shared/src/main/scala/zio/stream/ZSink.scala

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: And since `initial` doesn't take any arguments, there is no way to "inject" the old state.

iravid: Right, I see what you mean. Let's go ahead and merge this then.

PR #1161, opened 2019-07-10
Supervision overhaul; fixes #1081

### LINKED ISSUE #1081: Completely overhaul supervision
The drawbacks to the current supervision model are many:

 - It can consume lots of memory, potentially unlimited on Javascript
 - There are differences between JVM and Scala.js, because Scala.js does not support weak structures
 - Deterministic tests are difficult to write because of the non-determinism of garbage collection
 - It has to be explicitly turned on, and tracking child fibers has non-trivial costs

The main function of supervision is to interrupt child fibers when the parent has been interrupted.

This can be accomplished in a different way:

 - By allowing child fibers to share the same interruption variable as their parent

Because this behavior is not always desired, we can introduce a new `daemon` operator to severe the connection between parent and child.

### Benefits

 - No memory consumed for the feature
 - The model is the same across JVM and Scala.js
 - No reliance on non-deterministic garbage collection
 - Interruption-safe by default

As an example of the "safety by default", here is the simplest possible implementation of `zipPar`:

```scala
def zipPar[A, B](left: UIO[A], right: UIO[B]): UIO[(A, B)] = 
  for {
    fiberA <- left.fork
    fiberB <- right.fork
    a <- fiberA.join
    b <- fiberB.join
  } yield (a, b)
```

Notice how this implementation is "just correct" under the new model.

The new model makes it much safe to write resource-safe code using raw fibers.

### Drawbacks

We lose the ability to perform a "join" of the children fibers on exit of the parent. 

However, if this capability is desired, it can be implemented manually via explicit tracking of all child fibers.

### The Daemon Operator 

```scala
trait ZIO[R, E, A] { self =>
  def daemon: ZIO[R, E, A] = ZIO.daemon(self)
}
object ZIO {
   def daemon[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] = ???
}
```

### Forking All as Daemons 

```scala
ZIO.forkAll(list.map(_.daemon))
```

### Summary of Tasks

 - [ ] Introduce interruption variable inheritance
 - [ ] Introduce `daemon`
 - [ ] Delete `supervised`
 - [ ] Delete `unsupervised`
 - [ ] Delete `interruptChildren`
 - [ ] Delete `checkSupervised`
 - [ ] Delete `superviseStatus`
 - [ ] Delete `handleChildrenWith`

jdegoes:


jdegoes:
Will re-open this shortly.

Issue #1120, opened 2019-07-01
Enhance ZStream#spaced to emit the schedule's output into the stream as well


iravid:
Currently, `spaced` doesn't emit the schedule's output into the stream; it is ignored. To align the combinator more closely with `ZIO#repeat`, we should emit the schedule's output into the stream when the schedule finishes. 

The signatures for the new combinators should be:

```scala
def spacedEither[R1 <: R, B](schedule: ZSchedule[R1, A, B]): ZStream[R1 with Clock, E, Either[B, A]]
def spaced[R1 <: R, A1 >: A](schedule: ZSchedule[R1, A, A1]): ZStream[R1 with Clock, E, A1] = 
  spacedEither(schedule).map(_.merge)
```

dag2ww:
Hi,
if no one has started this one yet, I would like to take it as my first contribution.

iravid:
Hi @dag2ww, go ahead! I updated the description as the previous spec was not aligned with how `ZSchedule` works.

dag2ww:
OK, great! 
Thank you.

dag2ww:
Hi @iravid ,
Having the change implemented, I'm about to update test for `spaced` and would like to confirm the expected output/behaviour. 
What would you expect for `Stream(1, 2, 3).spaced(Schedule.recurs(1))` with Schedule's output emitted each time it finishes? 
Without Schedule's output emitted it was: 1, 1, 2, 2, 3, 3. 

Looking forward your comment on this. Thanks!

iravid:
@dag2ww Yes, every time it finishes.

PR #1162, opened 2019-07-10
Enhanced ZStream spaced to also emit schedule's output (#1120) 

dag2ww:
Hi,
@iravid, other contrivutors 

Please have a look at my changes and let me know your comments on these.

### DIFF
--- a/streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -114,8 +114,11 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     short circuits          $repeatShortCircuits
 
   Stream.spaced
-    spaced                  $spaced
-    short circuits          $spacedShortCircuits
+    spaced                        $spaced
+    spacedEither                  $spacedEither
+    repeated and spaced           $repeatedAndSpaced
+    short circuits in schedule    $spacedShortCircuitsWhileInSchedule
+    short circuits after schedule $spacedShortCircuitsAfterScheduleFinished
 
   Stream.take
     take                     $take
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -698,28 +698,39 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Analogical to `spaced` but with distinction of stream elements and schedule output represented by Either
+   */
+  def spacedEither[R1 <: R, B](schedule: ZSchedule[R1, A, B]): ZStream[R1 with Clock, E, Either[B, A]] =
+    new ZStream[R1 with Clock, E, Either[B, A]] {
+
+      override def fold[R2 <: R1 with Clock, E1 >: E, A1 >: Either[B, A], S]: Fold[R2, E1, A1, S] =
         ZManaged.succeedLazy { (s, cont, f) =>
-          def loop(s: S, sched: schedule.State, a: A): ZIO[R2, E1, S] =
+          def loop(s: S, schedSt: schedule.State, a: Either[B, A]): ZIO[R2, E1, S] =
```
iravid: The loop doesn't need to take an `Either[B, A]` as it will only ever be called with an `A`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -698,28 +698,39 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
             if (!cont(s)) ZIO.succeed(s)
             else
-              f(s, a).zip(schedule.update(a, sched)).flatMap {
-                case (s, decision) =>
-                  if (decision.cont && cont(s))
-                    loop(s, decision.state, a).delay(decision.delay)
-                  else IO.succeed(s)
+              f(s, a).zip(schedule.update(a.right.get, schedSt)).flatMap {
```
iravid: Doing `a.right.get` is unsafe, but it's actually unneeded here; once the definition of `loop` only uses an `A`, you won't need it.

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -776,7 +776,7 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       Stream(1, 2, 3)
         .spaced(Schedule.recurs(1))
         .run(Sink.collectAll[Int])
-        .map(_ must_=== List(1, 1, 2, 2, 3, 3))
+        .map(_ must_=== List(1, 1, 2, 2, 2, 2, 3, 3, 2))
```
iravid: Could you add a test that uses a schedule with a different output type?

streams/shared/src/main/scala/zio/stream/ZStream.scala
dag2ww: Sure, looks like it was too late for me to spot it ;P Thanks! Will change accordingly.

streams/shared/src/main/scala/zio/stream/ZStream.scala
dag2ww: In general I agree, however in this case we only put Left into Either upon stop, so no chance to get anything else than Right in the loop. 
Anyway, over all - fully agree, Either not needed here at all :)

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
dag2ww: Good idea - this one requires (at least I required) some time to find out why such output sequence...
Will try to come up with some more meaningful test for this.

iravid:
The failed test is known to be flaky, so merging this.

```
[error]     x supervise fibers in supervised
```

dag2ww:
Great! Really happy it's already into master :)
@iravid Thanks for review and hints!

Issue #1159, opened 2019-07-10
Ensure that two chunks can be compared for equality and hash code


jdegoes:
`Chunk#equals` and `Chunk#hashCode` should observe the following properties for chunks `c1: Chunk[A]`, `c2: Chunk[A]`, and `c3: Chunk[A]`:

1. **Transitivity**. If `c1 == c2` and `c2 == c3`, then `c1 == c3`.
2. **Symmetry**. If `c1 == c2`, then `c2 == c1`.
3. **Reflexivity**. `c1 == c1`.
3. **Negation**. `c1 != c2 == !(c1 == c2)`
4. **Substitutivity**. If `c1 == c2` then `f(c1) == f(c2)` for all functions `f`.
5. **Hash Consistency**. If `c1 == c2`, then `c1.hashCode == c2.hashCode`.

We need to add tests for all these properties in some test cases in `ChunkSpec.scala`.

For example, here is one test of transitivity:

```scala
def testTransitivity = {
  val c1 = Chunk(1, 2, 3)
  val c2 = Chunk(1, 2, 3)
  val c3 = Chunk(1, 2, 3)
  ((c1 == c2) && (c2 == c3) && (c1 == c3)) must beTrue
```

Other tests can be added in similar fashion by following the coding conventions in `ChunkSpec.scala`.

In addition, we should probably ensure that `Chunk#toString`

Vilkina:
I'll take it

Vilkina:
Sorry, I don't know how to add 'Scala' code here. So, something like:
<img width="959" alt="Screen Shot 2019-07-10 at 9 56 43 PM" src="https://user-images.githubusercontent.com/3358827/60997597-da3a8680-a35f-11e9-83d3-0fb6f871b0d9.png">

iravid:
Resolved in #1163

PR #1163, opened 2019-07-11
Ensure that two chunks can be compared for equality and hash code #1159

Vilkina:


jdegoes:
@Vilkina Looks superb! 

Just one small change: in order to _run_ the tests, they must be in the specification, which is a string at the beginning of a class (this is because we use `Specs2` for running tests, and this is the format it requires). 

So at the beginning of `ChunkSpec.scala`, you can see:

```scala
def is: SpecStructure =
    "ChunkSpec".title ^
      s2"""
  chunk apply $apply
  chunk length $length
  chunk equality prop $equality
  chunk inequality $inequality
  flatMap chunk $flatMap
  map chunk $map
  materialize chunk $materialize
  foldLeft chunk $foldLeft
  filter chunk $filter
  drop chunk $drop
  take chunk $take
  dropWhile chunk $dropWhile
  takeWhile chunk $takeWhile
  toArray $toArray
  foreach $foreach
  concat chunk $concat
  An Array-based chunk that is filtered empty and mapped must not throw NPEs. $nullArrayBug
  toArray on concat of a slice must work properly. $toArrayOnConcatOfSlice
  toArray on concat of empty and integers must work properly. $toArrayOnConcatOfEmptyAndInts
  Chunk.filter that results in an empty Chunk must use Chunk.empty $filterConstFalseResultsInEmptyChunk
    """
```

You can add a line in the "spec string" for each test, something like:

```
Chunk reflexivity $testReflexivity
...
```

etc. Then run `testOnly *ChunkSpec*` from `sbt`, to confirm the tests are being run.

Then we'll be ready to merge!

Vilkina:
Screen Shot 2019-07-11 at 4 52 33 PM

jdegoes:
@Vilkina Thank you for the contribution! 🎉 

giphy

Vilkina:
@jdegoes, looking forward for next ticket)

jdegoes:
@Vilkina How about [this one](https://github.com/zio/zio/issues/1164)? 😉

PR #1168, opened 2019-07-12
Cleanup nvm install and make sure paths set before use

### LINKED ISSUE #1167: Microsite publishing doesn't work
```
#!/bin/bash -eo pipefail
export TRAVIS_BUILD_NUMBER="${CIRCLE_BUILD_NUM}"
export TRAVIS_COMMIT="${CIRCLE_SHA1}"
sudo chown -R $USER:$USER /tmp
nvm use
node -v
./sbt docs/docusaurusPublishGhpages

/bin/bash: line 3: nvm: command not found
Exited with code 127
```

softinio:
This should resolve #1167 

- Machine executor comes with nvm so no need to try to install it via curl
- making sure nvm env vars are set before using nvm

Note: I actually couldn't reproduce the problem locally or by ssh into circleci failed build's rebuild with ssh. All worked fine when I did that. Hoping this cleanup will fix it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,6 @@ install_nodejs: &install_nodejs
   - run:
       name: Install node.js
       command: |
-        curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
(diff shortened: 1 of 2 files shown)


iravid:
Thanks @softinio

PR #1175, opened 2019-07-12
Enhance support for option and either - Fix #1160

### LINKED ISSUE #1160: Enhance support for Option and Either
# Getter Helpers

Sometimes we have values like `UIO[Option[A]]`, and we wish the option to be non-empty; if the option is empty, we wish to fail the effect. Similarly for structures like `Task[Either[A, B]]`.

## Option Getter Helpers

Add these methods to `ZIO` trait:

```scala
trait ZIO[-R, +E, +A] {
  def someOrFailWith[B, E1 >: E](f: Throwable => E1)(implicit ev: A <:< Option[B]): ZIO[R, E1, B] = ???
  
  def someOrFail[B, E1 >: NoSuchElementException](implicit ev: A <:< Option[B], ev2: E <:< E1): ZIO[R, E1, B] = ???
}
```
### Usage 

Add a variety of tests for `someOrFail` and `someOrFailWith`.

```scala
val task: Task[Int] = UIO(Some(42)).someOrFail
```

## Either Getter Helpers
Add these methods to `ZIO` trait:

```scala
trait ZIO[-R, +E, +A] {
  def leftOrFailWith[B, C, E1 >: E](f: Throwable => E1)(implicit ev: A <:< Either[B, C]): ZIO[R, E1, B] = ???
  
  def leftOrFail[B, E1 >: NoSuchElementException](implicit ev: A <:< Either[B, C], ev2: E <:< E1): ZIO[R, E1, B] = ???

  def rightOrFailWith[B, C, E1 >: E](f: Throwable => E1)(implicit ev: A <:< Either[B, C]): ZIO[R, E1, C] = ???
  
  def rightOrFail[B, E1 >: NoSuchElementException](implicit ev: A <:< Either[B, C], ev2: E <:< E1): ZIO[R, E1, C] = ???
}
```
### Usage 

Add a variety of tests for `leftOrFail`, `leftOrFailWith`, `rightOrFail`, and `rightOrFailWith`.

```scala
val task: Task[Int] = UIO(Right(42)).rightOrFail
```

# Constructor Helpers

It would be nice to construct things like `UIO[Option[Int]]` directly rather than using repeated function application. To do this, some helpers can be added into `ZIO` companion object, and other companion objects:

```scala
def some[A](a: A): UIO[Option[A]] = ???
val none: UIO[Option[Nothing]] = ???

def left[A](a: A): UIO[Either[A, Nothing]] = ???
def right[B](b: B): UIO[Either[Nothing, B]] = ???
```

With corresponding tests in `RTSSpec.scala`.

dorsev:
Resolves #1160.

### DIFF
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -48,6 +48,10 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `zipPar` method does not swallow exit causes of loser. $testZipParInterupt
    Check `zipPar` method does not report failure when interrupting loser after it succeeded. $testZipParSucceed
    Check `orElse` method does not recover from defects. $testOrElseDefectHandling
+   Check `someOrFail` method extracts the optional value. $testSomeOrFailExtractOptionalValue
+   Check `someOrFail` method fails when given a None. $testSomeOrFailWithNone
+   Check `someOrFailException` method extracts the optional value. $testSomeOrFailExceptionOnOptionalValue
+   Check `someOrFailException` method fails when given a None. $testSomeOrFailExceptionOnEmptyValue
    Check uncurried `bracket`. $testUncurriedBracket
    Check uncurried `bracket_`. $testUncurriedBracket_
    Check uncurried `bracketExit`. $testUncurriedBracketExit
@@ -134,8 +138,10 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
       (unsafeRunSync(IO.foldLeft(l)(0)((_, _) => IO.fail("fail"))) must_=== unsafeRunSync(IO.fail("fail")))
   }
 
+  private val exampleError = new Error("something went wrong")
+
   def testDone = {
-    val error                         = new Error("something went wrong")
+    val error                         = exampleError
     val completed                     = Exit.succeed(1)
     val interrupted: Exit[Error, Int] = Exit.interrupt
     val terminated: Exit[Error, Int]  = Exit.die(error)
@@ -277,6 +283,23 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
         .and(thn must_=== Exit.die(ex))
         .and(fail must_=== Exit.succeed(()))
     }
+  }
+
+  def testSomeOrFailWithNone = {
+    val task: Task[Int] = UIO(Option.empty[Int]).someOrFail(exampleError)
+    unsafeRun(task) must throwA[FiberFailure]
+  }
+
+  def testSomeOrFailExtractOptionalValue = {
+    val task: Task[Int] = UIO(Some(42)).someOrFail(exampleError)
+    unsafeRun(task) must_=== 42
+  }
+
+  def testSomeOrFailExceptionOnOptionalValue = unsafeRun(ZIO.succeed(Some(42)).someOrFailException) must_=== 42
+
+  def testSomeOrFailExceptionOnEmptyValue = {
+    val task = ZIO.succeed(Option.empty[Int]).someOrFailException
+    unsafeRun(task) must throwA[FiberFailure]
   }
 
   def testUncurriedBracket =
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -56,6 +56,10 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `rightOrFail` method fails when given a Left. $testRightOrFailWithLeft
    Check `rightOrFailException` method extracts the Right value. $testRightOrFailExceptionOnRightValue
    Check `rightOrFailException` method fails when given a Left. $testRightOrFailExceptionOnLeftValue
+   Check `leftOrFail` method extracts the Left value. $testLeftOrFailExtractsLeftValue
+   Check `leftOrFail` method fails when given a Right. $testLeftOrFailWithRight
+   Check `leftOrFailException` method extracts the Left value. $testLeftOrFailExceptionOnLeftValue
+   Check `leftOrFailException` method fails when given a Right. $testLeftOrFailExceptionOnRightValue
    Check uncurried `bracket`. $testUncurriedBracket
    Check uncurried `bracket_`. $testUncurriedBracket_
    Check uncurried `bracketExit`. $testUncurriedBracketExit
@@ -320,6 +324,23 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
 
   def testRightOrFailWithLeft = {
     val task: Task[Int] = UIO(Left(1)).rightOrFail(exampleError)
+    unsafeRun(task) must throwA[FiberFailure]
+  }
+
+  def testLeftOrFailExceptionOnLeftValue = unsafeRun(ZIO.succeed(Left(42)).leftOrFailException) must_=== 42
+
+  def testLeftOrFailExceptionOnRightValue = {
+    val task: Task[Int] = ZIO.succeed(Right(2)).leftOrFailException
+    unsafeRun(task) must throwA[FiberFailure]
+  }
+
+  def testLeftOrFailExtractsLeftValue = {
+    val task: Task[Int] = UIO(Left(42)).leftOrFail(exampleError)
+    unsafeRun(task) must_=== 42
+  }
+
+  def testLeftOrFailWithRight = {
+    val task: Task[Int] = UIO(Right(12)).leftOrFail(exampleError)
     unsafeRun(task) must throwA[FiberFailure]
   }
 
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -52,6 +52,10 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `someOrFail` method fails when given a None. $testSomeOrFailWithNone
    Check `someOrFailException` method extracts the optional value. $testSomeOrFailExceptionOnOptionalValue
    Check `someOrFailException` method fails when given a None. $testSomeOrFailExceptionOnEmptyValue
+   Check `rightOrFail` method extracts the Right value. $testRightOrFailExtractsRightValue
+   Check `rightOrFail` method fails when given a Left. $testRightOrFailWithLeft
+   Check `rightOrFailException` method extracts the Right value. $testRightOrFailExceptionOnRightValue
+   Check `rightOrFailException` method fails when given a Left. $testRightOrFailExceptionOnLeftValue
    Check uncurried `bracket`. $testUncurriedBracket
    Check uncurried `bracket_`. $testUncurriedBracket_
    Check uncurried `bracketExit`. $testUncurriedBracketExit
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -787,8 +787,37 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
 
   final def left[R1 <: R, C]: ZIO[Either[R1, C], E, Either[A, C]] = self +++ ZIO.identity[C]
 
+  /**
+   * Returns a successful effect if the value is 'Left', or fails with the error f.
```
```suggestion
   * Returns a successful effect if the value is `Left`, or fails with the error `f`.
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -787,8 +787,37 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+   */
+  def leftOrFail[B, C, E1 >: E](f: E1)(implicit ev: A <:< Either[B, C]): ZIO[R, E1, B] =
```
iravid: Can we name `f` as `e`?

core/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -279,6 +293,57 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
+
+  def testRightOrFailWithLeft = {
+    val task: Task[Int] = UIO(Left(1)).rightOrFail(exampleError)
+    unsafeRun(task) must throwA[FiberFailure]
+  }
+
+  def testLeftOrFailExceptionOnLeftValue = unsafeRun(ZIO.succeed(Left(42)).leftOrFailException) must_=== 42
+
+  def testLeftOrFailExceptionOnRightValue = {
+    val task: Task[Int] = ZIO.succeed(Right(2)).leftOrFailException
```
iravid: Does this work without annotating `task` as `Task[Int]`?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -787,8 +787,37 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+    self.flatMap(ev(_).fold[ZIO[R, E1, B]](ZIO.succeed(_), _ => ZIO.fail(f)))
+
+  /**
+   * Returns a successful effect if the value is 'Left', or fails with a [[NoSuchElementException]].
+   */
+  def leftOrFailException[B, C, E1 >: NoSuchElementException](
```
iravid: Hmm why `E1 >: NoSuchElementException` rather than just returning `ZIO[R, NoSuchElementException, B]`? Since `E1` doesn't show up in the function's parameters (other than the `ev2` but that doesn't influence inference, I think), this might not infer properly (or worse inferred as `Any`)

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -787,8 +787,37 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+    implicit ev: A <:< Either[B, C],
+    ev2: E <:< E1
+  ): ZIO[R, E1, B] = self.map(_.left.get).mapError(ev2)
```
iravid: 2 problems here - this is throwing in `map` and will cause a defect (rather than a typed exception), and projections are deprecated in 2.13 if I am not mistaken.

Can you fold/pattern match on the Either instead?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1108,6 +1108,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+   * Extracts the optional value, or fails with the given error 'f'.
+   */
+  def someOrFail[B, E1 >: E](f: E1)(implicit ev: A <:< Option[B]): ZIO[R, E1, B] =
+    self.flatMap(ev(_).fold[ZIO[R, E1, B]](ZIO.fail(f))(ZIO.succeed))
+
+  /**
+   * Extracts the optional value, or fails with a [[NoSuchElementException]]
+   */
+  def someOrFailException[B, E1 >: NoSuchElementException](implicit ev: A <:< Option[B], ev2: E <:< E1): ZIO[R, E1, B] =
+    self.map(_.get).mapError(ev2)
```
dorsev: I really don't like the fact that I am using '.get' here and in the other *orFailException functions, since `.get` is unsafe.

The other implementation I tried here was:
```scala
self.mapError(ev2).flatMap(ev(_).fold[ZIO[R, E1, B]](???)(ZIO.succeed))
```
and in the "???" I tried the following:

1. ```scala
self.mapError(ev2)
```
seemed promising, but it returns a ```ZIO[R,E1,A]``` instead of ```ZIO[R,E1,B]``` as needed.
2 . `ZIO.fail` requires an error, And I don't have one(?), just an error transformation function

I Would love some help on how to avoid this.

File: core/shared/src/main/scala/zio/ZIO.scala [hunk 7]
```diff
@@ -787,8 +787,37 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Returns a successful effect if the value is 'Right', or fails with the given error 'f'.
+   */
+  def rightOrFail[B, C, E1 >: E](f: E1)(implicit ev: A <:< Either[B, C]): ZIO[R, E1, C] =
+    self.flatMap(ev(_).fold[ZIO[R, E1, C]](_ => ZIO.fail(f), ZIO.succeed(_)))
+
+  /**
+   * Returns a successful effect if the value is 'Right', or fails with a [[NoSuchElementException]].
+   */
+  def rightOrFailException[B, C, E1 >: NoSuchElementException](
```
iravid: Same as with `leftOrFailException`

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1108,6 +1137,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Extracts the optional value, or fails with the given error 'f'.
+   */
+  def someOrFail[B, E1 >: E](f: E1)(implicit ev: A <:< Option[B]): ZIO[R, E1, B] =
+    self.flatMap(ev(_).fold[ZIO[R, E1, B]](ZIO.fail(f))(ZIO.succeed))
+
+  /**
+   * Extracts the optional value, or fails with a [[NoSuchElementException]]
+   */
+  def someOrFailException[B, E1 >: NoSuchElementException](implicit ev: A <:< Option[B], ev2: E <:< E1): ZIO[R, E1, B] =
```
iravid: Same comments as with `leftOrFailException`

core/jvm/src/test/scala/zio/IOSpec.scala
dorsev: Great question. No.
The return type by the compiler is ```[B, C, E1 >:NoSuchElementException]zio.ZIO[Any,E1,B]```. 🤔 

also, changing the signature to 
```scala
def leftOrFailException[B, C](
    implicit ev: A <:< Either[B, C]
  ): ZIO[R, NoSuchElementException, B]```
does not work too.

File: core/shared/src/main/scala/zio/ZIO.scala (hunk 5 above)
Comment (dorsev) [reply core/shared/src/main/scala/zio/ZIO.scala]: I agree with both points. please see my comment above, I would love some help here if possible.

Comment (dorsev) [reply core/shared/src/main/scala/zio/ZIO.scala]: I agree on both points. 

I am struggling with the fold/pattern matching on the '*exception' methods and opened this PR to get feedback on how to do this properly(please see my comment above).

would love some help here if possible :)

File: core/shared/src/main/scala/zio/ZIO.scala (hunk 8 above)
Comment (evis) [reply core/shared/src/main/scala/zio/ZIO.scala]: This method does almost the same as `ZIO#get`, do we really need both?

File: core/shared/src/main/scala/zio/ZIO.scala [hunk 9]
```diff
@@ -1108,6 +1137,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Extracts the optional value, or fails with the given error 'f'.
+   */
+  def someOrFail[B, E1 >: E](e: E1)(implicit ev: A <:< Option[B]): ZIO[R, E1, B] =
+    self.flatMap(ev(_).fold[ZIO[R, E1, B]](ZIO.fail(e))(ZIO.succeed))
+
+  /**
+   * Extracts the optional value, or fails with a [[scala.NoSuchElementException]]
+   */
+  def someOrFailException[B, E1 >: NoSuchElementException](implicit ev: A <:< Option[B], ev2: E <:< E1): ZIO[R, E1, B] =
```
evis: This method does almost the same as `ZIO#get`, do we really need both?

core/shared/src/main/scala/zio/ZIO.scala
iravid: Sorry, I take back what I said about the `E1` - the bounds are necessary to unify `NoSuchElementException` with the existing `E`.

This should work without `right.get`:
```scala
  /**
   * Returns a successful effect if the value is `Left`, or fails with a [[scala.NoSuchElementException]].
   */
  def leftOrFailException[B, C, E1 >: NoSuchElementException](
    implicit ev: A <:< Either[B, C],
    ev2: E <:< E1
  ): ZIO[R, E1, B] =
    self.foldM(
      e => ZIO.fail(ev2(e)),
      a =>
        ev(a).fold(
          ZIO.succeed(_),
          _ => ZIO.fail(new NoSuchElementException)
        )
    )
```

core/jvm/src/test/scala/zio/IOSpec.scala
iravid: Actually, I tested and it looks like it infers properly - scalac says it's a `ZIO[Any, NoSuchElementException, Nothing]` which is accurate.

dorsev:
Thanks for the help @iravid !!! much appreciated! 
Fixed the commits accordingly.

A thought:  Should I add a custom "message" for every "NoSuchElementException" like in `None.get` or leave it empty?

iravid:
Yeah, a custom message would be a nice touch 👍

core/shared/src/main/scala/zio/ZIO.scala
dorsev: @evis, I agree with you.
I think the difference is that: someOrFailException gives you an exception in case of a failure while .get just return the unit value.

what do you think about expressing `.get` using `someOrFailException`? does that make sense? 

Something like
```scala
final def get[E1 >: E, B](implicit ev: A <:< Option[B]): ZIO[R, Unit, B] =
    self.someOrFail(()).mapError(_ => ())
```

iravid: Good idea @dorsev

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -447,8 +447,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Unwraps the optional success of this effect, but can fail with unit value.
    */
-  final def get[E1 >: E, B](implicit ev1: E1 =:= Nothing, ev2: A <:< Option[B]): ZIO[R, Unit, B] =
-    ZIO.absolve(self.mapError(ev1).map(ev2(_).toRight(())))
+  final def get[E1 >: E, B](implicit ev: A <:< Option[B]): ZIO[R, Unit, B] =
```
iravid: Oh, I forgot this had the `E =:= Nothing` constraint. I think we should leave this method alone; we're discarding an existing error here.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -787,8 +787,50 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+   * Returns a successful effect if the value is `Right`, or fails with the given error 'e'.
+   */
+  def rightOrFail[B, C, E1 >: E](e: E1)(implicit ev: A <:< Either[B, C]): ZIO[R, E1, C] =
+    self.flatMap(ev(_) match {
+      case Right(value) => ZIO.succeed(value)
+      case Left(_)      => ZIO.fail(e)
+    })
+
+  /**
+   * Returns a successful effect if the value is `Right`, or fails with a [[scala.NoSuchElementException]].
```
```suggestion
   * Returns a successful effect if the value is `Right`, or fails with a [[java.util.NoSuchElementException]].
```

iravid:
Two small comments and this is good to go.

jdegoes:
@dorsev Thank you for your contribution!

PR #1185, opened 2019-07-15
Add Random.shuffle method

ioleo:
The `zio.random.Random` is missing `shuffle` method, probably due to implicit argument in `scala.util.Random.shuffle` interface (AFAIK ZIO is implicit-free by design).

Neverthless having a shuffle method for some basic types like: `Array` or `List` would be really nice.


ioleo:
I guess for other types like `List` we converting to/from array and reuse this implementation.

ioleo:
@jdegoes In general for `lists.length > 1` there is always a chance the shuffle will yield a list in the same order. Here I test the simplest of scenarios, where the `nextInt` always returns the same index as the one being swapped with, but obviously there are many ways that can form a "ring" and end up with the same list.

And obviously there are even more combinations which yield a list in diffrent order. Here I chose to only check a "reversed order" scenario.

jdegoes:
@ioleo Looks great! Thanks for this contribution, and congratulations on your first ZIO contribution!

giphy

Issue #1182, opened 2019-07-14
Optimize parallel methods


jdegoes:
The parallel methods of ZIO require optimization:

 - [ ] `foreachPar` & variants
 - [ ] `collectAllPar` & variants
 - [x] `raceAll` & variants
 - [ ] `forkAll` & variants
 - [ ] `reduceAllPar` & variants
 - [ ] `mergeAllPar` & variants

Right now, these methods are implemented via _binary composition_. 

This creates deeply nested and lop-sided trees:

```
   /\
  /  \
 *    \
      /\
    /   \
   *    /\
       /  \
      *    \
           /\
          /  \
         *    *
```

All the intermediate propagation of results takes a long time and generates heap.

The ideal way to optimize many of these parallel methods is to use an `Array` (a flat data structure) and coordinate results through a `Ref` and maybe a `Promise`.

If someone wants to take on this ticket but needs more details, let me know and we can pair on one of them (maybe `forkAll`) to see an example of how to apply performance optimization in this fashion.

galosa:
I would love to take it

galosa:
I will start from forkAll :)

hepin1989:
if the size is 2, then use binary operation, otherwise, use array.

galosa:
Hey @jdegoes,
This is my first OSS contribution, I tried this and I'm struggling with the implementation for `forkAll`.
I would love to pair up and do this together.

jdegoes:
@galosa Let's do it! What time zone are you in?

galosa:
@jdegoes Thank you, my timezone is GMT+3.

vasilmkd:
Any progress on this? Not necessarily looking to directly contribute, just interested.

galosa:
Not yet, we didn't manage to sit on this yet. Since it taking so long feel free to take it.

simpadjo:
@jdegoes Let me try to do it.
I have a question though. 
The size of the result is not known beforehand so we can't just preallocate an array of the right size. We need a growable structure that supports `add(idx: Int, value: B)` operation. `java.util.ArrayList` is a perfect candidate but it's not js-friendly.
I really wouldn't like to reimplement it.

I'm going to collect results together with task indices and into scala.mutable.ArrayBuffer[(Int, B)]` then convert it into `Array[B]` (linear complexity) and then to list.

What do you think? Maybe there are existing data structures to reuse?

simpadjo:
[...]

PR #1187, opened 2019-07-16
Fix #1165 - foreachParN should not use more than 'n' fibers

### LINKED ISSUE #1165: foreachParN should not use more than 'n' fibers.
The current implementation of foreachParN uses a semaphore to restrict the number of active fibers working concurrently. This doesn't restrict the number of fibers actually being created, and results in some surprising behavior. For example

```scala
val rt = new DefaultRuntime {}
val n = 10
val seq = 0 to 1000000
val boom = ZIO.foreachParN[Any,Nothing, Int, Int](n)(seq)(UIO.succeed(_))
rt.unsafeRun(boom)
```

The expectation here would be that this works, but instead we run into an out of memory exception. 
```scala
[error] (zio-default-async-268-1441532751) java.lang.OutOfMemoryError: GC overhead limit exceeded
[error] java.lang.OutOfMemoryError: GC overhead limit exceeded
[error] 	at java.lang.Integer.valueOf(Integer.java:832)
[error] 	at scala.runtime.BoxesRunTime.boxToInteger(BoxesRunTime.java:67)
[error] 	at zio.internal.PlatformLive$ExecutorUtil$.$anonfun$makeDefault$1$adapted(PlatformLive.scala:67)
[error] 	at zio.internal.PlatformLive$ExecutorUtil$$$Lambda$4633/467302532.apply(Unknown Source)
[error] 	at zio.internal.PlatformLive$ExecutorUtil$$anon$2.yieldOpCount(PlatformLive.scala:115)
[error] 	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:217)
[error] 	at zio.internal.FiberContext.$anonfun$fork$1(FiberContext.scala:586)
[error] 	at zio.internal.FiberContext$$Lambda$4847/1031667563.run(Unknown Source)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```
A possible fix is to change the implementation to use a queue with a worker pool instead of using semaphores.

The general idea would be something like

```scala
  final def foreachParN[R >: LowerR, E <: UpperE, A, B](n: Long)(as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] =
    for {
      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
      env   <- ZIO.environment[R]
      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a))) // preserve order
      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork // publish work
      _     <- ZIO.collectAll(List.fill(n.toInt)(q.take.flatMap { case (p, a) => p.done(fn(a).provide(env).onError(_ => q.shutdown)) }.forever.fork)) // worker handles work
      res   <- ZIO.collectAll(pairs.map(_._1.await)).ensuring(q.shutdown) // wait for results
    } yield res
```

Other approaches than the above one are probably possible as well.

adamgfraser:
Following the suggestion in #1165 uses a queue to ensure that `foreachParN` only uses `n` fibers and thus can successfully traverse very large lists. I also added a test for this.

I noticed that the performance is still slow compared to `scala.concurrent.Future` and `monix.eval.Task` but I believe that some of these issues are being addressed in #1182 so think it makes sense to resolve that before optimizing this.

### DIFF
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -611,11 +611,11 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
   def testForeachParN_Interruption = {
     import zio.duration._
     val actions = List(
-      clock.sleep(100.millis) *> ZIO.fail("A"),
+      clock.sleep(1000.millis) *> ZIO.fail("A"),
       ZIO.succeed(1),
       ZIO.fail("C")
     )
-    val io = ZIO.foreachParN(4)(actions)(a => a).timeout(50.millis)
+    val io = ZIO.foreachParN(4)(actions)(a => a).timeout(500.millis)
     unsafeRun(io.either) must_=== Left("C")
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -609,13 +609,15 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
   }
 
   def testForeachParN_Interruption = {
-    import zio.duration._
-    val actions = List(
-      clock.sleep(1000.millis) *> ZIO.fail("A"),
-      ZIO.succeed(1),
-      ZIO.fail("C")
-    )
-    val io = ZIO.foreachParN(4)(actions)(a => a).timeout(500.millis)
+    val io = for {
+      latch <- Promise.make[Nothing, Unit]
+      actions = List(
+        latch.await *> ZIO.fail("A"),
+        ZIO.succeed(1),
+        ZIO.fail("C").ensuring(latch.succeed(()))
+      )
+      res <- ZIO.foreachParN(4)(actions)(a => a)
+    } yield res
     unsafeRun(io.either) must_=== Left("C")
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -585,7 +585,7 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
 
   def testForeachParN_Threads = {
     val n   = 10L
-    val seq = 0 to 1000000
+    val seq = 0 to 100000
     val res = unsafeRun(IO.foreachParN(n)(seq)(UIO.succeed))
     res must be_===(seq)
   }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -63,6 +63,7 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `collectM` returns failure ignoring value $testCollectM
    Check `reject` returns failure ignoring value $testReject
    Check `rejectM` returns failure ignoring value $testRejectM
+   Check `foreachParN` works on large lists $testForeachParN
     """
 
   def functionIOGen: Gen[String => Task[Int]] =
@@ -513,5 +514,12 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
     ).left.map(_.failureOrCause) must_=== Left(Left("Partial failed!"))
 
     goodCase and partialBadCase and badCase
+  }
+
+  def testForeachParN = {
+    val n   = 10L
+    val seq = 0 to 1000000
+    val res = unsafeRun(IO.foreachParN(n)(seq)(UIO.succeed))
+    res must be_===(seq)
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -77,6 +77,7 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `rejectM` returns failure ignoring value $testRejectM
    Check `foreachParN` works on large lists $testForeachParN_Threads
    Check `foreachParN` runs effects in parallel $testForeachParN_Parallel
+   Check `foreachParN` propogates error $testForeachParN_Error
     """
 
   def functionIOGen: Gen[String => Task[Int]] =
@@ -596,5 +597,13 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
       _ <- p.await
     } yield true
     unsafeRun(io)
+  }
+
+  def testForeachParN_Error = {
+    val ints = List(1, 2, 3, 4, 5, 6)
+    val odds = ZIO.foreachParN(4)(ints) { n =>
+      if (n % 2 != 0) ZIO.succeed(n) else ZIO.fail(s"$n is not odd")
+    }
+    unsafeRun(odds.either) must_=== Left("2 is not odd")
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -75,7 +75,8 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `collectM` returns failure ignoring value $testCollectM
    Check `reject` returns failure ignoring value $testReject
    Check `rejectM` returns failure ignoring value $testRejectM
-   Check `foreachParN` works on large lists $testForeachParN
+   Check `foreachParN` works on large lists $testForeachParN_Threads
+   Check `foreachParN` runs effects in parallel $testForeachParN_Parallel
     """
 
   def functionIOGen: Gen[String => Task[Int]] =
@@ -581,10 +582,19 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
     goodCase and partialBadCase and badCase
   }
 
-  def testForeachParN = {
+  def testForeachParN_Threads = {
     val n   = 10L
     val seq = 0 to 1000000
     val res = unsafeRun(IO.foreachParN(n)(seq)(UIO.succeed))
     res must be_===(seq)
+  }
+
+  def testForeachParN_Parallel = {
+    val io = for {
+      p <- Promise.make[Nothing, Unit]
+      _ <- UIO.foreachParN(2)(List(UIO.never, p.succeed(())))(a => a).fork
+      _ <- p.await
+    } yield true
+    unsafeRun(io)
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -609,15 +609,12 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
   }
 
   def testForeachParN_Interruption = {
-    val io = for {
-      latch <- Promise.make[Nothing, Unit]
-      actions = List(
-        latch.await *> ZIO.fail("A"),
-        ZIO.succeed(1),
-        ZIO.fail("C").ensuring(latch.succeed(()))
-      )
-      res <- ZIO.foreachParN(4)(actions)(a => a)
-    } yield res
+    val actions = List(
+      ZIO.never,
+      ZIO.succeed(1),
+      ZIO.fail("C")
+    )
+    val io = ZIO.foreachParN(4)(actions)(a => a)
     unsafeRun(io.either) must_=== Left("C")
   }
 }
--- a/core/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core/jvm/src/test/scala/zio/IOSpec.scala
@@ -78,6 +78,7 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `foreachParN` works on large lists $testForeachParN_Threads
    Check `foreachParN` runs effects in parallel $testForeachParN_Parallel
    Check `foreachParN` propogates error $testForeachParN_Error
+   Check `foreachParN` interrupts effects on first failure $testForeachParN_Interruption
     """
 
   def functionIOGen: Gen[String => Task[Int]] =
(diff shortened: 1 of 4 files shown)


adamgfraser:
Hmmm, not sure why `OneShotSpec` is failing on one of the builds. It does not appear to be related to any changes in this PR.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1869,12 +1869,16 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-             semaphore.withPermit(fn(a))
-           }
-    } yield bs).refailWithTrace
+    for {
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      env   <- ZIO.environment[R]
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(List.fill(n.toInt)(q.take.flatMap {
+            case (p, a) => p.done(fn(a).provide(env).onError(_ => q.shutdown))
```
freskog: After some experimenting I realized that p.done(fn(a)) implies that the effect produced by fn(a) will be evaluated by the fiber that invokes p.await. In this case that is the fiber that sequentially awaits all the promises.

In practice this means we don't have any parallelism with the proposed implementation.

A reasonable fix is to move the fn(a) call outside of p.done. Something like

fn(a).onError( _ => q.shutdown) >>= (p.succeed)

Might be a good idea to verify manually that the behavior is parallel as well. I can post some code later for that (away from my computer at the moment)

adamgfraser: You're totally right! Good catch! I just made the change you suggested and manually confirmed that effects are now running in parallel. I also added a test for this.

freskog: Another observation is that 

`p.done(fn(a).provide(env).onError(_ => q.shutdown))`

doesn't propagate how fn(a) failed. I think a better way of doing this is

`fn(a).foldM(e => p.fail(e) *> q.shutdown, b => p.succeed(b)).onError(_ => q.shutdown`

That should propagate the error correctly (though some tests should be added to verify this as well)

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1770,12 +1763,17 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+    for {
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(
+            List.fill(n.toInt)(
+              q.take.flatMap { case (p, a) => fn(a).onError(_ => q.shutdown).flatMap(p.succeed) }.forever.fork
+            )
+          )
+      res <- ZIO.collectAll(pairs.map(_._1.await)).ensuring(q.shutdown)
```
freskog: This could be rewritten using ZIO.foreach

core/shared/src/main/scala/zio/ZIO.scala
iravid: Best use `foldCauseM` and `p.halt` for that.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I fixed this so that errors are propagated and added a test for it. Looks like everything is working now except one of the CI tests for `RTSSpec` is failing. I'm not sure what is going on here. It does not appear to be related to any of the changes in this PR.

freskog: When using foldCauseM you don't need the onError.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Got it. It seems using a collection size of 1,000,000 elements in the tests was causing some flakiness (at different times we were failing on 2.13 / JDK 8, 2.13 / JDK 11, and 2.12 / JDK 8 in what appeared to be unrelated tests) so I reduced it to 100,000 and everything is working now. What do you think?

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Got it. It seems using a collection size of 1,000,000 elements in the tests was causing some flakiness (at different times we were failing on 2.13 / JDK 8, 2.13 / JDK 11, and 2.12 / JDK 8 in what appeared to be unrelated tests) so I reduced it to 100,000 and everything is working now. What do you think?

freskog: If you compare it to the original implementation that's still a big improvement. Sounds more than fair to me.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1763,12 +1763,17 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-             semaphore.withPermit(fn(a))
-           }
-    } yield bs).refailWithTrace
+    for {
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(
+            List.fill(n.toInt)(
+              q.take.flatMap { case (p, a) => fn(a).foldCauseM(e => p.halt(e) *> q.shutdown, b => p.succeed(b)) }.forever.fork
```
freskog: I wonder if the correct thing to do here is to halt all the promises, rather than just the current one. `q.shutdown` doesn't seem to reliably interrupt the other fibers the way I thought it would. The following seems to work a lot better

`fn(a).foldCauseM(c => ZIO.foreach(promises)(_._2.halt(c)), p.succeed)`

This seems to handle interruption better.

adamgfraser: Is there an example you're looking at where you're seeing this?

freskog: If you have a list like

List(sleep10sAndThenFail, succeed, failDirectly)

I would expect that to fail as soon as it's evaluated by foreachParN (given n > 2). In my testing it didn't actually fail directly, it needed to wait for 10s before it failed. It's possible I messed up my test (away from my computer at the moment).

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1763,12 +1763,15 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-    } yield bs).refailWithTrace
+    for {
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(List.fill(n.toInt)(q.take.flatMap {
+            case (p, a) => fn(a).foldCauseM(c => ZIO.foreach(pairs)(_._1.halt(c)), b => p.succeed(b))
+          }.forever.fork))
+      res <- ZIO.foreach(pairs)(_._1.await).ensuring(q.shutdown)
+    } yield res
```
kaishh: Gotta keep `refailWithTrace`, so that an error in child fiber points to the location of `foreachParN`. If you could also add a test for foreachParN similar to "foreachPar fail" in StacktracesSpec to prevent future regressions, that'd be great! 👍

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Fixed now. I think the issue was that previous approach was imposing a sequencing on the failures. So if we had `List(ZIO.fail("A"), ZIO.fail("B"))` we were guaranteed to get `ZIO.fail("A")`, which is great for determinism but means we have to wait for the first failure in the list if it takes longer than the other failure. In your example, if you switched the ordering of sleep10sAndThenFail and failDirectly it returns immediately.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: This is fixed now and I added a test. Thanks for your feedback on this!

jdegoes:
The other way of doing this (aside from a queue) is to put the semaphore around construction of the fiber, rather than running of the effect.

The old `foreachPar` put the semaphore around the running, instead of the construction. 

I _think_ this one can be as fast or faster, but I don't know for sure.

Does this need anything before a final review???

adamgfraser:
I think this is ready to review unless we want to try implementing the version based on putting the semaphore around the construction of the fiber like you suggested and benchmark the two implementations. My thought was since there is ongoing work on optimizing the parallel methods it would make sense to resolve that first but happy to give it a go if you think it would be helpful.

core/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -579,4 +583,39 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
 
     goodCase and partialBadCase and badCase
   }
+
+  def testForeachParN_Threads = {
```
jdegoes: How long do these tests take? If it's close to the timeout, the test will be flaky on CI.

core/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -579,4 +583,39 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
+    val odds = ZIO.foreachParN(4)(ints) { n =>
+      if (n % 2 != 0) ZIO.succeed(n) else ZIO.fail("not odd")
+    }
+    unsafeRun(odds.either) must_=== Left("not odd")
+  }
+
+  def testForeachParN_Interruption = {
+    import zio.duration._
+    val actions = List(
+      clock.sleep(1000.millis) *> ZIO.fail("A"),
```
jdegoes: This takes a whole second. Can we reduce the timeout, or use a latch? Test speed is critical as the test suite grows forever without bound.

core/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -579,4 +583,39 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
+      ZIO.succeed(1),
+      ZIO.fail("C")
+    )
+    val io = ZIO.foreachParN(4)(actions)(a => a).timeout(500.millis)
+    unsafeRun(io.either) must_=== Left("C")
```
jdegoes: This timeout may result in a flaky test; it's also half a second which is long to wait for the happy path.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1764,11 +1764,14 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-           }
-    } yield bs).refailWithTrace
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(List.fill(n.toInt)(q.take.flatMap {
+            case (p, a) => fn(a).foldCauseM(c => ZIO.foreach(pairs)(_._1.halt(c)), b => p.succeed(b))
+          }.forever.fork))
+      res <- ZIO.foreach(pairs)(_._1.await).ensuring(q.shutdown)
+    } yield res).refailWithTrace
```
jdegoes: There is a fiber-leak potential here: the for comprehension can be interrupted after the queue is created, but before `ensuring(q.shutdown)`. 

To solve this problem, you can use `bracket_`, e.g.:

```scala
Queue.bounded(n.toInt).bracket(_.shutdown) { q =>
  ...
}
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1764,11 +1764,14 @@ private[zio] trait ZIOFunctions extends Serializable {
...
     (for {
-      semaphore <- Semaphore.make(n)
-      bs <- foreachPar[R, E, A, B](as) { a =>
-             semaphore.withPermit(fn(a))
-           }
-    } yield bs).refailWithTrace
+      q     <- Queue.bounded[(Promise[E, B], A)](n.toInt)
+      pairs <- ZIO.foreach(as)(a => Promise.make[E, B].map(p => (p, a)))
+      _     <- ZIO.foreach(pairs)(pair => q.offer(pair)).fork
+      _ <- ZIO.collectAll(List.fill(n.toInt)(q.take.flatMap {
```
jdegoes: Instead of filling a `List`, we could do this in one go. It's not easy now, but I will add a ticket for `ZIO.fill`.

adamgfraser: Sounds good. I would be happy to pick up that ticket next.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I fixed this.

core/jvm/src/test/scala/zio/IOSpec.scala
adamgfraser: I refactored this test to avoid using timeouts.

core/jvm/src/test/scala/zio/IOSpec.scala
adamgfraser: I refactored this test to avoid using timeouts.

core/jvm/src/test/scala/zio/IOSpec.scala
adamgfraser: The one with a collection size of 100,000 takes about 2 seconds on my local machine. The challenge here is that you need quite a large collection size to verify the improvement from the previous implementation. On my local machine the previous implementation would work on collection sizes up to 50,000. Not sure how to test this more efficiently.

jdegoes:
@adamgfraser Thank you for this contribution! 

@freskog Thanks for reporting / prototyping and the reviews (/cc @iravid @kaishh ).

iravid:
🎉🎉 nice work @adamgfraser @freskog!

adamgfraser:
Thanks @jdegoes, @freskog, @iravid, @kaishh! Great experience contributing for the first time!

PR #1190, opened 2019-07-16
[WIP] Improve forkAll performance

galosa:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1411,14 +1412,58 @@ private[zio] trait ZIOFunctions extends Serializable {
    * Returns an effect that forks all of the specified values, and returns a
    * composite fiber that produces a list of their results, in order.
    */
+  import collection.JavaConverters._
   final def forkAll[R >: LowerR, E <: UpperE, A](as: Iterable[ZIO[R, E, A]]): ZIO[R, Nothing, Fiber[E, List[A]]] =
+    as match {
```
jdegoes: A simpler place to start may be: `Fiber.collectAll(iterable: Iterable[Fiber[E, A]]): Fiber[E, Seq[A]]` (assuming we don't already have that).

This would make a nicer building lock to implement this `forkAll`.

Should we try to do that one first?

jdegoes:
@galosa I will close this one for now just to do cleanup, but please ping me if you'd like to resume work!

PR #1191, opened 2019-07-16
Issue-1177 TArray

LGLO:
`TArray[A]` wrapper for `Array[TRef[A]]`
List of functions (quick draft, for sure some will be added, maybe some will be removed):
- [x] Tests for each method
- [x] apply
- [x] update, updateM
- [x] fold, foldM
- [x] map, mapM - copies to new `TArray`
- [x] foreach (doesn't make sense for `A => Unit`)
- [x] transform, transformM

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TArray.scala
+++ b/core/shared/src/main/scala/zio/stm/TArray.scala
@@ -21,12 +21,12 @@ package zio.stm
  * */
 class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
 
-  /** Extracts value from ref in array. Returns None when index is out of bounds. */
-  def apply(index: Int): STM[Nothing, Option[A]] =
-    if (0 <= index && index < array.size) array(index).get.map(Some(_))
-    else STM.succeed(None)
+  /** Extracts value from ref in array. */
+  final def apply(index: Int): STM[Nothing, A] =
+    if (0 <= index && index < array.size) array(index).get
+    else STM.die(new ArrayIndexOutOfBoundsException(index))
 
-  def collect[B](pf: PartialFunction[A, B]): STM[Nothing, TArray[B]] =
+  final def collect[B](pf: PartialFunction[A, B]): STM[Nothing, TArray[B]] =
     this
       .foldM(List.empty[TRef[B]]) {
         case (acc, a) =>
@@ -36,12 +36,12 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
       .map(l => TArray(l.reverse.toArray))
 
   /* Atomically folds [[TArray]] with pure function. */
-  def fold[Z](acc: Z)(op: (Z, A) => Z): STM[Nothing, Z] =
+  final def fold[Z](acc: Z)(op: (Z, A) => Z): STM[Nothing, Z] =
     if (array.isEmpty) STM.succeed(acc)
     else array.head.get.flatMap(a => new TArray(array.tail).fold(op(acc, a))(op))
 
   /** Atomically folds [[TArray]] with STM function. */
-  def foldM[E, Z](acc: Z)(op: (Z, A) => STM[E, Z]): STM[E, Z] =
+  final def foldM[E, Z](acc: Z)(op: (Z, A) => STM[E, Z]): STM[E, Z] =
     if (array.isEmpty) STM.succeed(acc)
     else
       for {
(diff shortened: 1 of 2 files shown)


LGLO:
@jdegoes I've opened this for quick review. Could you just tell me if I'm generally doing it right or is this totally missed shot?
Moar functions will come, documentation also.

jdegoes:
@LGLO Yes, generally speaking, this is right on! Nice work. 👍

LGLO:
Should I re-write tests to new framework?

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -0,0 +1,94 @@
...
+  /** Creates [[TArray]] of new [[TRef]]s, mapped with pure function. */
+  def map[B](f: A => B): STM[Nothing, TArray[B]] =
+    this.mapM(f andThen STM.succeed)
+
+  /** Creates [[TArray]] of new [[TRef]]s, mapped with transactional effect. */
+  def mapM[E, B](f: A => STM[E, B]): STM[E, TArray[B]] =
+    STM.foreach(array)(_.get.flatMap(f).flatMap(b => TRef.make(b))).map(l => new TArray(l.toArray))
+
+  /** Atomically updates all [[TRef]]s inside this array using pure function. */
+  def transform(f: A => A): STM[Nothing, Unit] =
```
jdegoes: I'd make all methods `final`.

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -0,0 +1,94 @@
...
+  /** Atomically updates all elements using transactional effect. */
+  def transformM[E](f: A => STM[E, A]): STM[E, Unit] =
+    (0 to array.size - 1).foldLeft[STM[E, Unit]](STM.succeed(())) {
+      case (tx, idx) =>
+        val ref = array(idx)
+        ref.get.flatMap(f).flatMap(a => ref.set(a)).flatMap(_ => tx)
+    }
+
+  /** Updates element in the array with given function. None signals index out of bounds. */
+  def update(index: Int, fn: A => A): STM[Nothing, Option[A]] =
```
jdegoes: We could also use `STM.die` for out of bounds. Thoughts?

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -0,0 +1,94 @@
...
+        acc2 <- op(acc, a)
+        res  <- new TArray(array.tail).foldM(acc2)(op)
+      } yield res
+
+  /** Atomically performs side-effect for each item in array */
+  def foreach[E](f: A => STM[E, Unit]): STM[E, Unit] =
+    this.foldM(())((_, a) => f(a))
+
+  /** Creates [[TArray]] of new [[TRef]]s, mapped with pure function. */
+  def map[B](f: A => B): STM[Nothing, TArray[B]] =
```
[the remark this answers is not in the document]
LGLO: I can't see it working, because `duplicate` and `transform` wouldn't/don't change type of `TArray[A]`.
`transform` (named after `Array#transform`) updates values inside `Ref`s.
Maybe I should rename `transform` to `updateAll`?

core/shared/src/main/scala/zio/stm/TArray.scala
LGLO: Yeah, I think it will be better. User is still obliged to take care of it but not necessarily just after calling this method.
:+1:

LGLO: Umm, why `die` and not `fail`?

LGLO:
@jdegoes please take a look at https://github.com/zio/zio/pull/1191/commits/97a7d653e0a3cec3f15012f5d96f435558238a8d and tell me which parts do you like and which you don't like.
After implementing alternative solution, I think that mapping index out of bounds to `None` in result channel is not that bad. Error channel gets ugly for `updateM` if I use `STM.fail`.
`STM.die` isn't nice IMO.

I see a lot of changes in tests. Should I migrate `TArraySpec` to new zio tests?

jdegoes:
@LGLO I would not port tests yet. We can do that later after ZIO Test is further along.

LGLO:
@jdegoes OK - I'll leave tests refactoring for some later time.
Please choose if I should go with `Option`, `fail` or `die` for `apply`, `update` and `updateM`. I'll make changes this weekend!

jdegoes:
@LGLO I would prefer to `die`, because in any case I can imagine, array index out of bounds is a bug, and a user cannot sensibly recover from a bug in their code.

LGLO:
@jdegoes I've switched to `die`

jdegoes:
@LGLO Looks great! Sorry for the delay in getting this merged.

ghostdogpr:
Woops, some methods became deprecated so it doesn't compile on master. Will send a PR shortly.

PR #1195, opened 2019-07-17
Add ZStream#throttleShape, ZStream#throttleShapeM

vasilmkd:
Part of #1027

### DIFF
--- a/streams/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -249,7 +249,7 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       for {
         clock <- Ref.make(TestClock.Zero).map(ref => new Clock { val clock = TestClock(ref) })
         test <- ZSink
-                 .throttleShapeM[Any, Nothing, Int](1, 1.second)(n => UIO.succeed(n.toLong))
+                 .throttleShape[Int](1, 1.second)(_.toLong)
                  .use(sinkTest)
                  .provide(clock)
       } yield test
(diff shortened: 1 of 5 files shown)


vasilmkd:
Can anything be done about the failing tests? I believe they are unrelated.

iravid:
Hey Vasil, this looks great. I’ll have time to review this tomorrow. I’d just like to note that you don’t have to use RefM; you can use Ref and run the delay after modify has finished.

streams/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -225,4 +228,55 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       } yield test
     }
   }
+
+  private def throttleShape = {
+
+    def sinkTest(sink: ZSink[Clock, Nothing, Nothing, Int, Int]) =
```
iravid: very nice use of the TestClock!

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1128,4 +1134,59 @@ object ZSink extends ZSinkPlatformSpecific {
...
+                        else {
+                          val c = missing / units
+                          if (c * units < missing) c + 1 else c
+                        }
+                      val newTokens    = tokens + cycles * units - weight
+                      val newTimestamp = timestamp + cycles * duration.toNanos
+                      val delay        = Duration.Finite(newTimestamp - current)
+                      (delay, (newTokens, newTimestamp))
+                  }
+          _ <- state._2.succeed(a).delay(delay)
```
iravid: `delay` forces a bounce through the platform's scheduler; can we skip it if `delay <= 0`? You can also do `clock.sleep(delay)` directly, no need to delay a value.

streams/jvm/src/test/scala/zio/stream/SinkSpec.scala
vasilmkd: Thanks. And thank you for the help with the sbt setup.

iravid:
Amazing work! Thank you @vasilmkd!

PR #1197, opened 2019-07-18
fixed methods Both.equals & Then.equals

darl:


jdegoes:
@darl Thank you for this contribution, and congratulations on your first ZIO pull request!

giphy

Issue #913, opened 2019-05-29
Introduce super polymorphic version of Ref called 'ZRef'


jdegoes:
We need a version of `Ref` that is highly polymorphic called `ZRef`. The existing `Ref` can be implemented atop this, as a type alias, and pseudo-companion object.

The new `ZRef` will have the following type parameters:

```scala
abstract class ZRef[-A, +EA, +B, +EB] {
  final def get: IO[EB, B] = ???
  final def set(a: A): IO[EA, Unit] = ???
  final def setAsync(a: A): IO[EA, Unit] = ???
  final def update[E, A1 <: A](f: A => A1)(implicit ev1: EA <:< E, ev2: EB <:< E): IO[E, A1] = ???
  final def modify[B, A1 <: A](f: A => (B, A1))(implicit ev1: EA <:< E, ev2: EB <:< E): IO[E,B] = ???
  // etc.
}
```

etc. I don't know if this type encoding will work for errors, it remains to be investigated.

The old `Ref` will be implemented as:

```scala
type Ref[A] = ZRef[A, Nothing, A, Nothing]
val Ref = ZRef
```

to avoid breaking any code and to provide a simpler version for common cases.

Not (necessarily) part of this ticket, this polymorphic version will allow introducing `map`, `const`, `unit`, `mapError`, `zip`, `filter`, and numerous other methods.

Most excitingly, there is a connection between this formulation of `ZRef` and optics. 😱 💃

iravid:
<3 "ZIO - Profunctors for Scala" <3

regiskuckaertz:
I'll pick this one 💪

regiskuckaertz:
I'm struggling to write `contramap` because of the `update*` and `modify*` methods. Variance aside, it seems impossible to provide an `A => A` when you have a `C => C` and a `C => A`:
```scala
def contramap[C](g: C => A) = new ZRef[EA, EB, C, B] {
  def update(f: C => C): IO[EB, B] = // cannot call self.update: (A => A) => IO[EB, B]
}
```
It suggest to me the update mechanism needs to be abstract somehow. `ZSink` comes close with
```scala
  final def update(state: Step[State, Nothing]): ZSink[R, E, A0, A, B] =
    new ZSink[R, E, A0, A, B] {
      type State = self.State
      val initial: ZIO[R, E, Step[State, Nothing]]             = IO.succeed(state)
      def step(state: State, a: A): ZIO[R, E, Step[State, A0]] = self.step(state, a)
[...]

PR #1198, opened 2019-07-19
Refactor Ref to ZRef - Fix (#913)

dsebban:
- [x] Refactor `Ref` to `ZRef` (all tests should pass, no code changes besides `ZRef.scala`)
- [x] Add `map`, `contramap`, `const`, `unit`, `mapErrorA` ?, `mapErrorB` ?, `collect`, `filter`
- [x] Make sure combinators name/doc is consistant with `ZQueue`
- [x] Add tests for new combinators
- [x] Docs
- [x]  Introduce real companion object for `Ref`
- [ ]  Make derived operations atomic using case class encoding


core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,120 @@
...
+     */
+    def updateSome(pf: PartialFunction[A, A]): IO[E, A] =
+      zref.updateUnify[A](identity, identity)(pf.orElse { case a => a })
+
+    /**
+     * Atomically modifies the `Ref` with the specified function, which computes
+     * a return value for the modification. This is a more powerful version of
+     * `update`.
+     */
+    def modify[B](f: A => (B, A)): IO[E, B] = zref.updateUnify[B](???, f.andThen(_._1))(identity)
```
dsebban: I am asked for a `B => A` instead of the `???`, not sure how to proceed here

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,120 @@
...
+   */
+  def set(a: A): IO[EA, Unit]
+
+  /**
+   * Writes a new value to the `Ref` without providing a guarantee of
+   * immediate consistency.
+   */
+  def setAsync(a: A): IO[EA, Unit]
+
+  def updateUnify[C](ea: C => A, eb: B => C)(f: C => C): IO[EA, C]
```
dsebban: is that what you meant @jdegoes ? you said it should return a `Ref[C]` but that would seem to be the `map` combinator and would demand to create a new `AtomicReference[C]` ?

jdegoes:
@dsebban 

I sketched this out to show what I mean (and get the types right):

```scala

object test {
  type Ref[A] = ZRef[Nothing, Nothing, A, A]

  trait ZRef[+EA, +EB, -A, +B] { self =>
    def set(a: A): IO[EA, Unit]
    def get: IO[EB, B]

    def unifyError[E](ea: EA => E, eb: EB => E): ZRef[E, E, A, B] = 
      new ZRef[E, E, A, B] {
        def set(a: A): IO[E, Unit] = self.set(a).mapError(ea)
        def get: IO[E, B] = self.get.mapError(eb)
      }

    def unifyValue[C](ca: C => A, bc: B => C): ZRef[EA, EB, C, C] = 
      new ZRef[EA, EB, C, C] {
        def set(c: C): IO[EA, Unit] = self.set(ca(c))
        def get: IO[EB, C] = self.get.map(bc)
      }

    private[zio] def unsafeUpdate[E, C](f: Any => Any): IO[E, C] = ??? // Implement and use this for performance
  }
  object ZRef {
    implicit class UnifiedSyntax[E, A](eref: ERef[E, A]) {
      def update(f: A => A): IO[E, A] = 
        eref.get.flatMap(a => eref.set(f(a)).unit) *> eref.get

      def modify[B](f: A => (B, A)): IO[E, B] = 
        eref.get.flatMap { (a0: A) => 
          val (b, a) = f(a0)
          
          eref.set(a) *> ZIO.succeed(b)
        }
    }
  }
}
```

We should be able to define nice combinators on this structure and not alter any existing code.

Please play around and let me know if you run into something that doesn't work.

dsebban:
That looks awesome thanks ! Will take it from here

regiskuckaertz:
```scala
def update(f: A => A): IO[E, A] = eref.get.flatMap(a => eref.set(f(a)).unit) *> eref.get
```
I think this should be uninterruptible to guarantee atomicity, am I right? But will be slower than a CAS loop. In my latest attempts I had `ZRef` contain an `AtomicRef[A]` and an abstract member `A => IO[EB, B]` but that created problems with `contramap` and some such.

dsebban:
@regiskuckaertz I agree there, anyway the update/modify method delegates to the unsafe ones that uses CAS , let me know if it makes sense , thanks

jdegoes:
@regiskuckaertz 

You are right, I did it more to demonstrate its soundness, it will be replaced by an `unsafeUpdate` that assumes unification; it can be accessed exclusively from the syntax class (private).

jdegoes:
@dsebban @regiskuckaertz 

A little more...

```scala
object test {
  type Ref[A] = ZRef[Nothing, Nothing, A, A]

  trait ZRef[+EA, +EB, -A, +B] { self =>
    def set(a: A): IO[EA, Unit]
    def get: IO[EB, B]

    final def collect[C](pf: PartialFunction[B, C]): ZRef[EA, Option[EB], A, C] = 
      new ZRef[EA, Option[EB], A, C] {
        def set(a: A): IO[EA, Unit] = self.set(a)
        def get: IO[Option[EB], C] = self.get.foldM(
          error => ZIO.fail(Some(error)),
          value => pf.lift(value).fold[IO[Option[EB], C]](ZIO.fail(None))(ZIO.succeed(_))
        )
      }

    final def filter(f: B => Boolean): ZRef[EA, Option[EB], A, B] = 
      collect { case b if f(b) => b }

    final def map[C](f: B => C): ZRef[EA, EB, A, C] = 
      new ZRef[EA, EB, A, C] {
        def set(a: A): IO[EA, Unit] = self.set(a)
        def get: IO[EB, C] = self.get.map(f)
      }

    final def mapSet[C](f: C => A): ZRef[EA, EB, C, B] = 
      new ZRef[EA, EB, C, B] {
        def set(c: C): IO[EA, Unit] = self.set(f(c))
        def get: IO[EB, B] = self.get
      }

    final def mapError[E](f: EB => E): ZRef[EA, E, A, B] = ???

    final def mapErrorSet[E](f: EA => E): ZRef[E, EB, A, B] = ???

    final def unifyError[E](ea: EA => E, eb: EB => E): ZRef[E, E, A, B] = 
      new ZRef[E, E, A, B] {
        def set(a: A): IO[E, Unit] = self.set(a).mapError(ea)
        def get: IO[E, B] = self.get.mapError(eb)
      }

    final def unifyValue[C](ca: C => A, bc: B => C): ZRef[EA, EB, C, C] = 
      new ZRef[EA, EB, C, C] {
        def set(c: C): IO[EA, Unit] = self.set(ca(c))
        def get: IO[EB, C] = self.get.map(bc)
      }

    private[zio] def unsafeUpdate[E, C](f: C => C): IO[E, C] = ??? // Implement and use this for performance
  }
  object ZRef {
    implicit class UnifiedSyntax[E, A](eref: ERef[E, A]) {
      def update(f: A => A): IO[E, A] = 
        eref.get.flatMap(a => eref.set(f(a)).unit) *> eref.get

      def modify[B](f: A => (B, A)): IO[E, B] = 
        eref.get.flatMap { (a0: A) => 
          val (b, a) = f(a0)
          
          eref.set(a) *> ZIO.succeed(b)
        }
    }
  }
}
```

We should ensure naming consistency between this and `ZQueue`, which is the closest thing we have to this right now.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,238 @@
...
+      def setAsync(a: A): zio.IO[EA, Unit] = self.setAsync(a)
+      private[zio] def unsafeUpdate[E](f: Any => Any): zio.IO[E, C] =
+        self.unsafeUpdate(f).foldM(error => ZIO.fail(error), value => ZIO.succeed(f(value).asInstanceOf[C]))
+      private[zio] def unsafeModify[E, D](f: Any => (Any, Any)): zio.IO[E, D] = self.unsafeModify(f)
+    }
+
+  final def filter(f: B => Boolean): ZRef[EA, Option[EB], A, B] =
+    collect { case b if f(b) => b }
+
+  final def zip[EA1 >: EA, EB1 >: EB, A1, B1](that: ZRef[EA1, EB1, A1, B1]): ZRef[EA1, EB1, (A, A1), (B, B1)] =
```
jdegoes: I think `zip` could only be done sanely for `TRef`, not `Ref`, because of the cross-ref transactional guarantees you'd need...

dsebban: 👍

core/shared/src/test/scala/zio/RefSpec.scala
```diff
@@ -96,4 +118,71 @@ class RefSpec extends Specification with DefaultRuntime {
...
+      } yield value must_=== Changed
+    )
+
+  def e11 = {
+    val failure = new Error("expected")
+    unsafeRun(
+      for {
+        ref    <- Ref.make[Option[State]](Some(Active))
+        refEA  = ref.mapError(_ => "oh no !")
+        failed <- refEA.set(Some(Changed)).map(_ => Option.empty[State]).someOrFail(failure).either
```
dsebban: This test is wrong, I am not sure how to test `mapError`, need some help here @jdegoes

jdegoes: To test `mapError`, you'll have to make a `Ref` whose getter will fail. You can do that manually (`new ZRef { def get = ZIO.fail("foo") ... }`) or through another "lifting" operation that we have yet to add. Then you can map that error into another one, and test the old and new error.

Let me know which approach you'd prefer!

dsebban: cool I went with the manual way for now, not sure how this `lift` function would look, if it would make this cleaner I am game

dsebban:
@jdegoes I think it's ready for review, thanks

core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -18,126 +18,64 @@ package zio
...
+       * See [[ZRef#setAsync]]
+       */
+      final def setAsync(a: A): UIO[Unit] = IO.effectTotal(value.lazySet(a))
 
-  /**
-   * Writes a new value to the `Ref`, with a guarantee of immediate
-   * consistency (at some cost to performance).
-   */
-  final def set(a: A): UIO[Unit] = IO.effectTotal(value.set(a))
+      private[zio] def unsafeUpdate[E](f: Any => Any): UIO[A] = IO.effectTotal {
```
regiskuckaertz: spurious `E`?

dsebban: should I return an `IO[E,A]` ?  Ref return `UIO` by definition, not sure what is the convention in this case

regiskuckaertz: I only saw the `unsafeUpdate` definition after. Apparently the scala compiler allows you to change the return type, I didn't know that 🤔 I'm inclined to say it doesn't really matter then.

jdegoes: I'd delete the `E` since it isn't used, and `UIO[A]` is a subtype of `IO[E, A]` for all `E`.

core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -18,126 +18,64 @@ package zio
...
 
-    while (loop) {
-      val current = value.get
+          loop = !value.compareAndSet(current, next)
+        }
+        next
+      }
 
-      next = f(current)
+      private[zio] def unsafeModify[E, B](f: Any => (Any, Any)): UIO[B] = IO.effectTotal {
```
jdegoes: Delete `E`.

core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -18,126 +18,64 @@ package zio
...
-   * Atomically modifies the `Ref` with the specified partial function, which computes
-   * a return value for the modification if the function is defined in the current value
-   * otherwise it returns a default value.
-   * This is a more powerful version of `updateSome`.
-   */
-  final def modifySome[B](default: B)(pf: PartialFunction[A, (B, A)]): UIO[B] = IO.effectTotal {
-    var loop = true
-    var b: B = null.asInstanceOf[B]
+  private final def unsafeCreate[A](a: A): Ref[A] =
+    new Ref[A] {
```
jdegoes: Can't we just use `ZRef` here? Why the need to create a new `Ref`?

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,197 @@
...
+    def get: IO[EB, B]                                                      = self.get
+    def setAsync(a: A): zio.IO[E, Unit]                                     = self.setAsync(a).mapError(f)
+    private[zio] def unsafeUpdate[E](f: Any => Any): zio.IO[E, B]           = self.unsafeUpdate(f)
+    private[zio] def unsafeModify[E, C](f: Any => (Any, Any)): zio.IO[E, C] = self.unsafeModify(f)
+  }
+
+  /**
+   * Replace the content of the Ref with `c`
+   */
+  final def const[C](c: C): ZRef[EA, EB, A, C] = map(_ => c)
```
jdegoes: Be sure to sort all methods in here alphabetically. The order is:

 - abstract types (alphabetically in this category)
 - abstract operators (alphabetically in this category)
 - abstract named defs / vals  (alphabetically in this category)
 - concrete types (alphabetically in this category)
 - concrete operators (alphabetically in this category)
 - concrete named methods (alphabetically in this category)

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Transforms elements that are set into this Ref with a function.
+   */
+  final def contramap[C](ca: C => A): ZRef[EA, EB, C, B] =
+    new ZRef[EA, EB, C, B] {
+      def set(c: C): IO[EA, Unit]                                             = self.set(ca(c))
+      def get: IO[EB, B]                                                      = self.get
+      def setAsync(c: C): zio.IO[EA, Unit]                                    = self.setAsync(ca(c))
+      private[zio] def unsafeModify[E, C](f: Any => (Any, Any)): zio.IO[E, C] = self.unsafeModify(f)
```
jdegoes: Unfortunately, I can see this is unsound. Let me think about this.

dsebban: Can you elaborate 🙏 ? What do you mean by unsound ?

regiskuckaertz: @dsebban if you have a `Ref[Int]` that you `contramap(f: String => Int)`, the underlying atomic is still holding a value of type `Int`. However if you call `update`, you will pass a function `String => String`, which will be cast to `Any => Any`. The type checker will not complain but you'll get a runtime exception. There's also the fact that you can't produce a `String` from an `Int`, so that function can never be called 🤔

That's a real conundrum.

dsebban: Wow , that’s a nice one ! Thanks for the explanation :)

dsebban: Ok cool , will work on tests to show the unsoundness issue after the weekend

dsebban:
@jdegoes I wrote the tests and indeed it’s pretty easy to get a runtime exception , any pointers on how to refactor the design so we can solve the issue ?

regiskuckaertz:
What about changing update et al to:
```scala
trait ZRef[E, I, O] {
  def update(f: O => I): IO[E, O]
}
```
?

jdegoes:
@regiskuckaertz Yes, you are exactly right on this. We need safe update / modify operations added to the base class with a default implementation.

@dsebban See Regis' comment above. Between this and my other comment, we should get something up and running quickly.

dsebban:
Great thanks for the help to both of you will get to it sometime next week

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,213 @@
...
+  //  */
+  // private[zio] def unsafeUpdate[E](f: Any => Any): IO[E, B]
+
+  def update(f: B => A): IO[E, B] =
+    get.flatMap(b => set(f(b)) *> get).uninterruptible
+
+  def modify[C](f: B => (C, A)): IO[E, C] =
+    get.flatMap { b =>
+      val (c, a) = f(b)
+      set(a).map(_ => c)
```
regiskuckaertz: nit: 
```suggestion
      set(a).const(c)
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,213 @@
...
+    }.uninterruptible
+
+}
+object ZRef extends Serializable {
+  implicit class InvariantSyntax[E, A](ref: ZRef[E, A, A]) {
+    def updateSome(pf: PartialFunction[A, A]): IO[E, A] =
```
regiskuckaertz: Can these move back into `ZRef` to? As
```
def updateSome(pf: PartialFunction[B, A]): IO[E, A]
def modifySome[C](default: C)(pf: PartialFunction[B, (C, A)])
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * {{{
+ * for {
+ *   ref <- Ref.make(2)
+ *   v   <- ref.update(_ + 3)
+ *   _   <- console.putStrLn("Value = " + v) // Value = 5
+ * } yield ()
+ * }}}
+ */
+trait ZRef[+E, A, B] extends Serializable { self =>
```
regiskuckaertz:
```suggestion
trait ZRef[+E, -A, +B] extends Serializable { self =>
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  // /**
+  //  * Replace the content of the Ref with `c`
+  //  */
+  // final def const[C](c: C): ZRef[EA, EB, A, C] = map(_ => c)
+
+  // /**
+  //  * Transforms elements that are set into this Ref with a function.
+  //  */
+  // final def contramap[C](ca: C => A): ZRef[EA, EB, C, B] =
```
regiskuckaertz: Now you should be able to write `contramap` 🎉

dsebban:
Nice ! Thanks @regiskuckaertz

core/shared/src/main/scala/zio/ZRef.scala
dsebban: Compiler complains for 
for `def updateSome(pf: PartialFunction[B, A]): IO[E, A]`
`contravariant type A occurs in covariant position in type (pf: PartialFunction[B,A])zio.IO[E,A] of method updateSome`

For `modifySome` I am stuck here, I need an `A` : 
`def modifySome[C](default: C)(pf: PartialFunction[B, (C, A)]) = modify(pf.orElse[B, (C, A)] { case b => (default, ???) })`

core/shared/src/main/scala/zio/ZRef.scala
dsebban: yay even the test that failed because it was unsound passes now

core/shared/src/main/scala/zio/ZRef.scala
regiskuckaertz: My bad, it should be:
```scala
def updateSome(pf: PartialFunction[B, A]): IO[E, B]
def modifySome[C](default: C)(pf: PartialFunction[B, (C, A)]): IO[E, B]
```
just like the `update` and `modify` equivalents.

regiskuckaertz: There might be a formulation that does not use `orElse` so you don't fall into that problem.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+  /**
+   * Atomically modifies the `Ref` with the specified function. This is not
+   * implemented in terms of `modify` purely for performance reasons.
+   *
+   * @param f function to atomically modify the `Ref`
+   * @return `IO[E, B]` modified value of the `Ref`
+   */
+  def update[E](f: B => A)(implicit ea: EA <:< E, eb: EB <:< E): IO[E, B] = {
```
dsebban: @jdegoes @regiskuckaertz , is it what you had in mind ? tests seem to pass, not sure about this encoding

jdegoes: We have to override `update` and `modify` inside the `ZRef.make` constructor for the shortcut versions. Here, while we know the _types_ are equal, we don't know that they are "original" — they may have been mapped many times. So it's not safe to use the unsafe methods.

jdegoes:
This is looking really good and nearly there.

Unfortunately, we can't preserve atomic guarantees with this encoding. So what I think we should do is "finish" it and then move it into `TRef`, which the same API as `Ref`, so it should be an easy port. Then we can get atomic guarantees for free. But first, let's get this all the way there (just a few small tweaks and it'll be golden!).

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -160,4 +162,55 @@ object ZRef extends Serializable {
...
 
   }
+
+  final def make[E, A](a: A): UIO[ZRef[E, E, A, A]] = IO.effectTotal {
+    unsafeCreate(a)
+  }
+
+  private final def unsafeCreate[E, A](a: A): ZRef[E, E, A, A] =
+    new ZRef[E, E, A, A] {
+      val value: AtomicReference[A] = new AtomicReference[A](a)
```
regiskuckaertz: Let's make it private.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * A `ZRef[EA, EB, A, B]` is a mutable atomic reference for the `IO` monad. This is the `IO` equivalent of
```
regiskuckaertz: I don't think we still use "the `IO` monad" idiom anymore, how about:
```suggestion
 * A `ZRef[EA, EB, A, B]` is a purely functional mutable reference augmented with atomic operations.
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+    }.uninterruptible
+  }
+
+}
+object ZRef extends Serializable {
+
+  implicit class InvariantSyntax[E, A](ref: ZRef[E, E, A, A]) {
+
+    def updateSome(pf: PartialFunction[A, A]): IO[E, A] =
+      ref.update(pf.orElse { case a => a })
```
regiskuckaertz: I think you can write a default version like this in the class itself and move that version in `unsafeCreate`:
```suggestion
      get.flatMap(b => pf.lift(b).fold(ZIO.succeed(b))(a => set(a) *> get).uninterruptible
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+
+    def modifySome[C](default: C)(pf: PartialFunction[A, (C, A)]): IO[E, C] =
+      ref.modify(pf.orElse { case a => (default, a) })
```
regiskuckaertz: Same trick here using `lift`.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+    new ZRef[E, E, A, A] {
+      val value: AtomicReference[A] = new AtomicReference[A](a)
+
+      final def get: IO[E, A] = IO.effectTotal(value.get)
+
+      final def set(a: A): IO[E, Unit] = IO.effectTotal(value.set(a))
+
+      final def setAsync(a: A): IO[E, Unit] = IO.effectTotal(value.lazySet(a))
+
+      override def update[E1](f: A => A)(implicit ea: E <:< E1, eb: E <:< E1): zio.IO[E1, A] =
```
regiskuckaertz:
```suggestion
      override final def update[E1](f: A => A)(implicit ea: E <:< E1, eb: E <:< E1): zio.IO[E1, A] =
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+        unsafeUpdate(f.asInstanceOf[Any => Any]).mapError(e => ea.apply(e))
+
+      override final def modify[E1, C](f: A => (C, A))(implicit ea: E <:< E1, eb: E <:< E1): zio.IO[E1, C] =
+        unsafeModify(f.asInstanceOf[Any => (Any, Any)]).mapError(e => ea.apply(e))
+
+      private[zio] def unsafeModify[E, B](f: Any => (Any, Any)): IO[E, B] = IO.effectTotal {
```
regiskuckaertz:
```suggestion
      private[zio] final def unsafeModify[E, B](f: Any => (Any, Any)): IO[E, B] = IO.effectTotal {
```

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,216 @@
...
+
+          b = tuple._1.asInstanceOf[B]
+
+          loop = !value.compareAndSet(current, tuple._2.asInstanceOf[A])
+        }
+
+        b
+      }
+
+      private[zio] def unsafeUpdate(f: Any => Any): IO[E, A] = IO.effectTotal {
```
regiskuckaertz:
```suggestion
      private[zio] final def unsafeUpdate(f: Any => Any): IO[E, A] = IO.effectTotal {
```

core-tests/shared/src/test/scala/zio/RefSpec.scala
```diff
@@ -64,10 +64,125 @@ object RefSpec
...
+            ref                    <- Ref.make(1)
+            refStringInt           = ref.contramap((_: String).toInt)
+            refString: Ref[String] = refStringInt.map(_.toString)
+            value                  <- refString.update(_.trim())
+          } yield assert(value, equalTo("1"))
         }
       )
     )
 
+// def e9 =
```
regiskuckaertz: Looks like you can delete `e9` and `e10` and move the others in the suite above?

core/shared/src/main/scala/zio/ZRef.scala
dsebban: /zio/core/shared/src/main/scala/zio/ZRef.scala:210:17: a type was inferred to be `Any`; this may indicate a programming error.
[error]     unified.get.flatMap { f }.uninterruptible
[error]                 ^

dsebban: Not sure why the compiler is upset at me ...

regiskuckaertz:
@dsebban You may have to add some type annotations in order to get rid of that error message. Alternatively, you can try using the silencer plugin, as in [here](https://github.com/zio/zio/pull/1784/files#diff-ddc0e611b880d3cf1a5f645b19b5d5c0R219).

core-tests/shared/src/test/scala/zio/RefSpec.scala
```diff
@@ -45,29 +45,103 @@ object RefSpec
...
-            r     <- ref.modify[String](_ => ("hello", update))
+            r     <- ref.modify[Nothing, String](_ => ("hello", update))
             value <- ref.get
           } yield assert(r, equalTo("hello")) && assert(value, equalTo(update))
         },
         testM("modifySome") {
           for {
             ref   <- Ref.make[State](Active)
-            value <- ref.modifySome[String]("State doesn't change") { case Closed => ("active", Active) }
+            value <- ref.modifySome[Nothing, String]("State doesn't change") { case Closed => ("active", Active) }
```
dsebban: 😱 
is there a way to get this inferred @regiskuckaertz  ?

regiskuckaertz:
The failing test was flaky, if you merge upstream this PR should be good to go 🎸 🚀

core-tests/shared/src/test/scala/zio/RefSpec.scala
regiskuckaertz: Not sure 🤔

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,278 @@
...
+        unsafeModify(f.asInstanceOf[Any => (Any, Any)]).mapError(e => ea.apply(e))
+
+      override final def modifySome[E1, C](
+        default: C
+      )(pf: PartialFunction[A, (C, A)])(implicit ea: E <:< E1, eb: E <:< E1): zio.IO[E1, C] = {
+        val f: A => (C, A) = a => pf.lift(a).getOrElse((default, a))
+        self.modify(f)(ea, eb)
+      }
+
+      private[zio] final def unsafeModify[E, B](f: Any => (Any, Any)): IO[E, B] = IO.effectTotal {
```
regiskuckaertz: Does this need to be package-private? Or can we do `private[this]`?

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+  /**
+   * Atomically modifies the `ZRef` with the specified function. This is not
+   *
+   * @param f function to atomically modify the `ZRef`
+   * @tparam E type of the unified error value of the `ZRef`
+   * @return `IO[E,B]` modified value of the `ZRef`
+   */
+  def update[E](f: B => A)(implicit ea: EA <:< E, eb: EB <:< E): IO[E, B] = {
```
jdegoes: How good is type inference here?

dsebban: you can check out the tests, it works fine, it infers correctly.
The issue now is `modify` and `modifySome` because of the new `E` type param, it needs to be specified and leads to this atrocity 
https://github.com/zio/zio/pull/1198/commits/e7be76639dcceeec4287a31fb00d6ff7e2e65249#diff-e2c16046e05624c39e46b9e4c2870b9cR2138
anything that can be done to make it infer the `E` ?

jdegoes:
@dsebban I will try to take a look at this tomorrow or over the weekend.

This is really excellent work. To merge this in, we just need to figure out two things:

1. Type inference issues. I believe I can help fix those.
2. Atomicity issues. Currently ZRef will not be atomic for any derived reference. Thinking carefully about it, I _think_ we can make it work with `ZRef`; of course, it's trivial with `TRef`. I'll write up a few thoughts shortly.

core-tests/shared/src/test/scala/zio/RefSpec.scala
```diff
@@ -45,29 +45,103 @@ object RefSpec
         testM("modify") {
           for {
             ref   <- Ref.make(current)
-            r     <- ref.modify[String](_ => ("hello", update))
+            r     <- ref.modify[Nothing, String](_ => ("hello", update))
```
jdegoes: What happens if you omit _both_ type parameters? Does it work?

dsebban: Yes !

dsebban:
@jdegoes I think the inference issues are behind us, we can even remove type parameters from the old Ref methods and it infers fine.
Would appreciate to get your thoughts here on the atomicity issues, I would love to get this PR merged.

jdegoes:
@dsebban 

This looks good to me! We just have one final concern and then can merge:

 By design, both `update` and `modify` functions must be atomic, consistent and isolated. That is, no other fiber may "clobber" values during an update / modify.

Currently, this is not true, because `.uninterruptible` does not make an action consistent and isolated, it merely guarantees the action won't be interrupted.

To fix this problem, I previously thought we needed to switch the polymorphic machinery over to `TRef`, because it allows us to make the composition of get + set atomic. 

However, I think I can see a way to do it here in `ZRef` by modeling our derived operations differently.

In essence, we can think of a `ZRef` as being one of the following cases:

```scala
sealed trait ZRef[+EA, +EB, -A, +B]
object ZRef {
  final case class AtomicRef(ref: AtomicReference[A]) extends ZRef[Nothing, Nothing, A, A]
  final case class Derived[EA0, EB0, A0, B0, EA, EB, A, B](
    original: ZRef[EA0, EB0, A0, B0],
    getter: Either[EB0, B0] => Either[EB, B],
    setter: Either[EA0, A] => Either[EA, A0])
}
```

Now the `update` and `modify` operations on `AtomicRef` will be implemented in straightforward fashion; and the corresponding operations on `Derived` are only a little more complex, they have to dig into (looping) and apply the getters / setters, going through intermediate values.

Does this make sense? Can you see if I'm missing anything?

regiskuckaertz:
That encoding will simplify the code a lot! I observe that so far we never use `EA` or `EB`, except in  `collect` where the error out is lifted to `Option[EB]`. And I fail to see how we could enforce atomicity with a combinator like `mapM`? or is that  a  non  sequituur? If not, then we may want to drop those type parameters altogether and fix them to `Nothing` and `Unit` (for `collect`)?

dsebban:
@jdegoes @regiskuckaertz Thanks for the help ! I will implement a couple of combinators with the new encoding and see how it  goes

jdegoes:
> That encoding will simplify the code a lot! I observe that so far we never use EA or EB, except in collect where the error out is lifted to Option[EB]. And I fail to see how we could enforce atomicity with a combinator like mapM? or is that a non sequituur? If not, then we may want to drop those type parameters altogether and fix them to Nothing and Unit (for collect)?

I have a whole host of combinators that use the error channels... but we don't need to rush them in because they will be backward compatible. Also I have a plan for ZRef for "concurrent optics" that requires the error channels... HMU on Discord!

jdegoes:
@dsebban Let me know when you are done, I will be 100% ready to merge right away when we achieve the atomic/consistent/isolated guarantees of the previous version!

jdegoes:
@dsebban Btw will you be at ZIO Hackathon Berlin?

regiskuckaertz:
(Googling HMU 😅) can’t wait to see that 🤗🕺🏻

jdegoes:
With the new encoding you can do totally safe update/modify, because the `AtomicRef` subtype knows the types are equal, and the `Derived` one will just form a "big function" to pass upward until it gets to the `AtomicRef`, which will use the "big function" to do the update.

You'll need some private methods, I have some idea of their shape.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -114,11 +115,7 @@ trait ZRef[+EA, +EB, -A, +B] extends Serializable { self =>
    * @return `ZRef[EA, EB, A, C]` a new `ZRef` with the modified value
    */
   final def map[C](f: B => C): ZRef[EA, EB, A, C] =
-    new ZRef[EA, EB, A, C] {
-      def set(a: A): IO[EA, Unit]          = self.set(a)
-      def get: IO[EB, C]                   = self.get.map(f)
-      def setAsync(a: A): zio.IO[EA, Unit] = self.setAsync(a)
-    }
+    Derived[EA, EB, A, B, EA, EB, A, C](self, e => e.map(f), e => e)
```
dsebban: I am not sure if this is what you had in mind @jdegoes @regiskuckaertz, am I on the right track ?

regiskuckaertz: Yes I think so, and in addition to using the "big functions" in `get` and `set`, you will also have to use them in `update` and `modify`, as the definitions now only address interruptibility as per @jdegoes 's explanation. I assume to guarantee atomicity, you will have to create a function `bigF` that you pass to `original` as `original.update(bigF)` which, by hypothesis, is atomic.

However this will require some juggling, because `bigF` should have type `B0 => A0` and it doesn't seem possible to write it:

```scala
class Derived[EA0, EB0, A0, B0, EA, EB, A, B](
     original: ZRef[EA0, EB0, A0, B0],
     getter: Either[EB0, B0] => Either[EB, B],
     setter: Either[EA0, A] => Either[EA, A0]
   ) extends ZRef[EA, EB, A, B] {

  def update[E](f: B => A)(implicit ea: EA <:< E, eb: EB <:< E): IO[E, A] =
    original.update { b0 =>
      setter(getter(Right(b0)).map(f)) : Either[E, A0]
      // ^ we can't get an `A0` out since there may be an error somewhere in the pipeline
    }
}
```

So the best we can come up with is a function `B0 => Either[E, A0]`. Perhaps we can exploit the fact that `A ≡ Either[Nothing, A]` in the `Atomic` case and define:

```scala
trait ZRef[EA, EB, A, B] {
  def update(f: B => A)(implicit ea: EA <:< E, eb: EB <:< E): IO[E, A] =
    maybeUpdate(f.andThen(Right(_)))
  def maybeUpdate[E](f: B => Either[E, A])(implicit ea: EA <:< E, eb: EB <:< E): IO[E, A]
}

case AtomicRef[A](value: AtomicReference[A]) extends ZRef[Nothing, Nothing, A, A] {
  def maybeUpdate(f: A => Either[Nothing, A]) =
    UIO.effectTotal(value.updateAndGet(a => f(a).getOrElse(???)))
```

We can use `???` above since we know that the result of `f` must be a `Right`.

jdegoes:
In all likelihood, you will require a private `update` and `modify` method defined in `ZRef` trait and implemented in low-level fashion for `AtomicRef` that allow short-circuiting in the event that `Left` (error) is returned from the update / modify operation. In such a case, the state of the ref would not be modified, but the error would be returned.

Alternately, you could do that using `modify` always, i.e. `modify` is already powerful enough to give you the short-circuiting ability for the `Derived` case where you compose the "big (fallible) getters / setters".

dsebban:
Thanks for all the pointers, will work on the implementation at the end of the week

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -307,9 +307,32 @@ object ZRef extends Serializable {
     def set(a: A): zio.IO[EA, Unit] = ZIO.fromEither(setter(Right(a))).flatMap(a0 => original.set(a0).mapError(fea))
 
     def setAsync(a: A): zio.IO[EA, Unit] = ZIO.fromEither(setter(Right(a))).map(a0 => original.setAsync(a0)).unit
-    // def maybeUpdate[E](f: B => Either[E,A])(implicit ea: EA <:< E, eb: EB <:< E): IO[E,B] = original.map { o =>
+    def maybeUpdate[E](f: B => Either[E, A])(implicit evea: EA <:< E, eveb: EB <:< E): IO[E, B] =
```
dsebban: @regiskuckaertz , I am having a hard time  here, would appreciate any help to get me unstuck

regiskuckaertz: @dsebban sorry quite busy right now, will have a look asap 😇

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -0,0 +1,343 @@
...
+    setter: Either[EA0, A] => Either[EA, A0]
+  )(implicit fea: EA0 => EA)
+      extends ZRef[EA, EB, A, B] { self =>
+    def get: zio.IO[EB, B] =
+      original.get.either.flatMap(e => ZIO.fromEither(getter(e)))
+    def set(a: A): zio.IO[EA, Unit] = ZIO.fromEither(setter(Right(a))).flatMap(a0 => original.set(a0).mapError(fea))
+
+    def setAsync(a: A): zio.IO[EA, Unit] = ZIO.fromEither(setter(Right(a))).map(a0 => original.setAsync(a0)).unit
+
+    def modify[E, C](f: B => (C, A))(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] = get.mapError(eb.apply).flatMap{ b =>
```
dsebban: That seems too easy, it's probably wrong, I need guidance here @jdegoes 🙏

jdegoes: This unfortunately will not be atomic. Any time you separate `get` from `set`, i.e. do state changes outside `update` or `modify`, it will not have atomic guarantees. Gotta stitch together the `Derived` functions all the way back to the root, to create "big getter" and big setter", which can be fused into a function that can be passed to the `AtomicRef`'s update / modify.

regiskuckaertz: @dsebban You can apply the same reasoning here, i.e. the only thing you can do is use `original`:
```scala
def modify[E, C](f: B => (C, A))(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] =
  original.modify(???)
```
We need to provide a function `B0 => (C, A0)` however as you remember failure may occur at some point, so we need:
```scala
private def modifyOrFail[E, C](f: B => Either[E, (C, A)])(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] =
  original.modifyOrFail(???)

def modify[E, C](f: B => (C, A))(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] =
  modifyOrFail(b => Right(f(b)))
```
Then it's just a matter of following the compiler:
```scala
private def modifyOrFail[E, C](f: B => Either[E, (C, A)])(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] =
  original.modifyOrFail(b0 =>
    for {
      b <- getter(Right(b0))
      (c, a) <- f(b)
      a0 <- setter(Right(a))
    } yield (c, a0)
  )
```
Actually some future `recover*` combinator may change `setter` so I think this will need to be slightly modified to account for that.

dsebban: Thanks for the help will try that this weekend or next week

dsebban: @regiskuckaertz , @jdegoes I am having an issue with providing the right evidences: 
```
private def modifyOrFail[E, C](f: B => Either[E, (C, A)])(implicit ea: EA <:< E, eb: EB <:< E): zio.IO[E, C] =
  original.modifyOrFail(b0 =>
```
It's asking for the evidence of the original
```
def modifyOrFail[E, C](f: B0 => Either[E,(C, A0)])(implicit ea: EA0 <:< E, eb: EB0 <:< E): IO[E,C]
```
I have an `fea: EA0 => EA` and an `ea: EA <:< E` so it seems to me I should be able to get an 
`EA0 <:< E` , I havent figured out how though ...
And for `EB0 <:< E` I have no clue how to get a  hold of one.

regiskuckaertz: hi @dsebban and apologies for the late reply. I will find time this weekend to look into this in more detail and help you get this one over the line for 1.0 🎉

jdegoes:
@dsebban Need any help? I'm emerging again from another crunch. 😆

dsebban:
@jdegoes Sure! I am stuck in type tetris hell, you can checkout my previous comment and help me fill the missing ??? . There is a bunch of evidences missing, I might be approaching this wrong and there is a way to get rid of them somehow

regiskuckaertz:
I've noticed a few things:

- the  base case should never fail, i.e. be a `ZRef[Nothing, Nothing, A, A]`
- it is a pattern in ZIO to define a more generic version of functions with implicit evidence, i.e.
```scala
trait ZRef[EA, EB, A, B] {
  final def updateOrFail[E](f: B => Either[E, A])(implicit eva: EA <:< E, evb: EB <:< E): IO[E, B] =
    updateOrFailWith(f)(eva, evb)

  def updateOrFailWith[E](f: B => Either[E, A])(fea: EA => E, feb: EB => E): IO[E, B]
}
```
- I think `Derived` needs two additional members:
```scala
final case class Derived[EA0, EB0, A0, B0, EA, EB, A, B](
  original: ZRef[EA0, EB0, A0, B0],
  getter: Either[EB0, B0] => Either[EB, B],
  setter: Either[EA, A] => Either[EA0, A0],
  ea: EA0 => EA,
  eb: EB0 => EB
)
```
- the shape of `setter` and `getter` allow for error recovery, so a for comprehension won't do, e.g.

```scala
final case class Derived[EA0, EB0, A0, B0, EA, EB, A, B](...) {
  final def updateOrFailWith[E](f: B => Either[E, A])(fea: EA => E, feb: EB => E): IO[E, B] =
    original.updateOrFailWith { b0 =>
      val b1: Either[EB, B] = getter(Right(b0))
      val b2: Either[E, B] = b1.swap.map(feb).swap
      val a: Either[E, A] = b2.flatMap(f)
      setter(a) // <- not sure this is correct
    }(ea.compose(fea), eb.compose(feb))
}
```

What do you think @jdegoes ?

adamgfraser:
@dsebban Is there anything I can do to help with this? Would be great to get this in for the 1.0 release.

dsebban:
Yes , I was stuck there, I will try to come back to this today, write a list of issues I need help  with . 
thanks @adamgfraser

regiskuckaertz:
So I was looking at this the other day, and the nail in the coffin is that many combinators require total functions, e.g. `update(f: B => A)`. This combinator is essential to the meaning of `ZRef`, but building it for the derived case, where failure can occur, does not seem possible. It cannot be changed to `update(f: B => Either[Either[EA, EB], A])`, because then it becomes impossible to call the underlying method on the atomic reference and we lose the atomicity guarantee. Perhaps this needs to be encoded over STM, but then I think performance will plummet. Or perhaps I'm missing something.

adamgfraser:
How about something like this?

```scala
package zio

import java.util.concurrent.atomic.AtomicReference

sealed trait ZRef[+EA, +EB, -A, +B] { self =>

  def get: IO[EB, B]

  def set(a: A): IO[EA, Unit]

  def modify[E, C](f: B => (C, A))(implicit ev1: EA <:< E, ev2: EB <:< E): IO[E, C]

  def mapEither[EC >: EB, C](f: B => Either[EC, C]): ZRef[EA, EC, A, C]
}

object ZRef {

  final case class Atomic[A](value: AtomicReference[A]) extends ZRef[Nothing, Nothing, A, A] { self =>
    def get: UIO[A] =
      UIO.effectTotal(value.get)
    def set(a: A): UIO[Unit] =
      UIO.effectTotal(value.set(a))
    def modify[E, B](f: A => (B, A))(implicit ev1: Nothing <:< E, ev2: Nothing <:< E): IO[E, B] =
      UIO.effectTotal(f(value.getAndUpdate(a => f(a)._2))._1)
    def mapEither[E, B](f: A => Either[E, B]): ZRef[Nothing, E, A, B] =
      new Derived[Nothing, E, A, B] {
        protected type S = A
        protected val original                         = self
        protected def getter(s: S): Either[E, B]       = f(s)
        protected def setter(a: A): Either[Nothing, S] = Right(a)
      }
  }

  sealed trait Derived[+EA, +EB, -A, +B] extends ZRef[EA, EB, A, B] { self =>
    protected type S
    protected val original: Atomic[S]
    protected def getter(s: S): Either[EB, B]
    protected def setter(a: A): Either[EA, S]
    final def get: IO[EB, B] =
      original.get.map(getter).absolve
    final def set(a: A): IO[EA, Unit] =
      IO.fromEither(setter(a)).flatMap(original.set)
    final def modify[E, C](f: B => (C, A))(implicit ev1: EA <:< E, ev2: EB <:< E): IO[E, C] =
      original
        .modify[E, Either[E, C]] { s =>
          getter(s) match {
            case Left(eb) => (Left(ev2(eb)), s)
            case Right(b) =>
              val (c, a) = f(b)
              setter(a) match {
                case Left(ea)  => (Left(ev1(ea)), s)
                case Right(s1) => (Right(c), s1)
              }
          }
        }(ev1, ev2)
        .absolve
    final def mapEither[EC >: EB, C](f: B => Either[EC, C]): ZRef[EA, EC, A, C] =
      new Derived[EA, EC, A, C] {
        protected type S = self.S
        protected val original                    = self.original
        protected def getter(s: S): Either[EC, C] = self.getter(s).flatMap(f)
        protected def setter(a: A): Either[EA, S] = self.setter(a)
      }
  }
}
```

adamgfraser:
Basically a ZRef is an atomic reference plus an optic.

regiskuckaertz:
﻿That looks promising ! Elegant solution on modify 😍 Would that work for update? Perhaps we can just write it in terms of modify...

dsebban:
Looks nice indeed, will give it a shot and implement it this coming week, thanks !!!

adamgfraser:
@regiskuckaertz Yes we can just define it in terms of `modify` on `ZRef`:

```scala
final def update[E](f: B => A)(implicit ev1: EA <:< E, ev2: EB <:< E): IO[E, Unit] =
  modify(b => ((), f(b)))
```

adamgfraser:
@dsebban Great! Let me know if there is anything else I can do to help as you are working through it.

adamgfraser:
@dsebban Just want to check in on status here. We're getting close to 1.0 so need to get this over the finish line.

dsebban:
Will try to squeeze some work tonight, how much time is left ?

adamgfraser:
@dsebban We are probably a couple of weeks away from a release but we are already really pushing the boundaries in terms of adding new functionality very close to the release and not giving ourselves much time to get feedback and resolve any issues. So I would say ASAP. Can you resolve conflicts and get it to a point where we can merge, even if it doesn't have all the functionality we want, so we get your changes in and then we can do PRs on top of that?

dsebban:
@adamgfraser Done, I have a lot of `???` I need to fill,  but rebased tests should pass now.

adamgfraser:
@dsebban Awesome. Can you fix the failing tests?

dsebban:
Hmm weird,  tests passed on my local machine, will take care of it tommorow

dsebban:
@adamgfraser  This branch should be good to go it only introduces `ZRef` with feature parity to `Ref`. I kept the same docs, not sure if I should remove all references to Ref in the scaladocs.
I opened another branch 
https://github.com/dsebban/zio/tree/zref-combinators
There I  have a [WIP] commit with the `mapEither` combinator, I have some work left there, it compiles with a lot of `???

Let me know if this what you had in mind

adamgfraser:
@dsebban Great! I will take a look this morning. Maybe think about which combinators you have already done or want to do and which ones I can help with? We're doing another release on Monday so we could potentially get `ZRef` in for that if we make a big push today.

dsebban:
I won't have time to work on this until probably Thursday, you can take over my branch and try to fill in the `???` to get the `mapEither` in,  that would be cool, thanks!
Regardless hopefully this PR should make it in 🙏

jdegoes:
@dsebban @regiskuckaertz @adamgfraser 

Thanks to all for the enormous work that went into this PR!

Opens the door to nice liftings of optics into ZRef!

adamgfraser:
@dsebban Thanks for your work on this! Excited we were able to get this in!

PR #1205, opened 2019-07-19
Add burst control to the existing Stream throttling combinators

### LINKED ISSUE #1027: Add throttling combinators to ZStream

vasilmkd:
Also fix potential token overflow.

Should be enough to close #1027.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1232,19 +1232,22 @@ object ZSink extends ZSinkPlatformSpecific {
 
   /**
    * Creates a sink which delays input elements of type A according to the given bandwidth parameters
-   * using the token bucket algorithm. The weight of each element is determined by the `costFn` function.
+   * using the token bucket algorithm. The sink allows for burst in the processing of elements by allowing
+   * the token bucket to accumulate tokens up to a `max` threshold. The weight of each element is determined
+   * by the `costFn` function.
    */
-  final def throttleShape[A](units: Long, duration: Duration)(
+  final def throttleShape[A](units: Long, duration: Duration, max: Long = Long.MaxValue)(
     costFn: A => Long
   ): ZManaged[Clock, Nothing, ZSink[Clock, Nothing, Nothing, A, A]] =
-    throttleShapeM[Any, Nothing, A](units, duration)(a => UIO.succeed(costFn(a)))
+    throttleShapeM[Any, Nothing, A](units, duration, max)(a => UIO.succeed(costFn(a)))
 
   /**
    * Creates a sink which delays input elements of type A according to the given bandwidth parameters
-   * using the token bucket algorithm. The weight of each element is determined by the `costFn` effectful
-   * function.
+   * using the token bucket algorithm. The sink allows for burst in the processing of elements by allowing
+   * the token bucket to accumulate tokens up to a `max` threshold. The weight of each element is determined
+   * by the `costFn` effectful function.
    */
-  final def throttleShapeM[R, E, A](units: Long, duration: Duration)(
+  final def throttleShapeM[R, E, A](units: Long, duration: Duration, max: Long = Long.MaxValue)(
     costFn: A => ZIO[R, E, Long]
   ): ZManaged[Clock, Nothing, ZSink[R with Clock, E, Nothing, A, A]] = {
     import ZSink.internal._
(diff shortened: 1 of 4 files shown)


iravid:
@vasilmkd would it be possible to add burst as a default parameter on all the variants instead of a separate variant?

vasilmkd:
The current implementation of the token bucket algorithm should now closely match the one described here: [Token Bucket Wikipedia](https://en.wikipedia.org/wiki/Token_bucket#Algorithm).

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1271,10 +1278,15 @@ object ZSink extends ZSinkPlatformSpecific {
       def extract(state: State) = state._2.await
     }
 
-    for {
-      _       <- assertPositive(units).toManaged_
-      current <- clock.currentTime(TimeUnit.NANOSECONDS).toManaged_
-      bucket  <- Ref.make((units, current)).toManaged_
+    def checkTokens(sum: Long, max: Long): Long = if (sum < 0) max else math.min(sum, max)
```
iravid: Question: when `sum < 0`, should we subtract it from `max`? Otherwise when the bucket is depleted, there is always constant amount of `max` available in addition.

iravid:
Also, I believe the default value for `max` should reflect no burst capacity, so `Long.MaxValue` is probably not a good choice for that.

vasilmkd:
There are 2 ways to do this.

1. Use `max` as the maximum number of tokens allowed at any time in the bucket (currently implemented). At the moment, the default value of Long.MaxValue enables an _average bandwidth_ of `units / duration`. In order to restrict burst to 0, and offer _at best_ `units / duration` bandwidth, one needs to set `max` to be equal to `units`. I personally think that having the maximum number of tokens be configurable is more user friendly than the second option. However, the _no burst_ default needs to be implemented as an overloaded method, since `max` cannot be initialized to `units` in the type signature.

2. Use `threshold` as a number of tokens which can be added to `units` when burst occurs. In the _no burst_ default, it would be an argument with default value 0. This approach however, increases the danger of overflow, as two values of type `Long` need to be added together.

Interested to hear your thoughts on this. I'm happy to implement any of them.

vasilmkd:
Idea.

I can implement the second proposal from the last comment by wrapping the existing functions in
`throttleEnforceM(units: Long, duration: Duration, burst: Long = 0): ZManaged[ZSink]` which simply delegates to the _inner method_ `throttleEnforceM(units, duration, units + burst or Long.MaxValue in case of overflow)` (the current implementaion).

That way we get the semantics of the second proposal with the "overflow safety" of the first one, and as a bonus, it requires minimal code changes.

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: You're right, I'll fix this.

vasilmkd: Actually, on second thought, the `checkTokens` function is called to guard against overflow. `sum` is computed exclusively from positive values, whose total value might exceed `Long.MaxValue` (which is what the if statement checks). If that happens, we want `max` tokens to be available, otherwise, if we are dealing with positive values (`sum` and `max`), just get the minimum value.

vasilmkd:
I'm still confused about the burst control on the shaping throttle. It doesn't ever drop elements, it just delays them. At the moment, it feels like the API is lying to the user, since an element can arrive with a cost which is greater than the "maximum capacity" of the bucket, but it still goes through after some time. As far as I can see, this behavior is also present in akka-streams.

iravid:
Hi @vasilmkd, could you explain why you say that `sum` is computed exclusively of positive values? 

I infer from what you say that the `tokens` in the ref could never be negative. However:
- `remaining` is computed as `available - weight`; 
- if `available < weight`, then remaining is negative
- a negative number is stored in the ref
- at the next element, assuming not enough cycles have passed, `tokens + (cycles * units)` is still negative, so `maxTokens` is used
- assuming that element is heavy enough, we still go into negative values for `remaining`;
- the next round repeats the same, effectively not throttling.

WDYT?

vasilmkd:
Yes, `tokens` can become negative. However, the sink is sleeping for enough time required such that `tokens` amount of tokens can be accumulated, thus at the time when the next element is processed, there are at least 0 tokens.

iravid:
Oh that's true! Thank you for pointing that out. So I'll read your comments about the burst implementation again and reply.

vasilmkd:
No problem. I was reading through the source code of akka-streams and the implementation of this is quite similar. It seems that they don't care about clock precision as much though.

iravid:
@vasilmkd The current implementation (and the `units + burst` trick) looks good to me! It's a very elegant solution.

This is good to merge by me, unless there's anything else you wanted to address.

vasilmkd:
Happy with how everything works now.

iravid:
Awesome work, thank you @vasilmkd!

PR #1207, opened 2019-07-20
Suspend register in ZStream.effectAsyncInterrupt and harden test

### LINKED ISSUE #1206: Flaky stream test
```
[info]   Stream.effectAsyncInterrupt
[error]     x effectAsyncInterrupt Left
```

iravid:
Resolves #1206

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1536,15 +1536,17 @@ object ZStream extends ZStreamPlatformSpecific {
           for {
             output  <- Queue.bounded[Take[E1, A1]](outputBuffer).toManaged(_.shutdown)
             runtime <- ZIO.runtime[R].toManaged_
-            eitherStream = register(
-              k =>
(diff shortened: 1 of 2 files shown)


dariusrobson:
I see. Was the not wrapping in the UIO(..) the problem here or was it the test?

iravid:
Just the test :-) Wrapped the registration in `UIO` just as a best practice.

dariusrobson:
Ah Cool. Thanks for patching!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1536,15 +1536,17 @@ object ZStream extends ZStreamPlatformSpecific {
...
-            eitherStream = register(
-              k =>
-                runtime.unsafeRunAsync_(
-                  k.foldCauseM(
-                    cause => output.offer(Take.Fail(cause)).unit,
-                    b => output.offer(Take.Value(b)).unit
-                  )
-                )
-            )
+            eitherStream <- UIO {
```
jdegoes: Ouch.

PR #1210, opened 2019-07-22
Fix Queue's error type in docs

amir:


ghostdogpr:
@amir good catch, thank you!

PR #1215, opened 2019-07-22
Removing flaky tests in STM

LGLO:
Removed cases try to prove that navie implementation with `Ref`s won't work (thus we need STM).
Sometimes it happens that stars are against us and these cases fail.
I advocate for removing them because we can break STM in every possible way and these two cases will still pass (usually).


jdegoes:
🙏

PR #723, opened 2019-04-05
Add simple test framework

NeQuissimus:
Add `testFrameworks += new TestFramework("scalaz.zio.test.TestFramework")` to your `build.sbt` and include `scalaz-zio-testkit` as a test dependency.

This is more a POC than anything.
- The matchers are super simple
- The output is not super pretty

But things work and everything seems to integrate with SBT fine.

```scala
package bla

import scalaz.zio._
import scalaz.zio.testkit._

object Bla {
  def helloWorld = ZIO.succeed("Hello World")
  def foobar     = ZIO.fail(new Exception("This is no good"))
}

object BlaSpec extends TestSpec with DefaultRuntime with SimpleMatchers {
  val tests = Map(
    "first test"  -> first,
    "second test" -> second,
    "third test"  -> third
  )

  def first: Task[Result] =
    Bla.helloWorld.map(expect("Hello World"))

  def second: Task[Result] =
    Bla.helloWorld.map(expect("FooBar"))

  def third: Task[Result] =
    Bla.foobar.mapError { e =>
      compare[Throwable]("Did not receive the expected error")(a => a.getMessage == e.getMessage)(e)
    }.flip
}
```

#717

PR #1220, opened 2019-07-23
Prototype ZIO Test

jdegoes:
Unfinished but best to merge after review even in its unfinished state, as it will allow others to continue the project.


adamgfraser:
Would love to contribute to this.

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
```diff
@@ -0,0 +1,37 @@
...
+package zio.test
+
+import zio._
+
+abstract class ZIOTest[R, E] {
+
+  /**
+   * Construct your environment here.
+   */
+  val environment: Managed[Nothing, R]
```
runtologist: How about adding `runtime` as well, so for example execution contexts can be configured and error logs can be redirected?

jdegoes: My idea is to create the runtime in `main`. This way you can run the main function or you can use reflective mechanisms to load all tests and use a single runtime, rather than one runtime per class (which would be slower). That work?

jdegoes:
@adamgfraser Awesome! You could implement the parallel and sequential `ZSpecRunner` and submit a PR to my fork / branch of the project. Sound good?

adamgfraser:
@jdegoes Great!

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
runtologist: Exactly the right thing to do, but isn't that an orthogonal question? 

More specifically, I'm wondering whether it makes sense to be able to configure `Tracer` and `Platform`: For example I'm using a test runtime, that
* logs through slf4j instead of using println
* does not log interruption as failure
* logs only the stacktrace on Traces.

I could also imagine restricting parallelism due to limitations of the build system. 

Not all of these may make sense for a test runner. Most could be exposed as parameters I guess.

jdegoes: I think no matter what, the user needs some control over the platform. I think I can do that and _still_ have a main function by using a decorator. Look for commit incoming in the next 120 seconds...

runtologist: awesome. :)

jdegoes:
Anyone want to work of @NeQuissimus's [pull request](https://github.com/zio/zio/pull/723) (in particular, [here](https://github.com/zio/zio/pull/723/files)) to implement a `ZSpecRender` for JVM console?

It would loop over an executed spec and build the classic list hierarchy, with color-coded text:

```
+ Addition works fine
+ Subtraction works fine
- Value falls within range
  `52` did not satisfy `(equals(42) || (gt(5) && lt(10))`
  `52` did not satisfy `gt(5)`
```

The first failure is the failure of the whole value / whole predicate; the second is the failure of the fragment value / fragment predicate.

The predicate fragment `gt(5)` inside the whole predicate would be colored differently, to draw user's attention to the fragment that failed. This can be accomplished by looking for substrings in the rendered predicate that match the rendered fragment predicate.

Failed test label should be red; succeeded test label should be green.

NeQuissimus:
How do you plan on handling things like `eventually` (loop until true), `continually` (true for a set amount of time), `timeout` (fail if test takes too long) etc.?

testkit/jvm/src/main/scala/zio/test/Predicate.scala
```diff
@@ -0,0 +1,308 @@
...
+class Predicate[-A] private (render: String, val run: A => PredicateResult) extends (A => PredicateResult) { self =>
+  import AssertResult._
+
+  /**
+   * Returns a new predicate that succeeds only if both predicates succeed.
+   */
+  final def &&[A1 <: A](that: => Predicate[A1]): Predicate[A1] =
+    Predicate.predicateDirect(s"(${self} && ${that})") { actual =>
+      self.run(actual) match {
+        case Failure(l) => Failure(l)
```
NeQuissimus: Would this not unapply and recreate the object?
Would `f: Failure => f` be more efficient?

edvmorango:
Hey, @jdegoes I want to work in it, but I'll only be available next week, are you guys in a hurry?

jdegoes:
@NeQuissimus  

`continually` etc. are trivial to write because every test is just an effect, so you can do things like:

```scala
testM {
  (for { ... } yield asset).repeat(Schedule.recurs(5))
}
```

We can include the most common cases as methods on `ZSpec` so you can do:

```scala
testM { }.continually
...
testM { }.eventually
...
```

`timeout` is already implemented in such a fashion, it's trivial to do new ones.

testkit/jvm/src/main/scala/zio/test/Predicate.scala
jdegoes: Very true, but doesn't matter, the overhead of the test framework is inconsequential.

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Place all your tests here.
+   */
+  val tests: ZSpec[R, E, String]
+
+  /**
+   * The platform that will be used to run the tests if the tests are run
+   * through the main function.
+   */
+  def makePlatform(): Platform = PlatformLive.makeDefault().withReportFailure(_ => ())
```
NeQuissimus: :pray:

jdegoes:
@edvmorango Hopefully it will be already implemented by then, but if not, it's yours!

NeQuissimus:
I quite like this so far! :tada:

jdegoes:
@NeQuissimus Then please approve so I can merge, this will make outside help finishing it go much faster! Nothing's set in stone now anyway.

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import zio._
+import zio.internal.{ Platform, PlatformLive }
+
+abstract class ZIOTest[R, E] { self =>
+
+  /**
+   * Construct your environment here.
+   */
+  val environment: Managed[Nothing, R]
```
kaishh: This assumes that tests create their own environment, which feels too concrete (i.e. doesn't fit our usecase!)
Also, `final def main` removes control over running an individual test from the user, pinning a single strategy...
What do you think about `ZTest` referencing the `Runner` that it want to run on and the `Runner` providing an implementation of `main` instead ?

jdegoes: Here's the problem I'm trying to solve:

 - I want to be able to run a suite (existing inside a class) by running its main function

Everything else falls out of here.

There should be other ways, but you can see why we need the environment and other details inside this class: because otherwise we cannot provide runnability on subclasses.

Now, we can give up on runnability. But I think it's quite nice and most IDEs can already run files with a click or two. So it's out of the box ease-of-use.

We can tweak this design, but one of the important details I've been punting on is the extent to which the user must compose specs, and the extent to which the library does that using reflection.

If the user composes everything, it will be more verbose and painful for common cases, but also you can achieve other benefits, such as keeping `R` and so forth polymorphic and pushing that up higher and higher. OTOH with reflection, you need some unit at which everything has been packaged together in a form that's ready to run.

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Place all your tests here.
+   */
+  val tests: ZSpec[R, E, String]
```
kaishh: This should be a `def` to allow accumulation through mutable DSLs

testkit/jvm/src/main/scala/zio/test/ZSpecRunner.scala
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test
+
+import zio.UIO
+
+trait ZSpecRunner {
+  def apply[R, E, L](spec: ZSpec[R, E, L]): UIO[ExecutedSpec[R, E, L]]
```
kaishh: This signature feels too restrictive – ZSpecRunner cannot provide an environment... How about `ZSpecRunner[R]` or maybe another entity that can override/combine Suite-specific Env with a Global Env?

jdegoes: Why combine the runner with the provision of the environment? The provision can be deferred until much later.

Also: is your use case met by simply stitching together ordinary `Spec`, and having, e.g., an application-wide `ZIOTest`? You can compose Spec with `++`, for example, so you just need 1 per application / module.

jdegoes: Also Spec's already have `provide` / `provideSome`

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
jdegoes: Done... but eww! 😉

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
kaishh: That's fine – delegating `main` implementation to someone else doesn't hurt runnability – we can have a default implementor.

jdegoes: I will pull out logic of `main` into something better, and it can be used by `main` and something else. I still think it should be separate from `runner` since that class should be dead simple, going through the tree, running each thing, and collecting the outputs (either sequentially or in parallel).

testkit/jvm/src/main/scala/zio/test/ZSpecRunner.scala
kaishh: > The provision can be deferred until much later.

Not in our usecase, e.g. we want to setup shared components/resources before all tests happen, then close them. So in this case it's `global provision -> run suites -> per-test provision -> run tests ; close per-test ; close global`

> Also: is your use case met by simply stitching together ordinary Spec, and having, e.g., an application-wide ZIOTest

Everything can be done by stitching all tests together! But it would be a bit of a nightmare to write and use such tests, esp. if we can just add a type parameter instead...

jdegoes: If we want to allow R to participate globally, we'll need class tag on R.

I assume you want reflective discovery?

testkit/jvm/src/main/scala/zio/test/ZIOTest.scala
kaishh: Thanks! 👍

testkit/jvm/src/main/scala/zio/test/ZSpecRunner.scala
kaishh: Not sure if we do need a ClassTag:

- If you use non-reflective instantation: then the R is assembled in one place and checked statically – don't need a ClassTag.
- If you use reflective instantation – if ZTest includes a reference to the Runner that it prefers – it can be a direct instantiation, a label or some kind of a handle - it can be indexed by `R` and checked statically
- If you use reflective instantiation – AND if the decision is completely top-down, i.e. runner discovers all tests and tests do not choose their runner – a ClassTag would not really make it safe for the runner to manipulate the environment anyway, ClassTag contains only the java type which is not representative... Personally, in distage-testkit, we'd just process specific subclasses only and redirect non-distage tests to default runner, a ClassTag does not provide value in this case.

Issue #1217, opened 2019-07-23
Enhance combinators for working with sum types (Option, Either, etc.)


jdegoes:
A few operators can allow digging into terms of sum types, pushing error states onto the error channel; and a couple more can allow "popping" error states off the error channel.

```scala
trait ZIO[R, E, A] {
  def left[B, C](implicit ev: A <:< Either[B, C]): ZIO[R, Option[E], B] = ???
  def right[B, C](implicit ev: A <:< Either[B, C]): ZIO[R, Option[E], C] = ???
  def some[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], B] = ???
  def none[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], Unit] = ???
  def head[B](implicit ev: A <:< List[B]): ZIO[R, Option[E], B] = ???
  
  def fuseError[E1, E2 <: E1](default: E2)(implicit ev: E <: Option[E1]): ZIO[R, E1, B] = ???

  def peelError[E1, A1 >: A](default: A1)(implicit ev: E <: Option[E1]): ZIO[R, E1, A1] = ???
}
```

/cc @jczuchnowski who was thinking about these issues.

adamgfraser:
I'm happy to take this if no one else is.

jdegoes:
@adamgfraser 🙏

adamgfraser:
I’m on it.

jczuchnowski:
So just not to double the work, I'll post a PR with what I was working on.

jczuchnowski:
So here it is.
@adamgfraser I suggest I finish the work on Option (I have `some` and `peelError`) and you could take Either. What do you think?
@jdegoes we talked about different names before. I believe I have a kind of `peelError` but with a different signature, right?

adamgfraser:
@jczuchnowski That sounds great. I'll take `left`, `right`, and `head` and leave `some`, `none`, `fuseError`, and `peelError` to you. I just did implementations of `none` and `fuseError` so I will add them as a comment to your PR in case it is helpful to you.

jczuchnowski:
@adamgfraser yes, it'll be helpful, thanks!

jdegoes:
@jczuchnowski Yes, I am not really sure on good names here.

ghostdogpr:
Closing this, both PRs are merged.

PR #1221, opened 2019-07-23
#1217 - operators for option type

jczuchnowski:
Part of work for https://github.com/zio/zio/issues/1217 . In short - improve working with options:
```scala
    val value1: Task[Option[Int]] = Task(Some(1))
    val value2: Task[Option[Int]] = Task(Some(2))

    val result: IO[Option[Throwable], Int] = for {
      val1 <- value1.some
      val2 <- value2.some
    } yield val1 + val2
        
    val resultOption: Task[Option[Int]] = result.peelError
```


adamgfraser:
In case it is helpful, here is the work I did on implementing `none` and `fuseError`.

```
  def none[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], Unit] =
    self.foldM(
      e => ZIO.fail(Some(e)),
      a => ev(a).fold[ZIO[R, Option[E], Unit]](ZIO.succeed(()))(_ => ZIO.fail(None))
    )
```

```
  def fuseError[E1, E2 <: E1](default: E2)(implicit ev: E <:< Option[E1]): ZIO[R, E1, A] =
    self.mapError(e => ev(e).getOrElse(default))
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -395,6 +395,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def get[E1 >: E, B](implicit ev1: E1 =:= Nothing, ev2: A <:< Option[B]): ZIO[R, Unit, B] =
     ZIO.absolve(self.mapError(ev1).map(ev2(_).toRight(())))
 
+  final def some[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], B] = 
+    self.foldM(e => ZIO.fail(Some(e)), a => a.fold[ZIO[R, Option[E], B]](ZIO.fail(Option.empty[E]))(ZIO.succeed(_).mapError(_ => Option.empty[E])))
+
+  final def peelError[E1](implicit ev: E <:< Option[E1]): ZIO[R, E1, Option[A]] =
```
jdegoes: Think we should call this `optional`. Better name than `peelError`, anyway. Otherwise looks good!

jczuchnowski:
@jdegoes @adamgfraser what are the actual use cases for `none` and `fuseError`. I'm trying to come up with a meaningful test cases.

adamgfraser:
@jczuchnowski `none` is basically asserting that the optional value in a `ZIO` is empty and failing if it is not. So you should have:

```scala
IO.succeed(None).none === IO(())
IO.succeed(Some(1)).none === IO.fail(Option.empty)
IO.fail("Fail").none === IO.fail(Some("Fail"))
```

Fuse error is taking a `ZIO` that could fail with 'Some(error)' or 'None' and adding a default error value to fail with to replace the `None` case. So you should have:

```scala
IO.fail(Some("Fail")).fuseError("Default") === IO.fail("Fail")
IO.fail(None).fuseError("Default") === IO.fail("Default")
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -400,6 +400,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def forkOn(ec: ExecutionContext): ZIO[R, E, Fiber[E, A]] =
     self.on(ec).fork
 
+  final def fuseError[E1, E2 <: E1](default: E2)(implicit ev: E <:< Option[E1]): ZIO[R, E1, A] =
```
jdegoes: How about we call this `flattenErrorOption`?

jczuchnowski: Definitely more descriptive.

PR #1222, opened 2019-07-24
Add ZStream#flatMapParSwitch

### LINKED ISSUE #1203: Add ZStream#flatMapParSwitch
Comparing to Rx*, we've got `concatMap` (which is `ZStream#flatMap`) and `mergeMap` (which is `ZStream#flatMapPar`). We're only missing `switchMap`, with semantics as described here: https://www.learnrxjs.io/operators/transformation/switchmap.html.

We could generalize `switchMap` to a combinator called `flatMapParSwitch`:
```scala
trait ZStream[R, E, A] {
  def flatMapParSwitch[R1 <: R, E1 >: E, B](maxOpen: Int)(f: A => ZStream[R1, E1, B]): ZStream[R1, E1, B]
}
```

This operator maps every `A` to a new stream, and merges the output of up to `n` streams in parallel. So far, identical to `flatMapPar`. The difference is that whenever a new `A` is emitted, the oldest* stream of `B` to be opened is cancelled.

\* the strategy of choosing which stream to cancel can also be generalized, but that can be left for a follow-up

vasilmkd:
Resolves #1203

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -684,8 +684,8 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
                   for {
                     latch <- Promise.make[Nothing, Unit]
                     innerStream = Stream
-                      .bracket(permits.acquire *> latch.succeed(()))(_ => permits.release)
-                      .flatMap(_ => f(a))
+                      .bracket(permits.acquire)(_ => permits.release)
+                      .flatMap(_ => ZStream.fromEffect(latch.succeed(())).flatMap(_ => f(a)))
                       .foreach(b => out.offer(Take.Value(b)).unit)
                       .foldCauseM(
                         cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause),
@@ -755,10 +755,10 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
                           cancelers.size.flatMap { size =>
                             if (size < n) UIO.unit
                             else cancelers.take.flatMap(_.succeed(()))
-                          } *> cancelers.offer(canceler) *> permits.acquire *> latch.succeed(()).unit
+                          } *> cancelers.offer(canceler) *> permits.acquire
                         )
                       )(_ => permits.release)
-                      .flatMap(_ => f(a))
+                      .flatMap(_ => ZStream.fromEffect(latch.succeed(())).flatMap(_ => f(a)))
                       .foreach(b => out.offer(Take.Value(b)).unit)
                       .foldCauseM(
                         cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause).unit,
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -718,7 +718,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
                   )
                   // This finalizer makes sure that in all cases, the driver stops spawning new streams
                   // and the inner fibers are signalled to interrupt and actually exit.
-                  .ensuring(interruptInners.succeed(()) *> permits.withPermits(n)(ZIO.unit))
+                  .ensuringFirst(interruptInners.succeed(()) *> permits.withPermits(n)(ZIO.unit))
                   .fork
             s <- ZStream.fromQueue(out).unTake.fold[R2, E2, B1, S].flatMap(fold => fold(s, cont, g))
           } yield s
@@ -778,7 +778,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
                         )
                         .toManaged_
                   )
-                  .ensuring(interruptInners.succeed(()) *> permits.withPermits(n.toLong)(UIO.unit))
+                  .ensuringFirst(interruptInners.succeed(()) *> permits.withPermits(n.toLong)(UIO.unit))
                   .fork
             s <- ZStream.fromQueue(out).unTake.fold[R2, E2, B1, S].flatMap(fold => fold(s, cont, g))
           } yield s
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -728,6 +728,68 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
     }
 
   /**
+   * Maps each element of this stream to another stream and returns the non-deterministic merge
+   * of those streams, executing up to `n` inner streams concurrently. When a new stream is created
+   * from an element of the source stream, the oldest executing stream is cancelled. Up to `bufferSize`
+   * elements of the produced streams may be buffered in memory by this operator.
+   */
+  final def flatMapParSwitch[R1 <: R, E1 >: E, B](n: Int, bufferSize: Int = 16)(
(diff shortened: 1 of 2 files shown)


vasilmkd:
In the current implementation, the Streams produced from each element of the outer stream can enter the "execution window of `n` parallel streams" completely non-deterministically.

E.g. in `Stream(Stream(1), Stream.never).flatMapParSwitch`, the never ending stream can enter the execution window before the `Stream(1)`. Is this desired behavior or should the ordering be preserved?

As far as I know, flatMapPar is also non-deterministic in this way.

iravid:
Hi @vasilmkd, looks really good from a cursory look. I hope to have more time to dig in tomorrow. Thank you for contributing this so quickly!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -727,6 +727,68 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+                           innerFailure.await
+                             .raceWith(permits.withPermits(n.toLong)(UIO.unit))(
+                               leftDone =
+                                 (_, permitAcquisition) => interruptInners.succeed(()) *> permitAcquisition.interrupt,
+                               rightDone = (_, failureAwait) => out.offer(Take.End) *> failureAwait.interrupt
+                             )
+                             .toManaged_
+                       )
+                       .fork
+            _ <- ZManaged.finalizer(
```
iravid: Something I realized after developing `mapMPar` but haven't ported back to `flatMapPar`: using `ZManaged.finalizer` can be unsafe because the computation can be canceled between the previous step and the finalizer step, leading to the finalizer not running.

Let's instead use `.onExit` on the previous step to ensure that a finalizer is installed. And if you can make that change on `flatMapPar` too, that'd be nice :-)

iravid:
Regarding the ordering: I think it'd be a positive change to move the `latch.succeed(())` into `flatMap(_ => latch.succeed(()) *> f(a))`. I wonder if that is enough though; we have to make sure we don't move the latch release too deeply into `f(a)` or else we risk hangs due to streams like `Stream.never`.

vasilmkd:
> Regarding the ordering: I think it'd be a positive change to move the `latch.succeed(())` into `flatMap(_ => latch.succeed(()) *> f(a))` (in `flatMapPar` as well). I wonder if that is enough though; we have to make sure we don't move the latch release too deeply into `f(a)` or else we risk hangs due to streams like `Stream.never`.

I made this change, however, I see no difference in behavior.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -727,6 +727,68 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+            interruptInners <- Promise.make[Nothing, Unit].toManaged_
+            cancelers       <- Queue.bounded[Promise[Nothing, Unit]](n).toManaged(_.shutdown)
+            cancelersGuard  <- Semaphore.make(1).toManaged_
+            driver <- self.foreachManaged { a =>
+                       for {
+                         latch    <- Promise.make[Nothing, Unit]
+                         canceler <- Promise.make[Nothing, Unit]
+                         innerStream = Stream
+                           .bracket(
+                             cancelersGuard.withPermit(
```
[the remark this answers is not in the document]
vasilmkd: Thanks, I appreciate it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Done. Both flatMapPar and flatMapParSwitch.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -680,48 +680,106 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+                          leftDone =
+                            (_, permitAcquisition) => interruptInners.succeed(()) *> permitAcquisition.interrupt,
+                          // All fibers completed successfully, so we signal that we're done.
+                          rightDone = (_, failureAwait) => out.offer(Take.End) *> failureAwait.interrupt
+                        )
+                        .toManaged_
+                  )
+                  // This finalizer makes sure that in all cases, the driver stops spawning new streams
+                  // and the inner fibers are signalled to interrupt and actually exit.
+                  .ensuring(interruptInners.succeed(()) *> permits.withPermits(n)(ZIO.unit))
```
iravid: This needs to be an `ensuringFirst`, otherwise the finalizer ordering will be wrong; `ensuring` appends a finalizer so it runs at the end. This means that the finalizer for the driver stream runs first, and only then will the inner fibers be interrupted.

You can actually keep the equivalence to the previous formulation with `ZManaged#onExitFirst`.

iravid: Oh but the tests are still passing 🤨 Well this is interesting. I'll dig in some more.

iravid: Ah yes. This is probably just racey. There's nothing *preventing* the inners from completing before the outer in the finalizer ordering test.

vasilmkd: Wait, that's why we have the `n` permits semaphore (`permits`), the driver finalizer kicks off, tells the inner fibers to terminate and waits for their termination (until it succeeds in collecting `n` permits from the semaphore), at which point all of the inner fibers will have terminated and finalized.

Or am I missing something?

iravid: No, that's exactly it. However, consider that the driver stream has its own set of finalizers (from `self`) that need to run; for example, the stream you're flatMapping might have acquired a resource that the inner fibers are using.

When you do `ensuring`, the finalizer passed to `ensuring` is *appended* to the list of finalizers, not prepended. So the resulting set of finalizers for the driver stream is:
```
selfFinalizer *> interruptInners.succeed(()) *> permits.withPermits(n)(ZIO.unit)
```

Which means that the resources acquired in `self` would be released before the inner fibers.

vasilmkd: So just change `ensuring` to `ensuringFirst`? Both here and in `flatMapPar` right?

iravid: I'd go with `.fork.onExitFirst(...)`, and interrupt the driver fiber before `interruptInners.succeed(())` (if it is Exit.Success).

vasilmkd: Well, I may be wrong here, but the `driver` fiber is not available in its own definition block. Doesn't `fork` on a ZManaged interrupt the fiber itself, then run all of the finalizers?

iravid:
@vasilmkd Let's deal with the execution order issue you've noticed separately. Can you open an issue that describes what you're seeing?

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: You're right, that won't be possible with `ensuringFirst`, but `onExitFirst` lets you access the result. But I think you're right with your note about the interruption of the fiber, so `ensuringFirst` is actually sufficient. Thanks for being extra attentive here :-)

vasilmkd:
I will, let me finish up the other PRs first. I think I have some sample code too.

vasilmkd:
@iravid This is also ready for review.

PR #1223, opened 2019-07-24
Add ZIO#left, ZIO#right, and ZIO#head

adamgfraser:
Adds combinators for working with `Either` and `List` as proposed in #1217.

Should these methods be marked `final`? I noticed that some methods on `ZIO` are marked `final` and others are not and it is not clear to me what the logic is for determining this.

Also, I noticed there are existing `left` and `right` methods on `ZIO` and wonder if we should remove or rename them to avoid confusion. The `left` method takes an `A` value and constructs a `UIO[Either[A, Nothing]]`. This duplicates existing functionality that exists on the `ZIO` companion object and as a constructor seems to more properly belong on the companion object than an individual `ZIO` instance. The `right` method takes a `ZIO[R, E, A]` and a new type parameter `C` and returns a `ZIO[Either[C, R], E, Either[C, A]]`. It has no documentation, no tests, and no dependencies.


adamgfraser:
Actually the existing `right` method conflicts with the new one so I removed it for now. Can add it back with another name if people think that makes sense.

Also, it looks like there is some flakiness in an unrelated test. I'm getting a timeout in "supervise fibers in supervised" in `RTSSpec` in one of the CI builds and this PR is just adding a new methods and a few tests that should not take appreciable time.

This should be ready for review now.

adamgfraser:
@jdegoes Could I get a review of this when you have the chance?

jdegoes:
> Also, I noticed there are existing left and right methods on ZIO and wonder if we should remove or rename them to avoid confusion. 

👍 They come from `Arrow` type class / combinators. I'm not sure why `left` / `right` aren't symmetrical (they should be). Possibly better names for those would be `onLeft` / `onRight`, as they take the channel of information and push it to left or right; (also `onFirst`, `onSecond` for the tuple variations, which are probably called first / second now).

adamgfraser:
Done.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -645,6 +665,15 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       r <- e.fold(self.map(Left(_)) provide _, that.map(Right(_)) provide _)
     } yield r
 
+  /**
+   * Returns a successful effect if the value is `Left`, or fails with the error `None`.
+   */
+  def left[B, C](implicit ev: A <:< Either[B, C]): ZIO[R, Option[E], B] =
+    self.foldM(
+      e => ZIO.fail(Some(e)),
+      a => ev(a).fold(ZIO.succeed(_), _ => ZIO.fail(None))
```
ghostdogpr: Super minor but I think you don’t need the `(_)`

adamgfraser: Nice!

Issue #1227, opened 2019-07-24
Make `testkit` cross-platform (Scala.js)


jdegoes:
Right now, `testkit` is entirely JVM, but it can work on Scala.js as well, possibly without changes.

This ticket is to move source files inside `testkit` from the `jvm` directory to the `shared` directory and fix any issues that arise in the process.

vasilmkd:
I can take this.

vasilmkd:
This issue can be closed probably.

mgibowski:
Hi, while rebasing my branch I have disovered that some specs (`test.mock.*`) are not JS-compatible because of dependency on `TestRuntime`.

It was not visible, as in the current sbt configuration these tests were actaully nevery really compiled or run in JS.

If you want to see how current sbt could be changed to see them failing, I created a [branch](https://github.com/mgibowski/zio/tree/failing-js-test).

This will be also visible in the PR I whill shortly submit. In that case I decided it is best to bring these back to `jvm` folder.

Not sure if this issue should get reopened or not.  It is these 5 tests that are not JS-compatible:
* `ClockSpec.scala` 
* `ConsoleSpec.scala` 
* `RandomSpec.scala` 
* `SchedulerSpec.scala`
* `SystemSpec.scala`

jdegoes:
@mgibowski 

`TestRuntime` will go away soon, so I'd say let's not worry about this. Ultimately we want all these tests running on Scala.js, and that will happen shortly after ZIO Test is completed.

vasilmkd:
Sorry for all the confusion this caused.

mgibowski:
> Sorry for all the confusion this caused.

No worries! 😉

PR #1230, opened 2019-07-24
Make `test` compile to JS

vasilmkd:
Addresses #1227.


jdegoes:
@vasilmkd Thank you again! Great work. 🙏

LGLO:
@vasilmkd I'm bit late to the party. Could you check that when some project depends on `test` and is ScalaJS itself, then it will not "compile" zio test cases?
What I mean: we don't want zio-core test cases to be JS "compiled" when someone needs zio-test framework.

vasilmkd:
I'm sorry, I don't quite understand what you mean. These changes are already merged, maybe you could try this yourself? I'm sorry for the confusion.

LGLO:
Ok, I'll do!

LGLO:
@vasilmkd All is good, sorry for my false alarm!
zio-test jar contains only "framework" files!

vasilmkd:
I'm glad to hear it because I definitely did not foresee that. 😅

PR #1232, opened 2019-07-25
Add ZManaged.makeExit that allows to handle `Exit` by the finalizer.

### LINKED ISSUE #1231: Add ZManaged.makeExit
The signature might be: 
```scala
def makeExit[R, E, A](acquire: ZIO[R, E, A])(release: (A, Exit[_, _]) => ZIO[R, Nothing, _]): ZManaged[R, E, A]
```
or
```scala
def makeExit[R, E, A](acquire: ZIO[R, E, A])(release: (A, Option[Cause[_]]) => ZIO[R, Nothing, _]): ZManaged[R, E, A]
```

One possible use case for this can be the saga pattern.

Created as discussed on gitter.

vvlevykin:
Resolves #1231.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/ZManagedSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/ZManagedSpec.scala
@@ -22,6 +22,9 @@ class ZManagedSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends Test
     Constructs an uninterruptible Managed value. $uninterruptible
   ZManaged.reserve
     Interruption is possible when using this form. $interruptible
+  ZManaged.makeExit
+    Invokes with the failure of the use. $invokesWithUseFailure
+    Invokes with the failure of the subsequent acquire. $invokesWithAcquireFailure
   ZManaged.ensuring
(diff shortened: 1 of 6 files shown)


iravid:
This needs a rebase and a fix to mdoc and should be good to go!

vvlevykin:
This example from the doc becomes clunky, what do you suggest instead of `???` here?
```scala
val data: IO[IOException, String] = for {
  p <- Promise.make[IOException,File]
  res = Reservation(openFile("x", p), _ => closeFile(p))
  result <- (for {
      fiber  <- res.acquire.fork
      ex     <- fiber.interrupt
      data   <- ex.toEither.fold(e => IO.fail(new InterruptedIOException("Stop!")), file => readFile(file))
    } yield data).ensuring(res.release(???))
} yield result
```

iravid:
@vvlevykin I'd just delete the entire section on that. `Reservation` doesn't have any value outside the context of `ZManaged` and that example does not make sense to me.

iravid:
Nicely done - thanks @vvlevykin!

PR #1234, opened 2019-07-25
Implement ZSpecRunner

adamgfraser:
Implements parallel and sequential `ZSpecRunner` as discussed in #1220.

I changed the type signature to require a `ZSpec` with the environment already provided. Since we need to run the `ZSpec` to get a `TestResult` and the result type is a `UIO` that does not depend on any environment I think this is necessary. In `ZIOTestMain` we are providing the environment before running the tests so I don't think this is a problem.

Sorry if I have butchered anything here. Definitely open to feedback!


test/shared/src/main/scala/zio/test/ZSpecRunner.scala
```diff
@@ -28,15 +28,55 @@ object ZSpecRunner {
    * Runs tests in parallel, up to the specified limit.
    */
   def parallel(n: Int): ZSpecRunner = new ZSpecRunner {
-    val _ = n
-
-    def run[R, E, L](spec: ZSpec[R, E, L]): UIO[ExecutedSpec[R, E, L]] = ???
+    def run[E, L](spec: ZSpec[Any, E, L]): UIO[ExecutedSpec[Any, E, L]] =
```
jdegoes: Can you change this to be polymorphic in `R`? You just need to change the return type from `UIO` to `RIO` (sorry, my bad).

Otherwise, looks great!

test/shared/src/main/scala/zio/test/ZSpecRunner.scala
```diff
@@ -19,7 +19,7 @@ package zio.test
 import zio._
 
 trait ZSpecRunner {
-  def run[E, L](spec: ZSpec[Any, E, L]): UIO[ExecutedSpec[Any, E, L]]
+  def run[R, E, L](spec: ZSpec[R, E, L]): RIO[R, ExecutedSpec[R, E, L]]
```
jdegoes: And looks like I made yet another error: instead of `RIO` we can use `ZIO[R, Nothing, ...]` because running cannot (should not) fail.

Sorry about that!

PR #1235, opened 2019-07-25
Signal stream end ZStream.effectAsync

### LINKED ISSUE #1200: ZStream#effectAsync should provide a way for the caller to signal stream end
We can perhaps add an overload that uses `Option[A]` as the value for the ZIO, but that would unfortunately incur an allocation on each element.

@jdegoes any ideas? Maybe `Option[E]` on the error channel?

vasilmkd:
Resolves #1200. 

The PR currently does not introduce overloaded effectAsync methods. Let me know if you want to add them and keep the original type signatures.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1558,7 +1558,7 @@ object ZStream extends ZStreamPlatformSpecific {
                              register(
                                k =>
                                  runtime.unsafeRunAsync_(
-                                   k.foldM[R, E, Option[A]](
+                                   k.foldM(
                                        _.fold[ZIO[R, E, Option[A]]](UIO.succeed(None))(e => IO.fail(e)),
                                        a => UIO.succeed(Some(a))
                                      )
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1512,7 +1512,7 @@ object ZStream extends ZStreamPlatformSpecific {
 
   /**
    * Creates a stream from an asynchronous callback that can be called multiple times
-   * The registration of the callback itself returns an effect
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1493,10 +1496,14 @@ object ZStream extends ZStreamPlatformSpecific {
                             register(
                               k =>
                                 runtime.unsafeRunAsync_(
-                                  k.foldCauseM(
-                                    cause => output.offer(Take.Fail(cause)).unit,
-                                    b => output.offer(Take.Value(b)).unit
-                                  )
+                                  k.foldM(
```
iravid: Is it possible to merge the `foldM` and `foldCauseM` (through `Cause#failureOrCause`)? And same re the variants below

vasilmkd: I'm not sure it keeps the same semantics though. `Cause#failureOrCause` will `extract` the error from the cause, which opens a new logical branch, what to do when there is a type error that is not `None`. We used to just report it in the `foldCauseM`.

iravid: Well, in `foldCauseM`, you get a `Cause[Option[E]]`. `failureOrCause` on that would give you an `Either[Option[E], Cause[Nothing]`. Let's dismantle that:
- `Left(None)` - this means `Take.End`.
- `Left(Some(e: E))` - this is an actual error, so it's `Take.Fail(Cause.fail(e))`
- `Right(c)` - this is a defect or interruption, so it's still a failure that needs to be propagated downstream. So `Take.Fail(c)`.

Makes sense?

vasilmkd: Got it, thanks. I was just worried I was changing the semantics of the previous implementation.

vasilmkd:
@iravid This is now ready for review.

PR #1241, opened 2019-07-25
Fixes 'doc' failure because of unknown reference to 'weave'

LGLO:
It just fixes 'doc' and other dependent tasks (like publish)


jdegoes:
@LGLO Awesome. 🙏

PR #1246, opened 2019-07-26
Add whenCase and whenCaseM

ghostdogpr:
This is a helper for turning this:
```scala
someOption match {
  case Some(v) => someEffect(v)
  case _       => IO.unit
}
```
into this:
```scala
IO.whenCase(someOption) { case Some(v) => someEffect(v) }
```

This pattern comes pretty often in my code (not just with Option) and the `case _ => IO.unit` feels unnecessary.


jdegoes:
@ghostdogpr Looks super useful!

PR #1247, opened 2019-07-26
WIP: Ongoing work to ZIO Test

jdegoes:


test/shared/src/main/scala/zio/test/Predicate.scala
```diff
@@ -17,6 +17,7 @@
 package zio.test
 
 import zio.Exit
+import scala.reflect.ClassTag
```
NeQuissimus: Unused

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -51,6 +55,19 @@ object TestAspect {
       def apply[R >: R0 <: R0, E >: E0 <: E0](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] = f(test)
     }
 
+  /**
+   * An aspect that retries a test until success, without limit.
```
NeQuissimus: I can compose this with a limit?

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -27,12 +27,13 @@ package zio
  *
  * {{{
```
NeQuissimus: Should we introduce something that type-checks Scaladoc or move these examples into the microsite?

test/shared/src/main/scala/zio/test/TestAspect.scala
jdegoes: Sure, e.g. `retry >>> timeout(60.seconds)`, but if you want something fancier, you're better off using `repeat(schedule)`.

test/shared/src/main/scala/zio/test/package.scala
jdegoes: For sure, but let's wait till API stabilizes a bit, probably another week.

test/shared/src/main/scala/zio/test/TestAspect.scala
NeQuissimus: What I figured, just wanted to make sure :)

jdegoes: You know how I _LOVE_ composability. 😉

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -68,6 +68,25 @@ object TestAspect {
...
     }
 
+  /**
+   * An aspect that repeats the test a specified number of times, ensuring it
+   * is stable ("non-flaky").
+   */
+  def nonFlaky(n0: Int): TestAspectPoly =
+    new TestAspectPoly {
+      def apply[R >: Nothing <: Any, E >: Nothing <: Any](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] = {
+        def repeat(n: Int): ZIO[R, E, TestResult] =
```
NeQuissimus: You should be able to put a `@tailrec` here, no?

jdegoes: It's monadically tail recursive, so the compiler can't accept that as `@tailrec`.

NeQuissimus: Ah, that's true...

Issue #1225, opened 2019-07-24
Pull out all core tests into separate `core-tests` project


jdegoes:
We will use _ZIO Test_ for all tests. In order to avoid circular dependencies, we should pull out all `core` tests into a new project: `core-tests`. This will depend both on `testkit` and on `core` and will contain all tests currently in `core`.

Some basic level of familiarity with `sbt` is required for this ticket.

mgibowski:
I volunteer to take this.

jdegoes:
@mgibowski Thank you! 🙏

ghostdogpr:
Closed by #1248

jdegoes:
@ghostdogpr Thank you for all the cleanup! I should remind people to use `Fixes #xyz` syntax in pull requests...

PR #1248, opened 2019-07-26
Pull out all core tests into `core-tests`

mgibowski:
Addresses #1225

Some comments:

* The core module was publishing test classes as part of the artifact (`publishArtifact in (Test, packageBin) := true` ), so I moved that setting to the new `core-tests` module as this is where the core tests now reside.
* Most of the tests were moved to `core-tests` module, but some base classes (`TestRuntime`, `BaseCrossPlatformSpec`) were also used in the `test` module, so they landed over there.
* Moved back some tests in `test` module from `shared` to `jvm`, as previously they were actually not run in JS. More details [here](https://github.com/zio/zio/issues/1227#issuecomment-515421446)


jdegoes:
@mgibowski Looks great!

What did this end up doing with "not truly JS" tests?

mgibowski:
@jdegoes thanks for the review!

> What did this end up doing with "not truly JS" tests?

I needed to move back the tests from `test/shared/src/test/scala/zio/test/mock/` to `test/jvm/src/test/scala/zio/test/mock/`. Otherwise there were JS compilation errors.

The difference is that currently on the master the `test` module doesn't get compiled and tested in JS (some non-obvious `sbt` behavior). In this branch it does.

jdegoes:
@mgibowski That's totally fine, we can move them back once `TestRuntime` is gone, hopefully next week.

Ready to merge as soon as conflicts are fixed!

mgibowski:
@jdegoes ok, great!

Conflicts fixed.

Issue #1250, opened 2019-07-26
Change Promise#done to take an Exit[E, A]


iravid:
Rationale: we have `Promise#fail(E)`, `Promise#halt(Cause[E])`, `Promise#die(Throwable)`. These match the `ZIO.fail/halt/die` constructors.

The existing `Promise#done` should be renamed to `Promise#completeWith`.

Thoughts? @ghostdogpr @jdegoes

Signatures:
```scala
class Promise[E, A] {
  def done(exit: Exit[E, A]): UIO[Boolean]
  def completeWith(io: IO[E, A]): UIO[Boolean]
}
```

jdegoes:
💯 

Consistency matters a lot for usability.

mgibowski:
I would be glad to make this change.

mtsokol:
I would like to try to do it.

ghostdogpr:
Closed by #1251

PR #1251, opened 2019-07-26
Changed Promise#done to take an Exit

mtsokol:
Addresses https://github.com/zio/zio/issues/1250

@iravid I think it's done.

### DIFF
--- a/core/shared/src/main/scala/zio/Promise.scala
+++ b/core/shared/src/main/scala/zio/Promise.scala
@@ -70,13 +70,19 @@ class Promise[E, A] private (private val state: AtomicReference[State[E, A]]) ex
    * Kills the promise with the specified error, which will be propagated to all
    * fibers waiting on the value of the promise.
    */
-  final def die(e: Throwable): UIO[Boolean] = done(IO.die(e))
+  final def die(e: Throwable): UIO[Boolean] = completeWith(IO.die(e))
+
+  /**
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/zio/Promise.scala
```diff
@@ -70,13 +70,19 @@ class Promise[E, A] private (private val state: AtomicReference[State[E, A]]) ex
...
+   * fibers waiting on the value of the promise.
+   */
+  final def done(e: Exit[E, A]): UIO[Boolean] = completeWith(IO.done(e))
 
   /**
    * Completes the promise with the specified result. If the specified promise
    * has already been completed, the method will produce false.
    */
-  final def done(io: IO[E, A]): UIO[Boolean] =
+  final def completeWith(io: IO[E, A]): UIO[Boolean] =
```
jdegoes: I'd just call this `complete`. Also it would be nice to keep the old `done` (an overloaded method, as an alias) with a deprecation warning. Just to give people a chance to migrate. We can delete it on 1.0.

mtsokol: @jdegoes Sure, done it.

iravid:
Needs a rebase and then we can get this in.

mtsokol:
@iravid rebase done.

jdegoes:
@mtsokol Thank you!

PR #1253, opened 2019-07-27
Add a link to slf4zio

mlangc:
@NeQuissimus I decided not to migrate my code to [zio-slf4j](https://github.com/NeQuissimus/zio-slf4j) for the time being, because for the moment it would generate too much work for too little value. Also, I'm still experimenting with some concepts and APIs, and merging would limit the speed of iterations at this stage. In the long run though, having all ZIO SLF4J compatibility code in one library still makes sense, so we shouldn't loose sight of this.


jdegoes:
Looks good!

PR #1252, opened 2019-07-26
Add "Setter" / "Update" Methods to Mock Services

adamgfraser:
Adds "setter" / "update" methods to `TestConsole`, `TestSystem`, and `TestRandom` mock services and hides `Ref` implementation as proposed in #1240.

Issue #1129, opened 2019-07-03
Create TestEnvironment 


jdegoes:
Inside the `testkit` project, we should stitch together a complete `TestEnvironment` to make it easy to test apps that just use ZIO modules; and to make it easy to use a wired-up `TestEnvironment` in other larger environments.

To complete this ticket, familiarity with `Runtime` / `DefaultRuntime` / `App`, and ZIO environment is required.

LGLO:
In some interop projects we load all ZIO tests to have this enabled. This means scalajs tests from ZIO are compiled just to have base tests classes available.
I think testkit should be replacement for that pattern.

dkarlinsky:
Hi, I'm really new to ZIO, but I am very interested in how it could make testing effectful code nicer.
I'm a big fan of leaning by doing, so I think contributing would be a great way to get into this.
I would like to take this issue, but I would likely need a bit more guidance. 
What would be the best channel for this?

jdegoes:
@dkarlinsky Are you available to take this?

We need a version for JVM, something like this:

```scala
import zio.blocking._
  import zio.clock.Clock 
  import zio.console.Console
  import zio.scheduler.Scheduler
  import zio.random.Random
  import zio.system.System
  import zio.internal.Platform
  import zio.internal.PlatformLive   

object TestEnvironment {
  val Value: Managed[Nothing, DefaultRuntime#Environment] = 
    Managed.fromEffect {
      import zio.testkit._ 

      for {
        bootstrap <- ZIO.effect(PlatformLive.fromExecutionContext(scala.concurrent.ExecutionContext.Implicits.global))
        clockRef <- Ref.make(TestClock.DefaultData)
        consoleRef <- Ref.make(TestConsole.DefaultData)
        randomRef <- Ref.make(TestRandom.DefaultData)
        systemRef <- Ref.make(TestSystem.DefaultData)
      } yield new Blocking with Clock with Console with Random with Scheduler with System {
        val clock: TestClock = TestClock(clockRef)
        val console: TestConsole = TestConsole(consoleRef)
        val random: TestRandom = TestRandom(randomRef)
        val scheduler: TestScheduler = TestScheduler(clockRef, Runtime(clockRef, TestClock(clockRef)))
        val system: TestSystem = TestSystem(systemRef)
        val blocking: Blocking = new Blocking.Live {}
      }
    }
}
```

and one for `Scala.js`, without the `Blocking` (which does not exist for `Scala.js`). 

[...]

PR #1255, opened 2019-07-27
WIP: Create TestEnvironment

adamgfraser:
Creates a `TestEnvironment` as proposed in #1129. Will add more convenience functions and tests once #1252 is merged but wanted to get this out there.


test/js/src/main/scala/zio/test/mock/TestEnvironment.scala
```diff
@@ -0,0 +1,65 @@
...
+        clock     <- Ref.make(TestClock.DefaultData).map(TestClock(_))
+        console   <- Ref.make(TestConsole.DefaultData).map(TestConsole(_))
+        random    <- Ref.make(TestRandom.DefaultData).map(TestRandom(_))
+        scheduler = TestScheduler(clock.ref, Runtime(Clock(clock), bootstrap))
+        system    <- Ref.make(TestSystem.DefaultData).map(TestSystem(_))
+      } yield new TestEnvironment(clock, console, random, scheduler, system)
+    }
+
+  object testConsole {
+    def feedLines(lines: String*): ZIO[TestEnvironment, Nothing, Unit] =
```
jdegoes: This is fine for this pull request, but in general, I think I'd put these helpers inside the companion objects of the `Test*` classes.

e.g.:

```scala
TestConsole.feedLines(...)
```

Then there's a logical place for users to look for them.

test/js/src/main/scala/zio/test/mock/TestEnvironment.scala
```diff
@@ -0,0 +1,65 @@
...
+  system: TestSystem,
+) extends Clock
+    with Console
+    with Random
+    with Scheduler
+    with System
+
+object TestEnvironment {
+
+  val Value: Managed[Nothing, TestEnvironment] =
```
jdegoes: One more organizational tip:

Let's rename `TestEnvironment` => `TestEnvironmentInternals`, and make it package private (`private[test]`).

Then inside `zio.test.mock` package object, we can do:

```scala
package object mock {
  val TestEnvironment: TestEnvironment = TestEnvironmentInternals.Value
}
```

Then a user has type `TestEnvironment` and can use value `TestEnvironment`, which is more Scala-idiomatic.

jdegoes: Hmm, actually, I'm not so sure about this, because `TestEnvironment.Value` has type `Managed[...]`. So let's leave this the way it is. 

We can add this value in `package object mock`:

```scala
val TestEnvironmentManaged: Managed[TestEnvironment] = TestEnvironment.Value
```

In fact, it probably makes sense to rename `TestEnvironment.Value` to `TestEnvironment.ManagedDefault` (a managed, default test environment).

jdegoes:
@adamgfraser Can you do the `Mock` rename in this PR and merge the conflicts?

adamgfraser:
@jdegoes Yes.

jdegoes:
Er, mock rename for `TestEnvironment`, the other ones were taken care of by @dkarlinsky

test/js/src/main/scala/zio/test/mock/TestEnvironment.scala
adamgfraser: Yes I think this makes sense. One thing I am thinking about is whether it makes sense to have something like:

```
trait TestConsole extends Console {
  val console: TestConsoleImplementation
}

case class TestConsoleImplementation extends Console.Service[Any]
```

That way when we define these helper methods we can only depend on the specific environment type that these methods require and users could get the benefit of the method if they only depend on certain mock objects instead of the mock environment. Not sure about naming. What do you think?

jdegoes: Yes, you are right. Let's follow the ZIO conventions though:

```scala
trait MockConsole extends Console {
  val console: MockConsole.Service[Any]
}
object MockConsole {
  trait Service[R] extends Console.Service[R] { def feedLines(...) }

  def feedLines(lines: String*): ZIO[MockConsole, Nothing, Unit] = ...
}
```

adamgfraser: Great. I'm on it!

adamgfraser:

adamgfraser:
@jdegoes Thanks! Means a lot coming from you.

PR #1256, opened 2019-07-27
Make ZManaged#mergeAllParN and ZManaged#reduceAllParN tolerant to finalizer defects

### LINKED ISSUE #1254: ZManaged.mergeAllParN and ZManaged.reduceAllParN are not tolerant to finalizer defects
ZManaged.mergeAllParN and ZManaged.reduceAllParN should compose finalizers in a list and run them with`.run`. 

This is the same issue as #940.

adamgfraser:
Resolves #1254.


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1020,7 +1020,11 @@ object ZManaged {
               } yield b).ensuring((queue.shutdown *> ZIO.foreach_(fibers)(_.interrupt)).uninterruptible)
             }
           },
-          ZIO.flatten(finalizers.get)
+          for {
+            fs    <- finalizers.get
+            exits <- ZIO.foreach(fs)(_.run)
+            _     <- ZIO.done(Exit.collectAll(exits).getOrElse(Exit.unit))
```
iravid: Let's use `Exit.collectAllPar` here (and in `reduceAllParN`) as that would better reflect the nature of errors that may have occured.

adamgfraser: All set.

PR #1262, opened 2019-07-27
Add replicate method - Fix Issue #1261

### LINKED ISSUE #1261: Implement ZIO.replicate
A quite common idiom in ZIO is `List.fill(n)(effect)`, to replicate an effect `n` times.

However, this consumes unnecessary memory, since the effect is the same, there is no need to create `n` copies of it.

Instead, we can return an `Iterable` that _pretends_ to be that long, and whose iterator simply returns the same effect `n` times in a row.

It may be a challenge making this in a way that compiles on 2.12 and 2.13 simultaneously.

```scala
object ZIO {
  def replicate[R, E, A](n: Int)(effect: ZIO[R, E, A]): Iterable[ZIO[R, E, A]] = ???
}
```

Be sure to add delegations to this method inside the companion objects for the type aliases:

 - [ ] UIO
 - [ ] RIO
 - [ ] Task
 - [ ] IO

These methods will simply delegate to `ZIO.replicate`, with the appropriate type aliases used instead of `ZIO`.

dorsev:


dorsev:
Just realized my implementation is eager. 
1. If that is a problem maybe I'll implement it twice using
```scala
Stream.range(0,n).map(_ => effect)
```
But that requires a different implementation for 2.13:
```scala
LazyList.range(0,n).map(_ => effect)
```

2. maybe It's OK for ZIO to return an eager implementation and let ZStream return a lazy one? 
(Since the Iterable trait says nothing about eager or laziness?)

dorsev:
Another question:
`However, this consumes unnecessary memory, since the effect is the same, there is no need to create n copies of it.`

As far as I understand, List.fill(n)(effect) copies only pointers, not just the values.
Is that what we are trying to optimize for?

PR #1265, opened 2019-07-27
Guarantee ordering of ZStream#flatMapPar and ZStream#flatMapParSwitch

### LINKED ISSUE #1249: Guarantee deterministic ordering for ZStream#flatMapPar and ZStream#flatMapParSwitch
I've already started work on this.

vasilmkd:
Resolves #1249.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -682,22 +682,16 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
             //   - On completion, an inner fiber does nothing
             _ <- self.foreachManaged { a =>
                   for {
-                    latch <- Promise.make[Nothing, Unit]
+                    _ <- permits.acquire
                     innerStream = Stream
-                      .bracket(permits.acquire)(_ => permits.release)
-                      .flatMap(_ => ZStream.fromEffect(latch.succeed(())).flatMap(_ => f(a)))
+                      .bracket(UIO.unit)(_ => permits.release)
+                      .flatMap(_ => f(a))
                       .foreach(b => out.offer(Take.Value(b)).unit)
                       .foldCauseM(
-                        cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause),
+                        cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause).unit,
                         _ => ZIO.unit
                       )
                     _ <- (innerStream race interruptInners.await).fork
-                    // Make sure that the current inner stream has actually succeeded in acquiring
-                    // a permit before continuing. Otherwise, two bad things happen:
-                    // - we might needlessly fork inner streams without available permits
-                    // - worse, we could reach the end of the stream and acquire the permits ourselves
-                    //   before the inners had a chance to start
-                    _ <- latch.await
                   } yield ()
                 }.foldCauseM(
                     cause => (interruptInners.succeed(()) *> out.offer(Take.Fail(cause))).unit.toManaged_,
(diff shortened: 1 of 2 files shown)


vasilmkd:
@iravid Thanks for the help yesterday. This is now ready for review.

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -707,6 +710,12 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       slurp(leftStream) must_=== slurp(rightStream)
     }
 
+  private def flatMapParGuaranteeOrdering = prop { m: List[Int] =>
```
iravid: As I mentioned - super important test to have! Great thinking here.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -683,20 +683,16 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+                      .bracket(latch.succeed(()))(_ => permits.release)
+                      .flatMap(_ => f(a))
                       .foreach(b => out.offer(Take.Value(b)).unit)
                       .foldCauseM(
-                        cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause),
+                        cause => out.offer(Take.Fail(cause)) *> innerFailure.fail(cause).unit,
                         _ => ZIO.unit
                       )
                     _ <- (innerStream race interruptInners.await).fork
-                    // Make sure that the current inner stream has actually succeeded in acquiring
```
iravid: Could we keep the comment and update it instead of dropping it entirely? :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -683,20 +683,16 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
             _ <- self.foreachManaged { a =>
                   for {
                     latch <- Promise.make[Nothing, Unit]
+                    _     <- permits.acquire
```
iravid: Let's add a comment that explains why an unbracketed `permits.acquire` here is safe - from my reasoning this is ok because the fiber executing this could only be interrupted if the entire stream is interrupted, and at that point it doesn't matter if we leak an acquisition.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -744,21 +740,18 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
             innerFailure    <- Promise.make[Cause[E1], Nothing].toManaged_
             interruptInners <- Promise.make[Nothing, Unit].toManaged_
             cancelers       <- Queue.bounded[Promise[Nothing, Unit]](n).toManaged(_.shutdown)
-            cancelersGuard  <- Semaphore.make(1).toManaged_
             _ <- self.foreachManaged { a =>
                   for {
-                    latch    <- Promise.make[Nothing, Unit]
                     canceler <- Promise.make[Nothing, Unit]
+                    latch    <- Promise.make[Nothing, Unit]
+                    size     <- cancelers.size
```
iravid: I liked `cancelersGuard` but simpler is of course better :-) Are there any race conditions that could happen here with the handlers in the `foldCauseM` or the finalizer on the outer stream below?

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
vasilmkd: Thanks. Couldn't have done it without your help.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Ah yes, I forgot to put it back. Sorry.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: You're right, it's safer and easier to reason about to put it back in the bracket, since now the latch protects the continuation of the foreachManaged. Making the change.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Technically, and I realized this when developing this PR, the `cancelers` queue is protected by the `permits` semaphore, if we block the continutaion of the `foreachManaged`, which we have `Managed` to do here. Sorry for the bad joke.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: But nothing is changed if you put it back, no? Did the latch not protect the continuation before?

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Again, the `latch` makes sure that an inner stream gets at least a chance to run its `bracket` operation, which means that the finalizer on the outer stream has to compete with the last `n` inner streams for the `permits`, at which point this becomes semantically the same with `flatMapPar`, so I'd think that we are pretty safe from any race conditions.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Let me sync my local repo first so I can observe the changes. I'll comment in a few minutes.

vasilmkd: Alright, I've no idea what's going on now...

vasilmkd: So `flatMapPar` must have been good, we just missed the test that proves it, so I guess that's solved now.

vasilmkd: `flatMapParSwitch` and the `cancelersGuard` trick must have added non-determinism by waking up fibers out of order...

vasilmkd:
Ready for a final review.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -684,19 +684,17 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
                   for {
                     latch <- Promise.make[Nothing, Unit]
                     innerStream = Stream
-                      .bracket(permits.acquire)(_ => permits.release)
-                      .flatMap(_ => ZStream.fromEffect(latch.succeed(())).flatMap(_ => f(a)))
+                      .bracket(permits.acquire *> latch.succeed(()))(_ => permits.release)
```
iravid: I just realized that acquiring the permits in the bracket is actually unsafe. It makes the acquisition uncancelable.

vasilmkd: Alright, what do we do now?

iravid: For a perfect solution, we need a version of `Semaphore#withPermit` that can be lifted to `ZManaged`. It's pretty straightforward to implement.

But for this PR we can just put the permit acquisition back outside the stream. I think it's relatively safe even if not bracketed.

vasilmkd:
Blocked until #1289 is resolved.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Implemented in #1293 and used here. Done.

iravid:
Looks great. Just needs a quick rebase.

vasilmkd:
I rebased but let's not merge yet.

vasilmkd:
I went over the code one more time. Ready for merging.

Issue #1204, opened 2019-07-19
Add ZStream#zipWithLatest


iravid:
Like `ZStream#zipWith`, but holds onto the last item emitted on each side so there's always an element to combine with.

contrun:
I'd like to contribute. Do you mean a function like combineLatest in RxJava https://github.com/ReactiveX/RxJava/blob/3.x/docs/Combining-Observables.md#combineLatest  ?

iravid:
Yup!

PR #1270, opened 2019-07-28
add ZStream#zipWithLatest

contrun:
closes https://github.com/zio/zio/issues/1204

### DIFF
--- a/streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -203,6 +203,10 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     zipWithIndex                $zipWithIndex
     zipWith ignore RHS          $zipWithIgnoreRhs
     zipWith prioritizes failure $zipWithPrioritizesFailure
+    zipWithLatest                     $zipWithLatest
+    zipWithLatest ignore RHS          $zipWithLatestIgnoreRhs
+    zipWithLatest prioritizes failure $zipWithLatestPrioritizesFailure
+
   """
 
   def aggregate = unsafeRun {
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1455,6 +1455,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    f: (A, B) => C
+  ): ZStream[R1, E1, C] = {
+    var latestLeft  = null.asInstanceOf[A]
+    var latestRight = null.asInstanceOf[B]
+    val first = self.take(1).zip(that.take(1)).map { x =>
+      latestLeft = x._1
+      latestRight = x._2
+      f.tupled(x)
+    }
+    first ++ self
```
iravid: Unfortunately this would execute the original stream twice. So for example, if the stream acquires a resource, it would be acquired twice, etc. Would be pretty surprising.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1455,6 +1455,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
     }
 
+  /**
+   * Zips two streams together with a specified function.
+   * Use the latest element of another stream if no element is available from another stream.
+   */
+  final def zipWithLatest[R1 <: R, E1 >: E, B, C](that: ZStream[R1, E1, B])(
+    f: (A, B) => C
+  ): ZStream[R1, E1, C] = {
+    var latestLeft  = null.asInstanceOf[A]
```
iravid: Outside of the runtime of ZIO, we try not to use bare `var`s, `null`s or casts unless absolutely necessary. Let's try and rewrite the combinator without these.

contrun: `null` can be avoided with `Some`. I still need locally mutable variables. It seems update a `Ref` of ZIO requires extra overhead as it uses synchronization primitive. This is the reason I avoided `ref`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
contrun: I thought of using the same method of zipWith. Lately, I found mergeEither combinator is good enough for zipWithLatest. I've modified my implementation. I think it is a lot simpler.

contrun: I wonder what is the correct behaviour of zipWithLatest. What should I return in the following situation?
IMG_20190731_125743
If it is (f5, s1), then my implementation is correct.

contrun:
I checked the behavior of rxjava `combineLatest`. From  Reactive Programming with RxJava P86

screenshot-2019-07-31-23-13-54

rxjava discards all earlier events. The current implementation also does this. I've updated the test cases to align with this.

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1456,4 +1460,31 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       .either
       .map(_ must_=== Left("Ouch"))
   }
+
+  private def zipWithLatest = unsafeRun {
+    // In case this test fails, try increasing delay time propotionally.
+    val s1     = Stream("s1", "s2", "s3").tap(_ => ZIO.sleep(1200.milliseconds))
```
iravid: Let's rewrite this test to use latches instead of sleeps, as sleeps will definitely cause flakiness in the CI. Let me know if you need help with that.

streams/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1456,4 +1460,39 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       .either
       .map(_ must_=== Left("Ouch"))
   }
+
+  private def zipWithLatest = unsafeRun {
+    import java.util.concurrent.CountDownLatch
+    val latch1 = new CountDownLatch(2)
+    val latch2 = new CountDownLatch(2)
+    val latch3 = new CountDownLatch(1)
+    // Merely awaiting the latch does not guarantee runCollect will collect elements in the emission order.
```
contrun: I don't know why awaiting the latch still does not guarantee runCollect will collect elements in the emission order. So I added a sleep here. Any better ideas?

regiskuckaertz:
woops, I was taking a shot at this one and didn't see this PR. There's been much changes to ZStream meanwhile, will you update your branch @contrun ? Otherwise I may as well open a new PR, let me know 🤗

contrun:
@regiskuckaertz Thank you for the ask. I tried rebase my branch to the master. It seems Fold does not exist anymore. My main computer currently experiences a problem. I am not able to sort things out for the time being. Do make a new PR if you already implemented it. I'd love to see that. You can also implement a new solution or improve this one. Do whichever suits you.

Issue #1229, opened 2019-07-24
Implement Reporter for Console JVM 


jdegoes:
This ticket involves working from @NeQuissimus's [pull request](https://github.com/zio/zio/pull/723) (in particular, [here](https://github.com/zio/zio/pull/723/files)) to implement a `Reporter` for JVM console. The `Reporter` is essentially a function that takes an executed spec and reports it.

It would loop over an executed spec and print the classic list hierarchy, with color-coded text:

```
+ Addition works fine
+ Subtraction works fine
- Value falls within range
  `52` did not satisfy `(equals(42) || (gt(5) && lt(10))`
  `52` did not satisfy `gt(5)`
```

The first failure is the failure of the whole value / whole predicate; the second is the failure of the fragment value / fragment predicate.

The predicate fragment `gt(5)` inside the whole predicate would be colored differently, to draw user's attention to the fragment that failed. This can be accomplished by looking for substrings in the rendered predicate that match the rendered fragment predicate.

Failed test labels should be red, marked with a `-`; succeeded test labels should be marked with a green `+`.

<img width="537" alt="Screen Shot 2019-07-24 at 9 54 49 AM" src="https://user-images.githubusercontent.com/156745/61799441-1f979300-adf9-11e9-8aaf-afacdc8fb0a5.png">

**Notes**

 - See `testkit` `zio/test` for the source code for this ticket

vasilmkd:
I'd like to take this if I may.

adamgfraser:
I’m happy to take this.

NeQuissimus:
I would not render the succeeded test labels green. Just leave them default and render the + in green, just like the screenshot.

adamgfraser:
Sounds good.

NeQuissimus:
Oh, and this needs to have an option to disable colours altogether. Otherwise the output is going to look like crap in some CI

jdegoes:
@adamgfraser Fantastic!

jdegoes:
It would be nice to do a summary of failures:

<img width="540" alt="Screen Shot 2019-07-27 at 2 12 38 PM" src="https://user-images.githubusercontent.com/156745/61998120-1b79a880-b079-11e9-8519-67b45eab0d96.png">

In some test frameworks, you have to scroll back to find failures; there is no summary at the end.

NeQuissimus:
If you properly integrate with SBT and `test-framework`, it should give you a summary at the very end

jdegoes:
[...]

PR #1273, opened 2019-07-28
Implement test reporter

adamgfraser:
Implements a default console reporter for new test framework as discussed in #1229. Displaying successful and failed tests, as well as highlighting of failed predicate fragments, is implemented. Options for turning display of colors on / off and summary of failures is not yet implemented.

The only change I made to the rest of the API is adding a `R` type parameter to the `Reporter`, since the `Reporter` may need to access an environment (typically `Console` but I suppose it could be something else if the report writes the results to a file or something) to report the results. This also facilitates testing the `Reporter` itself using a `MockConsole`. Changes were pretty minor but open to feedback if you think there is a better way to do this.


jdegoes:
@adamgfraser Circle CI couldn't download the OpenJDK which screwed up a bunch of builds. Can you fix conflicts and I'll re-trigger a full build?

jdegoes:
On adding `R` to `TestReporter`, just for simplicity I might do:

```scala
object TestReporter {
  def console(console: Console): TestReporter = ...
}
```

Although I don't have a big preference. But if we have `R` in `TestReporter`, we should probably add to the other things as well, `TestExecutor`, etc. to describe the set of all dependencies required to run something.

jdegoes:
giphy

adamgfraser:
Thanks John!

Issue #1263, opened 2019-07-27
Add ZManaged#memoize


jdegoes:
Constructs an inner `ZManaged` that will lazily acquire the resource, but only a single time; releasing in the outer scope only.

```scala
trait ZManaged[R, E, A] {
  def memoize: ZManaged[R, E, ZManaged[R, E, A]] = ???
}
```

/cc @iravid

jdegoes:
This could also be called `duplicate`.

jdegoes:
It could also be called `shared`.

mtsokol:
I would like to try to do it.

jdegoes:
@mtsokol Awesome, thank you! 🙏

fsvehla:
Is this actually taken for the hackathon?

mtsokol:
@fsvehla I've started it here https://github.com/zio/zio/pull/1276 and came up with idea in PR last comment according to given tip and it still waits for feedback. I'm not sure about it so feel free to look at it and give you opinion on that snippet.

fsvehla:
@mtsokol are you still around? When will you leave tomorrow?

mtsokol:
Yeah I'm I our conference room available until 22:00. Tomorrow I'm leaving 15:00, also available whole day 🙂

fsvehla:
Aah too bad we miss ourselves then, we have left for a pub and I will need to leave at 09:00 tomorrow.

PR #1276, opened 2019-07-29
Added ZManaged#shared

mtsokol:
Addresses #1263 

@jdegoes I found this task a bit harder than the previous. 
I wish to ask whether I understood it correctly. The ugly example:

```scala
case class DB() {
  var isClosed = false

  def close: DB = {
    isClosed = true
    this
  }

  def open: DB = {
    println("DB open")
    this
  }
}

val dbIO = ZIO.succeed(DB())

val program = for {

  _ <- dbIO >>= (db => putStrLn(db.isClosed.toString))                            // false

  manageDb = Managed.make(dbIO.map(_.open))(db => ZIO.effectTotal(db.close)).shared

  _ <- manageDb.use(innerManage => for {                                          // DB open
    _ <- dbIO >>= (db => putStrLn(db.isClosed.toString))                          // false
    r <- innerManage.use(db => putStrLn(db.isClosed.toString) *> UIO.succeed(db)) // false
    _ <- putStrLn(r.isClosed.toString)                                            // false
  } yield ())

  _ <- dbIO >>= (db => putStrLn(db.isClosed.toString))                            // true

} yield ()

```

So according to the issue description there’s constructed an internal `ZManage` which `release` does not release the resource. As shown above the `isClosed` is false no sooner than the outer `use` is done (so it satisfies the requirement for releasing in outer scope). Resource is released (`DB open` println) only once and lazy which addresses first requirement. 

Is it correct? Or right direction? Or have I missed the point?

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -21,7 +21,7 @@ import zio.clock.Clock
 import zio.duration.Duration
 
 /**
- * A `Reservation[-R, +E, +A]` encapsulates resource aquisition and disposal
+ * A `Reservation[-R, +E, +A]` encapsulates resource acquisition and disposal
  * without specifying when or how that resource might be used.
  *
  * See [[ZManaged#reserve]] and [[ZIO#reserve]] for details of usage.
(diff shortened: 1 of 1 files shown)


iravid:
@mtsokol take a look at how `ZIO#memoize` is implemented; the concept should be similar.

mtsokol:
@iravid Thank you for this hint! I retraced it but I’m still kind of uncertain.
I come up with this:

```scala
def memoize: ZManaged[R, E, ZManaged[R, E, A]] = {
  val innerManaged = for {
    t <- Promise.make[E, A]
    r <- Promise.make[Nothing, Unit]
    _ <- (r.await *> self.reserve.>>=(_.acquire).to(t)).fork
    s <- r.succeed(()) *> t.await
  } yield s
  self.map(_ => ZManaged.fromEffect(innerManaged))
}
```

The inner `ZManaged` is constructed via `fromEffect` so the inner release action is unit and release occurs only by outer `use` scope. The `Promise` combination follows `ZIO#memoize` pattern, but it’s still kind of blurry to me.
Is that closer to the point?

mlangc:
@mtsokol are you still on this?

mtsokol:
@mlangc no, I'm not. I've been busy recently and forgot about it. Sorry for not writing that it's free.

mlangc:
@mtsokol no problem, thanks for the quick response!

jdegoes:
This one has been superceded, so closing. @mtsokol Thank you for your work on this, I am sure it was an inspiration to the new PR!

Issue #538, opened 2019-01-26
Remove Sink.Step in favor of methods on Sink


iravid:
We can (probably) do away with Sink.Step by adding `def leftover(s: State): Chunk[A0]` and `def cont(s: State): Boolean` to Sink.

vasilmkd:
I got pretty familiar with Step when working on throttling. Should I give this a shot, if it's still something worth looking into?

iravid:
It'd definitely save allocations at the cost of making some combinators harder to write. I'd try and write a parallel implementation of `Sink`, a few sink combinators and `Stream.{run,transduce}` to see how ergonomic it is.

vasilmkd:
Continuing work on stepless sink.

vasilmkd:
https://github.com/vasilmkd/zio/tree/zsink-stepless

Whole migration done. Just worried it might be too much for a single PR.

iravid:
We'll deal with it :-) Feel free to open a PR

Issue #429, opened 2018-12-05
Add comprehensive tests to Sink


iravid:
Preferably using property checks with generated streams.

edvmorango:
I would like to work on this.

ghostdogpr:
Closed by #1277 + follow-up issues #1312 and #1313

PR #1277, opened 2019-07-29
Add unit tests for Sink combinators

vasilmkd:
I know this doesn't contain property checks with generated streams, so it partially addresses #429.

These tests will be crucial for the refactoring in #538.

Some tests exhibit surprising behavior, at least to me. Maybe they show bugs in the implementation of some of the combinators, or I may have just misunderstood their use. Anyways, I'd appreciate a second look.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -117,6 +117,13 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       step error    $optionalStepError
       extract error $optionalExtractError
 
+    takeWhile
+      happy path      $takeWhileHappyPath
+      false predicate $takeWhileFalsePredicate
+      init error      $takeWhileInitError
+      step error      $takeWhileStepError
+      extract error   $takeWhileExtractError
+
   Constructors
     Sink.foldLeft                         $foldLeft
     Sink.fold                             $fold
@@ -506,6 +513,31 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
   private def optionalExtractError = {
     val sink = extractErrorSink.optional
     unsafeRun(sinkIteration(sink, 1).map(_ must_=== None))
+  }
+
+  private def takeWhileHappyPath = {
+    val sink = ZSink.identity[Int].takeWhile[Int](_ < 5)
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== 1))
+  }
+
+  private def takeWhileFalsePredicate = {
+    val sink = ZSink.identity[Int].takeWhile[Int](_ > 5)
+    unsafeRun(sinkIteration(sink, 1).option.map(_ must_=== None))
+  }
+
+  private def takeWhileInitError = {
+    val sink = initErrorSink.takeWhile[Int](_ < 5)
+    unsafeRun(sinkIteration(sink, 1).option.map(_ must_=== None))
+  }
+
+  private def takeWhileStepError = {
+    val sink = stepErrorSink.takeWhile[Int](_ < 5)
+    unsafeRun(sinkIteration(sink, 1).option.map(_ must_=== None))
+  }
+
+  private def takeWhileExtractError = {
+    val sink = extractErrorSink.takeWhile[Int](_ < 5)
+    unsafeRun(sinkIteration(sink, 1).option.map(_ must_=== None))
   }
 
   private def foldLeft =
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -95,6 +95,12 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
       step error      $filterMStepError
       extractError    $filterMExtractError
 
+    map
+      happy path    $mapHappyPath
+      init error    $mapInitError
+      step error    $mapStepError
+      extract error $mapExtractError
+
   Constructors
     Sink.foldLeft                         $foldLeft
     Sink.fold                             $fold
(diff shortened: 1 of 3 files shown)


vasilmkd:
This PR is not ready for review yet.

iravid:
Thanks for working on this @vasilmkd, this is invaluable coverage!

iravid:
Would it be possible to test assert on the errors themselves rather than using `.option`?

vasilmkd:
@iravid Will `.either.map(_ must_=== Left(something))` do the trick?

iravid:
Sure that works too.

vasilmkd:
@iravid Please take another look, thanks.

Btw, I checked coverage, it went up from ~40% to ~80% now.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== List(1)))
+  }
+
+  private def collectAllInitError = {
+    val sink = initErrorSink.collectAll
+    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
+  }
+
+  private def collectAllStepError = {
+    // This test needs to be verified for correctness.
```
iravid: Yeah I think the implementation for `Sink#collectAllWith` is surprising. It stores the error in the first element of the tuple, but never uses it in `extract`; and if subsequent errors occur in a step, the previous errors are discarded. We should probably change it to fail eagerly in `step`.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    val sink = ZSink.identity[Int].collectAllWhile[Int, Int](_ < 10)
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== List(1)))
+  }
+
+  private def collectAllWhileFalsePredicate = {
+    // This test needs to be verified for correctness.
+    // I find this behavior to be surprising.
+    val sink = ZSink.identity[Int].collectAllWhile[Int, Int](_ < 0)
+    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left(())))
+    // Fails instead of returning empty list.
```
iravid: The implementation of `collectAllWhile` looks right - it tries to extract the underlying sink if an element that doesn't match the predicate is stepped.

The problem is with `ZSink.identity`, however; that one fails with `Unit` if an element never passed through the sink. Maybe test it with a different sink?

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    val sink = extractErrorSink orElse ZSink.succeedLazy("Hello")
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== Right("Hello")))
+  }
+
+  private def orElseExtractErrorRight = {
+    val sink = ZSink.identity[Int] orElse extractErrorSink
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== Left(1)))
+  }
+
+  private def orElseExtractErrorBoth = {
```
iravid: `orElse` could probably benefit from some more tests around leftover handling and so forth, if you have the cycles for that here

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    val sink = extractErrorSink orElse extractErrorSink
+    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
+  }
+
+  private def raceBothLeft = {
```
iravid: `raceBoth` too, re testing leftover handling

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
+  }
+
+  private def takeWhileExtractError = {
+    val sink = extractErrorSink.takeWhile[Int](_ < 5)
+    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
+  }
+
+  private def untilOutputHappyPath = {
+    // This test needs to be verified for correctness.
```
iravid: The logic in `untilOutput`'s step looks backwards. should be `if (!Step.done(s)) ...`

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    val sink = ZSink.identity[Int].untilOutput(_ % 2 == 0)
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== 1))
+  }
+
+  private def untilOutputFalsePredicate = {
+    // This test needs to be verified for correctness.
+    val sink = ZSink.identity[Int].untilOutput(_ % 2 != 0)
```
iravid: Yes, it's surprising too. Extracting that should probably result in an error (or a `None` which means the output should be `Option[B]`).

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -17,29 +17,791 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
+    val sink = ZSink.identity[Int].zipLeft(ZSink.succeedLazy("Hello"))
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== 1))
+  }
+
+  private def zipRightHappyPath = {
+    val sink = ZSink.identity[Int].zipRight(ZSink.succeedLazy("Hello"))
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== "Hello"))
+  }
+
+  private def zipWithHappyPath = {
```
iravid: Leftover handling is also an interesting property of `zip*` to test.

vasilmkd:
Since I won't be able to do development work today or next week, I'd go with merging this PR as it is, and opening follow up tickets. Is that ok?

iravid:
@vasilmkd Definitely. How about we remove the tests that exhibit wrong behavior for now?

vasilmkd:
Sure, I can do that. Give me a second.

iravid:
But keep them somewhere so we can open the tickets :-)

vasilmkd:
Done.

iravid:
@vasilmkd thanks! enjoy your vacation!

PR #1278, opened 2019-07-29
Cleanup gitignore

asakaev:


mgibowski:
Sorry, I was the one that introduced that line. I did not notice the comment with a link to StackOverflow below. 

I had doubt if it's OK to put it there, and I got mislead by the presence of `.metals`. If we ignore `.metals` why can't we ignore `.idea`? What is the difference?

To me it seems that one extra line in `.gitignore` does no harm and is a bit more friendly to new contributors, who may not have the global ignore setting configured. Otherwise they need to go through some inconvenience.

ghostdogpr:
I agree with @mgibowski, IntelliJ and Metals being probably the 2 most used ones, I don’t see the harm having them in the gitignore file.

denisgarci:
What about using https://www.gitignore.io/ to generate the .gitignore file?

asakaev:
I like @mgibowski's point.
@denisgarci, I think it's overkill :)

PR #1290, opened 2019-07-30
Use an Int instead of a Long in parallelism combinators

### LINKED ISSUE #1288: Should ZIO#foreachParN and related combinators take an Int instead of a Long?
Right now `ZIO.foreachParN` and related combinators take a `Long` parameter to limit the degree of parallelism. Should this be changed to an `Int`? I have a hard time seeing when using an `n` value greater than `Int.MaxValue` would make sense. And the existing signature can lead to warnings about implicit numeric widening if an `Int` is passed that create unnecessary friction for users.

Happy to open a PR for this if there is agreement but wanted to check first since I am relatively new here and I know there has been a tremendous amount of prior work that has gone into this.

adamgfraser:
Resolves #1288.


jdegoes:
This had been bothering me too. Nice find!

PR #1238, opened 2019-07-25
Update ZIO Test based on feedback; cleanup code by adding TestAspect

jdegoes:
Further work to the development of ZIO Test.

/cc @kaishh

PR #1292, opened 2019-07-30
Generalize and add variance to a few ZIO Test signatures

kaishh:
@jdegoes 
Following up on https://github.com/zio/zio/pull/1238#issuecomment-515411952, I think you've succeeded at making ZIO Test general enough so that a `distage-testkit` TestExecutor is implementable with rather minimal fuss, thank you very much!  - https://github.com/7mind/izumi/pull/600/files#diff-322276ed019865eea4a966fbc5f03ef7R22 although using a non-ZIO type in `T` means losing access to `TestAspect`...

I think now we need to think about

1. test discovery & SBT integration
2. online test reporting – to support tooling test results must be reported during execution, not afterwards

### DIFF
--- a/test/shared/src/main/scala/zio/test/package.scala
+++ b/test/shared/src/main/scala/zio/test/package.scala
@@ -84,7 +84,7 @@ package object test {
   /**
    * Builds a suite containing a number of other specs.
    */
-  final def suite[R, E, L](label: L)(specs: ZSpec[R, E, L]*): ZSpec[R, E, L] = Spec.suite(label, specs.toVector, None)
+  final def suite[L, T](label: L)(specs: Spec[L, T]*): Spec[L, T] = Spec.suite(label, specs.toVector, None)
 
   /**
    * Builds a spec with a single pure test.
--- a/test/shared/src/main/scala/zio/test/package.scala
+++ b/test/shared/src/main/scala/zio/test/package.scala
@@ -95,5 +95,5 @@ package object test {
   /**
    * Builds a spec with a single effectful test.
    */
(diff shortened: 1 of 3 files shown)


test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -84,7 +84,7 @@ package object test {
   /**
    * Builds a suite containing a number of other specs.
    */
-  final def suite[R, E, L](label: L)(specs: ZSpec[R, E, L]*): ZSpec[R, E, L] = Spec.suite(label, specs.toVector, None)
+  final def suite[L, T](label: L)(specs: Spec[L, T]*): Spec[L, T] = Spec.suite(label, specs.toVector, None)
```
jdegoes: If you're going to generalize `suite`, I'd do the same for `testM`, which will become identity. 

It's a little strange that `testM` just passes through the test, which is not necessarily monadic, but it will work as expected for most users.

jdegoes:
Glad to hear that! I tried my best while still keeping in mind the end-user experience.

(1) is in development now, to be tweaked / improved later; and (2), we'll have to think about how to do incremental reporting, it should be straightforward to do a stream of test results as they are produced, but we want to preserve the hierarchy, too. Right now reporting would be done at the level of every top-level suite, which is coarse.

test/shared/src/main/scala/zio/test/package.scala
kaishh: Right! Didn't notice that testM was identity through the type aliases D:

kaishh:
@jdegoes Done!

PR #1293, opened 2019-07-30
Implement Semaphore#withPermitsManaged

### LINKED ISSUE #1289: Add a version of `Semaphore#withPermit/withPermits` that returns a Managed
This would be more composable compared to `withPermit(task)`, and safer compared to bare usage of `acquire/release`.

In the implementation, it's important to put the `prepare` step in the uncancelable reservation step, and `awaitAcquire` in the cancelable step.

vasilmkd:
Resolves #1289.

cc @iravid

### DIFF
--- a/core/shared/src/main/scala/zio/Semaphore.scala
+++ b/core/shared/src/main/scala/zio/Semaphore.scala
@@ -102,8 +102,7 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
     withPermits(1)(task)
 
   /**
-   * Acquires a permit in a [[zio.ZManaged]], executes the action and releases the
-   * permit in the finalizer.
+   * Acquires a permit in a [[zio.ZManaged]] and releases the permit in the finalizer.
    */
   final def withPermitManaged[R, E]: ZManaged[R, E, Unit] =
     withPermitsManaged(1)
@@ -115,8 +114,7 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
     prepare(n).bracket(_.release)(_.awaitAcquire *> task)
 
   /**
-   * Acquires `n` permits in a [[zio.ZManaged]], executes the action and releases the
-   * permits in the finalizer.
+   * Acquires `n` permits in a [[zio.ZManaged]] and releases the permits in the finalizer.
    */
   final def withPermitsManaged[R, E](n: Long): ZManaged[R, E, Unit] =
     ZManaged(prepare(n).map(a => Reservation(a.awaitAcquire, a.release)))
--- a/core/shared/src/main/scala/zio/Semaphore.scala
+++ b/core/shared/src/main/scala/zio/Semaphore.scala
@@ -96,16 +96,30 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
   }
 
   /**
-   * Acquires a permit, executes the action and releases the permits right after.
+   * Acquires a permit, executes the action and releases the permit right after.
    */
   final def withPermit[R, E, A](task: ZIO[R, E, A]): ZIO[R, E, A] =
-    withPermits(1L)(task)
+    withPermits(1)(task)
+
+  /**
+   * Acquires a permit in a [[zio.ZManaged]], executes the action and releases the
+   * permit in the finalizer.
+   */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -96,17 +96,31 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
...
    * Acquires `n` permits, executes the action and releases the permits right after.
    */
   final def withPermits[R, E, A](n: Long)(task: ZIO[R, E, A]): ZIO[R, E, A] =
     prepare(n).bracket(_.release)(_.awaitAcquire *> task)
 
+  /**
+   * Acquires `n` permits in a [[zio.ZManaged]], executes the action and releases the
+   * permits in the finalizer.
+   */
+  final def withPermitsManaged[R, E, A](n: Long)(task: ZIO[R, E, A]): ZManaged[R, E, A] =
```
iravid: `task` is actually redundant here. This method should return a `ZManaged[R, E, Unit]` and the same semantics can be achieved by flatMapping the returned managed

core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -96,17 +96,31 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
...
+    ZManaged.make(prepare(n))(_.release).flatMap(a => (a.awaitAcquire *> task).toManaged_)
```
iravid: You should use the raw `ZManaged` constructor here, as you need to place the `prepare` into the ZIO that creates the reservation, the `awaitAcquire` into the Reservation's acquire and the `release` in the Reservation's release

core/shared/src/main/scala/zio/Semaphore.scala
vasilmkd: Awesome, will do.

core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -96,17 +96,31 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
...
    */
   final def withPermits[R, E, A](n: Long)(task: ZIO[R, E, A]): ZIO[R, E, A] =
     prepare(n).bracket(_.release)(_.awaitAcquire *> task)
 
+  /**
+   * Acquires `n` permits in a [[zio.ZManaged]], executes the action and releases the
+   * permits in the finalizer.
+   */
+  final def withPermitsManaged[R, E](n: Long): ZManaged[R, E, Unit] =
+    ZManaged(prepare(n).map(a => Reservation(a.awaitAcquire, a.release)))
```
iravid: This is really nice.

core/shared/src/main/scala/zio/Semaphore.scala
vasilmkd: Done.

core/shared/src/main/scala/zio/Semaphore.scala
vasilmkd: Thanks. And thanks for all the tips.

core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -96,17 +96,31 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
...
+    withPermitsManaged(1)
 
   /**
    * Acquires `n` permits, executes the action and releases the permits right after.
    */
   final def withPermits[R, E, A](n: Long)(task: ZIO[R, E, A]): ZIO[R, E, A] =
     prepare(n).bracket(_.release)(_.awaitAcquire *> task)
 
+  /**
+   * Acquires `n` permits in a [[zio.ZManaged]], executes the action and releases the
```
iravid: Sorry could you just update the scaladoc?

vasilmkd: You're right, sorry.

vasilmkd: Done.

vasilmkd:
Let's now try this shiny new combinator in Streams! :smile:

iravid:
🎉

Issue #1294, opened 2019-07-30
ZIO Test: Pull stream tests into new `stream-tests` top-level project


jdegoes:
As part of preparation to make ZIO Test depend on Streams tests need to be pulled out into their own top-level project, following `core` / `core-tests`.

adamgfraser:
Mind if I take this? I will think we will need to do this to fully remove the Specs2 dependency from ZIO Test since right now the tests in streams depend on the test runtime in Test.

jdegoes:
@adamgfraser That would be awesome!

jdegoes:
@adamgfraser Any ETA on this? 

Trying to introduce Stream into ZIO Test but due to circular dependencies, can't be done until this one is closed. Wondering whether to grab this or pray your evening isn't booked. 😂

adamgfraser:
@jdegoes Sorry didn't see this early. Just submitted a PR.

iravid:
Done in #1296

PR #1296, opened 2019-07-31
Remove Specs2 dependency from ZIO Test

### LINKED ISSUE #1283: ZIO Test: Remove dependency on Specs 2
ZIO Test currently depends on Specs 2. The test suite should be refactored to use ordinary Scala `assert` in a main function, to avoid having any external dependencies.

adamgfraser:
Removes Specs2 dependency from ZIO Test, refactors all tests in ZIO Test to use Scala `assert` in a main function and moves streams tests into separate project. Resolves #1283 and #1294.

We need to add a task to Circle CI to run the new main function.


jdegoes:
@adamgfraser Wow! Thanks for your work on this!!!

PR #1302, opened 2019-07-31
Add ZIO#eventually

### LINKED ISSUE #1299: Add ZIO#eventually
```scala
trait ZIO[R, E, A] { self =>
  def eventually: ZIO[R, Nothing, A] = self orElse eventually
}
```

/cc @LGLO

vasilmkd:
Resolves #1299.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/IOSpec.scala
@@ -330,13 +330,11 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
 
   def testEventually = {
     def effect(ref: Ref[Int]) =
-      ref.get.flatMap(n => if (n < 10) IO.fail("Ouch") else UIO.succeed(n))
+      ref.get.flatMap(n => if (n < 10) ref.update(_ + 1) *> IO.fail("Ouch") else UIO.succeed(n))
 
     val test = for {
-      ref   <- Ref.make(0)
-      fiber <- effect(ref).eventually.fork
-      _     <- ref.update(_ + 1).repeat(Schedule.recurs(9)).fork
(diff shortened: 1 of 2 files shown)


LGLO:
Thank you! Could you add some spec for it?

vasilmkd:
Sure. Just, where do I do that exactly?

LGLO:
Sure!
I think here: https://github.com/zio/zio/blob/2853e800906914d19059da35346787d9bf8fdc7a/core-tests/jvm/src/test/scala/zio/IOSpec.scala

vasilmkd:
Added the test code.

core-tests/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -327,6 +328,20 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
 
+  def testEventually = {
+    def effect(ref: Ref[Int]) =
+      ref.get.flatMap(n => if (n < 10) IO.fail("Ouch") else UIO.succeed(n))
+
+    val test = for {
+      ref   <- Ref.make(0)
+      fiber <- effect(ref).eventually.fork
+      _     <- ref.update(_ + 1).repeat(Schedule.recurs(9)).fork
+      n     <- fiber.join
```
LGLO: There is a possibility, that this test passes ocassionally without even using `eventually`. If forked fiber doesn't progress when line below executes completely.
I think it's possible to update `ref` (instead of `get`) in the `effect` and then you don't need undeterministic behaviour of `fork`. What do you think?

vasilmkd: Done. Sorry.

vasilmkd:
Glad I could help. Thanks for the tips.

PR #1303, opened 2019-07-31
Experiment with stepless Sink

vasilmkd:
Opening an experimental PR for #538. 

cc @iravid 

WIP.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSteplessSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSteplessSink.scala
@@ -1,0 +1,123 @@
+package zio.stream
+
+import zio._
+
+trait ZSteplessSink[-R, +E, +A0, -A, +B] { self =>
+  type State
+
+  def cont(state: State): Boolean
+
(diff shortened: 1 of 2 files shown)


jdegoes:
Does it seem to eliminate some boxing or simplify code?

iravid:
This is definitely looking simpler, and also eliminates the `Step` from the `initial` like we thought of doing. I'd say this is worth pursuing!

vasilmkd:
@jdegoes I haven't profiled this code or anything. We also have to assume that some combinators will have to keep a more complex state (off the top of my head, a tuple with a `Boolean` for continuation, maybe an `Option[A0]` for the leftovers), which will incur some allocations (so in these cases probably nothing will be saved).

@iravid I got pretty used to the `Step` API, but yes, this literally took no time to implement, and I think it exposes less internal implementation. Also, reasoning about `cont` felt more natural when implementing this PR. I used to have to constantly double check the `Step` module for implementation details.

jdegoes:
@vasilmkd @iravid Will close this for now, feel free to re-open if it survives the refactor.

PR #1306, opened 2019-07-31
Refactor MockRandom to use a pseudo-random number generator

### LINKED ISSUE #1301: ZIO Test: Refactor MockRandom to use pseudo-random number generator (PRNG)
MockRandom (in `test` project) currently has simple lists, defined by users, of "random" data. This is useful sometimes, but is not a very good default mock implementation.

We should power the methods of MockRandom with a pseudo-random number generator (which can be kept track of in the `MockRandom.Data` structure that is currently fed into `MockRandom`), and implement all methods in terms of this.

There are [many](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/Random.java) sources of inspiration, [this one](https://github.com/NICTA/rng), perhaps, the closest.

**ZIO Test won't be useful for property checking until this ticket is completed.**

adamgfraser:
Resolves #1301.

### DIFF
--- a/core/shared/src/main/scala/zio/random/Random.scala
+++ b/core/shared/src/main/scala/zio/random/Random.scala
@@ -83,7 +83,7 @@ object Random extends Serializable {
       buffer <- bufferRef.get
     } yield buffer.toList
 
-  def nextLongWith(nextLong: UIO[Long], n: Long): UIO[Long] =
+  protected[zio] def nextLongWith(nextLong: UIO[Long], n: Long): UIO[Long] =
     if (n <= 0)
       UIO.die(new IllegalArgumentException("n must be positive"))
     else {
--- a/core/shared/src/main/scala/zio/random/Random.scala
+++ b/core/shared/src/main/scala/zio/random/Random.scala
@@ -31,6 +31,7 @@ object Random extends Serializable {
     def nextInt(n: Int): ZIO[R, Nothing, Int]
     val nextInt: ZIO[R, Nothing, Int]
     val nextLong: ZIO[R, Nothing, Long]
+    def nextLong(n: Long): ZIO[R, Nothing, Long]
     val nextPrintableChar: ZIO[R, Nothing, Char]
     def nextString(length: Int): ZIO[R, Nothing, String]
     def shuffle[A](list: List[A]): ZIO[R, Nothing, List[A]]
@@ -54,6 +55,7 @@ object Random extends Serializable {
       def nextInt(n: Int): UIO[Int]               = ZIO.effectTotal(SRandom.nextInt(n))
       val nextInt: UIO[Int]                       = ZIO.effectTotal(SRandom.nextInt())
       val nextLong: UIO[Long]                     = ZIO.effectTotal(SRandom.nextLong())
+      def nextLong(n: Long): UIO[Long]            = Random.nextLongWith(nextLong, n)
       val nextPrintableChar: UIO[Char]            = ZIO.effectTotal(SRandom.nextPrintableChar())
       def nextString(length: Int): UIO[String]    = ZIO.effectTotal(SRandom.nextString(length))
       def shuffle[A](list: List[A]): UIO[List[A]] = Random.shuffleWith(nextInt, list)
@@ -80,4 +82,21 @@ object Random extends Serializable {
           }
       buffer <- bufferRef.get
     } yield buffer.toList
+
+  def nextLongWith(nextLong: UIO[Long], n: Long): UIO[Long] =
+    if (n <= 0)
+      UIO.die(new IllegalArgumentException("n must be positive"))
+    else {
+      nextLong.flatMap { r =>
+        val m = n - 1
+        if ((n & m) == 0L)
+          UIO.succeed(r & m)
+        else {
+          def loop(u: Long): UIO[Long] =
+            if (u + m - u % m < 0L) nextLong.flatMap(r => loop(r >>> 1))
+            else UIO.succeed(u % n)
+          loop(r >>> 1)
+        }
+      }
+    }
 }
--- a/test/shared/src/main/scala/zio/test/mock/MockRandom.scala
+++ b/test/shared/src/main/scala/zio/test/mock/MockRandom.scala
@@ -31,7 +31,6 @@ object MockRandom {
 
   trait Service[R] extends Random.Service[R] {
     def setSeed(seed: Long): UIO[Unit]
-    def nextLong(n: Long): UIO[Long]
   }
 
   /**
@@ -121,21 +120,7 @@ object MockRandom {
       } yield ((i1.toLong << 32) + i2)
 
     def nextLong(n: Long): UIO[Long] =
-      if (n <= 0)
-        UIO.die(new IllegalArgumentException("n must be positive"))
-      else {
-        nextLong.flatMap { r =>
-          val m = n - 1
-          if ((n & m) == 0L)
-            UIO.succeed(r & m)
-          else {
-            def loop(u: Long): UIO[Long] =
-              if (u + m - u % m < 0L) nextLong.flatMap(r => loop(r >>> 1))
-              else UIO.succeed(u % n)
-            loop(r >>> 1)
-          }
-        }
-      }
+      Random.nextLongWith(nextLong, n)
 
     val nextPrintableChar: UIO[Char] =
       nextInt(127 - 33).map(i => (i + 33).toChar)
@@ -146,24 +131,7 @@ object MockRandom {
     }
 
     def shuffle[A](list: List[A]): UIO[List[A]] =
-      for {
-        bufferRef <- Ref.make(new scala.collection.mutable.ArrayBuffer[A])
-        _         <- bufferRef.update(_ ++= list)
-        swap = (i1: Int, i2: Int) =>
-          bufferRef.update {
-            case buffer =>
-              val tmp = buffer(i1)
-              buffer(i1) = buffer(i2)
-              buffer(i2) = tmp
-              buffer
-          }
-        _ <- ZIO.traverse(list.length to 2 by -1) { n: Int =>
-              nextInt(n).flatMap { k =>
-                swap(n - 1, k)
-              }
-            }
-        buffer <- bufferRef.get
-      } yield buffer.toList
+      Random.shuffleWith(nextInt, list)
 
     def setSeed(seed: Long): UIO[Unit] =
       randomState.set {
--- a/test/shared/src/main/scala/zio/test/mock/MockRandom.scala
+++ b/test/shared/src/main/scala/zio/test/mock/MockRandom.scala
@@ -16,7 +16,11 @@
 
 package zio.test.mock
 
-import zio._
+import java.lang.StrictMath.{ log, sqrt }
+
+import scala.collection.immutable.Queue
+
+import zio.{ Chunk, Ref, UIO, ZIO }
 import zio.random.Random
 
 trait MockRandom extends Random {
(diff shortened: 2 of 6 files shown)


test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -26,146 +30,141 @@ trait MockRandom extends Random {
...
-
-    private def shiftIntWithLimit(limit: Int)(data: Data) = {
-      val next = data.integers.headOption.fold(MockRandom.defaultInteger)(identity)
-      (Math.min(limit, next), data.copy(integers = shiftLeft(data.integers)))
+        i1 <- next(26)
+        i2 <- next(27)
+      } yield ((i1.toLong << 27) + i2) / (1L << 53).toDouble
+
+    def nextBytes(length: Int): UIO[Chunk[Byte]] = {
+      //  Our RNG generates 32 bit integers so to maximize efficieny we want to
```
jdegoes: Is this code original or derived? If derived, we should include credit, copyright notice, and make sure the license is Apache 2 compliant.

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -26,146 +30,141 @@ trait MockRandom extends Random {
...
+        case Data(seed, queue) =>
+          queue.dequeueOption.fold { (Option.empty[Double], Data(seed, queue)) } {
+            case (d, queue) => (Some(d), Data(seed, queue))
+          }
+      }.flatMap {
+        case Some(nextNextGaussian) => UIO.succeed(nextNextGaussian)
+        case None =>
+          def loop: UIO[(Double, Double, Double)] =
+            nextDouble.flatMap { d1 =>
+              nextDouble.flatMap { d2 =>
```
jdegoes: `(nextDouble zip nextDouble)` can be handy for this pattern.

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -26,146 +30,141 @@ trait MockRandom extends Random {
...
+              }
+          }
+      }
 
-    private def shiftFloats(data: Data) =
-      (data.floats.headOption.fold(MockRandom.defaultFloat)(identity), data.copy(floats = shiftLeft(data.floats)))
+    val nextInt: UIO[Int] =
+      next(32)
+
+    def nextInt(n: Int): UIO[Int] =
```
jdegoes: We have no `nextLong(n: Long)`. Do you think that would be hard to add?

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
```diff
@@ -1,363 +1,84 @@
...
-    )
-    SAssert(
-      shuffleListWithThreeElementsReversed,
-      "MockRandom shuffle returns the same list in different order sometimes when list contains three elements"
-    )
-    SAssert(
-      shuffleListWithManyElementsReversed,
-      "MockRandom shuffle returns the same list in different order sometimes when list contains many elements"
-    )
+    SAssert(referentiallyTransparent, "MockRandom referential transparency")
```
jdegoes: Really great work! At some point we may want to assert the "bounded" variants generate values within the bounds.

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
adamgfraser: It is based on `java.util.Random`. It looks like that is available under the GNU General Public License version 2. Does that work? Is there an example of including that information somewhere else in the project I could work off of?

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
adamgfraser: Yes!

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
adamgfraser: Shouldn't be a problem at all.

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
adamgfraser: Yes I will add a test for this.

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -16,7 +16,11 @@
 
 package zio.test.mock
 
-import zio._
+import java.lang.StrictMath.{ log, sqrt }
```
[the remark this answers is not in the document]
adamgfraser: You're right, it is not supported. I will swap this out for `scala.math`. Random numbers generated will be almost identical. Just have to add a tolerance to the tests when we compare the Gaussians we generate to those from `scala.util.Random`.

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
jdegoes: Unfortunately, we cannot include any GPL code. BSD/MIT/Apache 2 are all good licenses and we can incorporate that code with proper attribution and copyright notices.

ZIO is mostly original but there is some derived stuff here:

<https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/Semaphore.scala>

adamgfraser: Okay, I could use some guidance on how to proceed here. Right now I'm trying to generate the same sequence of random numbers as `scala.util.Random` for a given seed, which is great for testing to make sure we don't have subtle bugs in our random number generator but means there are necessarily going to be significant similarities between our implementations. The underlying random number generator is a linear congruential generator based on Knuth's The Art of Computer Programming and the algorithm for generating random Gaussians is the polar method from Box and Muller also described in Knuth. Is it sufficient to describe that or do we need to go back to the drawing board here?

jdegoes: Hmm, tough situation. Here are the options I see:

1. A clean room implementation following `Knuth`, without reference to the JDK source code.
2. Possibly another project like [Apache RNG](http://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/random/package-summary.html) has something already pre-written (also [here](https://github.com/encog/encog-java-core/blob/master/src/main/java/org/encog/mathutil/randomize/generate/LinearCongruentialRandom.java), [here](https://www.rosettacode.org/wiki/Knuth_shuffle), [here](https://typelevel.org/cats/datatypes/state.html))

adamgfraser: What if we wrap `java.util.Random` in a `private` `Ref`? I think I can give us the same interface and hide the mutability. Otherwise I can do the second option.

jdegoes: That should work!

jdegoes: Oh wait, it will not be purely functional. Let me see if there's a hook we can use to get access into the (deterministic) random generator inside `java.util.Random`.

jdegoes: Oh, I just found this:

<https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/Random.scala>

Still looking though...

jdegoes: All right, we can [get the seed](https://gist.github.com/wookietreiber/cb9117c9fb747be3fa270242302bd073) on the JVM using reflection.

This means we could construct a `new Random(seed)`, get the next random, get the new seed, and update the seed, all atomically inside a `Ref`.

So the `Ref` approach (delegating to `java.util.Random`) would be viable for JVM.

It would not work for Scala.js, though. 

I think our best option may be basing our work off the Scala.js implementation, which is Apache 2 licensed, and then providing proper attribution and copyright notice on that file. 

What do you think?

adamgfraser: Yes I think that makes sense. The Scala.js implementation is very similar to our existing one. The main difference is they break the seed down into two 24 bit parts instead of storing the seed as a 48 bit long, but the code for methods like `nextBytes` and `nextGaussian` is the same.

jdegoes: Looking from [the history](https://github.com/scala-js/scala-js/commit/4053f3a18fc7efc6c96b117bfc9fff6c7852dec6#diff-29a10fbf90dbeb23b7016d48d5411d99) it looks to be an original implementation; and if it's licensed with Apache 2, then it cannot be derived from the GPL implementation. Of course the math is the same, no way around that, but previous rulings of courts have decided in cases where there is only one "obvious" way to produce a desired result, a novel implementation is not subject to copyright infringement.

adamgfraser: Thanks for your guidance on this!

jdegoes: Happy to help!

adamgfraser:

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
+ * Copyright 2014-2019 EPFL
```
jdegoes: 👍

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -26,146 +30,173 @@ trait MockRandom extends Random {
...
-
-    def shuffle[A](list: List[A]): UIO[List[A]] = Random.shuffleWith(nextInt, list)
-
-    def feedInts(ints: Int*): UIO[Unit] =
-      randomState.update(data => data.copy(integers = ints.toList ::: data.integers)).unit
-
-    def feedBooleans(booleans: Boolean*): UIO[Unit] =
-      randomState.update(data => data.copy(booleans = booleans.toList ::: data.booleans)).unit
+  /**
+   * Adapted from @gzmo work in Scala.js (https://github.com/scala-js/scala-js/pull/780)
```
jdegoes: 👍

test/shared/src/main/scala/zio/test/mock/MockRandom.scala
```diff
@@ -26,146 +30,173 @@ trait MockRandom extends Random {
...
+          next(31).flatMap { i =>
+            val value = i % n
+            if (i - value + (n - 1) < 0) loop
+            else UIO.succeed(value)
+          }
+        loop
+      }
 
-    private def nextRandom[T](shift: Data => (T, Data)) =
+    val nextLong: UIO[Long] =
```
jdegoes: Did you push this up into `Random.Service`?

adamgfraser: No. Shall I do that?

jdegoes: Yes, please! 🙏

adamgfraser: Done!

core/shared/src/main/scala/zio/random/Random.scala
```diff
@@ -80,4 +82,21 @@ object Random extends Serializable {
           }
       buffer <- bufferRef.get
     } yield buffer.toList
+
+  def nextLongWith(nextLong: UIO[Long], n: Long): UIO[Long] =
```
jdegoes: Instead of having this helper function, you could make this the default on the `Random.Service` trait. Otherwise it will appear in the `Random` object and users may wonder what it is for.

How does that sound?

Otherwise, looks perfect and ready to merge!

adamgfraser: Oh shoot, I forgot to make that `protected[zio]` like the `shuffleWith` helper method above. Does that work?

jdegoes: Yes, that works!

PR #1307, opened 2019-07-31
Remove unused constraints in GenIO

joroKr21:
Should be able to generate `genIdentityTrans(genSuccess[Nothing, A])`


jdegoes:
@joroKr21 Looks great, Joro, thank you!

PR #1315, opened 2019-08-01
Map aliases

ioleo:
This proposition consists of two parts:

1. Adding the `as` and `asError` operators.

Similar to `*>` and `<*` operators, sometimes we just don't care about the value in the success/error channel and wish to replace it. With these operators, we can express that more clearly.

```scala
val answerToLive1 = getStrLn map (_ => "42")
// vs
val answerToLive2 = getStrLn as "42"
```

2. Adding symbolic aliases for `map`, `mapError`, `as` and `asError`.

The symbolic aliases are more readable when chaining lots of operations.

```scala
val result1 = foo flatMap bar flatMap baz map wrap
// vs
val result2 = foo >>= bar >>= baz +> wrap
// or even

val result3 = (
  foo 
    >>= bar
    >>= baz
    +> wrap
)
```

The choice of symbols is somewhat arbitrary. Among many alternatives I propose these (`+>`, `*+>`, `#>`, `*#>`) follwing this reasoning:

* the right arrow `>` indicates the direction of the flow, similar to `*>` operator
* the `+` and `#` indicates the channel being mapped (respectively success and error)
* the `*` prefix indicates the previous value is being ignored, similar to `*>` operator 

I'm open to discuss alternatives.

### DIFF
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -61,11 +61,7 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
   /**
    * Replaces the error value with the one provided.
    */
-  final def asError[E1](e1: E1): Exit[E1, A] =
-    self match {
-      case e @ Success(_) => e
-      case Failure(c)     => halt(c.map(_ => e1))
-    }
+  final def asError[E1](e1: E1): Exit[E1, A] = mapError(_ => e1)
 
   /**
    * Maps over both the error and value type.
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -54,17 +54,26 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
   final def <*>[E1 >: E, B](that: Exit[E1, B]): Exit[E1, (A, B)] = zipWith(that)((_, _), _ ++ _)
 
   /**
+   * Replaces the success value with the one provided.
+   */
+  final def as[B](b: B): Exit[E, B] = map(_ => b)
+
+  /**
+   * Replaces the error value with the one provided.
+   */
+  final def asError[E1](e1: E1): Exit[E1, A] =
+    self match {
+      case e @ Success(_) => e
+      case Failure(c)     => halt(c.map(_ => e1))
+    }
+
+  /**
    * Maps over both the error and value type.
    */
   final def bimap[E1, A1](f: E => E1, g: A => A1): Exit[E1, A1] = mapError(f).map(g)
 
   @deprecated("use as", "1.0.0")
   final def const[B](b: B): Exit[E, B] = as(b)
-
-  /**
-   * Replaces the value with the one provided.
-   */
-  final def as[B](b: B): Exit[E, B] = map(_ => b)
 
   /**
    * Flat maps over the value type.
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -58,10 +58,13 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
    */
   final def bimap[E1, A1](f: E => E1, g: A => A1): Exit[E1, A1] = mapError(f).map(g)
 
+  @deprecated("use as", "1.0.0")
+  final def const[B](b: B): Exit[E, B] = as(b)
+
   /**
    * Replaces the value with the one provided.
    */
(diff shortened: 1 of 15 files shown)


jdegoes:
I am not opposed to `as`. Note that it serves the same function as `const`, but `const` is inherited from Haskell; `as` reads better with OO syntax.

To make this change, however, we would need:

1. Deprecate _all_ the `const` methods
2. Add `asUnit`, and deprecate `unit`

Anyone else have thoughts on this?

jdegoes:
_Clarification_: I am opposed to the symbolic aliases. Developers outside FP complain loudly about these, and unlike, e.g., `>>=`, there is no precedent for new operators.

ghostdogpr:
I quite liked `.unit`. `.asUnit` is as long as than `.as(())`.

adamgfraser:
I like `as`. I agree that `asUnit` seems a little wordy for a common and simple method, maybe `skip` or just leave as is? I am thumbs down on the symbolic aliases for the reasons discussed above. If you want to write in that style you could could always do `fa *> succeed(b)` or `fa *> fail(e)`.

iravid:
👍 on `as`
👎 on `asUnit`
👎 on additional symbolics, especially ones that aren't established

jdegoes:
All right, so it looks like we're:

 - 👍 to `as`, but we need deprecations for `const`, and we need `as` added _everywhere_ `const` is used (and that is a _lot_ of places!)
 - 👎 to `asUnit`, leaving `as(())` with the name `unit` for brevity

ioleo:
Alright, thanks for the input! I will adjust the PR as soon as I have some spare time :)

ioleo:
Also deprecated `ZSchedule.const` and added `ZSchedule.as` for consistency (principle of least surprise).

ioleo:
@jdegoes Ready to review. `ci/circleci: test_212_jdk8_jvm` failure is in `BlockingSpec`, I believe its a flaky test and the failure is unrelated to this PR.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -470,6 +467,11 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    */
   def map[B](f: A => B): ZIO[R, E, B] = new ZIO.FlatMap(self, new ZIO.MapFn(f))
 
+  /**
+   * Returns an effect whose success is replaced by the specified `b` value.
+   */
+  final def as[B](b: B): ZIO[R, E, B] = self.flatMap(new ZIO.ConstFn(() => b))
```
ghostdogpr: Can you move `as` so that the alphabetical sorting is maintained?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -478,6 +480,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def mapError[E2](f: E => E2): ZIO[R, E2, A] =
     self.foldCauseM(new ZIO.MapErrorFn(f), new ZIO.SucceedFn(f))
 
+  /**
+   * Returns an effect whose error channel is replaced by the specified `e2` error.
+   * This can be used to lift a "smaller" error into a "larger" error.
+   */
+  final def asError[E2](e2: E2): ZIO[R, E2, A] = mapError(_ => e2)
```
ghostdogpr: This one too

ioleo:
@ghostdogpr Added `as` and `asError` (where `mapError` was defined) to `ZManaged`, `Assertion`, `Exit`, `Fiber`, `FunctionIO`, `Runtime` and `STM`.

core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -54,14 +54,26 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
...
-  final def bimap[E1, A1](f: E => E1, g: A => A1): Exit[E1, A1] = mapError(f).map(g)
+  final def as[B](b: B): Exit[E, B] = map(_ => b)
 
   /**
-   * Replaces the value with the one provided.
+   * Replaces the error value with the one provided.
    */
-  final def const[B](b: B): Exit[E, B] = map(_ => b)
+  final def asError[E1](e1: E1): Exit[E1, A] =
+    self match {
```
ghostdogpr: Any reason not to use `mapError` here?

ioleo: No, quite the opposite, good catch :+1:

ioleo:
@ghostdogpr Ready

ghostdogpr:
@ioleo thanks for the suggestion & implementation!

PR #1316, opened 2019-08-01
Proposal remove R from Service

wi101:
The packages that extend the Service are helpers and optional 
The proposal is to remove the `extends` there and we can access the Environment and be free in naming those helper functions, and basically each Service for a given environment doesn't require any external environment it's always `Service[Any]`
If you like this proposal we can make accessM take `R => IO[E, A]` :-)


wi101:
(Well, I have to fix all the errors🙃)
And what about making the Console, Random..an interface that contains directly the functions that we want to interact via ZIO environment?

jdegoes:
> The proposal is to remove the extends there and we can access the Environment and be free in naming those helper functions

"Be free" means that when a developer adds a function to, e.g. `Random.Service`, they will forget to add it to `random` object; so the two will become out of sync. 

Now if a user adds a method to any service, the compiler reminds them to add a corresponding helper method to the package object; which is quite handy.

However, I have seen how `R` is very confusing for developers new to ZIO, I would be in favor of removing it 👍 , even though it will make maintaining the code trickier.

(Personally, I would always use `R` in private code bases; it's just better coding style.)

> And what about making the Console, Random..an interface that contains directly the functions that we want to interact via ZIO environment?

👎 This would mean a single gigantic global namespace that includes all functions across all modules.

wi101:
@jdegoes Thanks for your answer, 
If we want to leave the package helpers in the code base we need to keep the `R` otherwise it would be tricky.

I think, you're right this would depend to the choice of the developer,
I will close the PR

jdegoes:
> If we want to leave the package helpers in the code base we need to keep the R otherwise it would be tricky.

Did some people complain about package helpers? 

I only use the package helpers. I never use the other style (`ZIO.accessM[Random] { env => import env.random._ ... }`), but maybe I am rare. 😄 

I'm still in favor of removing `R` even though it will make maintaining the code trickier. It will make ZIO less confusing to people and more experienced developers can still use the pattern in their code bases if they want.

It would be nice to hear feedback from others on this one.

wi101:
No one complained, but I came up with this idea, when someone asked me yesterday about why Service has a type parameter [Any] why do we need R in the trait Service and I told him it's a trick for the helper when we extend it to be able to define those functions there and access the R and it wasn't clear for him, 
Also I see `ZIO[R, E, A]` as `R => IO[E, A] `  in which the environment R has IO functions that we can access, it would be easier to see that definition in the code.

adamgfraser:
I've had some experience with the package helpers from working on ZIO Test. I am a big fan of them. They let you write code like this:

```scala
for {
  _      <- console.putStrLn("First line")
  _      <- console.putStrLn("Second line")
  output <- MockConsole.output
} yield output == Vector("First line\n", "Second line\n")
```

It makes it feel very much like you are just calling a normal method on an object like `scala.Console.println`. You actually have to think about it to realize that you are describing accessing some environment that you are threading through your code and can actually provide different implementations for. Very seamless.

LGLO:
I'm one of people for whom existence of `[Any]` in `extends ABC.Service[Any]` param for services wasn't clear.
I've implemented one and now I understand it.
I think that chapter in docs for "Module pattern" and "R" param (this is covered in Testing chapter to some extent) which explains why `ABC.Service[R]` over`ABC.Service` would do the trick.
If I ever manage to publish something, it will use `[R]`.

wi101:
Nice, I think there is no need to change anything in the code base, this would be different from developer to another.

I will close this PR

jdegoes:
Personally I'm still 👍 on removing `R` just to reduce confusion, but if there's consensus to leave it, I'm fine with that too. Either way!

Issue #1218, opened 2019-07-23
FiberFailure stack trace suggests wrong location


jdegoes:
Whenever a `FiberFailure` is constructed, it builds a stack trace, which is printed out by some third-party libraries (rather than, say, its `toString` method).

To avoid this, we should probably disable the stack trace, or delegate to the stack trace of the first / primary defect.

/cc @kaishh ???

kaishh:
Makes sense! I can do this if no one else wants to take it first 😉

mgibowski:
I am not sure I understand well the issue.

@jdegoes Do you mean something like that?

```scala
import scala.{ Console => SConsole }

final case class FiberFailure(cause: Cause[Any]) extends Throwable {

  override final def getMessage: String = cause.toString

  override def printStackTrace(): Unit = {
    SConsole.withErr(SConsole.err)(cause.prettyPrint)
    () // because of "discarded non-Unit value"
  }

  override def printStackTrace(s: PrintStream): Unit = {
    SConsole.withErr(s)(cause.prettyPrint)
    ()
  }


  override def printStackTrace(w: PrintWriter): Unit = {
    w.println(cause.prettyPrint)
    ()
  }

}
```

jdegoes:
@mgibowski Yes, and possibly even overriding `getStackTrace` and filling it with execution trace information. It's not a perfect mapping but I can see some frameworks just dump out the stack trace and it's not remotely helpful compared to the `Cause` with tracing.

kaishh:
@mgibowski 
`Throwable` has a constructor parameter to disable stacktrace capture, you may use that instead of overriding methods.

@jdegoes 
Have you encountered frameworks dumping just the trace and ommiting `.getMessage/.printStackTrace` - which should both print ZIO Trace? There's a mapping method `ZTrace.ZTraceElement.toStacktraceElement` that can convert a ZTrace to a Java stacktrace, but it may be unnecessary as long `message` is not omitted by the frameworks.

jdegoes:
@kaishh No, just `.printStackTrace`. Maybe we should just disable the stack trace capture.

mgibowski:
Ok, I opened the PR 👆

PR #1318, opened 2019-08-01
Create `FiberFailure` without stack trace (#1218)

mgibowski:
Addresses #1218 

`FiberFailure` now extends `Throwable` calling constructor that does not fill the stack trace.

### DIFF
--- a/core/shared/src/main/scala/zio/FiberFailure.scala
+++ b/core/shared/src/main/scala/zio/FiberFailure.scala
@@ -24,6 +24,4 @@ package zio
  * This class is used to wrap ZIO failures into something that can be thrown,
  * to better integrate with Scala exception handling.
  */
-final case class FiberFailure(cause: Cause[Any]) extends Throwable {
-  override final def getMessage: String = cause.prettyPrint
-}
+final case class FiberFailure(cause: Cause[Any]) extends Throwable(cause.toString, null, true, false)
--- a/core/shared/src/main/scala/zio/FiberFailure.scala
+++ b/core/shared/src/main/scala/zio/FiberFailure.scala
@@ -24,4 +24,6 @@ package zio
  * This class is used to wrap ZIO failures into something that can be thrown,
  * to better integrate with Scala exception handling.
  */
-final case class FiberFailure(cause: Cause[Any]) extends Throwable(cause.toString, null, true, false)
+final case class FiberFailure(cause: Cause[Any]) extends Throwable(null, null, true, false) {
+  override final def getMessage: String = cause.prettyPrint
+}


core/shared/src/main/scala/zio/FiberFailure.scala
```diff
@@ -24,6 +24,4 @@ package zio
  * This class is used to wrap ZIO failures into something that can be thrown,
  * to better integrate with Scala exception handling.
  */
-final case class FiberFailure(cause: Cause[Any]) extends Throwable {
-  override final def getMessage: String = cause.prettyPrint
-}
+final case class FiberFailure(cause: Cause[Any]) extends Throwable(cause.toString, null, true, false)
```
kaishh: Printing `cause` is rather expensive; we may want to pass `null` here and return the `getMessage` override.

mgibowski: Oh I did not consider that, thanx. Then, what should the message, just "FiberFailure"?

kaishh: `cause.prettyPrint`

kaishh: This will return a formatted ZIO Trace whenever user/libraries call `.getMessage` or `.printStackTrace` on a FiberFailure. (It will pretty-print all over for each call, but I think that's still better than using a val, since traces also tend to have a huge printed representation that you may not want to always keep in memory)

mgibowski: @kaishh  OK, I just made the change according to your instruction (thanks!).

Although looking back at the original description from @jdegoes I have a little doubt:

> Whenever a FiberFailure is constructed, it builds a stack trace, which is printed out by some third-party libraries (rather than, say, its toString method).

Now, the stack trace is no longer built in the constructor, however third-party libraries will still pretty-print the formatted ZIO Trace instead of something simple. So maybe it's a bit different from what John expected?

kaishh: Err, so the stack trace in question refers to the Native JVM Stack Trace – FiberFailure, being a Throwable, has one, but in FiberFailure's case it's completely useless, because a FiberFailure is always thrown from the same place in the RTS! So, when printing a ZIO failure you get something like:

```
Fiber n failed 
  future continuation blah
  future continuation bleh
  future continuation bluh

Fiber spawned by...
  at zio.FiberContext.evaluate
  at zio.FiberContext.evaluate
  at zio.FiberContext.evaluate
```

Where the last `at`'s are from FiberFailure's own JVM stack trace, not the ZIO Trace. They're generally always the same and hold no useful information unless you're `unsafeRun`ning constantly – but even then ZIO trace includes the location of the first `unsafeRun` and you could wrap `unsafeRun` with a rethrow if you need a more detailed stack...

If we don't print the ZIO Trace, then it's completely pointless to sample it in the first place – its only usecase is to be examined by a human when an unexpected failure happens to assist in debugging :)

Hmm, now that I think of it, a FiberFailure stack trace is not always useless – when using ZIO across the entire application / using zio.App – it is; But newcomers might find it useful (although one could argue that the location of unsafeRun is good enough), @jdegoes  thoughts?

kaishh: @mgibowski

jdegoes: A `FiberFailure` is almost always constructed by ZIO when no one has handled the error. Being so deep inside the guts of ZIO and having no relation to user code, I don't think it's that useful. It's distracting from the true cause of the problem.

Now, as long as `printStackTrace` is not printing out the stack trace, but rather the execution trace... maybe it's not so bad to leave it attached (`getStackTraceElements`?).

kaishh: Why not. In that case `printStackTrace` should be overriden to not print the synthetic `StackTraceElement`s

Issue #1284, opened 2019-07-30
ZIO Test: Develop example suites


jdegoes:
Example test suites will show people how to use ZIO Test, and provide a "compile time test" ensuring that the syntax is supported.

We should have 3 short specs, and 3 medium specs, in varying complexity levels.

### Short Simple

10 line of code test, which only tests simple truths (no effects).

### Short Moderate

10 line of code test, which tests effectful operations.

### Short Complex

15 line of code test, which tests effectful operations with aspects and resource management.

### Medium Simple

50 line of code test, which only tests simple truths (no effects).

### Medium Moderate

50 line of code test, which tests effectful operations.

### Medium Complex

65 line of code test, which tests effectful operations with aspects and resource management.

edvmorango:
I would like to give it a shot.

jdegoes:
@edvmorango Thank you!

jdegoes:
ScalaTest has some very nice examples on their website. They're not quite realistic (most of them), but they give a feel for the library.

Specs2 has a number, front and center on the home page.

edvmorango:
Hey, @jdegoes I'll start work in it today, but I have some doubts:

1 - Where should I put these new specs (I'm thinking about core-tests/shared)? 

2 - By `aspect` did you mean PBT?

3 - Do you think to create two different specs, one with realistic tests and one more similar to ScalaTest/Specs2 a good idea?

jdegoes:
1. I'd do a new top-level project `examples` that we can put ZIO examples, ZIO test examples, etc.
2. By Aspect, I mean `TestAspect`.
3. Yes, that sounds like a good idea to me.

edvmorango:
Hey, @jdegoes  I'm working at this issue, after creating the first suit:

```
   suite("Operations")(
     test("(1 + 1) must be 2") {
       assert[Int](1 + 1, Predicate.equals(2))
     }
      ,
     test("(10 - 5) must be 5") {
       assert[Int](10 - 5, Predicate.equals(5))
     }
   )
```

I executed using the `DefaultRunnableSpec`. When I run the tests individually they work perfectly, but when I execute the suit none of them run. I modified the `run` function inside `TestRunner`  from parallel execution strategy to sequential:

```final def run(spec: Spec[L, T]): UIO[ExecutedSpec[L]] =
    executor.execute(spec, ExecutionStrategy.Sequential).flatMap { results =>
      reporter.report(results) *> ZIO.succeed(results)
 }
```

[...]

PR #1328, opened 2019-08-02
ZIO Test: Fix bug in DefaultTestReporter

adamgfraser:
`DefaultTestReporter` normally prints one line with the whole failing predicate and a second line with the failing predicate fragment. This results in two identical lines being printed when the predicate is "simple" (i.e. it is not composed of other predicates). This PR fixes that by adding a check for whether the whole predicate is identical to the fragment and only printing one line in that case.

Identified by @ghostdogpr in #1284.

Copying @jdegoes.


ghostdogpr:
Just tested this, nice 👍

PR #1329, opened 2019-08-02
Test Examples

edvmorango:
#1284


zio-examples/jvm/src/test/scala/examples/PredicateExampleSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+object PredicateExampleSpec {
+
+  val spec1: ZSpec[Any, Nothing, String] =  test("(1 + 1) must be 2") {
+    assert[Int](1 + 1, Predicate.equals(2))
+  }
+
+  val spec2: Spec[String, ZTest[Any, Nothing]] = suite("That")(spec1)
+
+  val spec: Spec[String, ZTest[Any, Nothing]] = suite("Operations")(
+      test("(1 + 1) must be 2") {
```
jdegoes: Instead of the labels repeating the tests, I'd make them more high-level, e.g., this one could be called "addition operator", because that's what it is testing.

zio-examples/jvm/src/test/scala/examples/PredicateExampleSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+        assert[Int](1 + 1, Predicate.equals(2))
```
jdegoes: Do you need the `[Int]` type annotation here? Best to omit, I think, for increased concision.

zio-examples/jvm/src/test/scala/examples/PredicateExampleSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+     test("(2 * 5) must be 10") {
+       assert[Int](1 + 1, Predicate.equals(10))
+     },
+     test("(25 / 5) must be 5") {
+       assert[Int](25 / 5, Predicate.equals(5))
+     }
+   )
+
+})
+
```
jdegoes: The build is failing because the code is not formatted. To format the code, merely run `fmt` from SBT.

zio-examples/jvm/src/test/scala/examples/PredicateExampleSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+     },
+     test("(2 * 5) must be 10") {
+       assert[Int](1 + 1, Predicate.equals(10))
+     },
+     test("(25 / 5) must be 5") {
+       assert[Int](25 / 5, Predicate.equals(5))
+     }
+   )
+
+})
```
ghostdogpr: The { } can be avoided

jdegoes:
@edvmorango Is this ready for review?

edvmorango:
@jdegoes, I started to create a suite for effectful tests, I've created examples for the most basic tests (Success and failure), but right now I'm struggling to extend the environment of the `DefaultRunnableSpec`.

jdegoes:
@edvmorango We have an `examples` project now. Can you put the examples in there and follow its pattern for test examples? Then should be good to merge.

Thanks for your work on this!

build.sbt
```diff
@@ -50,6 +50,8 @@ lazy val root = project
     coreTestsJVM,
     coreTestsJS,
     docs,
+    examplesJS,
```
ghostdogpr: I think we don't want to include them here (see the note next to their declaration: `Examples sub-project that is not included in the root project.`), otherwise they'll get published.

edvmorango: @ghostdogpr, I fixed it, I would like to create an example suite of a real problem (I'm thinking in an Http API embedded into an Environment),  and I didn't create yet an example suite for `TestAspect`, can we postpone the merge?

ghostdogpr: 👌

adamgfraser:
@edvmorango What do you think about adding an example that uses the new property based testing functionality?

edvmorango:
@adamgfraser, sure, as soon I finish the suite for `TestAspect`, I will work on it.

edvmorango:
Hey guys, I'm trying to finish the `TestAspect` suite, but right now I'm having problems with the`eventually / flaky`, I tried to print and debug the retry, but when the test fails, the retries behavior doesn 't occur.

```
 val eventually: TestAspectPoly =
    new TestAspect.PerTest[Nothing, Any, Nothing, Any] {
      def perTest[R >: Nothing <: Any, E >: Nothing <: Any](
        test: ZIO[R, E, TestResult]
      ): ZIO[R, Nothing, TestResult] = {
        lazy val untilSuccess: ZIO[R, Nothing, TestResult] =
          test.foldM(_ => ZIO.effectTotal(println("failure")) *> untilSuccess, ZIO.succeed(_))

        untilSuccess
      }
    }
```

jdegoes:
@edvmorango How did the test fail? Using `ZIO.fail` or by throwing somewhere it should not?

edvmorango:
@jdegoes, I'm not throwing an exception or using `ZIO.fail`, the effect evaluation sometimes doesn't result in a valid input to the assertion, so I'm expecting the `eventually/flaky` aspect to retry until the effect evaluation result, be a valid input to the assertion predicate. It's the same idea, of the v1.0.0-RC11 release example (without the aspect composition):
```
    (flaky >>> timeout(10.seconds)) {
      testM("flaky random.nextDouble") {
        assertM(nextDouble, isGreaterThan(0.5))
      }
    }
````

**Comment by jdegoes** [comment]
@edvmorango Ah, you are right. We should fix `eventually`:

```scala
 val eventually: TestAspectPoly =
    new TestAspect.PerTest[Nothing, Any, Nothing, Any] {
      def perTest[R >: Nothing <: Any, E >: Nothing <: Any](
        test: ZIO[R, E, TestResult]
      ): ZIO[R, Nothing, TestResult] = {
        lazy val untilSuccess: ZIO[R, Nothing, TestResult] =
          test.foldM(_ => ZIO.effectTotal(println("failure")) *> untilSuccess, 
            // This is the crucial line to modify: if the test did not succeed, keep trying.
            testResult => if (testResult.succeeds) ZIO.succeed(testResult) else untilSuccess
          )

        untilSuccess
      }
    }
```

It occurs to me that `retry` has a similar "problem": it only considers typed failures (not test failures) to require retrying. If you could fix that too, it would be great!

edvmorango:
@jdegoes,  I think when you said `retry` (related do `Scheduling`)  you meant `flaky` (just an alias for `eventually`), this test is working now, but I have found an unexpected behavior when testing `timeout` when I run the code below using the `DefaultRunnableSpec` he always succeeds.

```
        timeout(3.seconds) {
          testM("Timeout test") {
            val blockingEffect = ZIO
              .accessM[Blocking with Console](
                env =>
                  putStrLn("Starting blocking point") *>
                    env.blocking.effectBlocking { Thread.sleep(5000); "value" }
                    <* putStrLn(
                      "Finished blocking point"
                  )
              )
              .either
              .provide(...)
            assertM(blockingEffect, Predicate.isRight(Predicate.equals("value")))
          }
        }
```


The definition of time `timeout` aspect seems correct, so I tried to understand `timeout` better and eventually I executed a "similar" effect using the `DefaultRunTime` and gotta the expected behavior: 

```
 def timeout(duration: Duration): TestAspect[Nothing, Clock, Nothing, Any] =
    new TestAspect.PerTest[Nothing, Clock, Nothing, Any] {
      def perTest[R >: Nothing <: Clock, E >: Nothing <: Any](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] =
        test.timeout(duration).map {
          case None =>
            Assertion
              .failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
          case Some(v) => v
        }
    }
```

```
object SomeApp extends App {
  override def run(args: List[String]): ZIO[SomeApp.Environment, Nothing, Int] = {
    val blockingEffect = ZIO
      .accessM[Blocking with Console](
        env =>
          putStrLn("Starting blocking point") *>
            env.blocking.effectBlocking { Thread.sleep(5000); "value" }
            <* putStrLn(
              "Finished blocking point"
          )
      )
      .either
      .provide(...)

    blockingEffect
      .timeout(1.seconds)
      .flatMap(r => ZIO.effectTotal(println(r))) // None
      .map(_ => 1)
  }
}
```

I think the reason for this behavior is related to `MockClock` and `MockEnvironemnt` but I haven't had success fixing it yet.

```
trait MockClock extends Clock with Scheduler {
  val clock: MockClock.Service[Any]
  val scheduler: MockClock.Service[Any] 
}
```
```
case class MockEnvironment(
  clock: MockClock.Mock,
  console: MockConsole.Mock,
  random: MockRandom.Mock,
  scheduler: MockClock.Mock,
  system: MockSystem.Mock,
  blocking: Blocking.Service[Any]
) extends Blocking
    with MockClock
    with MockConsole
    with MockRandom
    with Scheduler
    with MockSystem

object MockEnvironment {

  val Value: Managed[Nothing, MockEnvironment] =
    Managed.fromEffect {
      for {
        clock    <- MockClock.makeMock(MockClock.DefaultData)
        console  <- MockConsole.makeMock(MockConsole.DefaultData)
        random   <- MockRandom.makeMock(MockRandom.DefaultData)
        system   <- MockSystem.makeMock(MockSystem.DefaultData)
        blocking = Blocking.Live.blocking
      } yield new MockEnvironment(clock, console, random, clock, system, blocking)
    }
}
```
Could you give me some hint, please?

adamgfraser:
@edvmorango I can try to help with the `MockClock`. A few things to keep in mind:

- `MockClock` mocks the testing of time by scheduling tasks at a certain "time" in the future and then running those tasks when you set the time to be after that time with the `adjust` or `setTime` methods. If you don't change the time yourself time won't pass, but by the same token you can mock any amount of time passing instantaneously.
- `MockClock` can only mock time related methods provided by `ZIO`. If you call `Thread.sleep(5000)` that thread is still sleeping for 5 seconds.
- Make sure you're providing the `MockClock` environment at the right level of your application. If some of your time related code is using a `MockClock` and some is using `Live` that could lead to issues.

In your case, the following code works as expected for me, returning `None` because the block effect times out, and does not require the passage of any actual time for the test. Notice that I had to fork the blocking effect so that I could set the time with `MockClock` to trigger the timeout.

Does this help?

I will also open an issue to improve the documentation of `MockClock`.

```scala
import zio._
import zio.blocking._
import zio.console._
import zio.duration._
import zio.test.mock.MockClock

object SomeApp extends App {
  override def run(args: List[String]): ZIO[SomeApp.Environment, Nothing, Int] = {
    val blockingEffect = ZIO
      .accessM[Blocking with Console](
        env =>
          putStrLn("Starting blocking point") *>
            env.blocking.effectBlocking { Thread.sleep(5000); "value" }
            <* putStrLn(
              "Finished blocking point"
            )
      )
      .either

    val io = for {
      fiber <- blockingEffect.timeout(1.second).fork
      _     <- MockClock.setTime(2.seconds)
      r     <- fiber.await
      _     <- ZIO.effectTotal(println(r))
    } yield 1

    io.provideManaged(zio.test.mock.mockEnvironmentManaged)
  }
}
```

edvmorango:
@adamfgraser Thanks for the help! Your explanation was very enlightening. The idea behind `MockClock` is very useful for flow control, I will create one test to demonstrate a possible use (maybe a test suite specific to `MockEnvironment`). But I think the idea of the  `timeout` aspect is the same as `Await.result`, I can make a `race` using the `Blocking` with `Thread.sleep` but I'm sure there are better solutions. If we create the function `sleep(effect: ZIO[R,E,A], duration: TimeUnit) ` for `MockClock` should be enough, what you think?

adamgfraser:
@edvmorango If you want to delay an effect you would normally just do something like:

```scala
def delay[R <: Clock, E, A](zio: ZIO[R, E, A])(duration: Duration): ZIO[R, E, A] =
  sleep(duration) *> zio
```

Basically do the sleep and when that is done do the effect. Note that this is already implemented as a method on `ZIO`. All the more complex time related methods are built on top of `sleep`, which is why it is (hopefully! :smiley:) helpful to use `MockClock`, because then you get to control the passage of time for all these more sophisticated combinators in the same way. We just merged #1447 last night that has more documentation for `MockClock`. From that, here is an example of how you could test a `timeout`:

```scala
import zio.ZIO
import zio.duration._
import zio.test.mock.MockClock

for {
  mockClock <- MockClock.makeMock(MockClock.DefaultData)
  fiber <- ZIO.sleep(5.minutes).timeout(1.minute).fork
  _ <- mockClock.setTime(1.minute)
  result <- fiber.join
} yield result == None
```

There are a bunch more example in the test suite for `MockClock` you could look at as well. Does that get you what you need?

edvmorango:
@adamgfraser The definition of `delay` through `sleep(duration) *> zio`  is not what I want  because the `sleep` behavior that I need relies on the real `ZSchedule`, the idea of `timeout` as *aspect*, isn't to test possible failures/fallbacks related to `ZIO.timeout`, so I don't need to work with `MockClock.makeMock`, I just need to make an arbitrary test fail when its evaluation takes longer than was limited through `TestAspect.timeout`, e.g:

```
timeout(3.seconds) {
    testM("slow test") {
        val someEffectNotRelatedWithClock = ...
        assertM(someEffectNotRelatedWithClock, isTrue)
    }
}
```

To make the current implementation of `TestAspect.timeout` works, I need to make a race between a `sleep` and the `effect` being evaluated by the test, the evaluation of the test uses the `MockEnvironment`, so I was proposing to create a function inside `MockClock` who does a real `sleep` to make this race possible. 

```
  def timeout(duration: Duration): TestAspect[Nothing, Clock, Nothing, Any] =
    new TestAspect.PerTest[Nothing, Clock, Nothing, Any] {
      def perTest[R >: Nothing <: Clock, E >: Nothing <: Any](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] =
         MockClock
           .realSleep(duration)
           .map(_ => 
             Assertion.failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
           )
           .race(test)
    }
```

adamgfraser:
@edvmorango So could you do something like?

```scala
timeout(3.seconds) {
    testM("slow test") {
        ZIO.never *> assertM(someEffectNotRelatedWithClock, isTrue)
    }
}
```

`assertM` is just a `ZIO` that returns a test result so you can use all the normal combinators on it.

edvmorango:
@adamgfraser Yes, through the implementation below, It's possible but I need the `Clock.Live`.

```
def timeout(duration: Duration): TestAspect[Nothing, Clock, Nothing, Any] =
    new TestAspect.PerTest[Nothing, Clock, Nothing, Any] {
      def perTest[R >: Nothing <: Clock, E >: Nothing <: Any](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] =
        ZIO
          .sleep(duration)
          .map(
            _ =>
              Assertion
                .failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
          )
          .race(test)
    }
```

These three tests fail as expected:

```
timeout(3.seconds) {
    testM("Timeout test") {
        assertM(ZIO.never, equalTo("value"))
     }
}
```
```
timeout(3.seconds) {
    testM("Timeout test") {
        val blockingEffect = ZIO
              .accessM[Blocking](_.blocking.effectBlocking { Thread.sleep(5000); "value" })
              .either
              .provide(Blocking.Live) 

        ZIO.never *> assertM(blockingEffect, isRight(equalTo("value")))
    }
}
```

```
timeout(3.seconds) {
    testM("Timeout test") {
        val blockingEffect = ZIO
              .accessM[Blocking](_.blocking.effectBlocking { Thread.sleep(5000); "value" })
              .either
              .provide(Blocking.Live) 
            assertM(blockingEffect, isRight(equalTo("value")))
     }
}
```

adamgfraser:
This will get the timeout working for you with `MockClock`:

```scala
import zio.test.mock.MockClock
val test = timeout(3.seconds) {
  testM("Timeout test") {
    MockClock.adjust(3.seconds) *> ZIO.never *> assertM(ZIO.succeed(Right("value")), isRight(equalTo("value")))
  }
}
test.mapTest(test => test.provideSomeM(MockClock.make(MockClock.DefaultData)))
```

jdegoes:
@adamgfraser @edvmorango 

We are in a position where a single environment is not sufficient:

1. For application code, there is no need to actually sleep, do console IO, etc. (in general)
2. For test code, timeouts need to actually operate using clock time; timed tests need to actually measure system time; etc.

This raises the question: how does a test writer specify which environment to use? The mock one or the real one? I think I have an idea, will write it up separately, until then we should keep in mind that the current design has this issue.

edvmorango:
@adamgfraser  Sorry for the delay, I will try to finish in the next days, I will finish the bank example and add more examples to Gen.

adamgfraser:
@edvmorango No worries, that would be great!

jdegoes:
@edvmorango Just checking to see if this one is blocking on anything. Please let us know if you need any help or you need someone to take over these great test examples!

/cc @adamgfraser

edvmorango:
@adamgfraser @jdegoes Sorry for the delay, in the last few weeks I had some personal problems, but right now I'm ready to finish these examples.

adamgfraser:
@edvmorango No problem. That definitely comes first. Hope everything is okay now. Glad to have you back in the mix.

edvmorango:
@adamgfraser Thank you, I think I found an unexpected behavior at `TestAspect.sequential`

I'm using a "Ref" as a managed resource to test `around` aspect,  this is the idea: 
```
 val zRef: Ref[RefState] = runtime.unsafeRun(Ref.make(RefState(Created, Nil)))
 sequential {
      suite("Sequential examples") {         
      
	around(zRef.update(_.newState(Ready)), zRef.update(_.newState(Done))) {
        	testM("Around test ") {
             	val effect = zRef.update(_.newState(Happening))
                assertM(effect, equalTo(RefState(Happening, List(Ready, Created))))
              }
            }
        
        identity {
              testM("Identity test (around validation)") {
                val effect = zRef.get
                assertM(effect, equalTo(RefState(Done, List(Happening, Ready, Created))))
              }
        }
 }
```

I need to execute these tests sequentially to avoid a race condition, despite using the`sequential` aspect, the race condition still happening.


I achieved the expected behavior through: 

```
    identity {
          testM("Identity test (around validation inside sequential aspect)") {
            val effect = ZIO
              .accessM[Blocking](_.blocking.effectBlocking { Thread.sleep(1000) }) *> zRef.get
            assertM(effect, equalTo(RefState(Done, List(Happening, Ready, Created))))
          }
        }
```

What you think?

adamgfraser:
@edvmorango Can you do a rebase? Right now this is showing a huge diff with all the changes from the last couple of months that makes it hard to review.

adamgfraser:
@edvmorango I am not able to replicate that issue on `master`. I am using the snippet below to test. This fails if `sequential` is removed but passes if it is included. I had a hard time finding your code in the large diff but if you could either paste a working snippet with the issue you are dealing with or rebase and include it in your branch I can provide more help.

```scala
package zio.test

import zio.test.Assertion._
import zio.test.TestAspect._

import SequentialSpecUtils._

object SequentialSpec extends DefaultRunnableSpec(
  suite("SequentialSpec")(
    test("one") {
      x += 1
      assert(x, equalTo(1))
    },
    test("two") {
      x += 1
      assert(x, equalTo(2))
    },
    test("three") {
      x += 1
      assert(x, equalTo(3))
    }
  ) @@ sequential
)

object SequentialSpecUtils {
  var x = 0
}
```

edvmorango:
@adamgfraser Thank you for the help, It's working now, I think I did a mistake merging the branch before, All the tests are working as expected now, I will try to create more tests to GenSpec and for the Bank example. After this, I think the branch will be ready to be merged.

jdegoes:
@edvmorango We quite need these examples and I appreciate all the work on this pull request!

I will close this pull request temporarily until you have a chance to continue work, but feel free to re-open as your schedule allows (no rush!).

Issue #1259, opened 2019-07-27
Create `test-sbt` project for SBT integration with ZIO Test


jdegoes:
In a new `test-sbt` project, which depends on `test` project, we will provide integration with SBT, by providing an implementation of `TestFramework`.

This can be built off @NeQuissimus's work in [this pull request](https://github.com/zio/zio/pull/723/files). 

Using reflection, we should identify the set of all `RunnableSpec`, and then run them as per SBT integration requirements.

dkarlinsky:
I would like to give it a shot. 
@jdegoes What was missing from the original @NeQuissimus 's PR?

jdegoes:
@dkarlinsky It was a good start but there are some details I wanted to modify in creating ZIO Test. ZIO Test is a different code base than the original PR but also models tests as effects that produce results.

It's basically doing what Tim did, but for the current ZIO Test code base, and using `RunnableSpec` as the basis.

dkarlinsky:
Ok, I'll try to make it work with the new `RunnableSpec`.

jdegoes:
Awesome, thank you!

The new `RunnableSpec` doesn't actually run yet but you can get the code in place...

dkarlinsky:
Hmm... Is there a way for me to make it run somehow, to test what I'm doing?

jdegoes:
Put a `println("Hello World")` in the main function of `RunnableSpec`?

dkarlinsky:
Sounds pure :)
I think I'm missing a lot of context on how `RunnableSpec` is supposed to work, though...
The main method is final, and calls `System.exit()`, so it seems that the only way to run it from a custom sbt `Framework` is to create a new JVM process and wait for its result code, to figure out if the test succeeded...

jdegoes:
@dkarlinsky Probably we need a separate function, aside from `main`, to call. I will add that function, it will be called `run` and take some arguments.

jdegoes:
@dkarlinsky Do you need success / failure information? What's the ideal information you would obtain from `RunnerSpec.run`?

dkarlinsky:
Well, I see that sbt's `Runner.execute()` method (which I will need to implement) gets an event handler and seems to be expected to report test results as events to the handler.
Here's the [Event interface](https://github.com/sbt/test-interface/blob/master/src/main/java/sbt/testing/Event.java).
Looks like I will need need:  
[...]

PR #1335, opened 2019-08-02
#1259: Simple Test Framework for SBT

dkarlinsky:
https://github.com/zio/zio/issues/1259


test-sbt/jvm/src/main/scala/zio/test/runner/ZTestRunner.scala
```diff
@@ -0,0 +1,128 @@
...
+}
+
+class ZTestTask(val taskDef: TaskDef, testClassLoader: ClassLoader) extends Task {
+
+  private val platform: Platform = PlatformLive.makeDefault().withReportFailure(_ => ())
+
+  override def tags(): Array[String] = Array.empty
+
+  private def loadSpec[R, L]: RunnableSpec[R, L] = {
+    import scala.reflect.runtime.universe
```
kaishh: `scala-reflect` is not available on .js; might wanna look into https://github.com/portable-scala/portable-scala-reflect / look what the other test frameworks do right now.

dkarlinsky: Looked at how `utest` did it and switched to `portable-scala-reflect`.
This required adding `@EnableReflectiveInstantiation` to `RunnableSpec`.

dkarlinsky:
@jdegoes, I'm kinda stuck trying to get `sbt` to actually run a dummy `RunnableSpec` using the new test framework.
I've created a subproject called `testRunnerTesting` (which is not included in the root project).
It is defined as follows:
```scala
lazy val testRunnerTesting = crossProject(JVMPlatform)
  .in(file("test-sbt-testing"))
  .settings(stdSettings("zio-test-sbt-testing"))
  .settings(testFrameworks += new TestFramework("zio.test.runner.ZTestFramework"))
  .dependsOn(testRunner % "test->test;compile->compile")
```
The sub-project contains a single [RunnableSpec object](https://github.com/zio/zio/blob/262bd0e261b5f95b61246fa33e23db1369706dbc/test-sbt-testing/jvm/src/test/scala/zio/test/runner/DummyZSpec.scala) 
When I run `testRunnerTestingJVM/test` in sbt, the output is:
```
[info] ZIO-Test
[info] Done
```
Which shows that the test framework is picked up, but the spec is not discovered...

dkarlinsky:
Ok, I think I figured it out. 
It looks like sbt doesn't support parametrized types in `SubclassFingerprint.superclassName`.
Once I switched to `DefaultRunnableSpec` discovery started working

jdegoes:
@dkarlinsky Could it be name mangling or some other thing that's causing the problem? 

`DefaultRunnableSpec` is only for a common case when the `Runner` uses the ZIO default environment.

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
```diff
@@ -0,0 +1,87 @@
...
+package zio.test.runner
+
+import zio.{ RIO, ZIO }
+import zio.test.Spec.{ SuiteCase, TestCase }
+import zio.test.runner.ExecutedSpecStructure.{ ExecutedSuite, ExecutedTest, Stats }
+import zio.test.{ Assertion, ExecutedSpec, FailureDetails, TestResult }
+
+sealed abstract class ExecutedSpecStructure {
+  def stats: Stats
+  def traverse[R](
```
jdegoes: This seems to reinvent `foldM`, will look more further down.

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+  case class ExecutedSuite(label: String, stats: Stats, children: Vector[ExecutedSpecStructure])
+      extends ExecutedSpecStructure
+  case class ExecutedTest(label: String, result: ZTestResult) extends ExecutedSpecStructure {
+    override def stats: Stats = result.stats
+  }
+}
+
+sealed abstract class ZTestResult(val stats: Stats) {
```
jdegoes: I think we can reuse `Assertion` here. It has the same structure (you are using constants for `Ignore` and `Success`).

Otherwise we have 2 data structures with a similar hierarchy (success, ignore, failure w/value).

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
```diff
@@ -0,0 +1,87 @@
...
+  case class Failure(details: FailureDetails, override val rendered: String) extends ZTestResult(Stats(failed = 1))
+
+  def from(testResult: TestResult): ZTestResult =
+    testResult match {
+      case Assertion.Success     => Success
+      case Assertion.Failure(fd) => Failure(fd, describe(fd))
+      case Assertion.Ignore      => Ignored
+    }
+
+  private def describe(failureDetails: FailureDetails) = failureDetails match {
```
jdegoes: This message will be the least helpful one (`whole`). The `fragment` one will be more relevant. 

However, it would be great if we could use the existing `TestReporter` to render this. Maybe if we rendered into an intermediate form first, then you could grab that and use it here? What do you think?

dkarlinsky:
@jdegoes don't think so.
I managed to step through the sbt discovery code using a debugger.
[At this line](https://github.com/sbt/zinc/blob/develop/internal/zinc-apiinfo/src/main/scala/xsbt/api/Discovery.scala#L40) `c.structure.parents` looks for a DefaultRunnableSpec looks as follows:
```
0 = {Projection@18173} "Projection(prefix: Singleton(path: Path(components: [Lxsbti.api.PathComponent;@6741c768)), id: DefaultRunnableSpec)"
1 = {Parameterized@18461} "Parameterized(baseType: Projection(prefix: Singleton(path: Path(components: [Lxsbti.api.PathComponent;@1659053e)), id: RunnableSpec), typeArguments: [Lxsbti.api.Type;@53749d12)"
2 = {Projection@18462} "Projection(prefix: Singleton(path: Path(components: [Lxsbti.api.PathComponent;@28d4e9be)), id: Object)"
3 = {Projection@18463} "Projection(prefix: Singleton(path: Path(components: [Lxsbti.api.PathComponent;@41c7114b)), id: Any)"
```
As you can see `1` is the sbt's internal representation of `RunnableSpec` parent and it is type is `Parameterized`.
But next thing that happens, is that the parent types are flat mapped with [simpleName method](https://github.com/sbt/zinc/blob/develop/internal/zinc-apiinfo/src/main/scala/xsbt/api/Discovery.scala#L111), which doesn't have a case for `Parameterized`, returning None.
So `RunnableSpec` parent is never matched against the `SubclassFingerprit.superclassName` declared by the test framework...

test-sbt/shared/src/main/scala/zio/test/runner/ZTestFramework.scala
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package zio.test.runner
+
+import sbt.testing._
+import zio.test.DefaultRunnableSpec
+
+final class ZTestFramework extends Framework {
+  override val name = s"${Console.UNDERLINED}ZIO-Test${Console.RESET}"
```
jdegoes: Do we need the dash in `ZIO-Test`, or can we use, `ZIO Test`?

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
dkarlinsky: Yes this makes sense now.

test-sbt/shared/src/main/scala/zio/test/runner/ZTestFramework.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  val fingerprints: Array[Fingerprint] = Array(RunnableSpecFingerprint)
+
+  override def runner(args: Array[String], remoteArgs: Array[String], testClassLoader: ClassLoader): ZTestRunner =
+    new ZTestRunner(args, remoteArgs, testClassLoader)
+}
+
+object RunnableSpecFingerprint extends SubclassFingerprint {
+  val superclassName: String  = classOf[DefaultRunnableSpec].getCanonicalName
```
jdegoes: The `getCanonicalName` may be the problem here, at least with other classes. `getName` is the "reflective" name; `getCanonicalName` is the import statement name.

test-sbt/shared/src/main/scala/zio/test/runner/ZTestRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+final class ZTestRunner(val args: Array[String], val remoteArgs: Array[String], testClassLoader: ClassLoader)
+    extends Runner {
+  def done(): String = "Done"
+  def tasks(defs: Array[TaskDef]): Array[Task] =
+    defs.map(new ZTestTask(_, testClassLoader))
+}
+
+class ZTestTask(val taskDef: TaskDef, testClassLoader: ClassLoader) extends Task {
+
+  private val platform: Platform = PlatformLive.makeDefault().withReportFailure(_ => ())
```
jdegoes: Every `Runner` has its own `Platform`, can we use that?

jdegoes: Failing that, if you need a "bootstrap" platform, it may be best to create one from an `ExecutionContext` (Scala's global one), rather than to create a new thread pool.

test-sbt/shared/src/main/scala/zio/test/runner/ZTestFramework.scala
dkarlinsky: I tested and they return the same string: 
`classOf[RunnableSpec[_,_]].getName` :   zio.test.RunnableSpec
`classOf[RunnableSpec[_,_]].getCanonicalName`: zio.test.RunnableSpec

test-sbt/shared/src/main/scala/zio/test/runner/ZTestRunner.scala
dkarlinsky: Looks like we would need to make `TestRunner.platform` a val and add a `platform` method to `RunnableSpec`

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
dkarlinsky: Yes, it would be great to reuse the TestReporter.
When I wrote this the reporter didn't exist yet, but when I saw it a copied the `did not satisfy phrasing` :)

test-sbt/shared/src/main/scala/zio/test/runner/ZTestRunner.scala
```diff
@@ -0,0 +1,127 @@
...
+      .getOrElse(throw new ClassNotFoundException("failed to load object: " + fqn))
+      .loadModule()
+      .asInstanceOf[RunnableSpec[R, L]]
+  }
+
+  override def execute(eventHandler: EventHandler, loggers: Array[Logger]): Array[Task] = {
+
+    val spec = loadSpec[Any, Any]
+    val handle = for {
+      result <- spec.run
```
jdegoes: Try this instead:

```scala
spec.run.flatMap(_.foldM(...))
```
where `foldM` will handle each suite / test. Will clean up a lot of code and eliminate the need for the intermediate structures. 

Let me know if you need a hand or if I've missed something!

test/jvm/src/main/scala/zio/test/reflect/Reflect.scala
```diff
@@ -0,0 +1,6 @@
+package zio.test.reflect
+
+object Reflect {
+  type EnableReflectiveInstantiation =
+    org.portablescala.reflect.annotation.EnableReflectiveInstantiation
+}
```
jdegoes: This makes ZIO test have a dependency on portable reflection library. Hmm, we have to think if it's worth it or not. Does it support Dotty?

test/shared/src/main/scala/zio/test/RunnableSpec.scala
```diff
@@ -17,10 +17,12 @@
 package zio.test
 
 import zio.UIO
+import zio.test.reflect.Reflect.EnableReflectiveInstantiation
 
 /**
  * A `RunnableSpec` has a main function and can be run by the JVM / Scala.js.
  */
+@EnableReflectiveInstantiation
```
jdegoes: This makes ZIO test have a dependency on portable reflection library. Hmm, we have to think if it's worth it or not. Does it support Dotty?

dkarlinsky: Looks like it might not...
It says: `Scala 2.10.x, 2.11.x, 2.12.x and 2.13.0-M4` on their [readme](https://github.com/portable-scala/portable-scala-reflect#setup).
In any case it looks like I'm still missing some pieces to have sbt run specs under Scala.js

kaishh: Scalatest recently ported & published for dotty, might want to look at what they do. (Also, dotty works just fine with 2.12 test frameworks under DottyCompat mode, might not need any special support)

jdegoes: Well, let's leave it in for now, we can decide for sure later! Maybe some small Scala.js code is all we need to do the reflection (iterate keys in `global` / `window`).

test-sbt/shared/src/main/scala/zio/test/runner/ZTestFramework.scala
jdegoes: Makes sense, no mangling or otherwise funny business!

test-sbt/jvm/src/test/scala/zio/test/runner/ZTestFrameworkSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+import java.util.concurrent.atomic.AtomicReference
+
+import sbt.testing._
+import zio.BaseCrossPlatformSpec
+import zio.test.{ assert, suite, test, DefaultRunnableSpec, Predicate, TestAspect }
+import ZTestFrameworkSpec._
+
+import scala.collection.mutable.ArrayBuffer
+
+class ZTestFrameworkSpec extends BaseCrossPlatformSpec {
```
jdegoes: Seeing as how this is so foundational, I'd rather not rely on any testing framework, but rather do what ZIO test does (main function with asserts). Removes another dependency.

jdegoes:
> I managed to step through the sbt discovery code using a debugger.
At this line c.structure.parents, for a DefaultRunnableSpec looks as follows:

@dkarlinsky Great detective work!

I think this is a bug in SBT. However, we can work around it in ZIO test.

We can change `RunnerSpec` to use type members instead of type variables, something like:

```scala
abstract class RunnableSpec {
  type Label
  type Test

  def runner: TestRunner[Label, Test]
  def spec = Spec[Label, Test]

  /**
   * A simple main function that can be used to run the spec.
   *
   * TODO: Parse command line options.
   */
  final def main(args: Array[String]): Unit = { val _ = runner.unsafeRunSync(spec) }

  /**
   * Returns an effect that executes the spec, producing the results of the execution.
   */
  final val run: UIO[ExecutedSpec[Label]] = runner.run(spec)
}
object RunnableSpec {
  type Aux[L, T] = RunnableSpec { type Label = L; type Test = T }

  def apply[L, T](runner0: TestRunner[L, T])(spec0: => Spec[Label, Test]): RunnableSpec = 
    new RunnableSpec {
      type Label = L 
      type Test = T

      def runner = runner0
      def spec = spec0
    }
}
```

Now `RunnerSpec` can be used as before, but it has no type parameters.

What do you think?

test-sbt/shared/src/main/scala/zio/test/runner/ZTestRunner.scala
```diff
@@ -106,18 +101,32 @@ object ZTestEvent {
       0L
     )
 
-  private def maybeThrowable(testResult: ZTestResult) = testResult match {
-    case ZTestResult.Failure(FailureDetails.Runtime(cause), _) =>
+  private def maybeThrowable(testResult: TestResult) = testResult match {
+    case Assertion.Failure(FailureDetails.Runtime(cause)) =>
```
jdegoes: The full details of the failure would be nice to print out at some point, pretty print, if not here then elsewhere (I suppose in the reporter).

test-sbt/shared/src/main/scala/zio/test/runner/ZTestRunner.scala
```diff
@@ -106,18 +101,32 @@ object ZTestEvent {
...
+object ZTestTask {
+
+  private def describeResult(testResult: TestResult) = testResult match {
+    case f: Assertion.Failure[FailureDetails] => describeFailure(f.message)
+    case Assertion.Success                    => "SUCCESS"
+    case Assertion.Ignore                     => "IGNORED"
+  }
+
+  private def describeFailure(failureDetails: FailureDetails) = failureDetails match {
+    case FailureDetails.Runtime(cause)         => cause.prettyPrint
```
jdegoes: Ah, nice, prettyPrint!

dkarlinsky:
> ```scala
> object RunnableSpec {
>   type Aux[L, T] = RunnableSpec { type Label = L; type Test = T }
> 
>   def apply[L, T](runner0: TestRunner[L, T])(spec0: => Spec[Label, Test]): RunnableSpec = 
>     new RunnableSpec {
>       type Label = L 
>       type Test = T
> 
>       def runner = runner0
>       def spec = spec0
>     }
> }
> ```
This looks very nice as a constructor for a new `RunnableSpec`, but I don't see how this will work with subclassing. 
The specs picked up by sbt need to be objects that are subclasses of `RunnableSpec`...

test-sbt-testing/js/src/test/scala/zio/test/runner/DummyJSSpec.scala
```diff
@@ -0,0 +1,14 @@
+package zio.test.runner
+
+import zio.test.{DefaultRunnableSpec, Predicate, assert, suite, test}
+
+object DummyJSSpec extends DefaultRunnableSpec({
```
jdegoes: You shouldn't need both the parentheses here and the curly braces, just one, and by convention, I'd say use `{ }`.

test-sbt-testing/js/src/test/scala/zio/test/runner/DummyJSSpec.scala
```diff
@@ -0,0 +1,14 @@
+package zio.test.runner
```
jdegoes: I'd add this to the top-level `examples` project that depends on both ZIO Core and ZIO Tests. It can have examples of both ZIO Test and ZIO proper.

jdegoes:
> This looks very nice as a constructor for a new RunnableSpec, but I don't see how this will work with subclassing.

Hmm, can you try this:

```scala
abstract class AbstractRunnableSpec { ... } // As before
type RunnableSpec[L, T] = AbstractRunnableSpec { type Label = L; type Test = T }
```

Then perhaps the user can do:

```scala
new RunnableSpec(runner)(spec)
```

If we have the appropriate 2 parameter list constructor.

test-sbt-testing/js/src/test/scala/zio/test/runner/DummyJSSpec.scala
dkarlinsky: Just `{}` doesn't compile:
```
Error:(5, 28) not enough arguments for constructor DefaultRunnableSpec: (_spec: => zio.test.ZSpec[zio.clock.Clock with zio.console.Console with zio.system.System with zio.random.Random,Nothing,String])zio.test.DefaultRunnableSpec.
Unspecified value parameter _spec.
object DummyJSSpec extends DefaultRunnableSpec {
```
I'll change to just `()`

test-sbt-testing/js/src/test/scala/zio/test/runner/DummyJSSpec.scala
dkarlinsky: The idea for this project was to be able to manually run `testRunnerTestingJVM/test` to test the sbt runner.
I guess I could to the same on the examples project...

dkarlinsky:
> > This looks very nice as a constructor for a new RunnableSpec, but I don't see how this will work with subclassing.
> 
> Hmm, can you try this:
> 
> ```scala
> abstract class AbstractRunnableSpec { ... } // As before
> type RunnableSpec[L, T] = AbstractRunnableSpec { type Label = L; type Test = T }
> ```
> 
> Then perhaps the user can do:
> 
> ```scala
> class MyRunnableSpec extends RunnableSpec(MyRunner) { 
>   ...
> }
> ```
> 
> If we have the appropriate 2 parameter list constructor.

I don't think `AbstractRunnableSpec` can have a constructor that accepts Spec and Runner because we don't have the `Label` and `Test` types defined yet...
Currently, I just have [DefaultRunnableSpec](https://github.com/dkarlinsky/zio/blob/b793b60caa70aa14bbe22b84627bbe1ffb049ade/test/shared/src/main/scala/zio/test/DefaultRunnableSpec.scala#L26-L25) define the abstract types and the two parameter values directly.
Perhaps we can have `AbstractRunnableSpec` extending `RunnableSpec` to make it easier to implement specs with non default runner. Something like:

```scala
abstract class AbstractRunnableSpec[L, T](runner0: TestRunner[L, T])(spec0: => Spec[L, T]) extends RunnableSpec {
  override type Label = L
  override type Test = T

  override def runner = runner0
  override def spec =spec0
}
```
Then `DefaultRunnableSpec` can go back to looking like it did before.

kaishh:
@dkarlinsky Extending it should work; We use a lot of classes with type parameters that extend base Scalatest classes. I'd expect the bug with type parameter handling to apply only to the class that you're searching for, not to subclasses. (Coincidentally, using defs in `AbstractRunnableSpec` just makes it easier for me to plug a mutable DSL 😆 )

test-sbt/shared/src/main/scala/zio/test/runner/ZSpecStructure.scala
dkarlinsky: Exposed rendering to a  intermediate form in TestReporter and reused

jdegoes:
@dkarlinsky 

This is a good idea. 

So let's define the following:

```scala
abstract class AbstractRunnableSpec {
  type Label
  type Test

  def runner: TestRunner[Label, Test]
  def spec: Spec[Label, Test])

  ...
}
abstract class RunnableSpec[L, T](runner: TestRunner[L, T])(spec0: => Spec[L, T]) extends AbstractRunnableSpec[L, T] {
  type Label = L
  type Test = T

  def spec = spec0
}
```

Now we have the same syntax as before, just with a workaround for the bug.

Sound good?

/cc @kaishh

dkarlinsky:
@jdegoes I think I'm done with all the functionality and requested changes.
SBT runs tests under both JVM and JS

jdegoes:
@dkarlinsky Awesome work! ZIO Test is one more step closer to fully operational...

Issue #803, opened 2019-04-27
STM: Support Scala.js


jdegoes:
Currently "synchronized" is used in the implementation of STM, which is not portable to Scala.js. This functionality should be factored out into a simple data structure:

```scala
class LockedRef[A] {
  def get: A
  def set(a: A): A 
  def lock: Unit
  def unlock: Unit
}
```

The Scala.js implementation will obviously not use locking.

The data structure can be placed in the jvm and js packages `scalaz.zio.internal`.

Then the use of `synchronized` inside `STM.scala` can be replaced by the above structure, bringing full portability of STM to Scala.js.

heraklos:
I would like to take this issue :)

mgibowski:
I just checked and found that the implementation also uses `java.util.concurrent.Semaphore` (`internal.globalLock`) and the spec uses `java.util.concurrent.CountDownLatch` - both of these are not implemented in Scala.js as they contain blocking operations. 

@jdegoes do you consider them portable in some way in the context of zio?
Or should they rather be extracted into other simple data structures?

heraklos:
@jdegoes 
@mgibowski is right. When compiling Scala.js code containing STM method, the error `Referring to non-existent class java.uti.concurrent.Semaphore` was thrown. 
Am I supposed to create two data structure `LockedRef` and `Semaphore` to support STM in Scala.js and replace `synchronized` and `Semaphore` respectively?

heraklos:
@jdegoes 
I have one more question.
When I just removed all `done synchronized`s from `atomically` in `STM.scala`, `STMSpec` still worked.
Does that mean we don't need `done synchronized` in the first place, or I should add some tests that fails when `atomically` is controlled exclusively?

heraklos:
> When I just removed all done synchronizeds from atomically in STM.scala, STMSpec still worked.
Does that mean we don't need done synchronized in the first place, or I should add some tests that fails when atomically is controlled not exclusively?

It seems that removing `done synchronized` escapes test failures when there are not enough fibers forked to cause data confliction. 
I confirmed that changing the number of fibers from 10 to 100 caused test failure.
I may have to modify the number of fibers in some test case for STM.

jdegoes:
@heraklos 

Thank you for taking this project on!

> I think it is possible to replace synchronized and Semaphore with just LockedRef data type because Semaphore being used in STM is binary semaphore and can be regarded as lock.

You are right, we only need `LockedRef`.

[...]

Issue #1216, opened 2019-07-22
Flaky queue test


jdegoes:
```
[info]     make a bounded queue, offer 4 values, `takeUpTo` with max = 2, and then `takeUpTo` again with max = 2;
[info]      the first result must contain the first 2 values and the second one must
[error] java.lang.Exception: TIMEOUT: 60 seconds
[error] 	at zio.BaseCrossPlatformSpec$$anon$1.run(BaseCrossPlatformSpec.scala:66)
[error] 	at java.util.TimerThread.mainLoop(Timer.java:555)
[error] 	at java.util.TimerThread.run(Timer.java:505)
[error] Error: Total 620, Failed 0, Errors 1, Passed 619
```

Might be unrelated to queue, actually.

ghostdogpr:
Migrating the tests to ZIO Test may have revealed the actual error: https://circleci.com/gh/zio/zio/17876

@jdegoes looks like something wrong with `interruptChildren`. Since it will go away with #1161, it might not be a big deal. I could change the tests to interrupt the fork fibers explicitly in the meantime.

```
- offers are suspended by back pressure
[info]     Fiber failed.
[info]     An unchecked error was produced.
[info]     java.lang.NullPointerException
[info]     	at zio.Fiber$.$anonfun$interruptAll$2(Fiber.scala:296)
[info]     	at zio.ZIO$ZipLeftFn.apply(ZIO.scala:2490)
[info]     	at zio.ZIO$ZipLeftFn.apply(ZIO.scala:2488)
[info]     	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:270)
[info]     	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:602)
[info]     	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info]     	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info]     	at java.lang.Thread.run(Thread.java:748)
[info]     
[info]     Fiber:12547 was supposed to continue to:
[info]       a future continuation at zio.ZIO.ensuring(ZIO.scala:285)
[info]       a future continuation at zio.ZIO.run(ZIO.scala:1007)
[info]       a future continuation at zio.ZIO.bracket_(ZIO.scala:131)
[info]       a future continuation at zio.ZIO.run(ZIO.scala:1007)
[...]

Issue #1023, opened 2019-06-15
Flaky runtime test


NeQuissimus:
This test loves to time out

```
[info]   RTS concurrency correctness
[info]     + shallow fork/join identity
[info]     + deep fork/join identity
[info]     + asyncPure creation is interruptible
[info]     + asyncInterrupt runs cancel token on interrupt
[info]     + supervising returns fiber refs
[info]     + supervising in unsupervised returns Nil
[info]     + supervise fibers
[error]     x supervise fibers in supervised
[error]  TIMEOUT: 60 seconds (TestRuntime.scala:18)
```

Last observed with Scala 2.12 and JDK 8

jdegoes:
```scala
  def testSupervised =
    unsafeRun(for {
      pa <- Promise.make[Nothing, Int]
      pb <- Promise.make[Nothing, Int]
      _ <- (for {
            p1 <- Promise.make[Nothing, Unit]
            p2 <- Promise.make[Nothing, Unit]
            _ <- p1
                  .succeed(())
                  .bracket_[Any, Nothing]
                  .apply[Any](pa.succeed(1).unit)(IO.never)
                  .fork //    TODO: Dotty doesn't infer this properly
            _ <- p2.succeed(()).bracket_[Any, Nothing].apply[Any](pb.succeed(2).unit)(IO.never).fork
            _ <- p1.await *> p2.await
          } yield ()).interruptChildren
      r <- pa.await zip pb.await
    } yield r) must_=== (1 -> 2)
```

jdegoes:
This test, as convoluted as it is, looks correct to me. 

Meaning the problem may be in `bracket_` or `interruptChildren` or something lower level.

iravid:
Hmm FWIW ive seen “returns fiber refs” time out as well.

ghostdogpr:
There are several tests that are flaky in RTSSpec, the common point from what I’ve seen is that they all involve supervision. 

With the multiplication of test configurations (jdk11, 2.11, js, etc) it became extremely difficult not to run into it :/

wongelz:
The "returns fiber refs" test sometimes fails with : `'WrappedArray(zio.internal.FiberContext@867d0ad)' doesn't have size 2 but size 1`

jdegoes:
@ghostdogpr @wongelz See [here](https://github.com/zio/zio/pull/1063)

Still more work to do on the fundamental problem `supervise fibers in supervised`

wongelz:
[...]

PR #1338, opened 2019-08-03
STM supports scala.js

heraklos:
Issue : https://github.com/zio/zio/issues/803

now scala.js code with STM like the following works 
```scala
  def run(arg: List[String]) =
    (for {
      intVar <- TRef.make(14)
      _ <- intVar.set(42)
      v <- intVar.get
    } yield v ).commit
```

# Changes in code
- [x] Implement `LockedRef` class
  - simple data structure to replace `sychronized` and `java.util.concurrent.Semaphore` for locking, which are not supported in scala.js

- [x] Replace `forEach` method calls with iterator-based implementation
  - this is necessary since scala.js does not support jdk8 higher-order function for `HashMap` as of now

- [x] Replace `synchronized` and `java.util.concurrent.Semaphore` with `LockedRef`

# Changes in performance
 - [x] Comparison of benchmark results between master branch and this feature branch
   - done by executing `sbt:zio> benchmarks/jmh:run .*SemaphoreBenchmark.*` task locally

master branch

```sbt
[info] Result "zio.stm.SemaphoreBenchmark.semaphoreCatsContention":
[info]   141.237 ±(99.9%) 7.604 ops/s [Average]
[info]   (min, avg, max) = (120.844, 141.237, 155.422), stdev = 10.151
[info]   CI (99.9%): [133.633, 148.841] (assumes normal distribution)

[info] Result "zio.stm.SemaphoreBenchmark.semaphoreContention":
[info]   101.732 ±(99.9%) 5.197 ops/s [Average]
[info]   (min, avg, max) = (90.690, 101.732, 114.566), stdev = 6.938
[info]   CI (99.9%): [96.535, 106.929] (assumes normal distribution)

[info] Result "zio.stm.SemaphoreBenchmark.tsemaphoreContention":
[info]   114.949 ±(99.9%) 5.291 ops/s [Average]
[info]   (min, avg, max) = (98.940, 114.949, 124.836), stdev = 7.064
[info]   CI (99.9%): [109.658, 120.241] (assumes normal distribution)

[info] # Run complete. Total time: 00:25:06
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Benchmark                                   (fibers)  (ops)   Mode  Cnt    Score   Error  Units
[info] SemaphoreBenchmark.semaphoreCatsContention        10   1000  thrpt   25  141.237 ± 7.604  ops/s
[info] SemaphoreBenchmark.semaphoreContention            10   1000  thrpt   25  101.732 ± 5.197  ops/s
[info] SemaphoreBenchmark.tsemaphoreContention           10   1000  thrpt   25  114.949 ± 5.291  ops/s
```

feature branch
```sbt
[info] Result "zio.stm.SemaphoreBenchmark.semaphoreCatsContention":
[info]   145.410 ±(99.9%) 7.057 ops/s [Average]
[info]   (min, avg, max) = (121.232, 145.410, 158.190), stdev = 9.421
[info]   CI (99.9%): [138.353, 152.467] (assumes normal distribution)

[info] Result "zio.stm.SemaphoreBenchmark.semaphoreContention":
[info]   106.594 ±(99.9%) 3.541 ops/s [Average]
[info]   (min, avg, max) = (96.782, 106.594, 115.379), stdev = 4.727
[info]   CI (99.9%): [103.053, 110.135] (assumes normal distribution)

[info] Result "zio.stm.SemaphoreBenchmark.tsemaphoreContention":
[info]   99.766 ±(99.9%) 27.598 ops/s [Average]
[info]   (min, avg, max) = (65.356, 99.766, 152.758), stdev = 36.843
[info]   CI (99.9%): [72.168, 127.364] (assumes normal distribution)

[info] # Run complete. Total time: 00:25:07
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Benchmark                                   (fibers)  (ops)   Mode  Cnt    Score    Error  Units
[info] SemaphoreBenchmark.semaphoreCatsContention        10   1000  thrpt   25  145.410 ±  7.057  ops/s
[info] SemaphoreBenchmark.semaphoreContention            10   1000  thrpt   25  106.594 ±  3.541  ops/s
[info] SemaphoreBenchmark.tsemaphoreContention           10   1000  thrpt   25   99.766 ± 27.598  ops/s
```

the result is that average ops/s on `zio.stm.SemaphoreBenchmark.semaphoreCatsContention` and `zio.stm.SemaphoreBenchmark.semaphoreContention` got worse for about 5 ops/s as expected, and average ops/s on `zio.stm.SemaphoreBenchmark.tsemaphoreContention` unexpectedly got better for about 15 ops/s


ghostdogpr:
Great work! Could you move the STM tests from `jvm` to `shared`? That would make them run with ScalaJS and validate the compatibility.

heraklos:
I moved the STM tests from `jvm` to `shared`. 
But when I ran the tests, then I got errors like this : 
```sbt
[error] Referring to non-existent method java.util.concurrent.CountDownLatch.<init>(scala.Int)
[error]   called from zio.stm.STMSpec.$$anonfun$e28$1()zio.ZIO
[error]   called from zio.stm.STMSpec.e28()org.specs2.matcher.MatchResult
[error]   called from zio.stm.STMSpec.$$anonfun$is$27()org.specs2.matcher.MatchResult
[error]   called from zio.stm.STMSpec.$$anonfun$is$1()org.specs2.specification.core.Fragments
[error]   called from zio.stm.STMSpec.is()org.specs2.specification.core.SpecStructure
[error]   called from org.specs2.specification.core.SpecificationStructure.$$anonfun$structure$1(org.specs2.specification.core.Env)org.specs2.specification.core.SpecStructure
[error]   called from zio.FunctionIOSpec.$$anonfun$structure$1(org.specs2.specification.core.Env)org.specs2.specification.core.SpecStructure
[error]   called from org.specs2.specification.core.SpecificationStructure.structure()scala.Function1
[error]   called from org.specs2.Specification.org$specs2$specification$core$ImmutableSpecificationStructure$$super$structure()scala.Function1
[error]   called from org.specs2.specification.core.ImmutableSpecificationStructure.$$anonfun$structure$1(org.specs2.specification.core.Env)org.specs2.specification.core.SpecStructure
[error]   called from zio.FunctionIOSpec.$$anonfun$structure$1(org.specs2.specification.core.Env)org.specs2.specification.core.SpecStructure
[error]   called from org.specs2.specification.core.ImmutableSpecificationStructure.structure()scala.Function1
[error]   called from org.specs2.Specification.structure()scala.Function1
[error]   called from org.specs2.runner.SbtTask.$$anonfun$createSpecStructure$2(org.specs2.specification.core.Env,org.specs2.specification.core.SpecificationStructure)scala.Option
[error]   called from org.specs2.runner.SbtTask.createSpecStructure(sbt.testing.TaskDef,java.lang.ClassLoader,org.specs2.specification.core.Env)org.specs2.control.eff.Eff
[error]   called from org.specs2.runner.SbtTask.executeFuture(sbt.testing.EventHandler,[sbt.testing.Logger)scala.concurrent.Future
[error]   called from org.specs2.runner.SbtTask.execute(sbt.testing.EventHandler,[sbt.testing.Logger,scala.Function1)scala.Unit
[error]   called from org.scalajs.testinterface.internal.Bridge$.$$anonfun$executeFun$1(sbt.testing.Runner,scala.Int,org.scalajs.testcommon.ExecuteRequest)scala.concurrent.Future
[error]   called from org.scalajs.testinterface.internal.Bridge$.executeFun(scala.Int,sbt.testing.Runner)scala.Function1
[error]   called from org.scalajs.testinterface.internal.Bridge$.$$anonfun$createRunnerFun$1(scala.Boolean,org.scalajs.testcommon.RunnerArgs)scala.Unit
[error]   called from org.scalajs.testinterface.internal.Bridge$.createRunnerFun(scala.Boolean)scala.Function1
[error]   called from org.scalajs.testinterface.internal.Bridge$.start()scala.Unit
[error]   called from org.scalajs.testinterface.internal.Bridge$.__exportedInits
[error]   exported to JavaScript with @JSExport
[error] involving instantiated classes:
[error]   zio.stm.STMSpec
[error]   zio.FunctionIOSpec
[error]   zio.internal.OneShotSpec
[error]   zio.PromiseSpec
[error]   zio.FiberLocalSpec
[error]   zio.duration.DurationSyntaxSpec
[error]   zio.RetrySpec
[error]   zio.RepeatSpec
[error]   zio.RefMSpec
[error]   zio.internal.ExecutorSpec
[error]   zio.ParallelErrorsSpec
[error]   zio.internal.MutableConcurrentQueueSpec
[error]   zio.FiberRefSpec
[error]   zio.RefSpec
[error]   zio.duration.DurationSpec
[error]   zio.FiberSpec
[error]   zio.internal.StackBoolSpec
[error]   zio.QueueSpec
[error]   zio.ExitSpec
[error]   org.specs2.runner.SbtTask
[error]   org.scalajs.testinterface.internal.Bridge$
.
.
.
```

It seems that CountDownLatch use in STMSpec causes these errors. Please teach me if you know some workaround to this.

heraklos:
I got away with `CountDownLatch` by replacing it with `Promise`.
But then I got new errors like the following:
```sbt
[error] java.util.concurrent.ExecutionException: Boxed Error
```

I'm trying to fix this.

ghostdogpr:
@heraklos we could use a `Promise[Nothing, Unit]` instead of `CountDownLatch` but then `latch.countDown()` would have to be `unsafeRun(latch.succeed(()))` so that it is "unsafely" executed.

For example:
```scala
  def e21 =
    unsafeRun {
      for {
        done  <- Promise.make[Nothing, Unit]
        tvar1 <- TRef.makeCommit(0)
        tvar2 <- TRef.makeCommit("Failed!")
        fiber <- (STM.atomically {
                  for {
                    v1 <- tvar1.get
                    _  <- STM.succeedLazy(unsafeRun(latch.succeed(())))
                    _  <- STM.check(v1 > 42)
                    _  <- tvar2.set("Succeeded!")
                    v2 <- tvar2.get
                  } yield v2
                } <* done.succeed(())).fork
        _    <- latch.await
        old  <- tvar2.get.commit
        _    <- tvar1.set(43).commit
        _    <- done.await
        newV <- tvar2.get.commit
        join <- fiber.join
      } yield (old must_=== "Failed!") and (newV must_=== join)
    }
```

ghostdogpr:
Oops, you found it already. Are you getting those errors only with ScalaJS?

heraklos:
@ghostdogpr 
Thank you for helping me about `CountDownLatch` :)

And yes to your question.  The error pops out only when executing ScalaJS test.
I noticed that there is `[error] Caused by: java.lang.Error: Cannot block for result to be set in JavaScript` in errors shown in the stack trace, and this message seems to come from `OneShot.get` but I cannot see any connection between STMSpec and this class...

ghostdogpr:
Maybe those 3 tests that target concurrent behavior should stay in JVM since they don't make too much sense in JS?

@jdegoes what do you think?

jdegoes:
The whole STM test suite should run on Scala.js without issue. But `unsafeRunSync` cannot be used. 

Specs 2 supports returning Futures, so we need to `unsafeRunToFuture` and tap into the Specs 2 support. 

Does that make sense?

heraklos:
Thank you for your guide!
I'll replace all the tests that fail due to `unsafeRunSync` with  `unsafeRunToFuture` and spec2 Future support 👍

heraklos:
@ghostdogpr 
Now I got an error 
```
[error] java.lang.Exception: TIMEOUT: 60 seconds
[error] 	at zio.BaseCrossPlatformSpec$$anon$1.run(BaseCrossPlatformSpec.scala:67)
[error] 	at java.util.TimerThread.mainLoop(Timer.java:555)
[error] 	at java.util.TimerThread.run(Timer.java:505)
[error] Error: Total 626, Failed 0, Errors 1, Passed 625
[error] Error during tests:
[error] 	zio.QueueSpec
[error] (coreTestsJVM / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 174 s, completed Aug 4, 2019 1:14:06 AM
```

I think my changes does not affect `QueueSpec` but I'm not sure of it...
Could you run again CI tests for me?

ghostdogpr:
Yeah it’s probably the same issue as #1216. I restarted it.

heraklos:
Thanks a lot 😆

But another flaky test popped up... I'm sure this is the same issue as https://github.com/zio/zio/issues/1023
```
[error]     x supervise fibers in supervised
```
Would you run tests again?

ghostdogpr:
It’s green now :D

core-tests/shared/src/test/scala/zio/stm/STMSpec.scala
```diff
@@ -187,28 +185,24 @@ final class STMSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends Tes
...
   def e15 =
-    unsafeRun(
-      for {
-        tVar  <- TRef.makeCommit(0)
-        fiber <- ZIO.forkAll(List.fill(10)(incrementVarN(99, tVar)))
-        _     <- fiber.join
-        value <- tVar.get.commit
-      } yield value
-    ) must_=== 1000
+    for {
```
ghostdogpr: Just to be consistent, can you remove `unsafeRun` in all the tests?

heraklos: Fixed 😆

ghostdogpr:
LGTM, @jdegoes you wanna have a final look?

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -537,10 +547,9 @@ object STM {
           value match {
             case _: TRez.Succeed[_] =>
               if (analysis eq JournalAnalysis.ReadWrite) {
-                globalLock.acquire()
-
-                try if (isValid(journal)) commitJournal(journal) else loop = true
-                finally globalLock.release()
+                globalLock.synchronized {
```
jdegoes: I thought `.synchronized` threw exceptions on Scala.js. No???

core/shared/src/main/scala/zio/internal/LockedRef.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def synchronized(f: => Unit): Unit = {
+    lock()
+    try f
+    finally unlock()
+  }
+
+  def lock(): Unit = {
+    var loop = true
+    while (loop) {
```
jdegoes: This spin loop is not efficient. You should just use an ordinary Java `Semaphore` or `Lock`. An alternative would be to implement a lock atop `synchronized`, which would be more efficient than anything else. You have to know how to use `waitAll` / `notifyAll` and deal with spurious wakeups, etc. So the simple solution is to use `Semaphore` here the more complex one is to build something low-level

core/shared/src/main/scala/zio/internal/LockedRef.scala
```diff
@@ -0,0 +1,39 @@
+package zio.internal
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+class LockedRef[A](var ref: A) {
```
jdegoes: If this raw `var` is stored here, it needs to be marked as @volatile.

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -561,7 +570,7 @@ object STM {
 
     private[this] val txnCounter: AtomicLong = new AtomicLong()
 
-    final val globalLock = new java.util.concurrent.Semaphore(1)
+    final val globalLock = new zio.internal.LockedRef(1)
```
jdegoes: This is not even using the `Int` passed into `LockedRef`, which means `globalLock` is unnecessary. You may as well just synchronize on `STM` object, e.g. `STM.synchronized { ... }` or maybe make a private object, e.g. `final val globalLock = new AnyRef { }`.

core/shared/src/main/scala/zio/stm/STM.scala
heraklos: Yes, java built-in `synchronized` throws exception on Scala.js.
So, I implemented and used `synchronized` method on `LockedRef` class which wraps the procedure in `{}` code with locking/unlocking.

core/shared/src/main/scala/zio/internal/LockedRef.scala
heraklos: Thank you for your advice on locking!
Unfortunately, `java.util.concurrent.Semaphore` is not supported by scala.js...
Maybe I should implement lock with `ReentrantLock` here while learning `waitAll` / `notifyAll` and other concurrent matter to be cared for. (java's built-in `synchronized` is not supported as you taught us in https://github.com/zio/zio/issues/803)

But on reading your overall review on this PR, this class is not needed in the first place...?

core/shared/src/main/scala/zio/stm/STM.scala
heraklos: You're right! This `LockedRef` instance does not use the passed `Int` at all.
but my understanding is that we can't use java's built-in `synchronized` in scala.js.
So, I have to come up with some data structure similar to `LockedRef` which does not take an argument.

heraklos:
Thank you for your detailed review 😆
Your advice seems perfectly reasonable to me and I would like to implement that.

But here is the problem.  I actually used `LockedRef` for synchronizing some operations which are synchronized by java's built-in `synchronized` before. So, in my opinion, it's not just a container class.

So I'm afraid that we can get by simple with `AtomicBoolean` + `synchronized` for `done`, and a `val globalLock: AnyRef = new AnyRef { }` for the global lock as you recommended for me, because `synchronized` is `LockedRef`'s method (sorry for  confusing method name)

Instead, what about if separating `LockedRef` into `jvm`/`js` package and implement lock/unlock with `Semaphore` in `jvm` side, and implement lock/unlock with `Lock` or other possible solution in `js` side for better performance?

I may have taken your words wrongly again,

core/shared/src/main/scala/zio/stm/STM.scala
jdegoes: @heraklos How about this:

```scala
// JS
object Sync {
  def apply[A](anyRef: AnyRef)(f: => A): A = f
// JVM
object Sync {
  def apply[A](anyRef: AnyRef)(f: => A): A = anyRef.synchronized { f }
}
```

Then you can replace `x.synchronized { f }` with `Sync(x) { f }`

Sound good?

core/shared/src/main/scala/zio/internal/LockedRef.scala
jdegoes: Javascript will not need a semaphore. I think we can get by with purely `synchronized` (and `Sync` class above), without any waiting / awaiting.

Can you give it a try and let me know? Thank you!

core/shared/src/main/scala/zio/stm/STM.scala
heraklos: @jdegoes 
That sounds perfect to me!
And with `Sync`, `STMSpec` passed for both `jvm`/`js`!

The only obstacle here is the following error 
```
parameter value anyRef in method apply is never used
```
for `Sync` of `js`.
```
// JS
object Sync {
  def apply[A](anyRef: AnyRef)(f: => A): A = f
}
```
I'm gonna think of some workaround.

core/shared/src/main/scala/zio/internal/LockedRef.scala
heraklos: Yes, you're totally correct.
The mechanism of Javascript runtime slipped my mind.

core/js/src/main/scala/zio/internal/Sync.scala
```diff
@@ -0,0 +1,8 @@
+package zio.internal
+
+object Sync {
+  def apply[A](anyRef: AnyRef)(f: => A): A = {
```
heraklos: For `STM.scala` to be shared, this method should take the same arguments as `Sync` in `jvm` does.
But the compiler does not allow us to take `anyRef` and does nothing with it.
So I avoided the error just by asserting the existence of `anyRef`.

jdegoes: You can do, inside the body:

```scala
val _ = anyRef
```

cw-kajiwara: I learned another thing! Thank you so much :)

jdegoes:
giphy

ghostdogpr:
@heraklos Glad to see you went through this 👏👏👏 One less blocker for ZIO 1.0!

heraklos:
@jdegoes 
Thanks a lot for teaching me 😆  
I'm really glad that this PR was merged since this was my first contribution to application code in `zio`! 
Next time, I'll do it more quickly and nicely :)

@ghostdogpr 
Thank you so much for your warmful support 😆I finally made it!
Your continuous help enabled me to move forward at all times. Thanks!

jdegoes:
@heraklos Congratulations on your first merge to core... first of many, I hope! 😉

PR #1340, opened 2019-08-03
Change time dependency in core tests

### LINKED ISSUE #1325: Clock is not compatible with Scala.js
`Clock` uses `java.time`, which is not part of the Scala.js Java library. We get around this in the tests by depending on `scalajs-java-time` but this does not implement `OffsetDateTime` or `ZoneId` (see [implementation of Instant](https://github.com/scala-js/scala-js-java-time/blob/master/src/main/scala/java/time/Instant.scala)). I think this didn't come up because you don't see the error until runtime and almost all of our tests are on the JVM right now. But for example if I try to run the program below on Scala.js I get an error "Referring to non-existent class java.time.ZoneId", even with a dependency  on `scalajs-java-time`.

```scala
object TestMain  {

  def main(args: Array[String]): Unit = {

    println(java.time.LocalDate.now)      
    // 2019-08-02

    val runtime = new zio.DefaultRuntime {}
    val clock = zio.clock.Clock.Live
    println(runtime.unsafeRun(clock.clock.currentDateTime))
    // [error] Referring to non-existent class java.time.ZoneId
  }
}
```

Am I missing something here?

adamgfraser:
Changes time dependency in core tests from `scalajs-java-time` to `scala-java-time`. Resolves #1325.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -93,7 +93,7 @@ lazy val coreTestsJVM = coreTests.jvm
 
 lazy val coreTestsJS = coreTests.js
   .settings(
-    libraryDependencies += "org.scala-js" %%% "scalajs-java-time" % "0.2.5" % Test
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test
   )
 
 lazy val streams = crossProject(JSPlatform, JVMPlatform)


jdegoes:
Thank you! 🙏

PR #1341, opened 2019-08-03
WIP - Implement `FiberRef.asThreadLocal` for interop with side effecting code

mlangc:
The purpose if this PR is to discuss and eventually merge APIs that expose information about the current ZIO fiber stack to side effecting code. More concretely, I've added an impure method
```scala
def unsafeGetFiberStack: List[FiberId]
```
to `Runtime`, that can be used from side effecting code. The implementation, as well as the tests are just prototypes, that have known shortcomings (see below). I didn't want to invest too much time on the implementation before there is agreement on the API and the general direction of this PR. The primary use case I have in mind is implementing fully transparent, fiber aware MDC logging for `log4j2` (see https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ContextDataInjector).

These are the shortcomings of the current proof of concept implementation that I'm aware of:
* The implementation does not clear the `ThreadLocal` variable when the interpreter exits. 
* The implementation does not currently handle nested interpreter calls, that is when someone does `unsafeRun(Task(unsafeRun(whatever)))`. 
* I suspect that you are not too happy with the way I added `ThreadLocal` support for ScalaJS.

None of the implementation related points above poses a serious problem as far as I can see, so I would like to discuss the API first:
* Do you think an API like this makes sense at all, or would you prefer an entirely different approach? 
* Maybe it would make sense to indicate which fibers in the stack are still alive (for children that outlive their parents)?
* Everything that's exposed in an impure API should also be exposed in a pure API. Mabe `Fiber.Descripter` should be extended accordingly?

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
+++ b/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
@@ -1,0 +1,40 @@
+package zio
+
+import java.util.concurrent.atomic.AtomicReference
+
+import zio.FiberRefSpecUtil._
+import zio.test.Assertion._
+import zio.test._
+
+object FiberRefSpecJvm
+    extends ZIOBaseSpec(
+      suite("FiberRefSpecJvm")(
+        testM("unsafe handles properly invoke fallback functions if fiber specific data cannot be accessed") {
+          for {
+            fiberRef    <- FiberRef.make(initial)
+            handle      <- fiberRef.unsafeHandle
+            fallbackRef <- UIO(new AtomicReference(update1))
+            resRef      <- UIO(new AtomicReference("" -> ""))
+
+            unsafelyGetSetGet = new Runnable {
+              def run(): Unit = {
+                val v1 = handle.unsafeGet(fallbackRef.get())
+                handle.unsafeSet(update2, fallbackRef.set)
+                val v2 = handle.unsafeGet(fallbackRef.get())
+                resRef.set(v1 -> v2)
+              }
+            }
+
+            thread <- UIO(new Thread(unsafelyGetSetGet))
+            _      <- UIO(thread.start()).ensuring(UIO(thread.join()))
+
+            value0           <- fiberRef.get
+            values           <- UIO(resRef.get())
+            (value1, value2) = values
+          } yield assert((value0, value1, value2), equalTo((initial, update1, update2)))
+        }
+      )
+    ) {
+  override def runner =
+    super.runner.withPlatform(_.withFiberContextPropagation(true))
+}
--- a/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
+++ b/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
@@ -36,7 +36,4 @@ object FiberRefSpecJvm
           } yield assert((value0, value1, value2, value3), equalTo((update1, initial, update2, initial)))
         }
       )
-    ) {
-  override def runner =
-    super.runner.withPlatform(_.withFiberContextPropagation(true))
-}
+    )
--- a/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
+++ b/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
@@ -9,29 +9,31 @@ import zio.test._
 object FiberRefSpecJvm
     extends ZIOBaseSpec(
       suite("FiberRefSpecJvm")(
-        testM("unsafe handles properly invoke fallback functions if fiber specific data cannot be accessed") {
+        testM("unsafe handles behave properly if fiber specific data cannot be accessed") {
           for {
-            fiberRef    <- FiberRef.make(initial)
-            handle      <- fiberRef.unsafeHandle
-            fallbackRef <- UIO(new AtomicReference(update1))
-            resRef      <- UIO(new AtomicReference("" -> ""))
+            fiberRef <- FiberRef.make(initial)
+            handle   <- fiberRef.unsafeAsThreadLocal
+            resRef   <- UIO(new AtomicReference(("", "", "")))
 
             unsafelyGetSetGet = new Runnable {
               def run(): Unit = {
-                val v1 = handle.unsafeGet(fallbackRef.get())
-                handle.unsafeSet(update2, fallbackRef.set)
-                val v2 = handle.unsafeGet(fallbackRef.get())
-                resRef.set(v1 -> v2)
+                val v1 = handle.get()
+                handle.set(update2)
+                val v2 = handle.get()
+                handle.remove()
+                val v3 = handle.get()
+                resRef.set((v1, v2, v3))
               }
             }
 
+            _      <- fiberRef.set(update1)
             thread <- UIO(new Thread(unsafelyGetSetGet))
             _      <- UIO(thread.start()).ensuring(UIO(thread.join()))
 
-            value0           <- fiberRef.get
-            values           <- UIO(resRef.get())
-            (value1, value2) = values
-          } yield assert((value0, value1, value2), equalTo((initial, update1, update2)))
+            value0                   <- fiberRef.get
+            values                   <- UIO(resRef.get())
+            (value1, value2, value3) = values
+          } yield assert((value0, value1, value2, value3), equalTo((update1, initial, update2, initial)))
         }
       )
     ) {
--- a/core/shared/src/main/scala/zio/FiberRef.scala
+++ b/core/shared/src/main/scala/zio/FiberRef.scala
@@ -16,6 +16,9 @@
 
 package zio
 
+import zio.FiberRef.UnsafeHandle
+import zio.internal.FiberContext
+
 /**
  * Fiber's counterpart for Java's `ThreadLocal`. Value is automatically propagated
  * to child on fork and merged back in after joining child.
@@ -46,7 +49,7 @@ package zio
  * @tparam A
  */
 final class FiberRef[A] private[zio] (private[zio] val initial: A, private[zio] val combine: (A, A) => A)
-    extends Serializable {
+    extends Serializable { self =>
 
   /**
    * Reads the value associated with the current fiber. Returns initial value if
@@ -104,9 +107,68 @@ final class FiberRef[A] private[zio] (private[zio] val initial: A, private[zio] 
     (result, result)
   }
 
+  /**
+   * Returns a handle that can be used to interact with this `FiberRef` from side effecting code.
+   *
+   * This feature is meant to be used for integration with side effecting code, that needs to access fiber specific data,
+   * like MDC contexts and the like.
+   *
+   * Important: Enable `Platform.fiberContextPropagation` if you want to use this.
+   */
+  final def unsafeHandle: UIO[UnsafeHandle[A]] =
+    ZIO.effectSuspendTotalWith { platform =>
+      if (platform.fiberContextPropagation) ZIO.unit
+      else ZIO.dieMessage("Unsafe handles require Platform.fiberContextPropagation to be enabled.")
+    } *> ZIO.effectTotal {
+      new UnsafeHandle[A](initial) {
+        def unsafeGet(fallback: => A): A = {
+          val fiberContext = FiberContext.current.get()
+
+          Option {
+            if (fiberContext eq null) null
+            else fiberContext.fiberRefLocals.get(self)
+          }.map(_.asInstanceOf[A]).getOrElse(fallback)
+        }
+
+        def unsafeSet(a: A, fallback: A => Unit): Unit = {
+          val fiberContext = FiberContext.current.get()
+          val fiberRef     = self.asInstanceOf[FiberRef[Any]]
+
+          if (fiberContext eq null) fallback(a)
+          else fiberContext.fiberRefLocals.put(fiberRef, a)
+
+          ()
+        }
+      }
+    }
 }
 
 object FiberRef extends Serializable {
+
+  /**
+   * A handle that allows interaction with fiber refs from side effecting code
+   */
+  abstract class UnsafeHandle[A] private[FiberRef] (initial: A) {
+
+    /**
+     * Unsafely reads from the associated `FiberRef`
+     *
+     * @param fallback a fallback that is used if reading from the `FiberRef` is not possible, for example because the ZIO
+     *                 interpreter has already exited, or because the handle is read from a thread not managed by ZIO.
+     */
+    def unsafeGet(fallback: => A = initial): A
+
+    /**
+     * Unsafely writes to the associated `FiberRef`
+     *
+     * @param fallback a fallback that is used if writing to the `FiberRef` is not possible, for example because the ZIO
+     *                 interpreter has already exited, or because the handle is accessed from a thread not managed by ZIO.
+     */
+    def unsafeSet(
+      a: A,
+      fallback: A => Unit = _ => throw new IllegalStateException("Cannot set fiber ref via unsafe handle")
+    ): Unit
+  }
 
   /**
    * Creates a new `FiberRef` with given initial value.
--- a/core/shared/src/main/scala/zio/FiberRef.scala
+++ b/core/shared/src/main/scala/zio/FiberRef.scala
@@ -16,8 +16,6 @@
 
 package zio
 
-import zio.internal.FiberContext
-
 /**
  * Fiber's counterpart for Java's `ThreadLocal`. Value is automatically propagated
  * to child on fork and merged back in after joining child.
@@ -112,17 +110,12 @@ final class FiberRef[A] private[zio] (private[zio] val initial: A, private[zio] 
    * This feature is meant to be used for integration with side effecting code, that needs to access fiber specific data,
    * like MDC contexts and the like. The returned `ThreadLocal` will be backed by this `FiberRef` on all threads that are
    * currently managed by ZIO, and behave like an ordinary `ThreadLocal` on all other threads.
-   *
-   * Important: Enable `Platform.fiberContextPropagation` if you want to use this.
    */
   final def unsafeAsThreadLocal: UIO[ThreadLocal[A]] =
-    ZIO.effectSuspendTotalWith { platform =>
-      if (platform.fiberContextPropagation) ZIO.unit
-      else ZIO.dieMessage("Unsafe handles require Platform.fiberContextPropagation to be enabled.")
-    } *> ZIO.effectTotal {
+    ZIO.effectTotal {
       new ThreadLocal[A] {
         override def get(): A = {
-          val fiberContext = FiberContext.current.get()
+          val fiberContext = Fiber._currentFiber.get()
 
           Option {
             if (fiberContext eq null) null
@@ -131,7 +124,7 @@ final class FiberRef[A] private[zio] (private[zio] val initial: A, private[zio] 
         }
 
         override def set(a: A): Unit = {
-          val fiberContext = FiberContext.current.get()
+          val fiberContext = Fiber._currentFiber.get()
           val fiberRef     = self.asInstanceOf[FiberRef[Any]]
 
           if (fiberContext eq null) super.set(a)
@@ -141,7 +134,7 @@ final class FiberRef[A] private[zio] (private[zio] val initial: A, private[zio] 
         }
 
         override def remove(): Unit = {
-          val fiberContext = FiberContext.current.get()
+          val fiberContext = Fiber._currentFiber.get()
           val fiberRef     = self.asInstanceOf[FiberRef[Any]]
 
           if (fiberContext eq null) super.remove()
--- a/core/shared/src/main/scala/zio/FiberRef.scala
+++ b/core/shared/src/main/scala/zio/FiberRef.scala
@@ -16,7 +16,6 @@
 
 package zio
 
-import zio.FiberRef.UnsafeHandle
 import zio.internal.FiberContext
 
 /**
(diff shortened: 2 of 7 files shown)


jdegoes:
@mlangc How about we just add `asThreadLocal` to `FiberRef`? Which is guaranteed to be propagated if and only if invoked in some region embedded into a ZIO effect?

mlangc:
You mean adding
```scala
def asThreadLocal: UIO[ThreadLocal[A]]
```
to `FiberRef`, that returns a `ThreadLocal` that always stays in sync with the associated `FiberRef`? To make this more precise, I would expect the following specifications to pass:
```scala
def updateToRefIsVisibleToThreadLocal =
  for {
    fiberRef    <- FiberRef.make(initial)
    threadLocal <- fiberRef.asThreadLocal
    _           <- fiberRef.set(update)
    value       <- UIO(threadLocal.get())
  } yield value must_== update

def updatesToThreadLocalIsVisibleToRef =
  for {
    fiberRef    <- FiberRef.make(initial)
    threadLocal <- fiberRef.asThreadLocal
    _           <- UIO(threadLocal.set(update))
    value       <- fiberRef.get
  } yield value must_== update

def threadLocalKeepsInSyncWhenFiberSwitchesThreads =
  for {
    fiberRef    <- FiberRef.make(initial)
    threadLocal <- fiberRef.asThreadLocal
    _           <- fiberRef.set(update).lock(exec1)
    value       <- UIO(threadLocal.get()).lock(exec2)
  } yield value must_== update

def threadLocalKeepsInSyncWhenThreadSwitchesFiber =
  for {
    threadLocal <- fiberRef.asThreadLocal
    fiber1 <- (fiberRef.set(update) *> UIO(threadLocal.get())).lock(exec1).fork
    fiber2 <- UIO(threadLocal.get()).lock(exec1).fork
    value1 <- fiber1.join
    value2 <- fiber2.join
  } yield (value1 must_== update) and (value2 must_== initial)
```
If this is what you have in mind I definitely like the idea very much, and I think that it should be possible to implement fiber aware MDC logging on top of that too.

core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
```diff
@@ -0,0 +1,45 @@
+package zio
+
+import java.util.concurrent.Executors
+
+import zio.internal.Executor
+
+import scala.concurrent.ExecutionContext
+
+class FiberRefSpecJvm extends BaseCrossPlatformSpec {
+  override def is = "FiberRefSpec".title ^ s2"""
```
mlangc: I put the executor related tests into a separate, JVM only spec, because I had trouble to make this compile with ScalaJS (they don't have `java.util.concurrent.Executors`). If there is a better way, let me know.

jdegoes:
> If this is what you have in mind I definitely like the idea very much, and I think that it should be possible to implement fiber aware MDC logging on top of that too.

This is what I have in mind. I know how to do it, too, but can't help till the weekend at earliest.

mlangc:
I've just pushed an implementation of your proposal, that passes the tests scenarios mentioned above (and a few more).

mlangc:
@jdegoes I'm not in a hurry, so having some feedback on the weekend is fine. I can also ask around in the Gitter channel, to get some additional feedback.

hanny24:
I'm actually not sure that getting `ThreadLocal` out of `FiberRef` is very useful. AFAIK MDC uses its own `ThreadLocal` that can't be changed without nasty hacks. Even log4j implementation requires some sort of global variable.

What I would suggest is something like this:
```scala
trait FiberRefCallback[T]{
  def get: Option[T]
  def set(v: T): Unit
  def remove: Unit
}
```

and also add method to `FiberRef`:
```scala
def withCallback(callback: FiberRefCallback[T]): UIO[Unit] 
```

I'm not really sure if I like it as `FiberRef` has somehow pure API, while `FiberRefCallback` is completely unpure. But it would allow users some sort of flexibility regarding integration with other  libraries.

mlangc:
@hanny24 I think it should work at least with log4j2 without major hacks by implementing a custom [ThreadContextMap](https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ThreadContextMap_implementations), but maybe I'm mistaken. I'm currently working on a proof of concept based on the ZIO version from this PR. I'll post my findings here.

As for the callback: I think we would rather need
```scala
trait FiberRefCallback[T]{
  // To be informed in side effecting code about events from ZIO
  def onFiberEnter(t: T): Unit
  def onSet(t: T): Unit
  def onFiberExit(t: T): Unit
 
  // To update the value from side effecting code
  def set(t: T): T
}
```
Something like this might make the integration more straight forward indeed. On the other hand, it would embed arbitrary side effects, and therefore arbitrary ways to fail or kill performance, into every operation regarding fiber locals as well as the fiber life cycle. All the methods on `FiberRef` that currently return `UIO` would than have to be changed into returning `Task`, either in place, or preferably by creating a separate class, that is by changing your signature from above to
```scala
def withCallback(callback: FiberRefCallback[T]): UIO[FiberRefWithCallback[T]] 
```
Considering all this, I'm still in favor of the `ThreadLocal` approach, if I manage to implement a working [ThreadContextMap](https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ThreadContextMap_implementations) with it.

mlangc:
I have a fiber aware, proof of concept, [ThreadContextMap](https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ThreadContextMap_implementations) implementation based on the code of this PR, that seems to work and that I'll push to Github soon.

Apart from that, the following points still need addressing:
- [ ] The current tests where not able to detect various issues I ran into when doing the POC [ThreadContextMap](https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ThreadContextMap_implementations) implementation. All corner cases, especially involving asynchronous boundaries, should be checked.
- [ ] There is a race condition in `FiberRef.asThreadLocal`in my opinion, that could be fixed by using a `Semaphore`
- [ ] Someone with a deep understanding of the interpreter loop should take a careful look at the changes I've made, and especially at the places where I call `syncFiberLocalsToThreadLocals()` and `syncThreadLocalsToFiberLocals`. I'm almost sure that I've forgot something.
- [ ] Performance:
  - [ ] Currently everybody pays for the feature. Maybe the implementation should be changed, so that there is almost no performance impact when `asThreadLocal` is not used.
  - [ ] Maybe we should also do some benchmarks, to see in how far the feature impacts performance
- [ ] Last but not least: The documentation should be updated too!

hanny24:
> Something like this might make the integration more straight forward indeed. On the other hand, it would embed arbitrary side effects, and therefore arbitrary ways to fail or kill performance, into every operation regarding fiber locals as well as the fiber life cycle. All the methods on `FiberRef` that currently return `UIO` would than have to be changed into returning `Task`, either in place, or preferably by creating a separate class...

I believe it's not only methods on `FiberRef`, but many more methods directly on `ZIO`. On the other hand, we could treat those exceptions as a defect and let he fiber crash. Not really sure if that is a good idea though. 

> Considering all this, I'm still in favor of the ThreadLocal approach, if I manage to implement a working ThreadContextMap with it.

What still really concerns is rather limited usage. It might work for log4j, but it might fail for others such as [Logback](https://logback.qos.ch/) (it might not be the case; just an example)

mlangc:
If you are willing to get your hands dirty by using reflection, the `FiberRef.asThreadLocal` based approach should work with every logging framework supported by `slf4j`, by using a hack similar to the one described here: https://olegpy.com/better-logging-monix-1/

As for the arbitrary behavior/failures that could be injected into `ZIO` (and you are right - it would not be limited to `FiberRef` itself) by the callback: In the worst case, a problem in the logging framework would could then kill fibers randomly. Thus, I'm not sure if I have a good feeling about this.

mlangc:
Obviously I still have a race condition in the code somewhere :-/. The basic idea if my implementation of `FiberRef.asThreadLocal` is to update the associated `ThreadLocal` whenever the `FiberRef` is updated (see https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR525). To sync changes from the `ThreadLocal` back to the `FiberRef`, I always check if there is a value in the `ThreadLocal` before reading from the map of fiber refs, and prefer it if available. Before the fiber stops executing, I sync `FiberContext#fiberRefLocals` with the values in the associated thread locals (see https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR632), and before the fiber starts executing, I make sure that the thread locals contain the correct values from `FiberContext#fiberRefLocals` (see https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR624). Unfortunately, due to my limited understanding of the interpreter, there seems to be a data race in my implementation, that I cannot reproduce locally at the moment. I'll continue to look into this, but maybe someone with a deeper understanding of the interpreter can give me a hint?

jdegoes:
I am happy to help shortly.

I think the callback idea is worth considering, because we can implement `asThreadLocal` in terms of that; therefore it's more foundational. In addition, it could push out "thread local" -specific code from the interpreter. 

We can try / catch those callbacks and report any errors to the `Platform`. I even think we could make it so the callbacks are pure (`UIO`) and only have users pay for the cost if they use it.

hanny24:
> I always check if there is a value in the `ThreadLocal` before reading from the map of fiber refs, and prefer it if available. 

Hm, this kind behavior of behavior might be considered as very surprising. And I would even say it's incorrect. Suppose this:
```scala
for  {
  // assuming `fiberRef` get propagated to ThreadLocal `threadLocal`
  _ <- fiberRef.set("tmp")
  _ <- ZIO.effectTotal(threadLocal.set("tmp2"))
 _ <- fiberRef.set("tmp3")
 v <- fiberRef.get()
} yield v
```
If I understood you correctly, `v == "tmp2"` which does not seem right.

I think that in order to propagate changes back to `FiberRef`, we would need to propagate value after every `ZIO#effect*` method. I think that it would make propagation very costly, even though it might still be possible to implement it in a way that it affects your performance only when you use this feature.

mlangc:
> If I understood you correctly, v == "tmp2" which does not seem right.

No, you should get `v == "tmp3"` (see these tests: https://github.com/zio/zio/pull/1341/files#diff-b3f6f229206b73e705ea7e4c6e03784bR20), because `fiberRef.set("tmp3")` comes after `ZIO.effectTotal(threadLocal.set("tmp2"))` and the synchronization goes in both ways (that is writing to the `FiberRef` changes the `ThreadLocal` and the other way around).

> I think that in order to propagate changes back to FiberRef, we would need to propagate value after every ZIO#effect* method. I think that it would make propagation very costly, even though it might still be possible to implement it in a way that it affects your performance only when you use this feature.

The propagation as it is implemented now is not that costly, as it adds overhead only when a `FiberRef` is modified or read, and before certain fiber live cycle events (start, finish, async boundaries) - but then there is still a race condition that needs to be sorted out, so everything I say about the current implementation should be taken with a heavy grain of salt. I think that the implementation could be further adapted, so that the overhead becomes negligible if the feature is not used, and quite low otherwise.

jdegoes:
>  I think that the implementation could be further adapted, so that the overhead becomes negligible if the feature is not used, and quite low otherwise.

All right, will take a closer look at this as soon as I can (hopefully later next week / weekend).

mlangc:
I think that I've finally hunted down the race conditions I've been struggling with (see https://github.com/zio/zio/pull/1341/files#diff-d59b7265242698fc77688bc58cf0ac52R68). Looking at the implementation of `MirroringThreadLocal` (see https://github.com/zio/zio/pull/1341/files#diff-80bec549c1f5ec11d92e81f121ee8864R136), that isn't a real thread local anymore, I think that it would be better to define `getUnsafeHandle: UIO[UnsafeHandle[A]]` together with `FiberRef.UnsafeHandle` more or less like
```scala
trait UnsafeHandle[A] {
  def get(): A
  def set(a: A): Unit
}
```
The implementation of this trait would be quite similar to `MirroringThreadLocal`, but the API would be much cleaner.

If we decide to go in the direction of callbacks, what about making it possible to have notifications about the fiber lifecycle, via an interface like `Fiber.Observer`
```scala
trait Observer {
   def onEnter(id: FiberId, parendId: Option[FiberId]): UIO[Unit]
   def onSupend(id: FiberId): UIO[Unit]
   def onExit(id: FiberId): UIO[Unit]
}
```
I guess that we could implement fiber local storage for side effecting code, and even `FiberRef` itself, on top of that.

jdegoes:
> I guess that we could implement fiber local storage for side effecting code, and even FiberRef itself, on top of that.

I like any ideas that allow us to pull `FiberRef` out of core, in exchange for something as small or smaller; especially if we gain the ability to do other things, as well.

mlangc:
I've now refactored the code as outlined above from `asThreadLocal` to `getUnsafeHandle` and added another test. Apart from performance improvements and some documentation updates, this code could be ready for a review and hopefully merging soon.

Looking at the direction of the discussion, I'm not sure though that investing more time into this approach is worth the effort. There are two possibilities that I currently see:

1. We could try to get this PR merged and explore more general solutions (like `Fiber.Observer`) after that.
2. We could abandon this PR, and start exploring right away.

I'm OK with both approaches, if I have a solution for fiber aware MDC logging in the not too distant future. @jdegoes what is your take on this?

mlangc:
I've just resynchronized with master and squashed most of the stuff into one commit. Also note that this PR is dependent on #1341.

As for the discussion from above: I've experimented quite a bit with implementing support for hooks into the ZIO fiber life cycle for side effecting code via 
```scala
object Fiber {
  trait Observer {
    def onStart(id: FiberId, parentId: Option[FiberId]): Unit = ()
    def onExit(id: FiberId): Unit = ()
    def onSuspend(id: FiberId): Unit = ()
    def onResume(id: FiberId): Unit = ()
    def onFork(id: FiberId): Unit = ()
    def onJoin(id: FiberId, childId: FiberId): Unit = ()
  }
}
```
See https://github.com/mlangc/zio/tree/add-fiber-observers for details.

However, it turned out that enforcing guaranties that would make these observers actually useful (like `onExit` being the last call, `onSuspend` always called after `onResume` on the same thread, etc.) is far from easy and requires non trivial changes to the interpreter with potential performance impacts. I'm therefore not sure if this is really worth the effort, and rather stay with the `UnsafeHandle` based solution implemented in this PR. Apart from performance related improvements, there is one limitation with the current implementation: If `inheritFiberRefs` is called on a fiber before it exits, some changes made to the `FiberRef` via the `UnsafeHandle` might not get inherited. I have an idea on how to fix this, but it would introduce an asynchronous boundary into `inheritFiberRefs`.

mlangc:
Dear reviewers,

It is clear to me that this cannot be merged yet because of potential performance impacts. However, before I invest a considerable amount of time into benchmarking and optimizing, I really need your feedback in order to know if you agree with what I have so far in general and if it makes sense to continue working on this.

@jdegoes @iravid @adamgfraser @ghostdogpr @NeQuissimus @neko-kai  can anyone take a look at this in order to give me some feedback? If you are interested in theory, but don't want to go through the whole thread, just ping me, so that I can assembly a brief summary.

jdegoes:
@mlangc I've not abandoned this! I'm really thinking in detail here about the right way forward. I hope to have more comments soon.

mlangc:
I've just reduced `UnsafeHandle` to a read only view of the underlying `FiberRef`, simplifying the whole implementation considerably. Additionally, I've pushed a proof of concept implementation of fiber aware MDC logging for `log4j2` based on this PR (see https://github.com/mlangc/zio-interop-log4j2).

Note that in order to synchronize the unsafe handles, there is a loop over all `FiberRefs` at the start and the end of `evaluateNow` (see https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR239 and https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR546). I've not yet done any benchmarks, but thus would definitely have noticeable performance impacts if number of `FiberRefs` is large enough. I could try to change the implementation, so that only fiber refs where unsafe handles have been created are part of the iteration, at the cost of increased implementation complexity.

@jdegoes what is your take on this?

mlangc:
The last commit adapts the implementation of unsafe handles, so that unless they are used, the performance impact is only two empty checks against a synchronized map in `FiberContext.evaluateNow`. On the JVM, an additional optimization makes sure, that the performance impact increases linearly with the number of unsafe handles, and not linearly
with the number of `FiberRef`s. This optimization is not available for JS, since it is based on a `WeakHashMap`.

I could not measure any negative performance impact with these benchmarks:

```
jmh:run -i 20 -wi 4 -f 1 -r 1 -w 1 .*\.FiberRefBenchmarks
jmh:run -i 10 -wi 3 -f 1 -r 10 -w 5 .*\.ParallelMergeSortBenchmark\.zioSort$
jmh:run -i 10 -wi 4 -f 1 -r 1 -w 1 .*\.BubbleSortBenchmarks\.scalazBubbleSort
jmh:run -i 10 -wi 3 -f 1 -r 1 -w 1 .*\.IODeepFlatMapBenchmark\.scalazDeepFlatMap$
jmh:run -i 20 -wi 3 -f 1 -r 1 -w 1 .*\.IOForkInterruptBenchmark\.zioForkInterrupt$
```

core/shared/src/main/scala/zio/FiberRef.scala
```diff
@@ -103,9 +109,48 @@ final class FiberRef[A](private[zio] val initial: A, private[zio] val combine: (
     (result, result)
   }
 
+  /**
+   * Returns a handle that can be used to read the value of this `FiberRef` from side effecting code.
+   *
+   * This feature is meant to be used for integration with side effecting code, that needs to access fiber specific data.
+   */
+  final def getUnsafeHandle: UIO[UnsafeHandle[A]] =
```
jdegoes: Should be called `unsafeHandle` by ZIO conventions.

core/shared/src/main/scala/zio/FiberRef.scala
```diff
@@ -103,9 +109,48 @@ final class FiberRef[A](private[zio] val initial: A, private[zio] val combine: (
...
+      }
+
+      if (!threadLocalRef.compareAndSet(null, newCandidate)) getThreadLocal
+      else (newCandidate, true)
+    }
+  }
 }
 
 object FiberRef extends Serializable {
+  trait UnsafeHandle[A] {
```
jdegoes: The `get` should be renamed `unsafeGet` by ZIO conventions.

Also I'm not sure we really need it, since we could just return `Function0[A]` from `unsafeHandle: UIO[Function0[A]]`.

core/shared/src/main/scala/zio/FiberRef.scala
```diff
@@ -103,9 +109,48 @@ final class FiberRef[A](private[zio] val initial: A, private[zio] val combine: (
...
+          if (!isNew) ZIO.succeed(handle)
+          else
+            ZIO.effectTotal {
+              UnsafelyExposedFiberRefs.register(this)
+              handle
+            }
+      }
+    }
+
+  private[this] def getThreadLocal: (ThreadLocal[A], Boolean) = {
```
jdegoes: `@tailrec`

core/shared/src/main/scala/zio/FiberRef.scala
mlangc: I've renamed ` get` to `unsafeGet` as proposed. I'm also fine with `UIO[Function0[A]]`, but I'm slightly more inclined to having a separate trait, as at makes things more explicit.

mlangc:
@jdegoes I've addressed the issues you found. Please let me know if there is anything else.

mlangc:
@jdegoes your suggestion about passing the currently executing fiber to the implementation that unsafely updates it, brought me on the right path for implementing `unsafeSet`.

The new implementation uses only one, global `ThreadLocal`, to store the currently executing `FiberContext`, namely [FiberContext.current](https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR828). This not only greatly simplifies the code for `unsafeGet`, it also made implementing `unsafeSet` easily possible.

Since setting and unsetting a `ThreadLocal` for every call to `FiberContext.evaluateNow` has a performance impact in some situations, there is [a guard](https://github.com/zio/zio/pull/1341/files#diff-b53adcb54f013ee1630f875cdc6d23afR216), that makes sure this is only done when unsafe handles are in use. It's also because of this optimization, that I had to add the new, private [SetCurrentFiberContext](https://github.com/zio/zio/pull/1341/files#diff-9cfb50d60ba29ffb4a6a15c933dbf2c3R2873) command. We might consider changing this to `GetCurrentFiberContext`, so that the command is more widely usable, while still keeping it private.

I've not yet done any detailed benchmarking, but in theory, the impact of the feature if not used, are just two empty checks (I can most likely get rid of one) per `evaluateNow` call.

jdegoes:
@mlangc I like this approach much better.

Now I suggest we change the API to `unsafeAsThreadLocal: UIO[ThreadLocal[A]]`. Then we can return `null` if there is nothing, as per Java convention.

jdegoes:
@mlangc How much more work before ready to review?

mlangc:
@jdegoes There are still a few issues to be sorted out if we don't want to to have one call to `ThreadLocal.set` and one to `ThreadLocal.remove` at each evocation of `evaluateNow` even when the feature is not used at all. My plan is to have something that is ready to be reviewed in the next two weeks, hopefully sooner.

mlangc:
@jdegoes I dug into this some more, and finally decided to remove all the code that attempted to be smart by setting [FiberContext.current](https://github.com/zio/zio/pull/1341/commits/e9d9c71afe45ba20f72c03dec8255218de7e9d1a#diff-b53adcb54f013ee1630f875cdc6d23afR218)` only as needed, as this added quite some complexity, without even handling all corner cases.

With the latest version, the user has to enable the feature manually, using [Platform#fiberContextPropagation](https://github.com/zio/zio/pull/1341/commits/e9d9c71afe45ba20f72c03dec8255218de7e9d1a#diff-88fdb92ec07ae104f0a820d39bd60b48R98) (defaults to false). The only thing that I don't particularly like, is that currently, `FiberRef#unsafeHandle` dies if `fiberContextPropagation` is not enabled. If we only could turn that into a compile error...

Apart from that, I've updated `FiberRef.UnsafeHandle` so that the user can pass in [fallback functions](https://github.com/zio/zio/pull/1341/commits/e9d9c71afe45ba20f72c03dec8255218de7e9d1a#diff-d59b7265242698fc77688bc58cf0ac52R19), that are invoked when the current `FiberContext` cannot be accessed (for example because the handle is accessed from a thread not managed by ZIO).

To cut a long story short: A review would be great!

jdegoes:
@mlangc I will try to do this tomorrow. ZIO Hackathon is keeping me quite busy!

core/shared/src/main/scala/zio/FiberRef.scala
```diff
@@ -103,10 +109,92 @@ final class FiberRef[A](private[zio] val initial: A, private[zio] val combine: (
...
 
+  /**
+   * Returns a handle that can be used to interact with this `FiberRef` from side effecting code.
+   *
+   * This feature is meant to be used for integration with side effecting code, that needs to access fiber specific data,
+   * like MDC contexts and the like.
+   *
+   * Important: Enable `Platform.fiberContextPropagation` if you want to use this.
+   */
+  final def unsafeHandle: UIO[UnsafeHandle[A]] =
```
jdegoes: I think we need to change this to `UIO[ThreadLocal[A]]`. We can implement all methods:

 - `get` (`unsafeGet` in `UnsafeHandle`)
 - `set` (`unsafeSet` in `UnsafeHandle`)
 - `initialValue` (the private initial value in `FiberRef`)
 - `remove` (same as `set(null)`?)

`ThreadLocal` is widely understood by existing Java / Scala developers who would be most likely to use this feature. We can piggyback atop their expectations of how it should work.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -38,7 +39,7 @@ private[zio] final class FiberContext[E, A](
   startSStatus: SuperviseStatus,
   parentTrace: Option[ZTrace],
   initialTracingStatus: Boolean,
-  fiberRefLocals: FiberRefLocals
+  private[zio] val fiberRefLocals: FiberRefLocals
```
jdegoes: If this is a global variable, as it looks, we should just make it a singleton to avoid increasing memory consumption by the cost of a reference (per fiber).

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -510,12 +516,19 @@ private[zio] final class FiberContext[E, A](
...
+                    .popPendingUnsafeSet(zio.fiberRef)
+                    .orElse(Option(fiberRefLocals.get(zio.fiberRef)))
+
                   val (result, newValue) = zio.f(oldValue.getOrElse(zio.fiberRef.initial))
                   fiberRefLocals.put(zio.fiberRef, newValue)
 
                   curZio = nextInstr(result)
 
+                case ZIO.Tags.SetCurrentFiberContext =>
+                  FiberContext.current.set(this)
```
jdegoes: Why do we need this, since if `fiberContextPropagation` is set, then `FiberContext.current.set(this)` is run at the start of the method `evaluateNow`, and of course will not change until the end.

core/shared/src/main/scala/zio/FiberRef.scala
mlangc: Fair enough, I'll look into that over the weekend

core/shared/src/main/scala/zio/internal/FiberContext.scala
mlangc: Just is just a field of `FiberContext` as it was before (only `private[zio] val` was added to make it accessible from outside `FiberContext`)

core/shared/src/main/scala/zio/internal/FiberContext.scala
mlangc: Indeed, this is just a leftover from an older implementation that tried to be "smart". I removed it.

core/shared/src/main/scala/zio/FiberRef.scala
mlangc: I've just dropped `FiberRef.unsafeHandle` in favour of `FiberRef.unsafeAsThreadLocal` and adapted the implementation as outlined by you above.

mlangc:
@jdegoes I've addressed all your comments. In addition to that, I ran some benchmarks, and found the following: If `fiberContextPropagation` is disabled, we are adding just two comparisons against a boolean per `evaluateNow`. Even in the most unfavorable scenario (see below for an explanation), the performance impact of this PR kept below 5% in my measurements.

If enabled, the performance impact of `fiberContextPropagation` is negligible in most cases, especially if tracing, that has a much larger impact, is enabled. Even in the worst possible scenario, where tracing is disabled, and we are just yielding in a loop, so that every instruction is mapped to a single `evaluateNow` invocation, the performance penalty incurred by `fiberContextPropagation` is below 10% (see `jmh:run -i 20 -wi 4 -f 1 -r 1 -w 1 .*\.FiberRefBenchmarks\.unTracedJustYield`). Thus, we might even think about enabling, instead of disabling the feature by default.

mlangc:
@jdegoes I've just rebased and adapted the implementation to use `Fiber._currentFiber`.

Issue #1310, opened 2019-08-01
Add a combinator for parallel composition of sinks


iravid:

iravid:
cc @simpadjo_gitlab

simpadjo:
Thank you, it would be very useful.

@jdegoes 
>  The reason that probably doesn't exist now is because it's not obvious what to do with leftovers: each sink will, in general, have a different amount of leftover; and the leftovers will overlap. So in their composition, what do you do? Probably emit the shortest. All to say, possible but tricky to do correctly.

I had the same question in my head initially. Now I think that returning the shortest one looks pretty much consistent with respect to `ZIIO#zipPar` behavior. If a Stream is a timeline and Sinks are processes then `ZSink#zipPar` considered done at the latest point. Dual to `ZSink#raceBoth`.

jdegoes:
I do think shortest is the best ("minimal leftover"). 

Albeit there will be some nice laws that it can't satisfy.

It'd be nice if you could take `sink1 -> sink2` and `sink3 -> sink4` and `zipPar` `sink1` and `sink3`, and then `zipPar` `sink2` and `sink4`, and then do `(sink1 zipPar sink3) -> (sink2 zipPar sink4)` and have that give you the same result. But Sink appears not to have the structure required to allow that, because it would require tuple leftovers and tuple inputs.

simpadjo:
I'd like to try to implement it

iravid:
Closed in #1344

PR #1344, opened 2019-08-04
Adding `ZSink#zipPar`

simpadjo:
https://github.com/zio/zio/issues/1310

Added a combinator for aggregating a stream into two sinks at the same time. 
Please review my approach. I'll add more tests and squash commits afterwards.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -8,6 +8,8 @@ import zio.clock.Clock
 import zio.duration._
 import zio.test.mock.MockClock
 import java.util.concurrent.TimeUnit
+import org.specs2.matcher.MatchResult
+import org.specs2.matcher.describe.Diffable
 
 class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     extends TestRuntime
@@ -167,6 +169,17 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     zipLeft (<*)
       happy path $zipLeftHappyPath
 
+    zipPar
+       happy path 1 $zipParHappyPathBothDone
+       happy path 2 $zipParHappyPathOneNonterm
+       happy path 3 $zipParHappyPathBothNonterm
+       extract error $zipParErrorExtract
+       step error $zipParErrorStep
+       init error $zipParErrorInit
+       both error $zipParErrorBoth
+       remainder corner case 1 $zipParRemainderWhenCompleteSeparately
+       remainder corner case 2 $zipParRemainderWhenCompleteTogether
+
     zipRight (*>)
       happy path $zipRightHappyPath
 
@@ -236,6 +249,31 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     val initial                    = UIO.succeed(Step.more(()))
     def step(state: State, a: Int) = UIO.succeed(Step.done((), Chunk.empty))
     def extract(state: State)      = IO.fail("Ouch")
+  }
+
+  /** Searches for the `target` element in the stream.
+   * When met - accumulates next `accumulateAfterMet` elements and returns as `leftover`
+   * If `target` is not met - returns `default` with empty `leftover`
+   */
+  private def sinkWithLeftover[A](target: A, accumulateAfterMet: Int, default: A) = new ZSink[Any, String, A, A, A] {
+    override type State = Option[List[A]]
+
+    override def extract(state: Option[List[A]]): ZIO[Any, String, A] =
+      UIO.succeed(if (state.isEmpty) default else target)
+
+    override def initial: ZIO[Any, String, Step[Option[List[A]], Nothing]] = UIO.succeed(Step.more(None))
+
+    override def step(state: Option[List[A]], a: A): ZIO[Any, String, Step[Option[List[A]], A]] =
+      state match {
+        case None =>
+          val st = if (a == target) Some(Nil) else None
+          UIO.succeed(Step.more(st))
+        case Some(acc) =>
+          if (acc.length >= accumulateAfterMet)
+            UIO.succeed(Step.done(state, Chunk.fromIterable(acc)))
+          else
+            UIO.succeed(Step.more(Some(acc :+ a)))
+      }
   }
 
   private def sinkIteration[R, E, A0, A, B](sink: ZSink[R, E, A0, A, B], a: A) =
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1289,3 +1387,15 @@ object ZSink extends ZSinkPlatformSpecific {
     ZManaged.fromEffect(sink)
   }
 }
+
+/*
+object ASDF extends App {
```
simpadjo: Oh, sorry :)

simpadjo:
@iravid could you please take a look?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+                      val leftover: Chunk[A2] = (Step.leftover(st))
+                      Left((c, Some(leftover)))
+                    })
+                }
+              })
+        )
+
+        firstResult.flatMap(r1 => {
+          secondResult.map(r2 => {
+            r1 match {
```
iravid: Pattern matching on `(r1, r2)` could possibly make this nicer to read

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
     }
 
+  final def zipPar[R1 <: R, E1 >: E, A2 >: A0, A1 <: A, C](
+    that: ZSink[R1, E1, A2, A1, C]
+  ): ZSink[R1, E1, A2, A1, (B, C)] =
+    new ZSink[R1, E1, A2, A1, (B, C)] {
+      type State = (Either[B, self.State], Either[C, that.State])
+
+      override def extract(state: State): ZIO[R1, E1, (B, C)] = {
+        val b: ZIO[R, E, B]   = state._1.fold(ZIO.succeed, s1 => self.extract(s1))
```
iravid:
```suggestion
        val b: ZIO[R, E, B]   = state._1.fold(ZIO.succeed, self.extract)
```

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        val c: ZIO[R1, E1, C] = state._2.fold(ZIO.succeed, s2 => that.extract(s2))
```
iravid:
```suggestion
        val c: ZIO[R1, E1, C] = state._2.fold(ZIO.succeed, that.extract)
```

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        b.zip(c)
+      }
+
+      override def initial: ZIO[R1, E1, Step[State, Nothing]] =
+        self.initial.flatMap(s1 => {
```
iravid: 2 minor notes: 
- these types of blocks usually read nicer as `self.initial.flatMap { s1 =>`
- whenever you have multiple consecutive `flatMap`s, those usually read nicer as a `for` comprehension

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+          that.initial.flatMap(s2 => {
+            (Step.cont(s1), Step.cont(s2)) match {
+              case (false, false) => {
```
iravid: Another minor note: the bodies of `case` blocks don't need braces around them

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+              }))
+        )
+
+        val secondResult: ZIO[R1, E1, Either[(C, Option[Chunk[A2]]), that.State]] = state._2.fold(
+          c => ZIO.succeed(Left((c, None))),
+          s =>
+            that
+              .step(s, a)
+              .flatMap(st => {
+                if (Step.cont(st)) {
```
iravid: When your `if/else` block body consists of just one expression, you could drop the braces around it as well

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,114 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        )
+
+        firstResult.flatMap(r1 => {
+          secondResult.map(r2 => {
+            r1 match {
+              case Left((b, rem1)) => {
+                r2 match {
+                  case Left((c, rem2)) => {
+                    val minLeftover =
+                      if (rem1.isEmpty && rem2.isEmpty) Chunk.empty else (rem1.toList ++ rem2.toList).minBy(_.length)
```
iravid: Converting the remainders to a list could be potentially costly, so let's do this the old fashioned way by comparing the lengths with an `if`

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -167,6 +167,9 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     zipLeft (<*)
       happy path $zipLeftHappyPath
 
+    zipPar
+       zipPar somehow works $zipParSanity
```
iravid: Here are some additional test cases I suggest we add:
- left sink finishes earlier
- right sink finishes earlier
- both sinks finish at the same time
- left sink errors during step
- left sink errors during extract
- right sink errors during step
- right sink errors during extract

streams/shared/src/main/scala/zio/stream/ZSink.scala
simpadjo: I do agree about using `for` and initially tried to write it like that. But I faced `no method withFilter` error and then continued with `flatMap`s

streams/shared/src/main/scala/zio/stream/ZSink.scala
simpadjo: Maybe it's not obvious here but `rem1` and `rem2` here are `Option[Chunk]`, not `Chunk`. So it's relatively cheap and runs only once. 

The logic is following: if one sink terminated strictly before the other I drop remainder of the former anyway. I compare remainders' length only if sinks finished simultaneously.

But if you still think it's costly - I'll change it.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
simpadjo: Yeah. Must cover cases of non-terminating sinks as well. And also it would be nice to explicitly check the logic of choosing the shorter `leftover` when sinks terminated at the same time.
Haven't found test sinks with non-trivial leftovers to use here. Will invent one.

simpadjo:
@iravid thank you for kind words!

simpadjo:
@iravid I think I'm done with it. 
I found few laws that together more or less specify `zipPar`. But since we can't generate `Sink` I run them manually.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -780,6 +818,116 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     unsafeRun(sinkIteration(sink, 1).map(_ must_=== "1Hello"))
   }
 
+  private object ZipParLaws {
```
iravid: This is excellent, comprehensive work! Big like :-)

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,98 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+  final def zipPar[R1 <: R, E1 >: E, A2 >: A0, A1 <: A, C](
+    that: ZSink[R1, E1, A2, A1, C]
+  ): ZSink[R1, E1, A2, A1, (B, C)] =
+    new ZSink[R1, E1, A2, A1, (B, C)] {
+      type State = (Either[B, self.State], Either[C, that.State])
+
+      override def extract(state: State): ZIO[R1, E1, (B, C)] = {
+        val b: ZIO[R, E, B]   = state._1.fold(ZIO.succeed, self.extract)
+        val c: ZIO[R1, E1, C] = state._2.fold(ZIO.succeed, that.extract)
+        b.zip(c)
```
iravid: Let's use `zipPar` here

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,98 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+      }
+
+      override def initial: ZIO[R1, E1, Step[State, Nothing]] =
+        self.initial.flatMap { s1 =>
+          that.initial.flatMap { s2 =>
+            (Step.cont(s1), Step.cont(s2)) match {
+              case (false, false) =>
+                val zb = self.extract(Step.state(s1))
+                val zc = that.extract(Step.state(s2))
+                zb.zipWith(zc)((b, c) => Step.done((Left(b), Left(c)), Chunk.empty))
```
iravid:
```suggestion
                zb.zipWithPar(zc)((b, c) => Step.done((Left(b), Left(c)), Chunk.empty))
```

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -678,6 +678,98 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+                  that
+                    .extract(Step.state(st))
+                    .map(c => {
+                      val leftover: Chunk[A2] = (Step.leftover(st))
+                      Left((c, Some(leftover)))
+                    })
+              }
+        )
+
+        for {
```
iravid: Let's use `firstResult.zipWithPar(secondResult) { ...` here

PR #1347, opened 2019-08-04
[wip] Cause to disallow sequencial failures as parallel

### LINKED ISSUE #1073: Improve `Cause` to disallow Sequential failures being exposed as parallel
See https://github.com/zio/zio/pull/917#issuecomment-504782408

softinio:
This resolves #1073 

* DO NOT MERGE - WIP *

### DIFF
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -31,7 +31,7 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
   /**
    * Sequentially zips the this result with the specified result discarding the first element of the tuple or else returns the failed `Cause[E1]`
    */
-  final def *>[E1 >: E, B](that: Exit[E1, B]): Exit[E1, B] = zipWith(that)((_, _), _ ++ _).map(_._2)
+  final def *>[E1 >: E, B](that: Exit[E1, B]): Exit[E1, B] = zipWith(that)((_, _), _ && _).map(_._2)
 
   /**
    * Parallelly zips the this result with the specified result discarding the second element of the tuple or else returns the failed `Cause[E1]`
@@ -46,12 +46,12 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
   /**
    * Sequentially zips the this result with the specified result discarding the second element of the tuple or else returns the failed `Cause[E1]`
    */
-  final def <*[E1 >: E, B](that: Exit[E1, B]): Exit[E1, A] = zipWith(that)((_, _), _ ++ _).map(_._1)
+  final def <*[E1 >: E, B](that: Exit[E1, B]): Exit[E1, A] = zipWith(that)((_, _), _ && _).map(_._1)
 
   /**
    * Sequentially zips the this result with the specified result or else returns the failed `Cause[E1]`
    */
-  final def <*>[E1 >: E, B](that: Exit[E1, B]): Exit[E1, (A, B)] = zipWith(that)((_, _), _ ++ _)
+  final def <*>[E1 >: E, B](that: Exit[E1, B]): Exit[E1, (A, B)] = zipWith(that)((_, _), _ && _)
 
   /**
    * Replaces the success value with the one provided.
@@ -210,7 +210,7 @@ object Exit extends Serializable {
     exits.headOption.map { head =>
       exits
         .drop(1)
-        .foldLeft(head.map(List(_)))((acc, el) => acc.zipWith(el)((acc, el) => el :: acc, _ ++ _))
+        .foldLeft(head.map(List(_)))((acc, el) => acc.zipWith(el)((acc, el) => el :: acc, _ && _))
         .map(_.reverse)
     }
 
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -110,7 +110,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Returns an effect whose failure and success channels have been mapped by
    * the specified pair of functions, `f` and `g`.
    */
-  final def bimap[B](f: E => Nothing, g: A => B): ZIO[R, Nothing, B] = mapError(f).map(g)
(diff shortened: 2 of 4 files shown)


softinio:
@jdegoes How would I enforce `Exit.zipWith` to not enforce both fibers have an `E` ?

jdegoes:
@softinio 

We want to allow this:

```scala
(c1 : Cause[Nothing]) ++ (c2: Cause[E])
```

forall `E`.

We also want to allow this:

```scala
(c1 : Cause[E]) ++ (c2: Cause[Nothing])
```

For all `E`. 

This means there will be at most one sequential error `E`. 

Actually, we may be able to lock it down tighter: the failure `E` must always be first, before finalizer causes. So in reality, we only want to allow this:

```scala
(c1 : Cause[E]) ++ (c2: Cause[Nothing])
```

So this becomes a matter of modifying `++`, and the case class that backs it (`Then`) so as to not allow an `E` in the second term. 

That, and fix any compiler errors (which may be bugs!) may be all you have to do, unless I've missed something...

softinio:
@jdegoes your suggestion is exactly what I have done ... do you agree? 

ie `(c1 : Cause[E]) ++ (c2: Cause[Nothing])`

The compilation errors I get are all when `++` is used I get:

```
[error]  found   : zio.Cause[E]
[error]  required: zio.Cause[Nothing]
```
mainly for `zipWith` see for example here: https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/Exit.scala#L34

So what I was in effect asking was how would I fix this compilation error?

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -83,7 +83,7 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       case _                 => false
     }
 
-  final def map[E1](f: E => E1): Cause[E1] = self match {
+  final def map[E1 <: Nothing](f: E => E1): Cause[E1] = self match {
```
jdegoes: This type bound is not correct (there are no subtypes of `Nothing`). I'd delete it.

jdegoes:
> mainly for zipWith see for example here: 
> https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/Exit.scala#L34
> 
> So what I was in effect asking was how would I fix this compilation error?

I think this one should use `&&`, not `++`, because the failure occur in parallel (separate fibers).

Any others, let me know!

softinio:
> > mainly for zipWith see for example here:
> > https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/Exit.scala#L34
> > So what I was in effect asking was how would I fix this compilation error?
> 
> I think this one should use `&&`, not `++`, because the failure occur in parallel (separate fibers).
> 
> Any others, let me know!

But that would make it the same as `&>` line 29

jdegoes:
@softinio True, but errors from different fibers always happen in parallel, never in sequence. Perhaps the `zip` should be removed in favor of `zipPar`, if they are always the same / identical.

softinio:
@jdegoes Making this change one hurdle is how to handle a `Throwable` when its now expecting a `Nothing`, for example: 

https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZManaged.scala#L457

and 

https://github.com/zio/zio/blob/54b3acf13fae8c6501587f60131d2ae5074aee46/test/shared/src/main/scala/zio/test/mock/MockConsole.scala#L55

Any suggestions on what I should do here? 

Will push my latest changes in case off track though not convinced I am right expecting a `Nothing` in lieu of a `Throwable`. 

Your help appreciated. Thank you sir!

neko-kai:
@softinio 
I believe these signatures can't work:

in Cause https://github.com/zio/zio/pull/1347/files#diff-1135e3fcd6165ca11a68e2706ae3450dR86:
```scala
final def map(f: E => Nothing): Cause[Nothing]
```
in ZIO https://github.com/zio/zio/pull/1347/files#diff-ba99038d96380b834448ea7c9ddd3c05R392:
```scala
final def mapError(f: E => Nothing): ZIO[R, Nothing, A]
```

Note that `Nothing` is a type for programs that never finish, in Scala a function can return Nothing only if it throws an exception or loops forever, e.g. `def x(): Nothing = x()`.
There is no other way to return a `Nothing` and there's no way to obtain a value that has a class of `Nothing`, `Nothing` is a fiction of the compiler – it never exists at runtime and can't be returned.
That means that if `.mapError` accepts `E => Nothing`, then a user would only be able to supply a function `(_ => ???)` - and when evaluated that function will error. Effectively `mapError` must always error and cannot successfully return.

That means that whenever Nothing appears as a result of function that means you might want to look for a different way to make it work. Seems like just reverting all the Nothing-related changes makes this PR work, please check the PR I opened for your PR - https://github.com/softinio/zio/pull/1

softinio:
@neko-kai Thanks for your help with this. Valid point about `Nothing`. Merged your PR. Thoughts? Seems I over complicated my thinking of how to solve this problem.

neko-kai:
@softinio Seems like rebasing nuked the changes to Cause 😅

softinio:
> @softinio Seems like rebasing nuked the changes to Cause 😅

Yes ugh!

softinio:
@neko-kai Seems compared to the original Cause we now have a flatMap. What change do I need to make for this? It doesn't work as is? Thanks for direction and help on this.

jdegoes:
After a recent change to handling error composition in `orElse`, I am not sure we can make the strict version of this change anymore. 

The problem is that `orElse` attempts to preserve chronology by capturing the left failure as a `Die(FiberFailure(...))`.

This means that we can at most allow only `Cause[Nothing] ++ Cause[E]` and `Cause[E] ++ Cause[Nothing]`, but we cannot require than `E` be "first" in any sequence; and I wonder what that says about how other operators are implemented.

I think it may be best to close this for a while until we figure out the solution to the `orElse` problem.

What do you think?

neko-kai:
@jdegoes Why not compose the errors in `orElse` as parallel with `&&`?.. While they're not exactly parallel, they're also different from fiber failures that really are always expected to be structured as `(E|Die|Interrupt ++ ...Die*)`

PR #1355, opened 2019-08-05
Add ZStream#interleave, ZStream#interleaveWith, ZStream#combine

adamgfraser:
Adds an `interleave` method on `ZStream` for deterministic merging of streams. Uses a stream of boolean values to determine whether to pull from the left or right stream each time so could be used for simply alternating between two streams, pulling from each stream with different frequency, or using effects to determine which stream to pull from next.

@iravid Would you take a look when you get the chance?

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1402,7 +1402,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
       )
     }
 
-    ZStream.combine(self, that, lc = lc, rc = rc) { (left, right, dest) =>
+    ZStream.combine[R1, E1, A, B, C](self, that, lc = lc, rc = rc) { (left, right, dest) =>
       loop(false, false, left, right, dest)
     }
   }
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -507,6 +507,36 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   }
 
   /**
+   * Combines this stream and the specified stream by converting both streams
+   * to queues and repeatedly applying the function `f0` to extract
+   * an element from the queues and conceptually "offer" it to the destination
+   * stream. `f0` can maintain some internal state to control the combining
+   * process, with the initial state being specified by `s1`.
+   */
+  final def combine[R1 <: R, E1 >: E, A1 >: A, S1, B, C](that: ZStream[R1, E1, B], lc: Int = 2, rc: Int = 2)(
+    s1: S1
+  )(f0: (S1, Queue[Take[E1, A1]], Queue[Take[E1, B]]) => ZIO[R1, E1, (S1, Take[E1, C])]): ZStream[R1, E1, C] =
+    new ZStream[R1, E1, C] {
+      def fold[R2 <: R1, E2 >: E1, C1 >: C, S]: Fold[R2, E2, C1, S] =
+        ZManaged.succeedLazy { (s, cont, f) =>
+          for {
+            left  <- self.toQueue[E1, A1](lc)
+            right <- that.toQueue(rc)
+            result <- ZStream
+                       .unfoldM((s1, left, right)) {
+                         case (s1, left, right) =>
+                           f0(s1, left, right).flatMap {
+                             case (s1, take) =>
+                               Take.option(UIO.succeed(take)).map(_.map((_, (s1, left, right))))
+                           }
+                       }
+                       .fold[R2, E2, C1, S]
+                       .flatMap(_.apply(s, cont, f))
+          } yield result
+        }
+    }
+
+  /**
    * Appends another stream to this stream. The concatenated stream will first emit the
    * elements of this stream, and then emit the elements of the `other` stream.
    */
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1720,6 +1720,71 @@ object ZStream extends ZStreamPlatformSpecific {
...
+  /**
+   * Combines two streams `s1` and `s2` deterministically using the stream of
+   * boolean values `b` to control which stream to pull from next. `true`
+   * indicates to pull from the left stream and `right` indicates to pull from
+   * the right stream. Only consumes as many elements as requested by `b`. If
+   * either `s1` or `s2` are exhausted further requests for values from them
+   * will be ignored.
+   */
+  final def interleave[R, E, A](
+    b: ZStream[R, E, Boolean],
```
iravid: Is this expressive enough to handle two streams with different lengths? if all the caller of this combinator can do is provide a stream of booleans, surely they can't know if one of the streams ended and therefore they will not be able to "switch over" to reading just one side.

Maybe we need this to be a function that uses the queues themselves?

jdegoes:
> @adamgfraser Great work. This looks like a great combinator.
> 
> The only thought I have is that the implementations of `zipWith`, `mergeWith` and this combinator are all eerily similar. I wonder if we can abstract over them? @jdegoes could you have a look at this?

Perhaps a unification might look something like: `combine((dest, left, right) => ...)`, i.e. you get a destination queue (or at least the `offer` part of it), and you can push things there from either left, or right, in whatever order you choose; all scoped by Managed to ensure things cleanup properly.

iravid:
Right, I agree with that direction. What do you think @adamgfraser? Want to give this a shot?

adamgfraser:
@iravid Yes, I'm on it!

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: I was thinking that if the stream of booleans eventually requested values from both side like `Stream(true, false).forever` then the caller would be able to get all the values, but it is a little indirect.

What if we added a parameter `f: (ZStream[R, E, A], ZStream[R, E, A]) => ZStream[R, E, A)` where the first stream is the result of interleaving and the second is the "leftover" stream? So if the caller wanted to pull all remaining elements from the leftover stream they could use `_ ++ _`, if they wants to ignore to ignore them they could use `_._1`, or they could apply more complex logic.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1720,6 +1720,71 @@ object ZStream extends ZStreamPlatformSpecific {
...
 
+  /**
+   * Combines two streams `s1` and `s2` deterministically using the stream of
+   * boolean values `b` to control which stream to pull from next. `true`
+   * indicates to pull from the left stream and `right` indicates to pull from
+   * the right stream. Only consumes as many elements as requested by `b`. If
+   * either `s1` or `s2` are exhausted further requests for values from them
+   * will be ignored.
+   */
+  final def interleave[R, E, A](
```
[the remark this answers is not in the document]
adamgfraser: Definitely.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1720,6 +1720,71 @@ object ZStream extends ZStreamPlatformSpecific {
...
+                case Take.End => IO.succeed(s)
+              }
+
+          val resources = for {
+            queue      <- b.toQueue()
+            leftQueue  <- s1.toQueue()
+            rightQueue <- s2.toQueue()
+          } yield (queue, leftQueue, rightQueue)
+
+          ZManaged.fromEffect {
```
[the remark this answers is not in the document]
adamgfraser: Yes, that works. I see now that it doesn't make sense to use a resource and then just wrap it in another resource.

adamgfraser:
@iravid @jdegoes Take a look at this.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1516,6 +1472,31 @@ object ZStream extends ZStreamPlatformSpecific {
...
+                case Take.End      => IO.succeed(s)
+                case Take.Fail(e)  => IO.halt(e)
+                case Take.Value(c) => f(s, c).flatMap(s => loop(s, q))
+              }
+
+          for {
+            left   <- s1.toQueue(lc)
+            right  <- s2.toQueue(rc)
+            dest   <- Queue.unbounded[Take[E, C]].toManaged(_.shutdown)
+            _      <- f0(left, right, dest).toManaged_
```
iravid: Don't we need to fork here?

iravid: Do we actually need an output queue? I haven't tested if all combinators work with this, but what if we do the following:
- `f0: (Queue[Take[E, A]], Queue[Take[E, B]]) => ZIO[R, E, Take[E, C]]`
- run `f0` with `ZStream.unfoldM` and `unTake`

Does that make sense?

adamgfraser: Let me give it a try.

adamgfraser: Yes we can.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -998,62 +998,39 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+          }
+    } yield queue
 
-            s <- ZManaged.fromEffect(loop(false, false, s, queue))
-          } yield s
-        }
-    }
+    for {
+      queue  <- ZStream.managed(queue)
+      result <- ZStream.combine(self, that)((_, _) => queue.take)
```
iravid: 😂

I think this could be implemented if `ZStream.combine` allowed the function to keep a state between the invocations, like an unfold. Then you could use `raceWith` to race takes against the two queues. Whenever one side wins, you pass the pending side through the state to the next invocation. That would save the need for the output queue here.

That said, we can defer that to a follow-up ticket that enhances `combine` and implements `mergeWith` in terms of it if you'd like to get interleave in sooner.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1405,52 +1382,47 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-                )
-            )
-          }
-
-          for {
-            left  <- self.toQueue[E2, A](lc)
-            right <- that.toQueue[E2, B](rc)
-            s     <- ZManaged.fromEffect(loop(false, false, left, right, s))
-          } yield s
+  ): ZStream[R1, E1, C] = {
```
iravid: Nice code removal here!

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Sorry, I read more carefully and saw that you skip over a side that has ended. That makes sense! I think this is fine in this form then.

I suggest that we have a shorthand combinator for the common case of `Stream(false, true).forever` though :-) This combinator can be called `interleaveWith` and the shorter one `interleave`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1516,6 +1488,27 @@ object ZStream extends ZStreamPlatformSpecific {
   final def bracket[R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R, Nothing, _]): ZStream[R, E, A] =
     managed(ZManaged.make(acquire)(release))
 
+  final def combine[R, E, A, B, C](s1: ZStream[R, E, A], s2: ZStream[R, E, B], lc: Int = 2, rc: Int = 2)(
```
iravid: This one could also use an infix version.

iravid: It's also interesting to think of how would the generalized n-stream version would look like!

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: I agree.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Yes, I think that makes sense. I guess we could potentially use a `Ref` to pass the loser from one invocation to another like I did to manage state in the other combinators. But I was definitely feeling the lack of ability to maintain some state in writing them. It would probably be both easier to use and more performant to pass the state explicitly.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Is it better to just do everything infix? That seems more consistent with most of the other methods.

iravid: Yes, good point. Probably only variadic combinators should have a prefix form.

adamgfraser: And yes about generalizing further to multiple streams! But that I think we can save for another PR! 😃

adamgfraser:
This is ready for review.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -978,82 +1067,54 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-                }
+    def race(left: UIO[Take[E1, A]], right: UIO[Take[E1, B]]): UIO[(Take[E1, C], Loser)] =
+      left.raceWith(right)(
+        (exit, right) => ZIO.done(exit).map(a => (a.map(l), Right(right))),
+        (exit, left) => ZIO.done(exit).map(b => (b.map(r), Left(left)))
+      )
 
-            s <- ZManaged.fromEffect(loop(false, false, s, queue))
-          } yield s
+    self.combine(that)((false, false, Option.empty[Loser])) {
```
iravid: This is so nice!

adamgfraser:
@iravid Thanks for all your help! Excited to get this merged!

iravid:
@adamgfraser Happily and thank you for working on this - I was wondering for a long time how we'd abstract over all of these binary merging operators.

ghostdogpr:
@adamgfraser seems the test `interleaveWith` is flaky, I’ve seen it fail on an unrelated PR. 

https://circleci.com/gh/zio/zio/16550

adamgfraser:
@ghostdogpr Let me take a look.

iravid:
Might also be related to the `process` refactoring as I haven’t seen that fail before.

adamgfraser:
@ghostdogpr Have you rebased against master? I'm not able to reproduce locally.

ghostdogpr:
Yeah it happened on a PR that rebased against master. The failure only happened once and disappeared after re-running the build.

PR #1360, opened 2019-08-06
ZIO Test: Improve MockClock Handling of Sleep

adamgfraser:
The current implementation of `MockClock` does not handle sleep well. Currently, a call to `sleep` will increment the current time but not semantically block the calling fiber. This conflates modeling incrementing time with waiting for a future time to occur. In some circumstances the current implementation can be useful, such as in keeping track of how much "time" has elapsed through calls to `sleep` in a program. But it also has some major pitfalls. For example, we might expect the program below to print "Hello, World!" once when run with a `MockEnvironment`.

```scala
for {
  _ <- ZIO.effectTotal(println("Hello, World!")).delay(1.hour).forever.fork
  _ <- MockClock.adjust(90.minutes)
} yield true
```

However, it actually runs in an infinite loop printing "Hello, World!" again and again until it eventually crashes our program because there is no modeling of any delay. This seriously limits our ability to use `MockClock` to test abstractions dealing with time such as `Schedule`.

This pull request fixes this by having calls to `sleep` semantically block the calling fiber as long as the sleep duration is less than the current "clock time". Adjustment to the "clock time" through `adjust` or `setTime` check for any fibers with a sleep duration less than the new time and wake them up.


jdegoes:
@adamgfraser I really love this change. I also think that with this change, `MockClock` should also implement `Scheduler`, and we can delete the separate `MockScheduler`. Rather, scheduler will be triggered through `adjust`. This will also eliminate the need for cleanup actions to shut down the scheduler, and delete a lot of mock scheduler code.

Nice work! 💪

iravid:
This is really nice. I was looking for something like this when testing `ZStream#aggregateWithin`. Awesome work @adamgfraser!

adamgfraser:
@iravid Thanks! @jdegoes Yes I think you are right about `MockScheduler`. I refactored `MockClock` to extend `Scheduler` and deleted `MockScheduler` in the latest commit.

test/shared/src/main/scala/zio/test/mock/MockClock.scala
```diff
@@ -50,34 +53,98 @@ object MockClock {
...
+        }
+      }
+
+    private val wakeUp: UIO[Unit] =
+      clockState.modify { data =>
+        val (wakes, sleeps) =
+          data.sleeps.partition(_._1 <= Duration.fromNanos(data.nanoTime))
+        (wakes, data.copy(sleeps = sleeps))
+      }.flatMap { wakes =>
+        UIO.foreachPar_(wakes.sortBy(_._1))(_._2.succeed(())).fork.unit
```
jdegoes: `forkAll_` may be more efficient.

test/shared/src/main/scala/zio/test/mock/MockClock.scala
```diff
@@ -50,34 +53,98 @@ object MockClock {
...
+                            data.copy(
+                              sleeps =
+                                (Duration.fromNanos(data.nanoTime) + duration, latch) ::
+                                  data.sleeps
+                            )
+                          }
+                    } yield () => runtime.unsafeRun(cancel(latch))
+                  }
+              }
+            final def shutdown(): Unit = ()
```
jdegoes: Perhaps `shutdown` should run all pending tasks??? Not sure on that one.

test/shared/src/main/scala/zio/test/mock/MockClock.scala
adamgfraser: Yes.

test/shared/src/main/scala/zio/test/mock/MockClock.scala
adamgfraser: I wondered about that too. `java.util.concurrent.ScheduledExecutorService` does run pending tasks on shutdown so I updated to do the same.

PR #1361, opened 2019-08-07
Add ZIO#cached

adamgfraser:
@NeQuissimus Is this what you had in mind? There is a risk of a resource leak if the memoized value continues being updated when it is no longer needed so I returned a canceler as well. I think the test takes too long but I was getting flakiness if I used a shorter duration. We can refactor to use `MockClock` when we get #1360 in.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/IOSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/IOSpec.scala
@@ -7,6 +7,7 @@ import org.specs2.execute.Result
 import scala.collection.mutable
 import scala.util.Try
 import zio.Cause.{ die, fail, interrupt, Both }
+import zio.duration._
 
 class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime with GenIO with ScalaCheck {
   import Prop.forAll
@@ -45,6 +46,7 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
    Check `absolve` method on IO[E, Either[E, A]] returns the same IO[E, Either[E, String]] as `IO.absolve` does. $testAbsolve
    Check non-`memoize`d IO[E, A] returns new instances on repeated calls due to referential transparency. $testNonMemoizationRT
    Check `memoize` method on IO[E, A] returns the same instance on repeated calls. $testMemoization
+   Check `memoizeTimed` method on IO[E, A] returns new instances after duration. $testMemoizeTimed
    Check `raceAll` method returns the same IO[E, A] as `IO.raceAll` does. $testRaceAll
    Check `firstSuccessOf` method returns the same IO[E, A] as `IO.firstSuccessOf` does. $testfirstSuccessOf
    Check `zipPar` method does not swallow exit causes of loser. $testZipParInterupt
@@ -303,6 +305,23 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
         .flatMap(io => io <*> io)
         .map(tuple => tuple._1 must beTheSameAs(tuple._2))
     )
+  }
+
+  def testMemoizeTimed = {
+    def incrementAndGet(ref: Ref[Int]): UIO[Int] = ref.modify(n => (n + 1, n + 1))
+    unsafeRun {
+      for {
+        ref             <- Ref.make(0)
+        memoized        <- incrementAndGet(ref).memoizeTimed(100.milliseconds)
+        (cache, cancel) = memoized
+        a               <- cache
+        b               <- cache
+        _               <- clock.sleep(100.milliseconds)
+        c               <- cache
+        d               <- cache
+        _               <- cancel
+      } yield (a must_=== b) && (b must_!== c) && (c must_=== d)
+    }
   }
 
   def testRaceAll = {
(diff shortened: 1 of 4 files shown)


NeQuissimus:
Oh, I see what you did.
Not really what I had in mind.

I was more thinking along the lines of (pseudo-code)
`def memoize(d: Duration) = if cache.timestamp < clock.now - duration compute() else cache`

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -500,6 +500,25 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       _ <- (l.await *> ((self provide r) to p)).fork
     } yield l.succeed(()) *> p.await
 
+  /**
+   * Returns an effect that, if evaluated, will return the lazily computed result
+   * of this effect. The result of this effect will be recomputed with a fixed
+   * duration `d`. Also returns a canceler, which can be used to cancel the
+   * recomputation.
+   */
+  final def memoizeTimed(d: Duration): URIO[R with Clock, (IO[E, A], Canceler)] = {
```
jdegoes: The `Canceler` makes the API trickier to use.

I'd suggest an alternate approach:

1. Go with `Ref[Option[Promise[...]]]` as the core backing type.
2. When requested, fill the ref with a new promise that will be completed with the result of the effect.
3. After the promise is completed, set a timer, and after the timer elapses, the ref will be cleared to None.
4. Every get will trry to get the completion of the promise if it exists, but if it doesn't exist, it will go to (2).

This implementation will not require a canceler.

I would also call this method `cached` and not `memoizeTimed`.

adamgfraser:
Yes, updating the cache on client request if the time to live has expired is a much better approach that creating a background process to update the cache. Thanks for your feedback! This is ready for another review.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -147,6 +147,30 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
     )(use)
 
+  /**
+   * Returns an effect that, if evaluated, will return the cached result of
+   * this effect. Cached results will expire after `timeToLive` duration.
+   */
+  final def cached(timeToLive: Duration): ZIO[R with Clock, Nothing, IO[E, A]] = {
+
+    def get(cache: RefM[Option[Promise[E, A]]]): ZIO[R with Clock, E, A] =
+      cache.update {
```
ghostdogpr: You can use `updateSome` and only handle `case None`.

core-tests/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -305,6 +307,21 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
     )
   }
 
+  def testCached = {
+    def incrementAndGet(ref: Ref[Int]): UIO[Int] = ref.modify(n => (n + 1, n + 1))
```
ghostdogpr: Using `update` instead of `modify`, you won't have to repeat `n + 1`

core-tests/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -305,6 +307,21 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
+    unsafeRun {
+      for {
+        ref   <- Ref.make(0)
+        cache <- incrementAndGet(ref).cached(100.milliseconds)
+        a     <- cache
+        b     <- cache
+        _     <- clock.sleep(100.milliseconds)
```
jdegoes: This will be flaky due to non-determinism. We could use with `MockClock` or add `flaky` combinator to test (it's defined in `RTSSpec`).

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -147,6 +147,30 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+        case Some(p) =>
+          ZIO.succeed(Some(p))
+        case None =>
+          for {
+            p <- Promise.make[E, A]
+            _ <- self.to(p)
+            _ <- p.await.delay(timeToLive).flatMap(_ => cache.set(None)).fork
+          } yield Some(p)
+      }.flatMap(_.get.await)
```
jdegoes: Looks great! There are only two ways to improve this:

 - Make the ref a try state (empty, pending, or full), so two fibers don't try to fill it at the same time
 - Deal with interruption, i.e. ensure interruption of a `get` does not prevent the cache from _ever_ being filled

The first would necessitate the second, I think.

In any case, looks good to me, 👍  to ship!

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Great suggestion!

core-tests/jvm/src/test/scala/zio/IOSpec.scala
adamgfraser: Yes.

core-tests/jvm/src/test/scala/zio/IOSpec.scala
adamgfraser: Yes, trying to test this is actually what led me to identify the issues with the `MockClock`. I used the `flaky` combinator for now. If we get the `MockClock` improvements in first we can switch it to use that or otherwise we can do it subsequently.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I updated it to deal with interruption by using `uninterruptibleMask` so only the waiting on the promise is interruptible. Regarding your first point, I thought I had addressed this by using `RefM`. My thinking was that if two fibers called `get` at the same time the first would create the promise, fork the fiber to fill it, and take a reference to that promise, and then the second would wait for the promise creation / forking and take a copy of the same promise. So the empty / pending / full states would correspond to none, some pending promise, and some completed promise. Does this work or am I messing something up here?

jdegoes: Ah, yes, you did. I forgot you were using `RefM`. Effects will be synchronized appropriately.

adamgfraser:
Thanks!

PR #1362, opened 2019-08-07
Sinks for grouping elements into n sized lists, sets and maps. (#1309)

### LINKED ISSUE #1309: Add a Sink for grouping elements into n-sized lists, sets, maps
Signatures (happy to hear alternatives to naming):
```scala
def collectAllN(n: Int): ZSink[Any, Nothing, A, A, List[A]]
def collectAllToSet[A]: ZSink[Any, Nothing, Nothing, A, Set[A]]
def collectAllToSetN[A](n: Int): ZSink[Any, Nothing, A, A, Set[A]]
def collectAllToMap[K, A](key: A => K): ZSink[Any, Nothing, Nothing, A, Map[K, A]]
def collectAllToMapN[K, A](n: Int)(key: A => K): ZSink[Any, Nothing, A, A, Map[K, A]]
```

`collectAllN` can be a special-cased version of `Sink.foldUntil`. The latter 4 compute the stopping condition as a function of the state, not the elements, so they can't use `foldUntil`.

PawelGizka:
Resolves #1309

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -187,6 +187,16 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     foldM            $foldM
       short circuits $foldMShortCircuits
 
+    collectAllN $collectAllN
+
+    collectAllToSet $collectAllToSet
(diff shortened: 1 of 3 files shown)


iravid:
Just needs a quick `fmt`.

PawelGizka:
@iravid I formatted code.

PR #1363, opened 2019-08-07
Return Managed.apply & Managed.unapply methods lost when moving to separate object

kaishh:
We forgot to copy case class .apply/.unapply methods for ZManaged when moving to a separate Managed object

As reported on gitter:

```
Boris V.Kuznetsov @tampler 11:41
Hi guys. ZManaged has changed in RC11-1 and my code stopped to compile. I had an integration with Doobie Hikari, see below:
def mkTransactor(
    cfg: DBConfig,
    connectEC: ExecutionContext,
    transactEC: Blocker
  ): Managed[Throwable, Transactor[Task]] = {
    val xa =
      HikariTransactor.newHikariTransactor[Task](cfg.driver, cfg.url, cfg.user, cfg.password, connectEC, transactEC)

    val res = xa.allocated.map {
      case (transactor, cleanupM) =>
        Reservation(ZIO.succeed(transactor), cleanupM.orDie) // <<---------- Error 1 here
    }.uninterruptible

    Managed(res) // << ----------- Error 2 here
  }
Now, in RC11-1 I get an error:
[error] /home/bku/work/clover/zio_front/src/main/scala/config.scala:51:55: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [E1 >: Throwable]zio.ZIO[Any,Nothing,Unit]
[error]  required: zio.Exit[_, _] => zio.ZIO[?,Nothing,Any]
[error]         Reservation(ZIO.succeed(transactor), cleanupM.orDie)
[error]                                                       ^
[info] zio.ZIO[Any,Nothing,Unit] <: zio.Exit[_, _] => zio.ZIO[?,Nothing,Any]?
[info] false
[error] /home/bku/work/clover/zio_front/src/main/scala/config.scala:54:12: zio.Managed.type does not take parameters
[error]     Managed(res)
[error]            ^
[error] two errors found
How to fix ?
```

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -21,7 +21,7 @@ object IO {
   /**
    * @see See [[zio.ZIO.apply]]
    */
-  def apply[A](a: => A): Task[A] = ZIO.apply(a)
+  final def apply[A](a: => A): Task[A] = ZIO.apply(a)
 
   /**
    * @see See bracket [[zio.ZIO]]
--- a/core/shared/src/main/scala/zio/Managed.scala
+++ b/core/shared/src/main/scala/zio/Managed.scala
@@ -25,6 +25,12 @@ object Managed {
     ZManaged.absolve(v)
 
   /**
+   * See [[zio.ZManaged]]
+   */
+  final def apply[E, A](reserve: IO[E, Reservation[Any, E, A]]): Managed[E, A] =
+    ZManaged.apply(reserve)
+
+  /**
    * See [[zio.ZManaged.collectAll]]
    */
   final def collectAll[E, A1, A2](ms: Iterable[Managed[E, A2]]): Managed[E, List[A2]] =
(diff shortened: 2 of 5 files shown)


kaishh:
@iravid Hmm, dotty changes the return type of `unapply` and fails to compile... I'm tempted to remove `.unapply`'s type signature even though it's a public method to avoid creating split files, WDYT?
```
[info] Compiling 78 Scala sources and 1 Java source to /home/circleci/project/core/jvm/target/scala-0.16/classes ...
[error] -- [E007] Type Mismatch Error: /home/circleci/project/core/shared/src/main/scala/zio/Managed.scala:301:20 
[error] 301 |    ZManaged.unapply(v)
[error]     |    ^^^^^^^^^^^^^^^^^^^
[error]     |    Found:    zio.ZManaged[Any, E, A]
[error]     |    Required: Option[zio.IO[E, zio.Reservation[Any, E, A]]]
[warn] -- Warning: /home/circleci/project/core/shared/src/main/scala/zio/stm/STM.scala:587:6 
[warn] 587 |      final case object Retry               extends TRez[Nothing, Nothing]
[warn]     |      ^^^^^
[warn]     |      final modifier is redundant for objects
[warn] one warning found****
```

iravid:
Sounds fine to me.

PR #1365, opened 2019-08-07
Rename equals to equalTo

ioleo:
Becouse `equals` method is defined for Any object the `Predicate.equals` gets shadowed when imported:

```scala
import zio.test.{assert, suite, test, DefaultRunnableSpec}
import zio.test.Predicate.equals

object ExampleSpec {
  def squaredScenario(n: Int, expected: Int) =
    assert(n * n, equals(expected))

  object spec extends DefaultRunnableSpec(
    suite("square")(
      test("two squared") { squaredScenario(2, 4) }
    )
  )
}
```

This will fail at compile time:
```
[error] (...): type mismatch;
[error]  found   : Boolean
[error]  required: zio.test.Predicate[?]
[error]     assert(n * n, equals(expected))
[error]                         ^
```

The workaround is to explicitly state `Predicate.equals`, but this is clunky. It's also unexpected and might cause quite a headache to scala beginners.

I propose to rename it to `isEqual` following already estabilished pattern from `isSome`, `isNone`, `isTrue`, `isFalse`, etc.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZQueueSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZQueueSpec.scala
@@ -16,7 +16,7 @@ object ZQueueSpec
             v1    <- queue.take
             o2    <- queue.offer(20)
             v2    <- queue.take
-          } yield assert((v1, v2, o1, o2), Predicate.equals((10, 20, true, true)))
+          } yield assert((v1, v2, o1, o2), Predicate.equalTo((10, 20, true, true)))
         },
         testM("sequential take and offer") {
           for {
@@ -24,7 +24,7 @@ object ZQueueSpec
             f1    <- queue.take.zipWith(queue.take)(_ + _).fork
             _     <- queue.offer("don't ") *> queue.offer("give up :D")
             v     <- f1.join
-          } yield assert(v, Predicate.equals("don't give up :D"))
+          } yield assert(v, Predicate.equalTo("don't give up :D"))
         },
         testM("parallel takes and sequential offers ") {
           for {
@@ -33,7 +33,7 @@ object ZQueueSpec
             values = Range.inclusive(1, 10).toList
             _      <- values.map(queue.offer).foldLeft[UIO[Boolean]](IO.succeed(false))(_ *> _)
             v      <- f.join
-          } yield assert(v.toSet, Predicate.equals(values.toSet))
+          } yield assert(v.toSet, Predicate.equalTo(values.toSet))
         },
         testM("parallel offers and sequential takes") {
           for {
@@ -43,7 +43,7 @@ object ZQueueSpec
             _      <- waitForSize(queue, 10)
             l      <- queue.take.repeat(ZSchedule.recurs(9) *> ZSchedule.identity[Int].collectAll)
             _      <- f.join
-          } yield assert(l.toSet, Predicate.equals(values.toSet))
+          } yield assert(l.toSet, Predicate.equalTo(values.toSet))
         },
         testM("offers are suspended by back pressure") {
           for {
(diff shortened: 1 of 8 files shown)


adamgfraser:
I think you are definitely right that we need to change the name for the reason you describe. What about `equal`? I feel like people are going to be using that predicate a ton so might be nice to make it shorter.

ioleo:
@adamgfraser I would prefer `isEqual` to keep this pattern. Also `equal` could easily lead to typos (`equals`). Having most common predicates starting with `is` would be very IDE-friendly.

adamgfraser:
@ioleo Sounds good to me.

ioleo:
Note for reviewers, to spare your time. Changes in `PredicateSpec` are only:
* `equals` -> `isEqual`
* sorting tests alphabetically (just like combinators are)

NeQuissimus:
I would almost suggest naming it `isEqualTo` to keep the flow of `a isEqualTo b`.

ioleo:
Hmm :thinking: I'm not convinced. To me `a isEqual b` is clear enough and more concise. Also we have `isGreaterThanEqual` and `isLessThenEqual`.

I'm :+1: to `isEqual`
I'm :-1: to `isEqualTo`, becouse I feel if we went that way, we should also add the suffix to `isGreaterThanEqual` and `isLessThenEqual` and those seem a bit lenghy already (without the `To` suffix)

I'm curious what others think.

jdegoes:
`a isEqual b` reads fine, however, that's now how predicates are used. They are only used with one side. So it would be more like: `isSome(isEqual(3))`, in which case `isSome(isEqualTo(3))` is perhaps more grammatically correct. That said, it's used often, perhaps `equalTo` could suffice, since `equal` is already a verb; e.g.: `isSome(equalTo(4))`.

In any case, our goal should be good user experience out of the box, so a rename is definitely in order.

ioleo:
@NeQuissimus @jdegoes Alright, `isEqualTo` it is.

jdegoes:
@ioleo What do you think of `equalTo` instead of `isEqualTo`?

ioleo:
@jdegoes I think its OK iff we also rename `isGreater.../isLess` methods to drop the is prefix (provided in latest commit).

ioleo:
@jdegoes Is there anything missing or can we merge this?

adamgfraser:
I am definitely not the final decision maker here but I have a couple of comments on this.

I'm not sure about dropping the `is` from `isGreaterThan` and related predicates. We have a variety of other predicates that start with `is` (e.g, `isCase`, `isFalse`, `isLeft`). Do you just think that numerical comparisons are in some sense "more similar" conceptually to equality so you want to make these parallel? I would tend to think of equality as more of a special case because (1) it is going to be an extremely common predicate and (2) we are limited by things that are already in the namespace versus something we have to align everything else around.

Regardless of what we do, I would just change the names and not leave the old names with deprecation warnings. We're still at the very early stages of this project, I don't think anyone is really using this other than us internally for a few tests. So if we think a change makes sense we should just change it here and everywhere and move forward versus carrying around these slightly different method names into the future.

ioleo:
@adamgfraser Thank you for your input. I would also prefer to just rename things, but I assumed the deprecations are necessary, since we already have a release with these methods. 

I'd like to hear some directions from decision makers to move this forward.

ghostdogpr:
I agree with @adamgfraser on both points. 

There was only one release with ZIO Test (RC11) ans a bunch of things were still missing so I think there’s no harm just renaming these methods. It’s not like the rest of ZIO which is used by a lot of people.

ioleo:
@ghostdogpr Removed the deprecated methods. What about the `is` prefix discussion? I proposed that felt more natural/consistent to me, but I'll adapt to whatever is decided for. I feel like this (quite simple) PR is hanging too long already, so I'd prefer to get a decision, adapt the PR and get it merged.

ghostdogpr:
I think I prefer to keep the `is` for discoverability and only make an exception for `equalTo` which is used a lot.

ioleo:
@ghostdogpr Done.

ghostdogpr:
@ioleo thanks for going through this!

PR #1373, opened 2019-08-09
fix deadlock on maxYieldOps less than 3

### LINKED ISSUE #1372: Runtime deadlocks on maxYieldOps less than 3
When `maxYieldOps` is less than 3, `FiberContext` dead locks. The reason is that when `maxopcount` is reached, `curZio` is set to a `FlatMap(Yield, _)`. Processing this structure already takes two cycles and so never completes. Moreover, resuming from a `Yield` requires processing a `Success(Unit)` before the actual next ZIO. If `maxYieldOps` is less than 2, this triggers the next `Yield`. 

Steps to reproduce:

```scala

val rts = new DefaultRuntime {
  override val Platform = PlatformLive.Default.withExecutor(PlatformLive.ExecutorUtil.makeDefault(1))
}

rts.unsafeRun(
  for {
   _ <- UIO.unit
   _ <- UIO.unit
  } yield true
)
```

runtologist:
Fixes #1372

This PR removes the indirection via `FlatMap` when `maxopcount` is reached. For resuming, a new tag is introduced, which is handled before entering the evaluation loop and does not contribute to `maxopcount`.

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -238,12 +238,12 @@ private[zio] final class FiberContext[E, A](
         kTrace
       } else null
 
+    while ((curZio ne null) && (curZio.tag == ZIO.Tags.Resume)) { curZio = nextInstr(()) }
     while (curZio ne null) {
       try {
         var opcount: Int = 0
 
         while (curZio ne null) {
-          while (curZio.tag == ZIO.Tags.Resume) { curZio = nextInstr(()) }
           val tag = curZio.tag
 
           // Check to see if the fiber should continue executing or not:
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -243,6 +243,7 @@ private[zio] final class FiberContext[E, A](
         var opcount: Int = 0
 
         while (curZio ne null) {
+          while (curZio.tag == ZIO.Tags.Resume) { curZio = nextInstr(()) }
           val tag = curZio.tag
 
           // Check to see if the fiber should continue executing or not:
@@ -252,10 +253,9 @@ private[zio] final class FiberContext[E, A](
               // Cannot capture `curZio` since it will be boxed into `ObjectRef`,
               // which destroys performance. So put `curZio` into a temp val:
               val tmpIo = curZio
-
-              curZio = ZIO.yieldNow *> tmpIo
-
-              opcount = 0
+              pushContinuation(_ => tmpIo)
+              evaluateLater(ZIO.Resume)
+              curZio = null
             } else {
               // Fiber is neither being interrupted nor needs to yield. Execute
               // the next instruction in the program:
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -249,10 +249,7 @@ private[zio] final class FiberContext[E, A](
           if (tag == ZIO.Tags.Fail || !shouldInterrupt) {
             // Fiber does not need to be interrupted, but might need to yield:
             if (opcount == maxopcount) {
-              // Cannot capture `curZio` since it will be boxed into `ObjectRef`,
-              // which destroys performance. So put `curZio` into a temp val:
-              val tmpIo = curZio
-              evaluateLater(tmpIo)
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -243,6 +243,7 @@ private[zio] final class FiberContext[E, A](
         var opcount: Int = 0
 
         while (curZio ne null) {
+          while (curZio.tag == ZIO.Tags.Resume) { curZio = nextInstr(()) }
```
dreverri: It looks like `nextInstr` can return `null` which could cause `curZio.tag` to throw a `NullPointerException`?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -252,10 +253,9 @@ private[zio] final class FiberContext[E, A](
               // Cannot capture `curZio` since it will be boxed into `ObjectRef`,
               // which destroys performance. So put `curZio` into a temp val:
               val tmpIo = curZio
-
-              curZio = ZIO.yieldNow *> tmpIo
-
-              opcount = 0
+              pushContinuation(_ => tmpIo)
+              evaluateLater(ZIO.Resume)
```
dreverri: Could `tmpIo` be evaluated later instead of introducing a new tag?

core/shared/src/main/scala/zio/internal/FiberContext.scala
runtologist: I've added a guard condition.

core/shared/src/main/scala/zio/internal/FiberContext.scala
runtologist: The point is that we need to handle the parameter to `evaluateLater` before the while loop. I'd be happy about a solution without a special tag, but I couldn't think of one.

dreverri: why does the parameter to `evaluateLater` need to be handled before the while loop?

runtologist: Because we must not increase `maxopcount`. Otherwise we need to immediately yield again.

dreverri: I was thinking of something like:

```
if (opcount == maxopcount) {
    // Cannot capture `curZio` since it will be boxed into `ObjectRef`,
    // which destroys performance. So put `curZio` into a temp val:
    val tmpIo = curZio
    evaluateLater(tmpIo)
    curZio = null
} else {
```

With `curZio` set to `null` the while loop stops so we don't need to worry about opcount.

dreverri: Another idea is to add a `resume` to `Yield` and create a new `Yield` if `opcount == maxopcount` before matching on the tag. This also allows the addition of `yieldNow` to the trait so users can opt-in to yielding without creating an intermediate FlatMap.

https://github.com/zio/zio/compare/master...dreverri:djr/yield-resume?w=1

runtologist: I see. I like both ideas. I tend to prefer the first, because it is simpler. What do you think?

runtologist: Incorporated the first one. Thanks for the good suggestion. 🙏

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -252,10 +252,8 @@ private[zio] final class FiberContext[E, A](
               // Cannot capture `curZio` since it will be boxed into `ObjectRef`,
               // which destroys performance. So put `curZio` into a temp val:
               val tmpIo = curZio
-
-              curZio = ZIO.yieldNow *> tmpIo
-
-              opcount = 0
+              evaluateLater(tmpIo)
```
jdegoes: This solution does not require `tmpIo`, so you can simplify to:

```scala
evaluateLater(tmpIo)
curZio = null
```

kaishh:
Wow! Thank you! :pray:

core/shared/src/main/scala/zio/internal/FiberContext.scala
runtologist: done, thank you.

jdegoes:
@runtologist Thank you for fixing this! 🙏

PR #1377, opened 2019-08-10
Deprecate succeedLazy

### LINKED ISSUE #1346: Deprecate all succeedLazy
Too many people are confused about `succeedLazy`. Given a pure FP market, it could survive, but given the target audience of ZIO, I don't think it's a good fit.

We should deprecate all `succeedLazy` methods and prepare to delete them for 1.0.

ioleo:
Fixes #1346

* Deprecated all `succeedLazy` methods.
* Delegated all deprecated methods to `succeed`. 
* Added missing `succeed` to `ZSink`, `Sink` and `ZStreamChunk`.
* Removed `succeedLazy` test from `RTSSpec`
* Replaced `succeedLazy` with `effect` or `effectTotal` where laziness was essential and with  `succeed` where lazyness was not required
* Updated docs

### DIFF
--- a/benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
@@ -122,10 +122,10 @@ class IODeepFlatMapBenchmark {
 
   private[this] def zioDeepFlatMap(runtime: Runtime[Any]): BigInt = {
     def fib(n: Int): UIO[BigInt] =
-      if (n <= 1) ZIO.succeed[BigInt](n)
+      if (n <= 1) ZIO.effectTotal[BigInt](n)
       else
         fib(n - 1).flatMap { a =>
-          fib(n - 2).flatMap(b => ZIO.succeed(a + b))
+          fib(n - 2).flatMap(b => ZIO.effectTotal(a + b))
         }
 
     runtime.unsafeRun(fib(depth))
--- a/benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
@@ -122,10 +122,10 @@ class IODeepFlatMapBenchmark {
 
   private[this] def zioDeepFlatMap(runtime: Runtime[Any]): BigInt = {
     def fib(n: Int): UIO[BigInt] =
-      if (n <= 1) ZIO.succeedLazy[BigInt](n)
+      if (n <= 1) ZIO.succeed[BigInt](n)
       else
         fib(n - 1).flatMap { a =>
-          fib(n - 2).flatMap(b => ZIO.succeedLazy(a + b))
+          fib(n - 2).flatMap(b => ZIO.succeed(a + b))
         }
 
     runtime.unsafeRun(fib(depth))
--- a/benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
@@ -34,9 +34,9 @@ class IODeepLeftBindBenchmark {
 
   def zioDeepLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     var i  = 0
-    var io = IO.succeed(i)
+    var io = IO.effectTotal(i)
     while (i < depth) {
-      io = io.flatMap(i => IO.succeed(i))
+      io = io.flatMap(i => IO.effectTotal(i))
       i += 1
     }
 
--- a/benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
@@ -34,9 +34,9 @@ class IODeepLeftBindBenchmark {
 
   def zioDeepLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     var i  = 0
-    var io = IO.succeedLazy(i)
+    var io = IO.succeed(i)
     while (i < depth) {
-      io = io.flatMap(i => IO.succeedLazy(i))
+      io = io.flatMap(i => IO.succeed(i))
       i += 1
     }
 
--- a/benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
@@ -120,11 +120,11 @@ class IOLeftBindBenchmark {
 
   private[this] def zioLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     def loop(i: Int): UIO[Int] =
-      if (i % depth == 0) IO.succeed[Int](i + 1).flatMap(loop)
-      else if (i < size) loop(i + 1).flatMap(i => IO.succeed(i))
-      else IO.succeed(i)
+      if (i % depth == 0) IO.effectTotal[Int](i + 1).flatMap(loop)
+      else if (i < size) loop(i + 1).flatMap(i => IO.effectTotal(i))
+      else IO.effectTotal(i)
 
-    runtime.unsafeRun(IO.succeed(0).flatMap[Any, Nothing, Int](loop))
+    runtime.unsafeRun(IO.effectTotal(0).flatMap[Any, Nothing, Int](loop))
   }
 
   @Benchmark
--- a/benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
@@ -120,11 +120,11 @@ class IOLeftBindBenchmark {
 
   private[this] def zioLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     def loop(i: Int): UIO[Int] =
-      if (i % depth == 0) IO.succeedLazy[Int](i + 1).flatMap(loop)
-      else if (i < size) loop(i + 1).flatMap(i => IO.succeedLazy(i))
-      else IO.succeedLazy(i)
+      if (i % depth == 0) IO.succeed[Int](i + 1).flatMap(loop)
+      else if (i < size) loop(i + 1).flatMap(i => IO.succeed(i))
+      else IO.succeed(i)
 
-    runtime.unsafeRun(IO.succeedLazy(0).flatMap[Any, Nothing, Int](loop))
+    runtime.unsafeRun(IO.succeed(0).flatMap[Any, Nothing, Int](loop))
   }
 
   @Benchmark
--- a/benchmarks/src/main/scala/zio/IOMapBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IOMapBenchmark.scala
@@ -113,7 +113,7 @@ class IOMapBenchmark {
       if (n <= 1) t
       else sumTo(t.map(_ + n), n - 1)
 
-    runtime.unsafeRun(sumTo(IO.succeed(0), depth))
+    runtime.unsafeRun(sumTo(IO.effectTotal(0), depth))
   }
 
   @Benchmark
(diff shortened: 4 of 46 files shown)


docs/overview/creating_effects.md
```diff
@@ -23,16 +23,16 @@ You can also create use methods in the companion objects of the `ZIO` type alias
 val s2: Task[Int] = Task.succeed(42)
 ```
 
-The `succeed` method is eager, which means the value passed to `succeed` will be constructed _before_ the method is invoked. Although this is the most common way to construct a successful effect, you can achieve lazy construction using the `ZIO.succeedLazy` method:
+The `succeed` method is eager, which means the value passed to `succeed` will be constructed _before_ the method is invoked. Although this is the most common way to construct a successful effect, you can achieve lazy construction using the `ZIO.succeed` method:
```
shankarshastri: Lazy construction?

docs/datatypes/io.md
```diff
@@ -13,22 +13,16 @@ A value of type `IO[E, A]` describes an effect that may fail with an `E`, run fo
...
+You can lift pure values into `IO` with `IO.succeed`:
 
 ```scala mdoc:silent
 import zio._
 
-val lazyValue: UIO[String] = IO.succeedLazy("Hello World")
+val value: UIO[String] = IO.succeed("Hello World")
 ```
 
 The constructor uses non-strict evaluation, so the parameter will not be evaluated until when and if the `IO` action is executed at runtime, which is useful if the construction is costly and the value may never be needed.
```
shankarshastri: Non-Strict Evaluation?

docs/overview/creating_effects.md
```diff
@@ -23,16 +23,16 @@ You can also create use methods in the companion objects of the `ZIO` type alias
...
 ```scala mdoc:silent
 lazy val bigList = (0 to 1000000).toList
 lazy val bigString = bigList.map(_.toString).mkString("\n")
 
-val s3 = ZIO.succeedLazy(bigString)
+val s3 = ZIO.succeed(bigString)
 ```
 
-The value inside a successful effect constructed with `ZIO.succeedLazy` will only be constructed if absolutely required.
+The value inside a successful effect constructed with `ZIO.succeed` will only be constructed if absolutely required.
```
shankarshastri: Same here.

ioleo:
@shankarshastri Fixed docs.

kaishh:
> Delegated all deprecated methods to `succeed`.

I think we should delegate to `effectTotal` to not break existing usages and since `effectTotal` is now the only blessed way to delay evaluation.

benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
```diff
@@ -34,9 +34,9 @@ class IODeepLeftBindBenchmark {
 
   def zioDeepLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     var i  = 0
-    var io = IO.succeedLazy(i)
+    var io = IO.succeed(i)
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
```diff
@@ -122,10 +122,10 @@ class IODeepFlatMapBenchmark {
 
   private[this] def zioDeepFlatMap(runtime: Runtime[Any]): BigInt = {
     def fib(n: Int): UIO[BigInt] =
-      if (n <= 1) ZIO.succeedLazy[BigInt](n)
+      if (n <= 1) ZIO.succeed[BigInt](n)
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
```diff
@@ -120,11 +120,11 @@ class IOLeftBindBenchmark {
 
   private[this] def zioLeftBindBenchmark(runtime: Runtime[Any]): Int = {
     def loop(i: Int): UIO[Int] =
-      if (i % depth == 0) IO.succeedLazy[Int](i + 1).flatMap(loop)
-      else if (i < size) loop(i + 1).flatMap(i => IO.succeedLazy(i))
-      else IO.succeedLazy(i)
+      if (i % depth == 0) IO.succeed[Int](i + 1).flatMap(loop)
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

benchmarks/src/main/scala/zio/IOMapBenchmark.scala
```diff
@@ -113,7 +113,7 @@ class IOMapBenchmark {
       if (n <= 1) t
       else sumTo(t.map(_ + n), n - 1)
 
-    runtime.unsafeRun(sumTo(IO.succeedLazy(0), depth))
+    runtime.unsafeRun(sumTo(IO.succeed(0), depth))
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

benchmarks/src/main/scala/zio/IONarrowFlatMapBenchmark.scala
```diff
@@ -112,10 +112,10 @@ class IONarrowFlatMapBenchmark {
 
   private[this] def zioNarrowFlatMap(runtime: Runtime[Any]): Int = {
     def loop(i: Int): UIO[Int] =
-      if (i < size) IO.succeedLazy[Int](i + 1).flatMap(loop)
-      else IO.succeedLazy(i)
+      if (i < size) IO.succeed[Int](i + 1).flatMap(loop)
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

benchmarks/src/main/scala/zio/IOShallowAttemptBenchmark.scala
```diff
@@ -130,7 +130,7 @@ class IOShallowAttemptBenchmark {
   def scalazShallowAttempt(): BigInt = {
     def throwup(n: Int): IO[ScalazError, BigInt] =
       if (n == 0) throwup(n + 1).fold[BigInt](_ => 50, identity)
-      else if (n == depth) IO.succeedLazy(1)
+      else if (n == depth) IO.succeed(1)
```
kaishh: This should become `effectTotal` in line with other effect types that delay execution in their benchmarks (e.g. monix `Task.eval`)

core-tests/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -122,19 +122,19 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
     val effects = new mutable.ListBuffer[String]
-    val res     = unsafeRun(IO.foreach(list)(x => IO.effectTotal(effects += x) *> IO.succeedLazy[Int](x.toInt)))
+    val res     = unsafeRun(IO.foreach(list)(x => IO.effectTotal(effects += x) *> IO.succeed[Int](x.toInt)))
     (effects.toList, res) must be_===((list, List(1, 2, 3)))
   }
 
   def t3 = {
     val list = List("1", "h", "3")
-    val res  = Try(unsafeRun(IO.foreach(list)(x => IO.succeedLazy[Int](x.toInt))))
+    val res  = Try(unsafeRun(IO.foreach(list)(x => IO.effect[Int](x.toInt))))
```
kaishh: I think all `_.toInt` in this file should become either `effectTotal` or `succeed` consistently (since it's all the same call and there's no difference between them)

core-tests/jvm/src/test/scala/zio/IOSpec.scala
```diff
@@ -144,26 +144,26 @@ class IOSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntim
...
   def t6 = {
-    val list = List(1, 2, 3).map(IO.succeedLazy[Int](_))
+    val list = List(1, 2, 3).map(IO.succeed[Int](_))
     val res  = unsafeRun(IO.collectAllPar(list))
     res must be_===(List(1, 2, 3))
   }
 
   def t7 = {
-    val list = List(1, 2, 3).map(IO.succeedLazy[Int](_))
+    val list = List(1, 2, 3).map(IO.succeed[Int](_))
```
kaishh: I think it's better to replace most of these by `effectTotal` since otherwise the fiber workload becomes too trivial which may interfere with the original intention of the test writer...

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -285,11 +285,9 @@ object Fiber {
    */
   final def succeed[E, A](a: A): Fiber[E, A] = done(Exit.succeed(a))
 
-  /**
-   * Returns a fiber that is already succeeded with the specified lazily
-   * evaluated value.
-   */
-  final def succeedLazy[E, A](a: => A): Fiber[E, A] = done(Exit.succeed(a))
+  @deprecated("use succeed", "1.0.0")
```
kaishh: There is no `effectTotal` on `Fiber`, so I'm inclined to let this one stay since there's no alternative?.. /cc @jdegoes

core-tests/jvm/src/test/scala/zio/IOSpec.scala
ioleo: Actually I did initially try with `succeed`, but the test failed (aparently laziness matters here).

vasilmkd:
What is the impact of this on `ZStream`, given that `ZManaged.succeedLazy` is used to implement effectful folding? /cc @iravid

iravid:
`ZManaged.succeedLazy(a)` is just `ZManaged.fromEffect(ZIO.effectTotal(a))` so we can name that `ZManaged.effectTotal`. Can be done in a separate PR as other data types have a `succeedLazy` too.

benchmarks/src/main/scala/zio/IODeepLeftBindBenchmark.scala
ioleo: Fixed

benchmarks/src/main/scala/zio/IODeepFlatMapBenchmark.scala
ioleo: Fixed

benchmarks/src/main/scala/zio/IOLeftBindBenchmark.scala
ioleo: :heavy_check_mark:

benchmarks/src/main/scala/zio/IOMapBenchmark.scala
ioleo: :heavy_check_mark:

benchmarks/src/main/scala/zio/IONarrowFlatMapBenchmark.scala
ioleo: :heavy_check_mark:

benchmarks/src/main/scala/zio/IOShallowAttemptBenchmark.scala
ioleo: :heavy_check_mark:

core-tests/jvm/src/test/scala/zio/IOSpec.scala
ioleo: In `t3` lazyness matters, so changing all `_.toInt` to `effectTotal`.

core-tests/jvm/src/test/scala/zio/IOSpec.scala
ioleo: :heavy_check_mark:

ioleo:
@kaishh 
There is no `effectTotal` on `Fiber`, `STM`, `Schedule` and `ZSchedule`. For all others I've switched to `effectTotal` as you sugested. For `Managed` / `ZManaged` I introduced `effectTotal` as @iravid suggested.

jdegoes:
@ioleo Excellent work, thank you for this! 🙏

PR #1379, opened 2019-08-10
Adding ZStreamChunk#Collect

### LINKED ISSUE #1352: Add ZStreamChunk#collect
It would be nice to have a collect function on ZStreamChunk.

shankarshastri:
Fixes #1352

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
@@ -64,6 +64,16 @@ trait ZStreamChunk[-R, +E, @specialized +A] { self =>
         }
     })
 
+  final def collect[B](p: PartialFunction[A, B]): ZStreamChunk[R, E, B] =
+    ZStreamChunk(new ZStream[R, E, Chunk[B]] {
+      override def fold[R1 <: R, E1 >: E, B1 >: Chunk[B], S]: Fold[R1, E1, B1, S] =
+        ZManaged.succeedLazy { (s, cont, f) =>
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
```diff
@@ -64,6 +64,16 @@ trait ZStreamChunk[-R, +E, @specialized +A] { self =>
         }
     })
 
+  final def collect[B](p: PartialFunction[A, B]): ZStreamChunk[R, E, B] =
+    ZStreamChunk(new ZStream[R, E, Chunk[B]] {
+      override def fold[R1 <: R, E1 >: E, B1 >: Chunk[B], S]: Fold[R1, E1, B1, S] =
+        ZManaged.succeedLazy { (s, cont, f) =>
+          self.fold[R1, E1, A, S].flatMap { fold =>
```
iravid: This would unfortunately destroy the chunk structure. Could you try rewriting it with `self.chunks.fold` and running `collect(p)` on each chunk?

shankarshastri: `ZStreamChunk(self.chunks.map(e => e.collect(p))`, @iravid any thoughts about this one?

iravid: Yes, that'd be nice and short!

shankarshastri:
@iravid , Please take a look at it once, and let me know if you have any concern.

shankarshastri:
Sure, fixed the same.

PR #1382, opened 2019-08-11
Use MockClock in Test of ZIO#Cached

adamgfraser:


adamgfraser:
I realized that we don't actually need to fork a fiber to clear the cache, we can just keep track of the expiration time for a cached result and compute a new result if we receive a request after the previous expiration time. This results in a simpler and more deterministic implementation.

ghostdogpr:
@adamgfraser looks better indeed!

PR #1384, opened 2019-08-11
Add ZSink.splitLines

iravid:


iravid:
Oh this has a bug when `\r\n` is split across inputs. Don't merge yet.

iravid:
All done. @vasilmkd could you have a look?

vasilmkd:
It'll be my pleasure. I'll review it first thing in the morning.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1295,6 +1297,78 @@ object ZSink extends ZSinkPlatformSpecific {
         }
     }
 
+  /**
+   * Splits strings on newlines. Handles both `\r\n` and `\n`.
+   */
+  final val splitLines: ZSink[Any, Nothing, String, String, Chunk[String]] =
```
vasilmkd: Not to nitpick here, but I see this as a deviation from the previous sinks, which have all been defined with `def`. AFAIK `val` creates objects eagerly, and they stay in memory for the duration of the program which references this code. Is that something desirable in this case? I may be completely off base here, I just want to hear the reasoning. Same for every other `val`, including the ones in `Sink`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1295,6 +1297,78 @@ object ZSink extends ZSinkPlatformSpecific {
...
+    new SinkPure[Nothing, String, String, Chunk[String]] {
+      type State = (Chunk[String], Option[String], Boolean)
```
vasilmkd: On an urelated note, this is probably what the stepless sinks' code will look like after the refactoring. :)

vasilmkd:
> Looks useful! Could be also useful to generalize `splitOn(',')` or whatever.

Or a predicate function.

vasilmkd:
This needs a rebase and should be good for merging.

PR #1386, opened 2019-08-11
Test mapMPar ordering

### LINKED ISSUE #1367: ZStream.mapMPar(1) may process stream element out of order
Code to reproduce:
```scala
for {
  queue  <- Queue.unbounded[Int]
  _      <- Stream.range(0, 9).mapMPar(1)(queue.offer).runDrain
  result <- queue.takeAll
  _      <- if (result == result.sorted) putStrLn("Ok")
            else putStrLn(s"Out of order: ${result}")
} yield ()
```
With relatively high probability it prints something like:
```
Out of order: List(0, 1, 2, 3, 4, 5, 6, 7, 9, 8)
```
It doesn't violate any of the guarantees scaladocs give, but may be surprising.

vasilmkd:
Fixes #1367.

Interesting edge case.

`mapMPar` and `mapMParUnordered` have the exact opposite semantics when the parallelism factor is 1.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1,12 +1,14 @@
 package zio.stream
 
 import org.scalacheck.Arbitrary
+import org.scalacheck.Prop.forAll
 import org.specs2.ScalaCheck
 
 import scala.{ Stream => _ }
 import zio._
 import zio.duration._
 import zio.QueueSpec.waitForSize
+import org.scalacheck.Gen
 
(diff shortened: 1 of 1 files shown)


iravid:
`mapMParUnordered` uses your previous fix :-)

iravid:
Oh you noted that in the issue 😂

vasilmkd:
Got it. `mapMParUnordered` is correct in all cases. We need to fix the ordering for `mapMPar`, i.e. the ordering must be preserved for all `n` parallelism factors. Working on it.

vasilmkd:
@iravid Please take another look at the discussion #1367. This is ready for review.

PR #1387, opened 2019-08-11
 add ZStream#{groupBy, split, fanOut}

### LINKED ISSUE #1102: Add ZStream#partition

### LINKED ISSUE #1378: Add ZStream#fanOut?
I believe it makes sense to add a combinator that allows multiple consumers to process a stream at their own rate, with a configuration that determines how much the slowest consumer may lag behind. Proposed signature and implementation

```scala
  final def fanOut[E1 >: E, A1 >: A](n: Int, maximumLag: Int): ZManaged[R, Nothing, List[Queue[Take[E1, A1]]]] =
    ZManaged.collectAll(List.fill(n)(Queue.bounded[Take[E1, A1]](maximumLag).toManaged(_.shutdown))).flatMap { queues =>
      val offerAll = (a: Take[E1, A1]) =>
        ZIO.foreach_(queues)(_.offer(a))

      val driver = self
        .foreachManaged(a => offerAll(Take.Value(a)))
        .foldCauseM(
          cause => offerAll(Take.Fail(cause)).toManaged_,
          _ => offerAll(Take.End).toManaged_
      )
      driver.fork.as(queues)
    }
```

mschuwalow:
Resolves #1378
Resolves #1102

@iravid Would be great if you could take a look at this.

Based on the toQueuesManaged (naming things is hard, I'm very open to suggestions :D) one can implent quite a bit more powerful stuff. In fact I had a version with a partioner using a consitent hashing scheme that allowed adhoc up and down scaling. I don't think this is paticularly useful as is, but might be interesting to explore later


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -811,6 +842,52 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+      override def fold[R2 <: R1, E2 >: E1, B1 >: B, S]: Fold[R2, E2, B1, S] =
+        ZManaged.succeedLazy { (s, cont, g) =>
+          for {
+            out       <- Queue.bounded[Take[E1, B]](outputBuffer).toManaged(_.shutdown)
+            fibers    <- Ref.make[Vector[Fiber[Unit, Unit]]](Vector()).toManaged_
+            interrupt = fibers.get.flatMap(ZIO.foreach(_)(_.interrupt))
+            _ <- self.foreachManaged { a =>
+                  val innerStream = f(a)
+                    .foreach(b => out.offer(Take.Value(b)).unit)
+                    .foldCauseM(
```
ghostdogpr: `catchAllCause`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -864,6 +941,52 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
         }
     }
 
+  /**
+   * More powerful version of `ZStream#groupByKey`
+   */
+  def groupBy[R1 <: R, E1 >: E, K, V](
+    f: A => ZIO[R1, E1, (K, V)],
+    buffer: Int = 12
```
ghostdogpr: 16 is probably better than 12 since the underlying `RingBuffer` has better performance with power of 2s

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -864,6 +941,52 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+      } yield ZStream.fromQueue0(out).unTake
+    }
+
+  /**
+   * Group a stream using a function. This returns a stream of stream with each stream corresponding
+   * to a key value that has occurred. The driver stream will wait for all child streams to finish.
+   * Note that both the number of active keys and the number of buffered elements per key are limited.
+   * Prefer to use this with `ZStream#flatMapParSema` in order to prevent deadlocks.
+   */
+  def groupByKey[K](f: A => K, buffer: Int = 12): ZStream[R, E, (K, Stream[E, A])] =
```
ghostdogpr: same remark (12 -> 16)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1294,6 +1435,36 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
 
     }
 
+  /**
+   * Split a stream by a predicate. The faster stream may advance by up to buffer elements further than the slower one.
+   */
+  def split[B, C, R1 <: R, E1 >: E](
+    p: A => ZIO[R1, E1, Either[B, C]],
+    buffer: Int = 12
```
ghostdogpr: 12 -> 16

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1385,16 +1556,122 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   ): ZStream[R1 with Clock, E1, A] =
     transduceManaged(ZSink.throttleShapeM(units, duration, burst)(costFn))
 
+  /**
+   * Interrupts the stream if it does not produce a value after d duration.
+   */
+  final def timeout(d: Duration): ZStream[R with Clock, E, A] =
```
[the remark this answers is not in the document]
ghostdogpr: 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -811,6 +842,52 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
     }
 
+  /**
+   * Maps each element of this stream to another stream and returns the
+   * non-deterministic merge of those streams, executing all inner streams
+   * concurrently.
+   * Up to `outputBuffer` elements of the produced streams may be
+   * buffered in memory by this operator.
+   */
+  final def flatMapParUnbounded[R1 <: R, E1 >: E, B](
```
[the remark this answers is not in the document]
mschuwalow: Nope there isn't. Removed :D

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Ah nice. Updated

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -754,6 +764,27 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Maps each element of this stream to another stream and returns the
+   * non-deterministic merge of those streams, executing all inner streams
+   * concurrently.
+   * A semaphore with n permits is provided to limit the number of
+   * concurrent executions.
+   * Up to `outputBuffer` elements of the produced streams may be
+   * buffered in memory by this operator.
+   */
+  final def flatMapParSema[R1 <: R, E1 >: E, B](
```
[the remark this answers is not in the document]
mschuwalow: Not always. I found a use case for this while testing the group* combinators.
There I wanted to have a limit on the number of concurrent executions, but not on the number of groups that are actively consuming in order to avoid deadlocks. 
With this you can write something like this and get the desired behavior.
ZStream#withPermit would be too weak to do this as this requires  some 'external' semaphore.

I added ZStream#withSemaphore that makes it trivial to do this adhoc though, so maybe we should still drop this.
```scala
        .groupByKey(identity, 8192)
        .flatMapParSema(10, 16) {
          case (sema, (k, s)) =>
            s.withPermit(sema).transduce(Sink.foldLeft[String, Int](0) { case (acc: Int, _: String) => acc + 1 }).take(1).map((k -> _))
        }
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -886,6 +944,145 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+   * Split a stream into n streams using a hash function.
+   * The driver streamer will only ever advance of the maximumLag values before the
+   * slowest downstream stream.
+   * That means that the producer may backpressure easily if the partitions are very uneven.
+   * If a stream finishes the keys will be rebalanced in a consistent way.
+   * This returns a stream and an unsubsribe function that can be used to remove a stream manually.
+   * In that case that stream will receive no new elements and run to completion. Manually unsubscribing
+   * and waiting for the stream to finish ensures that no elements are lost.
+   */
+  def hashPartition(
```
mschuwalow: I still need to do quite a bit more testing here and must update the docstring before it gets merged.
I added this here for now mainly to get some discussion started on whether we want something like this.
It might make sense to rip this out and do this in a seperate pr though

mschuwalow: removed for now

mschuwalow:
i think this is ready for another review

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   ): ZStream[R1 with Clock, E1, A] =
     transduceManaged(ZSink.throttleShapeM(units, duration, burst)(costFn))
 
+  /**
+   * Interrupts the stream if it does not produce a value after d duration.
+   */
+  final def timeout(d: Duration): ZStream[R with Clock, E, A] =
+    ZStream.managed(self.toQueue()).flatMap { queue =>
```
iravid: This can now be written directly in terms of `process`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1486,6 +1726,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+   */
+  final def withPermit(sema: Semaphore): ZStream[R, E, A] =
+    withPermits(sema, 1)
+
+  /**
+   * Acquire n permits for every consumer task.
+   */
+  final def withPermits(sema: Semaphore, n: Long): ZStream[R, E, A] =
+    new ZStream[R, E, A] {
+      def fold[R1 <: R, E1 >: E, A1 >: A, S]: ZStream.Fold[R1, E1, A1, S] =
```
iravid: This too can be written in terms of `process`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1846,6 +2115,23 @@ object ZStream extends ZStreamPlatformSpecific {
         )
     }
 
+  /**
+   * Creates a stream from a [[zio.ZQueue]] of values
+   * The queue will be shut down once the stream is finished
+   */
+  final def fromQueue0[R, E, A](queue: ZQueue[_, _, R, E, _, A]): ZStream[R, E, A] =
```
iravid: I think this is `ZStream.fromInputStreamManaged(fromQueue(queue).process.ensuringFirst(queue.shutdown))`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1927,4 +2213,20 @@ object ZStream extends ZStreamPlatformSpecific {
    */
   final def unwrap[R, E, A](fa: ZIO[R, E, ZStream[R, E, A]]): ZStream[R, E, A] =
     flatten(fromEffect(fa))
+
+  /**
+   * Creates a stream produced from a [[ZManaged]]
+   */
+  final def unwrapManaged[R, E, A](fa: ZManaged[R, E, ZStream[R, E, A]]): ZStream[R, E, A] =
+    new ZStream[R, E, A] {
+      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] =
```
iravid: Rewrite in terms of `process`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1486,6 +1726,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+          self.fold[R1, E1, A1, S].flatMap { fold =>
+            fold(s, cont, (s, a) => sema.withPermits(n)(f(s, a)))
+          }
+        }
+    }
+
+  /**
+   * Adds a shared semaphore with n permits to the stream
+   */
+  final def withSemaphore(n: Long): ZStream[R, E, (Semaphore, A)] =
```
iravid: Not sold on this. This is can be just `ZStream.fromEffect(Semaphore.make(n)).flatMap { sem =>`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -561,6 +561,12 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
         }
     }
 
+  /**
+   * Immediately ends the stream without consuming an element
+   */
+  def discard: ZStream[R, E, Unit] =
+    self.take(0).drain
```
iravid: What's the usecase for this combinator? `take(0)` means the stream does not execute any element effects, just scope effects.

streams/shared/src/main/scala/zio/stream/ZStream.scala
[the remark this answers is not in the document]
iravid: Not sold on either - they're just special cases for `flatMap`. I suggest we drop them.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1282,6 +1383,12 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
    */
   def runCollect: ZIO[R, E, List[A]] = run(Sink.collectAll[A])
 
+  /**
+   * Immediately ends the stream without consuming an element
+   */
+  def runDiscard: ZIO[R, E, Unit] =
+    self.discard.runDrain
```
iravid: If we do this, all the stream does is execute its scopes and exit. Is it useful?

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: And we can call it `fromQueueWithShutdown` I suppose, as it isn't a "more powerful" version of `toQueue"

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+  /**
+   * Converts the stream to a managed list of queues. Every value will be replicated to every queue with the
+   * slowest queue being allowed to buffer maximumLag elements before the driver is backpressured.
+   * The downstream queues will be provided with elements in the same order they are returned, so
+   * the fastest queue might have seen up to (maximumLag + 1) elements more than the slowest queue if it
+   * has a lower index than the slowest queue.
+   * During the finalizer the driver will wait for all queues to shutdown in order to signal completion.
+   * After managed queue is used, the queue will never again produce values and should be discarded.
+   */
+  final def toQueues[E1 >: E, A1 >: A](n: Int, maximumLag: Int): ZManaged[R, Nothing, List[Queue[Take[E1, A1]]]] = {
```
iravid: Naming suggestion: `broadcastedQueues`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    val decider = ZIO.succeed((_: Int) => true)
+    toQueuesBalanced(n, maximumLag, _ => decider)
+  }
+
+  /**
+   * More powerful version of `ZStream#toQueues`. Allows to provide a function that determines what
+   * queues should receive which elements.
+   */
+  final def toQueuesBalanced[E1 >: E, A1 >: A](
```
iravid: `distributedWith`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+   * new queues and corresponding indices.
+   * You can also provide a function that will be executed after the final events are enqueued in all queues
+   * but before the queues are actually shutdown.
+   * Shutdown of the queues is handled by the driver.
+   */
+  final def toQueuesBalanced0[E1 >: E, A1 >: A](
+    maximumLag: Int,
+    decide: A => UIO[Int => Boolean],
+    done: Take[E1, Nothing] => UIO[_]
+  ): ZManaged[R, Nothing, UIO[(Int, Queue[Take[E1, A1]])]] =
```
iravid: `distributedDynamicWith`? As this version will allow creating additional queues dynamically

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1327,12 +1434,42 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
 
     }
 
+  /**
+   * Split a stream by a predicate. The faster stream may advance by up to buffer elements further than the slower one.
+   */
+  def split[B, C, R1 <: R, E1 >: E](
```
```suggestion
  def partitionEither[B, C, R1 <: R, E1 >: E](
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -632,6 +638,16 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
         }
     }
 
+  /**
+   * Fan out the stream, producing a list of streams that have the same elements as this stream.
+   * The driver streamer will only ever advance of the maximumLag values before the
+   * slowest downstream stream.
+   */
+  final def fanOut[E1 >: E, A1 >: A](n: Int, maximumLag: Int): ZManaged[R, Nothing, List[ZStream[Any, E, A]]] =
```
iravid: `broadcastedStreams`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -632,6 +638,16 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    for {
+      queues <- self.toQueues(n, maximumLag)
+    } yield queues.map(q => ZStream.fromQueue(q).unTake.ensuring(q.shutdown))
```
```suggestion
    } yield queues.map(q => ZStream.fromQueueWithShutdown(q).unTake)
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -791,7 +791,7 @@ object ZManaged {
    * specified text message. This method can be used for terminating a fiber
    * because a defect has been detected in the code.
    */
-  final def dieMessage(message: String): ZManaged[Any, Throwable, Nothing] = die(new RuntimeException(message))
+  final def dieMessage(message: String): ZManaged[Any, Nothing, Nothing] = die(new RuntimeException(message))
```
jdegoes: Good catch! 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    Ref
+      .make[Vector[Queue[Take[E1, A1]]]](Vector())
+      .toManaged(_.get.flatMap(qs => ZIO.foreach(qs)(_.shutdown)))
+      .flatMap { queues =>
+        val broadcast = (a: Take[E1, Nothing]) =>
+          for {
+            queues <- queues.get
+            _ <- ZIO.foreach_(queues) { q =>
+                  q.offer(a).catchAllCause {
+                    case c if (c.interrupted) => ZIO.unit
```
jdegoes: Partial function used where total one expected...

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1418,16 +1555,119 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+                }
+          } yield ()
+
+        val offer = (a: A) =>
+          for {
+            decider <- decide(a)
+            queues  <- queues.get
+            _ <- ZIO.foreach_(queues.zipWithIndex.collect { case (q, id) if decider(id) => q }) { q =>
+                  q.offer(Take.Value(a)).catchAllCause {
+                    case c if (c.interrupted) => ZIO.unit
```
jdegoes: Ditto

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Removed as I don't think this make sense with the new process encoding

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Removed as I don't think this make sense with the new process encoding.
The idea was to basically eagerly signal upstream that we are done with the queue. This was possible in fold because of cont, but I don't think it makes sense otherwise

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Agreed

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Hey why not? This can still work as:
```scala
ZStream.fromInputStreamManaged {
  self.process.map { as =>
    sema.withPermits(n)(as)
  }
}
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Removed. This would again only be useful for eagerly signaling that we are done

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: That is match better!

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Probaby not useful enough to warrant its own combinator. The usecase I had in mind is when we are handed a stream for somebody else and want to signal that we don't intend to process it any further

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Ugh that was close. Shameful :fearful:
I brainfarted and thought this was defined as a partialfunction on causes, halting with any that don't match

mschuwalow:
@iravid please take a look at the new signature of groupby.
I think much of the reason for the questionable new combinators was that groupby exposed an api that was very easy to deadlock with. I think this new one is better and makes them obsolete

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1483,6 +1680,51 @@ object ZStream extends ZStreamPlatformSpecific {
...
+          if (f(k)) ZIO.succeed(elem).as(true)
+          else q.shutdown.as(false)
+      }
+      new GroupBy(g1, buffer)
+    }
+
+    /**
+     * Run the function across all groups, collecting the results in an arbitrary order.
+     */
+    def apply[R1 <: R, E1 >: E, A](f: (K, Stream[E, V]) => ZStream[R1, E1, A]): ZStream[R1, E1, A] =
```
iravid: This API is very clever! Nice work :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -880,6 +976,61 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+   */
+  def groupBy[R1 <: R, E1 >: E, K, V, A1](
+    f: A => ZIO[R1, E1, (K, V)],
+    buffer: Int = 16
+  ): ZStream.GroupBy[R1, E1, K, V] = {
+    val qstream = ZStream.unwrapManaged {
+      for {
+        decider <- Promise.make[Nothing, (K, V) => UIO[Int => Boolean]].toManaged_
+        out <- Queue
+                .bounded[Take[E1, (K, ZQueue[Any, Nothing, Any, Nothing, Nothing, Take[E1, V]])]](buffer)
```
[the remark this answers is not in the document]
mschuwalow: Definitely makes things more readable here.
Do we want that type alias avaiable as part of the public API in the _root_.zio package object or similar?

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -393,6 +414,66 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
...
+  private def broadcastBackPressure =
+    flaky(
+      Stream
+        .range(0, 5)
+        .broadcast(2, 2)
+        .use {
+          case s1 :: s2 :: Nil =>
+            for {
+              ref      <- Ref.make[List[Int]](Nil)
+              _        <- s1.timeout(100.milliseconds).foreach(i => ref.update(i :: _)).ignore
```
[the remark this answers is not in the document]
mschuwalow: Hmm. I can't think of anything here. Maybe you can give an example of how you would test that it correctly backpressure here. Otherwise we can just throw it out I think

streams/shared/src/main/scala/zio/stream/ZStream.scala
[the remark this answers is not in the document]
iravid: I think that can be on `ZStream.GroupBy`'s companion object for now..

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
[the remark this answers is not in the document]
iravid: Something like this:
- The source is `Stream.range(0, 5).tap(i => ref.update(i :: _)).broadcast...`
- In the `use`:
```scala
for {
  latch1 <- Promise.make[Nothing, Unit]
  latch2 <- Promise.make[Nothing, Unit]
  fib <- s1.tap(i => latch1.succeed(()) *> latch2.await.when(i == 2)).runDrain.fork
  _ <- latch1.await
  snapshot1 <- ref.get
  _ <- latch2.succeed(())
  _ <- fib.await
  snapshot2 <- ref.get
} yield (snapshot1 must_=== <elems that are not backpressured>) and (snapshot2 must_=== <all elems>)
```
  
```

File: streams/shared/src/main/scala/zio/stream/ZStream.scala (hunk 25 above)
Comment (mschuwalow) [reply streams/shared/src/main/scala/zio/stream/ZStream.scala]: Thanks! :)

File: streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala (hunk 27 above)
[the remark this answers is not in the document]
Comment (mschuwalow) [reply streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala]: Yeah that works. Thank you!

**Comment by mschuwalow** [comment]
> A few minor changes and this is good to go. Amazing work @mschuwalow!

Great. I think you can take another look when you have the time :)

File: streams/shared/src/main/scala/zio/stream/ZStream.scala [hunk 28]
```diff
@@ -542,6 +569,75 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   final def concat[R1 <: R, E1 >: E, A1 >: A](other: => ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
     ZStream(UIO.succeed(self), UIO(other)).flatMap(ZStream.unwrap)
 
+  /**
+   * More powerful version of `ZStream#toQueues`. Allows to provide a function that determines what
```
iravid: scaladoc needs adjusting here

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -542,6 +569,75 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    decide: A => UIO[Int => Boolean]
+  ): ZManaged[R, Nothing, List[Queue[Take[E1, A1]]]] =
+    Promise.make[Nothing, A => UIO[Int => Boolean]].toManaged_.flatMap { prom =>
+      distributedDynamicWith[E1, A1](maximumLag, (a: A) => prom.await.flatMap(_(a)), _ => ZIO.unit).flatMap { next =>
+        (ZIO.collectAll(List.fill(n)(next.map(_._2))) <* prom.succeed(decide)).toManaged_
+      }
+    }
+
+  /**
+   * More powerful version of `ZStream#toQueuesBalanced`. This returns a function that will produce
```
iravid: and here

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Sure. Will try to do it tomorrow :)

PR #1389, opened 2019-08-12
Resolve #634

### LINKED ISSUE #634: Generalize ZStream with input/output error types to allow for combinators like `mapError`
mapError should allow for transforming errors before running the Stream. Proposed signature:

```scala
def mapError[E2](f: E => E2): ZStream[R, E2, A]
```

mschuwalow:


mschuwalow:
Sure, It was also a fun reason to learn more about the stream internals.
I think this can be closed and we instead use the inputstream encoding.
Or we can use this while the transition is taking place :thinking:

jdegoes:
@mschuwalow @iravid The other encoding seems to proceed so fast, perhaps best to close this one down? I am not sure, up to you two...

iravid:
Yes, I think we'll be done with the refactor by the end of the week. Thanks again for the work @mschuwalow :-)

Issue #1388, opened 2019-08-12
Lower-level, more expressive foundation for ZStream


jdegoes:
Lower-level foundation of Streams, which is stack-safe, does not require `Queue` for most combinators, and is readily understood as a "scoped" Java-style `InputStream`, without the unnecessary open / close operations.

It is trivially possible to implement `fold` in terms of `inputStream`; implementing `inputStream` in terms of `fold` would require a `Queue` (which is expected).

```scala
// An read on an input stream terminates normally because it is done (`None`) 
// or abnormally for some reason `Some(e: E)`:
type InputStream[+E, +A] = IO[Option[E], A]

trait ZStream[-R, +E, +A] {
  /**
   * Call `process` to obtain a managed input stream that can be used to read from 
   * the stream until it is empty (or possibly forever, if the stream is infinite). The 
   * provided `InputStream` is valid only inside the scope of the managed resource.
   */
  def inputStream: ZManaged[R, E, InputStream[E, A]]
}
```

Some interesting properties:

 - The failed read of an individual `A` can be retried (or ignored, or provided a fallback), independently of retrying or ignoring or providing a fallback for the whole stream
 - There is no need for a separate type parameter for client errors, because client errors occur only in consumption, and can be introduced on `run`, `fold`, `foreach`, etc., in which case supertype or manual unification could be used
 - Most operators are extremely trivial, and, indeed, the "guts" are simple enough for users to use:

```scala
stream1.zipWith(stream2) {
  case (inputStream1, inputStream2) =>
    ...
}
```

The only drawback I can see is that fusion is significantly harder to achieve; but indeed, there could exist two types of streams (subtypes of a sealed `ZStream`), one with `InputStream` as defined above, and one with `InputStream` as defined by `() => A` (very low level, not visible externally), which would throw both errors and end-of-stream; and this alternate constructor could perform extremely well for pure streams and synchronous effectful streams (async streams pretty much require the power of `ZIO` and will only be performant with `Chunk`).

/cc @iravid

jdegoes:
Oh, and obviously `uncons` is extremely simple; and fixed or varying headers can be pulled off and leave a stream remainder, which was previously insanely difficult (`peel`) and not performant.

iravid:
I like this! It's just an effectful iterator (e.g. `InputStream` is `next`) that establishes a scope.

I wonder if an equivalent to `hasNext` is needed?

jdegoes:
[...]

PR #1391, opened 2019-08-12
Introduce the ZStream.InputStream mechanism

vasilmkd:
First phase of the new low level implementation of `ZStream` as discussed in #1388.

The test code shows that the implementation is somewhat correct. It can be removed later, when `process` becomes a primitive combinator, or it can be removed as part of this PR, I have no preference.

/cc @iravid @jdegoes @mschuwalow


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1205,6 +1212,20 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
       }
   }
 
+  private final def processDefault: ZManaged[R, E, InputStream[E, A]] =
+    for {
+      queue <- Queue.bounded[Take[E, A]](1).toManaged(_.shutdown)
+      _     <- self.intoManaged(queue).fork
+    } yield queue.take.flatMap {
```
mschuwalow: There is an existing combinator called toQueue that does this.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -213,6 +216,18 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
     zipWith prioritizes failure $zipWithPrioritizesFailure
   """
 
+  def processRunCollect = {
+    def loop[E, A](effect: IO[_, A], ref: Ref[List[A]]): IO[Nothing, List[A]] =
+      (effect.flatMap(a => ref.update(a :: _)) *> loop(effect, ref)).catchAll(_ => ref.get)
```
mschuwalow: you can use ZIO#forever to make this a bit simpler

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -37,7 +37,14 @@ import zio.duration.Duration
...
 trait ZStream[-R, +E, +A] extends Serializable { self =>
-  import ZStream.Fold
+  import ZStream.{ Fold, InputStream }
+
+  /**
+   * Obtain a managed input stream that can be used to read from the stream until it is
+   * empty (or possibly forever, if the stream is infinite). The provided `InputStream`
+   * is valid only inside the scope of the managed resource.
+   */
+  def process: ZManaged[R, E, InputStream[E, A]] = processDefault
```
mschuwalow: The way I understood @iravid was that we keep this abstract and put process = processDefault on all anonymous subclasses that we create here.

vasilmkd: Yes, it'll have to be done sooner or later.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Thanks. Done.

iravid:
Ok, I did mean to add the `process = processDefault` override on all anonymous subclasses, but we can do it this way too.

Let's just do one more thing: add a `foldDefault` implementation that uses `process`. Then, when we migrate a combinator/constructor, we override its `process` method and forward its `fold` method to `foldDefault`.

So for example with `Stream.managed`:
```scala
// Before
  final def managed[R, E, A](managed: ZManaged[R, E, A]): ZStream[R, E, A] =
    new ZStream[R, E, A] {
      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] =
        ZManaged.succeedLazy { (s, cont, f) =>
          if (!cont(s)) ZManaged.succeed(s)
          else managed.mapM(f(s, _))
        }
    }

// After
  final def managed[R, E, A](managed: ZManaged[R, E, A]): ZStream[R, E, A] =
    new ZStream[R, E, A] {
      def process: ZManaged[R, E, IO[Option[E], A]] = managed.map(UIO.succeed)
      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault[...]
```

vasilmkd:
@iravid Ready for review. Unfortunately, I couldn't also make `foldDefault` private since it's technically dead code at the moment and remains unused.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
vasilmkd: Done.

iravid:
Nicely done @vasilmkd! Now the fun part begins!

vasilmkd:
Can't wait.

PR #1397, opened 2019-08-13
[WIP] Convert ZStream combinators and constructors to ZStream#process

vasilmkd:
Phase 2 of #1388.

WIP.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -44,7 +44,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
    * empty (or possibly forever, if the stream is infinite). The provided `InputStream`
    * is valid only inside the scope of the managed resource.
    */
-  def process: ZManaged[R, E, InputStream[E, A]] = processDefault
+  def process: ZManaged[R, E, InputStream[R, E, A]] = processDefault
 
   /**
    * Executes an effectful fold over the stream of values.
@@ -1227,7 +1227,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
       }
   }
 
-  private final def processDefault: ZManaged[R, E, InputStream[E, A]] =
+  private final def processDefault: ZManaged[R, E, InputStream[R, E, A]] =
     toQueue(1).map(_.take.flatMap {
       case Take.Value(a) => UIO.succeed(a)
       case Take.Fail(c) =>
@@ -1556,7 +1556,7 @@ object ZStream extends ZStreamPlatformSpecific {
    * Describes an effectful read from a stream. The optionality of the error channel denotes
    * normal termination of the stream when `None` and an error when `Some(e: E)`.
    */
-  type InputStream[+E, +A] = IO[Option[E], A]
+  type InputStream[-R, +E, +A] = ZIO[R, Option[E], A]
 
   /**
    * Describes an effectful fold over the elements of the stream. Conceptually it is an effectful state
@@ -1589,7 +1589,7 @@ object ZStream extends ZStreamPlatformSpecific {
     new Stream[Nothing, Nothing] {
       def fold[R, E, A, S]: Fold[R, E, A, S] = foldDefault
 
-      override def process: Managed[Nothing, InputStream[Nothing, Nothing]] =
+      override def process: Managed[Nothing, InputStream[Any, Nothing, Nothing]] =
         ZManaged.succeed(UIO.never)
     }
 
@@ -1644,9 +1644,9 @@ object ZStream extends ZStreamPlatformSpecific {
     new ZStream[R, E, A] {
       def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
 
-      override def process: ZManaged[R, E, InputStream[E, A]] =
+      override def process: ZManaged[R, E, InputStream[R, E, A]] =
         for {
-          output  <- Queue.bounded[InputStream[E, A]](outputBuffer).toManaged(_.shutdown)
+          output  <- Queue.bounded[InputStream[R, E, A]](outputBuffer).toManaged(_.shutdown)
           runtime <- ZIO.runtime[R].toManaged_
           maybeStream <- UIO(
                           register(
@@ -1692,9 +1692,9 @@ object ZStream extends ZStreamPlatformSpecific {
     new ZStream[R, E, A] {
       def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
 
-      override def process: ZManaged[R, E, InputStream[E, A]] =
+      override def process: ZManaged[R, E, InputStream[R, E, A]] =
         for {
-          output  <- Queue.bounded[InputStream[E, A]](outputBuffer).toManaged(_.shutdown)
+          output  <- Queue.bounded[InputStream[R, E, A]](outputBuffer).toManaged(_.shutdown)
           runtime <- ZIO.runtime[R].toManaged_
           _ <- register(
                 k =>
@@ -1725,9 +1725,9 @@ object ZStream extends ZStreamPlatformSpecific {
     new ZStream[R, E, A] {
       def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
 
-      override def process: ZManaged[R, E, InputStream[E, A]] =
+      override def process: ZManaged[R, E, InputStream[R, E, A]] =
         for {
-          output  <- Queue.bounded[InputStream[E, A]](outputBuffer).toManaged(_.shutdown)
+          output  <- Queue.bounded[InputStream[R, E, A]](outputBuffer).toManaged(_.shutdown)
           runtime <- ZIO.runtime[R].toManaged_
           eitherStream <- UIO(
                            register(
@@ -1765,7 +1765,7 @@ object ZStream extends ZStreamPlatformSpecific {
     new ZStream[R, Nothing, Nothing] {
       def fold[R1 <: R, E, A1, S]: Fold[R1, E, A1, S] = foldDefault
 
-      override def process: ZManaged[R, Nothing, InputStream[Nothing, Nothing]] =
+      override def process: ZManaged[R, Nothing, InputStream[R, Nothing, Nothing]] =
         ZManaged.succeed(IO.fail(None)).ensuring(finalizer)
     }
 
@@ -1792,7 +1792,7 @@ object ZStream extends ZStreamPlatformSpecific {
     new Stream[Nothing, A] {
       def fold[R, E, A1 >: A, S]: Fold[R, E, A1, S] = foldDefault
 
-      override def process: Managed[Nothing, InputStream[Nothing, A]] =
+      override def process: Managed[Nothing, InputStream[Any, Nothing, A]] =
         for {
           index <- Ref.make(0).toManaged_
           len   = c.length
@@ -1858,7 +1858,7 @@ object ZStream extends ZStreamPlatformSpecific {
     new ZStream[R, E, A] {
       def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
 
-      override def process: ZManaged[R, E, InputStream[E, A]] =
+      override def process: ZManaged[R, E, InputStream[R, E, A]] =
         for {
           done <- Ref.make(false).toManaged_
           a    <- managed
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1643,34 +1642,39 @@ object ZStream extends ZStreamPlatformSpecific {
...
-                  case None => ZStream.fromQueue(output).unTake.fold[R1, E1, A1, S].flatMap(fold => fold(s, cont, g))
-                }
-          } yield s
-        }
+                              )
+                          )
+                        ).toManaged_
+          is <- maybeStream.fold[ZManaged[R, E, InputStream[E, A]]](ZManaged.succeed(output.take.flatten))(_.process)
+          // versus
+          // is <- maybeStream.fold[ZManaged[R, E, InputStream[E, A]]](output.take.toManaged_)(_.process)
```
iravid: The difference is subtle, but the difference is this:
- `output.take` is `ZIO[Any, Nothing, IO[Option[E], A]`. The outer layer is a program that when run does a dequeue; the inner is just a value that encodes success/failure/end. When you flatten them, you get a program that does a dequeue and results in a value that encodes success/failure/end. When you wrap that in ZManaged, you just get a scope that gives you that program, which you can repeatedly run.
- `output.take.toManaged_` moves the dequeue operation to ZManaged, and that ZManaged yields the value. But since when we run ZStreams we only "enter" the ZManaged once, you'll only dequeue once.

So these aren't equivalent. Hope this helps!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1587,11 +1587,10 @@ object ZStream extends ZStreamPlatformSpecific {
...
   final val never: Stream[Nothing, Nothing] =
     new Stream[Nothing, Nothing] {
-      def fold[R1, E1, A1, S]: Fold[R1, E1, A1, S] =
-        ZManaged.succeed { (s, cont, _) =>
-          if (!cont(s)) ZManaged.succeed(s)
-          else ZManaged.never
-        }
+      def fold[R, E, A, S]: Fold[R, E, A, S] = foldDefault
+
+      override def process: Managed[Nothing, InputStream[Nothing, Nothing]] =
```
iravid: Looks good

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1643,34 +1642,39 @@ object ZStream extends ZStreamPlatformSpecific {
...
-                                    _.failureOrCause match {
-                                      case Left(None)    => output.offer(Take.End).unit
-                                      case Left(Some(e)) => output.offer(Take.Fail(Cause.fail(e))).unit
-                                      case Right(cause)  => output.offer(Take.Fail(cause)).unit
-                                    },
-                                    a => output.offer(Take.Value(a)).unit
-                                  )
+      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
+
+      override def process: ZManaged[R, E, InputStream[E, A]] =
```
iravid: Implementation looks good!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1760,15 +1764,14 @@ object ZStream extends ZStreamPlatformSpecific {
...
   final def finalizer[R](finalizer: ZIO[R, Nothing, _]): ZStream[R, Nothing, Nothing] =
     new ZStream[R, Nothing, Nothing] {
-      def fold[R1 <: R, E1, A1, S]: ZStream.Fold[R1, E1, A1, S] =
-        ZManaged.succeed { (s, _, _) =>
-          ZManaged.reserve(Reservation(UIO.succeed(s), _ => finalizer))
-        }
+      def fold[R1 <: R, E, A1, S]: Fold[R1, E, A1, S] = foldDefault
+
+      override def process: ZManaged[R, Nothing, InputStream[Nothing, Nothing]] =
+        ZManaged.succeed(IO.fail(None)).ensuring(finalizer)
```
iravid: Let's introduce a companion object for `InputStream` with common constructors:
```
object InputStream {
  val end = IO.fail(None)
  def emit[A](a: A) = ...
  def fail[E](e: E) = ...
  def halt[E](c: Cause[E]) = ...
  def die(t: Throwable) = ...
  def dieMessage(m: String) = ...
  def done[E, A](e: Exit[E, A])  = ...
}
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1792,23 +1795,22 @@ object ZStream extends ZStreamPlatformSpecific {
...
     new Stream[Nothing, A] {
-      def fold[R1, E1, A1 >: A, S]: Fold[R1, E1, A1, S] =
-        ZManaged.succeed { (s, cont, f) =>
-          ZManaged.fromEffect(c.foldMLazy(s)(cont)(f))
+      def fold[R, E, A1 >: A, S]: Fold[R, E, A1, S] = foldDefault
+
+      override def process: Managed[Nothing, InputStream[Nothing, A]] =
+        for {
+          ref <- Ref.make(c).toManaged_
+        } yield ref.get.flatMap { chunk =>
```
iravid: Slightly more efficient:
- read the chunk length
- keep the last read index in the ref
- bump it on every pull

`chunk.drop(1)` will allocate a new `Chunk.Slice` on every pull

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1858,10 +1860,15 @@ object ZStream extends ZStreamPlatformSpecific {
...
     new ZStream[R, E, A] {
-      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] =
-        ZManaged.succeed { (s, cont, f) =>
-          if (!cont(s)) ZManaged.succeed(s)
-          else managed.mapM(f(s, _))
+      def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] = foldDefault
+
+      override def process: ZManaged[R, E, InputStream[E, A]] =
+        for {
+          p <- Promise.make[Nothing, Unit].toManaged_
```
iravid: Why not just `Ref[Boolean]` that represents if the `InputStream` has been pulled or not?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1902,19 +1909,23 @@ object ZStream extends ZStreamPlatformSpecific {
...
-              f0(s) flatMap {
-                case None => ZIO.succeed(s -> s2)
-                case Some((a, s)) =>
-                  f(s2, a).flatMap(loop(s, _))
-              }
+      def fold[R1 <: R, E1 >: E, A1 >: A, S2]: Fold[R1, E1, A1, S2] = foldDefault
 
-          ZManaged.fromEffect(loop(s, s2).map(_._2))
-        }
+      override def process: ZManaged[R, E, InputStream[E, A]] =
```
iravid: Very nice

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1683,27 +1687,27 @@ object ZStream extends ZStreamPlatformSpecific {
...
+                      a => output.offer(UIO.succeed(a))
                     )
-                ).toManaged_
-            s <- ZStream.fromQueue(output).unTake.fold[R1, E1, A1, S].flatMap(fold => fold(s, cont, g))
-          } yield s
-        }
+                  )
+              ).toManaged_
+          is <- ZManaged.succeed(output.take.flatten)
+        } yield is
```
iravid: You can just yield `output.take.flatten` directly here

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Basically keep the index in a Ref?

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: It lives directly inside ZStream companion object alongside the type alias right?

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Sure, no preference, just wanted to have some fun.

vasilmkd:
Constructors are more or less done.

I will address the comments tomorrow and start going through the combinators.

StreamPure will not be a part of this PR.

jdegoes:
@vasilmkd @iravid 

Two quick comments:

1. I think we do need `R` inside `InputStream`, at least in the general case.
2. `PureStream` will probably disappear because fusion is different in this version; we'll need to create the new "pure" version in a different way, without the `foldPure` machinery, which won't be useful after the conversion is complete.
3. This work to me suggests we change from `Queue[Take[...]]` to `ZQueue[Any, Option[E], A]` (not initially, but something to keep in mind for things like `toQueue`, etc.).

iravid:
Agreed on all points @jdegoes. Also, the ZQueue work finally pays off! 🎉🚀

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Ah, ok, it's just much less efficient :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Yup!

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Yeah.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1721,40 +1724,33 @@ object ZStream extends ZStreamPlatformSpecific {
...
-                      .ensuring(canceler)
-                }
-          } yield s
-        }
+                               )
+                           )
+                         ).toManaged_
+          is <- eitherStream match {
+                 case Left(canceler) =>
+                   ZManaged.succeed(output.take.flatten).ensuring(UIO.effectTotal(println("canceler")) *> canceler)
```
iravid: Rogue `println` here

vasilmkd: Yes, this code seemed dubious to me when I wrote it, but it actually works properly. I'll remove it of course.

iravid: Read it as: "when we exit the scope defined by this InputStream, be sure to run this canceller"

vasilmkd: Yesss, we are running only the inner program in a loop, the scope stays the same. Great!

vasilmkd:
@iravid Can the queues in `effectAsync` be changed to `ZQueue[Any, Option[E], A]`? I've never used the full type alias.

/cc @jdegoes

vasilmkd:
> @iravid Can the queues in `effectAsync` be changed to `ZQueue[Any, Option[E], A]`? I've never used the full type alias.
> 
> /cc @jdegoes

I guess not really, because InputStream is an effect that needs to be run, and both the error channel and success have meaning. I guess you can't enqueue an Option on the other case.

iravid:
You can "push" the flatten into the queue by doing `Queue[InputStream[E, A]].bounded(16).mapM(identity)`

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Thanks again for this explanation.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Great, changing it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Thanks for teaching me the environment trick.

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: `halt` and `done` actually need `Option[E]` as the error type parameter of the `Cause` and `Exit` respectively. What should I do?

iravid: `halt = ZIO.halt(c).mapError(Some(_)), done = ZIO.done(e).mapError(Some(_))`

vasilmkd: Done. Thanks.

PR #1401, opened 2019-08-14
Convert more ZStream combinators to use ZStream#process

iravid:
The `buffer` test changed but that's benign because there's now an additional buffer introduced.

@vasilmkd


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -671,18 +688,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-        ZManaged.succeed { (s, cont, f) =>
-          self.fold[R2, E2, A, S].flatMap { foldOuter =>
-            foldOuter(s, cont, (s, a) => {
-              f0(a)
-                .fold[R2, E2, B1, S]
-                .flatMap { foldInner =>
-                  foldInner(s, cont, f)
+        foldDefault
+
+      override def process: ZManaged[R1, E1, ZStream.InputStream[R1, E1, B]] =
```
iravid: @jdegoes could you review this?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1858,12 +1909,22 @@ object ZStream extends ZStreamPlatformSpecific {
 
       override def process: ZManaged[R, E, InputStream[R, E, A]] =
         for {
-          done <- Ref.make(false).toManaged_
-          a    <- managed
-        } yield done.get.flatMap {
-          if (_) InputStream.end
-          else done.set(true) *> InputStream.emit(a)
-        }
+          doneRef      <- Ref.make(false).toManaged_
```
iravid: @jdegoes and this

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -671,18 +688,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-                .flatMap { foldInner =>
-                  foldInner(s, cont, f)
+        foldDefault
+
+      override def process: ZManaged[R1, E1, ZStream.InputStream[R1, E1, B]] =
+        for {
+          ref <- Ref.make[Option[(InputStream[R1, E1, B], Exit[_, _] => ZIO[R1, Nothing, Any])]](None).toManaged_
+          as  <- self.process
+          _   <- ZManaged.finalizerExit(e => ref.get.flatMap(_.map(_._2).getOrElse((_: Exit[_, _]) => UIO.unit).apply(e)))
+          pullOuter = ZIO.uninterruptibleMask { restore =>
```
jdegoes: Do we really need to put the pull of `a` inside `uninterruptible`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: About as terrible as I suspected `flatMap` would be... 😆  not all of course will increase, on average I expect we'll come out with much less code.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -671,18 +688,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+              } yield ()).mapError(Some(_))
+            }
+          }
+          bs = {
+            def go: InputStream[R1, E1, B] = ref.get.flatMap {
+              case None => pullOuter *> go
+              case Some((isB, finalizer)) =>
+                isB.catchAll {
+                  case e @ Some(e1) => (finalizer(Exit.fail(e1)) *> ref.set(None)).uninterruptible *> ZIO.fail(e)
+                  case None         => (finalizer(Exit.succeed(())) *> ref.set(None)).uninterruptible *> pullOuter *> go
```
jdegoes: Ensuring the `finalizer` is not interrupted is one thing, we also really need to ensure the `finalizer` will be run in all cases, i.e. that `go` will not be interrupted right before it even begins execution of the finalizer.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: The pull of `a` is restored, so it it'll run interruptibly. The rest has some regions that have to run uninterruptibly (e.g. storing the finalizer)

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Ouch, good point. Will use `uninterruptibleMask` here too.

iravid: @jdegoes Oh, I actually handled this. If `go` gets interrupted, the `ZManaged` will execute the stored finalizer (line 697).

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1858,12 +1909,22 @@ object ZStream extends ZStreamPlatformSpecific {
...
-          done <- Ref.make(false).toManaged_
-          a    <- managed
-        } yield done.get.flatMap {
-          if (_) InputStream.end
-          else done.set(true) *> InputStream.emit(a)
-        }
+          doneRef      <- Ref.make(false).toManaged_
+          finalizerRef <- Ref.make[Exit[_, _] => ZIO[R, Nothing, Any]](_ => UIO.unit).toManaged_
+          _            <- ZManaged.finalizerExit(e => finalizerRef.get.flatMap(_.apply(e)))
+          pull = ZIO.uninterruptibleMask { restore =>
```
jdegoes: If we had a helper:

```scala
object InputStream {
  def singleton(a: A): UIO[InputStream[Any, Nothing, A]] = 
    Ref.make[Option[A]](Some(a)).map(ref =>
      ref.modify {
        case None => ZIO.fail(()) -> None
        case Some(a) => ZIO.succeed(a) -> None
      }.flatten)
```

Then it might make this one easier to verify.

Thinking what you want to do is map the `A` inside the managed into the input stream. The current version seems quite aggressive about uninterruptibility...

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Ok, sounds good.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: That helper looks useful!

The logic for this constructor is almost what you wrote; I want the managed to allocate the `A` only when the InputStream is pulled. Does that make sense?

jdegoes: It does, but I can't convince myself there's a difference because of the contract on the original `ZManaged[..., A]` that is being lifted.

iravid: I *think* (not sure yet!) it matters when composing streams together. You're lifting a managed resource to a single-element stream and then flatMap it to create a multiple-element stream. So it should only allocate when you pull the composed stream.

vasilmkd:
Linking to the original discussion. #1388

vasilmkd:
We should rebase and run the full test suite.

Some possible regressions discussed here: https://github.com/zio/zio/pull/1356

vasilmkd:
I feel like the scope of this PR is getting a bit broad. I'm taking a look.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1160,27 +1161,14 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
...
-    val list: List[String] = slurp(merge).toEither.fold(
+    val list: List[String] = unsafeRunSync(merge.runCollect).toEither.fold(
       _ => List.empty,
       identity
     )
 
     list must containTheSameElementsAs(List("1", "2", "1", "2"))
   }
 
-  private def mergeWithShortCircuit = {
```
vasilmkd: Why is this test being dropped completely? Should we put a TODO or something?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -548,18 +555,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-          self.fold[R2, E2, A2, S].flatMap { foldLeft =>
-            foldLeft(s, cont, f).flatMap { s =>
-              if (!cont(s)) ZManaged.succeed(s)
-              else
-                other.fold[R2, E2, A2, S].flatMap(foldRight => foldRight(s, cont, f))
-            }
-          }
-        }
-    }
+    Stream(self, other).flatMap(identity)
```
vasilmkd: Should have been like this from the start. Nice.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -593,17 +588,25 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+
+    override def process =
+      for {
+        as              <- self.process
+        keepDroppingRef <- Ref.make(true).toManaged_
+        pull = {
+          def go: InputStream[R, E, A] =
+            as.flatMap { a =>
+              keepDroppingRef.get.flatMap { keepDropping =>
+                if (!keepDropping) ZIO.succeed(a)
```
vasilmkd: `InputStream.emit(a)`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1331,19 +1369,20 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+    new ZStream[R, E, A] {
+      override def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] =
+        foldDefault
+
+      override def process =
+        for {
+          as      <- self.process
+          counter <- Ref.make(0).toManaged_
+          pull = counter.modify { c =>
+            if (c >= n) (ZIO.fail(None), c)
```
vasilmkd: `InputStream.end`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1590,7 +1627,12 @@ object ZStream extends ZStreamPlatformSpecific {
    * The empty stream
    */
   final val empty: Stream[Nothing, Nothing] =
-    StreamPure.empty
+    new Stream[Nothing, Nothing] {
+      def fold[R, E, A, S]: Fold[R, E, A, S] = foldDefault
+
+      override def process: Managed[Nothing, InputStream[Any, Nothing, Nothing]] =
+        ZManaged.succeed(ZIO.fail(None))
```
vasilmkd: `InputStream.end`

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1766,7 +1808,11 @@ object ZStream extends ZStreamPlatformSpecific {
       def fold[R1 <: R, E, A1, S]: Fold[R1, E, A1, S] = foldDefault
 
       override def process: ZManaged[R, Nothing, InputStream[R, Nothing, Nothing]] =
-        ZManaged.succeed(InputStream.end).ensuring(finalizer)
+        for {
+          finalizerRef <- Ref.make[ZIO[R, Nothing, Any]](UIO.unit).toManaged_
+          _            <- ZManaged.finalizer[R](finalizerRef.get.flatten)
+          pull         = (finalizerRef.set(finalizer) *> ZIO.fail(None)).uninterruptible
```
vasilmkd: `InputStream.end`

iravid:
Agreed about the scope @vasilmkd - will get this in today :-)

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
iravid: Initially I thought it's not an important test when we switch to `process`, but I guess more coverage doesn't hurt.

ghostdogpr:
Not sure this change is responsible but `fast producer progress independently` seems flaky now:
https://circleci.com/gh/zio/zio/16869
https://circleci.com/gh/zio/zio/16868

iravid:
Most likely. Will look into that now.

iravid:
@ghostdogpr should be fixed in #1409

ghostdogpr:
@iravid masted build after that merge had no failures 👏

Issue #1403, opened 2019-08-15
`test_sbt`: mute regular console output from DefaultTestRunner


dkarlinsky:
`RunnableSpec`s come wired with their own `TestRunner` which has a `TestReporter`. 
This works great for running them via their object's `main` method, but under `sbt` it generates duplicated output like the following:
```
+ ZQueueSpec
  + simple offer and take
  + simple offer and take 2
[info] + ZQueueSpec
[info]   + simple offer and take
[info]   + simple offer and take 2
[info] ZIO Test
[info] Done
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
```
SBT test runner should probably run specs with a custom "muted" `TestReporter`, to prevent this.
cc: @ghostdogpr

dkarlinsky:
PR to follow soon

ghostdogpr:
Fixed by #1405

PR #1405, opened 2019-08-15
Fix for #1403 - mute regular console output from DefaultTestRunner

dkarlinsky:
* Made TestRunner a case class, so that reporter can be replaced.
* New `RunnableSpec` method: runWith(reporter).
* Also removed dependency by `testRunner` on `coreTests` (wasn't required).

#1403 

cc: @ghostdogpr

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -172,7 +172,6 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
   .jvmSettings(libraryDependencies ++= Seq("org.scala-sbt" % "test-interface" % "1.0"))
   .dependsOn(core % "test->test;compile->compile")
   .dependsOn(test % "test->test;compile->compile")
-  .dependsOn(coreTests % "test->test;compile->compile")
 
 lazy val testRunnerJVM = testRunner.jvm
 lazy val testRunnerJS  = testRunner.js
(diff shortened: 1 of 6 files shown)


ghostdogpr:
Nice 👍 
The CI fails because of an unrelated regression, it should be okay once #1404 is merged.

dkarlinsky:
@ghostdogpr thanks for the heads up!
I was trying to figure out how my change could've broken ` monad associativity` of `Gen` :)

ghostdogpr:
I just merged the fix, so a rebase should solve it.

ghostdogpr:
Thanks!

kaishh:
@dkarlinsky 
Hmm, that seems somewhat backwards, to be honest, is there no way to preserve the usefulness of supplying reporter to the spec and invert control such that the `spec` controls reporting when its ran from SBT, instead of being silenced?

ghostdogpr:
Would it make sense that we just pass an implementation of `Console` that uses sbt logger? So whatever is written using `Console` in the Reporter gets proper output, and there’s no need to output again in test-sbt?

kaishh:
Makes sense to me!

dkarlinsky:
@kaishh , @ghostdogpr how about this: https://github.com/zio/zio/pull/1426 ?

PR #1406, opened 2019-08-15
Loosen type constraints for either predicates

ioleo:
Current `isLeft` and `isRight` predicates are not very usable.

Example compile error when asserting on `IO[FooError, Unit].either`
```
[error] /.../ExampleSpec.scala:72:43: type mismatch;
[error]  found   : zio.test.Predicate[Either[FooError,Nothing]]
[error]  required: zio.test.Predicate[Either[FooError,Unit]]
[error] } yield assert(out, isLeft(isEqual(FooError)))
```


ghostdogpr:
Thanks!

PR #1407, opened 2019-08-16
ZIO Test: Improve Support for Sized Generators

adamgfraser:
Following up on discussion in #1356, implements sized generators as generators with a dependence on an environment `Size`. Adds basic combinators for sized collections and creating other sized generators. Not implemented yet are changes to the test runner to run sized tests or different common distributions of sizes (e.g. "small", "medium", "large").

Definitely open to different approaches to this, just thought it would be easier to discuss if we had something to react to.


adamgfraser:
@jdegoes I updated this to use a `FiberRef` as you suggested. I kept the name as `Size` for now. I'm a bit torn because I like something shorter like `Size` or `Sized` to keep things concise when working with `Gen`, where it is also obvious that the size you are talking about is with respect to the generator. But I see your point about namespacing when using `Size` in `MockEnvironment`, especially if we end up adding other dependencies.

One other idea is we could do something like:

```scala
trait SGen[-R, +A] = Gen[R with Size, A]
```

What do you think?

jdegoes:
@adamgfraser I like `Size` or perhaps `Sized`, they're short; the only real question is whether the user is likely to have a service of such a name (`size`) in their own `R`. Given how common the name `size` is, that was my only concern. 

Basically if a user had some service like this:

```scala
trait CacheConfig {
  def size: Int 
}
```

in the methods they are testing, then they would not be able to use mock environment or the generators with `Size` as a requirement, because `size: Int` and `size: FiberRef[Int]` cannot be unified (nor would you want them to, you don't want to tie one to another).

Still, these concerns are relatively minor and hopefully users have better namespacing discipline than putting a top-level `size` in their `R`.

adamgfraser:
@jdegoes I really like those suggestions! What do you think about using a name for `locally` that is more specific to our domain? Maybe `withSize`?

jdegoes:
Great idea! 👍

adamgfraser:
@jdegoes I incorporated your latest suggestions.

adamgfraser:
Thanks! No tech debt allowed here!

PR #1409, opened 2019-08-16
Convert ZStream#collectWhile to use ZStream#process

iravid:
cc @vasilmkd


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -506,18 +506,22 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+      foldDefault
+
+    override def process: ZManaged[R, E, ZStream.InputStream[R, E, B]] =
+      for {
+        as   <- self.process
+        done <- Ref.make(false).toManaged_
+        pfIO = pred.andThen(ZIO.succeed(_))
+        pull = for {
+          alreadyDone <- done.get
+          result <- if (alreadyDone) ZIO.fail(None)
```
vasilmkd: `InputStream.end`. Same in line 521.

vasilmkd:
Looks good.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Right!

PR #1410, opened 2019-08-16
Convert ZStream#peel to use ZStream#process

iravid:
@vasilmkd 

cc @jdegoes - say goodbye to that monstrosity!

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1163,92 +1163,35 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
 
   /**
    * Peels off enough material from the stream to construct an `R` using the
-   * provided `Sink`, and then returns both the `R` and the remainder of the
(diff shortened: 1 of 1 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1858,6 +1801,26 @@ object ZStream extends ZStreamPlatformSpecific {
   final def fromEffect[R, E, A](fa: ZIO[R, E, A]): ZStream[R, E, A] =
     managed(fa.toManaged_)
 
+  /**
+   * Creates a stream from an [[InputStream]].
+   */
+  final def fromInputStream[R, E, A](is: InputStream[R, E, A]): ZStream[R, E, A] =
```
vasilmkd: Why not implement in terms of `fromInputStreamManaged`?

iravid: Yes, good idea!

jdegoes:
@iravid Hurray! 🎉

Issue #1408, opened 2019-08-16
Revive StreamPure after Stream#process migration


iravid:

vasilmkd:
Work already started.

jdegoes:
Here's a sketch of the idea for embedding exceptions:

```scala
trait StreamEffect[+E, +A] {
  private[zio] def processEffect: Managed[E, () => A]

  def process: Managed[E, A => IO[Option[E], A]] = ???
}
object StreamEffect {
  case class StreamEffectFail[E](value: A) extends Throwable(value.toString, null, true, false)
  case object StreamEffectStop extends Throwable("Stream stop", null, true, false)
}
```

Inside `process` implementation, you `try` / `catch` the `() => A`: if it is `StreamEffectStop, you treat it as `InputStream.end`. If it is `StreamEffectFail`, you treat it as `InputStream.fail(e)` (a cast will be necessary; which makes this unsafe to expose to users).

Now the effectful streams can indicate failure or they can stop. Moreover, combinators on `StreamEffect` can use the lower-level representation for more efficient implementations.

Make sense? /cc @vasilmkd @iravid

jdegoes:
Corresponding to the new ZSink being essentially `InputStream[R, E, A] => ZIO[R, E, (A0, B)` (albeit modified because this one will infer horribly), we could also introduce a `ZSinkEffect` (package private) that would be something like `(() => A) => (A0, B)` (_give me an iterator, I'll give you the remainder and a B_). This will allow highly-performant specializations for stream / sink effect variants.

The applications for `ZStreamEffect` / `ZSInkEffect` would be limited to:

 - Pure unfolds
 - Pure "collection-style" data
 - Conversion from synchronous APIs, such as Java's `InputStream`
 - etc.

Probably the `InputStream` is the killer use-case, although the possibility of using `ZStream` as a more efficient collection (for larger data with laziness, interruption, etc.) might be somewhat interesting.

iravid:
@jdegoes the `A => IO[Option[E], A]]` is a typo, right? Should be `IO[Option[E], A]`?

vasilmkd:
Can we make a list of which combinators and constructors fall under `StreamEffect`?

vasilmkd:
@jdegoes can you explain why the cast is necessary for `StreamEffectFail(e: E)`?

jdegoes:
[...]

PR #1411, opened 2019-08-16
Remove StreamPure

vasilmkd:
Ongoing effort #1388.

StreamPure, StreamChunkPure and SinkPure will return in #1408.


vasilmkd:
Always happy to delete code.

vasilmkd:
> Ongoing effort #1388.
> 
> StreamPure, StreamChunkPure and SinkPure will return in #1408.

Hahahhaha, thanks for announcing their return. @iravid

PR #1413, opened 2019-08-16
Convert even more ZStream combinators to process

iravid:
@vasilmkd

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -909,17 +909,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
    * Repeats this stream forever.
    */
   def forever: ZStream[R, E, A] =
-    new ZStream[R, E, A] {
-      override def fold[R1 <: R, E1 >: E, A1 >: A, S]: Fold[R1, E1, A1, S] =
-        ZManaged.succeed { (s, cont, f) =>
-          def loop(s: S): ZManaged[R1, E1, S] =
-            self.fold[R1, E1, A, S].flatMap { fold =>
-              fold(s, cont, f).flatMap(s => if (cont(s)) loop(s) else ZManaged.succeed(s))
-            }
-
-          loop(s)
-        }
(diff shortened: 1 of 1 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -557,8 +558,8 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
    * Appends another stream to this stream. The concatenated stream will first emit the
    * elements of this stream, and then emit the elements of the `other` stream.
    */
-  final def concat[R1 <: R, E1 >: E, A1 >: A](other: ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
-    Stream(self, other).flatMap(identity)
+  final def concat[R1 <: R, E1 >: E, A1 >: A](other: => ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
+    ZStream(UIO.succeed(self), UIO(other)).flatMap(ZStream.unwrap)
```
iravid: @jdegoes is there another combinator constellation we could use to make this nicer? e.g. a lazy constructor like `construct(head: A, rest: => List[A])`

vasilmkd: Awesome trick.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -909,17 +909,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
-        ZManaged.succeed { (s, cont, f) =>
-          def loop(s: S): ZManaged[R1, E1, S] =
-            self.fold[R1, E1, A, S].flatMap { fold =>
-              fold(s, cont, f).flatMap(s => if (cont(s)) loop(s) else ZManaged.succeed(s))
-            }
-
-          loop(s)
-        }
-    }
+    self ++ forever
```
vasilmkd: And it's logical conclusion.

vasilmkd:
Looks good to me.

PR #1415, opened 2019-08-16
Fix comments on methods of ZIOFunctions trait

heraklos:
It seems some comments on ZIOFunctions' methods explain about different methods.


ghostdogpr:
Thanks for the fix!

PR #1416, opened 2019-08-16
Add isUnit predicate

ioleo:
`Task[Unit]` is a very common type for commands/statements (if we prefer more imperative style) or at least at the impure edges of our program (if we prefer functional style). Either way its cumbersome to assert with `isEqual(())`. In pair with `ZIO.unit` we should add `isUnit` for convinience.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Predicate.scala
+++ b/test/shared/src/main/scala/zio/test/Predicate.scala
@@ -296,6 +296,17 @@ object Predicate {
   }
 
   /**
+   * Makes a new predicate that requires the value be unit.
+   */
+  final def isUnit: Predicate[Any] =
+    Predicate.predicate(s"isUnit") { actual =>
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Predicate.scala
```diff
@@ -295,6 +295,17 @@ object Predicate {
     if (actual) Assertion.success else Assertion.failure(())
   }
 
+  /**
+   * Makes a new predicate that requires the value be unit.
+   */
+  final def isUnit: Predicate[Any] =
```
jdegoes: Is this inserted alphabetically?

ioleo: I believe it is. @jdegoes

ghostdogpr:
Thanks!

PR #1417, opened 2019-08-16
Convert ZStream parallel combinators to process

vasilmkd:
More combinators. #1388.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -803,51 +803,50 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
     f: A => ZStream[R1, E1, B]
   ): ZStream[R1, E1, B] =
     new ZStream[R1, E1, B] {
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1073,32 +1071,26 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+      override def process =
+        for {
+          out              <- Queue.bounded[InputStream[R1, E1, B]](n).toManaged(_.shutdown)
+          permits          <- Semaphore.make(n.toLong).toManaged_
+          interruptWorkers <- Promise.make[Nothing, Unit].toManaged_
+          _ <- self.foreachManaged { a =>
+                for {
+                  p <- Promise.make[E1, B]
+                  _ <- out.offer(InputStream.fromPromise(p))
+                  _ <- (f(a).to(p) race interruptWorkers.await).fork
```
iravid: `withPermit` is missing here on the `f(a).to(p)`

vasilmkd: Yes, mistake when reverting code...

PR #1419, opened 2019-08-16
Last but not least ZStream combinators conversion to process

vasilmkd:
#1388.
@iravid


adamgfraser:
Amazing how quickly you guys have gone through these.

PR #1426, opened 2019-08-17
Tweak to #1405 - instead of silent TestReporter pass one that reports to SBT

dkarlinsky:
@ghostdogpr , @kaishh How about this?

### DIFF
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
@@ -41,11 +41,14 @@ abstract class BaseTestTask(val taskDef: TaskDef, testClassLoader: ClassLoader) 
         }
       }
 
-    for {
-      result   <- spec.runWith(TestReporter.silent)
-      rendered = DefaultTestReporter.render(result.mapLabel(_.toString))
-      _        <- reportResults(rendered)
-    } yield ()
+    val testReporter: TestReporter[spec.Label] = result => {
+      val rendered = DefaultTestReporter.render(result.mapLabel(_.toString))
+      reportResults(rendered)
+        .catchAll(_ => ZIO.unit)
+        .unit
+    }
+
+    spec.runWith(testReporter)
   }
 
   override def execute(eventHandler: EventHandler, loggers: Array[Logger]): Array[Task] = {
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
@@ -1,11 +1,8 @@
 package zio.test.sbt
 
-import java.io.IOException
-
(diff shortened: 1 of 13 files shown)


test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
```diff
@@ -41,11 +41,14 @@ abstract class BaseTestTask(val taskDef: TaskDef, testClassLoader: ClassLoader)
         }
       }
 
-    for {
-      result   <- spec.runWith(TestReporter.silent)
-      rendered = DefaultTestReporter.render(result.mapLabel(_.toString))
-      _        <- reportResults(rendered)
-    } yield ()
+    val testReporter: TestReporter[spec.Label] = result => {
+      val rendered = DefaultTestReporter.render(result.mapLabel(_.toString))
```
ghostdogpr: What would be nice is if it could use `spec.runner.reporter` instead of overriding it with the default one, to respect what's defined by the user in the test. Maybe users might change the way their tests are reported and it's pity to lose it when using sbt.

We could change `TestReporter` to be `ExecutedSpec[L] => URIO[Console, Unit]`, remove the `Console` passed in `apply` and instead pass it when we `run`. When we run it from sbt, we would pass an implementation of `Console` that prints to sbt (`log.info`), otherwise we would just pass `Console.Live`.

dkarlinsky: Well `Console` doesn't map exactly to the sbt logger API.
First there is no way to implement a `putStr` - `log.info()` is `putStrLn`.
I guess I could accumulate the calls to `putStr` until I see `putStrLn` but this feels unnecessarily complicated...  
Also there is nothing for `getStrLn`. Of course there is no reason for a `TestReporter` to use it, but still, it doesn't feel very clean...

ghostdogpr: Maybe we extend `Console.Live` and only override `putStrLn`?
That way you get the expected behavior and only when appending a full line you forward it to sbt.

kaishh: > Well Console doesn't map exactly to the sbt logger API.

Does anything actually pins us to using Console instead of creating a separate interface?

ghostdogpr: Yep we can just use a new one 👍

dkarlinsky: Ok, I changed the definition of `TestReporter` to:
```scala
 type TestReporter[-L] = ExecutedSpec[L] => URIO[TestLogger, Unit]
```
Where `TestLogger.Service` looks like this:
```scala
trait Service {
    def logLine(line: String): UIO[Unit]
 }
```

Issue #1208, opened 2019-07-21
ZIO Test (remove Specs2 from project)


softinio:
Currently we cannot:
- Run dotty tests due to use of s2 macros in specs2
- Can't run tests in scala native due to no specs2 scala native compatibility. 

Should we explore an alternative? 

It was nice to see: 

https://twitter.com/RichardWhaling/status/1153004972012912640?s=20

jdegoes:
We need to move away from Specs2. I think it's time for _ZIO Test_. 😄

softinio:
ZIO Test is a great idea! I am in!

Fresh start or can we leverage: https://github.com/scalaz/testz

NeQuissimus:
Don't forget about #723 😜

pshirshov:
Hey guys,

Recently I've been rewriting our distage testkit  to support dependency sharing across tests. I hit all the possible limits of ScalaTest then tried to check if any other existing framework can satisfy our needs - and unfortunately they all cannot. 

Right now I have a proof of concept framework on top of scalatest with a lot workarounds. So, we are seriously considering writing our own one from scratch.

From my point of view an ideal framework should:

1. Allow user to implement custom discovery and registration (including static ones)
2. Allow user to implement custom DSLs
3. Allow user to implement custom initialization of arbitrary complexity (yes, we really need it, more details [here](https://github.com/scala/scala-dev/issues/641#issuecomment-511139819))
4. Online notifications about test lifecycle events

So, essentially such a framework ("metaframework") should 

1. Take care of all the reporting matters
2. Provide a test registry, allowing user to store additional/arbitrary metadata with each test
3. Provide a set of building blocks for implementing more specific "testkits"
4. Provide entrypoints/launchers and CLI option parser for all the target platforms
5. Does not depend on a single platform/compiler entirely (we need jvm for 2.12/2.13/dotty, and sjs for 2.12/2.13)

Another pain is IDE integration which includes:

1. Test ID extractors
2. Test report viewer

Each new framework has to implement at least these components for IDEA/VSCode and it's [not always trivial](https://github.com/JetBrains/intellij-scala/tree/e6441195268335090ae3278f9ad3732f232c8962/scala/runners/src/org/jetbrains/plugins/scala/testingSupport). It may be a good idea to try to build an abstraction similar to LSP for tests.

[...]

PR #1428, opened 2019-08-17
minimal scala native support - resolves ##639

rwhaling:
This provides support for Scala Native in zio core, utilizing our libuv executionContext scala-native-loop, and a compatability shim for the scalajs timer API.  Since we are line-for-line compatible with Scala.js for the purpose of Zio, I've imported the scalajs sources into the native subproject - no new code.

Open questions are -
1. Is it wise to implement just core and stacktracer to start, or should we try to cover more of the subprojects before merging?
2. Would it be better to copy/paste the scala.js platform code?
3. This published local artifacts fine with `coreNative/publishLocal` and `stacktracerNative/publishLocal`, but thats probably not enough for CI, right?


jdegoes:
@rwhaling Excited to get this merged! Can you run `fmt` to get pass the linter?

Meanwhile, any ZIO contributors / users familiar with Native and willing to give this a review / spin?

neko-kai:
briefly hijacked for fmt

neko-kai:
I guess we should wait until all tests migrate to ZIO Test since we can't (seemingly ?) run current tests

rwhaling:
thanks @neko-kai! Let me see what I can do about the tests.

adamgfraser:
Is there anything we need to do as we build out ZIO Test to make sure we can support Scala Native as well? Or since you are compatible with Scala.js are we good as long as we can support that?

rwhaling:
@adamgfraser - that's a really interesting question.  Testing on Native is kind of fiddly and a pain, since it is so far from what sbt expects - let me consult with a few other SN contributors and get back to you?

adamgfraser:
@rwhaling That would be great! We currently have an SBT runner but specifications are also their own executable programs. So if SBT was the issue you wouldn't necessarily need to use SBT to use ZIO Test. You can try doing `test:run` in the `coreTestsJVM` project to see what that looks like.

sideeffffect:
this is no longer blocked by https://github.com/zio/zio/issues/1208

also, this issue deserves the [`scala-native` badge](https://github.com/zio/zio/issues?q=label%3Ascala-native)

sideeffffect:
@rwhaling the PR is green and approved. Only some conflicts with sbt remain. Can you resolve them? Or would you like me to do that?

rwhaling:
@sideeffffect thanks for nudging this along, this PR fell off my radar a bit; have also been triaging some issues that affect this implementation over in SN.  

I've also seen a few quirks from the way I am re-using the JS tree verbatim here:
https://github.com/zio/zio/pull/1428/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR92

It looks like it is pulling in the wrong executionContext because of this, which would break a lot of things in a more complex program.

What I propose to do is fix the SBT conflicts, but also add a proper implementation of coreNative - it would be 90% identical to JS, but I think it'll be better in the long-run. Does that sound like a good plan to you?

rwhaling:
@sideeffffect I can take a first pass over the weekend - if you can review and help get it over the finish line, it shouldn't be a ton of work for either of us.

rwhaling:
@sideeffffect - I updated my PR, fixed the merge conflicts, but now it won't compile - specs2 4.8.0 isn't published for native yet: https://mvnrepository.com/artifact/org.specs2/specs2-core

Writing up an issue now so we can get this merged.

adamgfraser:
@rwhaling We're in the process of migrating away from Spec2 to ZIO Test and there are currently only two files that use Specs 2. We have an open ticket for that in #1647 so one solution is to just complete that migration.

But in advance of that would it be possible to just move those two files to a separate directory and include settings in `BuildHelper.scala` to only compile those two test files on JVM / ScalaJS? Then we could alter the `build.sbt` settings so that Specs2 was only included as a dependency for JVM and ScalaJS. Since it is only two files this would not result in much loss in test coverage and would let us push forward with this initiative immediately.

What do you think?

rwhaling:
@adamgfraser I'm definitely in favor of moving away from fragile dependencies.  

I like your idea a lot, but I'm probably not confident enough in my understanding of specs2 or the Zio build to execute it.  What do you think @sideeffffect?

We did get unblocked by specs2, that artifact should be up on Maven Central in an hour or two, but tbh it is likely to fail again.  

Depending on how long you think it will take to migrate those last two files, we can either merge as-is once we can get CI to pass (hopefully later today), or do the refactor if y'all think it is a quick win - but I've had this PR out for soooo long I probably want to merge as soon as is feasible.

adamgfraser:
I can try to help with the build tonight. Not sure what time zone you’re in. We have a ticket for those two tests but I don’t think anyone is working on them right now, so we shouldn’t wait for that. I’m coming into this a little from the outside so I am fine with whatever you decide to do but my experience from Dotty is there can be some weird corner cases so it would be nice if we could get most of the tests running so we could have some confidence that things were working properly.

adamgfraser:
@rwhaling We migrated the last tests to ZIO Test and just removed the Specs2 dependency so the way should be clear to merge this. Can you merge upstream changes and resolve any conflicts? Let me know if you need any help. Will be great to have this in for the 1.0.0 release!

build.sbt
```diff
@@ -199,7 +210,11 @@ lazy val stacktracerJVM = stacktracer.jvm
   .settings(dottySettings)
   .settings(replSettings)
 
-lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
+lazy val stacktracerNative = stacktracer.native
+  .settings(scalaVersion := "2.11.12")
+  .settings(skip in Test := true)
+  .settings(skip in doc := true)
+  lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
```
sideeffffect: unnecessarily changed indentation

core/native/src/main/scala/zio/internal/OneShot.scala
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Determines if the variable has been set.
+   */
+  final def isSet: Boolean = value != null
+
+  /**
+   * Retrieves the value of the variable, blocking if necessary.
+   */
+  final def get(): A = {
+    if (value == null) throw new Error("Cannot block for result to be set in JavaScript")
```
sideeffffect: this references "JavaScript"

core/native/src/main/scala/zio/scheduler/SchedulerLive.scala
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+
+package zio.scheduler
+
+import zio.ZIO
+import zio.duration.Duration
+import zio.internal.{ Scheduler => IScheduler }
+
+import scala.scalajs.js
```
sideeffffect: is it correct to import `scalajs` stuff here? (or is that what `native-loop-js-compat` is for?)

rwhaling: @sideeffffect - yep, exactly.  We have a compatability shim that presents Scala Native's async capabilities with a scalajs api.

core/native/src/main/scala/zio/internal/OneShot.scala
rwhaling: This might still be worth changing to Native

rwhaling:
@adamgfraser @sideeffffect - I should be able to do another pass Wednesday or Thursday, thanks for your patience and persistence 😄

adamgfraser:
@rwhaling Thank you!

core/native/src/main/scala/zio/internal/OneShot.scala
sideeffffect: or `native-loop`?

rwhaling:
@adamgfraser @sideeffffect - OK, up-to-date now and all tests passing.  Ready to merge?

build.sbt
```diff
@@ -95,6 +95,18 @@ lazy val coreJVM = core.jvm
 
 lazy val coreJS = core.js
 
+lazy val coreNative = core.native
+  .settings(scalaVersion := "2.11.12")
+  .settings(skip in Test := true)
+  .settings(skip in doc := true)
+  .settings(sources in (Compile, doc) := Seq.empty)
```
sideeffffect: maybe
```scala
Compile / doc / sources
```
(and similarly in other places)
but this is a trivial detail...

core/native/src/main/scala/zio/internal/Sync.scala
```diff
@@ -0,0 +1,8 @@
+package zio.internal
+
+private[zio] object Sync {
```
contrun, line 3: Can someone please explain the use of this object? I am really baffled.

PR #1432, opened 2019-08-18
The return of StreamPure as StreamEffect

### LINKED ISSUE #1408: Revive StreamPure after Stream#process migration

### LINKED ISSUE #1466: Port ZStream.unfold to StreamEffect
Follow up to #1432.

vasilmkd:
Just the basic combinators and constructors that don't need to discriminate between pure and effectful streams.

Resolves #1408 
Resolves #1466


vasilmkd:
@iravid Can you please take a look. I'm a bit confused about the failing tests.

streams/shared/src/main/scala/zio/stream/StreamPure.scala
```diff
@@ -0,0 +1,252 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.stream
+
+import zio._
+
+private[stream] trait StreamPure[+A] extends ZStream[Any, Nothing, A] { self =>
```
jdegoes: I would not call this `StreamPure`, but maybe, `StreamEffect`.

Also, I'd say the following:

 - We should permit `() => A` to throw `EndOfStreamException`, defined package-private so only we can use it
 - We should permit `() => A` to throw StreamFailedException(e: E)`, defined package-private so only we can use it
 - We should use the above to implement end of stream & failure, for both "pure streams" and synchronous effectful streams

Done properly, this will permit streams as fast as Scala collections, but lazy and potentially infinite; albeit only for blocking / synchronous, and pure streams.

Async streams are always gonna require chunking for performance.

vasilmkd:
Initial sketch. Will continue with the error implementations.

/cc @iravid @jdegoes

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,275 @@
...
+    new StreamEffect[E, A] {
+      def process = StreamEffect.super.takeWhile(pred).process
+
+      def processEffect =
+        self.processEffect.flatMap { it =>
+          Managed.effectTotal { () =>
+            {
+              val a = it()
+              if (pred(a)) a
+              else throw StreamEffect.End
```
jdegoes: Consider also:

```scala
object StreamEffect {
  private[zio] def end[A]: A = throw StreamEffect.End
}
```

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,275 @@
...
+
+  case class Failure[E](e: E) extends Throwable(e.toString, null, true, false)
+
+  case object End extends Throwable("stream end", null, true, false)
+
+  final val empty: StreamEffect[Nothing, Nothing] =
+    new StreamEffect[Nothing, Nothing] {
+      def process = ZManaged.succeed(InputStream.end)
+
+      def processEffect = Managed.effectTotal { throw End }
```
jdegoes: Wouldn't this be `() => throw End`???

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,275 @@
...
+        Managed.effectTotal {
+          val it = as.iterator
+
+          () => if (it.hasNext) it.next() else throw End
+        }
+    }
+
+  final def succeed[A](a: A): StreamEffect[Nothing, A] =
+    new StreamEffect[Nothing, A] {
+      def process =
```
jdegoes: The `var` trick you use for `take(n)` will be faster than `Ref` machinery.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Good catch, I will fix it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -687,25 +687,25 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   /**
    * Executes the provided finalizer after this stream's finalizers run.
    */
-  def ensuring[R1 <: R](fin: ZIO[R1, Nothing, _]): ZStream[R1, E, A] =
+  final def ensuring[R1 <: R](fin: ZIO[R1, Nothing, _]): ZStream[R1, E, A] =
     new ZStream[R1, E, A] {
-      override def process = self.process.ensuring(fin)
+      def process = self.process.ensuring(fin)
```
jdegoes: At some point, for stack safety, we should switch over to `class ZStream(val process: ZManaged[...])`.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,275 @@
...
+  override def filter(pred: A => Boolean): StreamEffect[E, A] =
+    new StreamEffect[E, A] {
+      def process = StreamEffect.super.filter(pred).process
+
+      def processEffect =
+        self.processEffect.flatMap { it =>
+          Managed.effectTotal {
+            def pull(): A = {
+              val a = it()
+              if (pred(a)) a else pull()
```
jdegoes: This needs a `tailrec` annotation, which I am not sure will work reliably when nested like this (the `scalac` bug may be fixed though).

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,275 @@
...
+            }
+
+            () => pull()
+          }
+        }
+    }
+
+  final def foldEffect[S](s: S)(cont: S => Boolean)(f: (S, A) => S): Managed[E, S] =
```
jdegoes: `effectFold`, perhaps? Any method that captures a side-effect must be prefixed by `effect`. Also in this case, it's not clear where the effect is—`cont`, `f`, or both?

Is this really needed? I think it will be confusing for users. Fine to use internally of course.

@iravid Any other name ideas?

vasilmkd: The method is not exported, visible only in `zio.stream` package.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Out of curiosity, would just changing the def to val be enough? Or is the trait to class change necessary as well?

jdegoes: Well, you need a way to "unwrap" and then "rewrap", rather than add one new stack frame.

Right now, the outer `process` delegates to the inner `process`, so you end up adding a stack frame for every operation. Eventually you run out of stack.

When you store it as a field that you can extract, map over, and stuff back it, it pushes 100% of the transformation to the ZIO effect type, which is already stack safe.

vasilmkd:
@jdegoes Can you comment on the feasibility of implementing `flatMap` and `concat` (with the non-strictness and deferred entering of the scope of the other stream) for StreamEffect?

My main concerns are guaranteed running of finalizers, resource management, interruption safety.

/cc @iravid

jdegoes:
> @jdegoes Can you comment on the feasibility of implementing flatMap and concat (with the non-strictness and deferred entering of the scope of the other stream) for StreamEffect?

For `flatMap`:

 - The function must always return a `Stream`, not a `StreamEffect`, because you cannot guarantee what happens at runtime. This implies it implements `process`. However, it can implement an optimized `process` that feeds an `A` to the `f`, and if it gets back a `StreamEffect`, it does something differently. However, I'm not sure I see any easy win here since if you do anything too clever, it won't be lazy anymore.

For `concat`:

If the RHS is also a `StreamEffect`, the two can be fused into a single `StreamEffect`. Otherwise, they will promote to `Stream`.

> My main concerns are guaranteed running of finalizers, resource management, interruption safety.

You shouldn't have to worry too much about resource management; this will function at a higher-level, on the `ZManaged`, via `process` (not `processEffect`). You will have to pay attention when you construct streams from effects (`Stream.effect` and friends). Packaging them correctly into the `ZManaged`, basically.

Interruption-safety is also probably OK to ignore at this level. If the stream is lazy, the most you are doing with these fusion operations is composing together effectful streams; they will be pulled via `process`, which is effectful, and interruption will happen there.

> And would a solution that matches on the type of the stream, running the quick case when possible be an adequate solution?

Yes, these "shortcuts" will be common; e.g. when calling `StreamEffect#run` on a hypothetical `SinkEffect`. They are key to squeezing out max perf.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: I get that completely. My question was wether this is the same.

`trait Stream { val process: ZManaged[...] }` and `class Stream(val process: ZManaged[...])`.

I always thought the fields in class constructors are only sugar, there is no actual difference.

vasilmkd: I will address this in a follow up PR.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Done.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Done.

iravid:
Writing what I wrote in DM here: we can add a default implementation of `def process` in `StreamEffect`, as follows (correct me if I've missed anything):
```scala
  def process: Managed[E, IO[Option[E], A]] =
    self.processEffect.map { thunk =>
      UIO {
        try UIO.succeed(thunk())
        catch {
          case StreamEffect.Failure(e) => ZIO.fail(Some(e.asInstanceOf[E]))
          case StreamEffect.End        => ZIO.fail(None)
        }
      }.flatten
    }
```

Now, with this implementation, nothing in the `StreamEffect` file needs to override or implement the `process` method.

In fact, without this implementation, whenever we have something like ...
```scala
Stream.fromIterable(list)
  .map(f)
  .filter(f)
  .run(effectfulSink)
```

We would never use the efficient implementations in this file. So this not only saves code, but makes things faster :-)

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -0,0 +1,289 @@
...
+            }
+          }
+        }
+    }
+}
+
+private[stream] object StreamEffect extends Serializable {
+  import ZStream.InputStream
+
+  case class Failure[E](e: E) extends Throwable(e.toString, null, true, false)
```
iravid: Do we need the `override def fillStacktrace() = this` trick on these two?

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
iravid: We could call this `foldLeftManaged`, and also add it to `ZStream` itself. I'll add a ticket.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: The original code had that, but then I switched to the code snippet that @jdegoes provided.

Relevant stackoverflow [link](https://stackoverflow.com/questions/11434431/exception-without-stack-trace-in-java).

It looks like we are good.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Ah, I see. No, there is no difference, so long as you are using `val` there (the initialization will happen eagerly and will not contribute to stack depth).

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: This is done already and compiling without issues on all Scala versions.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: As per @iravid's request I will address this in another PR and change to class Stream.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Not sure `foldLeftManaged` is appropriate. `foldLeft` doesn't have a continuation function.

I'm thinking more of dividing the folds in effectful and non-effectful, depending on the state transformation function, but I hate the name `foldMManaged` and others...

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Done.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Dug up the old StreamPure, this used to be called `foldLazyPure`. @iravid

PR #1434, opened 2019-08-18
Simplify ZStream#combine To Not Use Queues

### LINKED ISSUE #1423: Simplify ZStream#combine to not use queues
With the new encoding, the queues are not necessary. The pulling function can work directly on the `InputStream` values from both sides.

adamgfraser:
Resolves #1423.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/Take.scala
+++ b/streams/shared/src/main/scala/zio/stream/Take.scala
@@ -16,8 +16,8 @@
 
 package zio.stream
 
-import zio.IO
-import zio.Cause
+import zio.{ Cause, IO, ZIO }
+import zio.stream.ZStream.InputStream
 
 /**
  * A `Take[E, A]` represents a single `take` from a queue modeling a stream of
@@ -59,6 +59,9 @@ object Take {
   final case class Fail[E](value: Cause[E]) extends Take[E, Nothing]
   final case class Value[A](value: A)       extends Take[Nothing, A]
   case object End                           extends Take[Nothing, Nothing]
+
+  final def fromInputStream[R, E, A](is: InputStream[R, E, A]): ZIO[R, Nothing, Take[E, A]] =
+    is.fold(_.fold[Take[E, A]](Take.End)(e => Take.Fail(Cause.fail(e))), Take.Value(_))
 
   final def option[E, A](io: IO[E, Take[E, A]]): IO[E, Option[A]] =
     io.flatMap {
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -930,13 +930,17 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
         case Take.End => ZIO.succeed(((leftDone, rightDone, s), Take.End))
       }
 
-    for {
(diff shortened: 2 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -931,7 +931,7 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
       }
 
     for {
```
iravid: Slightly more efficient:

```scala
ZStream.fromInputStreamManaged {
  for {
    sides <- b.process
    result <- self.combine(...).process
  } yield result
}
```

streams/shared/src/main/scala/zio/stream/Take.scala
```diff
@@ -60,6 +60,9 @@ object Take {
   final case class Value[A](value: A)       extends Take[Nothing, A]
   case object End                           extends Take[Nothing, Nothing]
 
+  final def fromInputStream[R, E, A](is: InputStream[R, E, A]): ZIO[R, Nothing, Take[E, A]] =
```
iravid: 👍🏻

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1066,9 +1066,12 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
    * type with the specified mapping functions.
    */
   final def mergeWith[R1 <: R, E1 >: E, B, C](that: ZStream[R1, E1, B])(l: A => C, r: B => C): ZStream[R1, E1, C] = {
-    type Loser = Either[Fiber[Nothing, Take[E1, A]], Fiber[Nothing, Take[E1, B]]]
+    type Loser = Either[Fiber[Nothing, Take[E, A]], Fiber[Nothing, Take[E1, B]]]
 
-    def race(left: UIO[Take[E1, A]], right: UIO[Take[E1, B]]): UIO[(Take[E1, C], Loser)] =
+    def race(
```
iravid: I think we could probably get rid of `Take` entirely here, but we can do that in a followup.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1406,25 +1409,23 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
       leftDone: Boolean,
       rightDone: Boolean,
-      q1: Queue[Take[E1, A]],
-      q2: Queue[Take[E1, B]]
+      left: InputStream[R, E, A],
+      right: InputStream[R1, E1, B]
     ): ZIO[R1, E1, ((Boolean, Boolean), Take[E1, C])] = {
-      val takeLeft: ZIO[R1, E1, Option[A]]  = if (leftDone) IO.succeed(None) else Take.option(q1.take)
-      val takeRight: ZIO[R1, E1, Option[B]] = if (rightDone) IO.succeed(None) else Take.option(q2.take)
+      val takeLeft: ZIO[R, E, Option[A]]    = if (leftDone) IO.succeed(None) else left.optional
```
iravid: `optional` is a nice touch here!

adamgfraser:
@iravid Thanks for the comments! This is all set now.

PR #1436, opened 2019-08-18
Remove manual tests / print output in tests

mschuwalow:
@jdegoes can we throw these out as they clutter test output?


jdegoes:
@mschuwalow Works for me! 👍

PR #1438, opened 2019-08-18
ZIO Test: Add Combinators for Composing Assertions

### LINKED ISSUE #1433: ZIO Test: Add composition operators to Assertion
Currently in ZIO Test (`test` project / directory), it is quite inconvenient to make multiple assertions on different values. A pattern seen in `ZQueueSpec.scala` in `core-tests` is as follows:

```scala
assert((v1, v2, o1, o2), Predicate.equals((10, 20, true, true)))
```

This does not scale in general (although it works for `Predicate.equals`, specifically), and leads to overly verbose error messages in the event of failure.

While we can (and probably should) define `Predicate.tuple2` (etc.) predicates, a better solution for assertion composition is simply to define operators on `Assertion`.

In particular, we can define the following operators on the `Assertion` trait:

 - `&&` — This will compose two `Assertion` into one, requiring that both succeed for the result to succeed.
 - `||` — This will compose two `Assertion` into one, requiring that one or the other succeed for the result to succeed.

For now, we can simply choose one failure ("first failure") if both fail. Later, we may want to expand `Assertion` into a free semiring to collect all failures. 

The type signatures requiring implementation:

```scala
trait Assertion[+E] {
  def && [E1 >: E](that: Assertion[E1]): Assertion[E1] = ???
  def || [E1 >: E](that: Assertion[E1]): Assertion[E1] = ???
}
```

In addition, we should define two tests in the `test` package for ZIO Test; and finally, we should refactor `ZQueueSpec.scala` to use the new assertion operators, as proof the "mission" of this ticket has been successfully accomplished.

### LINKED ISSUE #1437: ZIO Test: Add traversal methods for Assertion
We need sequence / traverse on `Assertion` (in the `test` project / top-level directory):

```scala
Assertion.foreach[A, E](as: Iterable[A])(f: A => Assertion[E]): Assertion[List[E]]
Assertion.collectAll[E](as: Iterable[Assertion[E]]): Assertion[List[E]]
```

with corresponding unit tests.

adamgfraser:
Adds combinators for composing assertions including `&&`, `||`, `foreach`, and `collectAll`. Refactors tests in `ZQueueSpec.scala` to use new combinators. Creates a test suite for `Assertion`. Resolves #1433. Resolves #1437.

@jdegoes If there are any other combinators for `Assertion` that you think would be useful let me know and I can add them. Also, do you think we should just refactor it to be a free semiring now?


test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -43,6 +64,25 @@ sealed trait Assertion[+A] { self =>
       case (self, Success)            => self
     }
 
+  /**
+   * A named alies for `||`.
```
ghostdogpr: (minor) typo: `alias`

adamgfraser: Good catch! Thanks!

jdegoes:
The free semiring can be introduced:

```scala
final case class Or[E](left: Assertion[E], right: Assertion[E]) extends Assertion[E]
final case class And[E](left: Assertion[E], right: Assertion[E]) extends Assertion[E]
```

Note that this will require you to rewrite `hashCode` and `equals` to satisfy the appropriate laws for equality for the semiring (`Or` and `And` are commutative, associative, etc., and there are distribution laws, `a * (b + c) == a * b + a * c`).

`Skipped` is in fact the additive _and_ multiplicative "empty" for the semiring. So, e.g. `Or(Skipped, a) == a`, `Or(a, Skipped) == a`, etc. So equality must take this into consideration (see `Cause` `equals` / `hashCode` for an example of how to do this).

As for other combinators, the ones I can think of off-hand:

```scala
Assertion.all[E](as: List[Assertion[E]]): Assertion[E] // Requires all to succeed
Assertion.any[E](as: List[Assertion[E]]): Assertion[E] // Requires any to succeed
```

In addition, bottom-up folding over the `Assertion` could be useful:

```scala
trait Assertion[+E] {
  def fold[Z](
    caseSkipped: => Z, 
    caseSucceeded: => Z, 
    caseFailed: E => Z, 
    caseAnd: (Z, Z) => Z, 
    caseOr: (Z, Z) => Z): Z = ???
}
```

Although I suspect you will discover that (or possibly the need for other combinators) as you try to adapt the old code to the semiring structure.

Note that a free semiring has "too much structure" to support `Assertion#combineWith`. I'm not sure if that's used, but if it's used, it will require generalizing or rewriting the code that uses it.

ghostdogpr:
@adamgfraser I added a fix for flakiness in ZQueueTest which is causing conflicts 😬

adamgfraser:

The only thing that didn't entirely feel right to me was `foreach` and `collectAll`. To your point about having "too much structure", it felt like the signature made sense before when `Assertion` could only hold a single result, so we needed a `List` to store multiple failures in. But now `Assertion` has its own structure for composing assertions and it seemed very unclear how a list of assertion results was supposed to fit in with the tree like structure of an assertion. For now I changed the signature to compose all the assertions into one assertion, but let me know if you think something else makes sense here.

adamgfraser:
@jdegoes Thinking about this a little more I'm wondering about my treatment of `Ignore`.

Right now I'm keeping intermediate `Ignore` values as I build up the data structure but ignoring them for purposes of `equals` and `hashCode` to honor the "empty" contract. This preserves information (e.g. if I have an `And(a, Ignore)` I know I was supposed to run two sub-test and only ran one versus if I just have `a` I only have a single test result). But to a certain extent we're lying about equality in saying those two are equal right now because I could for example do a `fold` over each of them to count the number of ignored results.

The alternative would be to simplify as we go along so for example `Ignore && a` would just return `a` instead of `And(Ignore, a)`. We would lose information on intermediate ignores but it would definitely simplify the code. I just worry that if we don't do that we create some risk of unexpected behavior. For example, the `any` and `all` combinators use `Ignore` as their `zero` element, so if you use those you are going to get one extra `Ignore`. We could probably work around those specific situations but I worry we are creating issues for ourselves by not being fully consistent about whether this really is an empty element.

What do you think?

ghostdogpr:
@adamgfraser fyi I merged the rename of `equals` which created merge conflicts

adamgfraser:
@ghostdogpr Thanks for the heads up! Good to get that merged. I will resolve the conflicts tonight.

jdegoes:
@adamgfraser 

> The only thing that didn't entirely feel right to me was foreach and collectAll. 

You are right. Unfortunately, with the new tree `Assertion`, the signatures must change, e.g.:

```scala
// Uses `||` to glue the `f` results together
def foreach[E, A, B](as: Iterable[A])(f: A => Assertion[B]): Assertion[B]

// Uses `&&` to glue the `f` results together
def foreachPar[E, A, B](as: Iterable[A])(f: A => Assertion[B]): Assertion[B]
```

Then `collectAll` / `collectAllPar` can be implemented in terms of these.

Actually, `Par` is consistent but not a great suffix. Maybe `And` or `Conj` would be better; `foreachOr` / `foreachAnd`? I am not sure.

> But to a certain extent we're lying about equality in saying those two are equal right now because I could for example do a fold over each of them to count the number of ignored results.

One problem is that we have two distinct notions of equality:

 - Structural equality
 - Semantic equality

We want semantic equality to obey the laws. In a similar fashion, you might expect `And(And(x, y), z))` to be semantically equal to `And(x, And(x, y))`, if not structurally equal.

In my experience, it's more useful to have `==` / `hashCode` do semantic rather than structural equality (indeed, `Set` and many other structures do this; they don't compare representations).

> The alternative would be to simplify as we go along so for example Ignore && a would just return a instead of And(Ignore, a)

I have done this before and besides being more difficult to test (when you have `Gen` for the data type), it's fragile and surprises me in a different way (constructors != helper functions).

> I know I was supposed to run two sub-test and only ran one versus if I just have a I only have a single test result).

You are right this is an issue. To eliminate all surprises, I suppose we should eliminate `Ignore` as empty; this forces things like `any` and `all` to return `Option[...]`.

Alternately we could add an `Empty` separate from `Ignore`.

What do you think? What is the code telling you to do? 😄

adamgfraser:
@jdegoes That makes sense regarding `foreach` and `collectAll`. One idea I had was that you could do something like this:

```scala
// Uses `||` to glue the `f` results together
def exists[E, A, B](as: Iterable[A])(f: A => Assertion[B]): Assertion[B] =
  ???

// Uses `&&` to glue the `f` results together
def forall[E, A, B](as: Iterable[A])(f: A => Assertion[B]): Assertion[B] =
  ???

def any[E, A](as: Iterable[Assertion[A]]): Assertion[A] =
  exists(as)(identity)

def all[E, A](as: Iterable[Assertion[A]]): Assertion[B] =
  forall(as)(identity)

def foreach[E, A, B](as: Iterable[A])(f: A => Assertion[B]): Assertion[B] =
  forall(as)(f)

def collectAll[E, A](as: Iterable[Assertion[A]]): Assertion[A] =
  all(as)
```

I know we normally like to have related methods have the same prefix for discoverability, but `forall` and `exists` are common names from other domains like collections and express pretty well what those operators do (create a new assertion that all sub-assertions pass or that any sub-assertion passes). Then you can implement `any` and `all` in terms of these. `foreach` and `collectAll` are just aliases for the more common case in the testing domain of wanting to assert that multiple things all have to be true, or they could potentially be omitted. Agree that `Par` is not great as it is used to indicate a parallel execution strategy throughout the rest of the code base.

I think your points regarding equality make sense and the analogy to `Set` is a good one. Even if we did eliminate intermediate `Ignore` values you could still break the equality between `And(x, y)` and `And(y, x)` by doing a fold with any operator that is not commutative, and we have no way to enforce that with the type system. So we're already away from the perfect theoretical "right" answer and we should just do what is going to be best / least surprising for users. I am a little hesitant to add another subtype just for fear of making signatures for methods like `fold` even more intimidating, but let me think about this.

jdegoes:
> foreach and collectAll are just aliases for the more common case in the testing domain of wanting to assert that multiple things all have to be true, or they could potentially be omitted. 

This sounds like a good plan for me. 👍 

In fact maybe we can omit `any` / `all`.

>  I am a little hesitant to add another subtype just for fear of making signatures for methods like fold even more intimidating, but let me think about this.

Right.

We don't _have_ to have an empty; `Cause` does not. What made me think this has an empty is that we would obviously like `x && ignore` to succeed / fail with `x`, as well as `x || ignore` (plus we have commutativity, associativity). 

If we decided to _not_ have an empty, we could at least introduce `def normalize: Option[Assertion[E]]`, which would "normalize" it, removing all `Ignore`; structural equality on the normalized form would give you semantic equality under the originally proposed semantic.

adamgfraser:
@jdegoes Okay, so reflecting on this more, here is what I am thinking:

First, `Assertion` should not have an `Empty` subtype. Conceptually, an `Assertion` is the result of running a test. Running a test should have some result, even if just to say the test was ignored. It seems like the main purpose of having an `Empty` subtype is to avoid methods like `forall` and `exists` returning an `Option`, but we should address that either by accepting it, taking a nonempty list an an input, or just stating that `Ignore` will be returned by default (users can always test the collection for emptiness themselves before calling these methods if they care).

Second, I'm thinking that `Ignore` should actually not be an identity element and wondering if it should also be parameterized with a message value. Stepping back, the main purpose of an assertion I would argue is to facilitate more detailed reporting of the results of a test. Yes it ultimately is either a failure, success, or ignore, but the reason we have it instead of a simpler data structure is so we can see a "tree" of the failures that led to a test failing instead of a "list" trace. Right now `Ignore` seems like an empty element because there is nothing really to do with it since it contains no value. But if we really want to be able to report on the results of a test it seems like we would might like to know what assertion was ignored (e.g. a string representing the name of the ignored assertion or an integer representing its importance).

In terms of evaluating `isSuccess` or `isFailure` we would still ignore `Ignore` values and so in that sense it would still be an identity element. But for reporting this would give the consumer of the `Assertion` the ability to decide what to do with it (e.g. report just the failures or report some information about the ignored assertions as well).

What do you think?

regiskuckaertz:
sorry for this unwelcome incursion:
> the main purpose of having an Empty subtype is to avoid methods like `forall` and `exists` returning an `Option`
I think those names were perfectly chosen:
- `forall`: on an empty list, there doesn't exist an element to disprove the assumption, therefore the empty case is success (see [the `All` monoid](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Monoid.html#t:All))
- `exists`: on an empty list, there doesn't exist an element to prove the assumption, therefore the empty case is failure (see [the `Any` monoid](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Monoid.html#t:Any))

adamgfraser:
@regiskuckaertz Ideas are always welcome!

That is definitely interesting. The main issue I see is that constructing a failure requires a failure message of some type `A`. So while we could potentially do that with success we couldn't do it with failure. We would need the caller to provide some default failure message, at which point I think you might as well give them an `Option` and then they can call `getOrElse` if they want to or use other combinators. And I would think we would want to be symmetric between `exists` and `forall`.

jdegoes:
> Right now Ignore seems like an empty element because there is nothing really to do with it since it contains no value.

The main reason `Ignore` exists to support things like `pending` / `ignore` in other test frameworks. It's frequently useful to disable a test but not actually remove it from the test suite, so you have a reminder the test is there but is not actually running. 

Now, there's no reason we have to push this functionality to `Assertion`. 

For example, every test could be `Option[ZIO[R, E, TestResult]]` (as an example). The presence of a `None` would then indicates the test is not present, for whatever reason. (The label too could encode such information, but it's simpler to let users decide the structure of labels rather than require a given structure.)

So perhaps `Ignore` itself can be deleted. An advantage is that the structure becomes simpler and ignored tests can be known statically before running them. A disadvantage is that you cannot dynamically enable / disable certain tests based on, e.g. environment variables. Basically you are "committed" to either fail or succeed a test once you begin an effect. That is definitely less flexible.

As for `Ignore` taking parameters, I am not sure about that: to get, for example, details about a predicate, you'd have to run the test, and ignore its result; usually such a combinator is called `pending` and it runs the test and ignores it if it fails, but reports it if it succeeds, so you know you can remove the `pending` annotation.

Stepping back, I think external factors can help force the design of this: namely, a change I want to make to how failure is specified in tests.

Currently, both failure and success are specified in a `TestResult` (`Assertion[FailureDetails]`). I think this is fundamentally the wrong design, because effects already have a failure channel, so maintaining two different failure modes like this can lead to subtle gotchas (such as the recently discovered `eventually` and `retry` combinators not working on test failures, only on effect failures).

This means we should go from a test being `ZIO[R, E, TestResult]` to something like `ZIO[R, FailureDetails[E], Unit]`, which collapses the two error channels into one, allowing expected semantics for things like `eventually`. 

This means the type of `Assertion` will not appear inside a test at all. As a result, it is not necessary to model `Ignore` or anything else in `Assertion`, since it already has a representation at a higher-level (`Unit` in the above type). 

This suggests to me that we should simplify `Assertion` to be either `Success`, or `Failure` with information of the specified type; plus the semiring operations (`||` and `&&`), following the distributive and associativity laws; all without `Ignore` or `Empty`.

Then we refactor test to be `ZIO[R, FailureDetails[E], S]`, or something similar, where `S` (_status_ or _success_) can be `Unit`, but most code won't care about the return type of successful tests.

As a final refactor, we introduce `S` into `Assertion`, turning it into `Assertion[E, S]`, which ignoring the free semiring, is isomorphic to `Either`, representing a failure `E` or a success `S`, where in a common case `S` will be `Unit`. To convert such a structure to an `Either`, one needs an associative operation for transforming `(E, E) => E`, and `(S, S) => S` (i.e. a semigroup).

So either we fail with some details, or we succeed with some other details (like status); providing a source of the `S` values that can flow through in other combinators.

Now if we perform all these changes, then we have a fairly simple structure for `Assertion`: `Success[S]` or `Failure[E]`, plus the semiring operations. The question then becomes: should we include an `Empty`? I think not, because we can have our own empties inside `Success`, using the `S` type parameter; indeed, one could imagine:

```scala
sealed trait TestStatus
object TestStatus {
  case object Executed extends TestStatus
  case object Ignored extends TestStatus
  case class Pending(passed: Boolean) extends TestStatus
  case class Skipped extends TestStatus
  ...
}
```

The nature of these is more or less arbitrary, they are just statuses attached to a test which did not fail, so one would not expect them to be connected to any laws like the empty for semiring.

In summary:

**I think you are right, we should not add `Empty`, and we should probably delete `Ignore`, pushing such functionality higher in a polymorphic status `S`, which can feature in `Assertion[E, S]`, as well as `ZTest`.**

adamgfraser:
@jdegoes Yes, I like that. It is both simpler and more polymorphic. I will proceed on that basis.

jdegoes:
All right, to summarize these changes:

 - [ ] `Assertion` goes from `Assertion[E]` to `Assertion[E, S]`, briefly `data Assertion e s = Failure e | Success s | Or (Assertion e s) (Assertion e s) | And (Assertion e s) (Assertion e s)`, losing `Ignore` in the process and acquiring polymorphic information on success.
 - [ ] `FailureDetails` goes from `FailureDetails` (with `Cause[Any]`) to `FailureDetails[E]` (with `Cause[E]`), which allows us to submerge `E` into `FailureDetails[E]`.
 - [ ] `ZTest` goes from `ZIO[R, E, TestResult]` to something else. Perhaps `ZIO[R, Assertion[E, Nothing], Assertion[Nothing, S]]`, where `S` is a new type parameter that indicates the status of a **non-failed** test (_non-failed_ does not necessarily mean succeeded, but it means whatever happened, it should not fail the test / build). The type parameters of `Assertion[E, Nothing]` indicate it has at least one failure, and no successes; while the type parameters of `Assertion[Nothing, S]` indicate it has at least one success, and no failures. The types also allow us to preserve the full structure of the assertion composition all the way through producing results.
 - [ ] `ExecutedSpec[L]` will go from `Spec[L, TestResult]` to something like `Spec[L, Either[Assertion[E, Nothing], Assertion[Nothing, E]]]`.

After writing all this down, I think I see a further simplification.

Instead of having something like `Assertion` with two type parameters, we can make it have just 1. Then we can represent what is now called assertion as `Assertion[Either[E, S]]`. 

This would make `Assertion` something like `data Assertion a = Leaf a |  Or (Assertion a) (Assertion a) | And (Assertion a) (Assertion a)`.

More later, have to run!

jdegoes:
@adamgfraser 

False alarm, will run later. 😆 

Re-summary of the changes.

- [ ] `Assertion` goes from `Assertion[E]` to `Assertion[A]`, approximately `data Assertion a = Value a | Or (Assertion a) (Assertion a) | And (Assertion a) (Assertion a)`, losing `Ignore` in the process and becoming independent of the concept of success and failure, which can be layered on independently using the single type parameter.
 - [ ] `ZTest` goes from `ZIO[R, E, TestResult]` to something else. Perhaps `ZIO[R, TestFailure[E], Assertion[S]]`, where `S` is a new type parameter that indicates the status of a non-failed test (non-failed does not necessarily mean succeeded, but it means whatever happened, it should not fail the test / build), and `TestFailure[E]` is reimagined as:
    ```scala
    sealed trait TestFailure[+E]
    object TestFailure {
      final case class Predicate(assertion: Assertion[PredicateValue]) extends TestFailure[E]
      final case class Runtime(cause: Cause[E]) extends TestFailure[E]
    }
    ```
 - [ ] `ExecutedSpec[L]` will go from `Spec[L, TestResult]` to something like `Spec[L, Either[TestFailure[E], Assertion[S]]]`.

There's still a problem with this representation: namely, in the failure case, we forget the successes, and only know about the failures. It's possible to solve it but the naive way (making `TestFailure.Predicate` contain `Assertion[Either[PredicateValue, S]]`) will make it impossible to prove statically there's at least one failure in the `Assertion`.

jdegoes:
With newer names:

```scala
sealed trait AssertResult[+A]
object AssertResult {
  final case class Or[+A](left: AssertResult[A], right: AssertResult[A]) extends AssertResult[A]
  final case class And[+A](left: AssertResult[A], right: AssertResult[A]) extends AssertResult[A]
  final case class Value[+A](value: A) extends AssertResult[A]
}

sealed trait TestFailure[+E]
object TestFailure {
  final case class Predicate(result: AssertResult[PredicateValue]) extends TestFailure[Nothing]
  final case class Runtime[+E](cause: Cause[E]) extends TestFailure[E]
}

type ZTest[-R, +E, +S] = ZIO[R, TestFailure[E], S]

type ZSpec[-R, +E, +L, +S] = Spec[L, ZTest[R, E, S]]

type ExecutedSpec[+L, +E, +S] = Spec[L, Either[TestFailure[E], AssertResult[S]]]
```

Plus or minus.

It'd be nice if we could take a tree like `AssertResult`, peel it apart into two halves, perfectly interlocking, and then recombine them. Not possible in a type safe way though, because the two haves would have to align perfectly for recombination to make sense.

adamgfraser:
I'm on it.

adamgfraser:
@jdegoes One question on this. Above you wrote:

```scala
sealed trait TestFailure[+E]
object TestFailure {
  final case class Predicate(result: AssertResult[PredicateValue]) extends TestFailure[Nothing]
  final case class Runtime[+E](cause: Cause[E]) extends TestFailure[E]
}
```

But in #1475 you wrote `zio.test.FailureDetails` => `zio.test.TestFailure`, where `FailureDetails` is:

```scala
sealed trait FailureDetails
object FailureDetails {
  final case class Predicate(fragment: PredicateValue, whole: PredicateValue) extends FailureDetails
  final case class Runtime(cause: Cause[Any])                                 extends FailureDetails
}
```

So it seems like we are at risk of losing the whole / fragment concept which we use to drive "drilling down" into the specific sub-predicate that failed. Should it instead be?

```scala
final case class FailureDetails(fragment: PredicateValue, whole: PredicateValue)
// to be refactored to be a nonempty list

sealed trait TestFailure[+E]
object TestFailure {
  final case class Predicate(result: AssertResult[FailureDetails]) extends TestFailure[Nothing]
  final case class Runtime[+E](cause: Cause[E]) extends TestFailure[E]
}
```

adamgfraser:
Actually maybe it is better to have the runtime error as a subtype of failure details.

jdegoes:
@adamgfraser 

For every test failure, there can be at most one `TestFailure.Runtime` failure. The reason is that such failures come from the effects that make up the test; they indicate some sort of catastrophe that the user didn't handle. So for this reason, for maximum precision, we would say:

Either:

 - A test failed through normal means, i.e. by failed assertions, in which case, here's a tree of failures (and maybe some successes)
 - OR
 - A test failed through abnormal means, i.e. with some Cause, in which case, here is the cause

There can only ever be one `Cause` because it will halt the test immediately. But there can be multiple asserts all composed together, so there are a tree of results there.

Does that make sense?

adamgfraser:
@jdegoes Yes, thanks!

adamgfraser:
@jdegoes I think I have all the changes we talked about in here. I still need to do some clean up and I think I don't have some of the type signatures quite right, but everything is working except for ignoring tests. If you have time to take a look would definitely help in getting this over the finish line.

jdegoes:
@adamgfraser I'll take a look this morning. Thanks for your work on this!

test/shared/src/test/scala/zio/test/AssertResultSpec.scala
```diff
@@ -0,0 +1,183 @@
...
+      val assertion1 = randomAssertion()
+      val assertion2 = randomAssertion()
+      (assertion1 && assertion2) == (assertion2 && assertion1)
+    }
+
+  def anySucceedsIfAnyAssertionsSucceed: Future[Boolean] =
+    Future.successful(AssertResult.any(List(success1, failure1, failure2)).isSuccess)
+
+  // def asMapsFailureMessagestoConstantValue: Future[Boolean] =
+  //   Future.successful {
```
jdegoes: Dead code?

test/shared/src/test/scala/zio/test/AssertResultSpec.scala
```diff
@@ -0,0 +1,183 @@
...
+    // label(allSucceedsIfAllAssertionsSucceed, "all succeds if all assertions succeed"),
+    label(andDistributesOverOr, "and distributes over or"),
+    label(andIsAssociative, "and is associative"),
+    label(andIsCommutative, "and is commutative"),
+    // label(anySucceedsIfAnyAssertionsSucceed, "any succeeds if any assertion succeeds"),
+    // label(asMapsFailureMessagestoConstantValue, "as maps failure messages to constant value"),
+    label(bothReturnsConjunctionOfTwoAssertions, "both returns conjunction of two assertions"),
+    // label(collectAllCombinesMultipleAssertions, "collectAll combines multiple assertions"),
+    label(eitherReturnsDisjunctionOfTwoAssertions, "either returns disjunction of two assertions"),
+    // label(failuresReturnsListOfFailures, "failures returns list of failures"),
```
jdegoes: Dead code here and below?

test/shared/src/test/scala/zio/test/AssertResultSpec.scala
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+
+  def isFailureReturnsWhetherAssertionFailed: Future[Boolean] =
+    Future.successful(!success1.isFailure && failure1.isFailure)
+
+  def isSuccessReturnsWhetherAssertionSucceeded: Future[Boolean] =
+    Future.successful(success1.isSuccess && !failure1.isSuccess)
+
+  // def mapTransformsFailureMessages: Future[Boolean] =
```
jdegoes: Dead code?

jdegoes:
> Right now the ignore TestAspect basically just makes a test always succeed and the test reporter has no way to handle ignored tests or report how many tests were ignored because it just sees a generic S. Do we need some small data type that wraps the S and has an Ignore case?

I think we possibly need to introduce a true `TestStatus` and use that for `S` whenever we _must_ :

```scala
sealed trait TestStatus
object TestStatus {
  final case object Executed extends TestStatus
  final case object Ignored extends TestStatus
  ...
}
```

Theoretically we could add `Pending`, etc., too. 

Why not a wrapper? Because users are unlikely to use a custom `S` in addition to our pre-canned one above, so just choosing a data type is probably simplest. In addition, a test reporter won't know what to do with a generic `S`, anyway.

Does that seem good?

regiskuckaertz:
NB: it looks like you could add an `implies` predicate in addition to `&&` and `||` in order to promote it to [this structure](https://pursuit.purescript.org/packages/purescript-prelude/4.1.1/docs/Data.HeytingAlgebra#t:HeytingAlgebra) 😎

test/jvm/src/main/scala/zio/test/RunnableSpec.scala
```diff
@@ -21,9 +21,12 @@ import zio.test.Spec.TestCase
 /**
  * A `RunnableSpec` has a main function and can be run by the JVM / Scala.js.
  */
-abstract class RunnableSpec[L, T](runner0: TestRunner[L, T])(spec0: => Spec[L, T]) extends AbstractRunnableSpec {
-  override type Label = L
-  override type Test  = T
+abstract class RunnableSpec[L, T, E, S](runner0: TestRunner[L, T, E, S])(spec0: => Spec[L, T])
```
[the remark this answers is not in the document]
jdegoes: I mean using `+` and `-` on output / input channels, respectively. But we can do that later.

test/shared/src/main/scala/zio/test/AssertResult.scala
```diff
@@ -17,66 +17,234 @@
...
+   * function.
+   */
+  final def map[B](f: A => B): AssertResult[B] =
+    fold(f andThen value)(and, or)
+
+  final def isFailure(implicit ev: A <:< Either[_, _]): Boolean =
+    !isSuccess
+
+  final def isSuccess(implicit ev: A <:< Either[_, _]): Boolean =
+    fold(a => ev(a).isRight)(_ && _, _ || _)
```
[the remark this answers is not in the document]
jdegoes: Yes, that makes sense! Basically, "enough succeeded" rather than "all" or "some".

test/shared/src/main/scala/zio/test/TestReporter.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import zio.{ URIO, ZIO }
+import zio.duration.Duration
+
+/**
+ * A `TestReporter[L]` is capable of reporting test results annotated with
+ * labels `L`.
+ */
+trait TestReporter[-L] {
+  def apply[E, S](duration: Duration, executedSpec: ExecutedSpec[L, E, S]): URIO[TestLogger, Unit]
```
jdegoes: If we make `TestReporter` polymorphic in `E` and `S`, we won't be able to have reporters that do specific things with `E` or `S`. Since that could be useful, we may want to stick the type parameters in `TestReporter[-L, +E, +S]`, so some types of test reporters can take advantage of their knowledge of concrete error and status types.

test/shared/src/main/scala/zio/test/TestRunner.scala
```diff
@@ -37,7 +37,7 @@ case class TestRunner[L, -T](
   /**
    * Runs the spec, producing the execution results.
    */
-  final def run(spec: Spec[L, T]): URIO[TestLogger with Clock, ExecutedSpec[L]] =
+  final def run(spec: Spec[L, T]): URIO[TestLogger with Clock, ExecutedSpec[L, E, S]] =
```
jdegoes: One thing disturbs me about this: Where are these `E` and `S` values coming from? Of course we know they are in `T` but the code does not know that. I think we can still merge and figure it out later.

test/shared/src/main/scala/zio/test/TestStatus.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test
+
+sealed trait TestStatus[+S]
+
+object TestStatus {
+  final case class Executed[S](result: AssertResult[S]) extends TestStatus[S]
```
jdegoes: Maybe we should call this `Succeeded` like SBT?

test/shared/src/main/scala/zio/test/TestStatus.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test
+
+sealed trait TestStatus[+S]
```
jdegoes: Maybe we should call this `TestSuccess` to emphasize the symmetry with `TestFailure`?

adamgfraser:
All set.

test/shared/src/main/scala/zio/test/AssertResult.scala
```diff
@@ -17,66 +17,234 @@
 package zio.test
 
 /**
- * An `AssertResult[A]` is the result of running a test, which may be ignore,
- * success, or failure, with some message of type `A`.
+ * An `Assertion[A]` is the result of running a test, which may be ignore,
```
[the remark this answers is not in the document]
jdegoes: I like both of these ideas. `BoolAlg` or something might be useful as a name; `Not` / `Negate` another case of the type. Then type alias for `AssertionResult[E, S] = LogicTree[Either[E, S]]` or similar. Maybe that's all we need.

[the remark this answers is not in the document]
jdegoes: Do you want to get this in now, or merge and later?

ghostdogpr:
@adamgfraser can you rebase? Amazing work as usual 👍

adamgfraser:
@ghostdogpr Thanks! This is ready to go now.

PR #1444, opened 2019-08-20
Add tuple syntax that includes environment types

jeremyrsmith:
I did this so that `.map[2|3|4]` extension methods can be available on tuples which include environment types. The `R` of the result is simply the intersection of `R`s of the tuple elements.

I made new value classes and implicit defs for this, so as to avoid any binary compatibility complications. This entailed putting the new implicit conversions in a trait to be inherited, so that they wouldn't conflict with the existing implicit defs (they should only be used when one or more tuple elements have R =!:= Any, which currently wouldn't work at all). I think this ought to be binary compatible, but to be honest I'm not 100% sure.

### DIFF
--- a/core/shared/src/main/scala/zio/syntax/IOSyntax.scala
+++ b/core/shared/src/main/scala/zio/syntax/IOSyntax.scala
@@ -16,7 +16,7 @@
 
 package zio.syntax
 
-import zio.{ Fiber, IO, Task, UIO }
+import zio.{ Fiber, IO, Task, UIO, ZIO }
 
 object IOSyntax {
   final class IOCreationLazySyntax[A](val a: () => A) extends AnyVal {
@@ -39,12 +39,12 @@ object IOSyntax {
     def collectAll: IO[E, List[A]]                     = IO.collectAll(ios)
   }
 
-  final class IOTuple2[E, A, B](val ios2: (IO[E, A], IO[E, B])) extends AnyVal {
-    def map2[C](f: (A, B) => C): IO[E, C] = ios2._1.flatMap(a => ios2._2.map(f(a, _)))
+  final class IOTuple2[R, E, A, B](val ios2: (ZIO[R, E, A], ZIO[R, E, B])) extends AnyVal {
+    def map2[C](f: (A, B) => C): ZIO[R, E, C] = ios2._1.flatMap(a => ios2._2.map(f(a, _)))
   }
 
-  final class IOTuple3[E, A, B, C](val ios3: (IO[E, A], IO[E, B], IO[E, C])) extends AnyVal {
-    def map3[D](f: (A, B, C) => D): IO[E, D] =
+  final class IOTuple3[R, E, A, B, C](val ios3: (ZIO[R, E, A], ZIO[R, E, B], ZIO[R, E, C])) extends AnyVal {
+    def map3[D](f: (A, B, C) => D): ZIO[R, E, D] =
       for {
         a <- ios3._1
         b <- ios3._2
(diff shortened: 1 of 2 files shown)


jeremyrsmith:
@iravid Sure, that makes it way easier. Updated to just change the existing classes and methods.

core/shared/src/main/scala/zio/syntax/IOSyntax.scala
```diff
@@ -39,21 +39,23 @@ object IOSyntax {
     def collectAll: IO[E, List[A]]                     = IO.collectAll(ios)
   }
 
-  final class IOTuple2[E, A, B](val ios2: (IO[E, A], IO[E, B])) extends AnyVal {
-    def map2[C](f: (A, B) => C): IO[E, C] = ios2._1.flatMap(a => ios2._2.map(f(a, _)))
+  final class IOTuple2[E, RA, A, RB, B](val ios2: (ZIO[RA, E, A], ZIO[RB, E, B])) extends AnyVal {
```
kaishh: Would this widen correctly to the LUB error type if left and right IO in a tuple have different error types?

jeremyrsmith: I'm not sure... if it already did that, it should still do that. This could probably further modified to correctly infer the error LUB, if it doesn't already, but this PR is just for allowing environment types.

jeremyrsmith:
I'm not sure why test_232_jdk11_jvm failed... it seems unrelated?

ghostdogpr:
@jeremyrsmith indeed, I re-launched it

core/shared/src/main/scala/zio/syntax/IOSyntax.scala
jdegoes: Yes, it should infer correctly.

jdegoes: There is no need to use 2 different `R` type parameters: you can use one, and let Scala unify them. In fact, letting Scala do this initially instead of trying to force it later using `with` will lead to better inference.

On both this and the other tuple syntaxes, I'd go down to 1 `R` and let Scala do the heavy lifting...

jeremyrsmith: I tried it out and it does indeed infer the error LUB:

```scala
scala> import zio.syntax._
import zio.syntax._

scala> val a = ZIO(10)
a: zio.Task[Int] = zio.ZIO$EffectPartial@29dd4b0b

scala> val b = ZIO.succeed("hello")
b: zio.UIO[String] = zio.ZIO$Succeed@45db18dd

scala> val c = ZIO(20).refineOrDie { case err: Throwable => new IllegalArgumentException("Oh no!", err) }
c: zio.ZIO[Any,IllegalArgumentException,Int] = <function1>

scala> val all = (a, b, c).map3 {
     |   (i1, str, i2) => i1 + str.length + i2
     | }
all: zio.ZIO[Any with Any with Any,Throwable,Int] = zio.ZIO$FlatMap@363a9244
```

If the `Any with Any with Any` is bothersome, it can be solved with an additional implicit.

jeremyrsmith: Wow, you're right @jdegoes – I'm not sure why I'm surprised that the inference works, but it works just fine that way and is much simpler.  Updated the PR (if you can still call it that 😄)

jdegoes:
Looks great! Thanks for your contribution, and congratulations on you first ZIO pull request being merged! 🎉 

giphy

PR #1435, opened 2019-08-18
Report stats after each test suite execution

ghostdogpr:
This adds the following line (in cyan) to the reported results after each suite is executed:
```
Ran 70 tests in 3 seconds: 70 succeeded, 0 ignored, 0 failed
```

This is something that specs2 is doing and can be pretty useful (also, it visually separates the suites more clearly).

PR #1446, opened 2019-08-20
ZIO Test: Improve RunnableSpec

adamgfraser:
Following up on discussions in #1435 this PR addresses two issues with `RunnableSpec`.

First, running tests directly doesn't currently working on Scala.js because the `main` method is implemented in terms of `unsafeRun`, which is not safe on Scala.js. This PR provides separate platform specific implementations of `RunnableSpec` that use `unsafeRunAsync` on Scala.js to address this.

Second, running tests directly currently results in a success exit code even if one or more tests failed. This PR corrects that behavior on the JVM so the `main` method exits with an error code if any test failed.

### DIFF
--- a/test/jvm/src/main/scala/zio/test/RunnableSpec.scala
+++ b/test/jvm/src/main/scala/zio/test/RunnableSpec.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2019 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


test/jvm/src/main/scala/zio/test/RunnableSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * A simple main function that can be used to run the spec.
+   *
+   * TODO: Parse command line options.
+   */
+  final def main(args: Array[String]): Unit = {
+    val results     = runner.unsafeRun(spec)
+    val hasFailures = results.exists { case TestCase(_, test) => test.failure; case _ => false }
+    try if (hasFailures) sys.exit(1) else sys.exit(0)
```
ghostdogpr: I remember there was `sys.exit` before and we removed it because it was exiting the current sbt session when you ran the tests from an sbt shell. Isn't it the case here?

adamgfraser: You need to trap the `SecurityException` like I do here to prevent that. I just tested it locally and confirmed that this does not exit from the current SBT session.

ghostdogpr: Nice!

ghostdogpr:
@adamgfraser thanks!

PR #1447, opened 2019-08-20
ZIO Test: Add Documentation for MockClock

### LINKED ISSUE #1443: ZIO Test: Improve Documentation of MockClock
`MockClock` provides significantly enhanced functionality for testing code that involves the passage of time, but it requires users to adjust the time using the `adjust` and `setTime` methods instead of just waiting for "wall time" to pass. We need better documentation so users understand how to take advantage of this in writing tests. This recently came up in #1329.

I will take the ticket.

adamgfraser:
Resolves #1443.

### DIFF
--- a/test/shared/src/main/scala/zio/test/mock/MockClock.scala
+++ b/test/shared/src/main/scala/zio/test/mock/MockClock.scala
@@ -26,6 +26,47 @@ import zio.internal.{ Scheduler => IScheduler }
 import zio.internal.Scheduler.CancelToken
 import zio.scheduler.Scheduler
 
+/**
+ * `MockClock` makes it easy to deterministially and efficiently test effects
+ * involving the passage of time.
+ *
+ * Instead of waiting for actual time to pass, `sleep` and methods implemented
+ * in terms of it schedule effects to take place at a given time. Users can
+ * adjust the current clock time using the `adjust` and `setTime` methods, and
+ * all effects scheduled to take place on or before that time will automically
+ * be run.
+ *
+ * For example, here is how we can test [[ZIO.timeout]] using `MockClock:
+ *
+ * {{{
+ *  import zio.ZIO
+ *  import zio.duration._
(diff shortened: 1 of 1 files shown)


test/shared/src/main/scala/zio/test/mock/MockClock.scala
```diff
@@ -26,6 +26,47 @@ import zio.internal.{ Scheduler => IScheduler }
 import zio.internal.Scheduler.CancelToken
 import zio.scheduler.Scheduler
 
+/**
+ * `MockClock` makes it easy to deterministially and efficiently test effects
```
ghostdogpr: `deterministially` => `deterministically`

adamgfraser: Thanks!

test/shared/src/main/scala/zio/test/mock/MockClock.scala
```diff
@@ -26,6 +26,47 @@ import zio.internal.{ Scheduler => IScheduler }
...
+ * line below. Thus, a useful pattern when using `MockClock` is to fork the
+ * effect being tested, then adjust the clock to the desired time, and finally
+ * verify that the expected effects have been performed.
+ *
+ * One other thing to keep in mind is that `sleep` and related combinators
+ * schedule events to occur at an absolute point in time (e.g. 10 seconds from
+ * the `0` time), not relative time (e.g. 10 seconds from the current set clock
+ * time). This is necessary to avoid race conditions between scheduling effects
+ * and adjusting the time but means it sometimes may be necessary to use
+ * `adjustTime` to reset the time between effects you are testing to achieve
```
jozic: would be great to have an example here showing this use case

adamgfraser: Good suggestion. I will work on something and submit a follow up PR.

Issue #1348, opened 2019-08-04
Remove explicit delay in ZSchedule


jdegoes:
ZSchedule has an explicit `Duration` at each step. The original motivation for this was quite simple: it makes testing schedules easier, because there is no need for the schedule itself to sleep; rather, the driver for the schedule can sleep. 

However, at this point, with ZIO Environment, testing schedules is already easy, even if the schedules are sleeping. In fact, the `.jittered` combinator, which uses randomness, is also easy / fast / deterministic to test, when used with `MockRandom`.

Given the original motivation for making schedules explicit with `Duration` no longer applies, and given the potential to simplify #676, it may be worth implementing a formulation of `Schedule` that is responsible for its own "sleeps".

This will have the following effects:

 - Simplify some mechanics inside `Schedule`
 - Remove the `Clock` from `ZIO#repeat` / `ZIO#retry` and friends (only schedules that sleep will "contaminate" the effect with `Clock`)
 - Permit "sleep to fixed point in time" without added machinery
 - Make sleeping schedules responsible for their own sleeping

The most difficult change will be redefining semantics for `&&` and `||` operators.

### Intersection

The intersection operator `&&` repeats as long as _both_ continue, and for the maximum of the delay of the two schedules.

This operator could be replaced by something like `left.zipWith(right)(_ && _)`. 

### Union

The union operator `||` repeats as long as _either_ continue, and for the minimum of the delay of the two schedules.

This operator could be replaced by using `ZIO#raceWith`: if the first one to finish _wants_ to continue, the schedule as a whole continues, with the state of the first one; otherwise, the second one is joined, to get its decision, and continue with its state. Previous states can be reused to feed to the next race (although the loser of a race will have its state ignored—this may cause some meaningful differences that are worth exploring).

### Thoughts

Overall I am positive of this change because it makes schedule simpler and ties the semantics of `||` and `&&` to noted effect composition operators. 

Also stands to solve #1339  as well as #676, but in a fewer lines of code than currently possible.

mschuwalow:
I would like to take this

mschuwalow:
Sorry I got hung up on a few issues for streams and am now going to be busy for a few days :cry: .
Feel free to pick this up in the meantime.

alexandredantas:
[...]

PR #1451, opened 2019-08-21
ZIO Test: Further Improvements to MockClock Handling of Sleep

adamgfraser:
As suggested by @jozic in #1447, adds an additional example involving resetting the `MockClock` to test an effect recurring on a fixed schedule. Also adds examples to test suite to verify that they work correctly and prevent regressions.

### DIFF
--- a/test/shared/src/test/scala/zio/test/TestUtils.scala
+++ b/test/shared/src/test/scala/zio/test/TestUtils.scala
@@ -2,6 +2,8 @@ package zio.test
 
 import scala.concurrent.{ ExecutionContext, Future }
 
+import zio.ZIO
+
 object TestUtils {
 
   def label(f: Future[Boolean], s: String)(implicit ec: ExecutionContext): Future[(Boolean, String)] =
@@ -11,6 +13,18 @@ object TestUtils {
       else
         (p, fail(s))
     }.recover { case _ => (false, fail(s)) }
+
+  def nonFlaky[R, E](test: ZIO[R, E, Boolean]): ZIO[R, E, Boolean] = {
+    def repeat(n: Int): ZIO[R, E, Boolean] =
+      if (n <= 1) test
+      else
+        test.flatMap { result =>
+          if (result) repeat(n - 1)
+          else ZIO.succeed(result)
+        }
+
+    repeat(100)
+  }
 
   def scope(fs: List[Future[(Boolean, String)]], s: String)(
     implicit ec: ExecutionContext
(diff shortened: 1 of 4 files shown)


ghostdogpr:
Seems like the new test fails in most builds? Actually there’s no detail on which test fails :/

adamgfraser:
Yes I am reworking this. Unfortunately the test suite for ZIO Test doesn’t have nice error messages. May have to clean that up at some point.

adamgfraser:
After spending more time on this, I came to the conclusion that the previous `MockClock` implementation, while an improvement, was not sufficient.

While it was effective for testing some time related applications such as `timeout`, the fact that `sleep` always scheduled effects for an absolute time from zero time instead of relative time from some notion of the "current time" made it inconsistent with the actual `Clock` and created issues with testing code involving repetitive effects. For example, under the old implementation if we had an effect that executed repeatedly every hour and advanced the `MockClock` by an hour, further repetitions of that effect would execute without any further delay, requiring further coordination mechanisms such as latches that did not provide a good user experience.

This PR further improves `MockClock` so that `sleep` schedules actions for a relative time in the future. We avoid race conditions between calling `sleep` and updating the time by using a notion of "fiber time", backed by a `FiberRef`, that is incremented for each fiber when it sleeps and used as a relative basis for scheduling `sleep` calls, and "clock time" controlled by the user. With this we can test sophisticated concepts involving time in a straightforward way. For example, here is how we could test an effect that performs an action at a fixed interval:

```scala
val io = for {
  q <- Queue.unbounded[Unit]
  _ <- (q.offer(()).delay(60.minutes)).forever.fork
  a <- q.poll.map(_.isEmpty)
  _ <- MockClock.adjust(60.minutes)
  b <- q.take.as(true)
  c <- q.poll.map(_.isEmpty)
  _ <- MockClock.adjust(60.minutes)
  d <- q.take.as(true)
  e <- q.poll.map(_.isEmpty)
} yield a && b && c && d && e
```

@jdegoes Could I get a review on this when you have the chance?

iravid:
Very clever use of `FiberRef` @adamgfraser!

test/shared/src/test/scala/zio/test/TestUtils.scala
```diff
@@ -12,6 +14,18 @@ object TestUtils {
         (p, fail(s))
     }.recover { case _ => (false, fail(s)) }
 
+  def nonFlaky[R, E](test: ZIO[R, E, Boolean]): ZIO[R, E, Boolean] = {
```
jdegoes: Implementation could be: `test.repeat(Schedule.recurs(100) *> Schedule.identity[Boolean])`

adamgfraser: Thanks! Much nicer! Now if we can only get rid of that pesky `Clock` type constraint.

jdegoes: Which we can shortly, I hope, with #1348!

PR #1452, opened 2019-08-21
update docs and mark unsubscribe test as flaky

mschuwalow:


iravid:
Failure was in `RTSSpec`, so merging this.

PR #1454, opened 2019-08-22
Unify Signatures of App#run

adamgfraser:
Currently the `App` traits on the JVM and Scala.js define run methods with different signatures

```scala
def run(args: List[String]): ZIO[Environment, Nothing, Int]   // JVM
def run(args: List[String]): ZIO[Environment, Nothing, Unit]  // JS
```

Not only are they different, but they are conflicting, so it is actually impossible today to implement a program that runs on both the JVM and Scala.js by extending the `App` trait. This represents a failure of abstraction and forces users to concern themselves with lower level details of how a program is run and exit codes are interpreted on different platforms.

This PR addresses this by unifying the version of the trait on both platforms to return an `Int` representing an exit code. We don't currently do anything with that exit code on Scala.js, which is perhaps the best argument in favor of the status quo (at least we're honest that we're not even going to let you give us an exit code). But I think it is far better for us to have one trait that users can use to define executable programs across platforms.


jdegoes:
@adamgfraser Agreed, and good catch!

PR #1468, opened 2019-08-23
Fix queue shutdown race condition

### LINKED ISSUE #1449: Flaky test in ZStreamSpec
```
[info]   Stream.broadcast
[info]     + Values
[info]     + Errors
[info]     + BackPressure
[error]     x Unsubscribe
```
I've seen that `Unsubcribe` test fail in several builds.
Examples:
https://circleci.com/gh/zio/zio/18107
https://circleci.com/gh/zio/zio/18094

cc @mschuwalow @iravid

### LINKED ISSUE #1467: Taking fiber is not always interrupted during Queue#shutdown
this will deadlock
```scala
  def nonFlaky0(v: => ZIO[Environment, Any, org.specs2.matcher.MatchResult[Any]]): org.specs2.matcher.MatchResult[Any] =
  (1 to 1000).foldLeft[org.specs2.matcher.MatchResult[Any]](true must_=== true) {
    case (acc, _) =>
      acc and unsafeRun(v)
  }

  private def test =
    nonFlaky0 {
      for {
        q <- Queue.bounded[Int](2)
        f <- q.offer(1).forever.fork
        _ <- q.shutdown
        _ <- f.await
      } yield true must_=== true
    }
```
I'll try investigating the issue, but I'm not yet familar with  the queue internals

ghostdogpr:
Fixes #1467
Fixes #1449

In case `take` or `offer` were called concurrently with `shutdown`, the following was possible:
1) fiber A - take (or offer) passes `checkShutdownState`, the queue is valid
2) fiber B - shutdown changes the queue status to `shutdown`
3) fiber B - shutdown cancels all pending fibers (take and offer)
4) fiber A - take (or offer) awaits... forever

What I did is adding another call to `checkShutdownState` before `await`. If the queue was shutdown in the meantime, the current fiber will be interrupted. 

Added 2 new tests using `nonFlaky`, also tried at home with 1 million iterations.

cc @mschuwalow

### DIFF
--- a/core/shared/src/main/scala/zio/Queue.scala
+++ b/core/shared/src/main/scala/zio/Queue.scala
@@ -77,7 +77,7 @@ object Queue {
       p.unsafeDone(IO.succeed(a))
 
     sealed trait Strategy[A] {
-      def handleSurplus(as: List[A], queue: MutableConcurrentQueue[A]): UIO[Boolean]
+      def handleSurplus(as: List[A], queue: MutableConcurrentQueue[A], checkShutdownState: UIO[Unit]): UIO[Boolean]
 
       def unsafeOnQueueEmptySpace(queue: MutableConcurrentQueue[A]): Unit
 
@@ -87,7 +87,11 @@ object Queue {
     }
 
     case class Sliding[A]() extends Strategy[A] {
-      final def handleSurplus(as: List[A], queue: MutableConcurrentQueue[A]): UIO[Boolean] = {
+      final def handleSurplus(
+        as: List[A],
+        queue: MutableConcurrentQueue[A],
+        checkShutdownState: UIO[Unit]
+      ): UIO[Boolean] = {
         @tailrec
         def unsafeSlidingOffer(as: List[A]): Unit =
           as match {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Queue.scala
```diff
@@ -150,7 +162,7 @@ object Queue {
             _ <- (IO.effectTotal {
                   unsafeOffer(as, p)
                   unsafeOnQueueEmptySpace(queue)
-                } *> p.await).onInterrupt(IO.effectTotal(unsafeRemove(p)))
+                } *> checkShutdownState *> p.await).onInterrupt(IO.effectTotal(unsafeRemove(p)))
```
regiskuckaertz: Isn't there a likelihood of that action being interrupted in `*>` right after `checkShutdownState`? Perhaps that part should be uninterruptible?

ghostdogpr: I don’t see the problem if it was interrupted there? There’s a `onInterrupt` to clean it up. 

What I added is just a check that the queue wasn’t shut down before awaiting. If it was shut down, `checkShutdownState` will interrupt the current fiber (no endless waiting) and `onInterrupt` will remove the promise.

regiskuckaertz: Oh, right 👌

PR #1471, opened 2019-08-23
Improve documentation for ZIO#Lock

### LINKED ISSUE #1457: Improve Documentation for ZIO#Lock
There was a question on Gitter about how `lock` works and in particular how it composes (e.g. what happens if you lock an effect to a given executor and that effect itself locks part of its execution to a different executor?). I know there are a lot of other resources out there about `lock` and thread management in ZIO but I think we could have a bit more in the Scaladoc to address this directly since this is a differentiator for us versus other effect systems.

I will take the ticket.

adamgfraser:
Resolves #1457.


ghostdogpr:
@adamgfraser some linting issue it seems, otherwise 👍

Issue #1440, opened 2019-08-19
Error in Fiber.interruptAll


jdegoes:
Reported by @ghostdogpr 

```
- offers are suspended by back pressure
[info]     Fiber failed.
[info]     An unchecked error was produced.
[info]     java.lang.NullPointerException
[info]     	at zio.Fiber$.$anonfun$interruptAll$2(Fiber.scala:296)
[info]     	at zio.ZIO$ZipLeftFn.apply(ZIO.scala:2490)
[info]     	at zio.ZIO$ZipLeftFn.apply(ZIO.scala:2488)
[info]     	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:270)
[info]     	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:602)
[info]     	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info]     	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info]     	at java.lang.Thread.run(Thread.java:748)
[info]     
[info]     Fiber:12547 was supposed to continue to:
[info]       a future continuation at zio.ZIO.ensuring(ZIO.scala:285)
[info]       a future continuation at zio.ZIO.run(ZIO.scala:1007)
[info]       a future continuation at zio.ZIO.bracket_(ZIO.scala:131)
[info]       a future continuation at zio.ZIO.run(ZIO.scala:1007)
[info]       a future continuation at zio.ZIO.bracket_(ZIO.scala:131)
[info]       a future continuation at zio.ZIO.run(ZIO.scala:1007)
[info]       a future continuation at zio.ZManaged.use(ZManaged.scala:634)
[info]       a future continuation at zio.test.TestExecutor$.managed(TestExecutor.scala:25)
[info]       a future continuation at zio.test.Spec.foreachExec(Spec.scala:95)
[info]       a future continuation at zio.ZIOFunctions.foreachParN(ZIO.scala:1896)
[info]       a future continuation at zio.ZIO.forever(ZIO.scala:406)
[info]     
[info]     Fiber:12547 execution trace:
[info]       at zio.Fiber$.interruptAll(Fiber.scala:296)
[info]       at zio.ZIOFunctions.handleChildrenWith(ZIO.scala:2030)
[info]       at zio.internal.FiberContext.getFibers(FiberContext.scala:560)
[...]

PR #1473, opened 2019-08-24
This should fix the occasional NPEs reported in #1440

dkarlinsky:
caused by WeakHashMap (based Set) entries being replaced with nulls.

I assume the way java `Set` is converted to `IndexedSeq` (via an array), is intended to minimize memory churn... Not sure how `.filter(_ != null)` will affect this.

#1440


jdegoes:
Can you run `fmt` to format the code?

PR #1478, opened 2019-08-24
Rename InputStream to Pull

### LINKED ISSUE #1461: Find another name for ZStream.InputStream
Too easy to conflate with `java.io.InputStream`.

cc @jdegoes @vasilmkd @mschuwalow and all other naming experts :-)

vasilmkd:
Also, address stack safety by changing ZStream from being a trait to a class.

Resolves #1461.

@iravid @jdegoes

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -19,6 +19,7 @@ package zio.stream
 import zio._
 import zio.clock.Clock
 import zio.duration.Duration
+import ZStream.Pull
 
 /**
  * A `Stream[E, A]` represents an effectful stream that can produce values of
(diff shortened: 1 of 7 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2138,18 +2100,16 @@ object ZStream extends ZStreamPlatformSpecific {
...
 
   /**
-   * Creates a stream from a scoped [[InputStream]].
+   * Creates a stream from a scoped [[Pull]].
    */
-  final def fromInputStreamManaged[R, E, A](is: ZManaged[R, E, InputStream[R, E, A]]): ZStream[R, E, A] =
-    new ZStream[R, E, A] {
-      def process = is
-    }
+  final def fromPullManaged[R, E, A](pull: ZManaged[R, E, Pull[R, E, A]]): ZStream[R, E, A] =
```
iravid: Let's rename this to `def apply`

vasilmkd: `def apply` and keep curly braces or go with the new `ZStream()`?

iravid: The definition here has to be `def apply(pull: ..) = new ZStream(pull)`, but in the call sites you can then do:

```scala
ZStream {
  <expression that returns pull>
}
```

vasilmkd:
@iravid Ready for review.

PR #1481, opened 2019-08-25
Expand syntax package to cover ZIO

### LINKED ISSUE #1445: Question: Why is syntax for IO
Question that may turn into an easy ticket :)

Why are the implicit conversions in https://github.com/zio/zio/blob/v1.0.0-RC11/core/shared/src/main/scala/zio/syntax/syntax.scala only for `IO[E, A]` and not `ZIO[R, E, A]`?
Historical reasons (from before the introduction of `ZIO`)?

Can we fix those up?

NeQuissimus:
- Syntax now available with ZIO rather than IO
- Changed ZIO.require to return ZIO, not IO
- Fixed type parameters on RIO.require

Fixes #1445


core/shared/src/main/scala/zio/syntax/zio.scala
```diff
@@ -0,0 +1,64 @@
+/*
```
jdegoes: Can we call this file `ZIOSyntax.scala`?

PR #1482, opened 2019-08-25
Add ZStream#zipWithLatest

### LINKED ISSUE #1204: Add ZStream#zipWithLatest

### LINKED ISSUE #1492: Add combinator to tick elements of a ZStream at regular interval
The goal would be to emit the elements of a stream at regular intervals such as in this example.

```scala
val ticks = Stream.repeatEffectWith(IO.succeed(Tick), Schedule.spaced(1.second))
val numbers = Stream.fromIterable(1 to 10)

val tickedNumbers = numbers.zip(ticks).map(_._1)
```

`numbers` could be anything more complex than just emitting numbers for 1 to 10.

It could be having a `ZStream#ticking` combinator

```
trait ZStream[-R, +E, +A] {
  def ticking(initialDelay: Duration, interval: Duration): ZStream[R with Clock, E, A] = ???
}
```

And doing

```
val tickedNumbers = Stream.fromIterable(1 to 10).ticking(3.seconds, 1.second)
```

Or having simpler, more orthogonal methods, such as:

```
object ZStream {
  def tick[A](initialDelay: Duration, interval: Duration, a: A): Stream[Clock, Nothing, A] = ???
}
```

And doing

```scala
val tickedNumbers = Stream.fromIterable(1 to 10).zip(Stream.tick(3.seconds, 1.second, Tick)).map(_._1)
```

Also, having `<&>`, `<&`, `&>` combinators (backed by `zip`combinator for example) might be interesting, but it is questionable in 2 ways:
* `zip` combinator takes 2 queue sizes as additional parameters,
* this might not be the place to discuss those operators in this issue.

It would nevertheless allow to write:

```scala
val tickedNumbers = Stream.fromIterable(1 to 10) <& Stream.tick(3.seconds, 1.second, Tick)
```

regiskuckaertz:
Resolves #1204 
Resolves #1492 

Happy to close this if #1270 is updated (cc @contrun )

I have also added a few combinators I couldn't find how to express easily, but if I missed something I'll get rid of them.

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -58,7 +58,28 @@ object BuildHelper {
     }
   )
 
-  val replSettings = Seq(
+  val replSettings = makeReplSettings {
+    """|import zio._
+       |import zio.console._
+       |import zio.duration._
+       |object replRTS extends DefaultRuntime {}
+       |import replRTS._
+       |implicit class RunSyntax[R >: replRTS.Environment, E, A](io: ZIO[R, E, A]){ def unsafeRun: A = replRTS.unsafeRun(io) }
+    """.stripMargin
+  }
+
+  val streamReplSettings = makeReplSettings {
+    """|import zio._
+       |import zio.console._
+       |import zio.duration._
+       |import zio.stream._
+       |object replRTS extends DefaultRuntime {}
+       |import replRTS._
+       |implicit class RunSyntax[R >: replRTS.Environment, E, A](io: ZIO[R, E, A]){ def unsafeRun: A = replRTS.unsafeRun(io) }
+    """.stripMargin
+  }
+
+  def makeReplSettings(initialCommandsStr: String) = Seq(
     // In the repl most warnings are useless or worse.
     // This is intentionally := as it's more direct to enumerate the few
     // options we do want than to try to subtract off the ones we don't.
@@ -73,14 +94,7 @@ object BuildHelper {
       "-Xsource:2.13",
       "-Yrepl-class-based"
     ),
-    initialCommands in Compile in console := """
-                                               |import zio._
-                                               |import zio.console._
-                                               |import zio.duration._
-                                               |object replRTS extends DefaultRuntime {}
-                                               |import replRTS._
-                                               |implicit class RunSyntax[R >: replRTS.Environment, E, A](io: ZIO[R, E, A]){ def unsafeRun: A = replRTS.unsafeRun(io) }
-    """.stripMargin
+    initialCommands in Compile in console := initialCommandsStr
   )
 
   def extraOptions(scalaVersion: String, optimize: Boolean) =
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1774,8 +1774,8 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
   }
 
   private def zipWithLatest = unsafeRun {
-    val s1 = Stream.iterate(0)(_ + 1).delay(100.millis)
-    val s2 = Stream.iterate(0)(_ + 1).delay(70.millis)
+    val s1 = Stream.iterate(0)(_ + 1).fixed(100.millis)
+    val s2 = Stream.iterate(0)(_ + 1).fixed(70.millis)
 
     withLatch { release =>
       s1.zipWithLatest(s2)((_, _))
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -185,9 +185,10 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
     short circuits          $repeatShortCircuits
 
   Stream.schedule
+    scheduleElementsWith          $scheduleElementsWith
+    scheduleElementsEither        $scheduleElementsEither
     scheduleWith                  $scheduleWith
     scheduleEither                $scheduleEither
-    spaced                        $spaced
     repeated and spaced           $repeatedAndSpaced
     short circuits in schedule    $spacedShortCircuitsWhileInSchedule
     short circuits after schedule $spacedShortCircuitsAfterScheduleFinished
@@ -1472,26 +1473,34 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
       } yield result must_=== List(1, 1)
     )
 
-  private def spaced =
-    unsafeRun(
-      Stream("A", "B", "C")
-        .spaced(Schedule.recurs(0) *> Schedule.fromFunction((_) => "!"))
-        .run(Sink.collectAll[String])
-        .map(_ must_=== List("A", "!", "B", "!", "C", "!"))
-    )
-
   private def scheduleWith =
     unsafeRun(
       Stream("A", "B", "C")
-        .scheduleWith(Schedule.recurs(0) *> Schedule.fromFunction((_) => 123))(identity, _.toString)
+        .scheduleWith(Schedule.recurs(3) *> Schedule.fromFunction((_) => "!"))(_.toLowerCase, identity)
         .run(Sink.collectAll[String])
-        .map(_ must_=== List("A", "123", "B", "123", "C", "123"))
+        .map(_ must_=== List("a", "b", "c", "!"))
     )
 
   private def scheduleEither =
     unsafeRun(
       Stream("A", "B", "C")
-        .scheduleEither(Schedule.recurs(0) *> Schedule.fromFunction((_) => 123))
+        .scheduleEither(Schedule.recurs(3) *> Schedule.fromFunction((_) => "!"))
+        .run(Sink.collectAll[Either[String, String]])
+        .map(_ must_=== List(Right("A"), Right("B"), Right("C"), Left("!")))
+    )
+
+  private def scheduleElementsWith =
+    unsafeRun(
+      Stream("A", "B", "C")
+        .scheduleElementsWith(Schedule.recurs(0) *> Schedule.fromFunction((_) => 123))(identity, _.toString)
+        .run(Sink.collectAll[String])
+        .map(_ must_=== List("A", "123", "B", "123", "C", "123"))
+    )
+
+  private def scheduleElementsEither =
+    unsafeRun(
+      Stream("A", "B", "C")
+        .scheduleElementsEither(Schedule.recurs(0) *> Schedule.fromFunction((_) => 123))
         .run(Sink.collectAll[Either[Int, String]])
         .map(_ must_=== List(Right("A"), Left(123), Right("B"), Left(123), Right("C"), Left(123)))
     )
@@ -1499,7 +1508,7 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
   private def repeatedAndSpaced =
     unsafeRun(
       Stream("A", "B", "C")
-        .spaced(Schedule.recurs(1) >>> Schedule.fromFunction((_) => "!"))
+        .scheduleElements(Schedule.recurs(1) >>> Schedule.fromFunction((_) => "!"))
         .run(Sink.collectAll[String])
         .map(_ must_=== List("A", "A", "!", "B", "B", "!", "C", "C", "!"))
     )
@@ -1507,7 +1516,7 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
   private def spacedShortCircuitsAfterScheduleFinished =
     unsafeRun(
       Stream("A", "B", "C")
-        .spaced(Schedule.recurs(1) *> Schedule.fromFunction((_) => "!"))
+        .scheduleElements(Schedule.recurs(1) *> Schedule.fromFunction((_) => "!"))
         .take(3)
         .run(Sink.collectAll[String])
         .map(_ must_=== List("A", "A", "!"))
@@ -1516,7 +1525,7 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
   private def spacedShortCircuitsWhileInSchedule =
     unsafeRun(
       Stream("A", "B", "C")
-        .spaced(Schedule.recurs(1) *> Schedule.fromFunction((_) => "!"))
+        .scheduleElements(Schedule.recurs(1) *> Schedule.fromFunction((_) => "!"))
         .take(4)
         .run(Sink.collectAll[String])
         .map(_ must_=== List("A", "A", "!", "B"))
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1776,10 +1776,14 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
   private def zipWithLatest = unsafeRun {
     val s1 = Stream.iterate(0)(_ + 1).delay(100.millis)
     val s2 = Stream.iterate(0)(_ + 1).delay(70.millis)
-    s1.zipWithLatest(s2)((_, _))
-      .take(8)
-      .runCollect
-      .map(_ must_=== List(0 -> 0, 0 -> 1, 1 -> 1, 1 -> 2, 2 -> 2, 2 -> 3, 2 -> 4, 3 -> 4))
+
+    withLatch { release =>
+      s1.zipWithLatest(s2)((_, _))
+        .take(8)
+        .runCollect
+        .tap(_ => release)
+        .map(_ must_=== List(0 -> 0, 0 -> 1, 1 -> 1, 1 -> 2, 2 -> 2, 2 -> 3, 2 -> 4, 3 -> 4))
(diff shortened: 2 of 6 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2191,6 +2225,8 @@ object ZStream extends ZStreamPlatformSpecific {
    */
   final def halt[E](cause: Cause[E]): ZStream[Any, E, Nothing] = fromEffect(ZIO.halt(cause))
 
+  final def iterate[A](a: A)(f: A => A): ZStream[Any, Nothing, A] = ZStream.unfold(a)(a => Some(a -> f(a)))
```
iravid: Scaladoc? Pretty please? :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -650,6 +650,25 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   final def concat[R1 <: R, E1 >: E, A1 >: A](other: => ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
     ZStream(UIO.succeed(self), UIO(other)).flatMap(ZStream.unwrap)
 
+  /**
+   * Emits elements of this stream with a fixed delay in between, regardless of how long it
+   * takes to produce a value.
+   */
+  final def delay[R1 <: R, E1 >: E, A1 >: A](duration: Duration): ZStream[R1 with Clock, E1, A1] =
```
iravid: Nice 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -650,6 +650,25 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
   final def concat[R1 <: R, E1 >: E, A1 >: A](other: => ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
     ZStream(UIO.succeed(self), UIO(other)).flatMap(ZStream.unwrap)
 
+  /**
+   * Emits elements of this stream with a fixed delay in between, regardless of how long it
```
iravid: Hmm, if I'm not mistaken, this will only delay the first element, as the ref is updated to `true` afterwards.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1826,6 +1845,36 @@ trait ZStream[-R, +E, +A] extends Serializable { self =>
...
+          is    <- self.mergeEither(that).process
+          state <- Ref.make[(Option[A], Option[B])]((None, None)).toManaged_
+          pull: InputStream[R1, E1, C] = {
+            def go: InputStream[R1, E1, C] = is.flatMap { i =>
+              state
+                .modify[InputStream[R1, E1, C]] {
+                  case (previousLeft, previousRight) =>
+                    i match {
+                      case Left(a) =>
+                        previousRight.fold(go)(b => InputStream.emit(f0(a, b))) -> (Some(a) -> previousRight)
```
iravid: Nicely done! Look ma, no queues!

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1772,6 +1773,15 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
       .map(_ must_=== Left("Ouch"))
   }
 
+  private def zipWithLatest = unsafeRun {
```
iravid: Oh, could you rewrite this with a latch on one of the streams so it won't be flaky in the CI? Or alternatively with the new MockClock.

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: This is how I understand it:
- when the state is false, we just evaluate the action and turn the state to true
- when the state is true, we wait `duration` before evaluating the action and keep the state to true

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
regiskuckaertz: Oh yeah, good catch! I'm on it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: my first formulation was using `raceWith` and three times longer 😅 the `merge*` combinators are a godsend!

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Oh right!! Sorry, Boolean blindness :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Could this be implemented as `stream.spaced(Schedule.fixed(duration))`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1792,6 +1810,35 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    */
   final def zipWithIndex: ZStream[R, E, (A, Int)] =
     mapAccum(0)((index, a) => (index + 1, (a, index)))
+
+  /**
+   * Zips the two streams so that when a value is emitted by either of the two streams,
+   * it is combined with the latest value from the other stream to produce a result.
+   */
+  final def zipWithLatest[R1 <: R, E1 >: E, B, C](that: ZStream[R1, E1, B])(f0: (A, B) => C): ZStream[R1, E1, C] =
```
jdegoes: Nice. 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: I had tried that exact version initially but noticed the schedule would apply to each element of the stream, in this case repeat each element infinitely. I'll try with `stream.spaced(Schedule.fixed(duration) && Schedule.once)` 🤔

jdegoes: Ah right, yes, that should do the trick!

regiskuckaertz: Hmm, I don't think this is going to work. That will repeat once, so each element is repeated once, but also emit the schedule's output, so I am getting:
```scala
scala> Stream(1,2,3).spaced((ZSchedule.fixed(1.second) && Schedule.once).unit).runCollect.unsafeRun
res3: List[AnyVal] = List(1, 1, (), 2, 2, (), 3, 3, ())
```
IIUC also this will behave as follows:
```
1===1()2===2()3===3()
```
whereas what I'm looking for is
```
1===2===3
```
I don't think it is possible to express with `spaced`.

iravid: Right, `spaced` will always add an additional element if there's a delay.

project/BuildHelper.scala
```diff
@@ -58,7 +58,28 @@ object BuildHelper {
...
+       |import replRTS._
+       |implicit class RunSyntax[R >: replRTS.Environment, E, A](io: ZIO[R, E, A]){ def unsafeRun: A = replRTS.unsafeRun(io) }
+    """.stripMargin
+  }
+
+  val streamReplSettings = makeReplSettings {
+    """|import zio._
+       |import zio.console._
+       |import zio.duration._
+       |import zio.stream._
```
iravid: ❤️

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: @regiskuckaertz @iravid Do we need a new version of `spaced` that doesn't? 

This `delay` combinator seems like more common use case and it'd be nice if we could have a way to get all such delays (e.g. `fixed`), even if less common and not rising to the level of a separate specialization.

regiskuckaertz: Yes, I didn't understand `ZStream.spaced` at first because I was trying to map it to `ZSchedule.spaced` which seemed to do what I wanted.

vasilmkd: Can this be implemented in terms of `throttleShape`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1530,6 +1537,24 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    */
   final def runDrain: ZIO[R, E, Unit] = run(Sink.drain)
 
+  final def schedule[R1 <: R, E1 >: E, B](schedule: ZSchedule[R1, A, B]): ZStream[R1 with Clock, E1, B] =
```
iravid: This is really nice. Can it also replace `spaced`?

regiskuckaertz: Yeah I'll try to 👍

regiskuckaertz: There's some weird behaviour when I use a schedule, where the second stream in my test makes progress much faster than the first one in the init phase. I will investigate.

regiskuckaertz: Found it. The schedule will now contribute to deciding the end of the stream. So, if the schedule ends after 2 repeats, only 3 elements will be produced. It seems more intuitive to  me (see the definition of `fixed`), but I'm keen to hear your opinion. If ok I'll move on to look at `spaced`.

iravid: I actually found the previous behavior, in which `decision.cont == false` means that the schedule is restarted, to be more intuitive. This is also more in line with `aggregateWithin`, which restarts the schedule when it stops.

Could you explain what's the advantage?

By the way, we can attain this behavior with the previous formulation if you emit the schedule's output into the stream. We can then use `takeWhile` or `collectWhile` to end the stream when the schedule ends.

regiskuckaertz: I don't know if there's an advantage, I guess it's more that I got confused between the name and the intent. `spaced` has different meanings between schedule and streams and that's where I got confused: in the former it means there's a fixed duration between each cycle, in the latter it can mean any kind of schedule, even ones that don't wait.

iravid: Ok, I'm fairly certain that the previous formulation is more correct and we'd want to distribute the schedule's effect repeatedly over the stream rather than have it short-circuit the stream.

If you restart the schedule when it ends, that formulation lets you use `Schedule.recurs(2)` to repeat each element twice.

regiskuckaertz: Right 🤦‍♂ so I literally woke up during the night with a solution following your suggestion with `collectAll`:
- this function will be `scheduleWith`, emitting both the stream and schedule's output with two functions for aligning types
- `scheduleEither` will map to either (and replace the current `spacedEither`)
- I am not sure about my `fixed` and the existing `spaced`, there seems to be a semantic clash, but in any case the `fixed` is just `scheduleEither(...).collectAll(case Left(x) => x)` indeed.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1792,6 +1798,35 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+  /**
+   * Zips the two streams so that when a value is emitted by either of the two streams,
+   * it is combined with the latest value from the other stream to produce a result.
+   */
+  final def zipWithLatest[R1 <: R, E1 >: E, B, C](that: ZStream[R1, E1, B])(f0: (A, B) => C): ZStream[R1, E1, C] =
+    ZStream[R1, E1, C] {
+      for {
+        is    <- self.mergeEither(that).process
+        state <- Ref.make[(Option[A], Option[B])]((None, None)).toManaged_
+        pull: Pull[R1, E1, C] = {
```
quelgar: Nit: you could just put the `pull` expression directly in the `yield`.

regiskuckaertz: It's kind of a pattern in the `ZStream` class, I think it's a little easier to read.

regiskuckaertz:
@iravid I think this is ready for a final review, other schedule-based combinators can be documented in the ticket? I'm happy to pick them up right after... but I wonder if sth more useful would be to convert tests to the new framework?

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Hehe, I had a realization myself. There are two ways to apply a schedule to a stream with regards to the schedule's `cont`.

Given a stream `A B C` and a schedule `(Schedule.exponential(2.millis) && Schedule.recur(2) && Schedule.const("Done")).map(_._2)`, these are the interpretations:
1. As long as the schedule wants to continue, repeat the current element after a delay:
`A |-2ms-| A |-4ms-| A |-8ms-| "Done" B |-2ms-| B |-4ms-| B |-8ms-| "Done" ...`

2. As long as the schedule wants to continue, emit the stream's next element after a delay:
`A |-2ms-| B |-4ms-| C |-8ms-| "Done"`

The first interpretation is the existing `spaced`. I guess that this implementation should be the second interpretation.

regiskuckaertz: dang, I was quite happy with the terseness of `scheduleWith`. there may be some confusion as to which does what, are you ok with naming the one that acts on each elements `scheduleElementsWith` and the one that acts on the whole stream `scheduleWith` (I also think it would fix the confusion between `ZStream.spaced` and `ZSchedule.spaced`)?

iravid: Yeah, those names are good! We can bring `ZStream#repeat` inline with that too:
- The signature can be changed to:
```scala
final def repeat[R1 <: R, B](schedule: ZSchedule[R1, Any, B]): ZStream[R1 with Clock, E, Either[B, A]]
```
- And a `repeatWith` variant can be added.

Can be done in a follow-up of course :-)

iravid:
@regiskuckaertz Help with the tests conversion would be much much appreciated 🙏🏻 but really, pick up whatever you fancy the most :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: Let's do in a follow-up, I'd like to see where the ticket on combinators is going.

iravid:
@regiskuckaertz could you amend a commit to re-trigger the CI?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1531,58 +1540,121 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+            if (done) Pull.end
+            else
+              for {
+                a <- as.optional.mapError(Some(_))
+                c <- a match {
+                      case Some(a) =>
+                        for {
+                          decision <- schedule.update(a, sched)
+                          _        <- clock.sleep(decision.delay)
+                          _        <- state.set((!decision.cont, decision.state, Some(decision.finish)))
```
iravid: Hmm didn't we say that the schedule should not stop the stream?

iravid: Ah I see that my example was unclear. In this formulation, I meant that the schedule should restart between the elements. So for a stream `"A B C A B C"`, this should be result:
```
A |-2ms-| B |-4ms-| C |-8ms-| "Done" A |-2ms-| B |-4ms-| C |-8ms-| "Done"
```

regiskuckaertz: Nope, this is implement for point 2 above:

> 2. As long as the schedule wants to continue, emit the stream's next element after a delay

iravid: And the short circuiting can be attained by `collectWhile { case Right(v) => v }`, which would stop on the first `Left`.

iravid: (sorry for dragging this!)

iravid: Actually, let's merge this and address in a follow-up.

regiskuckaertz: That is the semantics of `scheduleElements`, isn't it? 🤔 I could add a test to prove that.

regiskuckaertz:
Thanks 🙏

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Ah not quite. `scheduleElements` repeats every element and restarts

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Ending with the schedule is more "foundational", since you can turn a terminating schedule into a forever schedule with `Schedule#forever`. However it is less convenient. Usually this argues for two methods: one to do the more essential thing, and the other, more visible and obvious one, to do the `Schedule#forever` thing.

PR #1483, opened 2019-08-25
Support a custom executor in StreamEffect

iravid:
This is needed to support running the thunk on the blocking executor in `fromInputStream`.

@jdegoes - wdyt? It's a bit patchy, but I'm not sure if we need anything more than this for now.

cc @vasilmkd

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -16,25 +16,34 @@
 
 package zio.stream
 
-import java.io.{ IOException, InputStream }
-
 import zio._
+import zio.internal.Executor
 
-private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E, () => A])
-    extends ZStream[R, E, A](
-      processEffect.map { thunk =>
-        UIO.effectTotal {
-          try UIO.succeed(thunk())
-          catch {
-            case StreamEffect.Failure(e) => IO.fail(Some(e.asInstanceOf[E]))
-            case StreamEffect.End        => IO.fail(None)
-          }
-        }.flatten
+private[stream] class StreamEffect[-R, +E, +A](
+  val processEffect: ZManaged[R, E, () => A],
+  val executor: ZIO[R, Nothing, Executor]
+) extends ZStream[R, E, A](
+      processEffect.flatMap { thunk =>
+        executor.toManaged_.map { exec =>
+          UIO.effectTotal {
+            try UIO.succeed(thunk())
+            catch {
+              case StreamEffect.Failure(e) => IO.fail(Some(e.asInstanceOf[E]))
+              case StreamEffect.End        => IO.fail(None)
+            }
+          }.flatten.lock(exec)
+        }
       }
     ) { self =>
 
+  def copy[R1, E1, A1](
+    processEffect0: ZManaged[R1, E1, () => A1] = processEffect,
+    executor0: ZIO[R1, Nothing, Executor] = executor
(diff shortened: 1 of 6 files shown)


iravid:
Ah, this forgets the executor between the combinators. Fixing this now.

iravid:
For interruption, we'll need to support different interruption strategies that propagate on the `StreamEffect` between the combinators. E.g. `fromInputStream` wants `Thread.interrupt`, `filter` wants `Fiber.unsafeIsInterrupted`, and `fromInputStream(...).filter(...)` wants both.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -17,22 +17,33 @@
...
-          catch {
-            case StreamEffect.Failure(e) => IO.fail(Some(e.asInstanceOf[E]))
-            case StreamEffect.End        => IO.fail(None)
-          }
-        }.flatten
+import zio.internal.Executor
+
+private[stream] class StreamEffect[-R, +E, +A](
+  val processEffect: ZManaged[R, E, () => A],
+  val executor: ZIO[R, Nothing, Executor]
```
jdegoes: As far as I can tell, this structure doesn't compose with well-defined semantics. Consider, e.g. `x ++ y` — what is the `executor` for such a composite?

It seems like we need to say three things:

 - Invocation of "open" occurs on the blocking thread pool, which we can do in the `reserve` of the managed
 - Invocation of "close" occurs on the blocking thread pool, which we can do in the `reserve` of managed (on finalization)
 - Invocation of the "pull" occurs on the blocking thread pool, which we can also do (below, using `effectBlocking` instead of `effectTotal`, etc.).

So unless I'm missing something, we can do this w/o new machinery. Right? Or wrong? 😄

iravid: Say `x: StreamEffect` uses a Blocking executor, `y: StreamEffect` uses a regular executor, and `z: Stream` that is asynchronous.

`x ++ y ++ z` does the following:
- opens `x` and starts pulling it. The `Pull` on `x` evaluates the `() => A` thunk while locked to the Blocking executor, so that pull runs on the blocking executor;
- opens `y` and starts pulling it. Similarly, it evaluates the thunk while locked to the default executor, so that pull runs there.
- opens `z` and starts pulling it. That `Pull` does whatever it wants because it's effectful.

From how I'm reading things, the semantics here are identical to the semantics of `lock`, as long as we always propagate the executor. Wdyt?

iravid: @jdegoes 

Sorry, I realized I didn't respond to your questions about whether it's possible to do this without new machinery.

The problem is as follows: how does a constructor (like `fromInputStream`), which only produces a thunk `Managed[E, () => A]`, indicate that when a stream which is based on it is run, it should be run on the Blocking threadpool? We can't use `effectBlocking` because we want to produce synchronous code that runs in the thunk.

iravid: @jdegoes Any thoughts on the above? :-)

iravid: We hashed it over deadlift sets. I believe @jdegoes is onboard. I'll pick this back up next week :-)

vasilmkd: Awesome. Please keep me in the review loop.

jdegoes:
@iravid What is the status of this one?

iravid:
I'll circle back to this soon. Let's close it out for now.

Issue #14, opened 2018-06-09
Demote `KleisliIO` from being a function


jdegoes:
Instead of being a function `A => IO[E, B]`, it should have a field called `run` which is a function `A => IO[E, B]`.

fabianpage:
Hi

Why do you want to have KleisliIO to have a run field insted of being a function?

I tried to do that but have the feeling it's to simple and i have probably done something wrong. ;) (Just added a val run to KleisliIO, removed the extends and made all test work again)

ktonga:
@fabianpage more info in this thread https://github.com/scalaz/scalaz/pull/1857#discussion_r193571323

PR #1488, opened 2019-08-26
ZIO Test: Allow Mock Environment To Access Real Environment

### LINKED ISSUE #1474: Allow tests to access real (non-mock) environment
Currently tests and test aspects (like `timeout`) will always access the environment provided to them, which will be a default mock environment in `DefaultRunnableSpec`.

This is a serious obstacle, as test combinators need to use real timeouts, real system clocks, real consoles, etc. To solve the problem, inside `zio.test.mock` package, we can introduce a new module and service:

```scala
package zio.test.mock

trait Live[+R] {
  def live: Live.Service[R]
}
object Live {
  trait Service[+R] {
    def provide[E, A](zio: ZIO[R, E, A]): IO[E, A]
  }
  def live[R, E, A](zio: ZIO[R, E, A]): ZIO[Live[R], E, A] = 
    ZIO.accessM[Live[R]](_.live.provide(zio))
}
```

Now all test combinators (for example, `TestAspect.timeout`) can move from requiring, e.g. `Clock`, to requiring `Live[Clock]`.

Then the default environment that we provide in tests (`zio.test.mock.mockEnvironmentManaged` / `zio.test.MockEnvironment.Value`) can provide `Live[DefaultRuntime#Environment]`, by supplying an implementation of the live environment module that is backed by a "live" environment.

This should provide a clean way to get access to the "real" environment when necessary—as it is necessary for all test machinery—while not increasing number of type parameters or adding any complexity for the common case.

/cc @adamgfraser

adamgfraser:
Resolves #1474.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -16,10 +16,12 @@
 
 package zio.test
 
-import zio.{ Cause, ZIO, ZManaged, ZSchedule }
+import java.util.concurrent.TimeoutException
+
+import zio.{ clock, Cause, ZIO, ZManaged, ZSchedule }
 import zio.duration.Duration
 import zio.clock.Clock
-import java.util.concurrent.TimeoutException
+import zio.test.mock.Live
 
 /**
  * A `TestAspect` is an aspect that can be weaved into specs. You can think of
@@ -183,12 +185,26 @@ object TestAspect {
   /**
    * An aspect that retries failed tests according to a schedule.
    */
-  def retry[R0, E0](schedule: ZSchedule[R0, E0, Any]): TestAspect[Nothing, R0 with Clock, Nothing, E0] =
-    new TestAspect.PerTest[Nothing, R0 with Clock, Nothing, E0] {
-      def perTest[R >: Nothing <: R0 with Clock, E >: Nothing <: E0](
+  def retry[R0, E0](schedule: ZSchedule[R0, E0, Any]): TestAspect[Nothing, R0 with Live[Clock], Nothing, E0] =
+    new TestAspect.PerTest[Nothing, R0 with Live[Clock], Nothing, E0] {
+      def perTest[R >: Nothing <: R0 with Live[Clock], E >: Nothing <: E0](
         test: ZIO[R, E, TestResult]
-      ): ZIO[R, E, TestResult] =
-        test.retry(schedule: ZSchedule[R0, E0, Any])
+      ): ZIO[R, E, TestResult] = {
+        def loop(state: schedule.State): ZIO[R with Live[Clock], E, TestResult] =
+          test.foldM(
+            err =>
+              schedule
+                .update(err, state)
+                .flatMap(
+                  decision =>
+                    if (decision.cont) Live.live(clock.sleep(decision.delay)) *> loop(decision.state)
+                    else ZIO.fail(err)
+                ),
+            succ => ZIO.succeed(succ)
+          )
+
+        schedule.initial.flatMap(loop)
+      }
     }
 
   /**
@@ -199,14 +215,18 @@ object TestAspect {
   /**
    * An aspect that times out tests using the specified duration.
    */
-  def timeout(duration: Duration): TestAspect[Nothing, Clock, Nothing, Any] =
-    new TestAspect.PerTest[Nothing, Clock, Nothing, Any] {
-      def perTest[R >: Nothing <: Clock, E >: Nothing <: Any](test: ZIO[R, E, TestResult]): ZIO[R, E, TestResult] =
-        test.timeout(duration).map {
-          case None =>
-            Assertion
-              .failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
-          case Some(v) => v
+  def timeout(duration: Duration): TestAspect[Nothing, Live[Clock], Nothing, Any] =
+    new TestAspect.PerTest[Nothing, Live[Clock], Nothing, Any] {
+      def perTest[R >: Nothing <: Live[Clock], E >: Nothing <: Any](
+        test: ZIO[R, E, TestResult]
+      ): ZIO[R, E, TestResult] =
+        ZIO.environment[R].flatMap { r =>
+          Live.live(test.provide(r).timeout(duration)).map {
+            case None =>
+              Assertion
+                .failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
+            case Some(v) => v
+          }
         }
     }
 
@@ -222,5 +242,4 @@ object TestAspect {
         case Spec.TestCase(label, test)  => Spec.TestCase(label, if (predicate(label)) perTest(test) else test)
       }
   }
-
 }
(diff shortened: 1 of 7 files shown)


ghostdogpr:
@adamgfraser should I make a dedicated ticket for improving the default behavior when a test is stuck (as discussed in #1474 )?

adamgfraser:
@ghostdogpr Yes let's do it.

test/shared/src/main/scala/zio/test/mock/Live.scala
```diff
@@ -0,0 +1,50 @@
...
+  def make: UIO[Live[DefaultRuntime#Environment]] =
+    makeService.map { service =>
+      new Live[DefaultRuntime#Environment] {
+        val live = service
+      }
+    }
+
+  def makeService: UIO[Live.Service[DefaultRuntime#Environment]] =
+    UIO.succeed {
+      val runtime     = new DefaultRuntime {}
```
jdegoes: I would accept `runtime` here (rather than creating a new default runtime), which gives users more flexibility and reduces total thread pools; and maybe even make this generic in `R`, something like, `def makeService[R](env: R): UIO[Live.Service[R]]`

test/shared/src/test/scala/zio/test/mock/LiveSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+  def run(implicit ec: ExecutionContext): List[Future[(Boolean, String)]] = List(
+    label(liveCanAccessRealEnvironment, "live can access real environment")
+  )
+
+  def liveCanAccessRealEnvironment: Future[Boolean] =
+    unsafeRunToFuture {
+      val io = for {
+        mock <- clock.currentTime(TimeUnit.MILLISECONDS)
+        live <- Live.live(clock.currentTime(TimeUnit.MILLISECONDS))
+      } yield mock == 0 && live != 0
```
jdegoes: Really love how this looks! We should consider re-exporting `live` as a method inside the `mock` package object, so a `import zio.test.mock._` brings it into scope.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -183,12 +185,26 @@ object TestAspect {
...
+                .update(err, state)
+                .flatMap(
+                  decision =>
+                    if (decision.cont) Live.live(clock.sleep(decision.delay)) *> loop(decision.state)
+                    else ZIO.fail(err)
+                ),
+            succ => ZIO.succeed(succ)
+          )
+
+        schedule.initial.flatMap(loop)
```
ghostdogpr: How about getting the live clock with `access`, then `provide` it, so that we reuse the implementation of `retry`?

adamgfraser: I tried that but was struggling to make it work. The issue is that the schedule also has a `R` parameter other than `Clock` but as far as I am aware there isn't a method like `provide` to eliminate that requirement. So between not being able to eliminate the `R` requirements by providing them and the limitations on `provideSome` this was the best I could come up with.

ghostdogpr: Oh yeah I see, there's no way to rebuild the R in `provideSome`...

test/shared/src/main/scala/zio/test/mock/Live.scala
adamgfraser: Great thought. Made it generic in the latest version.

test/shared/src/test/scala/zio/test/mock/LiveSpec.scala
adamgfraser: Yes, it is really nice. I re-exported it in the package object in the latest version.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -199,14 +215,18 @@ object TestAspect {
...
-            Assertion
-              .failure(FailureDetails.Runtime(Cause.fail(new TimeoutException(s"Timeout of ${duration} exceeded"))))
-          case Some(v) => v
+  def timeout(duration: Duration): TestAspect[Nothing, Live[Clock], Nothing, Any] =
+    new TestAspect.PerTest[Nothing, Live[Clock], Nothing, Any] {
+      def perTest[R >: Nothing <: Live[Clock], E >: Nothing <: Any](
+        test: ZIO[R, E, TestResult]
+      ): ZIO[R, E, TestResult] =
+        ZIO.environment[R].flatMap { r =>
+          Live.live(test.provide(r).timeout(duration)).map {
```
jdegoes: This is really interesting (get the environment, provide it on the inside, do an outside effect, lift that up to `Live`). 

It makes me think we can have a combinator:

```scala
def mockLive[R, R1, E, A](zio: ZIO[R, E, A])(f: IO[E, A] => ZIO[R with Live[R1], E, B]): ZIO[R, E, B] = 
  ZIO.environment[R].flatMap(r => Live.live(f(zio.provide(r))))
```

Then use it like so:

```scala
Live.mockLive(test, _timeout(duration))
```

I don't like the name and there may be an even more generic combinator for this....

adamgfraser: Maybe something like?

```scala
def withLive[R, R1, E, A, B](zio: ZIO[R, E, A])(f: IO[E, A] => ZIO[R1, E, B]): ZIO[R with Live[R1], E, B] =
  ZIO.environment[R].flatMap(r => live(f(zio.provide(r))))
```

The type inference on the second parameter seems to be pretty poor. I don't know if there is anything we can do about that.

jdegoes: We can improve type inference.

There are a few options.

### Option 1

Use one parameter list and require that `R <: Live[R1]`.

```scala
def withLive[R1, R <: Live[R1], E, A, B](zio: ZIO[R, E, A], f: IO[E, A] => ZIO[R1, E, B]): ZIO[R, E, B] =
```

### Option 2

Create a temporary class to allow incrementally introducing type parameters for each parameter list:

```scala
def withLive[R, E, A](zio: ZIO[R, E, A]): WithLive[R, E, A] = 
  WithLive[R, E, A](zio)

final case class WithLive[R, E, A](zio: ZIO[R, E, A]) {
  def apply[R1, B](f: IO[E, A] => ZIO[R1, E, B]): ZIO[R with Live[R1], E, B] =
    ZIO.environment[R].flatMap(r => live(f(zio.provide(r))))
}
```

### Option 3

Flip order so we can get `R1` before `R`. I don't think this will infer well so I won't right it (well, some parts of it will infer better, but overall worse).

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -183,12 +185,26 @@ object TestAspect {
...
         test: ZIO[R, E, TestResult]
-      ): ZIO[R, E, TestResult] =
-        test.retry(schedule: ZSchedule[R0, E0, Any])
+      ): ZIO[R, E, TestResult] = {
+        def loop(state: schedule.State): ZIO[R with Live[Clock], E, TestResult] =
+          test.foldM(
+            err =>
+              schedule
+                .update(err, state)
+                .flatMap(
```
jdegoes: It will be easier to eliminate this after #14

PR #1497, opened 2019-08-26
ZIO Test: Implement Live#withLive

adamgfraser:
Following up on conversations in #1488, adds a `withLive` combinator for transforming an effect with the "real" environment while guaranteeing that the effect itself will continue to use the mock environment.

@jdegoes I realized that the type inference issue was actually the result of the `E` type parameter. We had `IO[E, A] => ZIO[R1, E, B]` but we actually need `IO[E, A] => ZIO[R1, E1, B]`. The transformation function could potentially change the `E` to any type (e.g. catching all errors to eliminate it entirely or doing something dangerous to widen it to `Throwable`).


jdegoes:
@adamgfraser Just a conflict with `TestAspect` then it will be good to merge!

Issue #1462, opened 2019-08-22
Implement ZSink#andThen


iravid:
Signature:
```scala
trait ZSink[R, E, A0, A, B] {
  def andThen[R1 <: R, E1 >: E, A00 >: A0, C](that: ZSink[R1, E1, A00, B, C]): ZSink[R1, E1, A00, A, C]
}
```

Motivation: things like `Stream.fromInputStream(java.io.InputStream).transduce(ZSink.utf8DecodeChunk andThen ZSink.splitLines)` - voila, super-efficient fused UTF8 decoding and line splitting.

iravid:
This is currently impossible to implement generally.

PR #1498, opened 2019-08-26
#1462 ZSink#andThen - attempt [WIP]

dkarlinsky:
@iravid 
So, I gave #1462 a try, though it's all new to me and I'm pretty much fumbling in the dark here.
But it's fun to try :)
Here's what I've come up with so far.
It compiles and seems to be doing what you want, in case there is no leftovers at least (the first test `utf8DecodeChunkAndThenSplitLines` passes)
Leftover I can't really figure out. If there is leftover in the second sink (like `ZSink.splitLines`), I'm not sure how to reflect it as leftover in the combined sink...

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1883,4 +1883,9 @@ object ZSink extends ZSinkPlatformSpecific {
 
       def cont(state: State) = state._3
     }
+
+  implicit class Ops[R, E, A0, A, B](val self: ZSink[R, E, A0, A, B]) extends AnyVal {
+    final def andThen[R1 <: R, E1 >: E, C, A00](that: ZSink[R1, E1, B, B, C]): ZSink[R1, E1, A0, A, C] =
+      new AndThenSink[R1, R, E1, E, A, B, A0, C, A00](self, that)
+  }
 }
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1884,8 +1884,8 @@ object ZSink extends ZSinkPlatformSpecific {
       def cont(state: State) = state._3
     }
 
-  implicit class Ops[R, E, A0, A, B](val self: ZSink[R, E, A0, A, B]) extends AnyVal {
-    final def andThen[R1 <: R, E1 >: E, C, A00](that: ZSink[R1, E1, B, B, C]): ZSink[R1, E1, A0, A, C] =
-      new AndThenSink[R1, R, E1, E, A, B, A0, C, A00](self, that)
+  implicit class Ops[R, E, A, B](val self: ZSink[R, E, A, A, B]) extends AnyVal {
+    final def andThen[R1 <: R, E1 >: E, C](that: ZSink[R1, E1, B, B, C]): ZSink[R1, E1, A, A, C] =
+      new AndThenSink[R1, R, E1, E, A, B, C](self, that)
   }
 }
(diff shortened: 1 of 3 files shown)


iravid:
Hey @dkarlinsky, thanks for picking this up! I thought this through for a bit, and I think that the only viable strategy is to not allow for leftovers in `that`. This is not to say that we should require `A00 = Nothing`, because that would be too limiting. But rather, we should always signal `Step.more` when `that` has leftovers, and we should buffer those leftovers and re-step them when more input comes in from `this`.

Does that make sense?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -901,6 +901,36 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        that.initial
+
+      /**
+       * Steps through one iteration of the sink
+       */
+      override def step(state: State, a: A): ZIO[R1, E1, Step[State, A0]] =
+        for {
+          s0 <- self.initial
+          s1 <- self.step(Step.state(s0), a)
+          b  <- self.extract(Step.state(s1))
```
iravid: You'll need to check if `this` is actually done here, before extracting. Otherwise this may fail for sinks that do not allow for partial extracts.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -901,6 +901,36 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
   )(f: (B, C) => D)(implicit ev: A00 =:= A1): ZSink[R1, E1, A00, A1, D] =
     zip(that).map(f.tupled)
 
+  final def andThen[R1 <: R, E1 >: E, A00 >: A0, C, A000](that: ZSink[R1, E1, A00, B, C]): ZSink[R1, E1, A0, A, C] =
```
iravid: We should also add a non-IO version of this, overriden in `SinkPure`, for fusing together sinks that do not require IO effects.

streams/shared/src/main/scala/zio/stream/ZSink.scala
dkarlinsky: > Hey @dkarlinsky, thanks for picking this up! I thought this through for a bit, and I think that the only viable strategy is to not allow for leftovers in `that`. This is not to say that we should require `A00 = Nothing`, because that would be too limiting. But rather, we should always signal `Step.more` when `that` has leftovers, and we should buffer those leftovers and re-step them when more input can be extracted from `this`.
> 
> Does that make sense?

I think I understand the general idea, but I maybe be missing something:
If I buffer leftovers for `that`, which is of type `Chunk[A00]`, how do I try to feed them back into next `that.step()`, which expects type `B`?

iravid:
Good point! I think that’s good evidence that the `that` sink should have leftovers = inputs.

dkarlinsky:
Ok, I'm trying to make `A0` be the same as `B` for `that` sink, so I defined the method as follows:
```scala
final def andThen[R1 <: R, E1 >: E, C, A00](that: ZSink[R1, E1, B, B, C]): ZSink[R1, E1, A0, A, C]
```
But now I made the variance gods angry :)
```
[error] /Users/dmitryka/dev/com.github/zio/zio/streams/shared/src/main/scala/zio/stream/ZSink.scala:942:47: covariant type B occurs in contravariant position in type zio.stream.ZSink[R1,E1,B,B,C] of value that
[error]   final def andThen[R1 <: R, E1 >: E, C, A00](that: ZSink[R1, E1, B, B, C]): ZSink[R1, E1, A0, A, C] =
[error]                                               ^
```

jdegoes:
@dkarlinsky This one has badly bit-rotted, should I close for now? /cc @iravid

iravid:
Yes. We've come to the conclusion that a generalized solution is not possible at the moment.

Issue #1313, opened 2019-08-01
Test remainder logic of Sink combinators


vasilmkd:
`zip*`, `orElse` and `race*` combinators have non-trivial remainder logic that should be tested.

ghostdogpr:
Closed by #1560

PR #1501, opened 2019-08-27
Fix and test ZStream#optional leftover logic

vasilmkd:
Proposed solution for Sinks that produce remainder of the same type (with more combinators to be migrated later) and Sinks that produce no remainder at all, to help with type inference.

@iravid and I already tried adding implicit evidence, and there are other combinators that don't infer properly.

@iravid Please take a look.

Tests for lefotvers issue #1313.

cc @jdegoes

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -137,13 +137,12 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
    * will not improve performance, but can be used to adapt non-chunked sinks
    * wherever chunked sinks are required.
    */
-  final def chunked[A1 >: A0, A2 <: A]: ZSink[R, E, A1, Chunk[A2], B] =
-    new ZSink[R, E, A1, Chunk[A2], B] {
+  final def chunked: ZSink[R, E, A0, Chunk[A], B] =
+    new ZSink[R, E, A0, Chunk[A], B] {
       type State = self.State
-      val initial = self.initial
-      def step(state: State, a: Chunk[A2]): ZIO[R, E, Step[State, A1]] =
-        self.stepChunk(state, a)
-      def extract(state: State): ZIO[R, E, B] = self.extract(state)
+      val initial                         = self.initial
+      def step(state: State, a: Chunk[A]) = self.stepChunk(state, a)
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -904,6 +869,83 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
 }
 
 object ZSink extends ZSinkPlatformSpecific {
+
+  implicit class InvariantOps[R, E, A, B](val sink: ZSink[R, E, A, A, B]) extends AnyVal {
```
iravid: What happens if you mark `A` covariant here? `+A`

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -904,6 +869,83 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        val initial = sink.initial.map(Step.leftMap(_)(Option(_))) orElse
+          UIO.succeed(Step.done(None, Chunk.empty))
+
+        def step(state: State, a: A) =
+          state match {
+            case None => UIO.succeed(Step.done(state, Chunk.empty))
+            case Some(state) =>
+              sink
+                .step(state, a)
+                .foldM(
```
regiskuckaertz: It looks like you could just use `fold` and drop the `UIO.succeed` here.

vasilmkd: Thanks. Done.

vasilmkd:
@iravid Do you want me to drop the syntax classes until a decision has been made regarding #1509?

iravid:
No, let's keep them for now and revisit this with #1509. Just waiting for the CI to come back.

vasilmkd:
Since they are staying, can I push some other combinators as well? Of course, we're not merging while CI is unavailable?

iravid:
Sure! Let's add just the ones that have bad inference (e.g. have no parameter lists)

vasilmkd:
@iravid Ready for review.

vasilmkd:
`RTSSpec` is failing. Nothing new.

vasilmkd:
> Awesome. The tests are a great testament for the improved inference.

You know we are doing something right when the documentation needs to be amended! :laughing:

PR #1507, opened 2019-08-27
ZIO Test: Improve Handling of Timeouts

### LINKED ISSUE #1489: Improve default handling of timeouts in ZIO Test
Currently a flaky ZIO Test that doesn't have a timeout will cause the tests to hang forever with no output. The only option is to kill the tests and it won't tell you which test was hanging.

We could:
- add some message that tells you which test is causing the problem e.g. `In Suite "Integers", test "addition" has taken more than X minutes to execute. If this is not expected, consider using TestAspect.timeout to timeout runaway tests for faster diagnostics.`

- have a large enough default timeout so that CI jobs don't hang forever? CircleCI has a 10min timeout in case there is no output, but if we add the message above, it may keep it running forever.

cc @adamgfraser @jdegoes

adamgfraser:
This PR defines a new `TimeoutStrategy` type that applies logic for handling long-running or non-terminating tests to a Spec. There are currently three timeout strategies:

1. Error - fail the test after the specified duration
2. Warn - print a warning to the console identifying the test after the specified duration
3. Ignore - don't do anything

Timeout strategies are just functions so they can also compose using normal function composition (e.g. you could warn after 1 minute and timeout after 2 minutes). I also added a parameter to `DefaultRunnableSpec` to set the timeout strategy. Currently I have a default parameter of warning after 60 seconds and in our own tests I am timing out after 60 seconds, but we can definitely change.

Resolves #1489.


core-tests/shared/src/test/scala/zio/ZQueueSpec.scala
```diff
@@ -662,6 +662,7 @@ object ZQueueSpec
         testM("queue isShutdown") {
           for {
             queue <- Queue.bounded[Int](5)
+            _     <- ZIO.unit.delay(5.seconds).provide(Clock.Live)
```
ghostdogpr: This was for testing maybe?

ghostdogpr:
Now a conflict in this one ☝️

PR #1508, opened 2019-08-28
Refactor ZStream.fromQueue

### LINKED ISSUE #1380: ZStream#fromQueue interruption
Currently ZStream#fromQueue is implemented like this
```scala
  final def fromQueue[R, E, A](queue: ZQueue[_, _, R, E, _, A]): ZStream[R, E, A] =
    unfoldM(()) { _ =>
      queue.take
        .map(a => Some((a, ())))
        .foldCauseM(
          cause =>
            // Dequeueing from a shutdown queue will result in interruption,
            // so use that to signal the stream's end.
            if (cause.interrupted) ZIO.succeed(None)
            else ZIO.halt(cause),
          ZIO.succeed
        )
    }
```
It hides the effect that upstream has been interrupted / the queue has been shutdown.
As most of the ZStream code uses Take to precisely pass around when / why a stream has ended I wonder if this is still useful behavior

regiskuckaertz:
Closes #1380 

It seems like a more efficient formulation too.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2139,23 +2139,6 @@ object ZStream extends ZStreamPlatformSpecific {
     }
 
   /**
-   * Creates a stream from a [[zio.ZQueue]] of queue `take` operations
-   */
-  final def fromQueueTake[R, E, A](queue: ZQueue[_, _, R, E, _, Take[E, A]]): ZStream[R, E, A] =
-    ZStream[R, E, A] {
-      ZManaged.reserve(
-        Reservation(
-          UIO(queue.take.mapError(Some(_)).flatMap {
-            case Take.End      => Pull.end
-            case Take.Fail(c)  => Pull.halt(c)
-            case Take.Value(a) => Pull.emit(a)
-          }),
-          _ => queue.shutdown
-        )
-      )
-    }
-
-  /**
    * Creates a stream from a [[zio.ZQueue]] of values. The queue will be shutdown once the stream is closed.
    */
   final def fromQueueWithShutdown[R, E, A](queue: ZQueue[_, _, R, E, _, A]): ZStream[R, E, A] =
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2128,7 +2128,15 @@ object ZStream extends ZStreamPlatformSpecific {
     ZStream[R, E, A] {
       ZManaged.reserve(
         Reservation(
-          UIO(queue.take.mapError(Some(_))),
+          UIO(
+            queue.take.catchAllCause(
+              c => queue.isShutdown.flatMap(down => if (down && c.interrupted) Pull.end else Pull.halt(c))
+            )
+          ),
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
> we keep the semantics of end-on-interrupt in `fromQueue`

do you agree that this is only when the queue has been shutdown? if the interruption happens because of some other cause, then it should be propagated? something like this then:
```scala
queue.take.catchAllCause { c =>
  queue.isShutdown.flatMap(if (_ && c.interrupted) Pull.end else Pull.halt(c))
}
```

iravid:
Yes, sorry! Meant to write "end-on-shutdown".

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2125,17 +2125,21 @@ object ZStream extends ZStreamPlatformSpecific {
...
+          UIO(
+            queue.take.catchAllCause(
+              c => queue.isShutdown.flatMap(down => if (down && c.interrupted) Pull.end else Pull.halt(c))
+            )
+          ),
+          _ => queue.shutdown
         )
+      )
+    }
+          _ => queue.shutdown
```
regiskuckaertz: 🤔 that's not at all what I just pushed 🤣

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2125,17 +2125,34 @@ object ZStream extends ZStreamPlatformSpecific {
...
+          ),
+          _ => queue.shutdown
         )
+      )
+    }
+
+  /**
+   * Creates a stream from a [[zio.ZQueue]] of queue `take` operations
+   */
+  final def fromQueueTake[R, E, A](queue: ZQueue[_, _, R, E, _, Take[E, A]]): ZStream[R, E, A] =
```
regiskuckaertz: @iravid Did I understand your suggestion correctly? ☝️

iravid: Yup! Although, I realize now that this is just `fromQueue(q).unTake`. Do you think this justifies an additional constructor?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2125,17 +2125,34 @@ object ZStream extends ZStreamPlatformSpecific {
...
-          ZIO.succeed
+    ZStream[R, E, A] {
+      ZManaged.reserve(
+        Reservation(
+          UIO(
+            queue.take.catchAllCause(
+              c => queue.isShutdown.flatMap(down => if (down && c.interrupted) Pull.end else Pull.halt(c))
+            )
+          ),
+          _ => queue.shutdown
```
iravid: I also suggest not shutting down the queue - there's `fromQueueWithShutdown` for that :-)

regiskuckaertz: Oh, I didn't see that. Fair enough 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: Nope! ok, undoing that one.

iravid:
Thanks @regiskuckaertz!

PR #1516, opened 2019-08-29
Address a variety of ZSink combinator bugs and tests

### LINKED ISSUE #1312: Wrong behavior of Sink combinators
`collectAll`

```scala
private def collectAllStepError = {
  // This test needs to be verified for correctness.
  val sink = stepErrorSink.collectAll
  unsafeRun(sinkIteration(sink, 1).map(_ must_=== Nil))
}
```

> @iravid Yeah I think the implementation for Sink#collectAllWith is surprising. It stores the error in the first element of the tuple, but never uses it in extract; and if subsequent errors occur in a step, the previous errors are discarded. We should probably change it to fail eagerly in step.

```scala
private def collectAllWhileFalsePredicate = {
  // This test needs to be verified for correctness.
  // I find this behavior to be surprising.
  val sink = ZSink.identity[Int].collectAllWhile[Int, Int](_ < 0)
  unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left(())))
  // Fails instead of returning empty list.
  // I would presume that sinkIteration(sink, 1).map(_ must_=== Nil) is the correct behavior.
}
```
> @iravid The implementation of collectAllWhile looks right - it tries to extract the underlying sink if an element that doesn't match the predicate is stepped.

The problem is with ZSink.identity, however; that one fails with Unit if an element never passed through the sink. Maybe test it with a different sink?

`untilOutput`

```scala
private def untilOutputHappyPath = {
  // This test needs to be verified for correctness.
  val sink = ZSink.identity[Int].untilOutput(_ % 2 == 0)
  unsafeRun(sinkIteration(sink, 1).map(_ must_=== 1))
}
```
> @iravid The logic in untilOutput's step looks backwards. should be `if (!Step.done(s))` ...

```scala
private def untilOutputFalsePredicate = {
  // This test needs to be verified for correctness.
  val sink = ZSink.identity[Int].untilOutput(_ % 2 != 0)
  unsafeRun(sinkIteration(sink, 1).map(_ must_=== 1))
}
```
> @iravid Yes, it's surprising too. Extracting that should probably result in an error (or a `None` which means the output should be `Option[B]`).

vasilmkd:
Resolves #1312. Finally.

Also improves test coverage.

@iravid Please take a look when you have time.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -171,15 +171,15 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
     z: S
   )(f: (S, B) => S)(implicit ev: A00 =:= A1): ZSink[R, E, A00, A1, S] =
     new ZSink[R, E, A00, A1, S] {
-      type State = (Option[E], S, self.State)
+      type State = (S, self.State)
 
-      val initial = self.initial.map(step => Step.leftMap(step)((None, z, _)))
+      val initial = self.initial.map(step => Step.leftMap(step)((z, _)))
 
-      def step(state: State, a: A1): ZIO[R, E, Step[State, A00]] =
+      def step(state: State, a: A1) =
         self
-          .step(state._3, a)
+          .step(state._2, a)
           .flatMap { step =>
-            if (Step.cont(step)) IO.succeed(Step.more((state._1, state._2, Step.state(step))))
+            if (Step.cont(step)) UIO.succeed(Step.more((state._1, Step.state(step))))
             else {
               val s  = Step.state(step)
               val as = Step.leftover(step)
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -188,17 +190,13 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
                 self.initial.flatMap { init =>
                   self
                     .stepChunk(Step.state(init), as.map(ev))
-                    .fold(
-                      e => Step.done((Some(e), f(state._2, b), Step.state(init)), Chunk.empty),
-                      s => Step.leftMap(s)((state._1, f(state._2, b), _))
-                    )
+                    .map(Step.leftMap(_)((f(state._1, b), _)))
```
iravid: Nice. I noticed that swallowed error too :-)

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -838,24 +867,28 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
    * Creates a sink that produces values until one verifies
    * the predicate `f`.
    */
-  final def untilOutput(f: B => Boolean): ZSink[R, E, A0, A, B] =
-    new ZSink[R, E, A0, A, B] {
-      type State = self.State
+  final def untilOutput(f: B => Boolean): ZSink[R, Option[E], A0, A, B] =
```
iravid: Could you clarify why `Option[E]` is needed?

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: It's now in line with `collectAllWhile`, so that's something. I'm happy.

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: Fixed after discussion in DM.

vasilmkd:
@iravid Ready for review.

PR #1517, opened 2019-08-29
ZIO Test: Compile Test Examples in CI

adamgfraser:
The examples of using ZIO Test weren't compiling after we renamed `Predicate` to `Assertion`. This fixes that and adds compiling the test examples to CI to prevent regressions in the future.

@edvmorango


adamgfraser:
I refactored this so it doesn't compile against Dotty. That will take a little more work. Should pass as soon as the CI is back online.

edvmorango:
@adamgfraser I merged  the `master` yesterday and fixed all the finished tests.

adamgfraser:
@edvmorango Awesome!

adamgfraser:
@edvmorango One other thing I noticed as I was getting the CI working on the examples is right now they are in the JS and JVM folders. Everything should be in the shared folder because the entire API that ZIO Test exposes is available on both the JVM and Scala.js, so we only need to do things once.

ghostdogpr:
@adamgfraser can you rebase?

adamgfraser:
@ghostdogpr All set.

PR #1520, opened 2019-08-29
Catch defects in effectAsyncM (could cause deadlocks)

neko-kai:
added test did not pass before the change


neko-kai:
@iravid 
Actually no, this is completely wrong wrt Cats laws – the example should deadlock since the callback is not called; failures in `asyncF` action should be dropped/logged but should not be used in continuation.
Current implementation creates a race condition wrt what completes first – the callback or the failure.

Was there a discussion on deviating from cats-effect here or did it just happen?

iravid:
Really. That sounds strange.

I'm mostly unfamiliar with the typeclass laws of  cats-effect, so can't comment intelligently on that.

neko-kai:
Ok, sorry! The change was introduced here https://github.com/zio/zio/pull/823 - and discussion has moved to https://github.com/typelevel/cats-effect/pull/615 & https://github.com/typelevel/cats-effect/pull/612

iravid:
Looks like I was asking the same questions there ;-)

PR #1521, opened 2019-08-29
Do we actually need to fork in effectAsyncM?

neko-kai:
To me it seems like the essence of the operation is single-threaded – we must post a continuation of the Fiber on EC as a result of invoked callback – which we do via Promise, but there's no reason to evaluate the provided effect on a new fiber – and indeed cats-effect does not do that.


jdegoes:
Assuming tests pass, looks good to me. The `fork` may be a `pre` `uninterruptibleMask` artifact.

neko-kai:
Outdated following discussion in https://github.com/typelevel/cats-effect/pull/615 – we should revert `effectAsyncM` back to `Nothing` error type to avoid a race between failure and success outputs

neko-kai:
Actually, we certainly can't drop the fork since this would deviate from behavior of cats-effect IO (this part not yet covered by laws unfortunately https://github.com/typelevel/cats-effect/pull/615)

PR #1523, opened 2019-08-29
Prototype Safe Semaphore

### LINKED ISSUE #1511: Add Semaphore#permits
It would be useful to have a method in Semaphore to get the maximum number of permits available. We have Semaphore#available but it returns only free permits. Semaphore#permits or Semaphore#permitsTotal should return capacity of the Semaphore.

P.S. Use case is a read write lock on top of semaphore where you acquire all permits to have an exclusive access.

adamgfraser:
Resolves #1511. We actually didn't keep track of the number of permits before because it didn't matter as long as we started out with a certain number and didn't leak them. I added a value to record the number of permits. Change required was very minimal.

Only other thing to think about is in the case of `TRef` I had to abandon it extending `AnyVal` because now it has two parameters. We could try go put the value in the `TRef` instead or put them both in a case class but I think we lose as much as we gain that way.


adamgfraser:
This could use some feedback if we want to do anything on it for the 1.0 release.

core-tests/jvm/src/test/scala/zio/SemaphoreSpec.scala
```diff
@@ -3,58 +3,54 @@
...
+      `withPermitManaged` does not leak permits upon cancellation $e8
     """
 
   def e1 = {
     val n = 20L
     unsafeRun(for {
       semaphore <- Semaphore.make(n)
-      available <- IO.foreach((0L until n).toList)(_ => semaphore.acquire) *> semaphore.available
-    } yield available must_=== 0)
+      available <- IO.foreach((0L until n).toList)(_ => semaphore.withPermit(semaphore.available))
```
regiskuckaertz: Makes me want to add a `semaphore.withAllPermits`

regiskuckaertz: ... nothing to do with the test though, just thought that one might want a monopoly over all permits.

docs/datatypes/semaphore.md
```diff
@@ -3,14 +3,14 @@ id: datatypes_semaphore
 title:  "Semaphore"
 ---
 
-A `Semaphore` datatype which allows synchronization between fibers with `acquire` and `release` operations.
+A `Semaphore` datatype which allows synchronization between fibers with the `withPermit` operation, which safely acquires and releases a permit.
```
regiskuckaertz: Note that it is still possible to trick the system by asking for a negative number of permits.

regiskuckaertz:
You have my vote, I'd just like to clarify the behaviour when passing a negative number (right now `prepare` deals with the case where `n == 0`)

core-tests/jvm/src/test/scala/zio/SemaphoreSpec.scala
adamgfraser: Yes, if we went with this approach we could definitely maintain a `val` with the total number of permits and expose a `withAllPermits` method.

docs/datatypes/semaphore.md
adamgfraser: Wouldn't `assertNonNegative` cause the fiber to die at that point?

regiskuckaertz: Yes, on release, which seems wrong 🤔 It's like catching a bad guy when it's too late 😆

adamgfraser: I see. Yes we could definitely fix that.

adamgfraser:
@regiskuckaertz I added a fix for this.

core-tests/jvm/src/test/scala/zio/RTSSpec.scala
```diff
@@ -1215,10 +1215,10 @@ class RTSSpec(implicit ee: ExecutionEnv) extends TestRuntime with org.specs2.mat
 
   def testRaceBothInterruptsLoser =
     unsafeRun(for {
-      s      <- Semaphore.make(0L)
+      latch  <- Promise.make[Nothing, Unit]
```
regiskuckaertz: I can't believe that degenerate behaviour was actually used in a test as a poor man's promise 😆

adamgfraser: I know. I want to be humble because this is the behavior of Semaphore in Java and other implementations but I remember looking at those tests being surprised that they passed at all, much less verifying expected results.

adamgfraser:
@regiskuckaertz Sorry, don't we need it in `prepare`? Agree we can move it out of `releaseN`.

adamgfraser:
@jdegoes When you have the chance could I get some feedback on whether you want to move forward with this? If so I think the next steps would be adding a couple of convenience methods on `Semaphore` (`maxPermits` and `withAllPermits`) and deciding how, if at all, we want to change `TSemaphore` based on this.

core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -34,26 +34,6 @@ import scala.collection.immutable.{ Queue => IQueue }
...
-   */
-  final def acquire: UIO[Unit] = acquireN(1)
-
-  /**
-   * Acquires a specified number of permits.
-   *
-   * If the specified number of permits are not available, the fiber invoking
-   * this method will be suspended until the permits are available.
-   *
-   * Ported from @mpilquist work in Cats Effect (https://github.com/typelevel/cats-effect/pull/403)
```
jdegoes: We should preserve this attribution comment in any case.

core/shared/src/main/scala/zio/Semaphore.scala
```diff
@@ -62,39 +42,6 @@ final class Semaphore private (private val state: Ref[State]) extends Serializab
     case Right(n) => n
   }
 
-  /**
-   * Releases a single permit.
-   */
-  final def release: UIO[Unit] = releaseN(1)
```
jdegoes: Because code uses these functions, we should at least deprecate them until 1.0.

adamgfraser:
I incorporated your comments. The changes are pretty small at this point, just deprecating the methods on `Semaphore` and updating the tests to note use those methods, but good to deprecate them now if we are going to take them out. This is ready for another review.

adamgfraser:
@jdegoes Sorry could you approve one more time? Had to resolve a merge conflict.

adamgfraser:
@regiskuckaertz Thanks!

PR #1530, opened 2019-08-30
ZIO Test: Improve Ergonomics of Property Based Testing

### LINKED ISSUE #1527: Reduce boilerplate in Gen-based tests
Currently:
```scala
def check[R, A](rv: Gen[R, A])(assertion: Assertion[A]): ZTest[R, Nothing]
```

we change it to:

```scala
def check[R, A](rv: Gen[R, A])(test: (=> A) => TestResult): ZTest[R, Nothing]
```

This will allow test suites to be written in a property-based testing style, e.g.:

```scala
testM("Chunk.apply") {
  val chunkWithLength: Gen[Random, (Chunk[Int], Int)] = ???
  check(chunkWithLength) { case (chunk, i) => 
    assert(chunk.apply(i), equalTo(chunk.toSeq.apply(i)))
  }
}
```

In the same fashion, add `M`-suffixed variants:

```scala
def checkM[R, A](rv: Gen[R, A])(test: (=> A) => UIO[TestResult]): ZTest[R, Nothing]
```

### LINKED ISSUE #1528: ZIO Test: Add OverLoaded Check Methods for Multiple Generators
As suggested by @iravid it would be nice if we could write `check` statements for property based testing with multiple generators. This can already be done today by using `<*>` to tuple the generators but this requires the user to do the tupling and then write a match expression for every property they want to test. We can write some overloaded methods to do this automatically and provide a more pleasant experience for the user.

I will take the ticket.

adamgfraser:
Changes `check` methods to accept a function that produces a test result instead of an assertion. Adds `checkM`, `checkAllM`, and `checkSomeM` methods for testing effectual properties. Adds overloaded methods that accept multiple generators. Resolves #1527. Resolves #1528.

### DIFF
--- a/examples/jvm/src/test/scala/zio/examples/test/ExampleSpec.scala
+++ b/examples/jvm/src/test/scala/zio/examples/test/ExampleSpec.scala
@@ -1,14 +1,14 @@
 package zio.examples.test
 
-import zio.test.{DefaultRunnableSpec, Predicate, assert, suite, test}
+import zio.test.{ Assertion, DefaultRunnableSpec, assert, suite, test}
 
 object ExampleSpec extends DefaultRunnableSpec(
   suite("some suite") (
     test("failing test") {
-      assert(1, Predicate.equalTo(2))
+      assert(1, Assertion.equalTo(2))
     },
     test("passing test") {
-      assert(1, Predicate.equalTo(1))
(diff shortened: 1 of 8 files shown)


adamgfraser:
Good point. Will change now.

Issue #895, opened 2019-05-24
Type inference of the E type breaks down when using ZIO.retry in conjunction with ZIO.provide


CatsDogsAndOther:
import java.util.concurrent.TimeUnit

import scalaz.zio.clock.Clock
import scalaz.zio.duration.Duration
import scalaz.zio.{IO, Schedule, ZIO}

import scala.concurrent.Future

class Example {

  // what I have to do
  def fetchRemoteData1(): IO[RemoteClientException, RemoteData] = {
    val keys: List[String] = List("key1", "key2")

    val io: ZIO[Clock, RemoteClientException, RemoteData] =
      ZIO
        .fromFuture{ implicit ec => remoteCall(keys) }
        .timed
        .mapError { e: Throwable => RemoteCallFailedException(e) }
        .map { case (d, x: RemoteData) => x }
        .timeoutFail(RemoteCallTimeoutException())(Duration(100, TimeUnit.MILLISECONDS))
        .retry(Schedule.recurs(2))

    io.provide(Clock.Live)
  }

  // what I have want to do
  def fetchRemoteData2(): IO[RemoteClientException, RemoteData] = {
    val keys: List[String] = List("key1", "key2")

    ZIO
      .fromFuture{ implicit ec => remoteCall(keys) }
      .timed
      .mapError { e: Throwable => RemoteCallFailedException(e) }
      .map { case (d, x: RemoteData) => x }
      .timeoutFail(RemoteCallTimeoutException())(Duration(100, TimeUnit.MILLISECONDS))
      .retry(Schedule.recurs(2))
      .provide(Clock.Live)
  }

  // what also breaks
  def fetchRemoteData3(): IO[RemoteClientException, RemoteData] = {
    val keys: List[String] = List("key1", "key2")

    val io =
      ZIO
        .fromFuture{ implicit ec => remoteCall(keys) }
        .timed
        .mapError { e: Throwable => RemoteCallFailedException(e) }
        .map { case (d, x: RemoteData) => x }
        .timeoutFail(RemoteCallTimeoutException())(Duration(100, TimeUnit.MILLISECONDS))
        .retry(Schedule.recurs(2))

    io.provide(Clock.Live)
  }

  // the retry seems to contribute to the issue as this compiles fine
  def fetchRemoteData4(): IO[RemoteClientException, RemoteData] = {
    val keys: List[String] = List("key1", "key2")

    ZIO
      .fromFuture{ implicit ec => remoteCall(keys) }
      .timed
[...]

PR #1535, opened 2019-08-31
Fix Type Inference Issue with ZIO#retry

### LINKED ISSUE #1502: Loss of Type Inference
This toy example highlights a type inference issue.

```scala
import scalaz.{-\/, \/, \/-}
import zio.clock.Clock
import zio.duration._
import zio.{Schedule, ZIO}

import scala.concurrent._

case class Error(message: String) extends Exception

case class Failure(message: String)
case class Success[O](content: O)

def foo[O](v: O): ZIO[Any with Clock, Error, \/[Failure, Success[O]]] =
    ZIO
      .fromFuture { implicit ec =>
        Future.successful(v)
      }
      .foldM(
        error => ZIO.fail(Error("Some error")),
        ok => {
          ZIO.succeed(\/-(Success(ok)))
        }
      )
      .retry(Schedule.spaced(2.seconds) && Schedule.recurs(1))
      .catchAll(
        (e:Error) => ZIO.succeed(-\/(Failure(e.message)))
      )
```

Compiling this code leads to the following compiler error:

```scala
Error:(26, 19) type mismatch;
found   : Error => zio.UIO[scalaz.-\/[Failure]] (which expands to)  Error => zio.ZIO[Any,Nothing,scalaz.-\/[Failure]]
required: Any => zio.ZIO[Any with zio.clock.Clock,Error,Failure \/ Success[O]](e:Error) => ZIO.succeed(-\/(Failure(e.message)))
```
After talking to @ghostdogpr on the ZIO gitter channel, it was suggested to add type information to the retry method which I did as follows:

```scala
import scalaz.{-\/, \/, \/-}
import zio.clock.Clock
import zio.duration._
import zio.{Schedule, ZIO}

import scala.concurrent._

case class Error(message: String) extends Exception

case class Failure(message: String)
case class Success[O](content: O)

def foo[O](v: O): ZIO[Any with Clock, Error, Either[Failure, Success[O]]] =
    ZIO
      .fromFuture { implicit ec =>
        Future.successful(v)
      }
      .foldM(
        error => ZIO.fail(Error("Some error")),
        ok => {
          ZIO.succeed(Right(Success(ok)))
        }
      )
      .retry[Any with Clock, Error, (Int, Int)](Schedule.spaced(2.seconds) && Schedule.recurs(1))
      .catchAll(
        error => ZIO.succeed(Left(Failure(error.message)))
      )
```

This now compiles. This type inference issue only began to exhibit itself when I introduced a type parameter on the Success type i.e Success[O]. Before that change, the code compiled without any issues.

Creating this issue to see if something can be done to not specify the type parameters on the retry method.

adamgfraser:
Resolves #1502.

@jdegoes The issue actually wasn't with the `R` parameter but the `E` parameter.

The previous type signature of `retry` was wrong.

```scala
final def retry[R1 <: R, E1 >: E, S](policy: ZSchedule[R1, E1, S]): ZIO[R1 with Clock, E1, A]
```

Notice that we are widening `E` to the `E1` type of the `Schedule`, as if the `Schedule` produced errors so the resulting error type needs to be the union of the two. But the `Schedule` doesn't produce errors, it consumes them! Its `E1` type parameter needs to be a superset of `E` so it can handle all the errors that the effect produces, but any actual error that remains is guaranteed to be of type `E`. The correct type signature should be:

```scala
final def retry[R1 <: R, E1 >: E, S](policy: ZSchedule[R1, E1, S]): ZIO[R1 with Clock, E, A]
```

I made the appropriate changes to `retry`, `retryOrElse`, and `retryOrElseEither` and added a test to prevent regression. Interestingly, the same issue did not exist with `repeat` and related combinators.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/RetrySpec.scala
+++ b/core-tests/shared/src/test/scala/zio/RetrySpec.scala
@@ -1,5 +1,8 @@
 package zio
 
+import scala.concurrent.Future
+
+import zio.clock.Clock
 import zio.duration._
 import zio.random._
 
@@ -30,6 +33,7 @@ class RetrySpec extends BaseCrossPlatformSpec {
      retry exactly one time for `once` when second time succeeds - retryOrElse0 $retryOrElseEitherSucceed
   Retry a failed action 2 times and call `ensuring` should
      run the specified finalizer as soon as the schedule is complete $ensuring
+  Retry type parameters should infer correctly $retryTypeInference
   """
 
   def retryCollect[R, E, A, E1 >: E, S](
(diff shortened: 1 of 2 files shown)


jdegoes:
@adamgfraser Awesome work on this! It has been a big pain point for a while, and no one noticed the cause until now.

ghostdogpr:
Does that fix also https://github.com/zio/zio/issues/895 by any chance?

PR #1536, opened 2019-08-31
Migrate PromiseSpec, RefSpec and RefMSpec to ZIO Test

ghostdogpr:
Also added a new assertion `isInterrupted`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
@@ -1,165 +1,95 @@
 package zio
 
-class PromiseSpec extends BaseCrossPlatformSpec {
+import zio.duration._
+import zio.test._
+import zio.test.Assertion._
 
-  def is = "PromiseSpec".title ^ s2"""
-        Make a promise and retrieve its value correctly after complete it with:
-          `complete` to complete that promise with a specified value. $e1
-          `done` to complete that promise with a completed result.    $e2
-
-        Make a promise and retrieve its fail value after complete it with:
-          `error` to fail that promise with a specified error.  $e3
-          `done` to complete that promise with a failed result. $e4
-
-        Given a completed promise `done` returns false and get should return the first completed value. $e5
-
-        Make a promise and retrieve its Throwable value after interruption calling:
-          `done` to complete that promise with a terminated result.              $e6
-          `interrupt` and interrupt all other fibers.                            $e7
-
-        poll retrieves the final status immediately
-          it `fails' with Unit ` if the promise is not done yet.  $e8
-          Otherwise, it returns the `Exit`, whether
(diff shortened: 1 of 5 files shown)


core-tests/shared/src/test/scala/zio/PromiseSpec.scala
```diff
@@ -1,165 +1,95 @@
...
+        },
+        testM("isDone when a promise is failed") {
+          for {
+            p <- Promise.make[String, Int]
+            _ <- p.fail("failure")
+            d <- p.isDone
+          } yield assert(d, isTrue)
+        }
+      ),
+      timeout = TimeoutStrategy.Error(60.seconds)
```
jdegoes: Stylistically, I would do `timeout(60.seconds) { ... suite ... }`

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -244,6 +244,17 @@ object Assertion {
       else AssertResult.failure(())
     }
 
+  /**
+   * Makes a new assertion that requires an exit value to be interrupted.
+   */
+  final def isInterrupted: Assertion[Exit[Any, Any]] =
```
jdegoes: 👍

core-tests/shared/src/test/scala/zio/PromiseSpec.scala
ghostdogpr: @jdegoes @adamgfraser 
If we move it to `Runner`, that means we couldn't use `DefaultTestRunner` and `DefaultRunnableSpec` if we want to have a different one from the default. We would have to make a custom `Runner` and then a custom `RunnableSpec`, is that right?

adamgfraser: @jdegoes @ghostdogpr What if we made `TimeoutStrategy` a subtype of `TestAspect`? `TestAspect.timeout` is already isomorphic with `TestStrategy.Error`. `TestStrategy.Ignore` is a no-op, so that would be equivalent to `TestAspect.identity`, which we don't have but could easily add. So the only thing would be whether we could implement `Warn` as a `TestAspect`. I remember when I worked on the original PR I thought we couldn't for some reason but I think maybe I reworked it after that because the type signatures look the same. 

That would go a long way to unify things because then everything would be a `TestAspect` so at worst there are two different ways to add a `TestAspect` versus having this `TestAspect.timeout` and then `TimeoutStrategy` that are completely different things. I'm a little concerned with moving it to `Runner` because I think that makes it more difficult for us to have a default timeout policy that is also easily override by a user who is just extending `DefaultRunnableSpec` But maybe if we do that unification it addresses your concerns?

ghostdogpr:
Let’s merge this one and continue the timeout discussion in #1543 ?

PR #1539, opened 2019-08-31
ZIO Test: Support Platform Specific Test Configuration

adamgfraser:
We have been experiencing slow builds on ScalaJS due to the ZIO Test test suite. On the JVM the full test suite runs in about 30 seconds, we have a timeout of 60 seconds for the entire test suite, and to my knowledge this has never caused flakiness. In contrast, on ScalaJS we are observing running the test suite taking 4 to 5 minutes.

It appears that there are several root causes for this:

**1. Possible Excessive Parallelism** - We use `Future` as our asynchronous effect type for tests since we don't want to rely on ZIO while testing ZIO and have been using `sequence` to combine all the tests together. `Future.sequence` runs effects in parallel, so we may have had a large number of tests running in parallel, causing contention that was handled poorly in the ScalaJS single threaded environment. This PR addresses that by adding an `Async` type that provides a thin, purely functional wrapper around `Future` so that we can properly sequence effects. This appears to help in my local tests though I am interested to see what impact it has on CI build times.

**2. Poor Performance of Random Number Generation in ScalaJS** - It appears that random number generation in Scala.js is substantially slower than on the JVM. In benchmarking on my local machine random number generation takes approximately three times as long on ScalaJS as on the JVM. Note that this has nothing to do with `MockRandom`, this is just performance of `scala.util.Random` from the standard library and Scala.js. The parts of the test suite that take the longest all rely heavily on random number generation, particularly `GenSpec`, so I think this will explain most of the performance difference after the parallelism issue is fixed.

**3. Potential ZIO Performance Differences Across Platforms** - There may be differences in ZIO performance across platforms. For example, the snippet below ran in about 4 seconds for me on the JVM but took 36 seconds on ScalaJS. The random number generation issue is definitely a part of that but there may also be other cross platform differences. I haven't seen any comparison of ZIO performance across platforms so I don't have a sense of what we should expect here and how things like whether you do full optimization in ScalaJS impact that.

**4. Some Tests Take Disproportionate Time** - The tests that take the longest verify that fixed size collections shrink appropriately (all elements of the collection shrink to their "smallest" value). We currently do this on size 10 lists, vectors, and strings and repeat each test 100 times to verify non-flakiness. We could definitely cut this down though I do think these are important tests and reducing the number of tests is to a certain extent a one time improvement that doesn't address the root causes, especially since the time for the entire test suite is very reasonable on the JVM.

Let's see how much we get out of these changes and then we can make further adjustments if necessary.

```scala
def benchmark = 
  unsafeRunToFuture {
    import zio._
    val io = ZIO.collectAll(List.fill(100000)(random.nextInt(100))).repeat(Schedule.recurs(100))
    io.timed.tap { case (duration, _) => ZIO.effectTotal(println(duration.toMillis / 1000.0)) }
  }
```


ghostdogpr:
Looking at the test runs, it seems the first change (Async) was not that impactful (still around 4min, sometimes 5min), but the second one (nonflakiness) improved it a lot (~1min30).

adamgfraser:
That was my conclusion as well. I am going to try to get the `main` method of `TestMain` to accept a command line `size` parameter so we can pass a different `size` in for `testJVM` and `testJS`. Then the test suites can use that parameter how they want, so we can still run the additional tests on the JVM. I think that will strike a nice balance between getting the times down and maintaining coverage. And if contributors are making changes and testing interactively in the `testJVM` project they will still get strong assurances that everything is working correctly.

adamgfraser:
Okay, well my plan to use a command line argument kind of blew up in my face when I learned that ScalaJS doesn't accept command line arguments but I developed a workaround. This is ready to go from my perspective.

jdegoes:
So the reason why `Async` cannot help is because (among other things), Javascript only has 1 thread, so `sequence` cannot introduce "too much concurrency". Even being lazy about when these futures are started (construction versus "unsafeRun") is not going to impact performance. What will help is fewer futures: i.e. doing more inside each future, so you need fewer flat maps. It would also help to use ZIO instead of Future, although that's not possible / viable due to bootstrapping.

All monadic-heavy code will be quite a bit slower on Javascript. A bit surprising how much, though, and possibly worth profiling and doing a deep dive, but not now, I'd say (too little ROI).

I think we could benefit from `isJs` and `isJVM` being inside ZIO Test directly, btw.

test/js/src/test/scala/zio/test/TestPlatform.scala
```diff
@@ -2,10 +2,14 @@ package zio.test
...
+private[test] object TestPlatform {
 
   def await(f: Future[Boolean])(implicit ec: ExecutionContext): Unit =
     f.foreach(_ => ())
+
+  def fail(): Unit = System.exit(-1)
+
+  val isJS = true
+
+  val isJVM = false
```
jdegoes: Would like to see these two helpers (`isJS` and `isJVM`) inside `ZIOTest`, so we could make `aspects` and a combinator for platform-specific tests.

```scala
test("Javascript-only") { ... } @@ js
```

jdegoes:
One other thing: I think we can disable `nonFlaky` combinator for JS. It's enough to do that on one platform.

adamgfraser:
@jdegoes I incorporated your feedback, including the test aspects, here. Resolves #1541.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -204,6 +215,28 @@ object TestAspect {
...
+  def jvm[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS](
+    that: TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS]
+  ): TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS] =
+    if (TestPlatform.isJVM) that else identity
+
+  /**
+   * An aspect that only runs tests on the JVM.
+   */
+  val jvmOnly: TestAspectPoly =
+    if (TestPlatform.isJVM) identity else ignore
```
jdegoes: Beautiful

jdegoes:
@adamgfraser Looks great!

ghostdogpr:
@adamgfraser I fixed a typo to re-trigger the build, the CI is back. I'll merge if the tests pass.

ghostdogpr:
There's some issue downloading JDK from CircleCI, will re-try again later :/

adamgfraser:
Okay.

PR #1531, opened 2019-08-30
Upgrade Dotty Version

adamgfraser:

PR #1540, opened 2019-09-01
Upgrade Dotty Version

adamgfraser:
Another release of Dotty just hit, so upgrades to that and fixes an error I had made in #1531 where I did not update the Dotty version in the CI configuration. Thanks to @smarter for pointing both out!


adamgfraser:
@regiskuckaertz Sorry I didn't close the loop on this. I filed a bug report with the Dotty team related to this and they made some changes a couple of days ago to fix it so just getting a chance to test those changes now.

adamgfraser:
Okay, so this is working now. Only issue is there was a bug in the release so we are compiling against a nightly after the bug got fixed. Are we okay with doing that or do we want to hold off until the next release in another five weeks?

ghostdogpr:
Since we are not releasing for dotty yet, I'd say it's fine?

adamgfraser:
I think so.

PR #1551, opened 2019-09-02
ZIO Test: Add Structure to Assertion Display

### LINKED ISSUE #1542: ZIO Test: Add structure to `Assertion` display
Currently, `Assertion` has a `render` method that converts the assertion into a string. This is useful but a string has no structure at all, and with more structure comes greater potential (nicely formatted output, different renderings for different types, etc.).

This ticket is to change the `Assertion` type so that `Assertion#render` becomes a function that returns this more structured representation, while `Assertion#toString` becomes a function from this more structured representation, to the current string representation.

For the more structured representation, we can place an ADT similar to the following inside the `Assertion` companion object:

```scala
final case class Render(name: String, paramsLists: List[List[RenderParam]])
sealed trait RenderParam
object RenderParam {
  final case class Assertion[A](assertion: Assertion[A]) extends RenderArg
  final case class Value(value: Any) extends RenderArg
}
```

The main goals of this ADT are to capture both the name of an assertion, as it will be rendered to strings and as it exists in Scala, as well as the parameters to the assertion combinator (there may be multiple parameter lists, due to curried syntax!), both assertion parameters, and catch-all value parameters (later this could be improved, but for now, "catch all" is fine).

Then the various `Assertion.assertion*` methods can be modified, and the combinators adjusted, to produce the structured form, using helpers to simplify construction. 

We currently define assertions like this:

```scala
Assertion.assertion("hasField(\"" + name + "\"" + s", _.${name}, ${assertion})") { ... }
```

After the change, it will be something like:

```scala
Assertion.assertion("hasField")(param(name), param(assertion)) { ... }
```

The latter representation captures a LOT more structure, including enough for pretty-printing the output (wrapping / tree-like rendering), and treating string values specially (quoting them so they render as valid Scala strings; since currently `String#toString` will provide a rendering that is not valid Scala code).

adamgfraser:
Resolves #1542.

I made `Render` an ADT and added case classes for `Function` and `Infix` so we can pretty print assertion combinators like `&&` and `||` in infix form.


ghostdogpr:
You mind if I apply the suggestion and merge this?

adamgfraser:
Maybe give @jdegoes another day to weigh in since he raised the issue and otherwise we just go ahead? In the meantime I can fix that typo.

ghostdogpr:
Sure! I wanted to reduce the open PRs but let’s give it a little more time.

adamgfraser:
Yeah I hear you. You have been really good about that.

jdegoes:
On it in the AM!

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -74,22 +76,96 @@ class Assertion[-A] private (render: String, val run: (=> A) => AssertResult[Eit
...
+  /**
+   * `Render` captures both the name of an assertion as well as the parameters
+   * to the assertion combinator for pretty-printing.
+   */
+  sealed trait Render {
+    override final def toString: String = this match {
+      case Render.Function(name, paramLists) =>
+        name + paramLists.map(_.mkString("(", ", ", ")")).mkString
+      case Render.Infix(left, op, right) =>
+        "(" + left + " " + op + " " + right + ")"
```
jdegoes: This is good. Eventually we could embed precedence here to make rendering a bit nicer.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -74,22 +76,96 @@ class Assertion[-A] private (render: String, val run: (=> A) => AssertResult[Eit
...
+    /**
+     * Construct a `RenderParam` from an `Assertion`.
+     */
+    final def param[A](assertion: Assertion[A]): RenderParam =
+      RenderParam.Assertion(assertion)
+
+    /**
+     * Construct a `RenderParam` from a value.
+     */
+    final def param[A](value: Any): RenderParam =
```
jdegoes: `value` here should have type `A`.

adamgfraser:
Awesome! Good catch on that. Will make the change right now.

Issue #1549, opened 2019-09-02
Only the 2.11 builds are required for releasing


iravid:
The release workflow should require all builds to pass before releasing.

ghostdogpr:
Right, just noticed this as well.

It's pretty easy, in `.circleci/config.yml`:
```
      - release_211:
          context: Sonatype
          requires:
            - test_211_jdk8_jvm
            - test_211_jdk11_jvm
            - test_211_jdk8_js
            - test_211_jdk11_js
```
All the test steps should be added here.

PR #1553, opened 2019-09-02
Require all builds to pass for release

denisgarci:
Fixes https://github.com/zio/zio/issues/1549. 
@ghostdogpr the tests could also be required for each release (`release_212` etc) but that would make it possible to release for 2.11 even if 2.12 tests don't pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -416,6 +416,14 @@ workflows:
             - test_211_jdk11_jvm
             - test_211_jdk8_js
             - test_211_jdk11_js
+            - test_212_jdk8_jvm
+            - test_212_jdk11_jvm
+            - test_212_jdk8_js
+            - test_212_jdk11_js
+            - test_213_jdk8_jvm
+            - test_213_jdk11_jvm
+            - test_213_jdk8_js
+            - test_213_jdk11_js
           filters:
             branches:
               ignore: /.*/


.circleci/config.yml
```diff
@@ -417,6 +417,14 @@ workflows:
             - test_211_jdk11_jvm
             - test_211_jdk8_js
             - test_211_jdk11_js
+            - test_212_jdk8_jvm
```
softinio: - I would prefer `test_212` tests to be moved to the `require` section of `release_212` please

.circleci/config.yml
```diff
@@ -417,6 +417,14 @@ workflows:
...
+            - test_212_jdk11_jvm
+            - test_212_jdk8_js
+            - test_212_jdk11_js
+            - test_213_jdk8_jvm
```
softinio: Similarly:

- I would prefer `test_213` to be moved to the `require` section of `release_213` please

.circleci/config.yml
softinio: here https://github.com/zio/zio/blob/master/.circleci/config.yml#L437

.circleci/config.yml
softinio: here https://github.com/zio/zio/blob/master/.circleci/config.yml#L437

.circleci/config.yml
softinio: here https://github.com/zio/zio/blob/master/.circleci/config.yml#L428

ghostdogpr: Isn’t it better to stop releasing if any test fail? If we move it as you say, we might release 2.11 and not 2.12.

denisgarci: Yeah not sure which strategy you prefer. I'll let you decide.

ghostdogpr: It’s a mess when we release 2.11 and not 2.12 (happened a couple time in the past) so I’d really prefer checking all of them before releasing 2.11. 
@softinio can you confirm it’s ok with you?

softinio: Even though it infringes my ocd 🤣I do agree with your point which I had not thought of. Thank you both. Approving.

denisgarci:
Done

ghostdogpr:
Thanks!

PR #1555, opened 2019-09-02
ZIO Test: Make Timeout Strategies Test Aspects

### LINKED ISSUE #1543: ZIO Test: Replace `timeout` parameter to runnable spec with aspects
To ensure one clear way of doing thing, encode `TimeoutStrategy` cases as test aspects.

(This ticket lacks details, just to track ongoing discussions.)

/cc @adamgfraser @ghostdogpr

adamgfraser:
Resolves #1543.

I made `TimeoutStrategy.Warn` a test aspect and put it in `TimeoutVariants.scala` that `TestAspect.scala` extends, since there are a lot of implementation specific details for printing the timeout warnings that I didn't want to clutter up `TestAspect` with. `TimeoutStrategy.Error` and `TimeoutStrategy.Ignore` go away since they are redundant with `TestAspect.timeout` and `TestAspect.identity`.

I changed the second parameter in `DefaultRunnableSpec` to be `defaultTestAspects` and take a list of test aspects to apply to the spec, with a default parameter of `timeoutWarning(60.seconds)`. So it becomes a little more general if in the future there are other test aspects we want to have as overridable defaults.

We were still figuring out the details of the solution we wanted here so happy to take it in another direction but thought it would be good if we had a starting point.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
@@ -1,11 +1,10 @@
 package zio
 
-import zio.duration._
 import zio.test._
 import zio.test.Assertion._
 
 object PromiseSpec
-    extends DefaultRunnableSpec(
+    extends ZIOSpec(
       suite("PromiseSpec")(
         testM("complete a promise using succeed") {
           for {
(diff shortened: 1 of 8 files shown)


ghostdogpr:
Very nice to be able to pass any aspect rather than just `timeout` 👍

PR #1556, opened 2019-09-03
Keep only one jdk11 build

### LINKED ISSUE #1545: Reduce build times and costs
Our overall CI build times have become pretty large, and with the sustained activity on ZIO from many contributors (which is great! 🎉), the costs are getting important too.

Each individual build is not that long, but we have 12 combinations of Scala versions (2.11, 2.12, 2.13), JDK (8, 11) and platforms (JVM, JS). We expect to add Dotty (#1022) and Scala Native (#639) soon, which is going to increase the number of combinations exponentially. The question is: do we really need all combinations?

I had a look at a number of open-source Scala projects and most of them only build and test with jdk8. In the rare cases where they do it with jdk11, they only do for a single version (usually Scala 2.12 with JVM). Examples [here](https://travis-ci.org/scalaz/scalaz) and [there](https://travis-ci.org/http4s/http4s).

How about testing JDK11 only with Scala 2.12 + JVM? We would go from:
```
- test_211_jdk8_jvm:
- test_212_jdk8_jvm:
- test_213_jdk8_jvm:
- test_211_jdk11_jvm:
- test_212_jdk11_jvm:
- test_213_jdk11_jvm:
- test_211_jdk8_js:
- test_212_jdk8_js:
- test_213_jdk8_js:
- test_211_jdk11_js:
- test_212_jdk11_js:
- test_213_jdk11_js:
```
to
```
- test_211_jdk8_jvm:
- test_212_jdk8_jvm:
- test_213_jdk8_jvm:
- test_212_jdk11_jvm:
- test_211_jdk8_js:
- test_212_jdk8_js:
- test_213_jdk8_js:
```

Any other ideas?

ghostdogpr:
Fixes #1545

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,32 +208,11 @@ jobs:
       - <<: *scala_dotty
       - <<: *jdk_8
 
-  test_211_jdk11_jvm:
-    <<: *testJVM
-    <<: *machine_ubuntu
-    environment:
-      - <<: *scala_211
-      - <<: *jdk_11
-
   test_212_jdk11_jvm:
     <<: *testJVM
     <<: *machine_ubuntu
     environment:
       - <<: *scala_212
-      - <<: *jdk_11
-
-  test_213_jdk11_jvm:
-    <<: *testJVM
-    <<: *machine_ubuntu
-    environment:
-      - <<: *scala_213
-      - <<: *jdk_11
-
-  test_dotty_jdk11_jvm:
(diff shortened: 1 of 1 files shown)


vasilmkd:
It seems that they are still marked as `expected` and `Required`.

ghostdogpr:
@vasilmkd yeah it's like that because they're defined on master. We can ignore them and once this is merged to master, they will disappear.

PR #1564, opened 2019-09-03
[Proposal] Make suspending effects a capability

sideeffffect:
I've been always been bothered in ZIO, that side-effecting code can be present (suspended by `effect`, `effectAsync` & friends) in all places where the `ZIO` type is used

to avoid that I would usually use an abstract `F` (tagless final style), but that has other drawbacks compared to using `ZIO` directly

maybe the best of both worlds is if suspending effects were a capability that was tracked by the environment, just like any other capability, like working with time using `Clock`, etc

(I will write a better pitch for this proposal tomorrow, because now it's very late, but I wanted to at least open the PR)

this is still a rough sketch to start a discussion, but it should compile and pass tests


NeQuissimus:
I think I understand what you're trying to express but I'm not sure. Is the purpose to signal "This ZIO side-effects"?
If so, I would not call the marker `Effect`.

I am trying to wrap my head around why this is needed. 
If something uses the console, we have a `Console` in the environment, hence the effect is no longer side-effecting.
Where would you have an `Effect` that could not be better represented by something else?

jdegoes:
I implemented this in an early prototype. @wi101 and I thought it added too much noise for the benefit. I think we could consider this in 2.0 but we'd need more evidence that it pays for its cost.

Also due to the pain of effect elimination right now, the module pattern might require re-tinkering.

iravid:
Generally 👎 on this for 1.0.0.

sideeffffect:
I'm sorry guys I'm responding just now, it has been a busy time for me, so once I realized we can't make deep changes to ZIO before 1.x release, I let this go.

But I would like to make a case for treating effects like capability. I hope we could talk about it either tonight or tomorrow -- I here with you at the Warsaw hackathon :tada: :smile_cat: 

Normal ZIO style

```scala
case class CustomerOrder(data: String)

trait CustomerService {
  def collectOrders: Task[List[CustomerOrder]]
}

case class SupplierRequest(data: String)

trait SupplierService {
  def createRequest(supplierRequest: SupplierRequest): Task[Int]
}

object CustomerService {
  val make = new CustomerService {
    override val collectOrders: Task[List[CustomerOrder]] = {
      ZIO.effect(List(CustomerOrder("a"), CustomerOrder("abc"), CustomerOrder("qwer")))
    }
  }
}

object SupplierService {
  val make = new SupplierService {
    override def createRequest(supplierRequest: SupplierRequest): Task[Int] =
      ZIO.effect(supplierRequest.data.length)
  }
}

trait OrderProcessor {
  def processOrders: Task[List[Int]]
}

object OrderProcessor {
  def make(customerService: CustomerService, supplierService: SupplierService): OrderProcessor = new OrderProcessor {
    override def processOrders: Task[List[Int]] =
      for {
        orders <- customerService.collectOrders
        requestResponses <- ZIO.traverse(orders) { order =>
                             supplierService.createRequest(SupplierRequest(order.data))
                           }
        // can do that here! :(
        _ <- ZIO.effect(println("launch nukes"))
      } yield requestResponses
  }
}
```

Tagless final like
```scala
case class CustomerOrder(data: String)

trait CustomerService[F[_], C[_]] {
  def collectOrders: F[C[CustomerOrder]]
}

case class SupplierRequest(data: String)

trait SupplierService[F[_]] {
  def createRequest(supplierRequest: SupplierRequest): F[Int]
}

object CustomerService {
  val make: CustomerService[Task, List] = new CustomerService[Task, List] {
    override val collectOrders: Task[List[CustomerOrder]] = {
      Task.succeed(List(CustomerOrder("a"), CustomerOrder("abc"), CustomerOrder("qwer")))
    }
  }
}

object SupplierService {
  val make: SupplierService[Task] = new SupplierService[Task] {
    override def createRequest(supplierRequest: SupplierRequest): Task[Int] =
      Task.succeed(supplierRequest.data.length)
  }
}

trait OrderProcessor[F[_], C[_]] {
  def processOrders: F[C[Int]]
}

object OrderProcessor {
  def make[F[_]: Monad, C[_]: Traverse](
    customerService: CustomerService[F, C],
    supplierService: SupplierService[F]
  ): OrderProcessor[F, C] = new OrderProcessor[F, C] {
    override def processOrders: F[C[Int]] =
      for {
        orders <- customerService.collectOrders
        requestResponses <- orders.traverse { order =>
                             supplierService.createRequest(SupplierRequest(order.data))
                           }
      } yield requestResponses
  }
}

val OrderProcessorImpl = OrderProcessor.make[Task, List](CustomerService.make, SupplierService.make)
```

Effect as capability
```scala
case class CustomerOrder(data: String)

trait CustomerService {
  def collectOrders: Task[List[CustomerOrder]]
}

case class SupplierRequest(data: String)

trait SupplierService {
  def createRequest(supplierRequest: SupplierRequest): Task[Int]
}

object CustomerService {
  val make = ZIO.fromFunction[Effect, CustomerService] { effect =>
    new CustomerService {
      override val collectOrders: Task[List[CustomerOrder]] = {
        effect.effect(List(CustomerOrder("a"), CustomerOrder("abc"), CustomerOrder("qwer")))
      }
    }
  }
}

object SupplierService {
  val make = ZIO.fromFunction[Effect, SupplierService] { effect =>
    new SupplierService {
      override def createRequest(supplierRequest: SupplierRequest): Task[Int] = {
        effect.effect(supplierRequest.data.length)
      }
    }
  }
}

trait OrderProcessor {
  def processOrders: Task[List[Int]]
}

object OrderProcessor {
  def make(customerService: CustomerService, supplierService: SupplierService): OrderProcessor = new OrderProcessor {
    override def processOrders: Task[List[Int]] = {
      for {
        orders <- customerService.collectOrders
        requestResponses <- ZIO.traverse(orders) {order =>
          supplierService.createRequest(SupplierRequest(order.data))
        }
        // can't do that here! :D 🎉
        // _ <- ZIO.effect(println("launch nukes"))
      } yield requestResponses
    }
  }
}

val services = for {
  customerService <- CustomerService.make
  supplierService <- SupplierService.make
} yield (customerService, supplierService)

val orderProcessorImpl = for {
  (customerService, supplierService) <- {
    for {
      customerService <- CustomerService.make
      supplierService <- SupplierService.make
    } yield (customerService, supplierService)
  }.provide(Effect.Live)
  supplierService <- SupplierService.make
  orderProcessor = OrderProcessor.make(customerService, supplierService)
} yield orderProcessor
```

jdegoes:
@sideeffffect 

Keep in mind TF is not a real constraint:

```scala
def effect[F[_]: Applicative, A](effect: => A): F[A] = Applicative[F].pure(()).flatMap(_ => Applicative[F].pure(effect))

def foo[F[_]: Applicative]: F[Unit] = {
  println("Launch some nukes!")

  effect(println("Launch some stealth nukes!"))
}
```

It's a constraint in your mind and in pull request reviews only. 

That's not to say it is without benefit: if enforced by hand in a discipline team, TF can give you a list of effectful methods that each method calls. However you could also make an IDE plug-in that tells you which effectful methods each method calls, and puts it into documentation. Both have the same benefit (you can see what capabilities each method uses), but one is maintained by tool and other one is maintained by hand at great cost.

That said I want to find a way to make this work (if possible) for ZIO 2.0. We have some problems to solve to get there, ut we should track them in a separate ticket with your work here, and my prior work on the subject.

PR #1565, opened 2019-09-04
migrate zmanagedspec to zio test

mschuwalow:
ZIO test is a pleasure to work with.
Great work @adamgfraser!


core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
```diff
@@ -0,0 +1,760 @@
...
+              resultRef     <- Ref.make[Option[Exit[Nothing, String]]](None)
+              _ <- ZManaged
+                    .make(UIO.succeed("42"))(_ => finalizersRef.update("First" :: _))
+                    .onExitFirst(e => finalizersRef.update("Second" :: _) *> resultRef.set(Some(e)))
+                    .use_(ZIO.unit)
+              finalizers <- finalizersRef.get
+              result     <- resultRef.get
+            } yield assert(finalizers, equalTo(List("First", "Second"))) && assert(
+              result,
+              isSome(equalTo(Exit.succeed("42")))
```
ghostdogpr: You can use `succeeds`, `fails` and `isInterrupted` to apply predicates on `Exit`.

mschuwalow:
@ghostdogpr I ran into a few issues with scalajs compilation with this.
I for now increased timeouts in circleci and maximum heap space for sbt

ghostdogpr:
If you rebase you can now extend `ZIOSpec` which includes a timeout for test execution. You can see how it’s done in `ZQueueTest`.

I don’t think we should increase the 10 min timeout of CircleCI. If something is hanging for over 10 min without output it is likely going to hang forever and this is just costing more build credits. Increasing heap will probably help, we had similar issues in cats-interop and it was solved that way.

mschuwalow:
> If you rebase you can now extend `ZIOSpec` which includes a timeout for test execution. You can see how it’s done in `ZQueueTest`.
> 
> I don’t think we should increase the 10 min timeout of CircleCI. If something is hanging for over 10 min without output it is likely going to hang forever and this is just costing more build credits. Increasing heap will probably help, we had similar issues in cats-interop and it was solved that way.

Makes sense. I've updated the pr

PR #1572, opened 2019-09-04
Implement ZStreamChunk#drop and ZStreamChunk#take

### LINKED ISSUE #1568: Add ZStreamChunk#take

### LINKED ISSUE #1569: Add ZStreamChunk#drop

vasilmkd:
Resolves #1568.
Resolves #1569.

Fixed an interesting bug in `Chunk`. I wonder how it remained unnoticed for so long.
`(Chunk(1) ++ Chunk.empty ++ Chunk(2, 3, 4, 5, 6)).drop(8)` would fail with a `NegativeArraySizeException` (`Slice` chunks weren't checking their length when dropping elements).

@iravid

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -52,7 +52,7 @@ sealed trait Chunk[@specialized +A] { self =>
     val len = self.length
 
     if (n <= 0) self
-    else if (n == len) Chunk.empty
+    else if (n >= len) Chunk.empty
     else
       self match {
         case Chunk.Slice(c, o, l)        => Chunk.Slice(c, o + n, l - n)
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -52,7 +52,7 @@ sealed trait Chunk[@specialized +A] { self =>
     val len = self.length
 
     if (n <= 0) self
-    else if (n == len) Chunk.empty
+    else if (n >= len) Chunk.empty
```
iravid: Interesting that this did not show up in the property checks either. Good catch either way!

streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
```diff
@@ -44,6 +44,35 @@ class ZStreamChunk[-R, +E, @specialized +A](val chunks: ZStream[R, E, Chunk[A]])
   final def collect[B](p: PartialFunction[A, B]): ZStreamChunk[R, E, B] =
     ZStreamChunk(self.chunks.map(chunk => chunk.collect(p)))
 
+  /**
+   * Drops the specified number of elements from this stream.
+   */
+  final def drop(n: Int): ZStreamChunk[R, E, A] =
```
iravid: This made me change my mind on `StreamChunkEffect` - we probably do need that. There's no  way to implement this combinator  efficiently without an override.

PR #1575, opened 2019-09-05
Use Docker image to run tests and releases

ghostdogpr:
Installing the JDK frequently takes over 10 min, sometimes 25 min or even fails.
This PR uses pre-built docker images with JDK installed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ install_nodejs: &install_nodejs
   - run:
       name: Install node.js
       command: |
-        curl -sL https://deb.nodesource.com/setup_10.x | -E bash -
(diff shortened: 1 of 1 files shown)


ghostdogpr:
Giving up for now. I ran into ScalaJS errors `org.scalajs.testcommon.RPCCore$ClosedException: org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed`.

Using Docker also seems to increase the amount of flaky errors in RTSSpec.

NeQuissimus:
The flakiness is why we moved them in the first place :)
Can we add the JDK install into the cache maybe?

ghostdogpr:
@NeQuissimus we found the slowness was just Ubuntu servers being so slow. This fixes it https://github.com/zio/zio/pull/1578

PR #1576, opened 2019-09-05
[docs] textual introduction to zio error model

sidnt:


docs/advanced/error_model.md
```diff
@@ -3,4 +3,7 @@ id: advanced_error_model
 title:  "Error Model"
 ---
 
-Coming soon
+## Declared Errors vs Unforeseen Defects
+A ZIO value, iow a ZIO program, has a type parameter `E` which is the type of the declared errors it can fail with. E only covers those errors, which were specified at the outset. This does mean, that the same zio program can fail in ways that couldn't be foreseen. These unforeseen situations are aka _defects_ in a zio program, and they lie outside E.
```
iravid:
```suggestion
A ZIO value has a type parameter `E` which is the type of declared errors it can fail with. `E` only covers the errors which were specified at the outset. The same ZIO value could still throw exceptions in ways that couldn't be foreseen. These unforeseen situations are called _defects_ in a ZIO program, and they lie outside E.
```

docs/advanced/error_model.md
```diff
@@ -3,4 +3,7 @@ id: advanced_error_model
...
+
+Bringing abnormal app situations, from the domain of defects into that of E, enables the compiler to help us keep a tab on error conditions throughout the application, at compile time itself. This helps ensure the handling of domain errors in domain specific ways. Defects, on the other hand, can creep silently to higher levels in our application, and, if at all they get triggered, their handling might eventually be in more general ways, mostly leading to a loss in the richness of our application and a bad user experience
```
iravid:
```suggestion
Bringing abnormal situations from the domain of defects into that of E enables the compiler to help us keep a tab on error conditions throughout the application, at compile time. This helps ensure the handling of domain errors in domain specific ways. Defects, on the other hand, can creep silently to higher levels in our application, and, if at all they get triggered, their handling might eventually be in more general ways.
```

PR #1577, opened 2019-09-05
add apt cache to circle (wip)

mschuwalow:
so we don't download the entire internet every time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,9 @@ install_jdk: &install_jdk
 
 load_cache: &load_cache
   - restore_cache:
-      key: sbt-cache
+      keys:
+        - sbt-cache
+        - apt-cache
 
 save_cache: &save_cache
   - save_cache:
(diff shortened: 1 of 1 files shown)


yoohaemin:
Caching could be a way to go but I two problems comes to mind: 
- It's hard to determine when we update the caches (openjdk gets updates not so infrequently) and 
- it's hard to cherry pick openjdk packages from the cache directory given the limitations of the circleci config file. 

I chose a different approach to use a faster mirror and re-download (in a shorter amount of time), see #1578

mschuwalow:
> Caching could be a way to go but I two problems comes to mind:
> 
> * It's hard to determine when we update the caches (openjdk gets updates not so infrequently) and
> * it's hard to cherry pick openjdk packages from the cache directory given the limitations of the circleci config file.
> 
> I chose a different approach to use a faster mirror and re-download (in a shorter amount of time), see #1578

That looks like a saner choice. closing this one

PR #1578, opened 2019-09-05
Use faster Ubuntu mirrors during initial setup

yoohaemin:
Change the Ubuntu initial setup to use a mirror for jvm install, instead of the default `archive.ubuntu.com`. 

I assumed that the build server is located in the US, and chose `Princeton Mathematics` mirror that has a large bandwidth, and located in NJ. I expect that the download speed would be fast and stable across runs.

The concern that I have is that the mirror of choice might not have the latest packages. At this point, Princeton Mathematics is one day behind, which seems acceptable, but it might not always be the case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ install_jdk: &install_jdk
         deb http://mirror.math.princeton.edu/pub/ubuntu/ xenial main universe
         deb http://mirror.math.princeton.edu/pub/ubuntu/ xenial-updates main universe
         deb http://mirror.math.princeton.edu/pub/ubuntu/ xenial-backports main universe
-        deb http://security.ubuntu.com/ubuntu/ xenial-security main restricted universe
+        deb http://mirror.math.princeton.edu/pub/ubuntu/ xenial-security main restricted universe
         EOF
 
         if [ $JDK_VERSION == 11 ]; then sudo add-apt-repository ppa:openjdk-r/ppa -y; fi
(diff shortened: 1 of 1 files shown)


yoohaemin:
I noticed during the update phase, that the system has a lot of repositories registered. 
[This run](https://circleci.com/gh/zio/zio/21723?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) in particular spent 22 seconds downloading package list updates that is never used. It seems it's worth removing them.

yoohaemin:
```
Fetched 18.4 MB in 1min 5s (283 kB/s)
```
[link](https://circleci.com/gh/zio/zio/21748?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Maybe it's worth removing `security.ubuntu.com` (which is actually the same as `archive.ubuntu.com`) since it's just a throwaway machine, and we aren't running `apt update` anyway.

yoohaemin:
From what I see right now, all of the builds that runs on java 8 takes < 20 seconds to install JDK, and the one that required a separate ppa (java 11) took 40 seconds. 

I guess this is as fast as it'll get by switching to a mirror.

adamgfraser:
Thanks to everyone involved for your work on this! It is a huge collective benefit to all ZIO contributors. Just wanted to flag that I still saw one of my builds taking 20 minute to try to install JDK and ultimately timing out. My other builds were all much faster with the new changes.

https://circleci.com/gh/zio/zio/21826?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

yoohaemin:
Interesting, do you happen to know if the download was slow from the beginning or did it freeze in the middle?

adamgfraser:
I don't really unfortunately.

yoohaemin:
Ah it's the one from the PPA. My changes didn't affect that, and they are known to be very slow from time to time. I will try to come up with a solution for that too.

PR #1581, opened 2019-09-05
Adopt Scala Code of Conduct

neko-kai:
I want to open a discussion with regards to adopting a Code of Conduct across ZIO projects, I have 2 goals in mind with this PR:

1. Secure the long-term health of the project. ZIO is currently one of the most kind, safe and inclusive communities in Scala and I want to ensure that this continues to be the case, even as time goes by and new contributors join the project.
2. Secure cooperation with other projects and organizations. By adopting a Code of Conduct compatible with most other Scala projects we can help friendly and professional cooperation with other projects occur within and without the ZIO organization.

I think the Scala Code of Conduct is a natural choice towards achieving both of these goals, as it already promotes a set of values that has arisen naturally in ZIO and, quite luckily for us, it's already adopted by an overwhelming amount of other Scala projects, so we wouldn't have to worry about compatibility.

I would very much like to hear every ZIO contributor's thoughts about this change, and if you agree please approve the PR. :+1:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,19 @@
+# Code of Conduct
+
+We are committed to providing a friendly, safe and welcoming
+environment for all, regardless of level of experience, gender, gender
+identity and expression, sexual orientation, disability, personal
+appearance, body size, race, ethnicity, age, religion, nationality, or
+other such characteristics.
+
+Everyone is expected to follow the [Scala Code of Conduct] when
+discussing the project on the available communication channels. If you
+are being harassed, please contact us immediately so that we can
+support you.
+
+ ## Moderation
+
+For any questions, concerns, or moderation requests please contact a
+member of the project.
+
+[Scala Code of Conduct]: https://www.scala-lang.org/conduct/
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ To learn more about ZIO, see the following references:
 
 ---
 
+## Code of Conduct
+
+See the [Code of Conduct](CODE_OF_CONDUCT.md)
+
+---
+
 ### Legal
 
 Copyright 2017 - 2019 John A. De Goes and the ZIO Contributors. All rights reserved.


adamgfraser:
Me too.

LGLO:
I hope it won't turn out to be a dead code but even more I wish we'll never need to enforce it.
Who will judge if accusations are true or false?
We are adding law, that when enforced could lead to damage people instead of protecting them **because we don't have and will never have real judical system**. Even *real* judical systems are prone to error and are very expensive.
I mean law is somehow like JavaScript...
Or in other words: do we want a law that works correctly in 95% of cases?
I don't :-1: but also don't :+1: - this is very complex issue.

iravid:
If someone has the cycles, it'd be nice to add a few enhancements to the Scala CoC:
- "Drastic" measures (kicking, banning and the likes) against contributors and participants should require a quorum (size TBD) of core members
- @jdegoes should act as BDFL for the organization and as a tie-breaker for the above

I believe the Scala CoC, as I read it, does not address these safeguards.

jdegoes:
I'm on board, contingent on the above enhancements suggested by @iravid.

I would not expect reciprocity, and CoCs manifestly offer no protection or guarantees of any kind (see recent events), but it's a positive gesture that we can make, to set the right tone for the community and to show a willingness to be a team player, which are both important. 👍

NeQuissimus:
I'd suggest making a group in this Github org so that membership in such a "steering committee" is public.
Additionally, I suggest quorum to be ` > 2/3`

paulp:
Are you pre-committing to all future changes to the Scala code of conduct? Adopting their code of conduct is only going to embolden them. It achieves nothing and effectively acknowledges that obsession with superficial characteristics is a useful and necessary framing of existence, when in fact it is neither useful nor necessary. I note this only for posterity's sake.

LGLO:
If Scala CoC changes in unacceptable way, it's easy change.
Fr me at adding CoC here is like buying flowers - it doesn't solve any problems, but some people find it nice. Difference is that you cannot have physical allergy to Codes of Conduct.

paulp:
> Difference is that you cannot have physical allergy to Codes of Conduct.

Research is ongoing.

iravid:
I think everyone's on board with this addition, so merging this. I will submit a follow-up with the quorum and BDFL clauses.

jdegoes:
@paulp Wow, good to see you around these parts!

We are not committed to one CoC for life (i.e. all versions).

paulp:
@jdegoes thanks. I thought to myself, I wonder if there's anything going on in the world where absolutely nobody involved is thinking "this situation would improve if @paulp was here to comment on it." And what do you know, first place i checked!

neko-kai:
Just to clarify, I do not consider this Code of Conduct to be a "gesture", I believe it's a good instrument – indeed a defensive, exclusionary, but not evil instrument – for maintaining long-term social health of a project, especially if you remember that should any conflict arise, deescalation is preferrable to exclusion and infinitely preferrable to any possible harm.
I also, perhaps naively, think that Scala CoC cannot, realistically, change in an unacceptable way for ZIO. **Even** considering the current happenings – and should that really happen, we'll know and move away – it's not like a CoC can be used to hijack a project, it's a pledge that needs people to believe in it to work.

SethTisue:
there are some proposed CoC changes here: https://github.com/scala/scala-lang/pull/1079. they are intended to be minor/uncontroversial, but y'all might want to check it regardless

adamgfraser:
@SethTisue Thanks!

PR #1582, opened 2019-09-06
ZIO Test: Upgrade AssertResult to BoolAlgebra

adamgfraser:
Following up on discussions in #1438 this upgrades `AssertResult` to `BoolAlgebra` by adding a `Not` subtype to the algebraic data type. This lets us clean things up a lot because we don't need an `Either` embedded within it anymore to represent failure. It also allows us to preserve all information as we combine different assertions and implement all logical operations on assertion results including negation and more complex operators like implication, though that is not implemented here. We may want to rethink some of the other backend supporting data structures based on this but I think this is already a big enough PR. No significant changes to the API for users.


adamgfraser:
Thanks! Yeah, I think there is even more we can do but wanted to take it one step at a time.

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
```diff
@@ -0,0 +1,327 @@
...
+  final def either[A1 >: A](that: BoolAlgebra[A1]): BoolAlgebra[A1] =
+    or(self, that)
+
+  /**
+   * If this result is a success returns `None`. If it is a failure returns a
+   * new result containing all failures that are relevant to this result being
+   * a failure.
+   */
+  final def failures: Option[BoolAlgebra[A]] =
+    fold[Either[BoolAlgebra[A], BoolAlgebra[A]]](a => Right(success(a)))(
```
regiskuckaertz: Do you need to go through `Either` here?

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
```diff
@@ -0,0 +1,327 @@
...
+        case _                                    => false
+      }
+  }
+
+  /**
+   * Returns a result that is the logical conjunction of all of the results in
+   * the specified collection.
+   */
+  final def all[A](as: Iterable[BoolAlgebra[A]]): Option[BoolAlgebra[A]] =
+    if (as.isEmpty) None else Some(as.reduce(_ && _))
```
regiskuckaertz: I think the empty case does not disprove the assumption, so should be true by default.

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
```diff
@@ -0,0 +1,327 @@
...
+   */
+  final def and[A](left: BoolAlgebra[A], right: BoolAlgebra[A]): BoolAlgebra[A] =
+    And(left, right)
+
+  /**
+   * Returns a result that is the logical disjunction of all of the results in
+   * the specified collection.
+   */
+  final def any[A](as: Iterable[BoolAlgebra[A]]): Option[BoolAlgebra[A]] =
+    if (as.isEmpty) None else Some(as.reduce(_ || _))
```
regiskuckaertz: same here, there is nothing to prove so the empty case is failure.

regiskuckaertz: ha, same thing as before, you need an `A` for the failure. ok scratch that then.

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
adamgfraser: I think we do to deal with the `not` case. If we don't use `Either` then `notCase` turns `Some` into `None`, but what does it do with `None`? We've lost information by only using the option so for example if we had `not(not(a))` we could't recover the original value, whereas `a.swap.swap == a`. Open to more elegant ways for achieving that but I think we definitely need some intermediate structure and `Either` seems to work relatively well.

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
adamgfraser: Yeah, exactly.

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
regiskuckaertz: Of course 👍

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
regiskuckaertz: if I'm correct, it's not strictly a bool algebra though, as there should be one and only one "`false`" element.

adamgfraser: Yeah, I'm definitely not an expert on this. I think in boolean algebra as a branch of mathematics there is only one `true` and one `false` element. I was more trying to go for the idea that it is a reification, an "algebra", of boolean operations. I'm not committed to the name though not sure about a better one. Any ideas?

regiskuckaertz: oooh yeah, that makes sense. I don't have a better proposition either way 😆

adamgfraser:
Yeah, I agree with you about the name. What about `Result`? The two concrete types we have that instantiate it are:

```scala
type AssertResult = BoolAlgebra[AssertionValue]
type TestResult   = BoolAlgebra[FailureDetails]
```

So maybe this is a general "result" type that can be instantiated with specific types of result data. That is already very consistent with the Scaladoc I have now.

adamgfraser:
@jdegoes Any other thoughts on names for this? Would be nice to get it merged so we can start building on it.

PR #1583, opened 2019-09-06
ZIO Test: Implement TestAspect#timed

adamgfraser:
Following up on discussions on Gitter, there was interest in having an easy way to see which tests were taking longer to run. This PR adds a new `TestAspect` timed that reports the execution time of each test as it runs. We could also potentially do something more sophisticated like gather all the execution times, rank order them, and show what percent of total time they took so you could easily see an 80/20 analysis of what is driving your total testing time, but that is not implemented here. I also added a new subtype of `TestAspect`, `perTestLabeled`, that modifies each test but also provide access to the test label.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -18,10 +18,11 @@ package zio.test
 
 import java.util.concurrent.TimeoutException
 
-import zio.{ clock, Cause, ZIO, ZManaged, ZSchedule }
+import zio.{ clock, console, Cause, ZIO, ZManaged, ZSchedule }
 import zio.duration.Duration
 import zio.clock.Clock
-import zio.test.mock.Live
+import zio.console.Console
+import zio.test.mock._
 
 /**
  * A `TestAspect` is an aspect that can be weaved into specs. You can think of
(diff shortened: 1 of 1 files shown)


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -280,6 +281,25 @@ object TestAspect extends TimeoutVariants {
...
+    new TestAspect.PerTestLabeled[Nothing, Live[Clock] with Live[Console], Nothing, Any, Nothing, Any] {
+      def perTestLabeled[R <: Live[Clock] with Live[Console], E, S, L](
+        label: L,
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
+      ): ZIO[R, TestFailure[E], TestSuccess[S]] =
+        withLive(test.either)(_.timed).flatMap {
+          case (duration, ees) =>
+            live {
+              console.putStrLn {
+                s"$label took ${duration.toMillis / 1000.0} seconds"
```
ghostdogpr: I noticed we have similar code to render durations in `TimeoutVariants#renderDuration` and `DefaultTestReporter#logStats`. How about extracting it and using the same one everywhere? `TimeoutVariants` is probably the best one.

adamgfraser: Good idea. Will do that tomorrow.

mschuwalow:
Very nice!
I've looked into this as well after our discussion on gitter and I feel maybe printing after every test is not the right way to do this. @iravid also requested the option to have resources that are only allocated once per suite and I think those two can be easily combined.
This would it also make it trivial to postprocess results like the ranking you suggested.
I don't think this can be expressed by the current formulation of TestAspect but maybe you have some idea.

jdegoes:
I have been thinking about this. I have some code to help, basically a `TestAnnotation` so you can add arbitrary annotations (like timings) to each test, without altering any data structures. I'll try to get that PR up tomorrow and we can see if it might be useful for this.

adamgfraser:
@jdegoes That sounds interesting! I wonder if we could use it to add information about the random seed to each test and then if it was a failing property based test the test reporter could report it.

adamgfraser:
Closing in favor of #1951.

PR #1588, opened 2019-09-06
Add Gen#collect

ghostdogpr:
@adamgfraser I found `collect` would be pretty useful to me. I tried to follow the logic of existing operators, hope I didn't miss anything!


ghostdogpr:
Closing because of #1591. Since filter usage is not encouraged maybe it's better not to have this.

adamgfraser:
Yep I was thinking the same thing. Sorry I should have gotten documentation out about that earlier.

PR #1594, opened 2019-09-06
Correctly inherit fiber refs in `ZIO.raceWith`

### LINKED ISSUE #1515: FiberRef Not Being Inherited by Parent After Race
Normally if a `FiberRef` is updated by a child fiber that value is propagated back to the parent fiber on `join`, with a "last join wins" semantic if multiple child fibers have each updated their local copy of the `FiberRef`. For example, the following holds:

```scala
for {
  ref    <- FiberRef.make(0)
  fiber1 <- ref.set(10).fork
  fiber2 <- ref.set(20).fork
  _      <- fiber1.join
  _      <- fiber2.join
  result <- ref.get
} yield result == 20 // true
```

However, it appears that currently this is not happening after `race` and related combinators. For example, in the code below I would have expected the `FiberRef` to be set to either `10` or `20` depending on which fiber won the race, but in fact it is still `0`. I would have thought that the `FiberRef` would be updated to the winner state, and then if we joined the loser as in `zipPar` it would subsequently get updated to the loser state once the loser is joined.

```scala
for {
  ref    <- FiberRef.make(0)
  _      <- ref.set(10).race(ref.set(20))
  result <- ref.get
} yield result != 0 // false
```

A lot of combinators are defined in terms of `raceWith` so this significantly impacts the ability of users to rely on `FiberRef` updates being inherited after the fiber is joined.

mlangc:
Fixes #1515

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -867,22 +867,40 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E2, C] = {
     def arbiter[E0, E1, A, B](
       f: (Exit[E0, A], Fiber[E1, B]) => ZIO[R1, E2, C],
+      winner: Fiber[E0, A],
       loser: Fiber[E1, B],
-      race: Ref[Int],
+      waitForLeft: Boolean,
+      race: Ref[Boolean],
+      inherit: Ref[Option[Boolean]],
       done: Promise[E2, C]
-    )(res: Exit[E0, A]): ZIO[R1, Nothing, _] =
-      ZIO.flatten(race.modify((c: Int) => (if (c > 0) ZIO.unit else f(res, loser).to(done).unit) -> (c + 1)))
+    )(res: Exit[E0, A]): ZIO[R1, Nothing, _] = {
+
+      val handleRes =
+        (f(res, loser) <* {
+          ZIO.when(res.succeeded)(winner.inheritFiberRefs) *> inherit.set(Some(waitForLeft))
+        }).to(done)
+
+      ZIO.flatten(race.modify(b => (if (b) ZIO.unit else handleRes) -> true))
+    }
 
     for {
-      done <- Promise.make[E2, C]
-      race <- Ref.make[Int](0)
+      done    <- Promise.make[E2, C]
+      race    <- Ref.make[Boolean](false)
+      inherit <- Ref.make(None: Option[Boolean])
       c <- ZIO.uninterruptibleMask { restore =>
             for {
-              left  <- ZIO.interruptible(self).fork
-              right <- ZIO.interruptible(that).fork
-              _     <- left.await.flatMap(arbiter(leftDone, right, race, done)).fork
-              _     <- right.await.flatMap(arbiter(rightDone, left, race, done)).fork
-              c     <- restore(done.await).onInterrupt(left.interrupt *> right.interrupt)
+              left   <- ZIO.interruptible(self).fork
+              right  <- ZIO.interruptible(that).fork
+              left2  <- left.await.flatMap(arbiter(leftDone, left, right, true, race, inherit, done)).fork
+              right2 <- right.await.flatMap(arbiter(rightDone, right, left, false, race, inherit, done)).fork
+
+              inheritFiberRefs = inherit.get.flatMap {
+                case None        => ZIO.unit
+                case Some(true)  => left2.inheritFiberRefs
+                case Some(false) => right2.inheritFiberRefs
+              }
+
+              c <- restore(done.await <* inheritFiberRefs).onInterrupt(left.interrupt *> right.interrupt)
             } yield c
           }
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -867,22 +867,40 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
-              _     <- right.await.flatMap(arbiter(rightDone, left, race, done)).fork
-              c     <- restore(done.await).onInterrupt(left.interrupt *> right.interrupt)
+              left   <- ZIO.interruptible(self).fork
+              right  <- ZIO.interruptible(that).fork
+              left2  <- left.await.flatMap(arbiter(leftDone, left, right, true, race, inherit, done)).fork
+              right2 <- right.await.flatMap(arbiter(rightDone, right, left, false, race, inherit, done)).fork
+
+              inheritFiberRefs = inherit.get.flatMap {
+                case None        => ZIO.unit
+                case Some(true)  => left2.inheritFiberRefs
```
jdegoes: There is a more direct way to encode this: use `Ref[Task[Unit]]`. Then you can start that `Ref` off with `ZIO.unit`, and then you can either stick `left2.inheritFiberRefs` into that, or `right2.inheritFiberRefs` into it, depending on the race outcome.

This will simplify the code a _lot_.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -867,22 +867,40 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E2, C] = {
     def arbiter[E0, E1, A, B](
       f: (Exit[E0, A], Fiber[E1, B]) => ZIO[R1, E2, C],
+      winner: Fiber[E0, A],
       loser: Fiber[E1, B],
-      race: Ref[Int],
+      waitForLeft: Boolean,
+      race: Ref[Boolean],
+      inherit: Ref[Option[Boolean]],
```
jdegoes: This would change to `Ref[Task[Unit]]`, see below.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -867,22 +867,40 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E2, C] = {
     def arbiter[E0, E1, A, B](
       f: (Exit[E0, A], Fiber[E1, B]) => ZIO[R1, E2, C],
+      winner: Fiber[E0, A],
       loser: Fiber[E1, B],
-      race: Ref[Int],
+      waitForLeft: Boolean,
+      race: Ref[Boolean],
```
jdegoes: Can we rename this to `raceDone`?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -867,22 +867,40 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E2, C] = {
     def arbiter[E0, E1, A, B](
       f: (Exit[E0, A], Fiber[E1, B]) => ZIO[R1, E2, C],
+      winner: Fiber[E0, A],
       loser: Fiber[E1, B],
-      race: Ref[Int],
+      waitForLeft: Boolean,
```
jdegoes: Do we need this?

core/shared/src/main/scala/zio/ZIO.scala
mlangc: Done ;-)

core/shared/src/main/scala/zio/ZIO.scala
mlangc: See above

core/shared/src/main/scala/zio/ZIO.scala
mlangc: Without the refactoring proposed above - yes, we need this to have the necessary context information in `arbiter` if I'm not mistaken. I've renamed it to `leftWins` though...

mlangc:
> With these changes if we race two effects with `raceWith` and then `join` the loser as in `zipPar`, would the loser `FiberRef` be inherited?

@adamgfraser yes, if the looser wasn't a bad looser that is ;-) - I mean if the looser did not fail. There is a test that checks for this: https://github.com/zio/zio/pull/1594/files#diff-b3f6f229206b73e705ea7e4c6e03784bR23

adamgfraser:
@mlangc Thanks! It looks like right now we are testing that one or the other `FiberRef` gets inherited. Could we add a test where we `ZipPar` with two successful fibers, one of which is slower, and show that it is the loser `FiberRef` that gets inherited?

mlangc:
@adamgfraser thanks a lot for brining this up! After adding tests for `zipPar` (see https://github.com/zio/zio/pull/1594/files#diff-b3f6f229206b73e705ea7e4c6e03784bR25) I found out that the implementation needed another tweak.

Issue #1472, opened 2019-08-24
ZIO Test: Support Generators for Functions


adamgfraser:
I would like to add generators for functions to our property based testing framework so we could have generators of type `Gen[R, A => B]`. I think that, while there are definitely more combinators we could add, this is the last major feature to bring ourselves to parity with existing libraries like ScalaCheck in this area. Right now I'm struggling to do this.

The general approach used by ScalaCheck and QuickCheck is to use the `A` to perturb the random number generator in some way. Typically you have a `Cogen` type that basically represents a "hashing" function that converts an `A` to a `Long` and then you use `^` or some operator to combine that with the existing random number seed to get a new random number generator, and use that along with a `Gen[B]` to generate `B` values. This generally works well but in every implementation I have seen requires running the `Gen[B]`. Basically the problem is we have a `Gen[R, A => Gen[R, B]]` and at least I haven't found a way to convert this to a `Gen[R, A => B]` without unsafely running it. For other testing frameworks this is less of an issue because their effect type is generally a `State` monad that doesn't support other effects, so running it to eliminate the local effect is less of an issue.

@jdegoes Do you have any thoughts on how we might approach this?

jdegoes:
We can convert any total side-effecting method `f : A => B` into a pure function by doing the following:

1. Create a mutable map (a concurrent hash map, probably; at least on JVM, Scala.js has no such thing);
2. Create a memoizing wrapper around `f` that will invoke `f` at most once for a given `A`, storing the result inside the mutable map.

As long as edge conditions are properly handled (i.e. no race conditions), the function, let's call it `f'`, will now be referentially transparent and safe for exposing to user-land code.

So given a `Gen[R, B]`, we should be able to produce a `Gen[R, A => B]`, by building the `A => Gen[R, B]` (which will, for every unique `A`, generate a `B` effectfully); then using `ZIO.runtime` (for `unsafeRun`) and a memo table, we can safely convert the `A => Gen[R, B]` into an effect containing `A => B`.

The main challenge, perhaps, will be the distinction between "discrete" and "continuous" random variables: for the discrete case, the `Gen` stream for `B` will contain all values, deterministically specified; for the continuous case, the stream will contain a single effectful value that samples the state space randomly and non-deterministically. Perhaps you can unify them by doing `.forever` on the stream and then converting it to a `Queue`, which you can then use as a fresh source of `B`'s, one for every unique `A`.

[...]

Issue #1548, opened 2019-09-02
Flaky test for ZStream#zipWithLatest


ghostdogpr:
Build: https://circleci.com/gh/zio/zio/20915
```
[error]     x zipWithLatest
[error]  
[error]  (0,0)
[error]  (0,1)
[error]  (1,1)
[error]  (1,2)
[error]  (2,2)
[error]  (2,3)
[error]  - (3,3)
[error]  + (2,4)
[error]  (3,4) (StreamSpec.scala:1813)
[error] Added (1)
[error] (3,3)
[error] 
[error] Missing (1)
[error] (2,4)
[error] 
```

cc @regiskuckaertz @iravid

iravid:
Ouch, yeah. We should rewrite that test without any sleeps at all. Or switch to MockClock.

PR #1595, opened 2019-09-06
Migrate stream tests to ZIO Test

regiskuckaertz:
Right. I've gone through all the tests in the streams project. It's not ready yet but wanted to open for review in case I'm doing something awful.

Summary:

- tests expecting random functions are pending #1472  
- there's only one test in `StreamChunkSpec` and it takes ~50s on my machine. is there something wrong with my generator? I had to reduce `check` to only 50 samples
- equality on arrays does not work with `equalTo`, as does equality on some strings (see `utf8Decode`)
- tests using implication are pending #1582 
- recursive generators overflow the stack
- some tests are failing even though I did not change them
- not sure I got the tests using `fails` right, the errors don't seem to be properly detected
- had to set `sized` as a `def` or would always get an NPE


adamgfraser:
This is great! I'm going to try to do function generators this weekend. Will come back with more detailed thoughts on the other property based testing issues.

jdegoes:
@regiskuckaertz Thank you for this one! Tests are looking sweet.

I'll assume you and @adamgfraser have this one unless you pull me in for anything in particular...

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1,1851 +1,1807 @@
...
-
-  private def flatMapParSwitchOuterDefectsInterruptAllFibers = unsafeRun {
-    val ex = new RuntimeException()
-
-    for {
-      substreamCancelled <- Ref.make[Boolean](false)
-      latch              <- Promise.make[Nothing, Unit]
-      result <- (Stream(()) ++ Stream.fromEffect(latch.await *> ZIO.die(ex)))
-                 .flatMapParSwitch(2) { _ =>
-                   Stream.fromEffect((latch.s
```
ghostdogpr: Btw not sure you had seen this? https://github.com/zio/zio/issues/1548
`MockClock` could be used to ensure deterministic timings.

regiskuckaertz: Yes I did, it's on my todo list 👍

regiskuckaertz:
@iravid 
- equality in the `Sink.utf8Decode*` tests does not work, even though the strings "look" the same when printed. After converting to lists of bytes, I'm getting `List() did not satisfy equalTo(List(0))`. Is it possible string concatenation swallows some bytes and the state should be `(Chunk[Byte], Chunk[Byte]) in `utf8DecodeChunk`?
- also I'm getting consistent errors in the `zipPar` tests, I think that's because the current `ZipParLaws.remainders` is not correct in the assertion:
```scala
longer must_=== rem
rem.endsWith(shorter) must_=== true
```
only the last one is reported. When I do that indeed it works, but what I did is:
```scala
assert(longer, equalTo(rem)) && assert(rem.endsWith(shorter), isTrue)
```
which breaks four tests. Is it that the first assertion is not correct?

@adamgfraser in the point above, it was hard to isolate which assertion was breaking as the only message I got was `List(3, 4, 5, 6) did not satisfy equalTo(List(5, 6))`. I was under the impression that ZIO Test would lay out the whole stack of assertions from the root down to the failing one(s)?

adamgfraser:
@regiskuckaertz @jdegoes Your last point raises a good issue. Right now we have ways of combining assertions versus combining assertion results. If you combine assertions:

```scala
assert(52, equalTo(42) || (isGreaterThan(5) && isLessThan(10)))
```

You are going to get very nice reporting of the whole failing predicate versus the fragment. On the other hand if you combine assertion results:

```scala
assert(52, equalTo(42)) && assert(5, isLessThan(10))
```

We're going to store those two assertions in as `AssertResult` / `BoolAlgebra`, when we execute the tests if anything fails we'll actually throw away everything that isn't a failure right now, and then we will report the failures to you as you see.

I think there are a couple of things we could do to improve on this. First, once we get the `BoolAlgebra` or whatever we end up calling it merged I think it would be worth exploring a version where an executed test maintains one tree of all the successes and failures. Right now if a test is a success we throw away all the failures and vice versa so we are losing information. If we kept the whole tree the test reporter could have logic to see if the tree is a failure and if so print a drill down of the tree to the failing part.

Second, it goes a bit to the point that @jdegoes raised on Gitter about having a way to annotate tests with a message, though I think ideally the user wouldn't have to annotate here.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
adamgfraser: One thing to watch out for here is that the `MockEnvironment` provides the `MockClock` as its default `Clock` implementation. While the `MockClock` is great, it requires manually adjusting the time, so if you just drop in tests that previously relied on the live clock they are not likely to work. For tests that rely on the passage of time I would probably start by providing `Live[Clock]` to them to make sure that the test is working and has been migrated correctly and then look to move to the `MockClock`.

regiskuckaertz: oooh that's super useful, thanks! I'll do that and then we can look at `MockClock` in a follow-up. I tried using it when writing `zipWithLatest` but one fiber wasn't making progress, so I thought I was doing something wrong.

adamgfraser:
@regiskuckaertz I spent some time looking at `StreamChunkSpec`. The time seemed to be very dependent on the size of some of the elements you are generating. If I change

```scala
checkM(succeededStreamChunkGen(Gen.anyString), Gen.anyInt) { (s, n) =>
```

to

```scala
checkM(succeededStreamChunkGen(Gen.small(Gen.stringN(_)(Gen.alphaNumericChar))), Gen.int(-10, 10)) { (s, n) =>
```

and reduce the maximum size of the first generator in `chunkIxGen` from `100` to `10` the total time drops to a few seconds.

I know that isn't entirely satisfying since presumably the test was running without issues with the larger sizes before. I think part of the issue is right now we're wrapping every random number generation we do in a `ZIO`. We may be able to be smarter about that in cases where we are generating a collection, for example. I will open an issue to do some benchmarking and further analysis.

adamgfraser:
Are you still having issues with `Assertion#fails`? I looked at the test for number parsing in `SinkSpec` on your branch and it looked okay, but let me know if there is something we still need to investigate there.

adamgfraser:
For the implication tests I think you can do something like `if(condition) assert(p, isTrue) else assert((), anything)` until we get something more solid in.

regiskuckaertz:
> and reduce the maximum size of the first generator in chunkIxGen from 100 to 10 the total time drops to a few seconds.

Nice! I'll do that, we can see if these constraints can be relaxed later 🎉 

> Are you still having issues with `Assertion#fails`?

Nope, not anymore, though I have commented out loads to focus on one issue at a time 😅 

> For the implication tests

We have `a ⇒ b ≣ ¬a ⋁ b` so I can reformulate as follows perhaps
```scala
!assert(res1, cond1) || assert(res2, cond2)
```

regiskuckaertz:
@adamgfraser ha, yes that's because `fails` tests the error channel. I have added a `dies` combinator to test unchecked errors. Happy to raise a separate PR for it.

adamgfraser:
@regiskuckaertz `dies` looks like a good addition. A PR with that and maybe throw in `anyString` would be great. I'm also happy to do if it is easier. I know this is a huge PR.

So in terms of stuff you need from ZIO Test at this point I have: (1) better support for implications, (2) resolution for type inference issue with `assert` and `equalTo`, (3) better reporting of assertions combined using `&&`, and (4) continued performance optimizations, though not critical now that you are using smaller generators for some of your values. Anything else?

regiskuckaertz:
Yep that's it! I've got a few issues with compilation of some tests, where the compiler infers the type as `Any` but I have yet to get to the bottom of this.

regiskuckaertz:
I'll raise a separate PR for `die` and `anyString` 👍

adamgfraser:
Sounds great! Let me know if there are any issues you find on the ZIO Test side.

iravid:
@regiskuckaertz Could I get a sample input that fails the `utf8Decode` tests?

PR #1602, opened 2019-09-07
ZIO Test: Report timeouts in a prettier way

### LINKED ISSUE #1599: ZIO Test: Report timeouts in a prettier way
In case of timeout from TestAspect, we're displaying this:
```
[info]     Fiber failed.
[info]     An unchecked error was produced.
[info]     java.util.concurrent.TimeoutException: Timeout of Finite(1000000000) exceeded
[info]     	at zio.test.TestAspect$$anon$12.$anonfun$perTest$14(TestAspect.scala:293)
[info]     	at zio.internal.FiberContext.nextInstr(FiberContext.scala:709)
[info]     	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:332)
[info]     	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:601)
[info]     	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[info]     	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[info]     	at java.lang.Thread.run(Thread.java:748)
[info]     No ZIO Trace available.
```

Instead, display something like this:
```
[info] Timeout of 60 seconds exceeded
```

ghostdogpr:
Fixes #1599

### DIFF
--- a/core/shared/src/main/scala/zio/duration/Duration.scala
+++ b/core/shared/src/main/scala/zio/duration/Duration.scala
@@ -112,8 +112,8 @@ object Duration {
         case millis if millis < 1000 => s"$millis ms"
         case millis if millis < 60000 && millis % 1000 == 0 => s"${millis / 1000} s"
         case millis if millis < 60000 => s"${millis / 1000} s ${millis % 1000} ms"
-        case millis if (millis % 60000) / 1000 == 0 => s"${millis / 60000} min"
-        case millis => s"${millis / 60000} min ${(millis % 60000) / 1000} s"
+        case millis if (millis % 60000) / 1000 == 0 => s"${millis / 60000} m"
+        case millis => s"${millis / 60000} m ${(millis % 60000) / 1000} s"
       }
   }
 
--- a/core/shared/src/main/scala/zio/duration/Duration.scala
+++ b/core/shared/src/main/scala/zio/duration/Duration.scala
@@ -107,15 +107,14 @@ object Duration {
 
     override def asJava: JavaDuration = JavaDuration.ofNanos(nanos)
 
-    override def render: String = toMillis match {
-      case 0                         => "0 milliseconds"
-      case 1                         => "1 millisecond"
-      case millis if millis < 1000   => s"$millis milliseconds"
-      case millis if millis < 2000   => "1 second"
-      case millis if millis < 60000  => s"${millis / 1000} seconds"
-      case millis if millis < 120000 => "1 minute"
-      case millis                    => s"${millis / 60000} minutes"
-    }
+    override def render: String =
+      toMillis match {
+        case millis if millis < 1000 => s"$millis ms"
+        case millis if millis < 60000 && millis % 1000 == 0 => s"${millis / 1000} s"
+        case millis if millis < 60000 => s"${millis / 1000} s ${millis % 1000} ms"
+        case millis if (millis % 60000) / 1000 == 0 => s"${millis / 60000} min"
+        case millis => s"${millis / 60000} min ${(millis % 60000) / 1000} s"
+      }
   }
 
   case object Infinity extends Duration {
--- a/core/shared/src/main/scala/zio/duration/Duration.scala
+++ b/core/shared/src/main/scala/zio/duration/Duration.scala
@@ -60,6 +60,8 @@ sealed trait Duration extends Ordered[Duration] with Serializable with Product {
 
   /** The `java.time.Duration` returned for an infinite Duration is technically "only" ~2x10^16 hours long (`Long.MaxValue` number of seconds) */
   def asJava: JavaDuration
+
+  def render: String
 }
 
 object Duration {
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/zio/duration/Duration.scala
```diff
@@ -104,6 +106,16 @@ object Duration {
     override def asScala: ScalaDuration = ScalaFiniteDuration(nanos, TimeUnit.NANOSECONDS)
 
     override def asJava: JavaDuration = JavaDuration.ofNanos(nanos)
+
+    override def render: String = toMillis match {
```
adamgfraser: Are we okay with losing so much precision here? If a duration is 1500 milliseconds for example that is going to get displayed as 1 second. Seems like there are some use cases where that could be okay but others where it could be very undesirable. Maybe for values in between we divide by `1000.0` or `60000.0` and report the double value plus the unit, possibly with some formatting of the double value so it doesn't have too many digits?

ghostdogpr: I didn’t change how it was but yeah maybe we can do better. I’ll have a look at how it’s done elsewhere.

adamgfraser: You're right. I messed up [here](https://github.com/zio/zio/blob/e83b74cfef73cbf952a6bec5418d92855a92f1f1/test/shared/src/main/scala/zio/test/TimeoutVariants.scala#L80) by doing integral instead of fractional division. I'm just thinking this is what is going to show up for anyone who is using `Duration` for anything. So there is a lot of opportunity for improvement since seeing the nanoseconds today is not that useful but we want to make sure we don't cause problems for people who are doing different things with time.

ghostdogpr: I improved it as follows:
```
4 ms
3 s
3 s 459 ms
2 min
2 min 43 s
```
I'm only displaying 1 unit below except when it's zero.

adamgfraser:
This looks great. Do you think the `toString` method in `Duration` should call `render` now?

core/shared/src/main/scala/zio/duration/Duration.scala
```diff
@@ -104,6 +106,15 @@ object Duration {
     override def asScala: ScalaDuration = ScalaFiniteDuration(nanos, TimeUnit.NANOSECONDS)
 
     override def asJava: JavaDuration = JavaDuration.ofNanos(nanos)
+
+    override def render: String =
+      toMillis match {
+        case millis if millis < 1000 => s"$millis ms"
+        case millis if millis < 60000 && millis % 1000 == 0 => s"${millis / 1000} s"
+        case millis if millis < 60000 => s"${millis / 1000} s ${millis % 1000} ms"
+        case millis if (millis % 60000) / 1000 == 0 => s"${millis / 60000} min"
```
regiskuckaertz: just `m` for minutes usually

ghostdogpr:
> This looks great. Do you think the `toString` method in `Duration` should call `render` now?

@adamgfraser I did it initially but then I thought maybe some people are already relying on it and didn't want to change the behavior. Also I thought some people might want the actual `nanos` so it might be good to give them the choice, `toString` being the "raw" and `render` a nicer version for printing. I don't have a strong opinion on it though 😄

core/shared/src/main/scala/zio/duration/Duration.scala
ghostdogpr: Fixed, thanks!

test/shared/src/main/scala/zio/test/TestTimeoutException.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test
+
+final case class TestTimeoutException(message: String) extends Throwable(message, null, true, false)
```
jdegoes: Maybe make this extend `TimeoutException`?

ghostdogpr: It was the case initially but we changed for 2 reasons:
- not generating any unnecessary call stack
- since we don't display the full cause in `DefautTestReporter` for this error type, if a user uses `TimeoutException` in their code, they won't see the cause either. By using a separate class we can mute just that one.

adamgfraser: If we made `TestTimeoutException` extend `TimeoutException` and pattern matched on `TestTimeoutException` in `DefaultTestReporter` wouldn't that still get us what we want? Or do we lose the ability to specify no stack trace if we do that?

ghostdogpr: I think it would generate a stack trace then. But what would be the benefit?

adamgfraser: Yeah, I thought it might just be good for consistency if we ever want to do anything else with to handle that exception, but we will always know what it is. I'm good to merge.

PR #1607, opened 2019-09-07
Make required configuration stand out more in Testing docs

ioleo:
I missed `zio-test-sbt` dependency when upgrading my project. Perhaps with this change it will stand out more and may save someone some time.

### DIFF
--- a/docs/usecases/testing.md
+++ b/docs/usecases/testing.md
@@ -3,7 +3,17 @@ id: usecases_testing
 title:  "Testing"
 ---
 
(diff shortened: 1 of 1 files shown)


adamgfraser:
Thanks! This is a great suggestion and makes it much clearer. Just one thing. I recall when I was testing it that the SBT test runner couldn't just be a dependency for `"test"` and you had to do something like `"test->test;compile->compile"` or just leave that off entirely to be able to run tests with SBT. Could you double check that? Otherwise much appreciated and a nice improvement.

adamgfraser:
You're totally right, works fine as is. This will really help people get started quickly with ZIO Test! Thanks again for both identifying the issue and solving it!

Issue #1608, opened 2019-09-07
Unexpected behaviour of throws assertion


vilu:
```
  import zio.test._
  import zio.test.Assertion._
  val ex = new Exception
  val a = throws(equalTo(ex))
  println(assert(throw ex, a)) // expected Value(Right(())), got exception
```

Maybe I'm missing something here but this seems wrong to me.

adamgfraser:
@vilu Thanks for reporting! Let me take a look and will get back to you.

vilu:
I've checked already, I'm opening up a PR now. Maybe you want to check that first and just merge if you agree?

adamgfraser:
Oh awesome. You're one step ahead. Yes, will do.

vilu:
Done

ghostdogpr:
Closed by #1609

PR #1609, opened 2019-09-07
1608 Fix issue with throws assertion always throwing exception when true

vilu:
Suggested fix for issue #1608


adamgfraser:
Yes you are 100% right. This is much better.

PR #1610, opened 2019-09-07
Add ZIO.assert

ioleo:
The `assert` combinator is useful for introducing mock-style assertions in your tests.

Example use case:
* you have a `Command` type method which takes some parameters
* you want to check if it was called with expected arguments
* you do not want to introduce heavy libraries like Mockito

```scala
def send(recipient: String, content: String): ZIO[R, InvalidEmailAddress, Unit]

// for test purposes be implemented as follows
def send(recipient: String, content: String) =
  UIO.unit
    .assert(recipient == "foo@acme.com", s"""Invalid recipient $recipient, expected "foo@acme.com".""")
    .assert(content == "lorem ipsum dolor sit amet", s"""Invalid content $content, expected "lorem ipsum dolor sit amet".""")
```

@adamgfraser We might also have some special handling for `AssertionError` in `zio-test`, to replace long stack trace (for failed fiber) with a shorter one. What do you think?


adamgfraser:
I think this is interesting. It is kind of the ZIO equivalent of `assert` in the Scala standard library. A couple of thoughts:

- What if the type of the assertion parameter was `A => Boolean`? You could still support your use case by just ignoring the `A` value, but then it would support a lot more use cases where you want to check the value of an effect as you go.
- Could we call the `assert` method in the `scala.Predef` in the standard library so it elidable like that version?
- Could we provide a default argument so the user doesn't have to specify a failure message if they don't want to? Again the standard library version might provide good inspiration here.

neko-kai:
I feel like `filterOrDie` & `filterOrDieMessage` (see `filterOrFail`) could be a bit less ad-hoc and provide value on top of `Predef.assert`; as in here `ZIO.assert` is basically exactly the same as `.map(_ => assert(...))`

ioleo:
@neko-kai I like the `filterOrDie` and `filterOrDieMessage`, incorporated into the PR.
Also following the discussion on gitter, I added custom `AssertionFailure` (with no stack trace) and used `untraced` for the `assert` combinators, such that assertions made this way do not clutter the test result output with long stack/execution traces.

core/shared/src/main/scala/zio/AssertionFailure.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package zio
+
+/**
+ * Represents assertion failure.
+ *
+ * This class is used to avoid constructing stack trace and special handling by `zio-test`.
+ */
+final case class AssertionFailure(message: String)
```
jdegoes: It might be good to make this extend `AssertionError`.

jdegoes: In fact, maybe we could just use `AssertionError`? Then one less file, one less class we have to include.

adamgfraser: I like that idea. I think @ioleo was looking to not have the stack trace printed to make his mocking output clean, but I think that makes it a lot less general because in a lot of cases the user wants to see the stack trace. Maybe we can just provide a helper method to remove the stack trace or change the way it is rendered. Feels like this is coming up a couple of times in this and the timeout reporting PR.

ioleo: The reasoning for this as a seperate class was to:
* not pay the cost of stack trace construction
* have special handling in `zio-test` such that assertions failures made this way are simply reported in one line (instead of cluttering output with full stack and execution trace --> which are useful in scenarios where an "unexpected" exception was throws --> in case of test assertions this is an "expected" failure --> if the predicate does not hold, we expect the test to fail)

ioleo: See also @ghostdogpr comment on [#1602](https://github.com/zio/zio/pull/1602/files#r322037212)

ioleo: @jdegoes 

1) I was only wondering, if perhaps moving `AssertionFailure` to `zio-test` and also `assert` & `assert_` to implicit class in `zio-test`, to not clutter the "main ZIO" API.

For all other use cases (that don't need special handling in `zio-test` `filterOrDie` & `filterOrDieMessage` should do the job.

2) Or maybe have `assert` & `assert_` die with `java.lang.AssertionError` as you proposed, and have implicit class in `zio-test` with `testAssert` & `testAssert_` (or whatever diffrent name) handle the "special case".

WDYT?

ioleo:
Resolved conflicts.

adamgfraser:
@ioleo Here is my take, @jdegoes may disagree. I think special casing these throwables is not great. I would like to add a combinator `noStackTrace` on `Cause` that would cause a `Cause` with an underlying throwable to be rendered with just the throwable and not the stack trace (maybe removing tracing as well or maybe those should be separate and composed). Then for your use case and others someone can just call `noStackTrace` on the `Cause` when creating it to get the desired behavior. Then `filterOrDie` and `filterOrDieMessage` go on `ZIO` and you can use those along with `noStackTrace` to implement your desired behavior at the user level.

ioleo:
@adamgfraser that would be fine with me, I would be happy to implement it as part of this PR, but before I commit to early @jdegoes do you agree with this approach?

neko-kai:
@adamgfraser @ioleo 
This would be implementable as an additional flag in each Cause branch because
1. Throwables are mutable and don't have `copy`, you can't just make a new Throwable without a stacktrace
2. Not sure if we can mutate the stacktrace away, even if we can it's not particularly desireable, even wrapping the operation in IO doesn't make it sensible – mutating the throwables will mutate unrelated immutable values of Cause that contain the same Throwable.

So, the only way to achieve this would be a flag (or an additional branch) that changes behavior of `prettyPrint`.

Note that you can strip away ZIO traces with `Cause#untraced` method

adamgfraser:
@neko-kai Thanks! That is really helpful and sounds like a good approach.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -139,6 +139,7 @@ object DefaultTestReporter {
 
   private def renderCause(cause: Cause[Any], offset: Int): String =
     cause match {
+      case Cause.Die(AssertionFailure(message))     => message
```
jdegoes: @adamgfraser @ioleo 

What about consistently using `null` stack trace to indicate special rendering (i.e. rendering just the message)? Then maybe we can reduce never-ending lists of special cases?

adamgfraser: @jdegoes Yes I think that is definitely the direction we want to go in. See above comments from @neko-kai that I think get us to the same place with `Cause`.

jdegoes:
@adamgfraser  

After reading your other comments, I'll sketch an earlier idea:

```scala
final case class Stackless[E](cause: Cause[E]) extends Cause[E]
```

This would be a new case in the `Cause` enumeration, indicating that no stack trace should be rendering at any deeper levels.

I don't think copying throwables would be necessary. /cc @neko-kai (damn your new username, Kai!!! 😉 )

adamgfraser:
@jdegoes That sounds very logical to me.

neko-kai:
I'm not super happy about a new branch because it's a breaking change, while adding a new Boolean field to Fail/Die can be done in a backwards-compatible way. Note that we could use this opportunity to add not just a `Stackless` branch, but a more general `Meta` or `RenderingOptions` branch that could be extended with more flags in the future – possibly without breaking bincompat, by omitting fields in apply/unapply

> damn your new username, Kai!!! 😉 

Eh, but why? ;)

> I don't think copying throwables would be necessary.

I wish we could copy throwables, then we'd do just that; I think that's not doable in general on the JVM, although I hope I'm wrong. Note that we could just replace throwables with `new Exception(err.getMessage)` – but that would lose the runtime type of the exception

pshirshov:
> I wish we could copy throwables, then we'd do just that;

We could do that of course. At least we always have that old dirty trick "copy through serialization" which works in most of the cases.

>  but that would lose the runtime type of the exception
You may preserve the type as well.

I'm not advocating this, just checking in.

adamgfraser:
@ioleo To drive to some conclusion here, how about we do this? In this PR, just implement `filterOrDie and `filterOrDieMessage`. It seems like there is consensus that these would be useful and someone else just opened an issue in #1645 that this would resolve. Then we can open a follow up issue to implement some kind of "stackless" cause.

ioleo:
@adamgfraser Agreed, lets merge `filterOrDie` & `filterOrDieMessage` as a step forward and keep the discussion alive about the rest.

Issue #1587, opened 2019-09-06
STM that contains only `get` operations can see partial state


adamgfraser:
https://circleci.com/gh/zio/zio/21901?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
[info]         fold:
[error]           x is atomic
[error]  30 != 0; 30 != 1000 (TArraySpec.scala:99)
```

ghostdogpr:
@LGLO any idea about this one?

adamgfraser:
Just saw another issue here. This time a different method. Is there something we changed that could have impacted this?

https://circleci.com/gh/zio/zio/22084?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

[info]         map:
[error]           x creates new array atomically
[error]  19 -> 3 != 19 -> 19; 19 -> 3 != 3 -> 3 (TArraySpec.scala:137)

dkarlinsky:
@adamgfraser I think I managed to create a test that (pretty much) consistently reproduces the failure.
[Here the code](https://github.com/dkarlinsky/zio/blob/3e4edac589ae5d88bfd66bca28da3209797feff7/core-tests/jvm/src/test/scala/zio/stm/TArrayFlakySpec.scala)
It basically runs `TArraySpec.foldAtomic` up to 5000 times, with batches of 50 in parallel, until the assertion of `equalTo(0) or equalTo(N)` fails.
The parallelism seems to be important - before I introduced it I had no luck getting a failure.
I've also run this test on the [commit](https://github.com/zio/zio/commit/934967ad0b2fd32ba535fb3548df2543d042bb78) that introduced `TArray` and `TArraySpec` and it failed there as well.

LGLO:
Thank you @dkarlinsky . I added this code and I feel guilty - I'll take a look this weekend! Hundredfold thanks for providing code!

dkarlinsky:
Sure :) I was hoping to be able to figure out what was going on myself, but after digging through STM internals for a while and feeling out my depth,    I figured I'd provide the test for you or whoever takes this issue. Hope it helps :) 
Very interested to see what the bug ends up being.

LGLO:
I was able to reproduce this bug without TArray but only TRefs:
```
  def test = {
    for {
      r0 <- TRef.makeCommit(0)
      r1 <- TRef.makeCommit(0)
      r2 <- TRef.makeCommit(0)
      r3 <- TRef.makeCommit(0)
[...]

PR #1613, opened 2019-09-07
Issue-1587 STM - ReadOnly journal commits when it's invalid

LGLO:
Fix for #1587
When STM journal is `ReadOnly` in `tryCommit` there is no check if journal was made invalid by concurrent STM

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/stm/StmIssue1587Spec.scala
+++ b/core-tests/jvm/src/test/scala/zio/stm/StmIssue1587Spec.scala
@@ -1,0 +1,36 @@
+package zio.stm
+
+import zio.TestRuntime
+import scala.util.control.Breaks._
+
+class StmIssue1587Spec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime {
+
+  def is = "STM made of read only operations".title ^ s2"""
+     doen't commit when journal is invalid $getOnlyStmCheck
+  """
+
+  def getOnlyStmCheck = {
+    val iterations = 1000000
+    var badResult  = Option.empty[String]
+    breakable {
+      for (i <- 1 to iterations) {
+        val result = unsafeRun(test)
+        if (result != 0 && result != 2) {
+          badResult = Some(s"Result[$i/$iterations] $result")
+          break()
+        }
+      }
+    }
+    badResult aka "badResult" must beNone
+  }
+
+  val test = {
+    for {
+      r0        <- TRef.makeCommit(0)
+      r1        <- TRef.makeCommit(0)
+      sum1Fiber <- r0.get.flatMap(v0 => r1.get.map(_ + v0)).commit.fork
+      _         <- r0.update(_ + 1).flatMap(_ => r1.update(_ + 1)).commit
+      sum1      <- sum1Fiber.join
+    } yield sum1
+  }
+}
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -558,6 +558,10 @@ object STM {
                 Sync(globalLock) {
                   if (isValid(journal)) commitJournal(journal) else loop = true
                 }
+              } else {
+                Sync(globalLock) {
+                  if (isInvalid(journal)) loop = true
+                }
               }
 
             case _ =>


core-tests/jvm/src/test/scala/zio/stm/StmIssue1587Spec.scala
```diff
@@ -0,0 +1,65 @@
+package zio.stm
+
+import zio.{ TestRuntime, ZIO }
+import scala.util.control.Breaks._
+
+class StmIssue1587Spec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime {
```
LGLO: Props for this test to @dkarlinsky!

core-tests/jvm/src/test/scala/zio/stm/StmIssue1587Spec.scala
```diff
@@ -0,0 +1,65 @@
...
+                    r4.update(_ + 1)
+                  }
+                }
+              }
+            }
+            .commit
+      sum1 <- sum1Fiber.join
+    } yield sum1
+  }
+}
```
LGLO: I admit `val test` ~isn't pretty~ is ugly but I wanted to use as small number of functions as possible.

LGLO:
@wi101 Is there any chance that you look into it? I'm freshman to STM implementation.

core-tests/jvm/src/test/scala/zio/stm/StmIssue1587Spec.scala
LGLO: Test is good
Screenshot from 2019-09-08 08-48-18

LGLO:
@jdegoes Maybe you could take a look?

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -558,6 +558,10 @@ object STM {
                 Sync(globalLock) {
                   if (isValid(journal)) commitJournal(journal) else loop = true
                 }
+              } else {
+                Sync(globalLock) {
+                  if (isInvalid(journal)) loop = true
```
jdegoes: I'm surprised this fixes anything, because above, we already have that `analysis ne JournalAnalysis.Invalid`. In other words, it may fix something only non-deterministically, because it happens to be checking "again" to see if it became invalid (and more will be invalid if checked again, so you'll occasionally fix a race issue this way).

Could be the real fix is a bit deeper. 

What are your thoughts?

LGLO: Scenario:
One more try to convince you:
- please take a look at test, only `r0` and `r1` exists. STM that reads them both and other STM that updates them both, racing.
- `stm` being processed here is read-only only - it won't contain `Entry` that `isChanged`.
- the other, updating, `STM` on same `TRef`s, it's commiting at this moment - `r0` has been committed and thread pauses before committing `r1`.
- hence `value = stm.exec(journal)` assigns `Succeed(1)` to `value` and adds `r0` and `r1` to `journal`.
- `val analysis = analyzeJournal(journal)` assigns `ReadOnly` to `analysis` <- I think it's possible, because `stm.exec` started after other STM commited `r0`.
- now the other STM continues to commit and updates `r1.versioned` with `1`
- our `stm` enters `if` in L522, then without my fix it breaks the loop, having `value == Succeed(1)`
- if my fix was in place validity of `journal` would be checked again in critical section - not running concurrently with other thread commiting `TRef`s

LGLO: My proposal adds check of journal validity for ReadOnly journal **inside critical section**.
Journal check outside critical section can only decide that journal is invalid, but for decision that journal is valid, it has to be done inside critical section.

jdegoes:
@LGLO Thank you for your work here!

adamgfraser:
I just saw a failure of the test for this when running locally. It definitely seems to be uncommon because it is the first time I have seen it but we may just have made the problem rare instead of fixed it completely.

jdegoes:
@adamgfraser @LGLO I'll have to do another audit and make sure we didn't miss something... either in the test or the code. Dang it!

LGLO:
That's not good.
@adamgfraser do you have a copy of log of error by any chance?
Are you sure it wasn't test timeout?

I'm currently running test action from spec in a loop and high parallelism:
```scala
object Issue1587 extends zio.App {
  def run(args: List[String]): ZIO[Environment, Nothing, Int] =
    ZIO.sequenceParN(8)(List.fill(8)(test(0))).map(_ => 0)

  def test(i: Int): ZIO[Console, Nothing, Unit] =
    for {
      _         <- if (i % 1000000 == 0) zio.console.putStrLn(i.toString()) else ZIO.succeed(())
      r0        <- TRef.makeCommit(0)
      r1        <- TRef.makeCommit(0)
      sum1Fiber <- r0.get.flatMap(v0 => r1.get.map(_ + v0)).commit.fork
      _         <- r0.update(_ + 1).flatMap(_ => r1.update(_ + 1)).commit
      sum       <- sum1Fiber.join
      _         <- if (sum != 0 && sum != 2) zio.console.putStrLn(s"sum=$sum, i=$i") else ZIO.succeed(())
      _         <- test(i + 1)
    } yield ()
}
```
I'll leave it for few hours and then will run it again but without such parallelism (currently I have 32 fibers for 6 (HT) cores) to have different load on global lock.

adamgfraser:
@LGLO It looks like it was a timeout. Here is the test report. Is there something that could be causing it to not terminate?

```
<?xml version='1.0' encoding='UTF-8'?>
<testsuite hostname="Adams-MacBook.local" name="zio.stm.StmIssue1587Spec" tests="1" errors="0" failures="1" skipped="0" time="67.895" timestamp="2019-09-19T10:56:13">
                     <properties>
      <property name="jline.esc.timeout" value="0"/><property name="jna.platform.library.path" value="/usr/lib:/usr/lib"/><property name="java.runtime.name" value="Java(TM) SE Runtime Environment"/><property name="sun.boot.library.path" value="/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib"/><property name="java.vm.version" value="25.162-b12"/><property name="gopherProxySet" value="false"/><property name="java.vm.vendor" value="Oracle Corporation"/><property name="java.vendor.url" value="http://java.oracle.com/"/><property name="path.separator" value=":"/><property name="jna.loaded" value="true"/><property name="java.vm.name" value="Java HotSpot(TM) 64-Bit Server VM"/><property name="file.encoding.pkg" value="sun.io"/><property name="user.country" value="US"/><property name="sun.java.launcher" value="SUN_STANDARD"/><property name="sun.os.patch.level" value="unknown"/><property name="jna.nosys" value="true"/><property name="java.vm.specification.name" value="Java Virtual Machine Specification"/><property name="user.dir" value="/Users/adamfraser/zio"/><property name="java.runtime.version" value="1.8.0_162-b12"/><property name="java.awt.graphicsenv" value="sun.awt.CGraphicsEnvironment"/><property name="java.endorsed.dirs" value="/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/endorsed"/><property name="os.arch" value="x86_64"/><property name="java.io.tmpdir" value="/var/folders/23/7nn6tlgn1bqflg_ly1yrr3sc0000gn/T/"/><property name="line.separator" value="
"/><property name="java.vm.specification.vendor" value="Oracle Corporation"/><property name="log4j.ignoreTCL" value="true"/><property name="os.name" value="Mac OS X"/><property name="sun.jnu.encoding" value="UTF-8"/><property name="jnidispatch.path" value="/var/folders/23/7nn6tlgn1bqflg_ly1yrr3sc0000gn/T/jna-292228986/jna921781928935233178.tmp"/><property name="java.library.path" value="/Users/adamfraser/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:."/><property name="java.specification.name" value="Java Platform API Specification"/><property name="java.class.version" value="52.0"/><property name="sun.management.compiler" value="HotSpot 64-Bit Tiered Compilers"/><property name="os.version" value="10.14.6"/><property name="http.nonProxyHosts" value="local|*.local|169.254/16|*.169.254/16"/><property name="swoval.tmpdir" value="/Users/adamfraser/.sbt/1.0"/><property name="user.home" value="/Users/adamfraser"/><property name="user.timezone" value="America/New_York"/><property name="java.awt.printerjob" value="sun.lwawt.macosx.CPrinterJob"/><property name="file.encoding" value="UTF-8"/><property name="java.specification.version" value="1.8"/><property name="java.class.path" value="/Users/adamfraser/.sbt/launchers/1.3.0/sbt-launch.jar"/><property name="user.name" value="adamfraser"/><property name="jline.shutdownhook" value="false"/><property name="java.vm.specification.version" value="1.8"/><property name="sun.java.command" value="/Users/adamfraser/.sbt/launchers/1.3.0/sbt-launch.jar shell"/><property name="java.home" value="/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre"/><property name="sun.arch.data.model" value="64"/><property name="user.language" value="en"/><property name="java.specification.vendor" value="Oracle Corporation"/><property name="awt.toolkit" value="sun.lwawt.macosx.LWCToolkit"/><property name="java.vm.info" value="mixed mode"/><property name="java.version" value="1.8.0_162"/><property name="java.ext.dirs" value="/Users/adamfraser/Library/Java/Extensions:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/ext:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java"/><property name="sun.boot.class.path" value="/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/resources.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/rt.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/sunrsasign.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/jsse.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/jce.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/charsets.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/lib/jfr.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_162.jdk/Contents/Home/jre/classes"/><property name="java.vendor" value="Oracle Corporation"/><property name="file.separator" value="/"/><property name="java.vendor.url.bug" value="http://bugreport.sun.com/bugreport/"/><property name="sun.io.unicode.encoding" value="UnicodeBig"/><property name="sun.cpu.endian" value="little"/><property name="socksNonProxyHosts" value="local|*.local|169.254/16|*.169.254/16"/><property name="ftp.nonProxyHosts" value="local|*.local|169.254/16|*.169.254/16"/><property name="sun.cpu.isalist" value=""/>
    </properties>
                     <testcase classname="zio.stm.StmIssue1587Spec" name="doen't commit when journal is invalid" time="67.895">
                                                 <failure message="TIMEOUT: 60 seconds" type="java.lang.Exception">java.lang.Exception: TIMEOUT: 60 seconds
	at zio.TestRuntime.around(TestRuntime.scala:20)
</failure>
                                               </testcase>
                     <system-out><![CDATA[]]></system-out>
                     <system-err><![CDATA[]]></system-err>
                   </testsuite>

LGLO:
That's good. I usually runs few seconds but having it run 10x longer seems possible.
Meanwhile my tests didn't reveal bug in few hours of running. It doesn't mean there is no bug, ofc.

I'm migrating specs for STM to zio-test. When moving this one, I'll make this test without loop, but with barriers, so I shouldn't require so much running time. I'll make sure it fails when my recent fix is not present.

PR #1616, opened 2019-09-07
ZIO Test: Support Generators for Functions

### LINKED ISSUE #1472: ZIO Test: Support Generators for Functions

adamgfraser:
Adds `Gen#function` for generating functions from `A` to `B` given a generator of `B` values.

```scala
final def function[R, A, B](gen: Gen[R, B]): Gen[R, A => B]
```

We do this by evaluating the `Gen[R, B]` for each input and memoizing the results to ensure that the generated function is referentially transparent. Since in some cases the `A` value will not implement `hashCode` in a way that is consistent with equality I also added a variant `functionWith` that allows specifying a hashing function. This is all backed by a new type `Fun` that wraps an effectual functional and memoizes its results.

We could potentially reify the hashing function as the `Cogen` type that exists in other testing frameworks. I didn't do that for now because it is not necessary as long as the `A` type implements `hashCode` in a way that is consistent with your definition of equality. But if people think that would be helpful we could definitely add it along with some nice methods for composing them.

Resolves #1472.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Fun.scala
+++ b/test/shared/src/main/scala/zio/test/Fun.scala
@@ -1,0 +1,42 @@
+package zio.test
+
+import java.util.concurrent.ConcurrentHashMap
+
+import zio.ZIO
+
+/**
+ * A `Fun[A, B]` is a referentially transparent version of a potentially
+ * effectual function from `A` to `B`. Each invocation of the function will be
+ * memoized so the function is guaranteed to return the same value for any
+ * given input. The function should not involve asynchronous effects.
+ */
+final case class Fun[-A, +B] private (private val f: A => B, private val hash: A => Int) extends (A => B) {
+
+  final def apply(a: A): B = map.computeIfAbsent(hash(a), _ => f(a))
+
+  private[this] final val map = new ConcurrentHashMap[Int, B]()
+}
+
+object Fun {
+
+  /**
+   * Constructs a new `Fun` from an effectual function. The function should not
+   * involve asynchronous effects.
+   */
+  final def make[R, A, B](f: A => ZIO[R, Nothing, B]): ZIO[R, Nothing, Fun[A, B]] =
+    makeHash(f)(_.hashCode)
+
+  /**
+   * Constructs a new `Fun` from an effectual function and a hashing function.
+   * This is useful when the domain of the function does not implement
+   * `hashCode` in a way that is consistent with equality.
+   */
+  final def makeHash[R, A, B](f: A => ZIO[R, Nothing, B])(hash: A => Int): ZIO[R, Nothing, Fun[A, B]] =
+    ZIO.runtime[R].map(runtime => Fun(a => runtime.unsafeRun(f(a)), hash))
+
+  /**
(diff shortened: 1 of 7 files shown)


ghostdogpr:
Any idea what’s going on with the ScalaJS build? Do you have that locally?

adamgfraser:
Yes. Despite implementing `java.util.concurrent.ConcurrentHashMap`, the ScalaJS version does not implement the `computeIfAbsent` method. I'm working on a thin wrapper around it now to resolve.

jdegoes:
@adamgfraser Amazing work. I never expected to have this done so quickly!

test/shared/src/main/scala/zio/test/Fun.scala
```diff
@@ -0,0 +1,40 @@
+package zio.test
```
jdegoes: License header ^^^

test/shared/src/main/scala/zio/test/Fun.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import zio.ZIO
+
+/**
+ * A `Fun[A, B]` is a referentially transparent version of a potentially
+ * effectual function from `A` to `B`. Each invocation of the function will be
+ * memoized so the function is guaranteed to return the same value for any
+ * given input. The function should not involve asynchronous effects.
+ */
+final case class Fun[-A, +B] private (private val f: A => B, private val hash: A => Int) extends (A => B) {
```
jdegoes: We should make this and the companion object package private, so that it doesn't appear in the Scaladoc (unless it needs to be public?).

test/shared/src/main/scala/zio/test/Fun.scala
adamgfraser: Yes.

test/shared/src/main/scala/zio/test/Fun.scala
adamgfraser: We can make it private. I thought a next step could be exploring building this out into something along the lines of shrinkable and showable functions in QuickCheck, though even then I'm not sure we would need to expose it and we could always do it later.

PR #1617, opened 2019-09-08
ZIO Test: Improve Reporting of Property Based Testing Result

### LINKED ISSUE #1585: ZIO Test: Improve Reporting of Property Based Testing Result
This ticket is to improve reporting of property based testing results. A completed ticket will make two or possibly three improvements to existing test reporting:

- [ ] Report the shrunk random variables that caused the test to fail in addition to the failing test result. This will involve modifying the `checkStream` method in `CheckVariants.scala` so instead of just mapping the test over the random sample we keep a tuple of the sample and the test result and shrink those.
- [ ] Show the number of tests that passed before the failing test. This can also be done within the `checkStream` function, probably by using something like `zipWithIndex`.

A nice to have would be:
- [ ] Show the random seed for failing property based tests. This will be a little tricky because property based tests don't necessarily need to use a "random" variable (all the check methods are generic in `R` and could be done with an exhaustive sample). So we would need a way to at compile time identify which ones do and then access the `Random` to get the seed.

ghostdogpr:
Fixes #1585 

Example output:
```
[info]   - test
[info]     Test failed after 3 iterations with input: (0,0)
[info]     Original input before shrinking was: (-396148549,-1142042791)
[info]     0 did not satisfy isGreaterThan(0)
```

### DIFF
--- a/test/shared/src/main/scala/zio/test/CheckVariants.scala
+++ b/test/shared/src/main/scala/zio/test/CheckVariants.scala
@@ -208,9 +208,15 @@ trait CheckVariants {
   private final def checkStream[R, E, A](stream: ZStream[R, Nothing, Sample[R, A]], maxShrinks: Int = 1000)(
     test: A => ZIO[R, E, TestResult]
   ): ZIO[R, E, TestResult] =
-    stream
-      .mapM(_.traverse(test(_).either))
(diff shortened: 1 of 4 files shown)


adamgfraser:
This is great! This is going to make reporting of property based tests a lot nicer!

PR #1619, opened 2019-09-08
CI: Replace ppa with adoptopenjdk artifactory cloud, add timeout

yoohaemin:
Follow up to #1578 
There was at least one case [here](https://github.com/zio/zio/pull/1578#issuecomment-528485926) where the downloads from the canonical PPAs were too slow, and they aren't exactly very fast in most cases either. 

This attempts to fix it by relying on a different jdk package from adoptopenjdk. It seems to be hosted in Google's servers somewhere in US East, which should be faster than downloading from Canonical's servers in the UK.

Also, add timeout so that it fails early when the build is taking too long, rather than waste precious time from contributors (and money from sponsors).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,9 +36,17 @@ install_jdk: &install_jdk
         deb http://mirror.math.princeton.edu/pub/ubuntu/ xenial-security main restricted universe
         EOF
 
-        if [ $JDK_VERSION == 11 ]; then sudo add-apt-repository ppa:openjdk-r/ppa -y; fi
-        sudo apt update
-        sudo apt install openjdk-${JDK_VERSION}-jdk -y
+        if [ $JDK_VERSION == 11 ]; then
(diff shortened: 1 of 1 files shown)


jdegoes:

What about security considerations? Who runs adoptionjdk?

yoohaemin:
@jdegoes AdoptOpenJDK is a community effort for building and distributing OpenJDK. It has a TSC (Technical Steering Committee), see [here](https://github.com/AdoptOpenJDK/TSC) and here for the list of TSC [members](https://github.com/AdoptOpenJDK/TSC#tsc-members). The build scripts are all public at github, the images themselves are widely adopted and popular (1M+ pulls from docker hub), so I find it trustworthy enough for general uses. We (me and @ghostdogpr) actually use it in production.

Also, the repository I added in the PR is the one run by them, and not by a 3rd party. 

The downside is that it is not endorsed by any company, nor has it passed the Java TCK (due to licensing reasons). If you would prefer to rely on something that's been around for longer and is backed by a company, Azul's Zulu build also has a public repository that should be fast too.

PR #1625, opened 2019-09-08
ZIO Test: Add Size Combinators

### LINKED ISSUE #1601: ZIO Test: Improve Use of Size Parameter in Tests
Split out from #1585 for clarity. Currently sized collection generators generate collections of size between 0 and the `size` parameter, currently 100. This results in too many large tests, slowing down testing, and not enough small tests, creating issues with coverage for cases like empty or small sets. As discussed on Gitter, we can address this by creating a combinator the uses the size parameter to create an exponential distribution of sizes up to the maximum size so we get mostly small cases but some large cases.

adamgfraser:
Adds `small`, `medium`, and `large` sized combinators for generators to generate a distribution of sizes based on the size parameter. The `large` combinator mimics the current behavior and with the size configured at `100` generates uniform sizes between `0` and `100`. The `medium` combinator uses an exponential distribution and so generate values between `0` and `100` but with most values towards the smaller end of the range and some larger values. The `small` combinator uses a logarithmic distribution and generates values between `0` and `7` with the default configuration.

I changed the collection generators like `listOf` to use new `medium` combinator. I think this should be a win-win in terms of performance and test coverage. We will be generating fewer larger collections which should speed up tests. And there are a lot of bugs that can be easier to catch with smaller collections that I don't think we weren't testing enough before.

Resolves #1601.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -21,6 +21,7 @@ import scala.collection.immutable.SortedMap
 import zio.{ UIO, ZIO }
 import zio.random._
 import zio.stream.{ Stream, ZStream }
+import zio.test.MathUtils._
 
 /**
  * A `Gen[R, A]` represents a generator of values of type `A`, which requires
@@ -61,6 +62,21 @@ case class Gen[-R, +A](sample: ZStream[R, Nothing, Sample[R, A]]) { self =>
 
   final def map[B](f: A => B): Gen[R, B] =
     Gen(sample.map(_.map(f)))
+
+  /**
+   * Maps an effectual function over a generator.
+   */
+  final def mapM[R1 <: R, B](f: A => ZIO[R1, Nothing, B]): Gen[R1, B] =
+    Gen(sample.mapM(_.traverse(f)))
+
+  /**
+   * Discards the shrinker for this generator and applies a new shrinker by
+   * mapping each value to a sample using the specified function. This is
+   * useful when the process to shrink a value is simpler than the process used
+   * to generate it.
+   */
+  final def reshrink[R1 <: R, B](f: A => Sample[R1, B]): Gen[R1, B] =
+    Gen(sample.map(sample => f(sample.value)))
 
   final def zip[R1 <: R, B](that: Gen[R1, B]): Gen[R1, (A, B)] =
     self.flatMap(a => that.map(b => (a, b)))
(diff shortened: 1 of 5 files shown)


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -62,6 +63,21 @@ case class Gen[-R, +A](sample: ZStream[R, Nothing, Sample[R, A]]) { self =>
...
+  final def mapM[R1 <: R, B](f: A => ZIO[R1, Nothing, B]): Gen[R1, B] =
+    Gen(sample.mapM(_.traverse(f)))
+
+  /**
+   * Discards the shrinker for this generator and applies a new shrinker by
+   * mapping each value to a sample using the specified function. This is
+   * useful when the process to shrink a value is simpler than the process used
+   * to generate it.
+   */
+  final def reshrink[R1 <: R, B](f: A => Sample[R1, B]): Gen[R1, B] =
```
jdegoes: Nice thinking. 👍

test/shared/src/main/scala/zio/test/MathUtils.scala
```diff
@@ -0,0 +1,63 @@
...
+      if (i <= 1) j
+      else loop(i >> 1, j + 1)
+    loop(n, 0)
+  }
+
+  /**
+   * Returns two raised to the specified power. If the specified power is
+   * negative returns zero.
+   */
+  final def pow2(n: Int): Int = {
```
jdegoes: How about `2.pow(n)`?

test/shared/src/main/scala/zio/test/MathUtils.scala
```diff
@@ -0,0 +1,63 @@
...
+ * `MathUtils` provides common math functions used in ZIO Test.
+ */
+private[test] object MathUtils {
+
+  /**
+   * Returns the base two logarithm of the specified integer, rounding up to
+   * the nearest integer. If the specified integer is zero or negative returns
+   * zero.
+   */
+  final def log2Ceil(n: Int): Int = {
```
jdegoes: Nothing like this in Scala / Java math lib?

test/shared/src/main/scala/zio/test/MathUtils.scala
```diff
@@ -0,0 +1,63 @@
...
+      else loop(i >> 1, j + 1)
+    loop(n - 1, 0)
+  }
+
+  /**
+   * Returns the base two logarithm of the specified integer, rounding down to
+   * the nearest integer. If the specified integer is zero or negative returns
+   * zero.
+   */
+  final def log2Floor(n: Int): Int = {
```
jdegoes: Can also compose logs and floor, right? e.g. [here](https://www.techiedelight.com/calculate-log-base-2-in-java/)

adamgfraser:
Yes I think you are right about the math combinators. I went down that road because all the logarithm and power operators use `Double` values so I thought since we knew we only cared about integers we could be more efficient by doing bitwise operations. But doing some informal benchmarking that doesn't appear to be the case. I'll go back to using the native methods.

jdegoes:
Looks great!

PR #1626, opened 2019-09-08
ZIO Test: Enable Assertion#equalTo for Arrays

adamgfraser:
As pointed out by @regiskuckaertz, `Assertion#equalTo` was not working for arrays due to `equalTo` being implemented in terms of universal equality and arrays using reference versus value equality. Adds a pattern match to determine if both values are arrays and if so checks equality using `sameElements`.  It is a little ad hoc but I don't know how we do it in a more systematic way without some kind of equality type class that I think is out of scope. The alternative would be we could just say that `equalTo` is not supported for arrays and provide a `sameElements` assertion.


jdegoes:
Looks great!

PR #1628, opened 2019-09-09
Fix Bug in Sample#shrinkSearch and Add ZStream#takeUntil and ZStream#dropUntil

adamgfraser:
Fixes a bug in shrinking logic for property based testing that could lead to very slow tests in some cases. Previously, we had a parameter `maxShrinks = 1000`. However, `shrinkSearch` was producing a stream of failing shrinks and then we were taking 1,000 elements from that. In some cases where failures were rare, this could mean that `shrinkSearch` actually explored much more than 1,000 potential values (@ghostdogpr was reporting more than 5,000 in one case). This changes that to have `shrinkSearch` return a stream of all potential shrinks explored, whether or not they are failures, and then the client, in this case the `checkStream` method, can take 1,000 of those and filter them for failures.

To support this, I also implemented a `takeUntil` method on `ZStream` that basically takes one more value than `takeWhile` would. So `ZStream("A", "A", "A", "B", "B", "B").takeUntil(_ == "B") == ZStream("A", "A", "A", "B")`. This has bounced around discussions about the Scala collections library as either `takeUntil` or `takeTo` so I'm open to naming suggestions. I think this method is pretty useful for something like `geolocationData.takeUntil(_.precision < x)` to give you a stream with a value satisfying your predicate if one exists and all the intermediate results leading up to it, and I don't think it is easily implemented in terms of existing combinators. I also added a `dropUntil` method for symmetry but I think it is much less useful and is easily implemented in terms of `dropWhile` and `drop`.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -65,6 +65,10 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
 
   Stream.drain              $drain
 
+  Stream.dropUntil
+    dropUntil         $dropUntil
+    short circuits    $dropUntilShortCircuiting
+
   Stream.dropWhile
     dropWhile         $dropWhile
     short circuits    $dropWhileShortCircuiting
@@ -199,6 +203,8 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
     take short circuits      $takeShortCircuits
     take(0) short circuits   $take0ShortCircuitsStreamNever
     take(1) short circuits   $take1ShortCircuitsStreamNever
+    takeUntil                $takeUntil
+    takeUntil short circuits $takeUntilShortCircuits
     takeWhile                $takeWhile
     takeWhile short circuits $takeWhileShortCircuits
 
@@ -608,6 +614,24 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
         l   <- ref.get
       } yield l.reverse must_=== (0 to 10).toList
     )
+
+  private def dropUntil = {
+    def dropUntil[A](as: List[A])(f: A => Boolean): List[A] =
+      as.dropWhile(!f(_)).drop(1)
+    prop { (s: Stream[String, Byte], p: Byte => Boolean) =>
+      unsafeRunSync(s.dropUntil(p).runCollect) must_=== unsafeRunSync(s.runCollect.map(dropUntil(_)(p)))
+    }
+  }
+
+  private def dropUntilShortCircuiting =
+    unsafeRun {
+      (Stream(1) ++ Stream.fail("Ouch"))
+        .take(1)
+        .dropUntil(_ => false)
+        .runDrain
+        .either
+        .map(_ must beRight(()))
+    }
 
   private def dropWhile =
     prop { (s: Stream[String, Byte], p: Byte => Boolean) =>
@@ -1570,6 +1594,25 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
       for {
         ints <- (Stream(1) ++ Stream.never).take(1).run(Sink.collectAll[Int])
       } yield ints must_=== List(1)
+    )
+
+  private def takeUntil = {
+    def takeUntil[A](as: List[A])(f: A => Boolean): List[A] =
+      as.takeWhile(!f(_)) ++ as.dropWhile(!f(_)).take(1)
+    prop { (s: Stream[String, Byte], p: Byte => Boolean) =>
+      val streamTakeWhile = unsafeRunSync(s.takeUntil(p).runCollect)
+      val listTakeWhile   = unsafeRunSync(s.runCollect.map(takeUntil(_)(p)))
+      listTakeWhile.succeeded ==> (streamTakeWhile must_=== listTakeWhile)
+    }
+  }
+
+  private def takeUntilShortCircuits =
+    unsafeRun(
+      (Stream(1) ++ Stream.fail("Ouch"))
+        .takeUntil(_ => true)
+        .runDrain
+        .either
+        .map(_ must beRight(()))
     )
 
   private def takeWhile =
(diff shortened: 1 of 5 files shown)


streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -609,6 +615,24 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
...
+      as.dropWhile(!f(_)).drop(1)
+    prop { (s: Stream[String, Byte], p: Byte => Boolean) =>
+      unsafeRunSync(s.dropUntil(p).runCollect) must_=== unsafeRunSync(s.runCollect.map(dropUntil(_)(p)))
+    }
+  }
+
+  private def dropUntilShortCircuiting =
+    unsafeRun {
+      (Stream(1) ++ Stream.fail("Ouch"))
+        .take(1)
```
regiskuckaertz: you can take that one out, can't you?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1681,6 +1688,26 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+  /**
+   * Takes all elements of the stream until the specified predicate evaluates
+   * to `true`.
+   */
+  def takeUntil(pred: A => Boolean): ZStream[R, E, A] =
+    ZStream[R, E, A] {
+      for {
+        as            <- self.process
+        keepTakingRef <- Ref.make(true).toManaged_
+        pull = keepTakingRef.get.flatMap { p =>
```
regiskuckaertz: does this action need to be atomic? if yes then it might be useful to use `modify` instead

test/shared/src/main/scala/zio/test/CheckVariants.scala
```diff
@@ -222,7 +222,10 @@ trait CheckVariants {
       .run(ZSink.collectAll[Either[E, TestResult]]) // Collect all the shrunken failures
       .flatMap { failures =>
         // Get the "last" failure, the smallest according to the shrinker:
-        failures.reverse.headOption
+        failures
+          .filter(_.fold(_ => true, _.isFailure))
+          .reverse
```
regiskuckaertz: `lastOption`?

adamgfraser: Yes.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
adamgfraser: Actually, looking at it I am wondering if there is an issue with this and the existing `dropWhile` test. If we are only taking the first element we're never going to see the failure so we're not testing anything about the drop function, right?

regiskuckaertz: Hmm, yeah I agree 😄

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: I thought we were okay because we were inside the managed from `process`  but I'm not 100% sure. @iravid or @vasilmkd?

iravid: All bets are off with regard to interruption of a `Pull`.

If you can do things atomically that's better, but basically everything breaks if you interrupt a `Pull`.

adamgfraser: Okay. It looks like we use the same `get` and `set` pattern in the existing implementation of `dropWhile`.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
adamgfraser: So thinking about this more I think the entire `dropWhile` short circuiting property is not sound. `DropWhile` has to look at each element to determine if it should drop it, and all the elements it doesn't look at are by definition not dropped. The only way to get some kind of "short circuiting" would be if `dropWhile` looked at each value to see if it was an error and then dropped those two while it was operating, which seems inconsistent with the other combinators. `takeWhile` makes sense because you are only taking a certain number of values, so of course you would ignore any errors that existed in values past the ones you were taking.

adamgfraser: The `filter` short circuiting properties have the same issue.

adamgfraser:
@regiskuckaertz Okay so addressing your comments I did the `lastOption` one, I think we leave the `Ref` as is since it is analogous to `TakeWhile`, and I am going to open a separate issue on the ZStream tests and leave it as is here. Sound good?

iravid:
The weird short circuiting tests are artifacts of the old encoding where we had to take extra care to short circuit everything. So can probably be dropped where they don't make sense :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: We could, rewrite all cases where we can do things atomically. But yes, it's as grim as @iravid puts it. And at that point, I'm not sure if the extra work is worth it.

adamgfraser:
@iravid Thanks! There are a lot of them so I probably won't mess with the existing ones in this PR but I won't add new ones with respect to these methods.

vasilmkd:
> The weird short circuiting tests are artifacts of the old encoding where we had to take extra care to short circuit everything. So can probably be dropped where they don't make sense :-)



> @iravid Thanks! There are a lot of them so I probably won't mess with the existing ones in this PR but I won't add new ones with respect to these methods.

I also think they don't make sense anymore.

ghostdogpr (approved):

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -729,6 +729,13 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
   final def drop(n: Int): ZStream[R, E, A] =
     self.zipWithIndex.filter(_._2 > n - 1).map(_._1)
 
+  /**
+   * Drops all elements of the stream until the specified predicate evaluates
+   * to `true`.
+   */
+  def dropUntil(pred: A => Boolean): ZStream[R, E, A] =
+    dropWhile(!pred(_)).drop(1)
```
joroKr21, line 737: That's a really confusing implementation - it doesn't match the documentation.
"Until" in programming usually means not including when the predicate evaluates to `true`.
In fs2 this methods is called `dropThrough` and resp. `takeThrough`.

PR #1635, opened 2019-09-10
ZIO Test: Enable Tracing for Test Execution

adamgfraser:
Very minor follow up to #1633.  Previously we disabled tracing for random variable generation to improve performance. This PR just reenables tracing for the effect of actually evaluating the test once the random variables have been generated. I think this is nice because it gives the user the default tracing behavior with respect to the effects they create, we just aren't tracing our internal effects. Sorry for the multiple PRs, didn't think of this earlier.

Copying @ghostdogpr.


ghostdogpr:
Thanks!

PR #1638, opened 2019-09-10
Add Assertion.dies and Gen.anyString

regiskuckaertz:
Taken from #1595 

- `Assertion.dies` helps with checking the unchecked error channel in a `Cause[E]`.
- `Gen.anyString` seems like a recurring use case for a generator (perhaps `asciiString` too?)

cc @adamgfraser

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -142,6 +142,12 @@ object Gen {
     }
 
   /**
+   * A generator of shorts. Shrinks towards the empty string.
+   */
+  final val anyString: Gen[Random with Sized, String] =
+    Gen.string(Gen.anyChar)
+
+  /**
    * A generator of booleans. Shrinks toward 'false'.
    */
   final val boolean: Gen[Random, Boolean] =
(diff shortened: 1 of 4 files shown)


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -141,6 +141,12 @@ object Gen {
         .map(Sample.shrinkIntegral(0))
     }
 
+  /**
+   * A generator of shorts. Shrinks towards the empty string.
```
adamgfraser: Should be "strings".

test/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -68,6 +70,14 @@ object AssertionSpec {
       assert(Exit.fail("Other Error"), fails(equalTo("Some Error"))),
       message = "fails must fail when error value does not satisfy specified assertion"
     ),
+    testSuccess(
+      assert(Exit.die(someException), dies(equalTo(someException))),
+      message = "dies must succeed when exception satisfy specified assertion"
+    ),
+    testFailure(
+      assert(Exit.die(new RuntimeException("Bam!")), dies(equalTo(someException))),
+      message = "fails must fail when exception does not satisfy specified assertion"
```
adamgfraser: Should be "dies" here.

regiskuckaertz: yes

PR #1639, opened 2019-09-10
ZIO Test: Implement Showable Functions

adamgfraser:
Since our implementation of randomly generated functions is backed by a `Map`, we can render these functions as mappings between their inputs and outputs. We also get a good amount of shrinking "for free" in many cases as we minimize the inputs to the function. A next step would be adding additional shrinking logic for functions but that is not implemented here.

Sample output is below.

```scala
object MySpec
    extends DefaultRunnableSpec(
      suite("MySpec")(
        testM("languages") {
          check(genStringBoolFn) { p =>
            assert(p("Scala"), equalTo(p("Haskell")))
          }
        },
        testM("mapFilter") {
          check(genIntList, genIntIntFn, genIntBoolFn) { (as, f, p) =>
            assert(as.map(f).filter(p), equalTo(as.filter(p).map(f)))
          }
        }
      )
    )
```

```
[info] running MySpec 
- MySpec
  - languages
    Test failed after 1 iteration with input: Fun(Scala -> false, Haskell -> true)
    false did not satisfy equalTo(true)
  - mapFilterM / Compile / compileIncremental 0s
    Test failed after 1 iteration with input: ((List(-8),Fun(-8 -> 10)),Fun(-8 -> false, 10 -> true))
    Original input before shrinking was: ((List(-10, 8, 9, -1, 6, 2, -3, -6),Fun(-1 -> 3, 2 -> 1, -3 -> -6, -6 -> 7, 6 -> 5, 8 -> 7, -10 -> 6, 9 -
> 10)),Fun(-1 -> true, 1 -> false, 2 -> true, -3 -> true, 3 -> false, -6 -> false, 5 -> true, 6 -> false, 7 -> true, 8 -> false, -10 -> true, 9 -> true, 10 -> false))
    List(10) did not satisfy equalTo(List())
Ran 2 tests in 2 s 187 ms: 0 succeeded, 0 ignored, 2 failed
```


jdegoes:
@adamgfraser Wow. Really great idea.

PR #1640, opened 2019-09-11
ZIO Test: Default Sized Generators to Small

adamgfraser:
Following up on feedback from @jdegoes, changes the default sized combinator for sized generators from `medium` to `small`.  Also changes the `small` generator so that nine times out of ten it gives the same logarithmic distribution as before but the other time it uses a uniform distribution up to half the max size parameter. The idea is to strongly concentrate values on smaller sizes but still occasionally test larger sizes. This should speed up tests, especially in cases with complex types where generating large sized values for individual fields had little value.

I've included a sample run below.

```
5
2
0
7
3
7
7
4
5
5
9
3
6
2
1
5
6
3
1
0
24
1
3
2
5
4
4
5
1
4
2
1
1
0
1
7
7
5
4
5
4
7
0
7
12
22
5
5
7
6
42
5
1
4
28
1
2
7
2
1
4
5
4
1
0
6
24
6
5
4
6
0
0
46
0
3
5
6
2
3
2
3
5
0
0
29
7
7
3
6
11
7
6
5
2
7
2
39
4
0
1
45
2
6
0
3
1
1
5
2
2
3
5
5
28
7
4
3
3
4
5
2
1
2
3
7
44
0
6
3
2
2
5
7
2
0
2
6
4
5
3
7
3
5
3
4
7
0
4
2
2
3
2
7
5
5
2
1
4
0
1
0
1
27
44
5
4
6
0
6
3
5
7
5
1
16
6
32
7
3
6
2
2
7
2
5
2
6
49
6
7
4
4
1
0
2
3
27
5
0
```

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -279,13 +279,13 @@ object Gen {
    * large number of larger sizes will be generated.
    */
   final def large[R <: Random with Sized, A](f: Int => Gen[R, A], min: Int = 0): Gen[R, A] =
-    size.flatMap(max => Gen.int(min, max)).flatMap(f)
+    size.flatMap(max => int(min, max)).flatMap(f)
 
   final def listOf[R <: Random with Sized, A](g: Gen[R, A]): Gen[R, List[A]] =
-    medium(listOfN(_)(g))
+    small(listOfN(_)(g))
 
   final def listOf1[R <: Random with Sized, A](g: Gen[R, A]): Gen[R, List[A]] =
-    medium(listOfN(_)(g), 1)
+    small(listOfN(_)(g), 1)
 
   final def listOfN[R <: Random, A](n: Int)(g: Gen[R, A]): Gen[R, List[A]] =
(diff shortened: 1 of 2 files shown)


adamgfraser:
I wonder if we should just keep the old version of `small` and use that as the default.

jdegoes:
I think exponential falloff is fine, it's just a matter of tweaking parameters.

I imagined `small` giving us something like this:

Plot-of-the-exponential-falloff-of-the-maximum-degree-of-polarisation-for-turbidities-in

Is that what the old `small` did?

adamgfraser:
@jdegoes Thanks! That is helpful. No, the old `small` generated exclusively values in that 0 to 10 range. What you want looks I think more like the current `medium` with a scaling factor. What parameters did you use to generate that? I may be able to just do the same thing or calibrate to do something similar. I will put together a chart with the probability distributions for all three combinators so we can have a better view of this.

adamgfraser:
I tweaked the parameters to be more in line with your suggestions. `small` is exponential distribution with mean of 4 (decay factor of 0.25) and medium is exponential distribution with mean of 10 (decay factor of 0.1). Graph is below.

image

PR #1655, opened 2019-09-12
CancelableFuture

### LINKED ISSUE #1627: Proposal: CancelableFuture
Proposal: Adding a `CancelableFuture`, which extends `Future` with a `cancel` operation that triggers interruption of the underlying `IO`.

### Use case

> We've started introducing zio at the edges of our system. This results in having multiple `unsafeRunToFutures` running in different actors. I'm looking for advice on terminating these from outside of zio (e.g. when the parent actor is shutting down).

(reported by @dpennell)

### API

1. Introduce a `CancelableFuture`

```scala
trait CancelableFuture[A] extends scala.concurrent.Future {
  def cancel(): Unit
}
```

or (see https://github.com/zio/zio/issues/1627#issuecomment-529606773)

---

```scala
trait CancelableFuture[E, A] extends scala.concurrent.Future {
  def cancel: `UIO[Exit[E, A]]`
}
```

2. Return `CancelableFuture` in `ZIO` and the RTS

```diff
-- final def toFuture(implicit ev2: E <:< Throwable): ZIO[R, Nothing, scala.concurrent.Future[A]]
++ final def toFuture(implicit ev2: E <:< Throwable): ZIO[R, Nothing, zio.CancelableFutureA]]
```

### Implementation

* AFAICT all `Future` methods must be reimplemented, most will most likely simply delegate
* Interop code with Cats-Effect will most likely have code of very similar shape (for `ConcurrentEffect` `runCancelable`)

fsvehla:
Closes #1627

Delegates calls to a `Future` that it wraps. I’ve found a lightweight way to support Scala 2.11 as well.


dpennell:
Is there anything I can do to help with this?

fsvehla:
It will be ready soon. I am at the hackathon and will wrap it up tomorrow

dpennell:
@fsvehla I'm jealous - wish I could be there!

fsvehla:
addressed all comments.

core/shared/src/main/scala/zio/Cancelable.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+trait Cancelable[+E, +A] {
```
jdegoes: This is only used in `Cancelable`, right? In that case, I'd inline it completely so we can delete the class.

fsvehla: @jdegoes You were actually correct, there is `java.util.Cancelable`

```scala
def cancel(mayInterruptIfRunning: Boolean): Boolean
def isCancelled(): Boolean
def isDone(): Boolean
```

Let’s use that.

fsvehla: That JCP was never accepted,

fsvehla:
Inlined.

core/shared/src/main/scala/zio/CancelableFuture.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scala.concurrent.{ CanAwait, ExecutionContext, Future }
+import scala.concurrent.duration.Duration
+import scala.util.Try
+
+private[zio] abstract class CancelableFuture[+E, +A](val future: Future[A])
+    extends Future[A]
+    with FutureTransformCompat[A] {
+
+  def cancel: Exit[E, A]
```
neko-kai: Maybe this could return another `Future` instead of blocking the current thread?

core/shared/src/main/scala/zio/CancelableFuture.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package zio
+
+import scala.concurrent.{ CanAwait, ExecutionContext, Future }
+import scala.concurrent.duration.Duration
+import scala.util.Try
+
+private[zio] abstract class CancelableFuture[+E, +A](val future: Future[A])
```
neko-kai: We might want to make it public so that users can use the additional cancellation ability!

core/shared/src/main/scala/zio/CancelableFuture.scala
fsvehla: I have no strong opinion here, but see
https://github.com/zio/zio/pull/1655#discussion_r323885081

neko-kai: Well, if we go back to the original use case:
> We've started introducing zio at the edges of our system. This results in having multiple `unsafeRunToFutures` running in different actors. I'm looking for advice on terminating these from outside of zio (e.g. when the parent actor is shutting down).

I'm fairly sure introducing a potentially never-ending blocking operation inside an akka actor is a big no-no. Note that `Future`/`scala.Promise` is also impure and it's already running the moment it's returned, so it's no less ergonomic to use it in impure code. In fact, I'd expect that most users will not care about the `Exit` return AND might not want to wait until the fiber is interrupted, in which case the default `run-and-forget` behavior of `cancel` without `Await` is preferrable.

Note that there would be no way for a user to achieve `run-and-forget` without blocking a thread if we commit to a synchronous API.

fsvehla: 👍, Kai. My initial idea was to return an `UIO[Exit[E, A]]`, because the runtime would have been around anyway at the call site where the `CancelableFuture` is created. Returning a `Future` sounds like a good compromise

fsvehla:
Addressed comments

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -267,17 +267,20 @@ trait Fiber[+E, +A] { self =>
...
       def failure(cause: Cause[E]): UIO[p.type] = UIO(p.failure(cause.squashWith(f)))
       def success(value: A): UIO[p.type]        = UIO(p.success(value))
 
-      UIO.effectTotal(p.future) <* self.await
+      ZIO.runtime[Any].map { runtime =>
+        new CancelableFuture[E, A](p.future) {
+          def cancel: Future[Exit[E, A]] = runtime.unsafeRunToFuture(interrupt)
+        }
+      } <* self.await
         .flatMap[Any, Nothing, p.type](_.foldM[Any, Nothing, p.type](failure, success))
```
neko-kai: I wonder if type annotations could be avoided if `concurrent.Promise[A]` is used in `success/failure` returns instead of `p.type`

jdegoes:
@fsvehla Thank you for all your work on this feature, and congratulations on your first ZIO contribution! 

giphy

dpennell:
ZIO team - thanks for addressing this!

PR #1664, opened 2019-09-13
ZIO Test: Support Managed Resources Per Suite

### LINKED ISSUE #1632: ZIO Test: Support Managed Resources Per Suite
As discussed with @iravid we need a solution to be able to provide a managed resource to an entire suite. I spent some time on it tonight and struggled with it. The challenge as I see it is our core type is essentially (simplified a bit):

```scala
sealed trait ZSpec[-R, +E, +L, +S]
case class SuiteCase(label: L, specs: Vector[ZSpec[R, E, L, S]]) extends ZSpec[R, E, L, S]
case class TestCase(label: L, test: ZIO[R, E, S])
```

So right now each test is itself an effect, which gives us a very natural way of describing effectual transformations at the per test level, which we have used to implement the test aspects we have so far. But at the suite level a suite is just a collection of individual effectual tests, we don't have a way to model the entire set of them as being effectual and do something like provide a single resource to all of them. We could change the signature to something like `specs: ZIO[R, E, Vector[ZSpec[R, E, L, S]]]` and I think that would give us the ability to write suite level test aspects. I think that would require parameterizing `Spec` on `ZIO` in the same way.

@jdegoes Do you have other ideas? I think this is our top priority for ZIO Test right now.

adamgfraser:
Implements a new `ManagedRunnableSpec`, which takes a `Managed` as a parameter in addition to a `Spec` and provides the managed resource once to the entire suite. 

Resolves #1632.


iravid:
This can definitely work, but are we sure we want to go down the route of requiring a different "base class" for this sort of usecase?

adamgfraser:
@iravid I hear you. I think we are limited by our current encoding right now because individual tests are effects but suites aren't. So if I want to perform effects across an entire suite I have to modify the logic of the executor that is executing the tests, instead of just being able to operate on the tests themselves.

I think this is definitely an indicator that we may not have the optimal encoding yet. I think the solution would be to make suites themselves effectual in addition to tests, but I get the sense that @jdegoes is not bought into this. So I am conscious that you need a solution to this problem and I don't want to hold you up when I don't have a clear path forward on a better approach.

jdegoes:
@adamgfraser @iravid I am not a fan of new specs.

I think effectful suites is preferable, because then we can accomplish per-suite provision with `TestAspect`.

Already, we can do per-test provision with `TestAspect`, using something like:

```scala
suiteM("my suite") { ... } @@ TestAspect.provided(env)
```

The goal should be to enhance aspects until they can do this for the whole suite, rather than just per test, i.e. achieve cross-test sharing within a suite; and this requires effectful suites, which is something we have already discussed and seems to be broadly useful.

What do you two think?

adamgfraser:
@jdegoes @iravid I completely agree with this approach. Let me proceed on that basis.

iravid:
@adamgfraser @jdegoes Yes, effectful suites definitely look like the right way to go. Thanks guys 🙏🏻

adamgfraser:
@jdegoes I just pushed a commit to make suites effectual. I think everything went relatively smoothly, but a couple of things I'm thinking about:

1. Do we need more type parameters on `ZSpec` now? We now have the possibility of "suite level" resources and failures" and "test level" resource and failures. Right now I have these fixed to be the same in the definition of `ZSpec` but not sure that really works (we introduce dependencies and error types unnecessarily and if something changes the type at one level and not the other it may not be a `ZSpec` so things like test aspects won't work). That would really be a lot of type parameters for test aspects. Would be nice to say that the suite type is the supertype / subtype of the test type based on variance, but not sure how to do that with the test type of `Spec` being parameterized on `T` and not `ZIO`.
2. I'm still struggling a little to do managed resources and not have the release action trigger too early before test evaluation. Not sure if I just don't have the combinators right or if it has to do with the way we have effectual tests that we transform and then later execute them so the release action has triggered by the time we actually execute the specs.

Appreciate any thoughts you have as always.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -16,183 +16,277 @@
...
         val (z, caseValue) = f(z0, t)
-        z -> Spec(caseValue)
+        ZIO.succeed(z -> Spec(caseValue))
     }
 }
 object Spec {
-  sealed trait SpecCase[+L, +T, +A] { self =>
-    final def map[B](f: A => B): SpecCase[L, T, B] = self match {
-      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.map(f), exec)
+  sealed trait SpecCase[-R, +E, +L, +T, +A] { self =>
```
jdegoes: Ah, I see you propagated `R` and `E` into `SpecCase`. A lot of type parameters but inference should be extremely good due to lower order nature + variance annotations.

What the design really wants to be is:

```scala
sealed trait SpecCase[F[_], +L, +T, +A]
```

So we can have:

```scala
final case class Effect[F[_], +L, +T, +A](fa: F[A]) extends SpecCase[F, L, T, A]
```

where in a common case, `F[A] = ZIO[R, E, A]`. And in another case, `F[A] = A`.

The problem is ZIO eschews higher-kinded types for teachability / inference, so doing so here would be rather inconsistent with the rest of ZIO, and is almost certain to diminish inference. Although I suppose we could test it.

adamgfraser: Yes I was thinking the same thing but if we managed to do functional effects and streaming without higher order types it seems like we should be able to do make it work for testing. 😃

jdegoes:
@adamgfraser 

> I'm still struggling a little to do managed resources and not have the release action trigger too early before test evaluation. 

The tests are effectful, which means they need to be flattened into the overall suite effect before provision (otherwise out-of-ordering problems will occur). I think this would be simpler with a separate `EffectCase` constructor, because then `T` can become much simpler (no effect, just result). Doesn't solve the problem of "pure specs" but maybe that's a non-problem—i.e. maybe it's OK to require an effect just to traverse a spec.

test/shared/src/main/scala/zio/test/Spec.scala
jdegoes: True, we should keep trying. 😄

adamgfraser:
@jdegoes Yes, I think the next step is for me to rework this with the `EffectCase` constructor. I agree about the pure specs. It would be nice but in this version we're already doing everything with the specs being effectual.

jdegoes:
> Yes, I think the next step is for me to rework this with the EffectCase constructor. I agree about the pure specs. It would be nice but in this version we're already doing everything with the specs being effectual.

I think this makes sense... and this is the one thing to sacrifice without much user-level pain (users won't really notice this given the API; maybe power users).

adamgfraser:
@jdegoes I refactored the effectual suites along the lines we discussed and now have a working prototype of providing a managed resource once to an entire suite. It is still a little rough around the edges but we are getting closer.

A few things I am thinking about:

**1. Handling suite level errors** - Errors in suite level effects are pretty dangerous because unlike the prior iteration, if a suite fails we don't even have a label for the suite. In the previous version when we executed a Spec if there was a failure in the effect I took the label and the failure and constructed a new test node with a runtime exception. Now we can't really do anything with such a failure except log it since we are polymorphic in `L` so can't even construct a label for a node to put it in. I am inclined to think we should stay polymorphic in the error type but say all suite level errors have to be handled before the `TestExecutor` executes the tests.
**2. Type signature for `ZSpec`** - I tried your suggestion of pushing the effect type up so that you only have the effect in one place, but ran into some challenges in making that work with test aspects. In that setup we don't even know if we have a suite or a test until we run an effect, so it isn't clear how we would implement for example test level timeouts since by the time we "know" we have a test it has already run. We do need to do something though. Maybe we create a super polymorphic version of `ZSpec` and then a type alias that only has effects at the test level? I'm open to other ideas.
**3. Use of error versus value channels** - At this point do we want to push test failures into the value channel so we more clearly distinguish them from any errors that can occur with suite level effects. Answer may depend on where we come out on the above but I am leaning towards thinking that would be cleaner.

Other thoughts?

adamgfraser:
@jdegoes The other thing I realized here is I don't think we can currently provide an environment through test aspects because the test aspect can enlarge the types but can't narrow them.

adamgfraser:

adamgfraser:
One idea on this. What if we said that all tests were effectual? Then we could have:

```scala
final case class SuiteCase[-R, +E, +L, +T, +A](label: L, specs: ZIO[R, E, Vector[A]], exec: Option[ExecutionStrategy]) extends SpecCase[R, E, L, T, A]
final case class TestCase[-R, +E, L, T](label: L, test: ZIO[R, E, T]) extends SpecCase[R, E, L, T, Nothing]
```

Advantages: (1) We unify the suite and test level environment and error types, (2) we maintain the guarantee that all test nodes in `ZSpec` are effectual, which I think is important for implementing test aspects, (3) we guarantee that while specs can be created effectually there is always some label for them, which gives us the ability to recover from errors by turning suites with failed effects into test nodes with values indicating failure.

I know there is a ton of other stuff going on right now so no rush in responding. Just something to think about.

jdegoes:
@adamgfraser 

> The other thing I realized here is I don't think we can currently provide an environment through test aspects because the test aspect can enlarge the types but can't narrow them.

This is true, although _combinators_ can narrow them, e.g. `@@ myCominator(all >>> other >>> aspects)`, so it's not a huge limitation. We could also (probably) have a narrowing type of aspect, but it could not compose with the more common widening aspect, so combinator may be better (I'm not sure).

Will address other things soon, sorry for delay, been super busy with hackathon!

adamgfraser:
@jdegoes Yes, I don't think it is a huge limitation. It just means that some things like providing resources will need to be done through methods on `Spec` versus test aspects.

No worries at all. I know that the hackathon and ZIO 1.0 release is top priority right now. There will be plenty of time for more work on ZIO Test after that.

jdegoes:
@adamgfraser 

> Advantages: (1) We unify the suite and test level environment and error types, (2) we maintain the guarantee that all test nodes in ZSpec are effectual, which I think is important for implementing test aspects, (3) we guarantee that while specs can be created effectually there is always some label for them, which gives us the ability to recover from errors by turning suites with failed effects into test nodes with values indicating failure.

I think this makes a lot of sense. It's a simpler representation than the one that factors out the effect, and more uniform. Less flexile but the ergonomics will be higher. I think it's the right set of tradeoffs.

adamgfraser:
@jdegoes I'm on it!

adamgfraser:
@jdegoes Okay, now we're cooking with fire here. All specs are effectual, we have a nice method `provideManagedSuite` on `Spec` that provides a managed resource once to an entire suite, and everything is generally working pretty smoothly.

The one change I made is moving assertion failures into the value channel. Previously we had:

```scala
type ZTest[-R, +E, +S] = ZIO[R, TestFailure[E], TestSuccess[S]]
```

This didn't play well with the new encoding for a couple of reasons. First, the error type is now unified for both the tests and suites, so if the error type for tests is `TestFailure[E]` it doesn't work well with effects at the suite level that just have error type `E`. In addition to this, the `TestExecutor` has to be able to translate a `Spec[R, E, L, S]` into a `Spec[Any, Nothing, L, Either[TestFailure[E], TestSuccess[S]]` so if the `E` type is already `TestFailure[_]` executed specs end up having nested test failures which are difficult to work with. Finally, I think moving assertion failures to the value channel will make it easier to add features like test annotations where we want to aggregate information across both passing and failing tests.

As a result, `ZTest` goes to:

```scala
type ZTest[-R, +E, +S] = ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]
```

So basically runtime failures go in the error channel and assertion results, whether successes or failures, go in the value channel. Could I get a review on this when you have the chance?

jdegoes:
@adamgfraser 

> Okay, now we're cooking with fire here. All specs are effectual, we have a nice method provideManagedSuite on Spec that provides a managed resource once to an entire suite, and everything is generally working pretty smoothly.

Awesome. Original goal: accomplished. ✅ 

> The one change I made is moving assertion failures into the value channel.

I think this is a natural move. The unfortunate side effect is that `retry` and other operations will no longer work on the effectful tests directly. That said, we could always provide a "lift" operator that can treat test failures as effect failures for purposes of effect combinators.

> As a result, ZTest goes to:
> 
> type ZTest[-R, +E, +S] = ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]

This seems sensible. If `E` is always `Nothing` for `TestFailure`, we could consider monomorphizing it.

> Could I get a review on this when you have the chance?

On it, and apologies for the delay!

test-sbt/shared/src/main/scala/zio/test/sbt/ZTestEvent.scala
```diff
@@ -19,15 +21,15 @@ object ZTestEvent {
...
     fingerprint: Fingerprint
-  ): Seq[ZTestEvent] = {
-    def loop(executedSpec: ExecutedSpec[String, E, S]): Seq[ZTestEvent] =
-      executedSpec.caseValue match {
-        case Spec.SuiteCase(_, executedSpecs, _) => executedSpecs.flatMap(loop)
-        case Spec.TestCase(label, result) =>
+  ): UIO[Seq[ZTestEvent]] =
+    executedSpec.mapLabel(_.toString).fold[UIO[Seq[ZTestEvent]]] {
+      case Spec.SuiteCase(_, results, _) =>
+        results.flatMap(UIO.collectAll(_).map(_.flatten))
```
jdegoes: collectAll makes a choice of sequentiality—intentional?

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -19,75 +19,89 @@ package zio.test
...
+          for {
+            specs <- executedSpecs
+            failures <- UIO.foreach(specs)(_.exists {
+                         case Spec.TestCase(_, test) => test.map(_.isLeft);
+                         case _                      => UIO.succeed(false)
+                       })
+            hasFailures   = failures.exists(identity)
+            status        = if (hasFailures) Failed else Passed
+            renderedLabel = if (hasFailures) renderFailureLabel(label, depth) else renderSuccessLabel(label, depth)
+            rest          <- UIO.foreach(specs)(loop(_, depth + tabSize)).map(_.flatten)
```
jdegoes: Fixed to sequential here

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -19,75 +19,89 @@ package zio.test
...
         case Spec.SuiteCase(label, executedSpecs, _) =>
-          val hasFailures = executedSpecs.exists(_.exists {
-            case Spec.TestCase(_, test) => test.isLeft; case _ => false
-          })
-          val status        = if (hasFailures) Failed else Passed
-          val renderedLabel = if (hasFailures) renderFailureLabel(label, depth) else renderSuccessLabel(label, depth)
-          rendered(Suite, label, status, depth, renderedLabel) +: executedSpecs.flatMap(loop(_, depth + tabSize))
+          for {
+            specs <- executedSpecs
+            failures <- UIO.foreach(specs)(_.exists {
```
jdegoes: Fixed to sequential here

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -16,60 +16,77 @@
...
+      case TestCase(label, test)         => TestCase(Right(label), test)
+    }
 
   /**
    * Determines if any node in the spec is satisfied by the given predicate.
    */
-  final def exists(f: SpecCase[L, T, Unit] => Boolean): Boolean =
-    fold[Boolean] {
-      case c @ SuiteCase(_, specs, _) => specs.exists(identity) || f(c.map(_ => ()))
+  final def exists[R1 <: R, E1 >: E](f: SpecCase[R, E, L, T, Any] => ZIO[R1, E1, Boolean]): ZIO[R1, E1, Boolean] =
```
jdegoes: This maybe should be called `existsM`?

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -78,9 +95,9 @@ final case class Spec[+L, +T](caseValue: SpecCase[L, T, Spec[L, T]]) { self =>
   /**
    * Determines if all node in the spec are satisfied by the given predicate.
    */
-  final def forall(f: SpecCase[L, T, Unit] => Boolean): Boolean =
-    fold[Boolean] {
-      case c @ SuiteCase(_, specs, _) => specs.forall(identity) && f(c.map(_ => ()))
+  final def forall[R1 <: R, E1 >: E](f: SpecCase[R, E, L, T, Any] => ZIO[R1, E1, Boolean]): ZIO[R1, E1, Boolean] =
+    fold[ZIO[R1, E1, Boolean]] {
+      case c @ SuiteCase(_, specs, _) => specs.flatMap(ZIO.collectAll(_).map(_.forall(identity))).zipWith(f(c))(_ && _)
```
jdegoes: Sequential specialization. Possibly look at the strategy in the node?

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -89,110 +106,162 @@ final case class Spec[+L, +T](caseValue: SpecCase[L, T, Spec[L, T]]) { self =>
...
-   * Computes the size of the spec, i.e. the number of tests in the spec.
+   * Uses the specified `Managed` to provide each test in this spec with its
+   * required environment.
    */
-  final def size: Int = fold[Int] {
-    case SuiteCase(_, counts, _) => counts.sum
-    case TestCase(_, _)          => 1
+  final def provideManaged[E1 >: E](managed: Managed[E1, R]): Spec[Any, E1, L, T] =
+    transform[Any, E1, L, T] {
+      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.provideManaged(managed), exec)
```
jdegoes: It seems like `transform` is awfully low-level for this, do we have anything that just allows function application to `specs` and `test`?

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -89,110 +106,162 @@ final case class Spec[+L, +T](caseValue: SpecCase[L, T, Spec[L, T]]) { self =>
...
+      case TestCase(label, test)         => TestCase(label, test.provideManaged(managed))
+    }
+
+  /**
+   * Uses the specified `Managed` once to provide all tests in this spec with
+   * a shared version of their required environment. This is useful when the
+   * act of creating the environment is expensive and should only be performed
+   * once.
+   */
+  final def provideManagedSuite[E1 >: E](managed: Managed[E1, R]): Spec[Any, E1, L, T] = {
```
jdegoes: How about `provideManagedShared` if this is global sharing? `provideManagedSuite` suggests operation at the level of a suite, maybe each suite (i.e. each nested suite gets its own resource).

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -89,110 +106,162 @@ final case class Spec[+L, +T](caseValue: SpecCase[L, T, Spec[L, T]]) { self =>
...
+    def loop(r: R)(spec: Spec[R, E, L, T]): ZIO[Any, E, Spec[Any, E, L, T]] =
```
jdegoes: I feel like this `loop` suggests the need for another traversal scheme.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -89,110 +106,162 @@ final case class Spec[+L, +T](caseValue: SpecCase[L, T, Spec[L, T]]) { self =>
...
+        Spec.suite(label, managed.use(r => specs.flatMap(ZIO.foreach(_)(loop(r))).map(_.toVector).provide(r)), exec)
+      case TestCase(label, test) =>
+        Spec.test(label, test.provideManaged(managed))
+    }
   }
 
+  /**
+   * Computes the size of the spec, i.e. the number of tests in the spec.
+   */
+  final def size: ZIO[R, E, Int] =
```
jdegoes: I wonder just how useful all these become in an effectful world. Basically computing the size here could, in theory, allocate and de-allocate resources required only to run the tests!

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -303,9 +303,37 @@ object TestAspect extends TimeoutVariants {
...
+        case Spec.TestCase(label, test) =>
+          Spec.TestCase(label, if (predicate(label)) from(perTest(to(test))) else test)
       }
+
+    /**
+     * These methods evidence an isomorphism between our previous encoding of
+     * `ZTest` and the new encoding. We can remove them once we refactor test
+     * aspects to use the new encoding
+     */
+    private def to[R, E, S](
```
jdegoes: Will this be part of this pull request, or come along later?

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -110,13 +110,13 @@ package object test extends CheckVariants {
    * Creates a failed test result with the specified runtime cause.
    */
   final def fail[E](cause: Cause[E]): ZTest[Any, E, Nothing] =
-    ZIO.fail(TestFailure.Runtime(cause))
+    ZIO.halt(cause)
```
jdegoes: Nice simplification! 👍

test-sbt/shared/src/main/scala/zio/test/sbt/ZTestEvent.scala
adamgfraser: My thought was that since we are already dealing with an `executedSpec` here the potentially compute intensive work of running the test has already been done so we don't need parallelism here.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
adamgfraser: Similar thought to the above that since we've already run the tests we don't need parallelism in rendering the results.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
adamgfraser: Same as above.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: Yes we can definitely do that.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: Yes! Changed!

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: I was planning on doing it as part of this pull request, I just wanted to make sure we were aligned before I went through and changed all the test aspects. The one thing that may change this is if we are going to do a release today it could be nice to get this in so that projects that need the shared resource functionality can start using ZIO Test, even if we need to do a little cleanup on the backend afterwards.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: I totally agree. I wonder if we move these to some type of syntax class for executed Specs or require implicit evidence that the spec has been executed. The only variants of these `fold` type methods we use are `exists` and `forall` in checking if all executed specs passed.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: I agree. How about something like:

```scala
  final def mapTests[R1, E1, L1 >: L, T1](
    suiteCase: ZIO[R, E, Vector[Spec[R1, E1, L1, T1]]] => ZIO[R1, E1, Vector[Spec[R1, E1, L1, T1]]],
    testCase: ZIO[R, E, T] => ZIO[R1, E1, T1]
  ): Spec[R1, E1, L1, T1] =
    caseValue match {
      case SuiteCase(label, specs, exec) =>
        Spec.suite(label, suiteCase(specs.map(_.map(_.mapTests(suiteCase, testCase)))), exec)
      case TestCase(label, test) => Spec.test(label, testCase(test))
    }
```

Then we could do:

```scala
  final def provideManaged[E1 >: E](managed: Managed[E1, R]): Spec[Any, E1, L, T] =
    mapTests[Any, E1, L, T](_.provideManaged(managed), _.provideManaged(managed))
```

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: We could. The only time this method is used in the code base today is on executed specs. This goes a little to some of the other points about the value of these methods on specs that haven't been executed now that the suites are effectual.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: I agree. I've been struggling a bit to find the right encoding. I think what we want is a  function that takes a `ZIO[R, E, T]` in the test case to a `ZIO[R1, E1, T]`, where there are no constrains on `R1` and `E1` being subtypes / supertypes of `R` and `E`. And then I think for the suite case we basically need a universally quantified function that takes a `ZIO[R, E, X]` to a `ZIO[R1, E1, X]`, so the suite case can change the effect but should delegate changing the child specs to recursively calling the traversal. Otherwise we run into issues with the environment and error types getting widened to the union of `R` and `R1` and `E` and `E1` instead of just getting `R1 ` and `E`.

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: I will have this done shortly.

adamgfraser:
@jdegoes I updated the name of the method to `provideManagedShared` and updated all the test aspects to the new encoding. My strong inclination at this point is that we should get this in for this release. We have some potential users who really need this feature and we could benefit from having more time to get feedback and iterate on the effectual suites. I think we can definitely further simplify but can also pick that up in a subsequent PR if we don't get to it now.

jdegoes:
@adamgfraser Agreed, done! Thanks for all your hard work on this. It gets us that much closer to ZIO Test 1.0!

PR #1667, opened 2019-09-13
Add a synchronous effect construction for the ZManaged (#1666)

### LINKED ISSUE #1666: Add method that allows lift a synchronous effect into `ZManaged`
As we discussed with @iravid at Gitter channel, it would be nice if ZIO ZManaged and Managed will have construction method from a synchronous effect like this:

`Managed.makeEffect(Files.createTempFile("xxxx", "yyyy"))(f => f.delete())`

I think above construction will remove a lot of cases where we need to write unnecessary boilerplate code by boxing a synchronous effect to `Task` and then call `.orDie`

LukaszByczynski:
Resolves #1666.

### DIFF
--- a/core/shared/src/main/scala/zio/Managed.scala
+++ b/core/shared/src/main/scala/zio/Managed.scala
@@ -164,7 +164,7 @@ object Managed {
   /**
    * See [[zio.ZManaged.makeEffect]]
    */
-  final def makeEffect[E, A](acquire: => A)(release: A => _): Managed[E, A] =
+  final def makeEffect[A](acquire: => A)(release: A => _): Managed[Throwable, A] =
     ZManaged.makeEffect(acquire)(release)
 
   /**
--- a/core/shared/src/main/scala/zio/Managed.scala
+++ b/core/shared/src/main/scala/zio/Managed.scala
@@ -162,6 +162,12 @@ object Managed {
     ZManaged.make(acquire)(release)
 
   /**
+   * See [[zio.ZManaged.makeEffect]]
+   */
+  final def makeEffect[E, A](acquire: => A)(release: A => _): Managed[E, A] =
+    ZManaged.makeEffect(acquire)(release)
+
+  /**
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1007,6 +1007,12 @@ object ZManaged {
   final def make[R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R, Nothing, _]): ZManaged[R, E, A] =
     ZManaged(acquire.map(r => Reservation(IO.succeed(r), _ => release(r))))
 
+  /**
+   * Lifts a synchronous effect into `ZManaged[R, E, R]` with a release action.
+   */
+  final def makeEffect[R, E, A](acquire: => A)(release: A => _): ZManaged[R, E, A] =
```
```suggestion
  final def makeEffect[R, A](acquire: => A)(release: A => _): ZManaged[R, Throwable, A] =
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1007,6 +1007,12 @@ object ZManaged {
...
+    make(Task(acquire).orDie)(a => Task(release(a)).orDie)
```
```suggestion
    make(Task(acquire))(a => Task(release(a)).orDie)
```

core/shared/src/main/scala/zio/Managed.scala
```diff
@@ -161,6 +161,12 @@ object Managed {
   final def make[E, A](acquire: IO[E, A])(release: A => UIO[_]): Managed[E, A] =
     ZManaged.make(acquire)(release)
 
+  /**
+   * See [[zio.ZManaged.makeEffect]]
+   */
+  final def makeEffect[E, A](acquire: => A)(release: A => _): Managed[E, A] =
```
iravid:
```suggestion
  final def makeEffect[A](acquire: => A)(release: A => _): Managed[Throwable, A] =
```

LukaszByczynski:
@iravid  Thanks for the review. I did suggested changes

Issue #1647, opened 2019-09-12
Migrate all tests to ZIO Test


ghostdogpr:
Here's the list of tests to migrate to ZIO Test. Look at `ZQueueSpec` or `ZManagedSpec` for examples.
If you want to contribute this ticket, please comment to say which spec you're tackling.

- [x] CauseSpec
- [x] FiberLocalSpec
- [x] FiberRefSpec
- [x] FiberSpec
- [x] FunctionIOSpec
- [x] PromiseSpec
- [x] RefMSpec
- [x] RefSpec
- [x] ZIOSpec
- [x] ZManagedSpec
- [x] ZQueueSpec
- [x] ZScheduleSpec
- [x] DurationSpec
- [x] ExecutorSpec
- [x] StackBoolSpec
- [x] STMSpec
- [x] PlatformSpec
- [x] RTSSpec
- [x] SemaphoreSpec
- [x] SerializableSpec
- [x] ZIOSpecJvm
- [x] BlockingSpec
- [x] MutableConcurrentQueueSpec
- [x] OneShotSpec
- [x] TArraySpec
- [x] SystemSpec
- [x] js/MutableConcurrentQueueSpec
- [x] js/OneShotSpec
- [x] jvm/2.12/StacktracesSpec

When all are done, we can remove the dependency to specs2 as well as some common code.

LGLO:
I'll try with STMSpec and merge StmIssue1587Spec to it.

zezutom:
I'd like to take Duration and Serializable specs.

adamgfraser:
I'm doing CauseSpec as part of #1706.

lemastero:
I am attacking `FiberSpec`

kapke:
I'll try to migrate JS specific ones.

wojtek-tg:
I'd like to take a stab at `DurationSpec`

XLPI:
Ok, I'll try pick easy one. Which one can you advice? @ghostdogpr

XLPI:
I picked ExecutorSpec

kapke:
`MutableConcurrentQueueSpec`and `OneShotSpec` are migrated too.

kapke:
I'm starting migration for `FunctionIOSpec` now

adamgfraser:
You're on fire! 🔥

mijicd:
I'll start with `RTSSpec` now.

smart-ass:
I'm taking `StackBoolSpec`

[...]

PR #1668, opened 2019-09-13
Migrates some tests to ZIOTests

fagossa:
Migrates `SystemSpec` and `BlockingSpec` to `ZIOTests`.

Solves #1647

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
@@ -12,13 +12,13 @@ class BlockingSpec
     extends ZIOBaseSpec(
       suite("BlockingSpec")(
         suite("Make a Blocking Service and verify that")(
-          testM("effectBlocking` completes successfully") {
+          testM("effectBlocking completes successfully") {
             assertM(effectBlocking(()), isUnit)
           },
-          testM("effectBlockingCancelable` completes successfully") {
+          testM("effectBlockingCancelable completes successfully") {
             assertM(effectBlockingCancelable(())(UIO.unit), isUnit)
           },
-          testM("effectBlocking` can be interrupted") {
+          testM("effectBlocking can be interrupted") {
             assertM(effectBlocking(Thread.sleep(50000)).timeout(Duration.Zero), isNone)
           },
           testM("effectBlockingCancelable can be interrupted") {
--- a/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
@@ -3,33 +3,44 @@ package zio.blocking
 import java.util.concurrent.atomic.AtomicBoolean
 
 import zio.duration.Duration
-import zio.{ TestRuntime, UIO }
+import zio.{ UIO, ZIOBaseSpec }
+import BlockingSpecUtil._
+import zio.test._
+import zio.test.Assertion._
 
-final class BlockingSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime {
+class BlockingSpec
+    extends ZIOBaseSpec(
+      suite("BlockingSpec")(
+        suite("Make a Blocking Service and verify that that")(
+          testM("effectBlocking` completes successfully") {
+            for {
+              io <- effectBlocking(())
+            } yield assert(io, equalTo(()))
+          },
+          testM("effectBlockingCancelable` completes successfully") {
+            for {
+              io <- effectBlockingCancelable(())(UIO.unit)
+            } yield assert(io, equalTo(()))
+          },
+          testM("effectBlocking` can be interrupted") {
+            for {
+              io <- effectBlocking(Thread.sleep(50000)).timeout(Duration.Zero)
+            } yield assert(io, isNone)
+          },
+          testM("effectBlockingCancelable` can be interrupted") {
+            val release = new AtomicBoolean(false)
+            val cancel  = UIO.effectTotal(release.set(true))
+            for {
+              io <- effectBlockingCancelable(blockingAtomic(release))(cancel)
+                     .timeout(Duration.Zero)
+            } yield assert(io, isNone)
+          }
+        )
+      )
+    )
 
-  def is =
-    "BlockingSpec".title ^
-      s2"""
-    Make a Blocking Service and
-    verify that
-      `effectBlocking` completes successfully $e1
-      `effectBlockingCancelable` completes successfully $e2
-      `effectBlocking` can be interrupted $e3
-      `effectBlockingCancelable` can be interrupted $e4
-    """
-
-  def e1 =
-    unsafeRun(effectBlocking(())).must_===(())
-
-  def e2 =
-    unsafeRun(effectBlockingCancelable(())(UIO.unit)).must_===(())
-
-  def e3 = {
-    val res = unsafeRun(effectBlocking(Thread.sleep(50000)).timeout(Duration.Zero))
-    res must beNone
-  }
-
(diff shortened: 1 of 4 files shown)


test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -195,6 +195,9 @@ object Assertion {
   final def contains[A](element: A): Assertion[Iterable[A]] =
     Assertion.assertion("contains")(param(element))(_.exists(_ == element))
 
+  final def containsString(element: String): Assertion[String] =
```
fagossa: maybe should I rename this function to `include`? 🤷‍♂

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -195,6 +195,9 @@ object Assertion {
   final def contains[A](element: A): Assertion[Iterable[A]] =
     Assertion.assertion("contains")(param(element))(_.exists(_ == element))
 
+  final def include(element: String): Assertion[String] =
```
adamgfraser: `hasSubstring`? Whatever name we choose, we should put it in alphabetical order with the other assertions and I would make the string description of the assertion match its name (e.g. if we go with `hasSubstring` it would be `Assertion.assertion("hasSubstring")`. I would also add a couple of tests for this in `AssertionSpec`, one for failure and one for success, along the lines of the other tests there.

core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala
```diff
@@ -1,44 +1,38 @@
...
-    io must beNone
-  }
+import scala.reflect.io.File
 
-  def lineSep1 = unsafeRun(lineSeparator) must_=== (java.lang.System.lineSeparator)
-}
+class SystemSpec
+    extends ZIOBaseSpec(
+      suite("SystemSpec")(
+        testM("Fetch an environment variable and check that If it exists, return a reasonable value") {
```
adamgfraser: You can use nested suites to recover the same structure that the original spec had. See `ZManagedSpec` for an example of this.

```scala
suite("Fetch an environment variable and check that:")(
  testM("If it exists, returns a reasonable value") { ... }
  testM("If it does not exist, return None") { ... }
)
```

core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala
```diff
@@ -1,44 +1,44 @@
...
-  def lineSep1 = unsafeRun(lineSeparator) must_=== (java.lang.System.lineSeparator)
-}
+class SystemSpec
+    extends ZIOBaseSpec(
+      suite("SystemSpec")(
+        suite("Fetch an environment variable and check that")(
+          testM("If it exists, return a reasonable value") {
+            for {
+              io <- system.env("PATH")
+            } yield assert(io, isSome(hasSubstring(File.separator + "bin")))
```
ghostdogpr: How about `containsString`? cc @adamgfraser

core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
```diff
@@ -3,45 +3,49 @@ package zio.blocking
...
+import zio.test.Assertion._
+
+class BlockingSpec
+    extends ZIOBaseSpec(
+      suite("BlockingSpec")(
+        suite("Make a Blocking Service and verify that that")(
+          testM("effectBlocking` completes successfully") {
+            for {
+              io <- effectBlocking(())
+            } yield assert(io, equalTo(()))
```
ghostdogpr: There’s `isUnit`, even shorter

core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
```diff
@@ -3,45 +3,49 @@ package zio.blocking
...
+import zio.test._
+import zio.test.Assertion._
+
+class BlockingSpec
+    extends ZIOBaseSpec(
+      suite("BlockingSpec")(
+        suite("Make a Blocking Service and verify that that")(
+          testM("effectBlocking` completes successfully") {
+            for {
+              io <- effectBlocking(())
```
ghostdogpr: How about using `assertM` to make these tests even shorter? You wouldn’t need the for comprehension.

core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
```diff
@@ -3,45 +3,49 @@ package zio.blocking
...
-  def blocking(released: AtomicBoolean) =
+import zio.{ UIO, ZIOBaseSpec }
+import BlockingSpecUtil._
+import zio.test._
+import zio.test.Assertion._
+
+class BlockingSpec
+    extends ZIOBaseSpec(
+      suite("BlockingSpec")(
+        suite("Make a Blocking Service and verify that that")(
```
ghostdogpr: -that

core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala
adamgfraser: 👍

test/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -225,6 +225,14 @@ object AssertionSpec {
     test(
       !nameStartsWithA.test(sampleUser),
       message = "test must return false when given element does not satisfy assertion"
+    ),
+    testSuccess(
+      assert("this is a value", containsString("is a")),
+      message = "hasSubstring must succeed when string is found"
```
adamgfraser: Update name.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -443,6 +443,12 @@ object Assertion {
     BoolAlgebra.failure(AssertionValue(self, actual))
   }
 
+  /**
+   * Makes a new assertion that search a substring.
```
adamgfraser: Typo. Also can we put it in alphabetical order so it is easy for people to find in the future?

core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
```diff
@@ -3,45 +3,40 @@ package zio.blocking
...
+import zio.{ UIO, ZIOBaseSpec }
+import BlockingSpecUtil._
+import zio.test._
+import zio.test.Assertion._
+
+class BlockingSpec
+    extends ZIOBaseSpec(
+      suite("BlockingSpec")(
+        suite("Make a Blocking Service and verify that")(
+          testM("effectBlocking` completes successfully") {
```
adamgfraser: Backticks before method names here and below.

core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala
```diff
@@ -1,44 +1,34 @@
...
-  def lineSep1 = unsafeRun(lineSeparator) must_=== (java.lang.System.lineSeparator)
-}
+class SystemSpec
+    extends ZIOBaseSpec(
+      suite("SystemSpec")(
+        suite("Fetch an environment variable and check that")(
+          testM("If it exists, return a reasonable value") {
+            assertM(system.env("PATH"), isSome(containsString(File.separator + "bin")))
+          },
+          testM("Fetch an environment variable and check that If it does not exist, return None") {
```
ghostdogpr: You can remove `Fetch an environment variable and check that ` here

fagossa: Yep I forgot that one too :s

ghostdogpr:
@fagossa thanks! 🎉

PR #1670, opened 2019-09-13
ZIO Test: Implement Generators for Effects

### LINKED ISSUE #1653: ZIO Test: Add support for effect generators
Support `Gen[R, ZIO[Env, E, A]]`, or at least `Gen[R, IO[E, A]]`.

adamgfraser:
Implements generators for `ZIO` and its various type aliases.

I took a somewhat different approach from the existing generators in the test suites. I think we need to make a distinction between generators that `ZIO` uses internally to test its own correctness and generators that users of `ZIO` should use. I would argue that for a user, if a `ZIO` produces an `A` and doesn't perform any actual effects then it should be indistinguishable to a user how that `A` was produced (e.g. if it was there all along or created through a complex chain or transformations). If it is possible for a user to distinguish those then that is an internal error in `ZIO` that needs to be addressed. If this is correct then to generate `ZIO` values we only need to be concerned with producing the right results and not in producing different computations to produce those results. This simplifies the implementation considerably and will also make these generators faster and shrink better for users.

Internally, `ZIO` can use more complex generators to verify that various different sequences of computations produce the expected results, but that shouldn't be exposed to users. At least that is my current thinking. Interested if others have different points of view.

Resolves #1653.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -84,7 +84,7 @@ case class Gen[-R, +A](sample: ZStream[R, Nothing, Sample[R, A]]) { self =>
     self.zip(that).map(f.tupled)
 }
 
-object Gen {
+object Gen extends GenZIO {
 
   /**
    * A generator of alphanumeric characters. Shrinks toward '0'.
@@ -397,6 +397,12 @@ object Gen {
    */
   final def suspend[R, A](gen: => Gen[R, A]): Gen[R, A] =
     fromEffect(ZIO.effectTotal(gen)).flatten
+
+  /**
+   * A generator of throwables.
+   */
+  final def throwable: Gen[Random, Throwable] =
+    Gen.const(new Throwable)
 
   /**
    * A generator of uniformly distributed doubles between [0, 1].
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -401,7 +401,7 @@ object Gen extends GenZIO {
   /**
    * A generator of throwables.
    */
-  final def throwable: Gen[Random, Throwable] =
+  final val throwable: Gen[Random, Throwable] =
     Gen.const(new Throwable)
 
   /**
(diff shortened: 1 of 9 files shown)


iravid:
@adamgfraser 

> I would argue that for a user, if a ZIO produces an A and doesn't perform any actual effects then it should be indistinguishable to a user how that A was produced (e.g. if it was there all along or created through a complex chain or transformations).

It actually matters quite a bit when testing concurrent code. As a user of the library, I want to be in control of whether the generated IO contains concurrency and interruption, so I can test concurrent code against that. E.g., consider `Stream`'s combinators- it'd be a shame if we could not test them using `IO` values that generate concurrency.

iravid:
And, of course, we can exchange "concurrency" which I wrote above with any other feature of ZIO - interruption, defects, sleeps, etc.

adamgfraser:
@iravid I think maybe there is a synthesis between what we are saying.

When I look at the [existing generators of effects](https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/GenIO.scala). I see us building up a trees of sub-effects that are essentially no-ops (e.g. `race` against `never`, `map` with `identity`, etc...). Those are the kind of steps that I'm not sure add value to say testing `Stream`. On the other hand, I think you are 100% right that it would be good if we had generators for things like `die` effects, effects that `sleep` for an arbitrary amount of time, or effects that never terminate and then users could compose those together.

Does that make sense?

adamgfraser:
On interruption in particular there is a difference where `a` may not be interruptible but `a.flatMap(ZIO.succeed)` could be interrupted so we do need something for that.

test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -398,6 +398,12 @@ object Gen {
   final def suspend[R, A](gen: => Gen[R, A]): Gen[R, A] =
     fromEffect(ZIO.effectTotal(gen)).flatten
 
+  /**
+   * A generator of throwables.
+   */
+  final def throwable: Gen[Random, Throwable] =
```
jdegoes: This can be `val`.

adamgfraser: Yes.

jdegoes:
What about a combinator based approach:

 - `successes` (ZIO.succeed)
 - `failures` (ZIO.fail)
 - `chained(zio)` - effect chained with itself `n` times (`Sized`)
 - `died` (ZIO.die)
 - `concurrent` - random no-op concurrency applied (race, etc.)
 - `parallel` - random no-op parallelism applied (race, etc.)
 - etc.

adamgfraser:
I like that a lot.

iravid:
@adamgfraser Yes, completely agree!

@jdegoes @adamgfraser Yes - this is what I want! 😍

adamgfraser:

PR #1672, opened 2019-09-14
Update ZStream#aggregate,aggregateWithin to only extract once

### LINKED ISSUE #1649: Modify the `ZStream#aggregate` state machine
Follow up to #1560, which changes the encoding of `ZSink`.

After #1560 is merged, the `aggregate` combinators end up extracting the sink twice, in order to obtain the leftovers. This might be too expensive for some sinks and should be addressed. The solution most likely needs to modify the `State` sealed trait which encodes the possible states that the combinator can be in.

One further improvement, the `consume` inner method return type can be changed from `ZIO[R, E, Option[Chunk[B]]]` to `ZIO[R, Option[E], Chunk[B]]`. The `None` case is currently used to signal finish, but this creates unnecessary allocations on each result. `Option[E]` creates only one allocation.

iravid:
Resolves #1649.

@vasilmkd Could you have a look?


vasilmkd:
LGTM.

PR #1681, opened 2019-09-15
Refactor ZSink#chunked and ZSink#flatMap to not use stepChunk

### LINKED ISSUE #1677: Refactor ZSink#chunked to not use stepChunk
Should use stepChunkSlice instead.

### LINKED ISSUE #1678: Refactor ZSink#flatMap to not use stepChunk
Should use stepChunkSlice instead.

vasilmkd:
Resolves #1677.
Resolves #1678.

@iravid

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -621,8 +621,10 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
       step1  <- sink.step(init, 1)
       step2  <- sink.step(step1, 2)
       step3  <- sink.step(step2, 3)
-      result <- sink.extract(step3)
-    } yield result must_=== ((List(1, 2, 3), Chunk.empty))
+      step4  <- sink.step(step3, 4)
+      step5  <- sink.step(step4, 5)
+      result <- sink.extract(step5)
+    } yield result must_=== ((List(1, 2, 3), Chunk(4, 5)))
     unsafeRun(test)
   }
 
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -86,6 +86,10 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
       init error    $flatMapInitError
       step error    $flatMapStepError
       extract error $flatMapExtractError
+      self done     $flatMapSelfDone
+      self more     $flatMapSelfMore
+      pass leftover $flatMapPassLeftover
+      end leftover  $flatMapEndLeftover
 
     filter
       happy path      $filterHappyPath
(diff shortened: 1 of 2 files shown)


streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -393,6 +399,26 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
...
+    val test = for {
+      init   <- sink.initial
+      step   <- sink.step(init, Chunk(1, 2, 3, 4, 5))
+      result <- sink.extract(step)
+    } yield result must_=== ((List(1, 2), Chunk(3, 4, 5)))
+    unsafeRun(test)
+  }
+
+  private def chunkedLeftoverAppend = {
+    val sink = ZSink.ignoreWhile[Int](_ < 0).chunked
```
iravid: You know, I think the semantics of `ignoreWhile` should not be "return ignored elements as leftover", but rather "swallow elements that are ignored". What do you think?

vasilmkd: Technically, it returns the first non ignored element as leftover.

iravid: Oh, sorry, I misread the test. All good then.

vasilmkd: Maybe it would be better as `dropUntil` with a flipped predicate, but it's more or less the same.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -305,48 +301,77 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+                          Right((that, s3, chunk.map(ev2)))
+                      }
                     }
                 }
             }
 
-          case Right((that, s2)) =>
-            that.step(s2.asInstanceOf[that.State], a).map(s2 => Right((that, s2)))
+          // If `that` needs to continue, it will have already processed all
+          // of the leftovers from `self`.
```
iravid:
```suggestion
          // of the leftovers from `self`, because they were stepped in `initial` or `case Left` above.
```

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -305,48 +301,77 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
    */
   final def flatMap[R1 <: R, E1 >: E, A00 >: A0, A1 <: A, C](
```
iravid: Great work on this one.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -588,6 +614,54 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
+  private def flatMapSelfDone = {
+    val sink = ZSink.succeed(3).flatMap(n => ZSink.collectAllN[Int](n.toLong))
+    val test = for {
+      init   <- sink.initial
+      step1  <- sink.step(init, 1)
+      step2  <- sink.step(step1, 2)
+      step3  <- sink.step(step2, 3)
```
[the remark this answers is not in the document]
vasilmkd: Done.

streams/shared/src/main/scala/zio/stream/ZSink.scala

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: It's nothing.

PR #1683, opened 2019-09-15
Fixed #1314 - ZIO Test: Add Common Predicates

### LINKED ISSUE #1314: ZIO Test: Add Common Predicates
Following the existing code in Predicate.scala, inside the `/test/...` directory (_ZIO Test_), this ticket is to add common predicates.

A predicate is basically a function from some value (like a string) to a test result, either success or failure (with a message). Mostly, existing similar predicates can be copy / pasted / modified, so there is no need to engineer complex code.

A good place to find predicates is boolean-returning methods on Scala collections, the String type in Java / Scala, and so forth.

These would include:

 - `startsWith`
 - `endsWith`
 - `matches` (Regex)
 - Etc.

Note that each predicate must be a _verb_; so, for example, `initial` is not a good name for a predicate, but `startsWith` is a good name.

jaliss:


test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
         case t: Throwable => assertion(t)
       }
     }
+
+  final def isEmptyString: Assertion[String] = Assertion.assertion(s"isEmpty")() { actual =>
+    actual.isEmpty
+  }
+
+  final def nonEmptyString[A <: String]: Assertion[String] = Assertion.assertion(s"nonEmpty")() { actual =>
```
jdegoes: `isNonEmptyString`

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+  final def approximatelyEquals[A](range: NumericRange[A]): Assertion[A] =
+    Assertion.assertion("approximatelyEquals")(param(range)) { actual =>
+      range.isInRange(actual)
+    }
+
+  final def isEmpty: Assertion[Iterable[_]] = Assertion.assertion("isEmpty")() { actual =>
+    actual.isEmpty
+  }
+
+  final def nonEmpty: Assertion[Iterable[_]] = Assertion.assertion("nonEmpty")() { actual =>
```
jdegoes: `isNonEmpty`

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+  final def endsWith(suffix: String): Assertion[String] = Assertion.assertion("endsWith")(param(suffix)) { actual =>
+    actual.endsWith(suffix)
+  }
+
+  final def matchesRegex(regex: String): Assertion[String] = Assertion.assertion("matchesRegex")(param(regex)) {
+    actual =>
+      actual.matches(regex)
+  }
+
+  final def approximatelyEquals[A](range: NumericRange[A]): Assertion[A] =
```
adamgfraser: We generally try to avoid relying on implicit methods because they can be harder for the user to find. Can we change the signature to something like:

```scala
final def approximatelyEquals[A: Numeric](reference: A, tolerance: A): Assertion[A]
```

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+
+  final def isEmpty: Assertion[Iterable[_]] = Assertion.assertion("isEmpty")() { actual =>
+    actual.isEmpty
+  }
+
+  final def isNonEmpty: Assertion[Iterable[_]] = Assertion.assertion("isNonEmpty")() { actual =>
+    actual.nonEmpty
+  }
+
+  final def containsElement[A](element: A): Assertion[Iterable[A]] =
```
adamgfraser: How is this different from the existing `contains` assertion?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
         case t: Throwable => assertion(t)
       }
     }
+
+  final def isEmptyString: Assertion[String] = Assertion.assertion(s"isEmpty")() { actual =>
```
adamgfraser: I would make the name of the assertion match the method name (i.e. both "isEmptyString"). Also, you don't need the string interpolation here.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+    actual.isEmpty
+  }
+
+  final def isNonEmptyString: Assertion[String] = Assertion.assertion(s"isNonEmptyString")() { actual =>
```
adamgfraser: You can make this and the other assertions that don't have any parameters a `val` instead of a `def`.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+
+  final def startsWith(prefix: String): Assertion[String] = Assertion.assertion("startsWith")(param(prefix)) { actual =>
+    actual.startsWith(prefix)
+  }
+
+  final def endsWith(suffix: String): Assertion[String] = Assertion.assertion("endsWith")(param(suffix)) { actual =>
+    actual.endsWith(suffix)
+  }
+
+  final def matchesRegex(regex: String): Assertion[String] = Assertion.assertion("matchesRegex")(param(regex)) {
```
adamgfraser: Great!

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -473,4 +473,52 @@ object Assertion {
...
+    actual =>
+      actual.matches(regex)
+  }
+
+  final def approximatelyEquals[A](range: NumericRange[A]): Assertion[A] =
+    Assertion.assertion("approximatelyEquals")(param(range)) { actual =>
+      range.isInRange(actual)
+    }
+
+  final def isEmpty: Assertion[Iterable[_]] = Assertion.assertion("isEmpty")() { actual =>
```
adamgfraser: Can go to `Assertion[Iterable[Any]]`.

test/shared/src/main/scala/zio/test/Assertion.scala
jaliss: it's no different ... missed we already had this.

PR #1685, opened 2019-09-15
ZStream#buffer variations

### LINKED ISSUE #1659: Add ZStream#bufferUnbounded, ZStream#bufferDropping, ZStream#bufferSliding
These combinators should be identical to `ZStream#buffer`, but should construct a different underlying queue.

The same combinators should also be added to `ZStreamChunk`.

vasilmkd:
Resolves #1659.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -597,6 +597,78 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
     }
 
   /**
+   * Creates a stream that passes its elements through a queue that drops elements
+   * (dropping or sliding). It implements a promise that signals abrupt completion
+   * of the stream due to an error or the end of the stream. After the promise has
+   * been completed, it continues to offer the current (at the time) contents of the
+   * queue, until it has been exhausted, at which point it finishes with the signal
+   * from the promise.
+   */
+  private final def bufferSignal[A1 >: A](queue: Queue[A1]): ZStream[R, E, A1] =
+    ZStream.managed {
+      for {
+        // Signal stream error or end through the result type of the promise
+        // for better composability with other effects.
+        signal <- Promise.make[Option[E], Nothing].toManaged_
+        _ <- self
+              .foreachManaged(a => queue.offer(a).unit)
+              .foldM(
+                e => signal.fail(Some(e)).unit.toManaged_,
+                _ => signal.fail(None).unit.toManaged_
+              )
+              .fork
+      } yield (queue, signal)
+    }.flatMap {
+      case (queue, signal) =>
+        def takeFirst(exit: Exit[Nothing, A1], fiber: Fiber[Option[E], Nothing]): Pull[Any, Nothing, A1] =
+          fiber.interrupt *> exit.fold(Pull.halt, Pull.emit)
+
+        def finishStream(exit: Exit[Option[E], Nothing], fiber: Fiber[Nothing, A1]): Pull[Any, E, Nothing] =
+          fiber.interrupt *> exit.fold(
+            c => Pull.sequenceCauseOption(c).fold[Pull[Any, E, Nothing]](Pull.end)(Pull.halt),
+            identity
+          )
+
+        def signalFirst(exit: Exit[Option[E], Nothing], fiber: Fiber[Nothing, A1]): Pull[Any, E, A1] =
+          queue.size.flatMap { size =>
+            if (size > 0) fiber.join
+            else fiber.poll.flatMap(_.fold[Pull[Any, E, A1]](finishStream(exit, fiber))(_.fold(Pull.halt, Pull.emit)))
+          }
+
+        val pull: Pull[Any, E, A1] =
+          queue.take.raceWith(signal.await)(takeFirst, signalFirst)
+
+        ZStream.fromPull(pull)
+    }
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * up to `capacity` elements in a dropping queue.
+   *
+   * @note Prefer capacities that are powers of 2 for better performance.
+   */
+  final def bufferDropping(capacity: Int): ZStream[R, E, A] =
+    ZStream.managed(ZManaged.make(Queue.dropping[A](capacity))(_.shutdown)).flatMap(bufferSignal)
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * up to `capacity` elements in a sliding queue.
+   *
+   * @note Prefer capacities that are powers of 2 for better performance.
+   */
+  final def bufferSliding(capacity: Int): ZStream[R, E, A] =
+    ZStream.managed(ZManaged.make(Queue.sliding[A](capacity))(_.shutdown)).flatMap(bufferSignal)
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * elements into an unbounded queue.
+   */
+  final def bufferUnbounded: ZStream[R, E, A] =
+    ZStream.managed(self.toQueueUnbounded).flatMap { queue =>
+      ZStream.fromQueue(queue).unTake
+    }
+
+  /**
    * Switches over to the stream produced by the provided function in case this one
    * fails with a typed error.
    */
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamBufferSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamBufferSpec.scala
@@ -1,0 +1,179 @@
+package zio.stream
+
+import org.specs2.ScalaCheck
+
+import scala.{ Stream => _ }
+import zio._
+
+class StreamBufferSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime with GenIO with ScalaCheck {
+
+  import Exit.{ Cause => _, _ }
+  import zio.Cause
+
+  def is = "StreamBufferSpec".title ^ s2"""
+  Stream.buffer
+    buffer the Stream                    $bufferStream
+    buffer the Stream with Error         $bufferStreamError
+    fast producer progress independently $bufferFastProducerSlowConsumer
+
+  Stream.bufferDropping
+    buffer the Stream with Error         $bufferDroppingError
+    fast producer progress independently $bufferDroppingFastProducerSlowConsumer
+
+  Stream.bufferSliding
+    buffer the Stream with Error         $bufferSlidingError
+    fast producer progress independently $bufferSlidingFastProducerSlowConsumer
+
+  Stream.bufferUnbounded
+    buffer the Stream                    $bufferUnboundedStream
+    buffer the Stream with Error         $bufferUnboundedStreamError
+    fast producer progress independently $bufferUnboundedFastProducerSlowConsumer
+  """
+
+  private def bufferStream = prop { list: List[Int] =>
+    unsafeRunSync(
+      Stream
+        .fromIterable(list)
+        .buffer(2)
+        .run(Sink.collectAll[Int])
+    ) must_== (Success(list))
+  }
+
+  private def bufferStreamError = {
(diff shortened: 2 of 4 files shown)


vasilmkd:
@iravid can you please assist me with the tests? Thank you.

iravid:
@vasilmkd Hey, nice work on this! The test works exactly as you described it. Happy to accept improvements if you have any suggestions :-)

vasilmkd:
I'm curious why the test was changed.

iravid:
I think I changed it a while back because it wasn't very accurate before.

vasilmkd:
To be honest I have no idea how to test the other buffer implementations.

iravid:
For the unbounded case, I think your tests are fine. 

For the other implementations, you need to work with a couple of latches to control the stream's progress, apply backpressure after the first element and release the backpressure once enough elements have passed. Then you consume the rest of the stream and verify that the intermediate elements have been dropped, in  the sliding case, or that the last elements were dropped in the Dropping case.

vasilmkd:
@iravid I require help for the failing `bufferSliding` test. The implementation of `bufferSliding` seems fairly innocent to me and should work.
I was much more worried about `bufferDropping` but it seems I have done a decent enough job of implementing. I don't believe this was anticipated when this issue was opened.

iravid:
Ok, looking now!

iravid:
Oh yes, I definitely forgot about the fact that we're signalling `Take.End` and `Take.Fail` too.

vasilmkd:
@iravid I'm happy to leave this for after the hackathon.

vasilmkd:
I started a fresh `StreamSpec2` because any additions that I made to `StreamSpec` resulted in `UTF8 string too long` compiler error, very nondescript.

vasilmkd:
As a bonus, I have reduced allocations in `unTake` by not going through `Option`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -595,6 +595,72 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+              .foreachManaged(a => queue.offer(Take.Value(a)).unit)
+              .foldCauseM(
+                c => signal.fail(c).unit.toManaged_,
+                _ => signal.succeed(()).unit.toManaged_
+              )
+              .fork
+      } yield (queue, signal)
+    }.flatMap {
+      case (queue, signal) =>
+        def pull: UIO[Take[E1, A1]] =
```
iravid: Very clever! My only comment here is that this is a busy-wait when the queue is empty and the stream has not completed yet.

Have you considered implementing this using two queues? One for the signal and one for the data? Then you can race the takes.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -595,6 +595,72 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+   * (dropping or sliding). It implements a promise that signals abrupt completion
+   * of the stream due to an error or the end of the stream. After the promise has
+   * been completed, it continues to offer the current (at the time) contents of the
+   * queue, until it has been exhausted, at which point it finishes with the signal
+   * from the promise.
+   */
+  private final def bufferSignal[E1 >: E, A1 >: A](queue: Queue[Take[E1, A1]]): ZStream[R, E1, A1] =
+    ZStream.managed {
+      for {
+        signal <- Promise.make[Cause[E], Unit].toManaged_
```
iravid: The `Cause[E]` here is basically redundant because you can always recover it by using `Promise#halt` from the producer side and using `await.sandbox` or `await.run` from the consumer side

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: I'm aware yes, that's why I wanted your input on this. I'll rework it. Thanks.

iravid: Ok 👍 If you do end up using this approach, use `raceWith` and keep the losing fiber in a ref, don't cancel it. I think there are some edge conditions where if you cancel a `queue.take`, you might lose the element.

vasilmkd: What do I do with the ref in the end?

iravid: You re-race the fiber's `join` against `data.take` until the fiber succeeds, then you wait for the queue to empty as you've done here

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Thanks, I'll keep an eye out next time.

This has become irrelevant for the new implementation.

vasilmkd:
@iravid Thanks for your help. Ready for review.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -596,6 +596,68 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                _ => signal.succeed(Take.End).unit.toManaged_
+              )
+              .fork
+      } yield (queue, fiber)
+    }.flatMap {
+      case (queue, fiber) =>
+        val pull = queue.poll.flatMap {
+          _.fold[Pull[Any, Nothing, Take[E1, A1]]](
+            queue.take.raceWith(fiber.join)(
+              (l, _) => l.fold(Pull.halt, Pull.emit),
```
iravid: This leaks the joining fiber every time, so make sure you cancel it

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -596,6 +596,68 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+              (r, f) => f.interrupt *> r.fold(Pull.halt, Pull.emit)
```
iravid: This potentially skips over items that are still in the buffer.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -596,6 +596,68 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+   * (dropping or sliding). It implements a promise that signals abrupt completion
+   * of the stream due to an error or the end of the stream. After the promise has
+   * been completed, it continues to offer the current (at the time) contents of the
+   * queue, until it has been exhausted, at which point it finishes with the signal
+   * from the promise.
+   */
+  private final def bufferSignal[E1 >: E, A1 >: A](queue: Queue[Take[E1, A1]]): ZStream[R, E1, A1] =
+    ZStream.managed {
+      for {
+        signal <- Promise.make[Nothing, Take[E, Nothing]].toManaged_
```
iravid: I think you can get rid of `Take` entirely here by emitting plain values in the queue and making this promise a `Promise[Option[E], Nothing]`. Unifying the two gives you `Pull`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Isn't this the same fiber passed from the previous `flatMap`? It will get collected when the `Managed` scope ends. Or am I wrong?

vasilmkd: Or `raceWith` constructs another fiber always?

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: What about the `Cause`?

vasilmkd:
@iravid This is ready for review once again. Hopefully the code comments make sense.

I'd like to know whether we should keep `toUnboundedQueue`. Same in `ZStreamChunk`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -596,6 +596,109 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                  if (_) {
+                    // The stream has signalled error or end.
+                    take.isDone.flatMap {
+                      if (_) {
+                        // However, the queue is not empty, so we take the next element.
+                        // We don't interrupt any fibers here because they have both
+                        // completed their execution.
+                        take.await
+                      } else {
+                        // Otherwise, the queue is empty, so we interrupt the fiber
```
iravid: Why does `take.isDone == false` mean that the queue is empty?

iravid: Oh I understand why you went down that route. That's a race condition. There's no guarantee that the take has completed when you check it. It's safer to check the queue's size.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1996,15 +2099,25 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
     for {
       queue <- ZManaged.make(Queue.bounded[Take[E1, A1]](capacity))(_.shutdown)
       _     <- self.intoManaged(queue).fork
     } yield queue
 
+  /**
+   * Converts the stream into an unbounded managed queue. After the managed queue is used,
+   * the queue will never again produce values and should be discarded.
+   */
+  final def toUnboundedQueue[E1 >: E, A1 >: A]: ZManaged[R, Nothing, Queue[Take[E1, A1]]] =
```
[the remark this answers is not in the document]
vasilmkd: Will do.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Does that mean I can get rid of the `take` `Promise` as well?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2091,10 +2204,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    */
   final def unTake[E1 >: E, A1](implicit ev: A <:< Take[E1, A1]): ZStream[R, E1, A1] = {
     val _ = ev
-    self
-      .asInstanceOf[ZStream[R, E, Take[E1, A1]]]
-      .mapM(t => Take.option(UIO.succeed(t)))
-      .collectWhile { case Some(a) => a }
+    ZStream(self.process.map(_.flatMap(Pull.fromTake(_))))
```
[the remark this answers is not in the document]
vasilmkd: Can't wait to have this merged just because of this change. 😇

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec2.scala
```diff
@@ -0,0 +1,154 @@
+package zio.stream
+
+import org.specs2.ScalaCheck
+
+import scala.{ Stream => _ }
+import zio._
+import zio.ZQueueSpecUtil.waitForSize
+
+class StreamSpec2(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime with GenIO with ScalaCheck {
```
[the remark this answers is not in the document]
vasilmkd: Done.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Let's see if we're finally there.

iravid:
@vasilmkd Yes, this one looks good! There's still a slim chance for a skipped element there in `signalFirst` but I don't know how to solve that currently. For this type of combinator I think this is good enough though.

Could you do a rebase?

vasilmkd:
@iravid Rebased.

vasilmkd:
My pleasure. Just let's wait for CI. I don't trust my rebasing skills yet.

PR #1686, opened 2019-09-16
Fixes and inference improvements to ZSink combinators

### LINKED ISSUE #1509: Does Sink really need a distinct leftover type parameter?
*To be discussed on one of the hackathons. Or here, if anyone wants to chime in.*

Data points:
- All sinks I've used either had `A0 = A`, or `A0 = Nothing`
- We have just two constructors (`Sink.fold`, `Sink.foldM`) that create sinks with `A0 != A`

It looks like we could just do with always returning a `Chunk[A]` as the leftover?

### LINKED ISSUE #1676: Refactor ZSink#collectAllWhileWith to not use stepChunk
And use stepChunkSlice instead, which doesn't discard elements.

### LINKED ISSUE #1682: ZSink#takeWhile continuation function is wrong
Should be
`def cont(state: State) = state._2.isEmpty && self.cont(state._1)`.

The sink is done whenever a stream element does not pass the predicate and is returned as remainder.

iravid:
Resolves #1682 
Resolves #1676

@vasilmkd - could you have a look? Thanks!

### DIFF
--- a/docs/datatypes/sink.md
+++ b/docs/datatypes/sink.md
@@ -3,9 +3,16 @@ id: datatypes_sink
 title:  "Sink"
 ---
 
-A `Sink[E, A0, A, B]` is used to consume elements produced by `stream`
-with `E` error in case of failure, `A0` remainder, `A` input elements and
-`B` result type. `Sink` is passed as `stream.run` argument:
+A `Sink[E, A0, A, B]` is used to consume elements produced by a stream.
+You can think of this sink as a function that will consume a variable 
+amount of `A` elements (could be 0, 1, or many!), might fail with an
+error of type `E`, and will eventually yield a value of type `B`.
+
+The `A0` parameter describes the sink's *leftover* type. Sinks might
+not consume all their inputs, and unconsumed elements are returned as
+chunks of `A0` values.
+
+A `Sink` is passed to `ZStream#run` as an argument:
 
 ```scala mdoc:silent
 import zio._
@@ -76,7 +83,7 @@ A fold with control over continuation and the remainder produced by the sink.
 `foldLeft` is implemented as a fold that always continues and produces no remainder.
 
 ```scala mdoc:silent
-Sink.fold[Nothing, Int, Int](0)(_ => true)((acc, n) => (acc + n, Chunk.empty))
+Sink.fold(0)(_ => true)((acc, n: Int) => (acc + n, Chunk.empty))
 ```
 
 Mapping over the received input elements:
@@ -89,8 +96,9 @@ Sink.fromFunction[Int, Int](_ * 2).collectAll
 
 ```scala mdoc:silent
 Sink
-  .pull1[String, Int, Int, Int](IO.fail("Empty stream, no value to pull"))(init => Sink
-  .foldLeft[Int, Int](init)(_ + _))
+  .pull1[String, Int, Int, Int](IO.fail("Empty stream, no value to pull")) { init =>
+    Sink.foldLeft[Int, Int](init)(_ + _)
+  }
 ```
 
 `read1` tries to read head element from stream,
@@ -106,12 +114,12 @@ Sink.read1[String, Int] {
 ## Transforming sinks
 
 Having created the sink, we can transform it with provided operations.
-One of them already appeared in previous section - `collectAll` in `read1`.
+One of them already appeared in previous section - `collectAll` in `pull1`.
 
 Sink that after collecting input - filters it:
 
 ```scala mdoc:silent
-Sink.collectAll[Int].filter[Int](_ > 100)
+Sink.collectAll[Int].filter(_ > 100)
 ```
 
 Running two sinks in parallel and returning the one that completed earlier:
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -37,10 +37,12 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
       leftover append $chunkedLeftoverAppend
 
     collectAll
-      happy path    $collectAllHappyPath
-      init error    $collectAllInitError
-      step error    $collectAllStepError
-      extract error $collectAllExtractError
+      happy path                   $collectAllHappyPath
+      init error                   $collectAllInitError
+      step error                   $collectAllStepError
+      extract error                $collectAllExtractError
+      interaction with succeed     $collectAllSucceed
+      interaction with ignoreWhile $collectAllIgnoreWhile
 
     collectAllN
       happy path          $collectAllNHappyPath
@@ -376,12 +378,16 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
 
   private def chunkedHappyPath = {
     val sink = ZSink.collectAll[Int].chunked
-    unsafeRun(sinkIteration(sink, Chunk(1, 2, 3, 4, 5)).map(_ must_=== ((List(1, 2, 3, 4, 5), Chunk.empty))))
+    unsafeRun(sinkIteration(sink, Chunk(1, 2, 3, 4, 5)).map {
+      case (b, leftovers) => (b must_=== List(1, 2, 3, 4, 5)) and (leftovers must_=== Chunk.empty)
+    })
   }
 
   private def chunkedEmpty = {
     val sink = ZSink.collectAll[Int].chunked
-    unsafeRun(sinkIteration(sink, Chunk.empty).map(_ must_=== ((Nil, Chunk.empty))))
+    unsafeRun(sinkIteration(sink, Chunk.empty).map {
+      case (b, leftovers) => (b must_=== Nil) and (leftovers must_=== Chunk.empty)
+    })
   }
 
   private def chunkedInitError = {
@@ -402,10 +408,11 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
   private def chunkedLeftover = {
     val sink = ZSink.collectAllN[Int](2).chunked
     val test = for {
-      init   <- sink.initial
-      step   <- sink.step(init, Chunk(1, 2, 3, 4, 5))
-      result <- sink.extract(step)
-    } yield result must_=== ((List(1, 2), Chunk(3, 4, 5)))
+      init           <- sink.initial
+      step           <- sink.step(init, Chunk(1, 2, 3, 4, 5))
+      result         <- sink.extract(step)
+      (b, leftovers) = result
+    } yield (b must_=== List(1, 2)) and (leftovers must_=== Chunk(3, 4, 5))
     unsafeRun(test)
   }
 
@@ -439,15 +446,45 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
+  private def collectAllSucceed = {
+    val sink = ZSink.succeed[Int, Int](5).collectAll
+    val test = for {
+      init <- sink.initial
+      s <- sink
+            .step(init, 1)
+            .flatMap(sink.step(_, 2))
+            .flatMap(sink.step(_, 3))
+      result <- sink.extract(s)
+    } yield result must_=== (List(5, 5, 5, 5) -> Chunk(1, 2, 3))
+
+    unsafeRun(test)
+  }
+
+  private def collectAllIgnoreWhile = {
+    val sink = ZSink.ignoreWhile[Int](_ < 5).collectAll
+    val test = for {
+      result <- sink.initial
+                 .flatMap(sink.step(_, 1))
+                 .flatMap(sink.step(_, 2))
+                 .flatMap(sink.step(_, 3))
+                 .flatMap(sink.step(_, 5))
+                 .flatMap(sink.step(_, 6))
+                 .flatMap(sink.extract)
+    } yield result must_=== (List((), ()) -> Chunk(5, 6))
+
+    unsafeRun(test)
+  }
+
   private def collectAllNHappyPath = {
     val sink = ZSink.identity[Int].collectAllN(3)
     val test = for {
-      init   <- sink.initial
-      step1  <- sink.step(init, 1)
-      step2  <- sink.step(step1, 2)
-      step3  <- sink.step(step2, 3)
-      result <- sink.extract(step3)
-    } yield result must_=== ((List(1, 2, 3), Chunk.empty))
+      result <- sink.initial
+                 .flatMap(sink.step(_, 1))
+                 .flatMap(sink.step(_, 2))
+                 .flatMap(sink.step(_, 3))
+                 .flatMap(sink.step(_, 4))
+                 .flatMap(sink.extract)
+    } yield result must_=== ((List(1, 2, 3), Chunk(4)))
     unsafeRun(test)
   }
 
@@ -489,9 +526,8 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
   }
 
   private def collectAllWhileFalsePredicate = {
-    val errorMsg = "No elements have been consumed by the sink"
-    val sink     = ZSink.identity[Int].collectAllWhile(_ < 0).asError(errorMsg)
-    unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left(errorMsg)))
+    val sink = ZSink.identity[Int].collectAllWhile(_ < 0)
+    unsafeRun(sinkIteration(sink, 1).map(_ must_=== (List() -> Chunk.single(1))))
   }
 
   private def collectAllWhileInitError = {
@@ -570,47 +606,47 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
   }
 
   private def dropWhileHappyPath = {
-    val sink = ZSink.identity[Int].dropWhile[Int](_ < 5)
+    val sink = ZSink.identity[Int].dropWhile(_ < 5)
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left(())))
   }
 
   private def dropWhileFalsePredicate = {
-    val sink = ZSink.identity[Int].dropWhile[Int](_ > 5)
+    val sink = ZSink.identity[Int].dropWhile(_ > 5)
     unsafeRun(sinkIteration(sink, 1).map(_ must_=== ((1, Chunk.empty))))
   }
 
   private def dropWhileInitError = {
-    val sink = initErrorSink.dropWhile[Int](_ < 5)
+    val sink = initErrorSink.dropWhile(_ < 5)
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
   private def dropWhileStepError = {
-    val sink = stepErrorSink.dropWhile[Int](_ < 5)
+    val sink = stepErrorSink.dropWhile(_ < 5)
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
   private def dropWhileExtractError = {
-    val sink = extractErrorSink.dropWhile[Int](_ < 5)
+    val sink = extractErrorSink.dropWhile(_ < 5)
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
   private def flatMapHappyPath = {
-    val sink = ZSink.identity[Int].flatMap(n => ZSink.succeed(n.toString))
+    val sink = ZSink.identity[Int].flatMap(n => ZSink.succeed[Int, String](n.toString))
     unsafeRun(sinkIteration(sink, 1).map(_ must_=== (("1", Chunk.empty))))
   }
 
   private def flatMapInitError = {
-    val sink = initErrorSink.flatMap(n => ZSink.succeed(n.toString))
+    val sink = initErrorSink.flatMap(n => ZSink.succeed[Int, String](n.toString))
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
   private def flatMapStepError = {
-    val sink = stepErrorSink.flatMap(n => ZSink.succeed(n.toString))
+    val sink = stepErrorSink.flatMap(n => ZSink.succeed[Int, String](n.toString))
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
   private def flatMapExtractError = {
-    val sink = extractErrorSink.flatMap(n => ZSink.succeed(n.toString))
+    val sink = extractErrorSink.flatMap(n => ZSink.succeed[Int, String](n.toString))
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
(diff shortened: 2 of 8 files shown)


streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -439,15 +423,45 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
     unsafeRun(sinkIteration(sink, 1).either.map(_ must_=== Left("Ouch")))
   }
 
+  private def collectAllSucceed = {
```
vasilmkd: Thanks for this test.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -439,15 +423,45 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
...
+            .step(init, 1)
+            .flatMap(sink.step(_, 2))
+            .flatMap(sink.step(_, 3))
+      result <- sink.extract(s)
+    } yield result must_=== (List(5, 5, 5, 5) -> Chunk(1, 2, 3))
+
+    unsafeRun(test)
+  }
+
+  private def collectAllIgnoreWhile = {
```
vasilmkd: And this one.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -489,9 +503,8 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
   }
 
   private def collectAllWhileFalsePredicate = {
```
vasilmkd: Fixed!

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -509,88 +522,28 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
...
-    val sink = stepErrorSink.dimap[String, String](_.toInt)(_.toString.reverse)
-    unsafeRun(sinkIteration(sink, "123").either.map(_ must_=== Left("Ouch")))
-  }
-
-  private def dimapExtractError = {
-    val sink = extractErrorSink.dimap[String, String](_.toInt)(_.toString.reverse)
-    unsafeRun(sinkIteration(sink, "123").either.map(_ must_=== Left("Ouch")))
-  }
-
   private def dropWhileHappyPath = {
```
vasilmkd: Inference win, great to see.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -1027,8 +967,9 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
       step3  <- sink.step(step2, 3)
       step4  <- sink.step(step3, 4)
       step5  <- sink.step(step4, 5)
+      cont   = sink.cont(step5)
       result <- sink.extract(step5)
-    } yield result must_=== ((List(1, 2, 3, 4), Chunk.single(5)))
+    } yield (cont must_=== false) and (result must_=== ((List(1, 2, 3, 4), Chunk.single(5))))
```
vasilmkd: Nice.

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -1053,15 +994,17 @@ class SinkSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRunt
...
-      result <- sink.extract(step4)
-    } yield result must_=== ((Some(List(1, 2, 3, 4)), Chunk.empty))
+      under <- sink.initial
+                .flatMap(sink.stepChunk(_, Chunk(1, 2)).map(_._1))
+                .flatMap(sink.extract)
+      over <- sink.initial
+               .flatMap(sink.stepChunk(_, Chunk(1, 2)).map(_._1))
+               .flatMap(sink.stepChunk(_, Chunk(2, 2)).map(_._1))
+               .flatMap(sink.extract)
+    } yield (under must_=== (None -> Chunk.empty)) and (over must_=== (Some(List(1, 2, 2)) -> Chunk(2)))
```
vasilmkd: Awesome test.

streams/shared/src/main/scala/zio/stream/Sink.scala
```diff
@@ -176,120 +176,116 @@ object Sink {
...
     f: (S, A) => S
-  ): Sink[Nothing, A, A, S] =
+  ): Sink[Nothing, A, S] =
     ZSink.foldWeightedDecompose(z)(costFn, max, decompose)(f)
 
   /**
    * see [[ZSink.fromEffect]]
    */
-  final def fromEffect[E, B](b: => IO[E, B]): Sink[E, Nothing, Any, B] =
+  final def fromEffect[E, B](b: => IO[E, B]): Sink[E, Any, B] =
```
vasilmkd: I'm confused. `succeed` gets a type parameter `A`, but this doesn't? How does this compose?

docs/datatypes/sink.md
```diff
@@ -76,7 +79,7 @@ A fold with control over continuation and the remainder produced by the sink.
 `foldLeft` is implemented as a fold that always continues and produces no remainder.
 
 ```scala mdoc:silent
-Sink.fold[Nothing, Int, Int](0)(_ => true)((acc, n) => (acc + n, Chunk.empty))
+Sink.fold(0)(_ => true)((acc, n: Int) => (acc + n, Chunk[Int]()))
```
jdegoes: If this is going to be the new pattern, maybe `Chunk.empty` should take a type parameter like `List.empty`?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -56,40 +56,42 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+  final def <|[R1 <: R, E1, B1 >: B](
+    that: ZSink[R1, E1, A, B1]
+  ): ZSink[R1, E1, A, B1] =
     (self orElse that).map(_.merge)
 
-  private[ZSink] final def ?[A00 >: A0, A1 <: A](implicit ev: A1 =:= A00): ZSink[R, Nothing, A00, A1, Option[B]] =
-    new ZSink[R, Nothing, A00, A1, Option[B]] {
+  final def optional = ?
+
+  final def ? : ZSink[R, Nothing, A, Option[B]] =
```
vasilmkd: It seems that we have lost the scaladoc when deleting the syntax classes.

docs/datatypes/sink.md
vasilmkd: Agreed.

On the other hand, would something like `ZSink[-R, +E, -A, +B]` with `def extract[A1 <: A](state: State): ZIO[R, E, (B, Chunk[A1])]` work?

@jdegoes

streams/shared/src/main/scala/zio/stream/Sink.scala
iravid: Ah good catch. Every sink that has  `A = Any` should be audited. Most of them would need to be polymorphic in `A`. Will fix that.

docs/datatypes/sink.md
iravid: Yes, good point.

streams/shared/src/main/scala/zio/stream/ZSink.scala
iravid: 👍🏻

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1405,7 +1403,7 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
     val s2 = Stream(1, 2)
 
     s1.mergeWith(s2)(_.toString, _.toString)
-      .run(Sink.succeed("done"))
+      .run(Sink.succeed[String, String]("done"))
```
jdegoes: Arg. This is one of the places where it's a loss.

docs/datatypes/sink.md
iravid: @vasilmkd It has to be `[A1 >: A]`, actually. 

I'll give that a try in a follow-up. I tried that for a bit but some of the  combinators that extract an underlying sink get confused so I scrapped that direction for now.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
iravid: @jdegoes Oh it's good to have the grandmaster of variance and inference here.

There's a really annoying issue that the Stream's  element type parameter  isn't used to infer the Sink's input type parameter - in `run`, `transduce` and others. Any thoughts on that?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -142,148 +144,129 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
-  )(f: (S, B) => S)(implicit ev: A00 =:= A1, ev2: A1 =:= A00): ZSink[R, E, A00, A1, S] =
-    collectAllWhileWith[S, A00, A1](_ => true)(z)(f)
-
-  private[ZSink] final def collectAllWhile[A00 >: A0, A1 <: A](
-    p: A1 => Boolean
-  )(implicit ev: A00 =:= A1, ev2: A1 =:= A00): ZSink[R, E, A00, A1, List[B]] =
-    collectAllWhileWith[List[B], A00, A1](p)(List.empty[B])((bs, b) => b :: bs)
-      .map(_.reverse)
+  final def collectAll: ZSink[R, E, A, List[B]] =
+    collectAllWith[List[B]](List.empty[B])((bs, b) => b :: bs).map(_.reverse)
```
vasilmkd: Can be simplified. Either `List.empty[B]` with no type parameter or `Nil`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -142,148 +144,129 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
 
-  private[ZSink] final def collectAllWhileWith[S, A00 >: A0, A1 <: A](
-    p: A1 => Boolean
-  )(z: S)(f: (S, B) => S)(implicit ev: A00 =:= A1, ev2: A1 =:= A00): ZSink[R, E, A00, A1, S] =
-    new ZSink[R, E, A00, A1, S] {
-      type State = (S, self.State, Boolean, Chunk[A00])
+  final def collectAllN(i: Int): ZSink[R, E, A, List[B]] =
+    new ZSink[R, E, A, List[B]] {
+      case class State(s: self.State, bs: List[B], n: Int, leftover: Chunk[A], dirty: Boolean)
```
vasilmkd: This is totally new to me. One could implement abstract types as classes. Seems so simple and logical but wow. :boom:

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
jdegoes: Can you paste signatures and I'll make it work (if it's possible 😆 )? Thank you! 🙏

streams/shared/src/main/scala/zio/stream/Sink.scala
jdegoes: Good catch! 👍

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -142,148 +144,129 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
-      .map(_.reverse)
+  final def collectAll: ZSink[R, E, A, List[B]] =
+    collectAllWith[List[B]](List.empty[B])((bs, b) => b :: bs).map(_.reverse)
 
-  private[ZSink] final def collectAllWhileWith[S, A00 >: A0, A1 <: A](
-    p: A1 => Boolean
-  )(z: S)(f: (S, B) => S)(implicit ev: A00 =:= A1, ev2: A1 =:= A00): ZSink[R, E, A00, A1, S] =
-    new ZSink[R, E, A00, A1, S] {
-      type State = (S, self.State, Boolean, Chunk[A00])
+  final def collectAllN(i: Int): ZSink[R, E, A, List[B]] =
```
vasilmkd: I like this a lot.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
iravid:
```scala
trait ZSink[-R, +E, A, +B]
object ZSink {
  final def succeed[A, B](b: B): ZSink[Any, Nothing, A, B] = new ZSink[Any, Nothing, A, B] {}
}

trait ZStream[-R, +E, +A] {
  def run[R1 <: R, E1 >: E, A1 >: A, B](sink: ZSink[R1, E1, A1, B]): Unit = ()
}

val stream = new ZStream[Any, Nothing, String] {}
stream.run(ZSink.succeed("OK"))

// cmd0.sc:11: type mismatch;
// found   : ammonite.$sess.cmd0.ZSink[Any,Nothing,Nothing,String]
// required: ammonite.$sess.cmd0.ZSink[Any,Nothing,String,String]
```

iravid: @jdegoes ^

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -463,36 +446,48 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
       def extract(state: State)    = self.extract(state).mapError(f)
       def cont(state: State)       = self.cont(state)
     }
 
-  /**
-   * Effectfully maps the value produced by this sink.
-   */
-  final def mapM[R1 <: R, E1 >: E, C](f: B => ZIO[R1, E1, C]): ZSink[R1, E1, A0, A, C] =
-    new ZSink[R1, E1, A0, A, C] {
+  final def mapInput[C](f: C => A)(g: A => C): ZSink[R, E, C, B] =
```
vasilmkd: Could use a scaladoc. Same with `mapInputM`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -508,29 +503,29 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
    * Right: ===== SUCCEEDS!
    *             xxxxxxxxx <- Should NOT be consumed
    */
-  final def orElse[R1 <: R, E1, A00 >: A0, A1 <: A, C](
-    that: ZSink[R1, E1, A00, A1, C]
-  )(implicit ev: A1 =:= A00): ZSink[R1, E1, A00, A1, Either[B, C]] =
-    new ZSink[R1, E1, A00, A1, Either[B, C]] {
+  final def orElse[R1 <: R, E1, C](
```
vasilmkd: Feels good to see the complex combinators migrated without too much hassle. It looks like we got them right last time. :smile:

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1111,119 +973,119 @@ object ZSink extends ZSinkPlatformSpecific {
...
+    foldM[R, E, A, (List[A], Boolean)]((Nil, true))(_._2) {
       case ((as, _), a) =>
         p(a).map(if (_) ((a :: as, true), Chunk.empty) else ((as, false), Chunk.single(a)))
     }.map(_._1.reverse)
 
   /**
    * Creates a sink halting with the specified `Throwable`.
    */
-  final def die(e: Throwable): ZSink[Any, Nothing, Nothing, Any, Nothing] =
+  final def die(e: Throwable): ZSink[Any, Nothing, Any, Nothing] =
```
vasilmkd: Just to mark them for having `Any`, `die`, `dieMessage` and `drain`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1446,15 +1308,15 @@ object ZSink extends ZSinkPlatformSpecific {
    * Creates a sink by starts consuming value as soon as one fails
    * the predicate `p`.
    */
-  final def ignoreWhile[A](p: A => Boolean): ZSink[Any, Nothing, A, A, Unit] =
+  final def ignoreWhile[A](p: A => Boolean): ZSink[Any, Nothing, A, Unit] =
```
vasilmkd: I'd like this to be a `SinkPure` and not delegate through `ignoreWhileM`. Ignore if you think it's not worth it.

vasilmkd:
Another day another `ZSink` refactor. :laughing:

Initial review done. I will be around for follow up changes.

streams/shared/src/main/scala/zio/stream/ZSink.scala
iravid: Yeah we'll do that in a follow-up. There are several combinators that can be overriden in `SinkPure`.

iravid:
Hmm. So, I went over the changes again and I admit I am not sold on whether the inference changes are a win or not. So I'm going to backpedal on this and keep `A0` separate.

vasilmkd:
I stand by your decision.

iravid:
@vasilmkd I moved the definitions of combinators that use `A0 = A` and have problematic inference into `InputRemainderOps`. Figured keeping the private methods on the trait was redundant.

vasilmkd:
@iravid I saw that and like it a lot. Big :+1:. It will take some more consideration when implementing future combinators but the removal of implicit compiler evidence is worth it in my opinion.

Can I bother you to go through the tests again and try to remove as much of the type parameters as possible? Just as a testament to the improved inference.

streams/shared/src/main/scala/zio/stream/Sink.scala
```diff
@@ -242,13 +242,9 @@ object Sink {
   /**
    * see [[ZSink.succeed]]
    */
-  final def succeed[B](b: B): Sink[Nothing, Nothing, Any, B] =
+  final def succeed[A, B](b: B): Sink[Nothing, A, A, B] =
```
vasilmkd: `A` should be reverted now. 😄

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -947,127 +665,428 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
     that: ZSink[R1, E1, A00, A1, C]
   )(f: (B, C) => D)(implicit ev: A00 =:= A1, ev2: A1 =:= A00): ZSink[R1, E1, A00, A1, D] =
     zip(that).map(f.tupled)
-
 }
 
 object ZSink extends ZSinkPlatformSpecific {
 
-  implicit class InputRemainderOps[R, E, A, B](val sink: ZSink[R, E, A, A, B]) extends AnyVal {
+  implicit class InputRemainderOps[R, E, A, B](private val sink: ZSink[R, E, A, A, B]) {
```
vasilmkd: `extends AnyVal`

streams/shared/src/main/scala/zio/stream/Sink.scala
iravid: It's actually required because of the leftovers.

streams/shared/src/main/scala/zio/stream/ZSink.scala
iravid: Can't work because of the inner classes that are defined. Not a big loss.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -947,127 +665,428 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+        def extract(state: State) = self.extract(state._1)
+
+        def cont(state: State) = self.cont(state._1)
+      }
+
+    /**
+     * Filters the inputs fed to this sink.
+     */
+    def filter(f: A => Boolean): ZSink[R, E, A0, A, B] =
+      self match {
```
vasilmkd: Nitpicking. Not sure how I feel about having `SinkPure` implementations in this file from a discoverability and maintainability standpoint.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -947,127 +665,428 @@ trait ZSink[-R, +E, +A0, -A, +B] { self =>
...
+              val results   = list.map { case (k, (b, _)) => (k, b) }.toMap
+              val leftovers = Chunk.fromIterable(list.map { case (k, (_, chunk)) => (k, chunk) }).filter(_._2.notEmpty)
+              (results, leftovers)
+            }
+
+        def cont(state: State) = state.values.forall(self.cont)
+      }
   }
 
   private[ZSink] object internal {
```
vasilmkd: Can the `Optional` ADT be migrated to `?` as well? OTOH Not sure if `Side` is used in multiple combinators.

streams/shared/src/main/scala/zio/stream/Sink.scala
vasilmkd: Sorry yes, I skimmed over the implementation.

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: Sorry, can you point me to the defined classes? I can't find anything.

iravid: `case class State` in `collectAllN`

streams/shared/src/main/scala/zio/stream/ZSink.scala
iravid: Well, since `filter` was moved out to an implicit class to improve inference, it can no longer be overriden in `SinkPure.

streams/shared/src/main/scala/zio/stream/ZSink.scala
iravid: Yup! Moved it

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: Yes, thanks.

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: I'm well aware... we'll live with it. :-)

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1606,19 +1664,15 @@ object ZSink extends ZSinkPlatformSpecific {
...
    * Creates a single-value sink from a value.
    */
-  final def succeed[B](b: B): ZSink[Any, Nothing, Nothing, Any, B] =
-    new SinkPure[Nothing, Nothing, Any, B] {
-      type State = Unit
-      val initialPure                    = ()
-      def stepPure(state: State, a: Any) = ()
-      def extractPure(state: State)      = Right((b, Chunk.empty))
-      def cont(state: State)             = false
+  final def succeed[A, B](b: B): ZSink[Any, Nothing, A, A, B] =
```
vasilmkd: I'm guessing this cannot be inferred from the sink before this one? :/

iravid: No, unfortunately not. Same problem as with `ZStream#run`

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
iravid: Fixed in Dotty, at least

jdegoes: I think this case is beyond help. The covariance on `A` in `ZStream` is interacting with the invariance on `ZSink` to cause the problem, and we certainly don't want to eliminate covariance on `ZStream`

PR #1688, opened 2019-09-16
Migrate TarraySpec to ZIOTest

denisgarci:
Migrate TarraySpec to ZIOTest. A few tests are not passing.


core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -0,0 +1,204 @@
...
+            for {
+              tArray <- makeTArray(1)(42)
+              value  <- tArray(0).commit
+            } yield assert(value, equalTo(42))
+          },
+          testM("dies with ArrayIndexOutOfBounds when index is out of bounds $applyOutOfBounds") {
+            assertM(for {
+              tArray <- makeTArray(1)(42)
+              _      <- tArray(-1).commit
+            } yield (), throws(equalTo(FiberFailure)))
```
denisgarci: This doesn't work. Not sure how to write that assert.

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -0,0 +1,204 @@
...
+              last           <- tArray.array(N - 1).get.commit
+            } yield assert((first, last), equalTo(("a+b+c", "a+b+c")) || equalTo(("a+c+b", "a+c+b")))
+          },
+          testM("updates all or nothing") {
+            for {
+              tArray <- makeTArray(N)(0)
+              _      <- tArray.array(N / 2).update(_ => 1).commit
+              result <- tArray.transformM(a => if (a == 0) STM.succeed(42) else STM.fail(boom)).commit.either
+              first  <- tArray.array(0).get.commit
+            } yield assert(result.left.map(r => (first, r)), isLeft(equalTo((0, boom))))
```
denisgarci: This is not great too. Not sure how to improve that. The original was just
```
(first, result) === (0, Left(boom))
```

LGLO:
I originally implemented these tests. I'll try to help within 24 hours, I have to take a look at zio-test...

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
LGLO: Expected behavior is that `Fiber` running `TArray#appy(INVALID_INDEX)` will DIE, because it's considered an fatal error. So I _guess_ there are two things to change:
* use `.run` combinator to get some result or exit value from apply
* because we want to ensure DIE of a fiber, then `dies` instead of `throws`.

denisgarci:
@LGLO Thanks for writing those tests! Your help would be appreciated.

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
LGLO: It's quite more complicated, since `new ArrayIndexOutOfBoundsException(-1)` is not equal to `new ArrayIndexOutOfBoundsException(-1)`, we need some other assertion. I didn't manage to use `isSubtype` - I don't know how :man_shrugging: 

I managed to fix this with creating custom assertion:
`Assertion.assertion[Throwable]("isArrayIndexOutOfBoundsException")()(_.isInstanceOf[ArrayIndexOutOfBoundsException]))`

A note to my first comment: `run` needs to be applied on a ZIO that contains call to apply(-1) inside. Because calling `tArray(-1)` is analogous to `Future.succeed(throw new Excepion)`.

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
LGLO: Looks good for me!
Alternatively you could manually annotate type `assert((first, result), equalTo((0, Right(()): Either[Exception, Unit])))` but I don't find it superior in any way.
Maybe there is some way of improving type inference for `equalTo`, but perhaps out of scope here.

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
denisgarci: Here is what I have so far:
```
val result = for {
    tArray <- makeTArray(1)(42)
    r      <- ZIO.effect(tArray(-1)).run
} yield r.toEither
assertM(result, isLeft(Assertion.assertion[Throwable]("isArrayIndexOutOfBoundsException")()(_.isInstanceOf[ArrayIndexOutOfBoundsException])))
```
I get the following error while running the test:
```
[info]     - dies with ArrayIndexOutOfBounds when index is out of bounds
[info]       Left(zio.FiberFailure: Fiber failed.
[info] A checked error was not handled.
[info] java.lang.ArrayIndexOutOfBoundsException: Array index out of range: -1
[info] 	at zio.stm.TArray$.apply$extension(TArray.scala:27)
[info] 	at zio.stm.TArraySpec$$anonfun$$lessinit$greater$1.$anonfun$new$4(TArraySpec.scala:36)
[info] 	at zio.stm.TArraySpec$$anonfun$$lessinit$greater$1.$anonfun$new$4$adapted(TArraySpec.scala:36)
[info] 	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:403)
[info] 	at zio.internal.FiberContext.$anonfun$fork$1(FiberContext.scala:595)
[info] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info] 	at java.lang.Thread.run(Thread.java:745)
```
I couldn't get type inference to work with `dies` that's why I used `toEither`.

denisgarci: I made it work with the help @adamgfraser and your suggestion with the custom assertion. Thank you!

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -0,0 +1,204 @@
...
+            for {
+              tArray <- makeTArray(1)(42)
+              value  <- tArray(0).commit
+            } yield assert(value, equalTo(42))
+          },
+          testM("dies with ArrayIndexOutOfBounds when index is out of bounds") {
+            for {
+              tArray <- makeTArray(1)(42)
+              result     <- ZIO.effect(tArray(-1)).run
+            } yield assert(result, fails(Assertion.assertion[Throwable]("isArrayIndexOutOfBoundsException")()(_.isInstanceOf[ArrayIndexOutOfBoundsException])))
```
LGLO: I see you could get away from using `dies` by making `result` a `ZIO[_, _, STM[_, _]]` - nice!
I think you can make a helper `val` for this assertion and use it in two other tests.

In sbt please run `fmt` to auto-format code.

Thank you very much for your work!

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -0,0 +1,207 @@
...
+        ),
+        suite("collect")(
+          testM("is atomic") {
+            for {
+              tArray <- makeTArray(N)("alpha-bravo-charlie")
+              _      <- STM.foreach(tArray.array)(_.update(_.take(11))).commit.fork
+              collected <- tArray.collect {
+                            case a if a.length == 11 => a
+                          }.commit
+            } yield assert(collected.array.length, equalTo(0) || equalTo(N))
```
denisgarci: I couldn't make the new operator `hasSize` work I get the following error:
```
[error] /Users/dgarcia/personal/zio/core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala:48:65: type mismatch;
[error]  found   : zio.test.Assertion[Iterable[Nothing]]
[error]  required: zio.test.Assertion[Array[zio.stm.TRef[String]]]
[error]             } yield assert(collected.array, hasSize(equalTo(0)) || hasSize(equalTo(N)))
[error]                                                                 ^
```

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -0,0 +1,207 @@
...
+          },
+          testM("is safe for empty array") {
+            for {
+              tArray <- makeTArray(0)("nothing")
+              collected <- tArray.collect {
+                            case _ => ()
+                          }.commit
+            } yield assert(collected.array.isEmpty, isTrue)
```
denisgarci: Same here with `isEmpty` assertion.

PR #1694, opened 2019-09-17
Restore bracket on zio actions

regiskuckaertz:
The method existed only inside an implicit class for invariant types, but it doesn't seem necessary. Besides it's missing from the scaladoc which is quite surprising (since `bracket_` is there).

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -113,6 +113,19 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def bimap[E2, B](f: E => E2, g: A => B): ZIO[R, E2, B] = mapError(f).map(g)
 
   /**
+   * Shorthand for the uncurried version of `ZIO.bracket`.
+   */
+  final def bracket[R1 <: R, E1 >: E, B](
+    release: A => ZIO[R1, Nothing, _],
+    use: A => ZIO[R1, E1, B]
+  ): ZIO[R1, E1, B] = ZIO.bracket(self, release, use)
+
+  /**
+   * Shorthand for the curried version of `ZIO.bracket`.
+   */
+  final def bracket[R1 <: R, E1 >: E]: ZIO.BracketAcquire[R1, E1, A] = ZIO.bracket(self)
+
+  /**
    * A less powerful variant of `bracket` where the resource acquired by this
    * effect is not needed.
    */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2488,22 +2501,14 @@ object ZIO extends ZIOFunctions {
   private val _IdentityFn: Any => Any    = (a: Any) => a
   private[zio] def identityFn[A]: A => A = _IdentityFn.asInstanceOf[A => A]
 
-  implicit final class ZIOInvariant[R, E, A](private val self: ZIO[R, E, A]) extends AnyVal {
-    final def bracket: ZIO.BracketAcquire[R, E, A] =
-      new ZIO.BracketAcquire(self)
-
-    final def bracketExit: ZIO.BracketExitAcquire[R, E, A] =
```
neko-kai: forgot bracketExit!

regiskuckaertz: this one is already there in fact 🧐

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2488,22 +2501,14 @@ object ZIO extends ZIOFunctions {
...
-
   implicit final class ZIOAutocloseableOps[R, E, A <: AutoCloseable](private val io: ZIO[R, E, A]) extends AnyVal {
 
     /**
      * Like `bracket`, safely wraps a use and release of a resource.
      * This resource will get automatically closed, because it implements `AutoCloseable`.
      */
     def bracketAuto[R1 <: R, E1 >: E, B](use: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] =
-      io.bracket(a => UIO(a.close()))(use)
+      io.bracket[R1, E1, B](a => UIO(a.close()), use)
```
neko-kai: Doesn't that look like a reason for invariant syntax to be there? 🤔

jdegoes: Ah, right...

jdegoes: Oops. Backout?

regiskuckaertz: I guess so 😟 then remove `bracketExit` from the main class too?

jdegoes:
Probably so if we can show type inference issues...

regiskuckaertz:
I don't think it needs to be reverted: https://github.com/zio/zio/pull/1696

PR #1695, opened 2019-09-17
resolves #1533 update zstream repeat

### LINKED ISSUE #1533: Align ZStream#repeat with the other schedule combinators
The signatures should be:
```scala
final def repeatWith[R1 <: R, E1 >: E, B, C](schedule: ZSchedule[R1, Unit, B])(f: A => C, g: B => C): ZStream[R1 with Clock, E1, C]
final def repeatEither[R1 <: R, E1 >: E, B](schedule: ZSchedule[R1, Any, B]): ZStream[R1 with Clock, E1, Either[B,  A]] = repeatWith(schedule)(Left(_), Right(_))
final def repeat[R1 <: R, E1 >: E, B, C](schedule: ZSchedule[R1, Any, B]): ZStream[R1 with Clock, E1, A] = repeatEither(schedule) collect { case Right(a) => a }
```

The semantics should be as follows: when the stream ends, run the schedule. If the schedule wants to continue, apply the delay and repeat the stream. If the schedule wants to stop, end the stream.

mschuwalow:
One thing I'm not quite sure about is how we want to handle emitting schedule output.
Behavior in this pr is that the output will be emitted at the beginning of each repetition but not at
the end when the schedule decides to stop.
So:
```scala
      Stream(1)
        .repeatEither(Schedule.recurs(4))
        .run(Sink.collectAll[Either[Int, Int]])
        .map(_ must_=== List(Right(1), Left(1), Right(1), Left(2), Right(1), Left(3), Right(1), Left(4), Right(1)))
``` 

cc @iravid

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1471,7 +1471,7 @@ class ZStreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestR
       Stream(1)
         .repeatEither(Schedule.recurs(4))
         .run(Sink.collectAll[Either[Int, Int]])
-        .map(_ must_=== List(Right(1), Right(1), Left(1), Right(1), Left(2), Right(1), Left(3), Right(1), Left(4)))
+        .map(_ must_=== List(Right(1), Left(1), Right(1), Left(2), Right(1), Left(3), Right(1), Left(4), Right(1)))
     )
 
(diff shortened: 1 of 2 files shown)


iravid:
Hey @mschuwalow, the changes look good, but could you adjust them so the schedule's output is  emitted at the end of each repetition, including the last one?

mschuwalow:
@iravid 
Sure. I wasn't quite sure how the semantics should be.
Is this what you had in mind?

iravid:
Oh yeah, sorry about that, I should have wrote that in the ticket :-) Thanks, this looks good!

Issue #413, opened 2018-12-02
Improve Scaladoc


jdegoes:
The Scaladoc for classes and traits inside ZIO can use improvement. In particular:

 - All methods should have Scaladoc.
 - All functions should include Scaladoc descriptions for all parameters.
 - Where useful, methods should have code examples to show users how to use the method.

This is a great issue for new contributors and folks wanting to learn how to use ZIO! Even if you don't know how to use ZIO, don't let that stop you. If you have any questions, post them here and me or another ZIO contributor can help you out.

Sign up for documenting one of the following classes:

 - [ ] **scalaz.zio.Ref**
 - [ ] **scalaz.zio.RefM**
 - [ ] **scalaz.zio.Queue**
 - [ ] **scalaz.zio.ExitResult**
 - [ ] **scalaz.zio.ExitResult.Cause**
 - [ ] **scalaz.zio.Clock**
 - [ ] **scalaz.zio.Async**
 - [ ] **scalaz.zio.IO**
 - [ ] **scalaz.zio.Fiber**
 - [ ] **scalaz.zio.FiberLocal**
 - [ ] **scalaz.zio.FiberFailure**
 - [ ] **scalaz.zio.KleisliIO**
 - [ ] **scalaz.zio.Managed**
 - [ ] **scalaz.zio.Promise**
 - [ ] **scalaz.zio.Schedule**
 - [ ] **scalaz.zio.Semaphore**
 - [ ] **scalaz.zio.ZIO**
 - [ ] **scalaz.zio.system**
 - [ ] **scalaz.zio.console**

If you grab one of these, please write the documentation that you would appreciate as a new user of the library. 😄

hepin1989:
The scaladoc should descript the type parameter and parameters too, not just the behavior of the method.

samuelsayag:
Hello,
I am really a beginner with this library but thought I could do some scaladoc to learn it. Is there somebody already doing it?
That is also my first time contributing to anything so please give first aid/advice/thing not to do if relevant.
 
By the way, if it is of any importance:
I cloned the project and add it for the console just to see the look but it could not be generated. 
And then after searching I saw there is an open issue: https://github.com/scala/bug/issues/8124
console being a package object the documentation will not be generated.

ttymck:
Would we like to have the "superpolymorphic" types added to their respective `Data Types` pages? If so I can add notes on `ZSchedule` to `Schedule`.

saeltz:
Hello,

[...]

PR #1700, opened 2019-09-18
Rename ZIO.aggregateWithin to ZIO.aggregateWithinEither and add a simpler aggregateWithin

### LINKED ISSUE #1567: Rename ZIO.aggregateWithin to ZIO.aggregateWithinEither and add a simpler aggregateWithin
The current version of `aggregateWithin` emits the schedule's output into the stream, resulting in a stream of `Either[ScheduleOutput, Batch]`. This variant should be named `aggrgateWithinEither`, and `ZStream#aggregateWithin` should be defined as:
```scala
def aggregateWithin = aggregateWithinEither.collect {
  case Right(v) => v
}
```

saeltz:
This fixes #1567


iravid:
Needs a rebase on master

saeltz:
Finally, ready for your review, @iravid.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -250,7 +250,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * Aggregated elements will be fed into the schedule to determine the delays between
    * pulls.
    */
-  final def aggregateWithin[R1 <: R, E1 >: E, A1 >: A, B, C](
+  final def aggregateWithinEither[R1 <: R, E1 >: E, A1 >: A, B, C](
```
vasilmkd: We will need to add some more documentation to this method, to highlight the difference compared to the new `aggregateWithin` which currently doesn't have a scaladoc.

saeltz: I appreciate your feedback! I've added ScalaDoc for both methods in https://github.com/zio/zio/pull/1700/commits/a4b8851cafcfc94722a5f88180fec3d56bad1abe.

vasilmkd: @iravid Should we open an issue maybe for the hackathon, to add annotated documentation like this?

saeltz: See https://github.com/zio/zio/issues/413 where I started adding ScalaDoc to some files. It's a good first issue for a hackathon.

PR #1703, opened 2019-09-18
Fixed #1702 - Add hasSameElements assertion

### LINKED ISSUE #1702: Add containsTheSameElements assertion

jaliss:


test/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -305,6 +305,14 @@ object AssertionSpec {
     testFailure(
       assert(Seq(), isNonEmpty),
       message = "isNonEmpty must fail when the traversable is empty"
+    ),
+    testSuccess(
```
NeQuissimus: Could you add a test that actually has the elements in a different order?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -213,6 +213,11 @@ object Assertion {
   final def containsString(element: String): Assertion[String] =
     Assertion.assertion("containsString")(param(element))(_.contains(element))
 
+  final def containsTheSameElements[A](other: Iterable[A]): Assertion[Iterable[A]] =
```
adamgfraser: I think we need to do a little more here because we want to make sure that in addition to having all the same elements we have the same number of each element. For example, right now `List(1, 1, 2, 3)` would be counted as containing the same elements as `List(1, 2, 3, 3)`.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -213,6 +213,11 @@ object Assertion {
...
+    Assertion.assertion("containsTheSameElements")(param(other)) { actual =>
+      actual.size == other.size && actual.toSet.diff(other.toSet).isEmpty
```
adamgfraser: It would be nice to figure out a way to get this assertion to report some more information about which elements are different between the two collections. You could try using `AssertionRec` and having a sub-assertion that the difference is empty. May take a little playing around to be sure we get good error reporting for different combinations of the two collections.

test/shared/src/main/scala/zio/test/Assertion.scala
jaliss: The latest change will make the assertion fail for this case.

test/shared/src/main/scala/zio/test/Assertion.scala
jaliss: Sounds good, let me see how `AssertionRec` that works.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -213,6 +213,23 @@ object Assertion {
   final def containsString(element: String): Assertion[String] =
     Assertion.assertion("containsString")(param(element))(_.contains(element))
 
+  final def containsTheSameElements[A](other: Iterable[A]): Assertion[Iterable[A]] = {
```
adamgfraser: Could we just do `actual.diff(other).isEmpty && other.diff(actual).isEmpty`? That might fit well with how we want to report the cause of the assertion failure too.

test/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -305,6 +305,26 @@ object AssertionSpec {
...
+    ),
+    testFailure(
+      assert(Seq(1, 2, 3, 4), containsTheSameElements(Seq(1, 2, 3))),
+      message = "containsTheSameElements must fail when the iterables do not contain the same elements"
+    ),
+    testSuccess(
+      assert(Seq(4, 3, 1, 2), containsTheSameElements(Seq(1, 2, 3, 4))),
+      message = "containsTheSameElements must succeed when both iterables contain the same elements in different order"
+    ),
+    testFailure(
```
adamgfraser: Nice.

test/shared/src/main/scala/zio/test/Assertion.scala
adamgfraser: What do you think about calling this just `sameElements`? It is shorter and what @ghostdogpr originally suggested. Downside is it is the same as the collection method `sameElements` which checks the elements in order.

jaliss: My understanding is that assertion names should be verbs, that's why I used this. Eg: in a previous PR I had `nonEmpty` and we changed it to `isNonEmpty`.   Maybe rename it to `hasSameElements`?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -216,7 +216,7 @@ object Assertion {
   final def containsTheSameElements[A](other: Iterable[A]): Assertion[Iterable[A]] =
     Assertion.assertion("containsTheSameElements")(param(other)) { actual =>
       val actualSeq = actual.toSeq
-      val otherSeq = other.toSeq
+      val otherSeq  = other.toSeq
       actualSeq.diff(otherSeq).isEmpty && otherSeq.diff(actualSeq).isEmpty
```
adamgfraser: This definitely takes care of the assertion being accurate. What do you think about reporting? Right now if the assertion fails we are going to get something like:

```scala
List(1, 1, 2, 3) did not satisfy sameElements(List(1, 2, 3, 3))
```

In this case it is pretty obvious what the difference is but it could be harder to tell if the list is longer. I think if we made the two checks for the lists being empty into assertions themselves we could get something more like:

```scala
List(1, 1, 2, 3) did not satisfy sameElements(List(1, 2, 3, 3))
(List(1) did not satisfy isEmpty) && (List(3) did not satisfy isEmpty)
```

It isn't perfect because we're not exactly explaining to you what those sub-assertions are but we could put in some documentation about that. What do you think?

test/shared/src/main/scala/zio/test/Assertion.scala
adamgfraser: Yeah, that is definitely true though we have some exceptions like `forall`. I'm fine with `hasSameElements` unless anyone else has other ideas.

ghostdogpr: +1 for `hasSameElements`

adamgfraser: Alright. Well that is settled. 😃

test/shared/src/main/scala/zio/test/Assertion.scala
jaliss: +1 ... let's do as you suggest.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -213,6 +213,14 @@ object Assertion {
   final def containsString(element: String): Assertion[String] =
     Assertion.assertion("containsString")(param(element))(_.contains(element))
 
+  final def hasSameElements[A](other: Iterable[A]): Assertion[Iterable[A]] =
+    Assertion.assertionRec[Iterable[A]]("hasSameElements")(param(other)) { (_, actual) =>
+      val actualSeq = actual.toSeq
+      val otherSeq  = other.toSeq
+
+      isEmpty(actualSeq.diff(otherSeq)) && isEmpty(otherSeq.diff(actualSeq))
```
adamgfraser: Looks great. Can we get a test that the string representation of running the assertion looks like we expect? Then we will be good to go!

adamgfraser:
Looking at this a bit more, I don't think we have quite the right infrastructure in place yet to report the differences between the two collections cleanly. So I would suggest we go back to the previous version where we just did `self.diff(other).isEmpty && other.diff(self).isEmpty`, add a little Scaladoc and alphabetize with the new name, and then we can get this merged.

jaliss:
@adamgfraser I agree. Once we have infrastructure to provide better reporting we can review this again.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -300,6 +300,18 @@ object Assertion {
       assertion(proj(actual))
     }
 
+  /**
+   * Makes a new assertion that requires an Iterable to have the same elements
+   * than the specified Iterable.
```
adamgfraser: Minor typo ("than"). Maybe clarify that elements don't necessarily have to be in the same order? Then ready to go!

Issue #427, opened 2018-12-05
Check for binary compatibility in 1.x


jdegoes:
Once 1.0 is released, we can check for backward-compatible changes using [MiMa](https://github.com/lightbend/migration-manager).

jdegoes:
It was previously decided we will **NOT** use Mima for 2.11, only 2.12+

kitlangton:
I think this has been addressed

PR #1706, opened 2019-09-19
Add Meta Subtype to Cause

### LINKED ISSUE #1684: Add Meta term to Cause enumeration
A new `Meta` tag can be attached to `Cause` to allow attachment of metadata.

There are three obvious options here.

One is supporting only _boolean metadata_. For example:

 - Should this stack trace be printed?
 - Is this a primary failure?
 - Did this failure come from a finalizer? 
 - Etc.

```scala
final case class Meta[E](cause: Cause[E], contains: AnyRef => Boolean) extends Cause[E]
...
case object Stackless
def stackless[E](cause: Cause[E]): Cause[E] = Meta(cause, _ eq Stackless)
```

This is very easy to use but only supports boolean metadata. 

The second option is to allow typed information to be attached to a `Cause`. This is more powerful but much more cumbersome to use:

```scala
final case class Meta[M, E](cause: Cause[E], meta: M) extends Cause[M, E]
...
def meta[M, E](meta: M, cause: Cause[M, E]): Cause[M, E] = Meta(cause, meta)
```

A third option involves fixing the metadata type: to a sum type, created and maintained by ZIO. It offers an intermediate option between the other two options.

/cc @adamgfraser @neko-kai

adamgfraser:
Resolves #1684.

I set it up so that `Meta` is excluded for `equals` and `hashCode`. Key things to look at would be:

1. That I got things set up so we can add additional metadata in the future while retaining binary compatibility.
2. Any other types of metadata we want to add right now?
3. How we are rendering the `Cause` with `stackless`. Right now it is pretty aggressive where we just show the first cause of parallel or sequential errors, don't print the stack trace of a `Throwable`, and don't print tracing information. So every failure is three of four lines saying a fiber failed, whether or not it was unchecked, a string representation of the error if available, and a line that no ZIO tracing is available.

I also noticed that some of the tests in `CauseSpec` weren't being run because there were two assertions on separate lines that weren't joined with a `&&`. 😱 Fortunately there were no failing tests after I fixed it.

Copying @ioleo.


vasilmkd:
Not to sound negative or anything, it's just that I don't have much experience with keeping binary compatibility and I really admire the effort.

I also know that this has nothing to do with the fact that we are close to ZIO 1.0 but more with the fact that we would like to evolve this further, or in a completely different direction and this would allow for a smoother upgrade path. I also know that I'm not going to be involved with the review of this and maybe you shouldn't listen to my opinion, but I'd feel much more sure in the future "binary sustainability" if we took the time to set up some tooling that make this possible. I also think this would be beneficial to the releases post 1.0.

As I said, I don't have much experience with this, but I'm pretty sure there are other Java and Scala libraries that include these types of checks in the toolchain. I'd also be willing to help, but not before two weeks from now, as I have exams. I'd like to know everyone's thoughts on this. Thanks.

adamgfraser:
@vasilmkd I think those are great points. There is actually an open issue for the hackathon to add MIMA support to check for binary compatibility (#427).

core-tests/shared/src/test/scala/zio/CauseSpec.scala
```diff
@@ -56,4 +59,13 @@ class CauseSpec extends Specification with ScalaCheck {
   private def e8 = prop { (c: Cause[String]) =>
     c.untraced.traces.headOption must beNone
   }
+
+  private def e9 = prop { (c: Cause[String]) =>
+    (Cause.stackless(c) must_== c) &&
+    (c must_== Cause.stackless(c))
```
jdegoes: 👍

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -323,13 +353,25 @@ object Cause extends Serializable {
     override final def hashCode: Int = cause.hashCode()
     override final def equals(obj: Any): Boolean = obj match {
       case traced: Traced[_] => cause == traced.cause
+      case meta: Meta[_]     => cause == meta.cause
+      case _                 => cause == obj
+    }
+  }
+
+  // Meta is excluded completely from equals & hashCode
+  final case class Meta[E](cause: Cause[E], data: Data) extends Cause[E] {
```
jdegoes: It may be time to make all these case classes / objects private so we have more flexibility to change `Cause` during the 1.x lifecycle. What do you think?

A basic fold method that hides `Meta` could be used to allow people to traverse.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -378,6 +421,17 @@ object Cause extends Serializable {
     }
   }
 
+  final case class Data private[Cause] (stackless: Boolean)
```
jdegoes: I wonder if this is enough for binary backward compatibility (I don't think so, but I am not sure). You might want to make the `Data` class private too

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -378,6 +421,17 @@ object Cause extends Serializable {
...
+
+  object Data {
+    final def apply(stackless: Boolean): Data =
```
regiskuckaertz: I think this function is implied by using a `case class` isn't it?

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -197,6 +203,25 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
                 Failure("An error was rethrown with a new trace." :: renderTrace(Some(trace))) :: causeToSequential(c).all
               )
           }
+        case Meta(cause, data) =>
+          if (data.stackless) cause match {
```
jdegoes: You may want to pass `Data` in from above, that way you can have a default for all `Meta`, and each level can override it. For example, you might have `stack` nested inside `stackless` .That means you want the stack on the inside part but not on the outside part. This logic requires top-down inheritance of parent settings, with child override.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: We're trying to preserve the ability to augment `Data` with additional fields while preserving binary compatibility. So the thought from @neko-kai, if I understood it correctly, was that by overriding these methods we could then forward them to a new version and keep the old version private for compatibility. But I admit I am still learning about this so I am more just trying to follow the advice of others with more expertise at this point.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: I think that makes sense. So then `fold` would go to needing to take a parameter for each constructor of `Cause`, because we couldn't allow people to pattern match against the `Cause` constructors, right?

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: I tried to do that initially but ran into issues because `Data` is exposed when users pattern match against `Meta`. But maybe to your point above if we disallow that then we can make this completely private. We just need to make sure we give users enough functionality to work with `Cause` like in the `ZStream` snippet I updated..

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: I think we could do that by changing the last case to:

```scala
case m @ Meta(_, _) => causeToSequential(m)
```

The cases for all the other constructors that allow nesting already pass down the parent `Meta`. So then if you had nested `Meta` cases you would get to the `Meta(Meta)` branch and you would then trigger the rule that innermost scope takes precedence.

core/shared/src/main/scala/zio/Cause.scala
iravid: Did anyone say streams? 🤪

I'd be happy to move that snippet into `Cause`'s companion object.

adamgfraser: @iravid Awesome! 😃

jdegoes: @adamgfraser Good point. I'm not exactly sure what to do here, but I do know if we don't hide enough, we're gonna have real trouble changing functionality in 1.x. We'd need MiMa to test to for sure either way...

adamgfraser: Why don't I try make all the constructors private and implementing a new version of `fold`. Then we should be able to make `Data` private. I can also try to set up MiMa so we can test binary compatibility and then we can just turn it on when we go live with 1.0.

jdegoes: That would be fantastic! Thank you! 🙏

adamgfraser:

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: I tried to add MiMa but it looks like it needs to compare a current version against a released version. However, I was able to make the constructors and `Data` private so I think we should be in good shape from a binary compatibility perspective.

core-tests/shared/src/test/scala/zio/CauseSpec.scala
```diff
@@ -1,59 +1,102 @@
...
-      `Cause#untraced` removes all traces $e8
-    Then 
-      `Then.equals` satisfies associativity $e4
-      `Then.equals` satisfies distributivity $e5
-    Both
-      `Both.equals` satisfies associativity $e6
-      `Both.equals` satisfies commutativity $e7
-      """
+object CauseSpec
+    extends ZIOBaseSpec(
```
jdegoes: 💪

core/shared/src/main/scala/zio/Cause.scala
jdegoes: I'm not convinced you can do it without a stack (either real stack or fake stack), as you have "edges" that should inherit parent behavior and "wedges" that have their own behavior, and you have to flip flop between them. However, we can fix this later, I am happy to get this merged in for 1.0.

adamgfraser: Yes, we are definitely simplifying a little bit here in that we are throwing away the second branch of any `Both` or `Then` cases, which might themselves contain metadata indicating that they should be shown. But as you say we can clean this up later,

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -274,62 +327,133 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
     override final def equals(that: Any): Boolean = that match {
       case fail: Fail[_]     => value == fail.value
       case traced: Traced[_] => this == traced.cause
+      case meta: Meta[_]     => this == meta.cause
       case _                 => false
     }
   }
 
-  final case class Die(value: Throwable) extends Cause[Nothing] {
+  object Fail {
```
jdegoes: I'd make all the companion objects private, too.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -368,16 +498,35 @@ object Cause extends Serializable {
...
       case _                            => false
     }
   }
 
+  object Both {
+    def apply[E](left: Cause[E], right: Cause[E]): Cause[E] =
+      new Both(left, right)
+  }
+
+  private case class Data(stackless: Boolean)
```
jdegoes: This should be well-hidden from Scalac. 😆

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: We need these if we want to expose the `apply` methods on them. Normally I wouldn't do this but there is a good amount of code that uses them, and `then` is a reserved keyword so we can't use the normal lower case smart constructor for that. I could delete the companion objects and replace with methods with capitalized names or we could go through and change everything, maybe using `parallel` and `sequential` for the smart constructors, though that is wordier.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -274,62 +327,133 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+        (sequenceCauseOption(left), sequenceCauseOption(right)) match {
+          case (Some(cl), Some(cr)) => Some(Cause.Both(cl, cr))
+          case (None, Some(cr))     => Some(cr)
+          case (Some(cl), None)     => Some(cl)
+          case (None, None)         => None
+        }
+    }
 
-  final case class Fail[E](value: E) extends Cause[E] {
+  private case class Fail[E](value: E) extends Cause[E] {
```
neko-kai: private final?

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -274,62 +327,133 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+  object Traced {
+    def apply[E](cause: Cause[E], trace: ZTrace): Cause[E] =
+      new Traced(cause, trace)
+  }
+
+  // Meta is excluded completely from equals & hashCode
+  private case class Meta[E](cause: Cause[E], data: Data) extends Cause[E] {
+    override final def hashCode: Int = cause.hashCode
+    override final def equals(obj: Any): Boolean = obj match {
+      case traced: Traced[_] => cause == traced
```
neko-kai:
```suggestion
      case traced: Traced[_] => cause == traced.cause
```

neko-kai: Traced could probably be replaced by Meta, tbh...

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -36,15 +36,24 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+      case Meta(cause, _)    => cause.died
       case _                 => false
     }
 
+  /**
+   * Returns the `Throwable` associated with the first `Die` in this `Cause` if
+   * one exists.
+   */
+  final def dieOption: Option[Throwable] =
+    fold(_ => None, t => Some(t), None)(_ orElse _, _ orElse _, (z, _) => z)
```
neko-kai: Might want to _always_ use named parameters with the new fold 😆

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: Good idea. The ordering of type parameters is definitely arbitrary other than that I tried to curry it so that the `Z` parameter could be inferred.

adamgfraser:
@jdegoes I updated this to have `causeToSequential` take `Data` as a parameter. I think you were right about this being a better way to do it. Now the semantics of `stackless` are better defined where `stackless` causes all "leaves" of the `Cause` tree with `Throwable` failures to be rendered without stack traces and is overridable by inner scopes. This also makes it orthagonal to `untraced` where `stackless` deals exclusively with rendering of `Throwable` stack traces, `untraced` deals with removing ZIO traces, and you could combine the two if you want both behaviors.

The one issue that this still leaves open is the constructors. Normally my preference would be to just expose lower case "smart" constructors (e.g. `die`, `fail`, etc...). The complication here is that `when` is a reserved word and usage as an identifier is deprecated. So for now I have exposed companion object with `apply` methods to allow capitalized constructors (`Both`, `Then`). Are you okay with that or do you want to come up with different names? `sequential` and `parallel` seems most obvious but they are longer.

adamgfraser:
@jdegoes @iravid I included the fix for `unsandbox` here as well.

ghostdogpr:
@adamgfraser can you rebase this one? Some conflicts appeared. Then I guess this is good to merge?

adamgfraser:
@ghostdogpr Done. Yes, I think this is good to merge. Though now a couple of the builds on Scala.JS are failing. It looks unrelated, though I haven't seen these errors before:

```
:: problems summary ::
:::: ERRORS
	Server access Error: sun.security.validator.ValidatorException: PKIX path validation failed: java.security.cert.CertPathValidatorException: timestamp check failed url=https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/fusesource/fusesource-pom/1.11/fusesource-pom-1.11.jar

	Server access Error: sun.security.validator.ValidatorException: PKIX path validation failed: java.security.cert.CertPathValidatorException: timestamp check failed url=https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/fusesource/jansi/jansi-project/1.12/jansi-project-1.12.jar
```

And later:

```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

<--- Last few GCs --->

  122003 ms: Mark-sweep 1358.7 (1435.0) -> 1358.4 (1435.0) MB, 2487.3 / 0 ms [allocation failure] [GC in old space requested].
  124501 ms: Mark-sweep 1358.4 (1435.0) -> 1358.4 (1435.0) MB, 2497.6 / 0 ms [allocation failure] [GC in old space requested].
  127013 ms: Mark-sweep 1358.4 (1435.0) -> 1358.3 (1435.0) MB, 2512.2 / 0 ms [last resort gc].
  129557 ms: Mark-sweep 1358.3 (1435.0) -> 1358.2 (1435.0) MB, 2544.0 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0xdd8f6c9fa9 <JS Object>
    3: from__sc_IterableOnce__sci_List [/home/circleci/project/core-tests/js/target/scala-2.13/core-tests-test-fastopt.js:144739] [pc=0x1eae85cf5d63] (this=0x13d65fec9209 <a $c_sci_List$ with map 0xcd93277b211>,coll=0x2d8ac8a87801 <a $c_sjsr_WrappedVarArgs with map 0xcd9327865d9>)
    4: collectAll__sc_Iterable__s_Option [/home/circleci/project/core-tests/js/target/scala-2.13/core-tests-test-fast...

[error] org.scalajs.testcommon.RPCCore$ClosedException: org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed
[error] 	at org.scalajs.testcommon.RPCCore.helpClose(RPCCore.scala:224)
[error] 	at org.scalajs.testcommon.RPCCore.call(RPCCore.scala:165)
[error] 	at org.scalajs.testcommon.RunMuxRPC.call(RunMuxRPC.scala:45)
[error] 	at org.scalajs.testadapter.RunnerAdapter.$anonfun$done$1(RunnerAdapter.scala:58)
[error] 	at scala.collection.immutable.List.map(List.scala:290)
[error] 	at org.scalajs.testadapter.RunnerAdapter.done(RunnerAdapter.scala:58)
[error] 	at sbt.Defaults$.$anonfun$allTestGroupsTask$11(Defaults.scala:1222)
[error] 	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
[error] 	at scala.collection.immutable.Map$Map3.foreach(Map.scala:195)
[error] 	at scala.collection.TraversableLike.map(TraversableLike.scala:238)
[error] 	at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
[error] 	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error] 	at sbt.Defaults$.$anonfun$allTestGroupsTask$5(Defaults.scala:1220)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:280)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:19)
[error] 	at sbt.Execute.work(Execute.scala:289)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:280)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:178)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:37)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[error] Caused by: org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed
[error] 	at org.scalajs.jsenv.nodejs.AbstractNodeJSEnv$NodeComJSRunner.receive(AbstractNodeJSEnv.scala:313)
[error] 	at org.scalajs.jsenv.nodejs.AbstractNodeJSEnv$NodeComJSRunner.receive$(AbstractNodeJSEnv.scala:293)
[error] 	at org.scalajs.jsenv.nodejs.NodeJSEnv$ComNodeRunner.receive(NodeJSEnv.scala:70)
[error] 	at org.scalajs.jsenv.ComJSRunner.receive(ComJSRunner.scala:27)
[error] 	at org.scalajs.jsenv.ComJSRunner.receive$(ComJSRunner.scala:27)
[error] 	at org.scalajs.jsenv.nodejs.NodeJSEnv$ComNodeRunner.receive(NodeJSEnv.scala:70)
[error] 	at org.scalajs.testadapter.ComJSEnvRPC$$anon$1.run(ComJSEnvRPC.scala:34)
[error] Caused by: java.io.EOFException
[error] 	at java.io.DataInputStream.readInt(DataInputStream.java:392)
[error] 	at org.scalajs.jsenv.nodejs.AbstractNodeJSEnv$NodeComJSRunner.receive(AbstractNodeJSEnv.scala:303)
[error] 	at org.scalajs.jsenv.nodejs.AbstractNodeJSEnv$NodeComJSRunner.receive$(AbstractNodeJSEnv.scala:293)
[error] 	at org.scalajs.jsenv.nodejs.NodeJSEnv$ComNodeRunner.receive(NodeJSEnv.scala:70)
[error] 	at org.scalajs.jsenv.ComJSRunner.receive(ComJSRunner.scala:27)
[error] 	at org.scalajs.jsenv.ComJSRunner.receive$(ComJSRunner.scala:27)
[error] 	at org.scalajs.jsenv.nodejs.NodeJSEnv$ComNodeRunner.receive(NodeJSEnv.scala:70)
[error] 	at org.scalajs.testadapter.ComJSEnvRPC$$anon$1.run(ComJSEnvRPC.scala:34)
[error] (coreTestsJS / Test / executeTests) org.scalajs.testcommon.RPCCore$ClosedException: org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed
[error] Total time: 487 s (08:07), completed Oct 1, 2019 10:16:15 AM
Exited with code 1
```

ghostdogpr:
Oh you fixed the problem with master 👍

adamgfraser:
Yes, we had a conflict when I removed the ZIO internal test suite dependency last night and then the test migrations we merged were relying on the implementation of `nonFlaky` in that instead of the one in `TestAspect#nonFlaky`.

adamgfraser:
Hmmm, looks like we're getting the same error.

adamgfraser:
I think I see the problem. The previous version was not actually doing 100 runs on Scala.js, only on JVM. So when I switched it over we ended up doing way more runs of those tests on Scala.js, which I guess caused us to run out of heap? Anyway, let me adjust the settings so we get the same behavior as the original and push another change.

adamgfraser:
@ghostdogpr Hmmm. Still not working, and always those same two builds.. I think there are a few things we can pursue:

1. I can try to do a PR that just fixes the issue on master and does nothing else so we try to isolate the issue.
2. We can try to look at when those messages started popping up. The first message about the server access error appears even on the JS build that does complete successfully.
3. We can look into if there is any CI issue that is causing this or CI setting that we could change regarding heap or garbage collection.

ghostdogpr:
Maybe try your first suggestion first? There wasn't any issue in #1862 but this one looks systematic so that's really weird.

adamgfraser:
Note that this reverts the `ZStream#flatMap` implementation. If @iravid is okay with it this is ready to merge.

iravid:
Yes, totally fine! Thanks for spotting it @adamgfraser

PR #1713, opened 2019-09-19
Add ZStream#unNone

asakaev:
Usually using this technique, you have to write boilerplate code.
Inspired by @mpilquist implementation.


asakaev:
I don't get the purpose of using extension methods if we have access to sources. What is the reason?

There is only one method implemented as an extension. What makes it distinct from the others? Maybe it should be implemented as a method of a class?

iravid:
Thanks @asakaev! Keeping the methods on the class as you've done is preferable (better discoverability). Let's move `unTake` in as well.

vasilmkd:
Moving `unTake` works as well! Thanks.

iravid:
Nice work @asakaev! Thanks for reviewing  @vasilmkd!

vasilmkd:
@iravid my pleasure. I'm willing to donate review cycles, as you call them, generously for the duration of the hackathon.

PR #1716, opened 2019-09-20
Fix flaky test

### LINKED ISSUE #1707: Flaky FiberRef Test
It appears to be related to one of the tests we added in #1594. Copying @mlangc.

```
[info] FiberRefSpec
[info]    Create a new FiberRef with a specified value and check if:
[info]       + `get` returns the current value.
[info]       + `get` returns the current value for a child.
[info]       + `set` updates the current value.
[info]       + `set` by a child doesn't update parent's value.
[info]       + `locally` restores original value.
[info]       + `locally` restores parent's value.
[info]       + `locally` restores undefined value.
[info]       + its value is inherited on join.
[info]       + initial value is always available.
[info]       + `update` changes value.
[info]       + `updateSome` changes value.
[info]       + `updateSome` not changes value.
[info]       + `modify` changes value.
[info]       + `modifySome` not changes value.
[info]       + its value is inherited after simple race.
[info]       + its value is inherited after a race with a bad winner.
[info]       + its value is not inherited after a race of losers.
[error]       x the value of the looser is inherited in zipPar
[error]  'update1' != 'update2' (FiberRefSpec.scala:175)
[info]       + nothing gets inherited with a failure in zipPar
[info] Total for specification FiberRefSpec
[info] Finished in 249 ms
[info] 19 examples, 20 expectations, 1 failure, 0 error
```

https://circleci.com/gh/zio/zio/24880?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mlangc:
Greatly decrease the probability that a fiber that is meant to loose a race actually wins. This should fix #1707.

Note: I've run the affected tests 10_000 times in parallel and did not get any failures. A very unfair our outright vicious CPU scheduler could still make them fail though.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/FiberRefSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberRefSpec.scala
@@ -27,6 +27,7 @@ class FiberRefSpec extends BaseCrossPlatformSpec {
     """
 
   val (initial, update, update1, update2) = ("initial", "update", "update1", "update2")
+  val looseTimeAndCpu                     = ZIO.yieldNow.repeat(Schedule.spaced(Duration.fromNanos(1)) && Schedule.recurs(100))
 
   def e1 =
     for {
@@ -150,7 +151,7 @@ class FiberRefSpec extends BaseCrossPlatformSpec {
     for {
       fiberRef   <- FiberRef.make(initial)
       badWinner  = fiberRef.set(update1) *> ZIO.fail("ups")
-      goodLooser = fiberRef.set(update2) *> clock.sleep(Duration.fromNanos(100))
+      goodLooser = fiberRef.set(update2) *> looseTimeAndCpu
       _          <- badWinner.race(goodLooser)
       value      <- fiberRef.get
     } yield value must_=== update2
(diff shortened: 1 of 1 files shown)


adamgfraser:
@mlangc Is this just an issue with the test being flaky or is there some flakiness in `FiberRef`?

mlangc:
No, it's the test:
```scala
def e18 =
  for {
    latch    <- Promise.make[Nothing, Unit]
    winner   = fiberRef.set(update1) *> latch.succeed(()).unit
    looser   = latch.await *> fiberRef.set(update2) *> looseTimeAndCpu
    _        <- winner.zipPar(looser)
    value    <- fiberRef.get
  } yield value must_=== update2
```
As you can see, even with the latch, a very unfair scheduler might decide to suspend the thread running the winner directly after succeeding the promise until the looser is done. With the current implementation of `looseTimeAndCpu` this should be extremely unlikely though. If the flakiness remains nonetheless (I've run the test thousands of times in parallel without a single failure), we could lock both fibers to the same thread.

adamgfraser:
@mlangc I see. To do in a completely deterministically we would need to be able to attach a finalizer to the `winner` fiber that would get invoked when it was joined, but I don't think that exists today.

PR #1717, opened 2019-09-20
Add ZStream.paginate

### LINKED ISSUE #1704: Add ZStream#paginate
At the workshop on ZStream in Scala World, I wanted to demonstrate how to use ZStream to scroll through an S3 bucket listing.

The way S3's API works is like so:
- You query the `ListBucket` method and get back (simplified) `(List[Key], Option[ContToken])`.
- If the continuation token is defined, you need to query `ListBucket` again, and this time pass in the continuation token.

You'd think that this would easily work with `ZStream#unfoldM`, but that doesn't have the right shape: `unfoldM` is `unfold(s: S)(S => Option[(A, S)])`, but this says that the last fetch we'll make needs to return `None`; that's not how the API works. So you have to do an awkward dance and of course I couldn't figure it out during the workshop :-P

So extrapolating on the above signature from S3, we need something like `def paginate[R1 <: R, E1 >: E, A, S](s: S)(f: S => ZIO[R1, E1, (A, Option[S]))]` on `ZStream`. Then we could do:
```scala
def fetch: Task[(List[Key], Option[ContToken])]
def fetchNext(token: ContToken): Task[(List[Key], Option[ContToken])]

Stream.fromEffect(fetch).flatMap {
  case (keys, None) => Stream.fromIterable(keys)
  case (keys, Some(token)) => 
    Stream.fromIterable(keys) ++
    Stream.paginate(token)(fetchNext).mapConcat(Chunk.fromIterable)
}
```

regiskuckaertz:
Closes #1704

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -181,6 +181,8 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
     mergeWith short circuit       $mergeWithShortCircuit
     mergeWith prioritizes failure $mergeWithPrioritizesFailure
 
+  Stream.paginate            $paginate
+
   Stream.partitionEither
     values        $partitionEitherValues
     errors        $partitionEitherErrors
@@ -253,8 +255,6 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
     zipWith ignore RHS          $zipWithIgnoreRhs
     zipWith prioritizes failure $zipWithPrioritizesFailure
     zipWithLatest               $zipWithLatest
-
-  Stream.paginate            $paginate
   """
 
   def aggregate = unsafeRun {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2516,6 +2516,16 @@ object ZStream extends ZStreamPlatformSpecific {
     streams: ZStream[R, E, A]*
   ): ZStream[R, E, A] = mergeAll(Int.MaxValue, outputBuffer)(streams: _*)
 
+  final def paginate[R, E, A, S](s: S)(f: S => ZIO[R, E, (A, Option[S])]): ZStream[R, E, A] =
```
iravid: Scaladoc please? :-)

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -253,6 +253,8 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
     zipWith ignore RHS          $zipWithIgnoreRhs
     zipWith prioritizes failure $zipWithPrioritizesFailure
     zipWithLatest               $zipWithLatest
+
+  Stream.paginate            $paginate
```
iravid: Can you bump this up to keep the list sorted?

PR #1718, opened 2019-09-20
ZIO Test: Simplify TestAspect#around

adamgfraser:
Simplifies the type signature of `TestAspect#around` so it becomes a combined version of `TestAspect#before` and `TestAspect#after` where the `before` and `after` effects correspond to the `acquire` and `release` effects of the specified `Managed`. The existing version that also allows for modifying the test result goes to `TestAspect#aroundTest`.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -103,12 +103,15 @@ object TestAspect extends TimeoutVariants {
   /**
    * Constructs an aspect that evaluates every test inside the context of a `Managed`.
    */
-  def around[R0, E0](managed: ZManaged[R0, E0, Any]) =
-    new TestAspect.PerTest[Any, R0, E0, Any, Nothing, Any] {
-      def perTest[R >: Any <: R0, E >: E0 <: Any, S >: Nothing <: Any](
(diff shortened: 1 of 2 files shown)


adamgfraser:
Yes, @jdegoes came up with it. It is really nice.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -100,10 +100,21 @@ object TestAspect extends TimeoutVariants {
         test <* effect.catchAllCause(cause => ZIO.fail(TestFailure.Runtime(cause)))
     }
 
+  /**
+   * Constructs an aspect that evaluates every test inside the context of a `Managed`.
+   */
+  def around[R0, E0](managed: ZManaged[R0, E0, Any]) =
```
jdegoes: Since the `ZManaged` returns `Any`, what about taking `before` and `after` directly, then we can wrap them into a managed inside this function and pass them to the other `aroundTest`? Less work for user.

adamgfraser: I like it. Let me make that change now.

adamgfraser:
Done.

adamgfraser:
@jdegoes Are you good with this?

jdegoes:
Improvement to the default behavior, and breaking, but worth it, I think!

PR #1719, opened 2019-09-20
STMSpec and StmIssue1587Spec migrated to zio-test

LGLO:
I moved issue-1587 to same file as other STM specs


core-tests/shared/src/test/scala/zio/stm/STMSpec.scala
```diff
@@ -1,147 +1,423 @@
...
+            ) {
+              for {
+                tvar <- TRef.makeCommit(42)
+                join <- tvar.get.filter(_ == 42).commit
+                _    <- tvar.set(9).commit
+                v    <- tvar.get.commit
+              } yield assert(v, equalTo(9)) && assert(join, equalTo(42))
+            },
+            testM("resume after satisfying the condition") {
+              val barrier = new Barrier
```
LGLO: Barrier definition in util object. Could please tell me what do you think, orginally there was `unsafeRunSync` inside `STM.succeed`

core-tests/shared/src/test/scala/zio/stm/STMSpec.scala
```diff
@@ -1,147 +1,423 @@
...
+
+            val race = for {
+              r0       <- TRef.makeCommit(0)
+              r1       <- TRef.makeCommit(0)
+              sumFiber <- r0.get.flatMap(v0 => r1.get.map(_ + v0)).commit.fork
+              _        <- r0.update(_ + 1).flatMap(_ => r1.update(_ + 1)).commit
+              sum      <- sumFiber.join
+            } yield assert(sum, equalTo(0) || equalTo(2))
+
+            repeat(race, 10000)
```
LGLO: I have another report that this spec timed out again.
I'll remove this repeat thing and will add barriers inside STM to make sure that:
* updating fiber will update r0 before sum fiber reads it
* sum fiber reads r1 before updating fiber updates it
Without recent fix this would result in sum=1, because of fix, STM will retry sum and result will have to be 2.

adamgfraser: Yes. I have been seeing timeouts more frequently. I know we don't have this merged yet so that doesn't speak to whether your earlier fix would have been sufficient but it is definitely good to make sure this isn't taking too long.

LGLO: I was too optimistic with this approach. I can make sure that read-only fiber won't start STM before update fiber updates first value but I can't await inside STM, so it's hard to make sure that update thread will suspend until read-only thread is done.

PR #1720, opened 2019-09-20
Remove Explicit Delay in ZSchedule

adamgfraser:
I took a slightly different approach to this one. `ZSchedule` continues to be a description of a sequence of delays rather than delaying itself, but now each `Decision` is also responsible for providing an effect that the client can "run". Schedules that use time provide an effect that accesses the `Clock` and sleeps for the specified duration whereas schedules that don't use time provide an effect that does nothing. These effects compose so that combining effects that don't use the `Clock` yields another effect that doesn't use the `Clock`, whereas any other combination yields an effect that does use the `Clock`, similar to how environment types compose. This allows us to have a relatively minimal change from a client perspective while removing the hard coded dependency on `Clock` in methods like `repeat` and `retry`.

```scala
sealed trait Delay[-R] extends (Duration => ZIO[R, Nothing, Unit]) {
  def combine[R1 <: R](that: Delay[R1]): Delay[R1]
}
final case object Live extends Delay[Clock] {
  def apply(d: Duration): ZIO[Clock, Nothing, Unit] =
    clock.sleep(d)
  def combine[R1 <: Clock](that: Delay[R1]): Delay[R1] =
    this
}
final case object Mock extends Delay[Any] {
  def apply(d: Duration): ZIO[Any, Nothing, Unit] =
    ZIO.unit
  def combine[R1 <: Any](that: Delay[R1]): Delay[R1] =
    that
}
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
@@ -119,7 +119,7 @@ class ZScheduleSpec extends BaseCrossPlatformSpec {
       (for {
         ref <- Ref.make(0)
         _   <- incr(ref).repeat(Schedule.recurs(42))
-      } yield ()).foldM(
+      } yield ()).foldM[Any, Nothing, String](
         err => IO.succeed(err),
         _ => IO.succeed("it should not be a success at all")
       )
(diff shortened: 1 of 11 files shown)


mschuwalow:
@adamgfraser looks good. I don't think this is quite powerful enough to have sleeps whose duration is determined at runtime though. I think those will be necessary for example for the cron stuff

adamgfraser:
@mschuwalow Couldn't you pattern match against the prior `Decision` and then make a new `Decision` based on that? Or did you mean something else? Maybe I'm not understanding. Is there a simple example we could use?

mschuwalow:
On second thought I agree that this is probably the right way to go about this

jdegoes:
@adamgfraser I am helping @mschuwalow with a simpler encoding that removes a lot of cruft from Schedule, will show you a preview shortly and would love your feedback!

adamgfraser:
@jdegoes Awesome!

adamgfraser:
Closing in favor of #1798.

PR #1721, opened 2019-09-20
Migrate ZIOSpec to ZIO Test (#1647)

chlin501:
Try migrating ZIOSpec to ZIO Test. But I am not sure if this is the right way to do that, so attempting to start from a small patch first. Please let me know if any issues.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -11,14 +11,14 @@ object ZIOSpec
             f1     <- IO.fail("error1").fork
             f2     <- IO.succeed("success1").fork
             errors <- f1.zip(f2).join.parallelErrors[String].flip
-          } yield assert(errors, equalTo("error1" :: Nil))
+          } yield assert(errors, equalTo(List("error1")))
         },
         testM("allFailures") {
           for {
             f1     <- IO.fail("error1").fork
             f2     <- IO.fail("error2").fork
             errors <- f1.zip(f2).join.parallelErrors[String].flip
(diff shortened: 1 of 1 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -1,24 +1,24 @@
...
-  def oneFailure =
-    for {
-      f1     <- IO.fail("error1").fork
-      f2     <- IO.succeed("success1").fork
-      errors <- f1.zip(f2).join.parallelErrors[String].flip
-    } yield errors must_=== ::("error1", List())
-}
+object ZIOSpec
+    extends ZIOBaseSpec(
+      suite("ZIOSpec")(
```
LGLO: I think that whole suite is about "ZIO#parallelErrors", so I would rename it.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -1,24 +1,24 @@
...
+        testM("oneFailure") {
+          for {
+            f1     <- IO.fail("error1").fork
+            f2     <- IO.succeed("success1").fork
+            errors <- f1.zip(f2).join.parallelErrors[String].flip
+          } yield assert(errors, equalTo(::("error1", List())))
```
LGLO: Nitpick: first time I see prefix `::` together with `List()`, could you, please, use something more standard: `"error1" :: Nil` or `List("error1")`?
I'm aware it was like it before.

chlin501:
Please let me know if some more places that I can improve. I am happy to do that.

ghostdogpr:
@chlin501 thanks!

PR #1722, opened 2019-09-21
Implement ZManaged#withEarlyRelease

### LINKED ISSUE #1697: Add ZManaged#withEarlyRelease
This should modify a `ZManaged[R, E, A]` to `ZManaged[R, E, (UIO[_], A)]` and provide a canceler that can be used to eagerly execute the finalizer.

The canceller should obviously run uninterruptibly, and if completed, should set a `Ref` that will cause the regular installed finalizer to not run.

adamgfraser:
Resolves #1697.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -667,15 +667,22 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) 
   /**
    * Modifies this `ZManaged` to provide a canceler that can be used to eagerly
    * execute the finalizer of this `ZManaged`. The canceler will run
-   * uninterruptibly, and if completed will cause the regular finalizer to not
-   * run.
+   * uninterruptibly with an exit value indicating that the effect was
+   * interrupted, and if completed will cause the regular finalizer to not run.
    */
   final def withEarlyRelease: ZManaged[R, E, (URIO[R, _], A)] =
+    withEarlyReleaseExit(Exit.interrupt)
+
+  /**
+   * A more powerful version of `withEarlyRelease` that allows specifying an 
+   * exit value in the event of early release.
+   */
+  final def withEarlyReleaseExit(exit: Exit[_, _]): ZManaged[R, E, (URIO[R, _], A)] =
     ZManaged[R, E, (URIO[R, _], A)] {
       reserve.flatMap {
         case Reservation(acquire, release) =>
           Ref.make(true).map { finalize =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -664,6 +664,24 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
+   * execute the finalizer of this `ZManaged`. The canceler will run
+   * uninterruptibly, and if completed will cause the regular finalizer to not
+   * run.
+   */
+  final def withEarlyRelease: ZManaged[R, E, (URIO[R, _], A)] =
+    ZManaged[R, E, (URIO[R, _], A)] {
+      reserve.flatMap {
+        case Reservation(acquire, release) =>
+          Ref.make(true).map { finalize =>
+            val canceler  = (release(Exit.interrupt) *> finalize.set(false)).uninterruptible
```
iravid: Oh right I forgot about release using  an `Exit[_, _]`. How about we do 2 variants - `withEarlyReleaseExit(Exit[_, _] => URIO)` and `withEarlyRelease`?

I was going back and forth on defaulting to `Exit.interrupt` but I'm convinced this is good. Let's just add a note about that on the scaladoc.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -667,15 +667,22 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
+   * interrupted, and if completed will cause the regular finalizer to not run.
    */
   final def withEarlyRelease: ZManaged[R, E, (URIO[R, _], A)] =
+    withEarlyReleaseExit(Exit.interrupt)
+
+  /**
+   * A more powerful version of `withEarlyRelease` that allows specifying an 
+   * exit value in the event of early release.
+   */
+  final def withEarlyReleaseExit(exit: Exit[_, _]): ZManaged[R, E, (URIO[R, _], A)] =
```
regiskuckaertz: How about renaming both to `earlyRelease` and `earlyReleaseWith`, this seems to be more in line with other combinators that have a more powerful version.

adamgfraser: I think maybe `withEarlyRelease` indicates more clearly that it is giving you an option to do early release but not actually doing it? And we do have other `Exit` variants like `finalizer` and `finalizerExit`. But that is super subjective. I'm fine with either one. @iravid What do you think?

iravid: `withEarlyRelease` is probably more accurate as it "adds" the option to release early

PR #1723, opened 2019-09-21
use accessM in Arrow combinators

neko-kai:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -731,23 +731,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
-      a1     <- either.fold(self.provide, that.provide)
-    } yield a1
+    ZIO.accessM(_.fold(self.provide, that.provide))
 
   final def +++[R1, B, E1 >: E](that: ZIO[R1, E1, B]): ZIO[Either[R, R1], E1, Either[A, B]] =
-    for {
-      e <- ZIO.environment[Either[R, R1]]
-      r <- e.fold(self.map(Left(_)) provide _, that.map(Right(_)) provide _)
-    } yield r
+    ZIO.accessM[Either[R, R1]] {
```
regiskuckaertz: Not using `fold` here?

PR #1725, opened 2019-09-21
Add ZStream.bimap and either

regiskuckaertz:
Gonna add new combinators as I see a need for them writing cats instances.

cc @iravid

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2256,6 +2256,12 @@ object ZStream extends ZStreamPlatformSpecific {
     ZStream(ZManaged.succeed(UIO.never))
 
   /**
+   * The stream of units
+   */
+  final val unit: Stream[Nothing, Unit] =
+    ZStream(()).forever
+
+  /**
    * Creates a pure stream from a variable list of values
    */
   final def apply[A](as: A*): Stream[Nothing, A] = fromIterable(as)
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -545,6 +545,11 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
   }
 
   /**
+   * Maps the success values of this stream to the specified constant value.
+   */
+  final def as[B](b: B): ZStream[R, E, B] = map(_ => b)
+
+  /**
    * Returns a stream whose failure and success channels have been mapped by
    * the specified pair of functions, `f` and `g`.
    */
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -847,8 +847,10 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
 
   /**
    * Returns a stream whose failures and successes have been lifted into an
-   * `Either`.The resulting stream cannot fail, because the failures have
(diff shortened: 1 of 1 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -839,6 +845,14 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
       } yield pull
     }
 
+  /**
+   * Returns a stream whose failures and successes have been lifted into an
+   * `Either`.The resulting stream cannot fail, because the failures have
+   * been exposed as part of the `Either` success case.
```
iravid: Cool. Add a note that the resulting stream will emit the failure that occured, then end?

regiskuckaertz: Ha, right. Is that the correct behaviour? or instead we keep pulling even after error?

iravid: Yes, that's what we do right now because the contract of `Pull` (currently!) doesn't allow for pulling after a failure.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -812,7 +823,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * Drops all elements of the stream until the specified predicate evaluates
    * to `true`.
    */
-  def dropUntil(pred: A => Boolean): ZStream[R, E, A] =
+  final def dropUntil(pred: A => Boolean): ZStream[R, E, A] =
```
iravid: This can be overriden in StreamEffect, so let's it non-final

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1446,13 +1467,13 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * and the second one will contain all element evaluated to false.
    * The faster stream may advance by up to buffer elements further than the slower one.
    */
-  def partition(p: A => Boolean, buffer: Int = 16): ZManaged[R, E, (ZStream[R, E, A], ZStream[Any, E, A])] =
+  final def partition(p: A => Boolean, buffer: Int = 16): ZManaged[R, E, (ZStream[R, E, A], ZStream[Any, E, A])] =
```
iravid: This too can be overriden :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1787,7 +1808,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * Takes all elements of the stream until the specified predicate evaluates
    * to `true`.
    */
-  def takeUntil(pred: A => Boolean): ZStream[R, E, A] =
+  final def takeUntil(pred: A => Boolean): ZStream[R, E, A] =
```
iravid: This too can be overriden :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2252,6 +2279,15 @@ object ZStream extends ZStreamPlatformSpecific {
   final def bracket[R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R, Nothing, _]): ZStream[R, E, A] =
     managed(ZManaged.make(acquire)(release))
 
+  /**
+   * Creates a stream from a single value that will get cleaned up after the
+   * stream is consumed
+   */
+  final def bracketExit[R, E, A](
```
iravid: nice

regiskuckaertz:
@iravid done

PR #1727, opened 2019-09-21
Migrate tests for MutableConcurrentQueueSpec and OneShotSpec

kapke:
There's a question though - should shortcut `throwsA[E]` be added to `Assertions`?


ghostdogpr:
@kapke yeah, I think you can definitely add it to `Assertion`

Issue #1729, opened 2019-09-21
Add ZStream.fromIterator


iravid:

sideeffffect:
I'll do it :)

kamilkloch:
Will this work? ;)
```scala
def fromIterator[R, A](it: Iterator[A]): ZStream[R, Nothing, A] = {
      ZStream.unfold(it)(it => if (it.hasNext) Some((it.next(), it)) else None)
    }
```

sideeffffect:
has to be
```scala
final def fromIterator[A](iterator: ZIO[Any, Nothing, Iterator[A]]): StreamEffect[Nothing, A]
```
or maybe even 
```scala
final def fromIterator[A](iterator: ZManaged[Any, Nothing, Iterator[A]]): StreamEffect[Nothing, A]
```

the reason is that `Iterator` is impure, mutable by nature and thus the stream will not be re-runnable

vasilmkd:
How is this different from `fromIterable`, geniunely curious?

sideeffffect:
@vasilmkd `Iterable.iterator` always creates a fresh iterator (pointing to the first element). `Iterator` itself is a non-functional structure, which mutates on `.next()`

PR #1731, opened 2019-09-21
ZStream.fromIterator

sideeffffect:
fixes https://github.com/zio/zio/issues/1729

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/Stream.scala
+++ b/streams/shared/src/main/scala/zio/stream/Stream.scala
@@ -176,10 +176,10 @@ object Stream extends ZStreamPlatformSpecific {
     ZStream.fromIterator(iterator)
 
   /**
-   * See [[ZStream.fromIterator]]
+   * See [[ZStream.fromIteratorManaged]]
    */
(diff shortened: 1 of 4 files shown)


streams/shared/src/main/scala/zio/stream/Stream.scala
```diff
@@ -169,6 +169,18 @@ object Stream extends ZStreamPlatformSpecific {
...
     ZStream.fromIterable(as)
 
+  /**
+   * See [[ZStream.fromIterator]]
+   */
+  final def fromIterator[A](iterator: ZIO[Any, Nothing, Iterator[A]]): Stream[Nothing, A] =
+    ZStream.fromIterator(iterator)
+
+  /**
+   * See [[ZStream.fromIterator]]
```
adamgfraser: Duplicate method?

sideeffffect: one takes `Managed`, other takes `ZIO`

iravid:
Just needs a bit more polymorphism :-)

PR #1737, opened 2019-09-21
For Issue zio#1647

wojtek-tg:
For Issue zio#1647

### DIFF
--- a/core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
@@ -12,311 +12,327 @@ import scala.concurrent.duration.{ Duration => ScalaDuration, FiniteDuration => 
 object DurationSpec
     extends ZIOBaseSpec(
       suite("DurationSpec")(
-        test("Make a Duration from positive nanos and check that: The Duration is Finite") {
-          assert(Duration.fromNanos(1), isSubtype[Duration.Finite](Assertion.anything))
-        },
-        test("Make a Duration from positive nanos and check that: Copy with a negative nanos returns Zero") {
-          assert(Duration.fromNanos(1).asInstanceOf[Duration.Finite].copy(-1), isSubtype(equalTo(Duration.Zero)))
-        },
-        test("Make a Duration from positive nanos and check that: Multiplying with a negative factor returns Zero") {
-          assert(Duration.fromNanos(1) * -1.0, equalTo(Duration.Zero: Duration))
-        },
-        test("Make a Duration from positive nanos and check that: Its stdlib representation is correct") {
-          val duration: ScalaDuration = Duration.fromNanos(1234L).asScala
-          val expected: ScalaDuration = ScalaFiniteDuration(1234L, TimeUnit.NANOSECONDS)
-          assert(duration, equalTo(expected))
-        },
-        test("Make a Duration from positive nanos and check that: Its JDK representation is correct") {
-          assert(Duration.fromNanos(2345L).asJava, equalTo(JavaDuration.ofNanos(2345L)))
-        },
-        test("Make a Duration from positive nanos and check that: It identifies as 'zero'") {
-          assert(Duration.fromNanos(0L).isZero, equalTo(true))
-        },
-        test("Make a Duration from positive nanos and check that: Creating it with a j.u.c.TimeUnit is identical") {
-          assert(Duration(12L, TimeUnit.NANOSECONDS), equalTo(Duration.fromNanos(12L)))
-        },
-        test("Make a Duration from positive nanos and check that: It knows its length in ns") {
-          assert(Duration.fromNanos(123L).toNanos, equalTo(123L))
-        },
-        test("Make a Duration from positive nanos and check that: It knows its length in ms") {
-          assert(Duration.fromNanos(123000000L).toMillis, equalTo(123L))
-        },
-        test("Make a Duration from positive nanos and check that: max(1 ns, 2 ns) is 2 ns") {
-          assert(Duration.fromNanos(1L).max(Duration.fromNanos(2L)), equalTo(Duration.fromNanos(2L)))
-        },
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -3,416 +3,320 @@ package zio.duration
...
-
-}
+object DurationSpec
+    extends ZIOBaseSpec(
+      suite("DurationSpec")(
+        test("Make a Duration from positive nanos and check that: The Duration is Finite") {
+          assert(Duration.fromNanos(1), isSubtype[Duration.Finite](Assertion.anything))
+        },
+        test("Make a Duration from positive nanos and check that: Copy with a negative nanos returns Zero") {
+          assert(Duration.fromNanos(1).asInstanceOf[Duration.Finite].copy(-1), isSubtype(equalTo(Duration.Zero)))
```
adamgfraser: I don't think you need the `isSubtype` here since you are already checking for equality.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -3,416 +3,320 @@ package zio.duration
...
-    1.day must_=== Duration.fromNanos(86400000000000L)
-
-  def nano20 =
-    2.days must_=== Duration.fromNanos(172800000000000L)
-
-}
+object DurationSpec
+    extends ZIOBaseSpec(
+      suite("DurationSpec")(
+        test("Make a Duration from positive nanos and check that: The Duration is Finite") {
```
adamgfraser: You could make a suite here with a label of `"Make a Duration from positive nanos and check that:" to provide some organization and cut down on the text.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -3,416 +3,320 @@ package zio.duration
...
+        test("Make a Duration from positive nanos and check that: * with overflow to negative results in Infinity") {
+          assert(Duration.fromNanos(Long.MaxValue) * 2, equalTo(Duration.Infinity: Duration))
+        },
+        test("Make a Duration from positive nanos and check that: Folding picks up the correct value") {
+          assert(Duration.fromNanos(Long.MaxValue).fold("Infinity", _ => "Finite"), equalTo("Finite"))
+        },
+        test("Make a Duration from negative nanos and check that the Duration is Zero ") {
+          assert(Duration.fromNanos(-1), equalTo(Duration.Zero: Duration))
+        },
+        test("Take Infinity and check that: It returns -1 milliseconds") {
```
adamgfraser: Same thing for creating a suite here. I would just match the existing hierarchical structure.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
wojtek-tg: Thanks, good catch, I tried to do it without asInstanceOf and this is an old artefact from that.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
wojtek-tg: Thanks, didn't know I can next suites. Will change all of them to have the original structure.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -3,416 +3,320 @@ package zio.duration
...
+          assert(Duration.Infinity.toMillis, equalTo(Long.MaxValue / 1000000))
```
adamgfraser: I realize it was this was in the original but the statement is inconsistent with the code being tested.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -3,416 +3,320 @@ package zio.duration
...
+        },
+        test("Take Infinity and check that: It returns -1 nanoseconds") {
```
adamgfraser: This is also inconsistent.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -465,6 +465,11 @@ object Assertion {
 
   /**
    * Makes an assertion that requires a value have the specified type.
+   *
+   * Example:
+   * {{{
+   *   assert(Duration.fromNanos(1), isSubtype[Duration.Finite](Assertion.anything))
```
adamgfraser: Great additional documentation! I wonder if we should add a `isSubtypeOf` assertion that doesn't take another parameter to make this use case more ergonomic.

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -12,311 +12,327 @@ import scala.concurrent.duration.{ Duration => ScalaDuration, FiniteDuration =>
...
+            assert(Duration.fromNanos(Long.MaxValue) * 3, equalTo(Duration.Infinity: Duration))
+          },
+          test("* with overflow to negative results in Infinity") {
+            assert(Duration.fromNanos(Long.MaxValue) * 2, equalTo(Duration.Infinity: Duration))
+          },
+          test("Folding picks up the correct value") {
+            assert(Duration.fromNanos(Long.MaxValue).fold("Infinity", _ => "Finite"), equalTo("Finite"))
+          }
+        ),
+        suite("Make a Duration from negative nanos and check that:")(test("The Duration is Zero ") {
```
adamgfraser: Move the first test onto its own line.

test/shared/src/main/scala/zio/test/Assertion.scala
wojtek-tg: I was really confused about that when I first started coding it - to be honest this was the only odd thing in the whole PR after I've read the ZIO test spec. So as a new user of this I'd recommend having an option to call isSubtype[A]()

wojtek-tg:
> Looks great.

Thanks!

PR #1740, opened 2019-09-21
StreamEffect polymorphic in enviRonment

sideeffffect:


iravid:
Nice work @sideeffffect!

PR #1741, opened 2019-09-21
Fixed #1691 - ZIO Test: Make `failure` TestAspect

### LINKED ISSUE #1691: ZIO Test: Make `failure` TestAspect
Inspired by @adamgfraser's recent PR to Cats interop, I think two `failure` aspects would be nice:

 - `failure(p: Assertion[Cause[E]])` — Expects a failure that matches the predicate
 - `failure` — Expects a failure (shorthand for `failure(Assertion.anything)`

jaliss:


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -168,6 +169,41 @@ object TestAspect extends TimeoutVariants {
...
+      def perTest[R >: Nothing <: Any, E >: Nothing <: E0, S >: Unit <: Unit](
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
+      ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
+        lazy val failValue: ZIO[R, TestFailure[E], TestSuccess[S]] = ZIO.fail(
+          TestFailure.Runtime(zio.Cause.die(new RuntimeException("expected failure")))
+        )
+
+        test.foldCauseM(
+          _.untraced match {
+            case Cause.Fail(TestFailure.Runtime(Traced(e, _))) if p.run(e).isSuccess =>
```
ghostdogpr: `untraced` removes all inner `Traced` so this shouldn't be needed?

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -168,6 +169,41 @@ object TestAspect extends TimeoutVariants {
...
+              ZIO.succeed(TestSuccess.Succeeded(BoolAlgebra.unit))
+
+            case Cause.Fail(TestFailure.Runtime(e)) if p.run(e).isSuccess =>
+              ZIO.succeed(TestSuccess.Succeeded(BoolAlgebra.unit))
+
+            case other => {
+              println(s"other = $other")
```
ghostdogpr: You can remove this one

adamgfraser:
@ghostdogpr We are going to change the signature of this to take an `Assertion[TestFailure[E]]` which should let us remove most of this code and let the caller specify what subtypes of `TestFailure` they are accepting.

ghostdogpr:
@jaliss ah ok that makes sense 👍

test/shared/src/main/scala/zio/test/TestAspect.scala
jaliss: from what I tested I had to get rid of it before I could match things.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -543,6 +543,26 @@ object Assertion {
       }
     }
 
+  /**
+   * Makes a new assertion that verifies a TestFailure matches a given failure
+   */
+  final def testFails[E](failure: E): Assertion[TestFailure[E]] = {
```
ghostdogpr: maybe `failsWith`? @adamgfraser wdyt?

adamgfraser: I think we can simplify this. I am reviewing now.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -472,6 +472,11 @@ object Assertion {
       else BoolAlgebra.failure(AssertionValue(self, actual))
     }
 
+  final def isInstance[A](reference: A): Assertion[Any] =
```
adamgfraser: I would take this out and use `isSubType` in your test.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -168,6 +168,35 @@ object TestAspect extends TimeoutVariants {
       }
     }
 
+  /**
+   * An aspect that makes a test succeed if it fails for any cause.
```
adamgfraser: I think we need to update this a bit because it makes a test that failed for any reason into a success, it doesn't have to be because the test failed with a `Cause`. It could have just failed with an `AssertionFailure`. Also clarify that it makes a previously successful test into a failure.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -168,6 +168,35 @@ object TestAspect extends TimeoutVariants {
...
+   */
+  val failure: PerTest[Nothing, Any, Nothing, Any, Unit, Unit] = failure(Assertion.anything)
+
+  /**
+   * An aspect that makes a test succeed if it fails for a given cause.
```
adamgfraser: Same as above.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -168,6 +168,35 @@ object TestAspect extends TimeoutVariants {
...
+   */
+  def failure[E0](p: Assertion[TestFailure[E0]]): PerTest[Nothing, Any, Nothing, E0, Unit, Unit] =
+    new TestAspect.PerTest[Nothing, Any, Nothing, E0, Unit, Unit] {
+      def perTest[R >: Nothing <: Any, E >: Nothing <: E0, S >: Unit <: Unit](
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
+      ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
+        def fail(value: TestFailure[E]) = ZIO.fail(
```
adamgfraser: Maybe you can inline these definitions into `test.foldM` since they are each only used once?

PR #1743, opened 2019-09-21
Remove schedule output from ZStream.schedule[Elements]

### LINKED ISSUE #1534: Modify ZStream#schedule and ZStream#scheduleElements to discard the schedule's output

regiskuckaertz:
Closes #1534

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1628,7 +1628,7 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
   private def repeatedAndSpaced =
     unsafeRun(
       Stream("A", "B", "C")
-        .scheduleElements(Schedule.recurs(1) >>> Schedule.fromFunction((_) => "!"))
+        .scheduleElements(Schedule.once)
         .run(Sink.collectAll[String])
         .map(_ must_=== List("A", "A", "B", "B", "C", "C"))
     )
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1671,8 +1671,8 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * Repeats are done in addition to the first execution, so that `scheduleElements(Schedule.once)` means "emit element
    * and if not short circuited, repeat element once".
    */
-  final def scheduleElements[R1 <: R, A1 >: A](schedule: ZSchedule[R1, A, A1]): ZStream[R1 with Clock, E, A1] =
-    scheduleElementsEither(schedule).map(_.merge)
+  final def scheduleElements[R1 <: R, A1 >: A](schedule: ZSchedule[R1, A, _]): ZStream[R1 with Clock, E, A1] =
```
jdegoes: I'd use `Any` instead of `_`, as existentials are going away in Dotty.

iravid:
Yes! Awesome!

iravid:
Ah tests need to be adjusted

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Hmm, let's open an issue that checks for that. I have seen existentials in other places as well.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1628,27 +1628,27 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
   private def repeatedAndSpaced =
```
vasilmkd: Is there a possibility that this test is flaky? Can you take another look @regiskuckaertz ?

vasilmkd: Nevermind, bad rebase. Sorry.

PR #1744, opened 2019-09-21
WIP: issue-1182: foreachPar_

hanny24:
Part of #1182 
cc @LGLO

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOParCombinatorsSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOParCombinatorsSpec.scala
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package zio
+
+import zio.test._
+import zio.test.Assertion._
+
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/ZIOParCombinatorsSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+          for {
+            ref <- Ref.make(List.empty[Int])
+            _   <- ZIO.foreachPar_(as)(a => ref.update(_ :+ a))
+            rs  <- ref.get
+          } yield assert(rs.sorted, equalTo(as))
+        },
+        testM("foreachPar_ cancels other computations and propagates errors") {
+          val boom = new Exception("Boom!")
+          for {
+            semaphore <- Semaphore.make(0)
```
iravid: Could I bother you to use a `Ref[Int]` instead of a semaphore? We're going to get rid of `acquireN` and `releaseN`.

core-tests/shared/src/test/scala/zio/ZIOParCombinatorsSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+                        else (ZIO.never).onInterrupt(semaphore.release)
+                      }
+                      .flip
+            _ <- semaphore.acquireN(4)
+          } yield assert(error, equalTo(boom))
+        },
+        testM("foreachPar_ cancels other computations and propagates defects") {
+          val boom = new Exception("Boom!")
+          for {
+            semaphore <- Semaphore.make(0)
```
iravid: Same here

core-tests/shared/src/test/scala/zio/ZIOParCombinatorsSpec.scala
hanny24: I used `Semaphore` because `onInterrupt` might be called only after we already called `counter.get` where `counter` is `Ref[Int]`. We might be able to replace it with some sort of `eventually` call on `Ref[Int]`, but I'm not really sure if it is any better.

iravid: Ah I see. Ok. I'm fine with that. Let's tackle this when we change Semaphore.

PR #1751, opened 2019-09-21
 Checklist: Make everything private that can / should be private #1687

### LINKED ISSUE #1687: Checklist: Make everything private that can / should be private
All implementation details should be hidden away so they can be changed without breaking binary compatibility.

zezutom:
Closes #1687


core/shared/src/main/scala/zio/console/Console.scala
```diff
@@ -44,7 +44,7 @@ object Console extends Serializable {
       final def putStr(line: String): UIO[Unit] =
         putStr(SConsole.out)(line)
 
-      final def putStr(stream: PrintStream)(line: String): UIO[Unit] =
+      private[this] final def putStr(stream: PrintStream)(line: String): UIO[Unit] =
```
LGLO: It's public API and we want to putStr to streams other than Console sometimes. Could you, please, revert this and two other occurrences in this file?

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
```diff
@@ -76,7 +76,7 @@ private[stream] class StreamEffectChunk[-R, +E, +A](override val chunks: StreamE
   override def filter(pred: A => Boolean): StreamEffectChunk[R, E, A] =
     StreamEffectChunk(chunks.map(_.filter(pred)))
 
-  final def foldLazyPure[S](s: S)(cont: S => Boolean)(f: (S, A) => S): ZManaged[R, E, S] =
+  private[this] final def foldLazyPure[S](s: S)(cont: S => Boolean)(f: (S, A) => S): ZManaged[R, E, S] =
```
LGLO: I think it's public API as well. Isn't it @iravid ?

jdegoes:
@zezutom Thanks Tomas!

PR #1752, opened 2019-09-21
Moved raceAll test to ZIOSpec

LGLO:
@mijicd FYI I'm moving tests to where they belong.


mijicd:
Nice, I'll align with it.

PR #1753, opened 2019-09-21
WIP: Add a TestRuntime to search for race conditions and deadlocks

Jentsch:
This can replace TestUtils.nonFlacky

This is a spin off from: https://github.com/Jentsch/modelchecker

Things still to do:
- Add documentation
- Better names
- more tests
- ~~think about a total search~~


adamgfraser:
This is really interesting. I think my next step here would be to try to make `analyze` more functional by using a `Ref` instead of a mutable data structure, trying to use `ZIO.run` instead of `unsafeRunAsync` to describe running effects, and returning a lazily constructed stream of potential execution permutations instead of taking a specific number. Ideally the result of this would be a `Gen` where each evaluation would yield one execution path, possibly with two different versions where one returns a single element stream that effectually generates a random execution path each time and the other returns a finite stream of all execution paths.

adamgfraser:
I think we also want an actual `TestRuntime` that extends `Runtime` and could be used to execute effects with an "arbitrary" execution order in addition to a generator of effects with an arbitrary execution order. The first could just be used to run tests, including non-property based, with more "coverage" of different execution orders. The second could power property based testing with different legal execution orders for a program.

Jentsch:
The stream is now unlimited. The code got a bit shorter - a good sign.

Jentsch:
As performace is something to keep in mind, I added a simple benchmark. First results: analyse method makes the execution takes five times longer, than running with the normal `Runtime`.

| Benchmark    |       Cnt  |  Score |
|---------------|-----|----------:|
| defaultRuntime |   25 | 20.809 ± 0,098  ops/s |
| analyse       |   25  | 3.854 ± 0,090  ops/s |

adamgfraser:
@Jentsch Okay, that's very helpful in guiding the development. If that is right, which conceptually makes sense because we're taking this hand optimized runtime and adding logic between every step and mixing up the execution order, then this would definitely be something we used to power specific combinators that people could use to check for race conditions / nondeterminism rather than used for tests in general.

Jentsch:
I added the capability to just execute a subset of all test in test/.../test. Maybe this could be handled in a separate PR to keep this PR smaller.

With the latest change created `TestExecutionContext`. My idea was to reuse `ZIO#on(ExecutionContext)`. Than you can repeatedly call `TestExecutionContext.execNextStep` to slowly execute the watched ZIO. But `TestExecutionContext.execute` is never called. I tinkered with `fork`, but it didn't helped. Do you have a hint how to trigger the first `execute` call?

adamgfraser:
Yes, being able to run tests in ZIO Test internal test suite on a per test basis would be very nice. If we can let's break that out into a separate PR because that one we should be able to merge very quickly versus this one I think will take us more time to work through.

adamgfraser:
What was your thinking with the ExecutionContext? Are you trying to get better performance? It seems more complex than your prior version.

Jentsch:
I thought that would be nice to wrap everything `ZIO.on(new TestExecutionContext)` to construct a more random execution. But it didn't went that well.

adamgfraser:
@Jentsch Just wanted to let you know I haven't forgotten about this. I think I need to spend some more time on it myself to be able to give us a good path forward and will try to do that tomorrow. Sorry for the delay.

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,119 @@
...
+    val singleRun = runOnce(yielding)
+
+    Gen.fromEffect(singleRun)
+  }
+
+  private def runOnce[R <: Random, E, A](zio: ZIO[R, E, A]) =
+    for {
+      executionContext <- ZIO.effectTotal { new TestExecutionContext }
+      resultPlace      <- Ref.make(Option.empty[Exit[E, A]])
+      taks             = zio.run.flatMap(r => resultPlace.set(Some(r))).fork.on(executionContext)
```
adamgfraser: I don't think this is doing anything here.

Jentsch:
@adamgfraser: No problem, I was busy as well.
The latch example you gave works. I added it to my prototype: https://github.com/Jentsch/modelchecker/blob/master/zio/src/test/scala/zio/modelchecker/example/LatchSpec.scala. The `await` stops the execution until `succeed` lets it resume.

I don't want to say the isn't maybe a surprise in the way some effect types interact with each other, but this example works.

adamgfraser:
@Jentsch Nice! Does the direction I suggested make sense?

Jentsch:
Btw: I'm totally surprised how `Promise#complete` works. It allows to do arbitrary IO operations on `Promise#await`. For example this test is green, because the first `await` returns a different value than the second one:

```scala
for {
  p <- Promise.make[Nothing, Int]
  r <- zio.Ref.make(0)
  _ <- p.complete(r.update(_ + 1))
  d <- p.await
  e <- p.await
} yield assert(d, equalTo(1)) && assert(e, equalTo(2))
```
I couldn't find any test or documentation that specifies this behaviour. The inline documentation even says '[...] variable that can be set exactly once'. Is this intentional?

adamgfraser:
Wow, that is really interesting. Yeah, I kind of have the same reaction. I wouldn't have expected that at all and it definitely goes against the idea that a `Promise` is a "write once" data type. At the same time the signature does accept an arbitrary effect and I guess you could say that it is giving you the same effect each time, even though the evaluation of the effect gives a different result. Definitely one for broader discussion.

Jentsch:
@adamgfraser Broadly yes, it makes sense. But two minor things to remark:
1. The signature has be more like `def name(zio: ZIO[R, E, A]): ZStream[R, Nothing, Option[Exit[E, A]]]`, because the given `zio` may never terminate. I modelled this with result of `None`, otherwise with `Some[Exit]` value.
3. I would like to restore the [previous state](https://github.com/zio/zio/blob/183eb15d3ddde7716b120e4d298992cc3b776524/test/shared/src/main/scala/zio/test/TestRuntime.scala) and restart from there, any objections?
4. Did I have forgotten any questions from your side? This discussion toke some detours via other issues.

adamgfraser:
Yes, completely agree with you on working forward from the previous version.

On the `Option[Exit[E, A]]`, I didn't know how we would be able to tell whether an effect was non-terminating so it would be worth representing in the type signature, but maybe I'm not understanding what you are doing there.

Jentsch:
In `runOnce` we keep track of every `Runnable` that wants to be executed. If no more `Runnable` are available than nothing more is to do. If in this case `result` is still empty we can conclude that the final of the given `zio` never where executed. This can for example be the case if a `Promise#await` was invoked, but never a `Promise#complete`.

adamgfraser:
@Jentsch Okay, so here's one for you:

```scala
live(ZIO.sleep(10.seconds)) *> ZIO.succeed(1)
```

It looks like running this with `analyse` returns `None` because there are no pending runnables but the value has also not been filled.

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
+package zio.test
+
+import zio.Exit
+import zio.ZIO
+import zio.internal.Executor
+import zio.random.Random
+import zio.stream.ZStream
+
+object TestRuntime {
```
adamgfraser: I think we need a better name here because at this point we're not doing anything with the `Runtime`. What if we called it `RandomExecutor` and then we could have a trait that extended `Executor` and a companion object with methods for using the `RandomExecutor` along with the rewrite function to generate random execution paths?

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+            }
+            
+            isRunning = false
+          }
+          true
+        }
+    }
+  }
+
+  private def yieldingEffects[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
```
adamgfraser: What else do we need to do here?

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+  def paths[R <: Random, E, A](zio: ZIO[R, E, A]): ZStream[R, Nothing, Exit[E, A]] = {
+
+    val yielding = yieldingEffects(zio)
+
+    val singleRun = runOnce(yielding)
+
+    ZStream.repeatEffect(singleRun)
+  }
+
+  private def runOnce[R <: Random, E, A](zio: ZIO[R, E, A]) =
```
adamgfraser: Maybe we expose this as a public method as well with a more user friendly name. That way if a user just wants to go from a `ZIO` to a `ZIO` with a more "arbitrary" execution order they can call this method.

benchmarks/src/main/scala/zio/TestRuntimeBenchmark.scala
```diff
@@ -0,0 +1,40 @@
...
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import zio.IOBenchmarks._
+import zio.test.TestRuntime
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class TestRuntimeBenchmark {
```
adamgfraser: How do the benchmarks look at this point?

benchmarks/src/main/scala/zio/IOBenchmarks.scala
```diff
@@ -15,6 +15,34 @@ object IOBenchmarks extends DefaultRuntime {
     override val Platform = PlatformLive.Benchmark.withTracing(Tracing.enabled)
   }
 
+  val SyncRuntime = new DefaultRuntime {
```
adamgfraser: We may want to move this into the `TestRuntimeBenchmarks.scala` file as these are the main benchmarks we use for comparing `ZIO` against other effect systems so we don't necessarily want to clutter them up with work we are doing to evaluate our test implementations.

adamgfraser:
One other thing that would be helpful to motivate this is an example where this feature helps identify a bug or race condition that would be difficult to detect otherwise . In other words, why is this better than just running the test 100 times?

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+  private val makeTestExecutor: ZIO[Random, Nothing, Executor] =
+    ZIO.effectTotal {
+      new Executor {
+        val pendingRunnables = new java.util.Vector[Runnable]
+        var isRunning = false
+
+        def here         = true
+        def metrics      = None
+        def yieldOpCount = Int.MaxValue
+        def submit(runnable: Runnable): Boolean = synchronized {
```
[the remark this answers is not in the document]
Jentsch: Done

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+    ZStream.repeatEffect(singleRun)
+  }
+
+  private def runOnce[R <: Random, E, A](zio: ZIO[R, E, A]) =
+    makeTestExecutor.flatMap(yieldingEffects(zio).lock(_).run)
+
+  private val makeTestExecutor: ZIO[Random, Nothing, Executor] =
+    ZIO.effectTotal {
+      new Executor {
+        val pendingRunnables = new java.util.Vector[Runnable]
```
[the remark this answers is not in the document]
Jentsch: If I have to use mutation I usually prefer the Java collections. With the usage of AtomicReferences this is now an immutable Scala `Vector`. But I don't know what's the best collection for this task.

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+
+        def here         = true
+        def metrics      = None
+        def yieldOpCount = Int.MaxValue
+        def submit(runnable: Runnable): Boolean = synchronized {
+          if (isRunning)
+            pendingRunnables.add(runnable)
+          else {
+            isRunning = true
+            runnable.run()
```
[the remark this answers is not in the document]
Jentsch: I don't get what you mean. `submit` is called in three scenarios here:

1. the initial invocation where `isRunning` is `true`. Here we should start execution the `runnable` and start the evaluation loop.
2. an invocation while 1. is still running (e.g. `a fork or resolved promise). Here we should just add the `runnable` to `pendingRunnables`.
3. an invocation after 1. finished (e.g. `sleep`). Here we should resume the execution.

Does that makes sense?

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    val singleRun = runOnce(yielding)
+
+    ZStream.repeatEffect(singleRun)
+  }
+
+  private def runOnce[R <: Random, E, A](zio: ZIO[R, E, A]) =
+    makeTestExecutor.flatMap(yieldingEffects(zio).lock(_).run)
+
+  private val makeTestExecutor: ZIO[Random, Nothing, Executor] =
```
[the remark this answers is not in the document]
Jentsch: Done

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+import zio.internal.Executor
+import zio.random.Random
+import zio.stream.ZStream
+
+object TestRuntime {
+
+  /**
+   * Returns multiple possible outcomes of the given `zio`.
+   */
+  def paths[R <: Random, E, A](zio: ZIO[R, E, A]): ZStream[R, Nothing, Exit[E, A]] = {
```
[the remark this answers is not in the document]
Jentsch: Wouldn't that (re-) introduce the unneeded `Random` dependency to the user?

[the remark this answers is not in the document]
adamgfraser: No. `Gen` is polymorphic in the environment type so if a particular generator doesn't use `Random` it doesn't have to depend on it.

adamgfraser:
Looks like we are only seeing one execution path in the race example on ScalaJS now.

Jentsch:
I'll look at that

test/shared/src/main/scala/zio/test/TestRuntime.scala
Jentsch: Do you have any particular name in mind? Maybe `runArbitrary` or `arbitraryScheduling`?

adamgfraser: Not really. If we're going to have it return a `ZIO[R, Nothing, Exit[E, A]]` maybe just `run` so it would be `RandomExecutor#run` since `ZIO#run` already returns a `ZIO[R, Nothing, Exit[E, A]]`?

Jentsch: Done

Jentsch:
Yeah! The tests are green again. Took a while.

test/shared/src/main/scala/zio/test/TestRuntime.scala
```diff
@@ -0,0 +1,108 @@
...
+  private def runOnce[R <: Random, E, A](zio: ZIO[R, E, A]) =
+    makeTestExecutor.flatMap(yieldingEffects(zio).lock(_).run)
+
+  private val makeTestExecutor: ZIO[Random, Nothing, Executor] =
+    ZIO.effectTotal {
+      new Executor {
+        val pendingRunnables = new java.util.Vector[Runnable]
+        var isRunning = false
+
+        def here         = true
```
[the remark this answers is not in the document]
Jentsch: Okay, its now `false`.

test/shared/src/main/scala/zio/test/TestRuntime.scala
Jentsch: @adamgfraser the object is now called `RandomExecutor` and has only `paths` and `run` as public methods. Is this API ok?

test/shared/src/main/scala/zio/test/TestRuntime.scala
Jentsch: I think everything is done here. Would be good to rise the confidence.

Jentsch:
Updated benchmark results (higher is better):

|Benchmark                                   |           Score   |
|-----|---:|
|RandomExecutorInDefaultRuntime|        1,465 ± 0,024  ops/s |
|RandomExecutorInSyncRuntime |          1,601 ± 0,017  ops/s |
|runInDefaultRuntime    |       17,981 ± 0,450  ops/s |
|runInSyncRuntime       |       12,866 ± 0,105  ops/s |

So the `RandomExecutor` is roughly 10 times slower the direct simple execution. This could be improved, but I would put effort into to it for now.

benchmarks/src/main/scala/zio/TestRuntimeBenchmark.scala
Jentsch: I added new benchmark results below.

Jentsch:
To conquer the big question: 'why is this better than just running the test 100 times?' I tried a shut-gun approach and just let the `nonFlaky` `TestAspect` use the `RandomExecutor`.

The good news is that all tests stayed green., that shows shows that the implementation isn't totally flawed. The bad news are that all tests stayed green, no unknown behaviour was uncovered.

jdegoes:
@Jentsch I think this is extremely fascinating and could be developed into something that ships with ZIO Test as an option. Please re-open if you find time to bring it together!

PR #1754, opened 2019-09-21
Fix build issues introduced by bundling

mijicd:
Follow-up of #1728 .

As verified in `interop-monix`, the following changes needed to be introduced:
- `sbt-pgp` has to be downgraded to `1.1.0` in order to avoid issues with keys.
- `PGP_PASSPHRASE` needs to be explicitly loaded.
- Sonatype credentials has to be set on release.
- Only one workflow is required.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,6 +139,9 @@ release: &release
           name: Fetch git tags
           command: git fetch --tags
       - <<: *load_cache
+      - run: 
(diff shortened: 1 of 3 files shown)


mijicd:
Don't merge yet!

mijicd:
The fixes are verified in `interop-monix` and `interop-guava`. @ghostdogpr & co. please re-review!

Issue #1463, opened 2019-08-22
Add additional fold variants to ZStream


iravid:
Add additional `fold` variants to `ZStream`:
```scala
// These exist:
// A fold that can exit early
final def fold[R1 <: R, E1 >: E, A1 >: A, S](s: S)(cont: S => Boolean)(f: (S, A1) => ZIO[R1, E1, S]): ZIO[R1, E1, S]
// A fold that can exit early, that carries the stream's scope
final def foldManaged[R1 <: R, E1 >: E, A1 >: A, S](s: S)(cont: S => Boolean)(f: (S, A1) => ZIO[R1, E1, S]): ZManaged[R1, E1, S]
// A pure fold that consumes everything
final def foldLeft[A1 >: A, S](s: S)(f: (S, A1) => S): ZIO[R, E, S]

// These don't:
// An effectful fold that consumes everything
final def foldLeftM[R1 <: R, E1 >: E, A1 >: A, S](s: S)(f: (S, A1) => ZIO[R1, E1, S]): ZIO[R1, E1, S]
// A pure fold that consumes everything and carries the stream's scope
final def foldLeftManaged[A1 >: A, S](s: S)(f: (S, A1) => S): ZManaged[R, E, S]
// An effectful fold that consumes everything and carries the stream's scope
final def foldLeftManagedM[R1 <: R, E1 >: E, A1 >: A, S](s: S)(f: (S, A1) => ZIO[R1, E1, S]): ZManaged[R1, E1, S]
```

vasilmkd:
Let's first decide on the naming scheme.

kamilkloch:
How should it differ from the readily available `ZStream#foldManaged`?

iravid:
The signatures are:
```scala
// These exist:
// A fold that can exit early
final def fold[R1 <: R, E1 >: E, A1 >: A, S](s: S)(cont: S => Boolean)(f: (S, A1) => ZIO[R1, E1, S]): ZIO[R1, E1, S]
// A fold that can exit early, that carries the stream's scope
final def foldManaged[R1 <: R, E1 >: E, A1 >: A, S](s: S)(cont: S => Boolean)(f: (S, A1) => ZIO[R1, E1, S]): ZManaged[R1, E1, S]

// These don't:
// A pure fold that consumes everything
final def foldLeft[A1 >: A, S](s: S)(f: (S, A1) => S): ZIO[R, E, S]
// An effectful fold that consumes everything
[...]

PR #1756, opened 2019-09-21
Added additional fold variants to ZStream #1463.

kamilkloch:


vasilmkd:
Should we keep this consistency with `Chunk` as well? @iravid

iravid:
Yes, good point @vasilmkd

Issue #1759, opened 2019-09-21
Remove existential types from the ZIO codebase


vasilmkd:
Existential types are going away in Scala 3.

E.g. `ZStream#tap`.

LGLO:
@vasilmkd could you point me to some article about existential types removal?

vasilmkd:
https://dotty.epfl.ch/docs/reference/dropped-features/existential-types.html

LGLO:
`Existential types using forSome have been dropped.`
`Existential types that can be expressed using only wildcards (but not forSome) are still supported, but are treated as refined types.`
I think we are safe using `_` and don't have to change to `Any` - do you see any benefit of having `Any` instead of `_`?

LGLO:
I mean, there is no `forSome` in ZIO codebase.

vasilmkd:
#1743.

I opened this issue because of a remark by @jdegoes. I just like to see consistency in the codebase. I'm ok with both solutions I suppose. :-)

adamgfraser:
Are there any situations where `[_]` and `[Any]` would be semantically different?

vasilmkd:
I'm guessing it depends on the variance of the type parameter? So for a contravariant parameter, it probably needs to be `Nothing`.

LGLO:
I cannot contribute to the discussion about difference of `[_]` and `[Any]` I'd just hold on making massive changes without justification. Maybe someone has strong arguments in this discussion.

vasilmkd:
I don't have strong feelings on this subject. I opened the issue for documentation purposes mainly.

LGLO:
:+1: Good, I hope I'll learn something here!

neko-kai:
wrt difference between Any and _ - this has been discussed before in

https://github.com/zio/zio/issues/381
https://github.com/zio/zio/pull/414

adamgfraser:
Can we close this? Or maybe we should wrap it into #1298 regarding updating the contributor's guide with clear guidance for this and other areas. I definitely don't feel as qualified as some others to do #1298 but I'm happy to take a stab at it and people can tell me where I am wrong.

neko-kai:
Hmm, there's a Scala bug with an example where scala fails to unify types – https://github.com/scala/bug/issues/11769, I thought it would be a good example for where `_` is better than `Any` – but surprisingly in this case `Any` works well, while both `_` and a type parameter fail!

```scala
// scala 2.12 only
object example {
[...]

PR #1760, opened 2019-09-21
Fix Flaky ZManaged#withEarlyRelease Test

### LINKED ISSUE #1747: ZManaged#withEarlyRelease test is flaky
```
[info]   - withEarlyRelease
[info]     + Provides a canceler that can be used to eagerly evaluate the finalizer
[info]     - The canceler should run uninterruptibly
[info]       Right(Some(Success(()))) did not satisfy isRight(isNone())
[info]       Some(Success(())) did not satisfy isNone()
```

@adamgfraser could you take a look?

adamgfraser:
Resolves #1747.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
@@ -670,8 +670,8 @@ object ZManagedSpec
                            for {
                              fiber        <- canceler.fork
                              _            <- latch.await
+                             interruption <- withLive(fiber.interrupt)(_.timeout(5.seconds)).either
                              _            <- ref.set(false)
-                             interruption <- fiber.interrupt.timeout(5.seconds).provide(zio.clock.Clock.Live).either
                            } yield interruption
                        }
             } yield assert(result, isRight(isNone))
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -670,15 +670,15 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) 
    * uninterruptibly with an exit value indicating that the effect was
    * interrupted, and if completed will cause the regular finalizer to not run.
    */
-  final def withEarlyRelease: ZManaged[R, E, (URIO[R, _], A)] =
+  final def withEarlyRelease: ZManaged[R, E, (URIO[R, Any], A)] =
     withEarlyReleaseExit(Exit.interrupt)
 
   /**
    * A more powerful version of `withEarlyRelease` that allows specifying an
    * exit value in the event of early release.
    */
-  final def withEarlyReleaseExit(exit: Exit[_, _]): ZManaged[R, E, (URIO[R, _], A)] =
-    ZManaged[R, E, (URIO[R, _], A)] {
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -670,15 +670,15 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
+  final def withEarlyRelease: ZManaged[R, E, (URIO[R, Any], A)] =
     withEarlyReleaseExit(Exit.interrupt)
 
   /**
    * A more powerful version of `withEarlyRelease` that allows specifying an
    * exit value in the event of early release.
    */
-  final def withEarlyReleaseExit(exit: Exit[_, _]): ZManaged[R, E, (URIO[R, _], A)] =
-    ZManaged[R, E, (URIO[R, _], A)] {
+  final def withEarlyReleaseExit(exit: Exit[Any, Any]): ZManaged[R, E, (URIO[R, Any], A)] =
```
LGLO: Is this because of Scala 3?

core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
```diff
@@ -670,8 +670,8 @@ object ZManagedSpec
                            for {
                              fiber        <- canceler.fork
                              _            <- latch.await
+                             interruption <- withLive(fiber.interrupt)(_.timeout(5.seconds)).either
```
LGLO: What is semantic change here?
Before `fiber.interrupt` was running with `Clock.Live` and now only `.timeout` is running with it?
How does it help with flakiness?

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: Yes, just trying to implement #1759 and at least not add any more existential types.

core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
adamgfraser: No, you're totally fine. Changing `provide(Clock.Live)` to `withLive` doesn't have any semantic effect, it is just trying to clean up the code a little.

The semantically significant change is moving the `ref.set(false)` effect to after the interruption. I am using the `Ref` so that the cancelation action is non-terminating and we can check that it is uninterruptible. But since it never terminates the finalizer will still run at the end of the `Managed`, and we want that to terminate so that we don't hang. The problem is that created a race condition, because if the `Ref` is set to `false` before the forked canceler checks the `whenM` condition the canceler will become terminating and our test will fail, causing the flakiness. We fix that by moving the effect to set the Ref to after we have checked if we can interrupt the canceler.

Does that make sense?

LGLO: Yes, it does!
I think the `.either` doesn't bring any value, E is Nothing here and `.either` doesn't help with defects.

adamgfraser:
Let me make one change before merging please.

ghostdogpr:
@adamgfraser sorry I merged it right before I saw this :/

adamgfraser:
@ghostdogpr No problem. Sorry, was just going to remove the `either` call as @LGLO suggested but don't think it is worth changing at this point.

PR #1763, opened 2019-09-21
Parameterized environment on zio.Canceler

### LINKED ISSUE #1762: zio.Canceler should be parameterized
Canceler should be changed to
```scala
  type Canceler[R] = URIO[R, _]
```

ajaychandran:
Fixes #1762

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -164,7 +164,7 @@ object IO {
   /**
    * @see See [[zio.ZIO.effectAsyncInterrupt]]
    */
-  final def effectAsyncInterrupt[E, A](register: (IO[E, A] => Unit) => Either[Canceler, IO[E, A]]): IO[E, A] =
+  final def effectAsyncInterrupt[E, A](register: (IO[E, A] => Unit) => Either[Canceler[Any], IO[E, A]]): IO[E, A] =
     ZIO.effectAsyncInterrupt(register)
 
   /**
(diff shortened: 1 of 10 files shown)


adamgfraser:
I think we are trying to avoid existential types for Dotty compatibility so we would want to use `Any` instead of `_` in type parameters.

neko-kai:
@adamgfraser Wildcards are fine, dotty drops only explicit `forSome` syntax

adamgfraser:
@neko-kai Ah, I misunderstood. Thanks!

ajaychandran:
@neko-kai 
Hi,
I just noticed that the call to `Runtime.unsafeRunAsync_` in `ZStream.effectAsyncInterrupt` could be simplified.
```scala
// Current
                             runtime.unsafeRunAsync_(
                               k.foldCauseM(
                                 Pull.sequenceCauseOption(_) match {
                                   case None    => output.offer(Pull.end).unit
                                   case Some(c) => output.offer(Pull.halt(c)).unit
                                 },
                                 a => output.offer(Pull.emit(a)).unit
                               )
                             )
```
```scala
// .unit calls removed
            r.unsafeRunAsync_(k.foldCause(Pull.sequenceCauseOption(_) match {
              case None    => Pull.end
              case Some(c) => Pull.halt(c)
            }, Pull.emit) >>= q.offer))

```
Should this be changed?

neko-kai:
@ajaychandran
Sure, makes sense to me :+1:

ajaychandran:
Removing the `.unit` calls would shave off 1 `flatMap` per `Pull`. So I went with that version.

PR #1767, opened 2019-09-22
Restart schedule on end

### LINKED ISSUE #1532: Enhance ZStream#scheduleWith to not short-circuit the stream
Currently, when the schedule supplied to ZStream#scheduleWith signals completion, the stream is ended.

In this ticket, change the behavior to restart the schedule instead; for a stream with elements `"A B C A B C"`, and a schedule `(Schedule.exponential(2.millis) && Schedule.recur(2) && Schedule.const("Done")).map(_._2)`, this should be the resulting stream:
```scala
A |-2ms-| B |-4ms-| C |-8ms-| "Done" A |-2ms-| B |-4ms-| C |-8ms-| "Done"
```

regiskuckaertz:
Closes #1532 

cc @iravid

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
@@ -1600,16 +1600,16 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
 
   private def scheduleWith =
     unsafeRun(
-      Stream("A", "B", "C")
-        .scheduleWith(Schedule.recurs(3) *> Schedule.fromFunction((_) => "!"))(_.toLowerCase, identity)
+      Stream("A", "B", "C", "A", "B", "C")
+        .scheduleWith(Schedule.recurs(2) *> Schedule.fromFunction((_) => "Done"))(_.toLowerCase, identity)
         .run(Sink.collectAll[String])
-        .map(_ must_=== List("a", "b", "c", "!"))
+        .map(_ must_=== List("a", "b", "c", "Done", "a", "b", "c", "Done"))
     )
 
   private def scheduleEither =
     unsafeRun(
       Stream("A", "B", "C")
-        .scheduleEither(Schedule.recurs(3) *> Schedule.fromFunction((_) => "!"))
+        .scheduleEither(Schedule.recurs(2) *> Schedule.fromFunction((_) => "!"))
         .run(Sink.collectAll[Either[String, String]])
         .map(_ must_=== List(Right("A"), Right("B"), Right("C"), Left("!")))
     )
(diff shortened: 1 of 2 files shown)


streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1600,16 +1600,16 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
 
   private def scheduleWith =
     unsafeRun(
-      Stream("A", "B", "C")
-        .scheduleWith(Schedule.recurs(3) *> Schedule.fromFunction((_) => "!"))(_.toLowerCase, identity)
+      Stream("A", "B", "C", "A", "B", "C")
+        .scheduleWith(Schedule.recurs(2) *> Schedule.fromFunction((_) => "Done"))(_.toLowerCase, identity)
```
iravid: Hmm why did it change to `2`?

regiskuckaertz: `Schedule.recurs(2)` will consume three elements, as in the test for `scheduleWith`

iravid: Ok let me wrap my head around this for a sec..running your state machine manually ;)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1812,24 +1812,28 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
         init  <- schedule.initial.toManaged_
         state <- Ref.make[(Boolean, schedule.State, Option[() => B])]((false, init, None)).toManaged_
         pull = state.get.flatMap {
-          case (done, sched, decision0) =>
+          case (done, sched0, finish0) =>
```
iravid: We're not actually setting `done = true` anywhere aren't we?

regiskuckaertz: ha, correct, I'll get rid of it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1810,30 +1810,35 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                            decision <- schedule.update(a, sched0)
+                            _        <- clock.sleep(decision.delay)
+                            sched    <- if (decision.cont) UIO.succeed(decision.state) else schedule.initial
+                            finish   = if (decision.cont) None else Some(decision.finish)
+                            _        <- state.set((sched, finish))
+                          } yield f(a)
+
+                        // The stream ends when both the underlying stream ends and the final
+                        // schedule value has been emitted
+                        case None => Pull.end
```
iravid: This means that we don't emit the schedule's final value if the stream ended before the schedule ended. Is that intentional?

regiskuckaertz: Yes. If for instance the stream is completely empty, there never was a value of type `a` to update the schedule.

iravid: Ok, I'm good with that 👍🏻

Issue #1460, opened 2019-08-22
Port ZStream.fromInputStream (the one in platform.scala) to StreamEffect


iravid:
Follow-up to #1432.

jsilva:
I am taking this one

ghostdogpr:
Closed by #1772

PR #1772, opened 2019-09-22
Port ZStream.fromInputStream (the one in platform.scala) to StreamEffect #1460

jsilva:
Pull request for #1460

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -16,7 +16,7 @@
 
 package zio.stream
 
-import java.io.InputStream
+import java.io.{ IOException, InputStream }
 
 import zio._
 
@@ -56,10 +56,11 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E,
         Managed.effectTotal {
           var done = false
 
-          () => {
-            if (done) StreamEffect.end
-            else pred.applyOrElse(thunk(), (_: A) => { done = true; StreamEffect.end })
-          }
+          () =>
+            {
+              if (done) StreamEffect.end
+              else pred.applyOrElse(thunk(), (_: A) => { done = true; StreamEffect.end })
+            }
         }
       }
     }
@@ -80,7 +81,8 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E,
             } else pull()
           }
 
-          () => pull()
+          () =>
+            pull()
         }
       }
     }
@@ -95,7 +97,8 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E,
             if (pred(a)) a else pull()
           }
 
-          () => pull()
+          () =>
+            pull()
         }
       }
     }
@@ -138,11 +141,12 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E,
         Managed.effectTotal {
           var state = s1
 
-          () => {
-            val (s2, b) = f1(state, thunk())
-            state = s2
-            b
-          }
+          () =>
+            {
+              val (s2, b) = f1(state, thunk())
+              state = s2
+              b
+            }
         }
       }
     }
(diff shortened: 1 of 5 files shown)


streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -320,6 +322,25 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
+  final def fromInputStream(is: InputStream): StreamEffect[Any, Nothing, Byte] =
+    StreamEffect[Any, Nothing, Byte] {
+      Managed.effectTotal {
+        def pull(): Byte = {
+          val a = is.read()
```
regiskuckaertz: this can throw an `IOException`, why not `StreamEffect[Any, IOException, Byte]` then?

iravid: Yes, good point. We need to try/catch the `IOException` and re-throw it using in `StreamEffect.fail`.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -322,19 +331,22 @@ private[stream] object StreamEffect extends Serializable {
...
+    StreamEffect[Any, IOException, Chunk[Byte]] {
       Managed.effectTotal {
-        def pull(): Byte = {
-          val a = is.read()
-          if (a < 0) {
-            end
-          } else {
-            a.toByte
-          }
+        val buf = Array.ofDim[Byte](chunkSize)
```
iravid: The `buf` has to be re-created every time `pull` runs, or we'll be mutating chunks that have already been emitted.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -322,19 +331,22 @@ private[stream] object StreamEffect extends Serializable {
...
+        def pull(): Chunk[Byte] = {
+          val bytesRead = is.read(buf)
```
iravid: this needs to be wrapped in try/catch and should catch `IOException` and re-throw it  using `StreamEffect.fail`

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -322,19 +331,22 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def fromInputStream(is: InputStream): StreamEffect[Any, Nothing, Byte] =
-    StreamEffect[Any, Nothing, Byte] {
+  final def fromInputStream(
```
iravid: I think this needs to go into a JVM-specific file?

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -320,6 +322,25 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
+  final def fromInputStream(is: InputStream): StreamEffect[Any, Nothing, Byte] =
+    StreamEffect[Any, Nothing, Byte] {
+      Managed.effectTotal {
```
regiskuckaertz: This is not a total effect, I can see `Managed.makeEffect` which is probably more appropriate here.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
jsilva: I am not sure, however scala.js has https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/io/InputStream.scala

iravid: Oh well that simplifies everything. We can keep it here then :-)

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
iravid: Well the `Managed` only returns a thunk that, so in that regard it *should* be total

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
jsilva: Updated this

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
jsilva: updated, thanks for help

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: This is how we treated all `StreamEffect` code so far.

PR #1424, opened 2019-08-17
Implement raceAll without folding with `race` #1182

chikei:

PR #1773, opened 2019-09-22
issue-1182: foreachPar, foreachPar_, collectAllPar, reduceAllPar

hanny24:
Partially solves #1182, missing `raceAll` and `forkAll`.
Inspired by https://github.com/zio/zio/pull/1424


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -57,6 +57,31 @@ object ZIOSpec
...
             assertM(ZIO.fail(42).raceAll(List(IO.succeed(24) <* live(ZIO.sleep(100.millis)))), equalTo(24))
           }
+        ),
+        suite("foreachPar")(
+          testM("runs effects in parallel") {
+            assertM(for {
+              p <- Promise.make[Nothing, Unit]
+              _ <- UIO.foreachPar(List(UIO.never, p.succeed(())))(a => a).fork
+              _ <- p.await
+            } yield true, equalTo(true))
```
LGLO: Nitpick: `equalTo(true)` => `isTrue`

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -57,6 +57,31 @@ object ZIOSpec
...
+            val ints = List(1, 2, 3, 4, 5, 6)
+            val odds = ZIO.foreachPar(ints) { n =>
+              if (n % 2 != 0) ZIO.succeed(n) else ZIO.fail("not odd")
+            }
+            assertM(odds.flip, equalTo("not odd"))
+          },
+          testM("interrupts effects on first failure") {
+            val actions = List(
+              ZIO.never,
+              ZIO.succeed(1),
```
LGLO: Could you update some `Ref` with `delay` to check that this effect got interrupted (not updated the ref)?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,79 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+          todo.modify { t =>
+            buffer.update(idx, a)
+            if (t == 1) promise.succeed(buffer.toList).unit -> 0 else UIO.unit -> (t - 1)
+          }.flatten
+      )
+
+    (for {
+      size    <- UIO.effectTotal(as.size)
+      todo    <- Ref.make(size)
+      buffer  <- UIO.effectTotal(new Array[Any](size).asInstanceOf[Array[B]])
```
freskog: Will this not cause a java.lang.ClassCastException ? A quick experiment from the scala repl:

```
scala> new Array[Any](100).asInstanceOf[Array[String]]
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;
  ... 36 elided
```

hanny24: I think the trick here is not to cast it to specific type, but only to a generic one.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,79 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+      promise <- Promise.make[E, List[B]]
+      c <- ZIO.uninterruptibleMask { restore =>
+            for {
+              as <- ZIO.traverse(as)(a => ZIO.interruptible(fn(a)).fork)
+              _ <- as.view.zipWithIndex.foldLeft[ZIO[R, E, _]](ZIO.unit) {
```
freskog: The call to .view here seems unnecessary, or am I missing something?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,79 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+                    case (io, (f, idx)) =>
+                      io *> f.await.flatMap(arbiter(as, promise, buffer, todo, idx)).fork
```
freskog: An alternative to this is to combine this step with the ZIO.traverse(as)( ...) call above, so that one less fiber per element in as is created.

core/shared/src/main/scala/zio/ZIO.scala
hanny24: I did not want to materialize it. But if we change to `ZIO.traverse` we will need to materialize it anyway.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,79 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-      .succeed(as.iterator)
-      .flatMap { i =>
-        def loop(a: A): ZIO[R, E, Unit] =
-          if (i.hasNext) f(a).zipWithPar(loop(i.next))((_, _) => ())
-          else f(a).unit
-        if (i.hasNext) loop(i.next)
-        else ZIO.unit
-      }
-      .refailWithTrace
+  final def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, _]): ZIO[R, E, Unit] = {
```
freskog: This function is duplicating a lot from the foreachPar method, maybe this could be rewritten to leverage that function instead?

core/shared/src/main/scala/zio/ZIO.scala
hanny24: I don't think it actually that easy to combine it: we need to be able to interrupt all fibers from `arbiter` helper method.

core/shared/src/main/scala/zio/ZIO.scala
hanny24: I know it's possible, I just wanted to avoid allocation of the extra array with results.

core/shared/src/main/scala/zio/ZIO.scala
freskog: I still don't get how this works at runtime though, because the array will need to get a concrete type at some stage right?

```
scala> def test[A](a:A):Array[A] = new Array[Any](1).asInstanceOf[Array[A]]
test: [A](a: A)Array[A]

scala> test(1)
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [I
  ... 28 elided
```

core/shared/src/main/scala/zio/ZIO.scala
hanny24: Rewritten in a way that uses `ZIO.traverse_`

core/shared/src/main/scala/zio/ZIO.scala
hanny24: Isn't is caused by specialization? There is even a test that uses `foreachPar` on a `List[Int]`: https://github.com/hanny24/scalaz-zio/blob/35bd04f28cedeb3c988a83af206c37d48d3b58ad/core-tests/shared/src/test/scala/zio/ZIOSpec.scala#L70

freskog: Yeah saw that test now. I have to admit that I don't understand how this works. It clearly does though :)

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,78 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+          todo.modify { t =>
+            buffer.update(idx, a)
+            if (t == 1) promise.succeed(buffer.toList).unit -> 0 else UIO.unit -> (t - 1)
+          }.flatten
+      )
+
+    (for {
+      size    <- UIO.effectTotal(as.size)
+      todo    <- Ref.make(size)
+      buffer  <- UIO.effectTotal(new Array[Any](size).asInstanceOf[Array[B]])
```
```suggestion
      buffer  <- UIO.effectTotal(Array.ofDim[B](size))
```

regiskuckaertz: Also, it's probably best to move the buffer into a `Ref`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,78 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+      promise <- Promise.make[E, List[B]]
+      c <- ZIO.uninterruptibleMask { restore =>
+            for {
+              as <- ZIO.traverse(as)(a => ZIO.interruptible(fn(a)).fork)
```
regiskuckaertz: 🤔 Can we look for a solution that only traverses the structure once and forks one fiber per element?
```scala
ZIO.traverse(as.zipWithIndex){ case (a, i) => fn(a).flatMap(arbiter(...)).fork }
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1966,29 +1966,78 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+      promise: Promise[E, List[B]],
+      buffer: Array[B],
+      todo: Ref[Int],
+      idx: Int
+    )(res: Exit[E, B]): ZIO[R, Nothing, Unit] =
+      res.foldM[R, Nothing, Unit](
+        e => promise.halt(e).unit *> Fiber.interruptAll(fibers),
+        a =>
+          todo.modify { t =>
+            buffer.update(idx, a)
```
regiskuckaertz: That's an effect which should be captured, making `buffer` a `Ref[Array[B]]` will help here.

core/shared/src/main/scala/zio/ZIO.scala
LGLO: What is gain of keeping *mutable* array in `Ref`? I think that reference will never change, so `Ref#modify` will never discover that value was changed.
Secondly - we know that each fiber operates only on it's index, so that are really independent operations.
What can go wrong?

Re: `ofDim` - I understand this gives better performance for primitive types but requires `ClassTag[B]` - do we want to change method signatures?

core/shared/src/main/scala/zio/ZIO.scala
LGLO: For sure!

LGLO: Yes, in #1937

regiskuckaertz:
@hanny24 hi 👋  gonna close this as all these combinators have recently be refactor into stack-safe array-backed versions. Check out https://github.com/zio/zio/pull/2590

PR #1774, opened 2019-09-22
Specify that this fiber will be "inner interrupted" on joining an interrupted Fiber

neko-kai:


core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -60,7 +60,9 @@ trait Fiber[+E, +A] { self =>
   /**
    * Joins the fiber, which suspends the joining fiber until the result of the
    * fiber has been determined. Attempting to join a fiber that has errored will
-   * result in a catchable error, _if_ that error does not result from interruption.
+   * result in a catchable error. Joining an interrupted fiber will result in an
+   * "inner interruption" of this fiber, unlike interruption triggered by another
+   * fiber, "inner interruption" can be catched and recovered.
```
LGLO: toes This comment update reflects that calling `.interrupt` before `.join` places `Exit` into result channel?

neko-kai: It will place `Cause.interrupt` in the result channel and cause your fiber to terminate. I've read the previous comment as saying that joining the interrupted fiber will not result in error and got confused.

PR #1778, opened 2019-09-22
Align Chunk's folding methods with ZStream #1761.

### LINKED ISSUE #1761: Align Chunk's folding methods with ZStream

kamilkloch:
Resolves #1761

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
@@ -133,7 +133,7 @@ class ZStreamChunk[-R, +E, +A](val chunks: ZStream[R, E, Chunk[A]]) { self =>
    */
   final def flatMap[R1 <: R, E1 >: E, B](f0: A => ZStreamChunk[R1, E1, B]): ZStreamChunk[R1, E1, B] =
     ZStreamChunk(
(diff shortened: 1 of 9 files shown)


streams/shared/src/main/scala/zio/stream/ZStreamChunk.scala
```diff
@@ -141,36 +141,49 @@ class ZStreamChunk[-R, +E, +A](val chunks: ZStream[R, E, Chunk[A]]) { self =>
   final def flattenChunks: ZStream[R, E, A] = chunks.flatMap(ZStream.fromChunk)
 
   /**
-   * Executes an effectful fold over the stream of values.
+   * Executes a pure fold over the stream of values - reduces all elements in the stream to a value of type `S`.
```
iravid: I would like to suggest that we drop the fold variants that operate on chunks here entirely, because they're accessible by just doing something like `stream.chunks.foldM`. What do you think @kamilkloch?

Instead of those, let's expand the methods on `ZStreamChunk` to include parallels to all the fold variants on `ZStream`. These methods should ideally delegate to the folds on `Chunk` where possible.

kamilkloch: @iravid Please take a look. I had to force-push the changes due to conficlts.

Issue #1730, opened 2019-09-21
Add a method for converting ZStream[Byte] to a ZManaged[java.io.InputStream]


iravid:

sideeffffect:
I'll do it :)

Cereopsis:
@iravid Any requirements on the completeness  the InputStream implementation i.e should `markSupported()` return `true`?

Cereopsis:
Lol, answering my own question, I think no is the answer as it would seem problematic and anyway the default implementation is `false`. I guess the user (developer!) should be aware of the IOException thrown by `reset()`

sideeffffect:
still free to take? if yes, I will try to tackle this

Cereopsis:
@sideeffffect Please feel free, I’m a little light on ZStream knowledge so although I think the end solution won’t be that complex, there’ll be a good bit of back and forth to get there 🤪. Not to sway your own style but I had a brief chat with @iravid yesterday and he mentioned that ZStreamEffectChunk might be useful?

Cereopsis:
@sideeffffect ...you at the hackathon, by the way?

sideeffffect:
not anymore :D (at the airport now, but proudly attended :muscle: )

PR #1784, opened 2019-09-22
Add a method for converting ZStream[Byte] to a ZManaged[java.io.InputStream]

sideeffffect:
fixes https://github.com/zio/zio/issues/1730


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -5,6 +5,50 @@ import java.io.{ IOException, OutputStream }
 import zio._
 import zio.blocking._
 
+trait ZStreamPlatformSpecific {
```
regiskuckaertz: Can we mark this as package-private?

iravid:
@sideeffffect what do you think of mixing the platform-specific trait directly into ZStream and using implicit evidence instead of implicit classes?

iravid:
Oh actually someone pointed out in another PR that ScalaJS has InputStream. So we can move this code directly into ZStream.

fsvehla:
Should we read more than a single byte on every `read` call?

fsvehla:
`close` on the InputStream looks to me like a noop

sideeffffect:
there is `toInputStream` for all `ZStream`, `StreamEffect`, `ZStreamChunk` and `StreamEffectChunk`, as proper (non-extension) methods

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
```diff
@@ -90,6 +90,20 @@ private[stream] class StreamEffectChunk[-R, +E, +A](override val chunks: StreamE
   override def mapConcatChunk[B](f: A => Chunk[B]): StreamEffectChunk[R, E, B] =
     StreamEffectChunk(chunks.map(_.flatMap(f)))
 
+  final def processChunk: ZManaged[R, E, () => A] =
```
iravid: Oh nice. This will be useful.

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
```diff
@@ -90,6 +90,20 @@ private[stream] class StreamEffectChunk[-R, +E, +A](override val chunks: StreamE
...
+    chunks.processEffect.map { thunk =>
+      var counter         = 0
+      var chunk: Chunk[A] = Chunk.empty
+      def pull(): A = {
+        while (counter >= chunk.length) {
```
iravid:
```suggestion
        if (counter >= chunk.length) {
```

sideeffffect: this case doesn't cover case of an empty chunk! the same logic is used in the analogous method `ZStreamChunk.process`

iravid: Oh right! Good catch!

iravid:
Ok, needs a quick rebase and we're good

sideeffffect:
@iravid so are we sure that the overhead of `Chunk` is low enough that we don't need also `toInputStream` from `Stream[_, Array[Byte]]`?

iravid:
@sideeffffect Chunk definitely has some overhead, but I believe that's offset by the utility of combinators available on ZStreamChunk. We're also designing the rest of the library to always return chunks, rather than arrays, when applicable - e.g. `ZStream.fromInputStream`.

sideeffffect:
my builds are now failing with
```
[warn] Thread[scala-execution-context-global-672,5,zio-timer] loading zio/RTSSpec.class after test or run has completed. This is a likely resource leak.
```
```
java.lang.OutOfMemoryError: GC overhead limit exceeded
```
and
```
[warn] Thread[ForkJoinPool-1-worker-1,5,zio-default-blocking] loading zio.RTSSpec$$anonfun$testTimeoutOfLongComputation$1$$anonfun$apply$10$$anonfun$apply$mcZ$sp$2$$anonfun$apply$321$$anonfun$apply$11 after test or run has completed. This is a likely resource leak.
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at scala.collection.mutable.ListBuffer.$plus$eq(ListBuffer.scala:174)
	at scala.collection.mutable.ListBuffer.$plus$eq(ListBuffer.scala:45)
...
```

I don't know how my changes could introduce such a problem... :-/ does it happen also in `master`?

streams-tests/jvm/src/test/scala/zio/stream/ChunkUtils.scala
```diff
@@ -14,7 +14,7 @@ trait ChunkUtils {
       chunk  = Chunk.fromIterable(vector)
     } yield (chunk, i)
 
-  def chunkGen[R <: Random, A: ClassTag](a: Gen[R, A], max: Int): Gen[R with Sized, Chunk[A]] =
+  private def chunkGen[R <: Random, A: ClassTag](a: Gen[R, A], max: Int): Gen[R with Sized, Chunk[A]] =
```
iravid: Can we keep this public? No reason for it to be private.

sideeffffect:
@iravid ready to merge

PR #1785, opened 2019-09-22
ISSUE-1647: FiberRefSpec

zezutom:
#1647 FiberRefSpec


core-tests/shared/src/test/scala/zio/FiberRefSpec.scala
```diff
@@ -1,187 +1,186 @@
...
-    """
-
+import zio.test.Assertion._
+import zio.test._
+import zio.test.mock.MockClock
+import zio.duration._
+
+object FiberRefSpec
+    extends ZIOBaseSpec(
+      suite("Create a new FiberRef with a specified value and check if:")(
```
ghostdogpr: Can you put this in a root suite called `FiberRefSpec`? Just to be consistent with other tests.

PR #1786, opened 2019-09-22
Return and display summary of failed and ignored tests at the very en…

kapke:
…d of tests output

Example test output with failing tests from master (notice that you have to scroll through whole output to know which specs did fail):
```
[info] - FunctionIOSpec
[info]   - Check if the functions in `FunctionIO` work correctly
[info]     + `fromFunction` lifts from A => B into effectful function
[info]     + `identity` returns the identity of the input without modification
[info]     + `>>>` is a symbolic operator of `andThen`which does a Backwards composition of effectful functions
[info]     - `<<<` is a symbolic operator of `compose` which compose two effectful functions
[info]       13 did not satisfy equalTo(14)
[info]     + `zipWith` zips the output of two effectful functions
[info]     + `&&&` zips the output of two effectful functions and returns a tuple of their result
[info]     + `|||` computes two effectful functions left and right from from an Either input
[info]     + `first` returns a tuple: the output on the first element and input on the second element
[info]     + `second` returns a tuple: the input on the first element and output on the second element
[info]     + `left` takes an Either as input and computes it if it is Left otherwise returns the same value of the input
[info]     + `right`takes an Either as input and computes it if it is Right otherwise returns the same value of the input
[info]     + `asEffect` returns the input value
[info]     - `test` check a condition and returns an Either output: Left if the condition is true otherwise false
[info]       Right(List(1, 2, 5)) did not satisfy isRight(equalTo(List(1, 5)))
[info]       List(1, 2, 5) did not satisfy equalTo(List(1, 5))
[info]     + `ifThenElse`
[info]       + check an impure condition if it is true then computes an effectful function `then0` else computes `else0`
[info]       + check a pure condition if it is true then computes an effectful function `then0` else computes `else0`
[info]     + `whileDo`
[info]       + take a condition and run the body until the condition will be  false with impure function
[info]       + take a condition and run the body until the condition will be  false with pure function
[info]     + `_1` extracts out the first element of a tuple
[info]     + `_2` extracts out the second element of a tuple
[info]     - `fail` returns a failure
[info]       Left(error) did not satisfy isLeft(equalTo(errr))
[info]       error did not satisfy equalTo(errr)
[info]     + `effect` can translate an Exception to an error
[info]     + `ignore` ignores a effect failure
[info] Ran 22 tests in 961 ms: 19 succeeded, 0 ignored, 3 failed
[info] - ZManaged
[info]   + make
[info]     + Invokes cleanups in reverse order of acquisition.
[info]     + Properly performs parallel acquire and release
[info]     + Constructs an uninterruptible Managed value
[info]   + makeEffect
[info]     + Invokes cleanups in reverse order of acquisition.
[info]   + reserve
[info]     + Interruption is possible when using this form
[info]   + makeExit
[info]     + Invokes with the failure of the use
[info]     + Invokes with the failure of the subsequent acquire
[info]   + ensuring
[info]     + Runs on successes
[info]     + Runs on failures
[info]     + Works when finalizers have defects
[info]   + ensuringFirst
[info]     + Runs on successes
[info]     + Runs on failures
[info]     + Works when finalizers have defects
[info]   + flatMap
[info]     + All finalizers run even when finalizers have defects
[info]   - foldM
[info]     - Runs onFailure on failure
[info]       List(1, 1) did not satisfy equalTo(List(1))
[info]     + Runs onSuccess on success
[info]     - Invokes cleanups
[info]       List(1, 2, 2, 1) did not satisfy equalTo(List(1, 1, 2, 1))
[info]     + Invokes cleanups on interrupt - 1
[info]     + Invokes cleanups on interrupt - 2
[info]     + Invokes cleanups on interrupt - 3
[info]   + foreach
[info]     + Returns elements in the correct order
[info]     + Runs finalizers
[info]     + Invokes cleanups in reverse order of acquisition
[info]   + foreachPar
[info]     + Returns elements in the correct order
[info]     + Runs finalizers
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]   + foreachParN
[info]     + Returns elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]   + foreach_
[info]     + Runs finalizers
[info]   + foreachPar_
[info]     + Runs finalizers
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]   + foreachParN_
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]   + fork
[info]     + Runs finalizers properly
[info]     + Acquires interruptibly
[info]   + fromAutoCloseable
[info]     + Runs finalizers properly
[info]   + mergeAll
[info]     + Merges elements in the correct order
[info]     + Runs finalizers
[info]   + mergeAllPar
[info]     + Merges elements in the correct order
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]     + Runs finalizers
[info]   - mergeAllParN
[info]     + Merges elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]     - All finalizers run even when finalizers have defects
[info]       4 did not satisfy equalTo(3)
[info]   + onExit
[info]     + Calls the cleanup
[info]   + onExitFirst
[info]     + Calls the cleanup
[info]   + reduceAll
[info]     + Reduces elements in the correct order
[info]     + Runs finalizers
[info]   + reduceAllPar
[info]     + Reduces elements in the correct order
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]     + Runs finalizers
[info]   + reduceAllParN
[info]     + Reduces elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]     + All finalizers run even when finalizers have defects
[info]   + retry
[info]     + Should retry the reservation
[info]     + Should retry the acquisition
[info]     + Should share retries between both
[info]   + timed
[info]     + Should time both the reservation and the acquisition
[info]   + timeout
[info]     + Returns Some if the timeout isn't reached
[info]     + Returns None if the reservation takes longer than d
[info]     + Returns None if the acquisition takes longer than d
[info]     + Runs finalizers if returning None and reservation is successful
[info]     + Runs finalizers if returning None and reservation is successful after timeout
[info]   + withEarlyRelease
[info]     + Provides a canceler that can be used to eagerly evaluate the finalizer
[info]     + The canceler should run uninterruptibly
[info]     + If completed, the canceler should cause the regular finalizer to not run
[info]     + The canceler will run with an exit value indicating the effect was interrupted
[info]   + withEarlyReleaseExit
[info]     + Allows specifying an exit value
[info]   + zipPar
[info]     + Does not swallow exit cause if one reservation fails
[info]     + Runs finalizers if one acquisition fails
[info]     + Does not swallow acquisition if one acquisition fails
[info]     + Run finalizers if one reservation fails
[info]   + flatten
[info]     + Returns the same as ZManaged.flatten
[info]   + absolve
[info]     + Returns the same as ZManaged.absolve
[info] Ran 85 tests in 6 s 292 ms: 82 succeeded, 0 ignored, 3 failed
[info] ZIO Test
[info] Done
[error] Failed: Total 107, Failed 6, Errors 0, Passed 101
[error] Failed tests:
[error] 	zio.FunctionIOSpec
[error] 	zio.ZManagedSpec
[error] (coreTestsJVM / Test / testQuick) sbt.TestsFailedException: Tests unsuccessful
```

Example output after changes from this PR (all failures are reported at the end of output):
```
[info] - FunctionIOSpec
[info]   - Check if the functions in `FunctionIO` work correctly
[info]     + `fromFunction` lifts from A => B into effectful function
[info]     + `identity` returns the identity of the input without modification
[info]     + `>>>` is a symbolic operator of `andThen`which does a Backwards composition of effectful functions
[info]     - `<<<` is a symbolic operator of `compose` which compose two effectful functions
[info]       13 did not satisfy equalTo(14)
[info]     + `zipWith` zips the output of two effectful functions
[info]     + `&&&` zips the output of two effectful functions and returns a tuple of their result
[info]     + `|||` computes two effectful functions left and right from from an Either input
[info]     + `first` returns a tuple: the output on the first element and input on the second element
[info]     + `second` returns a tuple: the input on the first element and output on the second element
[info]     + `left` takes an Either as input and computes it if it is Left otherwise returns the same value of the input
[info]     + `right`takes an Either as input and computes it if it is Right otherwise returns the same value of the input
[info]     + `asEffect` returns the input value
[info]     - `test` check a condition and returns an Either output: Left if the condition is true otherwise false
[info]       Right(List(1, 2, 5)) did not satisfy isRight(equalTo(List(1, 5)))
[info]       List(1, 2, 5) did not satisfy equalTo(List(1, 5))
[info]     + `ifThenElse`
[info]       + check an impure condition if it is true then computes an effectful function `then0` else computes `else0`
[info]       + check a pure condition if it is true then computes an effectful function `then0` else computes `else0`
[info]     + `whileDo`
[info]       + take a condition and run the body until the condition will be  false with impure function
[info]       + take a condition and run the body until the condition will be  false with pure function
[info]     + `_1` extracts out the first element of a tuple
[info]     + `_2` extracts out the second element of a tuple
[info]     - `fail` returns a failure
[info]       Left(error) did not satisfy isLeft(equalTo(errr))
[info]       error did not satisfy equalTo(errr)
[info]     + `effect` can translate an Exception to an error
[info]     + `ignore` ignores a effect failure
[info] Ran 22 tests in 618 ms: 19 succeeded, 0 ignored, 3 failed
[info] - ZManaged
[info]   + make
[info]     + Invokes cleanups in reverse order of acquisition.
[info]     + Properly performs parallel acquire and release
[info]     + Constructs an uninterruptible Managed value
[info]   + makeEffect
[info]     + Invokes cleanups in reverse order of acquisition.
[info]   + reserve
[info]     + Interruption is possible when using this form
[info]   + makeExit
[info]     + Invokes with the failure of the use
[info]     + Invokes with the failure of the subsequent acquire
[info]   + ensuring
[info]     + Runs on successes
[info]     + Runs on failures
[info]     + Works when finalizers have defects
[info]   + ensuringFirst
[info]     + Runs on successes
[info]     + Runs on failures
[info]     + Works when finalizers have defects
[info]   + flatMap
[info]     + All finalizers run even when finalizers have defects
[info]   - foldM
[info]     - Runs onFailure on failure
[info]       List(1, 1) did not satisfy equalTo(List(1))
[info]     + Runs onSuccess on success
[info]     - Invokes cleanups
[info]       List(1, 2, 2, 1) did not satisfy equalTo(List(1, 1, 2, 1))
[info]     + Invokes cleanups on interrupt - 1
[info]     + Invokes cleanups on interrupt - 2
[info]     + Invokes cleanups on interrupt - 3
[info]   + foreach
[info]     + Returns elements in the correct order
[info]     + Runs finalizers
[info]     + Invokes cleanups in reverse order of acquisition
[info]   + foreachPar
[info]     + Returns elements in the correct order
[info]     + Runs finalizers
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]   + foreachParN
[info]     + Returns elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]   + foreach_
[info]     + Runs finalizers
[info]   + foreachPar_
[info]     + Runs finalizers
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]   + foreachParN_
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]   + fork
[info]     + Runs finalizers properly
[info]     + Acquires interruptibly
[info]   + fromAutoCloseable
[info]     + Runs finalizers properly
[info]   + mergeAll
[info]     + Merges elements in the correct order
[info]     + Runs finalizers
[info]   + mergeAllPar
[info]     + Merges elements in the correct order
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]     + Runs finalizers
[info]   - mergeAllParN
[info]     + Merges elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]     - All finalizers run even when finalizers have defects
[info]       4 did not satisfy equalTo(3)
[info]   + onExit
[info]     + Calls the cleanup
[info]   + onExitFirst
[info]     + Calls the cleanup
[info]   + reduceAll
[info]     + Reduces elements in the correct order
[info]     + Runs finalizers
[info]   + reduceAllPar
[info]     + Reduces elements in the correct order
[info]     + Runs reservations in parallel
[info]     + Runs acquisitions in parallel
[info]     + Runs finalizers
[info]   + reduceAllParN
[info]     + Reduces elements in the correct order
[info]     + Uses at most n fibers for reservation
[info]     + Uses at most n fibers for acquisition
[info]     + Runs finalizers
[info]     + All finalizers run even when finalizers have defects
[info]   + retry
[info]     + Should retry the reservation
[info]     + Should retry the acquisition
[info]     + Should share retries between both
[info]   + timed
[info]     + Should time both the reservation and the acquisition
[info]   + timeout
[info]     + Returns Some if the timeout isn't reached
[info]     + Returns None if the reservation takes longer than d
[info]     + Returns None if the acquisition takes longer than d
[info]     + Runs finalizers if returning None and reservation is successful
[info]     + Runs finalizers if returning None and reservation is successful after timeout
[info]   + withEarlyRelease
[info]     + Provides a canceler that can be used to eagerly evaluate the finalizer
[info]     + The canceler should run uninterruptibly
[info]     + If completed, the canceler should cause the regular finalizer to not run
[info]     + The canceler will run with an exit value indicating the effect was interrupted
[info]   + withEarlyReleaseExit
[info]     + Allows specifying an exit value
[info]   + zipPar
[info]     + Does not swallow exit cause if one reservation fails
[info]     + Runs finalizers if one acquisition fails
[info]     + Does not swallow acquisition if one acquisition fails
[info]     + Run finalizers if one reservation fails
[info]   + flatten
[info]     + Returns the same as ZManaged.flatten
[info]   + absolve
[info]     + Returns the same as ZManaged.absolve
[info] Ran 85 tests in 5 s 884 ms: 82 succeeded, 0 ignored, 3 failed
[info] ZIO Test
[info] - ZManaged
[info]   - foldM
[info]     - Runs onFailure on failure
[info]       List(1, 1) did not satisfy equalTo(List(1))
[info]     - Invokes cleanups
[info]       List(1, 2, 2, 1) did not satisfy equalTo(List(1, 1, 2, 1))
[info]   - mergeAllParN
[info]     - All finalizers run even when finalizers have defects
[info]       4 did not satisfy equalTo(3)
[info] - FunctionIOSpec
[info]   - Check if the functions in `FunctionIO` work correctly
[info]     - `<<<` is a symbolic operator of `compose` which compose two effectful functions
[info]       13 did not satisfy equalTo(14)
[info]     - `test` check a condition and returns an Either output: Left if the condition is true otherwise false
[info]       Right(List(1, 2, 5)) did not satisfy isRight(equalTo(List(1, 5)))
[info]       List(1, 2, 5) did not satisfy equalTo(List(1, 5))
[info]     - `fail` returns a failure
[info]       Left(error) did not satisfy isLeft(equalTo(errr))
[info]       error did not satisfy equalTo(errr)
[info] Done
[error] Failed: Total 107, Failed 6, Errors 0, Passed 101
[error] Failed tests:
[error] 	zio.FunctionIOSpec
[error] 	zio.ZManagedSpec
[error] (coreTestsJVM / Test / testQuick) sbt.TestsFailedException: Tests unsuccessful
```


test/jvm/src/main/scala/zio/test/RunnableSpec.scala
```diff
@@ -37,8 +38,11 @@ abstract class RunnableSpec[L, T, E, S](runner0: TestRunner[L, T, E, S])(spec0:
    * TODO: Parse command line options.
    */
   final def main(args: Array[String]): Unit = {
-    val results     = runner.unsafeRun(spec)
-    val hasFailures = results.exists { case TestCase(_, test) => test.isLeft; case _ => false }
+    val ExecutionResult(results, summary) = runner.unsafeRun(spec)
+    val hasFailures                       = results.exists { case TestCase(_, test) => test.isLeft; case _ => false }
+
+    println(summary)
```
adamgfraser: Can we use the `TestLogger` in the `TestRunner` here?

kapke: Fixed. Shouldn't this be done similarly for JS?

kapke: I just noticed that `RunnableSpec` differs between JS and JVM only by:
  - the way tests are run (sync vs async)
  - lack of reeporting and `sys.exit` on JS side.

Wouldn't it work to move `RunnableSpec` into `shared` and run differently basing on `TestPlatform`?

adamgfraser: Yes, that is a great idea. We introduced `TestPlatform` after we created `RunnableSpec` which is why we had to create two separate files but if we can integrate into one file that will definitely make things more maintainable in the future.

kapke:
I decided to run `RunnableSpec` async on both JS and JVM. It seems working properly. Is there any reason to make it synchronous back?

Also I've fixed `EnvironmentSpec` `nextInt` as it was failing for me when run on Node.js because of fresh mock being provided for both values

adamgfraser:
@kapke Will do a full review tonight but regarding your question about synchronous versus asynchronous, the tricky part is we are trying to obtain the following behavior, across both platforms:

1. If any test fails, running the test should result in a failure (e.g. non-zero exit code on the JVM) so that things like CI tools will fail the build.
2. If the user runs tests within SBT and there is a failure, this should not cause the SBT session to exit.

My recollection is that we needed to wait for the result on the JVM and then exit to get the appropriate semantics, but if you can get that behavior with staying asynchronous all the better.

test/shared/src/test/scala/zio/test/mock/EnvironmentSpec.scala
```diff
@@ -52,10 +52,10 @@ object EnvironmentSpec extends DefaultRuntime {
 
   def nextInt =
```
adamgfraser: This test is actually testing for something important so if it isn't passing we need to understand why. When the `MockEnvironment` is created the `MockRandom` is supposed to be seeded with the current live system time (see [here](https://github.com/zio/zio/blob/fd8311581928cbb15ef7275b5364be3a15e9c6f3/test/js/src/main/scala/zio/test/mock/MockEnvironment.scala#L49)) so that when the user runs tests different times they are testing more of the full random "space" instead of just always testing with the same values since the seed is the same. The idea of this test is that if we create two different `MockEnvironment` instances the system time should be at least slightly different so they should have different seeds and produce different values, which has works on JVM and at least the version of JS we have been testing on. Is there something different about node that is causing this not to work? Does the system time always return 0 or something?

kapke: Due to this: https://github.com/scala-js/scala-js/issues/3623 Scala.js implementation of `System.nanoTime` is limited to stable JS capabilites which `Date` resolution is 1ms. Does it make sense to add 1 ms sleep between these `nextInt` calls then?

kapke:
Does it work for you to run JS tests the same way you've described above (sbt shell -> project -> test:run -> select main) on master? I didn't for me - should I open an issue with this? Any easy/obvious solution didn't work either.

JVM tests run using `main` - fixed

test/shared/src/test/scala/zio/test/mock/EnvironmentSpec.scala
adamgfraser: Interesting. I think this is something we may have to think about more. We test everything on Scalaj.js but not with the Node.js environment, which is why this test is passing in CI but not for you locally. For now yes adding a 1 ms delay makes sense.

adamgfraser:
Awesome regarding JVM tests.

JS tests work with some caveats. Under ScalaJS you have to have `scalaJSUseMainModuleInitializer in Test := true` in your SBT settings for the project, which we currently only have for ZIO Test. Also from what I can tell ScalaJS only supports one `main` method in a project versus being able to select from multiple `main` methods on JVM. But for example if I put the following into the `test` folder of ZIO Test and comment out the existing `TestMain` folder so there is only one method, then this will run with `sbt`, `project testJS`, `test:run`:

```scala
package zio.test

object MySpec extends DefaultRunnableSpec(
  suite("MySuite")(
    test("MyTest") {
      assert(true, Assertion.isTrue)
    }
  )
)
```

```
sbt:zio-test> test:run
[info] Compiling 2 Scala sources to /Users/adamfraser/zio/test/js/target/scala-2.12/test-classes ...
[info] Fast optimizing /Users/adamfraser/zio/test/js/target/scala-2.12/zio-test-test-fastopt.js
[info] Running zio.test.MySpec
+ MySuite
  + MyTest
Ran 1 test in 31 ms: 1 succeeded, 0 ignored, 0 failed
[success] Total time: 25 s, completed Sep 25, 2019 10:29:58 PM
```

kapke:
I managed to make it work with sbt test runner for JS too.

adamgfraser:
Nice!

test/shared/src/test/scala/zio/test/mock/EnvironmentSpec.scala
```diff
@@ -54,6 +54,7 @@ object EnvironmentSpec extends DefaultRuntime {
     unsafeRunToFuture {
       for {
         i <- random.nextInt.provideManaged(MockEnvironment.Value)
+        _ <- Live.live(clock.sleep(1.millisecond)).provideManaged(MockEnvironment.Value)
```
adamgfraser: This can just be `clock.sleep(1.millisecond)`. We're not providing the `MockEnvironment` here so the real clock will be used automatically and the default runtime will provide this.

test/shared/src/main/scala/zio/test/TestRunner.scala
```diff
@@ -71,17 +72,27 @@ case class TestRunner[L, -T, E, S](
...
+    Runtime(buildEnv(loggerSvc, clockSvc), platform)
+
+  private def buildEnv(loggerSvc: TestLogger, clockSvc: Clock): TestLogger with Clock =
+    new TestLogger with Clock {
+      override def testLogger: TestLogger.Service = loggerSvc.testLogger
+      override val clock: Clock.Service[Any]      = clockSvc.clock
+    }
+}
+object TestRunner {
+  case class ExecutionResult[L, E, S](spec: ExecutedSpec[L, E, S], summary: String)
```
adamgfraser: Can we eliminate this and make it solely the responsibility of the `DefaultTestReporter` and the SBT test runner to filter for failing tests, convert them into string representations, aggregate them, and display them? I think this would give us a cleaner separation between "data" and "presentation of data". An `ExecutedSpec` should, I believe, have all of the underlying data on which tests passed and failed and the associated labels. Then it is up to the client to interpret those results however they want, including creating and displaying a summary like this.

The summary is very useful and we definitely want to include it in the default way we report test results but I think we want to avoid baking it into the return type for running tests to preserve flexibility for other clients to interpret the test results in different ways.

kapke: Of course. This makes a lot of sense. I should commit this change during this weekend.

adamgfraser:
Can you do a rebase?

PR #1792, opened 2019-09-22
Fix #785, Fix #1441, prevent race condition on `interrupted` variable by moving `interrupted` into atomic state

### LINKED ISSUE #785: Interrupt leak / deadlock bug under high concurrency
As of **v1.0-RC4**, there's a possible non-deterministic interruption leak / deadlock bug under high concurrency. I've started seeing interrupt IOs that permanently block once every thousand calls in the context of a large working app.

I've managed to create a bare minimum, standalone reproducer:

```scala
import java.util.concurrent.atomic.LongAdder
import scalaz.zio.console._
import scalaz.zio.{Task, UIO, ZIO, ZSchedule}
import scalaz.zio.duration._

object ZioInterruptLeakOrDeadlockRepo extends scalaz.zio.App {
    val work = UIO.succeed("whatever")

    // Simulates an external "kill-switch"
    val abort = Task.unit.delay(1.minute) 

    val completedCounter = new LongAdder
    val pendingGauge = new LongAdder
    val timeoutCounter = new LongAdder

    val leakOrDeadlockTest = for {
      _ <- UIO(pendingGauge.increment())
      valueFib <- work.fork
      abortFib <- (abort *> valueFib.interrupt).fork
      _ <- valueFib.join

      // This blocks / leaks once every 20,000 rounds or so
      _ <- abortFib.interrupt 

      _ <- UIO(pendingGauge.decrement())
    } yield ()

    val main = for {
      _ <- UIO(s"completed=${completedCounter.longValue()} pending=${pendingGauge.longValue()} timed-out=${timeoutCounter.longValue()}")
        .flatMap(putStrLn)
        .repeat(ZSchedule.fixed(1.second))
        .fork

      _ <- ZIO
        .foreachPar(1 to Runtime.getRuntime.availableProcessors()) { _ =>
          leakOrDeadlockTest
            .timeout(5.seconds)
            .repeat(ZSchedule.identity[Option[Unit]].logInput {
              case Some(_) => UIO(completedCounter.increment())
              case None => UIO(timeoutCounter.increment())
            })
        }
    } yield 0

    main
      .catchAll(e => UIO(println(s"Died with $e")) *> UIO.succeed(1))
  }
}
```

**Expected:** 

- Steady, relatively constant pending count, ~= the number of system threads.
- No task timeouts (exceeds 5 seconds in this test)

**Actual:**

- Pending task count gradually increases, so as the number of timed-out tasks.

It's worth noting that by reducing the CPU load, for example with:

```
val work = UIO.succeed("whatever").delay(1.milli)
```

this issue would not be visible. So far I've only been able to reproduce this under high CPU utilization, which might indicate it's a scheduling fairness problem?

### LINKED ISSUE #1441: Performance problem and abrupt killed threads in combination with FS2-kafka
Hi everyone,

We're currently trying out ZIO in a new application that we're building here at DHL and have really been loving it.

However, we're currently experiencing a weird issue using ZIO in combination with FS2-kafka. In our case we're producing records to kafka and have noticed that sometimes the streaming just stops. When this happens, the ZIO async threadpool seems to disappear.

We've then tried using cats IO and have noticed this doesn't happen, and at the same time it seems to be a lot faster. We're not really very experienced in the internals of both zio or cats effect, so were wondering if you could help us figure out why this happens.

We've created a reproduction where we just stream the same record over and over to kafka. In the same time the implementation with cats IO is about 3 to 4x faster (see screenshot, 675k records for cats vs 182k). Maybe you will also notice that at some point the threads get killed in case of ZIO (see screenshot), but it could take a while before this happens (but usually within 1 million records or so).

We've included instructions in the readme:
https://github.com/dhlparcel/zio-kafka-bug

Thanks in advance!

*Screenshot of differences between ZIO and IO in same time*
image (1)

*Threads that suddenly get killed*
image (2)

neko-kai:
Also, prevent potential volatility problems with accessing interruptStack from multiple threads by moving `interruptible` state into Suspended

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -413,7 +413,7 @@ private[zio] final class FiberContext[E, A](
                   val zio = curZio.asInstanceOf[ZIO.EffectAsync[Any, E, Any]]
 
                   // Enter suspended state:
-                  curZio = if (enterAsync(interruptible)) {
+                  curZio = if (enterAsync()) {
                     val k = zio.register
 
                     if (traceEffects && inTracingRegion) addTrace(k)
@@ -653,15 +653,19 @@ private[zio] final class FiberContext[E, A](
     }
 
   @tailrec
-  private[this] final def enterAsync(interruptible: Boolean): Boolean = {
+  private[this] final def enterAsync(): Boolean = {
     val oldState = state.get
 
     oldState match {
-      case Executing(_, observers, false) =>
-        val newState = Executing(FiberStatus.Suspended(interruptible), observers, false)
+      case Executing(_, observers, interrupt) =>
+        val newState = Executing(FiberStatus.Suspended(interruptible), observers, interrupt)
 
-        if (!state.compareAndSet(oldState, newState)) enterAsync(interruptible)
-        else true
+        if (!state.compareAndSet(oldState, newState)) enterAsync()
+        else if (shouldInterrupt) {
+          // Fiber interrupted, so go back into running state:
+          exitAsync()
+          false
+        } else true
 
       case _ => false
     }
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -145,7 +145,7 @@ private[zio] final class FiberContext[E, A](
       case v    => k(v)
     }
 
-  private object InterruptExit extends Function[Any, IO[E, Any]] {
+  private[this] object InterruptExit extends Function[Any, IO[E, Any]] {
     final def apply(v: Any): IO[E, Any] = {
       val isInterruptible = interruptStatus.peekOrElse(true)
 
@@ -159,7 +159,7 @@ private[zio] final class FiberContext[E, A](
     }
   }
 
-  private object TracingRegionExit extends Function[Any, IO[E, Any]] {
+  private[this] object TracingRegionExit extends Function[Any, IO[E, Any]] {
     final def apply(v: Any): IO[E, Any] = {
       // don't use effectTotal to avoid TracingRegionExit appearing in execution trace twice with traceEffects=true
       tracingStatus.popDrop(())
@@ -172,7 +172,7 @@ private[zio] final class FiberContext[E, A](
    * Unwinds the stack, looking for the first error handler, and exiting
    * interruptible / uninterruptible regions.
    */
-  final def unwindStack(): Unit = {
+  private[this] final def unwindStack(): Unit = {
     var unwinding = true
 
     // Unwind the stack, looking for an error handler:
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -46,9 +46,10 @@ private[zio] final class FiberContext[E, A](
   import FiberState._
 
   // Accessed from multiple threads:
-  private[this] val state = new AtomicReference[FiberState[E, A]](FiberState.Initial[E, A])
+  private[this] val state = new AtomicReference[FiberState[E, A]](FiberState.initial)
 
-  @volatile private[this] var interrupted = false
+  @noinline
+  private[this] def interrupted = state.get.interrupt
 
   private[this] val traceExec: Boolean =
     PlatformConstants.tracingSupported && platform.tracing.tracingConfig.traceExecution
@@ -412,7 +413,7 @@ private[zio] final class FiberContext[E, A](
                   val zio = curZio.asInstanceOf[ZIO.EffectAsync[Any, E, Any]]
 
                   // Enter suspended state:
-                  curZio = if (enterAsync()) {
+                  curZio = if (enterAsync(interruptible)) {
                     val k = zio.register
 
                     if (traceEffects && inTracingRegion) addTrace(k)
@@ -526,7 +527,7 @@ private[zio] final class FiberContext[E, A](
         }
       } catch {
         case _: InterruptedException =>
-          Thread.interrupted
+          Thread.interrupted()
           curZio = ZIO.interrupt
 
         // Catastrophic error handler. Any error thrown inside the interpreter is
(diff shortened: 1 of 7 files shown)


neko-kai:
A clean reproduction for testing:

```scala
import java.util.concurrent.atomic.{AtomicLong, LongAdder}

import zio.{UIO, _}

import scala.language.reflectiveCalls

object ZioInterruptLeakOrDeadlockRepo extends zio.App {

  val leakedCounter = new AtomicLong(0L)

  val startedCounter = new LongAdder
  val completedCounter = new LongAdder
  val awakeCounter = new LongAdder
  val pendingGauge = new LongAdder

    val leakOrDeadlockTest = for {
      _ <- UIO {
        startedCounter.increment()
        pendingGauge.increment()
      }

      sleepInterruptFiber <- IO.never.fork

      // This blocks / leaks once every 20,000 rounds or so
      _ <- sleepInterruptFiber.interrupt
        .ensuring(UIO {
          awakeCounter.increment()
          pendingGauge.decrement()
        })

      _ <- UIO(completedCounter.increment())
    } yield ()

    val main = for {
      _ <-
        ZIO.runtime[Any].map(_.Platform.executor.metrics.get)
          .flatMap {
            metrics => UIO(new Thread(() => {
              while (true) {
                println(s"started=${startedCounter.longValue()} awake=${awakeCounter.longValue()} completed=${completedCounter.longValue()} pending=${pendingGauge.longValue()} queued=${metrics.size}")
                Thread.sleep(1000L)
              }
            }).start())
          }

      _ <- leakOrDeadlockTest.forever
    } yield 0

    main
  }
}

```

regiskuckaertz:
wow, you found it! Awesome detective work 🕵

neko-kai:
Race scenario:
```
fiber1: kill0 sets from Running to Running, 
fiber2: BEFORE interruptible=false, but AFTER atomic update, sets state=Suspended and exits.
fiber1: Exits because it thinks fiber2 is Running from previous state
```
@regiskuckaertz 
Seems John also found it at the same time, you could say it was a race :D Let's wait for John's feedback first!

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -675,9 +676,8 @@ private[zio] final class FiberContext[E, A](
     val oldState = state.get
 
     oldState match {
-      case Executing(FiberStatus.Suspended, observers) =>
-        if (!state.compareAndSet(oldState, Executing(FiberStatus.Running, observers)))
-          exitAsync()
+      case Executing(_: FiberStatus.Suspended, observers, interrupt) =>
```
jdegoes: How about `FiberStatus.Suspended(_)`, to avoid the appearance of type casing?

neko-kai: I'm not sure if that would result in a call to .unapply, avoiding that is the only reason I used a type match here, will check...

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -743,28 +742,23 @@ private[zio] final class FiberContext[E, A](
...
         else {
-          interrupted = true
-
           evaluateLater(ZIO.interrupt)
-
           None
         }
 
-      case Executing(status, observers0) =>
+      case Executing(status, observers0, _) =>
```
jdegoes: This extra flag is necessary in your fix. In my fix the model changed a bit, so pushing the "set" prior to the atomic ref update should in theory take care of it. But this looks good for now!

jdegoes:
@neko-kai Can you add the test?

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Ah, performance is always a reason to break rules in ZIO. 😆

neko-kai:
@jdegoes I've added a test that just runs IO.fork.interrupt 10K times, multiplied by test builds it should probably be able to catch a regression over many builds.

I've also added a benchmark that does the same IO.fork.interrupt and should definitely start hanging on regression. I've also added a benchmark for `race` that highlights the perf issue in #1441 - our `race` is quite a bit slower 😱 
```
[info] Benchmark                                (size)   Mode  Cnt    Score     Error  Units
[info] IOEmptyRaceBenchmark.catsEmptyRace         1000  thrpt   10  315.196 ± 398.678  ops/s
[info] IOEmptyRaceBenchmark.monixEmptyRace        1000  thrpt   10  192.894 ±   7.672  ops/s
[info] IOEmptyRaceBenchmark.zioEmptyRace          1000  thrpt   10   33.152 ±   0.572  ops/s
[info] IOEmptyRaceBenchmark.zioTracedEmptyRace    1000  thrpt   10   25.166 ±   1.548  ops/s
```

core/shared/src/main/scala/zio/internal/FiberContext.scala
neko-kai: Hmm, I'm wrong! Scala is smart enough to avoid `unapply` not only here, but everywhere in `FiberContext` there are no unapply calls:
```java
if (var6 instanceof FiberContext$FiberStatus$Suspended) {
   if (!this.state.compareAndSet(oldState, new FiberContext$FiberState$Executing(FiberContext$FiberStatus$Running$.MODULE$, observers, interrupt))) {
      continue;
   }

   var2 = true;
   break;
}
```

neko-kai:
I believe another inconsistency with `exitAsync()` can happen as witnessed by the following code:
```scala
IO.effectAsyncMaybe[Nothing, Unit] {
  k =>
    Future {
      Thread.sleep(200L)
      k(IO.dieMessage("INCOHERENT CONTINUE"))
    }
    Some(IO.unit)
}
  .flatMap(_ => ZIO.never)
  .ensuring(IO.dieMessage("CAN'T HAPPEN!"))
  .uninterruptible
```

that "overtakes" the fiber from an async that was "cancelled", but only if the fiber enters suspension at an opportune time

neko-kai:
There's also a remaining race with `evaluateLater(ZIO.interrupt)` in `kill0`, illustration:
```scala
object ZioIncoherentInterruptAsyncRepro extends zio.App {

    val test =
      ZIO.effectAsyncMaybe[Environment, Nothing, Unit] {
        k =>
          Future {
            Thread.sleep(200L)
          }
          Some(IO.unit)
      }
        .flatMap(_ => ZIO.sleep(500.millis))
        .ensuring(console.putStrLn("Async finalizer!") *> ZIO.never)

    test.fork.flatMap {
      f =>
        f.interrupt.delay(50.millis)
    }.as(0)
  }
}
```

Basically, every single point where multi-threaded access can happen in non-atomic or non-guarded way is a problem (:
Luckily, we only really have async + interrupt as these points.

neko-kai:
Will address these two - https://github.com/zio/zio/pull/1792#issuecomment-535262826, https://github.com/zio/zio/pull/1792#issuecomment-535266947 - separately.

iravid:
@neko-kai Is the performance of race improved after these fixes?

neko-kai:
@iravid 
Nope, I just added a benchmark that showcases the bad perf. Will open an issue about race perf too!

iravid:
Ah too bad. Ok :-)

neko-kai:
@iravid @jdegoes 
Race perf moved here – https://github.com/zio/zio/issues/1844
More async races - https://github.com/zio/zio/issues/1842, https://github.com/zio/zio/issues/1843

Issue #1732, opened 2019-09-21
Add ZStream#chunkN


iravid:
A method for chunking elements into Chunks :-)

iravid:
@LukaszByczynski

PR #1795, opened 2019-09-22
Add ChunkN to ZStream

LukaszByczynski:
Adds a small helper method to provide user friendly way to chunk the input stream into batches.

Resolves: #1732

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -664,7 +664,8 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
     }
   }
 
-  /*** Chunks the stream with specifed chunkSize
+  /**
+   * Chunks the stream with specifed chunkSize
    * @parm chunkSize size of the chunk
    */
   def chunkN(chunkSize: Long): ZStream[R, E, Chunk[A]] =
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -666,7 +666,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
 
   /**
    * Chunks the stream with specifed chunkSize
-   * @parm chunkSize size of the chunk
+   * @param chunkSize size of the chunk
    */
   def chunkN(chunkSize: Long): ZStream[R, E, Chunk[A]] =
     transduce(ZSink.collectAllN[A](chunkSize).map(Chunk.fromIterable))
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -664,6 +664,12 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
     }
   }
 
+  /*** Chunks the stream with specifed chunkSize
+   * @parm chunkSize size of the chunk
+   */
+  def chunkN(chunkSize: Long): ZStream[R, E, List[A]] =
+    transduce(ZSink.collectAllN[A](chunkSize))
```
iravid: How about:
```scala
def grouped(size: Long): ZStream[R,  E, List[A]]
def chunkN(size: Long): ZStream[R, E, Chunk[A]]
```

?

iravid: `ZStream#grouped` being the same as `List#grouped`

LukaszByczynski: Yeah, I did'nt know that we have a Chunk (homework for me).

jdegoes:
@LukaszByczynski @iravid How is this one looking?

streams/shared/src/main/scala/zio/stream/ZStream.scala
LukaszByczynski: @iravid I have question. How do I should implement this new `ChunkN` version? Is there `ZSink` operator for the Chunk result  ? Maybe, I should imlement one for this case (based on how ZSkink does that for the List)?

iravid: Something like `ZSink.collectAllN[A](chunkSize).map(Chunk.fromIterable)`

iravid: There's a much more efficient implementation that uses a pre-allocated buffer but we can do that in a follow-up if someone cares about the performance

LukaszByczynski: Ok. Tomorrow I will push changes for a review

LukaszByczynski:
@iravid it's ready for a review

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -664,6 +664,12 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
     }
   }
 
+  /*** Chunks the stream with specifed chunkSize
```
```suggestion
  /**
   * Chunks the stream with specifed chunkSize
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1268,6 +1274,12 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
   ): GroupBy[R1, E1, K, A] =
     self.groupBy(a => ZIO.succeed((f(a), a)), buffer)
 
+  /*** Partitions the stream with specifed chunkSize
```
```suggestion
  /**
   * Partitions the stream with specifed chunkSize
```

iravid:
@LukaszByczynski Could you check the mdoc build failure?

LukaszByczynski:
Fixed

iravid:
Thank you @LukaszByczynski!

PR #1799, opened 2019-09-22
Removed explicit type parameter when invoking failure

jaliss:
@adamgfraser I changed the spec a bit so we don't need to pass the type parameter for `failure` explicitly.  I think it reads better but let me know what you think.

### DIFF
--- a/test/shared/src/test/scala/zio/test/TestAspectSpec.scala
+++ b/test/shared/src/test/scala/zio/test/TestAspectSpec.scala
@@ -82,7 +82,7 @@ object TestAspectSpec extends DefaultRuntime {
   def failureMakesTestsPassOnSpecifiedException: Future[Boolean] =
     unsafeRunToFuture {
       val spec = test("test")(assert(throw new NullPointerException(), isFalse)) @@ failure(
-        exceptionAssertion[NullPointerException]
+        failsWithException[NullPointerException]
       )
       succeeded(spec)
     }
@@ -91,7 +91,7 @@ object TestAspectSpec extends DefaultRuntime {
     unsafeRunToFuture {
       val spec = test("test")(
         assert(throw new NullPointerException(), isFalse)
-      ) @@ failure(exceptionAssertion[IllegalArgumentException])
+      ) @@ failure(failsWithException[IllegalArgumentException])
       failed(spec)
     }
 
@@ -114,7 +114,7 @@ object TestAspectSpec extends DefaultRuntime {
   def failureMakesTestsPassOnExpectedAssertionFailure: Future[Boolean] =
     unsafeRunToFuture {
       val spec = test("test")(assert(true, equalTo(false))) @@ failure(
-        isCase[TestFailure[Any], TestResult]("Assertion", {
+        isCase[TestFailure[Any], Any]("Assertion", {
           case TestFailure.Assertion(result) => Some(result); case _ => None
         }, anything)
       )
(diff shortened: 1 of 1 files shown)


test/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -117,28 +114,28 @@ object TestAspectSpec extends DefaultRuntime {
...
+        equalTo(assert(42, equalTo(42)))
+      )
+      val spec = test("failureDoesNotMakesTestsPassOnUnexpectedAssertionFailure")(assert(true, equalTo(false))) @@ failure(
+        assertion
+      )
       failed(spec)
     }
 
-  private def failsWithException[T <: Throwable](implicit ct: ClassTag[T]): Assertion[Any] =
+  private def exceptionAssertion[T <: Throwable](implicit ct: ClassTag[T]): Assertion[Any] =
```
adamgfraser: Should be `Assertion[TestFailure[Any]]`.

test/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -75,70 +75,64 @@ object TestAspectSpec extends DefaultRuntime {
...
 
   def failureMakesTestsPassOnExpectedAssertionFailure: Future[Boolean] =
     unsafeRunToFuture {
-      val spec = test("failureMakesTestsPassOnExpectedAssertionFailure")(assert(true, equalTo(false))) @@ failure[
-        Nothing
-      ] {
-        isCase("Assertion", { case TestFailure.Assertion(result) => Some(result); case _ => None }, anything)
-      }
+      val spec = test("test")(assert(true, equalTo(false))) @@ failure(
+        isCase[TestFailure[Any], TestResult]("Assertion", {
```
adamgfraser: Second type parameter can be `Any`.

test/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -75,70 +75,64 @@ object TestAspectSpec extends DefaultRuntime {
...
           "Assertion", { case TestFailure.Assertion(result) => Some(result); case _ => None },
           equalTo(assert(42, equalTo(42)))
         )
-      }
+      )
       failed(spec)
     }
 
-  private def failsWithException[T <: Throwable](implicit ct: ClassTag[T]): Assertion[Any] =
+  private def exceptionAssertion[T <: Throwable](implicit ct: ClassTag[T]): Assertion[TestFailure[Any]] =
```
adamgfraser: For what it's worth I thought your original name of `failsWithException` was actually clearer here.

test/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -124,15 +124,15 @@ object TestAspectSpec extends DefaultRuntime {
   def failureDoesNotMakesTestsPassOnUnexpectedAssertionFailure: Future[Boolean] =
     unsafeRunToFuture {
       val spec = test("test")(assert(true, equalTo(false))) @@ failure(
-        isCase[TestFailure[Boolean], TestResult](
```
adamgfraser: This one should be `TestResult` because we want the type safety of making sure the thing we're checking for equality against is also a `TestResult`, versus the one above the assertion was `anything` which just takes an `Any`. Otherwise all good.

PR #1800, opened 2019-09-23
Update testing.md

almendar:
Added information about Mocking environments


almendar:
I'll add sections on other Mock environments e.g. Console

adamgfraser:
Great. I would also take a look at the existing Scaladoc in the `mock` package. That is probably the area we have best documented in ZIO Test so we should be sure we are consistent and leveraging the content there when possible.

docs/usecases/testing.md
```diff
@@ -48,6 +48,68 @@ In **ZIO Test**, all tests are immutable values and tests are tightly integrated
 
 The library also includes built-in mock versions of all the standard ZIO environmental effects (`Clock`, `Console`, `System`, and `Random`), so when we tested our program above the `helloWorld` method didn't actually print a String to the console but instead wrote the String to a buffer that could access for testing purposes. If you ever do need to access the "live" environment just use the `live` method in the `mock` package or specify the live environment in your type signature like `Live[Console]`.
 
+_Mocking Random_
+
+When working with randomness it is easier to controll what the radnom producer supplies. This can be controlled by a number of methods that allow for filling up buffers with values.
```
Avasil:
```suggestion
When working with randomness it is easier to control what the random producer supplies. This can be controlled by a number of methods that allow for filling up buffers with values.
```

docs/usecases/testing.md
```diff
@@ -48,6 +48,68 @@ In **ZIO Test**, all tests are immutable values and tests are tightly integrated
...
+  } yield assert(
+    List(1, 9, 2, 8, 3, 7, 4, 6, 5),
+    equalTo(List(r1, r2, r3, r4, r5, r6, r7, r8, r9))
+  )
+}
+```
+
+_Mocking Clock_
+
+Sometimes one need to be able to control the flow of time. In most cases you want your unit test to be as fast as possible. Waiting for readl time to pass by is a real killer for this. ZIO exposes a MockClock in testing Environment that can controll time.
```
Avasil:
```suggestion
Sometimes one need to be able to control the flow of time. In most cases you want your unit test to be as fast as possible. Waiting for real time to pass by is a real killer for this. ZIO exposes a MockClock in testing Environment that can control time.
```

docs/usecases/testing.md
```diff
@@ -48,6 +48,68 @@ In **ZIO Test**, all tests are immutable values and tests are tightly integrated
...
+
+Example 1
+
+Below test will pass. If we woule remove the line `MockClock.adjust(Duration(10, TimeUnit.SECONDS))` it would fail. If we wouldn't have access to `MockClock` to make it pass we would need to put manual `ZIO.sleep` call and that test would take 10 seconds every time when run.
```
Avasil:
```suggestion
Below test will pass. If we would remove the line `MockClock.adjust(Duration(10, TimeUnit.SECONDS))` it would fail. If we wouldn't have access to `MockClock` to make it pass we would need to put manual `ZIO.sleep` call and that test would take 10 seconds every time when run.
```

docs/usecases/testing.md
```diff
@@ -48,6 +48,68 @@ In **ZIO Test**, all tests are immutable values and tests are tightly integrated
...
+  for {
+    ref     <- Ref.make(0)
+    _       <- (ZIO.sleep(Duration(10, TimeUnit.SECONDS)) *> ref.update(_+1)).fork
+    _       <- MockClock.adjust(Duration(11, TimeUnit.SECONDS)) 
+    readRef <- ref.get
+  } yield assert(1, equalTo(readRef))
+}
+```
+
+The above code cretes a shared mutable cell that will be incremented by 1 after 10 seconds. With call to `MockClock.adjust(Duration(11, TimeUnit.SECONDS))` we simulate passing of 11 seconds of time. Unfortunately *there is no gurantee* that the ref is already increamented. `ZIO.sleep` only guarantees that at least 10 seconds will pass. There is no upper bound. This depends on when the runtime decides to run fiber that control process of updating this cell.
```
Avasil:
```suggestion
The above code creates a shared mutable cell that will be incremented by 1 after 10 seconds. With call to `MockClock.adjust(Duration(11, TimeUnit.SECONDS))` we simulate passing of 11 seconds of time. Unfortunately *there is no guarantee* that the ref is already incremented. `ZIO.sleep` only guarantees that at least 10 seconds will pass. There is no upper bound. This depends on when the runtime decides to run fiber that control process of updating this cell.
```

jdegoes:
@adamgfraser @Avasil Thank you both for all your help reviewing!

@almendar Sorely needed!

docs/usecases/testing.md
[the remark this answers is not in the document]
almendar: Are you sure about `will "wake up" in order of earliest time first`? E.g. you would put two fibers to sleep and there would be third that was evaluated by the runtime longer than both sleep took. Isn't the runtime free to choose which fiber to pickup first for further evaluation?

almendar:
First of all, I've installed a spellchecker in my VSCode :D 
Thanks @Avasil  for being typo sniper!

@adamgfraser I think I've addressed most (all?) of the remarks. Let's go for another round.

As for your question about user perspective - ZIO is terribly documented  ;) but it's getting steadly better.
I think first time I wanted to pickup zio-testing the testing.md file wasn't there yet so I did what most people would do read the code. There were enough test of zio itself to have something running pretty fast. But other stuff like MockEnvrionment I would first need to look for. Scaladoc are fine when you know what stuff you want to use. Documents that show most common use cases are invaluable. 
I've been running ZIO in production for half a year now and I would gladly get back to such docs from time to time.

@jdegoes no problem. I have very selfish reasons for this :) I want to make on-boarding of people easier for us. E.g. now I would like to throw away ScalaTest, but if people will need to write zio-test code we need a docs for this. I will probably try to document ZStream also in near future.

docs/usecases/testing.md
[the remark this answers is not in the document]
adamgfraser: Good point. The guarantee we can provide is that we will always submit the tasks to the runtime in order of earliest time first (see [here](https://github.com/zio/zio/blob/9d5601de8adf1471ddfd49a0c364175b70d0b285/test/shared/src/main/scala/zio/test/mock/MockClock.scala#L263). So if you have three tasks that are supposed to wake in 9 seconds, 10 seconds, and 11 seconds and you advance the clock by 12 seconds, we will send the first one to the runtime, then the second, and then the third. But you're right that we are sending them to the runtime in that order but almost simultaneously and it is up to the runtime at that point what to do with them.

docs/usecases/testing.md
```diff
@@ -72,3 +72,113 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+Because of it we don't need to wait for the real 10 seconds to pass and thus our unit test can run faster. This is a pattern that will very often be used when `sleep` and `MockClock` are being used for testing of effects that are based on time. The fiber that needs to sleep will be forked and `MockClock` will used to adjust the time so that all expected effects are run in the forked fiber.
+
+*WARNING* 
+Notice that if we don't call `adjust` at all we'll get stuck. `MockClock` doesn't make any progress on its own.
+
+Also it is worth mentioning that adjusting the time on `MockClock` doesn't make us immune to the timing overheads introduced by the runtime and races this introduces. Effects are guaranteed to be waken up not earlier than the argument passed to `sleep` but from there the order of execution since scheduled on different threads is indeterministic and its up to the user code to use tools like `Promise` to guarantee proper sequencing. 
+Below code will fail because there is non-zero overhead when switching fibers thus reading of the value happens before it is set and its change is
```
adamgfraser: You may want to say this test will be flaky versus it will fail. Since there is a race condition between getting and setting the `Ref` the test may pass or fail each time it is run but it will not reliably pass which is what we want for our tests.

almendar: Right 👌

almendar:
I've created a section on mocks in general. It also incorporates your remarks. As for orginizing, sure let's do it. Probably best at the end when we'll see what we've got as I'm also learning along the way of writting :)

I'll start adding next paragraphs about Console ect.

docs/usecases/testing.md
```diff
@@ -72,3 +72,149 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+
+*WARNING* 
+Notice that if we don't call `adjust` at all we'll get stuck. `MockClock` doesn't make any progress on its own.
+
+Also it is worth mentioning that adjusting the time on `MockClock` doesn't make us immune to the timing overheads introduced by the runtime and races this introduces. Effects are guaranteed to be waken up not earlier than the argument passed to `sleep` but from there the order of execution since scheduled on different threads is indeterministic and its up to the user code to use tools like `Promise` to guarantee proper sequencing. 
+Below code will be flaky because there is non-zero overhead when switching fibers thus reading of the value might happen before it is set and its change is 
+propagated.
+
+```scala
+testM("THIS TEST WILL FAIL - Sleep and adjust can introduce races") {
```
jdegoes: Would be nice to show how to fix this by using a `Promise` that is set _after_ the `sleep` inside the forked fiber; then awaiting on the promise right before `MockClock.adjust`.

ghostdogpr:
Hey @almendar , thanks for writing all this!
I got some remarks during the Hackathon about the fact that this documentation page is a bit "hidden" on the website and hard to find. @adamgfraser were thinking it should be a top level category next to "Data Types" or "Use Cases", then you would be able to split it into multiple pages. What do you think?

If we do that, maybe we could move Interop within Ecosystem to avoid having too many menu items. Or maybe someone has another idea?

docs/usecases/testing.md
almendar: Isn't it exactly this example?
https://github.com/zio/zio/pull/1800/files#diff-22f248e1a7bb07e86d37efa910d22132R190

almendar:
@ghostdogpr I think the location is fine but the name is  bit misleading or maybe we should clarify what everyone expects under "use cases". 

Maybe testing here is not the best example but Streaming is.
I'm planning to write some more docs on Streaming. Question is what should be put under
https://zio.dev/docs/usecases/usecases_streaming and what under https://zio.dev/docs/datatypes/datatypes_stream? Should datatypes be raw information about purpose and some basic APIs and use-cases be more practical cross concernt cutting?

ghostdogpr:
I would click “Use Cases” if I’m interested in ZIO but not using it yet. I expect to see there practical situations where ZIO shines. 

I agree with your idea of having APIs described in Data Types and practical cases in Use Cases for Streaming, but ZIO Test isn’t part of Data Types so it needs a reference page easy to find.

adamgfraser:
I think the other thing that is relevant on the website organization is that at least right now every other page under use cases is empty. So we're kind of trying to sell lemonade in an empty neighborhood. That is the thing that pushed me to think we should do something here. We can always revisit in the future.

almendar:
Maybe use cases section should stay as you said high level overview for people who are interested in ZIO. And there could be a Tutorial section that would have things more in spirit of what I've added to testing.md in thir PR?

docs/usecases/testing.md
```diff
@@ -72,3 +72,202 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+payment APIs, object storages, http APIs ect. It is normal to hide these kind of services behind an interface and provide mock implementation to regain
+control and determinism. However there is another source of complexity that comes from the local infrastructure that is also hard to control without building prior abstractions. Things like stdin/stdout, clocks, random generators, schedulers can make writing tests hard or even impossible. Fortunately ZIO abstracted most of it in its runtime under `Environment` type. Thanks to this design `zio-test` could easily provide its own implementation named `MockEnvironment` which gives you mocks of mentioned infrastructure. In most of the cases when you'll be using `ZIO`s `testM` mocks are already created and should be controlled by exposed functions on companion object. If for some reason you would like to provide custom environment or are using other testing framework but still want to use mocks there are `make` functions on companion objects of mocks where you can construct your own. 
+
+A mistake that is easy to do is to create new instance of mock but use the one provided by the `MockEnvironment`.
```
adamgfraser: Maybe say accidentally use different mock instances at the same time?

almendar:
As for providing own environment I would like to address it but first I think there should be a paragraph about how test are values and how this gives extra powers comparing to other test frameworks, a bit how one can construct test  (suit that can have suit than can have testM), a paragraph about assertions (that they are also values) and about providing own environment. I'm not sure where to put that. Before the paragraph about mocks?

adamgfraser:
Yes I think that is a great idea.

docs/usecases/testing.md
almendar: 🆗

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -22,8 +22,8 @@ import Spec._
 
 /**
  * A `Spec[L, T]` is the backbone of _ZIO Test_. Every spec is either a suite,
- * which contains other specs, or a test of type `T`. All specs are annotated
- * with labels of type `L`.
+ * which contains other specs, or a test of type `T` or an already executed test with test result.
```
almendar: Not sure if this is the right wording but as per https://github.com/zio/zio/blob/master/test/shared/src/main/scala/zio/test/package.scala#L93

Spec is used for already run tests.

adamgfraser: A `Spec` is just a collection of tests of some type `T`. Depending on what the type is those tests could be executed or not. As a user when you are interacting with Specs they have generally not been executed yet but are a description of the tests (see for example `ZSpec`). That's why you can do all these nice operations on them like ignore them, retry them, provide them with environments, etc... An `ExecutedSpec` is one particular kind of Spec that describes a test that has been run. ZIO Test works with `ExecutedSpec` for reporting results and if want to modify ZIO Test for advanced use case like reporting results in a different way you might use it, but for most users the specs they are creating in test suites will not have been executed yet.

almendar: I'll revert this change.

almendar:
I've pushed a bit about general design of zio-test. It's a bit of a tricky writing because I was not there when it was designed but I'm trying to put stuff that I find relevant as an end user. I have one more section about running test and providing environment.

docs/usecases/testing.md
```diff
@@ -72,3 +72,271 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+`zio-test` is designed around the idea of making test first-class objects. What it means is that tests 
+(and other accompanying concepts like assertions) become ordinary values that can be passed around, transformed and composed together. This approach allows for greater flexibility comparing to some other testing frameworks where tests and additional logic around tests had to be put into callbacks so that framework could make use of them. This approach also fits better with other `ZIO` concepts like `ZManaged` which can only be used within a scoped block of code. This also created a mismatch between `BeforeAll`, `AfterAll` callback-like methods when there were resources that should be opened and closed during test suite execution.
+
+_Constructing tests_
+
+All below code assumes that you have imported `zio.test._`
+
+The backbone of `zio-test` is the `Spec[L, T]` class. Every spec is labeled with `L` and can be a suite which contains other specs or a test of type `T`.
+
+The most common and easy way to create suits and tests is to use the `suite` and `testM` functions.
```
adamgfraser: You might want to mention `test` for testing pure functions and `testM` for testing effectual functions.

docs/usecases/testing.md
```diff
@@ -72,3 +72,271 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+import Assertion.isGreaterThan
+
+val clockSuite = suite("clock") {
+  testM("time is non-zero") {
+    assertM(nanoTime, isGreaterThan(0))
+  }
+}
+```
+
+As you can see the whole suit was assigned to `clockSuite` val. As it was said suits can contain other suits so we can aggregate them as much as needed. Example, we can have multiple suits that test external HTTP apis and one big suit that will aggregate them all.
```
adamgfraser: "Suites". I think I saw that a couple of other places so would check for spelling throughout.

docs/usecases/testing.md
```diff
@@ -72,3 +72,271 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
 **Zero Dependencies**
 
 As a library with zero third party dependencies, this project is available on the JVM, ScalaJS, and will be available on Dotty and Scala Native in the near future. So you can write your tests once and make sure that your code works correctly across all platforms that you support.
+
+
+**How zio-test was designed**
+
+`zio-test` is designed around the idea of making test first-class objects. What it means is that tests
```
adamgfraser: I think this is definitely right. I might say that not only are tests values, they are also effects. This has a couple of advantages. First, we solve the asynchrony problem where in other frameworks you have to "run" your effects to test them, but then you can't block, especially on Javascript, so then you basically need to rewrap them in some new effect or quasi-effect type like `Future`, creating a lot of indirection. Second, making tests effects gives us access to all the tools of `ZIO` which lets use solve lots of problems in very simple and straightforward ways. Want a timeout on your tests? You don't need special logic. `ZIO` already has a `timeout` method`. Retry a test a bunch of times to make sure it isn't flaky? `repeat`? Use some type of fixture? `Managed`.

docs/usecases/testing.md
```diff
@@ -72,3 +72,271 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+```
+
+Real tests that run some logic and return testing result are created mostly with `testM` function. It expects two arguments, first one that will be the label of test
+which will be used for visual reporting back to the user and an assertion of type
+`ZIO[R, E, TestResult]`. This means writing test in `zio-test` mostly gets down to creating a `ZIO` object that 
+will produce `TestResult`. There is another variant of function for creating test that are pure called simply `test`.
+It expects a thunk of code that will just return a `TestResult` without packing it into `ZIO`.
+
+_Assertions - creating TestResults_
+As it was already mentioned tests should return `TestResult`. The most common way to produce a `TestResult`
```
adamgfraser: I think this is really good. The other thing I might add here is how a lot of assertions compose by taking other assertions as parameters. For example you could write `isRight(isSome(equalTo(2)))`. The idea with this and using methods like `&&` and `||` is that a lot of examples for testing frameworks have these really simply tests where you are comparing two integers or something, so if the test fails it is obvious what went wrong. But in real life maybe you are comparing part of two large case classes, or a bunch of different statements about some test. So it isn't that helpful to just say this test failed. We want to really zero in for you on what part of your test failed so it is easy for you to diagnose the problem and fix it.

docs/usecases/testing.md
almendar: 👍

docs/usecases/testing.md
almendar: 👍

docs/usecases/testing.md
almendar: 👍

almendar:
@adamgfraser 
OK I think I'm done. Unless you don't have anything that you could point me at I don't have any more material in my head planned :) 

When I will have some time to port some of my work tests maybe something new will pop up and I would like to try prop-based testing later on. When I get more acquainted with them I will probably write it down just like with this PR.

almendar:
Oh, one more thing I would like to remove my work email address before merging. I am not sure how it is best done. Normally I would squash all commits and fore push just before the merge. Is that ok?

adamgfraser:
@almendar That is totally fine.

almendar:
OK, squashed. Let's go for another round and do the reorganisation as you've suggested.

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+is to resort to `assert` or its effectful counterpart `assertM`. Both of them accept a value of type `A` (effectful version wrapped in a `ZIO`) and an `Assertion[A]`.
+To create `Assertion[A]` object one can use functions defined under `zio.test.Assertion`. There are already a number
+of useful assertions predefined like `equalTo`, `isFalse`, `isTrue`, `contains`, `throws` and more.
+What is really useful in assertions is that they behave like boolean values and can be composed with operators
+known from operating on boolean values like and (`&&`), or (`||`), negation (`negate`)
```
ghostdogpr: .
```suggestion
known from operating on boolean values like and (`&&`), or (`||`), negation (`negate`).
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+import Assertion.isGreaterThan
+
+val clockSuite = suite("clock") {
+  testM("time is non-zero") {
+    assertM(nanoTime, isGreaterThan(0))
+  }
+}
+```
+
+As you can see the whole suit was assigned to `clockSuite` val. As it was said suites can contain other suites so we can aggregate them as much as needed. Example, we can have multiple suites that test external HTTP apis and one big suit that will aggregate them all.
```
ghostdogpr:
```suggestion
As you can see the whole suit was assigned to `clockSuite` val. As it was said suites can contain other suites so we can aggregate them as much as needed. Example, we can have multiple suites that test external HTTP apis and one big suite that will aggregate them all.
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+
+```scala
+val paymentProviderABCSuite  = suit("ABC payment provider tests") {/* tests */}
```
ghostdogpr:
```suggestion
val paymentProviderABCSuite  = suite("ABC payment provider tests") {/* tests */}
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+val paymentProviderXYZSuite  = suit("XYZ payment provider tests") {/* tests */}
```
ghostdogpr:
```suggestion
val paymentProviderXYZSuite  = suite("XYZ payment provider tests") {/* tests */}
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+val allPaymentProvidersTests = suit("All payment providers tests")(paymentProviderABCSuite, paymentProviderXYZSuite)
```
ghostdogpr:
```suggestion
val allPaymentProvidersTests = suite("All payment providers tests")(paymentProviderABCSuite, paymentProviderXYZSuite)
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+test("Rich checking") {
+  assert(
+    User("Jonny", 26, Address("Denmark", "Copenhagen")),
+    hasField("age", (u:User) => u.age, isGreaterThanEqualTo(18)) &&
+    hasField("country", (u:User) => u.address.country, not(equalTo("USA")))
+  )
+}
+```
+There is also an easy way to test object's data for certain assertions with `hasField` which accepts besides a name, a mapping function from object to its tested property and `Assertion` object which will validate this property. Here our test checks if a person has at least 18 years and is not from USA. What is nice about those tests is that, test reporters will tell you exactly which assertion was broken. Let's say we would change `isGreaterThanEqualTo(18)` to `isGreaterThanEqualTo(40)` which will fail. Printout
+on console will be a nice detailed text what exactly went wrong:
```
ghostdogpr:
```suggestion
on console will be a nice detailed text explaining what exactly went wrong:
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+    s         <- Semaphore.make(1L)
+    permits   <- s.available
+  } yield assert(permits, equalTo(1L))
+}
+```
+
+
+_Running tests_
+
+When all of our test are constructed we need to have a way to actually execute them. Your first stop is the `zio.test.DefaultRunnableSpec` which accepts a single suite that will be executed. A single suite might seem to be limiting but as it was already said suit can hold any number of other suites. You may structure your tests like this:
```
ghostdogpr:
```suggestion
When all of our tests are constructed, we need to have a way to actually execute them. Your first stop is the `zio.test.DefaultRunnableSpec` which accepts a single suite that will be executed. A single suite might seem to be limiting but as it was already said suites can hold any number of other suites. You may structure your tests like this:
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+}
+
+object AllSuites extends DefaultRunnableSpec(suite("All tests")(suite1, suite2, suite3))
+```
+
+`DefaultRunnableSpec` is very similar in its logic of operations to `zio.App`. Instead of providing one `ZIO` application
+at the end of the world we provide a suite that can be a tree of other suites and tests. Another resemblance is that `DefaultRunnableSpec` provides an Environment. Here it is an instance of `MockEnvironment` which helps us with controling our systems infrastructure. More info on using mocks can be found in sections bellow. 
+Just like with `zio.App` where at the very end an instance of `ZIO[R,E,A]` is expected where `R` can be at maximum of type `Environment` in `DefaultRunnableSpec` `R` cannot be more than `MockEnvironment`. So just like in normal application if our
+`R` is composed of some other modules we need to provide them first before test can be executed. How can we provide our dependencies?
+Here again the design of `zio-test` shines. Since our tests are ordinary values we can jest transform them with call to `mapTest`.
```
ghostdogpr:
```suggestion
Here again the design of `zio-test` shines. Since our tests are ordinary values we can just transform them with a call to `mapTest`.
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+It accepts a lambda of type `ZIO[R with MockSystem, TestFailure[Throwable], TestSuccess[Unit] ] => T1`. Wihtout getting into too much details about types we can see that our lambda argument is a test instance (`ZIO`) that expects an environment of type `R with MockSystem`. This is no different from normal usage of ZIO in `zio.App`. We can use the same `provide`, `provideSome` methods to provide modules which `DefaultRunnableSpec` cannot provide itselt as those are users modules.
```
ghostdogpr:
```suggestion
It accepts a lambda of type `ZIO[R with MockSystem, TestFailure[Throwable], TestSuccess[Unit] ] => T1`. Without getting into too much details about types we can see that our lambda argument is a test instance (`ZIO`) that expects an environment of type `R with MockSystem`. This is no different from normal usage of ZIO in `zio.App`. We can use the same `provide`, `provideSome` methods to provide modules which `DefaultRunnableSpec` cannot provide itself as those are users modules.
```

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+
+
+
+
+_Using Mocks in tests_
+
+What we expect from tests (at least those that we consider unit tests) is to be stable i.e. consecutive runs should yield the same results and take
+more or less the same amount of time. Biggest source of complexity during testing comes from external services which we cannot control like external
+payment APIs, object storages, http APIs ect. It is normal to hide these kind of services behind an interface and provide mock implementation to regain
```
ghostdogpr:
```suggestion
payment APIs, object storages, http APIs etc. It is normal to hide these kind of services behind an interface and provide mock implementation to regain
```

almendar:
@ghostdogpr thanks for catching all the typos 👍 !

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+}
+val suite3 = suite("suite3") {
+  testM("s3.t1") {...}
+}
+
+object AllSuites extends DefaultRunnableSpec(suite("All tests")(suite1, suite2, suite3))
+```
+
+`DefaultRunnableSpec` is very similar in its logic of operations to `zio.App`. Instead of providing one `ZIO` application
+at the end of the world we provide a suite that can be a tree of other suites and tests. Another resemblance is that `DefaultRunnableSpec` provides an Environment. Here it is an instance of `MockEnvironment` which helps us with controling our systems infrastructure. More info on using mocks can be found in sections bellow.
```
adamgfraser: "below" instead of "bellow".

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
 **Zero Dependencies**
 
 As a library with zero third party dependencies, this project is available on the JVM, ScalaJS, and will be available on Dotty and Scala Native in the near future. So you can write your tests once and make sure that your code works correctly across all platforms that you support.
+
+
+**How zio-test was designed**
+
+`zio-test` is designed around the idea of making test first-class objects. What it means is that tests (and other accompanying concepts like assertions) become ordinary values that can be passed around, transformed and composed together. This approach allows for greater flexibility comparing to some other testing frameworks where tests and additional logic around tests had to be put into callbacks so that framework could make use of them. This approach also fits better with other `ZIO` concepts like `ZManaged` which can only be used within a scoped block of code. This also created a mismatch between `BeforeAll`, `AfterAll` callback-like methods when there were resources that should be opened and closed during test suite execution.
```
adamgfraser: "tests" instead of "test" in the first sentence.

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+Just like with `zio.App` where at the very end an instance of `ZIO[R,E,A]` is expected where `R` can be at maximum of type `Environment` in `DefaultRunnableSpec` `R` cannot be more than `MockEnvironment`. So just like in normal application if our
+`R` is composed of some other modules we need to provide them first before test can be executed. How can we provide our dependencies?
+Here again the design of `zio-test` shines. Since our tests are ordinary values we can just transform them with a call to `mapTest`.
+It accepts a lambda of type `ZIO[R with MockSystem, TestFailure[Throwable], TestSuccess[Unit] ] => T1`. Without getting into too much details about types we can see that our lambda argument is a test instance (`ZIO`) that expects an environment of type `R with MockSystem`. This is no different from normal usage of ZIO in `zio.App`. We can use the same `provide`, `provideSome` methods to provide modules which `DefaultRunnableSpec` cannot provide itself as those are users modules.
+
+
```
adamgfraser: Maybe talk a little more about how you can either run specs extending `DefaultRunnableSpec` with the main method using `test:run`` or with the SBT test runner that will automatically detect all your tests and run them?

docs/usecases/testing.md
```diff
@@ -72,3 +72,364 @@ Test aspects are powerful tools for modifying behavior of individual tests or ev
...
+The returned type is `ZIO[R with Clock, E, Option[A]]` where our environment is "some R plus a Clock".
+Before running this `Clock` has to be provided and the framework provides the Clock from the `MockEnvironment` not our instance variable as it is not aware that we created it.
+
+
+If you need to provide real implementations instead of the mocks to some part of your tests there is a `live` method which will transform your `ZIO[R, E, A]` to `ZIO[Live[R], E, A]`. Going from `R` to `Live[R]` instructs the framework that we really want to be provided with live implementations and not mocks.
+
+
+If you have a non-trivial flow of data from multiple streams that can produce at different intervals and would like to test
+snapshots of data in particular point in time a nice pattern is to join these streams with `zipWithLatest` and
```
adamgfraser: I would generalize this a little bit. `zipWithLatest` is the specific method we are testing here but it is a more general pattern. If we have a single value we're waiting for after a certain amount of time we can have the effect complete a promise, advance the clock, and wait on the promise. But if we want to advance time multiple times and wait on multiple results, using a queue gives us the same ability to "wait" for the next result without having to create multiple promises.

almendar: OK, will talk about it and I think I will move this example to MockClock section. It makes more sense there.

almendar:
https://github.com/zio/zio/pull/1800#pullrequestreview-294326347

I am planning to start working on the `streams.md` files soon so you can account for that one as well :)

almendar:
Looking at the diffs I think it is important to do line breaks at constant size because it makes reviews harder then they should be.

ghostdogpr:
Shall we do the org changes in this PR or merge it first and do it in a later one?

adamgfraser:
@ghostdogpr Either way is fine with me. @almendar Any preference?

almendar:
Nope, both are fine with me.

almendar:
I'll try to do it tonight or tomorrow on this pr

jdegoes:
@ghostdogpr 

> I like that, how about calling it How to? It could replace current Advanced and there would be pages like:

Interesting idea, I like it. 

"How to do" practical things are easy and obvious to use.

almendar:
OK, I've moved all from advanced to how to. The parts written by me were put into Effectful testing. Rest stayed as it was. Hope I did not break anything, don't have any experience with those sites generators.

docs/howto/effectful_testing.md
```diff
@@ -0,0 +1,368 @@
+---
+id: howto_effectful_testing
+title:  "Effectful testing"
```
ghostdogpr: How about we use verbs so that it sounds better after "How to...". Here "Test effects"

docs/howto/error_model.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: advanced_error_model
+id: howto_error_model
```
ghostdogpr: "Handle errors"

docs/howto/performance.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: advanced_performance
+id: howto_performance
```
ghostdogpr: This one doesn't really fit "how to", I suggest we move it to the end of "Overview"

docs/howto/rts.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: advanced_rts
+id: howto_rts
```
ghostdogpr: Let's get rid of this one (it's empty). We can add it back at the right place when someone writes it.

docs/howto/system.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: advanced_system
+id: howto_system
```
ghostdogpr: "Access system information"

website/sidebars.json
```diff
@@ -55,13 +55,15 @@
...
-      "advanced/advanced_system"
+  "howto-sidebar": {
+    "How to": [
+      "howto/howto_index",
+      "howto/howto_effectful_testing",
+      "howto/howto_error_model",
+      "howto/howto_performance",
+      "howto/howto_rts",
+      "howto/howto_system",
+      "howto/howto_effectful_testing"
```
ghostdogpr: This one is duplicated

almendar:
@ghostdogpr will CI test if webpage is ok?

docs/howto/handle_errors.md
```diff
@@ -1,6 +1,6 @@
 ---
-id: advanced_error_model
-title:  "Error Model"
+id: howto_handle_errors
+title:  "How to: Handle errors"
```
ghostdogpr: Ah, you don't need to repeat `How to: ` here since the title of the section will be `How to`

ghostdogpr: (same for the others)

almendar: Argh, I thought that this will only show in page title:
image

and I didn't understand it will also appear on side bar. Will fix it.

almendar:
The page title looks a bit strange now, but I guess that's ok.

almendar:
BTW:
```
[error] stack trace is suppressed; run last docs / docusaurusCreateSite for the full output
[error] (docs / docusaurusCreateSite) java.io.IOException: Cannot run program "yarn" (in directory "/Users/t.kogut/dev/github/zio/zio/website"): error=2, No such file or directory
[error] Total time: 100 s (01:40), completed Oct 1, 2019, 3:23:23 PM
sbt:zio> git sat^?^?^?[info] shutting down sbt server
```

my first thought was, why the hell generating WWW involves hadoop :D

ghostdogpr:
hadoop? you just need yarn installed ^^

almendar:
yeah, I know just I'm not really involved with JS-stuff :) I know there is yarn in JS-world but for me it's is a resource manager for hadoop 🙈

ghostdogpr:
I think we're good!

almendar:
Thanks guys!

PR #1812, opened 2019-09-24
Implement repeatUntil and repeatUntilM

YuvalItzchakov:
This PR adds two combinators to the API surface. These combinators came up as useful during John's ZIO Workshop and I thought it was worth implementing

- `repeatUntil`: Takes a `PartialFunction[A, B]` and repeats until the effect is in the codomain of the partial function
- `repeatUntilM`: Same as `repeatUntil` but takes an effect instead, evaluating it until it yields a success.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1055,6 +1055,24 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   }
 
   /**
+   * Repeats this effect until the result exists in the partial function
+   * codomain.
+   */
+  final def repeatUntil[B](f: PartialFunction[A, B]): ZIO[R, E, B] =
+    self.flatMap { a =>
+      f.lift(a) match {
+        case Some(value) => ZIO.succeed(value)
+        case None        => repeatUntil(f)
+      }
+    }
+
+  /**
+   * Effectful version of `repeatUntil`. Will repeat until `f` succeeds
+   */
+  final def repeatUntilM[R1 <: R, E2, B](f: A => ZIO[R1, E2, B]): ZIO[R1, E2, B] =
+    self.flatMap(f).catchAll(_ => repeatUntilM(f))
+
+  /**
    * Retries with the specified retry policy.
    * Retries are done following the failure of the original `io` (up to a fixed maximum with
    * `once` or `recurs` for example), so that that `io.retry(Schedule.once)` means
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1060,7 +1060,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def repeatUntil[R1 <: R, E1 >: E, B](f: PartialFunction[A, ZIO[R1, E1, B]]): ZIO[R1, Nothing, B] =
     self.flatMap { a =>
       f.lift(a) match {
-        case Some(value) => value.eventually
+        case Some(value) => value
         case None        => repeatUntil(f)
       }
     }.eventually
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1057,19 +1057,24 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Repeats this effect until the partial function succeeds on the result.
    */
-  final def repeatUntil[B](f: PartialFunction[A, B]): ZIO[R, E, B] =
+  final def repeatUntil[R1 <: R, E1 >: E, B](f: PartialFunction[A, ZIO[R1, E1, B]]): ZIO[R1, Nothing, B] =
     self.flatMap { a =>
       f.lift(a) match {
-        case Some(value) => ZIO.succeed(value)
+        case Some(value) => value.eventually
         case None        => repeatUntil(f)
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Can you rebase to the latest master? The error you have with the CI has been fixed.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1054,6 +1054,24 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     schedule.initial.flatMap(loop(None, _))
   }
 
+  /**
+   * Repeats this effect until the result exists in the partial function
+   * codomain.
```
neko-kai: I'd probably reword this as "until the partial function succeeds on the result"

YuvalItzchakov: Yup, done.

YuvalItzchakov:
Should these methods be implemented on the ZIO companion as well?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1054,6 +1054,23 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+      f.lift(a) match {
+        case Some(value) => ZIO.succeed(value)
+        case None        => repeatUntil(f)
+      }
+    }
+
+  /**
+   * Effectful version of `repeatUntil`. Will repeat until `f` succeeds
+   */
+  final def repeatUntilM[R1 <: R, E2, B](f: A => ZIO[R1, E2, B]): ZIO[R1, E2, B] =
```
iravid: Should the result be typed `ZIO[R1, Nothing, B]`? The function will never fail.

Also, this is essentially `flatMap(f).eventually`. Does this justify a dedicated combinator?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1054,6 +1054,23 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     schedule.initial.flatMap(loop(None, _))
   }
 
+  /**
+   * Repeats this effect until the partial function succeeds on the result.
+   */
+  final def repeatUntil[B](f: PartialFunction[A, B]): ZIO[R, E, B] =
```
iravid: Nice

core/shared/src/main/scala/zio/ZIO.scala
YuvalItzchakov: It should. The main use case was using the console to read input and make sure it's valid:

```scala
for {
    _ <- putStrLn("Please enter a time in 24 hour format (hh:mm): ")
    userTime <- getStrLn.repeatUntil(in => Task.effect(in.toString).tapError(_ => putStrLn("Please enter a valid value"))
  } yield userTime
```

Does this justify a dedicated combinator? I'm not sure :)

iravid: Ok, let's wait for another opinion :-)

YuvalItzchakov: I talked with @jdegoes about this during the workshop, perhaps he can add his 2 cents.

jdegoes: The return type should definitely omit the final error, and we can implement it using `(self >>= f).eventually`. 

The symmetry with `repeatUntil` is nice, although perhaps we could make it closer still by allowing `PartialFunction[A, ZIO[R1, E2, B]]`. The implementation of this one is also more complex and possibly justifying its existence.

Thoughts?

YuvalItzchakov: SGTM. I've reworked the implementation, WDYT now? @iravid @jdegoes @neko-kai

iravid: Did you mean to convert `repeatUntilM` to take a `PartialFunction`?

iravid: I.e. `repeatUntil` just uses a plain partial function, `repeatUntilM` allows evaluating an effectful partial function

YuvalItzchakov: @iravid If I understood correctly, I think he meant for `repeatUntil` to take a `PartialFunction[A, ZIO[R1, E1, B]]` and for `repeatUntilM` to take a total function `A => ZIO[R1, E1, B]`

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1054,6 +1054,23 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+      f.lift(a) match {
+        case Some(value) => value
+        case None        => repeatUntil(f)
+      }
+    }.eventually
+
+  /**
+   * Effectful version of `repeatUntil`. Will repeat until `f` succeeds
+   */
+  final def repeatUntilM[R1 <: R, E1 >: E, B](f: A => ZIO[R1, E1, B]): ZIO[R1, Nothing, B] =
```
jdegoes: Sorry for the confusion, I think this one should be `PartialFunction[A, ZIO[R1, E1, B]]`, and that `repeatUntil` should be `PartialFunction[A, B]`. Then there is symmetry between them and it matches the overall ZIO convention on the `M` suffix. Does that make sense?

jdegoes:
@YuvalItzchakov This one just needs a small fix and we can merge it. However, I will close the PR now to avoid clutter, please re-open when it makes sense for you to resume!

PR #1813, opened 2019-09-24
Fixed #1811 Add `effectRetry` to ZIO.

### LINKED ISSUE #1811: Add `effectRetry` to ZIO
Effect retry will attempt to import a synchronous effect and before retries, we compose an onFailure effect.  

An example of this would be 

```scala
val numberGuessingGame: RIO[Console with Random, Unit] = for {
    number    <- random.nextInt(5).map(_+1)
    _         <- putStrLn("Enter a number between 0 - 5: ")
    result    <- effectRetry(scala.io.StdIn.readLine().toInt, putStrLn("You didn't enter an integer!"))
                  .flatMap(guess =>
                      if (guess == number)
                        putStrLn("You guessed right! The number was " + number)
                      else
                        putStrLn("You guessed wrong! The number was " + number)
                  )
  } yield result
```

cloudmark:
Added `effectRetry` to ZIO, during the advanced programming course we used `effectRetry` quite a number of times and it feels missing.

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -151,6 +151,11 @@ object IO {
   final def done[E, A](r: Exit[E, A]): IO[E, A] = ZIO.done(r)
 
   /**
+   * @see See [[zio.ZIO.effectRetry]]
+   */
+  final def effectRetry[R, A](effect: => A, onFailure: URIO[R, Unit]): URIO[R, A] = ZIO.effectRetry(effect, onFailure)
+
+  /**
    * @see See [[zio.ZIO.effect]]
    */
   final def effect[A](effect: => A): Task[A] = ZIO.effect(effect)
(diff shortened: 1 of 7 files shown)


cloudmark:
@jdegoes well, now there is an `effectRetry` :)

iravid:
I don't think this is sufficiently differentiated from other combinators (e.g. `effect(...).eventually` or `effect(...).retry(...)`, or is more composable than what is currently available, so I am 👎 on this.

adamgfraser:
I'm inclined to agree. I think this is a symptom of the fact that `Schedule` isn't great for these sorts of uses cases right now because of the `Clock` dependency and we need to solve that versus developing workarounds.

jdegoes:
👍 I think I agree, `effect(...).tapError(...).eventually` works reasonably well enough.

@cloudmark Thanks for working this up though, I hope the experience was valuable. 🙏

PR #1814, opened 2019-09-24
Upgrade Dotty Version

adamgfraser:


iravid:
I think we need to turn off the no-braces syntax.

vasilmkd:
I don't even understand the error, what's wrong?

adamgfraser:
It has to do with the significant whitespace change. Now the indentation has a semantic effect, so the fact that the brace closing the match expression is at the same indentation level as the else in the if expression is throwing the compiler off. My understanding had been that all valid syntax was going to continue to be valid under the whitespace proposal so I am planning on opening a ticket on the Dotty repo.

vasilmkd:
Exactly my thoughts as well... Does this mean that significant whitespace is now the default mode of compilation??

adamgfraser:
Okay, so after discussions with the Dotty team it is not guaranteed that previously valid code will compile and in particular they view this kind of code as improperly indented and not supported.

The further complication is that not only is the formatting at issue legal in existing Scala versions, at least our formatting settings automatically format code from the version that is legal under Dotty to the version that is no longer legal.

I think there are two ways we can move forward:

1. There is a `-noindent` compiler option we can use which I believe will prevent these errors. Upside is it is easy, downside is that it sounds like `-noindent` silencing these sorts of warnings is a bit of an implementation detail that is not guaranteed to continue working and to a certain extent we aren't future proofing ourselves for Dotty by opting out of the direction they are going.
2. We can just add braces around the first clause of `if` statements like this. I'm happy to go through and clean this up. My concern more is that it creates a little friction for contributors in that people will think their code is fine compiling locally and won't get a warning until Dotty compilation fails. We can try to educate people and there are already other cases on the margin like this around things like new collection in 2.13 but it is not great.

My inclination is to do the first option until we can get some tooling in place to help us with this. I have reached out to the scalafmt team. If we can get our formatting settings to go the other way (rewriting invalid to valid indentation) then we can turn off the compiler flag and handle this very seamlessly.

```scala
object Example {

  val okay =
    if (true)
      List(1, 2, 3) match {
        case x :: xs => ???
        case Nil => ???
      }
    else {
      2
    }

  val notOkay = 
    if (true)
      List(1, 2, 3) match {
        case x :: xs => ???
        case Nil => ???
    } else {
      2
    }
}
```

iravid:
Most definitely `-noindent`.

vasilmkd:
First time I've ever wanted a sad reaction to a github comment. @adamgfraser

adamgfraser:
This is ready to go.

PR #1821, opened 2019-09-24
processChunk in toInputStream

sideeffffect:


iravid:
`StreamChunkSpec` is timing out locally for me too, so this looks like an actual problem.

iravid:
It appears that the `toInputStream` is hanging now

sideeffffect:
I suspect that `processChunk` is at fault here, but I don't really understand what could be the issue with it... I tried to rework it by wrapping it all in `Managed.effectTotal`
let's hope it helps

iravid:
Hi @sideeffffect, any luck with this?

sideeffffect:
@iravid sadly not, I have no idea why a simple, type-preserving, reorganization of the code suddenly causes tests to hang and whatnot :crying_cat_face:

iravid:
@sideeffffect 

The counter is never increased in `processChunk`. This does the trick after the while loop:
```scala
          val elem = chunk(counter)
          counter += 1
          elem
```

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
```diff
@@ -90,18 +90,25 @@ private[stream] class StreamEffectChunk[-R, +E, +A](override val chunks: StreamE
...
-          counter = 0
+      Managed.effectTotal {
+        var counter         = 0
+        var chunk: Chunk[A] = Chunk.empty
+        def pull(): A = {
+          while (counter >= chunk.length) {
+            chunk = thunk()
+            counter = 0
+          }
+          chunk(counter)
```
iravid:
```suggestion
          val elem = chunk(counter)
          counter += 1
          elem
```

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
```diff
@@ -90,18 +90,25 @@ private[stream] class StreamEffectChunk[-R, +E, +A](override val chunks: StreamE
...
         }
-        chunk(counter)
+        () => pull()
       }
-      ZManaged.fromEffect(ZIO.succeed(() => pull()))
+    }
+
+  final def flatten: StreamEffect[R, E, A] =
```
iravid: This is `flattenChunks` and can override the method defined in `StreamChunk`.

streams/shared/src/main/scala/zio/stream/StreamEffectChunk.scala
sideeffffect: :man_facepalming: of course! thank you

PR #1825, opened 2019-09-25
Clear up some Chunk hotspots

iravid:
I think we had to keep some things in `lazy val` or `def` before because of specialization. 

I took a quick flamegraph of `StreamChunkSpec` and `Chunk.Concat` is a big hotspot. Making things a bit more eager lessens that.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -407,12 +407,13 @@ sealed trait Chunk[+A] { self =>
   }
 
   def toSeq: Seq[A] = {
+    val c          = materialize
     val seqBuilder = Seq.newBuilder[A]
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -807,7 +808,7 @@ object Chunk {
 
     override def toArray[A1 >: A](implicit tag: ClassTag[A1]): Array[A1] = array.asInstanceOf[Array[A1]]
 
-    override def length: Int = array.length
+    override val length: Int = array.length
```
jdegoes: Another thing you can do: anytime `length` is referenced, instead put it into a local `val`. This way it won't go through `invokevirtual` like all (non-`private[this]`) `val` access does.

Issue #1809, opened 2019-09-24
Be less lossy about error composition in `orElse` and variants


jdegoes:
Currently `orElse` and variants are implemented in terms of `tryOrElse`, which will discard failure information on the left hand side, even if the right hand side fails. If the right hand side fails, it would be nice to have access to information on the failure of the left hand side, as well.

While discarding some information is inevitable — mainly because the left and right may have entirely different error types — holding onto some information may be useful.

In particular, we could wrap the `Cause` from the left hand side in `FiberFailure`, and make it a defect that is exposed chronologically prior to the errors of the right hand side. Or possibly, we could strip the `E1`s from the left hand side, leaving a `Cause[Nothing]`, which could precede the `Cause[E2]` of the right hand side.

```scala
  private[this] final def tryOrElse[R1 <: R, E2, B](
    that: => ZIO[R1, E2, B],
    succ: A => ZIO[R1, E2, B]
  ): ZIO[R1, E2, B] =
    new ZIO.Fold[R1, E, E2, A, B](
      self,
      ZIOFn(() => that)(_.stripFailures match {
        case None    => that
        case Some(c) => ZIO.halt(c)
      }),
      succ
```

Twizty:
Hi, I'd like to implement that.

vasilmkd:
The `ZSink#orElse` combinator closely follows these semantics as well. It would be nice if it continued to do so.

@jdegoes We used `Cause.Both` in other `ZSink` combinators. Would that make sense here?

Twizty:
@jdegoes Hi, I'd like to clarify, does this issue require to fix `orElse`, or to fix `tryOrElse` in general? To fix just `orElse` I came up with something like this:
```
  final def orElse[R1 <: R, E2, A1 >: A](that: => ZIO[R1, E2, A1]): ZIO[R1, E2, A1] =
    self.foldCauseM(cause => cause.stripFailures match {
      case None    => that.catchAllCause(cause2 => ZIO.halt(Cause.die(FiberFailure(Cause.Both(cause, cause2)))))
      case Some(c) => ZIO.halt(c)
    }, _ => that)
```

Is it something you were looking for?

neko-kai:
Done in https://github.com/zio/zio/pull/1828

Issue #1807, opened 2019-09-24
Timeout on ChunkGen Test


adamgfraser:
It looks like the third monad law test for `ChunkGen` is occasionally timing out, causing flakiness. @regiskuckaertz Could you take a look?

https://circleci.com/gh/zio/zio/27859?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

regiskuckaertz:
Ok, will do as soon as I get the opportunity either later today, or tomorrow as today this guy arrives in the family 😍 

![IMG_1590](https://user-images.githubusercontent.com/629976/65504503-70923900-dec7-11e9-849c-9353a61220e1.jpeg)

ghostdogpr:
@regiskuckaertz I was gonna ask you a picture 😄 Enjoy!

iravid:
😍

adamgfraser:
Whoa! Awesome!

adamgfraser:
This is quite old and I have not observed any flakiness in this test so going to close.

PR #1828, opened 2019-09-25
Issue1809

Twizty:
Implemented #1809


Twizty:
@jdegoes @iravid @neko-kai please review.

Twizty:
For some reason `StreamChunk.monadLaw3` failed in 2 checks. I can't re-run the build because I don't have a permission. Is it a known error? I can't reproduce it locally.

adamgfraser:
@Twizty Did it time out?

Twizty:
@adamgfraser yes, `Timeout of 1 m exceeded`.

adamgfraser:
Yes. There is an issue for it in #1807 and @regiskuckaertz is working on it. It is a computationally intensive test that sometimes takes longer than the timeout to complete, leading to flakiness. I expect the solution is to reduce the size of the inputs to the test.

Twizty:
@adamgfraser Is there a way to run tests again? What can I do?

adamgfraser:
I believe some people with write access to CI have the ability to rerun. You can push another change if there are more edits you want to make and it will run again (but you may have the same problem). If you have implemented your feature and all of your tests are passing I wouldn't worry about it too much, though I know it can be annoying. When someone reviews they can see that everything is passing except for this flaky test and merge if everything else is good.

Twizty:
Yes, my feature is implemented and tests are written.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -673,7 +673,19 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * otherwise executes the specified effect.
    */
   final def orElse[R1 <: R, E2, A1 >: A](that: => ZIO[R1, E2, A1]): ZIO[R1, E2, A1] =
-    tryOrElse(that, new ZIO.SucceedFn(() => that))
+    self.foldCauseM(
+      ZIOFn(() => that) { cause =>
+        cause.stripFailures match {
+          case None =>
+            that.catchSomeCause {
+              case c2 if c2.died => ZIO.halt(Cause.die(FiberFailure(Cause.Both(cause, c2))))
```
jdegoes: Why does it have `if c2.died` here?

Twizty: @jdegoes as far as I understood, I only need to catch died actions. Should I catch failed as well?

Twizty: If I catch both failed and died actions here, I would make the whole action died. I thought it isn't semantically correct.

Twizty: I could handle failed right hand side with something like this:
```scala
case c2 if c2.failed => ZIO.halt(Cause.Both(Cause.die(FiberFailure(cause)), c2))
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1357,10 +1357,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
     new ZIO.Fold[R1, E, E2, A, B](
       self,
-      ZIOFn(() => that)(_.stripFailures match {
-        case None    => that
-        case Some(c) => ZIO.halt(c)
-      }),
+      ZIOFn(() => that) { cause =>
+        cause.stripFailures match {
+          case None =>
+            that.catchSomeCause {
```
jdegoes: I think what you really want is `that.catchAll(cause2 => cause1 ++ cause2)`. There is no need to special case on die / failure (also they are not mutually exclusive, the same cause can be death and failure at the same time).

Twizty: If I just sum 2 causes, I would have a situation where `IO.fail(1) orElse IO.fail(2)` produce `ZIO.halt(Cause.die(FiberFailure(cause ++ cause2)))`, it's not possible to catch this with `either`, only with `catchAllCause`, and application dies.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1357,10 +1357,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E2, B] =
     new ZIO.Fold[R1, E, E2, A, B](
       self,
-      ZIOFn(() => that)(_.stripFailures match {
-        case None    => that
-        case Some(c) => ZIO.halt(c)
-      }),
+      ZIOFn(() => that) { cause =>
+        cause.stripFailures match {
+          case None    => that.catchAllCause(cause2 => ZIO.halt(Cause.die(FiberFailure(cause)) ++ cause2))
```
jdegoes: If we are going to wrap in `Cause.die` (I am not sure 😆 ), we could avoid `stripFailures` altogether.

Twizty: Are we? I thought `stripFailures` were used to continue execution if the error that happened previously wasn't fatal. Otherwise execution halts.

Twizty: For example `IO.die(1) orElse IO.fail(2)` would fail on the right side instead of dying on the left side.

jdegoes: `stripFailures` will remove the `E` from the `fail`. But if you are taking the _whole_ cause and putting it into `Die`, then that `E` will not be visible at the type level.

We will always do `that` right after `this` fails, it's only when both fail that we take the whole left `cause`, smash it into `Die`, and then follow it by the right's (typed) failure.

You can write a test here to see the behavior of `self.catchAllCause(cause1 => that.catchAllCause(cause2 => Cause.die(FiberFailure(cause1)) ++ cause2))`

Twizty: Yes, I understand it, and if I do `self.catchAllCause(cause1 => that.catchAllCause(cause2 => Cause.die(FiberFailure(cause1)) ++ cause2))` then `ZIOSpecJvm.testOrElseDefectHandling` fails, because `(ZIO.die(ex) <> IO.unit).run` isn't equal to `Exit.die(ex)` anymore. Should I change this behaviour or does it need to be preserved?

jdegoes: Hmm, what does the error change to?

Twizty:
```

Comment (Twizty) [reply core/shared/src/main/scala/zio/ZIO.scala]:
```
Success(()) != Failure(Die(java.lang.Exception: Died))
Expected :Failure(Die(java.lang.Exception: Died))
Actual   :Success(())
```

**Comment by Twizty** [comment]
@jdegoes could you please take a look at this issue once again, if there is anything you'd like to change.

**Comment by jdegoes** [comment]
@Twizty Will try to take a look at lunch today. Thanks for your work on this!!!

File: core/shared/src/main/scala/zio/ZIO.scala (hunk 3 above)
Comment (jdegoes) [reply core/shared/src/main/scala/zio/ZIO.scala]: OK, right. That's because we don't want `orElse` to "recover" from defects. So if there are any defects, we fail it. Well, we can do that in another way:

```scala
       ZIOFn(() => that) { cause =>
         if (cause.died) ZIO.halt(cause)
         else that.catchAllCause(cause2 => ZIO.halt(Cause.die(FiberFailure(cause)) ++ cause2))
       }
```

This is more direct and reflects intention better than `stripFailures`.

neko-kai: I agree, but `cause.died` would still recover from internal interrupts, might as well use `cause.failed` to be in line with other typed error recovery functions

Twizty: @jdegoes I'm sorry, I haven't seen your response. In my opinion, `stripFailures` is better because it returns `Option[Cause[Nothin]]`, and can pass it directly to `ZIO#halt`. Otherwise I would need to wrap `cause` into `Cause.die(FiberFailure(cause))`.

neko-kai: Seems like `stripFailures` is required here to typecheck, so I think there's not much to be changed here and it's probably good to go as is, so I'll merge. 👍

neko-kai: Another way would be to cast the Cause, but I guess it's not necessary tbh:

```scala
       ZIOFn(() => that) { cause =>
         if (!cause.failed) ZIO.halt(cause.asInstanceOf[Cause[E2])
         else that.catchAllCause(cause2 => ZIO.halt(Cause.die(FiberFailure(cause)) ++ cause2))
       }
```

PR #1830, opened 2019-09-25
Mocking framework

ioleo:
This PR introduces the framework for zero-dependency mocking of services. Kudos to @jdegoes for the design and helping me out with some type inference issues :heart:.

I think it's best explained by showing how it's supposed to be used:
(examples moved below)

There is still room for improvement:
- [x] fix type inference so you don't have to explicitly pass types on the last `MockSpec.expect`
- [x] fix implicit mockable discovery issue, so user does not have to manually import it into scope
- [x] update documentation (in general expand on the idea of environmental effects and module pattern)
- [x] handle `MockException` in `zio.Test` and display nicely the failed expectations
- [x] add more combinators to `MockSpec`, eg. one that ignores the `A` or one without predicate like here `MockSpec.expect(Cache.Service.Clear)(anything)(_ => UIO.unit)` since `clear` does not take any input values (input type is `Nothing`), but we're forced to add the `anything` predicate and ignore the value when producing the return value `_ => UIO.unit`
- [x] add capability tags and mockable implementation for all default services (Console, Random, Clock, etc)
- [x] add examples
- [x] add tests
- [x] rename/move mock modules to fake/stub following the semantics defined in Martin Fowlers [post](https://martinfowler.com/articles/mocksArentStubs.html)
- [x] uncomment & refactor to `zio.test` or deal otherwise with TestMail tests
- [x] ~switch to `ZIO.dieOption` to drop tracing info & refactor DefaultTestRenderer to match on `Cause.Die` directly (without wrapping `Cause.Traced`) when #1706 is merged~ no longer neccessary
- [x] fix CI bugs
- [x] remove the need to explicitly lift spec to managed env
- [x] check if expectsRef is empty when the program finishes (die with new mock excpetion & handle it in zio.test)

CC @adamgfraser @ghostdogpr @jdegoes


ioleo:
Hmm intereting failures

adamgfraser:
I think it is just we have a namespace conflict because all the existing `mock` objects use `Mock` as the name for their implementations. So like `MockRandom.Mock` is the actual implementation of the `MockRandom` service. We can probably come up with a different naming convention.

adamgfraser:
Anyway, will review in more detail later but very excited about this!

test/shared/src/main/scala/zio/test/mock/Mock.scala
```diff
@@ -0,0 +1,184 @@
...
+    o: O,
+    p: P,
+    q: Q,
+    r: R,
+    s: S,
+    t: T
+  ): IO[E0, A0] =
+    invoke(method)((a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t))
+
+  final def invoke22[E0, A0, A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U](
```
adamgfraser: This must have been fun. 😃

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
```diff
@@ -4,7 +4,7 @@ import scala.concurrent.Future
 import scala.util.{ Random => SRandom }
 
 import zio.{ Chunk, DefaultRuntime, UIO }
-import zio.test.mock.MockRandom.{ DefaultData, Mock }
+import zio.test.mock.MockRandom.{ DefaultData, Mock => RMock }
```
adamgfraser: We may want to find a new name for the `Mock` objects in `MockRandom` and friends to avoid namespace conflicts and confusion.

ioleo:
Added combinators and updated example in PR description.

iravid:
Damn this is nice!

test/shared/src/main/scala/zio/test/mock/Method.scala
```diff
@@ -0,0 +1,3 @@
+package zio.test.mock
+
+trait Method[+A, B]
```
adamgfraser: Can this be contravariant in `B`?

ioleo: I remember it blowing up in some places when I did that, but perhaps with some tweaks here and there it could work. Leaving this open to investigate in future.

ioleo: OK, I just tried it and it works now. Probably during the hackathon I had some types wrongly set in `expects` return type. Also, it seems like the type inference problem is gone now. :boom:

adamgfraser: Variance for the win!

test/shared/src/main/scala/zio/test/mock/Mock.scala
ioleo: Yes :)

ghostdogpr:
Looking good! I like how it reuses assertions 👍

test/shared/src/main/scala/zio/test/mock/Mock.scala
```diff
@@ -0,0 +1,212 @@
+package zio.test.mock
+
+import zio.{ IO, Promise, Ref, ZIO }
+
+trait Mock {
+
+  def invoke[E0, A0, A](method: Method[A, A0])(value: A): IO[E0, A0]
```
jdegoes: Maye we should rename these to `apply`? The overloads should work fine, then we'll have syntax like:

```scala
mock(Get)(1, 2, 3)
```

Which would be nicer than:

```scala
mock.invoke3(Get)(1, 2, 3)
```

test/shared/src/main/scala/zio/test/mock/MockSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+  final def zip[R1 <: R, E1 >: E, B](that: => MockSpec[R1, E1, B]): MockSpec[R1, E1, (A, B)] =
+    self.flatMap(a => that.map(b => (a, b)))
+
+  final def zipLeft[R1 <: R, E1 >: E, B](that: => MockSpec[R1, E1, B]): MockSpec[R1, E1, A] =
+    (self zip that).map(_._1)
+
+  final def zipRight[R1 <: R, E1 >: E, B](that: => MockSpec[R1, E1, B]): MockSpec[R1, E1, B] =
+    (self zip that).map(_._2)
+
+  final def <*[R1 <: R, E1 >: E, B](that: => MockSpec[R1, E1, B]): MockSpec[R1, E1, A] =
```
jdegoes: Operators go before the alpha-numeric ones.

test/shared/src/main/scala/zio/test/mock/Mock.scala
adamgfraser: Really nice for users not to have to worry about the arity of the methods they are calling.

test/shared/src/main/scala/zio/test/mock/MockException.scala
```diff
@@ -0,0 +1,15 @@
+package zio.test.mock
+
+import zio.test.{ AssertResult, Assertion }
+
+sealed trait MockException extends Throwable
+
+final case class PredicateFailedException(predicate: AssertResult) extends MockException
```
jdegoes: Can you place these two subtypes inside a companion object for `MockException`? That's the standard ZIO pattern (and a common Scala idiom).

test/shared/src/main/scala/zio/test/mock/MockException.scala
```diff
@@ -0,0 +1,15 @@
...
+
+final case class UnexpectedCallException(method: Method[Nothing, Any], args: Any) extends MockException
+
+final case class WrongCallException(
+  expected: Method[Nothing, Any],
```
jdegoes: You should make `AnyMethod`, a `sealed trait` that is extended only bby `Method[A, B]`, so you have a saner way to refer to these "any method" handles.

ioleo: Is there any reason to drop type information here? I made the exceptions parametrized in `A` and `B` and could drop type casts to `Method[Nothing, Any]` and everything still works.

ioleo:
Work in progress. Do not merge yet - I plan to add a few things today and tomorrow.

test/shared/src/test/scala/zio/test/TestMain.scala
```diff
@@ -1,12 +1,12 @@
...
+/*
 import scala.concurrent.ExecutionContext.Implicits.global
 
 import zio.test.mock._
 import zio.test.TestUtils.{ report, scope }
-
+ */
 object TestMain {
-
+  /*
```
jdegoes: Accidentally commented?

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
jdegoes: Maybe we should rename `Mock` to `Proxy`? This way, `MockRandom.Mock`, for example, can stay the same?

test/shared/src/test/scala/zio/test/TestMain.scala
ioleo: Temporarily. I'm testing `MockSpec` using ZIO Test (see `MockSpecSpec.scala`) and during development I have `~testJVM/test:run` running, but sbt could see two main classes and kept asking to choose one. Will restore this after I'm done.

ioleo:
Ready for another round of review. Still on TODO list:
- fix CI bugs
- add examples
- add docs
- add capability tags and mockable implementation for core modules (Clock, Console, etc)
- rename/move mock modules to fake/stub following the semantics defined in Martin Fowlers [post](https://martinfowler.com/articles/mocksArentStubs.html)

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
ioleo: I think we should keep the name `Mock` for this and rename current impl to `Fake` or `Stub` following the semantics defined [here](https://martinfowler.com/articles/mocksArentStubs.html).

ioleo: TLDR;

> - **Dummy** objects are passed around but never actually used. Usually they are just used to fill parameter lists.
> - **Fake** objects actually have working implementations, but usually take some shortcut which makes them not suitable for production (an in memory database is a good example).
> - **Stubs** provide canned answers to calls made during the test, usually not responding at all to anything outside what's programmed in for the test.
> - **Spies** are stubs that also record some information based on how they were called. One form of this might be an email service that records how many messages it was sent.
> - **Mocks** are what we are talking about here: objects pre-programmed with expectations which form a specification of the calls they are expected to receive.

adamgfraser:
Okay, will take a look tonight.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -144,9 +146,41 @@ object DefaultTestReporter {
...
+        ).mkString("\n")
+
+      case UnexpectedCallException(method, args) =>
+        List(
+          red(s"Unexpected call to $method"),
+          withOffset(tabSize)(s"with arguments ${blue(args.toString)}")
+        ).mkString("\n")
+    }
+
+  private def renderAssertResult(assertResult: AssertResult): String =
```
adamgfraser: Are you confident that there is never going to be an `AssertResult` with multiple failures (e.g. `p && q` where both `p` and `q` are false)? If there is one it is not going to match your pattern and you are not going to get any reporting. You may want to do something like:

```scala
private def renderAssertResult(assertResult: AssertResult): String =
  assertResult.failures.fold("")(
    _.fold(renderFragment(value, 0))(
      "(" + _ + " && " + _ + ")",
      "(" + _ + " || " + _ + ")",
      "~" + _
    )
  )
```

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -144,9 +146,41 @@ object DefaultTestReporter {
   private def renderCause(cause: Cause[Any], offset: Int): String =
     cause match {
```
adamgfraser: You should be able to just use `untraced` here and get rid of the `Cause.Traced` in your pattern match so you display the mock exceptions the way you want regardless of whether tracing is enabled.

ioleo:
> We should figure out what we want to do on the tests. (...) Regardless we should align one way or the other because it doesn't make much sense to have some tests written in ZIO Test and others written in the bootstrap testing framework.

As stated above, I think it's OK to test `zio.test` using itself as the test framework. Similar concept is largely estabilished in the compilers space, where once a compiler is able to compile itself it is considered a huge milestone and a testimony of its maturity.

I'd like to hear from others @iravid @ghostdogpr @jdegoes @neko-kai WDYT?

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
ioleo: That got me thinking, and I replaced `assertion.test(input)` with `assert(input, assertion)`, this way the exception holds `TestResult` instead of `AssertResult` and we can reuse the same rendering mechanism as in regular assertions.

adamgfraser: Nice!

adamgfraser: Nice!

test/shared/src/main/scala/zio/test/mock/MockableRandom.scala
```diff
@@ -0,0 +1,56 @@
...
+object MockableRandom {
+
+  object Service {
+    object NextBoolean       extends Method[Nothing, Boolean]
+    object NextBytes         extends Method[Int, Chunk[Byte]]
+    object NextDouble        extends Method[Nothing, Double]
+    object NextFloat         extends Method[Nothing, Float]
+    object NextGaussian      extends Method[Nothing, Double]
+    object NextInt           extends Method[Int, Int]
+    object NextInt1          extends Method[Nothing, Int]
```
ioleo: Hmm.. I'm wondering how to deal with method overloading. For the time being I just added a numeric suffix, but it's kinda arbitrary. Here the intuitive thing would be to use `NextInt` or `NextInt0` for the one without input and `NextInt1` for the one taking one input.... BUT:

```
def add(a: Long, b: Long): Long
def add(a: Int, b: Long): Long
def add(a: Int, b: Int): Long
```

are also valid method overloads, in which case the number of arguments does not help to uniquely identify the method.

It's not a problem for manualy created tags like here, but I'm wondering what strategy we should take with `Mockable` macro?

1) like above, simply add numeric suffix, even though the number has completely no meaning
* NextInt takes 1 param and NextInt1 takes 0 params
* but then NextLong takes 0 params and NextLong1 takes 1 param)

2) do not automatically handle overloaded methods with macro, add configuration param to the `@Mockable` annotation allowing the user to instruct the macro how to deal with overloads (this should still result in less boilerplate then manually writing the machinery)

@adamgfraser @jdegoes

test/shared/src/main/scala/zio/test/mock/Mockable.scala
```diff
@@ -0,0 +1,28 @@
...
+package zio.test.mock
+
+trait Mockable[Interface] {
+
+  def make(mock: Mock): Interface
+}
+
+object Mockable {
+
+  def apply[Interface](implicit mockable: Mockable[Interface]): Mockable[Interface] =
```
jdegoes: I think quite possibly we should not use the type class pattern for `Mockable`. That is, no implicits, make everything explicit. It doesn't really help us that much since you only need to provide a `Mockable` when a spec is run, not when it's constructed, so it's almost no savings.

Also the way the code is organized, the instances would not be found, anyway. So the user would need to import them implicitly, which is an anti-pattern.

test/shared/src/main/scala/zio/test/mock/MockableRandom.scala
jdegoes: @ioleo 

I'd just use numbers, probably. Yes, they are arbitrary, but a macro could allow this syntax:

```scala
expect(zio.clock.currentTime, ...)
```

then you can use Scala to choose the method and just look up which index you need for overloads (probabbly easier said than done!).

test/shared/src/test/scala/zio/test/mock/RandomSpec.scala
jdegoes: I am open to this but I am not sold on this level of precision because most people will not have read Fowler on the topic.

@adamgfraser Thoughts?

adamgfraser: I like the idea of having different names for the existing `Mock` implementations versus this new functionality because I do think they facilitate significantly different types of testing. For example, a "mock" implementation of `Clock` provided by this PR is going to behave very differently from the existing `MockClock` implementation. So it would be nice to have different names and possibly different packages within `test` to help us make this distinction and communicate it clearly to users in our documentation.

That said, I have a couple of concerns with the specifics of this framework:
- I think `Fake` and `Stub` names have negative connotations we want to avoid. I'm obviously somewhat biased here but I think the existing `Mock` implementations, especially `MockClock` and `MockRandom`, provide some nice functionality and would like to be able to talk about them. I think it is a lot easier to say "we have this awesome MockClock implementation" than we have an awesome "FakeClock" or "StubClock". Maybe it will actually make more sense to people who have read Fowler but a lot of people won't have.
- I think the full framework is too much complexity to educate users on. I think we can say there are two categories of things, the existing Mock implementations and this new mocking functionality, whatever we call them. I think the full five categories make things too complex, especially since some of our existing implementations cross the boundaries between the categories.
- We should at least be aware that if we change the names of the existing Mock implementations there is some cost to users and the education effort. There is no good time to change a name and sooner is probably better than later but we are just starting to educate people about using these things so changing the name right now is not the greatest.

My suggestion was to rename the existing `MockClock` and so on to `TestClock`. I realize this is a little overly general since everything we're doing has to do with testing, but I think it is a clear and non-perjorative and frees up the "mock" namespace for this new functionality.

iravid: I prefer `Test*` too.

iravid:
Re bootstrap testing: I suggest keeping the test suite for ZIO Test on the minimal testing framework, for as long as it isn't painful.

test/shared/src/main/scala/zio/test/mock/Mockable.scala
ioleo: Hmm I thought companion objects of type arguments (here `Interface`) should also be searched.
Will check if I can figure something out to fix this. If its not possible then I agree - no point in it being implicit.

ioleo: @jdegoes I was right, the problem was the `managedEnv[I]` contained an abstract type `I` that compiler had no idea how to fill (therefore it had no idea where to look for implicits). Fixed the issue in latest commit :grin:

test/shared/src/main/scala/zio/test/mock/MockableRandom.scala
ioleo: @jdegoes I'll add this as an enchancement proposal to macros project. For now I've decided to go with:
```
object overloadedMethod {
  object _0 extends Method[A, B]
  object _1 extends Method[B, B]
  object _2 extends Method[C, B]
  // ...
}
```

So for each overload we add a `_{0-index}`. This should be nicely discoverable by IDE if you just type `foo.bar.overloadedMethod` followed by dot, you should see a list of possible overloads.

ioleo: Added to `zio-macros` [issue #6](https://github.com/zio/zio-macros/issues/6#issuecomment-536263430).

ioleo:
Ready for another review. Left TODO:
- add examples
- add documentation
- rename `Mock*` implementations to `Test*` implementations

@jdegoes @iravid @adamgfraser

adamgfraser:
@ioleo Just a heads up we made minor changes to the ZIO Test internal test suite to allow running a single spec from the command line like `test:run MockSpecSpec` that are creating a merge conflict for you. Should be very easy to resolve and hopefully it is useful functionality for you but just wanted to let you know.

ioleo:
Rebased to master.

ioleo:
Rebased to master & squashed.

ioleo:
Last remaining task is adding some doc to `use cases`.

ioleo:
@jdegoes @ghostdogpr @iravid @adamgfraser I'd prefer to merge this now and have a new RC release, so I can depend on it in `zio-macros` project. I'll submit documentation in a seperate PR.

ghostdogpr:
@ioleo can you update this MD file though? https://github.com/zio/zio/blob/379532fe3c75d02eff334c9fcdb81e04ff55b326/docs/howto/test_effects.md
Just to apply the renaming. If we do an RC we need the docs to match the code 😄

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -513,6 +513,14 @@ object Assertion {
       actual.matches(regex)
   }
 
+  /**
+   * Makes a new assertion that requires the value to match specified partial function `f`.
+   */
+  final def matchesPattern[A](f: PartialFunction[A, Boolean]): Assertion[A] = Assertion.assertion("matchesPattern")() {
```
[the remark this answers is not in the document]
ioleo: Not anymore. I had used it at some point, but then I dropped it and forgot to cleanup. Nice catch.

examples/shared/src/test/scala/zio/examples/test/MockingExampleSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+
+object MockingExampleSpec
+    extends DefaultRunnableSpec(
+      suite("suite with mocks")(
+        testM("expect call returning output") {
+          val app = clock.nanoTime
+          val managedEnv: Managed[Nothing, MockClock] =
+            MockSpec.expectOut(MockClock.Service.nanoTime)(1000L)
+
+          managedEnv.use(app.provide _).map { result =>
```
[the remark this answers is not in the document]
ioleo: Didn't know this combinator. Loving it!

ioleo:
@adamgfraser Included all changes from your review. Regarding the docs, I see your point. I will try to provide a thorough documentation today (as a seperate `how to` page). It should be relatively easy for me to explain how this framework should be used. What will require further work is aligning the diffrent pages on testing and figuring out what message we want to deliver to ZIO users -- what is the recommended approach (for which scenario?).

In general, I believe for most cases where you want to assert something happened with the collaborator, you should use this mocking framework.

So, I propose I will add to this PR an isolated "how to" on mocking, and we work on smoothening the docs in the next PR. How does that sound @adamgfraser ?

ioleo:
Offending suite `ZIOSpec`

```
[info]   - RTS concurrency correctness
[info]     + shallow fork/join identity
[info]     + deep fork/join identity
[info]     + asyncPure creation is interruptible
[info]     + asyncInterrupt runs cancel token on interrupt
[info]     + supervising returns fiber refs
[info]     + supervising in unsupervised returns Nil
[info]     + supervise fibers
[info]     + supervise fibers in race
[info]     - supervise fibers in fork
```

Probably a flaky test?

adamgfraser:
Nice. Regarding the documents, they definitely don't need to be comprehensive at this point. I would love to just have a description and example of how someone could use the mocking framework to test an arbitrary collaborator, the "warehouse" from that article you shared with me. I was thinking it could go in the `test_effects.md` page and we could just build that out into a comprehensive user guide for ZIO Test over time but open to other thoughts.

I think the message I would have is that where the mocking framework really shines is its flexibility. If you are only using the standard ZIO environment types, using the test instances is probably going to be easier for you unless you really like the mocking style or want to use it consistently. We knew what those services were in advance and we wrote specialized test implementations for each of those services to make testing as easy as possible. But if you want to test anything else, especially anything with any degree of complexity, the mocking framework is really going to help you out because it is going to give you tools to specify expected inputs and outputs in a composable way without having to reimplement whatever it is you are trying to test.

The other thing I might emphasize is just that ZIO Test gives you the flexibility to use the testing style that works for you and what you are testing. I haven't gotten to work with the mocking framework yet but I've really been enjoying the way I can shift between writing conventional and property based tests seamlessly and I imagine the same thing would apply here.

It looks like there were some comments from the last review that didn't get resolved, maybe they were under the fold or something. Would you mind taking a look one more time? Getting close!

ioleo:
Sorry, didnt notice folded comments, indeed. Will get right into it.

test/shared/src/main/scala/zio/test/mock/MockClock.scala
```diff
@@ -16,333 +16,36 @@
...
- * Here we verify that no effect is performed before the recurrence period,
- * that an effect is performed after the recurrence period, and that the effect
- * is performed exactly once. The key thing to note here is that after each
- * recurrence the next recurrence is scheduled to occur at the appropriate time
- * in the future, so when we adjust the clock by 60 minutes exactly one value
- * is placed in the queue, and when we adjust the clock by another 60 minutes
- * exactly one more value is placed in the queue.
- */
-trait MockClock extends Clock with Scheduler {
+trait MockClock extends Clock {
```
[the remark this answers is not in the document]
ioleo: After adding the docs I will sketch out release notes for this feature :+1:

sideeffffect:
btw, here you can see the overview of failed tests on CircleCI
https://app.circleci.com/jobs/github/zio/zio/30354/tests

adamgfraser:
@sideeffffect That's really nice! Any idea why is seems to show the same failure twice?

jdegoes:
@ioleo @adamgfraser 

Personally, I have no real opinion on whether ZIO Test should be tested with ZIO Test itself. I think on the one hand, it makes it slightly more difficult for maintainers, because if one things break, many things break, and it can be hard to figure out what the root cause is; but on the other hand, it reduces code duplication and is a way of "dogfooding". Indeed, it is a sign of compiler maturity when a language can bootstrap itself, and this process generally increases contributions because the same people using the "language" can now contribute in the same "language".

My opinion is mainly that we should NOT mix two styles, that is, go all-in on one way or the other.

sideeffffect:
:man_shrugging: sorry, @adamgfraser no idea, I've just enabled this feature recently and I am no expert in CircleCI

ghostdogpr:
I have the same issue with bitbucket pipelines at work (results reported twice) so I guess the issue is in `test-reports`?

adamgfraser:
@ioleo One more substantive comment. Would it make sense to add sequence/traverse methods to `MockSpec`? Right now it is easy to expect `A, `B` , and then `C`. But if I want to expect a method to get called a hundred times there doesn't seem to be an easy way to do that other than implementing those combinators yourself. We could also definitely add in a subsequent PR.

adamgfraser:
@jdegoes Those are good points. I had liked the idea of having an "independent check" on ZIO Test but you are right that it would make it easier for contributors and would let us replace more ad hoc implementations of various testing functionality with the ones in ZIO Test. I also looked at some other testing frameworks and Scalacheck, Specs2, and Hedgehog all test against themselves (though Hedgehog has a [funny note](https://github.com/hedgehogqa/scala-hedgehog/blob/master/test/README.md) about this very issue).

Most of all I agree that we should only have one way of writing tests for ZIO Test. So if we are going to switch we should make a relatively concerted effort to move everything over. Maybe this could be a good issue for the next hackathon?

ioleo:
@ghostdogpr @adamgfraser @jdegoes Added documentation entries:
* `howto/module_pattern.md`
* `howto/mocking_services.md`

I'm not sure if I need to do something to have their links displayed in the left menu of "How to" section?
Please review.

Regarding the discussion:
* `add sequence/traverse` yes I intend to add more combinators, but this has already grown huge and I've already had to rebase it on several occasions (a lot of post-hackaton PRs getting merged :tada:), so lets leave this for future PRs (also, when it's "out" and we actually get to use it it will become more apparent which combinators we're missing)
* `testing ZIO Test with itself` I completely agree with @jdegoes and adding my 2 cents to this - during this PR I've had some issues (failing suites), however the "basic framework" output was not helpful, I had to add additional printlines and modify the test code a little to actually get an idea what went wrong - I know that ZIO Test's output is far more helpful (:heart:  @adamgfraser ) and had we used it I'd struggle less to get this resolved
* `^^^` but I believe this is an issue for another PR (if we agree to move to ZIO test testing itself)

ioleo:
@ghostdogpr regarding the release notes, here is a quick overview what has changed:
* moved package `zio.test.mock` to `zio.test.environment`
* renamed `MockConsole`, `MockClock`, `MockSystem` and `MockRandom` to `Test*` equivalents (within the moved package)
* added mocking framework under `zio.test.mock` package
* added rendering of `zio.test.mock.MockException` defects in `zio.test.DefaultTestReporter`

If you need more detailed info I'll be on gitter tomorrow (aka in ~8h from now).

ghostdogpr:
@ioleo you need to edit https://github.com/zio/zio/blob/master/website/sidebars.json for the sidebar.

I created a draft for v1.0.0-RC14 release notes with your information so that whoever does the release will have it.

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+    _    <- console.putStrLn("What is your name?")
+    name <- console.getStrLn.orDie
+    _    <- console.putStrLn(s"$name, your lucky number today is 42!")
+  } yield ()
+
+val result = combinedApp.provideManaged(combinedEnv)
+assertM(result, isUnit)
+```
+
+[doc-module-pattern](module_pattern.md)
```
ghostdogpr: Isn't the correct syntax `[doc-module-pattern]: module_pattern.md` ?

docs/howto/module_pattern.md
```diff
@@ -0,0 +1,285 @@
+---
+id: howto_module_pattern
+title:  "Module pattern"
```
ghostdogpr: How about having a title starting with a verb so that it sounds good when you read "How to ...". I think all the other sections are like that.

docs/howto/module_pattern.md
```diff
@@ -0,0 +1,285 @@
...
+
+Finally, the machinery behind _module pattern_ is simple Scala. There is **no magic behind the scenes**. The programmer is in full control and can rely
+on the compiler to hint him what environment he needs to provide to run the program he created.
+
+## Next Steps
+
+Depending on the role _service_ plays in the application you may need to either assert on values returned by _capacities_ or set expectations on
+_calls_ to the collaborating service. Luckily both approaches are very simple.
+
+First learn how to [test effecs][doc-test-effects] and once you are comfortable with testing effects, then the next step is to learn about
```
ghostdogpr: typo

ghostdogpr:
You can run `mdoc` and `docs/docusaurusCreateSite` in sbt to check locally that the docs are okay.

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -21,72 +21,72 @@ import zio.{ IO, ZIO }
 import zio.Managed
 
 /**
- * The `mock` package contains testable versions of all the standard ZIO
- * environment types through the [[MockClock]], [[MockConsole]],
- * [[MockSystem]], and [[MockRandom]] modules. See the documentation on the
+ * The `provided` package contains testable versions of all the standard ZIO
```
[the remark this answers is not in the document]
adamgfraser: I think we still need to get this one in.

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+id: howto_mocking
+title:  "Mocking services"
+---
+
+## How to test interactions between services?
+
+Whenever possible, we should strive to make our functions pure, which makes testing such function easy - you just need to assert on the return value.
+However in larger applications there is a need for intermediate layers that delegate the work to specialized services.
+
+For example, in an HTTP server the first layer of indirection are so called _routes_, whos job is to match the request and delegate the processing to
```
adamgfraser: "whose"

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+downstream layers. Often below there is a second layer of indirection, so called _controller_, which consits of several business logic units grouped
```
adamgfraser: "consists"

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+How can we test it did exacly what we expected it to do?
+
+## Mocking
+
+In this sort of situations we need mock implementations of our collaborator service. As _Martin Fowler_ puts it in his excellent article
+[Mocks Aren't Stubs][link-test-doubles]:
+
+> **Mocks** are what (...) objects pre-programmed with expectations which form a specification of the calls they are expected to receive.
```
adamgfraser: "Mocks are ... objects"?

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+
+ZIO Test provides a framework for mocking your modules.
+
+## Creating a mock service
+
+```scala mdoc:invisible
+trait AccountEvent
+```
+
+Assuming you are following the [module pattern][doc-module-pattern], for an `AccountListener` service we _capability tags_ and the accessor helper `>`
```
adamgfraser: Missing a verb here?

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+```
+implicit val mockable: Mockable[AccountObserver] = (mock: Mock) =>
+  new AccountObserver {
+    val accountObserver = new AccountObserver.Service[Any] {
+      def processEvent(event: AccountEvent): UIO[Unit] = mock(AccountObserver.Service.processEvent, event)
+    }
+  }
+```
+
+> **Note:** To make our mockable implementation automatically discovered, we need to place it inside `AccountObserver` modules companion object.
```
adamgfraser: "module's"

docs/howto/mocking_services.md
```diff
@@ -0,0 +1,273 @@
...
+## Setting up expectations
+
+Finally we're all set and can create ad-hoc mock environments with our services.
+
+```scala mdoc:silent
+import zio.test._
+import zio.test.Assertion._
+
+val event = new ProductEvent
+val app = ProductObserver.>.processEvent(event)
```
adamgfraser: Is it more clear to use `>` infix (e.g. `ProductObserver > processEvent(event)`)? I think I saw this a few other places so if you agree might be good to do a global find and replace

docs/howto/module_pattern.md
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+```
+
+```scala
+val result3 = app.provide(new EmailRendererLive with PostmanLive with NewsletterLive {})
+val result4 = app.provide(new PostmanLive with NewsletterLive with EmailRendererLive {})
+```
+
+This approach **allows for circual dependencies** between services. The environment is a single object with all the modules mixed in, it cannot be
```
adamgfraser: "circular"

adamgfraser:
@ioleo This looks great! Awesome job adding a ton of documentation. I think there is one comment from my previous review that is still outstanding and I left some minor comments on the documentation you added. Completely agree with you about adding the `sequence` and `traverse` methods in a subsequent PR. I'm a little concerned that the "how to" organizational structure is not going to scale as we get more  documentation, but I think that and migrating the ZIO Test internal test suite to ZIO test are issues for another day. From my end I'm good to merge once the latest round of comments are resolved.

ioleo:
Rebased & resolved conflicts. Will get to comments now.

docs/howto/mocking_services.md
ioleo: Yes :+1:

docs/howto/mocking_services.md
ioleo: Unfortunately as it is now, the infix notation does not work:
```scala
val app = ProductObserver > processEvent(event)
// raises compile error:
// ProductObserver.>.type does not take parameters
```

I'm not entirely sure what is going on there. Will research this later and (if possible) fix in another PR.

ioleo:
`mdoc` had failed on my examples as if it used old ZIO version (without the classes/traits from this PR), so I added the `:fail` tag to ignore the error

I think this can be merged now and we can deal with why was mdoc failing in a seperate PR.
@adamgfraser @ghostdogpr

docs/howto/use_module_pattern.md
```diff
@@ -0,0 +1,306 @@
...
+## Next Steps
+
+Depending on the role _service_ plays in the application, you may need to either assert on values returned by _capacities_ or set expectations on
+_calls_ to the collaborating service. Luckily both approaches are very simple.
+
+First learn how to [test effects][doc-test-effects] and once you are comfortable with testing effects, then the next step is to learn how to
+[mock services][doc-mock-services].
+
+[doc-test-effects]: test_effects.md
+[doc-mock-services]: mock_services.md
```
ghostdogpr: The links don't match no? like `[mock services]` vs `[doc-mock-services]`, etc

Issue #1790, opened 2019-09-22
A new TestAspect "only" to run single test in a Suite


dkarlinsky:
@adamgfraser It would be nice to have an aspect, called something like `only`, which when added to a single test in a suite, would case all other tests to be ignored. (Similar to `it.only' in mocha).
This could be very convenient during development, especially when debugging.

I think two aspects will be needed to pull this off. One on the outer suite that will do the filtering (ideally added to default aspects) and the `only` aspect that will mark the selected test somehow. 
The only way I see to mark a test in the `only` aspect,  is to add some magic mark to the label. (Unless we decide to add a notion of "tags" to `TestCase`)

What do you think @adamgfraser?
I would like to work on this, if this makes sense to you.

adamgfraser:
I think this would be a great feature to add. The challenge is how to implement it. Let me think about it, you can do the same, and we can use this issue to discuss ideas or feel free to open a PR if you have something you would like to use as a starting point. Would be great to work with you on this!

reibitto:
I was looking for this feature just now and I stumbled upon this issue. I like this idea.

I've been using the name matching approach up until now. The `-z` option in ScalaTest, and `-ex` in specs2:
https://etorreborre.github.io/specs2/guide/SPECS2-3.5/org.specs2.guide.Selection.html

These two approaches are ergonomically different, though I guess these features overlap enough that both don't need to be implemented?

adamgfraser:
@reibitto I actually think these are both good features that would be nice to have. They have similarities but also support slightly different styles / use cases. `@@ only` seems like it would be slightly more useful if, say, you knew you wanted to only run these tests for the next few hours. Whereas the command line option could be more useful if you did a patch change and just want to run one or two tests once to make sure they still work okay before submitting your PR.

The SBT Test Runner already supports running only a certain suite like `testOnly zio.ZManagedSpec` but doesn't support only running certain tests within a suite. If you were interested, a PR to add that feature would be very welcome.

@dkarlinsky wrote the SBT test runner and I believe there was some discussion of supporting more command line options at the time. Anything you would add here from your experience?

reibitto:
@adamgfraser Alright, thanks. I think I'll take a look at that then.

adamgfraser:
@reibitto Awesome!

dkarlinsky:
@adamgfraser I don't recall if there was actual discussion about adding a label filter command line argument. Maybe the possibility was just mentioned as "we may support this in the future"

[...]

PR #1838, opened 2019-09-26
#1790 - A suite annotated with `.only("some-label")`, will ignore all non matching labels

dkarlinsky:
New method `Spec.only()`, that when applied on a suite will ignore all tests that don't have the given substring in their label.\
Requires evidence that `L` to be String and the `Spec` is `ZSpec`

Example:
```scala
import zio.test.TestOnly._
suite("suite")(
    test("test1") {
      assert(1, equalTo(1))
    },
    test("test2") {
      assert(1, equalTo(1))
    }
  ).only("test2") 
```

~~This requires an outer suite to be marked with @@ thisOnlySupport.~~
~~Or `TeatAspects.runOnlySupport` needs to be added to default aspects.~~

~~To make this work I had to add a notion of `tags` to `TestCase`~~

~~@adamgfraser WDYT?~~


jdegoes:
I understand this is a useful scenario to support, I think we should reflect on how we can do this (and if we can, in the right way now), since in this design we cannot enforce the "outer suite" requirement.

adamgfraser:
One idea I had is once we get effectual suites in perhaps we can have a function like:

```scala
def testOnly[R, E, L, T](f: TestAspectPoly => Spec[R, E, L, T]): Spec[R, E, L, T] = {
  for {
    ref <- Ref.make[Vector[Spec[R, E, L, T]]
    only = new TestAspectPoly { ... }  // TestAspect that adds any tests it is called on to the ref
    _ <- f(only)
    specs <- ref.get
  } yield Spec.suite("only", specs, None)
```

Then you could use it like:

```scala
testOnly { only =>
  suite("MySuite")(
    test1,
    test2 @@ only,
    test3
  )
}
```

jdegoes:
@adamgfraser Yes, that's a much better API! It should work for now, and we can think about whether to do this with annotations later.

adamgfraser:
@dkarlinsky We just merged #1664 which makes suites effectual so I think now we really have the tools to implement your idea. Do you want to rework this PR along the lines I suggested? Would be great to get some feedback on working with the new effectual suites functionality and I think this would be a nice feature to have.

dkarlinsky:
Yes, I would like to work on this. I'm currently on vacation in Australia,
though :) I do have my laptop with me, so I'll try to get some work in, but
not sure how much.

adamgfraser:
@dkarlinsky Sounds good. No rush. Enjoy your vacation!

dkarlinsky:
Thanks!

jdegoes:
I will close this, @dkarlinsky, feel free to re-open when you have time to add this great feature!

dkarlinsky:
@jdegoes Back from vacation (landed yesterday) and now ready to give this another shot :) 
Looks like I don't have permissions to reopen the PR, though...

jdegoes:
Done!

dkarlinsky:
Thank you!

test/shared/src/main/scala/zio/test/TestOnly.scala
```diff
@@ -0,0 +1,88 @@
...
+     * ignores all tests that don't exactly match the given label.<br>
+     * If no matching labels are found, the test will fail.
+     */
+    def onlyWith(label: L): ZSpec[R, E, L, S] = only(_ == label)
+
+    /**
+     * ignores all tests with labels not satisfying the give predicate.<br>
+     * If no matching labels are found, the test will fail.
+     */
+    def only(p: L => Boolean): ZSpec[R, E, L, S] =
```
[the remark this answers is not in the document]
adamgfraser: @dkarlinsky You can do something like:

```scala
  final def only[S](
    s: String
  )(implicit ev1: L <:< String, ev2: T <:< Either[TestFailure[Nothing], TestSuccess[S]]): ZSpec[R, E, String, S] =
    mapLabel(ev1).mapTest(ev2).filterLabels(_.contains(s)).getOrElse(Spec.test("only", ignore))
```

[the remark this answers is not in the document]
dkarlinsky: Yes, this works! Thanks @adamgfraser and @jdegoes.
Useful trick to remember :)

test-tests/shared/src/test/scala/zio/test/TestOnlySpec.scala
```diff
@@ -0,0 +1,58 @@
+package zio.test
+
+import zio.test.Assertion._
+import zio.test.TestOnly._
+import zio.test.TestOnlySpecData._
+import zio.test.TestUtils.succeeded
+
+object TestOnlySpec
```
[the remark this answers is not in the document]
dkarlinsky: You probably mean `TestSpec` (didn't find a `SpecSpec`)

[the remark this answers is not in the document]
adamgfraser: I would just create a new file with that name. We don’t have other tests for spec but we probably should. 😃

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -28,6 +28,8 @@ import Spec._
  */
 final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E, L, T]]) { self =>
 
+  final def label: L = caseValue.label
```
adamgfraser: I think we can delete this. Looks like it isn't being used anywhere.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -312,10 +314,23 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
+    mapLabel(ev1)
+      .mapTest(ev2)
+      .mapError(ev3)
+      .filterLabels(_.contains(s))
+      .getOrElse(Spec.test("only", ignored))
 }
 
 object Spec {
   sealed trait SpecCase[-R, +E, +L, +T, +A] { self =>
+    val label: L
```
adamgfraser: We can delete this as well.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -312,10 +314,23 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
         val (z, caseValue) = f(z0, t)
         ZIO.succeed(z -> Spec(caseValue))
     }
+
+  /**
+   * Runs only tests whose labels (which must be strings) contain the given substring
+   */
+  final def only[S, E0](
```
adamgfraser: Would probably call this `E1` for consistency.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -312,10 +314,23 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
+    s: String
+  )(implicit ev1: L <:< String, ev2: T <:< TestSuccess[S], ev3: E <:< TestFailure[E0]): ZSpec[R, E0, String, S] =
+    mapLabel(ev1)
```
adamgfraser: If we have the implicit evidence we can use a cast to avoid doing multiple traversals of the `Spec`. `Spec.asInstanceOf[ZSpec[R, E1, String, S].filterLabels(_.contains(s)).getOrElse(Spec.test("only", ignored))`.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -312,6 +312,17 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
         val (z, caseValue) = f(z0, t)
         ZIO.succeed(z -> Spec(caseValue))
     }
+
+  /**
+   * Runs only tests whose labels (which must be strings) contain the given substring
+   */
+  final def only[S, E1](
+    s: String
+  )(implicit ev1: L <:< String, ev2: T <:< TestSuccess[S], ev3: E <:< TestFailure[E1]): ZSpec[R, E1, String, S] =
```
adamgfraser: Sorry, to be consistent can we flip the order here so the `TestFailure` evidence comes before the `TestSuccess` evidence?

test-tests/shared/src/test/scala/zio/test/SpecSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+import zio.test.Assertion.{ equalTo, isFalse, isTrue }
+import zio.test.TestOnlySpecData._
+import zio.test.TestUtils._
+
+object SpecSpec
+    extends ZIOBaseSpec(
+      suite("SpecSpec")(
+        suite(".only(pattern)")(
+          testM("ignores all tests except one matching the given label") {
+            zio.test.checkM(genSuite) { spec =>
```
adamgfraser: You can just do `checkM` now here and below.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -312,6 +312,17 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
         val (z, caseValue) = f(z0, t)
         ZIO.succeed(z -> Spec(caseValue))
     }
+
+  /**
+   * Runs only tests whose labels (which must be strings) contain the given substring
```
adamgfraser: We should probably clarify the contract with regard to suites. "Returns a new spec containing only specs with labels containing the specified string. If a suite label contains the specified string all specs in that suite will be included in the resulting spec. This method is only available on specs whose label type is `String`.

adamgfraser:
@dkarlinsky We merged #2192 which created the same test file and is creating a merge conflict. Should be easy to drop your tests in.

PR #1848, opened 2019-09-28
Simplify ZStream#flatMap using ZManaged.switchable

iravid:
And adding some missing ZManaged combinators.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1324,13 +1324,15 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self.fork >>= (_.toFutureWith(f))
 
   /**
-   * Converts this ZIO to [[zio.Managed]].
+   * Converts this ZIO to [[zio.Managed]]. This ZIO and the provided release action
+   * will be performed uninterruptibly.
    */
   final def toManaged[R1 <: R](release: A => ZIO[R1, Nothing, _]): ZManaged[R1, E, A] =
     ZManaged.make[R1, E, A](this)(release)
 
   /**
-   * Converts this ZIO to [[zio.ZManaged]] with no release action.
+   * Converts this ZIO to [[zio.ZManaged]] with no release action. It will be performed
+   * interruptibly.
    */
   final def toManaged_ : ZManaged[R, E, A] =
     ZManaged.fromEffect[R, E, A](this)
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -559,6 +559,19 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) 
   final def second[R1 <: R, A1 >: A]: ZManaged[R1, E, (R1, A1)] = ZManaged.identity[R1] &&& self
 
   /**
+   * Returns an effect that effectfully peeks at the acquired resource.
+   */
+  final def tap[R1 <: R, E1 >: E](f: A => ZManaged[R1, E1, _]): ZManaged[R1, E1, A] =
+    flatMap(a => f(a).as(a))
+
+  /**
+   * Like [[ZManaged#tap]], but uses a function that returns a ZIO value rather than a
+   * ZManaged value.
+   */
+  final def tapM[R1 <: R, E1 >: E](f: A => ZIO[R1, E1, _]): ZManaged[R1, E1, A] =
+    mapM(a => f(a).as(a))
+
+  /**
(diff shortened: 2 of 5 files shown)


iravid:
cc @vasilmkd who notes the similarity of ZManaged.switchable to flatMap :-)

mschuwalow:
Wow. This looks so much better :fireworks: 

One random thought: Couldn't we use the same idea in order to implement flatMapPar? We could keep a set of open streams around and pull from them in a round robin fashion / race them and keep extra results in a fixed size buffer. Once a stream is done we kick it out and open the next one.
This would also no longer require any queues :)
Old behavior with buffering of output could be easily restored using buffer.

I don't think switchable as is powerful enough to this, but I don't see why switchableN shouldn't be possible

iravid:
@mschuwalow 
Interesting that you bring this up :-) I looked into this when we switched over to `Pull`. My thought then is that racing was less efficient than forking off the streams and having them push into a queue. So this is definitely possible and possibly simpler; worth exploring if you have the time!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1324,13 +1324,15 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self.fork >>= (_.toFutureWith(f))
 
   /**
-   * Converts this ZIO to [[zio.Managed]].
+   * Converts this ZIO to [[zio.Managed]]. This ZIO and the provided release action
```
mschuwalow: This is very nice. Makes it far easier to understand interruption semantics without digging through ZManaged / bracket

mschuwalow:
> @mschuwalow
> Interesting that you bring this up :-) I looked into this when we switched over to `Pull`. My thought then is that racing was less efficient than forking off the streams and having them push into a queue. So this is definitely possible and possibly simpler; worth exploring if you have the time!

Cool. I'll play around with it and open a pr so we can see the differences. btw do we have any relevant benchmarks for this / do you have ideas what to use?

iravid:
No, I don't have anything smart to say about that. If you're using`flatMapPar` you're probably doing some sort of IO, so pure throughput is not very interesting. Allocation rate could be interesting.

iravid:
Any improvement on code complexity would be very welcome in any case. The combinator is super subtle currently.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -987,12 +1000,20 @@ object ZManaged {
...
    */
   final def fromEffect[R, E, A](fa: ZIO[R, E, A]): ZManaged[R, E, A] =
     ZManaged(IO.succeed(Reservation(fa, _ => IO.unit)))
 
+  /**
+   * Lifts a ZIO[R, E, A] into ZManaged[R, E, A] with no release action. The
+   * effect will be performed uninterruptibly. You usually want the [[ZManaged.fromEffect]]
+   * variant.
+   */
+  final def fromEffectUninterruptible[R, E, A](fa: ZIO[R, E, A]): ZManaged[R, E, A] =
```
vasilmkd: This reminds me of `Semaphore.withPermitManaged`. Cool.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1736,31 +1755,12 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
       .runCollect must_=== List()
   }
 
-  private def throttleEnforceShortCircuits = {
```
vasilmkd: Good that these are deleted.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -932,29 +932,19 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
   final def flatMap[R1 <: R, E1 >: E, B](f0: A => ZStream[R1, E1, B]): ZStream[R1, E1, B] =
```
vasilmkd: Finally we get to reason about `flatMap` on a higher level. Love it.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
iravid: Ah I had to get rid of something because of the `UTF8 string too long` error.

iravid:
Thanks for the review @mschuwalow and @vasilmkd!

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
vasilmkd: Finally, I thought it was just me and I was going crazy.

PR #1849, opened 2019-09-28
WIP: Update on the streaming.md file

almendar:
@iravid as we talked I've sketched the outline of the streaming document. I've started with describing internals of `ZStream`. I know it is not what benefits the end user most but I did it also for myself to get a better understanding when I'll be documenting other paragraphs. 

It will probably take me a week or even more to finish this but I'm sending early to get a feedback on what you would see yourself in such a document and catching my mistakes in description.

### DIFF
--- a/docs/usecases/streaming.md
+++ b/docs/usecases/streaming.md
@@ -3,4 +3,105 @@ id: usecases_streaming
 title:  "Streaming"
 ---
 
-Coming soon...
+All material on this pages assumes knowledge about the `ZIO[R, E, A]` type.
+
+## Why we need streams?
+TBD
+
+## How `zio-streams` are build
+
+The backbone of `zio-streams` are 3 types: `ZStream`, `Pull` and `ZSink`. 
+`ZStream[R, E, A]` represents the stream itself, which produces effectful values of type `A` and might end up with 
+error `E`. The `R` parameter has exactly the same meaning as as in the `ZIO` type that is to represent the environment
+requirements to run this stream.
+Second type, `Pull` should be understood as a handle by which consecutive effectful elements are being pulled into the
+stream (thus the name). `Pull`'s exact definition is a type alias for the `ZIO` type:
+`Pull[-R, +E, +A] = ZIO[R, Option[E], A]`. The contract states that each evaluation of the `Pull` type either produces a
+value `A` which will flow through the stream or produces an `Option[E]` which signals an actual error in `Some(_)` case 
+and flags the end of the stream in the`None` case. To emphasize this - the same `Pull` is being run multiple times to
+get items of the stream produced.
+
+How are `ZStream` and `Pull` related? `ZStream` is constructed based on the `Pull` type - passed as a constructor
+argument wrapped in `ZManaged`. Why the `ZManaged` indirection? Since our `Pull` contract says in most cases we need to assume it
+is not thread-safe so it should not be evaluated from multiple fibers. The evaluation also cannot be safely restarted after 
+interruption in all cases. When the `Pull` fails with a `None` it is not safe to evaluate it again. All of those seem
+very restrictive but keeping in mind that a stream will talk to external world possibly multiple times, which can be
+stateful, those assumptions are the only safe ones we can make. To give a concrete example, if our stream job is to
+interact with an external API that has time throttling of request it can accept from a single user then even if our 
+stream implementation does proper buffering and back-off in presence of two concurrent not coordinated clients both of
+them will probably end up with failures. Other example for getting `None` as error value. If our stream is responsible 
+for reading a file, `None` represents end of the file. Trying to pull more values after the file was wholly read makes
+no sense.
+
+The last type widely used in `zio-stream` is the `ZSink` type. Sink should be understood as a place that runs the stream
+and where the evaluated items are being passed down to. Sinks also have control over the evaluation of streams, possibly
+ending them earlier. `Sink[-R, +E, +A0, -A, +B]` contract states that sinks consume multiple values of type `A` coming
+from the `ZStream` type, producing either a single value of type B and reminders of type `A0` or an error of type `E`. 
+If we recall the `ZStream` and `Pull` definitions we can fast recognize that there is not space of "state" in both of
+them. This responsibility is being passed on to `ZSink` type. Example where we need such state would be a scenario where
+we want to count the number of elements that flow though the stream. Here we're moving out of the simple scenario of 
+transforming the incoming data into, keeping derived data based on the history of execution. Thus to support the need
+for such operators `ZSink` has a standardized interface to make this possible. First of all any sink may have some kind of
+state. In the general case `State` is a path-dependent type defined in `ZSink` that is hidden from the client of 
+the stream and might be anything that is needed so it is fully abstract. Next there are number of methods exposed on 
+`ZSink`. Let's enumerate them with simplified types:
+- `initial` - returns initial `state` of the stream which might be anything
+- `step` - is a function `(state, a) => state`. Step takes the current state, incoming item and calculates the next state.
+- `extract`: `state => (B, Chunk[A0])`, calculates the final value that is the result of running the stream and list of 
+  values that were not consumed.
+- `cont`: `state => boolean`, controls should the stream be evaluated further.
+
+Let's go through the `run` definition on `ZStream` which allows for connecting any `ZSink` to have a better 
+understanding how those two types work together.
+
+```scala
+def run[R1 <: R, E1 >: E, A0, A1 >: A, B](sink: ZSink[R1, E1, A0, A1, B]): ZIO[R1, E1, B] =
+    sink.initial.flatMap { initial =>
+      self.process.use { as =>
+        def pull(state: sink.State): ZIO[R1, E1, B] =
+          as.foldM(
+            {
+              case Some(e) => ZIO.fail(e)
+              case None    => sink.extract(state).map(_._1)
+            },
+            sink.step(state, _).flatMap { step =>
+              if (sink.cont(step)) pull(step)
+              else sink.extract(step).map(_._1)
+            }
+          )
+
+        pull(initial)
+      }
+    }
+```
+
+The line `sink.initial.flatMap { initial =>` extracts for us the initial state in which the sink is. The second line
+`self.process.use { as =>` exposes for us the `Pull` handle for generating values. At this point we need to read the
+values for as long as 
+- the stream doesn't end
+- there is no error
+- the `ZSink` decides that it doesn't want any more items
+
+We call `foldM` function on the `Pull` handle. First pair of parenthesis handles the erroneous case. For `Some(e)` there was
+an actual error during generation of value `A` and we finish with `ZIO` value set to error. For the `None` case there
+are no more values to generate so we should return what was already accumulated in the `ZSink`. For this purpose we're
+calling the `extract` method on sink which returns the `B` item and reminder which we throw away in this case as 
+irrelevant. The second pair of parenthesis handles the case where an item of type `A` was generated. We use this item
+to make progress in the sink calling the `step` method which generates for our our next state. That state is then used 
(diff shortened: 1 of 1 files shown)


docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
-Coming soon...
+All material on this pages assumes knowledge about the `ZIO[R, E, A]` type.
+
+## Why we need streams?
+TBD
+
+## How `zio-streams` are build
+
+The backbone of `zio-streams` are 3 types: `ZStream`, `Pull` and `ZSink`. 
+`ZStream[R, E, A]` represents the stream itself, which produces effectful values of type `A` and might end up with
```
pawelkaczor: What is "effectful value" ?

almendar: No problem, questions like this help me to write it in a more straight forward way.

I think the use of word 'effect' is unfortunate because of  confusion with side effects and those are different things, but well that's the terminology that people use so I'm sticking with it. 

Example:
We say that one can model absence of value when evaluating a function with 'Option'. And the fact that your function returns not plain A but Option[A] means your using effects to model computation. Another example would be using Either to model failure. So when you're returning Either[Error, A] from function you're modeling reporting errors with effect.

Back to your original question, our A is being created by modeling the computation with ZIO (allowing us to do whatever we want) so again we have an effect involved. Thus it is called effectful value as to get it we need to evaluate that effect.

pawelkaczor: Thanks. Still, I would expect the same terminology in both docs (see: https://zio.dev/docs/datatypes/datatypes_stream).

almendar: You mean this sentence?

"A Stream[E, A] represents an effectful stream that can produce values of type A, or potentially fail with a value of type E."?

That's a good point the docs should be consistent.
The word 'effect' is not a very well defined one I guess. When I was writing this I focused more on the internals and that each A comes from a separate evaluation. Will rephrase this.

jdegoes: Could be phrased, `effectfully produced value`, i.e. effects are required for the production of the value.

Roughly speaking, "functional effect" is the functional programming version of "side-effect", but in FP libraries, "functional effect" tends to be truncated to "effect", to avoid tedious repetition.

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
 title:  "Streaming"
```
iravid: @almendar How about we move the excellent breakdown you've written to the "Advanced" section?

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
 ---
 
-Coming soon...
+All material on this pages assumes knowledge about the `ZIO[R, E, A]` type.
+
+## Why we need streams?
```
iravid:
```suggestion
## Why do we need streams?
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+TBD
+
+## How `zio-streams` are build
```
iravid:
```suggestion
## How `zio-streams` are built
```

docs/usecases/streaming.md
iravid:
```suggestion
`ZStream[R, E, A]` represents the stream itself, which produces effectful values of type `A` and might fail with 
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+
+The backbone of `zio-streams` are 3 types: `ZStream`, `Pull` and `ZSink`. 
+`ZStream[R, E, A]` represents the stream itself, which produces effectful values of type `A` and might end up with 
+error `E`. The `R` parameter has exactly the same meaning as as in the `ZIO` type that is to represent the environment
```
iravid:
```suggestion
error `E`. The `R` parameter has exactly the same meaning as in the `ZIO` type and represents the environment
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+requirements to run this stream.
+Second type, `Pull` should be understood as a handle by which consecutive effectful elements are being pulled into the
```
iravid:
```suggestion
Second type, `Pull` should be understood as a handle by which consecutive elements can be pulled from the
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+stream (thus the name). `Pull`'s exact definition is a type alias for the `ZIO` type:
+`Pull[-R, +E, +A] = ZIO[R, Option[E], A]`. The contract states that each evaluation of the `Pull` type either produces a
+value `A` which will flow through the stream or produces an `Option[E]` which signals an actual error in `Some(_)` case 
+and flags the end of the stream in the`None` case. To emphasize this - the same `Pull` is being run multiple times to
```
iravid:
```suggestion
and flags the end of the stream in the `None` case. To emphasize this - the same `Pull` is being run multiple times to
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+get items of the stream produced.
+
+How are `ZStream` and `Pull` related? `ZStream` is constructed based on the `Pull` type - passed as a constructor
+argument wrapped in `ZManaged`. Why the `ZManaged` indirection? Since our `Pull` contract says in most cases we need to assume it
```
iravid: `ZManaged` is actually needed here to safely allocate and release resources. Would be good to note that.

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+is not thread-safe so it should not be evaluated from multiple fibers. The evaluation also cannot be safely restarted after 
+interruption in all cases. When the `Pull` fails with a `None` it is not safe to evaluate it again. All of those seem
+very restrictive but keeping in mind that a stream will talk to external world possibly multiple times, which can be
+stateful, those assumptions are the only safe ones we can make. To give a concrete example, if our stream job is to
```
iravid: I'd break this into another paragraph, starting at `To give a concrete example...`

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+interact with an external API that has time throttling of request it can accept from a single user then even if our 
+stream implementation does proper buffering and back-off in presence of two concurrent not coordinated clients both of
+them will probably end up with failures. Other example for getting `None` as error value. If our stream is responsible 
+for reading a file, `None` represents end of the file. Trying to pull more values after the file was wholly read makes
+no sense.
+
+The last type widely used in `zio-stream` is the `ZSink` type. Sink should be understood as a place that runs the stream
+and where the evaluated items are being passed down to. Sinks also have control over the evaluation of streams, possibly
+ending them earlier. `Sink[-R, +E, +A0, -A, +B]` contract states that sinks consume multiple values of type `A` coming
```
iravid:
```suggestion
ending them earlier. The `ZSink[-R, +E, +A0, -A, +B]` contract states that sinks consume multiple values of type `A`
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+from the `ZStream` type, producing either a single value of type B and reminders of type `A0` or an error of type `E`.
```
iravid:
```suggestion
and produce either a single value of type B and reminders of type `A0` or an error of type `E`. 
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+stream implementation does proper buffering and back-off in presence of two concurrent not coordinated clients both of
+them will probably end up with failures. Other example for getting `None` as error value. If our stream is responsible 
+for reading a file, `None` represents end of the file. Trying to pull more values after the file was wholly read makes
+no sense.
+
+The last type widely used in `zio-stream` is the `ZSink` type. Sink should be understood as a place that runs the stream
+and where the evaluated items are being passed down to. Sinks also have control over the evaluation of streams, possibly
+ending them earlier. `Sink[-R, +E, +A0, -A, +B]` contract states that sinks consume multiple values of type `A` coming
+from the `ZStream` type, producing either a single value of type B and reminders of type `A0` or an error of type `E`. 
+If we recall the `ZStream` and `Pull` definitions we can fast recognize that there is not space of "state" in both of
```
iravid:
```suggestion
If we recall the `ZStream` and `Pull` definitions we can fast recognize that there is no notion of "state" in either of
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+them. This responsibility is being passed on to `ZSink` type. Example where we need such state would be a scenario where
```
iravid:
```suggestion
them. This responsibility is passed on to the `ZSink` type. Example where we need such state would be a scenario where
```

docs/usecases/streaming.md
iravid: This is not entirely accurate because we have the `ZStream#mapAccum` combinator

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+interruption in all cases. When the `Pull` fails with a `None` it is not safe to evaluate it again. All of those seem
+very restrictive but keeping in mind that a stream will talk to external world possibly multiple times, which can be
+stateful, those assumptions are the only safe ones we can make. To give a concrete example, if our stream job is to
+interact with an external API that has time throttling of request it can accept from a single user then even if our 
+stream implementation does proper buffering and back-off in presence of two concurrent not coordinated clients both of
+them will probably end up with failures. Other example for getting `None` as error value. If our stream is responsible 
+for reading a file, `None` represents end of the file. Trying to pull more values after the file was wholly read makes
+no sense.
+
+The last type widely used in `zio-stream` is the `ZSink` type. Sink should be understood as a place that runs the stream
```
iravid: Let's add a small headline here:
```suggestion
### ZSink

The last type widely used in `zio-stream` is the `ZSink` type. Sink should be understood as a place that runs the stream
```

docs/usecases/streaming.md
```diff
@@ -3,4 +3,105 @@ id: usecases_streaming
...
+state. In the general case `State` is a path-dependent type defined in `ZSink` that is hidden from the client of 
+the stream and might be anything that is needed so it is fully abstract. Next there are number of methods exposed on 
+`ZSink`. Let's enumerate them with simplified types:
+- `initial` - returns initial `state` of the stream which might be anything
+- `step` - is a function `(state, a) => state`. Step takes the current state, incoming item and calculates the next state.
+- `extract`: `state => (B, Chunk[A0])`, calculates the final value that is the result of running the stream and list of 
+  values that were not consumed.
+- `cont`: `state => boolean`, controls should the stream be evaluated further.
+
+Let's go through the `run` definition on `ZStream` which allows for connecting any `ZSink` to have a better
```
iravid: I'm a bit wary of going through the definition of `run` here because it will undoubtedly change. Maybe it'd be better to conceptually describe how `run` consumes a stream using a sink?

almendar:
@iravid in other PR we decided to remove Advanced https://github.com/zio/zio/pull/1800/files#diff-cb1f944b76e14c5b994bf7231b98104aL59

I haven't dropped work on this, just thin on time letely. I should be able to get back to this next week.

iravid:
Ah, ok. So in that case I'd move the breakdown of ZStream to its page in the Data Types section, and similarly to sink.

You can put the breakdown in a separate section called `In-depth details` or something similar at the bottom of the pages.

iravid:
Hey @almendar - I'm closing this one out for now. Very interested in it, so please re-open when you get the time to take it over the finish line! Thanks again for picking it up!

PR #1850, opened 2019-09-29
Fix finalizer ordering on ZStream#flatMap

iravid:
The recent change to ZStream#flatMap caused the next stream to be allocated before the previous stream was deallocated. This is fixed by using ZManaged#withEarlyRelease for explicit deallocation before pulling the next `A`.

@vasilmkd


streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -933,18 +888,37 @@ class StreamSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRu
...
         _      <- latch.await
         _      <- fiber.interrupt
         result <- effects.get
-      } yield result must_=== List(3, 2, 1)
+      } yield result must_=== List(3, 3, 2, 1, 1)
+
+    unsafeRun(test)
+  }
+
+  private def flatMapFinalizerOrdering = {
```
vasilmkd: Great test.

Issue #1845, opened 2019-09-27
Improve ZStream/ZStreamChunk composition


iravid:
Currently, mixing ZStream and ZStreamChunk can be a bit clunky.

For example, this doesn't work:
```scala
Stream.fromEffect(someEffect) *> aStreamOfChunks
```

It's possible that adding a few overloads of the sequencing operators will solve this. Happy to hear other thoughts!

vasilmkd:
Just off the top of my head, would an implicit syntax class that exposes the inner stream of chunks work?

Twizty:
Hi, I'd like to implement it.

vasilmkd:
Go for it!

vasilmkd:
@iravid Here's an idea. Can we make `ZStreamChunk` be a syntax class over a stream of chunks?

vasilmkd:
@iravid Maybe not the most popular idea, but worth a discussion: https://docs.oracle.com/javase/8/docs/api/java/util/stream/BaseStream.html.

iravid:
Closing as this is not the right approach to the problem.

PR #1857, opened 2019-09-30
#1845 Added methods to compose ZStream with ZStreamChunk

Twizty:
Implemented #1845.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SteamWithChunksCombinationsSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SteamWithChunksCombinationsSpec.scala
@@ -3,9 +3,12 @@ package zio.stream
 import org.specs2.ScalaCheck
 import zio._
 
-import scala.{Stream => _}
+import scala.{ Stream => _ }
 
-class SteamWithChunksCombinationsSpec(implicit ee: org.specs2.concurrent.ExecutionEnv) extends TestRuntime with GenIO with ScalaCheck {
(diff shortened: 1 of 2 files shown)


iravid:
@jdegoes Could I get your opinion on this? Not sure what the right approach should be here

jdegoes:
@iravid Looking now...

jdegoes:
The main problem is that by introducing these few methods, we are setting up the user for some inconsistency problems:

1. First, that the methods are not universal: only a few cases have been covered, not all.
2. Second, that the methods aren't symmetrical. Ideally, users who can do `a op b` can also do `b op a`, but that will not be the case in this PR.

We could, using type classes or implicits or even subtyping (with more type parameters!), ensure all methods can be called on all combinations. That seems overkill given the massive impact on the API.

Ultimately a combinator approach may be more useful.

First, we want to carrot toward chunking the non-chunked side, so at least one side is chunked. If a user could type:

```scala
stream1.chunked zip stream2
```

Well, that's almost as good as `stream1 zip stream2`, isn't it?

Now such naive chunking may work fine, if it doesn't, you could do:

```scala
stream1.chunkWith(stream2)(_ zip _)
```

where you would then chunk `stream1` based on `stream2` chunk sizes. But that's more cumbersome / tedious to write, so perhaps we could just start off encouraging the first one?

What do people think?

iravid:
Thanks for your work on this @Twizty. I'm closing this one out as after some thought, it's not the right approach to the problem; we should make the translation between `ZStream/ZStreamChunk` easy, but not allow for sequencing between the types.

PR #1860, opened 2019-09-30
Add ZStream#toQueueUnbounded

vasilmkd:
Forked off from #1685.

@iravid

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1996,12 +1996,22 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
     }
 
   /**
-   * Converts the stream to a managed queue. After managed queue is used, the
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1996,15 +1996,25 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
     for {
       queue <- ZManaged.make(Queue.bounded[Take[E1, A1]](capacity))(_.shutdown)
       _     <- self.intoManaged(queue).fork
     } yield queue
 
+  /**
+   * Converts the stream into an unbounded managed queue. After the managed queue
+   * is used, the queue will never again produce values and should be discarded.
+   */
+  final def toUnboundedQueue[E1 >: E, A1 >: A]: ZManaged[R, Nothing, Queue[Take[E1, A1]]] =
```
iravid: Let’s make this a bit more completion friendly and name it `toQueueUnbounded`?

PR #1862, opened 2019-09-30
Migrate RTSSpec to zio-test

mijicd:
Part of #1647.


adamgfraser:
Can we move this to the `shared` folder so tests run on both platforms?

mijicd:
@adamgfraser Tried it moments ago, forgot about blocking-related tests which kinda spoil the happiness :).

ghostdogpr:
Ideally (but it could be done in a later PR if this is already enough), those tests would be moved to either `ZIOSpec` and `ZIOSpecJvm` when they only make sense on JVM.

It was previously attempted in https://github.com/zio/zio/pull/1641#issuecomment-530715911 but reverted because specs2 used a single string to define all tests and it was too big. This problem won't happen with ZIO Test.

mijicd:
@ghostdogpr shall we wait for `ZIOSpecJvm` to be migrated to zio test?

ghostdogpr:
Oops, it's not migrated yet. Maybe move the non-blocking ones to `ZIOSpec` now, and keep the rest in `RTSSpec`, and when there's the PR for `ZIOSpecJvm` we can ask the author to merge `RTSSpec` back in.

mijicd:
@ghostdogpr done.

PR #1863, opened 2019-09-30
Allow to specify a single spec of all test/test to be executed

Jentsch:
Invocation example: testJVM/test:run GenSpec

Throws an exception if more than one argument is given or the single one argument doesn't match a spec name.

This is a split of #1756 

Question: where should this behaviour / feature be documented?

### DIFF
--- a/test/shared/src/test/scala/zio/test/TestMain.scala
+++ b/test/shared/src/test/scala/zio/test/TestMain.scala
@@ -8,25 +8,41 @@ import zio.test.TestUtils.{ report, scope }
 object TestMain {
 
   def main(args: Array[String]): Unit = {
-    val testResults = List(
-      scope(AssertionSpec.run, "AssertionSpec"),
-      scope(BoolAlgebraSpec.run, "BoolAlgebraSpec"),
-      scope(CheckSpec.run, "CheckSpec"),
-      scope(ClockSpec.run, "ClockSpec"),
-      scope(ConsoleSpec.run, "ConsoleSpec"),
-      scope(DefaultTestReporterSpec.run, "DefaultTestReporterSpec"),
-      scope(EnvironmentSpec.run, "EnvironmentSpec"),
-      scope(FunSpec.run, "FunSpec"),
-      scope(GenSpec.run, "GenSpec"),
-      scope(GenZIOSpec.run, "GenZIOSpec"),
-      scope(LiveSpec.run, "LiveSpec"),
-      scope(RandomSpec.run, "RandomSpec"),
-      scope(SampleSpec.run, "SampleSpec"),
-      scope(SchedulerSpec.run, "SchedulerSpec"),
-      scope(SystemSpec.run, "SystemSpec"),
-      scope(TestAspectSpec.run, "TestAspectSpec"),
-      scope(TestSpec.run, "TestSpec")
+    val allTests: List[(String, BaseSpec)] = List(
+      ("AssertionSpec", AssertionSpec),
+      ("BoolAlgebraSpec", AssertionSpec),
+      ("CheckSpec", CheckSpec),
+      ("ClockSpec", ClockSpec),
+      ("ConsoleSpec", ConsoleSpec),
+      ("DefaultTestReporterSpec", DefaultTestReporterSpec),
+      ("EnvironmentSpec", EnvironmentSpec),
+      ("FunSpec", FunSpec),
+      ("GenSpec", GenSpec),
+      ("GenZIOSpec", GenZIOSpec),
+      ("LiveSpec", LiveSpec),
+      ("RandomSpec", RandomSpec),
+      ("SampleSpec", SampleSpec),
+      ("SchedulerSpec", SchedulerSpec),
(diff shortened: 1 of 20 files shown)


adamgfraser:
@Jentsch Great question! I would put it in the `contributing.md` file.

Jentsch:
@adamgfraser sure about the `contributing.md`? Nothing in there talks about how to run the specs of the specs. So I would add a new paragraph about `testJVM/test:run`.

adamgfraser:
@Jentsch Yeah, I think if it is going to go anywhere that is where it is going to have to live. This is a feature to make things easier for people who are contributing to ZIO, and in particular ZIO Test. As you say, the issue is that we haven't updated the contributor's guide in a while so it doesn't even talk about the fact that tests should be written using ZIO Test, best practices for doing that, how tests for ZIO Test itself should be written and how to run those tests, etc...

If you want I'm fine with just adding the feature right now and then we can open a separate issue to update the contributor's guide more broadly. Whichever you prefer.

Jentsch:
Yeah, I would prefer that. I scrolled through the `contributing.md` file and found no good place to add this peace of information.

adamgfraser:
Sounds good.

test/shared/src/test/scala/zio/test/BaseSpec.scala
```diff
@@ -0,0 +1,5 @@
+package zio.test
+
+trait BaseSpec {
```
adamgfraser: Can we have `BaseSpec` extend `DefaultRuntime` so Specs just have to extend one thing?

test/shared/src/test/scala/zio/test/TestMain.scala
```diff
@@ -4,29 +4,44 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
-      scope(GenSpec.run, "GenSpec"),
-      scope(GenZIOSpec.run, "GenZIOSpec"),
-      scope(LiveSpec.run, "LiveSpec"),
-      scope(RandomSpec.run, "RandomSpec"),
-      scope(SampleSpec.run, "SampleSpec"),
-      scope(SchedulerSpec.run, "SchedulerSpec"),
-      scope(SystemSpec.run, "SystemSpec"),
-      scope(TestAspectSpec.run, "TestAspectSpec"),
-      scope(TestSpec.run, "TestSpec")
+    val allTests: ListMap[String, BaseSpec] = ListMap(
```
adamgfraser: Can we just use a `List` here and do `List.filter(_._2 == selection)`?

Jentsch: Sure, just though this list has the semantic of a map, so maybe I should use one. But `ListMap` isn't that commonly known and used and it doesn't provide any big benefit in the code. I'll revert it back to a list.

test/shared/src/test/scala/zio/test/BaseSpec.scala
Jentsch: Sure, but three Specs don't extend `DefaultRuntime` (`AssertionSpec`, `GenSpec` and `GenZIOSpec`). If it's okay that they inherit the `DefaultRuntime`, than sure I can let `BaseSpec` extends `DefaultRuntime`.

An other option would be to have a third trait that extends both `DefaultRuntime` and `BaseSpec`, but in tests I rather prefer a bit of duplication.

Jentsch:
@adamgfraser can you trigger a re-run on circle-ci? The linter catches me often, but this times it looks like a network hiccup.

test/shared/src/test/scala/zio/test/BaseSpec.scala
adamgfraser: Yes I think it is fine to have them all extent `DefaultRuntime`. It is a little functionality they won't need but I think that is outweighed by the benefit of just having one thing contributors need to extend when adding new Specs. Could we also rename it `ZIOBaseSpec`? That should trigger the CI to run again as well.

Jentsch: Done

Jentsch:
The name `ZIOBaseSpec` is already in use (face-palm). @adamgfraser  Any suggestions for an other name?

adamgfraser:
Yeah, I was just looking at that. I think it is a symptom of another issue though. The other use is in the tests for the core project. But the tests for the core project should depend on ZIO Test's source files, not its test files. So I'm looking at the `build.sbt` file now.

adamgfraser:
Okay, so after fixing that apparently there are some places where the tests for ZIO Core are depending on code in the ZIO internal test suite, which they shouldn't be doing, but we aren't going to fix that as part of this PR. Let's go back to `BaseSpec` and get this in. Sorry about that!

adamgfraser:
Congratulations on your first contribution! Hopefully more to come soon!

Jentsch:
Not my first contribution, but some more useful than before. My next contribution already waits: #1864, I mad a mistake during the last renaming.

adamgfraser:
Sorry!

PR #1866, opened 2019-10-01
Fix surprising behaviour of promise complete

Jentsch:
Before this change using the `complete` of a `Promise` would store the provided `io` and every user of `await` would reevaluate this `io`, including all side effects. For promises this is an unusual behaviour.
With this changes `complete` evaluates the given `io` immediately and stores the `Exit` value.

Open questions:

- The new `done` code is the old `complete` code, but should be intensely reviewed.
- Maybe `await` are now slower, because it now wraps the given `k` in a new closure. What would be good benchmarks to show this? Similar is true for `complete`.
- Maybe `poll` should now return now an `UIO[Option[Exit[E, A]]]`
- Last but not least: is this the right solution?

Found while discussing #1753


Jentsch:
The failure of circleci looks like a race condition. The tested code uses Promises, but it should be independent of the version of Scala and the JDK. I suspect a re-run 'solves' the problem.

core-tests/shared/src/test/scala/zio/PromiseSpec.scala
```diff
@@ -88,6 +88,18 @@ object PromiseSpec
...
           } yield assert(d, isTrue)
+        },
+        testM("complete evaluates the provided io immediately") {
+          for {
+            p <- Promise.make[Nothing, Int]
+            r <- zio.Ref.make(0)
+            _ <- p.complete(r.update(_ + 1))
+            u <- r.get
+            d <- p.await
+            e <- p.await
```
LGLO: Could you add  `u2 <- r.get` here and assertion to make sure effect wasn't executed twice?

jdegoes:
@Jentsch This is not surprising to me given that ZIO's promise is the only one based on continuations. The majority of people will interact with this through `succeed` et al, and you can also achieve the "memoized" semantic by doing something like `value.memoize.flatMap(promise.complete(_))`. 

Would more documentation be useful fix against this, including perhaps a helper function to call `memoize` directly?

LGLO:
@Jentsch because I was surprised as well, I decided to make changes in docs. Could you please take a look at #1877 ?

jdegoes:
Could we potentially add `completeOnce` or `completeMemoized` for the "normal" use case?

mlangc:
This behavior just surprised me quite a bit as well. What about adding `succeedM(io: IO[E, A])` for the "normal" use case?

adamgfraser:
@Jentsch Just a heads up we merged #1954 to address this by renaming the previous implementation to `Promise#completeWith` and implementing a new `Promise#complete` combinator that completes a promise with the result of an effect rather than the effect itself. Thanks for raising the issue!

PR #1870, opened 2019-10-01
ZIO.option do not catch throwable

### LINKED ISSUE #1868: ZIO.option shadows unchecked exceptions
```scala
@ r.unsafeRun(ZIO.dieMessage("oops").option)
res5: Option[Nothing] = None
```

In my opinion recovering from unchecked exceptions should be explicit and can be achieved by user with method `.sandbox`:
```scala
ZIO.dieMessage("oops").sandbox.option
```

darl:
fixes #1868

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -646,7 +646,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Executes this effect, skipping the error but returning optionally the success.
    */
   final def option: ZIO[R, Nothing, Option[A]] =
-    self.foldCauseM(_ => IO.succeed(None), a => IO.succeed(Some(a)))
+    self.foldM(_ => IO.succeed(None), a => IO.succeed(Some(a)))
 
   final def optional[E1](implicit ev: E <:< Option[E1]): ZIO[R, E1, Option[A]] =
     self.foldM(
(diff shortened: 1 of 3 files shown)


iravid:
Thanks @darl. This makes sense to me. Let's wait for another reviewer's opinion. Could you resolve conflicts and run a format in the mean time?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -646,7 +646,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Executes this effect, skipping the error but returning optionally the success.
    */
   final def option: ZIO[R, Nothing, Option[A]] =
-    self.foldCauseM(_ => IO.succeed(None), a => IO.succeed(Some(a)))
+    self.foldM(_ => IO.succeed(None), a => IO.succeed(Some(a)))
```
jdegoes: This looks good, can you make sure `either` is consistent with this?

adamgfraser: Just checked and `either` is implemented in terms of `foldM` versus `foldCauseM`, so I think we are good here. Also reinforces that this is the right way to go as these methods should be consistent.

jdegoes:
@darl Thank you for your work on this!

darl:
👍 
p.s. looking for another free issue ;-)

adamgfraser:
@darl I'm looking for help on #1898 if you are interested.

darl:
@adamgfraser i'll look into it

PR #1871, opened 2019-10-01
Add ZSchedule.elapsed example to docs

reibitto:
Since I struggled with this one, I thought it would be helpful to include at least one example of a `ZSchedule`-specific method (such as needing `Clock`) so that people know to look at the additional methods in `ZSchedule` and not just `Schedule`.

### DIFF
--- a/docs/datatypes/schedule.md
+++ b/docs/datatypes/schedule.md
@@ -96,3 +96,9 @@ recur, using the minimum of the two delays between recurrences:
 ```scala mdoc:silent
 val expCapped = Schedule.exponential(100.milliseconds) || Schedule.spaced(1.second)
 ```
+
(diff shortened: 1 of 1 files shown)


docs/datatypes/schedule.md
```diff
@@ -96,3 +96,9 @@ recur, using the minimum of the two delays between recurrences:
 ```scala mdoc:silent
 val expCapped = Schedule.exponential(100.milliseconds) || Schedule.spaced(1.second)
 ```
+
+Stops trying to retry after a specified amount of time has elapsed:
```
LGLO: Maybe just `Stops retrying after...`? WDYT?

reibitto: Yeah, I like that better. Updated.

PR #1879, opened 2019-10-02
resolve #1847

### LINKED ISSUE #1847: Allow custom behavior on inheritFiberRefs?
It might be interesting to allow custom behavior for FiberRefs when joining a fiber.
For example one could instrument sleeps very nicely with a fiberref that tracks how long a fiber has slept and on join the maximum of both values is used.

The default case should be to just use the value of the joined fibers, as is currently done.

Looking at FiberContext I believe I understand how this could be implemented.

mschuwalow:
/cc @adamgfraser @jdegoes

### DIFF
--- a/core/shared/src/main/scala/zio/FiberRef.scala
+++ b/core/shared/src/main/scala/zio/FiberRef.scala
@@ -29,9 +29,9 @@ package zio
  *
  * `result` will be equal to "Hi!" as changes done by child were merged on join.
  *
- * FiberRef also forms a monoid using the combine function provided in FiberRef#make.
- * This will be used to merge the values on join and by default will just use the value
- * of the joined fiber.
+ * FiberRef also forms a semigroup using the combine function provided in FiberRef#make.
+ * It determines how values of two fibers will combine when joining. By default
+ * the value of the joined fiber will be used.
  * {{{
  * for {
  *   fiberRef <- FiberRef.make(0, math.max)
--- a/core/shared/src/main/scala/zio/FiberRef.scala
+++ b/core/shared/src/main/scala/zio/FiberRef.scala
@@ -29,10 +29,25 @@ package zio
  *
  * `result` will be equal to "Hi!" as changes done by child were merged on join.
  *
+ * FiberRef also forms a monoid using the combine function provided in FiberRef#make.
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/FiberRef.scala
```diff
@@ -29,10 +29,25 @@ package zio
  *
  * `result` will be equal to "Hi!" as changes done by child were merged on join.
  *
+ * FiberRef also forms a monoid using the combine function provided in FiberRef#make.
```
adamgfraser: Does the operation need to form a `Monoid` or just a `Semigroup`? It seems like we need an associative operation `combine` but don't actually need a `zero` value since the `FiberRef` will always be filled with an initial value on creation and children will inherit that. For instance, in all your examples you use `0` as the value but that isn't actually a `zero` value for `max` over the integers. I don't think it matters for the code but just how we describe it.

jdegoes: I'd say it forms (or should form) a monoid with the initial value as the 'zero'. Obviously we can't enforce that and to avoid scaring people away, the monoid thing should be a footnote rather than a headliner.

Also the examples we use should probably use the "real" zero for the provided combine

adamgfraser: What if, say, we want to create a `FiberRef` with an initial value of one and addition as the operation to count the number of fibers that an effect creates?

adamgfraser: Also our default implementation of `combine` is `last` which doesn't have a `zero`.

jdegoes: Good point. Maybe just leave this unspecified. It's more general than a monoid, but well-defined really only for monoids.

mschuwalow: You are right. I was thinking of the initial value as the zero value for monoid when writing this, but this does not work as it does not always serve as the identity. I'll update

mschuwalow: Oh, I updated on a stale version of the comments, this was directly replying to @adamgfraser initial comment. I'll take this out of the docs

PR #1881, opened 2019-10-02
Add Stream.chunked

mvv:
A straightforward way to group elements into `Chunk`s


mvv:
Oh, you already have the #1795 )
But this implementation seems to be more efficient.

mvv:
Test failure is unrelated to the changes (`BlockingSpec`)

iravid:
Thanks @mvv! I'll get #1795 merged, then we can replace the implementation there with this implementation.

Could you call the method `chunkN`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1932,6 +1932,35 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+   * Groups elements into chunks of the specified size.
+   */
+  final def chunked(chunkSize: Int): ZStream[R, E, Chunk[A]] =
+    ZStream {
+      self.process.mapM { as =>
+        Ref.make[Option[Option[E]]](None).flatMap { stateRef =>
+          def loop(chunk: Chunk[A]): Pull[R, E, Chunk[A]] =
+            as.foldM(
+              success = { a =>
+                val chunk1 = chunk ++ Chunk.single(a)
```
iravid: This would result in a left-associated "tree" of concats. Not the best structure for a chunk! Since the purpose of this implementation is to be performant, we can alternatively:
- preallocate an array when entering `loop`, set elements on it when pulling upstream and use `Chunk.fromArray` / `Chunk.fromArray(arr).slice(...)`
- Use `Pull.emit(chunk1.materialize)` which would copy the elements of the concatenated chunks into an array.

mvv: The first option, although it is the fastest, would require a `ClassTag[A]`, so I chose the latter. Using a `List` would also work, not sure which one is slower, `Chunk.fromIterable(List.reverse)` or `Chunk.materialize`.

mvv:
@iravid Merged master and re-implemented on top of arrays using `Chunk.Tags.fromValue`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -669,8 +669,45 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                stateRef.set(Some(e)) *> Pull.emit(Chunk.fromArray(acc).take(i))
+              }
+            )
+          def first: Pull[R, E, Chunk[A]] =
+            as.foldM(
+              success = { a =>
+                if (chunkSize <= 1) {
+                  Pull.emit(Chunk.single(a))
+                } else {
+                  val acc = Array.ofDim(chunkSize)(Chunk.Tags.fromValue(a))
```
iravid: Ah that's clever! Nicely done

PR #1893, opened 2019-10-03
ZIO Test: Implement Multiple Parameter Function Generators

### LINKED ISSUE #1889: ZIO test: multiple parameters function generators
Right now it is easy to write generators for functions of one argument, but not more. This result in having to write tests as follows:
```scala
check(Gen.function[R, (A1, A2), B]) { f =>
  (??? zipWith ???)(Function.untupled(f))
}
```

Perhaps we need to add `Gen.function[2|3|4|5]`.

adamgfraser:
Resolves #1889.


regiskuckaertz:
that was fast 🤯 thank you @adamgfraser 🙏

PR #1901, opened 2019-10-04
Add `Gen.anyUnicodeChar`

regiskuckaertz:
Some values in the `[Char.MinValue .. Char.MaxValue]` range are not valid Unicode characters. This was detected in the `utf8Decode` tests in the `ZSink` spec [here](https://github.com/zio/zio/pull/1887/files).

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -143,8 +143,26 @@ object Gen extends GenZIO with FunctionVariants {
   /**
    * A generator of strings. Shrinks towards the empty string.
    */
-  final val anyString: Gen[Random with Sized, String] =
-    Gen.string(Gen.anyChar)
+  final def anyString: Gen[Random with Sized, String] =
+    Gen.string(Gen.anyUnicodeChar)
+
+  /**
+   * A generator of Unicode characters. Shrinks toward '0'.
+   * Heavily inspired from:
+   * https://github.com/typelevel/scalacheck/blob/ab15a9fe012ca7c9feb48ee188e4d167df76d6ba/src/main/scala/org/scalacheck/Arbitrary.scala#L122-L134
+   */
+  final val anyUnicodeChar: Gen[Random, Char] = {
+    def unicodeRange(min: Int, max: Int): Gen[Random, Char] =
+      fromEffectSample {
+        // 0xFFFF and 0xFFFE are not characters in the Unicode standard.
+        // See http://www.unicode.org/charts/PDF/UFFF0.pdf
+        nextInt(max - min + 1)
+          .map(r => (min + r).toChar)
+          .map(Sample.shrinkIntegral(0))
+      }
+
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -143,8 +143,26 @@ object Gen extends GenZIO with FunctionVariants {
   /**
    * A generator of strings. Shrinks towards the empty string.
    */
-  final val anyString: Gen[Random with Sized, String] =
-    Gen.string(Gen.anyChar)
+  final def anyString: Gen[Random with Sized, String] =
```
regiskuckaertz: Turned into a `def` to avoid initialisation failures.

regiskuckaertz:
@vasilmkd Thank you mate 🙏

adamgfraser:
Wow, that was fast!

test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -143,8 +143,26 @@ object Gen extends GenZIO with FunctionVariants {
...
+    Gen.string(Gen.anyUnicodeChar)
+
+  /**
+   * A generator of Unicode characters. Shrinks toward '0'.
+   * Heavily inspired from:
+   * https://github.com/typelevel/scalacheck/blob/ab15a9fe012ca7c9feb48ee188e4d167df76d6ba/src/main/scala/org/scalacheck/Arbitrary.scala#L122-L134
+   */
+  final val anyUnicodeChar: Gen[Random, Char] = {
+    def unicodeRange(min: Int, max: Int): Gen[Random, Char] =
```
adamgfraser: This definitely works but in the future could be simplified to `Gen.char`.

test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -143,8 +143,26 @@ object Gen extends GenZIO with FunctionVariants {
...
+      fromEffectSample {
+        // 0xFFFF and 0xFFFE are not characters in the Unicode standard.
+        // See http://www.unicode.org/charts/PDF/UFFF0.pdf
+        nextInt(max - min + 1)
+          .map(r => (min + r).toChar)
+          .map(Sample.shrinkIntegral(0))
```
adamgfraser: Be careful about shrinking logic here. We want to make sure that shrinks obey the invariants of the generator. Here we would shrink to `0` even if the minimum value in the range is not equal to `0`. It doesn't matter in this case because this is a `private` method and it is being called with a `0` minimum value but just a heads up.

regiskuckaertz: ah that makes sense, I should use `min` instead here. Will update 👍

test/shared/src/main/scala/zio/test/Gen.scala
regiskuckaertz: Oh, I didn't see that. Yeah that's much better and safer.

PR #1905, opened 2019-10-04
Make `asError` argument call-by-name to match `as`

reibitto:
I stumbled upon this because I wanted to avoid constructing an unnecessary error object when doing something like `someSuccessfulIO.asError(createError(...))`. But I also noticed `ZIO.as` is currently call-by-name but not `asError`. So seems like a good thing to do for consistency's sake anyway.

I noticed this same mismatch in `ZSink`.


jdegoes:
I think for consistency reasons this makes sense. One can debate whether or not it should be eager or by name, but by name at least affords tracing potential.

jdegoes:
Thank you! 🙏

Issue #1898, opened 2019-10-04
ZIO Test: Migrate ZIO Test Internal Test Suite to ZIO Test


adamgfraser:
ZIO Test has reached a point of maturity where it makes sense to migrate the internal test suite to use ZIO Test itself rather than the minimal test framework we are currently using. This will make it easier for contributors in the future because they will be able to use one testing framework across all ZIO projects and take advantage of all the features in ZIO Test to write and run tests.

This is a great issue for people who are looking to learn more about writing tests in ZIO Test and some of how it works and is also a good first issue for new contributors since there are a lot of different specs of varying size and complexity so you can pick what works for you. If you are interested in working on this please respond here with which spec you would like to work on so we don't duplicate efforts, submit a PR for that spec, and then we will check it off here.

- [x] AssertionSpec
- [x] BoolAlgebraSpec
- [x] CheckSpec
- [x] ClockSpec
- [x] ConsoleSpec
- [x] DefaultTestReporterSpec
- [x] ExpectationSpec
- [x] EnvironmentSpec
- [x] FunSpec
- [x] GenSpec
- [x] GenZIOSpec
- [x] LiveSpec
- [x] ManagedSpec
- [x] RandomSpec
- [x] SampleSpec
- [x] SchedulerSpec
- [x] SummaryBuilderSpec
- [x] SystemSpec
- [x] TestAspectSpec
- [x] TestSpec
- [x] TimeVariantsSpec

unclebob418:
I'd like to help out with a few of these but may need some guidance (new to ZIO and I've never worked on a team before so the PR process is a bit unfamiliar). I'll start with TestSpec since that seems the simplest.

adamgfraser:
Great! Please feel free to reach out here or on Discord if you have questions or need help as you are working on it.

unclebob418:
I'll see if I can tackle the rest of these - the first one seemed easy enough and taking the time to understand the tests should help my understanding of how it all works

adamgfraser:
Sounds good. Do however many you want. They vary a lot in size so whenever you reach a stopping point let me know and we can merge what you have done. Feel free to reach out any time if you have questions as you are working on them.

unclebob418:
Working TestAspectSpec at the moment (understanding that will help with the doco). This one is... mch harder :) getting there slowly

unclebob418:
@adamgfraser can you please tick off the tests migrated under #1959?
TestSpec
ConsoleSpec
EnvironmentSpec
[...]

PR #1906, opened 2019-10-04
Migrate ZIO internal test to ZIO Test

darl:
#1898

- [x]  AssertionSpec
- [x]   BoolAlgebraSpec
- [ ]   CheckSpec
- [ ]   ClockSpec
- [ ]   ConsoleSpec
- [ ]   DefaultTestReporterSpec
- [ ]   EnvironmentSpec
- [x]   FunSpec
- [ ]   GenSpec
- [x]   GenZIOSpec
- [x]   LiveSpec
- [x]   ManagedSpec
- [ ]   MockSpecSpec
- [ ]   RandomSpec
- [ ]   SampleSpec
- [ ]   SchedulerSpec
- [ ]   SystemSpec
- [ ]   TestAspectSpec
- [ ]   TestSpec


test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
```diff
@@ -207,3 +198,15 @@ object BoolAlgebraSpec extends ZIOBaseSpec {
     if (a == b) (a, b) else randomEqualBoolAlgebra(size)
   }
 }
+
+object BoolAlgebraSpecHelper {
```
adamgfraser: Are you going to migrate this spec as part of this PR? If not I would leave this where it was.

darl: i think this will be last one for this pr

adamgfraser: Okay, great. That would be fantastic! Maybe call this `BoolAlgebraSpecUtil` to conform with the practice in `core` and `tests`?  Otherwise will review this file when it is done but everything else is good to go once the tests pass. It looks like there are some compilation issues right now. Let me know if you need any help in resolving.

darl:
wow. all test was moved from `src/test/scala` to `src/main/scala`

adamgfraser:
Yes, there were some issues with the minimal testing framework specs running multiple times on ScalaJS  if they were in the `test` folder.

darl:
@adamgfraser `DefaultRunnableSpec` also have `main` method and therefore will be executed twice on ScalaJS.

test-tests/shared/src/main/scala/zio/test/GenUtils.scala
```diff
@@ -1,10 +1,11 @@
...
-
-import zio.{ DefaultRuntime, Managed, UIO, ZIO }
+import zio.Exit.{ Failure, Success }
 import zio.random.Random
 import zio.test.environment.TestRandom
+import zio._
+
+import scala.concurrent.Future
 
 object GenUtils extends DefaultRuntime {
```
adamgfraser: Can we remove the `DefaultRuntime` here?

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
```diff
@@ -0,0 +1,167 @@
+package zio.test
+
+import zio.test.Assertion._
+import zio.test.BoolAlgebraSpecHelper._
+import zio.test.TestAspect._
+
+import scala.util.Random
+
+object BoolAlgebraSpec
```
adamgfraser: Now that we are using `ZIO Test` we can take advantage of its built in property based testing functionality instead of our homegrown one. So `randomBoolAlgebra` can return a `Gen[Random, BoolAlgebra[Int]]` and then in all the tests that use it we can use the `check` method instead of `nonFlaky`.

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
darl: > Maybe call this BoolAlgebraSpecUtil to conform with the practice in core and tests?

actually named it like `ZIOSpec` companion `ZIOSpecHelper`

test-tests/shared/src/main/scala/zio/test/GenUtils.scala
darl: `GenSpec` still use it

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
darl: great idea!

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
adamgfraser: Sounds good.

test-tests/shared/src/main/scala/zio/test/GenUtils.scala
adamgfraser: Good point.

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
```diff
@@ -1,34 +1,30 @@
...
-          val c = randomBoolAlgebra()
-          assert((a && b) && c, equalTo(a && (b && c)))
-        } @@ nonFlaky(100),
-        test("and is commutative") {
-          val a = randomBoolAlgebra()
-          val b = randomBoolAlgebra()
-          assert(a && b, equalTo(b && a))
-        } @@ nonFlaky(100),
+        testM("and distributes over or") {
+          zio.test.check(boolAlgebra, boolAlgebra, boolAlgebra) { (a, b, c) =>
```
darl: getting some weird compilation error without fully-qualified name

adamgfraser: Yes, it is because we have the `test` package object extend the `CheckVariants` trait so we don't clutter up the package object with all the different variants of the check methods. It looks like this issue only appears because the test suite uses the same package name. Not great but I don't have a better solution right now so we will have to go with it.

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+  val success2 = BoolAlgebra.success(value2)
+  val failure1 = BoolAlgebra.failure(value3)
+  val failure2 = BoolAlgebra.failure(value4)
+
+  val isSuccess: Assertion[BoolAlgebra[_]] = assertion("isSuccess")()(_.isSuccess)
+  val isFailure: Assertion[BoolAlgebra[_]] = assertion("isFailure")()(_.isFailure)
+
+  def boolAlgebra: Gen[Random with Sized, BoolAlgebra[Int]] = Gen.small(s => boolAlgebraOfSize(s))
+
+  def boolAlgebraOfSize(size: Int): Gen[Random, BoolAlgebra[Int]] =
```
adamgfraser: I think right now this is generating objects that are "too large" which is why you are having to reduce the size and number of samples in checking the "hashCode is consistent with equals" property. I think we want the concept of "size" to be that each term adds one to the size, so for example `And(Not(Value(1)), Value(2))` would have a size of four, and then `boolAlgebraOfSize` should always generate objects of the requested size. I think something like the below will do that. When I tried that I was able to change the size and number of sample parameters back to their previous values.

```scala
  def boolAlgebra: Gen[Random with Sized, BoolAlgebra[Int]] = Gen.small(s => boolAlgebraOfSize(s), 1)

  def boolAlgebraOfSize(size: Int): Gen[Random, BoolAlgebra[Int]] =
    if (size == 1) {
      Gen.int(0, 9).map(BoolAlgebra.success)
    } else if (size == 2) {
      boolAlgebraOfSize(size - 1).map(!_)
    } else {
      for {
        n <- Gen.int(1, size - 2)
        gen <- Gen.oneOf(
                (boolAlgebraOfSize(n) <*> boolAlgebraOfSize(size - n - 1)).map((p => p._1 && p._2)),
                (boolAlgebraOfSize(n) <*> boolAlgebraOfSize(size - n - 1)).map(p => p._1 || p._2),
                boolAlgebraOfSize(size - 1).map(!_)
              )
      } yield gen
    }
```

test-tests/shared/src/test/scala/zio/test/BoolAlgebraSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+        },
+        test("either returns disjunction of two values") {
+          assert(success1 || success2, isSuccess) &&
+          assert(success1 || failure1, isSuccess) &&
+          assert(failure1 || success1, isSuccess) &&
+          assert(failure1 || failure2, isFailure)
+        },
+        testM("hashCode is consistent with equals") {
+          zio.test.checkSome(equalBoolAlgebraOfSize(3))(n = 4) { pair =>
+            println(1)
```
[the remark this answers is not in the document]
darl: oops

adamgfraser:
@darl Great job on this! Thanks for contributing!

PR #1908, opened 2019-10-04
Deduplicate compier error message

ioleo:
The compier error is currently being displayed twice in the [use module pattern](https://zio.dev/docs/howto/howto_use_module_pattern) doc. My intention was to hide the original error and re-type it below, to clean it up a little (remove noise like `$anon: repl.Session.App0`) to get the message across better. 

image


jdegoes:
@ioleo Thank you!

PR #1912, opened 2019-10-05
Fix Issue with Tests Running Twice on ScalaJS

adamgfraser:
Tests in ZIO Test internal test suite were running twice on ScalaJS after #1895 was merged. To run the tests written with the minimal testing framework at all we need a `main` method which means on ScalaJS we need to set `scalaJSUseMainModuleInitializer` to `true`. Unfortunately this means that the `main` method is run when the object is initialized, not just when the `main` method is explicitly called. So the result is that the tests written with the minimal testing framework are run once when we call `test:run` and again when we call `test` to run the tests written with ZIO Test.  See example [here](https://circleci.com/gh/zio/zio/30961?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

To address this, I moved the tests written with the minimal testing framework to the `main` folder and only enabled `scalaJSUseMainModuleInitializer` there.

Sorry for not spotting this earlier.


ghostdogpr:
Woops, good catch!

PR #1915, opened 2019-10-05
resolve #1891 Add `ZEnv`

### LINKED ISSUE #1891: Add `ZEnv`
Currently, we lack a type alias to describe the set of all environmental effects that "ship" with ZIO.

Ideally, we would have this inside the `zio` package object:

```scala
type ZEnv = Clock with Console with System with Random with Scheduler with Blocking
```

It's not possible to do this naively, because Scala.js does not support `Blocking`. However, one could extract out a package-private `ZIOTopLevel` (or `AbstractZIOPackage`?) from the existing `package.scala` inside the `zio` package, and then extend from this class in both the Scala.js directory, and the JVM directory.

Then each extension could specify its own `type ZEnv`, as appropriate for the underlying platform.

Finally, existing code that uses, e.g. `DefaultEnvironment#Environment` or similar could be refactored to use the much simpler `ZEnv` type alias.

mschuwalow:
I went ahead and deleted DefaultRuntime#Environment as it's now useless in my opinion.

I had a quick skim through the docs and I don't think DefaultRuntime#Environment is mentioned anywhere else, but I might have missed something.

/cc @jdegoes

### DIFF
--- a/core/jvm/src/main/scala/zio/DefaultRuntime.scala
+++ b/core/jvm/src/main/scala/zio/DefaultRuntime.scala
@@ -23,9 +23,7 @@ import zio.random.Random
 import zio.blocking.Blocking
 import zio.internal.{ Platform, PlatformLive }
 
-trait DefaultRuntime extends Runtime[Clock with Console with System with Random with Blocking] {
-  type Environment = Clock with Console with System with Random with Blocking
-
-  val Platform: Platform       = PlatformLive.Default
-  val Environment: Environment = new Clock.Live with Console.Live with System.Live with Random.Live with Blocking.Live
+trait DefaultRuntime extends Runtime[ZEnv] {
+  val Platform: Platform = PlatformLive.Default
+  val Environment: ZEnv  = new Clock.Live with Console.Live with System.Live with Random.Live with Blocking.Live
 }
--- a/core/shared/src/main/scala/zio/AbstractZIOPackage.scala
+++ b/core/shared/src/main/scala/zio/AbstractZIOPackage.scala
@@ -1,3 +1,5 @@
+package zio
+
 /*
  * Copyright 2017-2019 John A. De Goes and the ZIO Contributors
  *
@@ -14,7 +16,7 @@
  * limitations under the License.
  */
 
-package object zio extends EitherCompat {
+abstract class AbstractZIOPackage extends EitherCompat {
   private[zio] type Callback[E, A] = Exit[E, A] => Unit
 
   type Canceler[R] = URIO[R, Any]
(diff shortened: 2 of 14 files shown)


mschuwalow:
Hmm I think it would be nicer to have ZEnv as an abstract type member in AbstractZIOPackage, but the  2.11 typechecker does not seem to like it for some reason :man_shrugging:

vasilmkd:
You can check if everything is ok with the documentation by running `doc` from the root project.

core/shared/src/main/scala/zio/AbstractZIOPackage.scala
```diff
@@ -1,3 +1,5 @@
+package zio
```
LGLO: Can we keep package object zio with all the stuff where people usually look for it?
It could be done`package object zio extends EitherCompat with ZEnvDefinition` where `ZEnvDefinition` is a trait containing `type ZEnv = ...`.

mschuwalow: If we go this route we might also just alias ZEnv as
```scala
type ZEnv         = DefaultRuntime#Environment
```
. I think this might be the nicest way to do this

core/jvm/src/main/scala/zio/DefaultRuntime.scala
```diff
@@ -23,9 +23,7 @@ import zio.random.Random
 import zio.blocking.Blocking
 import zio.internal.{ Platform, PlatformLive }
 
-trait DefaultRuntime extends Runtime[Clock with Console with System with Random with Blocking] {
-  type Environment = Clock with Console with System with Random with Blocking
-
-  val Platform: Platform       = PlatformLive.Default
-  val Environment: Environment = new Clock.Live with Console.Live with System.Live with Random.Live with Blocking.Live
+trait DefaultRuntime extends Runtime[ZEnv] {
```
jdegoes: Nice cleanup! 👍

core/shared/src/main/scala/zio/AbstractZIOPackage.scala
jdegoes: I think I like @LGLO's suggestion, since it makes it easier to find stuff for new contributors.

We could definitely do `type ZEnv = ...` (as per [here](https://dotty.epfl.ch/docs/reference/dropped-features/type-projection.html) it looks still valid in Dotty), although it will lead to a circular reference in `DefaultRuntime` so you'd have to revert that cleanup.

mschuwalow: True. All in all I think @LGLO's suggestion ends up being the cleanest

jdegoes:
@mschuwalow Great work!

PR #1918, opened 2019-10-06
Add test selection feature to ZIO Test

reibitto:
This introduces test selection to the ZIO Test runner. Essentially what `-z` does in ScalaTest and `-ex` in specs2. I named it `-t` for test selection, but I'm fine with anything really. I only did that because I could foresee somebody wanting `-s` for suite selection (I thought about adding this too, but it's more complicated since suites can be nested).

Example usage:

```
sbt:zio> coreTestsJVM/testOnly *.ZScheduleSpec -- -t "retry 0 time for `once` when first time succeeds"
[info] + ZScheduleSpec
[info]   + Retry on failure according to a provided strategy
[info]     + retry 0 time for `once` when first time succeeds
[info] Ran 1 test in 408 ms: 1 succeeded, 0 ignored, 0 failed
```

Two things:
- I wasn't sure what to do about the command-line parser. Normally I'd use a library for this, but I'm pretty sure we don't want to introduce any dependencies here. Plus it's overkill since we only have 1 option right now (which doesn't even need validation). Maybe it's fine to revisit the idea of creating a more robust parser once we need to add more options?
- Can somebody check the type params [here](https://github.com/reibitto/zio/blob/c7d22196fe468fe2137c6828f83c5d42b84d9544/test/shared/src/main/scala/zio/test/Spec.scala#L221-L223)? I'm still awful when it comes type parameter covariance/contravariance. I just do what the compiler tells me to. 😛


test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
```diff
@@ -18,7 +18,7 @@ abstract class BaseTestTask(val taskDef: TaskDef, testClassLoader: ClassLoader)
 
   protected def run(eventHandler: EventHandler, loggers: Array[Logger]) =
     for {
-      res    <- spec.run.provide(new SbtTestLogger(loggers) with Clock.Live)
+      res    <- spec.run(args).provide(new SbtTestLogger(loggers) with Clock.Live)
```
adamgfraser: Can we have the SBT test runner filter the spec here and then run the resulting spec? I think that would make this more modular where the testing framework is exposed methods like `filter` and `run` and then the client, in this case the test runner, can combine those to get the desired output instead of baking the concept of `run` taking a certain kind of command line parameter.

test/shared/src/main/scala/zio/test/AbstractRunnableSpec.scala
```diff
@@ -35,8 +36,19 @@ abstract class AbstractRunnableSpec {
   /**
    * Returns an effect that executes the spec, producing the results of the execution.
    */
-  final def run: URIO[TestLogger with Clock, ExecutedSpec[Label, Failure, Success]] =
-    runner.run(spec)
+  final def run(args: TestArgs): URIO[TestLogger with Clock, ExecutedSpec[Label, Failure, Success]] = {
```
adamgfraser: I think this logic can move into the SBT test runner.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -38,10 +38,12 @@ object DefaultTestReporter {
                          case Spec.TestCase(_, test) => test.map(_.isLeft);
                          case _                      => UIO.succeed(false)
                        })
-            hasFailures   = failures.exists(identity)
-            status        = if (hasFailures) Failed else Passed
-            renderedLabel = if (hasFailures) renderFailureLabel(label, depth) else renderSuccessLabel(label, depth)
-            rest          <- UIO.foreach(specs)(loop(_, depth + tabSize)).map(_.flatten)
+            hasFailures = failures.exists(identity)
+            status      = if (specs.isEmpty) Ignored else if (hasFailures) Failed else Passed
```
adamgfraser: What is the significance of returning `Ignored` versus passed here? It looks like we get the same output where we don't run any tests and exit with a success code.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -218,6 +218,24 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
       case t @ TestCase(_, _)            => Spec(f(t))
     }
 
+  final def filter[R1 <: R, E1 >: E, L1 >: L, T1 >: T](
```
adamgfraser: Type annotations look good. Can we get `f` to return a `Boolean` instead of an `Option`?

test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
reibitto: Hmm, I haven't found a great way to do this from `BaseTestTask`. I guess I could filter there and change `AbstractRunnableSpec.run` to accept the new spec and use that instead, but I'm not sure if that's what you were going for. Maybe I'm just not seeing the better solution. Because I can't swap out the spec in `AbstractRunnableSpec` (it being abstract and all).

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
reibitto: You're right, it doesn't seem to matter.

test/shared/src/main/scala/zio/test/Spec.scala
reibitto: Yeah, that's a good point. You'd expect `Boolean` when encountering a `filter`. There wasn't a need to use `Option` there.

test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
adamgfraser: You can do something like:

```scala
      res <- args.testSearchTerm
              .fold(spec.run)(
                s => spec.runner.run(spec.spec.filterTestLabels(_.toString.contains(s)).getOrElse(spec.spec))
              )
              .provide(new SbtTestLogger(loggers) with Clock.Live)
```

May make sense to pull the filtering logic out into a separate helper function.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -38,10 +38,12 @@ object DefaultTestReporter {
                          case Spec.TestCase(_, test) => test.map(_.isLeft);
                          case _                      => UIO.succeed(false)
                        })
-            hasFailures   = failures.exists(identity)
-            status        = if (hasFailures) Failed else Passed
-            renderedLabel = if (hasFailures) renderFailureLabel(label, depth) else renderSuccessLabel(label, depth)
-            rest          <- UIO.foreach(specs)(loop(_, depth + tabSize)).map(_.flatten)
+            hasFailures = failures.exists(identity)
+            status      = if (hasFailures) Failed else Passed
+            renderedLabel = if (specs.isEmpty) ""
```
adamgfraser: Is it possible for `specs` to be empty? Aren't we defaulting to running all the tests if the filter predicate isn't matching any tests? I don't think we need to make any changes to this file, though I could be wrong.

reibitto: This addition isn't really for when the root Suite's specs are empty. It's more for the nested Suites below that. If we don't have this, then you get output like:

```
+ ZScheduleSpec
  + Repeat on success according to a provided strategy
  + Collect all inputs into a list
  + Repeat an action 2 times and call `ensuring` should
  + Retry on failure according to a provided strategy
    + retry 0 time for `once` when first time succeeds
  + Retry according to a provided strategy
  + Return the result of the fallback after failing and no more retries left
  + Return the result after successful retry
  + Retry a failed action 2 times and call `ensuring` should
```

as opposed to:

```
+ ZScheduleSpec
  + Retry on failure according to a provided strategy
    + retry 0 time for `once` when first time succeeds
```

I much prefer the latter (only show the selected nodes). Well, you could color the ignored specs gray or something, but it's still a lot of noise in my opinion. Especially for large specs where you would have to scroll to see the part you care about.

test/shared/src/main/scala/zio/test/Spec.scala
reibitto: Ooh, that's much nicer. Thanks! I applied your suggestions.

I'd like to add the suite filter too, but if you want that as part of this PR I think it's going to take me some time. I need to think about it more. It seems like to do it properly you'd need to do a depth-first search while maintaining history of the nodes you've visited. But maybe I'm over-complicating it. I haven't given it too much thought yet.

adamgfraser: No I definitely don't think we need to do the suite filter as part of this PR. I know you had mentioned it above and just trying to think ahead a bit but I agree it has some additional complications and I have a feeling this will be the more commonly used feature.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
adamgfraser: Got it. That makes sense.

adamgfraser:
Can you resolve the merge conflicts?

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -47,6 +47,19 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
       case c @ TestCase(_, _)         => f(c)
     }
 
+  /**
+   * Traverses all nodes and leaves only the ones that have a path to tests that satisfy the given predicate.
```
adamgfraser: Isn't this false? Don't we keep all the suites, regardless of whether they contain tests that satisfy the predicate?

reibitto: Yeah, sorry, the wording isn't accurate. Hmm... maybe "Traverses all nodes and filters out TestCases with labels that do not satisfy the given predicate."? Do you have better suggestions?

adamgfraser: "Returns a new Spec containing only tests with labels satisfying the specified predicate"?

reibitto: I like that better. Thanks. 👍

adamgfraser:
Great work on this! Excited to have this feature!

reibitto:
And thanks for your help and catching my silly mistakes. Hoping my next PR will be a little smoother. 😅

PR #1924, opened 2019-10-07
fatal-warnings based on system property

reibitto:
This makes fatal-warnings customizable based on a system property (always on in CI though). The motivation being that getting unused import/variable errors while developing locally can get in the way when doing rework.

I was initially going to make the default `true`, but it seems like others are fine with it being `false`. I don't have strong preferences either way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,7 +99,7 @@ compile: &compile
     - <<: *load_cache
     - run:
         name: Compile code
-        command: ./sbt ++${SCALA_VERSION}! compileJVM
+        command: ./sbt -Dfatal.warnings=true ++${SCALA_VERSION}! compileJVM
     - <<: *clean_cache
     - <<: *save_cache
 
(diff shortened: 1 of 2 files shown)


jdegoes:
The problem with something like this is that it encourages use of the CI for build testing, which given that it costs $$$ to do that, is maybe not a great idea.

Maybe it'd be better to allow setting a "DEV MODE" environment variable locally, which disables warnings-as-errors, etc.?

reibitto:
That's one reason I initially made the setting default to true. I can sympathize with the cost issue. For one of my projects at work we're required to use our own servers and the whole test suite takes about 90 minutes to run. To save on costs and prevent unnecessary builds from queuing up, we did 3 things:
1. fatal-warnings is on by default (have to explicitly disable it)
2. Create a git pre-push hook to do a scalafmt check to prevent those "oops, forgot to format before pushing" (which I did just a little while ago 😅)
3. Only run the tests if you commit with `[ci run]` in the commit message

3 was the biggest win, though I can imagine it would be annoying for a public repo where new people constantly come and go.

PR #1925, opened 2019-10-07
Add SBT onLoadMessage

reibitto:
To make useful SBT tasks more discoverable for first-time contributors. Example:

welcome-message


reibitto:
Actually I just noticed my description of `compileJVM` is wrong. It doesn't compile all the JVM modules like I thought. But now I'm confused... Is it supposed to? I sort of feel like that was the intention (considering what `testJVM` looks like), but maybe the command simply wasn't kept up to date as new modules were added?

I don't know. It's tough to tell for me.

ghostdogpr:
It’s currently used by the CI only to compile with dotty. I think some projects might just be forgotten, while some others are left out because they don’t compile on dotty yet. Maybe we could make it compile all and have a dedicated one for dotty (+ update the CI config)?

adamgfraser:
I think most of the other projects are left out because they don't compile on Dotty yet. My understanding is that ZIO Streams doesn't, though work for that is ongoing. ZIO Test depends on Streams, and then most of the other modules like the SBT test runner, the ZIO Test examples, and the benchmarks depend on one or the other.

We could definitely add another task to compile everything.

reibitto:
Would it make sense to do it the other way around? That is making `compileJVM` compile everything and add a `compileDotty`/`compileJVMDotty`/or whatever it should be called?

adamgfraser:
Yes I think that would definitely be clearer. I was hoping I could get everything to compile on Dotty and make this issue moot but I just blew up the compiler so that may be a little farther away.

reibitto:
Alright, updated the aliases and config.yml.

adamgfraser:
Looks like it just needs a format.

reibitto:
Oops, I always forget that. I might need to set up a git pre-push hook. 😅

adamgfraser:
This looks great! Can you just make that last change to #1918 so we can get it merged first? Want to make sure we have the test selection feature implemented before we start advertising it every time a contributor opens SBT. 😃

reibitto:
Yeah, good point. I updated the other PR.

adamgfraser:
Great idea! This is really going to help new contributors.

PR #1928, opened 2019-10-07
Compile ZIO Streams and ZIO Test Against Dotty

### LINKED ISSUE #1524: ZIO Stream: Compile Against Dotty
Currently we only compile ZIO Core against Dotty. This ticket is to get ZIO Stream to compile against Dotty as well. This would involve identifying what issues cause ZIO Stream to fail to compile with Dotty today and resolving those issues. It may also be helpful to look at the settings in `build.sbt` and in particular `dottySettings` to see if those same settings can be applied to ZIO Stream or if other settings are needed.

Getting ZIO Stream to compile against Dotty is a blocking item for other downstream ZIO libraries, including ZIO Test, to do so as well. As the Scala ecosystem transitions from Scala 2 to Scala 3 over the next couple of years there is a big opportunity for ZIO to be nimble and add value for users with its zero dependency approach if we can provide a true cross-platform, cross-version solution.

cc @iravid

adamgfraser:
Resolves #1524.


iravid:
Exciting! Happy to help if needed :-)

adamgfraser:
@iravid Are you able to make anything of [this](https://circleci.com/gh/zio/zio/32058?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)? Previously it looked like there were a few compilation errors (one of which seems to be a Dotty type inference issue that I will open a ticket for and copy you on). But now the Dotty compiler is crashing.

jdegoes:
Whoa, this is big!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1354,7 +1354,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
         add <- self
                 .mapM(f)
                 .distributedDynamicWith(
-                  buffer, { kv: (K, V) =>
+                  buffer, { (kv: (K, V)) =>
```
neko-kai: This is highlighted as redundant by Intellij, someone might undo it by mistake later - a way around it would be to use parens instead of curlies to enclose this lambda. (actually, I think it's legal in this case to avoid even parens, since the argument is enclosed)

iravid:
@adamgfraser wild guess, but maybe it's because of the nested classes we define in the sinks. I think they all share the name `State`.

Maybe try as a first step to rename them? If that doesn't help maybe try making them not nested.

adamgfraser:
@iravid Will do!

adamgfraser:
@iravid I can confirm the issue is with `ZSink`. If I remove the implementations of the methods in `ZSink` I am able to successfully compile against Dotty. Working on further isolating the issue.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Thanks! I took out the brackets and parentheses entirely.

adamgfraser:
@iravid This is ready for review.

build.sbt
```diff
@@ -33,9 +33,12 @@ addCommandAlias("fmt", "all scalafmtSbt scalafmt test:scalafmt")
 addCommandAlias("check", "all scalafmtSbtCheck scalafmtCheck test:scalafmtCheck")
 addCommandAlias(
   "compileJVM",
-  ";coreJVM/test:compile;stacktracerJVM/test:compile;streamsTestsJVM/test:compile;testTestsJVM/test:compile;testTestsJVM/test:compile;testRunnerJVM/test:compile;examplesJVM/test:compile"
```
ghostdogpr: Why not coreJVM here?

adamgfraser: I was just trying to align with the other commands that included `streamsTestsJVM` and `testTestsJVM` so it seemed like there was an intention to compile the tests as well. I'm fine with not compiling the tests, we should probably just be consistent.

test/shared/src/main/scala/zio/test/mock/MockSpec.scala
```diff
@@ -95,7 +95,7 @@ sealed trait MockSpec[-R, +E, +A] { self =>
 
         case MockSpec.FlatMap(nestedSpec, continue) =>
           for {
-            _   <- state.continuationsRef.update(continue :: _)
+            _   <- state.continuationsRef.update(continue.asInstanceOf[Any => MockSpec[R, E, Any]] :: _)
```
iravid: Add a comment about Dotty?

test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -441,4 +441,7 @@ object Gen extends GenZIO with FunctionVariants {
     if (n < min) min
     else if (n > max) max
     else n
+
+  private def defaultShrinker: Any => ZStream[Any, Nothing, Nothing] =
```
iravid:
```suggestion
  private val defaultShrinker: Any => ZStream[Any, Nothing, Nothing] =
```

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -67,7 +67,7 @@ object AssertionSpec
             sampleUser,
             isCase[SampleUser, (String, Int)](
               termName = "SampleUser",
-              SampleUser.unapply,
+              { case SampleUser(name, age) => Some((name, age)) },
```
iravid: This is Dotty related too?

test/shared/src/main/scala/zio/test/mock/MockSpec.scala
adamgfraser: I think Dotty is actually right here. The type of `continue` is `A => MockSpec[R, E, Any]` for some concrete type `A`. So it isn’t sound to infer this as `Any => MockSpec[R, E, Any]`. It is only safe for us to do so because we know that we have built the list of continuations in such a way that the types match up.

iravid: Right, upcasting on the input is actually downcasting.

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
adamgfraser: Yes. Explicit calls to `unapply` no longer return an `Option`. See https://github.com/lampepfl/dotty/issues/2335.

PR #1929, opened 2019-10-07
Added choose and chooseByFrequency to Random.Service

ajaychandran:
```scala
    def choose[A](as: Chunk[A]): ZIO[R, Nothing, A]
    def choose[A](as: Iterable[A]): ZIO[R, Nothing, A]
    def chooseByFrequency[A](as: Iterable[A])(f: A => Int): ZIO[R, Nothing, A]
```

The last method allows random selection of an element based on a [frequency distribution](https://en.wikipedia.org/wiki/Frequency_distribution).
```scala
// a biased coin with a 0.75 probability for heads
val coin = Map("heads" -> 3, "tails" -> 1)
val toss = zio.random.chooseByFrequency(coin.keys)(coin)
```
Note that passing an empty chunk/collection would result in a [IllegalArgumentException at runtime](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt-int-).


adamgfraser:
Thanks for submitting a PR!

I'm a little hesitant about adding new methods to `Random` without a strong justification. To date we have kept the interfaces for the standard environment types (`Clock`, `Console`, `System`, and `Random`) relatively minimal, generally only implementing methods that were primitive or provided by the corresponding implementations in the Scala standard library. This has the advantage of keeping the number of methods that an instance extending these traits has to implement or override low. Then other types can build more advanced functionality on top of that (e.g. `Schedule` with `Clock`).

These methods definitely seem useful but I'm not sure they pull their weight to add to the interface of `Random`. `choose` can be implemented quite straightforwardly in terms of `nextInt` and if you think of a weighted version as taking a `Map[A, Double]` of values and frequencies that can be implemented fairly easily in terms of `nextDouble`. This functionality also already exists in `Gen#elements` and `Gen#weighted`. Would those satisfy your use case?

ajaychandran:
In practice, the need for randomness arises from either performing a random selection or shuffling. Including these methods would satisfy the first use case.

Having said that, I agree with the design philosophy and will close the PR if the justification above isn't sufficient.

Thanks for the feedback!

adamgfraser:
> In practice, the need for randomness arises from either performing a random selection or shuffling. Including these methods would satisfy the first use case.

`shuffle` is a method on `Random` today. Is that what you meant?

ajaychandran:
In applications, like games of chance and simulations, randomness is required to either perform a selection from a collection or to shuffle a collection. 

That is, `choose` and `shuffle` are the primitive operations from a usage standpoint. The existing implementation does not have a primitive for the former.

ajaychandran:
Generalizing `choose` would require defining a [distribution function](https://github.com/typelevel/spire/blob/master/core/src/main/scala/spire/random/Dist.scala). This is out of scope for the project.

@adamgfraser Thanks for the review.

Issue #945, opened 2019-06-02
Cats Effect interop needs updating in light of inherited interruption status


jdegoes:
In Cats Effect, interruption status is not inherited, and every "started" fiber starts off as interruptible. In the interop layer, we need to match Cats Effect semantics.

We should also look for other places where semantics have changed to warrant introduction of, e.g. `interruptibleMask`.

kaishh:
My opinion so far, although I'm very open to reconsider, is to consider simply dropping interruptibility inheritance altogether, because:

1. I find the examples not compelling enough, i.e. the example equations already hold with or without presence of inner forks in passed actions - as long as the fiber handles don't leak. Many combinators ensure that fiber handles can't leak, and fibers have no access to and cannot leak their own fiber handles when spawned by such a combinator.
2. It interacts badly with `bracket` and naive uses of `Resource` and `ZManaged` – even if `Concurrent` instance is fixed, monomorphic mixes of cats and ZIO primitives would interact badly in this case.
3. Some combinators change meaning in unexpected ways, e.g. `foreachPar`. Although one could argue that `foreachPar.uninterruptible` now behaves closer to `foreach.uninterruptible` due to uninterruptibility of inner action, in fact it behaves very differently now in another case – uninterruptiblity of inner forks means that sibling workers will NOT be interrupted after one of them fails – error propagation stops working when `foreachPar` is wrapped in `uninterruptible`, this is a very stark difference from `foreach`, where every failure stops the chain, and this behavior does not follow intuitively from `uninterruptible` modifier unless you're highly knowledgeable of internals!

Linking previous gitter discussion here - https://gitter.im/ZIO/Core?at=5cf408bc949a104c36901c42 

/cc @ghostdogpr

ghostdogpr:
To be honest, I have no strong opinion about it. The one that I think is really important is the supervision inheritance: i got bit by the way it works now and i had to comment some `.timeout` so that my code works properly. Didn’t get any case where I needed it on interruptibility.

kaishh:
Re: supervision inheritance, if we remove interruption inheritance, we can't make transitive child fibers visible to root parent's supervision, because then a they'll be able to interrupt 'hidden' fiber handles, e.g. in `foreachPar(...).uninterruptible.fork.interruptChildren` – foreachPar workers will be interrupted by the outermost supervisor, even though that directly contradicts its uninterruptible status.

[...]

PR #1937, opened 2019-10-08
Issue 1182: foreachPar, foreachPar_, collectAllPar, reduceAllPar

LGLO:
This is continuation to #1773 .

I'm wondering about using `ClassTag` to be able to create array "the proper way" vs `ArrayBuffer`.


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2022,29 +2022,84 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+                         .interruptible(fn(a))
+                         .run
+                         .flatMap { exit =>
+                           arbiter(promise, buffer, todo, idx)(exit)
+                         }
+                         .fork
+                   }
+              _ <- promise.succeed(Right(Nil)).when(as.isEmpty)
+              c <- restore(promise.await.absolve.unsandbox)
+                    .onError(_ => promise.interrupt *> Fiber.interruptAll(as))
```
LGLO: I think intention is to uninterruptibly fork fibers and register `aribter` on their completion but `await`ing and later can be interrupted.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2022,29 +2022,84 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+      c <- ZIO.uninterruptibleMask { restore =>
+            for {
+              as <- ZIO.traverse(as.zipWithIndex) {
+                     case (a, idx) =>
+                       ZIO
+                         .interruptible(fn(a))
+                         .run
+                         .flatMap { exit =>
+                           arbiter(promise, buffer, todo, idx)(exit)
+                         }
```
LGLO: I'll fix it to `.flatMap(arbiter(promise, buffer, todo, idx))`

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2022,29 +2022,84 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+            val updateBuffer = buffer.modify(b => (b.update(idx, a), b))
+            if (t == 1) (updateBuffer <* buffer.get.flatMap(b => promise.succeed(Right(b.toList)))) -> 0
+            else updateBuffer                                                                       -> (t - 1)
+          }.flatten
+      )
+
+    (for {
+      size    <- UIO.effectTotal(as.size)
+      todo    <- Ref.make(size)
+      buffer  <- Ref.make(new Array[Any](size).asInstanceOf[Array[B]])
```
LGLO: Should I use `Array.ofDim[B]` and add `ClassTag` to signature or use `ArrayBuffer` or something else?

LGLO:
StacktracesTest fails now. Can someone tell me, what's expected behavior regarding content of stack trace if some fiber fail?

iravid:
@LGLO I didn't review the code yet, but there's a trick you could use to get a ClassTag without requiring an implicit one. If you have a single element, you could use `Chunk.Tags.fromValue` to get back its class tag.

This means your state machine has to be a bit more complex (the first task to finish allocates the array), but can potentially be much more efficient for primitive data.

@mvv used the same trick in `ZStream.chunkN` if you want inspiration.

core/shared/src/main/scala/zio/ZIO.scala
LGLO: I'll go with `mutable.ArraySeq` because of:

> You would typically use an ArraySeq if you want an array for its performance characteristics, but you also want to create generic instances of the sequence where you do not know the type of the elements and you do not have a ClassTag to provide it at run-time. These issues are explained in the section on arrays.

>

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2022,29 +2022,80 @@ private[zio] trait ZIOFunctions extends Serializable {
...
-      .refailWithTrace
+  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] = {
+    def arbiter(
+      promise: Promise[Cause[E], List[B]],
+      buffer: Ref[Array[B]],
+      todo: Ref[Int],
+      idx: Int
+    )(res: Exit[E, B]): ZIO[R, Nothing, Unit] =
+      res.foldM[R, Nothing, Unit](
+        e => promise.fail(e).unit,
```
neko-kai: @LGLO 
This has a different, lossy behavior towards failures than `zipWith` before. I wonder if you could store failures in another collection and merge them into one Cause at the end when all interrupts have ran (discarding uninteresting `Cause.interrupt` failures without additional finalizer failures)
Note: you could reuse a single Array by storing `Either` in `Array` cells and in case of failure take only Lefts.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2022,29 +2022,80 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+        a =>
+          todo.modify { t =>
+            val updateBuffer = buffer.modify(b => (b.update(idx, a), b))
```
neko-kai: Using a mutable collection in `Ref` is pretty odd, I'm not sure there's any difference whatsoever here wrt a plain `val`, even volatility in Ref will apply to the array reference, not the contents of the array – since the reference does not change, it has no effect. Even `modify` here is completely equivalent to `buffer.get.map(b => b(idx) = a)` because `compareAndSet` spin-lock will never fail and retry, as `b` is always the same reference.

LGLO: I had same concern: https://github.com/zio/zio/pull/1773#discussion_r332588328 in original PR review. I think I'll revert this to `Array` or `mutable.ArraySeq` - depends on how I would handle error.

core/shared/src/main/scala/zio/ZIO.scala
LGLO: @neko-kai 
Thank you very much for that comment!

Now I'm wondering if previous behavior was intentional or just accidental because of implementation with `zipWithPar`. Effects run in concurrently so they can fail before receiving interruption (or be uninterruptible and complete) - simply we can have multiple failures. On the other hand there is only `E` in error channel so by types it looks OK, to collect only one (first to complete promise) failure.
Other problem for me is when effects are `uninterruptible` - should we then also await completion of other effects even if we discard their results? Sequential analog would be foreach not short-circuiting on failure.

That said, I'm going to implement what I'm told to implement - can you tell me what to do here?
If I have to merge all failures,  I'll try `Array[Exit[E, A]]` initially filled with `Failure(Cause.interrupt)` and also ignored during merge - WDYT about this?
In case of successes I can collect to List.

neko-kai: > Other problem for me is when effects are uninterruptible - should we then also await completion of other effects even if we discard their results? Sequential analog would be foreach not short-circuiting on failure.

We already always wait for completion – in case of success we must wait for all results, in case of failure `.onError(_ => Fiber.interruptAll)` will wait for all fibers to interrupt. This is consistent with current behavior of `foreachPar`, and with `race`, `zipWithPar` and other concurrent ops in ZIO, so it makes sense to me to do this here too!

> Sequential analog would be foreach not short-circuiting on failure.

This <strikethrough>is</strikethrough> was one of the shortcomings of `interruption inheritance` – `foreachPar(_).uninterruptible` behaves strange - https://github.com/zio/zio/issues/945#issuecomment-498068296 – this PR fixes that by wrapping workers with `interruptible`, but obviously you can reintroduce it with another `uninterruptible` inside – but in that case it'll behave same as all `race*` variants, `zipWithPar` and others so I think it's completely correct.

> Now I'm wondering if previous behavior was intentional or just accidental because of implementation with zipWithPar.

It's intentional, I'm sure. Being lossless about errors has always been one of the design principles

> On the other hand there is only E in error channel so by types it looks OK, to collect only one (first to complete promise) failure.

Well, `Cause` like `List` can always contain multiple E's without having to change the type so that's ok

neko-kai: @LGLO

neko-kai: @LGLO 
> I'll try Array[Exit[E, A]] initially filled with Failure(Cause.interrupt) and also ignored during merge - WDYT about this? In case of successes I can collect to List.

I feel like pre-filling the Array is not necessary, since you either get a single element on success or wait for it to fill up completely on failure anyway.

LGLO: Thank you hundredfold for this detailed response!
I'll do what you said.
I still have a problem with StacktracesSpec, but I'll try to understand why and eventually ask for help.

neko-kai: @LGLO Re: StacktracesSpec, I'd advise you turn `val debug = true` in the test, inspect the traces that fail and if they look fine on manual inspection, just change the predicate in failed tests to accept them. – the traces are expected to change with new implementation, the only thing you can do here is manually check that they're useful and then change the test expectation

jdegoes:
@LGLO Do you have benchmarks for this? Any questions on semantics?

LGLO:
No benchmarks yet, I'll do some when I'll be more confident about correctness of error channel.
I'll prepare examples of stack traces. I can see some differences to current implementation. Give me a day or two, pls.

LGLO:
Life has put a hammer on me and I'll be too busy to continue this in next two weeks. Feel free to reimplement it or take it from here during Berlin Hackathon.
I'm sorry!

regiskuckaertz:
@LGLO hi mate, hope you're all right! Both `foreachPar` and `foreachPar_` have received a makeover recently, do you think this is still relevant?

LGLO:
Perhaps this is obsolete. I'll close to not make any confusion.

PR #1938, opened 2019-10-08
effectAsyncMaybe should not resume fiber twice after synchronous result

### LINKED ISSUE #1842: `effectAsyncMaybe` can resume the Fiber incoherently even after returning the value synchronously
Reproduction:

```scala
import zio._
import zio.duration._

import scala.concurrent.Future

object ZioIncoherentAsyncRepro extends zio.App {

    val test =
      ZIO.effectAsyncMaybe[Environment, Nothing, Unit] {
        k =>
          Future {
            Thread.sleep(200L)
          }
          Some(IO.unit)
      }
        .flatMap {
          _ =>
            ZIO.sleep(500.millis) *>
              console.putStrLn("SUCCESS") *>
              ZIO.never
        }
        .uninterruptible

    test.as(0).delay(1.second)
  }
}
```

Expected:

Outputs `SUCCESS` and never terminates

Actual:

Outputs
```
INCOHERENT CONTINUE
CAN'T HAPPEN!
```
and terminates

darl:
resolves #1842

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -10,6 +10,7 @@ import zio.test.Assertion._
 import zio.test.TestAspect.{ flaky, ignore, jvm, nonFlaky }
 
 import scala.annotation.tailrec
+import scala.concurrent.ExecutionContext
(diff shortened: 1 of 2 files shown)


darl:
https://github.com/zio/zio/pull/1939 have a better solution

PR #1939, opened 2019-10-09
effectAsync should not resume fiber twice after interruption

### LINKED ISSUE #1843: Previously interrupted `effectAsync` can incoherently resume the fiber after it has been interrupted
Reproduction:

```scala
import zio._
import zio.duration._

import scala.concurrent.Future

object ZioIncoherentInterruptAsyncRepro extends zio.App {

    val test =
      ZIO.effectAsync[Environment, Nothing, Unit](k =>
        Future {
          Thread.sleep(200L)
        })
        .ensuring(console.putStrLn("Async finalizer!") *> ZIO.never)

    test.fork
      .flatMap(_.interrupt.delay(50.millis))
      .as(0)
  }
}
```

Expected output:

```
Async finalizer!
<non-termination>
```

Actual output:

```
Async finalizer!
INCOHERENT CONTINUE
CAN'T HAPPEN!

Process finished with exit code 0
```

darl:
resolves #1843 & #1842


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -606,9 +606,8 @@ private[zio] final class FiberContext[E, A](
    *
    * @param value The value produced by the asynchronous computation.
    */
-  private[this] final def resumeAsync: IO[E, Any] => Unit = {
-    val neverRan = new AtomicBoolean(true)
-    zio => if (neverRan.getAndSet(false) && exitAsync()) evaluateLater(zio)
+  private[this] final def resumeAsync(ref: AnyRef): IO[E, Any] => Unit = { zio =>
+    if (exitAsync(ref)) evaluateLater(zio)
```
jdegoes: The old approach seemed simpler. Of course, it didn't work, but it didn't work because the `AtomicBoolean` should have been allocated in the main interpreter loop, not here. If you allocate it there, and pass it into `resumeAsync`, it will solve the problem with fewer code changes.

darl: I have another PR with this approach https://github.com/zio/zio/pull/1938
It fixes #1842 but not fixes #1843

Also there are more generic ABA problem related to Fiber status.
Look at following steps:
1. Starting effectAsync `A` (now in `Suspended` status)
2. Fiber got interrupted (not in `Running` status)
3. Finalizer starts another EffectAsync `B` (now in `Suspended` status)
4. some code executed callback from `A`; it will try to unsuspend fiber

jdegoes: You are right. We used to have an "async counter" to deal with this problem. 

Let me take another look at the PR with this in mind...

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -654,31 +653,29 @@ private[zio] final class FiberContext[E, A](
...
+        else if (!state.compareAndSet(oldState, newState)) enterAsync(ref)
+        else true
 
       case _ => false
     }
   }
 
   @tailrec
-  private[this] final def exitAsync(): Boolean = {
+  private[this] final def exitAsync(ref: AnyRef): Boolean = {
```
jdegoes: We can type this more precisely than `AnyRef`; at least in your case it's `ZIO[_, _, _]`

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -413,13 +413,13 @@ private[zio] final class FiberContext[E, A](
                   val zio = curZio.asInstanceOf[ZIO.EffectAsync[Any, E, Any]]
 
                   // Enter suspended state:
-                  curZio = if (enterAsync()) {
+                  curZio = if (enterAsync(curZio)) {
```
jdegoes: You have to be careful to avoid the capture of `curZio` inside a lambda, as this will create an `ObjectRef` that forces access to the variable to go through the heap. In these cases you can use `zio` to avoid any possibility of capturing the wrong value.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -654,31 +653,29 @@ private[zio] final class FiberContext[E, A](
...
 
   @tailrec
-  private[this] final def exitAsync(): Boolean = {
+  private[this] final def exitAsync(zio: ZIO[_, _, _]): Boolean = {
     val oldState = state.get
 
     oldState match {
-      case Executing(FiberStatus.Suspended(_), observers, interrupt) =>
-        if (!state.compareAndSet(oldState, Executing(FiberStatus.Running, observers, interrupt))) exitAsync()
+      case Executing(FiberStatus.Suspended(_, oldZio), observers, interrupt) if zio eq oldZio =>
```
neko-kai: This assumes all ZIO values have unique references which is not true in general (consider a ZIO value stored in a `val` and reused), I'm not sure if that's relevant, but if it is, would it make sense to use a volatile counter?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -654,31 +653,29 @@ private[zio] final class FiberContext[E, A](
...
+        val interruptible = this.interruptible
+        val newState      = Executing(FiberStatus.Suspended(interruptible, zio), observers, interrupt)
 
-        if (!state.compareAndSet(oldState, newState)) enterAsync()
-        else if (shouldInterrupt) {
-          // Fiber interrupted, so go back into running state:
-          exitAsync()
-          false
-        } else true
+        if (interruptible && interrupt) false
```
neko-kai: This feels odd... What happens when this fiber is interrupted
```scala
ZIO.never.ensuring(ZIO.sleep(5.seconds).interruptible *> putStrLn("Waited 5s"))
```
 ? Does it wait 5 seconds? Did it wait 5 seconds with previous code (I think no)? Could be that we had another bug in this logic... @jdegoes

neko-kai: Hi, I checked this and yep the behavior is incorrect in ZIO 1.0.0-RC15 – "Waited 5s" is never printed - https://github.com/zio/zio/issues/2005

core/shared/src/main/scala/zio/internal/FiberContext.scala
darl: You are right.
Now I think a volatile counter is a better solution.
Another solution: pass `new Object` instead of `curZio` as sync pointer. But it will slightly increase allocation rate.

core/shared/src/main/scala/zio/internal/FiberContext.scala
neko-kai: Guess this should be ignored for now in this PR

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -412,15 +415,18 @@ private[zio] final class FiberContext[E, A](
                 case ZIO.Tags.EffectAsync =>
                   val zio = curZio.asInstanceOf[ZIO.EffectAsync[Any, E, Any]]
 
+                  val epoch = asyncEpoch
+                  asyncEpoch = epoch + 1
```
neko-kai: Just `asyncEpoch += 1` should be ok

darl:
@neko-kai they already here: https://github.com/zio/zio/pull/1939/files#diff-ce160572c9127744730d5a14dc428162R674-R729

core/shared/src/main/scala/zio/internal/FiberContext.scala
darl: need to capture current value to local `val`
and avoid double volatile reads

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -654,31 +659,29 @@ private[zio] final class FiberContext[E, A](
...
+        val interruptible = this.interruptible
+        val newState      = Executing(FiberStatus.Suspended(interruptible, epoch), observers, interrupt)
 
-        if (!state.compareAndSet(oldState, newState)) enterAsync()
-        else if (shouldInterrupt) {
-          // Fiber interrupted, so go back into running state:
-          exitAsync()
-          false
-        } else true
+        if (interruptible && interrupt) false
```
neko-kai: @darl
I have a feeling this optimization does not make sense, i.e. we can't make decisions on _partial_ observed state until after we're out of the spinlock – that is, after we're in control and have the _current final_ version of the state. (that is, I'd move this if after the `if (!set)` as before)

Also, the tests on CircleCI keep failing – since `master` is building fine for some time now, I think the problem may be with this PR, could you please look into that?

neko-kai: https://circleci.com/gh/zio/zio/35648 

```
[info]   - Stream.bufferSliding
[info]     + buffer the Stream with Error
[info]     - fast producer progress independently
[info]       Fiber failed.
[info]       A checked error was not handled.
[info]       None
[info]       
[info]       Fiber:145 was supposed to continue to:
[info]         a future continuation at zio.ZIO$$anonfun$run$2.apply(ZIO.scala:1120)
[info]         a future continuation at zio.ZIO$$anonfun$bracket_$1.apply(ZIO.scala:144)
[info]         a future continuation at zio.ZIO$$anonfun$run$2.apply(ZIO.scala:1120)
[info]         a future continuation at zio.ZIO$$anonfun$bracket_$1.apply(ZIO.scala:144)
[info]         a future continuation at zio.ZIO$$anonfun$run$2.apply(ZIO.scala:1120)
[info]         a future continuation at zio.ZManaged$$anonfun$use$1.apply(ZManaged.scala:649)
[info]         a future continuation at zio.test.Spec$$anonfun$foreachExec$1$$anonfun$apply$21.apply(Spec.scala:158)
[info]         a future continuation at zio.ZIOFunctions$$anonfun$foreachParN$2$$anonfun$apply$137$$anonfun$apply$139$$anonfun$apply$140$$anonfun$apply$141$$anonfun$apply$144.apply(ZIO.scala:2076)
[info]         a future continuation at zio.ZIO$$anonfun$forever$1.apply(ZIO.scala:454)
[info]       
[info]       Fiber:145 execution trace:
[info]         at zio.test.TestAspect$$anon$14$$anonfun$perTest$12$$anonfun$apply$19.apply(TestAspect.scala:427)
[info]         at zio.ZIO$$anonfun$raceWith$1$$anonfun$apply$79$$anonfun$apply$80$$anonfun$apply$92.apply(ZIO.scala:948)
[info]         at zio.internal.FiberContext$InterruptExit$$anonfun$apply$1.apply(FiberContext.scala:160)
[info]         at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:151)
[info]         at zio.ZIO$$anonfun$raceWith$1$$anonfun$apply$79$$anonfun$apply$80$$anonfun$apply$81$$anonfun$apply$82$$anonfun$apply$83$$anonfun$apply$85$$anonfun$apply$88$$anonfun$apply$91.apply(ZIO.scala:965)
```

darl: reverted it

neko-kai:
@darl 
Thank you very much for fixing this! 🙏 Sorry it took so long to merge!

PR #1945, opened 2019-10-10
collectWhile in Chunk, ZStreamChunk, StreamChunkEffect

vasilmkd:
I wanted to feel useful again.

@iravid 

I also have a suggestion. I noticed that `Chunk#collect` allocates an array of the same size as the original chunk, and slices it at the end. I'm curious to know if sacrificing runtime for improved memory complexity is a better alternative. We could count the number of _defined_ elements (the ones that satisfy `pf.isDefinedAt`) and then map them to the destination array by doing another pass over the original.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -47,16 +47,7 @@ sealed trait Chunk[+A] { self =>
   /**
    * Transforms all elements of the chunk for as long as the specified partial function is defined.
    */
-  def collectWhile[B](p: PartialFunction[A, B]): Chunk[B] = {
-    val len = self.length
-
-    var i = 0
-    while (i < len && p.isDefinedAt(self(i))) {
-      i += 1
-    }
-
-    Chunk.Slice(self, 0, i).map(p.apply)
-  }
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -44,6 +44,20 @@ sealed trait Chunk[+A] { self =>
...
   def collect[B](p: PartialFunction[A, B]): Chunk[B] = self.materialize.collect(p)
 
+  /**
+   * Transforms all elements of the chunk for as long as the specified partial function is defined.
+   */
+  def collectWhile[B](p: PartialFunction[A, B]): Chunk[B] = {
+    val len = self.length
+
+    var i = 0
+    while (i < len && p.isDefinedAt(self(i))) {
```
iravid: This needs to be formulated with `pf.applyOrElse` to avoid double evaluation of the PartialFunction

vasilmkd: I'm aware, same thing happens in `Arr#collect`. Should I fix it too?

iravid: That'd be good!

vasilmkd: Done.

PR #1946, opened 2019-10-10
ZIO Test: Fix Type Inference Issue with Assertion#equalTo

### LINKED ISSUE #1562: ZIO Test: Type Inference Issue with Assertion#equalTo
@regiskuckaertz raises the good point that code like the following does not compile:

```scala
assert(List(1, 2, 3).filter(_ => false), equalTo(List.empty))
// found: Assertion[List[Nothing]]
// required: Assertion[List[Int]]
```

Where:

```scala
class Assertion[-A]
final def assert[A](value: => A, assertion: Assertion[A])
final def equalTo[A](expected: A): Assertion[A]
```

It does work with an explicit type annotation on `assert`:

```scala
assert[List[Int]](List(1, 2, 3).filter(_ => false), equalTo(List.empty))
```

The problem is that the compiler is not zeroing in on the expected type being `List[Int]` early enough and so is not widening `List.empty` from `List[Nothing]` to `List[Int]`. We could potentially address this by adding a curried version of `assert`, perhaps:

```scala
final def assert_[A](value: => A)(assertion: Assertion[A])
```

However, this fractures our API and this method loses all type safety because the contravariant assertion type will get widened arbitrarily, so for example this would compile:

```scala
assert_("Adam")(equalTo(1))
```

If we were going to do that we almost might as well make `equalTo` return an `Assertion[Any]`.

Right now my inclination is that adding the explicit type parameter is the best option but that isn't entirely satisfactory so putting it out there to see if anyone else has better ideas.

adamgfraser:
Resolves #1562.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/RefMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/RefMSpec.scala
@@ -36,7 +36,7 @@ object RefMSpec
           for {
             refM  <- RefM.make[State](Active)
             value <- refM.updateSome { case Closed => IO.succeed(Active) }
-          } yield assert(value, equalTo[State](Active))
+          } yield assert(value, equalTo(Active))
         },
         testM("updateSome twice") {
           for {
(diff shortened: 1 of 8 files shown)


darl:
Can this also be simplified?
https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala#L750

adamgfraser:
@darl Yes.

adamgfraser:
Thanks to @jdegoes for posting the fix and to @regiskuckaertz and @ioleo for raising the issue. Really excited to finally have this one resolved!

PR #1949, opened 2019-10-11
Implement ZIO#forkInternal to Prevent Spurious Warnings

adamgfraser:
Currently we generate a lot of spurious fiber failure warnings. For example, consider:

```scala
ZIO.fail("fail").zipPar(ZIO.succeed("succeed")).foldCause(_ => 0, _ => 0)
```

This will succeed because `foldCause` recovers from all errors. But a fiber failure warning will still be printed to the console because one of the forked fibers in `raceWith`, which `zipPar` is implemented in terms of, ended in failure. This is undesirable in general and particularly confusing to new users who don't understand why the warning is displayed or that the success of their effect is independent of this warning.

This PR implements a new combinator `forkInternal` that forks an effect that will be executed without unhandled failures being reported. This can be use in combinators such as `raceWith` that implement their own logic for handling failures to avoid these spurious warnings being generated.

So far I have only used this in `raceWith`, which seems to be the cause of most of the spurious warnings that are being reported, but we should be able to use this in other combinators as well.

### DIFF
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -150,6 +150,12 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
   }
 
   /**
+   * Traverses the `Exit` with an effectual function.
+   */
+  def traverse[R, E1 >: E, B](f: A => ZIO[R, E1, B]): ZIO[R, Nothing, Exit[E1, B]] =
+    fold(c => ZIO.succeed(halt(c)), a => f(a).run)
+
+  /**
    * Discards the value.
    */
   final def unit: Exit[E, Unit] = as(())
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -99,6 +99,13 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
     }
 
   /**
+   * Applies the function `f` to the successful result of the `Exit` and
+   * returns the result in a new `Exit`.
+   */
+  final def foreach[R, E1 >: E, B](f: A => ZIO[R, E1, B]): ZIO[R, Nothing, Exit[E1, B]] =
+    fold(c => ZIO.succeed(halt(c)), a => f(a).run)
+
+  /**
    * Retrieves the `A` if succeeded, or else returns the specified default `A`.
    */
   final def getOrElse[A1 >: A](orElse: Cause[E] => A1): A1 = self match {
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -217,11 +217,21 @@ trait Fiber[+E, +A] { self =>
...
   final def map[B](f: A => B): Fiber[E, B] =
-    new Fiber[E, B] {
-      def await: UIO[Exit[E, B]]        = self.await.map(_.map(f))
-      def poll: UIO[Option[Exit[E, B]]] = self.poll.map(_.map(_.map(f)))
-      def interrupt: UIO[Exit[E, B]]    = self.interrupt.map(_.map(f))
-      def inheritFiberRefs: UIO[Unit]   = self.inheritFiberRefs
+    mapM(f andThen UIO.succeed)
+
+  /**
+   * Effectually maps over the favlue the fiber computes.
```
LGLO: Typo here

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1572,6 +1576,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def <*>[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[R1, E1, (A, B)] =
     self &&& that
 
+  /**
+   * Forks an effect that will be executed without unhandled failures being
+   * reported. This is useful for implementing combinators that handle failures
+   * themselves.
+   */
```
LGLO: Thanks for doc

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: Thanks! Fixed now.

core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -149,6 +149,12 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
     case Failure(cause) => Left(FiberFailure(cause))
   }
 
+  /**
+   * Traverses the `Exit` with an effectual function.
+   */
+  def traverse[R, E1 >: E, B](f: A => ZIO[R, E1, B]): ZIO[R, Nothing, Exit[E1, B]] =
```
jdegoes: 👍 Let's add `foreach` alias too, if it doesn't exist.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1572,6 +1576,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+  private[zio] final def forkInternal: ZIO[R, Nothing, Fiber[E, A]] =
+    run.fork.map(_.mapM(IO.done))
```
jdegoes: Beautiful!

PR #1951, opened 2019-10-11
ZIO Test: Support Test Annotations

### LINKED ISSUE #1899: ZIO Test: Add Support for Test Annotations
Now that we have #1664 merged I think the next area to work on for infrastructure for ZIO Test is supporting the test annotations functionality that @jdegoes introduced.

@jdegoes With all tests being effectual now, I am wondering if we can do this in a slightly different way. What if we wrapped the `TestAnnotationMap` in a `Ref` and made it available through the environment type as part of `TestEnvironment`? Conceptually, you could think of test annotations as a more structured logging service or alternatively as a super polymorphic version of the writer monad effect where the normal writer monad can support writing to a log with values of one type that forms a monoid, whereas test annotations support logging arbitrarily many types of values that each form a monoid. I think this could provide a very nice way for us to allow tests to write to and read from test annotations without having to further complicate the return type of tests, which is already not simple, or manually thread the state through. What do you think?

adamgfraser:
Resolves #1899.

Implements support for test annotations by creating a new `TestAnnotations` service that is automatically provided by the `TestEnvironment` and allows reading and writing to a test annotation map backed by a `Ref`. Uses this functionality to implement a new `TestAspect#timed` that times tests. For example, here is the output from using `timed` on `BoolAlgebraSpec` that would be displayed after the normal test results:

```
[info] Timed 21 tests in 16 s 289 ms:
[info]   + hashCode is consistent with equals: 4 s 145 ms (25.45%)
[info]   + and distributes over or: 2 s 421 ms (14.87%)
[info]   + and is associative: 2 s 390 ms (14.68%)
[info]   + De Morgan's laws: 2 s 270 ms (13.94%)
[info]   + and is commutative: 2 s 163 ms (13.28%)
[info]   + or is associative: 951 ms (5.84%)
[info]   + or distributes over and: 866 ms (5.32%)
[info]   + or is commutative: 603 ms (3.70%)
[info]   + double negative: 441 ms (2.71%)
[info]   + map transforms values: 4 ms (0.03%)
[info]   + isSuccess returns whether result is success: 4 ms (0.03%)
[info]   + implies returns implication of two values: 3 ms (0.02%)
[info]   + either returns disjunction of two values: 3 ms (0.02%)
[info]   + foreach combines multiple values: 3 ms (0.02%)
[info]   + isFailure returns whether result is failure: 2 ms (0.02%)
[info]   + collectAll combines multiple values: 2 ms (0.02%)
[info]   + failures collects failures: 2 ms (0.02%)
[info]   + both returns conjunction of two values: 2 ms (0.01%)
[info]   + all returns conjunction of values: 2 ms (0.01%)
[info]   + any returns disjunction of values: 2 ms (0.01%)
[info]   + as maps values to constant value: 2 ms (0.01%)
```

The tests are sorted by duration and the percentage of time spent on each test is shown so it is easy to identify which tests are driving overall time for test execution.


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -179,6 +179,17 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
       case TestCase(label, test)         => TestCase(label, test.map(f))
     }
 
+  /**
+   * Returns a new spec with tests remapped using the specified effectual function.
+   */
+  final def mapTestM[R1 <: R, E1 >: E, L1 >: L, T1](f: T => ZIO[R1, E1, T1]): Spec[R1, E1, L1, T1] =
+    caseValue match {
+      case SuiteCase(label, specs, exec) =>
+        Spec.suite(label, specs.map(_.map(_.mapTestM(f))), exec)
```
jdegoes: `.map(_.map(_.mapTest(...)))`

😆

test/shared/src/main/scala/zio/test/TestAnnotations.scala
```diff
@@ -0,0 +1,86 @@
...
+trait TestAnnotations {
+  val testAnnotations: TestAnnotations.Service[Any]
+}
+
+object TestAnnotations {
+
+  trait Service[R] {
+    def annotate[V](key: TestAnnotation[V], value: V): ZIO[R, Nothing, Unit]
+    def get[V](key: TestAnnotation[V]): ZIO[R, Nothing, V]
+    val testAnnotationMap: ZIO[R, Nothing, TestAnnotationMap]
```
jdegoes: Do we have to expose this `map`? It would be nice to make this as tiny a surface area as possible, which means `annotate` and `get`.

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -90,7 +90,7 @@ package object test extends CheckVariants {
   /**
    * An `ExecutedSpec` is a spec that has been run to produce test results.
    */
-  type ExecutedSpec[+L, +E, +S] = Spec[Any, Nothing, L, Either[TestFailure[E], TestSuccess[S]]]
+  type ExecutedSpec[+L, +E, +S] = Spec[Any, Nothing, L, (Either[TestFailure[E], TestSuccess[S]], TestAnnotationMap)]
```
jdegoes: I like the idea that executing the spec gives you back the annotation map (at each node) always. 👍 

It may be empty in some configurations, that's fine.

By the way, this structure is _beginning_ to get complex enough that we should consider a custom type:

```scala
sealed trait TestResult[+E, +S] {
  def annotations: TestAnnotationMap
}
object TestResult {
  final case class Failed[E](failure: TestFailure[E], annotations: TestAnnotationMap) extends TestResult[E, S]
  final case class Succeeded[A](success: TestSuccess[S], annotations: TestAnnotationMap) extends TestResult[E, S]
}
```

Then we can simplify to:

```scala
type ExecutedSpec[+L, +E, +S] = 
  Spec[Any, Nothing, L, TestResult[E, S]]
```

jdegoes: (I think we're already using `TestResult` name for something...)

test/shared/src/main/scala/zio/test/TestAnnotations.scala
```diff
@@ -0,0 +1,86 @@
...
+    makeService.map { service =>
+      new TestAnnotations {
+        val testAnnotations = service
+      }
+    }
+
+  /**
+   * Constructs a new `TestAnnotations` service.
+   */
+  def makeService: UIO[TestAnnotations.Service[Any]] =
```
jdegoes: Is this actually helpful, because there is just one annotation map here?

test/shared/src/main/scala/zio/test/TestAnnotations.scala
```diff
@@ -0,0 +1,86 @@
...
+   * Accesses a `TestAnnotations` instance in the environment and retrieves the
+   * annotation of the specified type, or its default value if there is none.
+   */
+  def get[V](key: TestAnnotation[V]): ZIO[TestAnnotations, Nothing, V] =
+    ZIO.accessM(_.testAnnotations.get(key))
+
+  /**
+   * Constructs a new `TestAnnotations` instance.
+   */
+  def make: UIO[TestAnnotations] =
```
jdegoes: Ditto for this — is it helpful?

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -77,6 +77,7 @@ object DefaultTestReporter {
       res <- render(executedSpec.mapLabel(_.toString))
       _   <- ZIO.foreach(res.flatMap(_.rendered))(TestLogger.logLine)
       _   <- logStats(duration, executedSpec)
+      _   <- renderTimed(executedSpec).flatMap(_.fold[URIO[TestLogger, Unit]](URIO.unit)(TestLogger.logLine))
```
jdegoes: I think we are missing some type of interface something like this:

```scala
type TestAnnotationRenderer[A] = 
  TestAnnotation[A] => RenderedResult
```

Then we can push knowledge of which aspects need rendering to a higher level, and get it out of here, so the user can pass their own rendering for custom test aspects (of course, we will supply the ones for the built in test annotations like timing, etc.).

test/shared/src/main/scala/zio/test/TestAnnotations.scala
adamgfraser: If we want executing a spec to give you the annotation map we need a way to retrieve the annotation map. We could make it package private, though that would mean any user implementing their own test executor would not be able to access the annotation map.

test/shared/src/main/scala/zio/test/package.scala
adamgfraser: Yes. After getting this merged my next priority is to simplify some of these types.

test/shared/src/main/scala/zio/test/TestAnnotations.scala
adamgfraser: It is definitely questionable. For the vast majority of users they are just going to use the `TestEnvironment` and maybe `provideSome` to enrich that with additional functionality. But if a power user wants to provide their own environment type this would allow them to include test annotations functionality in that type.

test/shared/src/main/scala/zio/test/TestAnnotations.scala
adamgfraser: See above.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
adamgfraser: Agreed, I was thinking about whether test annotations should contain their own logic for rendering them but was worried that was coupling data with presentation too much. This is a better way to do that. Let me work on adding. Though I think we may need to modify it slightly to be something like

```scala
type TestAnnotationRenderer[A] =
  (TestAnnotation[A], ExecutedSpec[Any, Any, Any]) => String
```

Conceptually I think we want the renderer to be able to introspect the test annotations from all the tests in a spec (e.g. to sort the rendered test times from longest to shortest or to report at the level of suites instead of individual tests).

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -109,6 +110,36 @@ object DefaultTestReporter {
...
   }
 
+  private def renderTimed[L, E, S](executedSpec: ExecutedSpec[L, E, S]): UIO[Option[String]] = {
+    val results = executedSpec.fold[UIO[Vector[(L, Duration)]]] {
+      case Spec.SuiteCase(_, executedSpecs, _) =>
+        executedSpecs.flatMap(UIO.collectAll(_).map(_.foldLeft(Vector.empty[(L, Duration)])(_ ++ _)))
+      case Spec.TestCase(label, test) =>
+        test.map {
+          case (_, annotationMap) =>
+            val d = annotationMap.get(TestAnnotation.Timing)
```
jdegoes: We should think how to pull this logic out of here and push it higher.

test/jvm/src/main/scala/zio/test/environment/TestEnvironment.scala
```diff
@@ -29,30 +30,33 @@ case class TestEnvironment(
...
-        size     <- Sized.makeService(100)
-        system   <- TestSystem.makeTest(TestSystem.DefaultData)
-        blocking = Blocking.Live.blocking
-      } yield new TestEnvironment(blocking, clock, console, live, random, clock, size, system)
+        clock           <- TestClock.makeTest(TestClock.DefaultData)
+        console         <- TestConsole.makeTest(TestConsole.DefaultData)
+        live            <- Live.makeService(new DefaultRuntime {}.Environment)
+        random          <- TestRandom.makeTest(TestRandom.DefaultData)
+        size            <- Sized.makeService(100)
+        testAnnotations <- TestAnnotations.makeService
```
jdegoes: Don't we need (in the most general possible case), one `TestAnnotationMap` per test? Of course, we can compose all `TestAnnotation` inside a suite into one, for the suite itself, and so forth; but it seems to me to have the most flexibility, we will need one per test.

(This, btw, explains my confusion about the interface for `TestAnnotations`, specifically the utility about the `make*` methods.)

adamgfraser: Yes, I think we want one `TestAnnotationMap` per test. Right now I am doing this because the `TestExecutor` provides a separate copy of the `TestEnvironment` to each test. Does that make sense or would you approach it differently?

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
adamgfraser: Yes. I think once we have `TestAnnotationRenderer`, `TestReporter` can take a `List[TestAnnotationRenderer]` as input, and the `DefaultTestReporter` in particular can have a set of default renderers in addition to whichever ones are passed to it. Then the `DefaultTestReporter` can traverse the executed spec, report all annotations it has renderers for, and either ignore or report an error for any others.

test/jvm/src/main/scala/zio/test/environment/TestEnvironment.scala
jdegoes: Ah, I didn't see that. Where is that code?

adamgfraser: The `TestExecutor` uses `Spec#provideManaged` [here](https://github.com/zio/zio/blob/ef2375e434dce8d2e8fa6c9955ba03f77f822454/test/shared/src/main/scala/zio/test/TestExecutor.scala#L27) and `provideManaged` (as opposed to `provideManagedShared`) provides a separate copy of the resources to each test [here](https://github.com/zio/zio/blob/ef2375e434dce8d2e8fa6c9955ba03f77f822454/test/shared/src/main/scala/zio/test/Spec.scala#L186).

jdegoes: So the only problem I see is that if the user chooses to share an environment across tests (e.g. to share database connection costs), then suddenly the annotations will not be per test, but rather, "global" per the sharing. 

It would be nice if we could guarantee the same annotations regardless of environmental sharing; and also, if the annotations for suites were computed through semigroup composition of the annotations of their contents (it's a stronger guarantee).

What do you think?

adamgfraser: I tried that initially and ran into some issues. For a lot of annotations we don't actually want to combine them using the semigroup for the test annotation. For example, for the `Timing` annotation it isn't that interesting to just add up the duration for all the tests, we want to combine them using more of a "map merge" semigroup with mappings from labels to durations. The more serious issue is that we want to write information to test annotations through test aspects but test aspects don't have access to test labels. So if we have each test write its duration to a shared annotation map we don't have any keys to associate with the durations in a sensible way to then aggregate.

Maybe another way we could address is by taking the `TestAnnotations` service out of the test environment and having the `TestExecutor` provide it separately to all the tests so it would be impossible or at least very hard for a user to provide a shared annotation map.

I do think the semigroup solution is the most natural one so would love to do that if possible but not sure how to overcome the issues there.

jdegoes: I'll write up some ideas here soon!

jdegoes: If we use a fiber ref, then during the evaluation of tests, the executor can create a fresh map, and replace the contents of the fiber ref with that fresh map, and then after the test, capture the map, and in the end, generate one map per test, and one map per suite. The suite maps would be generated by using the annotation append functionality, which merges collisions, which means that suite maps would always be an aggregation of their children.

Then the renderer can look at the map and decide what to render, per node, based on the desired rendering test annotations (timing, failure count, success count, etc.).

adamgfraser:

regiskuckaertz:
Looks like this need a rebase and then 🚀

jdegoes:
We spent some time on this and it looked really good.

adamgfraser:
Updated now. Sorry for the delay on this.

jdegoes:
Looks fantastic! Would be nice to see how the "new" approach renders the timings, but anyway, looks good to merge!

adamgfraser:
@jdegoes Here is an example of what it looks like right now. The annotations are shown along with each test and suite annotations are composed using the monoid.

I really like your idea of having some kind of more structured tabular output in HTML or something like that. With the current console output the rendering of the annotations are interleaved with the test results so if there are too many annotations it can make it difficult to see the test results and there also isn't any ability to, for example, sort by execution times.

It seems like an ideal end point might be where certain annotations useful for interactive testing were rendered in the console (e.g. seeds of failing tests), and other annotations with more "metrics" were written to a more structured format for visualization or analysis.

image

PR #1952, opened 2019-10-11
Eliminate space leaks on ZStreams that concatenate infinitely

iravid:


iravid:
This change fixes a leak with streams that concatenate infinitely. Previously, any stream that is used with `.forever`, which is defined as `def forever = self ++ forever`, would allocate a ZManaged scope and resources for every repetition. This would lead to leaks of `zio.internal.Stack` frames. 

Instead, we now keep concatenations in an initial representation that can be introspected before evaluation. During evaluation, we take care to reuse the same ZManaged scope and not introduce additional ones.

This change currently does not fix space leaks caused by streams that recurse infinitely with ZStream#flatMap. It would seem that recursing with ZStream#flatMap isn't that useful (because this sort of infinite "downward" recursion indicates that no progress is made on the outer stream), but it's actually useful for embedding effects that control the recursion in the stream.

For example:
```scala
def go: Stream[Random, Int] = 
  Stream.fromEffect(random.nextInt).flatMap { i =>
    if (i % 2 == 0) Stream.succeed(i) ++ go
    else Stream.succeed(i)
  }
```

This will introduce another scope for every iteration until an odd number is hit. It is tempting to see if the same approach can solve this issue. I'll look into that over the next few weeks but if anyone would like to tackle that, feel free.

iravid:
Also, big thanks to @sebver who noticed the issue and shared the heap dumps with us.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -226,16 +226,12 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
-            resR <- reserve
-                     .flatMap(res => finalizers.update(res.release :: _).as(res))
-                     .uninterruptible
-            r <- resR.acquire
-            resR1 <- f0(r).reserve
-                      .flatMap(res => finalizers.update(res.release :: _).as(res))
-                      .uninterruptible
-            r1 <- resR1.acquire
-          } yield r1,
+          acquire = ZIO.uninterruptibleMask { restore =>
```
mschuwalow: :+1:

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -847,7 +850,7 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * elements of this stream, and then emit the elements of the `other` stream.
    */
   final def concat[R1 <: R, E1 >: E, A1 >: A](other: => ZStream[R1, E1, A1]): ZStream[R1, E1, A1] =
```
mschuwalow: What exactly ended up being the issue with the implementation of concat based on ZManaged#switchable that you posted in discord? I though it was a bit nicer than introducing structure

regiskuckaertz:
Massive 💪

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,61 @@ object ZStream {
...
+        def go(
+          doneRef: Ref[Boolean],
+          currPull: Ref[Pull[R, E, A]],
+          nextPull: Ref[Option[() => Structure[R, E, A]]],
+          switchPull: ZManaged[R, E, Pull[R, E, A]] => ZIO[R, E, Pull[R, E, A]]
+        ): ZIO[R, Option[E], A] =
+          doneRef.get.flatMap { done =>
+            if (done) IO.fail(None)
+            else
+              currPull.get.flatten.foldM(
```
regiskuckaertz: You can use `catchAll` to avoid having to deal with the happy path.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,61 @@ object ZStream {
...
+                {
+                  case e @ Some(_) => ZIO.fail(e)
+                  case None =>
+                    nextPull.get.flatMap {
+                      case None => doneRef.set(true) *> IO.fail(None)
```
regiskuckaertz:
```suggestion
                      case None => doneRef.set(true) *> Pull.end
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,61 @@ object ZStream {
       }
   }
 
+  private[stream] sealed abstract class Structure[-R, +E, +A] {
+    def process: ZManaged[R, E, ZIO[R, Option[E], A]]
```
regiskuckaertz:
```suggestion
    def process: ZManaged[R, E, Pull[R, E, A]]
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
-        stream = {
-          def repeated: ZStream[R1 with Clock, E, C] = ZStream.unwrap {
-            for {
-              scheduleState <- schedStateRef.get
-              decision      <- schedule.update((), scheduleState)
-              s2 = if (decision.cont)
-                ZStream
-                  .fromEffect(schedStateRef.set(decision.state) *> clock.sleep(decision.delay))
-                  .drain ++ self.map(f) ++ Stream.succeed(g(decision.finish())) ++ repeated
+        switchPull    <- ZManaged.switchable[R1 with Clock, E, ZIO[R1 with Clock, Option[E], C]]
```
regiskuckaertz:
```suggestion
        switchPull    <- ZManaged.switchable[R1 with Clock, E, Pull[R1 with Clock, E, C]]
```
😄

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+        currPull      <- switchPull(self.map(f).process).flatMap(as => Ref.make(as)).toManaged_
+        doneRef       <- Ref.make(false).toManaged_
+        pull = {
+          def go: ZIO[R1 with Clock, Option[E], C] =
+            doneRef.get.flatMap { done =>
+              if (done) ZIO.fail(None)
```
regiskuckaertz:
```suggestion
              if (done) Pull.end
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
-              decision      <- schedule.update((), scheduleState)
-              s2 = if (decision.cont)
-                ZStream
-                  .fromEffect(schedStateRef.set(decision.state) *> clock.sleep(decision.delay))
-                  .drain ++ self.map(f) ++ Stream.succeed(g(decision.finish())) ++ repeated
+        switchPull    <- ZManaged.switchable[R1 with Clock, E, ZIO[R1 with Clock, Option[E], C]]
+        currPull      <- switchPull(self.map(f).process).flatMap(as => Ref.make(as)).toManaged_
+        doneRef       <- Ref.make(false).toManaged_
+        pull = {
+          def go: ZIO[R1 with Clock, Option[E], C] =
```
regiskuckaertz:
```suggestion
          def go: Pull[R1 with Clock, E, C] =
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+            doneRef.get.flatMap { done =>
+              if (done) ZIO.fail(None)
               else
-                ZStream.empty
-            } yield s2
-          }
-          self.map(f) ++ repeated
+                currPull.get.flatten.foldM(
+                  {
+                    case e @ Some(_) => ZIO.fail(e)
```
regiskuckaertz: Same, just in case:
```suggestion
                    case Some(e) => Pull.fail(e)
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                    case None =>
+                      schedStateRef.get
+                        .flatMap(schedule.update((), _))
+                        .flatMap { decision =>
+                          if (!decision.cont) doneRef.set(true) *> ZIO.fail(None)
```
regiskuckaertz:
```suggestion
                          if (!decision.cont) doneRef.set(true) *> Pull.end
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,23 +1827,39 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+                          else {
+                            val nextPull =
+                              (ZStream.fromEffect(clock.sleep(decision.delay)).drain ++
+                                self.map(f) ++ Stream.succeed(g(decision.finish()))).process
+
+                            switchPull(nextPull).mapError(Some(_)).tap(currPull.set(_)) *>
```
regiskuckaertz: Have you noticed that pattern is coming up all the time with `switchable`? I wonder if we could refactor without having to keep `currPull` and `nextPull` refs around.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1045,38 +1048,42 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+            bs          <- restore(reservation.acquire)
+            _           <- finalizer.set(reservation.release)
+            _           <- currPull.set(bs)
+          } yield ()).mapError(Some(_))
+        }
+      }
+
+      currPull.get.flatten.catchAll {
+        case e @ Some(e1) =>
+          (finalizer.get.flatMap(_(Exit.fail(e1))) *> finalizer.set(_ => UIO.unit)).uninterruptible *> ZIO.fail(
```
regiskuckaertz:
```suggestion
          (finalizer.get.flatMap(_(Exit.fail(e1))) *> finalizer.set(_ => UIO.unit)).uninterruptible *> Pull.fail(e1)
```

vasilmkd:
I second @regiskuckaertz's suggestions.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Yeah, there's some convenience to be had here: we could have the `switchable` constructor return something like:
```
trait Switchable[R, E, A] {
  def switch(m: ZManaged[R, E, A]): ZIO[R, E, A]
  def get: ZIO[R, E, A]
}
```

So the `Ref` is embedded in that.

iravid:
Thanks for reviewing @regiskuckaertz! At some point in this PR I deleted `Pull` so those are the artifacts ;-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Good question and good to have it written in this PR for posterity.

The previous `++` would do something like this (written minimally on purpose):
```scala
def ++(other: => ZStream) = 
  ZStream {
    for {
      switched <- Ref.make(false).toManaged_
      currPull <- Ref.make(...)
      pull = {
        // get the current pull from the Ref, and pull it
        // if it ended and !switched, switch over to `other`
        // otherwise end
      }
    } yield pull
  }
```

As you can see, it allocates a bunch of resources that are retained for the existence of the stream `(self ++ other)`.

Now, let's consider what happens in `self.forever` (defined as `forever = self ++ forever`) - or any stream that concatenates recursively. When we evaluate `++`, it allocates these refs and starts pulling `self`. Once `self` is done, the ZManaged for `other` is opened and that one starts being pulled.

But, `other` is actually another `++`, so more `Ref` instances are allocated! This happens on every switch to the RHS, so every recursion allocates more `Ref`. That would be fine if we could somehow stop retaining the current `Ref`, but we can't "exit" the scope of the current ZManaged. This is the root of the problem. Once we're inside a scope, we can't exit it.

So this formulation is heap-unsafe. I tried pretty hard, but I couldn't work around it properly. I think that the way to solve it is to interpret the recursive structure "from outside", before entering the scope that is recursive.

I hope this makes sense!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,61 @@ object ZStream {
...
+          currPull: Ref[Pull[R, E, A]],
+          nextPull: Ref[Option[() => Structure[R, E, A]]],
+          switchPull: ZManaged[R, E, Pull[R, E, A]] => ZIO[R, E, Pull[R, E, A]]
+        ): ZIO[R, Option[E], A] =
+          doneRef.get.flatMap { done =>
+            if (done) IO.fail(None)
+            else
+              currPull.get.flatten.foldM(
+                {
+                  case e @ Some(_) => ZIO.fail(e)
```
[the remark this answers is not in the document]
iravid: It compiles, so I don't think it's problematic :-) Just trying to avoid an extra allocation here with the `e @ Some(_)`

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Right!

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Ah it's actually not that simple. We would have to start requiring an initial resource, or `get` becomes `UIO[Option[A]]`. Beyond that, we need to think what happens if `switch` fails to acquire the resource. What would `get` return afterwards? That also pushes us towards `UIO[Option[A]]`. The problem is that sticking an `Option` in there can be a real pain for calling code.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,61 @@ object ZStream {
...
+                              nextPull.set(None) *> go(doneRef, currPull, nextPull, switchPull)
+                          case Concat(hd, tl) =>
+                            // It is extremely important in this case to *NOT* recurse using
+                            // tl.process, as that would introduce a new ZManaged scope; this
+                            // will cause space leaks when used with streams that concatenate infinitely.
+                            // Instead, we re-use the same ZManaged scope introduced by the ZManaged.switchable
+                            // below to swap the current stream being pulled with tl.
+                            switchPull(hd.process).mapError(Some(_)).tap(currPull.set) *>
+                              nextPull.set(Some(tl)) *> go(doneRef, currPull, nextPull, switchPull)
+                        }
```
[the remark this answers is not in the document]
iravid: I'm actually going to avoid pulling more things up into `Structure` because I think this is going to end up as a mini-algebra, with an additional `FlatMap` constructor.

Then I'll move the interpreter up from `Concat` (which is the only composite term requiring interpretation) into `Structure`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: Ah, I think I understand. But this is a very subtle issue O.o

Thanks for the explanation and nice fix!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1045,38 +1048,42 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
    * Returns a stream made of the concatenation in strict order of all the streams
    * produced by passing each element of this stream to `f0`
    */
-  final def flatMap[R1 <: R, E1 >: E, B](f0: A => ZStream[R1, E1, B]): ZStream[R1, E1, B] =
+  final def flatMap[R1 <: R, E1 >: E, B](f0: A => ZStream[R1, E1, B]): ZStream[R1, E1, B] = {
+    def go(
+      as: ZIO[R1, Option[E1], A],
+      finalizer: Ref[Exit[_, _] => URIO[R1, _]],
+      currPull: Ref[ZIO[R1, Option[E1], B]]
```
regiskuckaertz:
```suggestion
      currPull: Ref[Pull[R1, E1, B]]
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1045,38 +1048,42 @@ class ZStream[-R, +E, +A](val process: ZManaged[R, E, Pull[R, E, A]]) extends Se
...
+    ): ZIO[R1, Option[E1], B] = {
```
regiskuckaertz:
```suggestion
    ): Pull[R1, E1, B] = {
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2424,6 +2452,58 @@ object ZStream {
       }
   }
 
+  private[stream] sealed abstract class Structure[-R, +E, +A] {
```
jdegoes: Good idea to keep all the guts private so we can experiment throughout `1.x`.

iravid:
@jdegoes Would you like to have another look on this? Or is this good for merging?

iravid:
Merging this because it's needed to solve a memory leak with zio-kafka. @jdegoes Happy to discuss further if you have any additional thoughts!

PR #1954, opened 2019-10-12
Implement Promise#completeWith

### LINKED ISSUE #1932: Promise.complete violates Promise "set exactly once" rule
I suppose that `Promise.complete` method has very special behavior:

```
val promiseTest = 
	for {
	    p      <- Promise.make[Nothing, Int]
	    action =  random.nextInt(10).provide(Random.Live)
	    _      <- p.complete(action)
	    v1     <- p.await
	    v2     <- p.await
	    v3     <- p.await
	  } yield (v1,v2,v3) 
```

As documentation says: `A promise represents an asynchronous variable, of [[zio.IO]] type, that can be set exactly once...`, so I expect the result of `promiseTest` to be three equal values, but actually it's not:
```
run(promiseTest) 
res13: (Int, Int, Int) = (4, 0, 1)
```
And that result violates the `can be set exactly once` rule. Also we have safe method `ZIO.to` to complete promise with effect which internally use `Promise.complete`. And this version works correctly:
```
for {
    p      <- Promise.make[Nothing, Int]
    action =  random.nextInt(10).provide(Random.Live)
    _      <- action.to(p).fork
    v1     <- p.await
    v2     <- p.await
    v3     <- p.await
  } yield (v1,v2,v3) 
```
 Maybe `Promise.complete` should be the part of private API? 
 Because I spent a lot of time to find this subtle bug trying to complete promise with some effect

adamgfraser:
Resolves #1932.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
@@ -15,10 +15,25 @@ object PromiseSpec
         },
         testM("complete a promise using complete") {
           for {
-            p <- Promise.make[Nothing, Int]
-            s <- p.complete(IO.succeed(14))
-            v <- p.await
-          } yield assert(s, isTrue) && assert(v, equalTo(14))
+            p  <- Promise.make[Nothing, Int]
+            r  <- Ref.make(13)
+            s  <- p.complete(r.update(_ + 1))
+            v1 <- p.await
+            v2 <- p.await
+          } yield assert(s, isTrue) &&
+            assert(v1, equalTo(14)) &&
+            assert(v2, equalTo(14))
+        },
+        testM("complete a promise using completeWith") {
+          for {
+            p  <- Promise.make[Nothing, Int]
+            r  <- Ref.make(13)
+            s  <- p.completeWith(r.update(_ + 1))
+            v1 <- p.await
+            v2 <- p.await
+          } yield assert(s, isTrue) &&
+            assert(v1, equalTo(14)) &&
+            assert(v2, equalTo(15))
         },
         testM("fail a promise using fail") {
           for {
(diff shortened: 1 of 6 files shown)


core-tests/shared/src/test/scala/zio/PromiseSpec.scala
```diff
@@ -29,10 +44,25 @@ object PromiseSpec
         },
         testM("fail a promise using complete") {
           for {
-            p <- Promise.make[String, Int]
-            s <- p.complete(IO.fail("error with done"))
-            v <- p.await.run
-          } yield assert(s, isTrue) && assert(v, fails(equalTo("error with done")))
+            p  <- Promise.make[String, Int]
+            r  <- Ref.make[List[String]](List("first error", "second error"))
```
LGLO: Cosmetic: do we need type ascription here?

adamgfraser: Great point! We definitely don't need that, Fixed now.

adamgfraser:
@LGLO Good point about the documentation. I updated in my latest commit. Let me know if there is anything you would change.

LGLO:
Great work!
@jdegoes Since people were lost with previous API and now it's more intuitive (subjective) maybe it will be worth a tweet that `complete` will change?

hepin1989:
add one more `completeAsyncWith`

adamgfraser:
@hepin1989 Sorry, can you clarify the signature and functionality of the proposed `completeAsyncWith`?

`completeWith` already takes an `IO[E, A]` that could itself involve asynchronous effects. `completeWith` is also already asynchronous in some sense in that the effect will not be evaluated until a fiber retrieves the promise. For example, this test will pass:

```scala
testM("effect in completeWith is not evaluated until value of promise is retrieved") {
  for {
    p <- Promise.make[String, Int]
    _ <- p.completeWith(IO.never *> IO.succeed(1))
  } yield assert(true, isTrue)
}
```

Finally, you could always fork `completeWith` or any other combinator.

core/shared/src/main/scala/zio/Promise.scala
```diff
@@ -86,10 +86,26 @@ class Promise[E, A] private (private val state: AtomicReference[State[E, A]]) ex
   final def done(io: IO[E, A]): UIO[Boolean] = complete(io)
 
   /**
-   * Completes the promise with the specified result. If the specified promise
-   * has already been completed, the method will produce false.
+   * Completes the promise with the result of the specified effect. If the
+   * promise has already been completed, the method will produce false.
```
jdegoes: Maybe add a note that `completeWith` will be much faster, so investigate using that if it makes sense?

adamgfraser: Done.

adamgfraser:
@jdegoes At this point the only code that uses `complete` is the deprecated `Promise.done` method so we are good there.

PR #1957, opened 2019-10-12
introduce ZEnv decorators

### LINKED ISSUE #1953: Introduce ZEnv decorators for different ZIO modules
Introducing a `ZEnv.scala` file, which will house a companion object for the type alias `ZEnv` (defined in `zio` package object, through `ZEnvDefinition`), we can place a number of case classes that can offer decorating behavior.

```scala
object ZEnv {
  def clockDecorator(f: Clock.Service[Any] => Clock.Service[Any]): ZEnv => ZEnv = ???
  def systemDecorator(f: System.Service[Any] => System.Service[Any]): ZEnv => ZEnv = ???
  ...
}
```

With this functionality in place, it will become possible to write:

```scala
effect.provideSome(ZEnv.clockDecorator(...))
```

for example, to "weave" an aspect into `Clock`.

mschuwalow:
#resolve #1953 

Also, there was a stray reference to DefaultRuntime#Environment that I missed in the BuildHelper.

Interestingly, we can write a fully generic version of this in zio-macros without introducing any new macros. I'll open a pr there later.

/cc @adamgfraser @jdegoes

### DIFF
--- a/core/js/src/main/scala/zio/ZEnv.scala
+++ b/core/js/src/main/scala/zio/ZEnv.scala
@@ -53,6 +53,9 @@ object ZEnv {
   def randomDecorator(f: Random.Service[Any] => Random.Service[Any]): ZEnv => ZEnv =
     zEnvDecorator(randomDecorator = f)
 
+  /**
+   * Create a decorator which modifies chosen parts of a Zenv.
+   */
   def zEnvDecorator(
     clockDecorator: Clock.Service[Any] => Clock.Service[Any] = identity,
     consoleDecorator: Console.Service[Any] => Console.Service[Any] = identity,
--- a/core/js/src/main/scala/zio/ZEnv.scala
+++ b/core/js/src/main/scala/zio/ZEnv.scala
@@ -33,63 +33,37 @@ object ZEnv {
    * }}}
    */
   def clockDecorator(f: Clock.Service[Any] => Clock.Service[Any]): ZEnv => ZEnv =
-    old =>
-      new ZEnvImpl(
-        clock0 = f(old.clock),
-        console0 = old.console,
-        system0 = old.system,
-        random0 = old.random
-      )
+    zEnvDecorator(clockDecorator = f)
 
   /**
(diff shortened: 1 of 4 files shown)


core/js/src/main/scala/zio/ZEnv.scala
```diff
@@ -0,0 +1,95 @@
...
+   */
+  def randomDecorator(f: Random.Service[Any] => Random.Service[Any]): ZEnv => ZEnv =
+    old =>
+      new ZEnvImpl(
+        clock0 = old.clock,
+        console0 = old.console,
+        system0 = old.system,
+        random0 = f(old.random)
+      )
+  private[ZEnv] class ZEnvImpl(
```
jdegoes: If you accept 4 functions here, to decorate all, you can pass identity in for all but one.

I might also call the class ZEnvDecorator

core/jvm/src/main/scala/zio/ZEnv.scala
```diff
@@ -0,0 +1,112 @@
...
+import zio.blocking.Blocking
+
+object ZEnv {
+
+  /**
+   * Lifts a function that decorates a [[Clock.Service[Any]]] to a function that decoratec an entire ZEnv.
+   *
+   * Use this with [[ZIO#provideSome]] for maximum effect.
+   * {{{
+   *   clock.sleep(1.second).provideSome(ZEnv.clockDecorator(oldClock => ???))
```
jdegoes: Great ScalaDoc

neko-kai:
Just my 2c here on naming – why `decorator`? Could be `mapBlocking`, `mapRandom`, etc...

jdegoes:
I like Kai's suggestion for names. Then we can have mapAll for the one that maps all the services.

mschuwalow:
> I like Kai's suggestion for names. Then we can have mapAll for the one that maps all the services.

Agreed. I'll push an updated version in a bit

jdegoes:
Really nice work! Going to make pinpoint modifications of part of a standard environment really easy.

PR #1959, opened 2019-10-13
WIP migrate to Zio Test

unclebob418:
#1898 

Think I've got the first one sorted. @adamgfraser can you please take a look?
If I'm on the right track I'll tackle a few more

Some questions that I wasn't able to figure out for myself while doing this:

1. When do I use '{ ... }' vs '( ... )' for suite (for the 2nd parameter list)? [doco](https://zio.dev/docs/howto/howto_test_effects) contains both but scalafmt and intellij don't seem to like the braces on suite for this one
2. What's launching these tests when I run sbt test? build.sbt seems to indicate they're being launched from zio.test.TestMain but, following @darl's lead, I've removed TestSpec from TestMain and it's still running these tests (and passing! I must've done something right...)
3. How do you pronounce "@@", and more importantly, what does it mean?


adamgfraser:
@unclebob418 This looks great! To answer your questions:

1. Use parentheses for the second parameter for `suite` if you want to take advantage of Scala's variable arguments syntax. The signature of `suite` is 

```scala
final def suite[R, E, L, T](label: L)(specs: Spec[R, E, L, T]*): Spec[R, E, L, T]
```

The second parameter takes a variable number of arguments to make it easy for including multiple tests in a suite. But variable arguments are only supported if you use parentheses.

```scala
suite("brackets okay") {
  test("first test")(???)
},
suite("brackets won't work") {
  test("first test")(???), // does not compile
  test("second test")(???)
},
suite("parentheses will")(
  test("first test")(???),
  test("second test")(???)
)
```

2. There is a ZIO Test SBT Runner that automatically detects any object that extends `AbstractRunnableSpec` and runs it when you use standard SBT commands such as `test` and `testOnly`. Your spec extends `ZIOBaseSpec`, which extends `DefaultRunnableSpec`, which in turn extends `AbstractRunnableSpec` so it automatically run when you run tests. The minimal test suite that we were migrating from didn't have an SBT test runner tests that haven't been migrated have to be included in `TestMain` and then we have a separate task for running that.

3.  I I normally just call it "at at". It is an alias for `TestAspect#apply` so you could also call it that. A test aspect is something that modifies a test in some way. It could time out a test if it doesn't pass in a certain amount of time, run a test 100 times to make sure it is not flaky, only run a test on javascript, or require a test to fail. You can see the test aspects we currently have in `TestAspects.scala`. `@@` is just a convenient way to write test aspects that allows you to do things like `test("myTest)(...) @@ timeout(60.seconds)`. You can always rewrite a statement like this to `timeout(60.seconds) { test("myTest")(...) }` using the `TestAspect#apply`.

unclebob418:
Thanks @adamgfraser; sorry, I should have picked up on that variable arguments one sorry!

I'll take a look at TestAspect and see if I can add something useful to the how-to documentation. Is that OK to do under this PR or should I add another issue/open a new PR since it's not really part of migrating to the new test suite?

adamgfraser:
More documentation would be fantastic! Why don't you open a separate PR for that so we can keep things clean? Probably the best place for the content to go would be the `test-effects.md` file, which we're trying to make a bit of a user guide for ZIO Test.

unclebob418:
@adamgfraser  TestAspectSpec has been moved. I did the move as a separate commit to make reviewing the changes easier. I learnt a lot but this one didn't feel quite right - I'm hoping it was just the runabout with testing the aspects, but it felt a bit hacky for a lot of tests. They do pass though so that's a win!

Particular concerns I have:

1. I couldn't create a `private def failsWithException` in the spec scope - I've just thrown this in a package object which works, but I'm hoping there's a better way to do this - do you have any suggestions?
2. For the last 2 tests related to timeouts, I had to change the type of the ZSpec to get the code to compile - in both cases I don't understand what the previous line is doing (`assertM(ZIO.never *> ZIO.unit, equalTo(()))`) so I'm not sure if I've broken the integrity of the test by doing this. I've put comments in with the original type.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+          val spec = test("test")(assert(true, equalTo(false))) @@ failure(
+            isCase[TestFailure[Any], Any]("Assertion", {
+              case TestFailure.Assertion(result) => Some(result); case _ => None
+            }, anything)
+          )
+          succeeded(spec).map(assert(_, isTrue))
+        },
+        testM("timeout makes tests fail after given duration") {
+          val spec = (testM("test") {
+            assertM(ZIO.never *> ZIO.unit, equalTo(()))                    //todo what is this doing?
```
adamgfraser: We're just creating a test that will never terminate to simulate a situation where a user accidentally wrote a non-terminating test (e.g. a bug created a deadlock condition) to make sure that we can timeout the test after the specified duration.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+          }: ZSpec[TestEnvironment, Any, String, Any]) @@ timeout(1.nanos) //todo confirm type change didn't break anything (was typed ZSpec[Live[Clock], Any, String, Any])
```
adamgfraser: There is a type inference issue with using the syntax here. I just opened an issue for it in #1978. For now you can resolve by using `TestAspect.timeout(1.nanos) { // your test goes here }`.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+        },
+        testM("failure does not make a test pass if the specified failure does not match") {
+          val spec = test("test")(assert(throw new RuntimeException(), isFalse)) @@ failure(
+            isCase[TestFailure[String], Cause[String]]("Runtime", {
+              case TestFailure.Runtime(e) => Some(e); case _ => None
+            }, equalTo(Cause.fail("boom")))
+          )
+          failed(spec).map(assert(_, isTrue))
+        },
+        testM("failure makes tests pass on any assertion failure") {
```
adamgfraser: I think for this one you don't need to create a separate test and execute it. Your test could just be something like:

```scala
test("failure makes tests pass on any assertion failure) {
  assert(true, isFalse)
} @@ failure
```

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+          val spec = test("test")(assert(true, equalTo(false))) @@ failure
+          succeeded(spec).map(assert(_, isTrue))
+        },
+        testM("failure makes tests pass on an expected assertion failure") {
```
adamgfraser: See above.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+          val spec   = test("JVM-only")(assert(TestPlatform.isJVM, isTrue)) @@ jvmOnly
+          val result = if (TestPlatform.isJVM) succeeded(spec) else ignored(spec)
+          result.map(assert(_, isTrue))
+        },
+        testM("failure makes a test pass if the result was a failure") {
+          val spec   = test("test")(assert(throw new java.lang.Exception("boom"), isFalse)) @@ failure
+          val result = succeeded(spec)
+          result.map(assert(_, isTrue))
+        },
+        testM("failure makes a test pass if it died with a specified failure") {
```
adamgfraser: See below.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+            _      <- execute(spec)
+            result <- ref.get
+          } yield assert(if (TestPlatform.isJVM) result else !result, isTrue)
+        },
+        testM("jvmOnly runs tests only on the JVM") {
+          val spec   = test("JVM-only")(assert(TestPlatform.isJVM, isTrue)) @@ jvmOnly
+          val result = if (TestPlatform.isJVM) succeeded(spec) else ignored(spec)
+          result.map(assert(_, isTrue))
+        },
+        testM("failure makes a test pass if the result was a failure") {
```
adamgfraser: See below.

test-tests/shared/src/main/scala/zio/test/TestMain.scala
```diff
@@ -1,10 +1,10 @@
 package zio.test
 
-import scala.concurrent.ExecutionContext.Implicits.global
-
-import zio.test.mock.MockSpecSpec
+import zio.test.TestUtils.{report, scope}
```
adamgfraser: This needs a `fmt`.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
unclebob418: Everything made a lot more sense with this!

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
unclebob418: Intellij still doesn't like it but it compiles!

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
unclebob418: thanks for the explanation.

test-tests/shared/src/test/scala/zio/test/package.scala
```diff
@@ -0,0 +1,17 @@
+package zio.test
+
+import zio.test.Assertion.{anything, isCase, isSubtype}
+
+import scala.reflect.ClassTag
+
+//todo confirm this is ok - moved to package object because I couldn't figure out how to create a private def in suite scope
```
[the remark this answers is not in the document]
unclebob418: That's an easy fix!

test-tests/shared/src/main/scala/zio/test/TestMain.scala
unclebob418: I had "optimize imports" enabled for committing... unfortunately it "optimizes" the formatting too :/ I've disabled this now

adamgfraser:
Can you resolve the merge conflict?

unclebob418:
@adamgfraser much happier with most of these now - your comment about in-lining the tests really helped, thanks! The timeout ones still look a bit messy; once the type inference issues are resolved I think this could be tidied up a lot, particularly if we only care that there was a TestTimeoutException, and not the message that it contains, I think this could be as simple as: 
```testM("timeout makes tests fail after given duration") {
  assertM(ZIO.never *> ZIO.unit, equalTo(()))
} @@ timeout(1.nanos)
  @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException]),
testM("timeout reports problem with interruption") {
  assertM(ZIO.never.uninterruptible *> ZIO.unit, equalTo(()))
} @@ timeout(10.millis, 1.nano)
  @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException])```

File: test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala [hunk 9]
```diff
@@ -0,0 +1,62 @@
...
+            output <- TestConsole.output
+          } yield assert(output, equalTo(Vector("First line\n", "Second line\n")))
+        },
+        testM("Console reads line from input") {
+          for {
+            _      <- TestConsole.feedLines("Input 1", "Input 2")
+            input1 <- console.getStrLn
+            input2 <- console.getStrLn
+          } yield {
+            assert(input1, equalTo("Input 1"))
```
adamgfraser: You need an `&&` here. Since assertions are just values a bare statement has no effect.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+            assert(input2, equalTo("Input 2"))
+          }
+        },
+        testM("Random returns next pseudorandom integer") {
```
adamgfraser: We lost something with this test. In the previous version we called `provideManaged(TestEnvironment.Value`) on each of the effects. That is important because what we are testing for in this assertion, which admittedly isn't labeled as clearly as it should be, is that the random number generators for different copies of the `TestEnvironment` are seeded with different seeds. If we don't do this both calls will use the same copy of the `TestEnvironment` so we won't be testing the same thing.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+          for {
+            i <- random.nextInt
+            //_ <- clock.sleep(1.millisecond) //TODO this causes test to fail with a timeout which is odd, but not sure why we needed to sleep here
```
adamgfraser: The issue here is that by default the `Clock` instance provided by the `TestEnvironment` is the `TestClock` which doesn't advance time unless you manually advance it with methods like `setTime` or `adjust`. To fix this you can use `Live.live(clock.sleep(1.millisecond)`. By the way, the reason we need this here is that apparently Node.js only has a resolution of 1 millisecond for its system clock and we use the live clock to seed the random number generator, so if we won't do this the test could be flaky on ScalaJS.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+            ref    <- Ref.make(false)
+            spec   = test("test")(assert(true, isTrue)) @@ js(after(ref.set(true)))
+            _      <- execute(spec)
+            result <- ref.get
+          } yield if (TestPlatform.isJS) assert(result, isTrue) else assert(result, isFalse)
+        },
+        testM("jsOnly runs tests only on ScalaJS") {
+          val spec   = test("Javascript-only")(assert(TestPlatform.isJS, isTrue)) @@ jsOnly
+          val result = if (TestPlatform.isJS) succeeded(spec) else ignored(spec)
+          result.map(assert(_, isTrue))
```
adamgfraser: You can use `assertM` so you don't have to map the result.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+            ref    <- Ref.make(false)
+            spec   = test("test")(assert(true, isTrue)) @@ jvm(after(ref.set(true)))
+            _      <- execute(spec)
+            result <- ref.get
+          } yield assert(if (TestPlatform.isJVM) result else !result, isTrue)
+        },
+        testM("jvmOnly runs tests only on the JVM") {
+          val spec   = test("JVM-only")(assert(TestPlatform.isJVM, isTrue)) @@ jvmOnly
+          val result = if (TestPlatform.isJVM) succeeded(spec) else ignored(spec)
+          result.map(assert(_, isTrue))
```
adamgfraser: `assertM`

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+          )
+        ),
+        testM("timeout makes tests fail after given duration") {
+          val spec = TestAspect.timeout(1.nanos) {
+            testM("test") {
+              assertM(ZIO.never *> ZIO.unit, equalTo(()))
+            }
+          }
+          val result = failedWith(spec, cause => cause == TestTimeoutException("Timeout of 1 ns exceeded."))
+          result.map(assert(_, isTrue))
```
adamgfraser: `assertM`

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+          }
+          val result =
+            failedWith(
+              spec,
+              cause =>
+                cause == TestTimeoutException(
+                  "Timeout of 10 ms exceeded. Couldn't interrupt test within 1 ns, possible resource leak!"
+                )
+            )
+          result.map(assert(_, isTrue))
```
adamgfraser: `assertM`

test-tests/shared/src/test/scala/zio/test/environment/SystemSpec.scala
```diff
@@ -0,0 +1,76 @@
+package zio.test.environment
+
+import zio.test.Assertion._
+import zio.test.environment.TestSystem.Data
+import zio.test.{ ZIOBaseSpec, suite, _ }
+
+object SystemSpec
```
adamgfraser: Same comment as on `ConsoleSpec` that you could just access the `TestSystem` in the environment instead of making your own if you just using the default data.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
unclebob418: Ah, thought I was being clever by figuring out I could use the TestEnvironment provided generator - I'll update the test description

unclebob418:
Thanks for the feedback and all the tips! I was a little concerned my skills weren't up to contributing to a project like this but I feel like the pieces are slowly coming together.

I'll stop migrating specs now and address the issues you've raised so we can get this one merged. That'll be a job for tomorrow though, it's past my bedtime!

test-tests/shared/src/test/scala/zio/test/environment/ConsoleSpec.scala
```diff
@@ -0,0 +1,82 @@
+package zio.test.environment
+
+import zio.test.environment.TestConsole.Data
+import zio.test.Assertion._
+import zio.test._
+
+object ConsoleSpec
```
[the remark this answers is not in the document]
unclebob418: that simplifies things a lot! I've left "reads from input" as is otherwise it would be identical to "feeds lines to input" and would fail if `feedLines` broke

adamgfraser:
We merged #1986 so you should be able to `timeout` with the `@@` syntax after merging upstream changes.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
unclebob418: sounds like another doco update!
I think most people generally expect clocks to tick by themselves... though I do understand why you'd want to manually adjust a clock for tests

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
unclebob418: i still have my imperative hat on :(

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
adamgfraser: I think that one already has fairly extensive documentation [here](https://zio.dev/docs/howto/howto_test_effects#testing-clock) but if you think it could be improved PRs are always welcome.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
unclebob418: I think I'll actually leave this test as is and create another one to check different copies of `TestEnvironment` are seeded with different seeds. I think it's important that we test the random number generator isn't just returning the same number for successive calls...

adamgfraser: Sounds good. We also have a ton of additional tests in `RandomSpec`.

test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
unclebob418: sorry, you're right, that's probably the bit with the most comprehensive doco!

unclebob418:
#1986 seems to cause other issues so have left this bit as is for now

adamgfraser:
Sounds good.

test-tests/shared/src/test/scala/zio/test/environment/ConsoleSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+import zio.test._
+import zio.test.environment.TestConsole._
+
+object ConsoleSpec
+    extends ZIOBaseSpec(
+      suite("ConsoleSpec")(
+        testM("outputs nothing") {
+          for {
+            output <- TestConsole.output
+          } yield assert(output.length, equalTo(0))
```
adamgfraser: You could do `assert(output, isEmpty)` here.

test-tests/shared/src/test/scala/zio/test/environment/ConsoleSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+            assert(message, equalTo("There is no more input left to read"))
+          }
+        },
+        testM("clears lines from output") {
+          for {
+            _      <- putStr("First line")
+            _      <- putStr("Second line")
+            _      <- clearOutput
+            output <- TestConsole.output
+          } yield assert(output, equalTo(Vector.empty))
```
adamgfraser: See comment above.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+        test("failure makes tests pass on an expected assertion failure") {
+          assert(true, equalTo(false))
+        } @@ failure(
+          isCase[TestFailure[Any], Any](
+            "Assertion",
+            { case TestFailure.Assertion(result) => Some(result); case _ => None },
+            anything
+          )
+        ),
+        /*doesn't compile, aspects do seem to work as intended below though...
```
adamgfraser: Can this be deleted? Looks like it is compiling in the test below unless I am missing something. Is there a type inference issue with test aspects we still need to address?

unclebob418: Never mind - user error! I was using `test` instead of `testM`

Intellij still gets confused with the types but the compiler has it sorted

unclebob418:
@adamgfraser I'm happy with how most of this works except for the failure on "timeout reports problem with interruption" test. Is there a better way to express this?
```
@@ failure(
            isCase[TestFailure[Any], Throwable](
              "Runtime",
              { case TestFailure.Runtime(e) => Some(e).map(_.dieOption.head); case _ => None },
              equalTo(
                TestTimeoutException(
                  "Timeout of 10 ms exceeded. Couldn't interrupt test within 1 ns, possible resource leak!"
                )
              )
            )
          )
```
it would be nice to have something like `dies` that works with `@@ failure` so we could do something like `@@ failure(diesWith(TestTimeoutException("Timeout of 10 ms exceeded")))`

I don't think this should be too hard to implement - If this isn't currently able to be done, is that something that would be useful for me to work on?

adamgfraser:
@unclebob418 The problem is that the current implementation of `failsWithException` is too specific. It combines the concept of asserting that a test dies with an exception with asserting that the exception is a certain type. As a result, you can't use it to express other assertions about the exception, such as that it has a particular message.

We can improve this by generalizing `failsWithException` to `diesWith`:

```scala
  def diesWith(assertion: Assertion[Throwable]): Assertion[TestFailure[Any]] =
    isCase(
      "Runtime", {
        case TestFailure.Runtime(c) => c.dieOption
        case _                      => None
      },
      assertion
    )
```

Notice that this is identical to the existing `failsWithException` but instead of hardcoding the assertion to `isSubtype[E](anything)` we are allowing the caller to specify the assertion.

Then we can implement the functionality of the current `failsWithException` in terms of this:

```scala
  def diesWithSubtypeOf[E](implicit C: ClassTag[E]): Assertion[TestFailure[Any]] =
    diesWith(isSubtype[E](anything))
```

This will allow you to write your code above very concisely as:

```scala
object testAspectSpec {
  test @@ fails(diesWith(equalTo(interruptionTimeoutFailure)))
}

object TestAspectSpecUtil {
  val interruptionTimeoutFailure =
    TestTimeoutException(
      "Timeout of 10 ms exceeded. Couldn't interrupt test within 1 ns, possible resource leak!"
    )
}
```

You should be able to simplify the "failure does not make a test pass if the specified failure does not match" test the same way.

unclebob418:
@adamgfraser thanks for that, my quick and dirty attempt was effectively factoring out everything I had in the `failure` assertion to a separate function in `Assertion` that took a `Throwable`:
```
final def diesWith(failure: Throwable): Assertion[TestFailure[Any]] =
    isCase[TestFailure[Any], Throwable](
      "fails with",
      { case TestFailure.Runtime(e) => Some(e).map(_.dieOption.head); case _ => None },
      equalTo(failure)
    )
```
That allows things to be a little more succinct than your implementation (`@@ failure(diesWith(TestTimeoutException("Timeout of 10 ms exceeded")))`), but suffers a similar problem to `failsWithException` to a less extent - it's expecting a value instead of another assertion, so isnt's flexible e.g. your implementation would allow:
```
object testAspectSpec {
  test @@ fails(diesWith(equalTo(interruptionTimeoutFailure) || equalTo(someOtherFailure)))
}
```
Are you happy for me to add your version to `zio.test.Assertion`? I think this is something that would be useful outside testing the test framework.

adamgfraser:
@unclebob418 I would keep this in `TestAspectSpec` for now. We can definitely open a separate PR to expose assertions like this but I think if we do that we want to be even more compositional and do something like:

```scala
def isRuntimeFailure[E](assertion: Assertion[Cause[E]]): Assertion[TestFailure[E]]
```

The cause in `TestFailure.Runtime` is not necessarily a `Die`, although it was in our tests, so this will give the user maximum flexibility to assert whether the cause was a `Die` or something else.

unclebob418:
@adamgfraser ok, I think this certainly tidies up the tests a lot anyway. Can we merge this one now?

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+        } @@ timeout(1.nanos)
+          @@ failure(diesWith(equalTo(timeoutFailure))),
+        testM("timeout reports problem with interruption") {
+          assertM(ZIO.never.uninterruptible *> ZIO.unit, equalTo(()))
+        } @@ timeout(10.millis, 1.nano)
+          @@ failure(diesWith(equalTo(interruptionTimeoutFailure)))
+      )
+    )
+object TestAspectSpecUtil {
+  def failsWithException[E](implicit ct: ClassTag[E]): Assertion[TestFailure[E]] =
```
adamgfraser: You can implement this in terms of `diesWith`. Let's update the name to something like `diesWithSubtypeOf` so it is clear how this and `diesWith` are related if someone else is working in this file.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+          assert(throw nullPointerException, isFalse)
+        } @@ failure(diesWith(equalTo(nullPointerException))),
+        test("failure does not make a test pass if it failed with an unexpected exception") {
+          assert(throw new NullPointerException(), isFalse)
+        } @@ failure(failsWithException[IllegalArgumentException])
+          @@ failure,
+        test("failure does not make a test pass if the specified failure does not match") {
+          assert(throw new RuntimeException(), isFalse)
+        } @@ failure(
+          isCase[TestFailure[String], Cause[String]]("Runtime", {
```
adamgfraser: You can implement this in terms of `diesWith`.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      "Runtime", {
+        case TestFailure.Runtime(c) => c.dieOption
+        case _                      => None
+      },
+      isSubtype[E](anything)
+    )
+  def diesWith(assertion: Assertion[Throwable]): Assertion[TestFailure[Any]] =
+    isCase(
+      "Runtime", {
+        case TestFailure.Runtime(c) => Some(c).map(_.dieOption.head)
```
adamgfraser: Can be simplified to `cause TestFailure.Runtime(c) => c.dieOption`.

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
unclebob418: that makes things much neater!

unclebob418:
done and done!

test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
adamgfraser: One more I think.

unclebob418:
hopefully that's it :)

unclebob418:
No worries, thanks for all your help & guidance! I'll start on some others in a new PR next week - have a good weekend

PR #1962, opened 2019-10-13
ZIO Test: Implement TestAspect#ifEnv

### LINKED ISSUE #1960: ZIO Test: Add ifEnv TestAspect
It would be useful to have a test aspect that runs a test only if an environment variable satisfies a provided assertion.

```scala
object TestAspect {
  ...
  def ifEnv(env: String, assertion: Assertion[String]): TestAspect[...] = ...
  ...
}
```

To be used as:

```scala
testM("A test of something") {
  ...
} @@ ifEnv("TRAVIS_CI", equalTo("true"))
```

Marks the test as ignored if the environment variable does not exist or is not set to the specified environment variable. Will need to use the live environment.

adamgfraser:
Resolves #1960.


adamgfraser:
All set.

jdegoes:
🎉

PR #1967, opened 2019-10-14
Remove Wildcard Types

### LINKED ISSUE #1759: Remove existential types from the ZIO codebase

adamgfraser:
Resolves #1759.

I found it relatively straightforward to to eliminate most uses of wildcard types in our code base by replacing uses for covariant parameters with `Any` and uses for contravariant parameters with `Nothing`.

The areas where I have not currently replaced wildcard types are in pattern matches and invariant type parameters.

Replacing wildcard types with `Any` in pattern matches results in a type erasure warning which I think is spurious but I don't know how we could silence in that particular case. With regard to arrays using `Any` actually won't generate the type erasure warning but results in different semantics where the pattern won't match. That one ends up being quite important in doing the "deep" equality check for arrays in `Assertion#equalTo`.

Invariant types don't appear very frequently in our code base. The two areas where the are used along with wildcard types is `TRef` and some of the private methods related to boxing and unboxing in `Chunk`. Since `TRef` is invariant `TRef[A]` is not a subset of `TRef[Any]` so trying to use `Any` there results in quite a few errors. It looked like it might be possible to make `TRef` covariant but I wasn't entirely sure with the `var` in `TRef` and the fact that it currently wasn't. The issues with `Chunk` could probably be resolved by casting since we are already doing a bunch of casting there but I'm not sure how much value there would be there.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/BracketTypeInferrenceSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/BracketTypeInferrenceSpec.scala
@@ -10,37 +10,37 @@ object BracketTypeInferrenceSpec {
   class E1 extends E
 
   def infersEType1: ZIO[R, E, B] = {
-    val acquire: ZIO[R, E, A]            = ???
-    val release: A => ZIO[R, Nothing, _] = ???
-    val use: A => ZIO[R, E1, B]          = ???
+    val acquire: ZIO[R, E, A]              = ???
+    val release: A => ZIO[R, Nothing, Any] = ???
+    val use: A => ZIO[R, E1, B]            = ???
     acquire.bracket(release)(use)
   }
 
   def infersEType2: ZIO[R, E, B] = {
-    val acquire: ZIO[R, E1, A]           = ???
-    val release: A => ZIO[R, Nothing, _] = ???
-    val use: A => ZIO[R, E, B]           = ???
+    val acquire: ZIO[R, E1, A]             = ???
+    val release: A => ZIO[R, Nothing, Any] = ???
+    val use: A => ZIO[R, E, B]             = ???
     acquire.bracket(release, use)
   }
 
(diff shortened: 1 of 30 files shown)


smarter:
> Replacing wildcard types with Any in pattern matches 

Wildcard types in pattern matches aren't really wildcard types, they're type patterns that can match anything, which imo is a perfectly fine thing to do.

adamgfraser:
@smarter I was wondering if those were different things and now I am more educated. Thank you!

jdegoes:
This should be discussed but it looks good to me, and use of `Any` / `Nothing` is certainly consistent with how we use it for effect elimination. 

We would continue using existentials for invariant types and wildcards for pattern matching.

core-tests/shared/src/test/scala/zio/BracketTypeInferrenceSpec.scala
```diff
@@ -10,37 +10,37 @@ object BracketTypeInferrenceSpec {
   class E1 extends E
 
   def infersEType1: ZIO[R, E, B] = {
-    val acquire: ZIO[R, E, A]            = ???
-    val release: A => ZIO[R, Nothing, _] = ???
-    val use: A => ZIO[R, E1, B]          = ???
+    val acquire: ZIO[R, E, A]              = ???
+    val release: A => ZIO[R, Nothing, Any] = ???
```
neko-kai: Doesn't that change the test itself?..

LGLO: As far as I remember, the point is that `acquire.bracket(release)(use)` doesn't require type annotations.

jdegoes:
If no further comments today, I propose we merge.

PR #1968, opened 2019-10-14
ZIO Test: Improve Documentation Regarding Using SBT Test Runner

adamgfraser:
A user on Discord today was having trouble running their tests with the SBT test runner because the `test_effects.md` page only mentions adding ZIO Test to `testFrameworks` and not adding a dependency on `zio-test-sbt`. Updates that page to clarify this.


ghostdogpr:
How about repeating the snippet of code that is in the other page? It's much more convenient (can copy paste) and it stands out (many people don't read long texts).

adamgfraser:
Good idea. Will do right now.

PR #1935, opened 2019-10-08
TestClock depends on Java Time API, adding dependency for ScalaJS project

LGLO:
I'm trying to fix #1934 
Please run `sbt` and then `test` to see linking error.
TestClock clearly depends on Java Time API that is not in vanilla ScalaJS.

PR #1969, opened 2019-10-14
Fix Test Command

adamgfraser:
When we added a command alias for `test` in #1935 I didn't realize that alias would apply globally. The result is that using `test` within SBT currently runs tests for all projects even if you are in a particular project, which is not great when a contributor is working on a certain project and only wants to run those tests. This PR removes that command alias and adds `scala-java-time` as a test dependency to the SBT test runner, which makes it available when `test` is called from the root directory so that all tests can be successfully run.

@LGLO Would you please take a look?


adamgfraser:
If we wanted we could add a separate `testAll` command alias that would run `testJVM` and `testJS`.

PR #1976, opened 2019-10-15
Fix scaladoc example on provideSome

ioleo:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -722,9 +722,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * {{{
    * val effect: ZIO[Console with Logging, Nothing, Unit] = ???
    *
-   * effect.provideSome[Console](console =>
+   * effect.provideSome[Console](env =>
    *   new Console with Logging {
-   *     val console = console
+   *     val console = env.console
    *     val logging = new Logging {
```
adamgfraser: Doesn't this need to be `Logging.Service`?

ioleo: True :+1:

PR #1985, opened 2019-10-16
ZIO Test: Refactor FailureDetails

### LINKED ISSUE #1332: ZIO Test: Refactor FailureDetails
### 1. Rename `FailureDetails` => `TestFailure`

Every `FailureDetails` represents the failure of a single test, so `TestFailure` is more descriptive and more consistent with other constructs inside the library.

### 2. Allow stack of predicate values

When a `FailureDetails` represents the failure of a predicate, currently there are two bits of information stored:

```scala
final case class Predicate(fragment: PredicateValue, whole: PredicateValue) extends FailureDetails
```

The `fragment` `PredicateValue`, and the `whole`. The first is the "inner most" failure, while the second is the "outer most" failure. For example, in a composite predicate like `isSome(isGreaterThan(4))`, if the value is a `Some`, but is 3, then there would be two failures:

 - `Some(3) did not satisfy isSome(isGreaterThan(4))`
 - `3 did not satisfy isGreaterThan(4)`

Instead of restricting to just 2, we should have a stack, so we can have 1 or more `PredicateValue`, potentially allowing a whole stack, each level "drilling down" into the failure.

This would also allow us to have 1 failure, rather than 2. In some cases, like `equals(32)`, there is just one failure, not two, so rather than de-duplicating that failure (as per work of @adamgfraser), we should just not create it.

The new representation would be:

```scala
final case class Predicate(values: ::[PredicateValue]) extends FailureDetails
```

where `::` is the "cons" constructor of `List`, which guarantees a head and a tail.

This change will necessitate modifying `Predicate.scala`, to "push" a new `PredicateValue` onto the stack in every combinator (for example, `isSome(pred)` is a combinator, because it is a function that takes a predicate and returns a predicate). 

I can assist with this if it is unclear.

adamgfraser:
Resolves #1332.

All inner failures when now be reported when there are multiple nested failures. For example, this assertion failure would now be reported like so:

```scala
assert(Right(Some(3)), isRight(isSome(isGreaterThan(4)))

Right(Some(3)) did not satisfy isRight(isSome(isGreaterThan(4)))
Some(3) did not satisfy isSome(isGreaterThan(4))
3 did not satisfy isGreaterThan(4)
```

Previously only the "inner most" and "outer most" failures would be reported.


adamgfraser:
Good idea. Done now.

jdegoes:
🎉

PR #1986, opened 2019-10-16
ZIO Test: Fix Type Inference Issue With ZSpec Syntax

### LINKED ISSUE #1978: ZIO Test: Type Inference Issue with ZSpecSyntax
The following snippet does not compile:

```scala
test("simple test") {
  assert(true, isTrue)
} @@ timeout(60.seconds)
```

While this snippet works fine:

```scala
TestAspect.timeout(60.seconds) {
  test("simple test") {
    assert(true, isTrue)
  }
}
```

The issue is with `ZSpecSyntax`:

```scala
implicit class ZSpecSyntax[R, E, L, S](spec: ZSpec[R, E, L, S]) {
  def @@[LowerR <: R, UpperR >: R, LowerE <: E, UpperE >: E, LowerS <: S, UpperS >: S](
    aspect: TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS]
  ): ZSpec[R, E, L, S] =
    aspect(spec)
}
```

In an example like the above, the test itself does not require any environment, so `R` is inferred as `Any`. But since `@@` requires `UpperR >: R` and `UpperR =:= Live[Clock]` the type constraints are not satisfied. Since the `ZSpec` doesn't require any environment it is perfectly legitimate to narrow it to `Live[Clock]` which is why the second example works fine but because the implicit class is resolving the `R` value too early this isn't working with the syntax.

adamgfraser:
Resolves #1978.

I think we can fix the type inference issue with `ZSpecSyntax` by moving `@@` from an implicit syntax class to a concrete method on `Spec` with implicit evidence that the spec is a `ZSpec`. This seems to work but the types are tricky so would appreciate someone else taking a look to make sure this is sound.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -34,8 +34,8 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
    * test("foo") { assert(42, equalTo(42)) } @@ ignore
    * }}}
    */
-  final def @@[R1 <: R, E1 >: E, T1 >: T, S](
-    aspect: TestAspect[Nothing, R1, Nothing, E1, S, S]
+  final def @@[R1 <: R, E1 >: E, T1 >: T, S, S1 <: S, S2 >: S](
+    aspect: TestAspect[Nothing, R1, Nothing, E1, S1, S2]
   )(implicit ev: T <:< Either[TestFailure[Nothing], TestSuccess[S]]): Spec[R1, E1, L, T1] = {
     val _ = ev
     aspect(self.asInstanceOf[ZSpec[R, E, L, S]]).asInstanceOf[Spec[R1, E1, L, T1]]
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -29,6 +29,19 @@ import Spec._
 final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E, L, T]]) { self =>
 
   /**
+   * Syntax for adding aspects.
+   * {{{
+   * test("foo") { assert(42, equalTo(42)) } @@ ignore
+   * }}}
+   */
+  final def @@[R1 <: R, E1 >: E, T1 >: T, S](
+    aspect: TestAspect[Nothing, R1, Nothing, E1, S, S]
+  )(implicit ev: T <:< Either[TestFailure[Nothing], TestSuccess[S]]): Spec[R1, E1, L, T1] = {
+    val _ = ev
+    aspect(self.asInstanceOf[ZSpec[R, E, L, S]]).asInstanceOf[Spec[R1, E1, L, T1]]
+  }
+
+  /**
    * Returns a new spec with the suite labels distinguished by `Left`, and the
    * test labels distinguished by `Right`.
    */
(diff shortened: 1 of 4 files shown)


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -28,6 +28,19 @@ import Spec._
...
+   * Syntax for adding aspects.
+   * {{{
+   * test("foo") { assert(42, equalTo(42)) } @@ ignore
+   * }}}
+   */
+  final def @@[R1 <: R, E1 >: E, T1 >: T, S, S1 <: S, S2 >: S](
+    aspect: TestAspect[Nothing, R1, Nothing, E1, S1, S2]
+  )(implicit ev: T <:< Either[TestFailure[Nothing], TestSuccess[S]]): Spec[R1, E1, L, T1] = {
+    val _ = ev
+    aspect(self.asInstanceOf[ZSpec[R, E, L, S]]).asInstanceOf[Spec[R1, E1, L, T1]]
```
jdegoes: I think you should `mapTest` with `ev` to avoid the unsafe cast.

adamgfraser: Good idea. I will submit a follow up PR.

unclebob418:
@adamgfraser I think we still have an issue here with being able to compose aspects

This does compile (but fails due to timeout):
```
testM("timeout makes tests fail after given duration") {
          assertM(ZIO.never *> ZIO.unit, equalTo(()))
        } @@ timeout(1.nanos)
```
This does compile (though would never finish without the suite timeout):
```
testM("timeout makes tests fail after given duration") {
          assertM(ZIO.never *> ZIO.unit, equalTo(()))
        } @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException])
```
This doesn't compile:
```
testM("timeout makes tests fail after given duration") {
          assertM(ZIO.never *> ZIO.unit, equalTo(()))
        } @@ timeout(1.nanos)
          @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException])
```
```
[error] C:\Users\rober\IdeaProjects\zio\test-tests\shared\src\test\scala\zio\test\TestAspectSpec.scala:88:11: inferred type arguments [zio.test.environment.Live[zio.clock.Clock],zio.test.TestTimeoutException,Either[zio.test.TestFailure[Nothing],zio.test.TestSuccess[Unit]],Unit,Unit,Unit] do not conform to method @@'s type parameter bounds [R1 <: zio.test.environment.Live[zio.clock.Clock],E1 >: Any,T1 >: Either[zio.test.TestFailure[Nothing],zio.test.TestSuccess[Unit]],S,S1 <: S,S2 >: S]
[error]           @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException]),
[error]           ^
[error] C:\Users\rober\IdeaProjects\zio\test-tests\shared\src\test\scala\zio\test\TestAspectSpec.scala:88:21: type mismatch;
[error]  found   : zio.test.TestAspect.PerTest[Nothing,Any,Nothing,zio.test.TestTimeoutException,Unit,Unit]
[error]  required: zio.test.TestAspect[Nothing,R1,Nothing,E1,S1,S2]
[error]           @@ failure(TestAspectSpecUtil.failsWithException[TestTimeoutException]),
[error]  
```

adamgfraser:
@unclebob418 Thanks for raising! I will have to spend some more time on this tomorrow.

unclebob418:
Thanks, won't have time to do much today but I think #1959 could possibly be merged anyway (sorry I forgot to ask for a review) - I can revisit these couple of tests in my next PR.

PR #1988, opened 2019-10-16
Fixed #1981 - corrected type signature of `zio.stream.ZSink#optional`

### LINKED ISSUE #1981: Error in type signature of `zio.stream.ZSink#optional`
The problem is best illustrated by the following snippet:
```scala
val stream = Stream.fromIterable(List(1))
val sink = Sink.identity[Int].optional

// Should type to IO[Nothing, Option[Int]] instead:
val maybeInt: IO[Unit, Option[Int]] = stream.run(sink)
```

As pointed out by @iravid, the there is an error in the type signature of
```scala
def ? : ZSink[R, E, A, A, Option[B]]
```
where `E` should be replaced with `Nothing`.

gerdreiss:


gerdreiss:
Thanks @iravid, looking forward to making more of those ;-)

PR #1990, opened 2019-10-16
Make Cause#foldLeft Stack Safe

### LINKED ISSUE #1984: Cause#failureOrCause can stackoverflow but should use
`Cause#failureOrCause` stack-overflows when it's made of many nested `Both`s.
`failures` would overflow as well in this case but case for `failureOrCause` is that is requires only first `Fail`.
Basically there is problem with `foldLeft`.

I stumped on it when I created 10000 failing effects and tried to `foreachPar` them and they completed before being interrupted, so each had time to fail before interruption.

adamgfraser:
Resolves #1984.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/CauseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/CauseSpec.scala
@@ -32,12 +32,10 @@ object CauseSpec
               assert(c.untraced.traces.headOption, isNone)
             }
           },
-          zio.test.test("`Cause.foldLeft is stack safe") {
-            val n = 100000
-            def causeN(n: Int, acc: Cause[String] = Cause.fail("fail")): Cause[String] =
-              if (n == 1) acc
-              else causeN(n - 1, Both(acc, Cause.fail("fail")))
-            assert(causeN(n).failures.length, equalTo(n))
+          zio.test.test("`Cause.failures is stack safe") {
+            val n     = 100000
+            val cause = List.fill(n)(Cause.fail("fail")).reduce(_ && _)
+            assert(cause.failures.length, equalTo(n))
           }
         ),
         suite("Then")(
--- a/core-tests/shared/src/test/scala/zio/CauseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/CauseSpec.scala
@@ -31,6 +31,13 @@ object CauseSpec
             check(causes) { c =>
               assert(c.untraced.traces.headOption, isNone)
             }
+          },
+          zio.test.test("`Cause.foldLeft is stack safe") {
+            val n = 100000
+            def causeN(n: Int, acc: Cause[String] = Cause.fail("fail")): Cause[String] =
+              if (n == 1) acc
+              else causeN(n - 1, Both(acc, Cause.fail("fail")))
+            assert(causeN(n).failures.length, equalTo(n))
           }
         ),
         suite("Then")(
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -16,6 +16,8 @@
 
 package zio
 
+import scala.annotation.tailrec
+
 sealed trait Cause[+E] extends Product with Serializable { self =>
   import Cause._
 
@@ -334,15 +336,22 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       }
       .reverse
 
-  private def foldLeft[Z](z: Z)(f: PartialFunction[(Z, Cause[E]), Z]): Z =
-    (f.lift(z -> self).getOrElse(z), self) match {
-      case (z, Then(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Both(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Traced(cause, _))  => cause.foldLeft(z)(f)
-      case (z, Meta(cause, _))    => cause.foldLeft(z)(f)
-
-      case (z, _) => z
-    }
+  private def foldLeft[Z](z: Z)(f: PartialFunction[(Z, Cause[E]), Z]): Z = {
+    @tailrec
+    def loop(z: Z, cause: Cause[E], stack: List[Cause[E]]): Z =
+      (f.lift(z -> cause).getOrElse(z), cause) match {
+        case (z, Then(left, right)) => loop(z, left, right :: stack)
+        case (z, Both(left, right)) => loop(z, left, right :: stack)
+        case (z, Traced(cause, _))  => loop(z, cause, stack)
+        case (z, Meta(cause, _))    => loop(z, cause, stack)
+        case (z, _) =>
+          stack match {
+            case hd :: tl => loop(z, hd, tl)
+            case Nil      => z
+          }
+      }
+    loop(z, self, List.empty)
+  }
 }
 
 object Cause extends Serializable {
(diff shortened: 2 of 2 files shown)


core-tests/shared/src/test/scala/zio/CauseSpec.scala
```diff
@@ -31,6 +31,13 @@ object CauseSpec
             check(causes) { c =>
               assert(c.untraced.traces.headOption, isNone)
             }
+          },
+          zio.test.test("`Cause.foldLeft is stack safe") {
```
neko-kai: Confused since the test doesn't use `Cause.foldLeft` 😕

adamgfraser: Ah, sorry. `failures` is implemented in terms of `foldLeft` so I was just using it that to test it. Previously this would stack overflow.

adamgfraser: The issue is that `foldLeft` is private so we can't test it directly.

jozic: `private[zio]`?

core-tests/shared/src/test/scala/zio/CauseSpec.scala
```diff
@@ -31,6 +31,13 @@ object CauseSpec
...
+            val n = 100000
+            def causeN(n: Int, acc: Cause[String] = Cause.fail("fail")): Cause[String] =
+              if (n == 1) acc
+              else causeN(n - 1, Both(acc, Cause.fail("fail")))
```
LGLO: No need to change but could be more compact, `List.fill(100000)(Cause.fail("fail")).reduce(_ && _)` or something like that. Recursion is OK.

core-tests/shared/src/test/scala/zio/CauseSpec.scala
adamgfraser: I don't think it makes sense to change the signature just to test it when we can easily test other methods that are implemented in terms of it. I can always rename the test to "Cause.failures is stack safe" if that would be clearer.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -334,15 +336,22 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
-      case (z, Then(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Both(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Traced(cause, _))  => cause.foldLeft(z)(f)
-      case (z, Meta(cause, _))    => cause.foldLeft(z)(f)
-
-      case (z, _) => z
-    }
+  private def foldLeft[Z](z: Z)(f: PartialFunction[(Z, Cause[E]), Z]): Z = {
+    @tailrec
+    def loop(z: Z, cause: Cause[E], stack: List[Cause[E]]): Z =
```
LGLO: Nice trick!

core-tests/shared/src/test/scala/zio/CauseSpec.scala
adamgfraser: I updated the test label to make this more clear.

core-tests/shared/src/test/scala/zio/CauseSpec.scala
adamgfraser: You're right. I was experimenting with different types of cause "trees". It is only left leaning trees like this that cause stack overflows and I wanted to avoid stack overflows myself as I created this giant tree. Your suggestion accomplishes both these things more concisely. Updated now.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: Thanks!

adamgfraser:
@LGLO Yes you're right. In addition to making the methods stack safe we could terminate the traversal early in some cases. I added a new combinator `find` and implemented the methods that could terminate early in terms of it.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -325,24 +309,50 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
-  private def foldLeft[Z](z: Z)(f: PartialFunction[(Z, Cause[E]), Z]): Z =
-    (f.lift(z -> self).getOrElse(z), self) match {
-      case (z, Then(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Both(left, right)) => right.foldLeft(left.foldLeft(z)(f))(f)
-      case (z, Traced(cause, _))  => cause.foldLeft(z)(f)
-      case (z, Meta(cause, _))    => cause.foldLeft(z)(f)
+  private def find[Z](f: PartialFunction[Cause[E], Z]): Option[Z] = {
+    @tailrec
+    def loop(cause: Cause[E], stack: List[Cause[E]]): Option[Z] =
+      f.lift(cause) match {
```
regiskuckaertz: I suggest `f.lift(cause) orElse ...` to avoid pattern matching

adamgfraser: The issue is if you do that the recursive call to `loop` is no longer in tail position.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -325,24 +309,50 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+        case Some(z) => Some(z)
+        case None =>
+          cause match {
```
regiskuckaertz: This is `cause.foldLeft(None) { case (_, cause) if f.isDefined(cause) => Some(f(cause)) }`, so find could be written as
```scala
private def find[Z](f: PartialFunction[Cause[E], Z]): Option[Z] = {
  f.lift(self) orElse foldLeft(None){ case (_, cause) if f.isDefined(cause) => Some(f(cause)) }
```

core/shared/src/main/scala/zio/Cause.scala
regiskuckaertz: Yes indeed 👍 I should have put the comment below with this one 🤦‍♂

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: I think this would require us to traverse the entire data structure if `f(lift(self)` is `None`, because then we call `foldLeft` which always traverses the entire data structure. The `PartialFunction` in `foldLeft` controls whether a value from a particular cause will be added to the accumulator but can't terminate the traversal early.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -325,24 +309,50 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+      }
+    loop(self, Nil)
+  }
 
-      case (z, _) => z
-    }
+  private def foldLeft[Z](z: Z)(f: PartialFunction[(Z, Cause[E]), Z]): Z = {
+    @tailrec
+    def loop(z: Z, cause: Cause[E], stack: List[Cause[E]]): Z =
+      (f.lift(z -> cause).getOrElse(z), cause) match {
```
regiskuckaertz: You may want to memoize `f.lift` as it create a new object on each invocation.

regiskuckaertz: In fact this is just `f.applyOrElse(z -> cause, _ => z)` in hiding, without the `Option` allocation

adamgfraser: Good point. Will update.

core/shared/src/main/scala/zio/Cause.scala
regiskuckaertz: Ha yeah right, you would need lazyness to get around that. Shame, that is elegant:
```scala
private def find[Z](f: PartialFunction[Cause[E], Z]): Option[Z] = {
  f.lift(self) orElse foldLeft(None){ case (z, cause) if f.isDefinedAt(cause) => z orElse Some(f(cause)) }
```
Out of curiosity, why a `PartialFunction` in `foldLeft`?

PR #1999, opened 2019-10-18
Add managed aliases

ioleo:
I found myself using `Managed[Nothing, A]` very often.

I propose we add all possible managed aliases (same naming schema as for IO).

~A shorter alias `Managed` => `Man` is also included, please let me know if you think its a good idea.
I think having both longer aliases (`Managed`-family) and shorter (`Man`-family) is a good idea - let the developer (user) choose wheater he needs more explicity or more convinience.~ removed after feedback

### DIFF
--- a/core/shared/src/main/scala/zio/package.scala
+++ b/core/shared/src/main/scala/zio/package.scala
@@ -25,13 +25,18 @@ package object zio extends ZEnvDefinition with EitherCompat {
   type IO[+E, +A]   = ZIO[Any, E, A]
   type UIO[+A]      = ZIO[Any, Nothing, A]
   type Task[+A]     = ZIO[Any, Throwable, A]
-  @deprecated("use RIO", "1.0.0")
-  type TaskR[-R, +A] = RIO[R, A]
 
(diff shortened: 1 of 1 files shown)


iravid:
I like the `*Managed` ones, but find `Man` a pretty weird abbreviation

iravid:
Also, if we add these aliases, should we add companion objects like with `ZIO`? I think it's not necessary, but bringing it up for discussion.

adamgfraser:
Should `TManaged` be `TaskManaged`? I realize it is longer but when I first read it I thought of STM functionality like `TRef` and `Ref`.

ioleo:
@adamgfraser valid point, we probably should have a discussion on the naming schemes we use
do we reserve `T` prefix for transactional (STM-related) stuff?

adamgfraser:
@ioleo Also `TPromise`, `TQueue`, `TSemaphore`, and work in progress on `TMap`. I would suggest we use `Task` prefix and reserve `T` prefix for STM functionality.

ioleo:
@adamgfraser @iravid I think we can merge this as-is.

PR #2000, opened 2019-10-18
Mock improvements

ioleo:


ioleo:
Will provide updates to documentation as a seperate PR.

ioleo:
@adamgfraser Ready for review.

test/shared/src/main/scala/zio/test/mock/ArgumentExpactation.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package zio.test.mock
+
+import zio.test.Assertion
+
+/**
+ * An `ArgumentExpectation[M, I, A]` represents an expectation on input `I` arguments
+ * for capability of module `M` that returns an effect that may produce a single `A`.
+ */
+final case class ArgumentExpectation[M, I, A](method: Method[M, I, A], assertion: Assertion[I]) {
```
adamgfraser: The file name is misspelled. Should be `ArgumentExpectation.scala` not `ArgumentExpactation.scala`.

test/shared/src/main/scala/zio/test/mock/Expectation.scala
```diff
@@ -16,9 +16,215 @@
...
+import zio.test.mock.Expectation.{ AnyCall, Call, Empty, FlatMap, Next, State }
+import zio.test.mock.ReturnExpectation.{ Fail, Succeed }
+import zio.test.mock.MockException.UnmetExpectationsException
+
+/**
+ * A `Expectation[-M, +E, +A]` is an immutable data structure that represents
+ * expectations on module `M` capabilities.
+ *
+ * This structure is a monad, becouse we need the sequential composability and
+ * in Scala we get the the convinient for-comprehension syntax for free.
```
adamgfraser: "the" is repeated". "becouse" should be "because". "convinient" to "convenient".

test/shared/src/main/scala/zio/test/mock/Expectation.scala
```diff
@@ -16,9 +16,215 @@
...
+
+  /**
+   * Effectfully maps the input arguments `I` to a return expecation to succeed with `A`.
+   */
+  final def valueM[I, A](f: I => IO[Nothing, A]): Succeed[I, A] = Succeed(f)
+
+  /**
+   * Implicitly converts Expectation to ZManaged mock environment.
+   */
+  implicit final def toManagedEnv[M, E, A](
```
adamgfraser: Should we put this in some separate object that can be optionally imported like `ZIOSyntax`? I understand why you want to have it for a more convenient API but using implicits, especially implicit conversions as opposed to implicit syntax and implicit conversions that are automatically imported when you import a top level object, seems like an anti-pattern for ZIO.

Alternatively. What if we just have syntax that adds a `provideExpectation` method on `ZIO` that would take an expectation, convert it to a mock environment, and then provide that to the effect. I think that would have the same benefits for usability but have less of the "magic" of implicit conversions.

test/shared/src/main/scala/zio/test/mock/Method.scala
```diff
@@ -16,7 +16,34 @@
...
+trait Method[M, I, A] { self =>
+
+  import Method._
+
+  /**
+   * Provides the `Assertion` on method arguments `I` to produce `ArgumentExpectation`.
+   *
+   * Available only for methods that do take arguments.
+   */
+  @silent("parameter value ev in method apply is never used")
```
adamgfraser: Instead of using `@silent` you can also do `val _ = ev` in the body of the method. I'm not sure which is better here.

test/shared/src/main/scala/zio/test/mock/Method.scala
```diff
@@ -31,3 +58,15 @@ trait Method[+A, -B] {
     }
   }
 }
+
+object Method {
+
+  /**
+   * Evidence type `A` is not equal to type `B`.
+   */
+  trait =!=[A, B]
```
adamgfraser: Maybe add some `implicitNotFound` or `implicitAmbiguous` annotations here? Right now if this doesn't exist I think the user is going to get a somewhat confusing message about an implicit he or she probably doesn't even realize exists.

test/shared/src/main/scala/zio/test/mock/Expectation.scala
ioleo: I've put it in `Expectation` on purpose. I'd like that implicit to be always available without the user haveing to import it. Basically, it allows to use the for-comprehension builder directly in `provideManaged`.

I don't think we should introduce a seperate `provideExpectation` for this.
The `Expectation` structure is supposed to be only a supporting builder, it's always going to be converted to managed environment. While I understand that haveing implicit magic is not always the best experience for the user, in this case I believe its beneficial:
* the user always wants to perform this conversion
* the user does not have to import anything, this conversion will always be in scope becouse it's placed in`Expectation`'s companion object
* the implicit conversion is total -> for every expectation it will yield a managed env
* there is no scenario where the user might want to perform this conversion on his own or replace it with a diffrent conversion

To summarize: this is to reduce the boilerplate and should be transparent to the user.

test/shared/src/main/scala/zio/test/mock/Method.scala
ioleo: Since the silencer was already in the dependencies I leaned towards using it. But I can change that if you guys prefer other solution.

test/shared/src/main/scala/zio/test/mock/Method.scala
ioleo: Good idea! Definitely will do!

ioleo: @adamgfraser turns out `implicitAmbiguous` is only available since scala 2.12, so I cannot use it :(

The `implicitNotFound` does not apply here, as the usage is in `Method` trait and implicit for all `A, B` where `A != B` is also defined in `Method` companion object.

test/shared/src/main/scala/zio/test/mock/Method.scala
adamgfraser: Fine with me.

test/shared/src/main/scala/zio/test/mock/Method.scala
adamgfraser: Good point about `implicitAmbiguous`. I was playing with `implicitNotFound` and in 2.11 the `implicitNotFound` warning will be triggered if the implicit can't be found because it is ambiguous. So you could have a 2.11 and a 2.12+ version of the file where one uses `implicitAmbiguous` and the other uses `implicitNotFound`. But I know that is a bit of a hassle, so up to you whether you think it is worth it.

test/shared/src/main/scala/zio/test/mock/Expectation.scala
adamgfraser: Those are great points. I'll defer to you on this one. I was just thinking if you go back to the examples you currently have:

```scala
testM("expect call returning output") {
  val app     = clock.nanoTime
  val mockEnv = MockClock.nanoTime returns value(1000L)
  val result  = app.provideManaged(mockEnv)
  assertM(result, equalTo(1000L))
},
```

Where the implicit conversion is necessary to make the call to `provideManaged` work. With a `provideMock` syntax method you could instead write this as:

```scala
testM("expect call returning output") {
  val app     = clock.nanoTime
  val mockEnv = MockClock.nanoTime returns value(1000L)
  val result  = app.provideMock(mockEnv)
  assertM(result, equalTo(1000L))
}
```

At least to me that seems as convenient and maybe even has some value in clarifying intent that you are providing a mock implementation which behaves significantly differently than other implementations.

But happy to do what you think makes sense. I don't know this as well as you so maybe there are other use cases where this would not work as well.

test/shared/src/main/scala/zio/test/mock/Method.scala
ioleo: Alright, I think I got this. Thanks, just learned something new :)

test/shared/src/main/scala/zio/test/mock/Expectation.scala
ioleo: Let me sleep with this one, I'll think about possible use cases and if provideMock solves them.

core/shared/src/main/scala-2.11/zio/=!=.scala
```diff
@@ -0,0 +1,18 @@
+package zio
+
+import scala.annotation.implicitNotFound
+
+/**
+ * Evidence type `A` is not equal to type `B`.
+ *
+ * Based on https://stackoverflow.com/a/6944070.
```
ioleo: @adamgfraser @jdegoes Are we allowed to include snippets like that?
A bit more expanded version of this is part of Miles shapeless which is licenced under Apache 2.0.
https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/package.scala#L41

Though I did find it on SO first. Question is, do we need to add some header, is this comment enough for attribution? Or perhaps no attribution is needed as the technique to introduce implicit ambiguity is well estabilished in Scala?

adamgfraser: If we give attribution I would suggest we do it to Shapeless as opposed to Stack Overflow since Shapeless is available under an Apache license which ZIO also uses whereas Stack Overflow content is available under a Creative Commons license that is not compatible with the Apache license. Miles Sabin is the author of the Stack Overflow post and also of Shapeless, which appears to be the original source of the work in question, so I think this is also the most accurate attribution.

test/shared/src/main/scala/zio/test/mock/Expectation.scala
ioleo: I think we should leave it as is. To introduce `provideMock` we'd have to modify `ZIO` which is in `core`, while `Expectation` is in `test`. We could ofcourse extend `ZIO` via implicit class, but... that would trade one implicit for another, so there is no point.

The other solution would be to move Expectation from `test` to `core`, but I feel thats not a good idea too.

Regarding the use cases I have some ideas what else could be improved, but nothing concrete yet, so let's merge it as is, and if I come up with some better design I'll put that into next PR.

core/shared/src/main/scala-2.11/zio/=!=.scala
ioleo: Added attribution and copyright in the header, though I'm not sure its the right way. Correct me if I'm wrong please :)

ioleo:
@adamgfraser I think it's ready to be merged.

test/shared/src/main/scala/zio/test/mock/Expectation.scala
adamgfraser: Okay. I think if we went with `provideMock` we would definitely want to do it as a syntax class the user could import, something like `zio.test.mock.syntax._` analogous to `zio.syntax._`. That would avoid changing anything in `core`.

You're right we would still have an implicit but I think an implicit syntax class is better than an implicit conversion from a ZIO perspective. It just gives the user an additional method they can call on something versus "magically" changing types. I think that is reflected in how there is some optional implicit syntax in ZIO today, for example in `ZIOSyntax` but to my knowledge there is no use of implicit conversions.

I'm okay with keeping it as is but just wanted to make sure I was clear about what I was proposing.

PR #2006, opened 2019-10-20
Add `ZSink.splitOn`

### LINKED ISSUE #1965: ZSink.splitDelimiter
It would be nice to have a `splitDelimiter` operation on ZSink, to be able to do something close to Akka Streams `Framing.delimiter`, i.e, change the way elements are separated in a ZStream by merging and re-splitting them.

Akka Streams `Framing.delimiter` usage example : 

```scala
source.via(Framing.delimiter(ByteString("\n"), maximumFrameLength = 256)
```

It could be something like : 

 ```scala
stream.transduce(ZSink.splitDelimiter("\n", maximumFrameLength = 256))
```
Note : `maximumFrameLength` should have a default (unbounded ?) value.

Thanks

### PR DESCRIPTION
**Pull request by regiskuckaertz**
I can add a maximum frame length if it makes sense, let me know what you think.

Closes #1965

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1678,15 +1678,18 @@ object ZSink extends ZSinkPlatformSpecific {
   final val splitLinesChunk: ZSink[Any, Nothing, Chunk[String], Chunk[String], Chunk[String]] =
     splitLines.contramap[Chunk[String]](_.mkString).mapRemainder(Chunk.single)
 
+  /**
+   * Splits strings on a delimiter.
+   */
   final def splitOn(delimiter: String): ZSink[Any, Nothing, String, String, Chunk[String]] =
     new SinkPure[Nothing, String, String, Chunk[String]] {
       type State = SplitOnState
       case class SplitOnState(
         // Index into the delimiter
-        i: Int,
+        delimiterPointer: Int,
         // Index into the current frame
-        j: Int,
-        // Signals when extraction is
+        framePointer: Int,
+        // Signals when extraction of the delimiter is ongoing
         cont: Boolean,
         // Accumulated strings from previous pulls
         leftover: String,
@@ -1711,8 +1714,8 @@ object ZSink extends ZSinkPlatformSpecific {
         val m     = frame.length
 
         var start = 0
-        var i     = s.i
-        var j     = s.j
+        var i     = s.delimiterPointer
+        var j     = s.framePointer
 
         val buf = mutable.ArrayBuffer[String]()
 
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1724,15 +1724,14 @@ object ZSink extends ZSinkPlatformSpecific {
 
           if (i == l) {
             // We've found a new frame, store it (_without_ the delimiter),
-            // reset the delimiter pointer and advance the frame pointer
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: streams/shared/src/main/scala/zio/stream/ZSink.scala [hunk 1]
```diff
@@ -1678,6 +1678,80 @@ object ZSink extends ZSinkPlatformSpecific {
...
 
+  final def splitOn(delimiter: String): ZSink[Any, Nothing, String, String, Chunk[String]] =
+    new SinkPure[Nothing, String, String, Chunk[String]] {
+      type State = SplitOnState
+      case class SplitOnState(
+        // Index into the delimiter
+        i: Int,
+        // Index into the current frame
+        j: Int,
+        // Signals when extraction is
```
vasilmkd: // Missing sentence end.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1678,6 +1678,80 @@ object ZSink extends ZSinkPlatformSpecific {
...
     splitLines.contramap[Chunk[String]](_.mkString).mapRemainder(Chunk.single)
 
+  final def splitOn(delimiter: String): ZSink[Any, Nothing, String, String, Chunk[String]] =
+    new SinkPure[Nothing, String, String, Chunk[String]] {
+      type State = SplitOnState
+      case class SplitOnState(
+        // Index into the delimiter
+        i: Int,
+        // Index into the current frame
+        j: Int,
```
vasilmkd: Nitpicking: I know that the variables are nicely commented but I feel like we could gain readability by giving them slightly more descriptive names, which I think is ok because in this case we are dealing with a specific use case. Feel free to ignore this comment.

P.S. I also wanted to try out multi line comments on GitHub! 😆

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1678,6 +1678,80 @@ object ZSink extends ZSinkPlatformSpecific {
   final val splitLinesChunk: ZSink[Any, Nothing, Chunk[String], Chunk[String], Chunk[String]] =
```
vasilmkd: Do you mind adding a quick test for this combinator?

vasilmkd (commented):
Great addition.

Just an idea. Would a `final def splitWith(p: A => Boolean): ZSink[Any, Nothing, A, A, Chunk[A]` combinator make sense?

streams/shared/src/main/scala/zio/stream/ZSink.scala
regiskuckaertz: Good point 👍

streams/shared/src/main/scala/zio/stream/ZSink.scala
regiskuckaertz: ditto

streams/shared/src/main/scala/zio/stream/ZSink.scala
regiskuckaertz: I did, but I must have misunderstood because it is failing in some cases 🤔

regiskuckaertz: Solved

vasilmkd (approved):
:+1:

regiskuckaertz:
Yes `splitWith` makes sense imo 🤔

iravid:
Will review tonight!

iravid (approved):
Love it! Awesome work @regiskuckaertz!

PR #2007, opened 2019-10-20
Streams tests cleanup

vasilmkd:
@regiskuckaertz  @iravid

There were build warnings, one unused import in `SinkSpec` and some type inference warnings in `StreamSpec` which I reported to @adamgfraser.


regiskuckaertz:
Thank you 🙏

PR #2008, opened 2019-10-20
Expand Stream's Structure to include FlatMap and Effect terms

iravid:
As a follow up to the introduction of `Structure` and `Concat`, this commit now expands the terms on `Structure` to include `FlatMap` and `Effect`. The interpreter for `Structure` has also been refactored to accomodate these.

This representation now supports recursion patterns like the following:
```
  def recursive: ZStream[random.Random, Nothing, Int] = 
    ZStream.fromEffect(random.nextInt).flatMap { i =>
      if (i % 2 == 0) Stream.succeed(i) ++ go
      else Stream.succeed(i)
    }
```

This is achieved by differentiating between structures with potentially multiple elements (like `Concat` and `Iterator`) and scopes, and scope-less single effects (`Effect`). By differentiating between them, we can avoid allocating a stack frame when evaluating `FlatMap(Effect(z), k)`. This is because once `z` is evaluated, we need to evaluate `k` against it only once.

With this optimization, the above `recursive` function uses constant stack space.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/Structure.scala
+++ b/streams/shared/src/main/scala/zio/stream/Structure.scala
@@ -1,0 +1,148 @@
+package zio.stream
+
+import zio._
+import ZStream.Pull
+
+private[stream] sealed abstract class Structure[-R, +E, +A] { self =>
+  import Structure._
+
+  def ++[R1 <: R, E1 >: E, A1 >: A](other: => Structure[R1, E1, A1]): Structure[R1, E1, A1] =
+    Concat(self, () => other)
+
+  def map[B](f: A => B): Structure[R, E, B] =
+    self match {
+      case Iterator(iter)          => Structure.Iterator(iter.map(_.map(f)))
+      case Concat(hd, tl)          => Structure.Concat(hd.map(f), () => tl().map(f))
+      case s: FlatMap[R, E, a0, A] => Structure.FlatMap(s.fa, (el: a0) => s.f(el).map(f))
+      case Effect(zio)             => Structure.Effect(zio.map(f))
+    }
+
+  def mapM[R1 <: R, E1 >: E, B](f: A => ZIO[R1, E1, B]): Structure[R1, E1, B] =
+    self match {
+      case Concat(hd, tl)          => Concat(hd.mapM(f), () => tl().mapM(f))
+      case Iterator(iter)          => Iterator(iter.map(_.flatMap(f(_).mapError(Some(_)))))
+      case s: FlatMap[R, E, a0, A] => Structure.FlatMap(s.fa, (el: a0) => s.f(el).mapM(f))
+      case Effect(zio)             => Structure.Effect(zio.flatMap(f))
+    }
+
+  def flatMap[R1 <: R, E1 >: E, B](f: A => Structure[R1, E1, B]): Structure[R1, E1, B] =
+    FlatMap[R1, E1, A, B](self, f)
+
+  def process: ZManaged[R, E, Pull[R, E, A]] =
+    self match {
+      case Iterator(iter) => iter
+      case _ =>
+        for {
+          refs <- (Ref.make[Structure[R, E, _]](self) <*>
+                   Ref.make[Pull[R, E, _]](Pull.end) <*>
+                   Ref.make[Exit[_, _] => URIO[R, _]](_ => UIO.unit) <*>
+                   Ref.make[List[(Pull[R, E, _], Exit[_, _] => URIO[R, _], Structure[R, E, _])]](Nil)).toManaged_
+          (currStruct <*> currPull <*> currFin <*> stack) = refs
+          pull <- ZManaged.reserve(
+                   Reservation(
+                     open(self, currFin, currPull).as(pull(currStruct, currFin, currPull, stack)),
+                     e =>
+                       for {
+                         fs    <- currFin.get.zipWith(stack.get.map(_.map(_._2)))(_ :: _)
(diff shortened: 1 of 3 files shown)


iravid:
This scheme still isn't perfect, btw, because "innocent" combinators like `take(1)` can destroy the optimization. For example, `ZStream.fromEffect(random.nextInt).take(1)` will no longer use constant stack space.

The reason is that most of the combinators now "compile" the `Structure` to a `ZManaged[R, E, Pull[R, E, A]]`, which discards all of the structure. This is why I changed `map` and `mapM` to delegate directly to `Structure` and preserve it.

There might be some follow-ups needed in this area.

vasilmkd:
I feel guilty that the only thing that is broken is `buffer`. I'm ok with dropping that code and pursuing the `Queue` improvements down the line.

Awesome PR. Big :+1:.

iravid:
@vasilmkd Those tests might have surfaced a bug, so I'll look into them, no worries :-)

streams/shared/src/main/scala/zio/stream/Structure.scala
```diff
@@ -0,0 +1,148 @@
...
+  def process: ZManaged[R, E, Pull[R, E, A]] =
+    self match {
+      case Iterator(iter) => iter
+      case _ =>
+        for {
+          refs <- (Ref.make[Structure[R, E, _]](self) <*>
+                   Ref.make[Pull[R, E, _]](Pull.end) <*>
+                   Ref.make[Exit[_, _] => URIO[R, _]](_ => UIO.unit) <*>
+                   Ref.make[List[(Pull[R, E, _], Exit[_, _] => URIO[R, _], Structure[R, E, _])]](Nil)).toManaged_
+          (currStruct <*> currPull <*> currFin <*> stack) = refs
```
regiskuckaertz: You can deconstruct using `<*>`? 🤯

streams/shared/src/main/scala/zio/stream/Structure.scala
```diff
@@ -0,0 +1,148 @@
...
+        _           <- currFin.set(reservation.release)
+        _           <- restore(reservation.acquire.tap(currPull.set))
+      } yield ()
+    }
+
+  private[this] def pull(
+    currStruct: Ref[Structure[R, E, Any]],
+    currFin: Ref[Exit[_, _] => URIO[R, _]],
+    currPull: Ref[Pull[R, E, Any]],
+    stack: Ref[List[(Pull[R, E, _], Exit[_, _] => URIO[R, _], Structure[R, E, _])]]
```
regiskuckaertz: I feel like some type aliases for finalizers and stack frames would improve readability here.

streams/shared/src/main/scala/zio/stream/Structure.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  def process: ZManaged[R, E, Pull[R, E, A]] =
+    self match {
+      case Iterator(iter) => iter
+      case _ =>
+        for {
+          refs <- (Ref.make[Structure[R, E, _]](self) <*>
+                   Ref.make[Pull[R, E, _]](Pull.end) <*>
+                   Ref.make[Exit[_, _] => URIO[R, _]](_ => UIO.unit) <*>
+                   Ref.make[List[(Pull[R, E, _], Exit[_, _] => URIO[R, _], Structure[R, E, _])]](Nil)).toManaged_
```
regiskuckaertz: Just as a side note, we now have so many of these `Ref.make(...).toManaged_` that we should add a `Ref.makeManaged`.

streams/shared/src/main/scala/zio/stream/Structure.scala
```diff
@@ -0,0 +1,148 @@
...
+              (for {
+                st  <- stack.get
+                p   <- currPull.get
+                fin <- currFin.get
+                s   <- currStruct.get
+                _   <- stack.set((p, fin, s) :: st)
+                _   <- currStruct.set(next)
+              } yield ()) *> restore(open(next, currFin, currPull).mapError(Some(_)))
+            } *> pull(currStruct, currFin, currPull, stack)
+        }
```
regiskuckaertz: This frame is going to be on the hot path? I wonder if using a `tag` akin to `zio.tag` would be a good optimisation here.

iravid:
I'll just note that I'm not yet completely sold on this approach for the following reasons:
- Allowing infinite recursion through `ZStream.fromEffect(zio).flatMap(...)` is nice, but not essential for writing streaming programs. This is evidenced by the fact that I was able to rewrite `ZStream.repeatWith` to not use it.
- The interpreter is becoming pretty nasty, and that's not the end of it. We would still need to support interpretation of a `catchAll` structure, which means adding error handles and stack unwinding to the interpreter. Yikes!
- I still don't know how to properly support combinators that want to view the structure as an `ZManaged[Pull]` without destroying the structure. I have a few ideas but they'd make the interpreter even nastier!

In short, the benefits are apparent, but I'm not sure they outweigh the costs of maintaining the interpreter over time.

vasilmkd:
Are we keeping the current interpreter that is currently in place?

iravid:
So I think infinite, recursive concatenation is a much more common usecase that is harder to re-implement without recursion. That part has much more justification, in my opinion.

iravid:
I picked @jdegoes' brain on this and the conclusion is that we should try and move the initial-ness of the representation into `ZManaged`. This way we can do the optimizations there and keep `ZStream` simpler.

Closing this out for now - thanks for reviewing @vasilmkd @regiskuckaertz!

PR #2009, opened 2019-10-21
WIP migrate remaining environment specs

unclebob418:
#1898 migrating ClockSpec, SchedulerSpec, RandomSpec


unclebob418:
@adamgfraser can you please review? SchedulerSpec passes but ClockSpec has some flakiness and I'm not sure why. Doco mentions it's easy to accidentally use different clocks in the Test environment but without creating new clocks I'm not sure how. I'll get started on RandomSpec soon.

unclebob418:
@adamgfraser, providing new instances of TestClock seems to have resolved the issues with the flaky tests, but I'm not sure why. I initially changed this thinking `nonFlaky` might be running the tests in parallel which could get things confused, but looking at the definition of `nonFlaky` seems to indicate that the tests are run sequentially. Can you please explain why this is?

I also noticed there are no tests for a few of the TestAspects (`nonFlaky` being one). I'll try and address this in a separate PR.

The last test is still timing out. I don't really understand how this is testing for race conditions so I'm not really sure how to investigate this one further.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+import zio.test.TestAspect._
+import zio.test._
+import zio.test.environment.TestClock._
+
+object ClockSpec
+    extends ZIOBaseSpec(
+      suite("ClockSpec")(
+        testM("sleep does not require passage of clock time") {
+          for {
+            latch <- Promise.make[Nothing, Unit]
```
adamgfraser: I would use `TestClock.setTime(0.hours)` here. I think the reason you were having issues with some of these tests is that the testing framework provides a separate copy of the `TestEnvironment` to each test, but that environment is shared between successive invocations of the same test via `nonFlaky`. So like in this test on the first run the clock time will start at 0 hours and then you will sleep until the time is 10 hours. But on the second run the clock time will start at 11 hours so now an effect that sleeps for 10 hours will immediately run.

I think if you generally use the `TestClock` in the `TestEnvironment` and set the time to zero as your first action in any test where you are using `nonFlaky` that should get you to a pretty good place.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+            time <- currentDateTime
+          } yield assert(time.toInstant.toEpochMilli, equalTo(1.millis.toMillis))
+        },
+        testM("setTime does not produce sleeps") {
+          for {
+            _      <- setTime(1.millis)
+            sleeps <- sleeps
+          } yield assert(sleeps, isEmpty)
+        },
+        testM("setTimeZone correctly sets timeZone") {
```
adamgfraser: I don't think this test is working correctly right now because setting the time is an effect so it needs to be chained in a `flatMap` or a for comprehension to the assertion. The test is just passing right now because the default time zone is "UTC".

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+            _ <- TestClock.adjust(60.minutes)
+            b <- q.take.as(true)
+            c <- q.poll.map(_.isEmpty)
+            _ <- TestClock.adjust(60.minutes)
+            d <- q.take.as(true)
+            e <- q.poll.map(_.isEmpty)
+          } yield a && b && c && d && e
+          assertM(example, isTrue)
+        },
+        testM("fiber time is not subject to race conditions") { //times out
```
adamgfraser: The issue is that you are creating and adjusting your own `TestClock` but `sleep` is using the `TestClock` in the environment so is never waking up. Use the pattern above of accessing the `TestClock` in the environment and setting it to zero and I think this should work fine.

test-tests/shared/src/test/scala/zio/test/environment/SchedulerSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+            canceled  <- ZIO.effectTotal(cancel())
+          } yield assert(canceled, isFalse)
+        ),
+        testM("scheduled tasks get executed before shutdown")(
+          for {
+            scheduler <- scheduler
+            promise   <- Promise.make[Nothing, Unit]
+            _         <- ZIO.effectTotal(runTask(scheduler, promise, 10.seconds))
+            _         <- ZIO.effectTotal(scheduler.shutdown())
+            _         <- promise.await //how does the test clock advance without an adjust?
```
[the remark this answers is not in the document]
unclebob418: that makes sense, thanks

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: >... the testing framework provides a separate copy of the TestEnvironment to each test, but that environment is shared between successive invocations of the same test via nonFlaky.

This is handy to know. I'll add a note to the aspect doco. seems like a good case for `@@ after(setTime(0.hours))`

> So like in this test on the first run the clock time will start at 0 hours and then you will sleep until the time is 10 hours. But on the second run the clock time will start at 11 hours so now an effect that sleeps for 10 hours will immediately run.

I don't think this example is quite right but I get the point :) (second run the clock starts at `11.hours`, then `sleep(10.hours)` sleeps to `21.hours`).

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: i still have my imperative hat on :(

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: @adamgfraser  I'm looking at "sleep correctly handles multiple sleeps" and my confusion is increasing... my (now improved) understanding of `nonFlaky` this is it's effectively checking the assertion over and over in the same environment, with the clock (in this particular case) increasing in 2 hr increments. Tasks are scheduled to start on odd hours, time is always even. The latches prevent the creations of race conditions caused by swapping fibers that the documentation mentions. The latches and ref are new for each iteration. Nothing seems dependent on the absolute time (sleep & adjust are both relative to the current time) so resetting the clock should have no effect here, yet this is still flaky without ` @@ after(setTime(0.hours))`

Is there some way of debugging in ZIO? I'm really missing being able to pause at arbitrary locations and check state.
```
for {                                    // time is 0 hours
  latch1 <- Promise.make[Nothing, Unit]  // new latch
  latch2 <- Promise.make[Nothing, Unit]  // new latch
  ref    <- Ref.make("")                 // new ref
                                         // set up fiber 1, delayed till 3.hours
  _      <- (sleep(3.hours) *> ref.update(_ + "World!") *> latch2.succeed(())).fork
                                         // set up fiber 2, delayed till 1.hour
  _      <- (sleep(1.hours) *> ref.update(_ + "Hello, ") *> latch1.succeed(())).fork 
  _      <- adjust(2.hours)              // adjust time to 2.hours, starts fiber 2
  _      <- latch1.await                 // wait for fiber 2 to finish, ref should now be "Hello, "
  _      <- adjust(2.hours)              // adjust time to 4.hours, starts fiber 1
  _      <- latch2.await                 // wait for fiber 1 to finish, ref should now be "Hello, World!"
  result <- ref.get
} yield assert(result, equalTo("Hello, World!"))
for {                                    // time is still 4.hours
  latch1 <- Promise.make[Nothing, Unit]  // new latch
  latch2 <- Promise.make[Nothing, Unit]  // new latch
  ref    <- Ref.make("")                 // new ref
                                         // set up fiber 3, delayed till 7.hours
  _      <- (sleep(3.hours) *> ref.update(_ + "World!") *> latch2.succeed(())).fork
                                         // set up fiber 4, delayed till 5.hours
  _      <- (sleep(1.hours) *> ref.update(_ + "Hello, ") *> latch1.succeed(())).fork
  _      <- adjust(2.hours)              // adjust time to 6.hours, starts fiber 4
  _      <- latch1.await                 // wait for fiber 4 to finish, ref should now be "Hello, "
  _      <- adjust(2.hours)              // adjust time to 8.hours, starts fiber 3
  _      <- latch2.await                 // wait for fiber 3 to finish, ref should now be "Hello, World!"
  result <- ref.get
} yield assert(result, equalTo("Hello, World!"))
//rinse, repeat 98 more times
```

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: @adamgfraser That makes sense. Fixed the timeout issue but created another :) fibertime seems to be 100x what it should be "200 did not satisfy equalTo(2)" (in millis)

unclebob418:
getting there. Remaining issues:

1. "sleep correctly handles multiple sleeps" works with `@@ after(setTime(0.hours))`, but I don't think it should be required for this test as it doesn't appear to rely on the absolute time.

1.  "sleep does sleep instantly when sleep duration less than set time" - this worked ok when I was creating a new test clock for each test. Now it's failing, even with `@@ after(setTime(0.hours))` (which I do think is required for this test)

1. "fiber time is not subject to race conditions" This looks like a unit conversion issue to me as the result is always 100x larger than the largest sleep in the `zipPar`. Fails the assertion with "200 did not satisfy equalTo(2)" - when I used the closest I could get to the original implementation (`yield assert(result.toNanos, equalTo(2000000L))`) it also failed.

adamgfraser:
So I think the key to understanding these issues is the distinction between "fiber time" and "clock time" in the `TestClock` and how those interact with `nonFlaky`.

The `TestClock` is based on a concept of "clock time" and "fiber time". Clock time you could think of as the equivalent of the wall clock time that you can control. Fiber time always starts at zero, is incremented by the total duration of `sleep` calls, is inherited by child fibers when created, and is inherited by parent fibers when the child fiber is joined if the child fiber time is greater. So you could think of fiber time as the total duration a fiber and its ancestors have spent sleeping to reach the current state, properly accounting for forks and joins. `Sleep` calls delay effects until the clock time is greater than the fiber time plus the sleep duration.

Unlike the previous version of `ClockSpec` where we were unsafely running each test 100 times and so always getting a new fiber, repeated runs can now run on the same fiber so if we join a fiber after sleeping it the fiber time will be increased. So we need to be careful to keep the clock time in sync with the fiber time at the beginning of each run so we are starting the test from the same state.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+            b <- q.take.as(true)
+            c <- q.poll.map(_.isEmpty)
+            _ <- TestClock.adjust(60.minutes)
+            d <- q.take.as(true)
+            e <- q.poll.map(_.isEmpty)
+          } yield a && b && c && d && e
+          assertM(example, isTrue)
+        },
+        //todo fix - looks like fiberTime is 100x larger than it's supposed to be - conversion error somewhere?
+        testM("fiber time is not subject to race conditions") {
```
adamgfraser: The problem here is that since `zipPar` is joining back the fibers after each run the fiber time is being incremented on each run by 2 milliseconds, which over the course of 100 runs becomes 200 milliseconds, hence your off by a factor of 100 error. This is getting further complicated by the bug in `nonFlaky` I am fixing in #2013 which is causing the test to run for all 100 iterations instead of failing early.

You can fix this by keeping the clock time in sync with the fiber time.

```scala
testM("fiber time is not subject to race conditions") {
  for {
    _        <- adjust(2.millis)
    _        <- sleep(2.millis).zipPar(sleep(1.millis))
    result   <- fiberTime
    expected <- clock.currentTime(TimeUnit.MILLISECONDS)
  } yield assert(result.toMillis, equalTo(expected))
} @@ nonFlaky(100)
```

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+          for {
+            latch <- Promise.make[Nothing, Unit]
+            _     <- (sleep(10.hours) *> latch.succeed(())).fork
+            _     <- setTime(11.hours)
+            _     <- latch.await
+          } yield assert((), anything)
+        } @@ nonFlaky(100)
+          @@ timeout(10.seconds),
+        //todo fix - times out
+        testM("sleep does sleep instantly when sleep duration less than set time") {
```
adamgfraser: The problem here is that since the `sleep(10.hours)` call is occurring on the main fiber the fiber time is being incremented, so in the second run when you sleep for another 10 hours the fiber is not going to wake up until the clock time is 20 hours, but you are only setting the time to 11 hours. You can fix this by changing `setTime(11.hours)` to `adjust(10.hours)` so the clock time stays in sync with the fiber time.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+          for {
+            ref    <- Ref.make(true)
+            _      <- (sleep(10.hours) *> ref.set(false)).fork
+            _      <- adjust(9.hours)
+            result <- ref.get
+          } yield assert(result, isTrue)
+        } @@ nonFlaky(100),
+        //todo  remove `@@ after(setTime(0.hours))` if possible. This doesn't depend on absolute time so I don't think
+        // it should rely on resetting the clock. Fails intermittently without it.
+        testM("sleep correctly handles multiple sleeps") {
```
adamgfraser: This isn't working without resetting the clock because the `sleep` calls never occur on the main fiber, so the `sleep` calls are always setting up the first fiber to wake up at a clock time of 3 hours and the second fiber to wake up after a clock time of 1 hour. That works fine for the first run, but by the second run the clock time is already set to 4 hours so both effects will run immediately, leading to flakiness.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: that'll make it run, but we wouldn't be testing the same thing anymore.
spec is "sleep does sleep instantly when sleep duration _less_ than _set time_"
but the changes you've proposed would remove set time all together and the sleep duration would be identical rather than less than.

I think there's a bigger issue here with `nonFlaky` - is there a way to make it sync the clock with fiberTime at the end of each iteration? or provide a new environment? or at least the option for one? At the moment is has the rather ironic consequence of making otherwise solid tests flaky unless you dance the dance of sleep and adjust perfectly every time!

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: as below I don't think this is testing the same thing anymore. The original test was pushing the clock time out to infinity, so it'd never be in sync with fiber time even the first time through. I think what it was trying to test is that with the clock set to infinity, fiber time still advances by the expected amount (2ms) each time

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: I'll have to do some more reading of comments/doco on this one, but happy enought to leave the clock reset as it doesn't change the meaning of the test. I thought that forking would result in child fibers that would have the same fiber time as the fiber that created them, & sleep would sleep for the specified duration from whatever the current clock time is set to. Is there anywhere this interaction is documented in one place? I can't find anything on the website that even mentions `fiberTime`

adamgfraser: Yes, I think fine to leave as is. The reason we need the fiber time in addition to the clock time is otherwise in code like this there is a race condition.

```scala
for {
  fiber <- TestClock.sleep(1.second).fork
  _ <- TestClock.adjust(1.second)
  _ <- fiber.join
yield ()
}
```

The `sleep` call and the `adjust` call are occurring on separate fibers so we can't guarantee the ordering between them. If the sleep call happens first we sleep until the time is 1 second and then adjust to 1 second so everything is fine. But if the `adjust` call completes first the clock gets set to 1 second, now the `sleep` call is sleeping until 2 seconds, and we hang forever on `join`. To get around this we need the need the fiber time concept and we need the inheritance behavior so that we get the expected behavior on forks and joins (e.g. if you fork a fiber and have it sleep for 2 seconds, then fork a fiber from that fiber and have it sleep for 2 seconds, then advance the clock by three seconds, the second fiber should no wake up.

We have some ideas of how we could simplify this but they require changing the runtime to allow the `TestClock` to examine the queue of pending actions to avoid the race conditions.

There is some documentation on this [here](https://github.com/zio/zio/blob/48891c29d0ba58e953e30cde6c598e381d84ffbd/test/shared/src/main/scala/zio/test/environment/TestClock.scala#L61).

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
adamgfraser: Maybe we change the test label to "sleep does sleep instantly when sleep duration less than or equal to clock time"? The idea of the test is more about making sure effects run than the specific `setTime` versus `adjust` method and making sure tests effects run if the time is less than or equal is actually a stronger guarantee that we want. There are other tricks we could do to keep the clock time and fiber time in sync if you want but this seemed like the simplest way to do.

I had the same thought about `nonFlaky`. Unfortunately I don't think there is a general solution. The problem is that from the perspective of `nonFlaky` `R` is completely polymorphic. It is easy to think in a case like this why can't we just provide a new copy of the environment. But the environment might be something like a database connection where we can't just create a new database connection. We could create a new `TestClock` in each test here and then be sure to `provide` it to each effect but that doesn't solve the problem in general.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
adamgfraser: The goal of this test is to check for a race condition in `ZipPar` with regard to the fiber time. As long as the clock time is equal to or greater than 2 milliseconds (ignoring repeated runs), both the left and right fibers in `zipPar` will wake up. There is a potential race condition that normally when child fibers are joined the `FiberRef` of the "last child" joined takes precedence. So there could be a situation where the left join takes places before the right join, leading to the right fiber time of 1 millisecond being inherited by the parent and leading to data corruption. We solved that by adding logic to `FiberRef` allowing to specify custom logic for joining and using a "greater duration wins" rule and this test is to prevent regressions in that. So in the new test the only way the result is equal to the expected value each time is if the fiber time is always incremented by 2 milliseconds, which is what we want.

There are definitely other way you could do this if you want but this seemed like a natural one to me.

test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
unclebob418: > the testing framework provides a separate copy of the TestEnvironment to each test, but that environment is shared between successive invocations of the same test via nonFlaky.

how does this work with database connections? surely it's not providing 100s of connections to run the tests. Could this same mechanism be exploited for `nonFlaky`?

Conceptually I think `nonFlaky` could turn a test into a suite consiting of `n0` copies of the test. 

e.g.
```
object concept
    extends ZIOBaseSpec(
      suite("ClockSpec")(
        testM("sleep does not require passage of clock time") {
          ???
        } @@ nonFlaky(100)
      )
    )
```
and turning it into this:
```
object conceptTranslated
    extends ZIOBaseSpec(
      suite("ClockSpec")(
        suite("sleep does not require passage of clock time")(
          testM("sleep does not require passage of clock time") {
            ???
          },
          testM("sleep does not require passage of clock time") {
            ???
          },
          testM("sleep does not require passage of clock time") {
            ???
          }
          /// 97 more copies
        )
      )
    )
```
you'd probably want a way to suppress the logging of this to the lowest suite level. This might mean moving it away from an aspect, more like `Spec#provideManagedShared`

adamgfraser: With something like a database connection you would provide it once to the entire suite with `provideShared`. So then if you repeat a test 100 times you are necessarily doing it with the same database connection and may need to perform cleanup actions after each run using `after`. My point is just that `nonFlaky` has to be polymorphic in `R`, so it has to work for any data type whether it is one we can create new copies of like `TestClock` or one that is outside of our control like a database connection.

For that reason, I don't think it really works to create more copies of the test. That happens to work for the standard ZIO environment types because the `TestExecutor` provides a separate copy of the `TestEnvironment` to each test but doesn't work for any other environment type. It would also require a lot of special case logic in different places because now we need to keep track of whether tests are regular tests or "copied" tests that need to be reported differently.

I think a better solution is to document the fact that the same environment will be shared across runs and that the user may want to provide a separate copy of a required environment for separate runs or use combinators like `after` to ensure a consistent state for each one.

core/shared/src/main/scala-2.11/zio/clock.scala
```diff
@@ -16,8 +16,9 @@
 
 package zio
 
-import zio.duration.Duration
```
adamgfraser: Really minor thing can you order imports? Then let's get this merged!

unclebob418: done!

unclebob418:
Thanks for all your help once again. I'm starting to get a better understanding of how it all works! I have to do some real work for a change but should get to RandomSpec tomorrow

PR #2012, opened 2019-10-21
Properly name ZStream#mapConcat* combinators

vasilmkd:
I noticed a discrepancy between the function signatures of
`ZStream#mapConcat` and `ZStreamChunk#mapConcat` (which is actually `ZStreamChunk#mapConcatChunk`.

@iravid WDYT we should do? Is there a good use case for `ZStreamChunk#mapConcat` that accepts a function that returns an `Iterable`, instead of a `Chunk`?

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -1079,7 +1079,7 @@ object SinkSpec
               assertM(
                 Stream("\n")
                   .transduce(ZSink.splitLines)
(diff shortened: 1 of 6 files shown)


iravid:
Hmm yes, I thought we fixed that before. The `mapConcat` functions should use iterables, while the `mapConcatChunk` versions should use a chunk. Both in ZStream and ZStreamChunk. Would be great if you could align that :-)

vasilmkd:
No problem, disregard this PR in its current state, because it's not correct according to that requirement.

vasilmkd:
@iravid I did as you suggested. Ready for review.

PR #2016, opened 2019-10-22
Implement `ZManaged.memoize`

### LINKED ISSUE #1263: Add ZManaged#memoize

mlangc:
Originally, I tried to implement this like `ZIO.memoize`. While this approach looks mean and lean for the happy path, getting this right for fatal errors and interruptions is far less trivial (see https://github.com/mlangc/zio/commit/016e2640712feb49c9c49a6fd3a0bab495b2a365#diff-ba99038d96380b834448ea7c9ddd3c05R776 to get an impression).

The current implementation uses a different approach, that is based on a `RefM` (this PR also fixes a `RefM` bug related to fatal errors that hit me). Maybe a normal `Ref` together with one or more `Promises` could be used too, at the cost of added complexity.

One last note on the naming: I choose `ZManaged.memoize` over `ZManaged.shared`, to emphasize that `ZManaged.memoize` corresponds directly to `ZIO.memoize`. Also, `shared` reminds me of reference counters (maybe from my old C++ times), and might be used to implement different semantics in the future.

Fixes #1263

### DIFF
--- a/core/shared/src/main/scala/zio/RefM.scala
+++ b/core/shared/src/main/scala/zio/RefM.scala
@@ -169,7 +169,7 @@ object RefM extends Serializable {
     for {
       ref   <- Ref.make(a)
       queue <- Queue.bounded[Bundle[_, A, _]](n)
-      _     <- queue.take.flatMap(b => ref.get.flatMap(a => b.run(a, ref, onDefect))).forever.fork
+      _     <- queue.take.flatMap(b => ref.get.flatMap(a => b.run(a, ref, onDefect))).forever.fork.daemon
     } yield new RefM[A](ref, queue)
 
 }
--- a/core/shared/src/main/scala/zio/RefM.scala
+++ b/core/shared/src/main/scala/zio/RefM.scala
@@ -146,7 +146,7 @@ object RefM extends Serializable {
       interrupted.get.flatMap {
         case Some(cause) => onDefect(cause)
         case None =>
-          update(a).foldM(e => onDefect(Cause.fail(e)) <* promise.fail(e), {
+          update(a).foldCauseM(c => onDefect(c).ensuring(promise.halt(c)), {
             case (b, a) => ref.set(a) <* promise.succeed(b)
           })
       }
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -886,10 +886,8 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
               }
           }.flatMap(ZIO.done)
 
-        val release1 = (_: Exit[_, _]) => ZIO.unit
-
         val acquire2: ZIO[R, E, ZManaged[R, E, A]] =
-          ZIO.succeed(ZManaged(ZIO.succeed(Reservation(acquire1, release1))))
+          ZIO.succeed(acquire1.toManaged_)
 
         val release2 = (_: Exit[_, _]) =>
           ref.updateSome {
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -872,6 +872,33 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
       }
     }
 
+  def memoize: ZManaged[R, E, ZManaged[R, E, A]] =
+    ZManaged {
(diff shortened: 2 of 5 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -773,6 +773,33 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
+                self.reserve.flatMap { res =>
+                  restore(res.acquire).run.map(e => e -> Some(res -> e))
+                }
+              }
+          }.flatMap(ZIO.done)
+
+        val release1 = (_: Exit[_, _]) => ZIO.unit
+
+        val acquire2: ZIO[R, E, ZManaged[R, E, A]] =
+          ZIO.succeed(ZManaged(ZIO.succeed(Reservation(acquire1, release1))))
```
```suggestion
          ZIO.succeed(acquire1.toManaged_)
```

core/shared/src/main/scala/zio/RefM.scala
```diff
@@ -146,7 +146,7 @@ object RefM extends Serializable {
       interrupted.get.flatMap {
         case Some(cause) => onDefect(cause)
         case None =>
-          update(a).foldM(e => onDefect(Cause.fail(e)) <* promise.fail(e), {
+          update(a).foldCauseM(c => onDefect(c).ensuring(promise.halt(c)), {
```
iravid: Can you add a test for this change?

mlangc: Of course, I will look into this over the weekend.

core/shared/src/main/scala/zio/ZManaged.scala
mlangc: :-) this looks better indeed - thanks for spotting

mlangc:
@iravid thanks for the review. Strangely enough, rebasing introduced a regression:
```
[info] ZIO Test
[info] - ZManaged
[info]   - memoize
[info]     - behaves properly if use is interrupted
```
I will investigate over the weekend, but if you have an idea about the possible cause let me know.

iravid:
It's probably related to the autosupervision work.

Another suggestion before you dive into this: I think you can replace RefM with a Ref and a Semaphore. It's a simpler mechanism that is easier to introspect and figure out what goes wrong; RefM has a bunch of moving parts that make it harder to diagnose problems in.

mlangc:
Thanks for the hints, I'll take that into consideration! I'm still wondering if the regression I see is because of a subtle, intended semantic change, or a because of a regression in another component.

core/shared/src/main/scala/zio/RefM.scala
mlangc: I've just added a test (see fecf6b412ac00fb7afe3b8c7bb23067656500eb5). Without the change from above, this test times out.

mlangc:
@iravid I found out more about the test running into a timeout after rebasing with auto supervision: The reason was not that finalizers where not run, but that `Fiber.interrupt` blocked, because the fiber that consumes the queue in `RefM` just keeps running. I first fixed the test by forking the interrupt call (see 545768bd344aebf07941efde9ac19dfa6fdb9251) - although this seems more like a workaround than a proper solution. The better approach in my opinion is to add `daemon` to the fiber consuming the `Queue` in `RefM` (see 358b1885c664c70036da5647947aca0094189fc6) . What is your take on this?

Note that I can still look into dropping `RefM` in favor of `Ref` and `Semaphore` if you prefer so, but I wanted to find out why auto supervision introduced a regression before.

iravid:
@mlangc Ah good catch. Yes, that's one of the "intricacies" with RefM. It may make sense to mark that consumer fiber as `ZIO.daemon`, but I'd need to think that through.

If you don't want to wait, I would suggest just getting rid of RefM in this combinator for now.

iravid:
Actually, I think marking that fiber as `ZIO.daemon` does make sense, as it should be tied to the lifecycle of `RefM`, and not to that of the fiber that created `RefM`.

mlangc:
@iravid indeed, tying the internal fiber to the live-cycle of the parent fiber is actually wrong, and I've just added a test (see 1b921d39a89384fb916e41754f522dc2e48161de) that demonstrates that (it fails without the `daemon`).

iravid:
Ok, sounds good - thanks for your work on this @mlangc. Will merge when CI is done.

PR #2017, opened 2019-10-22
Add TMap

### LINKED ISSUE #1178: Implement `TMap`
Implement `zio.stm.TMap`:

```scala
class TMap[K, V] private (hashmap: TRef[TArray[List[(K, V)]]) { ... }
```

with methods similar to `Map` inside it, allowing access and updating by key, as well as folding and other types of functionality compatible with STM.

mijicd:
Closes #1178 (finally)

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -103,12 +103,9 @@ class TMap[K, V] private (
    */
   final def put[E](k: K, v: V): STM[E, Unit] = {
     def upsert(bucket: List[(K, V)]): STM[Nothing, List[(K, V)]] = {
-      val newBucket = bucket match {
-        case Nil => List(k -> v)
-        case xs  => xs.map(kv => if (kv._1 == k) (k, v) else kv)
-      }
-
-      val diff = newBucket.size - bucket.size
+      val hasKey    = bucket.find(_._1 == k).isDefined
+      val newBucket = if (hasKey) bucket.map(kv => if (kv._1 == k) (k, v) else kv) else (k, v) :: bucket
+      val diff      = newBucket.size - bucket.size
 
       if (diff == 0) STM.succeed(newBucket) else tSize.update(_ + diff).as(newBucket)
     }
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -48,9 +48,10 @@ class TMap[K, V] private (buckets: TRef[TArray[List[(K, V)]]]) { self =>
     get(k).map(_.getOrElse(default))
 
   final def map[K2, V2](f: ((K, V)) => (K2, V2)): STM[Nothing, TMap[K2, V2]] =
-    self.fold(List.empty[(K2, V2)])((acc, pair) => f(pair) :: acc).flatMap(TMap(_))
+    self.fold(List.empty[(K2, V2)])((acc, kv) => f(kv) :: acc).flatMap(TMap(_))
 
-  final def mapM[E, K2, V2](f: ((K, V)) => STM[E, (K2, V2)]): STM[E, TMap[K2, V2]] = ???
+  final def mapM[E, K2, V2](f: ((K, V)) => STM[E, (K2, V2)]): STM[E, TMap[K2, V2]] =
+    self.foldM(List.empty[(K2, V2)])((acc, kv) => f(kv).map(_ :: acc)).flatMap(TMap(_))
 
   final def put[E](k: K, v: V): STM[E, Unit] = {
     def update(bucket: List[(K, V)]): List[(K, V)] =
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -59,12 +59,6 @@ class TMap[K, V] private (buckets: TArray[List[(K, V)]]) { self =>
 }
 
 object TMap {
-
-  /**
-   * Makes an empty `TMap`.
-   */
-  final def empty[K, V]: STM[Nothing, TMap[K, V]] = apply(List.empty[(K, V)])
-
   /**
    * Makes a new `TMap` that is initialized with the specified values.
    */
@@ -82,7 +76,12 @@ object TMap {
     STM.collectAll(stmBuckets).map(refs => new TMap(TArray(refs.toArray)))
   }
 
-  private final def indexOf[K](k: K): Int = k.hashCode() % DefaultSize
+  /**
+   * Makes an empty `TMap`.
+   */
+  final def empty[K, V]: STM[Nothing, TMap[K, V]] = apply(List.empty[(K, V)])
 
   private final val DefaultSize = 1000
+
+  private final def indexOf[K](k: K): Int = k.hashCode() % DefaultSize
 }
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -174,10 +174,10 @@ object TMap {
     }
 
     for {
-      tBuckets  <- STM.collectAll(buckets.map(b => TRef.make(b)))
-      tArray    <- TRef.make(TArray(tBuckets.toArray))
+      tChains   <- STM.collectAll(buckets.map(b => TRef.make(b)))
+      tBuckets  <- TRef.make(TArray(tChains.toArray))
       tCapacity <- TRef.make(capacity)
-    } yield new TMap(tArray, tCapacity)
+    } yield new TMap(tBuckets, tCapacity)
   }
 
   private final val DefaultCapacity = 100
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -45,8 +45,6 @@ class TMap[K, V] private (table: TArray[List[(K, V)]]) {
 }
 
 object TMap {
-  private final val DefaultSize = 1000
-
   /**
    * Makes an empty `TMap`.
    */
@@ -70,4 +68,6 @@ object TMap {
   }
 
   private final def indexOf[K](k: K): Int = k.hashCode() % DefaultSize
+
+  private final val DefaultSize = 1000
 }
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -46,7 +46,8 @@ class TMap[K, V] private (buckets: TRef[TArray[List[(K, V)]]]) { self =>
   final def getOrElse[E](k: K, default: => V): STM[E, V] =
     get(k).map(_.getOrElse(default))
 
-  final def map[K2, V2](f: ((K, V)) => (K2, V2)): STM[Nothing, TMap[K2, V2]] = ???
+  final def map[K2, V2](f: ((K, V)) => (K2, V2)): STM[Nothing, TMap[K2, V2]] =
+    accessM(_.map(_.map(f))).flatMap(t => TRef.make(t)).map(r => new TMap(r))
 
   final def mapM[E, K2, V2](f: ((K, V)) => STM[E, (K2, V2)]): STM[E, TMap[K2, V2]] = ???
 
@@ -60,7 +61,7 @@ class TMap[K, V] private (buckets: TRef[TArray[List[(K, V)]]]) { self =>
     accessM(_.update(indexOf(k), update)).as(self)
   }
 
-  private def accessM[E, A](f: TArray[List[(K, V)]] => STM[E, A]) = buckets.get.flatMap(f)
+  private def accessM[E, A](f: TArray[List[(K, V)]] => STM[E, A]): STM[E, A] = buckets.get.flatMap(f)
 }
 
 object TMap {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,185 @@
...
+    tCapacity.get.map(c => k.hashCode() % c)
+}
+
+object TMap {
+
+  /**
+   * Makes a new `TMap` that is initialized with the specified values.
+   */
+  final def apply[K, V](data: => List[(K, V)]): STM[Nothing, TMap[K, V]] = {
+    val capacity = if (data.isEmpty) DefaultCapacity else 2 * data.size
```
jdegoes: I would change to varargs (like most collections) and add a fromIterable method that takes an Iterable (not List).

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,185 @@
...
+   * Retrieves value associated with given key or default value, in case the
+   * key isn't present.
+   */
+  final def getOrElse[E](k: K, default: => V): STM[E, V] =
+    get(k).map(_.getOrElse(default))
+
+  /**
+   * Creates new [[TMap]] by mapping all bindings using pure function.
+   */
+  final def map[K2, V2](f: ((K, V)) => (K2, V2)): STM[Nothing, TMap[K2, V2]] =
```
jdegoes: Personally I find these pure methods confusing. I would delete map and mapM, leaving only transform and transformM

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,185 @@
...
+        _          <- self.tBuckets.set(newBuckets)
+        _          <- self.tCapacity.set(newCapacity)
+      } yield ()
+    }
+
+    for {
+      buckets     <- tBuckets.get
+      idx         <- indexOf(k)
+      _           <- buckets.update(idx, upsert)
+      data        <- self.fold(List.empty[(K, V)])((acc, kv) => kv :: acc)
```
jdegoes: Doesn't this grab the whole contents of the map, even if there is no need to resize? We should aim for maximum laziness to avoid doing unnecessary worm

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Done.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Agreed, done.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Good point, stored the size in `TRef`.

mijicd:
@jdegoes Done with the changes. Note that the PR uses a part of #2019 , therefore that one should be reviewed and (hopefully) merged first.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * Tests whether or not map contains a key.
+   */
+  final def contains[E](k: K): STM[E, Boolean] =
+    get(k).map(_.isDefined)
+
+  /**
+   * Removes binding for given key.
+   */
+  final def delete[E](k: K): STM[E, Unit] = {
```
mschuwalow: What is the reason for not just using Nothing?

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,211 @@
...
+      capacity    <- tCapacity.get
+      needsResize = capacity * TMap.LoadFactor < size
+      _           <- if (needsResize) resize(capacity * 2) else STM.unit
+    } yield ()
+  }
+
+  /**
+   * Atomically updates all bindings using pure function.
+   */
+  final def transform(f: ((K, V)) => (K, V)): STM[Nothing, Unit] =
```
mschuwalow: Does it make sense to also include mapValues?

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Good point. Applied the same change to `get`, `getOrElse` and `put`.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: I'm undecided about this one, clearly it's a low-hanging fruit:

```scala
final def mapValues(f: V => V): STM[Nothing, Unit] =
  transform(kv => kv._1 -> f(kv._2))
```

The reason I'm hesitating to put it is the API "consistency" as we've removed `map` / `mapM` already.

jdegoes: Should be called transformValues, right?

mijicd: Introduced `transformValues` and `transformValuesM`.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+class TMap[K, V] private (
+  private val tBuckets: TRef[TArray[List[(K, V)]]],
+  private val tCapacity: TRef[Int],
+  private val tSize: TRef[Int]
+) { self =>
+
+  /**
+   * Tests whether or not map contains a key.
+   */
+  final def contains[E](k: K): STM[E, Boolean] =
```
jdegoes: Replace `E` with `Nothing`.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+      capacity    <- tCapacity.get
+      needsResize = capacity * TMap.LoadFactor < size
+      _           <- if (needsResize) resize(capacity * 2) else STM.unit
+    } yield ()
+  }
+
+  /**
+   * Atomically updates all bindings using pure function.
+   */
+  final def transform(f: ((K, V)) => (K, V)): STM[Nothing, Unit] =
```
jdegoes: I think this doesn't have to be a tuple `((K, V))` and can be just a multi-parameter list `(K, V)` (`Function2`).

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Done.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+      data       <- self.fold(List.empty[(K, V)])((acc, kv) => f(kv) :: acc)
+      tmap       <- TMap.fromIterable(data)
+      newBuckets <- tmap.tBuckets.get
+      _          <- self.tBuckets.set(newBuckets)
+    } yield ()
+
+  /**
+   * Atomically updates all bindings using effectful function.
+   */
+  final def transformM[E](f: ((K, V)) => STM[E, (K, V)]): STM[E, Unit] =
```
jdegoes: I think this doesn't have to be a tuple `((K, V))` and can be just a multi-parameter list `(K, V)` (`Function2`).

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+      _           <- if (needsResize) resize(capacity * 2) else STM.unit
+    } yield ()
+  }
+
+  /**
+   * Atomically updates all bindings using pure function.
+   */
+  final def transform(f: ((K, V)) => (K, V)): STM[Nothing, Unit] =
+    for {
+      data       <- self.fold(List.empty[(K, V)])((acc, kv) => f(kv) :: acc)
```
jdegoes: This code is probably duplicated a few times and should be a `toList` or something method. Also we don't have to fix it here, of course, but in theory, you can be _much_ more efficient, because you can reuse the same `TArray` and just change the entries.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+      newBuckets <- tmap.tBuckets.get
+      _          <- self.tBuckets.set(newBuckets)
+    } yield ()
+
+  /**
+   * Atomically updates all bindings using effectful function.
+   */
+  final def transformM[E](f: ((K, V)) => STM[E, (K, V)]): STM[E, Unit] =
+    for {
+      data       <- self.foldM(List.empty[(K, V)])((acc, kv) => f(kv).map(_ :: acc))
```
jdegoes: Again in theory, you can be _much_ more efficient, because you can reuse the same `TArray` and just change the entries.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+      tmap       <- TMap.fromIterable(data)
+      newBuckets <- tmap.tBuckets.get
+      _          <- self.tBuckets.set(newBuckets)
+    } yield ()
+
+  /**
+   * Atomically updates all values using pure function.
+   */
+  final def transformValues(f: V => V): STM[Nothing, Unit] =
```
jdegoes: This one could in theory be super-efficient implementation which just modifies the values in-place.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+    transform(kv => kv._1 -> f(kv._2))
+
+  /**
+   * Atomically updates all values using effectful function.
+   */
+  final def transformValuesM[E](f: V => STM[E, V]): STM[E, Unit] =
```
jdegoes: Ditto.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+    transformM(kv => f(kv._2).map(v => kv._1 -> v))
+
+  /**
+   * Removes bindings matching predicate.
+   */
+  final def removeIf(p: ((K, V)) => Boolean): STM[Nothing, Unit] =
```
jdegoes: Why a tuple `((K, V))` instead of `(K, V)` (`Function2`).

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,223 @@
...
+    tBuckets.get.flatMap(_.transform(_.filterNot(p)))
+
+  /**
+   * Retains bindings matching predicate.
+   */
+  final def retainIf(p: ((K, V)) => Boolean): STM[Nothing, Unit] =
```
jdegoes: Why a tuple `((K, V))` instead of `(K, V)` (`Function2`).

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Replaced.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Replaced.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Replaced.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Replaced.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Something like this?

```scala
  final def transform(f: (K, V) => (K, V)): STM[Nothing, Unit] =
    for {
      data     <- self.fold(List.empty[(K, V)])((acc, kv) => f(kv._1, kv._2) :: acc)
      buckets  <- self.tBuckets.get
      capacity <- self.tCapacity.get
      _        <- buckets.transform(_ => Nil)
      updates  = data.map(kv => buckets.update(kv._1.hashCode() % capacity, kv :: _))
      _        <- STM.collectAll(updates)
    } yield ()
```

mijicd:
@jdegoes removed tupled parameters, refactored `transform*` functions. I've also exposed `toList`, `keys` and `values`. 

Some of the changes resulted in removal of duplicated code, but decreased efficiency (e.g. expressing `fold` in terms of `foldM` and `transform` in terms of `transformM`). These can be easily reverted if needed.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -0,0 +1,250 @@
...
+    for {
+      buckets <- tBuckets.get
+      res     <- buckets.foldM(zero)((acc, bucket) => loopM(STM.succeed(acc), bucket))
+    } yield res
+  }
+
+  /**
+   * Atomically performs side-effect for each binding present in map.
+   */
+  final def foreach[E](f: ((K, V)) => STM[E, Unit]): STM[E, Unit] =
```
jdegoes: Nested tuple `((K, V))` instead of 2 param function here will make it harder to use in general.

core/shared/src/main/scala/zio/stm/TMap.scala
jdegoes: Yes, exactly.

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Good catch, removed.

jdegoes:
Awesome work! 🎉

PR #2023, opened 2019-10-23
Safer effect building code in Chunk

vasilmkd:
Don't use side effecting code in pure functions.

@adamgfraser

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -146,14 +146,18 @@ sealed trait Chunk[+A] { self =>
     while (i < len) {
       val elem = self(i)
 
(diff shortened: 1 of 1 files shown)


jdegoes:
Btw if I am wrong about the mutation not escaping, let me know, a quick glance did not reveal anything suspicious.

vasilmkd:
I agree. I was thinking about this as well. In that case, maybe we can get away with removing the `Managed.effectTotal`s in `StreamEffect.scala` as well?

jdegoes:
> I agree. I was thinking about this as well. In that case, maybe we can get away with removing the Managed.effectTotals in StreamEffect.scala as well?

Anywhere it is safe to do so, we absolutely should do that. 👏 👍

PR #2025, opened 2019-10-24
WIP fix nonFlaky fiberTime issue

unclebob418:
@adamgfraser, sorry to bug you with more nonFlaky stuff but you said something the other day that got me thinking:

> This isn't working without resetting the clock because the sleep calls never occur on the main fiber

This made me think if I changed `nonFlaky` so the tests weren't running on the main fiber, we'd have a `fiberTime` of 0 at the start of each iteration. I did some fiddling and (after more time than I'd care to admit) I think I've got it sorted. This should eliminate all the issues with keeping clock time and fiber time in sync - a simple `@@ after(setTime(0.hours))` should suffice!

Can you please check this is solid and won't break all the things?

### DIFF
--- a/test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
+++ b/test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
@@ -100,6 +100,7 @@ case class TestEnvironment(
         def currentTime(unit: TimeUnit): UIO[Long]           = live.provide(zio.clock.currentTime(unit))
         val fiberTime: UIO[Duration]                         = UIO.succeed(Duration.Zero)
         val nanoTime: UIO[Long]                              = live.provide(zio.clock.nanoTime)
+        override val save: UIO[UIO[Unit]]                    = UIO(UIO.unit)
         def setDateTime(dateTime: OffsetDateTime): UIO[Unit] = UIO.unit
         def setTime(duration: Duration): UIO[Unit]           = UIO.unit
         def setTimeZone(zone: ZoneId): UIO[Unit]             = UIO.unit
@@ -125,6 +126,7 @@ case class TestEnvironment(
         val output: UIO[Vector[String]]          = UIO.succeed(Vector.empty)
         def putStr(line: String): UIO[Unit]      = live.provide(zio.console.putStr(line))
         def putStrLn(line: String): UIO[Unit]    = live.provide(zio.console.putStrLn(line))
+        override val save: UIO[UIO[Unit]]        = UIO(UIO.unit)
       }
     }
 
@@ -163,6 +165,7 @@ case class TestEnvironment(
         def nextLong(n: Long): UIO[Long]                = live.provide(zio.random.nextLong(n))
         val nextPrintableChar: UIO[Char]                = live.provide(zio.random.nextPrintableChar)
         def nextString(length: Int): UIO[String]        = live.provide(zio.random.nextString(length))
+        override val save: UIO[UIO[Unit]]               = UIO(UIO.unit)
         def setSeed(seed: Long): UIO[Unit]              = UIO.unit
         def shuffle[A](list: List[A]): UIO[List[A]]     = UIO.succeed(list)
       }
@@ -183,6 +186,7 @@ case class TestEnvironment(
         def property(prop: String): Task[Option[String]]                 = live.provide(zio.system.property(prop))
         def putEnv(name: String, value: String): UIO[Unit]               = UIO.unit
         def putProperty(name: String, value: String): UIO[Unit]          = UIO.unit
+        override val save: UIO[UIO[Unit]]                                = UIO(UIO.unit)
         def setLineSeparator(lineSep: String): UIO[Unit]                 = UIO.unit
       }
     }
--- a/test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
+++ b/test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
@@ -100,7 +100,7 @@ case class TestEnvironment(
         def currentTime(unit: TimeUnit): UIO[Long]           = live.provide(zio.clock.currentTime(unit))
         val fiberTime: UIO[Duration]                         = UIO.succeed(Duration.Zero)
         val nanoTime: UIO[Long]                              = live.provide(zio.clock.nanoTime)
-        override val save: UIO[UIO[Unit]]                    = UIO(UIO.unit)
+        val save: UIO[UIO[Unit]]                             = UIO.succeed(UIO.unit)
         def setDateTime(dateTime: OffsetDateTime): UIO[Unit] = UIO.unit
         def setTime(duration: Duration): UIO[Unit]           = UIO.unit
         def setTimeZone(zone: ZoneId): UIO[Unit]             = UIO.unit
@@ -126,7 +126,7 @@ case class TestEnvironment(
         val output: UIO[Vector[String]]          = UIO.succeed(Vector.empty)
         def putStr(line: String): UIO[Unit]      = live.provide(zio.console.putStr(line))
         def putStrLn(line: String): UIO[Unit]    = live.provide(zio.console.putStrLn(line))
-        override val save: UIO[UIO[Unit]]        = UIO(UIO.unit)
+        val save: UIO[UIO[Unit]]                 = UIO.succeed(UIO.unit)
       }
     }
 
@@ -165,7 +165,7 @@ case class TestEnvironment(
         def nextLong(n: Long): UIO[Long]                = live.provide(zio.random.nextLong(n))
         val nextPrintableChar: UIO[Char]                = live.provide(zio.random.nextPrintableChar)
         def nextString(length: Int): UIO[String]        = live.provide(zio.random.nextString(length))
-        override val save: UIO[UIO[Unit]]               = UIO(UIO.unit)
+        val save: UIO[UIO[Unit]]                        = UIO.succeed(UIO.unit)
         def setSeed(seed: Long): UIO[Unit]              = UIO.unit
         def shuffle[A](list: List[A]): UIO[List[A]]     = UIO.succeed(list)
       }
@@ -186,7 +186,7 @@ case class TestEnvironment(
         def property(prop: String): Task[Option[String]]                 = live.provide(zio.system.property(prop))
         def putEnv(name: String, value: String): UIO[Unit]               = UIO.unit
         def putProperty(name: String, value: String): UIO[Unit]          = UIO.unit
-        override val save: UIO[UIO[Unit]]                                = UIO(UIO.unit)
+        val save: UIO[UIO[Unit]]                                         = UIO.succeed(UIO.unit)
         def setLineSeparator(lineSep: String): UIO[Unit]                 = UIO.unit
       }
     }
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -300,7 +300,7 @@ object TestAspect extends TimeoutVariants {
    * satisfies the specified assertion.
    */
   def ifEnv(env: String, assertion: Assertion[String]): TestAspectAtLeastR[Live[System]] =
-    new PerTest.AtLeast[Live[System]] {
+    new PerTest.AtLeastR[Live[System]] {
       def perTest[R <: Live[System], E, S](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -338,7 +338,7 @@ object TestAspect extends TimeoutVariants {
     prop: String,
     assertion: Assertion[String]
   ): TestAspectAtLeastR[Live[System]] =
-    new PerTest.AtLeast[Live[System]] {
+    new PerTest.AtLeastR[Live[System]] {
       def perTest[R <: Live[System], E, S](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -530,7 +530,7 @@ object TestAspect extends TimeoutVariants {
     duration: Duration,
     interruptDuration: Duration = 1.second
   ): TestAspectAtLeastR[Live[Clock]] =
-    new PerTest.AtLeast[Live[Clock]] {
+    new PerTest.AtLeastR[Live[Clock]] {
       def perTest[R >: Nothing <: Live[Clock], E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
@@ -575,7 +575,7 @@ object TestAspect extends TimeoutVariants {
     /**
      * A `PerTest.AtLeast[R]` is a `TestAspect.PerTest` that that requires at least an R in its environment
      */
-    type AtLeast[R] =
+    type AtLeastR[R] =
       TestAspect.PerTest[Nothing, R, Nothing, Any, Nothing, Any]
 
     /**
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -106,7 +106,7 @@ object TestAspect extends TimeoutVariants {
    * An aspect that marks tests as ignored.
    */
   val ignore: TestAspectPoly =
-    new PerTestPoly {
+    new PerTest.Poly {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -126,11 +126,11 @@ object TestAspect extends TimeoutVariants {
 
   /**
    * Constructs an aspect that evaluates every test between two effects, `before` and `after`,
-   * where the result of `before` can be `used` in after.
+   * where the result of `before` can be used in `after`.
    */
   def around[R0, E0, A0](
     before: ZIO[R0, E0, A0]
-  )(after: A0 => ZIO[R0, Nothing, Any]): PerTest[Nothing, R0, E0, Any, Nothing, Any] =
+  )(after: A0 => ZIO[R0, Nothing, Any]): TestAspect[Nothing, R0, E0, Any, Nothing, Any] =
     new TestAspect.PerTest[Nothing, R0, E0, Any, Nothing, Any] {
       def perTest[R >: Nothing <: R0, E >: E0 <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
@@ -144,7 +144,7 @@ object TestAspect extends TimeoutVariants {
   def around_[R0, E0](
     before: ZIO[R0, E0, Any],
     after: ZIO[R0, Nothing, Any]
-  ): PerTest[Nothing, R0, E0, Any, Nothing, Any] = around(before)(_ => after)
+  ): TestAspect[Nothing, R0, E0, Any, Nothing, Any] = around(before)(_ => after)
 
   /**
    * Constructs an aspect that evaluates every test inside the context of the managed function.
@@ -201,8 +201,8 @@ object TestAspect extends TimeoutVariants {
   /**
    * An aspect that retries a test until success, without limit.
    */
-  val eventually: TestAspectAtLeastR[ZRTestEnv] = {
-    val eventually = new PerTestPoly {
+  val eventually: TestAspectAtLeastR[ZTestEnv] = {
+    val eventually = new PerTest.Poly {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -283,7 +283,7 @@ object TestAspect extends TimeoutVariants {
    * An aspect that retries a test until success, with a default limit, for use
    * with flaky tests.
    */
-  val flaky: TestAspectAtLeastR[ZRTestEnv] =
+  val flaky: TestAspectAtLeastR[ZTestEnv] =
     flaky(100)
 
   /**
@@ -292,7 +292,7 @@ object TestAspect extends TimeoutVariants {
    */
   def flaky(
     n: Int
-  ): TestAspectAtLeastR[ZRTestEnv] =
+  ): TestAspectAtLeastR[ZTestEnv] =
     retry(Schedule.recurs(n))
 
   /**
@@ -300,7 +300,7 @@ object TestAspect extends TimeoutVariants {
    * satisfies the specified assertion.
    */
   def ifEnv(env: String, assertion: Assertion[String]): TestAspectAtLeastR[Live[System]] =
-    new PerTestAtLeastR[Live[System]] {
+    new PerTest.AtLeast[Live[System]] {
       def perTest[R <: Live[System], E, S](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -338,7 +338,7 @@ object TestAspect extends TimeoutVariants {
     prop: String,
     assertion: Assertion[String]
   ): TestAspectAtLeastR[Live[System]] =
-    new PerTestAtLeastR[Live[System]] {
+    new PerTest.AtLeast[Live[System]] {
       def perTest[R <: Live[System], E, S](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -398,7 +398,7 @@ object TestAspect extends TimeoutVariants {
    * An aspect that repeats the test a default number of times, ensuring it is
    * stable ("non-flaky"). Stops at the first failure.
    */
-  val nonFlaky: TestAspectAtLeastR[ZRTestEnv] =
+  val nonFlaky: TestAspectAtLeastR[ZTestEnv] =
     nonFlaky(100)
 
   /**
@@ -407,8 +407,8 @@ object TestAspect extends TimeoutVariants {
    */
   def nonFlaky(
     n: Int
-  ): TestAspectAtLeastR[ZRTestEnv] = {
-    val nonFlaky = new PerTestPoly {
+  ): TestAspectAtLeastR[ZTestEnv] = {
+    val nonFlaky = new PerTest.Poly {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
@@ -462,20 +462,20 @@ object TestAspect extends TimeoutVariants {
    * An aspect that restores the [[TestSystem]]'s state to its starting state after the test is run.
    * Note that this is only useful when repeating tests.
    */
-  def restoreTestSystem: TestAspectAtLeastR[ZRTestEnv] = restore[TestSystem](_.system)
+  def restoreTestSystem: TestAspectAtLeastR[ZTestEnv] = restore[TestSystem](_.system)
 
   /**
    * An aspect that restores all state in the standard provided test environments
    * ([[TestClock]], [[TestConsole]], [[TestRandom]] and [[TestSystem]]) to their starting state after the test is run.
    * Note that this is only useful when repeating tests.
    */
-  def restoreTestEnvironment: TestAspectAtLeastR[ZRTestEnv] =
+  def restoreTestEnvironment: TestAspectAtLeastR[ZTestEnv] =
     restoreTestClock >>> restoreTestConsole >>> restoreTestRandom >>> restoreTestSystem
 
   /**
    * An aspect that retries failed tests according to a schedule.
    */
-  def retry[R0 <: ZRTestEnv, E0, S0](
+  def retry[R0 <: ZTestEnv, E0, S0](
     schedule: Schedule[R0, TestFailure[E0], S0]
   ): TestAspect[Nothing, R0, Nothing, E0, Nothing, Any] = {
     val retry = new TestAspect.PerTest[Nothing, R0, Nothing, E0, Nothing, Any] {
@@ -510,7 +510,7 @@ object TestAspect extends TimeoutVariants {
    * An aspect that converts ignored tests into test failures.
    */
   val success: TestAspectPoly =
-    new PerTestPoly {
+    new PerTest.Poly {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
@@ -530,7 +530,7 @@ object TestAspect extends TimeoutVariants {
     duration: Duration,
     interruptDuration: Duration = 1.second
   ): TestAspectAtLeastR[Live[Clock]] =
-    new PerTestAtLeastR[Live[Clock]] {
+    new PerTest.AtLeast[Live[Clock]] {
       def perTest[R >: Nothing <: Live[Clock], E >: Nothing <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
@@ -555,6 +555,7 @@ object TestAspect extends TimeoutVariants {
 
   trait PerTest[+LowerR, -UpperR, +LowerE, -UpperE, +LowerS, -UpperS]
       extends TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS] {
+
     def perTest[R >: LowerR <: UpperR, E >: LowerE <: UpperE, S >: LowerS <: UpperS](
       test: ZIO[R, TestFailure[E], TestSuccess[S]]
     ): ZIO[R, TestFailure[E], TestSuccess[S]]
@@ -568,5 +569,19 @@ object TestAspect extends TimeoutVariants {
         case Spec.TestCase(label, test) =>
           Spec.TestCase(label, if (predicate(label)) perTest(test) else test)
       }
+  }
+  object PerTest {
+
+    /**
+     * A `PerTest.AtLeast[R]` is a `TestAspect.PerTest` that that requires at least an R in its environment
+     */
+    type AtLeast[R] =
+      TestAspect.PerTest[Nothing, R, Nothing, Any, Nothing, Any]
+
+    /**
+     * A `PerTest.Poly` is a `TestAspect.PerTest` that is completely polymorphic,
+     * having no requirements ZRTestEnv on error or environment.
+     */
+    type Poly = TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any]
   }
 }
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -90,15 +90,12 @@ trait TestAspect[+LowerR, -UpperR, +LowerE, -UpperE, +LowerS, -UpperS] { self =>
     self >>> that
 }
 object TestAspect extends TimeoutVariants {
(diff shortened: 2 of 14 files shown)


unclebob418:
@adamgfraser added docs and merged in the rest of the clock specs. I think this should be good to go now :)

docs/howto/test_effects.md
```diff
@@ -407,4 +407,46 @@ object MySpec
           @@ timeout(20.nanos) //it's a good idea to compose `eventually` with `timeout`, or the test may never end
       ) @@ timeout(60.seconds)   //apply a timeout to the whole suite
     )
-``` 
\ No newline at end of file
+``` 
+
+#### @@ nonFlaky
```
adamgfraser: Great documentation. Can we add something about how we can't just create extra copies of arbitrary environment types? This does a great job of explaining the semantics but I think risks leaving the impression that the user has to do this just because we didn't do the work on our end to reset things rather than providing a fresh environment being conceptually something we can't do for a polymorphic environment type.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -223,6 +222,30 @@ object TestAspect extends TimeoutVariants {
...
+  def nonFlaky(n0: Int): TestAspectPoly =
+    new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
+      def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
+        test: ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]
+      ): ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]] = {
+
+        def repeat(n: Int): ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]] =
+          if (n <= 1) test
+          else {
+            for {
```
adamgfraser: This can be simplified to `test.fork.flatMap(_.await).flatMap(ZIO.done)`.

unclebob418:
Thanks - I like the fork idea - that could be handy for other cases as well! I'll get started on this tomorrow.

docs/howto/test_effects.md
unclebob418: good idea - wouldn't want to be accused of being lazy. On the topic of documentation, I've sort of rationalised the clock time vs fiber time issue in my head as vectors, with clock time being on the x axis and fibertime on the y axis, and the area above the identity function = sleepytime. I think it might be useful to others in quickly understanding how they interact. Here's a quick sketch of the idea:

image

I haven't seen any graphics in the doco - is there a recommended approach/format/style for this sort of thing?

adamgfraser: Yes that is a great idea. I haven't seen a lot of graphics in this project either. Maybe someone on Discord could suggest a good solution?

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -223,6 +222,46 @@ object TestAspect extends TimeoutVariants {
    */
   val flaky: TestAspectPoly = eventually
 
+  /**
+   * An aspect that runs a test on a forked fiber. This is handy to use with aspects that repeat tests such as
```
adamgfraser: I would say something about how this ensures that each test is run with a fresh set of `FiberRefs`. Normally when people think of `fork` they expect it to introduce additional parallelism but that is not the case here. Then you can mention `fiberTime` as one example of this.

docs/howto/test_effects.md
```diff
@@ -155,7 +155,7 @@ control and determinism. However there is another source of complexity that come
 It is easy to accidentally use different test instances at the same time.
 
 ```scala
-testM("`acquire` doesn't leak permits upon cancellation") {
+testM("THIS TEST WILL FAIL -`acquire` doesn't leak permits upon cancellation") {
```
adamgfraser: This test will hang or not terminate instead of fail?

docs/howto/test_effects.md
```diff
@@ -407,4 +407,49 @@ object MySpec
...
+import zio.duration._
+import zio.test.Assertion.anything
+import zio.test.TestAspect.{after, nonFlaky}
+import zio.test.environment.TestClock.{adjust, setTime}
+import zio.test.{ZIOBaseSpec, assert, suite, testM}
+
+object ClockSpec2
+    extends ZIOBaseSpec(
+      suite("ClockSpec")(
+        testM("THIS TEST WILL FAIL - sleep does not require passage of clock time") {
```
adamgfraser: Won't the test pass but just not really test what we want to test?

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -223,6 +222,46 @@ object TestAspect extends TimeoutVariants {
...
+    new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
+
+      def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
+        test: ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]
+      ): ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]] =
+        test.fork.flatMap(_.await).flatMap(ZIO.done)
+    }
+
+  /**
+   * An aspect that runs each test on a separate fork and resets the TestClock to 0.
```
adamgfraser: Should we generalize to `resetTestEnvironment` and provide fresh copies of all the standard ZIO environment types?

adamgfraser: I just merged #2043 which will provide some additional tools for us to do this in a clean way.

unclebob418:
@adamgfraser revisiting this one now but getting stuck with the usage of mapAll - am I on the right track here?

```
  def resetTestEnvironment: TestAspectPoly =
    fork >>>
      new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
        def perTest[R <: TestEnvironment, E, S](test: ZIO[R, TestFailure[E], TestSuccess[S]]) =
          for {
            testClock   <- TestClock.makeTest(TestClock.DefaultData)
            testConsole <- TestConsole.makeTest(TestConsole.DefaultData)
            testRandom  <- TestRandom.makeTest(TestRandom.DefaultData)
            testSystem  <- TestSystem.makeTest(TestSystem.DefaultData)
            _ <- ZIO
                  .environment[TestEnvironment]
                  .mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
          } yield test
      }
```
compilation error I'm getting is
```
[error] C:\Users\rober\IdeaProjects\zio\test\shared\src\main\scala\zio\test\TestAspect.scala:259:20: value mapAll is not a member of zio.URIO[zio.test.environment.TestEnvironment,zio.test.environment.TestEnvironment]
[error] possible cause: maybe a semicolon is missing before `value mapAll'?
[error]                   .mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
[error]                    ^
[error] one error found
```

adamgfraser:
@unclebob418 Try using `test.provideSome(_.mapAll(???))`.

unclebob418:
Thanks @adamgfraser, looks like I took a step backwards (I had `test.provide` before but that didn't seem correct as it always results in an IO).
Unfortunately it's still not loving life. Just changing the `yield` to `test.provideSome(...)` caused it to return a ZIO in a ZIO. A couple of tweaks later and I think I'm close but still not quite right.
```
 def resetTestEnvironment: TestAspectPoly =
    fork >>>
      new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
        def perTest[R <: TestEnvironment, E, S](
          test: ZIO[R, TestFailure[E], TestSuccess[S]]
        ): ZIO[R, TestFailure[E], TestSuccess[S]] =
          for {
            testClock   <- TestClock.makeTest(TestClock.DefaultData)
            testConsole <- TestConsole.makeTest(TestConsole.DefaultData)
            testRandom  <- TestRandom.makeTest(TestRandom.DefaultData)
            testSystem  <- TestSystem.makeTest(TestSystem.DefaultData)
            pt <- test.provideSome[TestEnvironment](_.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem))
          } yield pt
      }
```

```
[error] C:\Users\rober\IdeaProjects\zio\test\shared\src\main\scala\zio\test\TestAspect.scala:258:61: type mismatch;
[error]  found   : zio.test.environment.TestEnvironment
[error]  required: R
[error]             pt <- test.provideSome[TestEnvironment](_.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem))
[error]                                                             ^
[error] one error found
```

adamgfraser:
@unclebob418 I would start by getting it working with just a normal ZIO effect and then translate that to a test aspect because working with test aspects can be more complicated with their upper and lower bound type parameters. For just a normal effect, you can do it like this:

```scala
import zio._
import zio.test._
import zio.test.environment._

object Example {

  def resetEnvironment[R >: TestEnvironment, E, A](zio: ZIO[R, TestFailure[E], TestSuccess[A]]): ZIO[TestEnvironment, TestFailure[E], TestSuccess[A]] =
    zio.provideSomeM { testEnvironment =>
      for {
        testClock   <- TestClock.makeTest(TestClock.DefaultData)
        testConsole <- TestConsole.makeTest(TestConsole.DefaultData)
        testRandom  <- TestRandom.makeTest(TestRandom.DefaultData)
        testSystem  <- TestSystem.makeTest(TestSystem.DefaultData)
      } yield testEnvironment.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
      test
    }
}
```

Notice that we need the original environment to not require anything more than the test environment because otherwise we couldn't build it, and we will end up depending on having a whole test environment. Now try to translate that to a test aspect with the appropriate type constraints.

unclebob418:
@adamgfraser I'm guessing that last line of `provideSomeM` was supposed to be `zio` to match the argument rather than `test` but even changing that and throwing it in the same package this example doesn't compile. Is this my incompetence shining through or is this not behaving as expected?

```
package zio.test

import zio._
import zio.test._
import zio.test.environment._

object Example {

  def resetEnvironment[R >: TestEnvironment, E, A](zio: ZIO[R, TestFailure[E], TestSuccess[A]]): ZIO[TestEnvironment, TestFailure[E], TestSuccess[A]] =
    zio.provideSomeM { testEnvironment =>
      for {
        testClock   <- TestClock.makeTest(TestClock.DefaultData)
        testConsole <- TestConsole.makeTest(TestConsole.DefaultData)
        testRandom  <- TestRandom.makeTest(TestRandom.DefaultData)
        testSystem  <- TestSystem.makeTest(TestSystem.DefaultData)
      } yield testEnvironment.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
      zio
    }
}
```

```
[error] C:\Users\rober\IdeaProjects\zio\test\shared\src\main\scala\Example.scala:8:24: missing parameter type
[error]     zio.provideSomeM { testEnvironment =>
[error]                        ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

adamgfraser:
@unclebob418 I think I was working off a version from before we had made making the `TestClock` return a `Managed`. Try this on `master`:

```scala
package zio.test

import zio._
import zio.test.environment._

object Example {

  def resetEnvironment[R >: TestEnvironment, E, A](zio: ZIO[R, TestFailure[E], TestSuccess[A]]): ZIO[TestEnvironment, TestFailure[E], TestSuccess[A]] =
    zio.provideSomeManaged {
      for {
        testEnvironment <- ZIO.environment[TestEnvironment].toManaged_
        testClock   <- TestClock.makeTest(TestClock.DefaultData)
        testConsole <- TestConsole.makeTest(TestConsole.DefaultData).toManaged_
        testRandom  <- TestRandom.makeTest(TestRandom.DefaultData).toManaged_
        testSystem  <- TestSystem.makeTest(TestSystem.DefaultData).toManaged_
      } yield testEnvironment.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
    }
}
```

unclebob418:
@adamgfraser still no dice :(

```
[error] C:\Users\rober\IdeaProjects\zio\test\shared\src\main\scala\zio\test\TestAspect.scala:416:25: type mismatch;
[error]  found   : zio.ZIO[zio.test.environment.TestEnvironment,zio.test.TestFailure[E],zio.test.TestSuccess[S]]
[error]  required: zio.ZIO[R,zio.test.TestFailure[E],zio.test.TestSuccess[S]]
[error]         resetEnvironment(test)
[error]                         ^
[info] zio.ZIO[zio.test.environment.TestEnvironment,zio.test.TestFailure[E],zio.test.TestSuccess[S]] <: zio.ZIO[R,zio.test.TestFailure[E],zio.test.TestSuccess[S]]?
[info] false
```

Here's my attempt with some notes on how I arrived at this - can you please let me know what I've got wrong?
image
and the text edition:
```
  def resetEnvironment
  : PerTest[TestEnvironment, Any, Nothing, Any, Nothing, Any] =
    new TestAspect.PerTest[TestEnvironment, Any, Nothing, Any, Nothing, Any] {
      def perTest[R >: TestEnvironment, E, S](test: ZIO[R, TestFailure[E], TestSuccess[S]]): ZIO[R, TestFailure[E], TestSuccess[S]] =
        resetEnvironment(test)
    }
```

adamgfraser:
In this case you need the `TestEnvironment` as both an upper and lower bound. The way test aspects encode universally quantified functions is a little tricky. The thing to keep in mind is that a test aspect always has to take a `ZTest[R, E, A]` and return a `ZTest[R, E, A]` with the same `R`, `E`, and `A`. The type parameters of the test aspect are just constraining what those types can be. Here, we need the `TestEnvironment` to be a lower bound because as you correctly noted if the environment required was anything more than the test environment we couldn't call `mapAll` on it. However, the test environment is also an upper bound because resetting the environment returns an effect with an `R` type of `TestEnvironment`, so it wouldn't be possible for `R` to be any more general type than the test environment.

Notice that this still works okay because `R` is contravariant, so if the actual environment type for an effect is less constrained than `TestEnvironment` the compiler can always narrow it to `TestEnvironment`. See example below.

```scala
import zio.test.environment._

def resetEnvironment0[R >: TestEnvironment, E, A](zio: ZIO[R, E, A]): ZIO[TestEnvironment, E, A] =
  zio.provideSomeManaged {
    for {
      testEnvironment <- ZIO.environment[TestEnvironment].toManaged_
      testClock   <- TestClock.makeTest(TestClock.DefaultData)
      testConsole <- TestConsole.makeTest(TestConsole.DefaultData).toManaged_
      testRandom  <- TestRandom.makeTest(TestRandom.DefaultData).toManaged_
      testSystem  <- TestSystem.makeTest(TestSystem.DefaultData).toManaged_
    } yield testEnvironment.mapAll(_ => testClock, _ => testConsole, _ => testRandom, _ => testSystem)
  }

def resetEnvironment
: PerTest[TestEnvironment, TestEnvironment, Nothing, Any, Nothing, Any] =
  new TestAspect.PerTest[TestEnvironment, TestEnvironment, Nothing, Any, Nothing, Any] {
    def perTest[R >: TestEnvironment <: TestEnvironment, E, S](test: ZIO[R, TestFailure[E], TestSuccess[S]]): ZIO[R, TestFailure[E], TestSuccess[S]] =
      resetEnvironment0(test)
  }

val a: ZSpec[Any, Nothing, String, Unit] = ???
a @@ resetEnvironment
val b: ZSpec[String, Nothing, String, Unit] = ???
b @@ resetEnvironment // does not compile
```

unclebob418:
Thanks for yet another detailed explanation @adamgfraser! very helpful!

jdegoes:
```scala
trait Restorable {
  val save: UIO[UIO[Unit]]
}

object TestClock {
  trait Service extends Restorable {
    ..
  }
  def save: ZIO[TestClock, Nothing, UIO[Unit]] = ZIO.accessM[TestClock](_.clock.save)
}
```

```scala
val resetTestEnv: TestAspec[Nothing, TestEnvironment, Nothing, Any, Nothing, Any] = 
  ???

// Repeat combinators automatically call resetTestEnv!
```

adamgfraser:
Nice! @unclebob418 You good to run with this? I can also jump in if you are backed up.

unclebob418:
I think I should be able to sort this one (hopefully tomorrow)

adamgfraser:
@unclebob418 Awesome!

unclebob418:
sorry for the delay - here's a first attempt (well, at this approach :P ) which appears to be working for `TestClock`/`@@ nonFlaky` :) I think this is a much neater approach - thanks @jdegoes 

At the moment this is saving/restoring `clockState` and `fiberState` only. I don't think we want to mess with the `live`, but should I be including `warningState` in the reset?

@adamgfraser can you please take a look?

still to do:
1. make `console/system/random`  `restorable`
2. add `resetTestEnvironment` and adding to all repeat combinators
2. tests for `console/system/random`
2. removed unused imports -_-

Should I also remove `mapAll` & co? I think they were only introduced to enable this

jdegoes:
@unclebob418 I think you should reset all state for the known test environments. Because otherwise the behavior under repetition is observably different, which breaks encapsulation of the repeating combinators.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -441,50 +480,82 @@ object TestAspect extends TimeoutVariants {
...
+          test.retry(schedule)
+      }
 
   /**
-   * An aspect that executes the members of a suite sequentially.
+   * An aspect that restores the [[TestClock]]'s state to its starting state after the test is run.
+   * Note that this is only useful when repeating tests.
    */
-  val sequential: TestAspectPoly = executionStrategy(ExecutionStrategy.Sequential)
+  def restoreTestClock: TestAspect[Nothing, TestClock, Nothing, Any, Nothing, Any] =
```
adamgfraser: `val` instead of `def` here and below.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -441,50 +480,82 @@ object TestAspect extends TimeoutVariants {
...
   /**
-   * An aspect that applies the specified aspect on Scala 2.
+   * An aspect that restores the [[TestConsole]]'s state to its starting state after the test is run.
+   * Note that this is only useful when repeating tests.
    */
-  def scala2[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS](
-    that: TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS]
-  ): TestAspect[LowerR, UpperR, LowerE, UpperE, LowerS, UpperS] =
-    if (TestVersion.isScala2) that else identity
+  def restoreTestConsole: TestAspect[Nothing, TestConsole, Nothing, Any, Nothing, Any] =
```
adamgfraser: Can we write a `restore` test aspect that works for any `R` that extends `Restorable` and then implement all these methods in terms of that?

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -16,12 +16,11 @@
 
 package zio.test
```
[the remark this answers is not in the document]
unclebob418: I experimented with intellij's reordering on TestConsole which was all out of order - I must've accidentally hit this one as well. Results were excellent... if you don't mind your `val`s and `def`s separated. I've fixed this now.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -113,6 +112,120 @@ object TestAspect extends TimeoutVariants {
...
+  val nonFlaky
+    : TestAspect[Nothing, TestClock with TestConsole with TestRandom with TestSystem, Nothing, Any, Nothing, Any] =
+    nonFlaky(100)
+
+  /**
+   * An aspect that runs the test a default number of times in parallel,
+   * ensuring it is stable ("non-flaky"). Stops at the first failure. This can
+   * be useful for testing effects during periods of high contention.
+   */
+  val nonFlakyPar: TestAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
```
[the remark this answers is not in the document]
unclebob418: If needed, you can very easily create a test that does a bunch of stuff in parallel within one test. You'd be on your own when it comes to managing environments etc. between fibers, but I think that's better than the broken encapsulation. I'm happy to remove this if you'd like.

test/shared/src/main/scala/zio/test/TestAspect.scala
unclebob418: @adamgfraser  this is a great idea but the short answer is "not successfully to date" :) Can you please take a look at my attempt so far and see if you can point me in the right direction?

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -273,14 +273,17 @@ object TestAspect extends TimeoutVariants {
    * An aspect that retries a test until success, with a default limit, for use
    * with flaky tests.
    */
-  val flaky: TestAspectPoly =
```
adamgfraser: There is also `eventually` that we need to use with `restoreTestEnvironment`.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -386,25 +389,39 @@ object TestAspect extends TimeoutVariants {
...
-          else test.flatMap(_ => repeat(n - 1))
-
-        repeat(n)
+  def nonFlaky(
+    n: Int
+  ): TestAspect[Nothing, TestClock with TestConsole with TestRandom with TestSystem, Nothing, Any, Nothing, Any] =
+    restoreTestEnvironment >>>
+      new TestAspect.PerTest[
+        Nothing,
+        TestClock with TestConsole with TestRandom with TestSystem,
```
adamgfraser: You don't need the type constraints here. This test aspect is just repeating a test, it doesn't have to know anything about the test environment. It may also make this read a little better to pull it out into a separate definition.

```scala
def nonFlaky(
    n: Int
  ): TestAspect[Nothing, TestClock with TestConsole with TestRandom with TestSystem, Nothing, Any, Nothing, Any] = {
  val nonFlaky = ???
  restoreTestEnvironment >>> nonFlaky
}
```

Same thing could probably help with the other cases where we have to define these inner test aspects.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -439,18 +456,67 @@ object TestAspect extends TimeoutVariants {
...
-   * An aspect that retries failed tests according to a schedule.
+   * An aspect that restores a given [[Restorable]]'s state to its starting state after the test is run.
+   * Note that this is only useful when repeating tests.
    */
-  def retry[R0, E0, S0](
-    schedule: Schedule[R0, TestFailure[E0], S0]
-  ): TestAspect[Nothing, R0, Nothing, E0, Nothing, Any] =
-    new TestAspect.PerTest[Nothing, R0, Nothing, E0, Nothing, Any] {
-      def perTest[R >: Nothing <: R0, E >: Nothing <: E0, S >: Nothing <: Any](
+  def restore[R0 <: Restorable](r: R0): TestAspect[Nothing, R0, Nothing, Any, Nothing, Any] =
```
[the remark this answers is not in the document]
unclebob418: @adamgfraser I did initially look at `around` but for the reasons you mentioned moved to `r.flatmap(test <* _)`. I do like the solution for generalising `around` though - it'll be much more useful to everyone like that.

RE the service being `Restorable` and not the e.g. `TestClock`,  `save` has the same signature on the service and the helper methods - I thought I should be able to do something like this:
```scala
trait TestClock extends Clock with Scheduler with Restorable {
   ???
}

def restore[R0 <: Restorable](service:R0): TestAspect[Nothing, R0, Nothing, Any, Nothing, Any] = ???

val restoreTestClock: TestAspect[Nothing, TestClock, Nothing, Any, Nothing, Any] = restore(TestClock)
```
but the compiler said no:
```
[error] C:\Users\rober\IdeaProjects\zio\test\shared\src\main\scala\zio\test\TestAspect.scala:447:93: type mismatch;
[error]  found   : zio.test.TestAspect[Nothing,R0,Nothing,Any,Nothing,Any]
[error]  required: zio.test.TestAspect[Nothing,zio.test.environment.TestClock,Nothing,Any,Nothing,Any]
[error]   val restoreTestClock: TestAspect[Nothing, TestClock, Nothing, Any, Nothing, Any] = restore(TestClock)
```
is there a fundamental reason this can't work that I'm not understanding?

your approach of `service: R0 => Restorable` does work but doesn't feel right to me when we have helper methods on e.g. `TestClock` which aren't even used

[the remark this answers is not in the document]
adamgfraser: The issue is with the distinction between the companion object and the trait, and the service.  In your snippet `restore` is looking for a value that extends `Restorable` but neither the `TestClock` trait or the companion object `TestClock` implement `Restorable`.

I don't think you can implement `Restorable` at the level of the companion object because you need to get the environment to access the `TestClock` you want to restore so you can't satisfy the type signature of `Restorable`. I think you could potentially make `TestClock` the trait extend `Restorable` (the implementation would just call the `restore` method on the service). But I don't think it would be a good idea because when you mixed together different traits (e.g. `TestClock with TestConsole` you would create issues of which implementation of `Restorable` to use.

test/shared/src/main/scala/zio/test/TestAspect.scala
unclebob418: To do this I had to change restoreTestEnvironment to be lazy val due to the ordering (warning about referencing an uninitialized val) but I don't think this should be an issue

test/shared/src/main/scala/zio/test/TestAspect.scala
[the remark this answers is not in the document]
unclebob418: > But I don't think it would be a good idea because when you mixed together different traits (e.g. TestClock with TestConsole you would create issues of which implementation of Restorable to use.

I didn't think of this - I'll  leave it how it is

unclebob418:
@adamgfraser  unfortunately it looks like I've managed to break the test system :/

I've tried various changes to build.sbt based on the error message (e.g. added  `.settings(Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat)` to the `testTests` project) but I don't really know what I'm doing here and if that may break other things. Do you have any suggestions?

```
[error] Test suite zio.test.environment.EnvironmentSpec failed with java.lang.NoClassDefFoundError: Could not initialize class zio.test.TestAspect$.
[error] This may be due to the ClassLoaderLayeringStrategy (ScalaLibrary) used by your task.
[error] To improve performance and reduce memory, sbt attempts to cache the class loaders used to load the project dependencies.
[error] The project class files are loaded in a separate class loader that is created for each test run.
[error] The test class loader accesses the project dependency classes using the cached project dependency classloader.
[error] With this approach, class loading may fail under the following conditions:
[error]
[error]  * Dependencies use reflection to access classes in your project's classpath.
[error]    Java serialization/deserialization may cause this.
[error]  * An open package is accessed across layers. If the project's classes access or extend
[error]    jvm package private classes defined in a project dependency, it may cause an IllegalAccessError
[error]    because the jvm enforces package private at the classloader level.
[error]
[error] These issues, along with others that were not enumerated above, may be resolved by changing the class loader layering strategy.
[error] The Flat and ScalaLibrary strategies bundle the full project classpath in the same class loader.
[error] To use one of these strategies, set the  ClassLoaderLayeringStrategy key
[error] in your configuration, for example:
[error]
[error] set testTestsJVM / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.ScalaLibrary
[error] set testTestsJVM / Test / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat
[error]
[error] See ClassLoaderLayeringStrategy.scala for the full list of options.
```

adamgfraser:
@unclebob418 I think it is another early initialization error. Try changing all the restore test aspects to `def`.

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Yes, I remember running into that too when I was playing with it. No problem. I would just change to `def` since they are not computationally expensive and will probably only be invoked once if at all in a typical application.

unclebob418:
@adamgfraser much cleaner now :) i reckon this is probably pretty close to a merge

test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
```diff
@@ -100,6 +100,7 @@ case class TestEnvironment(
         def currentTime(unit: TimeUnit): UIO[Long]           = live.provide(zio.clock.currentTime(unit))
         val fiberTime: UIO[Duration]                         = UIO.succeed(Duration.Zero)
         val nanoTime: UIO[Long]                              = live.provide(zio.clock.nanoTime)
+        override val save: UIO[UIO[Unit]]                    = UIO(UIO.unit)
```
adamgfraser: `UIO.succeed(UIO.unit)`. Also I don't think you need the `override` here.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -127,13 +127,24 @@ object TestAspect extends TimeoutVariants {
   /**
    * Constructs an aspect that evaluates every test inside the context of a `Managed`.
    */
-  def around[R0, E0](before: ZIO[R0, E0, Any], after: ZIO[R0, Nothing, Any]) =
+  def around[R0, E0](
```
adamgfraser: Let's call this `around_` to mirror `bracket_` and describe them similarly.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -193,13 +204,16 @@ object TestAspect extends TimeoutVariants {
   /**
    * An aspect that retries a test until success, without limit.
    */
-  val eventually: TestAspectPoly =
-    new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
+  val eventually
+    : TestAspect[Nothing, TestClock with TestConsole with TestRandom with TestSystem, Nothing, Any, Nothing, Any] = {
```
adamgfraser: I am wondering if we just want to constrain it to `TestEnvironment` here and in the other combinators. It is adding a little bit of an additional constraint but it would make these type signatures less fiercesome and I don't really see these combinators being used with having implementations of all the test services and not the test environment. Also, conceptually we are trying to reset the entire test environment (e.g. if there was ever anything else added to the test environment that maintained state we would reset that too).

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -127,13 +127,24 @@ object TestAspect extends TimeoutVariants {
...
+   * be used in `after`.
+   */
+  def around[R0, E0, A0](
+    before: ZIO[R0, E0, A0]
+  )(after: A0 => ZIO[R0, Nothing, Any]): PerTest[Nothing, R0, E0, Any, Nothing, Any] =
     new TestAspect.PerTest[Nothing, R0, E0, Any, Nothing, Any] {
       def perTest[R >: Nothing <: R0, E >: E0 <: Any, S >: Nothing <: Any](
         test: ZIO[R, TestFailure[E], TestSuccess[S]]
       ): ZIO[R, TestFailure[E], TestSuccess[S]] =
         ZManaged
```
adamgfraser: I think we can use `ZIO#bracket` here instead of `ZManaged`. Since we're just using the resource and not composing it with other resources there is no need for the overhead of a `ZManaged`.

test/shared/src/main/scala/zio/test/TestAspect.scala
jdegoes: Would it make sense to introduce something like:

```scala
type TestAspectTE = TestAspect[Nothing, TestClock with TestConsole with TestRandom with TestSystem, Nothing, Any, Nothing, Any]
```

to describe "at least" a test environment?

unclebob418: I just changed it to use
```scala
type ResTestEnv = TestClock with TestConsole with TestRandom with TestSystem
def restoreTestEnvironment: TestAspect[Nothing, ResTestEnv, Nothing, Any, Nothing, Any] = ???
```
but I think your solution is cleaner

adamgfraser: @jdegoes  Yes, we could definitely do that. We could also do something like:

```scala
type ZTestEnv = TestClock with TestConsole with TestRandom with TestSystem
```

jdegoes: Ah, I like that... probably it's already repeated in several places.

Maybe even make `TestAspectAtLeastR` type alias so we can plug in `ZTestEnv` and other things in other cases.

adamgfraser: Yes that is a good idea. We don't need to do it right now but we could also think of some other aliases like that. `TestAspectPoly` is really nice right now but if the test aspect isn't fully polymorphic you go straight from that to six type parameters.

unclebob418: image

```scala
  type ZTestEnv = TestClock with TestConsole with TestRandom with TestSystem
  type TestAspectTE =
    TestAspect[Nothing, ZTestEnv, Nothing, Any, Nothing, Any]
```

then we get
```scala
def retry[R0 <: ZTestEnv, E0, S0](schedule: Schedule[R0, TestFailure[E0], S0]) = ???
def restoreTestEnvironment: TestAspectTE = ???
```

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -90,6 +90,9 @@ trait TestAspect[+LowerR, -UpperR, +LowerE, -UpperE, +LowerS, -UpperS] { self =>
     self >>> that
 }
 object TestAspect extends TimeoutVariants {
+  type ZTestEnv = TestClock with TestConsole with TestRandom with TestSystem
```
adamgfraser: Let's move these to the package object with the other type aliases.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -450,39 +450,38 @@ object TestAspect extends TimeoutVariants {
    * An aspect that restores the [[TestClock]]'s state to its starting state after the test is run.
    * Note that this is only useful when repeating tests.
    */
-  def restoreTestClock = restore[TestClock](_.clock)
+  def restoreTestClock: TestAspect[Nothing, TestClock, Nothing, Any, Nothing, Any] = restore[TestClock](_.clock)
```
adamgfraser: We may want to use the `TestAspectAtLeastR` alias here. I think that could help us clean up some of these other type signatures as well.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -125,16 +128,8 @@ object TestAspect extends TimeoutVariants {
...
-   */
-  def around[R0, E0](
-    before: ZIO[R0, E0, Any],
-    after: ZIO[R0, Nothing, Any]
-  ): PerTest[Nothing, R0, E0, Any, Nothing, Any] = around(before)(_ => after)
-
-  /**
-   * Constructs an aspect that evaluates every test inside the context of a `Managed` where the result of `before` can
-   * be used in `after`.
+   * Constructs an aspect that evaluates every test is evaluated between two effects, `before` and `after`,
```
adamgfraser: Constructs an aspect that evaluates every test between two effects, `before` and `after`, where the result of `before` can be used in `after`.

test/shared/src/main/scala/zio/test/TestAspect.scala
unclebob418: I missed that comment about `TestAspectAtLeastR` before. I've done a couple of the other obvious ones while I was at it.

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -43,8 +44,34 @@ import zio.stream.{ ZSink, ZStream }
...
  */
 package object test extends CompileVariants {
+
   type AssertResult = BoolAlgebraM[Any, Nothing, AssertionValue]
-  type TestResult   = BoolAlgebraM[Any, Nothing, FailureDetails]
+
+  /**
+   * A `PerTestAtLeastR[R]` is a `TestAspect.PerTest` that that requires at least an R in its environment
+   */
+  type PerTestAtLeastR[R] =
```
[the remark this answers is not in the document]
unclebob418: how about my last commit? It's only used in construction of test aspects; all return types are are still `TestAspect`s, but saves a lot of `[Nothing, Any, Nothing, Any, Nothing, Any]` clogging up the file and locates it with the `PerTest` trait. I think it reads much nicer like this rather than ditching them altogether.

unclebob418:
Excellent! Thanks @adamgfraser and @jdegoes for your help on this one :)

jdegoes:
@unclebob418 Thank you for your patience and perseverance, and this excellent improvement!

adamgfraser:
@unclebob418 Thank you!

PR #2031, opened 2019-10-25
Enable fatal warnings for all compile tasks

reibitto:
`fatal.warnings` is currently only set for Dotty for some reason. This PR enables it for all of them.

We could also flip the logic so the default is true to prevent issues like this creeping up in the future. Slightly more work for the person who wants to turn it off locally, but safer against refactorings. Though if we do that, I think I would prefer an env var fallback (easier to enable in your exports).

There was also John's concern here:
https://github.com/zio/zio/pull/1924#issuecomment-539113881

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ testJVM: &testJVM
     - <<: *install_jdk
     - run:
         name: Run tests
-        command: ./sbt ++${SCALA_VERSION}! testJVM
+        command: ./sbt -Dfatal.warnings=true ++${SCALA_VERSION}! testJVM
     - <<: *clean_cache
     - <<: *save_cache
     - store_test_results:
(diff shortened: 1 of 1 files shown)


jdegoes:
What are the implications of this change?

adamgfraser:
This just restores what I think was our understanding of the status quo behavior which is that if there are any compiler warnings CI should fail. Due to a regression introduced when we changed the default for fatal warnings locally this is not currently happening so we have seen a couple of situations where PRs were making it in to `master` with compiler warnings for things like unused imports.

My recommendation would be that we merge this immediately and then we can revisit your earlier point of whether we should go back to the previous behavior of having fatal warnings enabled locally as well.

reibitto:
We want `-Xfatal-warnings` to always be on for CI. Making them not fatal was only intended for local development. Currently only the Dotty build on CI has them set to fatal.

jdegoes:
> Making them not fatal was only intended for local development. 

We need to fix that, SoftwareMill shouldn't be paying for these things that are easily caught by the developer.

reibitto:
Oh, are there additional costs? It's not a fixed monthly bill? I wasn't aware of that. In that case, yeah, it would make sense to revert it.

Even if it was fixed, you still would want to take precautions as to not queue up unnecessary builds. Which is why I wasn't particularly a fan of defaulting it to off. Since new contributors wouldn't likely know about this (or knowing to run `fmt` which I think we could catch earlier than we do now).

Issue #1987, opened 2019-10-16
Flaky ZIO test


regiskuckaertz:
[check this out](https://app.circleci.com/jobs/github/zio/zio/33869):
```
ZIO Test
[info] - ZIO
[info]   - raceAll
[info]     - returns last failure
[info]       25 did not satisfy equalTo(24)
```

LGLO:
I was thinking a lot about this test some time ago. My only conclusion was that we can increase delay only.
This is because thread can be paused after passing some barrier/latch/whatever we add in test and before registering itself as a looser.

regiskuckaertz:
Also done.

Issue #1882, opened 2019-10-02
Flaky BlockingSpec Test


adamgfraser:
https://circleci.com/gh/zio/zio/30419?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

regiskuckaertz:
ha, got this one too
https://app.circleci.com/jobs/github/zio/zio/34311

hearnadam:
No longer blocking CI

PR #2037, opened 2019-10-26
WIP migrate RandomSpec

unclebob418:
#1898 
@adamgfraser can you please review? I'm reasonably happy with most of it except 

1. I couldn't figure out how to use the environment's TestRandom. using zio.random resulted in compilation errors. I've ended up leaving the `TestRandom.makeTest(DefaultData)` lines in there.

1. I think some of the assertions could be rewritten to provide better errors, but I couldn't quite get it to work. e.g. for `forAllBounded` I tried to turn 
`yield assert(actual.zip(bounds).forall { case (a, n) => zero <= a && a < n }, equalTo(true))`
into 
`yield assert(actual.zip(bounds), forall(case (a, n) => assert(a, isWithin(zero, n))))`
but the compiler said no
I think I'm getting my `Assertion`s and my `TestResults` mixed up here, but not quite sure how to nest the assertions


test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -0,0 +1,168 @@
+package zio.test.environment
+
+import zio.test.Assertion._
+import zio.test.environment.RandomSpecUtil._
+import zio.test.environment.TestRandom.DefaultData
+import zio.test.{ assert, suite, test, testM, TestResult, ZIOBaseSpec }
+import zio._
+
+object RandomSpec
+    extends ZIOBaseSpec(
```
jdegoes: You should be able to do:

```scala
object RandomSpec extends ZIOBaseSpec {
  // Helpers here

  suite(
    // Tests here
  )
}

Comment (unclebob418) [reply test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala]: Thanks @jdegoes, that's actually what I tried to do in my first PR but unfortunately it doesn't compile
```
[info] Compiling 1 Scala source to C:\Users\rober\IdeaProjects\zio\test-tests\jvm\target\scala-2.12\test-classes ...
[error] C:\Users\rober\IdeaProjects\zio\test-tests\shared\src\test\scala\zio\test\environment\RandomSpec.scala:10:27: not enough arguments for constructor ZIOBaseSpec: (spec: => zio.test.ZSpec[zio.test.environment.TestEnvironment,Any,String,Any])zio.test.ZIOBaseSpec.
[error] Unspecified value parameter spec.
[error] object RandomSpec extends ZIOBaseSpec {
[error]                           ^
```
this is the approach that @adamgfraser suggested I take here https://github.com/zio/zio/pull/1959#discussion_r334951983
and also the approach shown here https://zio.dev/docs/usecases/usecases_testing
but if we can change `DefaultRunnableSpec` so this works that'd be excellent

I think the issue is with `RunnableSpec` - the class is abstract so doesn't need to define `spec` but I suspect it's been done so you don't have `override def spec` scattered throughout your tests. Perhaps @adamgfraser can shed some light on the background? 
```
abstract class RunnableSpec[R, L, T, E, S](runner0: TestRunner[R, L, T, E, S])(spec0: => Spec[R, E, L, T])
    extends AbstractRunnableSpec {
  ...
  override def spec   = spec0
```

Comment (adamgfraser) [reply test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala]: To use John's approach you need one more set of parentheses.

```scala
package zio.test.environment

import zio.test._
import zio.test.Assertion._

object RandomSpec
    extends ZIOBaseSpec({

      // Helpers here
      def plus(x: Int, y: Int): Int =
        x + y

      suite("RandomSpec")(
        // Tests here
        test("addition works fine") {
          assert(plus(1, 1), equalTo(2))
        }
      )
    })
```

You need the parentheses after `ZIOBaseSpec` (or `DefaultRunnableSpec`) to tell the compiler the code that follows is providing the required parameter to `ZIOBaseSpec` rather than the body of `RandomSpec` and then you need the brackets to be able to provide multiple statements. We did this to enable the user to not have to override anything but just be able to write tests in a declarative style.

```scala
objectMySpec extends DefaultRunnableSpec(
  suite("MySpec")(
    test("myTest")
```

It seems like so far users have preferred the separate object with helper methods style but both are fine.

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+        testM("check nextInt")(forAllEqual(_.nextInt)(_.nextInt())),
+        testM("check nextLong")(forAllEqual(_.nextLong)(_.nextLong())),
+        testM("check nextString")(forAllEqualN(_.nextString(_))(_.nextString(_))),
+        testM("check nextPrintableChar")(forAllEqual(_.nextPrintableChar)(_.nextPrintableChar())),
+        testM("bounded nextInt")(forAllEqualN(_.nextInt(_))(_.nextInt(_))),
+        testM("bounded nextInt generates values within the bounds")(forAllBounded(_.nextInt)(_.nextInt(_))),
+        testM("bounded nextLong generates values within the bounds")(forAllBounded(_.nextLong)(_.nextLong(_))),
+        testM("shuffle")(forAllEqualShuffle(_.shuffle(_))(_.shuffle(_))),
+        test("referential transparency") {
+          val rt   = new DefaultRuntime {}
```
adamgfraser: You can use `ZIO.runtime` to access the runtime instead of creating your own.

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+    val num = implicitly[Numeric[A]]
+    import num._
+    val seed    = SRandom.nextLong()
+    val sRandom = new SRandom(seed)
+    for {
+      testRandom <- TestRandom.makeTest(DefaultData)
+      _          <- testRandom.setSeed(seed)
+      bounds     = List.fill(100)(num.abs(bound(sRandom)) max one)
+      actual     <- UIO.foreach(bounds)(f(testRandom, _))
+    } yield assert(actual.zip(bounds).forall { case (a, n) => zero <= a && a < n }, equalTo(true))
```
adamgfraser: If you want to have a more specific assertion here, you are on the right track with `assert(actual.zip(bounds), forall(...)`. The challenge is that `within` is not going to work as is because `isWithin` takes two parameters for the bounds and is expecting an `A`, but you need an `Assertion[(A, B)]`. You need a combinator that, given an `Assertion[A]` and a way to convert an `(A, B)` to an `A` and an `Assertion[A]`, gives you an `Assertion[(A, B)]`. The closest thing we have to this is `AssertionRec` but it is not quite powerful enough because it does not let you generate the assertion dynamically.

I will think about this for the future but I think this is fine for now.

unclebob418:
Another random failure :/ I don't think that's anything I've changed

[info] ZIO Test
[info] - BlockingSpec
[info]   - Make a Blocking Service and verify that
[info]     - effectBlocking can be interrupted
[info] Timeout of 1 m exceeded. Couldn't interrupt test within 1 s, possible resource leak!

doesn't look like anything too complicated!

```
testM("effectBlocking can be interrupted") {
            assertM(effectBlocking(Thread.sleep(50000)).timeout(Duration.Zero), isNone)
          },
```

adamgfraser:
Definitely not you. We have an open issue for that in #1882. Can you give that test a shorter timeout and mark it as flaky?

I had one other thought. We should change the helper methods to be property based tests using `Gen`. That will help us clean up some of the boilerplate and also make the assertions work better in situations like the one you asked about because we will only be making an assertion about a single value instead of a collection of values.

unclebob418:
Good point. I'll have to learn how to do that now! Really enjoying working this issue because of how many different aspects (pun intended) it covers :)

unclebob418:
I've done the shorter timeout and marked it as flaky. Is it OK to merge as is? I'm going to be busy with work for probably the next week so probably won't have time to swap it all to `Gen` in the immediate future.

unclebob418:
This isn't my understanding of how @@ flaky works - shouldn't it repeat until it succeeds (or at least until it hits the default 60s timeout)?
```
[info] ZIO Test
[info] - BlockingSpec
[info]   - Make a Blocking Service and verify that
[info]     - effectBlocking can be interrupted
[info] Timeout of 1 ms exceeded.
```

```
 testM("effectBlocking can be interrupted") {
            assertM(effectBlocking(Thread.sleep(50000)).timeout(Duration.Zero), isNone)
          } @@ timeout(1.millis) @@ flaky, //todo fix in #1882 - shouldn't be flaky or require timeout
```

jdegoes:
@unclebob418 Sure, but `timeout(1.minute) @@ flaky` will give the test a LOT of time between retries, whereas `timeout(1.millis) @@ flaky` will minimize wasted time. They'll both always succeed (assuming the test sometimes passes), but the first one will waste a lot of time.

unclebob418:
@jdegoes my last commit added the `@@ timeout(1.millis) @@ flaky`; previously it was just running once and occasionally failing. My issue is that I expected the error from CI tests (if there was one) to say that the timeout of 60s from the suite had failed, meaning it tried thousands of times and never succeeded (that's actually another issue - if we're not able to interrupt the sleeping thousands of threads that's not ideal :/). It seems that @@ flaky hasn't restarted the test when it should have and the 1ms timeout failure has been allowed through to the keeper, which should never happen unless I'm misunderstanding something. Is this a bug with @@ flaky or am I missing something?

jdegoes:
@unclebob418 Hmm, `@@ flaky` may not be updated to use the new model in which failures of the test are _not_ reflected in the failure of the test effect — /cc @adamgfraser 

In any case, `@@ flaky` with a 1 minute timeout would take a really long time to make any progress...

unclebob418:
One thought on this before I go, (without really understanding how the whole fiber system works other than potentially many fibers share a thread and can jump between them) - what happens if the fiber that's running the tests happens to share the fiber that executes the `Thread.sleep(5000)`? Or are these guaranteed to be from separate thread pools?

adamgfraser:
I just pushed #2045 to fix the issue with `flaky`. Wit that merged we should be able to do `timeout(1.millis) @@ flaky` here.

unclebob418:
@adamgfraser same again :/ I just discovered another issue

the tests were always passing on my computer, so I thought I'd try to get it to fail. 

and I did probably take this a bit far with ` @@ timeout(1.millis)  @@ nonFlaky(1000000)` and then another `nonFlaky(1000)` on the spec...

but it still passed!
```
[info] Passed: Total 713, Failed 0, Errors 0, Passed 713, Ignored 2
[error] There were linking errors
[error] (streamsTestsJS / Test / fastOptJS) There were linking errors
[error] (coreTestsJVM / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 61 s (01:01), completed 28/10/2019 11:48:47 PM
[IJ]sbt:zio>
```
trick is, it didn't really. It's been failing the whole time but it's not getting reported.
```
about 3k lines down?
...
[info] + RefMSpec
[info]   + get
[info]   + set
[info]   + update
[info]   + update with failure
[info]     + effectBlocking completes successfully
[info]     + effectBlockingCancelable completes successfully
[info]     - effectBlocking can be interrupted
[info] Timeout of 1 ms exceeded.
[info]     - effectBlockingCancelable can be interrupted
[info]       Some(()) did not satisfy isNone()
[info] Ran 4 tests in 4 s 790 ms: 2 succeeded, 0 ignored, 2 failed
...
```
that's not even the right spec for it to be running in!

side note, how do I just run one test or spec?
the welcome message method doesn't seem to work for me
```
coreTestsJVM/testOnly *.ZIOSpec -- -t "effectBlocking can be interrupted"
[info] + ZIO
[info] Ran 0 tests in 575 ms: 0 succeeded, 0 ignored, 0 failed
[info] ZIO Test
[info] Done
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[success] Total time: 3 s, completed 28/10/2019 11:58:08 PM
```

unclebob418:
edit: maybe I've just been reading this wrong - it does say some tests failed, but I've been looking at this line: `[info] Passed: Total 713, Failed 0, Errors 0, Passed 713, Ignored 2`

adamgfraser:
@unclebob418 You need to use the correct file name for the test. So in this case you would do:

```
project coreTestsJVM`
testOnly zio.blocking.BlockingSpec -- -t "effectBlocking can be interrupted"
```

adamgfraser:
I spent some time looking at this.

First, I think we need to use a timeout of longer than 1 millisecond. 1 millisecond does not seem to give enough time for the test to complete. 10 milliseconds seemed to work well for me locally. That is probably enough for us to merge this PR.

Second, I think there definitely is some kind of resource leak here. When I used the longer timeout with `flaky` the test passed but when I ran that for 100 iterations I observed unusual slowdowns. Given the simplicity of the test I am worried there is something going on where the blocking effect is leaking into the main ZIO thread pool.

unclebob418:
@adamgfraser  I can reproduce the slowdown - I used `@@ timeout(10.millis) @@ flaky @@ nonFlaky(100)` and all subsequent tests took way longer than usual. it's passing now though (failure on CI is due to #1987)
```

Some thoughts that may help - When the test fails, the timeout aspect is failing not with the expected `timeoutFailure`, but with `interruptionTimeoutFailure` (after the default 1 second). I think this means it's unable to kill off the running tests with their `sleep(50000)` calls and then continues with the retry from `@@ flaky` regardless. Looks like `Blocking#interruptThread` is running a while loop repetitively trying to interrupt these threads; if it can't for some reason, we're potentially running 100 loops in parallel for ~50 seconds. There does appear to be a 2ms backoff on each attempt (up to 50ms) which limits the impact a bit, I guess the question is does it limit the impact enough? 

BTW is this backoff supposed to be exponential? the variable names `base` and `n` imply that to me but they're just multiplied.

adamgfraser:
@unclebob418 Thanks! That's exactly right regarding the `interruptionTimeoutFailure`. One of the tricky things about this test is that it is testing that we can interrupt the `effectBlocking`, so if the test fails and the effect is uninterruptible then the test framework isn't going to be able to do anything to interrupt it either. The best it can do is report that there is a possible resource leak, which it is doing.

I don't know the answer to your question regarding `effectBlocking`. I haven't worked with that part of the codebase before so I can't be as helpful there as I can be in some other areas. Maybe when you free up if we haven't already fixed it we can work on fixing it, but for now I think it is outside of this PR. I am going to go ahead and merge this and we can convert those tests to use `Gen` in a follow up PR.

PR #2038, opened 2019-10-26
Make Fiber Composition Parallel

### LINKED ISSUE #2033: Fiber composition should be parallel
Fiber composition operators should _always_ compose in parallel.

The reason is that if in a composite fiber, one fiber fails on some operation like `join` or `await`, then all must fail; rather than waiting for the other members of the composite to complete, it makes more sense to terminate them immediately like other parallel operators.

Mostly, this will be a matter of replacing `zipWith` in `Fiber.scala` with `zipWithPar`.

The following code:

```scala
  def shard[R, E, A](queue: Queue[A], n: Int, worker: A => ZIO[R, E, Unit]): ZIO[R, E, Nothing] = {
    val worker1 = queue.take.flatMap(a => worker(a).uninterruptible).forever

    ZIO.forkAll(List.fill(n)(worker1)).flatMap(_.join) *> ZIO.never
  }
```

should exhibit improved behavior: namely, if one worker fails, then the others will be terminated, rather than running until they fail.

We should put this or other simpler examples in a test to demonstrate correctness of fiber composition.

adamgfraser:
Resolves #2033.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/FiberSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberSpec.scala
@@ -50,7 +50,7 @@ object FiberSpec
               child1    <- (fiberRef.set("child1") *> semaphore.release).fork
               child2    <- (fiberRef.set("child2") *> semaphore.release).fork
               _         <- semaphore.acquireN(2)
-              _         <- child1.zip(child2).inheritFiberRefs
+              _         <- child1.zipPar(child2).inheritFiberRefs
               value     <- fiberRef.get
             } yield assert(value, equalTo("child1"))
           }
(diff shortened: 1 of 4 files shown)


adamgfraser:
So I think the CI failure is actually because a test is asserting a property we can no longer guarantee. The flaky test is:

```scala
testM("allFailures") {
  for {
    f1     <- IO.fail("error1").fork
    f2     <- IO.fail("error2").fork
    errors <- f1.zip(f2).join.parallelErrors[String].flip
  } yield assert(errors, equalTo(List("error1", "error2")))
}
```

Basically, if there are two failures in a `f1.zip(f2)` the `Cause` resulting from their failure will contain both errors, preserving the order.

Looking at the implementation of `ZIO#zipWithPar` that `Fiber#zip` is implemented in terms of, there are two problems with this. First, notice that we aren't flipping the order of the causes if the right side fails first so `zipParWith` is returning a `Cause` with the first fiber failure on the left hand side, versus the original failure of the original left hand fiber on the left hand side. I'm not sure if this is intentional or not. The more serious problem is that in our new parallel implementation one side could fail and interrupt the other side before it completes or potentially before it even starts executing, so we aren't actually guaranteed to get both errors at all.

```scala
  final def zipWithPar[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] = {
    def coordinate[A, B](f: (A, B) => C)(winner: Exit[E1, A], loser: Fiber[E1, B]): ZIO[R1, E1, C] =
      winner match {
        case Exit.Success(a) => loser.join.map(f(a, _))
        case Exit.Failure(cause) =>
          loser.interrupt.flatMap {
            case Exit.Success(_)          => ZIO.halt(cause)
            case Exit.Failure(loserCause) => ZIO.halt(cause && loserCause)
          }
      }
    val g = (b: B, a: A) => f(a, b)
    (self raceWith that)(coordinate(f), coordinate(g))
  }
```

jdegoes:
1. We should preserve the order.
2. Due to non-determinism, we can't guarantee both errors anymore. Our test could account for that by allowing all possibilities.

adamgfraser:
All set.

jdegoes:
🎉🕺🙏

PR #2039, opened 2019-10-27
Add "run ci" trigger logic

reibitto:
@jdegoes Feel free to close this if you don't like the idea, but I wanted to give it a test run to see how the idea fairs in practice.

This attempts to solve a few problems:
- Circle CI seems to have a usage-based pricing scheme now. Meaning minimizing unnecessary work is beneficial in terms of cost savings (but probably also wait time if many builds are queued up).
- We also have a lot of unnecessary CI runs due to things like WIP commits and forgetting to run `fmt` manually. Then there's also the the `-Xfatal-warnings` being off by default issue when working locally.

What this PR does is:
- CI does not run by default (well, except the prerequites like `checkout` and so on). `[run ci]` is needed to trigger it (basically the opposite of [Circle CI's built-in [skip ci]](https://circleci.com/docs/2.0/skip-build/)).
- Makes `-Xfatal-warnings` on by default (but can still be overridden if it's getting in the way of your productivity).
- Added an SBT task called `ci` that adds the `[run ci]` message. But before doing that, it does the scalafmt check, as well as ensuring you haven't overridden `-Xfatal-warnings`.

My concern was that running the `ci` command is an extra step, but it might not be that big of a deal considering `fmt` is also a manual step currently. I don't know... whether this is a deal-breaker is up to you all. I can see arguments for either side.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,9 +93,19 @@ install_nodejs: &install_nodejs
         nvm use
         node -v
 
+check_should_run: &check_should_run
+  - run: |
+      if git log -1 --format=oneline $CIRCLE_SHA1 | grep --ignore-case '\[run ci\]'; then
+        echo 'CI set to run'
+      else
+        echo 'Early-exiting since CI is not set to run (no "[run ci]" in commit message). Use the "ci" SBT task.'
+        exit 1
+      fi
+
 lint: &lint
   steps:
     - checkout
+    - check_should_run
     - <<: *load_cache
     - run:
         name: Lint code
@@ -122,7 +132,7 @@ testJVM: &testJVM
     - <<: *install_jdk
     - run:
         name: Run tests
-        command: ./sbt -Dfatal.warnings=true ++${SCALA_VERSION}! testJVM
+        command: ./sbt ++${SCALA_VERSION}! testJVM
     - <<: *clean_cache
     - <<: *save_cache
     - store_test_results:
@@ -134,7 +144,7 @@ testJVMDotty: &testJVMDotty
     - <<: *load_cache
     - run:
         name: Run tests
-        command: ./sbt -Dfatal.warnings=true ++${SCALA_VERSION}! testJVMDotty
+        command: ./sbt ++${SCALA_VERSION}! testJVMDotty
     - <<: *clean_cache
     - <<: *save_cache
 
@@ -146,7 +156,7 @@ testJS: &testJS
     - <<: *install_nodejs
     - run:
         name: Run tests
-        command: ./sbt -Dfatal.warnings=true ++${SCALA_VERSION}! testJS
+        command: ./sbt ++${SCALA_VERSION}! testJS
     - <<: *clean_cache
     - <<: *save_cache
     - store_test_results:
--- a/build.sbt
+++ b/build.sbt
@@ -31,6 +31,7 @@ ThisBuild / publishTo := sonatypePublishToBundle.value
 
 addCommandAlias("fmt", "all root/scalafmtSbt root/scalafmtAll")
 addCommandAlias("check", "all root/scalafmtSbtCheck root/scalafmtCheckAll")
+addCommandAlias("ci", "check; checkCanRunCI; git commit --only --allow-empty -m \"[run ci]\"")
 addCommandAlias(
   "compileJVM",
   ";coreTestsJVM/test:compile;stacktracerJVM/test:compile;streamsTestsJVM/test:compile;testTestsJVM/test:compile;testRunnerJVM/test:compile;examplesJVM/test:compile"
@@ -54,6 +55,7 @@ lazy val root = project
     skip in publish := true,
     console := (console in Compile in coreJVM).value,
     unusedCompileDependenciesFilter -= moduleFilter("org.scala-js", "scalajs-library"),
+    checkCanRunCISetting,
     welcomeMessage
   )
   .aggregate(
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -151,6 +151,12 @@ If all the tests are passing, then you can format your code:
 fmt
 ```
 
+And then set CI to run with:
+
+```bash
+ci
+```
+
 If your changes altered an API, then you may need to rebuild the microsite to make sure none of the (compiled) documentation breaks:
 
 ```bash
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -37,11 +37,14 @@ object BuildHelper {
       )
     else Nil
 
-  private def propertyFlag(property: String, default: Boolean) =
-    sys.props.get(property).map(_.toBoolean).getOrElse(default)
+  private def propertyFlag(property: String) = sys.props.get(property).map(_.toBoolean)
+
+  private def envVarFlag(property: String) = sys.env.get(property).map(_.toBoolean)
+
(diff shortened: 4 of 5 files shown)


reibitto:
Also, technically you can make the `ci` command stricter by making sure the code compiles (or even make sure the tests pass locally), but that might be too much?

I'd like to get some opinions before cleaning this up a bit and testing more. Just so I know whether I shouldn't spend time on this if it's going to be rejected. For example, I think it would also make sense to always run CI on `master` regardless of `[run ci]` or not. Little things like that need to be thought through.

unclebob418:
I like the idea. I think you could solve the issue of the extra step by making the `ci` sbt task dependent on `fmt`. It still needs to  run `check` though. Bonus points for making it re`fmt` and re`check` if it fails :) I've come across the scalafmt not-quite idempotent issue a few times now that I know it exists... including just now!

```
[IJ]sbt:test-tests> fmt
[info] Formatting 1 Scala sources...
[info] Formatting 1 Scala sources...
[success] Total time: 1 s, completed 28/10/2019 12:39:35 AM
[IJ]sbt:test-tests> fmt
[info] Formatting 1 Scala sources...
[success] Total time: 1 s, completed 28/10/2019 12:39:38 AM
[IJ]sbt:test-tests> fmt
[success] Total time: 1 s, completed 28/10/2019 12:39:40 AM
```

softinio:
Just taken a look at this PR and I don't see the value in it right now as to me its seems we are trying to fix something that is not broken. I am not aware of us going over our monthly usage allowance or the performance of builds being a concern to warrant this.

I think there is value in ci running to highlight issues sooner.

One idea is instead introduce an optional precommit hook to force linting and tests to run locally before pushing.

reibitto:
If it's not a problem and won't be in the foreseeable future, then I guess it's fine to revist this idea (or some other one) later when it makes sense to do so.

I was testing this idea because of John's comments [here](https://github.com/zio/zio/pull/1924#issuecomment-539113881) and [here](https://github.com/zio/zio/pull/2031#issuecomment-546463248). My fatal-warnings change is what brought this up in the first place, so I felt responsible to do something about it. Maybe making fatal-warnings on by default is enough for now?

But I was also trying to address how we're currently using CI to catch scalafmt issues, when that really should be done locally. We're relying on discipline right now when it comes to manually running `fmt`. Much like how we're trusting the programmer not to ignore compiler warnings. I know multiple people have expressed concerns about the latter. Not too sure about the former though.

> One idea is instead introduce an optional precommit hook to force linting and tests to run locally before pushing.

A pre-push hook can work, but it has the problem in that I think it would be basically go unnoticed and not used by most people. It's entirely an opt-in thing, isn't it?

softinio:
the linting step of the current build should be catching scalafmt errors. Is it not?

softinio:
@reibitto thanks for sharing the links to those comments also as I was not aware of them. I have more context now to think about.

reibitto:
> the linting step of the current build should be catching scalafmt errors. Is it not?

Yeah, it is catching them. Compiler warnings are also being caught during CI (well, now at least... there was a bug in the linked issue). The concern was more about just offloading these checks to CI only rather than giving it a pass locally first. Especially if there is a financial cost to this (though I don't actually have insight into that, so I can't say for sure).

NeQuissimus:
I think it would be much easier if you instead used pre-commit hooks.
They stay client-side and you can enforce (unless somebody does `--no-verify`) linting and formatting very easily.

softinio:
I still think this adds complexity to our build pipeline and adds another level of knowledge new contributors will need to learn to utilize. 

I vote for a precommit hook personally.

Resources: 

https://blog.softwaremill.com/reusable-pre-commit-hooks-in-scala-projects-d8bb327047ee

iravid:
I have great distaste for pre-commit hooks as they make my editor hang. Big :-1:. But am willing to compromise if someone shows me how to disable them, by default, with no extra steps required to disable them on every commit.

iravid:
Also, @reibitto - can you rephrase what this PR is trying to solve? Are fatal warnings not enabled in CI?

If this is not a problem we're actually having, let's not change anything. The CI is in great shape and we haven't run out of credits as of late.

softinio:
> I have great distaste for pre-commit hooks as they make my editor hang. Big 👎. But am willing to compromise if someone shows me how to disable them, by default, with no extra steps required to disable them on every commit.

I totally agree thats why I suggested **optional**. We can have the script for the pre-commit hook in the repo and those interested in using it just need to copy or link to in their `.git/hooks` directory at work I have setup a `Makefile` for copying it for those interested.

softinio:
Quite honestly I don't think we even need a precommit hook but an optional one is fine. I won't be using it :-)

unclebob418:
One situation it would benefit is failing fast. I get annoyed if I've pushed, requested a review from someone, then go start something else, or go to bed only to have an email arrive 10 mins later telling me I missed an unused important (which recently happed to me in a separate file to the one my commit changed) or fmt didn't work (or I just forgot it). Personally I'm using intellij to format my code as I go, not fmt, so if this new task is changing the final step to running ci instead of fmt also saves me a keystroke :) (assuming fmt gets run by ci)

I've never used a precommit hook before but I sometimes like to commit code that doesn't even compile before trying a different approach to a problem so I can easily revert if it doesn't work out. It sounds like this idea would prevent it.

reibitto:
Yeah, I imagined this would be contentious. Especially since it affects everyone.

> I have great distaste for pre-commit hooks as they make my editor hang. Big 👎. But am willing to compromise if someone shows me how to disable them, by default, with no extra steps required to disable them on every commit.

Since hooks are opt-in, not opt-out, you don't need to do anything. Just don't register them. But this is also one of the reasons why hooks aren't very effective... 😛

But yeah, I agree pre-commit hooks are annoying if they're doing an operation that takes a while (like in this case here). pre-push hooks less so since they're less frequent, but frequent enough to still be annoying.

> Also, @reibitto - can you rephrase what this PR is trying to solve? Are fatal warnings not enabled in CI? I thoroughly enjoyed having fatal warnings disabled locally, but we should definitely have them enabled in Circle :-)

@iravid Yeah, fatal warnings are properly enabled in CI as of #2031. I was trying to address the issue of not encouraging using CI to do checks that we should be doing locally first. Like not bothering to run scalafmt and not addressing compiler warnings because "CI will check for that anyway". Which normally isn't a big problem in my opinion... but if compute resources are limited, or it's costing more money, then there's a case to be made to be more strict about it.

But like I said, I don't actually know what the costs/credits situation actually looks like with this project. If it's not a problem currently, then I have no issue tabling this until it starts becoming one. It's really not my call to make though. I was just suggesting one potential solution.

iravid:
Anecdotally I found (on other projects) that `scalafmtOnCompile` is a vastly more effective solution to making sure code hits the CI already formatted. But that also can be annoying with some editors.

My personal opinion is that the status quo is pretty good.

reibitto:
I'm still of the opinion that fatal-warnings should be on by default (while providing an override). I like having it be both a property + env var. The latter can be easier to enable globally.

But if nobody else sees it as a problem, I guess the status quo is fine for now.

softinio:
I would prefer to leave things as they are also. We worked hard to get our ci into such good state lets not touch something that is not broken.

jdegoes:
I too am not a fan of this specific approach to solving the problem (and it seems like many others are not a fan, too!). 

Personally I feel that the default local build experience should be very strict to catch most problems on the user's machine, and that a "power mode" should be opt-in, which disables things for faster local development by power users who know what they are doing.

jdegoes:
@reibitto Thank you for your work on this! 🙏

PR #2045, opened 2019-10-28
ZIO Test: Fix Bug in TestAspect#flaky

adamgfraser:
We were using `foldM` in `TestAspect#flaky` so we were not retrying tests that died. In particular, this is relevant to using `flaky` in combination with `TestAspect#timeout`, since `timeout` causes a test to die if the timeout is exceeded. This PR fixes this by using `foldCauseM` and adds a test to prevent regression.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
@@ -77,6 +77,26 @@ object TestAspectSpec
             anything
           )
         ),
+        testM("flaky retries a test that fails") {
+          for {
+            ref <- Ref.make(0)
+            spec = testM("flaky test") {
+              assertM(ref.update(_ + 1), equalTo(100))
+            } @@ flaky
+            result <- succeeded(spec)
+            n      <- ref.get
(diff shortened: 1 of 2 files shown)


adamgfraser:
Thanks to @unclebob418 for reporting!

jdegoes:
These type of issues were what originally motivated the effect failure = test failure. After living with the new model, do you think it is on the net still an improvement?

adamgfraser:
I think it is less than ideal to have assertion failures and runtime failures in separate channels. The challenge is that with making `Spec` effectual there is an `E` type there that we need to keep polymorphic for user errors. So we can't put test failures there while remaining fully polymorphic in the error type in `Spec`.

One idea is that we could specialize `Spec` to `ZSpec`, so that the error channel of `Spec` was fixed to `TestFailure[E]`, similar to how the error channel in `ZStream` is specialized to `Option[E]`. Then we would have to map the error from any effects into a `TestFailure` but it seems from the experience with `ZStream` that would be doable. The downside is we would be losing a little more generality in `Spec` but I don't think the generality is buying us much given how much of our other functionality like test aspects is already specialized to `ZSpec`.

What do you think?

PR #2046, opened 2019-10-28
Add ZManaged.makeInterruptible

neko-kai:
* use bracketExit in ZManaged.exit/First

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -406,14 +406,9 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) 
     ZManaged {
       Ref.make[Exit[Any, Any] => ZIO[R1, Nothing, Any]](_ => UIO.unit).map { finalizer =>
         Reservation(
-          acquire = ZIO.uninterruptibleMask { restore =>
-            for {
-              res   <- self.reserve
-              exitA <- restore(res.acquire).run
-              _     <- finalizer.set(exitU => res.release(exitU).ensuring(cleanup(exitA)))
-              a     <- ZIO.done(exitA)
-            } yield a
-          },
+          acquire = ZIO.bracketExit(self.reserve)(
+            (res, exitA: Exit[E, A]) => finalizer.set(exitU => res.release(exitU).ensuring(cleanup(exitA)))
+          )(_.acquire),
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1044,28 +1034,37 @@ object ZManaged {
...
+   * Lifts a `ZIO[R, E, A]` into `ZManaged[R, E, A]` with a release action that handles `Exit`.
    * The acquire and release actions will be performed uninterruptibly.
    */
   final def makeExit[R, E, A](
     acquire: ZIO[R, E, A]
   )(release: (A, Exit[Any, Any]) => ZIO[R, Nothing, Any]): ZManaged[R, E, A] =
     ZManaged(acquire.map(r => Reservation(IO.succeed(r), e => release(r, e))))
 
+  /**
+   * Lifts a ZIO[R, E, A] into ZManaged[R, E, A] with no release action. The
```
iravid: Scaladoc needs a fix :-)

neko-kai: @iravid done!

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -406,14 +406,9 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
         Reservation(
-          acquire = ZIO.uninterruptibleMask { restore =>
-            for {
-              res   <- self.reserve
-              exitA <- restore(res.acquire).run
-              _     <- finalizer.set(exitU => res.release(exitU).ensuring(cleanup(exitA)))
-              a     <- ZIO.done(exitA)
-            } yield a
-          },
+          acquire = ZIO.bracketExit(self.reserve)(
```
iravid: Curious, what's the motivation for the change? The differences are mainly correct usage of `ZIOFn` and error preservation (which shouldn't matter here because `finalizer.set` is a `UIO`)

neko-kai: @iravid Less low-level code; also this is the only instance of `uninterruptibleMask(restore(_).catch)` aside of `bracketExit` in core and turns out it just duplicates it – making it use `bracketExit` also safeguards it against any changes to finalizer behavior :)

iravid: Ok, makes sense!

PR #2049, opened 2019-10-28
Implement CanFail

### LINKED ISSUE #2047: Compile-time errors for useless combinators
Various combinators don't make sense for infallible errors. We should try to produce compile-time errors for:

```scala
uio orElse zio // Error!
uio catchAll f 
uio catchSome f
uio.option
uio.either
uio retry s
// etc
```

adamgfraser:
Resolves #2047.

Adds a new implicit instance `CanFail[E]` which provides implicit evidence that an effect with an error type `E` can fail, that is that `E` is not equal to `Nothing`. Requires such evidence to exist for ZIO combinators such as `orElse` that only make sense if an effect can fail.

Not implemented yet is adding annotations so that the implicit type does not appear in the Scaladoc. It looks like this has to be done through the `@useCase` annotation, but one watch out there is it requires you to specify the "simplified" type signature as a text string, so there is some risk that the documentation gets out of sync with the actual type signature since it is not tied to the code.

Also not implemented is applying this technique to other effect types such as `ZManaged` and `ZStream`.

### DIFF
--- a/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
+++ b/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
@@ -264,7 +264,7 @@ class StacktracesSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
         _ <- ZIO.unit
         _ <- ZIO.unit
       } yield t)
-        .foldCauseM(
+        .foldM(
           failure = _ => IO.fail(()),
           success = t =>
             IO.trace
--- a/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
+++ b/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
@@ -264,9 +264,8 @@ class StacktracesSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
         _ <- ZIO.unit
         _ <- ZIO.unit
       } yield t)
-        .foldM(
-          failure = _ => IO.fail(()),
-          success = t =>
+        .flatMap(
+          t =>
             IO.trace
               .map(tuple(t))
         )
--- a/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
+++ b/core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
@@ -264,7 +264,7 @@ class StacktracesSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
         _ <- ZIO.unit
         _ <- ZIO.unit
       } yield t)
-        .foldM(
+        .foldCauseM(
           failure = _ => IO.fail(()),
           success = t =>
             IO.trace
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -181,13 +181,15 @@ object ZIOSpec
         ),
         suite("option")(
           testM("return success in Some") {
-            assertM(ZIO.succeed(11).option, equalTo(Some(11)))
+            val zio: ZIO[Any, String, Int] = ZIO.succeed(11)
+            assertM(zio.option, equalTo(Some(11)))
           },
           testM("return failure as None") {
             assertM(ZIO.fail(123).option, equalTo(None))
           },
           testM("not catch throwable") {
-            assertM(ZIO.die(ExampleError).option.run, dies(equalTo(ExampleError)))
+            val zio: ZIO[Any, String, Int] = ZIO.die(ExampleError)
+            assertM(zio.option.run, dies(equalTo(ExampleError)))
           },
           testM("catch throwable after sandboxing") {
             assertM(ZIO.die(ExampleError).sandbox.option, equalTo(None))
@@ -623,7 +625,7 @@ object ZIOSpec
             assertM(fib(1000), equalTo(expected))
           },
           testM("deep absolve/attempt is identity") {
-            val io = (0 until 1000).foldLeft(IO.succeed(42)) { (acc, _) =>
+            val io = (0 until 1000).foldLeft[ZIO[Any, String, Int]](IO.succeed(42)) { (acc, _) =>
               IO.absolve(acc.either)
             }
 
@@ -833,15 +835,15 @@ object ZIOSpec
                 r <- pa.await zip pb.await
               } yield r
 
-            assertM(io.eventually, equalTo((1, 2)))
+            assertM(io, equalTo((1, 2)))
           } @@ ignore,
           testM("race of fail with success") {
             val io = IO.fail(42).race(IO.succeed(24)).either
             assertM(io, isRight(equalTo(24)))
           },
           testM("race of terminate with success") {
-            val io = IO.die(new Throwable {}).race(IO.succeed(24)).either
-            assertM(io, isRight(equalTo(24)))
+            val io = IO.die(new Throwable {}).race(IO.succeed(24))
+            assertM(io, equalTo(24))
           },
           testM("race of fail with fail") {
             val io = IO.fail(42).race(IO.fail(42)).either
@@ -870,7 +872,7 @@ object ZIOSpec
               winner = s.await *> IO.fromEither(Right(()))
               loser  = IO.bracket(s.succeed(()))(_ => effect.succeed(42))(_ => IO.never)
               race   = winner raceAttempt loser
-              _      <- race.either
+              _      <- race
               b      <- effect.await
             } yield assert(b, equalTo(42))
           },
@@ -1047,7 +1049,7 @@ object ZIOSpec
             for {
               cont <- Promise.make[Nothing, Unit]
               p1   <- Promise.make[Nothing, Boolean]
-              f1   <- (cont.succeed(()) *> IO.never).catchAll(IO.fail).ensuring(p1.succeed(true)).fork
+              f1   <- (cont.succeed(()) *> IO.never).catchAllCause(IO.fail).ensuring(p1.succeed(true)).fork
               _    <- cont.await
               _    <- f1.interrupt
               res  <- p1.await
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -181,15 +181,15 @@ object ZIOSpec
         ),
         suite("option")(
           testM("return success in Some") {
-            val zio: ZIO[Any, String, Int] = ZIO.succeed(11)
-            assertM(zio.option, equalTo(Some(11)))
+            import zio.CanFail.canFail
+            assertM(ZIO.succeed(11).option, equalTo(Some(11)))
           },
           testM("return failure as None") {
             assertM(ZIO.fail(123).option, equalTo(None))
           },
           testM("not catch throwable") {
-            val zio: ZIO[Any, String, Int] = ZIO.die(ExampleError)
-            assertM(zio.option.run, dies(equalTo(ExampleError)))
+            import zio.CanFail.canFail
+            assertM(ZIO.die(ExampleError).option.run, dies(equalTo(ExampleError)))
           },
           testM("catch throwable after sandboxing") {
             assertM(ZIO.die(ExampleError).sandbox.option, equalTo(None))
@@ -625,7 +625,8 @@ object ZIOSpec
             assertM(fib(1000), equalTo(expected))
           },
           testM("deep absolve/attempt is identity") {
-            val io = (0 until 1000).foldLeft[ZIO[Any, String, Int]](IO.succeed(42)) { (acc, _) =>
+            import zio.CanFail.canFail
+            val io = (0 until 1000).foldLeft(IO.succeed(42)) { (acc, _) =>
               IO.absolve(acc.either)
             }
 
@@ -1045,11 +1046,12 @@ object ZIOSpec
               r <- done.await.timeoutTo(42)(_ => 0)(60.second)
             } yield assert(r, equalTo(0))
           },
-          testM("redeem + ensuring + interrupt") {
+          testM("catchAll + ensuring + interrupt") {
+            import zio.CanFail.canFail
             for {
               cont <- Promise.make[Nothing, Unit]
               p1   <- Promise.make[Nothing, Boolean]
-              f1   <- (cont.succeed(()) *> IO.never).ensuring(p1.succeed(true)).fork
+              f1   <- (cont.succeed(()) *> IO.never).catchAll(IO.fail).ensuring(p1.succeed(true)).fork
               _    <- cont.await
               _    <- f1.interrupt
               res  <- p1.await
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1049,7 +1049,7 @@ object ZIOSpec
             for {
               cont <- Promise.make[Nothing, Unit]
               p1   <- Promise.make[Nothing, Boolean]
-              f1   <- (cont.succeed(()) *> IO.never).catchAllCause(IO.fail).ensuring(p1.succeed(true)).fork
+              f1   <- (cont.succeed(()) *> IO.never).ensuring(p1.succeed(true)).fork
               _    <- cont.await
               _    <- f1.interrupt
               res  <- p1.await
--- a/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
@@ -116,10 +116,10 @@ object ZScheduleSpec
         ),
         suite("Retry on failure according to a provided strategy")(
           testM("retry 0 time for `once` when first time succeeds") {
-            def update(ref: Ref[Int]): ZIO[Any, String, Int] = ref.update(_ + 1)
+            import zio.CanFail.canFail
             for {
               ref <- Ref.make(0)
-              _   <- update(ref).retry(Schedule.once)
+              _   <- ref.update(_ + 1).retry(Schedule.once)
               i   <- ref.get
             } yield assert(i, equalTo(1))
           },
--- a/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
@@ -116,9 +116,10 @@ object ZScheduleSpec
         ),
         suite("Retry on failure according to a provided strategy")(
           testM("retry 0 time for `once` when first time succeeds") {
+            def update(ref: Ref[Int]): ZIO[Any, String, Int] = ref.update(_ + 1)
             for {
               ref <- Ref.make(0)
-              _   <- ref.update(_ + 1).retry(Schedule.once)
+              _   <- update(ref).retry(Schedule.once)
               i   <- ref.get
             } yield assert(i, equalTo(1))
           },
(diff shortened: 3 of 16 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -833,15 +835,15 @@ object ZIOSpec
...
 
-            assertM(io.eventually, equalTo((1, 2)))
+            assertM(io, equalTo((1, 2)))
           } @@ ignore,
           testM("race of fail with success") {
             val io = IO.fail(42).race(IO.succeed(24)).either
             assertM(io, isRight(equalTo(24)))
           },
           testM("race of terminate with success") {
-            val io = IO.die(new Throwable {}).race(IO.succeed(24)).either
```
mschuwalow: Nice cleanup 👍

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -16,15 +16,18 @@
...
 import zio.clock.Clock
 import zio.duration._
 import zio.internal.tracing.{ ZIOFn, ZIOFn1, ZIOFn2 }
 import zio.internal.{ Executor, Platform }
-import scala.concurrent.ExecutionContext
-import scala.util.{ Failure, Success }
-import scala.reflect.ClassTag
-import zio.{ TracingStatus => TrasingS }
 import zio.{ InterruptStatus => InterruptS }
+import zio.{ TracingStatus => TrasingS }
```
mschuwalow: probably a typo: TracingS

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -104,13 +107,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Maps the error value of this effect to the specified constant value.
    */
-  final def asError[E1](e1: => E1): ZIO[R, E1, A] = mapError(new ZIO.ConstFn(() => e1))
+  @silent("never used")
```
mschuwalow: Can't we turn of warning for unused implicit params with `-Ywarn-unused:params.-implicits`?
Might be a bit nicer than silencing everywhere

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Good catch! Fixed in my version now.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Yes, I think it would definitely be nice to do something so we don't have to have those annotations everywhere. I think we could either do that or there is an option in the silencer to accept a global regex so we could like always call the parameter `canFail` and silence errors that say that was never used. Personally I think the compiler option is cleaner but I want to make sure other people are good with that because it could potentially be useful in some cases. Though I do see we are already silencing some other warnings about implicit parameters not being used.

core/shared/src/main/scala-2.11/zio/CanFail.scala
```diff
@@ -0,0 +1,35 @@
...
+ * A value of type `CanFail[E]` provides implicit evidence that an effect with
+ * error type `E` can fail, that is, that `E` is not equal to `Nothing`.
+ */
+@implicitNotFound("This operation only makes sense for effects that can fail.")
+sealed trait CanFail[-E]
+
+object CanFail {
+
+  implicit final def canFail[E]: CanFail[E] =
+    new CanFail[E] {}
```
neko-kai: Cache value? (Or return `null` :D)

adamgfraser:
@vasilmkd Yes definitely.

neko-kai:
@adamgfraser 
What do you think about adding tests that the combinators actually fail when called with `Nothing`? 

* For Scala 2 you can use `shapeless.test.illTyped` in test scope
* Dotty apparently has a builtin `scala.compiletime.testing.typeChecks` function - [example](https://github.com/milessabin/shapeless/blob/shapeless-3-staging/core/src/main/scala/shapeless/test/typechecking.scala)

adamgfraser:
@neko-kai Yes, we should definitely test this. I also wonder if we should add this functionality to ZIO Test.

You’re right about Dotty. We need to use a slight variation of this technique to work across all platforms. I’ll push a commit to fix that shortly.

jdegoes:
@neko-kai @adamgfraser Please _no_ dependency on Shapeless. If we do anything for this, it should be possible to support Scala 2.x and 3.x simultaneously.

core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
```diff
@@ -264,7 +264,7 @@ class StacktracesSpec(implicit ee: org.specs2.concurrent.ExecutionEnv)
         _ <- ZIO.unit
         _ <- ZIO.unit
       } yield t)
-        .foldM(
+        .foldCauseM(
```
jdegoes: This change isn't necessary.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -1047,7 +1049,7 @@ object ZIOSpec
             for {
               cont <- Promise.make[Nothing, Unit]
               p1   <- Promise.make[Nothing, Boolean]
-              f1   <- (cont.succeed(()) *> IO.never).catchAll(IO.fail).ensuring(p1.succeed(true)).fork
+              f1   <- (cont.succeed(()) *> IO.never).catchAllCause(IO.fail).ensuring(p1.succeed(true)).fork
```
jdegoes: This change won't be necessary, I don't think. If you can't call `catchAll` then it's the same as deleting it, e.g. `eff.catchAll(IO.fail(_))` is identity.

core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
```diff
@@ -116,9 +116,10 @@ object ZScheduleSpec
         ),
         suite("Retry on failure according to a provided strategy")(
           testM("retry 0 time for `once` when first time succeeds") {
+            def update(ref: Ref[Int]): ZIO[Any, String, Int] = ref.update(_ + 1)
             for {
               ref <- Ref.make(0)
-              _   <- ref.update(_ + 1).retry(Schedule.once)
+              _   <- update(ref).retry(Schedule.once)
```
jdegoes: Why was this one necessary? Seems a lot of boilerplate.

core/shared/src/main/scala-2.11/zio/CanFail.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package zio
+
+import scala.annotation.implicitNotFound
+
+/**
+ * A value of type `CanFail[E]` provides implicit evidence that an effect with
+ * error type `E` can fail, that is, that `E` is not equal to `Nothing`.
+ */
+@implicitNotFound("This operation only makes sense for effects that can fail.")
```
jdegoes: 👍

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -104,13 +105,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Maps the error value of this effect to the specified constant value.
    */
-  final def asError[E1](e1: => E1): ZIO[R, E1, A] = mapError(new ZIO.ConstFn(() => e1))
+  final def asError[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
```
jdegoes: This one makes sense because for an effect with `E = Nothing`, it can already be treated as a subtype of a more specific error type.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -104,13 +105,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+    mapError(new ZIO.ConstFn(() => e1))
 
   /**
    * Returns an effect whose failure and success channels have been mapped by
    * the specified pair of functions, `f` and `g`.
    */
-  final def bimap[E2, B](f: E => E2, g: A => B): ZIO[R, E2, B] = mapError(f).map(g)
+  final def bimap[E2, B](f: E => E2, g: A => B)(implicit ev: CanFail[E]): ZIO[R, E2, B] = mapError(f).map(g)
```
jdegoes: This one makes sense but seems more trouble than it's worth, because the right hand side is always useful; so at worse this is "half useful".

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -444,7 +449,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * The error parameter of the returned `IO` may be chosen arbitrarily, since
    * it will depend on the `IO`s returned by the given continuations.
    */
-  final def foldM[R1 <: R, E2, B](failure: E => ZIO[R1, E2, B], success: A => ZIO[R1, E2, B]): ZIO[R1, E2, B] =
+  final def foldM[R1 <: R, E2, B](failure: E => ZIO[R1, E2, B], success: A => ZIO[R1, E2, B])(
```
jdegoes: I would not change this one either, it can be used for proofs that the error cannot happen.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -558,7 +565,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * function. This can be used to lift a "smaller" error into a "larger"
    * error.
    */
-  final def mapError[E2](f: E => E2): ZIO[R, E2, A] =
+  final def mapError[E2](f: E => E2)(implicit ev: CanFail[E]): ZIO[R, E2, A] =
```
jdegoes: This one is useful since there is no need to map the error when `E` = Nothing, since it's already a subtype of all other errors.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Okay.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1157,7 +1164,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def some[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], B] =
     self.foldM(
       e => ZIO.fail(Some(e)),
-      a => a.fold[ZIO[R, Option[E], B]](ZIO.fail(Option.empty[E]))(ZIO.succeed(_).mapError(_ => Option.empty[E]))
+      a => a.fold[ZIO[R, Option[E], B]](ZIO.fail(Option.empty[E]))(ZIO.succeed(_))
```
jdegoes: I like how this change fixed some issues like this (dead code). ❤️

core/shared/src/main/scala/zio/ZIO.scala
jdegoes: Hmm, I guess such `bimap` can always be replaced by `map`, so maybe it's useful to leave this one (sorry!).

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1259,7 +1266,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * readFile("data.json").tapBoth(logError(_), logData(_))
    * }}}
    */
-  final def tapBoth[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, Any], g: A => ZIO[R1, E1, Any]): ZIO[R1, E1, A] =
+  final def tapBoth[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, Any], g: A => ZIO[R1, E1, Any])(
+    implicit ev: CanFail[E]
```
jdegoes: A failing example can always be replaced by `tap`, so I think this is useful.

docs/datatypes/ref.md
```diff
@@ -90,7 +90,7 @@ object S {
               case false => IO.fail(())
               case true  => IO.unit
             }
-        } <> P).either.unit
+        } <> P).unit
```
jdegoes: ❤️

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1093,8 +1093,8 @@ object StreamSpec
             val s = Stream.fromIterable(data)
 
             for {
-              l <- s.mapM(f).runCollect.either
-              r <- IO.foreach(data)(f).either
+              l <- s.mapM(f).runCollect
+              r <- IO.foreach(data)(f)
```
jdegoes: ❤️

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1177,18 +1177,16 @@ object StreamSpec
...
           },
           testM("mergeWith") {
             val s1 = Stream(1, 2)
             val s2 = Stream(1, 2)
 
             for {
-              merge <- s1.mergeWith(s2)(_.toString, _.toString).runCollect.either
-              list  = merge.fold[List[String]](_ => Nil, identity)
-            } yield assert(list, hasSameElements(List("1", "2", "1", "2")))
+              merge <- s1.mergeWith(s2)(_.toString, _.toString).runCollect
```
jdegoes: ❤️

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: No problem! I struggled with the same thing on some of these.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -418,7 +423,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * does not fail, but succeeds with the value returned by the left or right
    * function passed to `fold`.
    */
-  final def fold[B](failure: E => B, success: A => B): ZIO[R, Nothing, B] =
+  final def fold[B](failure: E => B, success: A => B)(implicit ev: CanFail[E]): ZIO[R, Nothing, B] =
```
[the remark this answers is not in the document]
neko-kai: Do you mean use an implicit `E =:= Nothing`? They can still use this method If E is not literal a Nothing, but only by evidence. Otherwise I'm missing what's the additional power vs. `.map(success)`, for E=Nothing the type is always `ZIO[R, Nothing, A] => ZIO[R, Nothing, B]`, i.e. `map`

adamgfraser:
Okay, so to summarize I rolled back `fold` and `foldM`.

Open questions / next steps:

1. Making similar changes to other ZIO data types.
2. Do we want to use the `@useCase` annotation for these? Personally I am somewhat inclined to leave it as is. First, the name `CanFail` now makes it very clear what it is, thanks to your good suggestion to use a more specific type constraint, and we can turn it into a feature. Second, it will make sure the type signature stays in sync for any changes to these methods. Third, it avoids the "that isn't actually the type signature" critique that I've seen applied to other uses of `@useCase`. But happy to do whatever you think is best.
3. Testing - what do you think about pursuing a separate PR to add some "does not compile" functionality to ZIO Test. We would have to do some platform specific work for Scala 2 versus Dotty but I think we could do all of that on the backend and expose a consistent API across platforms that the entire ZIO ecosystem could use so they didn't have to rely on third part libraries. ScalaTest offers this functionality.

adamgfraser:
It may not be worth it but we could potentially do the same thing with the `R` type (e.g. it doesn't make sense to call `provide` on an effect that does not require any environment).

core-tests/shared/src/test/scala/zio/ZScheduleSpec.scala
jdegoes: Ah right! 👍

jdegoes:
> Okay, so to summarize I rolled back fold and foldM.

@neko-kai argued to leave them on grounds that `fold` can always be replaced by `map` when the error is `Nothing`, and `foldM` can always be replaced by `flatMap` when the error is `Nothing`.

I think it is a valid argument, it just concerned me when I saw, e.g. `foldM` being replaced by `foldCauseM` merely to work around the (new) compiler error.

What do you think?

> Making similar changes to other ZIO data types.

Makes sense.

> First, the name CanFail now makes it very clear what it is, thanks to your good suggestion to use a more specific type constraint, and we can turn it into a feature. 

I think you're right about this, `CanFail` is good enough.

> Testing - what do you think about pursuing a separate PR to add some "does not compile" functionality to ZIO Test. 

Agreed, if we had something that was cross Scala (2.11 - Dotty), it could make sense to add to ZIO Test, and that's precisely the case when I'd feel comfortable adding it here.

vasilmkd:
Probably a good idea to watch this talk, I found it pretty fascinating when I first watched it. There might be some good hints when it comes to naming as well:

[Plain Functional Programming by Martin Odersky](https://www.youtube.com/watch?v=YXDm3WHZT5g).

adamgfraser:
@jdegoes I don’t think `uio.fold` or `uio.foldM` should compile since they can always be rewritten in terms of `map` and `flatMap`. It was my mistake in not rewriting one of the tests correctly. What we should do is put together documentation showing how to rewrite code for this change. I will put together a table that we can reference in the release notes or elsewhere.

I will add back the changes to `fold` and `foldM` and make similar changes to other data types.

I have also been working on something to test that code does not compile so I will post that shortly, though that will be in a separate PR.

Did you have any interest in doing something similar for the environment type?

adamgfraser:
Code | Rewrite 
--- | ---
`uio.asError(e)` | `uio`
`uio.bimap(f, g)` |  `uio.map(g)`
`uio.catchAll(f)` | `uio`
`uio.catchSome(pf)` | `uio`
`uio.either` | `uio`*
`uio.eventually` | `uio`
`uio.flatMapError(f)` | `uio`
`uio.fold(f, g)` | `uio.map(g)`
`uio.foldM(f, g)` | `uio.flatMap(g)`
`uio.mapError(f)` | `uio`
`uio.option` | `uio`*
`uio.orDie` | `uio`
`uio.orDieWith(f)` | `uio`
`uio.orElse(zio)` | `uio`
`uio.orElseEither(zio)` | `uio`*
`uio.refineOrDie(pf)` | `uio`
`uio.refineOrDieWith(pf)` | `uio`
`uio.refineToOrDie` | `uio`
`uio.retry(s)` | `uio`
`uio.retryOrElse(s, f)` | `uio`
`uio.retryOrElseEither(s, f)` | `uio`*
`uio.tapBoth(f, g)` | `uio.tap(g)`
`uio.tapError(f)` | `uio`

* `either`, `option`, `orElseEither`, and `retryOrElseEither` wrap their results in `Some` or `Right` so after rewriting, code calling these methods can be simplified to accept an `A` rather than an `Option[A]` or `Either[E, A]`.

jdegoes:
>  I don’t think uio.fold or uio.foldM should compile since they can always be rewritten in terms of map and flatMap. It was my mistake in not rewriting one of the tests correctly. What we should do is put together documentation showing how to rewrite code for this change. I will put together a table that we can reference in the release notes or elsewhere.

OK, sounds great!

> Did you have any interest in doing something similar for the environment type?

Yes, I think it's worthwhile to do that. Maybe `RequiresEnv` for that one?

adamgfraser:
Okay, great! I will open a separate PR to do the same thing for the environment type. Maybe `NeedsEnv`? Happy to do either.

jdegoes:
Like `NeedsEnv` 👍

core/shared/src/main/scala-2.11/zio/CanFail.scala
```diff
@@ -0,0 +1,34 @@
...
+sealed trait CanFail[-E]
+
+object CanFail extends CanFail[Any] {
+
+  implicit final def canFail[E]: CanFail[E] = CanFail
+
+  implicit final val canFailAmbiguous1: CanFail[Nothing] = CanFail
+  implicit final val canFailAmbiguous2: CanFail[Nothing] = CanFail
```
regiskuckaertz: For my own understanding, why is there two such values? Also it may be useful to add an explanatory comment for the future.

adamgfraser:
I'm on it!

PR #2059, opened 2019-10-30
ZIO Test: Implement assertCompiles

adamgfraser:
Adds a new method `assertCompiles` that requires the specified string to be valid Scala code.


ioleo:
@adamgfraser Thank you! :pray:

test/jvm/src/main/dotty/zio/test/AssertionVariants.scala
```diff
@@ -0,0 +1,30 @@
...
+package zio.test
+
+import scala.compiletime.testing.typeChecks
+
+trait AssertionVariants {
+
+  /**
+   * Makes a new assertion that requires the specified string to be valid Scala
+   * code.
+   */
```
neko-kai: hmm, why in `jvm` ? I'd guess this should in dotty `--js` mode too?

test/shared/src/main/scala2/zio/test/Macros.scala
```diff
@@ -0,0 +1,32 @@
...
+
+private[test] object Macros {
+
+  def assertCompiles_impl(c: Context)(code: c.Expr[String]): c.Expr[TestResult] = {
+    import c.universe._
+    try {
+      c.typecheck(c.parse(c.eval(c.Expr[String](c.untypecheck(code.tree)))))
+      c.Expr(q"assert(true, Assertion.isTrue)")
+    } catch {
+      case _: Throwable => c.Expr(q"assert(false, Assertion.isTrue)")
```
neko-kai: Is it possible to splice the error message here?

build.sbt
```diff
@@ -151,9 +151,16 @@ lazy val test = crossProject(JSPlatform, JVMPlatform)
   .dependsOn(core, streams)
   .settings(stdSettings("zio-test"))
   .settings(
-    libraryDependencies ++= Seq(
-      "org.portable-scala" %%% "portable-scala-reflect" % "0.1.0"
-    )
+    scalacOptions += "-language:experimental.macros",
+    libraryDependencies ++= {
+      if (isDotty.value)
+        Seq("org.portable-scala" %%% "portable-scala-reflect" % "0.1.0")
```
jdegoes: This dependency occurs in both cases, so perhaps you can push the conditional lower, e.g. `++= Seq("org...") ++ if (isDotty.value()) Seq() else Seq(...)`

adamgfraser: Done.

test/jvm/src/main/dotty/zio/test/AssertionVariants.scala
adamgfraser: Good thought. Moved to `shared` folder.

test/shared/src/main/scala2/zio/test/Macros.scala
adamgfraser: I think it is. The reason I didn't is that on Dotty `typeChecks` just returns whether or not code compiles. So if we want to have a consistent API across platforms we would have to do something a little fancier on Dotty to get the error messages there as well. While I haven't looked into it I assume that is doable. Is that a direction we want to go?

adamgfraser:
I addressed the comment from @jdegoes and moved the Dotty specific code to the `shared` folder. I also added a test suite for `CanFail`. I think the only open question is whether we want to support making assertions about the error returned if code does not compile versus just whether code compiles. If we want to do the first I have a little more work to do. Otherwise I think this is ready to merge.

jdegoes:
🎉

test/shared/src/main/scala2/zio/test/Macros.scala
neko-kai: @adamgfraser 
I think No, because the staging/quote/tasty APIs are not stable yet, so we'd have to fix them when upgrading version.
However, nearly all users are going to use this API from Scala 2, in which case they'd probably find an error message in their test very helpful, as otherwise on a test failure they'd have to move the failing code outside of the string to get the error message.

So, my opinion is that we should leave dotty as is because the relevant APIs aren't stable and there are very few users who'll notice, but it would be pretty helpful for our users to get the error messages on Scala 2.

adamgfraser: @neko-kai I think those are excellent points. Let me work on a follow up PR to add that functionality.

neko-kai: @adamgfraser Thanks!

PR #2061, opened 2019-10-31
readd ZSchedule#foldM

mschuwalow:
Seems we missed a few things in the larger schedule pr.

Also changes the behavior of both fold and foldM to until consider values if the schedule intends to continue.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -68,19 +68,19 @@ object ScheduleSpec extends ZIOBaseSpec {
     suite("Collect all inputs into a list")(
       testM("as long as the condition f holds") {
         def cond: Int => Boolean = _ < 10
-        checkRepeat(Schedule.collectWhile(cond), expected = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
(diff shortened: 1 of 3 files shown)


mschuwalow:
> @mschuwalow In this PR, can you allow "narrowing" of methods that require a fixed environment? If the user is only using `Clock`, we don't want to blow that up to `ZEnv`, which means we need only implicit evidence `R1 <:< ZEnv` for some `R1`

I discussed this with @adamgfraser briefly in discord. I believe that this is impossible to do because the ZEnv.mapX functions will force this widening

jdegoes:
@mschuwalow We can change the map functions to be more general. You can always treat a bigger thing as a smaller thing, i.e. return `R2` as `R1` when `R2 <: R1`.

mschuwalow:
@jdegoes Could you provide a code sample for it? I feel like I'm not seeing something here

mschuwalow:
Independently of any encoding changes to env I would like to get this one merged.

jdegoes:
You're right we can't narrow, we will deal with that in other tickets.

iravid:
@mschuwalow Can you rebase on master?

mschuwalow:
> @mschuwalow Can you rebase on master?

done 👍

PR #2066, opened 2019-10-31
Address Flaky Tests

adamgfraser:
Makes the following changes based on my review of recent build failures to make CI a reliable indicator for new PRs in advance of Berlin hackathon:

- Marks  "effectBlockingCanBeInterrupted" in `BlockingSpec` as `ignore`. It appears that `effectBlocking` can't always be interrupted, leading to a resource leak, which is hampering our ability to address by just marking it as `flaky`.
- Marks "read only STM shouldn't return partial state of concurrent read-write STM" in `STMSpec` as flaky. I have seen this fail several times on recent builds.
- Marks "cancellation is guaranteed" in `RTSSpec` as `flaky` and removes previous `nonFlaky` test aspect. The fact that it was still only failing sometimes even with `nonFlaky` indicates that this is an infrequent bug but still something going on here.

Obviously we should treat these as things we need to fix, maybe while we are in Berlin, but it will be a lot nicer for everyone if red means there is something wrong with your code instead of a spurious test failure.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
@@ -7,7 +7,7 @@ import zio.clock.Clock
 import zio.duration._
 import zio.test._
 import zio.test.Assertion._
-import zio.test.TestAspect.{ jvm, nonFlaky }
+import zio.test.TestAspect.{ flaky, jvm, nonFlaky }
 
 object RTSSpec
     extends ZIOBaseSpec(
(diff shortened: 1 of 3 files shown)


jdegoes:
Can you create tickets for all the tests so marked? Also maybe it would be worthwhile to introduce a `flaky` Specs 2 combinator so we can ensure they at least pass once in a while until we fix them?

adamgfraser:
"effectBlockingCanBeInterrupted" has a ticket open in #1882.
I just opened tickets for the STM issue in #2073 and for the RTS issue in #2074.

These tests have all been migrated to ZIO Test so I don't think there is anything with regard to Specs2. They are being marked as `flaky` so we are testing that they pass at least some of the time with the exception of the `effectBlockingCanBeInterrupted` test which seems to be creating a resource leak so `flaky` hasn't been successful in preventing timeouts.

PR #2075, opened 2019-11-01
#1630 ZIO Test: Gen Benchmark

### LINKED ISSUE #1630: ZIO Test: Benchmark and Explore Optimization Opportunities for Random Generators
As more people start using property based testing in ZIO Test we should benchmark how long it takes to generate common random data types with ZIO Test versus other property based testing frameworks and natively using `scala.util.Random`. Once we understand where we are, we can explore opportunities for optimization. For example, right now every random number we generate is wrapped in a `ZIO` which is itself modifying the `Ref` that underlies `MockRandom`. We could potentially generate data in batches in certain cases (e.g. if we know we need a list with 100 integers generate all of them in one effect). This would be somewhat less composition but could be worth it if we got a significant speedup.

A completed ticket would:

- [ ] Create a new `GenBenchmarks.scala` file in the `benchmarks` project set up the same way as existing benchmarks.
- [ ] Define and benchmark a few common random variable generation tasks (e.g. generate a random double, generate a list of random integers of a certain size, generate a random string of a certain size) using both the combinators and defined in ZIO Test and native methods on `sclala.util.Random`.
- [ ] Benchmark the same tasks using other Scala property based testing frameworks in Scala, in particular ScalaCheck and Hedgehog.

pk044:
Resolves #1630


benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,81 @@
...
+  @Benchmark
+  def zioIntListOfSizeN: Gen[Random, List[Int]] =
+    Gen.listOfN(size)(Gen.anyInt)
+
+  @Benchmark
+  def zioStringOfSizeN: Gen[Random, String] =
+    Gen.stringN(size)(Gen.anyChar)
+
+  @Benchmark
+  def scalaRandomDouble: Double =
```
jdegoes: It's not useful to include these in the benchmark, they aren't `Gen`-like.

benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,81 @@
...
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class GenBenchmarks {
+
+  @Param(Array("1000"))
+  var size: Int = _
+
+  @Benchmark
+  def zioDouble: Gen[Random, Double] =
+    Gen.double(0, size)
```
jdegoes: Things that just return `Gen` aren't useful to benchmark. You should use the `Gen` to generate a whole bunch of numbers (e.g. `size`.numbers).

benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,81 @@
...
+  def scalaCheckDouble: scalacheck.Gen[Double] =
+    scalacheck.Gen.choose(0, size)
+
+  @Benchmark
+  def scalaCheckIntListOfSizeN: scalacheck.Gen[List[Int]] =
+    scalacheck.Gen.listOfN(size, scalacheck.Gen.choose(0, size))
+
+  @Benchmark
+  def scalaCheckStringOfSizeN: scalacheck.Gen[String] =
+    scalacheck.Gen.listOfN(size, scalacheck.Gen.alphaChar).map(_.mkString)
```
jdegoes: These are all just benchmarking the cost of creating the `Gen`, not the cost of creating N number of values using the `Gen`. I suggest to change to the latter since that's what matters for performance of the tests.

benchmarks/src/main/scala/zio/GenBenchmarks.scala
pk044: #1630 
`generate common random data types with ZIO Test versus other property based testing frameworks and natively using scala.util.Random.`

do you mean we don't want to use scala.util.Random here at all anymore or am I just using it incorrectly in the benchmark?

pk044:
I updated the PR according to the comments, hopefully I didn't miss anything.

didn't want to call .get() on Option but I noticed that in a certain benchmark it is done identically on Either, so I think it should be fine here? :) 

I'm also unsure whether we should have a seed list just like in https://github.com/typelevel/scalacheck/blob/master/bench/src/main/scala/org/scalacheck/bench/GenBench.scala any thoughts on this?

benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,82 @@
...
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class GenBenchmarks {
+
+  @Param(Array("1000"))
+  var size: Int = _
+
+  @Benchmark
+  def zioDouble: Double =
```
adamgfraser: I think we want to generate a `List[Double]` here.

benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,82 @@
...
+    unsafeRun(Gen.uniform.sample.map(_.value).runHead.get)
+
+  @Benchmark
+  def zioIntListOfSizeN: List[Int] =
```
adamgfraser: `List[List[Int]]`

benchmarks/src/main/scala/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,82 @@
...
+    unsafeRun(Gen.listOfN(size)(Gen.anyInt).sample.map(_.value).runHead.get)
+
+  @Benchmark
+  def zioStringOfSizeN: String =
```
adamgfraser: `List[String]`

benchmarks/src/main/scala/zio/GenBenchmarks.scala
pk044: oh, ok! when reading the issue description I thought we want to generate a single `Double` - I'll change that then :)

jdegoes: Yes I would add a parameter that defaults to 1000 and generate that many of each type.. WIll help amortize the cost of setup / teardown.

jdegoes: Also if you use `List` but construct in the wrong way (e.g. appending), you will end up measuring the performance of `List`. Probably if you rely on each library's facilities to generate a list of something (rather than collecting the results yourself), it will be fine.

pk044:
I'm not sure if it's possible to include `nyaya-gen` as of today - I don't see any version that would support all Scala versions (2.11, 2.12, 2.13) - is there any workaround or should I delete the benchmarks?

adamgfraser:
You should be able to move it to a `scala-2.12+` folder so the code is only available on Scala 2.12 or higher. Look in the `core` project for an example of this pattern.

pk044:
> You should be able to move it to a `scala-2.12+` folder so the code is only available on Scala 2.12 or higher. Look in the `core` project for an example of this pattern.

Unfortunately moving it to another folder is not working here - it fails on

```
[error] (benchmarks / update) sbt.librarymanagement.ResolveException: Error downloading com.github.japgolly.nyaya:nyaya-gen_2.11:0.9.0-RC1
[error]   Not found
[error]   Not found
[error]   not found: /home/circleci/.ivy2/local/com.github.japgolly.nyaya/nyaya-gen_2.11/0.9.0-RC1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/github/japgolly/nyaya/nyaya-gen_2.11/0.9.0-RC1/nyaya-gen_2.11-0.9.0-RC1.pom
[error]   not found: https://dl.bintray.com/hedgehogqa/scala-hedgehog/com.github.japgolly.nyaya/nyaya-gen_2.11/0.9.0-RC1/ivys/ivy.xml
[error] Total time: 1 s, completed Nov 12, 2019 10:06:01 AM
```

so it's more like we want to exclude `nyaya-gen` dependency when compiling against `2.11`, which I think is impossible?

adamgfraser:
Sorry, you can use conditional logic to add dependencies in the `build.sbt` file. For example:

```scala
      Seq("org.portable-scala" %%% "portable-scala-reflect" % "0.1.0") ++ {
        if (isDotty.value) Seq()
        else Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided")
      }
```

pk044:
benchmark results with current changes:
```
[info] Benchmark                                (elementSize)  (listSize)  Mode  Cnt            Score            Error  Units
[info] GenBenchmarks.hedgehogDoubles                     1000        1000  avgt   25       161597.680 ±       4188.702  ns/op
[info] GenBenchmarks.hedgehogIntListsOfSizeN             1000        1000  avgt   25    127013877.546 ±    8348101.063  ns/op
[info] GenBenchmarks.hedgehogStringsOfSizeN              1000        1000  avgt   25    310839636.294 ±   34653329.192  ns/op
[info] GenBenchmarks.nyayaDoubles                        1000        1000  avgt   25          423.444 ±          1.231  ns/op
[info] GenBenchmarks.nyayaIntListsOfSizeN                1000        1000  avgt   25        14819.416 ±        434.518  ns/op
[info] GenBenchmarks.nyayaStringsOfSizeN                 1000        1000  avgt   25        13953.223 ±         23.565  ns/op
[info] GenBenchmarks.scalaCheckDoubles                   1000        1000  avgt   25       127506.596 ±       1447.932  ns/op
[info] GenBenchmarks.scalaCheckIntListsOfSizeN           1000        1000  avgt   25    162623079.619 ±    1246259.761  ns/op
[info] GenBenchmarks.scalaCheckStringsOfSizeN            1000        1000  avgt   25       389201.768 ±       4747.312  ns/op
[info] GenBenchmarks.zioDouble                           1000        1000  avgt   25      7210370.053 ±      38679.580  ns/op
[info] GenBenchmarks.zioIntListsOfSizeN                  1000        1000  avgt   25  23329423925.480 ± 1839338377.172  ns/op
[info] GenBenchmarks.zioStringsOfSizeN                   1000        1000  avgt   25  23164799394.920 ± 2699075263.731  ns/op
[success] Total time: 7547 s (02:05:47), completed Nov 14, 2019 9:17:56 AM```

File: benchmarks/src/main/scala-2.12/zio/GenBenchmarks.scala [hunk 7]
```diff
@@ -0,0 +1,92 @@
...
+class GenBenchmarks {
+
+  @Param(Array("1000"))
+  var listSize: Int = _
+
+  @Param(Array("1000"))
+  var elementSize: Int = _
+  @Benchmark
+  def zioDouble: List[Double] =
+    unsafeRun(Gen.listOfN(listSize)(Gen.uniform).sample.map(_.value).runHead.get)
```
adamgfraser: I don't know if it will make much of a difference but instead of doing `Gen.listOfN` and then `runHead I think I would do `Gen.uniform.sample.map(_.value).take(1000).run`. In other words, generate doubles and generate 1000 of them instead of Generating just one big list. You can do the same thing with the other benchmarks.

benchmarks/src/main/scala-2.12/zio/GenBenchmarks.scala
```diff
@@ -0,0 +1,92 @@
...
+  def scalaCheckIntListsOfSizeN: List[List[Int]] =
+    scalacheck.Gen
+      .listOfN(listSize, scalacheck.Gen.listOfN(elementSize, scalacheck.Gen.choose(Int.MinValue, Int.MaxValue)))
+      .sample
+      .get
+
+  @Benchmark
+  def scalaCheckStringsOfSizeN: List[String] =
+    scalacheck.Gen
+      .listOfN(listSize, scalacheck.Gen.listOfN(elementSize, scalacheck.Gen.alphaChar).sample.mkString)
```
adamgfraser: Is this benchmark off because of the inner `sample` call? In the results you showed generating a string is dramatically faster than generating a list of integers of the same size which doesn't make sense.  I think it is because inside you are just generating one character instead of generating a new one each time. I think you want `.map(_.mkString)` instead of `.sample.mkString`.

jdegoes:
@pk044 @adamgfraser Looks pretty good to me!

We have some optimization to do, clearly. 😉 

Although ZIO's Gen will never be as fast as some of the others, because it offers:

 - Pure effectful generation, e.g. you could pull back test data from databases, read them from files, etc.
 - Auto-shrinking like Hedgehog

That said, through additional laziness and some performance optimization, we should be able to get _closer_ to Hedgehog, despite the fact that we have effectful generation.

We could also do something clever around reusing past generated data and using the `Gen` data only on failure (for shrinking) or on change of the `Gen`.

PR #2087, opened 2019-11-02
Adapt `ZIO.someOrFailException` to improve interop with error types

mlangc:
This PR adapts the signature of `ZIO.someOrFailException` to better work with commonly used error types like `Throwable`. For example, without this PR,
```scala
Task(Option(3)).someOrFailException
```
does not compile, but with this PR, it does.

Something that does not yet compile is `(io: IO[E, Option[A]]).someOrFailException` if `E` is not a supertype of `NoSuchElementException`, like `(io: IO[IllegalArgumentException, Option[A]]).someOrFailException`. Ideally, this should work, with the error type of the result being the most specific common supertype of `E` and `NoSuchElementException`, but I didn't yet find a way to achieve this.

In any way, this PR is an improvement over the existing state of affairs.


iravid:
Thanks @mlangc!

PR #2089, opened 2019-11-02
Improve onDecision signature

### LINKED ISSUE #2053: Revise ZSchedule#onDecision
Current signature:
```scala
def onDecision[A1 <: A](f: (A1, self.State) => UIO[Unit]): ZSchedule[R, A1, B]
```

Should be:
```scala
def onDecision[A1 <: A](f: (A1, Option[self.State]) => URIO[R, Any]): ZSchedule[R, A1, B]
```

YuvalItzchakov:
Resolves #2053

### DIFF
--- a/core/shared/src/main/scala/zio/ZSchedule.scala
+++ b/core/shared/src/main/scala/zio/ZSchedule.scala
@@ -496,8 +496,15 @@ trait ZSchedule[-R, -A, +B] extends Serializable { self =>
    * for every decision of this schedule. This can be used to create schedules
    * that log failures, decisions, or computed values.
    */
-  final def onDecision[A1 <: A](f: (A1, self.State) => UIO[Unit]): ZSchedule[R, A1, B] =
-    updated(update => (a, s) => update(a, s).tap(step => f(a, step)))
+  final def onDecision[A1 <: A, R1 <: R](f: (A1, Option[self.State]) => URIO[R1, Any]): ZSchedule[R1, A1, B] =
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZSchedule.scala
```diff
@@ -496,8 +496,8 @@ trait ZSchedule[-R, -A, +B] extends Serializable { self =>
    * for every decision of this schedule. This can be used to create schedules
    * that log failures, decisions, or computed values.
    */
-  final def onDecision[A1 <: A](f: (A1, self.State) => UIO[Unit]): ZSchedule[R, A1, B] =
-    updated(update => (a, s) => update(a, s).tap(step => f(a, step)))
+  final def onDecision[A1 <: A, R1 <: R](f: (A1, Option[self.State]) => URIO[R1, Any]): ZSchedule[R1, A1, B] =
+    updated(update => (a, s) => update(a, s).tap(step => f(a, Some(step))))
```
jdegoes: Why change to `Option`? It looks like you are always wrapping into `Some`?

The expansion to allow use of `R` in the updater is good.

iravid: The idea is to `fold` the result of `update` so that a `fail(())` results in `None` and a success results in `Some`

iravid: @YuvalItzchakov ^^

PR #2091, opened 2019-11-02
Reimplement RaceWith in the ZIO runloop

### LINKED ISSUE #1844: ZIO.race performance problem
According to `IOEmptyRaceBenchmark` in `benchmarks`, `ZIO.race` is currently noticeably slower than Monix and Cats Effect IO Race's. On my machine:
```
[info] Benchmark                                (size)   Mode  Cnt    Score     Error  Units
[info] IOEmptyRaceBenchmark.catsEmptyRace         1000  thrpt   10  315.196 ± 398.678  ops/s
[info] IOEmptyRaceBenchmark.monixEmptyRace        1000  thrpt   10  192.894 ±   7.672  ops/s
[info] IOEmptyRaceBenchmark.zioEmptyRace          1000  thrpt   10   33.152 ±   0.572  ops/s
[info] IOEmptyRaceBenchmark.zioTracedEmptyRace    1000  thrpt   10   25.166 ±   1.548  ops/s
```

Sampled via `sbt 'benchmarks / jmh:run zio.IOEmptyRaceBenchmark'`

This makes a difference in workloads with a huge amount of `race`'s, such as `fs2.Stream.parJoin` – https://github.com/zio/zio/issues/1441

YuvalItzchakov:
Resolves #1844 by lifting the RaceWith to the run loop.

Original benchmark by @neko-kai

```
info] Benchmark                                (size)   Mode  Cnt    Score     Error  Units
[info] IOEmptyRaceBenchmark.catsEmptyRace         1000  thrpt   10  315.196 ± 398.678  ops/s
[info] IOEmptyRaceBenchmark.monixEmptyRace        1000  thrpt   10  192.894 ±   7.672  ops/s
[info] IOEmptyRaceBenchmark.zioEmptyRace          1000  thrpt   10   33.152 ±   0.572  ops/s
[info] IOEmptyRaceBenchmark.zioTracedEmptyRace    1000  thrpt   10   25.166 ±   1.548  ops/s
```

After a code review with @jdegoes and further improvements, we've even improved more! (almost 5x)

```
[info] Benchmark                                (size)   Mode  Cnt    Score    Error  Units
[info] IOEmptyRaceBenchmark.catsEmptyRace         1000  thrpt   10  279.041 ± 10.432  ops/s
[info] IOEmptyRaceBenchmark.monixEmptyRace        1000  thrpt   10  290.857 ±  5.608  ops/s
[info] IOEmptyRaceBenchmark.zioEmptyRace          1000  thrpt   10  149.221 ±  1.637  ops/s
[info] IOEmptyRaceBenchmark.zioTracedEmptyRace    1000  thrpt   10   96.610 ±  0.357  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1502,12 +1502,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       }
 
     val g = (b: B, a: A) => f(a, b)
-    (self raceWith that)(coordinate(f, true), coordinate(g, false)).fork.flatMap { f =>
-      f.await.flatMap { exit =>
-        if (exit.succeeded) f.inheritFiberRefs *> ZIO.done(exit)
-        else ZIO.done(exit)
-      }
-    }
+    (self raceWith that)(coordinate(f, true), coordinate(g, false))
   }
 
   /**
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -924,57 +924,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Returns an effect that races this effect with the specified effect, calling
    * the specified finisher as soon as one result or the other has been computed.
    */
-  final def raceWith[R1 <: R, E1, E2, B, C](
-    that: ZIO[R1, E1, B]
-  )(
+  final def raceWith[R1 <: R, E1, E2, B, C](that: ZIO[R1, E1, B])(
     leftDone: (Exit[E, A], Fiber[E1, B]) => ZIO[R1, E2, C],
     rightDone: (Exit[E1, B], Fiber[E, A]) => ZIO[R1, E2, C]
-  ): ZIO[R1, E2, C] = {
-    def arbiter[E0, E1, A, B](
-      f: (Exit[E0, A], Fiber[E1, B]) => ZIO[R1, E2, C],
-      winner: Fiber[E0, A],
-      loser: Fiber[E1, B],
-      leftWins: Boolean,
-      raceDone: Ref[Boolean],
-      inherit: Ref[Option[Boolean]],
-      done: Promise[E2, C]
-    )(res: Exit[E0, A]): URIO[R1, Any] = {
-
-      val handleRes =
-        winner.poll.flatMap {
-          case Some(Exit.Success(_)) => winner.inheritFiberRefs
-          case _                     => ZIO.unit
-        } *> (f(res, loser) <* inherit.set(Some(leftWins))).to(done)
-
-      ZIO.flatten(raceDone.modify(b => (if (b) ZIO.unit else handleRes) -> true))
-    }
-
-    for {
-      done     <- Promise.make[E2, C]
-      raceDone <- Ref.make[Boolean](false)
-      inherit  <- Ref.make(None: Option[Boolean])
-      c <- ZIO.uninterruptibleMask { restore =>
-            for {
-              left  <- ZIO.interruptible(self).forkInternal
-              right <- ZIO.interruptible(that).forkInternal
-              left2 <- left.await
-                        .flatMap(arbiter(leftDone, left, right, true, raceDone, inherit, done))
-                        .forkInternal
-              right2 <- right.await
-                         .flatMap(arbiter(rightDone, right, left, false, raceDone, inherit, done))
-                         .forkInternal
-
-              inheritFiberRefs = inherit.get.flatMap {
-                case None        => ZIO.unit
-                case Some(true)  => left2.inheritFiberRefs
-                case Some(false) => right2.inheritFiberRefs
-              }
-
-              c <- restore(done.await <* inheritFiberRefs).onInterrupt(left.interrupt *> right.interrupt)
-            } yield c
-          }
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1539,8 +1500,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
               else ZIO.halt(loserCause && cause)
           }
       }
+
     val g = (b: B, a: A) => f(a, b)
-    (self raceWith that)(coordinate(f, true), coordinate(g, false))
+    (self raceWith that)(coordinate(f, true), coordinate(g, false)).fork.flatMap { f =>
+      f.await.flatMap { exit =>
```
jdegoes: This looks like new functionality—isn't `raceWith` supposed to do the inheritance?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -206,6 +206,58 @@ private[zio] final class FiberContext[E, A](
...
+            case exit0: Exit.Success[Exit[_, _]] =>
+              complete(right, left, race.rightWins, exit0.value, raceIndicator, cb)
+            case exit: Exit.Failure[_] => complete(right, left, race.rightWins, exit, raceIndicator, cb)
+          }
+
+          if (rightRegister ne null)
+            complete(right, left, race.rightWins, rightRegister, raceIndicator, cb)
+        }
+      }
+      .onInterrupt(left.interrupt *> right.interrupt)
```
jdegoes: This will get faster after the supervision work I'm doing. 

Note you can do a little micro-optimization, replacing `left.interrupt *> right.interrupt` with `left.interrupt.flatMap(_ => right.interrupt)`.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -522,6 +574,9 @@ private[zio] final class FiberContext[E, A](
 
                   curZio = nextInstr(result)
 
+                case ZIO.Tags.RaceWith =>
+                  val zio = curZio.asInstanceOf[ZIO.RaceWith[Any, Any, Any, Any, Any, Any, Any]]
```
jdegoes: You probably don't need this cast because `raceWithImpl` is polymorphic.

core/shared/src/main/scala/zio/internal/StackBool.scala
```diff
@@ -96,14 +96,20 @@ private[zio] final class StackBool private () {
...
+  def apply(bool: Boolean): StackBool = {
     val stack = StackBool()
 
-    bools.reverse.foreach(stack.push(_))
+    stack.push(bool)
 
     stack
   }
 
+  def apply(bools: List[Boolean]): StackBool = {
```
jdegoes: This one should be called `fromIterable` and should take `Iterable[Boolean]` to match ZIO conventions.

neko-kai:
Still slower than competition... @YuvalItzchakov Do you think you could profile to find out what's really causing this?

<sub>and why am i @kaichaosun again? ._.</sub>

core/shared/src/main/scala/zio/ZIO.scala
YuvalItzchakov: This was from a previous implementation I did which was not on the run loop, I forgot to remove it.

core/shared/src/main/scala/zio/internal/FiberContext.scala
YuvalItzchakov: `raceWithImpl` needs a `RaceWith` instance. `curZio` would be `IO[E, Any]`

core/shared/src/main/scala/zio/ZIO.scala
YuvalItzchakov: Without this fix, a test in `FiberRefSpec` fails:

```scala
testM("nothing gets inherited with a failure in zipPar") {
            for {
              success  = fiberRef.set(update)
              failure1 = fiberRef.set(update1) *> ZIO.fail(":-(")
              failure2 = fiberRef.set(update2) *> ZIO.fail(":-O")
              _        <- success.zipPar(failure1.zipPar(failure2)).orElse(ZIO.unit)
              value    <- fiberRef.get
            } yield assert(value, equalTo(initial))
          }
```
I wasn't sure breaking existing behavior was the way to go. This test assumes that if one of fibers fail, we don't carry the fiberRef the successful either, even if it won the race. What do you think should be done here?

YuvalItzchakov:
Sorry @neko-kai, my bad on the wrong handle. It is still slower then competition, but I couldn't think of any way to further micro optimize. I talked to @jdegoes about this in the Berlin Hackathon and we agreed that an optimization to `FiberContext` is in order and that will also squeeze in more peformance juice.

I will profile this further and see if I can pinpoint anything specific.

core/shared/src/main/scala/zio/ZIO.scala
neko-kai: @YuvalItzchakov zip and race are different operations. Failure in zip cancels the computation and prevents ref inheritance – there is no "winner" in a failed zip.

YuvalItzchakov: @neko-kai As long as zip uses race underneath, it's going to need to apply these additional semantics. That's why I added added this change to compensate.

PR #2093, opened 2019-11-02
Removed for-comprehensions from TSemaphore

### LINKED ISSUE #2082: Remove for-comprehensions from TSemaphore
All for-comprehensions used internally should be de-sugared.

kellydavid:
fix #2082

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TSemaphore.scala
+++ b/core/shared/src/main/scala/zio/stm/TSemaphore.scala
@@ -20,12 +20,13 @@ class TSemaphore private (val permits: TRef[Long]) extends AnyVal {
   final def acquire: STM[Nothing, Unit] = acquireN(1L)
 
   final def acquireN(n: Long): STM[Nothing, Unit] =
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/stm/TSemaphore.scala
```diff
@@ -20,12 +20,13 @@ class TSemaphore private (val permits: TRef[Long]) extends AnyVal {
   final def acquire: STM[Nothing, Unit] = acquireN(1L)
 
   final def acquireN(n: Long): STM[Nothing, Unit] =
-    (for {
-      _     <- assertNonNegative(n)
-      value <- permits.get
-      _     <- STM.check(value >= n)
-      _     <- permits.set(value - n)
-    } yield ())
+    assertNonNegative(n).flatMap { _ =>
```
mijicd: You can probably save some typing by replacing `flatMap { _ => ... }` with `*>`, other than that it looks good.

PR #2101, opened 2019-11-02
Add mapMPartitioned to ZStream

alexvanolst:
When performing operations on a stream, it is quite a common use case to want to perform some operation concurrently on elements of the stream while maintaining the order of messages within some partition.

This is currently achievable using the `ZStream.GroupBy` but this is not immediately obvious to new users.

This PR proposes adding a `mapMPartitioned` method to the `ZStream` class to cover this.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1649,7 +1649,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    */
   final def mapMPartitioned[R1 <: R, E1 >: E, B, K](
     keyBy: A => K,
-    buffer: Int
+    buffer: Int = 16
   )(f: A => ZIO[R1, E1, B]): ZStream[R1, E1, B] =
     groupByKey(keyBy, buffer).apply { case (_, s) => s.mapM(f) }
 
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1643,9 +1643,10 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
 
   /**
    * Maps over elements of the stream with the specified effectful function,
-   * partitioned by `p` executing invocations of `f` concurrently on up
-   * to `buffer` partitions. Transformed elements may be reordered but the order within
-   * a partition is maintained.
+   * partitioned by `p` executing invocations of `f` concurrently. The number
+   * of concurrent invocations of `f` is determined by the number of different
+   * outputs of type `K`. Up to `buffer` elements may be buffered per partition.
+   * Transformed elements may be reordered but the order within a partition is maintained.
    */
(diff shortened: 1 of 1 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1641,6 +1641,17 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   final def mapMParUnordered[R1 <: R, E1 >: E, B](n: Int)(f: A => ZIO[R1, E1, B]): ZStream[R1, E1, B] =
     self.flatMapPar[R1, E1, B](n)(a => ZStream.fromEffect(f(a)))
 
+  /**
+   * Maps over elements of the stream with the specified effectful function,
+   * partitioned by `p` executing invocations of `f` concurrently on up
+   * to `n` partitions. Transformed elements may be reordered but the order within
+   * a partition is maintained.
+   */
+  final def mapMPartitioned[R1 <: R, E1 >: E, B, K](
```
iravid: this signature would be nicer,  I think:
```scala
def mapMPartitioned(keyBy: A => K, buffer: Int = 16)(f: A => ZIO[...])
```

alexvanolst: Sure, have updated it

mschuwalow:
Agreed this covers the common use case and is quite intuitive I think.
@alexvanolst nice addition!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1641,6 +1641,18 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   final def mapMParUnordered[R1 <: R, E1 >: E, B](n: Int)(f: A => ZIO[R1, E1, B]): ZStream[R1, E1, B] =
     self.flatMapPar[R1, E1, B](n)(a => ZStream.fromEffect(f(a)))
 
+  /**
+   * Maps over elements of the stream with the specified effectful function,
+   * partitioned by `p` executing invocations of `f` concurrently on up
+   * to `buffer` partitions. Transformed elements may be reordered but the order within
```
iravid: One last comment: `buffer` does not prescribe how many partitions; that is prescribes by however many different values of `K` are returned from `keyBy`.

The buffer describes the amount of elements that can be buffered in each partition.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1641,6 +1641,18 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+   * a partition is maintained.
+   */
+  final def mapMPartitioned[R1 <: R, E1 >: E, B, K](
+    keyBy: A => K,
+    buffer: Int
```
```suggestion
    buffer: Int = 16
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
alexvanolst: Thanks, good catch. I have updated this to be more correct about the buffer and tried to be more explicit about how the partition function affects how many concurrent invocations of `f` there are.

Issue #2084, opened 2019-11-01
Ensure all factories return STM


mijicd:
Make sure all structures are created in STM, e.g. in `TSet`:

```scala
object TSet {
  final def apply[A](data: A*): STM[Nothing, TSet[A]] = ...

  // other methods
}
```

Implementers should prefer `apply` to `make` and provide additional factories where applicable (as in `TMap` and `TSat`). The issue can be split among multiple persons, in which case the following checklist apply:

- [ ] `TArray`
- [x] `TPromise`
- [x] `TQueue`
- [x] `TRef`
- [x] `TSemaphore`

To avoid duplicate work, please state what you intend to pick in the comments.

smart-ass:
I take TArray

vitaly-pushkar:
I take TRef

env0der:
I take TQueue

vitaly-pushkar:
@mijicd `TRef` seems to already have a `make` that returns `STM`. Is it relevant?

mijicd:
@vitaly-pushkar let's just rename it to `apply`. I'd leave `makeCommit` as it is.

env0der:
@mijicd here is `TQueue` part: https://github.com/zio/zio/pull/2103

env0der:
@mijicd and same for `TSemaphore` and `TPromise`: https://github.com/zio/zio/pull/2104

PR #2103, opened 2019-11-02
TQueue: rename make to apply #2084

env0der:


mijicd:
Nice work, will merge once build is green.

PR #2113, opened 2019-11-02
Add new shorthands for running ZStream

### LINKED ISSUE #1491: Add more shorthands for running streams
```scala
class ZStream[R, E, A] {
  def head: ZIO[R, E, Option[A]]
  def last: ZIO[R, E, Option[A]]
}
```

alexvanolst:
Resolves #1491 

Since there was an "open" question on the behaviour of `head` two variants were added.

`head` which is equivalent to take(1) and is safe for infinite streams
`first` which consumes the entire stream (i.e. the inverse of last)

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1910,6 +1910,22 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
     }
 
   /**
+   * Returns the first element of the stream.
+   */
+  def head: ZIO[R, E, Option[A]] = take(1).runCollect.map(_.headOption)
+
+  /**
+   * Runs the stream returning the first element and discarding all elements thereafter.
+   */
+  def first: ZIO[R, E, Option[A]] =
+    run(ZSink.foldLeft[A, Option[A]](None) { case (s, a) => if (s.isDefined) s else Some(a) })
+
+  /**
+   * Runs the stream returning the last element and discarding all earlier elements.
+   */
+  def last: ZIO[R, E, Option[A]] = run(ZSink.foldLeft[A, Option[A]](None) { case (_, a) => Some(a) })
+
+  /**
    * Runs the stream and collects all of its elements in a list.
    *
    * Equivalent to `run(Sink.collectAll[A])`.
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1915,12 +1915,6 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   def head: ZIO[R, E, Option[A]] = take(1).runCollect.map(_.headOption)
 
   /**
-   * Runs the stream returning the first element and discarding all elements thereafter.
-   */
-  def first: ZIO[R, E, Option[A]] =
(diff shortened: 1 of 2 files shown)


adamgfraser:
Just my two cents `first` seems like a dangerous operation. I don't know if we should have a shortcut for it.

alexvanolst:
I wouldn't say I'm sold on it either. @iravid Since you raised the question on whether it should have that behaviour, what do you think?

iravid:
Yes, I'm not sold either. Let's drop `first`. Anyone who figures out they need it can probably implement it manually :-)

Could you add tests for `head` and `last`?

regiskuckaertz:
It seems to me these should go into `ZSink` instead no?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1909,6 +1909,22 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       }
     }
 
+  /**
+   * Returns the first element of the stream.
+   */
+  def head: ZIO[R, E, Option[A]] = take(1).runCollect.map(_.headOption)
```
```suggestion
  def runHead: ZIO[R, E, Option[A]] = take(1).runCollect.map(_.headOption)
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1909,6 +1909,22 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+  /**
+   * Runs the stream returning the first element and discarding all elements thereafter.
+   */
+  def first: ZIO[R, E, Option[A]] =
+    run(ZSink.foldLeft[A, Option[A]](None) { case (s, a) => if (s.isDefined) s else Some(a) })
+
+  /**
+   * Runs the stream returning the last element and discarding all earlier elements.
+   */
+  def last: ZIO[R, E, Option[A]] = run(ZSink.foldLeft[A, Option[A]](None) { case (_, a) => Some(a) })
```
```suggestion
  def runLast: ZIO[R, E, Option[A]] = run(ZSink.foldLeft[A, Option[A]](None) { case (_, a) => Some(a) })
```

alexvanolst:
@regiskuckaertz I don't think `head` makes too much sense as part of `ZSink`.
For `last` we could add it to `ZSink` and keep the shorthand using it.

iravid:
Yeah, `ZSink.head` and `ZSink.last` would be useful. I'll add another ticket for those.

iravid:
As currently implemented, `head` wouldn't make sense on ZSink, no. It can be implemented using `run` and a `ZSink` that keeps the first element. No need to do it in this PR though.

alexvanolst:
I've added the tests. @iravid Feel free to assign / ping me when you create that other ticket.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1909,6 +1909,16 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       }
     }
 
+  /**
+   * Runs the stream returning the first element of the stream. Later elements will not be consumed / have effects run
+   */
+  def runHead: ZIO[R, E, Option[A]] = take(1).runCollect.map(_.headOption)
```
simpadjo: This implementation collects entire stream into a list. It can run out of memory. 
Should it be `ZSink.foldLeft(None){ case (a, _) => a }` ?

alexvanolst: The entire stream is at most 1 element after `take(1)` so this cannot run out of memory and does work infinite streams

PR #2114, opened 2019-11-02
Add withFilter in ZIO

wi101:
That enables to extract tuples or check conditions in io values that fails with Exception 
if the condition is not satisfied it fails with `NoSuchElementException`


neko-kai:
+1 on this. One could argue that `better-monadic-for` is a better implementation of the same, however with the dotty portability situation and the general no-dependencies spirit of zio I think we're better off supporting the standard of doing this too 👍

PR #2115, opened 2019-11-02
Hand specialized Chunk for primitive value types

vasilmkd:
I'm sorry that I'm polluting the Chunk companion object namespace, but unfortunately I fell victim to the https://github.com/scala/bug/issues/2034 bug.

Here is a proof that this works.
https://imgur.com/a/iUDc9t2

The singletons need some more work. Currently most of the operations such as `map, flatMap, filter` on chunks allocate an array, so they turn 1 element arrays into `Arr` instances, instead of `Singletons`, rendering `Singletons` not as useful IMO. I also had to remove `Singleton` creation from `Chunk.apply(as: A*)` because it would always box to `RefSingleton`.

@jdegoes

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -684,7 +684,7 @@ object Chunk {
    */
   final def fromIterable[A](it: Iterable[A]): Chunk[A] =
     if (it.size <= 0) Empty
-    else if (it.size == 1) Singleton(it.head)
+    else if (it.size == 1) single(it.head)
     else {
       it match {
         case l: Vector[A] => VectorChunk(l)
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -719,7 +797,9 @@ object Chunk {
     case x: VectorChunk[A] => x.classTag
   }
 
-  private case class Arr[A](private val array: Array[A]) extends Chunk[A] {
+  private sealed trait Arr[A] extends Chunk[A] {
```
iravid:
```suggestion
  private sealed abstract class Arr[A] extends Chunk[A] {
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -975,19 +1122,87 @@ object Chunk {
     override def toArray[A1](implicit tag: ClassTag[A1]): Array[A1] = Array.empty
   }
 
-  private case class Singleton[A](a: A) extends Chunk[A] {
+  private sealed trait Singleton[A] extends Chunk[A] {
```
iravid:
```suggestion
  private sealed abstract class Singleton[A] extends Chunk[A] {
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -928,12 +1008,79 @@ object Chunk {
...
 
+    override def toArray[A1 >: A](n: Int, dest: Array[A1]): Unit =
+      Array.copy(array, 0, dest, n, length)
+  }
+
+  private object Arr {
+    def apply[A](array: Array[A]): Arr[A] = {
+      val clss = array.getClass().getComponentType()
+      val res =
+        if (clss == java.lang.Boolean.TYPE) BooleanArr(array.asInstanceOf[Array[Boolean]])
```
iravid: `java.lang.Boolean/Byte/etc.` are the boxed versions of primitives. Are you sure this isn't boxing the data in the array?

vasilmkd: https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#TYPE

https://repl.it/repls/StridentHealthyVirus

vasilmkd:
No love for traits.

core/shared/src/main/scala/zio/Chunk.scala
iravid: Aha! TIL! Name was funny ;)

vasilmkd: Just Java things.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: Done.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: Done.

vasilmkd:
Pushed latest changes. @iravid

vasilmkd:
https://imgur.com/S8LNrO4

Logically "unboxing" from a Ref type to a Val type.

vasilmkd:
Closing because it is not the right way to solve this problem.

Issue #2077, opened 2019-11-01
Port combinators from ZIO to ZManaged


mschuwalow:
ZManageds public API still lacks behind ZIOs API and we should explore how many of the combinators on ZIO can also be defined on ZManaged. The first part of this issue would be to collect all functions that are currently missing and in a second step investigate each one of them.

saraiva132:
Methods that I identified are part of ZIO and not ZManaged. I am pretty sure I missed some or I just ignored them because it just does not even make sense to mention them here. I guess we can identify the ones that make more sense and create tickets for those and I can work on these straight away :)

- absorb
- absorbWith
- asError
- Bracket related
- cached
- catchAllCause
- catchSomeCause
- collect
- eventually
- filter related
- forever
- forkOn
- flattenErrorOption
- handleChildrenWith
- left/right leftOrFail/rightOrFail
- race related methods
- reject / rejectM
- repeat/ repeatOrElse / repeatOrElseEither
- retryOrElse / retryOrElseEither
- some / someOrFail
- summarized
- supervised
- tapBoth/ tapError
- timedWith
- timeoutFail / timeoutFork / timeoutTo
- to
- traced / tracingStatus / untraced
- uninterruptible
- void

mijicd:
I propose to open this ticket for multiple contributors. Each one of them would be able to tackle a single combinator. WDYT?

@mschuwalow @saraiva132 ^

saraiva132:
There is still work left to actually identify which ones make sense to implement in ZManaged. Are you suggesting that each separate ticket for each specific combinator would just be closed if it does not make sense or should that be discussed in the context of this issue?

mijicd:
No, just keep the task list in this issue, perhaps in you comment. People could refer to this issue (e.g. part of ...) in their PRs.

mijicd:
@iravid could you take a look at the list above?

mijicd:
Guys, please raise your hand to the combinator you're taking and refer to this issue in the PR, otherwise there'll be duplicated work :).

saraiva132:
@mijicd Guilty. yesterday I was not paying attention and did the same one. Should I update my issue up there with which ones are already done / being done ? Also at some point probably discard the ones that make no sense to port but I am unsure exactly which ones.

mijicd:
That would be great, just convert the list to checklist.

saraiva132:
@mijicd  done. If someone at some point can come up and tell which ones do not make sense to port so I check them it would be great.

[...]

PR #2116, opened 2019-11-02
Port absorb and absorbWith combinators to ZManaged

alexvanolst:
Part of #2077 

These were trivial to port over and seem to make sense to me.


iravid:
Would be great if you could add a test for `absorbWith` as well.

alexvanolst:
@iravid I've alphabetized the methods and added tests.

There were no existing tests for `absorbWith` on `ZIO` , so added those first. Had to also port `run` to ZManaged to be able to test on failures. 

Now we're in the same situation again where `run` has no tests on `ZIO` or `ZManaged` and I've not got much time left this weekend.

If the `run` on `ZManaged` seems reasonable to you can I suggest we open a ticket for all the missing tests on `ZIO` and `ZManaged`?

iravid:
@alexvanolst I tried to fix the conflicts but messed up the formatting. Can you run an fmt here? Then we can merge this. Great work here!

iravid:
Thank you @alexvanolst!

PR #2120, opened 2019-11-03
【WIP】Migrate checkspec

TapanVaishnav:


TapanVaishnav:
@unclebob418 please have a look when you have time

unclebob418:
@TapanVaishnav sorry I think this one is above my pay grade - @adamgfraser can you please take a look? I'd be happy to know the issue myself, as the approach @TapanVaishnav has taken is the same as I'd have done

unclebob418:
@TapanVaishnav one comment is you don't need two levels of suites there - I'm guessing you tried that based on  `StreamSpec` having two levels. I'm actually now trying to use `checkM` in `RandomSpec` - it'll be interesting to see if I have the same trouble.

TapanVaishnav:
> @TapanVaishnav one comment is you don't need two levels of suites there - I'm guessing you tried that based on `StreamSpec` having two levels. I'm actually now trying to use `checkM` in `RandomSpec` - it'll be interesting to see if I have the same trouble.

will try the same in my case and let you know.

unclebob418:
@TapanVaishnav, no issues on my part but I had some inspiration and I think I've partially solved your issue. I suspect one of your imports is bringing in something you don't want. I'll fiddle a bit more and see if I can get it sorted, but in the meantime, this should work to get through the rest of the tests:
```
testM("checkM is polymorphic in error type") {
      zio.test.checkM(Gen.int(1, 100)) { n =>
```

TapanVaishnav:
@adamgfraser @unclebob418 
> The fix would be to remove the CheckVariants file and move the check methods directly into the package object, but I haven't done that so far because it would clutter up the package object with a lot of methods and it only requires us to be slightly more verbose when writing these internal tests, which we should only have to do once.

Thank you so much for your insights and help. I'll progress using that. 
And about the `CheckVariatns`, as you mentioned, it's really troublesome to move in this PR as other files changes would also merge in so will create a new issue afterwards.

adamgfraser:
Yes. To be clear I would not make any changes to `CheckVariants` and just use `zio.test.check` instead of `check`.

TapanVaishnav:
@adamgfraser @unclebob418 
I've made the respective changes to migrate the `CheckSpec`.
All the tests of `CheckSpec` are passing but CI is failing at `BlockSpec`, which is beyond the scope of this PR.
So, Should I fix that(or wait for someone to do so) or will I still be able to merge this PR.

Also, regarding the above error `Implementation restriction: access of method checkM....`, I think it was because of the fact that most of the packages have the same signature `zio.test` and as you both mentioned, we have to explicitly write the pkg signature for the spec to find the corresponding method. In a nutshell, instead of prefixing each method with `zio.test`, I update the imports to achieve the same behavior.  
Please review the PR when you get time.

adamgfraser:
@TapanVaishnav Can you merge upstream changes? That should solve your problem with that flaky test.

test-tests/shared/src/test/scala/zio/test/CheckSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+        testM("overloaded check methods work") {
+          Check(Gen.anyInt, Gen.anyInt, Gen.anyInt) { (x, y, z) =>
+            assert((x + y) + z, equalTo(x + (y + z)))
+          }
+        },
+        testM("max shrinks is respected") {
+          val gen = Gen.listOfN(10)(Gen.int(-10, 10))
+          for {
+            ref <- Ref.make(0)
+            _ <- execute {
```
adamgfraser: You can simplify this by removing the `execute` and `testM` calls. Since `checkM` itself returns an effect with a test result you can just include that in your for comprehension directly.

test-tests/shared/src/test/scala/zio/test/CheckSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+          Check(Gen.anyInt) { a =>
+            assert(a, isGreaterThan(0))
+          }.map {
+            _.failures match {
+              case Some(BoolAlgebra.Value(details)) => details.gen.fold(false)(_.shrinkedInput == 0)
+              case _                                => false
+            }
+          }.map(assert(_, isTrue))
+        },
+        testM("implicationWorksCorrectly") {
```
adamgfraser: "implication works correctly"

test-tests/shared/src/test/scala/zio/test/CheckSpec.scala
TapanVaishnav: got it, will try, thanks

adamgfraser:
@TapanVaishnav Can you rebase? Right now it is a little hard to review because it is including all changes to `master` from the last few days.

TapanVaishnav:
> @TapanVaishnav Can you rebase? Right now it is a little hard to review because it is including all changes to `master` from the last few days.

Ah, I thought I did rebase, I guess it was miss from my side, sorry for the trouble.
I will rebase and update the code according to your comment and then will mention you again for the review.

adamgfraser:
@TapanVaishnav Thanks!

TapanVaishnav:
Thank you so much for the quick response though :)

TapanVaishnav:
@adamgfraser I was testing something and PR got closed, my bad(really really sorry), will create a new PR.

adamgfraser:
Can you just reopen?

TapanVaishnav:
I don't have access to reopen.

TapanVaishnav:
I can just create a new PR if that's not an issue.

adamgfraser:
Let me try to reopen.

adamgfraser:
@TapanVaishnav I have permissions to reopen but Github won't let me since you changed the base branch. Can you rebase on `master`?

TapanVaishnav:
> Can you rebase on master?
I don't have any permission, the only thing I can do it to delete this branch.

adamgfraser:
You should be able to do it through git on the command line locally and then push to your branch from there.

TapanVaishnav:
> You should be able to do it through git on the command line locally and then push to your branch from there.

I deleted my `migrate-checkspec` branch(local and forked remote) and create a new one(with the same name) with cleaner commit history. Hence, if I try to rebase my local with `upstream/master` , it does nothing as it's up-to-date.
I can create a new PR with the new branch and reference this closed PR in that(to leave the traces of comments).

adamgfraser:
You should be able to recover it. If you Google how to restore a deleted branch there are a lot of good resources for doing that.

unclebob418:
@TapanVaishnav I can help you with this tomorrow if you're stuck. Worst case scenario I've cloned your repo so you won't lose any work :)

PR #2124, opened 2019-11-03
Make parts of ZManaged private

### LINKED ISSUE #2076: Make parts of ZManaged private
This is necessary to evolve ZManaged's API in a backwards compatible way.

The following bits need to be changed:
- `ZManaged` itself needs to be a plain `final class` (and not a `case class`)
- The constructor needs to be hidden (`case class ZManaged[R, E, A] private ...`)
- An `apply(reservation: ZIO[R, E, Reservation[R, E, A]])` method should be provided on the companion object
- The `reserve: Reservation` constructor parameter should be made private (though I think that's the default once it's a plain class)
- A method `reserve: ZIO[R, E, Reservation[R, E, A]]` should be provided on the class

stkeky:
closes #2076

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -39,7 +39,7 @@ final case class Reservation[-R, +E, +A](acquire: ZIO[R, E, A], release: Exit[An
  * has been consumed, the resource will not be valid anymore and may fail with
  * some checked error, as per the type of the functions provided by the resource.
  */
-final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) { self =>
+final class ZManaged[-R, +E, +A] private (val reserve: ZIO[R, E, Reservation[R, E, A]]) { self =>
 
   /**
    * Symbolic alias for zip.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -776,6 +776,18 @@ final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]])
...
+  /**
+   * Creates new [[ZManaged]] from wrapped [[Reservation]].
+   */
+  final def apply[R, E, A](reservation: ZIO[R, E, Reservation[R, E, A]]): ZManaged[R, E, A] =
+    new ZManaged(reservation)
+
+  /**
+   * Extracts [[Reservation]] from within.
+   */
+  final def unapply[E, A](v: Managed[E, A]): Option[ZIO[Any, E, Reservation[Any, E, A]]] =
```
iravid: We can skip `unapply`. No real use for it

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -39,7 +39,7 @@ final case class Reservation[-R, +E, +A](acquire: ZIO[R, E, A], release: Exit[An
  * has been consumed, the resource will not be valid anymore and may fail with
  * some checked error, as per the type of the functions provided by the resource.
  */
-final case class ZManaged[-R, +E, +A](reserve: ZIO[R, E, Reservation[R, E, A]]) { self =>
+final class ZManaged[-R, +E, +A] private (val reserve: ZIO[R, E, Reservation[R, E, A]]) extends Serializable { self =>
```
iravid: Let's make `reserve` just a constructor parameter, and not a field:
```suggestion
final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, E, A]]) extends Serializable { self =>
```

And add an accessor method:
```scala
def reserve: ZIO[R, E, Reservation[R, E, A]] =  reservation
```

stkeky:
@iravid Thank you for your review, I addressed the issues.

PR #2125, opened 2019-11-03
Make ZStream chunkN return an instance of ZStreamChunk

saraiva132:


iravid:
Tests need a fix, otherwise this is good to go!

PR #2128, opened 2019-11-03
refactor RandomSpec to use Gen

unclebob418:
RandomSpec round 1 had some messy assertions which would be improved by using property based testing
#1898, #2037 

@adamgfraser I think I've cracked `forAllBounded` - before I tackle the others, can you please review to confirm this is what you're after? Some things I'm concerned about that you may be able to assist with:

1. is the state of testRandom ever going to update during this test? I tried to debug to find out but my breakpoints were being ignored for some reason.
2. `next` is asking for a `(Random.Service[Any], A) => UIO[A]` mainly because that's how I'm getting my `Random`. my gut feel is this should really be `TestRandom.Test` instead of `Random.Service[Any]`, because we're operating in the test environment but I couldn't get this to work. I guess I'm not using any features of Test that aren't part of the standard interface so perhaps this doesn't matter.
3. looks like `fmt` went to town with the new lines - is this intentional?

### DIFF
--- a/core/jvm/project/build.properties
+++ b/core/jvm/project/build.properties
@@ -1,1 +1,0 @@
-sbt.version=1.2.8
--- a/core/jvm/project/build.properties
+++ b/core/jvm/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=1.2.8
--- a/test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
@@ -1,32 +1,79 @@
 package zio.test.environment
 
+import zio._
+import zio.random.Random
 import zio.test.Assertion._
 import zio.test.environment.RandomSpecUtil._
-import zio.test.environment.TestRandom.DefaultData
-import zio.test.{ assert, suite, testM, TestResult, ZIOBaseSpec }
-import zio._
+import zio.test.environment.TestRandom.{ DefaultData, Test }
+import zio.test._
+
+import scala.util.{ Random => SRandom }
 
 object RandomSpec
     extends ZIOBaseSpec(
       suite("RandomSpec")(
-        testM("check clearBooleans")(checkClear(_.nextBoolean)(_.feedBooleans(_: _*))(_.clearBooleans)(_.nextBoolean)),
-        testM("check clearBytes")(checkClear(nextBytes(1))(_.feedBytes(_: _*))(_.clearBytes)(_.nextBytes(1))),
-        testM("check clearChars")(
-          checkClear(_.nextPrintableChar)(_.feedChars(_: _*))(_.clearChars)(_.nextPrintableChar)
+        testM("check clearBooleans")(
+          checkClear(_.nextBoolean)(_.feedBooleans(_: _*))(_.clearBooleans)(
+            _.nextBoolean
+          )
         ),
-        testM("check clearDoubles")(checkClear(_.nextDouble)(_.feedDoubles(_: _*))(_.clearDoubles)(_.nextDouble)),
-        testM("check clearFloats")(checkClear(_.nextFloat)(_.feedFloats(_: _*))(_.clearFloats)(_.nextFloat)),
-        testM("check clearInts")(checkClear(_.nextInt)(_.feedInts(_: _*))(_.clearInts)(_.nextInt)),
-        testM("check clearLongs")(checkClear(_.nextLong)(_.feedLongs(_: _*))(_.clearLongs)(_.nextLong)),
-        testM("check clearStrings")(checkClear(_.nextString(1))(_.feedStrings(_: _*))(_.clearStrings)(_.nextString(1))),
-        testM("check feedBooleans")(checkFeed(_.nextBoolean)(_.feedBooleans(_: _*))(_.nextBoolean)),
-        testM("check feedBytes")(checkFeed(nextBytes(1))(_.feedBytes(_: _*))(_.nextBytes(1))),
-        testM("check feedChars")(checkFeed(_.nextPrintableChar)(_.feedChars(_: _*))(_.nextPrintableChar)),
-        testM("check feedDoubles")(checkFeed(_.nextDouble)(_.feedDoubles(_: _*))(_.nextDouble)),
-        testM("check feedFloats")(checkFeed(_.nextFloat)(_.feedFloats(_: _*))(_.nextFloat)),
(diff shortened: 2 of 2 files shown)


core/jvm/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.2.8
```
adamgfraser: I don't think you want to change this.

unclebob418: I don't recall changing that *sigh* I've had so many issues with intellij today I've done almost nothing useful. Might be time to start using something else.

adamgfraser: SBT version is 1.3.3 on `master` so don't think we want to change to an older SBT version or really do anything with it at all as part of this PR, right?

adamgfraser:
This looks good. To respond to your questions:

1. The state is being updated. You can see it by adding a line like `_ < ZIO.effectTotal(println(nextRandom))` right before the `yield` clause in `forAllBounded`. Generally as long as you aren't calling `unsafeRun` or its equivalent the random state will automatically be updated after you get any value and you should never have to worry about that.

2. I think it is fine to just ask for a `Random.Service[Any]`. You aren't actually using any of the functionality of `TestRandom` in this method so principle of least power says only ask for what you need.

3. I don't know about the `fmt`. To my knowledge there weren't any changes to `scalaFmt` that would be driving this. The methods may have just gotten slightly longer to be above the line length? Maybe looks slightly nicer to use brackets after the test label `test("label") { ... }` but otherwise I wouldn't worry about it.

core/jvm/project/build.properties
adamgfraser: Sorry, didn't see your response so wasn't sure if you were confused about my comment or what happened.

unclebob418: I'll change it back. I have no idea how it even changed since I don't believe I've even opened that file. Still I should have been checking the commit screen more closely. Another lesson learnt.

unclebob418:
Thanks, I'll continue the rest tomorrow.

unclebob418:
@adamgfraser I've done `forAllEqualShuffle` but I'm a bit stuck for the rest of them. Most of them are checking that the values generated by the `testRandom` are the same as those generated by `SRandom`. If I'm using `Gen` instead of `testRandom`, I'm not sure how to set the seed to get results consistent with `SRandom`. I've tried using `Gen.fromRandom(testRandom)` but `testRandom` is a `TestRandom.Test` instead of the required `Random.Service[Any]`. Do you have any suggestions for this?

adamgfraser:
@unclebob418 You can do those by generating one random long using the property based testing framework, and then in the test you set the random in the environment to have that seed, create a `scala.util.Random` and set it with the same seed, generate values from each of them, and assert that those two values are equal. That way the testing framework will do that 200 times with a different random value each time.

unclebob418:
@adamgfraser I think I was overcomplicating things by trying to use Gen for too much (i.e. for all random values), forgetting that we're testing TestRandom here, not Gen. Can you please take a look? I also managed to break the timeout on the tests - I'll raise a separate issue shortly, but I suspect it might be related to the issues we were having with `BlockingSpec` here https://github.com/zio/zio/pull/2037#issuecomment-547218975

unclebob418:
@adamgfraser I've removed the impure code - hopefully good to go now!

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -1,49 +1,120 @@
 package zio.test.environment
 
+import zio._
+import zio.ZIO._
```
adamgfraser: I wouldn't do this import. It works okay here but in other cases can create issues because similar methods exist on, for example, `ZIO`, `ZManaged`, and `ZStream` so it can make it less clear what is being done.

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -1,49 +1,120 @@
...
-import zio.test.{ assert, suite, testM, TestResult, ZIOBaseSpec }
-import zio._
+import zio.test.environment.TestRandom.{ DefaultData, Test }
+import zio.test._
+
+import scala.util.{ Random => SRandom }
 
 object RandomSpec
     extends ZIOBaseSpec(
       suite("RandomSpec")(
```
adamgfraser: I think we can revert these changes and keep each test on one line.

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -55,117 +126,119 @@ object RandomSpec
...
+        sRandom    <- effectTotal(new SRandom(seed))
+        testRandom <- TestRandom.makeTest(DefaultData)
+        _          <- testRandom.setSeed(seed)
+        values     <- effectTotal(List.fill(100)(generate(sRandom)))
+        _          <- feed(testRandom, values)
+        results    <- UIO.foreach(List.range(0, 100))(_ => extract(testRandom))
+        random     <- extract(testRandom)
+      } yield {
+        assert(results, equalTo(values)) &&
+        assert(random, equalTo(generate(new SRandom(seed))))
```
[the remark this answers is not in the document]
unclebob418: the good news is I did know this! the bad news is I missed them :/

unclebob418:
@adamgfraser should be good now. I've now re-read the overview which is making a lot more sense now I've got my hands dirty with some of the code.

test-tests/shared/src/test/scala/zio/test/environment/RandomSpec.scala
```diff
@@ -55,117 +58,121 @@ object RandomSpec
...
-  }
+  def forAllEqualGaussian: ZIO[Random, Nothing, TestResult] =
+    checkM(Gen.anyLong) { seed =>
+      for {
+        sRandom    <- ZIO.effectTotal(new SRandom(seed))
+        testRandom <- TestRandom.makeTest(DefaultData)
+        _          <- testRandom.setSeed(seed)
+        actual     <- testRandom.nextGaussian
+        expected   <- ZIO.effectTotal(sRandom.nextGaussian)
+      } yield assert(math.abs(actual - expected), isLessThan(0.01))
```
adamgfraser: You can use `Assertion.approximatelyEquals` here.

unclebob418: much nicer!

PR #2138, opened 2019-11-03
Enable tests on Dotty

### LINKED ISSUE #2070: Run ZIO Core Test Suite on Dotty
Currently ZIO Core is being compiled against Dotty but we are not running the test suite on Dotty. This ticket is to enable that by changing the `testDottyJVM` command alias in `build.sbt`, get the test suite to compile on Dotty, and resolve any issues that result.

tenzki:
Closes #2070

Following tests have been ignored on Dotty build:

- `DurationSpec#10 ns * NaN = Infinity` - due to an issue with evaluating `(Double.NaN <= 0 || 10L <= 0)` on Dotty.
- `SerializableSpec#Chunk.empty is serializable` - due to an issue with deserializing `private case object` defined inside of an `object`.
- `SerializableSpec#Schedule is serializable`.
- `ZIOSpec#withFilter doesn't compile with IO that fails with type other than Throwable` - compiles on dotty


adamgfraser:
Thanks for working on this! A lot of great progress. It looks like there are some compilation errors on Scala 2.13. It looks like maybe you moved some files into the 2.12 folder that were needed for 2.13? If you want to make a file available to all Scala 2 versions but not Dotty you can use a directory called `scala-2.x` in the `shared` folder. If you need a Java specific file on all versions except Dotty you will need to add new logic for that to the `BuildHelper.scala` file. Let me know if you need any help doing that.

tenzki:
Using `scala-2.x` causes issue with finding `specs2` classes.

tenzki:
Using `scala-2.x` causes issue with finding `specs2` classes in `BaseCrossPlatformSpec`. Tests pass if `BaseCrossPlatformSpec` is placed in both `scala-2.12` and `scala-2.13` folders. Help how solve this would be appreciated.

adamgfraser:
@tenzki You need to delete [this]( https://github.com/zio/zio/blob/e078e4405e4e2091c94ffd232a06b2b6f3241dc5/project/BuildHelper.scala#L190) line and the identical ones on lines 199 and 209 in the `BuildHelper.scala` file and move them to the `Test` scope directly below. Then is should work.

core-tests/jvm/src/test/scala/zio/RTSSpec.scala
```diff
@@ -81,21 +81,25 @@ object RTSSpec
           assertM(io, isTrue)
         } @@ jvm(nonFlaky(100)),
         testM("deadlock regression 1") {
-          import java.util.concurrent.Executors
+          if (TestVersion.isDotty) {
```
adamgfraser: We just merged a PR to make tests more lazy so I think this should work with the test aspect now. Can you try after merging upstream changes?

core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
```diff
@@ -59,7 +60,7 @@ object DurationSpec
           },
           test("10 ns * NaN = Infinity") {
             assert(Duration.fromNanos(10L) * Double.NaN, equalTo(Duration.Infinity: Duration))
```
adamgfraser: Can you write up a quick summary of the issue here so we can raise an issue with the Dotty team if necessary?

tenzki:
Here's a simple example of the issue:
```
val a: Double = Double.NaN
val eval = a <= 0 || 10L <= 0
println(s"when Double.NaN is extracted to val, expression is evaluated to: $eval")
val eval2 = Double.NaN <= 0 || 10L <= 0
println(s"when Double.NaN is used directly, expression is evaluated to: $eval2")
```

adamgfraser:
Can you merge upstream changes one more time? We just renamed `ZSchedule` to `Schedule` so want to make sure we don't have any conflicts there. Also `ZIOSpecJVM` got migrated to ZIO Test so I think that can run on Dotty now.

adamgfraser:
@tenzki We also just removed Specs2 entirely so that should make things even easier.

Issue #1765, opened 2019-09-22
Make ZStream's combinators resilient to failures


iravid:
For issues such as #573 and #1711, we need to first audit the existing combinators and constructors and make sure they can handle being pulled again after a pull resulted in a failure.

After this is done, we can "loosen" `ZStream#process`'s and `Pull`'s contract to allow pulling again after a failure.

To verify that a combinator is resilient to failures, this is a good testcase (of course, after `++` itself is made resilient):
```scala
(Stream(1) ++ Stream.fromEffect(ZIO.fail("Ouch")) ++ Stream(2))
  .<combinator>
  .process
  .use { as =>
    for {
      a <- as
      _ <- as.run
      b <- as
    } yield (a, b)
  }
  .unsafeRun must_=== (1, 2)
```

vasilmkd:
I'd like to take a look at this, but please treat this as a tentative request. Anyone else interested can still take over.

vasilmkd:
Just a heads up, I'm experimenting with this issue, as discussed with @iravid.

StelaLeon:
I am giving this a try, it's my first issue, since I am getting used to the code base a bit more feel free to open/merge anything related to this (fyi @vasilmkd  :) )

vasilmkd:
Feel free to tag me in the PR, I will review it. I suggest you take a look at the `StreamPullSafetySpec.scala` file for inspiration and feel free to contact me on the Discord server, should you need any help.

vasilmkd:
Stream combinators that keep running state that have been made safe so far:

- [ ] `++` (`concat`)
- [ ] `absolve`
- [x] `aggregate`
- [ ] `aggregateAsync`
- [ ] `aggregateAsyncWithinEither`
- [ ] `aggregateAsyncWithin`
- [x] `aggregateManaged`
- [x] `buffer`
- [x] `bufferDropping`
- [x] `bufferSliding`
- [x] `bufferUnbounded`
- [ ] `catchAll`
- [ ] `catchAllCause`
- [ ] `chunkN`
- [ ] `collectWhile`
- [ ] `collectWhileM`
- [ ] `combine` (no state but would be nice to be tested)
- [ ] `drainFork` (would be nice to be tested)
- [x] `dropWhile`
- [ ] `either`
- [x] `flatMap`
- [ ] `flatMapPar`
- [ ] `flatMapParSwitch`
- [ ] `foreachWhileManaged`
- [ ] `groupBy`
- [ ] `haltWhen`
[...]

PR #2139, opened 2019-11-03
Make ZStream.effectAsync safe to pull again

vasilmkd:
Slowly but surely towards #1765.

@iravid


vasilmkd:
@iravid I'd appreciate your thoughts and if you could take a look at the failing tests, that would be great.

PR #2140, opened 2019-11-03
ZIO Test: Implement Overloaded Version of TestAspect#nonFlaky

adamgfraser:
`nonFlaky` is a commonly used test aspect. Currently the user has to specify the number of times for the test to be repeated. This PR adds an overloaded version that repeats a test a default number of times determined by the test framework (currently set at 100). I think this makes it a little nicer. Users can still specify a number of repetitions if they want.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -340,6 +340,13 @@ object TestAspect extends TimeoutVariants {
     if (TestPlatform.isJVM) identity else ignore
 
   /**
+   * An aspect that repeats the test a default number of times, ensuring it is
+   * stable ("non-flaky"). Stops at the first failure.
+   */
+  val nonFlaky: TestAspectPoly =
+    nonFlaky(100)
+
+  /**
    * An aspect that repeats the test a specified number of times, ensuring it
    * is stable ("non-flaky"). Stops at the first failure.
    */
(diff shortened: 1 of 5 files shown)


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -339,6 +339,13 @@ object TestAspect extends TimeoutVariants {
   val jvmOnly: TestAspectPoly =
     if (TestPlatform.isJVM) identity else ignore
 
+  /**
+   * An aspect that repeats the test a default number of times, ensuring it is
+   * stable ("non-flaky"). Stops at the first failure.
+   */
+  val nonFlaky: TestAspectPoly =
+    nonFlaky(100)
```
vasilmkd: Not that I don't appreciate this change, far from it, but why didn't you use default arguments, of course, in the original function declaration, like so:

`def nonFlaky(n0: Int = 100): TestAspectPoly = ???`

adamgfraser: If you do that then you would need to call the method as `nonFlaky()` instead of `nonFlaky`. That is somewhat irregular (no other test aspect is called that way and it gives the impression that it is possibly an impure method) so I think it is nicer to do it this way, especially when it is only a few extra lines of code.

vasilmkd: Well, you answered my question 😄

PR #2145, opened 2019-11-04
ZIO Test: Refactor ZSpec to Move All Test Failures to Error Channel

adamgfraser:
When we refactored suites to be effectual we moved to an encoding where assertion failures were represented by a `Left` value whereas runtime failures were represented captured in the error channel. While I thought it was necessary at the time it has created some issues with bugs in properly keeping track of the different failure modes and I think at this point it is hindering our ability to add other useful test aspects because the encoding of all test failures in the error channel lends itself very naturally to many ZIO combinators.

This PR refactors ZSpec to move all test failures back to the error channel. I was able to do it in a relatively minimalistic way. The only thing we are really giving up is that a `TestExecutor` now takes a `ZSpec` rather than a `Spec` because we need `ZSpec` to have the structure where `TestFailure` is in the error channel and we also need to know that `ExecutedSpec` has the same structure for reporting and interface with the SBT test runner. I think this is a pretty minor cost considering that to my knowledge no one has implemented any other `TestExecutor` instance and doing one that didn't use `ZSpec` would also mean losing out on test aspects and a lot of other ZIO Test functionality.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -37,7 +37,7 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
   final def @@[R0 <: R1, R1 <: R, E0, E1, E2 >: E0 <: E1, S0, S1, S >: S0 <: S1](
     aspect: TestAspect[R0, R1, E0, E1, S0, S1]
   )(implicit ev1: E <:< TestFailure[E2], ev2: T <:< TestSuccess[S]): ZSpec[R1, E2, L, S] =
-    aspect(self.bimap(ev1, ev2))
+    aspect(self.asInstanceOf[ZSpec[R1, E2, L, S]])
 
   /**
    * Returns a new spec with remapped errors and tests.
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -16,7 +16,7 @@
 
 package zio.test
 
-import zio.{ Cause, Managed, NeedsEnv, ZIO }
+import zio._
 
 import Spec._
 
@@ -34,10 +34,19 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
    * test("foo") { assert(42, equalTo(42)) } @@ ignore
    * }}}
    */
-  final def @@[R0 <: R1, R1 <: R, E0 >: E, E1 >: E0, S0 <: S, S, S1 >: S](aspect: TestAspect[R0, R1, E0, E1, S0, S1])(
-    implicit ev: T <:< Either[TestFailure[Nothing], TestSuccess[S]]
-  ): Spec[R1, E0, L, Either[TestFailure[Nothing], TestSuccess[S]]] =
-    aspect(mapTest(ev))
+  final def @@[R0 <: R1, R1 <: R, E0, E1, E2 >: E0 <: E1, S0, S1, S >: S0 <: S1](
+    aspect: TestAspect[R0, R1, E0, E1, S0, S1]
+  )(implicit ev1: E <:< TestFailure[E2], ev2: T <:< TestSuccess[S]): ZSpec[R1, E2, L, S] =
+    aspect(self.bimap(ev1, ev2))
+
+  /**
+   * Returns a new spec with remapped errors and tests.
+   */
+  final def bimap[E1, T1](f: E => E1, g: T => T1)(implicit ev: CanFail[E]): Spec[R, E1, L, T1] =
+    transform[R, E1, L, T1] {
+      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.mapError(f), exec)
+      case TestCase(label, test)         => TestCase(label, test.bimap(f, g))
+    }
 
   /**
    * Returns a new spec with the suite labels distinguished by `Left`, and the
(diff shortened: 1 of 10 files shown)


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -34,10 +34,19 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
    * test("foo") { assert(42, equalTo(42)) } @@ ignore
    * }}}
    */
-  final def @@[R0 <: R1, R1 <: R, E0 >: E, E1 >: E0, S0 <: S, S, S1 >: S](aspect: TestAspect[R0, R1, E0, E1, S0, S1])(
-    implicit ev: T <:< Either[TestFailure[Nothing], TestSuccess[S]]
-  ): Spec[R1, E0, L, Either[TestFailure[Nothing], TestSuccess[S]]] =
-    aspect(mapTest(ev))
+  final def @@[R0 <: R1, R1 <: R, E0, E1, E2 >: E0 <: E1, S0, S1, S >: S0 <: S1](
+    aspect: TestAspect[R0, R1, E0, E1, S0, S1]
+  )(implicit ev1: E <:< TestFailure[E2], ev2: T <:< TestSuccess[S]): ZSpec[R1, E2, L, S] =
```
jdegoes: While generally I like implicit evidence, given the cost of the bimap, may be worth pushing `@@` to be a syntax (extension method) in the companion object, which is added only to specs with `TestFailure` and `TestSuccess` as the S / T. Alternately we could consider specializing all of `Spec` to these types, but the more polymorphic version is nicer because most methods don't need to know about that structure.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -108,9 +108,9 @@ object TestAspect extends TimeoutVariants {
   val ignore: TestAspectPoly =
     new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
-        test: ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]
-      ): ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]] =
-        ZIO.succeed(Right(TestSuccess.Ignored))
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
```
jdegoes: Looks cleaner and more symmetrical now. ❤️

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -201,10 +196,10 @@ object TestAspect extends TimeoutVariants {
   val eventually: TestAspectPoly =
     new TestAspect.PerTest[Nothing, Any, Nothing, Any, Nothing, Any] {
       def perTest[R >: Nothing <: Any, E >: Nothing <: Any, S >: Nothing <: Any](
-        test: ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]]
-      ): ZIO[R, E, Either[TestFailure[Nothing], TestSuccess[S]]] = {
-        lazy val untilSuccess: ZIO[R, Nothing, Either[TestFailure[Nothing], TestSuccess[S]]] =
-          test.foldCauseM(_ => untilSuccess, _.fold(_ => untilSuccess, s => ZIO.succeed(Right(s))))
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
+      ): ZIO[R, TestFailure[E], TestSuccess[S]] = {
+        lazy val untilSuccess: ZIO[R, Nothing, TestSuccess[S]] =
```
jdegoes: This could be rewritten as `test.eventually`, now.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -238,17 +233,18 @@ object TestAspect extends TimeoutVariants {
...
-            case Left(e) => if (p.run(e).isSuccess) succeed else Left(TestFailure.Assertion(assert(e, p)))
-            case _       => Left(TestFailure.Runtime(Cause.die(new RuntimeException("did not fail as expected"))))
-          }
+            case testFailure if p.run(testFailure).isSuccess => succeed
+            case other =>
+              ZIO.fail(
+                TestFailure.Assertion(assert(other, p))
+              )
+          },
+          _ => ZIO.fail(TestFailure.Runtime(zio.Cause.die(new RuntimeException("did not fail as expected"))))
```
jdegoes: Looks much simpler.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -372,33 +368,12 @@ object TestAspect extends TimeoutVariants {
...
-          )
-
-        Live.live(schedule.initial).flatMap(loop)
-      }
+  ): TestAspect[Nothing, R0, Nothing, E0, Nothing, S0] =
+    new TestAspect.PerTest[Nothing, R0, Nothing, E0, Nothing, S0] {
+      def perTest[R >: Nothing <: R0, E >: Nothing <: E0, S >: Nothing <: S0](
+        test: ZIO[R, TestFailure[E], TestSuccess[S]]
+      ): ZIO[R, TestFailure[E], TestSuccess[S]] =
+        test.retry(schedule)
```
jdegoes: ❤️ ❤️ ❤️

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Yes!

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: If we are worried about performance could we use `asInstanceOf` since we have the implicit evidence?

I explored specialized `Spec` when I was working on this PR and I don't think we want to go down that road. If we specialize `Spec` on a type that has a error we lose the ability to set the error channel to `Nothing` in `ExecutedSpec` to express that those specs can't fail, which is quite important now that they are effectual to be able to safely traverse them.

Also we moved `@@` to `Spec` originally to solve type inference issues, so not to say we can't come up with another way but I'm worried we are going to take a step back there.

jdegoes: All right, cast sounds good to me as long as it is sound.

jdegoes:
❤️ Awesome cleanup! 💪

PR #2148, opened 2019-11-04
Ignore should not recover from unchecked exceptions

### LINKED ISSUE #2142: Ignore should not recover from unchecked exceptions
Recently in [this issue](https://github.com/zio/zio/issues/1868) we fixed `ZIO#option` to make it NOT handle unchecked exceptions. The point was that it should be explicit using something like `sandbox`.

I think the same reasoning applies to `ZIO#ignore`. What do you think?

alexvanolst:
Resolves #2142

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -121,14 +121,12 @@ object ZIOSpec
         ),
         suite("ignore")(
           testM("return success as Unit") {
-            import zio.CanFail.canFail
             assertM(ZIO.succeed(11).ignore, equalTo(()))
           },
           testM("return failure as Unit") {
             assertM(ZIO.fail(123).ignore, equalTo(()))
           },
           testM("not catch throwable") {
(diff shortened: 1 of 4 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -119,6 +119,19 @@ object ZIOSpec
             assertM(ZIO.fail("Fail").head.either, isLeft(isSome(equalTo("Fail"))))
           }
         ),
+        suite("ignore")(
+          testM("return success as Unit") {
+            import zio.CanFail.canFail
```
ghostdogpr: These imports aren’t needed (that’s what makes the CI fail).

alexvanolst: Thanks 👍 I had those in from adapting the `.option` tests

ghostdogpr:
The change breaks the tests “Invokes cleanups on interrupt“ in ZManagedSpec because they expect the old behavior. I think you can just adapt those tests.

PR #2149, opened 2019-11-04
#2077 - added collect and collectM to ZManaged

dkarlinsky:
#2077


saraiva132:
So I was not paying attention and also wrote these combinators. Code is exactly the same except my tests. https://github.com/zio/zio/pull/2154

Facepalm

Issue #2127, opened 2019-11-03
Add Chunk/ZStreamChunk#collectM and collectWhileM


iravid:

saraiva132:
Taking this one

PR #2150, opened 2019-11-04
Chunk  add collectM collectWhileM combinators

saraiva132:
I wonder if the Arr implementation of collectM and collectWhileM is OK like this or should not be done in terms of Array combinators but more low-level. @iravid 

Solves half of  #2127

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -842,7 +842,10 @@ object Chunk {
       var i      = 0
       var j      = 0
       var done   = false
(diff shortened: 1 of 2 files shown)


saraiva132:
@iravid I am quite unsure about how decent the triggering of the done flag in the collectWhileM method is. Do you propose anything better? It feels like a hack.

Also, I had mentioned about the ability to get a list or vector from a chunk and also added in the context of this PR. I am not sure about the builder function though. Let me know what you think

Last but not least there seems to be quite a lot of common behaviour across different methods so perhaps opening a clean-up ticket might be a good idea. I did not want to change anything else in the context of this PR

PR #2156, opened 2019-11-04
Adding hasFirst, hasLast and hasAt assertions

13h3r:


test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -72,6 +72,36 @@ object AssertionSpec
...
+        test("hasLast must fail when an iterable is empty") {
+          assert(Seq(), hasLast(anything))
+        } @@ failure,
+        test("hasLast must succeed when a last is equal to a specific assertion") {
+          assert(Seq(1, 2, 3), hasLast(equalTo(3)))
+        },
+        test("hasLast must fail when a last is not equal to specific assertion") {
+          assert(Seq(1, 2, 3), hasLast(equalTo(100)))
+        } @@ failure,
+        test("hasAt must fail when an idnex is outside of a sequence range") {
```
adamgfraser: Spelling on index.

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -72,6 +72,36 @@ object AssertionSpec
...
+          assert(Seq(1, 2, 3), hasAt(-1)(anything))
+        } @@ failure,
+        test("hasAt must fail when an idnex is outside of a sequence range 2") {
```
adamgfraser: Spelling on index.

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -72,6 +72,36 @@ object AssertionSpec
...
+          assert(Seq(1, 2, 3), hasAt(3)(anything))
+        } @@ failure,
+        test("hasAt must fail when head is not equal to a specific assertion") {
```
adamgfraser: I think you want to call this `hasAt`.

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -72,6 +72,36 @@ object AssertionSpec
...
+          assert(Seq(1, 2, 3), hasAt(1)(equalTo(100)))
+        } @@ failure,
+        test("hasAt must succeed when a head is equal to a specific assertion") {
```
adamgfraser: You want to have at least one test that succeeds, right?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -326,6 +326,25 @@ object Assertion {
       assertion.test(actual.size)
     }
 
+  final def hasFirst[A](assertion: Assertion[A]): Assertion[Iterable[A]] =
```
adamgfraser: When you construct an assertion that takes another assertion as a parameter, you want to use `assertionRec` so that the sub-assertion and any further sub-assertions it contains are recursively embedded in the assertion result so we get nice hierarchical reporting of test failures.

13h3r:
I changed `assertion` to `assertionRec`. Should we change the tests in a way that they will check this? Not in this PR, but in general.

Could you please also tell what is the proper way to upload new changes? Just new commit in a branch or force push to my branch? I did the last one.

Thanks!

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
13h3r: Fixed

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
13h3r: Fixed

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
13h3r: Fixed

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
13h3r: Right! I created a new one

test/shared/src/main/scala/zio/test/Assertion.scala
13h3r: Switched to `assertionRec`

adamgfraser:
@13h3r Great idea on adding a helper function to check for this. We have a test for it in general [here](https://github.com/zio/zio/blob/4d97e8874f3b4bbb7800082497649dfef00bfab9/test-tests/shared/src/main/scala/zio/test/DefaultTestReporterSpec.scala#L22) but it is more of an integration test to check that the `DefaultTestReporter` reports it correctly. Would be nice to have a method you could use with any assertion that takes another assertion as a parameter to check that the sub-assertion is properly embedded. Do you want to do this here or as a follow up?

Regarding your other question, just pushing a new commit to your branch is usually best.

13h3r:
* Tests improvement. I would like to do this outside of the CR.
* Alphabetize. I am making an assumption that it is about sorting members in lexicographical order. Done. Is it any way to do this with `scalafmt`?

adamgfraser:
Sounds good on doing the test improvements separately.

Ordering looks great now. Unfortunately we haven't found a good way to automate this because if there is a `val` then naively reordering can result in a forward reference error.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -293,6 +293,19 @@ object Assertion {
       BoolAlgebra.success
     )
 
+  /**
+   * Makes a new assertion that requires a sequence to contain an element
+   * satysfying the given assertion on the given position
```
adamgfraser: Typo in "satysfying" here and below.

PR #2157, opened 2019-11-04
add more ZStream#distributed based combinators

mschuwalow:
Also updates the implementation of ZStream#distributedDynamicWith
/cc @iravid


iravid:
Sorry for the delay here. This PR requires slightly higher levels of energy from me so I'll get to it in a day or two.

PR #2158, opened 2019-11-04
add ZManaged#allocated

mschuwalow:
Maybe we can find a better name for this.
My vote still goes to ZManaged#dunno

/cc @iravid @pshemass

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -139,18 +139,6 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
     ZManaged.absolve(ev(self))
 
   /**
-   * Preallocates the managed resource, resulting in a ZManaged that reserves and acquires immediately and cannot fail.
-   */
-  final def allocated: ZIO[R, E, Managed[Nothing, A]] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -143,6 +138,18 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
   ): ZManaged[R1, E1, B] =
     ZManaged.absolve(ev(self))
 
+  /**
+   * Preallocates the managed resource, resulting in a ZManaged that reserves and acquires immediately and cannot fail.
+   */
+  final def allocated: ZIO[R, E, Managed[Nothing, A]] =
```
pshemass: just for my curiosity what happen with ```E``` in this case? is it swallowed?

pshemass:
so let go back to the name :)

for me ```allocated``` doesn't sound bad. it could be also ```acquired``` or  ```reserved```.

/cc: @iravid @mijicd

mijicd:
+1 for preallocate.

pshemass:
```preallocate``` let it be :)

mschuwalow:
Cool, I'll update the pr once I get home

PR #2161, opened 2019-11-05
Add eventually to ZManaged

saraiva132:
@adamgfraser Does this make sense?


saraiva132:
@adamgfraser I read the code for orElse and indeed I was not feeling too good about it. I am still getting acquainted with the codebase but I noticed exactly what you mentioned hence my comment :D 

So I do not think it makes sense for ZManaged eventually to also affect the release block. Based on your code sample it seems you agree. What do you think?

adamgfraser:
@saraiva132 Yes I do not think `eventually` should modify the behavior of the `release` action. The release action has a failure type of `Nothing` so it is guaranteed to not fail except for an unchecked exception which `eventually` would not recover from anyway.

PR #2165, opened 2019-11-06
Chunk#flatMap construct result array from the back

vasilmkd:
@iravid @jdegoes


iravid:
I’m all for optimizations but this one needs to be benchmarked, as not many things are faster than preprnding to a list

iravid:
See http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#construction-performance

vasilmkd:
Hmmm, TIL. Thanks. I have another idea.

vasilmkd:
@iravid How about now?

PR #2167, opened 2019-11-06
ZIO Test: Allow Inheritance Style Specification of Specs

### LINKED ISSUE #2085: ZIO Test: Allow inheritance-style specification of spec
```scala
object MySpec extends DefaultRunnableSpec {
  def spec = ...
}
```

adamgfraser:
Resolves #2085. I think this works relatively well. You can either specify the spec using constructor arguments or inheritance. If you don't do either you will get a spec that won't print anything and will be produce one ignored test. If you use both a constructor argument and inheritance the definition takes precedence. The inheritance specification needs to be a `def` or a `lazy val` or it will trigger an early initialization error (this seems to be the case even if we create a new `RunnableSpec` that doesn't take any constructor arguments in my tests). It also needs an `override` modifier.

@jdegoes What do you think?


iravid:
Love this @adamgfraser!

jdegoes:
I like it, but I feel maybe we should even go further, and deprecate/remove the old pattern (which eliminates the possibility of supplying nothing, and also eliminates the `override` keyword from uses of the inheritance method). But possibly we should do that in an RC with deprecation to allow people time to upgrade....

adamgfraser:
I agree it would be nice to get to a point where there is only one way to do this. Let me work on a version that goes in that direction.

adamgfraser:
@jdegoes How about this? If you are good with it I can deprecate `defaultRunnableSpec` and migrate our tests to use the inheritance style.

mschuwalow:
This is super nice. Also resolves the issue with defining helper functions, etc. for tests that we had until now

adamgfraser:
@jdegoes I deprecated `DefaultRunnableSpec` and updated all the tests to use the new style. If you are good with this I would suggest we merge as soon as possible. With the change to the new style of tests almost any other PR we merge will create a conflict with this one.

neko-kai:
If we're going inheritance-style, we may as well reexport functions from `zio.test` in `RunnableSpec` and remove the boilerplate `import zio.test._` import everywhere. What do you think?

jdegoes:
@adamgfraser Sorry for the delay, was stuck at the conference and then traveling back home. Looks good to me, feel free to merge after the conflicts are resolved.

@neko-kai Maybe as an option, but I think mixing them in for the user cannot be undone.

adamgfraser:
@jdegoes No worries. If we are removing deprecated code anyway, how about we change `DefaultRunnableSpec` to be the trait that users inherit when writing tests? This would make migration slightly easier (you just have to change from parentheses to curly braces and add `def spec = ...` without changing the name) and let us keep `RunnableSpec` polymorphic.

jdegoes:
@adamgfraser I like that!

jdegoes:
Fantastic work! Merged as soon as possible because of the size...

PR #2168, opened 2019-11-06
Proof of concept of spy mocking approach

13h3r:
# About this PR
This PR shows PoC for spy mocking. 

I would like to get a feedback on the idea and figure out do we want to work further in this direction or not.

# Motivation

With the current approach we need to specify an exact mock behavior before running a program. It works well for some cases, but for some cases you want to verify what your program did with a mock and validate calls after the execution.

The closest analog is `.when()` and `.then()` vs `.verify()`.

Another application of this kind of tests is to build property based tests with dependencies when we want to check a certain property of how the dependency is being called.

# Design

## Concept

High level overview of such kind of test:
1. define a behavior of the mock
2. instantiate the mock
3. running a program with the mock
4. obtaining traces of methods being called from the mock
5. validating methods being called

We need to implement `1.`, `2.` and `4.`. `3.` is a plan function call and `5.` could be done with the assertion framework that we have already.

## Define a behavior of the mock
In PoC I am piggybacking on current `Expectation` API and providing `anything` as assertion.

## Instantiate the mock
Instantiation is done with `spy` method which returns a tuple of a mocked service and an object to access execution traces.

## Obtaining traces 
`Spy` is an entry point to get execution traces. Its API is very simple now, could and should be extended to cover other cases.

# Further work

## Improving mock setup API
There are couple of disadvantages with a usage of `Expectation` API here:
* there is no way to skip assertions on function parameters
* there is no way to enforce that a methods behavior will be defined only once
* there is no way to use state or generators as part of a generation of a returning value in the mock

The better API here is something like this:
```
type Behavior[S, I, O] = (S, I) => UIO[(S, O)]
// we also can have specialized APIs too
type GenBehavior[O] = Gen[O]
type StateBehavior[S, I, O] = (S, I) => (S, O)

type MockBehavior[M, S] = Map[Method[M, Any, Any], Behavior[S, Any, Any]]
```

## Move towards generalized `Mock`
We may want to introduce generic state based `Mock` implementation and replace current `Mock.make` with it. In this case our state is a list of expected calls.

## Instantiate the mock
Returning a tuple is not really great API when we do not have `withFilter` for ZIO. I am open for any suggestions here.


adamgfraser:
@13h3r Thanks for writing this up! Let me think about it. Copying @ioleo as well.

ioleo:
@13h3r Hello! Thanks for these ideas. I really like the `setupMock / use` pattern. I'm looking forward to introduce changes to `zio-macros` to allow autogenerating mockable instance (and method definitions) anywhere (preferably in `src/test` rather than `src/main`) and promoting this way.

Regarding the **spying** proposition I think that if the only change is when the expectations/assertions are declared (`when/then` vs `verify`) then I don't see any benefit, it's just a matter of style.

However, if we defined a `spy` as a `proxy` for the underlaying implementation which logs the calls and their return values, but delegates the actual calculation I believe it could be useful to test some scenarios.

A seperate issue is to enhance current mocking API allowing more generic stubs like:
`Math.square(equalTo(anything)) return valueF(in => in * in) times(atLeast(2) && atMost(5))`

Last but not least, as we've talked before, integrating with generators would also be great.

13h3r:
@ioleo Thanks for you reply!

* `setupMock / use`. I am glad you like it! Feel free to send me PoC/full CR and I will provide a feedback
* spying. Typically I am creating a simple implementation of the service (for example HashMap based implementation instead of real KV store) and then want to validate what was called. My initial idea was to keep together a state based interface implementation and spy capabilities, but looks like if we can decouple them it will be even better, because then we can combine spying with mocks and spying with ad-hoc implementations. Do you agree that we should scope it to generic spying?
* `times` support. Agree this is a separate issue. Do you think we can also support a custom state in userspace to allow stateful mocks?
* `Gen` in mocks. I realized the idea is not really great in the form that we discussed, because it does not work well with current shrinking capabilities and we need somehow let `check` know that there are more generators involved into test and they also needs to be shrinked. I think it will be nice to use a more generic mechanism (like state based mocks) to support this usecase, rather that build a dedicated support.

13h3r:
Do we agree that we need to have `Spy` that can capture function calls? @ioleo @adamgfraser

ioleo:
@13h3r Yes, I would go that way.

adamgfraser:
@13h3r Sorry for the delay in getting back to you on this. Spent some time thinking about this and I think we actually have more power than I realized in `Expectation`.  One of the cases of `Expectation` is `Call`, which has the following signature:

```scala
  private[mock] final case class Call[M, I, E, A](
    method: Method[M, I, A],
    assertion: Assertion[I],
    returns: I => IO[E, A]
  ) extends Expectation[M, E, A]
```

`Call` lets us perform effects, and thus we can also already maintain state. We don't have the ability to access the environment, but I think that would be relatively simple to add. So conceptually we could create a `Spy` for any type we have an actual implementation of that delegates all calls to the underlying implementation and also logs calls and results. Furthermore, especially if we have access to the environment type we could have a `Spy` for services we don't have live implementations of by creating an ADT for the possible inputs and having `Call` pattern match against the inputs and generate the appropriate random value.

Here is a simple implementation of a spy for a cache that doesn't use the mocking framework, but we could relatively easily modify to use it. There is obviously some boilerplate here but I think there are some things we could do to clean it up, especially with macros.

What do you think?

```scala
import zio._
import zio.test._
import zio.test.Assertion._

object SpySpec
    extends DefaultRunnableSpec(
      suite("SpySpec") {
        testM("spy") {
          val expected = List(
            Put(3, "Bob"),
            Get(1, Some("Adam")),
            Get(2, Some("Amy")),
            Get(4, None)
          )
          for {
            test   <- Cache.makeTest(1 -> "Adam", 2 -> "Amy")
            spy    <- Spy.make(test)
            cache  = spy.cache
            _      <- cache.put(3, "Bob")
            _      <- cache.get(1)
            _      <- cache.get(2)
            _      <- cache.get(4)
            result <- cache.extract
          } yield assert(result, equalTo(expected))
        }
      }
    )

sealed trait Cache[K, V] {
  val cache: Cache.Service[K, V]
}

object Cache {

  trait Service[K, V] {
    def get(k: K): UIO[Option[V]]
    def put(k: K, v: V): UIO[Unit]
  }

  def get[K, V](k: K): ZIO[Cache[K, V], Nothing, Option[V]] =
    ZIO.accessM[Cache[K, V]](_.cache.get(k))

  def put[K, V](k: K, v: V): ZIO[Cache[K, V], Nothing, Unit] =
    ZIO.accessM[Cache[K, V]](_.cache.put(k, v))

  case class Test[K, V](data: Ref[Map[K, V]]) extends Service[K, V] {
    def get(k: K): UIO[Option[V]] =
      data.get.map(_.get(k))
    def put(k: K, v: V): UIO[Unit] =
      data.update(_.updated(k, v)).unit
  }

  def makeTest[K, V](kvs: (K, V)*): UIO[Cache[K, V]] =
    for {
      ref <- Ref.make(kvs.foldLeft(Map.empty[K, V]) { case (map, (k, v)) => map.updated(k, v) })
    } yield new Cache[K, V] { val cache = Test(ref) }

}

sealed trait Spy[K, V] extends Cache[K, V] {
  val cache: Spy.Service[K, V]
}

object Spy {

  case class Service[K, V](cache: Cache.Service[K, V], log: Ref[List[Request[K, V]]]) extends Cache.Service[K, V] {
    def get(k: K): UIO[Option[V]] =
      cache.get(k).tap(v => log.update(Get(k, v) :: _))
    def put(k: K, v: V): UIO[Unit] =
      cache.put(k, v) <* log.update(Put(k, v) :: _)
    def extract: UIO[List[Request[K, V]]] =
      log.get.map(_.reverse)
  }

  def extract[K, V]: ZIO[Spy[K, V], Nothing, List[Request[K, V]]] =
    ZIO.accessM[Spy[K, V]](_.cache.extract)

  def make[K, V](underlying: Cache[K, V]): UIO[Spy[K, V]] =
    for {
      log <- Ref.make(List.empty[Request[K, V]])
    } yield new Spy[K, V] { val cache = Service(underlying.cache, log) }
}

sealed trait Request[+K, +V]
final case class Get[K, V](k: K, v: Option[V]) extends Request[K, V]
final case class Put[K, V](k: K, v: V)         extends Request[K, V]
```

13h3r:
@adamgfraser Looks good! I was thinking about the same scope/approach.

13h3r:
@adamgfraser @ioleo I did some quick prototyping, but stuck with a simple question: how spying should behave in case of pure functions? In your example you can utilize an effect in your trait to encode spying effect, but what if code has no effect? 

I cant come up with any good idea here. Thoughts?

ioleo:
@13h3r No, so far it's only possible to mock effects, but I think with a few tweaks it should be possible to include pure functions mocks.

13h3r:
@ioleo could you share the idea of your tricks? How can we represent an effect with a pure function without changing function signature?

jdegoes:
@13h3r Closed in favor of #2660 

Great work on this and thank you for inspiring #2660!

PR #2169, opened 2019-11-06
Autosupervision

jdegoes:


adamgfraser:
Hallelujah!

jdegoes:
For some reason the js tests all hang... looking into that now.

jdegoes:
Hmm, maybe they don't hang so much as take an eternity to pass...

regiskuckaertz:
Simple and elegant, beautiful! Will take a closer look later today.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2736,7 +2746,8 @@ object ZStream extends Serializable {
                                   .unit
                               )
                             } catch {
-                              case FiberFailure(Cause.Interrupt) =>
+                              case FiberFailure(Cause.Interrupt(_)) =>
+                              case t: Throwable                     => t.printStackTrace()
```
iravid: This needed?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -770,7 +774,8 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
 
   /**
    * Switches over to the stream produced by the provided function in case this one
-   * fails. Allows recovery from all errors, except external interruption.
+   * fails. Allows recovery from all causes of failure, including interruption of the
+   * stream is uninterruptible.
```
iravid: Scaladoc can be rolled back?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -121,84 +121,88 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
-            notifyConsumer <- Promise.make[Nothing, Unit]
-            extractResult  <- sink.extract(state)
-            (b, leftovers) = extractResult
-            nextState = if (leftovers.isEmpty) State.Empty(initial, notifyConsumer)
-            else State.Leftovers(initial, leftovers, notifyConsumer)
-          } yield (notifyProducer.succeed(()).as(Chunk.single(b)), nextState)).mapError(Some(_))
-
-        case e @ State.Error(cause) => ZIO.succeed((ZIO.halt(cause.map(Some(_))), e))
-        case State.End              => ZIO.succeed((ZIO.fail(None), State.End))
+      withStateVar(stateVar, permits) { s =>
```
iravid: has anything changed here except the addition of the explicit match? If not, let's roll it back so the diff is smaller

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -121,84 +121,88 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
-                       )
-                     else
-                       UIO.succeed(
-                         (
-                           // Notify the consumer, wait for them to take the aggregate so we know
-                           // it's time to progress, and process the leftovers
-                           notifyConsumer.succeed(()) *> notifyProducer.await.as(true),
-                           State.BatchEnd(step, notifyProducer)
+      withStateVar(stateVar, permits) { s =>
+        (s match {
```
iravid: Same here re `match`

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -171,6 +225,8 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
         val sw = new StringWriter()
         val pw = new PrintWriter(sw)
         e.printStackTrace(pw)
+        e.printStackTrace()
+        pw.flush()
```
iravid: Think we can drop this too

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1039,7 +1039,7 @@ object Chunk {
       else if (isFloat(c)) FloatClass.asInstanceOf[ClassTag[A]]
       else if (isDouble(c)) DoubleClass.asInstanceOf[ClassTag[A]]
       else if (isChar(c)) CharClass.asInstanceOf[ClassTag[A]]
-      else classTag[java.lang.Object].asInstanceOf[ClassTag[A]]
+      else classTag[AnyRef].asInstanceOf[ClassTag[A]] // TODO: Find a better way
```
iravid: This was the problem with the `ArrayStoreException`?

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -307,18 +329,17 @@ object Fiber {
...
+   * @param children      The fiber's forked children. This will only be populated if the fiber is supervised (via [[ZIO#supervised]]).
    */
   final case class Descriptor(
     id: FiberId,
-    interrupted: Boolean,
+    interruptors: Set[FiberId],
     interruptStatus: InterruptStatus,
-    superviseStatus: SuperviseStatus,
-    executor: Executor,
-    children: UIO[IndexedSeq[Fiber[Any, Any]]]
```
regiskuckaertz: 🎉

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -31,25 +32,23 @@ import scala.collection.JavaConverters._
...
 
   import FiberContext._
   import FiberState._
 
   // Accessed from multiple threads:
-  private[this] val state = new AtomicReference[FiberState[E, A]](FiberState.initial)
-
-  @noinline
-  private[this] def interrupted = state.get.interrupt
+  private val state = new AtomicReference[FiberState[E, A]](FiberState.initial)
```
regiskuckaertz: Not `private[this]`?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1243,6 +1278,21 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   ): ZIO[R1, E1, A] =
     self.foldCauseM(new ZIO.TapErrorRefailFn(f), new ZIO.TapFn(g))
 
+  /**
+   * Taps into the children of this fiber (collected into a single fiber),
+   * guaranteeing the specified callback will be invoked, whether or not
+   * this effect succeeds.
+   */
+  final def tapChild[R1 <: R](f: Fiber[Any, List[Any]] => ZIO[R1, Nothing, Any]): ZIO[R1, E, A] =
```
neko-kai: `ensuringChild`?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -581,7 +575,7 @@ private[zio] final class FiberContext[E, A](
             }
           } else {
             // Fiber was interrupted
-            curZio = ZIO.interrupt
+            curZio = ZIO.halt(state.get.interrupted)
```
neko-kai: line info is lost in this change, use `haltWith` or `refailWithTrace` here?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -745,16 +709,54 @@ private[zio] final class FiberContext[E, A](
...
+    val _ = daemonStatus.popOrElse(false)
+  }
+
+  final def propagateAncestorInterruption(): Unit = {
+    var fiber = self: FiberContext[_, _]
+
+    while (fiber ne null) {
+      addInterruptor(fiber.state.get.interrupted)
+
+      fiber = fiber.parentFiber
```
neko-kai: It will need to be documented later that propagation is not hierarchical – a nested fiber will be interrupted even when all of its superiors are uninterruptible/unresponsive, as long as any one of them is targeted by kill0.

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: Yes. And I dug in deep, it literally has no stack trace attached, probably because it's generated from Scala code.

It happens when we have an array like `Array(Dog, Animal, Cat)` and detect the type as `Dog` and so specialize to `[LDog` but then putting `Animal` or `Cat` in there throws `ArrayStoreException`. So we either need to be not lazy or just use `AnyRef`. There's no reason I can tell for specializing the array, may as well just always do `AnyRef` arrays. But the types are a lie so we should encode that more nicely.

core/shared/src/main/scala/zio/ZIO.scala
jdegoes: 👍

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: I had to weaken it, because now I have to access the parent state from the child.

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Fixed

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Only if the children are interruptible, which by default they would not be, because interruptibility status is inheritable.

It's only if someone specifically made some section (or fiber) interruptible that the children would potentially terminate first, but that's the same as in-fiber semantics: if you call downstream code and there's an interruptible section in there, it could interrupt yourself unless you're dealing with it.

It's still worth documenting, of course.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -535,10 +605,16 @@ object Cause extends Serializable {
   private[Cause] def sym(f: (Cause[Any], Cause[Any]) => Boolean): (Cause[Any], Cause[Any]) => Boolean =
     (l, r) => f(l, r) || f(r, l)
 
-  private[Cause] def flatten(c: Cause[Any]): Set[Cause[Any]] = c match {
-    case Then(left, right) => flatten(left) ++ flatten(right)
-    case Both(left, right) => flatten(left) ++ flatten(right)
-    case Traced(cause, _)  => flatten(cause)
+  private[Cause] def flattenSeq(c: Cause[_]): Vector[Cause[_]] = c match {
+    case Then(left, right) => flattenSeq(left) ++ flattenSeq(right)
```
iravid: Missing Both?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1919,9 +1966,8 @@ private[zio] trait ZIOFunctions extends Serializable {
       r <- ZIO.runtime[R]
       a <- ZIO.uninterruptibleMask { restore =>
             val f = register(k => r.unsafeRunAsync_(k.to(p)))
-            restore(f.catchAllCause(p.halt)).fork.flatMap { f =>
-              restore(p.await).onInterrupt(f.interrupt)
-            }
+
+            restore(f.catchAllCause(p.halt)).fork *> restore(p.await) //.onInterrupt(f.interrupt)
```
iravid: Comment unneeded, right? `f` should get interrupted automatically

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Is  the `t.printStackTrace()` needed here?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -63,12 +62,12 @@ private[zio] final class FiberContext[E, A](
   private[this] val traceEffects: Boolean =
     traceExec && platform.tracing.tracingConfig.traceEffectOpsInExecution
 
-  private[this] val fiberId         = FiberContext.fiberCounter.getAndIncrement()
   private[this] val stack           = Stack[Any => IO[Any, Any]]()
   private[this] val environments    = Stack[AnyRef](startEnv)
   private[this] val executors       = Stack[Executor](startExec)
   private[this] val interruptStatus = StackBool(startIStatus.toBoolean)
-  private[this] val supervised      = Stack[SuperviseStatus](startSStatus)
+  private[this] val children        = new AtomicReference(Set[FiberContext[_, _]]())
```
iravid: Long term this could be replaced with a more efficient mutable Set, right?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -400,13 +392,14 @@ private[zio] final class FiberContext[E, A](
...
                     // Error not caught, stack is empty:
-                    curZio = null
+                    val cause = {
+                      // Add interruption information into the cause, if it's not already there:
+                      val interrupted = state.get.interrupted
 
-                    val cause =
-                      if (interrupted && !cause0.interrupted) cause0 ++ Cause.interrupt
-                      else cause0
+                      if (!cause0.contains(interrupted)) cause0 ++ interrupted else cause0
```
iravid: That contains call could get pretty expensive for large causes. I don't have a good idea on how to avoid it, but it's only relevant for internal interruption, right?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -745,16 +709,54 @@ private[zio] final class FiberContext[E, A](
...
+      addInterruptor(fiber.state.get.interrupted)
+
+      fiber = fiber.parentFiber
+    }
+  }
 
   @inline
-  private[this] final def interruptible: Boolean = interruptStatus.peekOrElse(true)
+  private final def isInterrupted(): Boolean = {
+    propagateAncestorInterruption()
```
iravid: Question: Why is this method (which is only called from `shouldInterrupt`) doing the propagation logic?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -858,13 +893,8 @@ private[zio] final class FiberContext[E, A](
...
 
-    // To preserve fair scheduling, we submit all resumptions on the thread
-    // pool in order of their submission.
-    observers.reverse.foreach(
-      k =>
-        platform.executor
-          .submitOrThrow(() => k(result))
-    )
+    // For improved fairness, we resume in order of submission:
+    observers.reverse.foreach(k => k(result))
```
iravid: Hmm why did we stop resuming on the threadpool here? Does this not execute the resumptions sequentially?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -795,32 +799,63 @@ private[zio] final class FiberContext[E, A](
...
+  private[this] final def kill0(fiberId: FiberId): UIO[Exit[E, A]] = {
+    @tailrec
+    def setInterruptedLoop(): Unit = {
+      val oldState = state.get
+
+      oldState match {
+        case Executing(FiberStatus.Suspended(true, _), observers, interrupted) =>
+          if (!state.compareAndSet(
+                oldState,
+                Executing(FiberStatus.Running, observers, interrupted ++ Cause.interrupt(fiberId))
```
iravid: The state is getting switched to `FiberStatus.Running` here - just verifying that is correct

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -768,25 +770,27 @@ private[zio] final class FiberContext[E, A](
...
 
     oldState match {
       case Executing(_, observers: List[Callback[Nothing, Exit[E, A]]], _) => // TODO: Dotty doesn't infer this properly
         if (!state.compareAndSet(oldState, Done(v))) done(v)
         else {
-          notifyObservers(v, observers)
           reportUnhandled(v)
+          notifyObservers(v, observers)
+          // Disconnect this node from the tree for GC reasons:
+          children.get.foreach { child =>
```
iravid: <3

core/shared/src/main/scala/zio/Cause.scala
jdegoes: `flattenSeq` should not flatten both. @adamgfraser fixed all this logic in the new one.

core/shared/src/main/scala/zio/ZIO.scala
jdegoes: Fixed.

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: I'm going to fix that soon. It is costly so it cannot be done for every iteration of the eval loop (like it is now).

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: It does, but now async operations just "trigger" resumption of the computation on the normal ZIO thread pool, so this improves performance because it eliminates the extra unnecessary redundant submission.

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Yes, we set it to `FiberStatus.Running` because we immediately call `evaluateNow` which means the fiber will be running again.

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Yes, although when the children are retrieved, then you have to copy the whole thing into an immutable set, which probably implies making that retrieval a separate operation (now it is bundled into descriptor).

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: Even external interruption. There must be a better way to avoid it. I got closer this time to eliminating the hackiness but there's still a few small issues that have to be dealt with.

core/shared/src/main/scala/zio/internal/FiberContext.scala
iravid: Agreed on that. Retrieving the children would also be a far less frequent operation after this is merged.

core/shared/src/main/scala/zio/internal/FiberContext.scala
iravid: Ok 👍

jdegoes:
@iravid @neko-kai This got a few upgrades if you want to take a look. One more coming later today maybe in a separate PR though.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -363,6 +519,11 @@ object Fiber {
    */
   final def fail[E](e: E): Fiber[E, Nothing] = done(Exit.fail(e))
 
+  /**
+   * A `FiberRef` that stores the name of the fiber, which defaults to `None`.
+   */
+  val fiberName: FiberRef[Option[String]] = new FiberRef(None, (old, _) => old)
```
iravid: Ohh that's pretty cool

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -266,397 +258,396 @@ private[zio] final class FiberContext[E, A](
...
-                  val zio = curZio.asInstanceOf[ZIO.RaceWith[Any, Any, Any, Any, Any, Any, Any]]
-                  curZio = raceWithImpl(zio).asInstanceOf[IO[E, Any]]
+                  case ZIO.Tags.RaceWith =>
+                    val zio = curZio.asInstanceOf[ZIO.RaceWith[Any, Any, Any, Any, Any, Any, Any]]
+                    curZio = raceWithImpl(zio).asInstanceOf[IO[E, Any]]
+                }
               }
+            } else {
+              // Fiber was interrupted
+              curZio = ZIO.halt(state.get.interrupted)
```
neko-kai: Ok, if `.refailWithTrace`, why not `.haltWith`? As is there's no info on what the fiber has been executing immediately before interruption

jdegoes: It can't be a composite instruction or it will end in an infinite loop. So we have to build the full cause right here. I could attach the trace here, I suppose?

neko-kai: @jdegoes Yeah you can construct with one op:

```scala
val snapshot = state.get.interrupted
ZIO.haltWith(trace => Cause.traced(snapshot, trace())
```

jdegoes: Actually I see `ZIO.halt` will generate `Fail` which will be filled with the trace. Should be good enough, right?

neko-kai: @jdegoes `fail` and `die` do generate the trace, but`halt` passes the raw Cause, see impl - https://github.com/zio/zio/blob/4b7225dd36f6a4b16f170a778cb553e18cb461d1/core/shared/src/main/scala/zio/ZIO.scala#L2279

PR #2170, opened 2019-11-07
ZIO Test: Limit Flaky

adamgfraser:
`TestAspect#flaky` currently retries a flaky test without limit. This can result in a test that consistently fails timing out, creating the risk of confusing assertion failures for nontermination. This PR changes `flaky` to retry a test up to 100 times. This should be plenty of time for a flaky test to pass while preventing failing tests from timing out. Users who still want to retry failing tests without limit can use `TestAspect#eventually`.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -270,10 +270,11 @@ object TestAspect extends TimeoutVariants {
     }
 
   /**
-   * An aspect that retries a test until success, without limit, for use with
-   * flaky tests.
+   * An aspect that retries a test until success, with a limit specified by the
+   * test framework, for use with flaky tests.
    */
-  val flaky: TestAspectPoly = eventually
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -270,10 +270,11 @@ object TestAspect extends TimeoutVariants {
     }
 
   /**
-   * An aspect that retries a test until success, without limit, for use with
-   * flaky tests.
+   * An aspect that retries a test until success, with a limit specified by the
+   * test framework, for use with flaky tests.
    */
-  val flaky: TestAspectPoly = eventually
+  val flaky: TestAspectPoly =
```
jdegoes: I'd add `def flaky(n: Int)` and then define `val flaky = flaky(100)`.

adamgfraser: Good idea!

PR #2172, opened 2019-11-07
Test: migrate checkspec

TapanVaishnav:
             https://github.com/zio/zio/pull/2120


unclebob418:
@TapanVaishnav nice work. I have mixed feelings about your implication restriction hack `import zio.test.{ check => Check, checkM => CheckM }`. On one hand, it's great in that it makes things look clean, on the other it may cause some confusion vs using e.g. `zio.test.checkM(...)`, since everywhere else it's referred to lower case. That said, there's a reason we can alias imports so perhaps it's ok. @adamgfraser what do you think?

adamgfraser:
@unclebob418 I think it is okay. This is only in the test suite so it doesn't effect users and having to write out `zio.test.check` everywhere isn't great either. In `ZIOSpec` we aliased `test` as `pureTest` to get around a similar issue (ZIO Test package name conflicts with name of `test` method). Can we come up with a name we like for this to consistently use in the test suite to make things easy for contributors?

unclebob418:
Just capitalising seems to work ok, but I think it makes them look like case classes instead of methods. Another option would be to prepend with test e.g. testCheckM.

A comment up in the imports to say why it was aliased wouldn't hurt.

TapanVaishnav:
> Another option would be to prepend with test e.g. testCheckM.

I also considered that one but then it was relatively longer so went with `CheckM`

PR #2173, opened 2019-11-07
ZIO Test: Add generators for time datatypes

dieproht:
This is an extension of Gen implementing generators for basic time datatypes: ZIO duration, Instant, LocalDateTime and OffsetDateTime. The implementation is done in the way like diskussed with @adamgfraser at Berlin hackathon.


test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import zio.duration.Duration
+import zio.random.Random
+
+trait TimeVariants {
+
+  /**
+   * A generator of ZIO duration values. Shrinks toward Duration.Zero.
+   */
+  final def anyDuration: Gen[Random, Duration] = Gen.long(0L, Long.MaxValue).map(Duration.Finite(_))
```
regiskuckaertz: `anyFiniteDuration`?

dieproht: That makes sense. I've updated the PR

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * A generator java.time.OffsetDateTime values.
+   */
+  final def anyOffsetDateTime: Gen[Random, OffsetDateTime] = offsetDateTime(OffsetDateTime.MIN, OffsetDateTime.MAX)
+
+  /**
+   * A generator of ZIO duration values inside the specified range: [min, max].
+   */
+  final def duration(min: Duration, max: Duration): Gen[Random, Duration] =
```
adamgfraser: `finiteDuration`? Should we add a generator of infinite values (it would just be a constant generator) and possibly a generator of duration values that produces both? Not sure what ratio of finite to infinite values would make sense.

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -0,0 +1,82 @@
...
+
+trait TimeVariants {
+
+  /**
+   * A generator of ZIO duration values. Shrinks toward Duration.Zero.
+   */
+  final def anyFiniteDuration: Gen[Random, Duration] = Gen.long(0L, Long.MaxValue).map(Duration.Finite(_))
+
+  /**
+   * A generator java.time.Instant values.
```
adamgfraser: "generator of" instead of "generator". I would put the class names like `java.time.Instant` in back ticks. Also briefly state what the generator shrinks towards.

test-tests/shared/src/main/scala/zio/test/TimeVariantsSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.concurrent.Future
+
+import zio.duration._
+import zio.test.GenUtils.{ checkSample, checkShrink }
+import zio.test.TestUtils.label
+
+object TimeVariantsSpec extends AsyncBaseSpec {
+
+  val run: List[Async[(Boolean, String)]] = List(
+    label(anyFiniteDurationShrinksToZero, "anyFiniteDuration shrinks to zero"),
```
adamgfraser: Can we add properties about what any of the other generators shrink to?

dieproht: Thank you for your feedback! I'm still working on improvements...

adamgfraser: No worries. Just reach out whenever you're ready for another review.

dieproht:
Hi @adamgfraser, generators for (finite) Duration, Instant, LocalDatetime and OffsetDateTime are ready for another review now. I would not add more datatypes in this PR because one can derive many similar types from the existing ones in a trivial way (i.e. LocalDateTime.toLocalTime). 

However I would like to add anyDuration, but therefor we have to combine a constant generator that emits Infinity with the existing anyFiniteDuration generator reasonably and it's unclear to me how to do that. Do you have an clue about this? Otherwise let`s add that later in another PR...

adamgfraser:
@dieproht To combine a generator of finite values with a generator of infinite values we could either do something like `booleans.flatMap(p => if (p) finite else infinite)` (if we want an equal distribution between finite and infinite values) or `weighted` which lets us specify some percentage of finite versus infinite values to generate. So mechanically it seems straightforward. The harder question is what ratio to generate. The only non-arbitrary ratio seems to be half and half, but that also feels like potentially "too many" infinite values. Maybe that is a reason to just leave it up to the user to combine them in the appropriate ratio for them or maybe we try to just pick some ratio so we can have a generator that covers both finite and infinite values for easy testing of functions that use duration concepts. What do you think?

dieproht:
Thank you @adamgfraser for supporting this!

PR #2174, opened 2019-11-07
TRef documentation

felipehaack:
First version for the TRef data type documentation.

@mijicd, I'm not sure about the structure (I tried to follow the other examples), would be great to hear your thoughts.

### DIFF
--- a/docs/datatypes/tref.md
+++ b/docs/datatypes/tref.md
@@ -1,0 +1,174 @@
+---
+id: datatypes_tref
+title:  "TRef"
+---
+
+A `TRef[A]` is a mutable reference to an immutable value, which can participate in transactions in STM. The mutable reference can be retrieved and set from within transactions, with strong guarantees for atomicity, consistency, and isolation from other transactions.
+
+`TRef` provides the low-level machinery to create transactions from modifications of STM memory.
+
+## Create a TRef
+
+Creating a `TRef` inside a transaction:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val createTRef: STM[Nothing, TRef[Int]] = TRef.make(10)
+```
+
+Or creating a `TRef` inside a transaction, and immediately committing the transaction, which allows you to store and pass along the reference.
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val commitTRef: UIO[TRef[Int]] = TRef.makeCommit(10)
+```
+
+## Retrieve the value out of a TRef
+
+Retrieving the value in a single transaction: 
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val retrieveSingle: UIO[Int] = (for {
+  tRef <- TRef.make(10)
+  value <- tRef.get
+} yield value).commit
+```
+
+Or on multiple transactional statements:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val retrieveMultiple: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
+  value <- tRef.get.commit
+} yield value
+```
+
+## Set a value to a TRef
+
+Setting the value overwrites the existing content of a reference.
+
+Setting the value in a single transaction:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val setSingle: UIO[Int] = (for {
+  tRef <- TRef.make(10)
+  _ <- tRef.set(20)
+  nValue <- tRef.get
+} yield nValue).commit
+```
+
+Or on multiple transactions:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val setMultiple: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
+  nValue <- tRef.set(20).flatMap(_ => tRef.get).commit
+} yield nValue
+```
+
+## Update the value of the TRef
+
+The update function `A => A` allows computing a new value for the `TRef` using the old value.
+
+Updating the value in a single transaction:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val updateSingle: UIO[Int] = (for {
+  tRef <- TRef.make(10)
+  nValue <- tRef.update(_ + 20)
+} yield nValue).commit
+```
+
+Or on multiple transactions:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val updateMultiple: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
+  nValue <- tRef.update(_ + 20).commit
+} yield nValue
+```
+
+## Modify the value of the TRef
+
+The modify function `A => (B, A): B` works similar to `update`, but allows extracting some information (the `B`) out of the update operation. 
+
+Modify the value in a single transaction:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val modifySingle: UIO[(String, Int)] = (for {
+  tRef <- TRef.make(10)
+  mValue <- tRef.modify(v => ("Zee-Oh", v + 10))
+  nValue <- tRef.get
+} yield (mValue, nValue)).commit
+```
+
+Or on multiple transactions:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val modifyMultiple: UIO[(String, Int)] = for {
+  tRef <- TRef.makeCommit(10)
+  tuple2 <- tRef.modify(v => ("Zee-Oh", v + 10)).zip(tRef.get).commit
+} yield tuple2
+```
+
+## Example usage
+
+Here is a scenario where we use a `TRef` to hand-off a value between two `Fiber`s
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+def transfer(tSender: TRef[Int],
+             tReceiver: TRef[Int],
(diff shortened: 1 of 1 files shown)


docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
+---
+id: datatypes_tref
+title:  "TRef"
+---
+
+A `TRef[A]` is part of the `STM` bundle that represents a transactional reference that can perform reads and writes inside a transactional effect.
+
+`TRef` gives the basic machinery for others `SMT` transactional data types.
```
mijicd: Fix typo, `SMT` => `STM`.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
...
+`TRef[A]` where `A` is an immutable value.
+
+## Create a TRef
+
+Creating a `TRef` without committing its value:
+
+```scala mdoc:silent
+import zio.stm.{STM, TRef}
+
+val tVar: STM[Nothing, TRef[Int]] = TRef.make(10)
```
mijicd: `make` has been renamed to `apply`, hence it's `TRef(10)`.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
+---
+id: datatypes_tref
+title:  "TRef"
+---
+
+A `TRef[A]` is part of the `STM` bundle that represents a transactional reference that can perform reads and writes inside a transactional effect.
+
+`TRef` gives the basic machinery for others `SMT` transactional data types.
+
+`TRef[A]` where `A` is an immutable value.
```
mijicd: I would remove this sentence.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
...
+## Retrieve the value out of a TRef
+
+By retrieving the value in a single transactional statement: 
+
+```scala mdoc:silent
+import zio.UIO
+import zio.stm.TRef
+
+val tVal: UIO[Int] = (for {
+  tRef <- TRef.make(10)
```
mijicd: See the comment about renaming.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+Or on multiple transactional statements:
+
+```scala mdoc:silent
+import zio.UIO
+import zio.stm.TRef
+
+val tVal: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
```
mijicd: I'd consider renaming `makeCommit` method to `committed`. 

@jdegoes wdyt?

docs/datatypes/tref.md
```diff
@@ -0,0 +1,223 @@
...
+  value <- tRef.get.commit
+} yield value
+```
+
+## Set a value to a TRef
+
+By setting a value, it will commit a new value without any relation with the actual value at certain point in time.
```
mijicd: I'd rephrase this to something like `Setting the value overwrites the existing content of a reference`

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+val tVal: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
+  _ <- tRef.set(20).commit
+  value <- tRef.get.commit
+} yield value
+```
+
+## Update the value of the TRef
+
+The update function `A => A`, differently from `set`, it cares about the actual value in the `TRef` by performing changes over the value and returning the same type at certain point in time.
```
mijicd: I'd rephrase to `Update modifies the content of a reference by applying function to it`.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
+---
+id: datatypes_tref
+title:  "TRef"
+---
+
+A `TRef[A]` is part of the `STM` bundle that represents a transactional reference that can perform reads and writes inside a transactional effect.
```
jdegoes:
```suggestion
A `TRef[A]` is a mutable reference to an immutable value, which can participate in transactions in STM. The mutable reference can be retrieved and set from within transactions, with strong guarantees for atomicity, consistency, and isolation from other transactions.
```

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+
+`TRef` gives the basic machinery for others `SMT` transactional data types.
```
jdegoes:
```suggestion
`TRef` provides the low-level machinery to create transactions from modifications of STM memory.
```

docs/datatypes/tref.md
jdegoes: Arg, we should change this back (I am sorry I didn't notice that). In Scala, `TRef.apply` should give you a `TRef`, not a `TRef` inside `STM`. So it's a pervasive ZIO custom that `make` gives you something inside an effect.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+
+`TRef[A]` where `A` is an immutable value.
+
+## Create a TRef
+
+Creating a `TRef` without committing its value:
```
jdegoes:
```suggestion
Creating a `TRef` inside a transaction:
```

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+
+```scala mdoc:silent
+import zio.stm.{STM, TRef}
+
+val tVar: STM[Nothing, TRef[Int]] = TRef.make(10)
+```
+
+Or creating a `TRef` and immediately committing its value to extract out the value:
```
jdegoes:
```suggestion
Or creating a `TRef` inside a transaction, and immediately committing the transaction, which allows you to store and pass along the reference.
```

docs/datatypes/tref.md
mijicd: Gotcha, I'll polish these up asap.

docs/datatypes/tref.md
jdegoes: It is `make` followed by `commit`. So `makeCommit`. If we kept with `apply` for `make` then `committed` could work well, or even just `commit`, but I am not a fan of that for reasons listed above.

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+val tVal: UIO[Int] = (for {
+  tRef <- TRef.make(10)
+  _ <- tRef.set(20)
+  nValue <- tRef.get
+} yield nValue).commit
+```
+
+Or on multiple transactional statement:
+
+```scala mdoc:silent
```
jdegoes: This example is not really realistic because each thing is done in isolation from the others, so there are no atomic guarantees across the set and get.

docs/datatypes/tref.md
jdegoes:
```suggestion
The update function `A => A` allows computing a new value for the `TRef` using the old value.
```

docs/datatypes/tref.md
```diff
@@ -0,0 +1,176 @@
...
+val tVal: UIO[Int] = for {
+  tRef <- TRef.makeCommit(10)
+  _ <- tRef.update(_ + 20).commit
+  value <- tRef.get.commit
+} yield value
+```
+
+## Modify the value of the TRef
+
+The modify function `A => (B, A): B` works similar to `update`, but it gives the possibility to perform computation over the actual value and updating to a new value.
```
jdegoes:
```suggestion
The modify function `A => (B, A): B` works similar to `update`, but allows extracting some information (the `B`) out of the update operation.
```

docs/datatypes/tref.md
felipehaack: Sorry, I didn't pull the latest changes. Thanks for noticing. +1

docs/datatypes/tref.md
felipehaack: Nice, I prefer your version, clarifies more.

docs/datatypes/tref.md
felipehaack: Perfect, let me change to your version.

felipehaack: Thank you for rephrasing and making even more clarified. (same applies for the other comments too)

docs/datatypes/tref.md
felipehaack: Sure, you're completely right (also applies for the other examples), for each statement, they're working in isolation. Let me just make sure I'm not wrong, but to have guarantees across the set and get, should I enclosure the statements in a single atomic transaction?
```
for {
  tRef <- TRef(10).commit
  nValue <- STM.atomically {
    for {
      _ <- tRef.set(12)
      nValue <- tRef.get
    } yield nValue
  }
} yield nValue
```
then it gives the guarantees about atomicity of the `set and get`.

docs/datatypes/tref.md
mijicd: Reverted the `apply` renaming in #2179. Please take a look.

docs/datatypes/tref.md
felipehaack: As you made changes to rolling back to `make`, I will leave as it is.

felipehaack:
@mijicd, thank you for all the comments, they're always welcome.
I pushed the changes just now. Let me know if I'm missing something else.

mijicd:
@felipehaack looks good to me, examples are fixed, changes are incorporated. We can polish it later in case someone notices any issues with the contents.

PR #2175, opened 2019-11-07
ZIO Test: Fix Type Inference Issue in `checkM` Variants

adamgfraser:
The current signature of `checkM` is:

```scala
final def checkM[R, E, A](rv: Gen[R, A])(test: A => ZIO[R, E, TestResult]): ZIO[R, E, TestResult]
```

The curried syntax is quite nice for ergonomics and writing tests in a property based style, but there can be a type inference issue if the test requires an additional environment beyond what the generator does because the compiler determines the `R` type after looking at the first parameter without considering the second.

This PR resolves this by changing the signature to be:

```scala
final def checkM[R, R1 <: R, E, A](rv: Gen[R, A])(test: A => ZIO[R1, E, TestResult]): ZIO[R1, E, TestResult]
```

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/CheckSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/CheckSpec.scala
@@ -83,6 +83,13 @@ object CheckSpec
             val sorted   = assert(nss.zip(nss.tail).exists { case (a, b) => a > b }, isFalse)
             nonEmpty ==> sorted
           }
+        },
+        testM("checkM effect type is correctly inferred") {
+          CheckM(Gen.unit) { _ =>
+            for {
+              _ <- random.nextInt
+            } yield assertCompletes
+          }
         }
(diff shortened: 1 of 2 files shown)


adamgfraser:
Yes, I thought about that and agree that would work too.

In my experiments the type inference worked quite well. The main thing I was worried about was that since we are stating that `R1 <: R` there would be an issue if `R1` was actually a supertype of `R` (e.g. the generator requires `Random` but the test doesn't require any effect). However, since `R` is contravariant the type can automatically be narrowed which I think may make this work a little better for the `R` parameter than the other parameters. I tested it with an effect that required a completely different environment (e.g. generator requires random, test requires clock) and that worked as well. We have also used this pattern in `Live#withLive` and I haven't seen any issues there.

jdegoes:
@adamgfraser All right, sounds great!

PR #2179, opened 2019-11-07
Align STM factory names

mijicd:
Ensure all factories returning STM are named `make` rather than `apply`.


jdegoes:
Superb! Thanks for making this consistent with the rest of ZIO.

PR #2180, opened 2019-11-07
ZIO Test: Unify Order of Type Parameters

### LINKED ISSUE #2178: ZIO Test: Unify order of type parameters
We did this once before, but unfortunately it looks like there is some divergence, e.g.:

```scala
TestRunner[R, L, T, E, S]
ZSpec[R, E, L, T]
```

We should look at all types and see if there is an ordering we could standardize on that would be familiar to those who know `R, E, A` in ZIO.

adamgfraser:
Resolves #2178.

This was relatively straightforward. The only thing that I think sometimes is a little awkward is the ordering of the `E` parameter because sometimes it can be more of an effect error type like in `Spec[R, E, L, T]` and in other cases it can be more of an output or capability. For example a `TestReporter[E, L, S]` can't fail at all but rather can report on tests that fail with an `E` or succeed with an `S`. So in isolation I do think that reads better as `TestReporter[L, E, S]`. But I recognize there is tremendous value in always having the type parameters in the same order, and even I found that the consistent ordering was nice while doing this refactoring.


adamgfraser:
@jdegoes Could you take a look at this when you get the chance?

jdegoes:
Looks great!

PR #2183, opened 2019-11-08
Fix Issue on Master and Update CI To Prevent Regressions

adamgfraser:
The benchmarks project is not currently compiling on master because of the recent change to the factory methods to the STM types. We're not currently compiling the benchmarks project in CI which creates the risk of regressions like this. Corrects the current issue and adds a task to CI to compile the benchmarks.


ghostdogpr:
It seems some libraries used in benchmarks are not available on Scala 2.13.

PR #2184, opened 2019-11-08
ZIO Test: Fix Naming Conflict with TestAspect#ignore

adamgfraser:
Currently we have `ignore` in the ZIO Test package object that creates an ignored test result and `TestAspect#ignore`. If you do some common wildcard imports like below this creates a naming conflict so you can't use `TestAspect#ignore` without hiding things / aliasing things / using a qualified name, which is not very nice. In my experience `TestAspect#ignore` is relatively commonly used versus `ignore` in the package object is quite uncommonly used. This PR changes the `ignore` in the package object to `ignored` to avoid the conflict.

```scala
import zio.test._
import zio.test.TestAspect._
```

### DIFF
--- a/test-tests/shared/src/main/scala/zio/test/TestUtils.scala
+++ b/test-tests/shared/src/main/scala/zio/test/TestUtils.scala
@@ -2,7 +2,7 @@ package zio.test
 
 import scala.concurrent.{ ExecutionContext, Future }
 
-import zio.{ Schedule, UIO, ZIO }
+import zio.{ UIO, ZIO }
 import zio.test.environment.TestEnvironment
 
 object TestUtils {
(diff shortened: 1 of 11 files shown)


test-tests/shared/src/main/scala/zio/test/TestUtils.scala
```diff
@@ -26,14 +23,25 @@ object TestUtils {
       results.forall { case Spec.TestCase(_, test) => test.map(f); case _ => ZIO.succeed(true) }
     }
 
-  final def ignored[L, E, S](spec: ZSpec[environment.TestEnvironment, E, L, S]): ZIO[Any, Nothing, Boolean] = {
+  final def isFailed[L, E, S](spec: ZSpec[environment.TestEnvironment, E, L, S]): ZIO[Any, Nothing, Boolean] =
```
ghostdogpr: Was that also conflicting?

Maybe `failedWith` should be renamed too?

adamgfraser: These combinators are only used in the internal test suite for ZIO Test. Once I renamed `ignore` to `ignored` I created another conflict with an internal combinator used here so I changed that to `isIgnored` and then just changed the other internal combinators for consistency. Yes I think we could rename `failedWith` too.  Let me do that now.

adamgfraser: Actually realized several of those combinators are not being used anymore now that the tests that relied on them have been migrated to ZIO Test so was able to remove them entirely.

jdegoes:
Worthwhile reviewing the others at the same time and being consistent about naming and avoiding name clashes.

adamgfraser:
I went through everything else in the package object. The only other potential issues are see are:

1. `fail` in the package object constructs a failed test. It is used quite infrequently in my experience. Shall we name it `failed` to be consistent with `ignored`? We don't have anything else that this would conflict with.
2. `test` method in the package object is the same as the package name. The only time this comes up is when trying to use `test` method in the `zio` package. In that case you have to do an alias import like `import zio.{ test => pureTest }`. This only comes up when writing tests in `zio` package (not a package like `zio.stream` and using `test` versus `testM`, so it really only impacts us internally and has come up quite infrequently, but is a little annoying. Not sure what to do about this since we are kind of committed to the package name and `test` is a very nice name for the rest of the world so I think we kind of have to live with that unless anyone has better ideas.

adamgfraser:
Changed `fail` to `failed` to be consistent with `ignored`.

Issue #1802, opened 2019-09-24
ZIO Test: Too many new lines printed during test runs


jdegoes:

adamgfraser:
Is there a description or example of undesired behavior for this?

adamgfraser:
@jdegoes Just wanted to follow up on this.

PR #2190, opened 2019-11-08
ZIO Test: Suspend Effects in testM

adamgfraser:
Currently `testM` accepts a `assertion: ZIO[R, E, TestResult]`. This works fine when the assertion really is an effect but can result in unintended behavior when the assertion contains side effects that are not properly suspended. For example, consider:

```scala
testM("test) {
  val x = 1
  println("x = " + x)
  ZIO.succeed(assert(x, equalTo(1))
}
```

This will only print "x = 1" once when run using a `main` method but twice when run with the SBT test runner. A similar issue can also come up when filtering tests, where the side effect can get executed even for tests that are supposed to be filtered out. You could argue that it is a contract violation to be creating effects that are not properly suspended like this but I think it is pretty common to want to do a debug statement in testing and users may not always properly suspend them so with this PR we suspend the entire assertion.

@jdegoes Is this the issue you identified in #1802?

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/TestSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestSpec.scala
@@ -4,6 +4,7 @@ import zio.ZIO
 import zio.clock._
 import zio.test.Assertion._
 import zio.test.TestAspect.failure
+import zio.test.TestUtils.execute
 
 object TestSpec
     extends ZIOBaseSpec(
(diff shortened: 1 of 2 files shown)


jdegoes:
@adamgfraser I don't think this is the issue. Actually I think the issue is caused by some combination of Metals / VS Code, because I have observed it with non-ZIO Test runs inside VS Code.

jdegoes:
Great change though and perfect implementation (using `effectSuspendTotal`).

adamgfraser:
Okay, well at least we fixed something. 😃

Is there anything else we should explore with regard to #1802 or can we close that?

jdegoes:
@adamgfraser Closed for now, if I can detect a pattern I'll re-open.

adamgfraser:
@jdegoes Okay, sounds good. If you can give me any lead, maybe an example that caused it, I am happy to investigate as well.

PR #2192, opened 2019-11-08
ZIO Test: Improvements to provideManagedShared

adamgfraser:
Makes two improvement to `provideManagedShared`. First, fixes a bug where an unchecked exception in the managed resource could cause the test framework to die instead of being caught and reported. Second, makes `provideManagedShared` more polymorphic by accepting a `ZManaged[R0, E1, R]`, allowing for partial provision of the environment and in particular accessing the `TestEnvironment` to help construct a managed resource. Technically I think this should be `provideManagedSomeShared` but that is a mouthful. Also since all tests require the `TestEnvironment` I think the value of demonstrating full elimination of environment requirements is less. But happy to change.

Also removes an unused method that I added at some point when we were making suites effectual but I don't think we need to be exposing at this point.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -227,13 +227,13 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
     }
 
   /**
-   * Uses the specified `Managed` to provide each test in this spec with its
+   * Uses the specified `ZManaged` to provide each test in this spec with its
    * required environment.
    */
-  final def provideManaged[E1 >: E](managed: Managed[E1, R])(implicit ev: NeedsEnv[R]): Spec[Any, E1, L, T] =
-    transform[Any, E1, L, T] {
-      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.provideManaged(managed), exec)
-      case TestCase(label, test)         => TestCase(label, test.provideManaged(managed))
+  final def provideManaged[R0, E1 >: E](managed: ZManaged[R0, E1, R])(implicit ev: NeedsEnv[R]): Spec[R0, E1, L, T] =
+    transform[R0, E1, L, T] {
+      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.provideSomeManaged(managed), exec)
+      case TestCase(label, test)         => TestCase(label, test.provideSomeManaged(managed))
     }
 
   /**
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -118,22 +118,22 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
       case SuiteCase(label, specs, exec) =>
         exec.getOrElse(defExec) match {
           case ExecutionStrategy.Parallel =>
-            specs.foldM(
-              e => f(SuiteCase(label, ZIO.fail(e), exec)),
+            specs.foldCauseM(
+              c => f(SuiteCase(label, ZIO.halt(c), exec)),
               ZIO
                 .foreachPar(_)(_.foldM(defExec)(f))
                 .flatMap(z => f(SuiteCase(label, ZIO.succeed(z.toVector), exec)))
             )
           case ExecutionStrategy.ParallelN(n) =>
-            specs.foldM(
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -118,22 +118,22 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
+            specs.foldCauseM(
+              c => f(SuiteCase(label, ZIO.halt(c), exec)),
               ZIO
                 .foreachParN(n)(_)(_.foldM(defExec)(f))
                 .flatMap(z => f(SuiteCase(label, ZIO.succeed(z.toVector), exec)))
             )
           case ExecutionStrategy.Sequential =>
-            specs.foldM(
-              e => f(SuiteCase(label, ZIO.fail(e), exec)),
+            specs.foldCauseM(
```
jdegoes: Great change!

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -227,34 +227,36 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
   /**
-   * Uses the specified `Managed` to provide each test in this spec with its
+   * Uses the specified `ZManaged` to provide each test in this spec with its
    * required environment.
    */
-  final def provideManaged[E1 >: E](managed: Managed[E1, R])(implicit ev: NeedsEnv[R]): Spec[Any, E1, L, T] =
-    transform[Any, E1, L, T] {
-      case SuiteCase(label, specs, exec) => SuiteCase(label, specs.provideManaged(managed), exec)
-      case TestCase(label, test)         => TestCase(label, test.provideManaged(managed))
+  final def provideManaged[R0, E1 >: E](managed: ZManaged[R0, E1, R])(implicit ev: NeedsEnv[R]): Spec[R0, E1, L, T] =
```
jdegoes: I'd be tempted to leave the old one, and have it delegate to a new `provideSomeManaged`, just for consistency with ZIO.

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -227,34 +227,36 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
   /**
-   * Uses the specified `Managed` once to provide all tests in this spec with
+   * Uses the specified `ZManaged` once to provide all tests in this spec with
    * a shared version of their required environment. This is useful when the
    * act of creating the environment is expensive and should only be performed
    * once.
    */
-  final def provideManagedShared[E1 >: E](managed: Managed[E1, R])(implicit ev: NeedsEnv[R]): Spec[Any, E1, L, T] = {
-    def loop(r: R)(spec: Spec[R, E, L, T]): ZIO[Any, E, Spec[Any, E, L, T]] =
+  final def provideManagedShared[R0, E1 >: E](
```
jdegoes: I'd probably call this `provideSomeManagedShared` even though it's getting horrendously long (just for consistency).

The other way to approach "sharing" would be a combinator, e.g.:

```scala
shared {
  ...
}
```

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: Okay. Do you want to do the same thing with `provideSomeShared` and have it delegate to `provideSomeManagedShared`? The only reason I did it this was was `provideSomeManagedShared` seemed really long and then I wanted to be consistent within Spec.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: Okay. Agree about the value of consistency.

adamgfraser:
Updated the names.

test/shared/src/main/scala/zio/test/Spec.scala
jdegoes: Right, I think ideally we'd achieve consistency within spec, and across Spec and ZIO. Your call though!

Another idea is introducing shorter aliases, both here and within ZIO. From my POV, it's really the `Shared` that pushes it over the edge into `almostUnreasonableMethodNameLength`. 😄 

But of course we need a shared variant for these things or a way to do it.

adamgfraser: Do you have an idea for what you would like the shorter aliases to be? Happy to add those across core, streams, and tests.

PR #2194, opened 2019-11-09
【WIP】Test: migrate DefaultTestReporterSpec

TapanVaishnav:


TapanVaishnav:
@adamgfraser Hi, I'm trying to migrate `DefaultTestReporterSpec` but as this one requires to actually test the reports from the console and the code for that is a bit messier(I believe), Can you guide me in a direction to make it more readable.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
-package zio.test
```
adamgfraser: You want to keep this within the `zio.test` package object.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
-
-import zio._
 import zio.clock.Clock
-import zio.test.environment._
 import zio.test.Assertion._
 import zio.test.ReportingTestUtils._
-import zio.test.TestUtils.label
+import zio.test._
```
adamgfraser: You shouldn't need this since you are in the same package.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
+import zio.test.environment.{TestClock, TestConsole, TestEnvironment, testEnvironmentManaged}
```
adamgfraser: This needs a `fmt`.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
+        testM("report success") {
+          assert(ZIO.effect(DefaultTestReporterSpecUtil
+            .check(test1, Vector(test1Expected, reportStats(1, 0, 0)))), isFalse)
+        }
+      )
+    )
+
+object DefaultTestReporterSpecUtil {
 
   def makeTest[L](label: L)(assertion: => TestResult): ZSpec[Any, Nothing, L, Unit] =
```
adamgfraser: This method isn't really doing anything. We can just replace all occurrences of `makeTest` with `test` using the existing method in the `zio.test` package object.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
     zio.test.test(label)(assertion)
 
   val test1 = makeTest("Addition works fine") {
-    assert(1 + 1, equalTo(2))
+    assert(1 + 0, equalTo(2))
```
adamgfraser: I don't know if you were just testing here but I would not change the actual assertions.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -110,35 +107,7 @@ object DefaultTestReporterSpec extends AsyncBaseSpec {
...
-    )
-
-  def simpleAssertion =
-    check(test5, test5Expected :+ reportStats(0, 0, 1))
-
-  def multipleNestedFailures =
-    check(test6, test6Expected :+ reportStats(0, 0, 1))
-
-  def check[E](spec: ZSpec[TestEnvironment, String, String, Unit], expected: Vector[String]): Future[Boolean] =
+  def check[E](spec: ZSpec[TestEnvironment, String, String, Unit], expected: Vector[String]) =
```
adamgfraser: Eliminate the `unsafeRunWith` and just have this return a `ZIO`. Delete the `unsafeRunWith` method below. Don't take a parameter for the expected result and instead just have the `ZIO` return the actual output. Rename so people don't confuse it with the `check` methods for property based testing. Maybe `runLog`.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
-    label(reportSuite2, "correctly reports failed test suite"),
-    label(reportSuites, "correctly reports multiple test suites"),
-    label(simpleAssertion, "correctly reports failure of simple assertion"),
-    label(multipleNestedFailures, "correctly reports multiple nested failures")
-  )
+object DefaultTestReporterSpec
+    extends ZIOBaseSpec(
+      suite("DefaultTestReporterSpec")(
+        testM("report success") {
+          assert(ZIO.effect(DefaultTestReporterSpecUtil
```
adamgfraser: Since the new `runLog` method will be returning a `ZIO`, we can get rid of the `ZIO.effect` call. I would import the `DefaultTestReporterSpecUtil` object so you don't need to type it out every time. Then since `runLog` will be returning a `Vector[String]` you can assert that the result is the same as the expected result.

test-tests/shared/src/test/scala/zio/test/DefaultTestReporterSpec.scala
```diff
@@ -1,32 +1,29 @@
...
+            .check(test1, Vector(test1Expected, reportStats(1, 0, 0)))), isFalse)
+        }
```
adamgfraser: Assume this is just work in process but we need to add the other tests here.

TapanVaishnav:
@adamgfraser Hi Adam, thank you for reviewing the PR, I'll resolve all the comments.

TapanVaishnav:
@adamgfraser I tried removing the `unsaferun` and added a simple test but somehow it's not working, can you please take a look at this?

adamgfraser:
@TapanVaishnav `TestClock.make` returns a `Managed` now so you don't need to call `toManaged_` on it anymore. `runLog` returns an effect so you need to either use `flatMap`, a for comprehension, or `assertM` to compare the result of the effect to the expected result, rather than the effect itself to the expected result. The following test passes on `master`:

```scala
package zio.test

import zio._
import zio.clock.Clock
import zio.test.Assertion._
import zio.test.ReportingTestUtils._
import zio.test.environment.{ testEnvironmentManaged, TestClock, TestConsole, TestEnvironment }

object DefaultTestReporterSpec
    extends ZIOBaseSpec(
      suite("DefaultTestReporterSpec")(
        testM("correctly reports a successful test") {
          val x             = zio.test.test("Addition works fine")(assert(1 + 1, equalTo(2)))
          val test1Expected = expectedSuccess("Addition works fine")
          assertM(
            DefaultTestReporterSpecUtil.runLog(x),
            equalTo(Vector(test1Expected, DefaultTestReporterSpecUtil.reportStats(1, 0, 0)))
          )
        }
      )
    )

object DefaultTestReporterSpecUtil {

  def reportStats(success: Int, ignore: Int, failure: Int) = {
    val total = success + ignore + failure
    cyan(
      s"Ran $total test${if (total == 1) "" else "s"} in 0 ns: $success succeeded, $ignore ignored, $failure failed"
    ) + "\n"
  }

  def runLog[E](spec: ZSpec[TestEnvironment, String, String, Unit]) = {
    val zio = for {
      _ <- TestTestRunner(testEnvironmentManaged)
            .run(spec)
            .provideSomeManaged(for {
              logSvc   <- TestLogger.fromConsoleM.toManaged_
              clockSvc <- TestClock.make(TestClock.DefaultData)
            } yield new TestLogger with Clock {
              override def testLogger: TestLogger.Service = logSvc.testLogger

              override val clock: Clock.Service[Any] = clockSvc.clock
            })
      output <- TestConsole.output
    } yield output
    zio
  }

  def TestTestRunner(testEnvironment: Managed[Nothing, TestEnvironment]) =
    TestRunner[TestEnvironment, String, Unit, String, Unit](
      executor = TestExecutor.managed[TestEnvironment, String, String, Unit](testEnvironment),
      reporter = DefaultTestReporter()
    )
}
```

TapanVaishnav:
> runLog returns an effect so you need to either use flatMap, a for comprehension, or assertM to compare the result of the effect to the expected result, rather than the effect itself to the expected result.

I see, guess I need to read the documentation again. Thank you so much.

TapanVaishnav:
Hi Adam, as always thank you so much for your review, I'll update the spec and will tag you for the final review.

TapanVaishnav:
@adamgfraser Hi, I've migrated all the tests(coz of your help obviously), can you again have a look over the PR and after that, I believe we should be good to go.

PR #2199, opened 2019-11-09
Make ZStream.unfold safe to pull again

vasilmkd:
@iravid

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2512,26 +2512,6 @@ object ZStream extends Serializable {
         case Take.Fail(e)  => halt(e)
         case Take.End      => end
       }
-
-    /**
-     * A managed [[Pull]] constructor that monitors the state of the wrapped [[Pull]].
-     * In case of error or stream end, it sets a guard that makes sure that any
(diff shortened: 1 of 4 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3066,19 +3066,25 @@ object ZStream extends Serializable {
...
-        .flatMap(f0)
-        .foldM(
-          e => Pull.fail(e),
-          opt =>
-            opt match {
-              case Some((a, s)) => ref.set(s) *> Pull.emit(a)
-              case None         => Pull.end
-            }
-        )
+        done <- Ref.make(false).toManaged_
```
iravid: So we can use `Pull.memoizeEnd` here right?

```
Pull.memoizeEnd {
  Ref.make(s).toManaged_.map {
    // previous code here
  }
}
```

vasilmkd:
I started disliking `memoizeEnd` because it also captures the error, so it's not completely correct in this case. We should resume normally after a potential error in `f0`, not end.

iravid:
So let's fix it, by not capturing the error, or delete it?

vasilmkd:
I vote for deleting

PR #2204, opened 2019-11-10
Add .DS_Store to .gitignore

serragnoli:
Sometimes I browse the file system with `Finder` and end up with a clutter of these `.DS_Store` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ test-output/
 .metals
 .idea
 coursier
+.DS_Store
 
 # if you are here to add your IDE's files please read this instead:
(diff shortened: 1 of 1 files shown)


softinio:
Git allows you to have your own global .gitignore. This I find better to include in that as this happens on any git project you may work on. 

The way I have it setup is in my mac home directory I have a `.gitignore` file. Then in the same home directory in `.gitconfig` I point to it like this: 

https://github.com/softinio/dotfiles/blob/master/git/gitconfig-me#L9

Think you will lead a happier life if you set this up whether this PR is merged or not :-)

PR #2210, opened 2019-11-11
ZIO Test: Prevent Spurious TestClock Warnings

adamgfraser:
Since we added warnings to the test clock I have been seeing some spurious warnings in our internal tests. It appears this is related to tests that that use the `TestClock` to do sleeping but complete without adjusting it (e.g. delaying an effect, forking it, then checking that it can be interrupted). To prevent this, this PR has making a `TestClock` return a `Managed` where the release action is canceling the warning.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/TestRuntime.scala
+++ b/core-tests/jvm/src/test/scala/zio/TestRuntime.scala
@@ -46,4 +46,7 @@ abstract class TestRuntime(implicit ee: org.specs2.concurrent.ExecutionEnv)
 
   def unsafeRunWith[R, E, A](r: UIO[R])(zio: ZIO[R, E, A]): A =
     unsafeRun(r.flatMap[Any, E, A](zio.provide))
+
+  def unsafeRunWithManaged[R, E, A](r: UManaged[R])(zio: ZIO[R, E, A]): A =
+    unsafeRun(zio.provideManaged(r))
 }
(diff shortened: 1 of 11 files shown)


adamgfraser:
Thinking about this more I am not really happy with this change. It feels like we are leaking too many implementation details by having to change the signature of `TestClock.make` and introducing irregularity with the constructors for the other test implementations. It looks like a good amount of the issue is that sometimes an effect is put to sleep in an uninterruptible block which by default makes the warning fiber uninterruptible. That is a very easy fix that we can do purely internally. Then any other warnings would be cases where `sleep` was called but `adjust` or its equivalents was never called, in which case `sleep` could just be replaced by `never` so we don't need to change all of our signatures to support that. I will update shortly.

adamgfraser:
@jdegoes Sorry I had some second thoughts about this that I was in the process of making.

The version we merged has creating a `TestClock` return a `Managed`, so advantage is that warning will get canceled if test completes even though sleep is used and clock is not adjusted. Disadvantage is that `TestClock.make` now returns a `Managed` unlike the constructors for all the other test implementations and is a little harder to use.

I was thinking of going with a more minimalistic change where we just made the warning fiber interruptible (it could be uninterruptible before if the clock adjustment occurred in an uninterruptible region). Disadvantage is warning could still be displayed if a test completed with a sleep and no adjustment, but in that case the sleep effect could just be replaced with `never` (though that could be harder if the `sleep` is composed into some larger effect you are calling).

I see the pluses and minuses of both and am happy either way, just want to make sure we are aligned.

Issue #2208, opened 2019-11-10
[error] Referring to non-existent class java.time.ZoneId$


serragnoli:
First time I'm running ZIO's `test` from `master` and I'm getting: [sbt-test-logs.txt](https://github.com/zio/zio/files/3828812/sbt-test-logs.txt)

I did a little digging and found a closed issue https://github.com/zio/zio/issues/1934 with a PR to fix it.
Are the errors I'm seeing due to something wrong with my env or an actual issue?

softinio:
Can you put the contents of your log in a gist or just paste in this issue?

serragnoli:
@softinio
Thanks for replying.
Here is the [GIST](https://gist.github.com/serragnoli/920cd860720d31ba0303b39ab93994c7l) with more details.

adamgfraser:
@serragnoli Are you using ZIO Test on ScalaJS? If so you need to add a dependency that provides the `java.time` package. You can use [scala-java-time](https://github.com/cquiroz/scala-java-time) for this.

serragnoli:
@adamgfraser 
I followed this [section](https://github.com/zio/zio/blob/master/docs/about/contributing.md#prepare-your-code) of the contributor's guide.
The [dependency you mentioned](https://github.com/zio/zio/blob/master/build.sbt#L123) exists in `master` plus the [plugin](https://github.com/zio/zio/blob/master/project/plugins.sbt#L5).

What am I missing please?

adamgfraser:
@serragnoli Sorry, I should have looked at the logs more closely. It looks like you are trying to run the tests for ZIO streams on ScalaJS, is that right? Those tests haven't been enabled on ScalaJS yet and I believe `streamsTests` doesn't have that dependency currently.

serragnoli:
@adamgfraser 
I am planning to start working on one of the open issues and I followed the contributors guide to verify whether my environment is OK. Therefore, I started SBT from the root of the project and did a  `clean; compile; test:compile; test` and noticed the errors in the GIST.

At the minute I'm under the impression the documentation may need some tweaking on how to avoid/ignore those errors to help newcomers (like me) but I'm not too sure what kind of tweak.

serragnoli:
I'll clock out for a couple of hours now, but I'll be back afterwards.

adamgfraser:
Sounds good. I will see if I can reproduce locally and come up with a recommendation. I will also be on Discord and can help there when you are back.

adamgfraser:
[...]

PR #2211, opened 2019-11-11
Add library dependency to avoid JS test errors

serragnoli:
#2208

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -149,6 +149,9 @@ lazy val streamsTests = crossProject(JSPlatform, JVMPlatform)
 lazy val streamsTestsJVM = streamsTests.jvm.dependsOn(coreTestsJVM % "test->compile")
 
 lazy val streamsTestsJS = streamsTests.js
+  .settings(
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test
+  )
 
 lazy val test = crossProject(JSPlatform, JVMPlatform)
   .in(file("test"))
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -164,7 +167,10 @@ lazy val test = crossProject(JSPlatform, JVMPlatform)
   )
 
 lazy val testJVM = test.jvm.settings(dottySettings)
-lazy val testJS  = test.js
+lazy val testJS = test.js
+  .settings(
```
adamgfraser: I don't think we want to add this. This will force everyone using ZIO Test for ScalaJS to bring in the `scala-java-time` dependency instead of bringing their own. Whether we should take a different approach to our zero dependencies policy here is a good question, but for this PR let's just fix the issue so contributors don't have an issue compiling the project.

serragnoli: Fair enough about the dependency. Alternatively (or in addition) to this PR, I could make a note of this in the contributors README. What do you think?

adamgfraser: Well I think once we merge this it should just work for new contributors working on the project, right? The need to bring your own time dependency should only apply to end users depending on ZIO Test.

serragnoli: Correct, the dependencies fixed the errors I reported yesterday in my env.
For end users, the documentation about bringing your own dependency is [here](https://github.com/zio/zio/blob/master/docs/interop/javascript.md)

adamgfraser: Yes, so let's just add the dependency to the streams tests and merge this.

PR #2215, opened 2019-11-11
ZIO Test: Support Assertions about Compilation Errors

adamgfraser:
Currently with `assertCompiles` we allow users to assert whether code compiles but we don't make it easy to make assertions about the resulting compilation errors. This can be less than ideal because code can fail to compile for other than the intended reasons (e.g. because of a typo or refactoring). To support this we change from `assertCompiles` to `compile`:

```scala
def compile(code: String): UIO[Either[String, Unit]]
```

`compile` can either return a `Right` if the code is valid Scala code or a `Left` with an error message if it is not. Example usage:

```scala
testM("compile must return Left with an error message otherwise") {
  val result = compile("1 ++ 1")
  val expected = "value ++ is not a member of Int"
  assertM(result, isLeft(equalTo(expected)))
}
```


mschuwalow:
Very nice! I think this will have much better usability for users

jdegoes:
This is a great direction! I wondered how many `assertCompiles` tests were made useless by my recent refactorings.

Maybe instead of a `String` message, which is somewhat fragile, we could do:

```scala
sealed trait CompileError
final case object SyntaxError extends CompileError
final case object ResolutionError extends CompileError
final case object TypeCheckError extends CompileError
```

or something similar? In fact, maybe just `assertTypeChecks` would be sufficient, since most times a user will want it to pass syntax checking and resolution (finding all symbols, etc.).

adamgfraser:
@jdegoes Yes I think that is definitely the direction we want to go. The challenge is that I think users want to be able to check that code didn't type check for a particular reason (e.g. because this combinator doesn't make sense for effects that can't fail). And what we get back is either a `TypecheckException` or a `ParseException` with a message so we don't have as structured information as we would like.

So it seems to me that you want users to be able to know whether code type checked or not, and if not why, and then push other failures like not parsing into the runtime failure channel. I thought there was some advantage to not having a separate algebraic data type so users can reuse existing assertions but very open to feedback on the best API for this feature.

jdegoes:
Sounds good. I suppose we can deal with "compiler-specific" messages in other ways, e.g. combinators for assertions `assertion1.jvmOnly && assertion2.dottyOnly` or something TBD.

adamgfraser:
I like your idea of `typeChecks`. Makes it clearer what the contract is.

ioleo:
:tada:

PR #2217, opened 2019-11-12
Fixed #2216 - Chunk.size always returns 0

### LINKED ISSUE #2216: Chunk.size always returns 0
```scala
println(Chunk.fromArray(Array(1, 2, 3)))                      // Arr(1,2,3)
println(Chunk.fromArray(Array(1, 2, 3)).length)               // 3
println(Chunk.fromArray(Array(1, 2, 3)).size)                 // 0
println((Chunk.fromArray(Array(1, 2, 3)) ++ Chunk(4)).length) // 4
println((Chunk.fromArray(Array(1, 2, 3)) ++ Chunk(4)).size)   // 0
```

I think its because `size` is defined as a `final val` that references `length`. Will test the hypothesis.

ajaychandran:
Changed `Chunk.size` from `val` to `def`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -1,0 +1,40 @@
+package zio
+
+import zio.test.{ assert, suite }
+import zio.test.Assertion.equalTo
+
+object ChunkSpec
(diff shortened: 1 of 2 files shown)


jdegoes:
Thanks @ajaychandran !

PR #2222, opened 2019-11-12
Not for merge (yet): Simple InMemoryCache implementation

sirthias:
Judging from the discord channel archives it appears as if the topic of how to implement an in-memory-cache with ZIO has already come up several times.

This PR proposes a simple implementation that is meant as a starting point for a discussion on how to best implement such a cache and whether an addition to ZIO core even makes sense.

At least these things are still missing:

- [ ] Decision on general value of inclusion in ZIO core
- [ ] Interruption handling
- [ ] Tests
- [ ] Scaladocs + Docs

I'll happily take care of the last two points after having received some feedback on the first two.
Especially regarding the second point (how to handle/fence-in/propagate interruptions) I think I'd need some help.

What are the points that one generally has to look out for wrt interruptions?
Are there any gotchas that frequently come up?
Have some patterns already emerged wrt to problematic constructs as well as solutions?

### DIFF
--- a/core/shared/src/main/scala/zio/InMemoryCache.scala
+++ b/core/shared/src/main/scala/zio/InMemoryCache.scala
@@ -1,0 +1,49 @@
+package zio
+
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.tailrec
+
+final class InMemoryCache[E, K, V] private (private val ref: AtomicReference[Map[K, Promise[E, V]]])
+    extends AnyVal
+    with Serializable {
+
+  def clear: UIO[Unit] =
+    IO.effectTotal(ref.set(Map.empty))
+
+  def get(k: K): UIO[Option[IO[E, V]]] =
+    IO.effectTotal(ref.get.get(k).map(_.await))
+
+  def set(k: K, value: IO[E, V]): UIO[Unit] =
+    for {
+      p <- Promise.make[E, V]
+      _ <- p.complete(value).fork
+      _ <- IO.effectTotal {
+            @tailrec def rec(current: Map[K, Promise[E, V]]): Unit = {
+              val updated = current.updated(k, p)
+              if (!ref.compareAndSet(current, updated)) rec(ref.get)
+            }
+            rec(ref.get)
+          }
+    } yield ()
+
+  def getOrUpdate(k: K)(f: K => IO[E, V]): IO[E, V] =
+    for {
+      map <- IO.effectTotal(ref.get)
+      value <- map.get(k) match {
+                case Some(p) => p.await
+                case None =>
+                  Promise.make[E, V].flatMap { p =>
+                    if (ref.compareAndSet(map, map.updated(k, p))) {
+                      f(k).to(p).fork *> p.await
+                    } else getOrUpdate(k)(f) // someone interfered with our update, so try again
+                  }
+              }
+    } yield value
+}
+
+object InMemoryCache extends Serializable {
+
+  final def make[E, K, V]: UIO[InMemoryCache[E, K, V]] =
+    IO.effectTotal(new InMemoryCache[E, K, V](new AtomicReference(Map.empty)))
+}


plokhotnyuk:
@sirthias Please, add to the missing things list an eviction strategy with some safe defaults to avoid out of memory errors.

Also, please, take in account that [Scala's `HashMap`/`HashSet` are vulnerable](https://github.com/scala/bug/issues/11203) on untrusted input under DoS/DoW attacks that exploit hash collision vulnerabilities of them directly or indirectly through Scala collections methods that use them under hood: `Map.empty`, `toMap`, `keys`, `distinct`, `groupBy`, etc..

sirthias:
Thank you for the feedback, @plokhotnyuk!
Yes, some kind of eviction logic would be an important feature, but whether it's a _required_ feature for the very first implementation should be one point in the discussion of the general merit of having a cache implementation in zio-core at all.

It might well be that the decision is that having a cache impl in zio-core would be beneficial but only if it includes eviction logic right out of the gates.

In my use case (where this impl is currently running) the cache already adds all the value I need without any eviction logic.
In other cases it might suffice to simply clear the cache in regular or irregular intervals.

I am not familiar enough (yet) with ZIO, its community and the goals and decision making processes of the core maintainers to be able to form a clear opinion on things here myself...

iravid:
My 2 cents: I'd love to have this in core along with all the features needed to drop the `scaffeine` dependency from my projects.

The entire implementation, including eviction strategies, will probably have a sufficiently small code footprint.

regiskuckaertz:
Isn't that a `ZRef` in disguise (or even a `TMap`)? 🤔 imo, once #1198 is released this use case and many more will become easier to express. I'm 👎 on that as this seems too simple to be in core, but having a purely functional cache with matching scaffeine/guava performance would be a blessing.

sirthias:
@regiskuckaertz I wasn't able to implement the `getOrUpdate` method with a `Ref` alone, since the "construct the value if not cached" function is effectful and all the atomic modify/update methods on `Ref`s aren't.

> I'm 👎 on that as this seems too simple to be in core

Ok. What features would have to be added in order for you to consider a cache valuable enough?

iravid:
I don’t think this can be expressed with TMap either, as you’d need to interleave arbitrary effects.

regiskuckaertz:
> I don’t think this can be expressed with TMap either, as you’d need to interleave arbitrary effects.

@jdegoes hinted at a whole class of transactional effects which I think would be more appropriate here, but I may be wrong.

@sirthias Yeah, currently you'd have to use `RefM`, but `ZRef` will eventually subsume it.

> What features would have to be added in order for you to consider a cache valuable enough?

It would have to be competitive with industry-grade alternatives (agrona, guava, etc.), I guess. (CPU) Cache friendliness, eviction strategies, telemetry, etc.

sirthias:
Ah, yes, `RefM` should do it.
I hadn't even see it yet.
Thank you!

>> What features would have to be added in order for you to consider a cache valuable enough?
>
> It would have to be competitive with industry-grade alternatives (agrona, guava, etc.), I guess. > (CPU) Cache friendliness, eviction strategies, telemetry, etc.

Ok, that is an entirely reasonable stance but I think that with this scope a cache implementation should better live its own module or even project, as the release cycles should probably be independent from each other.

regiskuckaertz:
@sirthias I'm reopening as this is an entirely valid question to ask and you've gone through the effort of defining an initial implementation, I think this is fantastic that you've done that so please don't rely on my single opinion to make a judgment. Also, I don't check Discord quite often and there may have been a whole conversation which I'm not aware of 🤗

sirthias:
@regiskuckaertz No worries at all!
I kind of expected your take that inclusions to ZIO should be competitive with existing alternatives. This only shows the professional aspirations of the project, which is of course a good thing!
At this point, however, I don't have the capacity to drive an effort to implement a full-blown cache implementation with all the bells and whistles, especially with regard to benchmarking and perf optimizations (which, in my experience, can be a huge time sink).

This PR was simply meant as a vehicle for floating an initial, naive implementation that might serve as a starting point for sth better or even just a discussion, and maybe learn sth about interruption in the process... :)

> Also, I don't check Discord quite often and there may have been a whole conversation which I'm not aware of...

I don't think you missed anything there. I simply searched the archives and saw questions around implementing caching coming up several times, but without any deeper discussions...

jdegoes:
I'd love this! And am willing to help on design. It would ideally be super low level and optimized like our zero-dep ringbuffer.

core/shared/src/main/scala/zio/InMemoryCache.scala
```diff
@@ -0,0 +1,49 @@
+package zio
```
vasilmkd: Make sure you copy the licence from some other file.

core/shared/src/main/scala/zio/InMemoryCache.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.annotation.tailrec
+
+final class InMemoryCache[E, K, V] private (private val ref: AtomicReference[Map[K, Promise[E, V]]])
```
vasilmkd: I see this implementation is very close to the `Promise`, or `Ref` implementations. I wonder if this is a higher level concept and could be represented in terms of some other, lower level primitive.

core/shared/src/main/scala/zio/InMemoryCache.scala
sirthias: Yeah, sorry, I'm used to `sbt-header` based "auto-headering", so I never really think about these kinds of things anymore.
Should ZIO maybe rely on sbt-header as well?

sirthias:
@jdegoes What feature set do you envision for such an implementation?
Competing with sth like caffeine on features and/or perf will be quite a challenge and likely warrant its own project outside of the zio codebase, I'd imagine.
But it'd sure be interesting to see how ZIO's fiber-based concurrency primitives fare wrt implementation complexity and bottom line perf in comparison w/ caffeine.

ben-manes:
good luck! Always fun too see new ideas thrown at an old, gnarly problem.

jdegoes:
@sirthias After thinking through a possible design for such a project, I too think that it is best done outside ZIO, in `ZIO Cache`. I think the combination of a low-level, highly-performant cache, together with a declarative and functional API design on top, would be much more attractive to Scala users than wrappers around caffeine. But it's clear it will be a few thousand lines of code to get there. Please let me know if you'd like to help work on that project! Would love a hand.

ben-manes:
> After thinking through a possible design for such a project, I too think that it is best done outside ZIO, in ZIO Cache

What are the possible designs and how would you incorporate ZIO to offer functionality not available as a Caffeine wrapper? I'd be very curious to learn your needs and see an original take on the problem.

PR #2223, opened 2019-11-12
Fix chunk tests

iravid:
ZIO#zipWith became lazy in the RHS in the Autosupervision branch, and the implementations in Chunk were relying on side-effects.

cc @saraiva132


saraiva132:
Thanks for the fix @iravid

PR #2228, opened 2019-11-13
Add pretty print for dump

jdegoes:


jdegoes:
@iravid Agreed! We should integrate that into zmx, a machine readable mode for all the dumps and metrics.

softinio:
> @iravid Agreed! We should integrate that into zmx, a machine readable mode for all the dumps and metrics.

I'll work on that next week

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -417,7 +418,55 @@ object Fiber {
...
+        case Running => "Running"
+        case Suspended(interruptible, epoch, _, asyncTrace) =>
+          val in = if (interruptible) "interruptible" else "uninterruptible"
+          val ep = s"${epoch} asyncs"
+          val as = asyncTrace.map(_.prettyPrint).mkString(" ")
+          s"Suspended(${in}, ${ep}, ${as})"
+      }
+
+      s"""
+      ${name}#${fiberId.seqNumber} (${lifeMsg}) ${waitMsg}}
```
regiskuckaertz: Don't you have to use `stripMargin` here?

jdegoes: Thank you! 🙏

regiskuckaertz: No worries 🤗  You will have to prefix with `|` every line though, like this:
```scala
s"""
   |${name}#${fiberId.seqNumber} (${lifeMsg}) ${waitMsg}}
   |  Status: ${statMsg}
   |    ${trace.prettyPrint}
   |""".stripMargin

PR #2229, opened 2019-11-13
ZIO Test: Support Pairwise Combination of Generators

adamgfraser:
Almost all of our existing combinators for `Gen` are based on `flatMap`, which does a cartesian product of the elements in the two generators. This doesn't matter much for random generates and generally works great, but for deterministic generators it can be desirable to combine values from two generators in more of a pairwise zipping fashion.

This PR implements a new combinator `zipWithPar` that conceptually does this. The guarantee is that that the combined generator will always include all elements from the "longer" generator, pulling again from the "shorter" generator if necessary.


adamgfraser:
@iravid It would be great to align with Stream. Two potential concerns:

1. What is the right default when someone who hasn't really thought about it calls `zip` or `zipWith`? We can shrink generators composed this way more efficiently since we can shrink each generator independently so that is a point in favor. But it may also lead to some unintuitive results depending on your expectations as in the example below.
2. It is inconsistent with the "zip is derived from flatMap" principle from Core.

```scala
val booleans = Gen.fromIterable(List(true, false)) // deterministic generator
val boolPair = booleans.zip(booleans) // List((true, true), (false, false))???

**Comment by adamgfraser** [comment]
I implemented an optimized shrinking algorithm for parallel generators. This is dramatically more efficient for applications such as creating large case classes where generators do not depend on each other. Because we know the generators do not depend on each each other we can shrink the generators in parallel instead of shrinking the first generator and shrinking the second generator based on that, which conceptually has M + N complexity instead of M * N.

One thing we should think about is whether we want to do this as an optimization for the normal `zip`. Conceptually we have three combinators:

- `flatMap` - constructs a generator that depends on another generator. When combining two generators returns cartesian product. Less efficient shrinking.
- `zipWith` - combines two generators that don't depend on each other to return their cartesian product. Implemented in terms of `flatMap` but could be implemented to continue to return product but use more efficient shrinking algorithm.
- `zipWithPar` combines two generators that don't depend on each other to return their pairwise combination. More efficient shrinking.

**Comment by iravid** [comment]
@adamgfraser 

My intuition is that for type constructors that represent something with more than one value, like streams, lists and gens - the intuitive operation for `zipWith` is actually point-wise zipping.

It is indeed inconsistent with the `zip <-> flatMap` law, but I don't think that really matters.

**Comment by adamgfraser** [comment]
@iravid I think you are 100% right about streams and lists. I'm just not sure whether people think about generators the same way, since you can conceptualize them either as a collection of values or as a single effect, and a lot of other property based testing frameworks treat them more like a single effect. I don't have a strong feeling on this but that is my uncertainty.

**Comment by iravid** [comment]
Ok, no worries- just throwing this thought in the mix. I can testify that I personally think of `genA zip genB` as zipping two lists together. But that's just me :-)

**Comment by mschuwalow** [comment]
I would have the same intuition as @iravid here

**Comment by jdegoes** [comment]
I think this is incredibly useful. It does concern me a bit that we don't have internal consistency:

 - is `zipWith` sequential or parallel?
 - is `zipWith` pairwise or cartesian? (doesn't matter for `1-element` sources)
 - what are the names for the parallel variants? sequential variants?

But I suppose we can deal with those inconsistencies in a separate PR.

Personally because I view `Gen` as basically a stream, my intuition is that `gen1 zip gen2` would do pairwise zipping.

**Comment by adamgfraser** [comment]
Okay, sounds like there is consensus on the stream analogy and I was coming around to that as well. I will update accordingly.

File: core-tests/shared/src/test/scala/zio/ZIOSpec.scala [hunk 1]
```diff
@@ -766,7 +766,7 @@ object ZIOSpec extends ZIOBaseSpec {
           assert(unexpected, isEmpty) &&
           assert(result, isNone) // timeout happens
         }
-      },
+      } @@ flaky,
```
jdegoes: Does this need a ticket?

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -849,7 +850,7 @@ object StreamSpec extends ZIOBaseSpec {
                    ).flatMapParSwitch(2)(identity).runDrain.either
           cancelled <- substreamCancelled.get
         } yield assert(cancelled, isTrue) && assert(result, isLeft(equalTo("Ouch")))
-      },
+      } @@ flaky,
```
jdegoes: Does this need a ticket?

test-tests/shared/src/test/scala/zio/test/GenSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+  def spec = suite("GenSpec")(
+    suite("zipWith")(
+      testM("left preservation") {
+        checkM(deterministic, deterministic) { (a, b) =>
+          for {
+            left  <- sample(a.zip(b).map(_._1))
+            right <- sample(a)
+          } yield assert(left, startsWith(right))
+        }
+      } @@ scala2Only,
```
jdegoes: Not dotty?

test-tests/shared/src/test/scala/zio/test/GenSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+      testM("right preservation") {
+        checkM(deterministic, deterministic) { (a, b) =>
+          for {
+            left  <- sample(a.zip(b).map(_._2))
+            right <- sample(b)
+          } yield assert(left, startsWith(right))
+        }
+      } @@ scala2Only,
```
jdegoes: Not dotty?

jdegoes:
Minor comments but good to merge! Thanks for working on this.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
adamgfraser: Opened a ticket for it here: https://github.com/zio/zio/issues/2243.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
adamgfraser: Opened a ticket for it here: https://github.com/zio/zio/issues/2243.

test-tests/shared/src/test/scala/zio/test/GenSpec.scala
adamgfraser: Blocked by: https://github.com/zio/zio/issues/2232.

test-tests/shared/src/test/scala/zio/test/GenSpec.scala
adamgfraser: Blocked by: https://github.com/zio/zio/issues/2232.

PR #2230, opened 2019-11-13
Issue #2129 - ZSink.head and ZSink.last

### LINKED ISSUE #2129: Add ZSink.head and ZSink.last

pk044:
Resolves #2129 

not sure why the type was needed, e.g. - `ZSink.last[Int]` I wanted it to work by simply using `ZSink.last` but it didn't compile. any ideas? :)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1283,6 +1283,18 @@ object ZSink extends ZSinkPlatformSpecific with Serializable {
   final def drain: ZSink[Any, Nothing, Nothing, Any, Unit] =
     foldLeft(())((s, _) => s)
 
+  /**
+   * Creates a sink containing the first value.
+   */
+  final def head[A]: ZSink[Any, Nothing, A, A, Option[A]] =
```
jdegoes: There is no remainder so you should use `Nothing` for the `A0` slot. This will improve type inference, too.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1283,6 +1283,18 @@ object ZSink extends ZSinkPlatformSpecific with Serializable {
...
+    foldUntil[Option[A], A](None, 1) { case (_, a) => Some(a) }
+
+  /**
+   * Creates a sink containing the last value.
+   */
+  final def last[A]: ZSink[Any, Nothing, A, A, Option[A]] =
```
jdegoes: There is no remainder so you should use `Nothing` for the `A0` slot. This will improve type inference, too.

streams/shared/src/main/scala/zio/stream/ZSink.scala
pk044: @jdegoes `foldUntil` returns `ZSink[Any, Nothing, A, A, S]` I guess I should approach it differently then?

streams/shared/src/main/scala/zio/stream/ZSink.scala
pk044: I tried that, but using `ZSink.last` instead of  e.g. `ZSink.last[Int]` still results in the same compilation error as before.

streams/shared/src/main/scala/zio/stream/ZSink.scala
jdegoes: Even if it doesn't infer better, if a sink doesn't leave a remainder, it should still use `Nothing` for `A0`.

pk044: I think I can't use foldUntil here then (when I use `Nothing` for the `A0` slot, it doesn't compile) - I'll come up with something else

adamgfraser: Shouldn't the leftover type of `head` be `A`? `head` will only look at one element of the stream, so it will potentially return a leftover of type `A` if the stream has more than one element. Also, shouldn't this just be implemented as `identity.optional`. From the microsite: "Collecting the first element into an option (returns None for empty streams): `Sink.identity[Int].optional`".

adamgfraser: Also, I think we can simplify the signature of `run` from:

```scala
def run[R1 <: R, E1 >: E, A0, A1 >: A, B](sink: ZSink[R1, E1, A0, A1, B]): ZIO[R1, E1, B]
```

to:

```scala
def run[R1 <: R, E1 >: E, A1 >: A, B](sink: ZSink[R1, E1, Any, A1, B]): ZIO[R1, E1, B]
```

Since we are just discarding the leftovers the leftover type doesn't matter and it is covariant so we widen to `Any`.

pk044: > Also, I think we can simplify the signature of `run` from:
> 
> ```scala
> def run[R1 <: R, E1 >: E, A0, A1 >: A, B](sink: ZSink[R1, E1, A0, A1, B]): ZIO[R1, E1, B]
> ```
> 
> to:
> 
> ```scala
> def run[R1 <: R, E1 >: E, A1 >: A, B](sink: ZSink[R1, E1, Any, A1, B]): ZIO[R1, E1, B]
> ```
> 
> Since we are just discarding the leftovers the leftover type doesn't matter and it is covariant so we widen to `Any`.

should I make a new PR for that?

adamgfraser: I think it is fine to just include in this PR. It is a pretty minor change.

pk044: done!

PR #2236, opened 2019-11-13
Rewrite ZStream's concurrency combinators to use supervision

### LINKED ISSUE #2235: Flaky Stream Test
```
[info] ZIO Test
[info] - StreamSpec
[info]   - Stream.flatMapPar / flattenPar / mergeAll
[info]     - outer defects interrupt all fibers
[info] Timeout of 1 m exceeded. Couldn't interrupt test within 1 s, possible resource leak!
```

iravid:
Resolves #2235

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1200,10 +1200,9 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   ): ZStream[R1, E1, B] =
     ZStream[R1, E1, B] {
       for {
-        out             <- Queue.bounded[Pull[R1, E1, B]](outputBuffer).toManaged(_.shutdown)
-        permits         <- Semaphore.make(n.toLong).toManaged_
-        innerFailure    <- Promise.make[Cause[E1], Nothing].toManaged_
-        interruptInners <- Promise.make[Nothing, Unit].toManaged_
+        out          <- Queue.bounded[Pull[R1, E1, B]](outputBuffer).toManaged(_.shutdown)
+        permits      <- Semaphore.make(n.toLong).toManaged_
+        innerFailure <- Promise.make[Cause[E1], Nothing].toManaged_
 
         // - The driver stream forks an inner fiber for each stream created
         //   by f, with an upper bound of n concurrent fibers, enforced by the semaphore.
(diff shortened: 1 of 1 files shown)


iravid:
@jdegoes Would be nice to have a shortcut for `ZIO.children.flatMap(Fiber.interruptAll)`. Maybe `ZIO.interruptChildren`?

iravid:
Btw I'm actually not sure why the previous formulation was broken, but I couldn't figure it out in less than 30 minutes and I don't want to risk RC17 going out without a fix for this

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1219,44 +1219,52 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
                       // All fibers completed successfully, so we signal that we're done.
                       rightDone = (_, failureAwait) => out.offer(Pull.end) *> failureAwait.interrupt.unit
                     )
                     .toManaged_
               )
               // This finalizer makes sure that in all cases, the driver stops spawning new streams
               // and the inner fibers are signalled to interrupt and actually exit.
-              .ensuringFirst(interruptInners.succeed(()) *> permits.withPermits(n.toLong)(ZIO.unit))
+              .ensuringFirst(
+                stopIteration.set(true) *> ZIO.children.flatMap(Fiber.interruptAll) &> permits
```
regiskuckaertz: this is where I get lost a bit: isn't there a risk that `ZIO.children` picks up the fibers created in `&>` and kills the whole thing?

iravid: Hmm I did not consider this. Let me consult the implementation

iravid: Ok, good catch, this is a pretty subtle pitfall. It's not that `children` here picks up the fibers forked from `&>` but rather that the entire finalizer is run on a different fiber from the one that spawned the inner fibers.

So it's going to actually require an effort to make this interrupt the fibers. I'm wondering though if I should actually be doing this manually though.

iravid: Thanks Regis, this actually helped me simplify the combinators quite a bit. Will push the changes tomorrow.

regiskuckaertz: Ha, I'm thrilled this helped! Looking forward to see your simplification. Yes it feels like interrupting fibers when the parent is done is part of autosupervision.

iravid: Ok, there it is. We still have to interrupt them if one of them failed or if the outer stream failed. I think we can't rely on supervision there because it has to be done eagerly.

iravid: There's also a pitfall with accessing `ZIO.children` inside the continuation to `raceWith` - we have to capture those outside.

The scoping for `children` can be definitely confusing, as it is not lexical. Wonder if we can do something about that?

regiskuckaertz: Yeah I just saw your change, I see the subtlety there. Interruption on error makes definitely sense.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1288,28 +1285,29 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
                 _ <- cancelers.offer(canceler)
                 innerStream = Stream
                   .managed(permits.withPermitManaged)
-                  .flatMap(_ => Stream.bracket(latch.succeed(()))(_ => UIO.unit))
+                  .flatMap(_ => Stream.fromEffect(latch.succeed(())))
```
regiskuckaertz: `.tap(_ => latch.succeed(()))` :-)

iravid: Nice one!

iravid:
@regiskuckaertz Yes, that's a good observation. There's a generalization to the parent-child interruption that can be used to perform what you're saying: you can specify a region for fiber creation and an action to be performed when exiting that region. But I'm not sure if it's wise to introduce before 1.0.

PR #2249, opened 2019-11-15
Issue #2238 - adding access and accessM to ZStream

### LINKED ISSUE #2238: Add access / accessM to ZStream
Looking for the best way to access the environment of a `ZStream`.

We can currently do 

```scala
trait FooZIO {
    def fooZio: ZIO[Any, Nothing, Unit]
}

object FooZIO {
    def bar: ZIO[FooZIO, Nothing, Unit] = ZIO.accessM[FooZIO](_.fooZio)
}
```

But, it doesn't appear that we have the equivalent ability on `ZStream`. E.g.

```scala
trait FooZStream {
    def fooZStream: ZIO[Any, Nothing, Unit]
}

object FooZStream {
    def bar: ZStream[FooZStream, Nothing, Unit] = ???
}
```

I propose adding `access` and `accessM` to `ZStream`. but I'm open to other suggestions.

pk044:
Resolves #2238 

I was also thinking of having something like

 `final def access[R, A](f: R => A): ZStream[R, Nothing, A]`

but I'm not sure if that would result in too much boilerplate for the user?
e.g. when accessing the environment I'd not know what the output type would be from `f`, so that also can be cumbersome when changing calls to something else, since I'd have to fix the types afterwards.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -16,7 +16,7 @@
 
 package zio.stream
 
-import java.io.{ IOException, InputStream }
+import java.io.{IOException, InputStream}
 
 import com.github.ghik.silencer.silent
 import zio._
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2646,6 +2646,18 @@ object ZStream extends Serializable {
   final def apply[R, E, A](pull: ZManaged[R, E, Pull[R, E, A]]): ZStream[R, E, A] =
     new ZStream[R, E, A](ZStream.Structure.Iterator(pull))
 
+  /**
+   * Creates a stream from a value received from the accessed environment.
+   */
+  final def access[R](f: R => Any): ZStream[R, Nothing, Any] =
```
adamgfraser: I think you want to parameterize this on `A` as well and have `f: R => A` to return a `ZStream[R, Nothing, A]`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2646,6 +2646,18 @@ object ZStream extends Serializable {
   final def apply[R, E, A](pull: ZManaged[R, E, Pull[R, E, A]]): ZStream[R, E, A] =
     new ZStream[R, E, A](ZStream.Structure.Iterator(pull))
 
+  /**
+   * Creates a stream from a value received from the accessed environment.
+   */
+  final def access[R, A](f: R => A): ZStream[R, Nothing, A] =
```
adamgfraser: We should use the partial type application trick. This works but requires the user to specify both type parameters (e.g. `ZStream.accessM[Random, Int](_.random.nextInt)`). If we have `access` return a new `AccessPartiallyAppliedType` like in ZIO then we can allow this to be written as `ZStream.accessM[Random](_.random.nextInt)` and the return type will be inferred. This is more convenient for the user and still preserves the type information. Sorry didn't think of this earlier.

pk044: yeah - that's what I mentioned in the issue description, but I didn't know any solution to that. thanks, I'll try to do it as suggested :)

pk044:
done!

PR #2251, opened 2019-11-16
ZIO Test: Warn when no tests executed

### LINKED ISSUE #2245: testOnly happily runs things that do not exist
This is more a UX thing than anything.
When using `testOnly` in SBT, our test runner seems to just report a happy status if the intended test suite does not exist.
I don't think it should be this way and a "missing test suite" error should be returned.

```
sbt:Kafka Commons> testOnly Bla
[info] ScalaTest
[info] Run completed in 15 milliseconds.
[info] Total number of tests run: 0
[info] Suites: completed 0, aborted 0
[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
[info] No tests were executed.
[info] ZIO Test
[info] Done
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for Test / testOnly
[success] Total time: 0 s, completed Nov 15, 2019 11:10:06 AM
```

reibitto:
Resolves #2245

From what I can tell there isn't a way to fail the run if no tasks were executed (besides resorting to exceptions which doesn't feel appropriate here). I don't seem to have access to SBT's event handler since `Task.execute` won't even be called [here](https://github.com/zio/zio/blob/8359a6ab26f0e4ee0854a322d05945227d3b1ffb/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala#L42) in this case.

The best I could do was print a yellow warning of "No tests were executed". This is the same thing ScalaTest does. I'm guessing for the same reason I described above. Unless there's another approach that I'm not seeing?

@adamgfraser Also, is it safe to assume no summary = no tests executed? I guess I could add another atomic var for this, but I was trying to avoid that if it wasn't necessary.


adamgfraser:
I think this looks good. Can we display the same warning when the user does `testOnly` with a command line argument and no tests contain the specified string? Also, what do you think the behavior should be if all tests were ignored?

reibitto:
Yeah, I think that's reasonable. Because if the total of `success + ignore + failure` is 0, then that seems to indicate that something isn't right with the test command that was run (`-t` args that don't make sense).

If all tests were ignored though, you'll see output like "2 ignored" and so on, so that _might_ be enough. At worst, the user probably just forgot to remove some `TestAspect.ignore`. It might not hurt to show a warning though. I went with that, but I could see the argument for not showing any warnings in that case.

adamgfraser:
Looking good. A couple of comments:

1.  Can we move the warning from the reporting of each spec to the summary? I think if we run four test files and one of them is completely ignored there is no need for a warning. But if we run four files and we haven't had any successful or failed tests across them a warning is appropriate.
2. Can we display the same warning message in both cases? That way if the user does something like `Spec.only` that results in no tests being executed it will be the same as doing `testOnly` in SBT.
3. Can we get this working on ScalaJS too?

reibitto:
Oops, I forgot there is a separate runner for JS. I'll add it for that too.

As for warning at the end as a summary (not per spec), that looks doable, just slightly trickier. From the looks of it, I think I have to accumlate each SBT Task's stats (number of tests ignored and so on) in ZTestRunner much like how it's being done with with the summaries `Vector[String]`. Then sum them up in the end in `ZTestRunner.done()` and if the total is still zero return the "No tests were executed" string.

I'm thinking something like:

```scala
case class TestTaskStats(success: Int, failure: Int, ignored: Int)
...
val testTaskStats: AtomicReference[Vector[TestTaskStats]] = new AtomicReference(Vector.empty)
```

Or instead of having another AtomicReference in addition to the summaries, I could pull in the summary string into this new case class?

```scala
case class TestTaskStats(success: Int, failure: Int, ignored: Int, summary: String)
```

Is that alright? Any possible better alternatives you can see?

adamgfraser:
I'm wondering if we should make `SummaryBuilder` return something more structured than a string that contains rendered failures and statistics. Then it would be available to both the `done` method in `ZSpecRunner` as well as to the `main` method in `RunnableSpec`.

reibitto:
Hmm, yeah, I think that works and may result in less code actually. I can give it a shot. Thanks!

reibitto:
Still a work in progress. I mean, it seems to work for the commands I tested (both JVM and JS), but the JS side presented some problems [here](https://github.com/zio/zio/blob/b7ae66eadb1c9b59ce6a1f4d9b4fb4741919ec2a/test-sbt/js/src/main/scala/zio/test/sbt/ZTestFramework.scala#L35) and [here](https://github.com/zio/zio/blob/b7ae66eadb1c9b59ce6a1f4d9b4fb4741919ec2a/test-sbt/js/src/main/scala/zio/test/sbt/ZTestRunner.scala#L49). It forces `String`. The thing is, I don't understand when these are actually used. I think I need to understand what this master/slave deal is with JS.

Also, there is some repetition of code in `buildSummary`. I can extract that out to a method. Or doing it in one pass (just haven't thought about how to do that nicely yet).

reibitto:
Alright, there's room for improvement but for now I just went with a simple tab-delimited format (and making sure to escape/unescape them).

I also cleaned up the code a little.

And I ran into some leftovers from the unifying of type params PR (#2180), so I changed those too.

PR #2256, opened 2019-11-18
Add content to How To summary page

ghostdogpr:
[This page](https://zio.dev/docs/howto/howto_index) is empty saying content is "coming soon" which is misleading because all the sub-items have content.

### DIFF
--- a/docs/howto/index.md
+++ b/docs/howto/index.md
@@ -5,8 +5,8 @@ title: "Summary"
 
 Here are a few guides for common patterns with ZIO:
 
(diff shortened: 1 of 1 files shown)


adamgfraser:
Should we add a little more content on this page, like a one sentence bullet point description of each of the topics on the left hand side, like we do for the "data types" page, for example?

ghostdogpr:
@adamgfraser Okay, I was hesitating to remove the page altogether but if we're keeping it some description doesn't hurt. I'll add it.

PR #2260, opened 2019-11-18
Fix subscribe button hover opacity when minimized

wongelz:
The docusaurus build/minification process seems to change all percentage opacity values to 1%

https://github.com/facebook/create-react-app/issues/7980


ghostdogpr:
@wongelz thanks for the fix! I applied it on the website manually.

PR #2274, opened 2019-11-19
Include All Fibers in Fiber Dump

### LINKED ISSUE #2266: Allow `Fiber.dump` to dump all fibers (even from other `unsafeRunSync`)
Today (1.0-RC17), `Fiber.dump` does not print fiber dumps for fibers started from an other `unsafeRun`. This is extremely problematic in app which need to integrate with existing non `ZIO` code (typically: integration with existing Java API, porting a big non `ZIO` app to `ZIO`). In these case, you don't (and can't easely) now where a deadlock comes from. 

But with today behavior, you need to know where the problematic fiber(s) are:
```
object TestTwoRunNow {
  def main (args: Array[String] ): Unit = {
    val runtime = new DefaultRuntime {}

    val deadlock = {
      def plop(s: String, sem: Semaphore) = sem.withPermit((s + "plop").succeed)
      def foo (s: String, sem: Semaphore) = sem.withPermit(plop(s + "foo" , sem))

      for {
        sem <- Semaphore.make(1)
        res <- foo("oups", sem)
        _   <- IO.effect(println(res))
      } yield ()
    }

    val printDumps = (for {
      dumps <- Fiber.dump.delay(2.seconds).provide(runtime.environment)
      _     <- IO.effect(println("************* fiber dump  ************"))
      s     <- ZIO.traverse(dumps) { _.prettyPrintM}
      _     <- IO.effect(println(s.mkString("\n")))
      _     <- IO.effect(println("************* /fiber dump ************"))
    } yield ())

    // case 1
    runtime.unsafeRun(deadlock.fork *> printDumps)

    // case 2
//    runtime.unsafeRun(deadlock.fork)
//    runtime.unsafeRun(printDumps)

    Thread.sleep(5*1000)
  }
}
```

Case 1 leads to (GOOD!):
```
************* fiber dump  ************

#0 (2s2686ms) 
   Status: Running
Fiber:Id(1574086710145,0) was supposed to continue to:
  a future continuation at zio.ZIO.$amp$amp$amp(ZIO.scala:1642)
  a future continuation at zio.Fiber$.dump(Fiber.scala:574)
  a future continuation at zio.ZIO.zipWith(ZIO.scala:1543)
  a future continuation at zio.Fiber$.dump(Fiber.scala:579)
  a future continuation at zio.ZIO.run(ZIO.scala:1166)
  a future continuation at zio.ZIO.bracket_(ZIO.scala:147)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:568)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:574)

Fiber:Id(1574086710145,0) execution trace:
  at zio.internal.FiberContext.trace(FiberContext.scala:704)
  at zio.Fiber.dump(Fiber.scala:115)
  at zio.internal.FiberContext.status(FiberContext.scala:702)
  at zio.Fiber.dump(Fiber.scala:114)
  at zio.internal.FiberContext.id(FiberContext.scala:689)
  at zio.Fiber.dump(Fiber.scala:113)
  at zio.internal.FiberContext.getRef(FiberContext.scala:682)
  at zio.ZIO.zipWith(ZIO.scala:1543)
  at zio.internal.FiberContext.children(FiberContext.scala:675)
  at zio.UIO$.effectSuspendTotal(UIO.scala:189)
  at <unknown>.<unknown>(ZIO.scala:0)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:157)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:150)
  at zio.ZIO.ensuring(ZIO.scala:352)
  at zio.ZIO.onInterrupt(ZIO.scala:689)
  at zio.ZIO.ensuring(ZIO.scala:350)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:150)
  at zio.ZIO$._IdentityFn(ZIO.scala:2630)
  at zio.clock.Clock$Live$$anon$1.sleep(Clock.scala:49)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.clock.Clock$Live$$anon$1.sleep(Clock.scala:48)
  at zio.clock.package$.sleep(clock.scala:50)
  at zio.ZIO.bracket_(ZIO.scala:147)
  at zio.internal.FiberContext.evaluateNow(FiberContext.scala:531)
  at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:574)

Fiber:Id(1574086710145,0) was spawned by: <empty trace>


#1 (2s2516ms) waiting on #1
   Status: Suspended(interruptible, 0 asyncs, zio.Promise.await(Promise.scala:49))
Fiber:Id(1574086710324,1) was supposed to continue to:
  a future continuation at zio.ZIO$._IdentityFn(ZIO.scala:2630)
  a future continuation at zio.ZIO.ensuring(ZIO.scala:350)
  a future continuation at zio.Semaphore.withPermits(Semaphore.scala:102)
  a future continuation at zio.ZIO.run(ZIO.scala:1166)
  a future continuation at zio.Semaphore.withPermits(Semaphore.scala:101)
  a future continuation at zio.ZIO.run(ZIO.scala:1166)
  a future continuation at zio.Semaphore.withPermits(Semaphore.scala:101)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:562)

Fiber:Id(1574086710324,1) execution trace:
  at zio.Promise.await(Promise.scala:49)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.Semaphore.withPermits(Semaphore.scala:102)
  at zio.Ref$.modify$extension(Ref.scala:53)
  at zio.Semaphore.prepare(Semaphore.scala:131)
  at zio.Promise$.makeAs(Promise.scala:248)
  at zio.Promise$.make(Promise.scala:241)
  at zio.ZIOFunctions.fiberId(ZIO.scala:2054)
  at zio.ZIOFunctions.descriptor(ZIO.scala:1874)
  at zio.Semaphore.withPermits(Semaphore.scala:102)
  at zio.Semaphore.withPermits(Semaphore.scala:102)
  at zio.Ref$.modify$extension(Ref.scala:53)
  at zio.Semaphore.prepare(Semaphore.scala:131)
  at zio.Promise$.makeAs(Promise.scala:248)
  at zio.Promise$.make(Promise.scala:241)
  at zio.ZIOFunctions.fiberId(ZIO.scala:2054)
  at zio.ZIOFunctions.descriptor(ZIO.scala:1874)
  at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:561)
  at zio.Semaphore$.make(Semaphore.scala:174)
  at zio.Ref$.make(Ref.scala:169)

Fiber:Id(1574086710324,1) was spawned by:

Fiber:Id(1574086710145,0) was supposed to continue to:
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:574)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:574)

Fiber:Id(1574086710145,0) ZIO Execution trace: <empty trace>

Fiber:Id(1574086710145,0) was spawned by: <empty trace>

************* /fiber dump ************
```

Case 2 leads to (BAD):
```
************* fiber dump  ************

#2 (2s2442ms) 
   Status: Running
Fiber:Id(1574086880664,2) was supposed to continue to:
  a future continuation at zio.ZIO.$amp$amp$amp(ZIO.scala:1642)
  a future continuation at zio.Fiber$.dump(Fiber.scala:574)
  a future continuation at zio.ZIO.zipWith(ZIO.scala:1543)
  a future continuation at zio.Fiber$.dump(Fiber.scala:579)
  a future continuation at zio.ZIO.run(ZIO.scala:1166)
  a future continuation at zio.ZIO.bracket_(ZIO.scala:147)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:568)
  a future continuation at com.normation.TestTwoRunNow$.main(ZioCommonsTest.scala:579)

Fiber:Id(1574086880664,2) execution trace:
  at zio.internal.FiberContext.trace(FiberContext.scala:704)
  at zio.Fiber.dump(Fiber.scala:115)
  at zio.internal.FiberContext.status(FiberContext.scala:702)
  at zio.Fiber.dump(Fiber.scala:114)
  at zio.internal.FiberContext.id(FiberContext.scala:689)
  at zio.Fiber.dump(Fiber.scala:113)
  at zio.internal.FiberContext.getRef(FiberContext.scala:682)
  at zio.ZIO.zipWith(ZIO.scala:1543)
  at zio.internal.FiberContext.children(FiberContext.scala:675)
  at zio.UIO$.effectSuspendTotal(UIO.scala:189)
  at <unknown>.<unknown>(ZIO.scala:0)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:157)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:150)
  at zio.ZIO.ensuring(ZIO.scala:352)
  at zio.ZIO.onInterrupt(ZIO.scala:689)
  at zio.ZIO.ensuring(ZIO.scala:350)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:150)
  at zio.ZIO$._IdentityFn(ZIO.scala:2630)
  at zio.clock.Clock$Live$$anon$1.sleep(Clock.scala:49)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.ZIOFunctions.effectAsyncInterrupt(ZIO.scala:1952)
  at zio.clock.Clock$Live$$anon$1.sleep(Clock.scala:48)
  at zio.clock.package$.sleep(clock.scala:50)
  at zio.ZIO.bracket_(ZIO.scala:147)
  at zio.internal.FiberContext.evaluateNow(FiberContext.scala:531)

Fiber:Id(1574086880664,2) was spawned by: <empty trace>

************* /fiber dump ************
```

adamgfraser:
Resolves #2266 

As I looked into it appears that we are already placing child fibers on the grandparent's node when the fiber is terminated. However, the issue is what we do when there is no grandparent fiber (i.e. when the fiber that is done is a direct child of the root). Currently, we set the parent of the child fibers equal to the grandparent, which results in setting them to `null` and the child fibers being "lost" in the fiber dump.

To correct this, we check if the grandparent fiber exists. If it does we simply connect the children to the grandparent as we were doing before. If not we track the children and add them to the set of root notes.


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -812,7 +812,8 @@ private[zio] final class FiberContext[E, A](
           while (iterator.hasNext()) {
             val child = iterator.next()
 
-            child.parentFiber = self.parentFiber
+            if (self.parentFiber ne null) child.parentFiber = self.parentFiber
+            else Fiber.track(child)
```
iravid: I think we need an `onDone` here that clears the child from the tracked set.

regiskuckaertz: Yes It is part of `track`

iravid:
Hmm. What happens if I interrupt the parent fiber after its children are transplanted to the grandparent? Will they still be interrupted?

core/shared/src/main/scala/zio/internal/FiberContext.scala
iravid: Ah right

adamgfraser:
@iravid I think the answer is no. But at that point the parent fiber would already be done. So conceptually it makes sense that you couldn't interrupt an effect that has already completed, and I think that is consistent with the prior semantics around interruption. But I can also see how it would be useful to be able to make sure that all non-daemon fibers forked by a fiber have been terminated. I think right now it is the responsibility of the parent fiber to arrange for any termination of its children before completing normally if it is not interrupted.

I think achieving the ability to interrupt children after completion would require maintaining the parent fiber at least until all of its children have terminated, which could result in fiber graphs that are much larger, so not sure what the trade-offs are there.

jdegoes:
@adamgfraser @iravid I think we kind of need this ability: if you interrupt a "dead" parent it should probably always interrupt all children, which should already work; but also if you interrupt an alive parent of a dead parent (a grandparent), then it should probably interrupt all children too. Otherwise we will have "non-deterministic auto-interruption" and it will be difficult for a user to do compositional cleanup.

I guess that implies the "dummy node" solution may be the best one. I just don't necessarily like the implications for GC...

iravid:
Yeah it's pretty important that we can interrupt grandchildren through the grandparent. Think of effects that fork off, fork some never-ending fibers and exit. Interruption is the only way to clean them up.

iravid:
Consequently this will also keep the fiber dump truer to reality.

Re the GC - if the fiber is done, it will discard its observers. Those are a pretty big GC pitfall as they can keep references to other things around. We can also consider nulling out the stack and any other stacks that aren't needed to minimize GC interference.

jdegoes:
@iravid Right, it almost seems like we need a 'makeZombie' type method to strip everything out and nullify everything possible except the set of children.

iravid:
Yes but FiberContext has to stay a class :-) no additional dispatches!

jdegoes:
We can also do "root relocation". If your parent is root, then when you are done, your children can be relocated to root; but still accessible through dead parent. That would prevent infinite ancestry assuming all ancestors finish eventually

iravid:
I admit I like keeping the original hierarchy in tact as it leads to fiber traces that are more useful and recognizable. But that is only assuming the GC impact is tractable.

adamgfraser:
Actually I think I misspoke about the effect of interrupting a terminated fiber. Consider the following snippet:

```scala
package zio

import zio.duration._
import zio.clock.Clock
import zio.console.Console

object Example extends App {

  def run(args: List[String]): ZIO[ZEnv, Nothing, Int] =
    for {
      fiber <- program.fork
      _     <- ZIO.sleep(5.seconds)
      _     <- fiber.interrupt
      _     <- ZIO.sleep(5.seconds)
    } yield 0

  val program: ZIO[Clock with Console, Nothing, Unit] =
    for {
      _ <- console.putStrLn("Hello, World!").delay(1.second).forever.fork
    } yield ()
}
```

This will print "Hello, World!" exactly five times. Even though `program` has clearly terminated by the time we interrupt it, it correctly interrupts the forked child. The issue is I was confusing the parent's reference to its children with the child's reference to its parent. So I think we are doing what John suggested, moving the child's parent to being the root node but leaving the parent's reference to its children unchanged.

adamgfraser:
Okay, so to summarize here is where I think we are:

Current behavior is that on completion a fiber retains a reference to its children but the child reference to its parent is updated to point to the grandparent. This allows interruption of a parent to propagate to children even after the parent completes but keeps fiber ancestry shorter. Perhaps we can optimize this to null out more fields of the terminated parent but this seems to basically work okay.

Where things break down in the version on `master` is at the root node. `track` in `Fiber` adds a fiber to the set of roots but removes the fiber from the root set on completion, resulting in any unfinished children being lost when we traverse the root nodes and their children in `dump`. So we either need to (1) never remove a fiber from the set of root nodes or (2) add the children to the set of root nodes when the parent finishes. This PR implements (2), which seems slightly more efficient.

So I think we are good to merge here if you both agree. Sorry again for my mistake earlier.

jdegoes:
@adamgfraser 

We have one more issue:

Fiber A creates fiber B.

Fiber B creates fiber C.

Fiber B ends life.

Fiber A is interrupted.

Should C be interrupted?

I think I am coming around to the answer, "yes", because whether or not C should be interrupted should be deterministic, not dependent on "how soon" B finishes.

Otherwise it will be difficult to make applications that do not leak fibers, if supervision isn't always entirely hierarchical.

This requirement (if indeed it should be a requirement) is the "strongest" in that it nearly forces Option 2 from the original issue.

Does all this make sense? If so, anyone have comments? /cc @iravid

adamgfraser:
@jdegoes I agree that this is the correct behavior. I think we can honor that contract now but let me check and then if so I will add some tests to prevent regression.

jdegoes:
The problem is that the children will point to their grandparent correctly, but the grandparent will have no knowledge of the grandchildren, so when the grandparent is interrupted, it will not interrupt its grandchildren, only its immediate children, who are in turn responsible for propagating the signal downward.

adamgfraser:
But isn't that fine because the parent will in turn interrupt its children and we showed before that the parent can be "interrupted" and interrupt its children even after terminating?

adamgfraser:
You're right, this is not happening today. I am working on it.

iravid:
Sorry I dropped out from the conversation. I still think option 2 from the original issue - keeping dead fibers around as "zombie" fibers - is the least surprising and most consistent. Both in terms of interruption semantics and in terms of how nice and familiar the fiber dump would look.

iravid:
Thanks again for picking this up @adamgfraser!

adamgfraser:
@jdegoes @iravid This is ready for another review.

regiskuckaertz:
> it almost seems like we need a 'makeZombie' type method

This may be stupid but why don't we create a whole separate class for that?

```scala
final class FiberZombie[E, A](
  id: Fiber.Id,
  exit: Exit[E, A],
  fiberRefLocals: FiberRefLocals,
  _children: JSet[FiberContext[Any, Any]]
) extends Fiber[E, A] {
  final def await = UIO(exit)
  final def children = UIO(_children.asScala.toSet)
  final def getRef[A](ref: FiberRef[A]) = UIO {
    val oldValue = Option(fiberRefLocals.get(ref))
    oldValue.asInstanceOf[Option[A]].getOrElse(ref.initial)
  }
  final def id = UIO(id)
  final def inheritRefs = UIO.unit
  final def interruptAs(fiberId: Fiber.Id) = UIO(exit)
  final def poll = UIO(Some(exit))
  final def status = UIO(Fiber.Status.Done)
  final def trace = UIO(Some(captureTrace(null)))
}
```

adamgfraser:
@regiskuckaertz I played with a version where I kept the zombie nodes and I found it led to quite unintuitive results because the set of children you have grows to include all these completed fibers. For example, the test below fails because on subsequent runs with `nonFlaky` the set of children includes all the done children from previous runs. Perhaps you could address this by having combinators treat zombie and non-zombie nodes differently but that is creating additional complexity. The other disadvantage is it leads to a fiber graph that grows without bound.

```scala
      testM("supervise fibers") {
        def makeChild(n: Int): URIO[Clock, Fiber[Nothing, Unit]] =
          (clock.sleep(20.millis * n.toDouble) *> IO.never).fork

        val io =
          for {
            counter <- Ref.make(0)
            _ <- (makeChild(1) *> makeChild(2)).handleChildrenWith { fs =>
                  fs.foldLeft(IO.unit)((acc, f) => acc *> f.interrupt.ignore *> counter.update(_ + 1).unit)
                }
            value <- counter.get
          } yield value

        assertM(io.provide(Clock.Live), equalTo(2))
      } @@ nonFlaky(100),
```

regiskuckaertz:
Yeah that would be the tradeoff, I guess.

> Perhaps you could address this by having combinators treat zombie and non-zombie nodes

If I'm not mistaken (😅), the only change would be in `Fiber#children` to filter out the zombies. But in light of your example, I now see the point with GC becoming an issue 👍

adamgfraser:
@jdegoes Could you take a look at this when you have the chance?

PR #2284, opened 2019-11-19
Small fix to sbt imports

hmemcpy:


docs/usecases/testing.md
```diff
@@ -9,8 +9,8 @@ title:  "Testing"
 libraryDependencies ++= Seq(
   "dev.zio" %% "zio-test"     % zioVersion % "test",
   "dev.zio" %% "zio-test-sbt" % zioVersion % "test"
-),
-testFrameworks += Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
+)
+testFrameworks ++= Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
```
softinio: Why is this change necessary? 

Docs suggest how it is now is correct: 

https://www.scala-sbt.org/1.x/docs/Testing.html#Using+Extensions

adamgfraser: Actually aren't we slightly off? It looks like the documentation says to use `+=` but then `new TestFramework(...)` instead of `Seq(new TestFramework(...))` so maybe we should just delete the `Seq`?

adamgfraser: @hmemcpy Could you make the change suggested above and we can merge this? We just had another user raise the same issue in #2375 so would be nice to get this resolved.

hmemcpy: @adamgfraser Done. `Seq` removed.

ghostdogpr:
Thanks!

PR #2286, opened 2019-11-19
Updated docs for RC17 to remove fold from the main App

tampler:


ghostdogpr:
The examples use `getStrLn` which can fail with a `IOException`, so we need to keep `fold` to transform the error to `Nothing`.

adamgfraser:
Yes, the code is compiling for me on `master`. @tampler Can you please double check?

ghostdogpr:
Closing but let us know if we missed something.

PR #2291, opened 2019-11-19
change welcome message

unclebob418:
Follow on from a discussion with @reibitto regarding the welcome message, I'm proposing we include a bit more on `testOnly` to make usage clearer. I and at least several others have asked on discord how to use this (super helpful) command.

This option provides both a version to explain usage and a version that can be copied and pasted directly.
```
[info] testOnly *.YourSpec -- -t "YourLabel" - Only runs tests with matching term e.g. coreTestsJVM/testOnly *.ZIOSpec -- -t "happy-path"
```


ghostdogpr:
Nice, thanks!

PR #2294, opened 2019-11-20
Add Schedule.doWhile example to microsite

reibitto:
Thought we should have at least 1 `doWhile` example since this question comes up a lot.

### DIFF
--- a/docs/datatypes/schedule.md
+++ b/docs/datatypes/schedule.md
@@ -102,3 +102,12 @@ Stops retrying after a specified amount of time has elapsed:
 ```scala mdoc:silent
 val expMaxElapsed = Schedule.exponential(10.milliseconds) && Schedule.elapsed.whileOutput(_ < 30.seconds)
 ```
+
+Retry only when a specific exception occurs:
+
+```scala mdoc:silent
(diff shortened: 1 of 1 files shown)


docs/datatypes/schedule.md
```diff
@@ -102,3 +102,12 @@ Stops retrying after a specified amount of time has elapsed:
 ```scala mdoc:silent
 val expMaxElapsed = Schedule.exponential(10.milliseconds) && Schedule.elapsed.whileOutput(_ < 30.seconds)
 ```
+
+Retry only when a specific exception occurs:
+
+```scala mdoc:silent
+val whileTimeout = Schedule.exponential(10.milliseconds) && Schedule.doWhile[Throwable] {
+  case _: TimeoutException => true
```
mschuwalow: import is missing

reibitto: Oh interesting. I didn't realize code in the snippets were compiled. Hmm, not exactly ideal but I guess I can use the `TimeoutException` in `scala.concurrent` to get around this. I didn't really want to specify one because usually web clients/servers define their own TimeoutException, so choosing one is pretty arbitrary.

Anyway, I went with that for now. Or I could make up a completely different example altogether.

PR #2301, opened 2019-11-21
Improve Error Messages for CanFail and NeedsEnv

### LINKED ISSUE #2285: Provide a more helpful hint in the CanFail message
I just upgraded to RC17, and my scratch program stopped compiling with the following error:
```
This operation only makes sense for effects that can fail.
[error]     program.fold(_ => 1, _ => 0)
[error]                 ^
```
This is due to #2049, but I feel the failure message can be more helpful. I didn't know what to do until someone suggested mapping the result to a success value. 
This suggestion could be reflected in the error message.

adamgfraser:
Resolves #2285.

I wasn't able to find a way to just generate a warning instead of an error so I expanded the error message and included a link to a new page on the website with more detailed information on how to replace useless error handling combinators.

### DIFF
--- a/docs/canfail.md
+++ b/docs/canfail.md
@@ -1,0 +1,77 @@
+---
+id: can_fail
+title:  "Compile Time Errors for Handling Combinators"
+---
+
+ZIO provides a variety of combinators to handle errors such as `orElse`, `catchAll`, `catchSome`, `option`, `either`, and `retry`. However, these combinators only make sense for effects that can fail (i.e. where the error type is not `Nothing`). To help you identify code that doesn't make sense, error handling combinators require implicit evidence `CanFail[E]`, which is automatically avaiable for all types except `Nothing`. The table below includes a list of combinators that only make sense for effects that can fail along with value preserving rewrites.
+
+**ZIO**
+
+Code | Rewrite 
+--- | ---
+`uio <> zio` | `uio`
+`uio.asError(e)` | `uio`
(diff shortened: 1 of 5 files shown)


docs/canfail.md
```diff
@@ -0,0 +1,77 @@
+---
+id: can_fail
+title:  "Compile Time Errors for Handling Combinators"
+---
+
+ZIO provides a variety of combinators to handle errors such as `orElse`, `catchAll`, `catchSome`, `option`, `either`, and `retry`. However, these combinators only make sense for effects that can fail (i.e. where the error type is not `Nothing`). To help you identify code that doesn't make sense, error handling combinators require implicit evidence `CanFail[E]`, which is automatically avaiable for all types except `Nothing`. The table below includes a list of combinators that only make sense for effects that can fail along with value preserving rewrites.
```
regiskuckaertz:
```suggestion
ZIO provides a variety of combinators to handle errors such as `orElse`, `catchAll`, `catchSome`, `option`, `either`, and `retry`. However, these combinators only make sense for effects that can fail (i.e. where the error type is not `Nothing`). To help you identify code that doesn't make sense, error handling combinators require implicit evidence `CanFail[E]`, which is automatically available for all types except `Nothing`. The table below includes a list of combinators that only make sense for effects that can fail along with value preserving rewrites.
```

adamgfraser:
Fixed now.

PR #2303, opened 2019-11-22
ZIO Test: Implement Gen#run

### LINKED ISSUE #2270: Single random element
Would it make sense to add this method?

```scala
  final def genValue[R, A](gen: Gen[R, A]): ZIO[R, Unit, A] =
    gen.sample.map(_.value).forever.take(1).runCollect.flatMap(l => ZIO.fromOption(l.headOption))
```

Sometimes I just want a single random element from a `Gen` rather than a full-blown `checkM` test.

adamgfraser:
Resolves #2270. Adds methods to run a generator to produce a single value, all values, or a specified number of values.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/GenSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/GenSpec.scala
@@ -19,7 +19,7 @@ object GenSpec extends ZIOBaseSpec {
         b <- gen.runAll
       } yield assert(a, equalTo(domain)) &&
         assert(b, equalTo(domain))
-    },
+    } @@ scala2Only,
     testM("runSome") {
       val gen = Gen.int(-10, 10)
       for {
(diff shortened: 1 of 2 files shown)


adamgfraser:
We definitely could. I thought about that and things that led me in this direction and trying to align more with `check`, `checkSome`, and `checkAll` (though in retrospect I think you were right about `zip` so very open to this):

* `run` / `runHead` - getting a single value seemed like a very common use case and the original motivator so wanted to keep that short. Also understand the collection analogy but not sure it is entirely intuitive that a generator has a "head".
* `runAll` / `runCollect` - this one only makes sense for deterministic generators so wanted to try to be explicit about that, though you could say the same thing about `runCollect` and infinite streams
* `runSome` / `???` - not sure what the right name would be to align with stream because of the way we're doing `forever` and `take`.

iravid:
For the last one we could call it `runCollectN`. I wanted to add such a combinator to stream too.

In theory it could also be useful to add a `Gen[A]#run(s: Sink[A])` method, but let's do that when someone actually needs it. Just means that we should reserve the name `run`.

adamgfraser:
I like that! Implemented in the latest version.

jdegoes:
Incidentally, I was thinking `checkSome` => `checkN` is possibly more consistent with the rest of ZIO...

adamgfraser:
I will change that!

PR #2305, opened 2019-11-22
Fix #2261 race interruption inside `daemon`

### LINKED ISSUE #2261: `race` interruption doesn't work inside `daemon` region
https://github.com/zio/interop-cats/pull/112#issuecomment-554846476

Since interruption of child fibers on interruption of this fiber in `race` now depends on supervision, in regions where it's off `race` stops working correctly

Twizty:
Implemented #2261.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1477,72 +1477,118 @@ object ZIOSpec extends ZIOBaseSpec {
           interrupted <- interruptionRef.get
         } yield assert(interrupted, equalTo(2)) && assert(res, equalTo(0))
 
-        io.provide(Clock.Live)
+        io.daemon.provide(Clock.Live)
       },
       testM("daemon mask") {
-        def sleep500(latch: Promise[Nothing, Unit]) =
-          latch.succeed(()) *> clock.sleep(500.milliseconds)
-        def forkSleep500 =
+        def forkAwait =
           for {
-            latch <- Promise.make[Nothing, Unit]
-            _     <- sleep500(latch).fork
-          } yield latch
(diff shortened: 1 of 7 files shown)


Twizty:
@neko-kai @ghostdogpr please review. Some tests are flaky, I'll fix it, but I guess the core functionality is implemented.

neko-kai:
@Twizty Looks fine. I guess you should use Promises instead of sleeps here

Twizty:
@neko-kai sure, I'll fix it.

Twizty:
@neko-kai done.

ghostdogpr:
@Twizty your first test doesn’t enable the daemon mode so it’s not actually testing that it works. Also as @neko-kai said, you can get rid of all the sleeps and `flaky` by using latches. I posted the test that shows the problem here: https://github.com/zio/zio/issues/2261#issuecomment-557383549 , maybe you re-use it?

Twizty:
@ghostdogpr @neko-kai done.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -1451,6 +1451,144 @@ object ZIOSpec extends ZIOBaseSpec {
           fibs <- ZIO.children
         } yield assert(fibs, isEmpty)
       },
+      testM("daemon fiber race interruption") {
+        def plus1(ref: Ref[Int], latch: Promise[Nothing, Unit]) =
+          latch.succeed(()) *> clock.sleep(200.milliseconds) *> ref.update(_ + 1)
```
ghostdogpr: The sleep could cause flakiness here. Actually you don't really need this `ref`, you could simply do `latch.succeed(()) *> ZIO.never` so that it wouldn't finish.

Twizty:
Thank you!

PR #2306, opened 2019-11-22
ZIO Test: Rename checkSome to checkN

adamgfraser:
Renames `checkSome` to `checkN` for consistency with other ZIO methods as suggested by @jdegoes in #2303. I think the renaming is quite natural for `checkSome`. The thing that is less ideal is `checkSomeM`, which goes to `checkNM`. I'm not sure if it is just me, if there is something else we should come up with for that, or if that is a reason to retain the existing names.


jdegoes:
@adamgfraser Good to merge when conflicts are resolved!

adamgfraser:
All set.

Issue #2122, opened 2019-11-03
Document the STM


mijicd:
STM documentation is missing from microsite. The issue can be split to the following subsections:

- [x] STM
- [x] TArray
- [x] TMap
- [x] TPromise
- [x] TQueue
- [x] TRef
- [x] TSemaphore
- [x] TSet
- [x] TReentrantLock

Note that this issue has been extracted from #417, and the parent will be updated once we're done with the changes.

Raise your hand for specific subsection you'd like to work on.

felipehaack:
Hi,
I'm new here, but I would like to work on a few items from the list `TArray, TMap, TPromise and TQueue`. 
I'm just not sure if these items are already on the website (I didn't find there actually). Or, would you point me the place where these items should be placed?

mijicd:
Hi @felipehaack, thanks for stepping up!

None of these are currently documented, and we really need them. They should be listed with other data types. Reach out if you need any help!

felipehaack:
Thanks, I will come up with the documentation for these items and on the way there I will definitely reach you out. (I'm taking this pull request as example #1013)

drmarjanovic:
Hi, I would like to take`TSet`.

Also, @felipehaack I can help you with `TArray`, `TMap` and `TQueue` if you agree.

mijicd:
@drmarjanovic `TSet` is yours.

felipehaack:
@drmarjanovic, please, thank you.

felipehaack:
Hi @mijicd,
I would like to take the `TSemaphore` this time. Is it ok?

mijicd:
@felipehaack sure thing, go for it!

calvinlfer:
hello @mijicd, @francisToth and myself would like to pick up whatever is remaining 😸

adamgfraser:
@calvinlfer Awesome!

mijicd:
@calvinlfer amazing, thanks!

PR #2313, opened 2019-11-23
TSet documentation

drmarjanovic:
Document `TSet` (part of #2122 ).

### DIFF
--- a/docs/datatypes/tset.md
+++ b/docs/datatypes/tset.md
@@ -233,7 +233,7 @@ import zio.stm._
 val tSetToList: UIO[List[Int]] = (for {
   tSet <- TSet.make(1, 2, 3, 4)
   list <- tSet.toList
-} yield tSetToList).commit
+} yield list).commit
 ```
 
 ## Size of a `TSet`
--- a/docs/datatypes/tset.md
+++ b/docs/datatypes/tset.md
@@ -1,0 +1,253 @@
+---
+id: datatypes_tset
+title: "TSet"
+---
+
+A `TSet[A]` is transactional data structure built on top of `TMap[A]`.
+
+## Create a TSet
+
+Creating an empty `TSet`:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val emptyTSet: STM[Nothing, TSet[Int]] = TSet.empty[Int]
+```
+
+Or creating a `TSet` with specified values:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val specifiedValuesTSet: STM[Nothing, TSet[Int]] = TSet.make(1, 2, 3)
+```
+
+Or creating a `TSet` from existing `Iterable` structure:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val iterableTSet: STM[Nothing, TSet[Int]] = TSet.fromIterable(List(1, 2, 3))
+```
+
+## Put an element to a `TSet`
+
+Putting new element to a `TSet` extends it if specified element is not already present:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val putElem: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2)
+  _    <- tSet.put(3)
+} yield tSet).commit
+```
+
+## Remove an element from a `TSet`
+
+The simplest way to remove an element from `TSet` is using `delete` method:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val deleteElem: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3)
+  _    <- tSet.delete(1)
+} yield tSet).commit
+```
+
+Also, it is possible to remove every element that matches predicate function:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val removedEvenElems: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  _    <- tSet.removeIf(_ % 2 == 0)
+} yield tSet).commit
+```
+
+Or you can keep all elements that match predicate function:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val retainedEvenElems: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  _    <- tSet.retainIf(_ % 2 == 0)
+} yield tSet).commit
+```
+
+## Union of a `TSet`
+
+Union of the sets A and B represents the set of distinct element belongs to set A or set B, or both.
+
+```scala mdoc:silent
+import zio._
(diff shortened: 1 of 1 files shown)


docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
+---
+id: datatypes_tset
+title: "TSet"
+---
+
+A `TSet[A]` is transactional data structure built on top of `TMap[A]`.
```
mijicd: I would rather describe it as a mutable set that can participate in transactions. See `tref.md` as an inspiration.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+Or creating a `TSet` with specified values:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val specifiedValuesTSet: STM[Nothing, TSet[Int]] = TSet.make(1, 2, 3)
+```
+
+Or creating a `TSet` from existing `Iterable` structure:
```
mijicd: Rephrase as `Alternatively, you can create a `TSet` by providing a collection of values:`. Here's the right place to put the remark about handling duplicates (last one "wins").

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+val emptyTSet: STM[Nothing, TSet[Int]] = TSet.empty[Int]
+```
+
+Or creating a `TSet` with specified values:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val specifiedValuesTSet: STM[Nothing, TSet[Int]] = TSet.make(1, 2, 3)
```
mijicd: I'd use `tset` as names for example variables, unless necessary otherwise.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val iterableTSet: STM[Nothing, TSet[Int]] = TSet.fromIterable(List(1, 2, 3))
+```
+
+## Put an element to a `TSet`
+
+Putting new element to a `TSet` extends it if specified element is not already present:
```
mijicd: Split in two, e.g.:

```
New element can be added to the set in the following way:

<example>

In case the set already contains the element, no modification will happen.
```

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+import zio._
+import zio.stm._
+
+val deleteElem: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3)
+  _    <- tSet.delete(1)
+} yield tSet).commit
+```
+
+Also, it is possible to remove every element that matches predicate function:
```
mijicd: `...that satisfies provided predicate`.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+
+val retainedEvenElems: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  _    <- tSet.retainIf(_ % 2 == 0)
+} yield tSet).commit
+```
+
+## Union of a `TSet`
+
+Union of the sets A and B represents the set of distinct element belongs to set A or set B, or both.
```
mijicd: `..the set of elements belonging to ...`. Emphasize which of the sets is modified.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+val unionTSet: UIO[TSet[Int]] = (for {
+  tSetA <- TSet.make(1, 2, 3, 4)
+  tSetB <- TSet.make(3, 4, 5, 6)
+  _     <- tSetA.union(tSetB)
+} yield tSetA).commit
+```
+
+## Intersection of a `TSet`
+
+Intersection of the sets A and B represents the set of elements belongs to both A and B (set of the common elements in A and B).
```
mijicd: Drop the parentheses, `belongs => belonging`, and emphasize which of the sets is modified by the operation. Also, `represents the set` should be changed to `is the set`.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+val intersectionTSet: UIO[TSet[Int]] = (for {
+  tSetA <- TSet.make(1, 2, 3, 4)
+  tSetB <- TSet.make(3, 4, 5, 6)
+  _     <- tSetA.intersect(tSetB)
+} yield tSetA).commit
+```
+
+## Difference of a `TSet`
+
+Difference between sets A and B represents the set containing elements of set A but not in B (all elements of A except the element of B).
```
mijicd: The same remarks apply here.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Union of a `TSet`
+
+Union of the sets A and B represents the set of distinct element belongs to set A or set B, or both.
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val unionTSet: UIO[TSet[Int]] = (for {
```
mijicd: It'd be useful to show the result. That can be achieved either through comments or by writing zio-test assertions.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+import zio.stm._
+
+val shrinkTSet: UIO[TSet[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  _    <- tSet.transform(_ => 1)
+} yield tSet).commit
+```
+Resulting set in example above has only one element.
+
+Using `transformM` is similar to `transform` except using pure function `A => A`, effectful function `A => STM[E, A]` is used:
```
mijicd: `The elements can be mapped effectfully via transformM`.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+import zio._
+import zio.stm._
+
+val foldTSet: UIO[Int] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  sum  <- tSet.fold(0)(_ + _)
+} yield sum).commit
+```
+
+Using `foldM` is similar to `fold` except using pure function `(B, A) => B`, effectful function `(B, A) => STM[E, B]` is used:
```
mijicd: The same remark as with `transformM`.

docs/datatypes/tset.md
```diff
@@ -0,0 +1,251 @@
...
+
+val tSetToList: UIO[List[Int]] = (for {
+  tSet <- TSet.make(1, 2, 3, 4)
+  list <- tSet.toList
+} yield tSetToList).commit
+```
+
+## Size of a `TSet`
+
+Calculating the size of a `TSet`:
```
mijicd: `Set's size can be obtained as follows:`. Similar rephrasing can be applied to the examples above.

mijicd:
Thank you for working on this!

Issue #811, opened 2019-04-30
Supervision is unintuitive due to internal forks


ghostdogpr:
If you execute the following code:
```scala
UIO(println("ok")).fork.timeout(30 seconds).superviseWith { list =>
  IO.unit
}
```
You expect `list` to contain the fiber just created with `fork`. 
However, because `timeout` uses `race` internally and `race` forks some fibers, you receive the list of those fibers instead of the expected one. In other words, internal implementation is affecting the behavior or supervision, which leads to unexpected results for users.

jdegoes:
In general, I think that monadic regions should be inheritable, and that the lack of inheritability causes confusion for users, requires global reasoning about whether or not effects "fork", and may violate an interpretation of fork/join identity law.

iravid:
@jdegoes which is to say, a supervised fiber always forks supervised fibers?

jdegoes:
More generally, a fiber with a regional effect will always propagate it on forking.

That means inheriting:

 - Current interruptibility status
 - Current executor (I think this is _already_ true)
 - Current supervision status
 - Soon, next-gen fiber locals

Any other stacks I missed?

ghostdogpr:
If I understand your proposal, when we fork, we pass to the new `FiberContext` the current state of supervision (`supervising` and `supervised.peek`) and interruptibility (`interruptStatus.peek`), the same way we currently do with the executor (`platform`) and the environment (`environment.peek`).

I tried to do that, it works well for supervision (my example above returns all the fibers and not just the `race` ones) but it breaks a number of tests when I add the interruptibility status 🤔

ghostdogpr:
Example of problem with interruptibility inheritance:
```scala
val io =
      for {
        promise <- Promise.make[Nothing, Unit]
        fiber   <- IO.bracket(promise.succeed(()) <* IO.never)(_ => IO.unit)(_ => IO.unit).fork
        res     <- promise.await *> fiber.interrupt.timeoutTo(42)(_ => 0)(1.second)
      } yield res
```

`fiber` in uninterruptible, and `timeoutTo` uses `race` and `interrupt` internally, which is not working anymore because it inherited the uninterruptibility.

kaishh:
Inheriting interruptibility is not immediately intuitive to me, personally I always expect a fiber to start out interruptible...

jdegoes:
These issues do require thought.

[...]

PR #2317, opened 2019-11-24
Prevent finalizer from interrupting itself

### LINKED ISSUE #2226: Race in a finalizer is not executed in case of interruption
Basically, what the title says.

Running the spec on a fresh sbt instance once runs fine. Running it a second time times out. Running it a third time, it goes into a `stacktracerJVM compileIncremental` loop and never executes again.

This is a test that usually takes less than 2 seconds to run. It involves interruption. I'm pretty sure it has something to do with #2169.

-- Edit by @iravid --

The symptom here is that the tests in that suite use the global pool to run blocking code. That global pool is eventually exhausted after a few runs, because we're not getting an interruption when offering to a shutdown queue.

-- Edit by @ghostdogpr 

Changed the title to describe the root problem which is that `race` inside a finalizer is not executed properly in case of interruption.

ghostdogpr:
Fixes #2226

### The problem
When an effect is interrupted, we may run some finalizers. If one of these finalizers has a fork with an interruptible region (for example because of a `race`), the fiber running the fork interrupts itself and doesn't run the finalizer logic properly.

This is visible when we shutdown a queue inside a finalizer (like we do a lot in `ZStream`), the shutdown uses `foreachPar` (which uses `race`) to interrupt pending offers and takers. Because of that problem, the pending offers and takers are not interrupted properly and stay there forever. There might be other places where this happens.

### The cause
There is some logic in `FiberContext` called `propagateAncestorInterruption`. What this function does is look at the parent fiber to see if it's interrupted, and inherit that state. That means that if the parent is interrupted and the fiber itself is interruptible, it proactively interrupts itself. The parent fiber already interrupts its children but that is a way to do it a little faster in some cases (not waiting for the parent). 

The problem is that fibers forked by finalizers apply the same logic and interrupt themselves if they have an interruptible region. `propagateAncestorInterruption` doesn't distinguish "normal" children from those created in finalizers.

### The fix
`race` needs to have interruptible regions because it needs to be able to cancel the losing side. We could potentially modify `race` to run fibers in daemon mode and interrupt them manually. But the problem described might occur in other places and in user code, so I think this is not enough. I propose to get rid of `propagateAncestorInterruption` altogether because it is not able to discriminate well cases where we want / don't want to inherit the parent state and leads to unexpected/unintuitive behavior.

What do you think? Others suggestions to fix the problem?

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1235,6 +1235,16 @@ object ZIOSpec extends ZIOBaseSpec {
           } yield test
 
         assertM(io.provide(Clock.Live), isTrue)
+      },
+      testM("race inside a finalizer is executed") {
+        for {
+          ref   <- Ref.make(0)
+          latch <- Promise.make[Nothing, Unit]
+          fiber <- latch.succeed(()).bracket(_ => ref.set(1) race ZIO.never)(_ => ZIO.never).fork
+          _     <- latch.await
+          _     <- fiber.interrupt
+          ref   <- ref.get
+        } yield assert(ref, equalTo(1))
       }
     ),
     suite("RTS synchronous stack safety")(
@@ -1697,7 +1707,7 @@ object ZIOSpec extends ZIOBaseSpec {
           res <- p1.await
         } yield assert(res, isTrue)
       },
-      testM("interruption of raced") {
+      testM("interruption of race") {
         for {
           ref   <- Ref.make(0)
           cont1 <- Promise.make[Nothing, Unit]
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -285,9 +285,6 @@ private[zio] final class FiberContext[E, A](
           kTrace
         } else null
 
-      // Propagate ancestor interruption every once in a while:
-      propagateAncestorInterruption()
-
       Fiber._currentFiber.set(this)
 
       while (curZio ne null) {
@@ -746,16 +743,6 @@ private[zio] final class FiberContext[E, A](
     val _ = daemonStatus.popOrElse(false)
   }
 
-  private[this] final def propagateAncestorInterruption(): Unit = {
-    var fiber = self: FiberContext[_, _]
-
-    while (fiber ne null) {
-      addInterruptor(fiber.state.get.interrupted)
-
-      fiber = fiber.parentFiber
-    }
-  }
-
   @inline
   private final def isInterrupted(): Boolean = !state.get.interrupted.isEmpty
 
@@ -765,23 +752,6 @@ private[zio] final class FiberContext[E, A](
   @inline
   private[this] final def shouldInterrupt(): Boolean =
     isInterrupted() && isInterruptible()
-
-  @tailrec
-  private[this] final def addInterruptor(cause: Cause[Nothing]): Unit =
-    if (!cause.isEmpty) {
-      val oldState = state.get
-
-      oldState match {
-        case Executing(status, observers, interrupted) =>
-          val newInterrupted = if (!interrupted.contains(cause)) interrupted ++ cause else interrupted
-
-          val newState = Executing(status, observers, newInterrupted)
-
-          if (!state.compareAndSet(oldState, newState)) addInterruptor(cause)
-
-        case _ =>
-      }
-    }
 
   @inline
   private[this] final def nextInstr(value: Any): IO[E, Any] =


adamgfraser:
This seems logical to me. I don't know the nuances of the implementation of the new supervision model as well as @jdegoes so would defer to him if this causes problems we aren't thinking about, but conceptually this seems to allow easier and more local reasoning.

With the removal of this code, we essentially have a model where each parent fiber is responsible for supervising and potentially interrupting its children, and interruption propagates through fiber trees from parent to child. This allows a level of "local" reasoning where to understand the interruptibility of a fiber you only need to understand the behavior of its parent whereas with `propagateAncestorInterruption` the behavior can potentially depend on logic much further up in the fiber tree that is difficult to control.

I know we criticize the actor model on various grounds, but this also seems very similar to the actor supervision model which seems like it has been relatively successful and is something that people already understand.

jdegoes:
This is not a reliable fix to the problem. I think we actually need to fix this in `race` by using `uninterruptibleMask`, to allow the parent to decide whether or not to perform the race interruptible or not. That may require we be even more cautious inside ZIO about too much uninterrutibility.

I'll respond at length when I have time (not for the next couple of days, I'm underwater...).

neko-kai:
@jdegoes 
> I think we actually need to fix this in race by using uninterruptibleMask, to allow the parent to decide whether or not to perform the race interruptible or not.

How do you imagine performing the race uninterruptibly? The loser in `race` must be interrupted – most of the time it's not a finite computation either, but something like `Promise.await`. I don't think there's a viable way to perform an 'uninterruptible race'.

jdegoes:
@neko-kai @ghostdogpr @adamgfraser 

The problem is not with `propagateAncestorInterruption` per se. The purpose of `propagateAncestorInterruption` is to propagate parent interruption status into child interruption status. Note that such propagation will occur naturally anyway: when the parent fiber is successfully interrupted, it will then attempt to interrupt all children, in sequential order.

This means that even if `propagateAncestorInterruption` is deleted, eventually parent signals will propagate to children. The only question is when. And the answer to that question is non-deterministic since it depends on the lifespan of the parent. 

Therefore, deleting `propagateAncestorInterruption` will not solve the problem deterministically, although you may observe it to solve the problem in your cases. There will be cases where users still have the same issues they do now.

Nor do I want to change the interruption model. Any interruption model that has anything as powerful as `ZIO.interruptible` is going to allow punching holes in uninterruptible regions. This is a good thing, because not only does the symmetry of `ZIO.interruptible` / `ZIO.uninterruptible` allow for some very powerful and useful algebraic laws (there is a ticket open to finish them), but it's a sign of good design; this power has enabled us to implement many combinators successfully; moreover, I'd point to the fact that ZIO can already emulate CE3 semantics easily as a sign it's the more expressive basis for programming with interruption.

Beyond this, even if we disallowed punching interruptible holes in uninterruptible regions, it doesn't solve any real problems because there are other ways of punching holes in uninterruptible regions: e.g. `ZIO.die`. Any code unsafe in the presence of interruptible holes is going to be unsafe more generally in the presence of defects. Essentially if you are sequencing "untrusted" effects into an uninterruptible region, then in order to reason locally about correctness, you need to `run` those effects or use appropriate combinators like `onError` / `onInterrupt`, etc. to ensure correct logic.

Now, all that said, this is a problem. The behavior of `race` is confusing. I think it definitely is broken and that we need to fix it.

How to fix it? I have a proposal below.

### Alternate Proposal

Here's what I want to do: change `race` to _not_ make both sides interruptible; rather, it will use the parent region's interruptibility status to decide what to do:

1. If the parent region is NOT interruptible, then we will not interrupt either left or right sides; rather, we will allow the loser of the race to continue operating in the background without affecting the execution of either side.
2. If the parent region IS interruptible, this is the user specifying that it is OK to interrupt, so we will make both left and right sides interruptible and interrupt the loser of the race.

In this fashion, we defer the decision of interruptibility to the call-site, which gives much more flexibility and leads to less surprising behavior:

 * If I want to make a race uninterruptible, then I can do so, and I can have the guarantee that I get the "fastest success", but it will not interfere with the loser's lifecycle.
 * If I want to make a race interruptible, or if I don't care (the default is interruptibility), then loser will be interrupted as per current semantics.

Separately we need to see if `race` is the only problematic combinator. I think any combinator which "forces" interruption without considering the parent region's status is likely to have confusing edge cases like this. So a code audit may identify other areas of potential improvement.

adamgfraser:
@jdegoes I think your proposal definitely works.

Just to push a bit more on `propagateAncestorInterruption`, it seems like there is a deterministic difference. Consider a situation where Fiber A forks Fiber B, which is uninterruptible, and then Fiber B forks Fiber C, which is interruptible. With `propagateAncestorInterruption` Fiber C will be interrupted. But without that Fiber C will not be interrupted, even eventually, because Fiber B will not be interrupted and so will not propagate its interruption to its children. This seems like a powerful and compositional way to create nested interruptible and uninterruptible scopes, but maybe I am missing something here.

jdegoes:
@adamgfraser 

> Consider a situation where Fiber A forks Fiber B, which is uninterruptible, and then Fiber B forks Fiber C, which is interruptible. With propagateAncestorInterruption Fiber C will be interrupted. But without that Fiber C will not be interrupted, even eventually, because Fiber B will not be interrupted and so will not propagate its interruption to its children.

This is true if the "whole" of Fiber B is uninterruptible. Consider a more common scenario where part of Fiber B is uninterruptible: then as soon as it exits an uninterruptible region, it begins shutdown, which then propagates the interruption to Fiber C.

You're right though that it's not exactly the same. 

Personally, I am fine to delete `propagateAncestorInterruption`. The reason it is there is because shutting down a large graph by sequentially terminating each child is an extremely slow process, and `propagateAncestorInterruption` can accelerate this by allowing faster detection of a desired termination state. I also kind of like that it forces us to deal with "questionable" interruption behavior: having interruptible regions that are only "accidentally" not interrupted, but which in theory could be interrupted if someone had access to the child fiber (which they can get with `ZIO.children`).

adamgfraser:
That's fair. In some cases that ability to delay interruption could be very powerful and could be implemented in a deterministic way. For example, in `race`, Fiber B forks the two raced fibers and makes itself uninterruptible until the race has completed, so although it will propagate interruption at some point it guarantees that it will not interrupt the racing fibers due to propagated interruption.

That being said you make a good point about `ZIO.children` breaking this encapsulation. The direct ancestor interruption model seemed interesting in that it allowed more "local" reasoning about interruption but if we can get a good model of interruption without that maybe we don't need to pay the performance price.

ghostdogpr:
@jdegoes Can you clarify one point for me: when a fiber is interrupted, does it kill its children AFTER running the finalizers, or at the same time (nondeterministically)?

If it's AFTER (like I thought it was), then `propagateAncestorInterruption` acts "too soon"  by killing the children before it's time. Removing it would completely prevent the children from being killed while finalizers are still being processed.

But if those steps happens in parallel, then I understand why removing it won't solve the problem.

ghostdogpr:
Another remark about the proposal (`we will not interrupt either left or right sides`). `raceWith` does not actually interrupt anything, it just returns the losing fiber. Changing the interruption logic would mean changing `raceEither`, `raceAttempt`, `timeoutFork` and a few more occurrences in `ZStream` and ZIO Test because they all call `interrupt` on something that will be uninterruptible. There's also user code that might do the same thing. Doesn't that indicate a problem? It seems pretty easy to do the "wrong" thing and call `interrupt` where you shouldn't.

jdegoes:
> In some cases that ability to delay interruption could be very powerful and could be implemented in a deterministic way. 

I agree it's useful. The question is as follows:

 - If I make a part of my code interruptible, but it is not safe to do so, and if no one "happens" to obtain access to my fiber to interrupt me, is this a bug in my code or not?

Part of me says that if code is interruptible then it should have no problem being interrupted. 

But another part of me says: it's a nice contract to have that anyone who interrupts me should have direct knowledge of my existence. That can't be true with the `ZIO.children` but it could be made more difficult to exploit by deleting `propagateAncestorInterruption`.

@ghostdogpr 

> Can you clarify one point for me: when a fiber is interrupted, does it kill its children AFTER running the finalizers, or at the same time (nondeterministically)?

It kills the children after running the finalizers. 

> If it's AFTER (like I thought it was), then propagateAncestorInterruption acts "too soon" by killing the children before it's time. Removing it would completely prevent the children from being killed while finalizers are still being processed.

Not "prevent", see ZIO.children conversation above. 

Are we sure that children should rely on outliving their parents?

> Another remark about the proposal (we will not interrupt either left or right sides). raceWith does not actually interrupt anything, it just returns the losing fiber.

Right, I was referring strictly to `race`, not to `raceWith`. `raceWith` is safe in the sense that it does not use `ZIO.interruptible`.

>  There's also user code that might do the same thing. Doesn't that indicate a problem? It seems pretty easy to do the "wrong" thing and call interrupt where you shouldn't.

I have an alternate, alternate proposal: 

We change `ZIO.interruptible` to return an always interruptible effect, but one which merely "disconnects" the interruption of the fiber from waiting for the interruption of the underlying effect.

Then we call the old thing `ZIO.interruptibleForce` or something. 

Finally, we change combinators to have the behavior I specified in the alternate proposal: if the parent is uninterruptible, we use `ZIO.interruptible`, otherwise, we use `ZIO.interruptibleForce`. 

Requires more thought for sure....

neko-kai:
There's a solution that allows leaving the semantic of race unchanged:
Make forks in race `daemon`ic and rely on a finalizer for interruption as before auto-supervision.
That way `race` forks are unaffected by _either_ `propagateAncestorInterruption` or `ZIO.children` – they're now good behaving private forks.
##
@jdegoes 
> It kills the children after running the finalizers.
> Are we sure that children should rely on outliving their parents?

As aside, right now children outlive parents as a rule since they're only destroyed after finalizers.
Right now the outermost finalizers will execute first, then inner, the graph folds as in `foldLeft` – which is directly the opposite order to when you control child fibers with finalizers, or old supervision's `.interruptChildren`, which is `foldRight` order!

Maybe instead children should be finalized _first_, before all finalizers, e.g.:
``setInterrupt *> interruptChildren *> await <* interruptChildren`
instead of:
`setInterrupt *> await <* interruptChildren`
?

> If the parent region is NOT interruptible, then we will not interrupt either left or right sides; rather, we will allow the loser of the race to continue operating in the background without affecting the execution of either side.

This will cause the following code to leak the fiber and to never execute `finalizerB`if `errPromise` is never signalled:

```scala
a.race(errPromise.await.ensuring(finalizerB)).uninterruptible
```

IMHO this is a _large_ departure from a semantic of race, the two semantics of race under different interruptibility would be effectively unrelated. As @ghostdogpr notes, this won't be a local change either, but impacts every combinator on top of `raceWith`

> I also kind of like that it forces us to deal with "questionable" interruption behavior: having interruptible regions that are only "accidentally" not interrupted, but which in theory could be interrupted if someone had access to the child fiber (which they can get with ZIO.children).

I think effects must be able to privately use interruption on their internal forks, even when the effect is in an uninterruptible section – IMHO that ability is not "questionable", but desirable!

> it's a sign of good design; this power has enabled us to implement many combinators successfully; moreover, I'd point to the fact that ZIO can already emulate CE3 semantics easily as a sign it's the more expressive basis for programming with interruption.

Both `ZIO.interruptible` and `interruptibleMask` are so far **unused** in ZIO core. Their only uses on the current `master` branch are with `.fork` – only to negate interruption inheritance. What _IS_ being used is `uninterruptibleMask`, which is strictly less powerful.

> Any code unsafe in the presence of interruptible holes is going to be unsafe more generally in the presence of defects.

AFAIK the `race` in question uses only internal trusted code. If I can reason that my finalizer will complete as long as I'm only using trusted code and ZIO built-ins, that's less things to think about.

iravid:
Just wanted to thank you @ghostdogpr @adamgfraser @neko-kai and @jdegoes for spending many brain cycles on this problem :-)

ghostdogpr:
> There's a solution that allows leaving the semantic of race unchanged:
> Make forks in race `daemon`ic and rely on a finalizer for interruption as before auto-supervision.
> That way `race` forks are unaffected by _either_ `propagateAncestorInterruption` or `ZIO.children` – they're now good behaving private forks.

@neko-kai but then we get back to the problem described in #811 where if you just stick a `.timeout` in your chain, you lose track of your children.

> Part of me says that if code is interruptible then it should have no problem being interrupted.

@jdegoes I agree. But in that case (finalizer), we never asked to interrupt that code. The problem isn't that it's interrupted per se but that there was no reason to interrupt it.

> That can't be true with the ZIO.children but it could be made more difficult to exploit by deleting propagateAncestorInterruption.

@jdegoes The difference is that with `propagateAncestorInterruption`, this happens without you doing anything and you have no control over it, but using `ZIO.children` it's only going to be interrupted if you said so.

ghostdogpr:
Another idea: what if instead of getting the value of `interrupted` from the parent state, `propagateAncestorInterruption` would get a similar value from the parent but that is set only after the finalizers are done (just before `interruptChildren`). We would still get the speed benefit of proactively killing children but without affecting the finalizers.

vasilmkd:
Any way to revive the discussion?

jdegoes:
@vasilmkd Yes, but it is too busy right now. I will get to this as soon as I can post conference. It's a very important topic and needs to be handled correctly in 1.0, which will be my 100% time focus beginning from December 15th.

vasilmkd:
Not rushing anybody, just interested.

ghostdogpr:
Closing this one since John is working on an alternative approach.

PR #2322, opened 2019-11-24
Add ZIO.merge

reibitto:
I was looking for `merge` on ZIO but couldn't find it. I've been doing `zio.fold(identity, identity)` up until now, but I figured this is a common enough method that might be worth adding since you find across many libraries (stdlib, Scalaz, Cats) with the same name: `Either.merge`, `EitherT.merge`, `\/.merge`, etc.

What do you think? Do you think it's worth it? If so, we might also want to consider if there's value in adding this to any other applicable data types. I only had `ZIO` in mind when I started this.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -637,8 +637,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Returns a new effect where the error channel has been merged into the
    * success channel to their common combined type.
    */
-  final def merge[A1 >: A](implicit ev: E <:< A1): URIO[R, A1] =
-    self.foldM(e => ZIO.succeed(ev(e)), ZIO.succeed)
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -633,6 +633,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       _ <- (l.await *> ((self provide r) to p)).fork
     } yield l.succeed(()) *> p.await
 
+  /**
+   * Returns a new effect where the error channel has been merged into the
+   * success channel to their common combined type.
+   */
+  final def merge[A1 >: A](implicit ev: E <:< A1): URIO[R, A1] =
```
adamgfraser: You could use `CanFail` here. It doesn't make sense to `merge` a `ZIO[R, Nothing, A]`.

reibitto: Yeah, good point. Thanks.

PR #2324, opened 2019-11-24
add ZManaged.access

mschuwalow:


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -922,6 +926,24 @@ object ZManaged {
   final def absolve[R, E, A](v: ZManaged[R, E, Either[E, A]]): ZManaged[R, E, A] =
     v.flatMap(fromEither(_))
 
+  /**
+   * Create a managed that acesses the environment.
```
ghostdogpr:
```suggestion
   * Create a managed that accesses the environment.
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -922,6 +926,24 @@ object ZManaged {
...
+   */
+  final def access[R]: AccessPartiallyApplied[R] =
+    new AccessPartiallyApplied
+
+  /**
+   * Create a managed that acesses the environment.
```
ghostdogpr:
```suggestion
   * Create a managed that accesses the environment.
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -867,41 +893,19 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-   */
-  final def apply[R, E, A](reservation: ZIO[R, E, Reservation[R, E, A]]): ZManaged[R, E, A] =
-    new ZManaged(reservation)
+  final class AccessPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
+    def apply[A](f: R => A): ZManaged[R, Nothing, A] =
+      ZManaged.environment.map(f)
+  }
+
+  final class AccessMPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
+    def apply[E, A](f: R => ZManaged[R, E, A]): ZManaged[R, E, A] =
```
adamgfraser: Should `f` here be `R => ZManaged[R, E, A]` or `R => ZIO[R, E, A]`? I ran into the same thing with regard to `accessM` on `ZStream`. It seems like it could just be the first one, and that is more consistent with other `M` methods, but I'm not sure.

mschuwalow: My feeling is that we are more flexible with the current signature.
I think the proposed signature would be equivalent to ZIO.accessM(???).toManaged_ which I feel is something you need less often.

adamgfraser: Added a comment related to this in #2337.

mschuwalow: I'll rename this to accessManaged and add accessM as per John's comment

adamgfraser: Sounds good. I am doing the same thing on `ZStream`.

mschuwalow:
> Weren't we going to add `accessManaged` and change the signature of `accessM`?

Don't ask 🤦‍♂ 
(Pushing commit had failed because I had rebased but I completely ignored it)

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -867,41 +893,24 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-  final def apply[R, E, A](reservation: ZIO[R, E, Reservation[R, E, A]]): ZManaged[R, E, A] =
-    new ZManaged(reservation)
+  final class AccessPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
+    def apply[A](f: R => A): ZManaged[R, Nothing, A] =
+      ZManaged.environment.map(f)
+  }
+
+  final class AccessMPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
+    def apply[E, A](f: R => ZIO[R, E, A]): ZManaged[R, E, A] =
+      ZManaged.environment[R].flatMap(f(_).toManaged_)
```
adamgfraser: `ZManaged.environment.mapM(f)`?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -922,6 +931,30 @@ object ZManaged {
...
+    new AccessPartiallyApplied
+
+  /**
+   * Create a managed that acesses the environment.
+   */
+  final def accessM[R]: AccessMPartiallyApplied[R] =
+    new AccessMPartiallyApplied
+
+  /**
+   * Create a managed that acesses the environment.
```
adamgfraser: Typo in "accesses" here and above.

PR #2326, opened 2019-11-24
implemented await(timeout) in fiber

wi101:
current state:

```scala
for {
  f <- putStrLn("hello").forever.fork
  _ <- f.await.timeout(10.millis)
} yield ()
```
the fiber will run forever.

Now it would be interrupted using `await(duration)` 

```scala
for {
  f <- putStrLn("hello").forever.fork
  _ <- f.await(10.millis)
} yield ()
```
For futures the current implementation of ` interrupt` will join the future and return its result..
Currently `await(duration)` in `Fiber.fromFuture(f)` will interrupt the fiber but the future will be still flying how could we have control over it? Could we use `Promise` in ` fromFuture` ?


wi101:
I will close this PR, and reopen it when I will back working on it

Issue #2269, opened 2019-11-18
Benchmark Performance of Chunk


adamgfraser:
We should benchmark the performance of `Chunk` versus other collection types. This will help us both demonstrate the value it adds versus existing collection types and give us a point of comparison as we explore further optimization opportunities for primitive types. Benchmarks should cover a variety of use cases (e.g. appending roughly equally sized chunks to create a tree like structure, repeatedly prepending or appending, traversing, etc...) with both primitive and non-primitive types. This blog post on [benchmarking Scala collections](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html) may be a good source of inspiration.

fanf:
there is that pretty exhausitve benchmark on scala vector tested with a potential replacement: https://github.com/scala/scala/pull/8534
The comments are pretty amazing about all the pitfall of benchmarking scala collection performance, from JVM to CPU optimization you want to control.

Vilkina:
I need some help with benchmark ...

Vilkina:
https://github.com/zio/zio/pull/2672#partial-pull-merging

Vilkina:
The following results are obtained by running benchmarks on the original Fold implementation:
```Scala
Benchmark                  (chunkSize)  Mode  Cnt   Score    Error  Units
ChunkBenchmarks.arrayFold          100  avgt   15  19.583 ± 13.368  ns/op
ChunkBenchmarks.arrayFold         1000  avgt   15  46.130 ± 38.537  ns/op ChunkBenchmarks.arrayFold        10000  avgt   15   7.204 ±  0.067  ns/op
ChunkBenchmarks.chunkFold          100  avgt   15   5.910 ±  1.782  ns/op
ChunkBenchmarks.chunkFold         1000  avgt   15   5.411 ±  0.038  ns/op
ChunkBenchmarks.chunkFold        10000  avgt   15   6.004 ±  0.955  ns/op
```

Vilkina:
```Scala
[info] Benchmark                     (chunkSize)  Mode  Cnt    Score    Error  Units
[info] ChunkBenchmarks.arrayFind             100  avgt   15    9.608 ±  0.916  ns/op
[info] ChunkBenchmarks.arrayFind            1000  avgt   15    9.250 ±  0.043  ns/op
[info] ChunkBenchmarks.arrayFind           10000  avgt   15   10.450 ±  1.991  ns/op
[...]

PR #2330, opened 2019-11-25
Chunk#apply considered harmful

vasilmkd:
Alternative implementation that I would like to see benchmarked as part of #2269, if anyone is actively working on that.

Let me try to explain my reasoning.

All of the Chunk combinators were implemented by accessing each element of the chunk that we are reading from using `def apply(n: Int): A`, which is a polymorphic method on the JVM that _must_ return a reference type, and therefore incur boxing overhead for primitives.

Now, after extensive debates with @iravid (many thanks), we came to the conclusion that most of the higher order functions in the combinators will force boxing for some of the primitive types, and if we are lucky, no boxing for `Int, Long, Float, Double`. If you are interested, check the specialization of [`Function1`](https://github.com/scala/scala/blob/v2.13.1/src/library/scala/Function1.scala#L65).

But, examining the specialization of `Function1` gave me an idea. Namely, `Function1`s that return `Boolean` (predicates) and `Unit` (Java void functions) are specialized and don't incur boxing overhead for some of the primitive input types. I also noticed that we had an implementation of `foreach` for Chunk that was lying around unused.

This PR removes the old `apply` method (there is one transient implementation that is used in `ZStream` code for now, to avoid changing the code) and defines all Chunk interactions through bulk traversal methods like `loopWhile`, `loop`, `foreach`, `foreachWhile`. These methods accept a predicate and a void function, whose side effects are used for propagating change and building new Chunks internally. This allows the implementation of these methods (the actual Chunk type, like Arr, Concat, Slice and in the future `java.nio.Buffer`) to use the internal magic of the backing type (array, vector, buffer, other chunk) to avoid boxing as much as possible. Of course, in the event of an unspecialized type like `Byte` (which is very sad to me), this would still incur boxing, but is unavoidable due to how Scala `Function1` works.

The good news is that we already did special overrides for Array backed chunks which didn't go through `def apply` and this PR does not change any of them. The PR aims to improve the performance of Slice and Concat Chunks, by allowing the underlying chunks to sort out the boxing requirements, instead of forcing a boxing through `apply`.

The code is very dirty, and I took some shortcuts in places (`def equals`) in order to pass the tests so there is still room for improvements, but the combinators are working reliably. I wanted to gather some feedback first.

I'm interested to hear some thoughts on this.

@iravid @jdegoes @Vilkina 

Edit: Added `java.nio.Buffer` backed Chunks.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -18,6 +18,7 @@ package zio
 
 import scala.collection.mutable.Builder
 import scala.reflect.{ classTag, ClassTag }
+import java.nio._
 
 /**
  * A `Chunk[A]` represents a chunk of values of type `A`. Chunks are designed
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -334,26 +294,22 @@ sealed trait Chunk[+A] { self =>
...
-    var s: S1          = s1
-    var i              = 0
-    var dest: Array[B] = null.asInstanceOf[Array[B]]
-    val len            = self.length
+    var s    = s1
+    var dest = null.asInstanceOf[Array[B]]
 
-    while (i < len) {
-      val a = self(i)
+    var i = 0
```
iravid: @vasilmkd The problem with this approach is that capturing a `var` in a closure causes it to be boxed inside an `ObjectRef`. That slows down the loops considerably. I am not sure this tradeoff is worth it for partially avoiding boxing in some cases.

iravid: Do you think you could benchmark the affect of this change?

vasilmkd: Learning things the hard way. 😄 Worth a try. Feel free to close the PR.

jdegoes:
@vasilmkd We do need to fix this problem.

However, would you be interested in exploring something slightly different?

```scala
trait SpecializedIndex[A] {
  def apply(i: Int)(implicit ev: IsInt[A]): Int = ???
  def apply(i: Int)(implicit ev: IsLong[A]): Int = ???
  def apply(i: Int)(implicit ev: IsDouble[A]): Double = ???
  def apply(i: Int)(implicit ev: IsFloat[A]): Double = ???
  ...
}
```

The types `IsInt`, `IsDouble`, etc., would not have a common supertype. Scala would treat them as distinct and be capable of resolving them. Since we have compile-time proof of type equality, we can "throw" from the "wrong" apply method. 

Then override the "right" one inside the appropriate `Chunk` subtype.

No API changes but full specialization and type safety.

vasilmkd:
What about directly adding these methods to the base trait `Chunk[A]`? @jdegoes

jdegoes:
@vasilmkd I don't really care where they are added but it's nice to avoid clutter in main classes.

In any case maybe worth prototyping...

vasilmkd:
I will definitely get to fleshing this out nicely. I'm curious how we could expose these methods to end users, who would like to drop down to quite a low level, getting individual `Byte`s from a `Chunk`. @iravid

jdegoes:
See #2526 

I don't think that's the full story actually but we should move conversation to there.

PR #2332, opened 2019-11-25
TMap documentation

drmarjanovic:
Document `TMap` (part of #2122).

### DIFF
--- a/docs/datatypes/tmap.md
+++ b/docs/datatypes/tmap.md
@@ -100,6 +100,8 @@ val retainedEvenValues: UIO[TMap[String, Int]] = (for {
 } yield tMap).commit
 ```
 
+Note that `retainIf` and `removeIf` serve the same purpose as `filter` and `filterNot`. The reason for naming them differently was to emphasize a distinction in their nature. Namely, both `retainIf` and `removeIf` are destructive - calling them can modify the collection.
+
 ## Retrieve the value from a TMap
 
 Value by key can be obtained as follows: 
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Btw is it intended to add the new pages but not adding them to the index and the sidebar? Do you plan to to do it at the end or was it just missed?

Talking about these:
https://github.com/zio/zio/blob/master/docs/datatypes/index.md
https://github.com/zio/zio/blob/master/website/sidebars.json

drmarjanovic:
@ghostdogpr, thanks for pointing it out. 

Since I'm planning to document most of the STM types, and as I can see there's not even `TRef` added to the index and the sidebar, I can provide separate PR with linking all datatypes after all of them are individually merged. 

But, if you want, I can add `TRef`, `TArray`, `TSet` and `TMap` to those files as part of this PR. 

Just let me know. 

Thanks in advance.

ghostdogpr:
You can do as you prefer, I just wanted to make sure you knew about it. Thanks for writing this! 👍

docs/datatypes/tmap.md
```diff
@@ -0,0 +1,279 @@
...
+
+Or you can keep all key-value pairs that match predicate function:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val retainedEvenValues: UIO[TMap[String, Int]] = (for {
+  tMap <- TMap.make(("a", 1), ("b", 2), ("c", 3), ("d", 4))
+  _    <- tMap.retainIf((_, v) => v % 2 == 0)
```
mijicd: Point out the parity between `removeIf` / `retainIf` and `filter` / `filterNot`, as well as with `transform`s and `map`s. This remark applies for sets as well.

docs/datatypes/tmap.md
```diff
@@ -0,0 +1,283 @@
...
+import zio._
+import zio.stm._
+
+val transformValuesMTMap: UIO[TMap[String, Int]] = (for {
+  tMap <- TMap.make(("a", 1), ("b", 2), ("c", 3))
+  _    <- tMap.transformValuesM(v => STM.succeed(v * v))
+} yield tMap).commit
+```
+
+Note that both `transform` and `transformValues` serve the same purpose as `map`. The reason for naming them differently was to emphasize a distinction in their nature. Namely, both `transform` and `transformValues` are destructive - calling them can modify the collection.
```
mijicd: ...as `map` and `mapValues`.

PR #2333, opened 2019-11-25
Support Scoped Changes to Environment

adamgfraser:
Exploration of using `FiberRef`s to support scoped modification of part of an environment. This is currently implemented for a `Logger` interface but if we like this we could apply it to `Clock` to solve the `jittered` problem.


jdegoes:
While not an "ultimate" solution to the problem of modifying environments of unknown shape, this seems straightforward and will solve `jittered` and problems like it. 

We might have to augment signature of the decorator a bit, make it a bit more powerful, but I don't see any blockers.

@adamgfraser What do you think?

adamgfraser:
@jdegoes Yes I think this works relatively well. In cases where the environmental effect has more methods there may be more boilerplate here because we need a data type to capture what to do with each method (e.g. for `Random` do we need to have separate fields or data types for what each method should do?) but that is not a major problem and worst case is just a little more one time work on our end.

My main question with this is whether we should implement this or go straight to the lens encoding. If we feel we have a clear path forward on the lens encoding I can implement that relatively quickly so it may make sense to go straight there but if not I think this provides an effective solution for the problems we have been encountering.

adamgfraser:
Closing in favor of #2594.

PR #2335, opened 2019-11-26
ZIO Test: Support Effectual Assertions

### LINKED ISSUE #1888: ZIO test: Impedence mismatch with && and ||
While it is possible to write
```scala
assert(x, assertionX) && assert(y, assertionY)
```
it is not to write
```scala
assertM(x, assertionX) && assertM(y, assertionY)
```
or indeed any combination of `assert`, `assertM`, `||`, `&&`, etc.

One is forced to write more boilerplate:

```scala
(assertM(x, assertionX) zipWith assertM(y, assertionY))(_ && _)
```
which doesn't scale well with more than two assertions.

### LINKED ISSUE #2182: ZIO Test: Explore Effectual Assertions
It would be nice to be able to use effects inside of an assertion. For example, I was thinking it could be good to have an `isSerializable` assertion that would serialize an object, deserialize it, and check that the objects are the same to support our efforts to make sure all our data types are serializable. But we can't do that today because serializing and deserializing involves an effect.

I am working on this now and expect to have something to share soon.

adamgfraser:
Adds support for effectual assertions by implementing a new `BoolAlgebraM` class that reifies an effectually computed assertion result. Doing this also allows us to offer combinators such as `&&` that works across `assert` and `assertM`. This still needs some cleanup and I think can be further simplified but wanted to get it out there.

Resolves #1888. Resolves #2182.


regiskuckaertz:
You rock 💪

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -293,11 +293,21 @@ object TestAspect extends TimeoutVariants {
...
-            .filter(assertion.test)
-            .fold[ZIO[R, TestFailure[E], TestSuccess[S]]](ZIO.succeed(TestSuccess.Ignored))(
+          val x = value match {
+            case None => ZIO.succeed(None)
+            case Some(v) =>
+              assertion.test(v).map { p =>
+                if (p) value else None
+              }
+          }
+          //value
```
regiskuckaertz: can be removed

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -176,8 +175,23 @@ object Assertion {
...
+    }
+    assertion
+  }
+
+  /**
+   * Makes a new `Assertion` from a pretty-printing and a function.
+   */
+  final def assertionM[R, E, A](
+    name: String
+  )(params: RenderParam*)(run: (=> A) => ZIO[Any, Nothing, Boolean]): Assertion[A] = {
```
jdegoes: `UIO[Boolean]`?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -18,8 +18,7 @@ package zio.test
 
 import scala.reflect.ClassTag
 
-import zio.Cause
-import zio.Exit
+import zio.{ Cause, Exit, ZIO }
```
jdegoes: I suppose we can add assertions on effects directly now, e.g. `failsM(zio)` and the like...

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
```diff
@@ -97,6 +99,13 @@ sealed trait BoolAlgebra[+A] extends Product with Serializable { self =>
   final def flatMap[B](f: A => BoolAlgebra[B]): BoolAlgebra[B] =
     fold(f)(and, or, not)
 
+  /**
+   * Returns a new result, with all values mapped to new results using the
+   * specified effectual function.
+   */
+  final def flatMapM[R, E, B](f: A => ZIO[R, E, BoolAlgebra[B]]): ZIO[R, E, BoolAlgebra[B]] =
```
jdegoes: 👍

test/shared/src/main/scala/zio/test/BoolAlgebra.scala
```diff
@@ -148,6 +157,13 @@ sealed trait BoolAlgebra[+A] extends Product with Serializable { self =>
   final def map[B](f: A => B): BoolAlgebra[B] =
     flatMap(f andThen success)
 
+  /**
+   * Returns a new result, with all values mapped by the specified effectual
+   * function.
+   */
+  final def mapM[R, E, B](f: A => ZIO[R, E, B]): ZIO[R, E, BoolAlgebra[B]] =
```
jdegoes: 👍

jdegoes:
@adamgfraser Really awesome work! It'd be great to see an example of where we can use this new power, e.g. #1888 and maybe new assertions (effectful versions of `fails`, etc.). Just enough to get a taste for the benefits.

PR #2337, opened 2019-11-26
ZIO Stream: Improve ZStream#accessM signature

adamgfraser:
Someone pointed out on Discord that the existing signature of `ZStream#accessM` is a bit too narrow in that it requires the `E` type to be `Nothing` instead of allowing it to be polymorphic.

When I was looking at that I realized the current signature is also inconsistent with other combinators in that right now it takes a `R => ZStream[R, E, A]` whereas all other `M` combinators take a function that returns a `ZIO`. I think this makes more sense and then if users want to create a different stream based on the value accessed they can just flatMap it.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2695,16 +2695,22 @@ object ZStream extends Serializable {
     new ZStream[R, E, A](ZStream.Structure.Iterator(pull))
 
   /**
-   * Creates a stream from a value received from the accessed environment.
+   * Accesses the environment of the stream.
    */
   final def access[R]: AccessPartiallyApplied[R] =
     new AccessPartiallyApplied[R]
 
   /**
-   * Creates a stream from a value received from the effectfully accessed environment.
+   * Accesses the environment of the stream in the context of an effect.
    */
   final def accessM[R]: AccessMPartiallyApplied[R] =
     new AccessMPartiallyApplied[R]
+
+  /**
+   * Accesses the environment of the stream in the context of a stream.
+   */
+  final def accessStream[R]: AccessStreamPartiallyApplied[R] =
+    new AccessStreamPartiallyApplied[R]
 
   /**
    * Creates a stream from a single value that will get cleaned up after the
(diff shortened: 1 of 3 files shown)


mschuwalow:
I think having ZStream in the signature is more flexible. You can always go from ZIO => ZStream painlessly but the inverse is not true.

Also why would a user be using ZStream.accessM and not ZIO.accessM in this case?

adamgfraser:
I don't have a particularly strong view. All of this functionality can be done pretty easily with existing combinators. For example here is the original version of `accessM`:

```scala
ZStream.fromEffect(ZIO.environment[R]).flatMap(f)
ZStream.environment[R].flatMap(f) // If we implemented ZStream.environment
```

So these are all convenience methods for saving a couple of relatively short method calls. I think in some cases the newer version is slightly easier to work with because it automatically wraps the environment type in a single element stream and then you can manipulate it with normal stream combinators versus having to construct the stream manually yourself:

```scala
val helloWorld = accessM[Console](_.console.putStrLn("Hello, World!")).forever
val helloWorld = accessM[Console](console => ZStream.repeatEffect(console.putStrLn("Hello, World!")))
```

I can see how in some cases where you wanted to do fancier stuff it would be nice to be able to create the whole stream manually but you're not saving many keystrokes versus `ZStream.environment[R].flatMap(f)`.

Maybe we should have both and have different names for them? Or I'm fine with leaving as is. It just kind of jumped out at me that I think this is the only `M` method that doesn't use a `ZIO`. I do think we should be consistent with whatever we do on `ZStream` and `ZManaged`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3141,8 +3141,8 @@ object ZStream extends Serializable {
   }
 
   final class AccessMPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
-    def apply[E, A](f: R => ZStream[R, Nothing, A]): ZStream[R, Nothing, A] =
-      ZStream.unwrap(ZIO.access(f))
+    def apply[E, A](f: R => ZIO[R, E, A]): ZStream[R, E, A] =
```
jdegoes: I think we have to decide what `M` means. If we mean `M` as the "host monad", it makes sense to lock this to `ZStream`", but if we intend it to be "ZIO monad", then it makes more sense to change to `ZIO` as in this pull request.

In any case, the generalization on the error channel is very useful! Probably an oversight in the preceding version.

If all the `mapM` etc. functions on` ZStream` are expecting `ZIO`, then maybe we should change this one to `ZIO`. We could always add `accessStream` for the stream variant, if we decided it was too cumbersome to use the combinators.

adamgfraser: I think that makes sense. I updated accordingly.

Issue #2309, opened 2019-11-23
Spurious Fiber Failure Warning when ZIO.repeat is used


reibitto:
I get a fiber failure warning for this when it completes:
```scala
val schedule = Schedule.recurs(10) && Schedule.fixed(100.millis)
putStrLn("test").repeat(schedule)
```

Output:
```
A checked error was not handled.
()

Fiber:Id(1574499603310,35) was supposed to continue to: <empty trace>

Fiber:Id(1574499603310,35) execution trace:
  at zio.Schedule.check(Schedule.scala:238)
...
```

But this completes fine without any warnings:
```scala
val schedule = Schedule.recurs(10)
putStrLn("test").repeat(schedule)
```

Possibly related to #1949.

reibitto:
It looks like it's the introduction of `&&` that causes this, not `Schedule.fixed` itself. For example, you can also see it with something like `Schedule.once && Schedule.once`. A plain `Schedule.once` is fine though.

Twizty:
Hi, I'd like to implement it.

reibitto:
@Twizty I haven't started this beyond looking for a minimal example. Sure, you can take it. Thanks!

Twizty:
@reibitto ok, then I start working on it.

thiloplanz:
Is this also the cause of the issue reported at https://stackoverflow.com/q/59063170/14955 ?

adamgfraser:
@thiloplanz I think it is related but slightly different. We have a general problem where fiber failures are reported when sometimes they shouldn't be because they are "intentional" in some sense (e.g. `race` is supposed to interrupt the loser). This issue in particular is caused by `race` because `Schedule#&&` is implemented in terms of `zipPar` which is implemented in terms of `race`. The Stack Overflow example involves `orElse` which doesn't use `race` and raises a slightly different issue of a fiber that should not initially be silent but should be silent after `orElse` is called on it. I will open a ticket for it.

adamgfraser:
Closed by #2339.

PR #2339, opened 2019-11-26
#2309 Introduced quite mode for fibers

Twizty:
As described in the issue #2309, forked fiber may print an error message if it fails. To avoid it I introduced quite mode for fiber races. I even considered using quite mode as a default mode for all fibers except the main one. Please suggest improvements.

### DIFF
--- a/core/js/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/js/src/main/scala/zio/internal/PlatformLive.scala
@@ -40,7 +40,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly)
+        if (!cause.interruptedOnly && cause.died)
           println(cause.prettyPrint)
 
       val tracing = Tracing(Tracer.Empty, TracingConfig.disabled)
--- a/core/js/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/js/src/main/scala/zio/internal/PlatformLive.scala
@@ -40,7 +40,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly && cause.died)
+        if (cause.died)
           println(cause.prettyPrint)
 
       val tracing = Tracing(Tracer.Empty, TracingConfig.disabled)
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -519,6 +519,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def fork: URIO[R, Fiber[E, A]] = new ZIO.Fork(self)
 
   /**
+   * Like [[fork]] but handles an error with the provided handler.
+   */
+  final def forkWithErrorHandler(handler: E => UIO[Unit]): URIO[R, Fiber[E, A]] =
+    onError(new ZIO.FoldCauseMFailureFn(handler)).forkInternal
+
+  /**
    * Forks the effect into a new independent fiber, with the specified name.
    */
   final def forkAs(name: String): URIO[R, Fiber[E, A]] =
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -522,7 +522,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Like [[fork]] but handles an error with the provided handler.
    */
   final def forkWithErrorHandler(handler: E => UIO[Unit]): URIO[R, Fiber[E, A]] =
-    onError(new ZIO.FoldCauseMFailureFn(handler)).forkInternal
+    onError(new ZIO.FoldCauseMFailureFn(handler)).run.fork.map(_.mapM(IO.done))
 
   /**
    * Forks the effect into a new independent fiber, with the specified name.
@@ -1628,14 +1628,6 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    */
   final def <*>[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[R1, E1, (A, B)] =
     self &&& that
-
-  /**
-   * Forks an effect that will be executed without unhandled failures being
-   * reported. This is useful for implementing combinators that handle failures
-   * themselves.
-   */
-  private[zio] final def forkInternal: URIO[R, Fiber[E, A]] =
-    run.fork.map(_.mapM(IO.done))
 }
 
 private[zio] trait ZIOFunctions extends Serializable {
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -540,7 +540,7 @@ object ZIOSpec extends ZIOBaseSpec {
       testM("calls provided function when task fails") {
         for {
           p <- Promise.make[Nothing, Unit]
-          _ <- ZIO.fail(()).forkWithErrorHandler(_ => p.succeed(()).unit)
+          _ <- ZIO.fail(()).forkWithErrorHandler(p.succeed(_).unit)
           _ <- p.await
         } yield assertCompletes
       }
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -536,6 +536,15 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(result, equalTo(Cause.die(boom)))
       } @@ flaky
     ),
+    suite("forkWithErrorHandler")(
+      testM("calls provided function when task fails") {
+        for {
+          p <- Promise.make[Nothing, Unit]
+          _ <- ZIO.fail(()).forkWithErrorHandler(_ => p.succeed(()).unit)
+          _ <- p.await
+        } yield assertCompletes
+      }
+    ),
     suite("head")(
       testM("on non empty list") {
         assertM(ZIO.succeed(List(1, 2, 3)).head.either, isRight(equalTo(1)))
--- a/core/jvm/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/jvm/src/main/scala/zio/internal/PlatformLive.scala
@@ -59,7 +59,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly && cause.died)
+        if (cause.died)
           System.err.println(cause.prettyPrint)
 
     }
--- a/core/jvm/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/jvm/src/main/scala/zio/internal/PlatformLive.scala
@@ -59,7 +59,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly)
+        if (!cause.interruptedOnly && cause.died)
           System.err.println(cause.prettyPrint)
 
     }
--- a/core/native/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/native/src/main/scala/zio/internal/PlatformLive.scala
@@ -40,7 +40,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly)
+        if (!cause.interruptedOnly && cause.died)
           println(cause.prettyPrint)
 
       val tracing = Tracing(Tracer.Empty, TracingConfig.disabled)
--- a/core/native/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/native/src/main/scala/zio/internal/PlatformLive.scala
@@ -40,7 +40,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly && cause.died)
+        if (cause.died)
           println(cause.prettyPrint)
 
       val tracing = Tracing(Tracer.Empty, TracingConfig.disabled)


Twizty:
@adamgfraser please take a look.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -41,7 +41,8 @@ private[zio] final class FiberContext[E, A](
   startDStatus: Boolean,
   parentTrace: Option[ZTrace],
   initialTracingStatus: Boolean,
-  val fiberRefLocals: FiberRefLocals
+  val fiberRefLocals: FiberRefLocals,
+  quiteMode: Boolean = false
```
adamgfraser: Maybe call this `silent`?

neko-kai:
This is the same as [forkInternal](https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZIO.scala#L1648)

Perhaps `forkInternal` should be un-privated and possibly renamed to `forkQuiet`? (rename will break bincompat for next release as `forkInternal` is already used in interop-cats)

neko-kai:
@adamgfraser 
Same logic as in `forkInternal` can be re-used inside the run loop, we may probably get away without adding a state boolean:

```scala
val left  = fork[EL, A](race.left.interruptible.asInstanceOf[IO[EL, A]].run, true).mapM(IO.done)
```

`continue` accepts abstract Fiber so we can still use `mapM` and lose the precise type of FiberContext

adamgfraser:
@neko-kai My worry with that is that we moved the `race` implementation into the run loop to improve performance and if we are doing `run` and `mapM` for every forked fiber we are going to take a hit there, though I haven't benchmarked how much and you are right we could save allocating a boolean. What do you think?

I agree about `forkInternal`. I think the original idea is that having fibers that don't report failures is dangerous and we wanted to control it, but you're right it can be useful in other applications and if we call it something like `forkSilent` I think it is pretty clear what you're getting.

Twizty:
@neko-kai @adamgfraser I think using `forkInternal` is a better option. I will change my implementation to use it instead.

Twizty:
@neko-kai @adamgfraser I've reimplmented it like in the example above, but there is a problem: all `StackTracesSpec` fail because there is an additional stack frame for each task. And the code itself is a bit scarry because `fork[EL, A](race.left.interruptible.asInstanceOf[IO[EL, A]].run, true).mapM(IO.done)` returns `Fiber` instead of `FiberContext`. Please take a look at the new implemetation. In my opinion the previous one is better.

adamgfraser:
I think we need to consider a different approach here. #2354 should remove all fiber failure warnings for interruptions. However, in implementing that I realized that it actually still does not solve the problem here. The issue is that `Schedule` encodes `update` as:

```scala
val update: (A, State) => ZIO[R, Unit, State]
```

Here `Fail(())` represents that the schedule should not continue.  So the fiber failure warning is being generated even without logging any pure interruptions because the `Cause` contains a `Fail` when we do `zipPar` in `Schedule#&&` and one `Schedule` terminates.

At this point it isn't clear to me that we should be making any changes to `race`. Not logging fiber failure warnings for interruption of the loser makes sense because interruption is intended but that will be accomplished once we merge #2354. After that I don't think it is right as a rule that we don't want to log other fiber failures in `race`. It just happened to be the case here that we didn't want to log the failure from the `Schedule` not continuing and suppressing all logging accomplished that.

So I think we need to step back and think about how the caller in functions like `Schedule.check` or `Fiber.orElse` can specify that a certain fiber failure should not be logged. One idea would be to expand `Cause.Meta` to include a `silent` field to indicate that a fiber failure should not be logged but there may be other approaches.

Copying @mschuwalow regarding `Schedule`.

neko-kai:
@adamgfraser 
> One idea would be to expand Cause.Meta to include a silent field to indicate that a fiber failure should not be logged 

I like that! But what if such a Cause is then rethrown as part of a Cause that should be logged - should it only be silent if the topmost node is silent?
Another, ad-hoc way, would be to create zipParInternal and use it in Schedule

adamgfraser:
@neko-kai What if we process the `Cause` recursively and silence everything below a silent node? You could implement it as filtering out any silent nodes and their children from the cause tree in `reportFailure` and then if there is anything left logging that.

Twizty:
@adamgfraser I changed race function because simple examples like `ZIO.fail(()).race(ZIO.succeed(()))` would still log failure. Is it an expected behavior?

adamgfraser:
I would say that is expected behavior. It is a little bit of an unusual case because normally we would interrupt the loser so we wouldn't see whether or not it failed, but in this case the `succeed` could win and the `fail` could already be completed.

But I would say the expectation is that while one effect may be interrupted neither effect should fail and if it does we want to report it somewhere. In this case it doesn't really matter but say the loser was an uninterruptible effect that you were expecting to do something. I don't think it makes sense to silently drop that failure.

There are a bunch of other `race` combinators to deal with other scenarios of the effects being raced possibly failing.

Twizty:
I guess misunderstood this issue. Then it seems like a good idea to me to do something with `Schedule` combinators (such as `&&` or `||`) to catch failures on the level of their `update` function. I'll try to implement that then.

adamgfraser:
I think we're all figuring this one out. Can you take a look at my suggestion regarding `Cause.Meta`? I think that could give us a nice way to address this.

Twizty:
@adamgfraser I'm sorry, I missed it. Seems like a very good idea. And I like the idea of silencing everything below a silent node.

Twizty:
@adamgfraser @neko-kai please take a look at the implementation of the `Cause.Meta` expansion.

mschuwalow:
Sorry for being quiet here, haha :D
I like the idea of tracking this in Cause.Meta. @Twizty I'll take a look once I get home

Twizty:
It might also be a good idea to use `catchSomeCause` instead `catchAllCause` and label Schedule failures with some specific error type.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2045,6 +2042,9 @@ private[zio] trait ZIOFunctions extends Serializable {
    */
   final def fail[E](error: E): IO[E, Nothing] = haltWith(trace => Cause.Traced(Cause.Fail(error), trace()))
 
+  private[zio] final def failSilently[E](error: E): IO[E, Nothing] =
+    haltWith(trace => Cause.silent(Cause.Traced(Cause.Fail(error), trace())))
```
neko-kai: If you're silencing it might as well skip the trace!

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -609,7 +616,7 @@ object Cause extends Serializable {
       new Both(left, right)
   }
 
-  private final case class Data(stackless: Boolean)
```
mschuwalow: Does it ever make sense to have a stackful, silent cause? If not, I propose to use a custom coproduct here for better readability

Twizty: Do you mean something like that?
```scala
sealed trait Data
case class Stackless(value: Boolean) extends Data
case class Silent(value: Boolean) extends Data
```

Twizty: Or even
```
sealed trait Data
case object Stackless extends Data
case object Silent extends Data
```

mschuwalow: Yeah, like this 👍

adamgfraser: One thing we're losing here is the ability to "turn on" and "turn off" stack traces. In the previous implementation stackless was compositional with inner scopes taking precedence (like almost everything else), so high up in the Cause tree  someone could say I don't want to show stack traces but then lower down someone could say no I really do want to show the trace for this cause.

Right now we don't need to do that for `silent` because we are only calling it on simple causes but there may be value in that in the future. There is also some value in having all `Meta` options work the same way. So I would suggest we have `final case class Data(stackless: Boolean, silent: Boolean)`.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -130,6 +130,12 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       case (_, Interrupt(_))    => false
     }
 
+  final def isTopLevelSilent: Boolean =
```
adamgfraser: I think we need to do more here. Because if we have, say `Both(cause, silent)` we want to report `cause` but not `silent`. So I think we want to walk the cause tree recursively and eliminate the children of silent nodes. Per my comment below, I think it might be nice to have `silent` have the same compositional semantics as `stackless` where an inner cause can say it doesn't want to be silent. Either way, think this is really something like:

```scala
final def stripSilent: Option[Cause[E]]
```

test/shared/src/main/scala/zio/test/GenZIO.scala
```diff
@@ -30,7 +30,7 @@ trait GenZIO {
     val empty          = Gen.const(Cause.empty)
     val interrupt      = Gen.anyLong.crossWith(Gen.anyLong)((l, r) => Cause.interrupt(Fiber.Id(l, r)))
     def traced(n: Int) = Gen.suspend(causesN(n - 1).map(Cause.Traced(_, ZTrace(Fiber.Id(0L, 0L), Nil, Nil, None))))
-    def meta(n: Int)   = Gen.suspend(causesN(n - 1).flatMap(c => Gen.elements(Cause.stack(c), Cause.stackless(c))))
+    def meta(n: Int)   = Gen.suspend(causesN(n - 1).flatMap(c => Gen.elements(Cause.silent(c), Cause.stackless(c))))
```
adamgfraser: We should generate cases with all types of `Meta` here.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2042,6 +2042,9 @@ private[zio] trait ZIOFunctions extends Serializable {
    */
   final def fail[E](error: E): IO[E, Nothing] = haltWith(trace => Cause.Traced(Cause.Fail(error), trace()))
 
+  private[zio] final def failSilently[E](error: E): IO[E, Nothing] =
```
adamgfraser: How about the following method on ZIO instead:

```scala
final def silence: ZIO[R, E, A] = mapErrorCause(cause => Cause.silent(cause))
```

Then in `Schedule` you could do:

```scala
ZIO.fail(()).silent
```

This would be a bit more compositional and at that point I think you could expose it as a pubic method which would address @neko-kai's earlier point of wanting to make `forkInternal` public.

Twizty:
@adamgfraser thank you for the review. And I think that I can fix #2349 in this branch as well.

core/shared/src/main/scala/zio/Cause.scala
jdegoes: Loss of compositional power is a serious red flag...

adamgfraser: @jdegoes If we do what I am suggesting we will maintain the current compositional power of `stackless` and `silent` will compose the same way.

core/shared/src/main/scala/zio/ZIO.scala
neko-kai: @adamgfraser Great idea!

Twizty:
@adamgfraser I've tried to #2349 in this issue, but there is too much work to do, I wasn't able to implement it in terms of `Cause.Silent`, but I think it's possible to implement it as a certain type of `Callback[E, A]`. I think it's better to do it in a separate branch, and I will start doing it right after this issue.

jdegoes:
Will comment momentarily!

Twizty:
I think I know a better solution, I will reimplement it when I come home.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -140,6 +140,36 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       case (_, Interrupt(_))    => false
     }
 
+  final def stripTopSilent: Cause[E] =
```
adamgfraser: Dead code?

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -140,6 +140,36 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
+    self match {
+      case Meta(cause, Data(_, true)) => cause
+      case other                      => other
+    }
+
+  final def stripSilent: Option[Cause[E]] =
+    self match {
+      case Meta(_, Data(_, true)) => None
```
adamgfraser: To be compositional we want to say that we won't report children of silent nodes unless a lower level `Meta` indicates that they should not be silent. So for example if we have:

```scala
Cause.silent(Cause.Both(Cause.Fail("fail1"), Cause.nonSilenct(Cause.Fail("fail2')))
```

`stripSilent` should return `Some(Cause.Fail("fail2"))`. So we need to recurse in the `Meta(_, true)` case.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -429,8 +459,9 @@ object Cause extends Serializable {
   final def die(defect: Throwable): Cause[Nothing]              = Die(defect)
   final def fail[E](error: E): Cause[E]                         = Fail(error)
   final def interrupt(fiberId: Fiber.Id): Cause[Nothing]        = Interrupt(fiberId)
-  final def stack[E](cause: Cause[E]): Cause[E]                 = Meta(cause, Data(false))
-  final def stackless[E](cause: Cause[E]): Cause[E]             = Meta(cause, Data(true))
+  final def stack[E](cause: Cause[E]): Cause[E]                 = Meta(cause, Data(false, false))
+  final def stackless[E](cause: Cause[E]): Cause[E]             = Meta(cause, Data(true, false))
+  final def silent[E](cause: Cause[E]): Cause[E]                = Meta(cause, Data(false, true))
```
adamgfraser: We should add `nonSilent` here.

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -264,7 +294,7 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
 
     def causeToSequential(cause: Cause[Any], maybeData: Option[Data]): Sequential =
       cause match {
-        case Empty => Sequential(Nil)
+        case Empty | Meta(_, Data(_, true)) => Sequential(Nil)
```
adamgfraser: I think we want to keep the logic for pretty printing a `Cause` separate from whether the `DefaultRuntime` should report a failure. So I would just ignore `silent` for purpose of pretty printing.

core/shared/src/main/scala/zio/Cause.scala
mschuwalow: @adamgfraser Good point. In that case I agree that the previous encoding was superior

core/shared/src/main/scala/zio/Cause.scala
mschuwalow: agreed.
Also currently silent will also enable stacktraces which I don't believe should be the case?

jdegoes:
@Twizty @adamgfraser @neko-kai @mschuwalow 

One way to deal with this problem:

`effect.fork` will report a failure if `effect` doesn't handle it's own errors.

If we don't want that, we can call, `effect.run.fork.mapM(IO.done)` as suggested by @neko-kai. This is a generic technique for suppressing errors, which works because the error is handled.

On the other hand, it adds notable overhead (quite significant in the case of a performance sensitive operation like `race`).

Another option is to say that, really: an error should be reported only if no one sees it. The problem is that due to race conditions, you may inspect a fiber and "see" its error only after it has already completed and yielded its exit value. There is no reliable means of detecting this.

Another option is, as proposed, make the error reporting a property of the error itself. However, after thinking about this in the afternoon, I could not convince myself that, in general, the code that generates the error knows whether or not it should be silent: this seems more use-case driven, e.g., how am I using the effect, rather than with what error it fails with. 

Indeed, coming up with rules for what to do with "disagreements" in composed errors seems a little arbitrary: in a `Both`, if one side wants to report, but the other side is silent, what do you do? How about `Then`? You can of course pick rules but why pick those? If there is no reason to pick one versus the other it's a warning sign.

Now while it's a property of the error, you can "refail" to make it more use case driven: catch the cause and make it silent and rethrow it. Yet if we were trying to avoid `z.run.fork.mapM(IO.done)` overhead, we came close to failing with the rethrowing.

Here's another idea:

 - Add a parameter to each fiber which is the unhandled error reporter
 - By default, in `Runtime`, start out the root effect with the `Platform` unhandled error reporter
 - When fibers are forked, pass the parent unhandled error reporter
 - In `raceWith` implementation, pass a dummy unhandled error reporter to the children

There's a more powerful variant of this (stack of unhandled error reporters, to allow region-based setting), but it'd require a "pure" unhandled error reporter, which causes other problems.

adamgfraser:
@jdegoes I think there is a relatively simple way to address your concerns about the error reporting. Allow each error to specify whether it should be reported and only remove from reporting the ones that specify they should not be reported. So in `Both` we don't need any special logic. If one side says it should be reported and the other side says it shouldn't we report the side that says it should. We also allow inner scopes that specify they should or should not be reported to override outer scopes. So for example a caller could specify that an effect that normally does not have any special error reporting logic should not be reported in this case. This mirrors the existing composition logic of `stackless` and I think provides a powerful tool to control error reporting in a compositional way.

That being said, I think your proposal of each fiber having an unhandled reporter would also work and is quite straightforward. One factor to consider might be that on the margin including the unhandled error reporter makes all fibers slightly more heavyweight versus adding a property to the error which only impacts the case that is not on the "happy path".

Twizty:
@jdegoes I missed your post earlier today, but made quite similar implementation: basic idea of handled and unhandled errors is the same, but I suggest making fibers unhandled by default, and add a handler when `.await` or `.join` is used or when we know for sure that we would handle an error in future (i.e. in `.race` implementation).

core/js/src/main/scala/zio/internal/PlatformLive.scala
```diff
@@ -40,8 +40,9 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interrupted)
+        cause.stripSilent.filterNot(_.interrupted).foreach { cause =>
```
adamgfraser: I think we want to use `interruptedOnly` instead of `interrupted` here and in the versions for other platforms.

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: That is a good point. Currently we are forcing every time we call `Meta` to fully specify all the parameters versus just one. I think this would suggest going back to an encoding where `Meta` is a sum type with subtypes for `Stackless` and `Silent` and then we may just have to change the implementation of the pretty print and `stripSilent` functions with regard to how we propagate the information about the current state when we recurse down.

Twizty:
@adamgfraser I think that it might be better to revert changes of `Meta` and keep fixes at `FiberContext`, because they fix both #2309 and #2349.

core/js/src/main/scala/zio/internal/PlatformLive.scala
Twizty: I need to merge master into my branch to fix it

Twizty:
@adamgfraser @jdegoes please take a look at this issue once again. I've made some changes at `FiberContext`, allowing to specify if the fiber is handled or not, and I've also changed logging of failed fibers, now only defects are being logged.

adamgfraser:
So after all this I think we can simplify this PR dramatically. If we are going with the model that only defects are reported, and then optionally there is this convenience method to fork a fiber with an error handler, then I think we can do the following:

- In `PlatformLive.reportFailure`, only log `Die` causes.
- Implement a new method on `ZIO`, `forkWithErrorHandler` that is similar to `forkInternal` but just calls the specified error handler in the event of a failure.
- Revert all other changes, including changes to `Cause` and changes to `FiberContext`.

Can we verify that this resolves both #2309 and #2349?

@Twizty Thanks for all your work on this as we worked through the different potential solutions here!

Twizty:
@adamgfraser done. And thank you for the opportunity to take part in the development of such a good library!

core/js/src/main/scala/zio/internal/PlatformLive.scala
```diff
@@ -40,7 +40,7 @@ object PlatformLive {
       }
 
       def reportFailure(cause: Cause[Any]): Unit =
-        if (!cause.interruptedOnly)
+        if (!cause.interruptedOnly && cause.died)
```
mschuwalow: Isn't `!cause.interruptedOnly` redundant now?

Twizty: Yes, you are right! Will fix that in a moment.

regiskuckaertz: 🤔 `!cause.interruptedOnly` is true if either the fiber died or failed, so `!cause.interruptedOnly && cause.died` would fail if the fiber failed.

regiskuckaertz: So yeah 😅

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -518,6 +518,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    */
   final def fork: URIO[R, Fiber[E, A]] = new ZIO.Fork(self)
 
+  /**
+   * Like [[fork]] but handles an error with the provided handler.
+   */
+  final def forkWithErrorHandler(handler: E => UIO[Unit]): URIO[R, Fiber[E, A]] =
+    onError(new ZIO.FoldCauseMFailureFn(handler)).forkInternal
```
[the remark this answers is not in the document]
Twizty: @neko-kai said that it is used at [interop cats](https://github.com/zio/interop-cats/blob/16a0cb5a472df862046029b113a9c4d516bbd6b8/interop-cats/shared/src/main/scala/zio/interop/cats.scala#L188).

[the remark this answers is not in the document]
adamgfraser: @Twizty Good point.

@neko-kai Are you okay with us removing `forkInternal`? You could get the same behavior using `forkWithErrorHandler(_ => ZIO.unit)`. You may also not need to do anything at all under the new model depending on what you want to do with `Die` causes.

Twizty:
@adamgfraser yes, I've tested all the cases (races, combined schedules, orElse on fibers), spurious failure messages are gone.

adamgfraser:
@Twizty Awesome! Let's just sync up on what we want to do with `forkInternal` and then I think this will be good to merge.

Twizty:
@adamgfraser maybe deprecate it, until interop-cats would get rid of it?

adamgfraser:
@Twizty It is already `private[zio]` so one one is using it except `cats-interop` and they won't have an issue until they upgrade to the next release candidate at which point they can just use `forkWithErrorHandler`.

Twizty:
@adamgfraser ok then, I'll make a fix in a moment.

adamgfraser:
Let's just hold off until @neko-kai confirms he is good with the change.

adamgfraser:
@Twizty Or if you want to make the change and we can wait for approval before merging that works too.

Twizty:
@adamgfraser sure, I can just revert my commit if we decide to keep `.forkInternal`.

neko-kai:
@adamgfraser I'm fairly sure next release will break bincompat anyway so interops will be re-released. No point taking interop dependencies into account in general IMHO

adamgfraser:
@neko-kai That's what I thought, just wanted to make sure!

adamgfraser:
@Twizty Thanks again for all your work on this! This is going to make things much nicer for users of a lot of different combinators.

PR #2340, opened 2019-11-26
java.nio.Buffer backed Chunks

### LINKED ISSUE #1479: Add java.nio.Buffer-backed Chunk subtypes
To support reading buffers without copying them into arrays.

Minimally, support ByteBuffer.
Ideally, support all java.nio.Buffer subtypes. (maybe using specialization?)

vasilmkd:
Resolves #1479.

I tried to minimize code duplication as much as possible.

The implementation copies the buffer into an array backed Chunk.

I implemented it this way for two reasons. First, if we read from the buffer directly, each read will incur boxing overhead if we are going through the `Chunk` polymorphic methods like `def apply(n: Int): A`. Second, the chunk combinators are hand optimized on array backed chunks, so further boxing will be avoided as much as possible.

The tests are running on both JVM and JS.

Hopefully, this is a nice addition for `zio-nio`.

@iravid @mijicd

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -1178,108 +1178,59 @@ object Chunk {
       }
 
     final def manualByteBuffer(buffer: ByteBuffer): Chunk[Byte] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Byte](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Byte](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualCharBuffer(buffer: CharBuffer): Chunk[Char] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Char](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Char](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualDoubleBuffer(buffer: DoubleBuffer): Chunk[Double] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Double](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Double](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualFloatBuffer(buffer: FloatBuffer): Chunk[Float] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Float](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Float](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualIntBuffer(buffer: IntBuffer): Chunk[Int] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Int](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Int](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualLongBuffer(buffer: LongBuffer): Chunk[Long] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Long](len)
-      var i   = pos
-      var j   = 0
-      while (i < lim) {
-        dst(j) = buffer.get(i)
-        i += 1
-        j += 1
-      }
-      Chunk.fromArray(dst)
+      val dest = Array.ofDim[Long](buffer.remaining())
+      val pos  = buffer.position()
+      buffer.get(dest)
+      buffer.position(pos)
+      Chunk.fromArray(dest)
     }
 
     final def manualShortBuffer(buffer: ShortBuffer): Chunk[Short] = {
-      val pos = buffer.position()
-      val lim = buffer.limit()
-      val len = lim - pos
-      val dst = Array.ofDim[Short](len)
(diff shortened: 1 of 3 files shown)


vasilmkd:
Not ready for merging, there is a bug when slicing buffers. Working on a fix.

jdegoes:
@vasilmkd Love all the work you're doing with chunk! 💪

vasilmkd:
The pleasure is all mine.

iravid:
I have some comments on these so don’t merge yet 🙏🏻

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1117,6 +1160,136 @@ object Chunk {
...
+      else {
+        val pos = buffer.position()
+        val lim = buffer.limit()
+        val cap = buffer.capacity()
+        val arr = buffer.array().asInstanceOf[Array[A]]
+        val off = buffer.arrayOffset()
+
+        val chunk = if (off == 0 && pos == 0 && cap == arr.length && lim == cap) {
+          // best case, reuse the underlying array
+          fromArray(arr)
```
iravid: We can't use the backing array - gotta copy it. The general pattern for reading from File/Socket channels with NIO is based on reusing the same buffer over and over. So if we're grabbing the backing array for the buffer, we will observe mutations as the buffer is re-used.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1117,6 +1160,136 @@ object Chunk {
...
+        } else {
+          implicit val classTag: ClassTag[A] = ClassTag(arr.getClass.getComponentType)
+          val len                            = lim - pos
+          val dest                           = Array.ofDim[A](len)
```
iravid: Rather than re-implement the slicing logic, let's allocate an array of size `remaining()`, and use `Buffer#get(arr)` to copy the data into the array. Also, to avoid mutating the position, use `mark()` before `get` and `reset` afterwards.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1117,6 +1160,136 @@ object Chunk {
...
+
+    final def manualByteBuffer(buffer: ByteBuffer): Chunk[Byte] = {
+      val pos = buffer.position()
+      val lim = buffer.limit()
+      val len = lim - pos
+      val dst = Array.ofDim[Byte](len)
+      var i   = pos
+      var j   = 0
+      while (i < lim) {
+        dst(j) = buffer.get(i)
```
iravid: This is really slow. Let's use `Buffer#get(arr)` instead.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1117,6 +1160,136 @@ object Chunk {
...
+        }
+
+        Some(chunk.asInstanceOf[Chunk[A]])
+      }
+
+    final def manualByteBuffer(buffer: ByteBuffer): Chunk[Byte] = {
+      val pos = buffer.position()
+      val lim = buffer.limit()
+      val len = lim - pos
+      val dst = Array.ofDim[Byte](len)
```
iravid: You can just do `Array.ofDim(buffer.remaining())` here

iravid:
cc @pshemass, @mijicd, @mschuwalow if you'd like to review as well.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: I thought about this and you are completely right. There is another related question to this. Should we do this too in `Chunk.fromArray`?

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: The reason I implemented this is because I was not sure if we are allowed to use the `mark()` and bulk `get` methods.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: At least we should document that the Array has to be effectively immutable.

jdegoes: We should document that but not change it INO.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: To me, `mark()`ing a Buffer is conceptually the same as mutating it.

core/shared/src/main/scala/zio/Chunk.scala
pshemass: @iravid @vasilmkd I would fixed those two methods and tests for this because it will be very confusing and finding bugs in bigger application will be nightmare.

vasilmkd: I'm definitely fixing this here. Not sure about `fromArray`, it's not my call to make.

mschuwalow: I think I would err on the side of making an extra copy here and trading it for performance.

core/shared/src/main/scala/zio/Chunk.scala
mschuwalow: I think given how buffers work I think the 'expected api' would be:
```scala
buf.mark()
val chunk = Chunk.manualByteBuffer(buf)
buf.reset()
```
This would of course make manualByteBuffer non-rt, but I think its fine as we are in mutable land anyway here. Maybe call it unsafeFromBuffer ?

iravid: Marking and resetting is a no-op (or should be) so I wouldn’t say it makes this method non-RT.

iravid: Err that is unless mark was already used before.

vasilmkd: Yes, that's my point. Also, `reset()` doesn't remove the mark.

vasilmkd: How about:
```scala
val pos = buffer.position()
buffer.get(dest)
buffer.position(pos)
```
?

iravid: Yup! That’s better.

mschuwalow: +1. This is nice

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: Maybe we could add `fromArrayClone` to `fromArray`, but I feel like in most cases, the array passed in will be "write once". It's not like buffers which are often reused.

iravid: For arrays I would also prefer that we not copy (by default) and document the caveat.

core/shared/src/main/scala/zio/Chunk.scala
mschuwalow: > Marking and resetting is a no-op (or should be) so I wouldn’t say it makes this method non-RT.

One point is overwriting the mark. Another is that I was thinking of leaving this to the user completely. So I think we could change mark and position at will, and document that if desired the user has to call mark/reset

vasilmkd:
Fresh out of the oven.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1117,6 +1160,80 @@ object Chunk {
     override def toArray[A1 >: A](n: Int, dest: Array[A1]): Unit = { val _ = vector.copyToArray(dest, n, length) }
   }
 
+  private[zio] object Buff {
+    final def arrayBacked[A](buffer: Buffer): Option[Chunk[A]] =
+      if (!buffer.hasArray()) None
+      else {
+        val pos = buffer.position()
```
iravid: Is there an advantage to copying the array manually rather than using `get`? It appears that this method is redundant if not.

vasilmkd: And we are allocating an Option... Yeah, I guess we could get rid of it.

iravid: that option allocation was going to be my next remark ;)

iravid: Finally do you think inlining these methods into the `from*Buffer` methods would make sense?

vasilmkd: Doing that as we speak.

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: Are we ok with the current (new) implementation?

core/shared/src/main/scala/zio/Chunk.scala
vasilmkd: Love it, most of the code is gone.

vasilmkd:
Great collaboration everyone. Thank you so much.

PR #2350, opened 2019-11-27
Adding neotypes to the ecosystem libraries list

BalmungSan:
While neotypes itself is agnostic to your effect & stream type. It provides two modules that add direct support to **ZIO** and **ZStream**.


ghostdogpr:
Thanks!

PR #2354, opened 2019-11-28
Do Not Log Pure Interruptions

### LINKED ISSUE #2352: DefaultRuntime: Do not log pure interruptions
We should add a method on `Cause` to check to see if the cause is a "pure interruption". A pure interruption is a `Cause` formed solely through nodes that are neither `Die` nor `Fail` (although it may have any combination of `Both`, `Then`, `Trace`, `Meta`, and `Interrupt`).

A cause that is a pure interruption (even a complex one) should not, by default, be logged to standard error, in the `DefaultRuntime`.

Currently, while simple interruptions are not logged, more complex interruptions are logged.

`Cause` is defined in `Cause.scala` in `core`, and `DefaultRuntime` is defined in `DefaultRuntime.scala` in `core`.

adamgfraser:
Resolves #2352.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -111,6 +111,16 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
     find { case Interrupt(_) => () }.isDefined
 
   /**
+   * Determines if the `Cause` contains only interruptions and not any `Die` or
+   * `Fail` causes.
+   */
+  final def interruptedOnly: Boolean =
+    find {
+      case Die(_)  => false
+      case Fail(_) => false
+    }.getOrElse(true)
+
+  /**
    * Returns a set of interruptors, fibers that interrupted the fiber described
    * by this `Cause`.
    */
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -130,6 +130,16 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       case (_, Interrupt(_))    => false
     }
 
+  /**
+   * Determines if the `Cause` is a pure interruption, a `Cause` that does not
+   * contain any `Die` or `Fail` causes.
+   */
+  final def isPureInterruption: Boolean =
```
jdegoes: I might call this one, `justInterrupted` or even `interruptedOnly` for greater symmetry with `interrupted: Boolean`.

Otherwise, looks fantastic!

adamgfraser: Good idea. Updated now.

jdegoes:
Superb. Thank you for this!

PR #2360, opened 2019-11-28
Fix Type Safety Issue with ZIO#refineToOrDie

### LINKED ISSUE #2355: refineToOrDie does not constrain E1 witn E
```scala
final def refineToOrDie[E1: ClassTag](implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZIO[R, E1, A]
```
this signature does not constrain `E1` with `E`, letting unsound code like the one below compile
```scala
  ZIO
    .fail(new RuntimeException("BOO!"))
    .refineToOrDie[Error]
```

Naively this can be fixed with something like this
```scala
  implicit final class ZioOps[R, E, A](self: ZIO[R, E, A]) {
    final def refineToOrDie1[E1 <: E : ClassTag](implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZIO[R, E1, A] =
      self.refineOrDieWith { case e: E1 => e }(ev1)
  }
```
and it will result in a compile error
```
type arguments [Error] do not conform to method refineToOrDie1's type parameter bounds [E1 <: RuntimeException]
```

adamgfraser:
Resolves #2355.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1048,12 +1048,6 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self catchAll (err => (pf lift err).fold[ZIO[R, E1, A]](ZIO.die(f(err)))(ZIO.fail(_)))
 
   /**
-   * Keeps some of the errors, and terminates the fiber with the rest.
-   */
-  final def refineToOrDie[E1: ClassTag](implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZIO[R, E1, A] =
-    refineOrDieWith { case e: E1 => e }(ev1)
-
-  /**
    * Fail with the returned value if the `PartialFunction` matches, otherwise
    * continue with our held value.
    */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2644,6 +2638,15 @@ object ZIO extends ZIOFunctions {
     def toManaged: ZManaged[R, E, A] = ZManaged.fromAutoCloseable(io)
   }
 
+  implicit final class ZioRefineToOrDieOps[R, E <: Throwable, A](private val self: ZIO[R, E, A]) extends AnyVal {
+
+    /**
+     * Keeps some of the errors, and terminates the fiber with the rest.
+     */
+    final def refineToOrDie[E1 <: E: ClassTag](implicit ev: CanFail[E]): ZIO[R, E1, A] =
```
neko-kai: Is the implicit class necessary? Does
```scala
final def refineToOrDie[E1 <: E: ClassTag](implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZIO[R, E1, A] =
```
behave differently?

adamgfraser: That won't compile because the covariant `E` appears in contravariant position in the type bound, which is why we need the implicit class.

neko-kai: Got it!

PR #2361, opened 2019-11-28
Refactor Syntax Methods

### LINKED ISSUE #2356: ZIO Syntax: Refactor map2, map3, ..., to mapN
The zio.syntax package defines `map2`, `map3` syntax, etc., on tuples, which utilize sequential zip. It is not convenient for refactoring to have to change the method name from, e.g., `map2` to `map3` because one added another element in the tuple. So instead, we should call this method `mapN` so no matter how many elements the tuple has, the user calls the same method name.

### LINKED ISSUE #2357: ZIO Syntax: Add mapParN
Inside the `zio.syntax` package, for tuples up to some arity, we should add `(a, b, ..., z).mapParN(...)` for performing parallel zip operations.

adamgfraser:
Resolves #2356. Resolves #2357.

@jdegoes What do you think about moving these methods to the companion object instead so they are always available and the user doesn't have to rely on importing the implicit syntax?

Whatever we decide to do we should probably do it for other ZIO data types as well. I am thinking `ZIO`, `ZManaged`, `ZStream`, and `Gen`.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2334,6 +2334,75 @@ private[zio] trait ZIOFunctions extends Serializable {
     new ZIO.Lock(executor, zio)
 
   /**
+   * Sequentially zips the specified effects using the specified combiner
+   * function.
+   */
+  final def mapN[R, E, A, B, C](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B])(f: (A, B) => C): ZIO[R, E, C] =
+    zio1.zipWith(zio2)(f)
+
+  /**
+   * Sequentially zips the specified effects using the specified combiner
+   * function.
+   */
+  final def mapN[R, E, A, B, C, D](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B], zio3: ZIO[R, E, C])(
+    f: (A, B, C) => D
+  ): ZIO[R, E, D] =
+    for {
+      a <- zio1
+      b <- zio2
+      c <- zio3
+    } yield f(a, b, c)
+
+  /**
+   * Sequentially zips the specified effects using the specified combiner
+   * function.
+   */
+  final def mapN[R, E, A, B, C, D, F](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B], zio3: ZIO[R, E, C], zio4: ZIO[R, E, D])(
+    f: (A, B, C, D) => F
+  ): ZIO[R, E, F] =
+    for {
+      a <- zio1
+      b <- zio2
+      c <- zio3
+      d <- zio4
+    } yield f(a, b, c, d)
+
+  /**
+   * Returns an effect that executes the specified effects in parallel,
+   * combining their results with the specified `f` function. If any effect
+   * fails, then the other effects will be interrupted.
+   */
+  final def mapParN[R, E, A, B, C](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B])(f: (A, B) => C): ZIO[R, E, C] =
+    zio1.zipWithPar(zio2)(f)
+
+  /**
+   * Returns an effect that executes the specified effects in parallel,
+   * combining their results with the specified `f` function. If any effect
+   * fails, then the other effects will be interrupted.
+   */
+  final def mapParN[R, E, A, B, C, D](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B], zio3: ZIO[R, E, C])(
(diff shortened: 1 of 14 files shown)


neko-kai:
I think it's a good idea to move all syntax that can be picked up from the companion object into the companion object (i.e. everything except `EagerCreationSyntax` and `LazyCreationSyntax`)

adamgfraser:
Great! I will move in that direction.

jdegoes:
@adamgfraser Totally agreed!

adamgfraser:
Okay, I moved the syntax methods to the companion objects and created versions for `ZManaged`, `ZStream`, and `Gen`. For the single element types I called them `mapN` and `mapParN`. For the collection types I called them `crossN` and `zipN`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2333,6 +2333,75 @@ private[zio] trait ZIOFunctions extends Serializable {
   final def lock[R, E, A](executor: Executor)(zio: ZIO[R, E, A]): ZIO[R, E, A] =
     new ZIO.Lock(executor, zio)
 
+  /**
+   * Sequentially zips the specified effects using the specified combiner
+   * function.
+   */
+  final def mapN[R, E, A, B, C](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B])(f: (A, B) => C): ZIO[R, E, C] =
```
neko-kai: @adamgfraser 
Hmm, I thought the idea was to move the implicit classes on tuples to the companion object? That way they will be picked up without a wildcard import, because conversions are searched also for type parameters and ZIO is present as a type parameter of tuple/Iterable.

adamgfraser: @neko-kai Sorry I misunderstood you. We could definitely do that, or potentially both variants.

My thinking was that the problem we're trying to solve here is providing a way for users to combine multiple effects. I thought this was better provided through a concrete method on the companion object that doesn't require any implicits. If I have:

```scala
val zio1: ZIO[R, E, A] = ???
val zio2: ZIO[R, E, B] = ???
val zio3: ZIO[R, E, C] = ???
val zio4: ZIO[R, E, D] = ???
val f: (A, B, C, D) => E = ???
```

It seems slightly more natural to do:

```scala
ZIO.mapN(zio1, zio2, zio3, zio4)(f)
```

than:

```scala
(zio1, zio2, zio3, zio4).mapN(f)
```

Both of them are concise and read well but the first is probably more easily discoverable just being a normal method on an object instead of implicit syntax.

Where the new signature is not as convenient is where you already have a tuple of effects, because then you need to destructure the tuple to feed to `mapN`, but I'm not sure how common that is.

I'm happy to do either but that why I thought this was the better solution.

neko-kai: @adamgfraser 
I'm fine either way (or both!), just checking if that's intentional. Note though, that plain methods are not equivalent in cases with chaining or variables e.g. ```io.flatMap(_.mapN { ... }))``` – Scala has auto-tupling, but no auto-untupling and I believe `ZIO.mapN.tupled(_) { ... }` will probably not work due to direction of type inference.

jdegoes: I like `ZIO.mapN` more as it's obvious where to find it. It's not obvious we add methods to tuples. I've never really liked that style.

I do think `io.flatMap(_.mapN(...))` working only with the syntax is a good point, though. Although in practice it only occurs where an effect produces a tuple. The most obvious candidates for tuple production would be `zip` and `zipPar`, but if we added `ZIO.zipN`, and `ZIO.zipParN`, then suddenly tuple creation doesn't become common anymore, which lessens the argument for it.

🤷‍♂

streams/shared/src/main/scala/zio/stream/Stream.scala
```diff
@@ -59,6 +59,33 @@ object Stream extends Serializable {
   final def bracketExit[E, A](acquire: IO[E, A])(release: (A, Exit[Any, Any]) => UIO[Any]): Stream[E, A] =
     ZStream.bracketExit(acquire)(release)
 
+  /**
+   *  @see [[zio.ZStream.crossN]]
+   */
+  final def crossNN[E, A, B, C](stream1: Stream[E, A], stream2: Stream[E, B])(f: (A, B) => C): Stream[E, C] =
```
jdegoes: `crossNN`???

adamgfraser: Oops, fixed now.

PR #2364, opened 2019-11-28
TPromise documentation

drmarjanovic:
Document `TPromise` (part of #2122).

### DIFF
--- a/docs/datatypes/tpromise.md
+++ b/docs/datatypes/tpromise.md
@@ -42,7 +42,7 @@ val tPromiseFail: UIO[TPromise[String, Int]] = for {
 } yield tPromise
 ```
 
-Alternatively, you can use `Either[E, A]` in combining with a `done` function:
+Alternatively, you can use `done` combinator and complete the promise by passing it `Either[E, A]`:
 
 ```scala mdoc:silent
 import zio._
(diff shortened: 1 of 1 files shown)


docs/datatypes/tpromise.md
```diff
@@ -0,0 +1,90 @@
...
+import zio._
+import zio.stm._
+
+val tPromiseFail: UIO[TPromise[String, Int]] = for {
+  tPromise <- TPromise.make[String, Int].commit
+  _        <- tPromise.fail("failed").commit
+} yield tPromise
+```
+
+Alternatively, you can use `Either[E, A]` in combining with a `done` function:
```
mijicd: Replace with "Alternatively, you can use `done` combinator and complete the promise by passing it `Either[E, A]`"

docs/datatypes/tpromise.md
```diff
@@ -0,0 +1,90 @@
...
+  _        <- tPromise.done(Right(0)).commit
+} yield tPromise
+
+val tPromiseDoneFail: UIO[TPromise[String, Int]] = for {
+  tPromise <- TPromise.make[String, Int].commit
+  _        <- tPromise.done(Left("failed")).commit
+} yield tPromise
+```
+
+Note that a value can be set only once and every following attempt will return `false` as a result.
```
mijicd: Replace with "Once the value is set, following attempts to set it will result in `false`".

Issue #2334, opened 2019-11-26
Add fromEffectfulIterator to Stream


alexknvl:
```scala
  final def fromEffectfulIterator[R, A]
  (iterator: ZManaged[R, Throwable, Iterator[A]]
  ): StreamEffect[R, Throwable, A] =
    StreamEffect {
      iterator.flatMap { iterator =>
        Managed.fromEffect(ZIO.runtime).flatMap { rt =>
          Managed.effectTotal { () =>
            val hasNext: Boolean = try iterator.hasNext catch {
              case e: Throwable if !rt.platform.fatal(e) =>
                StreamEffect.fail(e)
            }

            if (hasNext) {
              try iterator.next() catch {
                case e: Throwable if !rt.platform.fatal(e) => 
                  StreamEffect.fail(e)
              }
            } else StreamEffect.end
          }
        }
      }
    }
```

Allows handling exceptions that arise during iterator's `hasNext` and `next` methods.

The existing method 
```
final def fromIteratorManaged[R, E, A](iterator: ZManaged[R, E, Iterator[A]]): ZStream[R, E, A] =
    StreamEffect.fromIterator(iterator)
```
returns `Stream[E, A]` for any `E`, which means it must die if the underlying iterator throws.

vasilmkd:
@alexknvl Technically, fixing the `Throwable` as the error type on the `ZManaged` denotes that the iterator itself has can fail to be produced with a `Throwable` error. It doesn't actually say anything about the error that can be thrown by pulling the iterator. However, I agree that this is confusing and I'm looking for ways to fix it. @iravid I'd like your opinion on how to proceed. This is related to #2234.

iravid:
@vasilmkd @alexknvl Looks like a much-needed addition. In fact, we should rename the existing `fromIteratorManaged` to `fromIteratorTotal`, and name this method `fromIterator`. Would be good to get this in before 1.0.

vasilmkd:
I will still advise to be cautious. There is a lot of nuance in the implementation, considering that we have to adapt it to support element level error handling.

alexknvl:
Adapted my original snippet to `RC18-2`:

```scala
package zio.stream

import zio.stream.StreamEffect.Failure
import zio.{Managed, ZIO, ZManaged}

object CsvExtra {
  final def fromEffectfulIterator[R, A]
  (iterator: ZManaged[R, Throwable, Iterator[A]]
  ): ZStream[R, Throwable, A] =
[...]

PR #2372, opened 2019-11-30
Eliminate all notions of scope errors from ZStream

### LINKED ISSUE #2234: ZStream scope error handling
Currently, `ZStream` is defined in terms of the `val process: ZManaged[R, E, Pull[R, E, A]]` method, which defines most streams as scoped effectful iterators.

There are some constructors that accept scoped resources, immediately try to enter the scope and build a stream inside it (`ZStream.unwrapManaged`, `ZStream.fromIteratorManaged`). Passing a `ZManaged.fail` to these constructors fails with an unhandled checked error.

There are also other constructors that don't assume anything and guard against failures that occur when entering the scope (`ZStream.fromEffect`, `ZStream.managed`). These constructors are safer, but currently produce a stream that will throw the scope entering failure when pulled. Honestly, I feel that this behavior is wrong, because the stream should not output the error, but rather, the stream should not even be constructed, because this error conceptually occurs before the stream even exists.

In that sense, the constructors that I mentioned first in this description are more correct, they don't allow the stream to be constructed.

Therefore, I would like to suggest the following. We should force the user to handle the scope errors before constructing the Stream, by requiring that any managed effect passed to a stream constructor has all checked errors handled (error type Nothing).

Further, this should be visible in the type signature of `val process: ZManaged[R, Nothing, Pull[R, E, A]]` and should be visible in the `managed`, `fromEffect`, `unwrapManaged` and other constructors as needed.

One caveat. This would probably change the signature of `bracket` to accept a non-failing `acquire` effect.

On the other hand, if this is not the direction that we wish to continue in, we should use this opportunity to define exactly how to unify these constructors and other stream combinators that accept a scoped pull somewhere in their implementation. This is very important for #1765.

@iravid

iravid:
Resolves #2234.

This changes the representation of `ZStream` from `ZManaged[R, E, Pull[R, E, A]]` to `ZManaged[R, Nothing, Pull[R, E, A]]`. It turns out that we don't actually need to represent failures on the scope level; the only thing we're using scopes for is finalizer management.

Incidentally, this also makes a bunch of operations lazier as they don't eagerly allocate ZManaged values but rather wait for them to be pulled.

With this out of the way, we can properly eliminate errors from a stream by handling element-level errors.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -20,7 +20,7 @@ import java.io.{ IOException, InputStream }
 
 import zio._
 
-private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, E, () => A])
+private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, Nothing, () => A])
     extends ZStream[R, E, A](
       ZStream.Structure.Iterator(
         processEffect.map { thunk =>
@@ -323,7 +323,7 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def apply[R, E, A](pull: ZManaged[R, E, () => A]): StreamEffect[R, E, A] =
+  final def apply[R, E, A](pull: ZManaged[R, Nothing, () => A]): StreamEffect[R, E, A] =
     new StreamEffect(pull)
 
   final def fail[E](e: E): StreamEffect[Any, E, Nothing] =
@@ -365,12 +365,10 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def fromIterator[R, E, A](iterator: ZManaged[R, E, Iterator[A]]): StreamEffect[R, E, A] =
+  final def fromIterator[R, E, A](iterator: Iterator[A]): StreamEffect[R, E, A] =
     StreamEffect {
-      iterator.flatMap { iterator =>
-        Managed.effectTotal { () =>
-          if (iterator.hasNext) iterator.next() else end
-        }
+      Managed.effectTotal { () =>
+        if (iterator.hasNext) iterator.next() else end
       }
     }
 
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -365,7 +365,7 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def fromIterator[R, E, A](iterator: Iterator[A]): StreamEffect[R, E, A] =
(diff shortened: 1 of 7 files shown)


iravid:
Big thanks to @vasilmkd for pushing us to think this through :-)

iravid:
Also - @adamgfraser - the CanFail was super useful when refactoring this!

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -365,12 +365,10 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def fromIterator[R, E, A](iterator: ZManaged[R, E, Iterator[A]]): StreamEffect[R, E, A] =
+  final def fromIterator[R, E, A](iterator: Iterator[A]): StreamEffect[R, E, A] =
```
vasilmkd: How can we make this to work for #2334?

iravid: I think we can't do this on StreamEffect right now, as it requires `unsafeRun`.

vasilmkd:
Love to see team effort and orthogonal solutions coming together.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: Ok, then let's simplify the function signature. `R = Any, E = Nothing`?

iravid: Oh yeah. Good catch.

vasilmkd:
Also, it's nice to see my essays (like #2234) actually realized in code. :laughing:

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -365,12 +365,10 @@ private[stream] object StreamEffect extends Serializable {
       }
     }
 
-  final def fromIterator[R, E, A](iterator: ZManaged[R, E, Iterator[A]]): StreamEffect[R, E, A] =
+  final def fromIterator[A](iterator: Iterator[A]): StreamEffect[Any, Nothing, A] =
```
vasilmkd: Can we implement `StreamEffect.fromIterable` in terms of this (by calling `Iterable#iterator`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -581,7 +581,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   final def aggregateManaged[R1 <: R, E1 >: E, A1 >: A, B](
     managedSink: ZManaged[R1, E1, ZSink[R1, E1, A1, A1, B]]
   ): ZStream[R1, E1, B] =
-    ZStream[R1, E1, B] {
```
vasilmkd: I know it's nitpicking but I really dislike these type hints...

PR #2374, opened 2019-12-01
Explore Lens Encoding For Environment Type

adamgfraser:
This doesn't feel quite right yet but it provides the basic functionality and can serve as a basis for discussion.


core/jvm/src/main/scala/zio/prototype.scala
```diff
@@ -0,0 +1,130 @@
...
+          new LoggingEffect.Live {
+            val environment: Ref[ZEnv with Logging] = ref
+          }
+        }
+      }
+
+    def log(line: String): ZIO[LoggingEffect[ZEnv with Logging], Nothing, Unit] =
+      ZIO.accessM[LoggingEffect[ZEnv with Logging]](service => service.useM(service.logging)(_.log(line)))
+
+    def addPrefix(prefix: String): ZIO[LoggingEffect[ZEnv with Logging], Nothing, Unit] =
```
jdegoes: I'd make this polymorphic: `addPrefix[R <: Logging]` to show that `addPrefix` doesn't know what the environment type is.

adamgfraser: Yes, done now.

core/jvm/src/main/scala/zio/prototype.scala
```diff
@@ -0,0 +1,130 @@
...
+object Example extends App {
+  import prototype._
+
+  def run(args: List[String]): ZIO[ZEnv, Nothing, Int] =
+    myAppLogic.provideSomeM(LoggingEffect.make).as(0)
+
+  val myAppLogic =
+    for {
+      _ <- LoggingEffect.log("All clear.")
+      _ <- LoggingEffect.addPrefix[ZEnv with Logging]("Warning: ")
```
adamgfraser: Type inference not great here because the `Lens` and thus `EnvironmentalEffect` are invariant.

jdegoes:
It seems like this solves all problems but, when pushed to its logical conclusion, would require changing environments to things like:

```scala
Console[R] with Logging[R] with Clock[R]
```

And making all user code polymorphic in `R` (no more simple value effects).

Going back to the original sketch:

```scala
  // All environmental effects extend this trait:
  trait EnvironmentalEffect {
    type Environment
    def modify[A, B](lens: Lens[Environment, A])(f: A => (B, A)): UIO[B]

    final def update[A](lens: Lens[Environment, A])(f: A => A): UIO[Unit] = 
      modify(lens)(a => ((), f(a)))
  }
  object EnvironmentalEffect {
    def modify[R <: EnvironmentalEffect, A, B](lens: Lens[R#Environment, A])(f: A => (B, A)): ZIO[R, Nothing, B] = ZIO.accessM[R](_.modify(lens)(f))

    def modify[R <: EnvironmentalEffect, A, B](lens: Lens[R#Environment, A])(f: A => (B, A)): ZIO[R, Nothing, B] = 
      ZIO.accessM[R](_.modify(lens)(f))

    def update[R <: EnvironmentalEffect, A](lens: Lens[R#Environment, A])(f: A => A): ZIO[R, Nothing, Unit] = 
      ZIO.accessM[R](_.update(lens)(f))
  }
  
  trait Logging extends EnvironmentalEffect {
    def logging: Lens[Environment, Logging.Service]
  }
  object Logging {
    trait Service {
      def log(line: String): UIO[Unit]
    }
  }
```

The use of (unsound) type projection occurs only in the helpers. It occurs to me we don't actually need the helpers. If we call `accessM`, we get a concrete `R`, which is a stable identifier. We should be able to, from that `R`, modify the environment directly without knowing the type of `R`.

Unless I'm missing something.

core/jvm/src/main/scala/zio/prototype.scala
adamgfraser: Making the environment type existential eliminates the type inference problem here.

adamgfraser:
I think this is looking pretty good now. There is a little bit of work to do on naming because now we have three things:

1. The actual service that knows how to log a line (`Logging.Service`)
2. The module that includes the actual service (`Logging`)
3. The lens that knows how to zoom in from an unknown environment type to the logging service (`LoggingEffect`).

I think we want to rename the lens to `Logging` because that is what users are going to interact with in most method signatures. Then the module needs to be renamed to something else, maybe `LoggingModule` or `HasLogging`.

We will also need to work on some good examples and documentation for this because it is giving us a lot of additional power, some of which I don't think we are really exploring yet, but it is adding another layer of indirection to an area that people already struggle with. Hopefully if we do it right we can actually make a lot of use cases simpler though.

adamgfraser:
Simplified and refined based on work with @jdegoes to only use two types like today and limit the power to modify the environment so all changes are locally scoped. We should see where we are with the `HSet` based encoding and then make a decision on how we want to proceed here.

adamgfraser:
Shall we come back to this and how we want to move forward on making it easier to work with the environment? Is there more work we should do to explore the HSet encoding?

adamgfraser:
Closing in favor of #2594.

PR #2377, opened 2019-12-01
Compile Scala Native in CI

adamgfraser:
We recently added Scala Native support. We're still working through some issues on how to enable ZIO Test for Scala Native but until then we should at least compile Core on Scala Native to prevent regressions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,7 +321,7 @@ jobs:
       - <<: *jdk_8
 
 workflows:
-  version: 3
+  version: 2
   build:
     jobs:
       - lint:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -303,7 +321,7 @@ jobs:
       - <<: *jdk_8
 
 workflows:
-  version: 2
+  version: 3
```
ghostdogpr: Shouldn't it be 2? https://circleci.com/docs/2.0/configuration-reference/#version-1

It works though.

adamgfraser: You're right. Fixed now.

softinio:
Nice work!

Issue #2295, opened 2019-11-20
Should ZIO#fromEither be By Name?


adamgfraser:
Currently `ZIO.fromEither` takes its parameter by name and wraps it in `effectTotal`. This is to a certain extent safer but I would have expected `fromEither` to just life a pure `Either` value into a `ZIO`. This approach would also be more compositional because I could always wrap the `fromEither` call in `effectSuspendTotal` if I want the lazy behavior. Should we change this? The same issues also applies to other constructors such as `fromOption` and `fromTry`.

neko-kai:
Previous discussion (when it was changed from strict to by-name): https://github.com/zio/zio/pull/577
^Note: I think the above was alone before tracing, there's an added aspect now that a lambda wrapper brings with it a line number

jdegoes:
The laziness buys us the ability to "push" the errors into the ZIO runtime system.

Otherwise users write code like this:

```scala
def foo = {
  val zio = ZIO.fromEither(mySideEffectingMethodCall)
  ...
}
```

and now `foo` blows up, rather than `foo` returning a `ZIO`. 

ZIO has good mechanisms for error reporting and recovery so I think it makes sense to push as many failures as possible into ZIO.

reibitto:
Just to add on to this, should the same principle apply for things like `someOrFail(...)` and so on? I'm not exactly sure where the line should be drawn and what would be deemed "consistent" when it comes to all these methods. Right now I don't have an intuition for which ones are/should be call-by-name and which shouldn't.

I actually ran into `someOrFail` that had error logging with untracked side-effects contained in it yesterday (still in the middle of a conversion to ZIO at work). I scratched my head for a while wondering how my `Some` was "failing" (it wasn't). That was totally a user error, but it did make me wonder about this issue.

neko-kai:
@adamgfraser 
> Also, does it make sense to have something like succeedEither?

I like it! Maybe with a name like `fromEitherValue` or `fromEitherTotal`?

PR #2378, opened 2019-12-01
Make error arguments call-by-name.

vvlevykin:
Make `collect`, `collectM`, `leftOrFail`, `rightOrFail`, `someOrFail` error argument in ZIO consistent with `filterOrFail`.
```scala
final def filterOrFail[E1 >: E](p: A => Boolean)(e: => E1): ZIO[R, E1, A]
```
Same for `asError`, `collect`, `collectM` in ZManaged.

Similar to #1905.

Also the creation of the error (e.g. Exception) may be somewhat expensive.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -263,14 +263,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * Fail with `e` if the supplied `PartialFunction` does not match, otherwise
    * succeed with the returned value.
    */
-  final def collect[E1 >: E, B](e: E1)(pf: PartialFunction[A, B]): ZIO[R, E1, B] =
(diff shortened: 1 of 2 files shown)


reibitto:
Just for reference, I also asked about this here: https://github.com/zio/zio/issues/2295#issuecomment-556985105

I feel like it would be overall more consistent, but let's see what others have to say.

jdegoes:
@vvlevykin I like this and think we should tend toward more laziness to keep more defects in ZIO (as well as, of course, improve efficiency when exceptions are not used).

vvlevykin:
@neko-kai I see #1910, do I need to change something for tracing here?

neko-kai:
@vvlevykin 
You may write `ZManaged.asError` similarly to `ZIO.asError` to propagate the lambda line number:
```scala
  final def asError[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
    mapError(new ZIO.ConstFn(() => e1))
```

PR #2379, opened 2019-12-01
Add ZStream.groupedWithin

### LINKED ISSUE #2336: Add ZStream#groupedWithin
Signature:
```
def groupedWithin(n: Int, d: Duration): ZStream[R, E, List[A]]
```

The function should delegate to `aggregateAsyncWithin` with a `Sink.collectAllN` and a `Schedule.spaced`.

reibitto:
Resolves #2336

The test based on time passing was a bit finicky. I feel like there should be a better way to do this. I made a TestClock isolated to the test because otherwise it looks like the other tests were interfering with it (running the single test was passing, but not when other tests ran alongside it). Either that, or it's just general flakiness? I'm not sure, but I'll explore alternatives.

Or is there a different strategy I could use to test that part?


adamgfraser:
There shouldn't be any difference between using the `TestClock` in the environment and creating your own in this case. Every test gets provided with its own copy of the `TestClock` so there shouldn't be any interference from other tests. If you are using `nonFlaky` there is some additional work for now in resetting the `TestClock` but that isn't applicable here.

I think there may be some flakiness in the test itself. In particular, I note that `<&` zips the two streams together which guarantees that one pair of effects will be evaluated before the next pair but doesn't guarantee ordering of evaluating the left and right side effects. So for example with one stream doing `Emit` and the other doing `Adjust`, the following are all value orderings of the first four effects:

```scala
List(Adjust, Emit, Adjust, Emit)
List(Adjust, Emit, Emit, Adjust)
List(Emit, Adjust, Emit, Adjust)
List(Emit, Adjust, Adjust, Emit)
```

I tried adding taps to each of the streams using the `TestClock` in the environment and got the below outputs in a failing case. So it appears that the `TestClock` is being adjusted appropriately and you can see my point about the ordering of effects from the two streams.

```scala
Emit(1)
Adjust(1000000000)
Emit(2)
Adjust(2000000000)
Emit(3)
Adjust(3000000000)
Emit(4)
Adjust(4000000000)
Adjust(5000000000)
Emit(List(1, 2, 3))
Emit(1)
Emit(2)
Adjust(6000000000)
Emit(List(4, 1, 2))
Emit(3)
Adjust(7000000000)
Emit(4)
Adjust(8000000000)
Adjust(9000000000)
Emit(List(3))
Emit(List(4))
```

reibitto:
Thanks @adamgfraser. That was helpful. I see what you mean about the order issue when using `zip`. I'm able to duplicate that.

I changed the test to adjust using `tap` instead, but I'm seeing similar issues.

Running this test individually works for me:
`testOnly *.StreamSpec -- -t "group based on time passed"`

When run together with other tests, that same passing test starts becoming flaky:
`testOnly *.StreamSpec -- -t "Stream.groupedWithin"`

The same is true for `group immediately when chunk size is reached`.

I need to think about this. I'm not sure what's making the test flaky.

And perhaps unrelated (still need to debug), but I added a new test (`group before chunk size is reached due to time window`) which is broken 100% of the time. I can't make sense of that one. The time window is smaller than the spacing of the elements in the stream. I don't see how more than 1 element could be grouped together, but that's exactly what I'm seeing. Maybe I'm misunderstanding how `aggregateAsyncWithin` works. The implementation is pretty complex.

streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala
```diff
@@ -1090,10 +1091,28 @@ object StreamSpec extends ZIOBaseSpec {
     ),
     testM("Stream.grouped")(
       assertM(
-        Stream(1, 2, 3, 4).grouped(2).run(ZSink.collectAll[List[Int]]),
+        Stream(1, 2, 3, 4).grouped(2).runCollect,
```
mschuwalow: nice cleanup!

reibitto:
Thanks @mschuwalow. I'm still struggling with the tests on this one. I must be misunderstanding something. I've tried various things, but here's one attempt:

```scala
testM("group before chunk size is reached due to time window") {
  val zio = for {
    queue  <- Queue.unbounded[Unit]
    stream = ZStream.fromIterable(1 to 3).tap(_ => TestClock.adjust(1.second) *> queue.take)
    result <- stream
               .groupedWithin(10, 10.millis)
               .tap(_ => queue.offer(()))
               .runCollect
  } yield result

  assertM(zio, equalTo(List(List(1), List(2), List(3))))
}
```

This just hangs though. I might be putting the `queue.take` in the wrong place. What I'm trying might not even make sense. Also, I'm thinking this might need a `ZIO.sleep` somewhere.

reibitto:
It makes sense why it hangs. I'm just not sure what to do about it. This might need a `fork` call.

reibitto:
@mschuwalow Oops, I completely forgot about this. I could take another look, but to be honest I still don't have any new ideas on how to test this correctly. So yeah, I'd appreciate the help if you don't mind.

I should at least merge in master and fix these conflicts. One sec, I'll do that now.

mschuwalow:
> @mschuwalow Oops, I completely forgot about this. I could take another look, but to be honest I still don't have any new ideas on how to test this correctly. So yeah, I'd appreciate the help if you don't mind.
> 
> I should at least merge in master and fix these conflicts. One sec, I'll do that now.

please do. I'll take a look this evening

reibitto:
Alright, done. Thanks!

jdegoes:
@reibitto Any interest in reviving this?

@iravid Any feedback on the design of this combinator?

jdegoes:
@reibitto Please re-open when conflicts are fixed if you still want to push ahead with this, and thanks for your effort in developing this combinator!

reibitto:
@jdegoes It's already in. With the help of @mschuwalow who fixed the tests (which was the hard part).

jdegoes:
@reibitto Thanks!

PR #2380, opened 2019-12-01
Overload foreach for Option

vvlevykin:
Its type is:
```scala
def foreach[R, E, A, B](in: Option[A])(f: A => ZIO[R, E, B]): ZIO[R, E, Option[B]]
```
Sometimes `IO.fromOption` is not convenient if flatMap after it might fail. Using `fold` directly is also clunky.
If the result is not needed, the existing `foreach_` can be used over the Option.


adamgfraser:
Isn't this just `ZIO.foreach(in)(f).map(_.headOption)` or `ZIO.foreach(in)(f).head.option`? Maybe we add `ZIO.headOption` so you can do `ZIO.foreach(in)(f).headOption`. I'm a little worried this is too specific and too easily implemented in terms of existing combinators.

vvlevykin:
You are right, it's specific. I use `.map(_.headOption)` mostly or `traverse` from cats with interop for this.

Maybe it's better if there are more convenient ways to work with Option in zio. I don't have a strong opinion on this and I'm ok with this PR being closed.

regiskuckaertz:
Perhaps we could change `foreach` to take an `IterableOnce` instead. We don't need more than one traversal.

adamgfraser:
@regiskuckaertz `ZIO.foreach` already works with an `Option` with the current type signature. I think the potential issue is just that when you do `ZIO.foreach(option)(f)` you get back a `List` whereas often you would like to get back an `Option`

regiskuckaertz:
That's because of an implicit conversion from `Option` to `Iterable`; that's not a pattern we should encourage imo. It's worth noting the same choice was made in the [standard library](https://www.scala-lang.org/api/current/scala/Option.html):

> Many of the methods in here are duplicative with those in the Traversable hierarchy, but they are duplicated for a reason: the implicit conversion tends to leave one with an Iterable in situations where one could have retained an Option

I'm in favour of this change, however I would just call it `foreach`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2103,6 +2103,13 @@ private[zio] trait ZIOFunctions extends Serializable {
       loop
     }
 
+  /**
+   * Applies the function `f` if the argument is non-empty and
+   * returns the results in a new `Option[B]`.
+   */
+  final def foreachOption[R, E, A, B](in: Option[A])(f: A => ZIO[R, E, B]): ZIO[R, E, Option[B]] =
```
neko-kai: I guess this also needs the discarding variants, `foreach/traverseOption_`

neko-kai: Actually, there's no difference between a `foreachOption_` and `foreach`, I guess a discarding variant is redundant in anything but naming consistency.

neko-kai:
@regiskuckaertz 
> (Also, IterableOnce is only in 2.13, I didn't see that initially)

In 2.12 it's called `TraversableOnce`

> I'm in favour of this change, however I would just call it `foreach`.

I am *slightly* in favor of overloading `foreach` too, simply because then we wouldn't need `foreachOption_` & `traverseOption_` discarding variants – and not having them would be somewhat inconsistent

adamgfraser:
I worry about the proliferation of methods here and I'm not sure what we're getting.

I agree in principle that we shouldn't be relying on implicit conversions but treating an `Option` as an `Iterable` is perfectly valid, the conversion is always in scope, and it is kind of the Scala standard library's way of dealing with the fact that they don't provide some kind of type class to abstract over various "collection" types.

Beyond this all we're saving is the additional method call to transform the resulting `List` into an `Option`, which is the same thing we would do if we wanted any other collection type like a `Vector`.

neko-kai:
@adamgfraser 
IMHO Option is a bit of a special case. IME nearly every time i reach for `cats` `.traverse` syntax instead of IO.traverse, it's because of Option

jdegoes:
I don't think `foreach` overloads are a bad way to solve this problem. I think we have three useful cases:

 - `Option`, because it promises to construct exactly one, although it's true we can get that with `_.headOption`. 
 - `List`, because, well, that's the standard collection type in FP
 - `Chunk`, as a better array / vector (even though right now it doesn't have better append / prepend, but that can be fixed one day)

With `Option` or other error types, however, we have another choice: to push the `Option` into the error, and then do the `foreach`, and then "unpush" the option. 

I think we already have methods to do this. They're not very obvious though so it's not necessarily an argument against `foreach`.

The point about method proliferation is quite valid. Though if we can keep people from ever having to reach for a heavyweight `F.traverse`, maybe it's worth it?

mschuwalow:
What about providing our own typeclass? It doesn't have to do much and I think would be a nicer option than overloading

regiskuckaertz:
Looks like we are edging towards three specialisations then? @vvlevykin are you happy to pick this up?

vvlevykin:
Yes, I can pick this up. Correct me if I misunderstood: I'm going to add overloads of `foreach` for `Option` and `Chunk` and overloads of `foreach_`, `foreachPar`, `foreachPar_`, `foreachParN`, `foreachParN_` for `Chunk`.

regiskuckaertz:
Yeah, I think there’s a case for treating `Chunk` as a first class citizen since its raison d’être is, well, this library 😁

vvlevykin:
What would be the best way to implement `*Par*` methods? I think that Chunk is missing `iterator`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2115,6 +2140,39 @@ private[zio] trait ZIOFunctions extends Serializable {
...
 
+  /**
+   * Applies the function `f` to each element of the `Chunk[A]` in parallel,
+   * and returns the results in a new `Chunk[B]`.
+   *
+   * For a sequential version of this method, see `foreach`.
+   */
+  final def foreachPar[R, E, A, B](as: Chunk[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, Chunk[B]] = {
+    val len                        = as.length
+    var array: ZIO[R, E, Array[B]] = IO.succeed(null.asInstanceOf[Array[B]])
```
regiskuckaertz: Mixing mutation and IO here is bound to create unpredictable behaviour, how about matching the list version with something like:
```scala
as.foldRight(succeed(Chunk.empty))((a, bs) => fn(a).zipWithPar(bs)(_ +: _)).refailWithTrace
```
With the missing `+:`
```scala
def +:[A1 >: A](a: A1): Chunk[A1] = Chunk.single(a) ++ self
```

vvlevykin: This function matches mapM: https://github.com/zio/zio/blob/6e1960e9223b513cdd86ef5fce69b64c878cbbe3/core/shared/src/main/scala/zio/Chunk.scala#L542

regiskuckaertz: 😮 It’s baffling that scope is retained as such. My bad, but still I’d suggest either to add an `mapMPar` or use the fold.

jdegoes:
I wouldn't be opposed to introducing our own tiny type class(s) for this purpose. It looks like we have the following cases:

 - Scala Collections
 - Chunk
 - Option
 - Non-Empty List (`::`)

I think `Option` can be unified with Scala collections. Leaving three classes of "specializations". That's not terrible. It's at least more manageable.

The old `CanBuildFrom` would have worked for pre 2.13 (for collections), but not after the redesign.

regiskuckaertz:
There's a bit of duplication for sure, but it doesn't feel like this is going to be a maintenance burden, does it? One can't know but it seems unlikely these functions will change (famous last words).

jdegoes:
@vvlevykin Thank you for your work on this—it looks good to merge! 

Can your resolve the conflicts so we can get it merged in?

vvlevykin:
@jdegoes I've missed the notification, sorry. I've resolved the conflicts now.

adamgfraser:
@vvlevykin Can you format by running `sbt prepare`?

adamgfraser:
@vvlevykin It looks like with the new method signatures some of the Scaladoc links are ambiguous. You can fix them by adding enough of the method signature to disambiguate them. See #1171 for an example.

adamgfraser:
@vvlevykin Thank you for your work on this!

PR #2389, opened 2019-12-02
ZIO Test: Fix Colors for Multiline Strings in SBT Test Runner

adamgfraser:
Resolves #2363 

This issue is a little tricky. The problem is that when the SBT test logger displays a string, it prepends each line with a potentially colored string like `[info]` or `[warning]` and it resets the console color after that, which is why we lose the coloring on subsequent lines of multiline strings. The SBT test logger implementation is outside of our control, so to address we need to prepend the ANSI escape code for the correct color to every line of multiline strings so we restore the correct color.


adamgfraser:
Good idea! Will add.

PR #2391, opened 2019-12-02
Add ZStream#via

### LINKED ISSUE #2370: ZIO Stream: Consider adding `.via` operator
Other streaming libraries has the concept of a `Flow`, that allows you to map one stream to the other in a single chain, like:
```
source.via(flow).to(sink)
```

It would be nice to have `.via` operator in ZIO Streams that allows mapping `ZStream[A] => ZStream[B]`.

Without that, the flow of operations looks not so nice:
```
mapStream(
  stream
    .map(...)
    .flatMap(...)
).aggregateAsyncWithin(...)
```
but it could be something like:
```
stream
  .map(...)
  .flatMap(...)
  .via(mapStream)
  .aggregateAsyncWithin(...)
```

vasilmkd:
Resolves #2370.

@iravid @gchudnov Am I being too restrictive with `self.type`?

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamChunkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamChunkSpec.scala
@@ -317,6 +317,16 @@ object StreamChunkSpec extends ZIOBaseSpec {
         }
       }
     },
+    suite("StreamChunk.via")(
+      testM("happy path") {
+        val s = StreamChunk.fromChunks(Chunk(1), Chunk.empty, Chunk(2, 3, 4), Chunk(5, 6))
+        s.via(_.map(_.toString)).runCollect.map(assert(_, equalTo(List("1", "2", "3", "4", "5", "6"))))
+      },
+      testM("introduce error") {
+        val s = StreamChunk.fromChunks(Chunk(1), Chunk.empty, Chunk(2, 3, 4), Chunk(5, 6))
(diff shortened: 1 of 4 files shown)


iravid:
Yes, `self.type` is unnecessarily narrow.

vasilmkd:
Changed to `ZStream[R, E, A]` and `ZStreamChunk[R, E, A]` respectively.

gchudnov:
wow, thank you so much!

neko-kai:
@vasilmkd @iravid @gchudnov 
`.via` is the exact same thing as `.pipe` from the 2.13+ standard library - https://github.com/scala/scala/blob/841f45f8d4a29fd593699654973d12a62b087659/src/library/scala/util/ChainingOps.scala#L62

I feel like this can lead to bikeshedding between people whether to use `.via` or the more polymorphic `.pipe`. What do you think about replacing `.via` by a copy of `.pipe` instead? 

```scala
def pipe[B](f: ZStream[R, E, A] => B): B = f(self)
```

This will only make the `import scala.util.chaining._` redundant when working with streams, but would otherwise make the styles indistinguishable.

vasilmkd:
@neko-kai I was completely unaware of this. I agree with you. For Scala 2.13 this will hide the implicit right?

gchudnov:
agree with @vasilmkd 
`pipe` is more generic, you can pipe to anything, but with `via` you're still living within a stream.

IMHO with `pipe` it is easier to make a mistake a the compiler won't help you.

PR #2397, opened 2019-12-03
ZIO Test: Provide Combinators to Use Live Environment by Default

adamgfraser:
One of the pain points that I hear about regarding ZIO Test is people who want to use live implementations of the environment types by default.

I think this shows up in two ways.

One is that the `TestClock` in particular can make for a more difficult migration from existing test frameworks. We generally have a version of everything that exists in other testing frameworks so most of migration is just relatively mechanically replacing code with the ZIO Test version of things. But writing tests involving time with the `TestClock` is significantly different and while ultimately more powerful can require some thinking, before which your tests are going to be failing which is not very fun.

The second is there is a subset of users that just don't want one or more of the test implementations (e.g. you don't have any `Console` effects you want to test but you use a lot of them to print debug or information messages that are now no longer showing up, or you interact with a lot of third party services that need live clock effects and think that is a better default for you).

This PR addresses this by providing a set of combinators on `TestEnvironment` that replace one of the test implementations with a version of the live one that just ignores any "test" effects. For example, you can call `testEnvironment.withLiveClock` and get a `TestEnvironment` that does live sleeps.

I have somewhat mixed feelings about this because I think the `TestEnvironment` and the test implementations are great and want to make them even better, but I think if we can offer a relatively easy "escape hatch" like this we can both ease migration and provide a better for users who like everything else about ZIO Test but want to do something different here.

Other opinions very much appreciated.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/environment/EnvironmentSpec.scala
@@ -106,6 +106,29 @@ object EnvironmentSpec extends ZIOBaseSpec {
                    .env("k1")
                    .provide(testEnvironment.mapTestSystem(_ => testSystem))
       } yield assert(result, isNone)
-    }
+    },
+    testM("withLiveClock maps the `TestClock` implementation to one that uses the live environment") {
+      for {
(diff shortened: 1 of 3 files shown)


TimPigden:
awesome, thanks!

PR #2398, opened 2019-12-03
Remove Constructors Masquerading As Class Methods

### LINKED ISSUE #2388: `ZIO#succeed` (the one on the class) is redundant and confusing
`putStrLn("foo").succeed(0)` is equivalent to `ZIO.succeed(0)`.

The signature is almost identical to `as`, except for strictness of the argument.

As per the discord discussion on zio-dev, it should be removed. I'll submit a PR within a couple days.

adamgfraser:
Resolves #2388.


jdegoes:
Yikes, how did these get in? Bad merge maybe

kubukoz:
Thanks!

PR #2399, opened 2019-12-03
Fix Bracket Type Inference Issue

### LINKED ISSUE #2394: Type Inference Regression on ZIO.bracket between RC9 and RC17
Just recently got around to updating code that was originally written against zio-core RC9 to the latest (RC17). It appears that the type inference on `ZIO.bracket` has regressed. It's not hugely impactful for me, but I figured I'd open up an issue in case this was unintentional.

Consider the usecase of a Scalatest unit test that runs some blocking process that needs to be alive for the lifetime of the tests, and then released (for example, a mock DB implementation). On ZIO RC9, this could be expressed pretty cleanly using bracket to state that a resource (the DB) should be acquired, then used in the context of running the tests, and finally have the fiber that is running it be interrupted to shut it down:

```
class ExampleSpec extends FlatSpec with DefaultRuntime {

  override def run(testName: Option[String], args: Args) = {
    import zio.blocking._

    // environment isn't really important for this example, but it is important that it be specific
    val runDb: ZIO[Blocking with system.System, Throwable, Unit] = // some code that runs the process until it is interrupted

  // fork the process, run the test, interrupt fiber upon completion to kill the process
  val runTest = runDb.fork.bracket(_.interrupt) { _ =>
    ZIO.effectTotal(super.run(testName, args))
  }

  // write unit tests as per normal
}
```

This code worked fine on RC9. However, on RC17 (and presumably some earlier releases as well), it fails to typecheck due to the inferred type of `bracket` - specifically, because `Fiber.interrupt` is returning `UIO`, the environment gets inferred to `Any`, violating the constraint on `bracket` that `R1 <: R`.

```
error: inferred type arguments [Any,Nothing,Any] do not conform to method bracket's type parameter bounds [R1 <: zio.system.System with zio.blocking.Blocking,E1,R1 <: R1]
     val runTest = runDb.fork.bracket(_.interrupt) { _ =>
                              ^
```

Note that this failure is happening for me on 2.12.9 - I haven't had the opportunity to test on a later version of Scala yet due to a separate scala compiler bug preventing me from updating the compiler.

Either explicitly declaring the inferred type, i.e.:

```
runDb.fork.bracket[ZEnv, Status](_.interrupt) { _ =>
  ZIO.effectTotal(super.run(testName, args))
}
```

or using the uncurried version:

```
runDb.fork.bracket(_.interrupt, { _ =>
  ZIO.effectTotal(super.run(testName, args))
})
```

both work, so again, this isn't super pressing - but it seems like it may have been unintentional (in general, documentation throughout ZIO states that curried versions of many of these operations are preferable because they have better type inference, yet here they are worse), so I wanted to raise an issue just in case

adamgfraser:
Resolves #2394. We should review the other bracket related combinators. On initial inspection it appears that several of them have similar issues.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -126,7 +126,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Shorthand for the curried version of `ZIO.bracket`.
    */
-  final def bracket[R1 <: R, E1 >: E]: ZIO.BracketAcquire[R1, E1, A] = ZIO.bracket(self)
+  final def bracket: ZIO.BracketAcquire[R, E, A] = ZIO.bracket(self)
 
   /**
    * A less powerful variant of `bracket` where the resource acquired by this
@@ -2745,8 +2745,8 @@ object ZIO extends ZIOFunctions {
   }
 
   final class BracketAcquire[-R, +E, +A](private val acquire: ZIO[R, E, A]) extends AnyVal {
-    def apply[R1 <: R](release: A => URIO[R1, Any]): BracketRelease[R1, E, A] =
-      new BracketRelease[R1, E, A](acquire, release)
+    def apply[R1](release: A => URIO[R1, Any]): BracketRelease[R with R1, E, A] =
+      new BracketRelease[R with R1, E, A](acquire, release)
   }
   final class BracketRelease[-R, +E, +A](acquire: ZIO[R, E, A], release: A => URIO[R, Any]) {
     def apply[R1 <: R, E1 >: E, B](use: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2745,8 +2745,8 @@ object ZIO extends ZIOFunctions {
   }
 
   final class BracketAcquire[-R, +E, +A](private val acquire: ZIO[R, E, A]) extends AnyVal {
-    def apply[R1 <: R](release: A => URIO[R1, Any]): BracketRelease[R1, E, A] =
-      new BracketRelease[R1, E, A](acquire, release)
+    def apply[R1](release: A => URIO[R1, Any]): BracketRelease[R with R1, E, A] =
+      new BracketRelease[R with R1, E, A](acquire, release)
```
jdegoes: Clever. Could be more widely used

PR #2409, opened 2019-12-03
Fix Type Inference Issue With ZManaged#Make

### LINKED ISSUE #2407: ZManaged: Fix ZManaged.make
`ZManaged.make` does not infer well when the environment type of the "acquire" is different than the environment type of "release".

adamgfraser:
Resolves #2407.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1226,7 +1226,7 @@ object ZManaged {
    * Lifts a `ZIO[R, E, A]` into `ZManaged[R, E, A]` with a release action.
    * The acquire and release actions will be performed uninterruptibly.
    */
-  final def make[R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R, Nothing, Any]): ZManaged[R, E, A] =
+  final def make[R, R1 <: R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R1, Nothing, Any]): ZManaged[R1, E, A] =
     ZManaged(acquire.map(r => Reservation(IO.succeed(r), _ => release(r))))
 
   /**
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1226,7 +1226,7 @@ object ZManaged {
    * Lifts a `ZIO[R, E, A]` into `ZManaged[R, E, A]` with a release action.
    * The acquire and release actions will be performed uninterruptibly.
    */
-  final def make[R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R, Nothing, Any]): ZManaged[R, E, A] =
+  final def make[R, R1 <: R, E, A](acquire: ZIO[R, E, A])(release: A => ZIO[R1, Nothing, Any]): ZManaged[R1, E, A] =
```
mschuwalow: Wow, this is a quite unintuitive fix O.o

adamgfraser: I've used it in a couple of places and it works really well with contravariant types. You would think there would be an issue with the `R1 <: R` constraint if the environment type of the `release` is actually a supertype of the environment type of the `acquire`. But since `ZIO` is contravariant in `R` the compiler can arbitrarily narrow the type of `R1` to satisfy the constraint.

jdegoes: How about your other trick of making `R1` independent of `R` and returning `R with R1` in the result?

adamgfraser: Yes just confirmed that works too.

I thought this was slightly nicer because since you are just returning one type you should avoid the problem of getting back an environment type of `Clock with Clock`. But I guess if you did manually specify the environment type or looked at the type signature it could be confusing.

As far as I can tell they are identical from a type inference perspective and I think I have included tests for all possible relationships between the environment types.

Whichever one we use may be good to say what we think is best practice here and try to do more consistently.

adamgfraser: Thinking about this a bit more, I'm actually not sure this encoding should be surprising, at least given everything else we do in ZIO. Almost everything we do in ZIO where we compose two effects with environment types follows this pattern. For example, the signature of `ZIO#flatMap` is:

```scala
def flatMap[R1 <: R, E1 >: E, B](k: A => ZIO[R1, E1, B]): ZIO[R1, E1, B]
```

Here again we have a constraint that `R1 <: R` but that doesn't actually mean `k` has to use all the environmental effects that `self` does because of contravariance. So this "trick" is really just taking what we do in method type signatures all the time and pulling the `R` from a type parameter of `self` to a type parameter of the method since there is no `self` type here.

PR #2414, opened 2019-12-04
Implement Custom Extractors for Cause

### LINKED ISSUE #2353: Cause: Add extractors for terms of Cause
We could use some custom extractors (`unapply` methods in companion objects) for `Die`, `Fail`, `Both`, `Then`, and `Interrupt`). These custom extractors should strip `Meta` information from the nodes, and possibly `Trace` information as well (todo: discuss).

These would help people interact more naturally with the terms of the `Cause` subtype.

adamgfraser:
Resolves #2353. Extractors will look through `Meta` causes when matching. I left an extractor for `Traced` and didn't ignore that as I think there can be some cases where you want to introspect on the trace and it is easy to remove it with `untraced`.


core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -425,272 +425,356 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
...
-      case c @ Both(_, _)    => sym(empty)(this, c)
-      case traced: Traced[_] => this == traced.cause
-      case meta: Meta[_]     => this == meta.cause
-      case _                 => false
-    }
+  object Empty {
+    def unapply[E](cause: Cause[E]): Boolean =
+      cause.find {
+        case Internal.Empty        => true
+        case Internal.Fail(_)      => false
```
regiskuckaertz: It looks like you could merge all the false cases in a single one `case _: Cause[_] => false`, same for all the other ones?

regiskuckaertz: even using a simple pattern match instead of reaching for `find`.

adamgfraser: The reason we need this is to deal with the `Meta` case. Let's say the cause is `Internal.Meta(Internal.Fail(e))`. If we just pattern match on `Internal.Fail`  we won't match, but we want to ignore `Meta` and succeed with a match here. So then we might think okay, let's just pattern match on that. But `Meta` causes can be nested arbitrarily deeply so we need something to look through them, which is where `find` comes in.

This is also why we need to enumerate the other cases. `find` takes a partial function, so the partial function being defined for a cause means "we've found what we're looking for, stop searching" whereas it not being defined means "we haven't found it yet, keep recursing". So if we're looking for `Internal.Fail(e)` and we see `Internal.Both(l, r)` we want to stop searching immediately and fail the match. Whereas if we see `Meta(cause, _)` we want to recurse into `cause` to see if it satisfies the match.

regiskuckaertz: Lovely. I missed the addition of `Meta`.

jdegoes:
@adamgfraser Looks superb!

PR #2421, opened 2019-12-04
StreamEffect optimizations

vasilmkd:
I'm aware that this is not the nicest looking Scala code (manually instantiating `Function0`) but I believe it is worth it.

1. I have removed a whole layer of `Managed` instances on _each_ combinator.
2. I have removed all captured variables in closures (thus avoiding boxing), by making them local class fields (something that the JVM should be absolutely best at).

@jdegoes @iravid 

EDIT: See benchmark results below. I'm just about ready to admit defeat.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -37,68 +37,61 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, No
 
   override def collect[B](pf: PartialFunction[A, B]): StreamEffect[R, E, B] =
     StreamEffect {
-      self.processEffect.flatMap { thunk =>
-        Managed.effectTotal { () =>
-          {
-            var b = null.asInstanceOf[B]
+      self.processEffect.map { thunk => () =>
+        {
+          var b = null.asInstanceOf[B]
 
-            while (b == null) {
-              b = pf.applyOrElse(thunk(), (_: A) => null.asInstanceOf[B])
-            }
-
-            b
+          while (b == null) {
+            b = pf.applyOrElse(thunk(), (_: A) => null.asInstanceOf[B])
           }
+
+          b
         }
       }
     }
 
   override def collectWhile[B](pred: PartialFunction[A, B]): StreamEffect[R, E, B] =
     StreamEffect {
-      self.processEffect.flatMap { thunk =>
-        Managed.effectTotal {
+      self.processEffect.map { thunk =>
+        new Function0[B] {
           var done = false
 
-          () => {
+          def apply(): B =
             if (done) StreamEffect.end
             else pred.applyOrElse(thunk(), (_: A) => { done = true; StreamEffect.end })
-          }
         }
       }
     }
 
   override def dropWhile(pred: A => Boolean): StreamEffect[R, E, A] =
     StreamEffect {
-      self.processEffect.flatMap { thunk =>
-        Managed.effectTotal {
+      self.processEffect.map { thunk =>
+        new Function0[A] {
           var drop = true
 
           @annotation.tailrec
-          def pull(): A = {
+          def apply(): A = {
             val a = thunk()
             if (!drop) a
             else if (!pred(a)) {
               drop = false
               a
-            } else pull()
+            } else apply()
           }
-
-          () => pull()
         }
       }
     }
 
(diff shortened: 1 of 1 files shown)


vasilmkd:
I'll work on some benchmarks.

vasilmkd:
```scala
  @Benchmark
  def scalazFilterMapSum = {
    val stream = Stream.iterate(1L)(_ + 1L).take(10000000).filter(_ % 2 == 0).map(_ * 2).filter(_ % 2 == 0).map(_ * 2)
    val sink = Sink.foldLeft(0L)((sum, n: Long) => sum + n)
    val result = stream.run(sink)
    unsafeRun(result)
  }
```

[sum-bench-master.txt](https://github.com/zio/zio/files/3924327/sum-bench-master.txt)
[sum-bench-opt.txt](https://github.com/zio/zio/files/3924328/sum-bench-opt.txt)

vasilmkd:
Definitely not an expert, but I suspect that by instantiating real objects instead of lambdas (as Scala 2.12 source code is translated to jvm8 bytecode), we are foregoing `invokedynamic` for `invokeinterface`.

iravid:
Hmm it appears to be slower 🧐

vasilmkd:
Yeah, it appears so. Worth a try. Feel free to close the PR.

iravid:
Would you like to investigate why it is slower? Or not interested?

vasilmkd:
I think it's worth a look. Can you give me a few pointers?

iravid:
Try checking what those `Function0` objects are compiling into. Ideally the `var`s should be simple fields with no getter/setters. Also compare the bytecode to master to see if it's actually simplifying it.

vasilmkd:
I did that before. There was a getter. Can I maybe force direct access with a `public var`?

See some of my previous comments. I think the biggest change is losing `invokedynamic` lambdas, and instead doing an `invokeinterface` on a `Function0` object.

iravid:
`private[this]` might force direct field access

vasilmkd:
I'll try that. Other than that, there was definitely no boxing through `ObjectRef` and friends.

vasilmkd:
You're right, `private[this]` does force direct access, but it's still slower.

Master:
[sum-bench-master-private-this.txt](https://github.com/zio/zio/files/3925986/sum-bench-master-private-this.txt)
---
This PR:
[sum-bench-pr-private-this.txt](https://github.com/zio/zio/files/3925987/sum-bench-pr-private-this.txt)

vasilmkd:
@iravid Here are the .class files if you'd like to take a look. They are too long to paste here.

[classes.zip](https://github.com/zio/zio/files/3926235/classes.zip)

vasilmkd:
@iravid Browsing the bytecode, I noticed that `ZSink` methods are being called with `invokeinterface`. Since we have migrated `ZStream` to be a class, do you think it's time to migrate `ZSink` to become an `abstract class` to unlock some more optimizations from the JVM?

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -37,68 +37,61 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, No
...
         }
       }
     }
 
   override def collectWhile[B](pred: PartialFunction[A, B]): StreamEffect[R, E, B] =
     StreamEffect {
-      self.processEffect.flatMap { thunk =>
-        Managed.effectTotal {
+      self.processEffect.map { thunk =>
+        new Function0[B] {
```
regiskuckaertz: Can you memoise the object instead of recreating it each time?

vasilmkd: That one, maybe, but what I really set out to fix with this PR was to see if the variable capture boxing penalty could be reduced, and for those, I don't think that memoization of the function is an option, given the mutable state.

regiskuckaertz: Ah yeah, I see. So you defining a `Function2` and then currying it wouldn't help I guess 😞

vasilmkd: I thought about that too haha. The thing is, as soon as you introduce a ref type to a function, regardless of whether it is an argument or result, the function is not specialized anymore.

vasilmkd:
This was a fun experiment.

PR #2422, opened 2019-12-04
Add reentrant lock

jdegoes:


core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+ * STM transactions. However, this lock can be quite useful in effectful code,
+ * to provide consistent read/write access to mutable state; and being in STM
+ * allows this structure to be composed into more complicated concurrent
+ * structures that are consumed from effectful code.
+ */
+class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
+
+  /**
+   * Acquires a read lock. The transaction will suspend until no other fibers
+   * are holding write locks. Succeeds with the number of write locks held by
```
mschuwalow: s/write locks/read locks?

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+     * aside from the specified fiber id, then it is safe to upgrade the
+     * read lock into a write lock.
+     */
+    final def noOtherHolder(fiberId: Fiber.Id): Boolean =
+      readers.isEmpty || (readers.size == 1 && readers.contains(fiberId))
+
+    /**
+     * Computes the number of read locks held by the specified fiber id.
+     */
+    final def readLocks(fiberId: Fiber.Id): Int = readers.get(fiberId).fold(0)(identity)
```
regiskuckaertz:
```suggestion
    final def readLocks(fiberId: Fiber.Id): Int = readers.get(fiberId).getOrElse(0)
```

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Computes the number of write locks held by fibers.
+   */
+  final def writeLocks: STM[Nothing, Int] = data.get.map(_.fold(_ => 0, _.writeLocks))
+
+  private def adjustRead(fiberId: Fiber.Id, delta: Int): STM[Nothing, Int] =
+    (data.get.collect {
+      case Left(readLock)                    => Left(readLock.adjust(fiberId, delta))
+      case Right(WriteLock(w, r, `fiberId`)) => Right(WriteLock(w, r + delta, fiberId))
```
regiskuckaertz: Is it fine here if the number of read locks goes below 0?

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+   * write locks held by this fiber.
+   */
+  lazy val acquireWrite: STM[Nothing, Int] =
+    for {
+      fiberId <- STM.fiberId
+      w <- data.get.collect {
+            case Left(readLock) if (readLock.noOtherHolder(fiberId)) =>
+              WriteLock(1, readLock.readLocks(fiberId), fiberId)
+
+            case Right(WriteLock(n, m, `fiberId`)) =>
```
regiskuckaertz: Is there a reason for the backticks?

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * Releases a write lock held by this fiber. Succeeds with the outstanding
+   * number of write locks held by this fiber.
+   */
+  lazy val releaseWrite: STM[Nothing, Int] =
+    STM.fiberId.flatMap(
+      fiberId =>
+        data.modify {
+          case Right(WriteLock(1, m, `fiberId`)) =>
+            0 -> Left(ReadLock(fiberId, m))
```
regiskuckaertz: Yes, I think here if `m < 0` because of how `adjustRead` deals with the `WriteLock` situation, then you reach a bad state.

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+
+          case Right(WriteLock(n, m, `fiberId`)) if n > 1 =>
+            val newCount = n - 1
+
+            newCount -> Right(WriteLock(n - 1, m, fiberId))
```
regiskuckaertz:
```suggestion
            newCount -> Right(WriteLock(newCount, m, fiberId))
```

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Releases a write lock held by this fiber. Succeeds with the outstanding
+   * number of write locks held by this fiber.
+   */
+  lazy val releaseWrite: STM[Nothing, Int] =
+    STM.fiberId.flatMap(
+      fiberId =>
+        data.modify {
+          case Right(WriteLock(1, m, `fiberId`)) =>
```
regiskuckaertz: Should you test here that the two `fiberId` are equal?

iravid: That’s what the bzckticks do. Instead of binding a new value to fiberId, it checks that the value is equal to the existing fiberId value.

Very strange pattern matching syntax 🤷🏼‍♂️

regiskuckaertz: 🤯

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
jdegoes: Done!

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
jdegoes: Fixed by throwing useful die message.

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
jdegoes: To match for equality for the variable of the same name in the outer scope.

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
jdegoes: Fixed!

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
jdegoes: Done!

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -0,0 +1,225 @@
...
+ * Although located in the STM package, there is no need for locks within
+ * STM transactions. However, this lock can be quite useful in effectful code,
+ * to provide consistent read/write access to mutable state; and being in STM
+ * allows this structure to be composed into more complicated concurrent
+ * structures that are consumed from effectful code.
+ */
+class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
+
+  /**
+   * Acquires a read lock. The transaction will suspend until no other fibers
```
adamgfraser: Shouldn't this be "The transaction will suspend until no other fiber is holding a write lock."?

mijicd:
Very nice!

PR #2423, opened 2019-12-05
Implement ZIO#fromFutureInterrupt

### LINKED ISSUE #2419: Add ZIO.fromFutureInterrupt
Add a method `ZIO.fromFutureInterrupt((ec: ExecutionContext) => ... : Future[A])`, which passes it a specially crafted `ExecutionContext`, which will disable `submit` when the parent ZIO effect has been interrupted. This will allow interruption of running `Future`, albeit without any guarantee of resource cleanup, and without any interruption between async boundaries.

adamgfraser:
Resolves #2419.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -494,16 +494,20 @@ object ZIOSpec extends ZIOBaseSpec {
     ),
     suite("fromFutureInterrupt")(
       testM("running Future can be interrupted") {
+        import java.util.concurrent.atomic.AtomicInteger
+        import scala.concurrent.{ ExecutionContext, Future }
+        def infiniteFuture(ref: AtomicInteger)(implicit ec: ExecutionContext): Future[Nothing] =
+          Future(ref.getAndIncrement()).flatMap(_ => infiniteFuture(ref))
         for {
-          p1     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
-          p2     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
-          fiber  <- ZIO.fromFutureInterrupt(ec => p1.future.map(p2.success)(ec)).fork
-          _      <- fiber.interrupt
-          _      <- ZIO.effectTotal(p1.success(()))
-          result <- ZIO.effectTotal(p2.isCompleted)
-        } yield assert(result, isFalse)
+          ref   <- ZIO.effectTotal(new AtomicInteger(0))
+          fiber <- ZIO.fromFutureInterrupt(ec => infiniteFuture(ref)(ec)).fork
+          _     <- fiber.interrupt
+          v1    <- ZIO.effectTotal(ref.get)
+          _     <- Live.live(clock.sleep(10.milliseconds))
+          v2    <- ZIO.effectTotal(ref.get)
+        } yield assert(v1, equalTo(v2))
       }
-    ) @@ nonFlaky,
+    ),
     suite("head")(
       testM("on non empty list") {
         assertM(ZIO.succeed(List(1, 2, 3)).head.either, isRight(equalTo(1)))
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -506,6 +506,12 @@ object ZIOSpec extends ZIOBaseSpec {
           _     <- Live.live(clock.sleep(10.milliseconds))
           v2    <- ZIO.effectTotal(ref.get)
         } yield assert(v1, equalTo(v2))
+      },
+      testM("interruption blocks on interruption of the Future") {
+        for {
+          fiber  <- ZIO.fromFutureInterrupt(_ => scala.concurrent.Future.never).fork
+          result <- Live.withLive(fiber.interrupt)(_.timeout(10.milliseconds))
+        } yield assert(result, isNone)
       }
     ),
     suite("head")(
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -508,8 +508,13 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(v1, equalTo(v2))
       },
       testM("interruption blocks on interruption of the Future") {
+        import scala.concurrent.Promise
         for {
-          fiber  <- ZIO.fromFutureInterrupt(_ => scala.concurrent.Future.never).fork
+          latch <- ZIO.effectTotal(Promise[Unit]())
+          fiber <- ZIO.fromFutureInterrupt { _ =>
+                    latch.success(()); Promise[Unit]().future
+                  }.fork
+          _      <- ZIO.fromFuture(_ => latch.future).orDie
           result <- Live.withLive(fiber.interrupt)(_.timeout(10.milliseconds))
         } yield assert(result, isNone)
       }
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -492,6 +492,18 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assertCompletes
       }
     ),
+    suite("fromFutureInterrupt")(
+      testM("running Future can be interrupted") {
+        for {
+          p1     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          p2     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          fiber  <- ZIO.fromFutureInterrupt(ec => p1.future.map(p2.success)(ec)).fork
+          _      <- fiber.interrupt
+          _      <- ZIO.effectTotal(p1.success(()))
+          result <- ZIO.effectTotal(p2.isCompleted)
+        } yield assert(result, isFalse)
+      }
+    ) @@ nonFlaky,
     suite("head")(
       testM("on non empty list") {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2240,6 +2240,29 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+          if (!interrupted.get) ec.execute(runnable) else ()
+        def reportFailure(cause: Throwable): Unit =
+          ec.reportFailure(cause)
+      }
+      effect(make(interruptibleEC)).flatMap { f =>
+        f.value
+          .fold(
+            Task.effectAsync { (cb: Task[A] => Unit) =>
+              f.onComplete {
+                case Success(a) => cb(Task.succeed(a))
```
regiskuckaertz: may be worth check `interrupted` here too

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -492,6 +492,18 @@ object ZIOSpec extends ZIOBaseSpec {
...
     ),
+    suite("fromFutureInterrupt")(
+      testM("running Future can be interrupted") {
+        for {
+          p1     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          p2     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          fiber  <- ZIO.fromFutureInterrupt(ec => p1.future.map(p2.success)(ec)).fork
+          _      <- fiber.interrupt
+          _      <- ZIO.effectTotal(p1.success(()))
+          result <- ZIO.effectTotal(p2.isCompleted)
```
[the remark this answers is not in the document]
adamgfraser: I think the interruption is guaranteed to occur before `p2` can complete. `fiber.interrupt` is not going to return until the finalizer of the interrupted fiber has completed, which means that `interrupted` will have been set to `true` before we begin evaluating `p1.success(())`. Then we complete `p1` and evaluate the future continuation but since `interrupted` is true the continuation will never be submitted for execution.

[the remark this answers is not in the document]
iravid: Oh right! I missed that `p1` is only resolved after the interruption.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I think this is already taken care of by `effectAsync`. If `effectAsync` is interrupted before the callback is invoke, any continuation from `effectAsync` will not be executed.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -492,6 +492,18 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assertCompletes
       }
     ),
+    suite("fromFutureInterrupt")(
+      testM("running Future can be interrupted") {
+        for {
+          p1     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          p2     <- ZIO.effectTotal(scala.concurrent.Promise[Unit])
+          fiber  <- ZIO.fromFutureInterrupt(ec => p1.future.map(p2.success)(ec)).fork
```
jdegoes: I would create a future chain that is infinite, e.g.:

```scala
def infiniteFuture(implicit ec: ExecutionContext): Future[Nothing] = 
  Future(()).flatMap(_ => infiniteFuture)
```

This way you know it will never terminate unless called with `fromFutureInterrupt`.

adamgfraser: I thought about doing that. How would you verify that it did terminate? If we installed a finalizer on the effect wouldn't it be triggered on interruption even if the future didn't terminate.

regiskuckaertz: The test won't time out if it terminates?

jdegoes: Ideally `interrupt` would not return until `Future` as killed, but to verify that:

```scala
def infiniteFuture(ref: AtomicInteger)(implicit ec: ExecutionContext): Future[Nothing] = 
  Future(ref.getAndIncrement()).flatMap(_ => infiniteFuture(ref))
```

Then after interrupt returns, you verify that `ref` doesn't change for some small amount of real time (10 millis or so).

adamgfraser: All set now.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2240,6 +2240,29 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+                case Failure(t) => cb(Task.fail(t))
+              }(interruptibleEC)
+            }
+          )(Task.fromTry(_))
+      }.onInterrupt(Task.effectTotal(interrupted.set(true)))
```
jdegoes: This looks great. The only thing I would suggest is to "block" interruption on the interruption of the `Future`. This will allow very precise knowledge of when the future is terminated to propagate upward (of course users can always do `.interrupt.fork` if they don't want that).

To do this, you can create an outer async effect, and complete it from inside the `interruptibleEC` (after the Future is interrupted, i.e. on failed submit), but also complete it if the Future is completed.

Does that make sense?

adamgfraser: I think it raises the question of when we consider the future as being interrupted.

I had been conceptualizing it as the future being interrupted when `interrupted` is set to `true`, because then we know that no further steps can occur in the chain of continuations.

It sounds like you are proposing a concept where the future counts as being interrupted when the first runnable submitted for execution is rejected.

I'm not sure that actually gives us a stronger guarantee and I think it raises some potential issues.

I think the stronger guarantee you would want is that the future is not doing anything anymore. But the original future could have spawned multiple futures so even if one runnable submitted for execution is rejected there could be other futures that are still "in flight".

The other issue is that it makes the timing of interruption dependent on the timing of the future. If we have the equivalent of `Future.never` interruption would not terminate because the future would never complete or submit to the execution context. Or if the future only continues at a long interval interruption has to wait for the next interval.

jdegoes: What you say is true. Waiting for the next submit would guarantee that _some_ `Future` has been interrupted, but it's possible the implicit `ec` would have been passed and used in many `Future`'s, so we can't say it's the only continuation of the originally created `Future`. 

My only concern about the current implementation is that:

 - Interrupting a ZIO effect may never terminate: that's a risk of calling `interrupt` without also forking that interrupt
 - Interrupting a ZIO effect may not terminate all children; if it's children were created with `daemon` mode on
 - Interrupting a ZIO effect doesn't return until we know for sure it's done executing

So it seems to me there are similar risks with interrupting a `Future`. 

Also if you look at `effectBlocking`, when we interrupt some synchronous code, we actually do not return from `interrupt` until it's actually interrupted, despite the potential thread leak resulting from this behavior.

So it seems more consistent to _try_ to emulate that semantic for `Future`, together with its inherent risks. Although your point above is quite valid: at best we can only know we interrupted _some_ continuation of the `Future`, and if there are many continuations, we won't know and can't wait around for them.

Thoughts?

adamgfraser: I think those are all good points. This seems like one of those situations where we are interacting with code that is not principled in the same way we are so we aren't going to be able to do something that is completely principled and just have to pick the best thing. And I think you have done a good job of laying out as good a set of principles as we can have in that situation.

My only worry is that the implications of that can be counterintuitive. For example, interrupting a fiber forked from `ZIO.never` is completely safe and will return almost immediately. On the other hand, using `ZIO.fromFutureInterrupt` with `Future.never` and attempting to interrupt that will hang. That makes sense if you say we need some at least "best efforts" evidence that the Future is done executing but that can be a subtle point for two things that look similar on one level.

Totally happy to do as you suggest though.

adamgfraser: I changed it as you suggested. I struggled a little bit with the implementation here. I tried to use an outer asynchronous effect as you suggested but I found that I needed one effect to signal the interruption, which needed to be the "outer" effect because I needed it in scope when creating the `interruptibleEC`, but then that ended up being the result type of the composed effect.

jdegoes: > My only worry is that the implications of that can be counterintuitive. 

I agree with you here. It's counterintuitive. 

The flipside is the counterintuition from the other solution: you interrupt a future but it's still doing things in the background (which means, for one: it's not safe to attempt to do the same thing in some solutions without leaking resources, e.g. consider an http request being thrown away, it is not "resource safe" to begin another one until the old one killed).

Like you say, it's a war of competing tradeoffs and I don't think any solution will be optimal, because future. On the net I believe — but do not know for sure — that an infinite future will be comparatively rare, while interruption is so automatic and composable that if the user has composed a ZIO effect from many ZIO effects and some futures, the fact that interrupting the whole chain has definitely left things running in the background due to the futures may be detectable in more situations.

Anyway, if you're happy with this, I am good to merge, looks great!

adamgfraser: I'm good with it.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2240,6 +2240,35 @@ private[zio] trait ZIOFunctions extends Serializable {
       }
     }
 
+  final def fromFutureInterrupt[A](make: ExecutionContext => scala.concurrent.Future[A]): Task[A] =
+    Task.descriptorWith { d =>
+      val ec          = d.executor.asEC
+      val interrupted = new java.util.concurrent.atomic.AtomicBoolean(false)
```
hepin1989: It's better to use AtomicBooleanFieldUpdater

PR #2424, opened 2019-12-05
Fix URIO#bracket Type Signature

adamgfraser:
The type signature was unnecessarily widening the error type from `Nothing` to `Throwable`.


jdegoes:
Oh my, good catch! 😅

PR #2425, opened 2019-12-05
ZStream.flatten, unwrap, unwrapManaged are safe to pull again

vasilmkd:
Verifying more Stream constructors for pull safety.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamPullSafetySpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamPullSafetySpec.scala
@@ -529,6 +529,46 @@ object StreamPullSafetySpec extends ZIOBaseSpec {
           .use(nPulls(_, 3))
           .map(assert(_, equalTo(List(Left(Some("Ouch")), Left(None), Left(None)))))
       }
+    ),
+    suite("Stream.unwrapManaged")(
+      testM("is safe to pull again after success") {
+        for {
+          ref <- Ref.make(false)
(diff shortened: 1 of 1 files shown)


streams-tests/jvm/src/test/scala/zio/stream/StreamPullSafetySpec.scala
```diff
@@ -63,6 +63,29 @@ object StreamPullSafetySpec extends ZIOBaseSpec {
         } yield assert(fin, isTrue) && assert(pulls, equalTo(List(Left(Some("Ouch")), Left(None), Left(None))))
       }
     ),
+    suite("Stream.flatten")(
+      testM("is safe to pull again after success") {
+        Stream
+          .flatten(Stream(Stream.fromEffect(UIO.succeed(5))))
+          .process
+          .use(nPulls(_, 3))
+          .map(assert(_, equalTo(List(Right(5), Left(None), Left(None)))))
```
regiskuckaertz: not that it matters but you can write as:
```scala
assertM(
  Stream.flatten(Stream(Stream.fromEffect(UIO.succeed(5)))).process.use(nPulls(_, 3)),
  equalTo(List(Right(5), Left(None), Left(None))
)
```

vasilmkd: I'm aware, however your formulation tends to show the warning "type inferred to be Any" for the environment, which has since been fixed in upstream scalac. I mainly avoid it for the warning.

PR #2428, opened 2019-12-06
Add ZStream constructors for Java Iterators and Streams

iravid:


streams/shared/src/main/scala/zio/stream/Stream.scala
```diff
@@ -225,6 +226,30 @@ object Stream extends Serializable {
   final def fromIteratorManaged[E, A](iterator: Managed[E, Iterator[A]]): Stream[E, A] =
     ZStream.fromIteratorManaged(iterator)
 
+  /**
+   * See [[ZStream.fromJavaIterator]]
+   */
+  final def fromJavaIterator[E, A](iterator: IO[E, ju.Iterator[A]]): Stream[E, A] =
```
regiskuckaertz: Shouldn't these go under the `jvm` branch?

iravid: Hmm it compiles on JS but that might be misleading. I’ll check if Iterator and Stream are implemented there.

vasilmkd:
This looks like a lot of repetitive code. Could we maybe use some of the methods defined here and delegate to the old implementations of the iterator code?
https://www.scala-lang.org/api/2.13.1/scala/collection/JavaConverters$.html

iravid:
Yes, good point @vasilmkd!

streams/shared/src/main/scala/zio/stream/Stream.scala
iravid: So there's a ju.Iterator implementation but not Stream. Weird that it compiles. Anyway I moved the Stream things to platform-specific.

vasilmkd: What about this? Not very familiar with converters though. Maybe you could take a look.
https://www.scala-lang.org/api/2.13.1/scala/jdk/StreamConverters$.html

vasilmkd: Ah, maybe it's only available for 2.13. :/

iravid:
Dropped the use of the JavaConverters to avoid 2.12/2.13 compatibility shenanigans.

@vasilmkd StreamConverters only provides an eager transformation from Java's Stream to a Scala Collection, so not what we need here.

Merging when CI is green.

vasilmkd:
TIL, thanks.

PR #2435, opened 2019-12-06
Add Documentation for ZIO#fromFutureInterrupt

adamgfraser:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2225,6 +2225,15 @@ private[zio] trait ZIOFunctions extends Serializable {
       }
     }
 
+  /**
+   * Imports a function that creates a [[scala.concurrent.Future]] from an
+   * [[scala.concurrent.ExecutionContext]] into a `ZIO`. The provided
+   * `ExecutionContext` will interrupt the `Future` between asynchronous
+   * operations such as `map` and `flatMap` if this effect is interrupted. Note
+   * that no attempt will will be made to interrupt a `Future` blocking on a
```
wi101: Awesome, a small comment otherwise it looks good to me
```suggestion
   * that no attempt will be made to interrupt a `Future` blocking on a
```

adamgfraser: Fixed now!

PR #2442, opened 2019-12-07
Add ZStream.drainFork, haltWhen, interruptWhen

iravid:
cc @vasilmkd

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1094,7 +1094,19 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * }}}
    */
   final def drain: ZStream[R, E, Nothing] =
-    ZStream[R, E, Nothing](self.process.map(_.forever))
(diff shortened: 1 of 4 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1645,6 +1678,24 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+   *
+   * If the promise completes with a failure, the stream will emit that failure.
+   */
+  final def interruptWhen[E1 >: E](p: Promise[E1, _]): ZStream[R, E1, A] =
+    ZStream {
+      for {
+        as   <- self.process
+        done <- Ref.make(false).toManaged_
+        pull = done.get flatMap {
+          if (_) Pull.end
```
svroonland: Could this be simplified to:

```scala
      ZStream {
        for {
          as <- self.process
          pull = p.isDone.flatMap {
            if (_) Pull.end
            else as.raceAttempt(p.await.mapError(Some(_)) *> Pull.end)
          }
        } yield pull
      }
```
?

Issue #2436, opened 2019-12-06
Fibers' `onTermination` effects are only run on sbt `exit`


mtsokol:
@adamgfraser So here's the code snippet with sbt output, thanks once again for looking at this together 🙂 

For such program:
```scala
import zio._
import zio.console.Console.Live.console

object Example extends App {

  def run(args: List[String]): ZIO[ZEnv, Nothing, Int] =
    myAppLogic.as(0)

  val myAppLogic =
    for {
      fib <- IO.never.onTermination(_ => console.putStrLn("First")).fork
      _   <- fib.interrupt
      _   <- IO.never.onTermination(_ => console.putStrLn("Second")).fork
    } yield ()
}
```
The output is:
```
[info] Running Main
First
[success] Total time: 1 s, completed Dec 7, 2019 12:12:28 AM
```
So after completing the run the second `onTermination` is not executed?

And the rest is completed after sbt `exit` (here accumulated after few runs):
```
sbt:zio-workspace> exit
[info] shutting down server
Second
Second
Second
Second
```
[EDIT] Also after trying few times the output sometimes is with hanging sbt:
```
sbt:zio-workspace> run
[info] Compiling 1 Scala source to /Users/mateusz/IdeaProjects/play-json-workspace/target/scala-2.12/classes ...
[info] running Example
First
[success] Total time: 2 s, completed Dec 7, 2019 12:36:16 AM
sbt:zio-workspace> exit
[info] shutting down sbt server
[warn] Thread[Thread-4,5,run-main-group-0] loading zio/App$$anon$1.class after test or run has completed. This is a likely resource leak
```

jdegoes:
You aren't interrupting the 2nd one, which means it is only being run when the JVM is shutdown, via the shutdown hook installed by `zio.App`. SBT controls when that JVM is terminated.

Maybe we could add a method `zio.App.forceShutdown` to do the same thing that's happening right now when the JVM is shutdown. That'd probably be a good ticket!

PR #2445, opened 2019-12-08
Added interruption in App main

mtsokol:
Regards #2436 

@jdegoes @adamgfraser  Hi!
I've started working on #2436. The first idea (and the simplest) that came to my head is just interrupting fiber after `.join`, instead of introducing new method. Thanks to this we can be sure that after running App.main we interrupt fibers.  

Such modification fixes example in the issue, but I don't know whether it's correct solution.
Also I'm not sure whether this shutdown hook is needed then.

What do you think about it?

### DIFF
--- a/core/jvm/src/main/scala/zio/App.scala
+++ b/core/jvm/src/main/scala/zio/App.scala
@@ -60,6 +60,7 @@ trait App extends DefaultRuntime {
                 }
               }))
           result <- fiber.join
+          _      <- fiber.interrupt
         } yield result
       )
     )


mtsokol:
@jdegoes So what do you think about such solution?

jdegoes:
I think seeing this code revealed a bad design choice in ZIO: right now `interrupt` after `join` could actually do something: i.e. interrupt all descendants. That's extremely surprising and needs to be fixed. So I am good to merge this change even though I think ultimately we'll want to change the `fiber.interrupt` to something like `fiber.interruptDescendants`.

jdegoes:
@mtsokol Thank you for this contribution! 🙏

Issue #2368, opened 2019-11-30
STM: Enhance TArray


jdegoes:
`zio.stm.TArray` is missing many useful methods:

 - `find` to find an element
 - `indexOf` to get the index of a matching element
 - etc.

Probably 10 or so helpful utility methods could be added to `TArray` to make summarizing and retrieving information from the array more useful.

I would look on methods of `Array` / `Vector` in Scala and see which ones would be a good fit for `TArray`.

horned-sphere:
I'd like to have a go at this one. I've made a start but need to add in some tests yet.

horned-sphere:
I've made a [PR for this](https://github.com/zio/zio/pull/2446). Hopefully I haven't gone too far with adding methods.

mijicd:
Closed by #2446.

PR #2446, opened 2019-12-08
New search and aggregation methods on TArray.

horned-sphere:
This will complete #2368.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TArray.scala
+++ b/core/shared/src/main/scala/zio/stm/TArray.scala
@@ -26,6 +26,65 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
     if (0 <= index && index < array.length) array(index).get
     else STM.die(new ArrayIndexOutOfBoundsException(index))
 
+  /** Finds the result of applying a partial function to the first value in its domain. */
+  final def collectFirst[B](pf: PartialFunction[A, B]): STM[Nothing, Option[B]] =
+    find(pf.isDefinedAt).map(_.map(pf))
+
+  /** Finds the result of applying an STM partial function to the first value in its domain. */
+  final def collectFirstM[E, B](pf: PartialFunction[A, STM[E, B]]): STM[E, Option[B]] =
+    find(pf.isDefinedAt).flatMap {
+      case Some(a) => pf(a).map(Some(_))
+      case _       => STM.succeed(None)
+    }
+
+  /** Determine if the array contains a specified value. */
+  final def contains(a: A): STM[Nothing, Boolean] = exists(_ == a)
+
+  /** Count the values in the array matching a predicate. */
+  final def count(predicate: A => Boolean): STM[Nothing, Int] = fold(0) { (n, a) =>
+    if (predicate(a)) n + 1 else n
+  }
+
+  /** Count the values in the array matching an STM predicate. */
+  final def countM[E](predicate: A => STM[E, Boolean]): STM[E, Int] = foldM[E, Int](0) { (n, a) =>
+    predicate(a).map(result => if (result) n + 1 else n)
+  }
+
+  /** Determine if the array contains a value satisfying a predicate. */
+  final def exists(predicate: A => Boolean): STM[Nothing, Boolean] = find(predicate).map(_.isDefined)
+
+  /** Determine if the array contains a value satisfying an STM predicate. */
+  final def existsM[E](predicate: A => STM[E, Boolean]): STM[E, Boolean] =
+    countM(predicate).map(_ > 0) //Existence should not be order dependent so the entire array needs to be observed.
+
+  /** Find the first element in the array matching a predicate. */
+  final def find(predicate: A => Boolean): STM[Nothing, Option[A]] =
+    if (array.isEmpty) STM.succeed(None)
+    else
+      array.head.get.flatMap { a =>
+        if (predicate(a)) STM.succeed(Some(a))
+        else new TArray(array.tail).find(predicate)
+      }
+
+  /** Find the last element in the array matching a predicate. */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -39,10 +98,124 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+
+  /** Atomically reduce the array by a binary operator. */
+  final def reduce(op: (A, A) => A): STM[Nothing, A] =
+    if (array.isEmpty) STM.die(new UnsupportedOperationException("reduction of empty array"))
+    else
+      array.head.get
+        .flatMap(h => new TArray(array.tail).fold(h)((m, a) => op(m, a)))
+
+  /** Atomically reduce the array by an STM binary operator. */
+  final def reduceM[E](op: (A, A) => STM[E, A]): STM[E, A] =
```
jdegoes: I would nuke this one and `reduce`, because they can blow up. The safer versions are much nicer to work with!

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -39,10 +98,124 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+    val len = array.length
+    def forIndex(i: Int): STM[E, Int] =
+      if (i >= len) STM.succeed(-1)
+      else apply(i).flatMap(a => predicate(a).flatMap(result => if (result) STM.succeed(i) else forIndex(i + 1)))
+
+    if (from >= 0) forIndex(from) else STM.succeed(-1)
+  }
+
+  /** The last entry in the array. */
+  final def last: STM[Nothing, A] =
```
jdegoes: I would make this `lastOption` and nuke `last`.

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -39,10 +98,124 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+  /** Atomically evaluate the conjunction of an STM predicate across the members of the array. */
+  final def forallM[E](predicate: A => STM[E, Boolean]): STM[E, Boolean] =
+    countM(predicate).map(_ == array.length) //Universal quantification should not be order dependent so the entire array needs to be observed.
+
   /** Atomically performs side-effect for each item in array */
   final def foreach[E](f: A => STM[E, Unit]): STM[E, Unit] =
     this.foldM(())((_, a) => f(a))
 
+  /** The head of the array. */
+  final def head: STM[Nothing, A] =
```
jdegoes: I would nuke this one and only keep `headOption`, but instead rename it to `firstOption` for symmetry with `lastOption`.

core/shared/src/main/scala/zio/stm/TArray.scala
horned-sphere: I agree entirely; I was just trying to stay as close to Seq as possible. More than happy to get rid of them!

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -26,6 +26,65 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+  /** Determine if the array contains a value satisfying a predicate. */
+  final def exists(predicate: A => Boolean): STM[Nothing, Boolean] = find(predicate).map(_.isDefined)
+
+  /** Determine if the array contains a value satisfying an STM predicate. */
+  final def existsM[E](predicate: A => STM[E, Boolean]): STM[E, Boolean] =
+    countM(predicate).map(_ > 0) //Existence should not be order dependent so the entire array needs to be observed.
+
+  /** Find the first element in the array matching a predicate. */
+  final def find(predicate: A => Boolean): STM[Nothing, Option[A]] =
+    if (array.isEmpty) STM.succeed(None)
```
mijicd: Maybe express it in terms of `fold`?

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -26,6 +26,65 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+  /** Find the last element in the array matching a predicate. */
+  final def findLast(predicate: A => Boolean): STM[Nothing, Option[A]] =
+    new TArray(array.reverse).find(predicate)
+
+  /** Find the last element in the array matching an STM predicate. */
+  final def findLastM[E](predicate: A => STM[E, Boolean]): STM[E, Option[A]] =
+    new TArray(array.reverse).findM(predicate)
+
+  /** Find the first element in the array matching an STM predicate. */
+  final def findM[E](predicate: A => STM[E, Boolean]): STM[E, Option[A]] =
```
mijicd: This one can probably be expressed in terms of `foldM`.

core/shared/src/main/scala/zio/stm/TArray.scala
horned-sphere: The reason I did not was so that it could return early at the first match. TArray doesn't have a foldWhile (like Chunk, for example). I'm happy to change it to use fold if you think it's better to keep it simpler. What do you think?

mijicd: Good point. Let's keep it this way, and add the combinators like foldWhile once we make an iteration to optimizer the transactional structures.

mijicd: Good point. Let's keep it this way, and add the combinators like foldWhile once we make an iteration to optimizer the transactional structures.

horned-sphere:
All the unsafe methods are now gone (except apply which was there already). headOption is now firstOption.

PR #2447, opened 2019-12-08
Make ZStream#aggregate safe to pull again

vasilmkd:
I am extremely grateful for how many tests we have covering all subtle edge cases of this combinator.

@iravid

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamEffect.scala
@@ -234,7 +234,7 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, No
   override def aggregate[R1 <: R, E1 >: E, A1 >: A, B](
     sink: ZSink[R1, E1, A1, A1, B]
   ): ZStream[R1, E1, B] = {
-    import StreamEffect.internal.AggregateState
+    import ZStream.internal.AggregateState
 
     sink match {
       case sink: SinkPure[E1, A1, A1, B] =>
@@ -327,17 +327,6 @@ private[stream] object StreamEffect extends Serializable {
 
   case object End extends Throwable("stream end", null, true, false) {
     override def fillInStackTrace() = this
-  }
-
-  private[StreamEffect] object internal {
-    sealed abstract class AggregateState[+S, +A]
-    object AggregateState {
-      final case class Pull[S](s: S, dirty: Boolean)                      extends AggregateState[S, Nothing]
-      final case class Extract[S, A](s: S, leftovers: Chunk[A])           extends AggregateState[S, A]
-      final case class Drain[S, A](s: S, leftovers: Chunk[A], index: Int) extends AggregateState[S, A]
-      final case class DirtyDone[S](s: S)                                 extends AggregateState[S, Nothing]
-      case object Done                                                    extends AggregateState[Nothing, Nothing]
-    }
   }
 
   def end[A]: A = throw End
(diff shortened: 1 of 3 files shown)


vasilmkd:
I'll fix 2.11 compilation issues but can anyone take a look what the dotty issue is? @adamgfraser

vasilmkd:
@iravid you have written a lot of local classes and path dependent types before, can you please take a look at the compilation issues? I'm running out of ideas.

regiskuckaertz:
What if you remove the `final` modifier?

vasilmkd:
> What if you remove the `final` modifier?

Tried, sadly no. :cry:

regiskuckaertz:
Hmm, another idea would be to hoist the definition at the top of the method, rather than there inside the pattern match 🤔

vasilmkd:
Yes, I thought of that too, but I need the matched sink value's `State` type.

vasilmkd:
Dotty seems to really dislike local definitions. The error had something to do with the significant whitespace syntax, which I thought we had solved before.

Anyways, it's finally ready for review. @iravid

adamgfraser:
@vasilmkd I don't think it is related to the significant whitespace proposal, since we have disabled that. I believe it is related to defining a local type that shadows an abstract type in an enclosing scope (`sealed trait State` in `aggregate` and `type State` in `ZSink`). You should be able to avoid it by just renaming your ADT to something like `AggregateState` and then creating a type alias `type State = AggregateState` as done [here](https://github.com/zio/zio/pull/1928/files#diff-7e76fb4cea322437475668ac73762d6aR729).

vasilmkd:
@adamgfraser Thanks, I'll try that. The thing is, I think `ZStream.scala` compiles correctly without these workarounds.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
```diff
@@ -232,58 +232,73 @@ private[stream] class StreamEffect[-R, +E, +A](val processEffect: ZManaged[R, No
...
-                      val (newState, newLeftovers) = sink.stepChunkPure(state, leftovers)
-                      leftovers = newLeftovers
-                      go(newState, true)
-                    } else {
-                      val a = thunk()
-                      go(sink.stepPure(state, a), true)
-                    }
+              var state: AggregateState[sink.State, A1] = AggregateState.Pull(sink.initialPure, false)
+
+              @annotation.tailrec
```
regiskuckaertz: 🚀 ✨

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -586,70 +586,92 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
-                              leftoversRef
-                                .update(_ ++ leftovers)
-                                .when(leftovers.nonEmpty)
-                                .as(b)
-                          }
-                    },
+    ZStream
+      .managed(managedSink)
+      .flatMap { sink =>
+        sealed abstract class State
```
regiskuckaertz: Looks like you could reuse `AggregateState` here?

vasilmkd: Sure, why not.

streams/shared/src/main/scala/zio/stream/StreamEffect.scala
vasilmkd: 🤣 🤣 🤣

vasilmkd:
I tried to minimize code duplication as much as possible. It took quite some time to write though. :)

PR #2449, opened 2019-12-08
Implement ZIO#fallback

### LINKED ISSUE #2448: ZIO Core: Add fallback (non-effectful orElse)
```scala
sealed trait ZIO[-R, +E, +A] {
  def fallback(a: => A): ZIO[R, Nothing, A] = self fold (_ => a, identity)
}
```

In hindsight this should be called `orElse` and the other one `orElseM`, but oh well! 😆

adamgfraser:
Resolves #2448.

I also added `fallback` to `ZManaged` and `STM`. I noticed that the error handling methods in `STM` weren't using `CanFail` so I added that as well.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -373,6 +373,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self orElse eventually
 
   /**
+   * Executes this effect and returns its value, if it succeeds, but otherwise
+   * returns the specified value.
+   */
+  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): ZIO[R, Nothing, A1] =
+    fold(_ => a, identity)
+
+  /**
    * Dies with specified `Throwable` if the predicate fails.
    */
   final def filterOrDie(p: A => Boolean)(t: => Throwable): ZIO[R, E, A] =
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -157,7 +157,7 @@ final class STM[+E, +A] private[stm] (
    * Tries this effect first, and if it fails, succeeds with the specified
    * value.
    */
-  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): STM[E, A1] =
+  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): STM[Nothing, A1] =
     fold(_ => a, identity)
 
   /**
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -18,7 +18,7 @@ package zio.stm
 
 import java.util.concurrent.atomic.{ AtomicBoolean, AtomicLong }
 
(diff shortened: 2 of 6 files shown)


core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -151,6 +153,13 @@ final class STM[+E, +A] private[stm] (
   final def ensuring(finalizer: STM[Nothing, Any]): STM[E, A] =
     foldM(e => finalizer *> STM.fail(e), a => finalizer *> STM.succeed(a))
 
+  /**
+   * Tries this effect first, and if it fails, succeeds with the specified
+   * value.
+   */
+  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): STM[E, A1] =
```
mijicd: Do we need `E` in return type?

adamgfraser: Good catch! Fixed now.

mijicd: Nice!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -372,6 +372,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def eventually(implicit ev: CanFail[E]): URIO[R, A] =
     self orElse eventually
 
+  /**
+   * Executes this effect and returns its value, if it succeeds, but otherwise
+   * returns the specified value.
+   */
+  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): ZIO[R, Nothing, A1] =
+    fold(_ => a, identity)
```
regiskuckaertz: `catchAll(_ => a)`

adamgfraser: I'm not sure this is better if we need to wrap the result in an additional effect constructor.

adamgfraser:
@mijicd I think the issue here is that we already are using the `orElse` name very widely in ZIO where the other value is an effect. So we either have to change `orElse` in `ZIO`, which is a relatively significant change at this point, or come up with another name. Per the ticket it seems like we want to use another name. And then once we are using `fallback` in ZIO I think it makes sense to use it in other data types for consistency, even though I agree the STM API is less settled so if it was STM in isolation it would be easier to just change it to `orElse` and `orElseM`.

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: There’s no difference in that regard, one additional inspection in `fold` in fact. But it’s a shame it’s longer, perhaps there should have been a `catchAllM` too.

adamgfraser: You're right. Aren't they the the same though? Both are implemented in terms of `foldM` and have to wrap the success or failure in an effect constructor. `fold` does both of them internally whereas `catchAll` only does the success case itself and then we would need to wrap the failure case in an effect constructor here.

regiskuckaertz: Yep, they are. You must be getting tired of hearing this, but: awesome work 😅

adamgfraser: Thanks! Lot of people including you doing awesome work on this project! Just trying to contribute and learn. 😃

PR #2450, opened 2019-12-09
Make STM stack-safe

mijicd:
The current implementation of STM suffers from stack safety issues. STM stack safety suite was able to verify these on the depth of 10000. 

This PR introduces resumable exceptions as a way to mitigate these problems. 

### Summary

- Operations use stack up to predefined number of frames, after which they're continued via `Resumable`.
- Renamed `TRez` to `TExit`.
- Introduced `STM.done`.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -231,8 +231,6 @@ final class STM[+E, +A] private[stm] (
   final def orElse[E1, A1 >: A](that: => STM[E1, A1]): STM[E1, A1] =
     new STM(
       (journal, fiberId, stackSize) => {
-        val framesCount = stackSize.incrementAndGet()
-
         val reset = prepareResetJournal(journal)
 
         val continueM: TExit[E, A] => STM[E1, A1] = {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -303,10 +299,68 @@ final class STM[+E, +A] private[stm] (
...
+      }
+    }
+
+    result.asInstanceOf[TExit[E, A]]
+  }
 }
 
 object STM {
 
+  private final class Resumable[E, E1, A, B](val stm: STM[E, A], val ks: Stack[internal.TExit[E, A] => STM[E1, B]])
```
mschuwalow: Should we use https://github.com/scala/scala/blob/2.13.x/src/library/scala/util/control/ControlThrowable.scala for this?

mijicd: I wouldn't go for it. Its representation has been changed in 2.13, from trait inheriting `NoStackTrace` to an abstract class, and according to the docs, it seems that its purpose is also changed a bit. 

Related to this, the reason to avoid `NoStackTrace` in the first place was the possibility to override the suppression via system properties.

adamgfraser:
Very interesting! Why did you choose to use this approach versus a more traditional trampoline based solution stack safety.

core/shared/src/main/scala/zio/stm/STM.scala
mschuwalow: makes sense 👍

mijicd:
@adamgfraser Mainly for the sake of performance. That being said, here's the output of flatMap benchmark:

```
Result "zio.stm.STMFlatMapBenchmark.deepFlatMap":
   1139.607 ±(99.9%) 1.670 ops/s [Average]
   (min, avg, max) = (1137.734, 1139.607, 1142.280), stdev = 1.562
   CI (99.9%): [1137.937, 1141.278] (assumes normal distribution)

Benchmark                                          (depth)   Mode  Cnt     Score   Error  Units
STMFlatMapBenchmark.deepFlatMap       20    thrpt    15  1139.607 ± 1.670  ops/s
```

Comparing it to the results reported in #2457, it shows the slowdown of 1.47 times.

jdegoes:
Awesome job! A new era for STM in Scala.

PR #2461, opened 2019-12-10
ZSink.collectAllToMap/N now are able to merge values

### LINKED ISSUE #2454: ZSink.collectAllToMap/N should merge values
Instead of:
```scala
  final def collectAllToMap[K, A](key: A => K): ZSink[Any, Nothing, Nothing, A, Map[K, A]]
```

Should be:
```scala
  final def collectAllToMap[K, A](key: A => K)(f: (A, A) => A): ZSink[Any, Nothing, Nothing, A, Map[K, A]]
```

saraiva132:
Unlike before, collectAllToMap now receives a merging function. This means that now collectAllToMapN keeps consuming elements discarding those who would see its map grow over the size limit.

In this context I changed max size constraint on Stream#range in order to make it behave similarly to standard library Range.

closes #2454

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1236,23 +1236,34 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
   /**
    * Creates a sink accumulating incoming values into a map.
    * Key of each element is determined by supplied function.
+   * Combines elements with same key with supplied function f.
+   *
    */
-  final def collectAllToMap[K, A](key: A => K): ZSink[Any, Nothing, Nothing, A, Map[K, A]] =
-    foldLeft[A, Map[K, A]](Map.empty[K, A])((map, element) => map + (key(element) -> element))
+  final def collectAllToMap[K, A](key: A => K)(f: (A, A) => A): Sink[Nothing, Nothing, A, Map[K, A]] =
+    foldLeft[A, Map[K, A]](Map.empty)((curMap, a) => {
+      val k = key(a)
+      curMap.get(k).fold(curMap.updated(k, a))(v => curMap.updated(k, f(v, a)))
+    })
 
   /**
    * Creates a sink accumulating incoming values into a map of maximum size `n`.
    * Key of each element is determined by supplied function.
+   *
+   * Combines elements with same key with supplied function f.
    */
-  final def collectAllToMapN[K, A](n: Long)(key: A => K): Sink[Nothing, A, A, Map[K, A]] = {
-    type State = (Map[K, A], Boolean)
-    def f(state: State, a: A): (State, Chunk[A]) = {
-      val newMap = state._1 + (key(a) -> a)
-      if (newMap.size > n) ((state._1, false), Chunk.single(a))
-      else if (newMap.size == n) ((newMap, false), Chunk.empty)
-      else ((newMap, true), Chunk.empty)
+  final def collectAllToMapN[K, A](n: Long)(key: A => K)(f: (A, A) => A): Sink[Nothing, A, A, Map[K, A]] = {
+
+    def inner(curMap: Map[K, A], a: A): (Map[K, A], Chunk[A]) = {
+      val k = key(a)
+      curMap
+        .get(k)
+        .fold(
+          if (curMap.size >= n) (curMap, Chunk.single(a))
+          else (curMap.updated(k, a), Chunk.empty)
+        )(v => (curMap.updated(k, f(v, a)), Chunk.empty))
+
     }
-    fold[A, A, State]((Map.empty, true))(_._2)(f).map(_._1)
+    fold[A, A, Map[K, A]](Map.empty)(_ => true)(inner)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1227,24 +1227,30 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
   /**
    * Creates a sink accumulating incoming values into a map.
    * Key of each element is determined by supplied function.
+   * Combines elements with same key with supplied function f.
+   *
    */
-  final def collectAllToMap[K, A](key: A => K): ZSink[Any, Nothing, Nothing, A, Map[K, A]] =
-    foldLeft[A, Map[K, A]](Map.empty[K, A])((map, element) => map + (key(element) -> element))
+  final def collectAllToMap[K, A](key: A => K)(f: (A, A) => A): ZSink[Any, Nothing, Nothing, A, Map[K, A]] =
```
vasilmkd: Because this and `collectAllToMapN` are now implemented in terms of `foldLeft`, which is a Sink that has no leftover, please change the leftover of these two sinks from `A` to `Nothing` to better document them.

saraiva132: Good catch. thanks @vasilmkd

saraiva132:
I noticed that there is no method returning Sink in ZSink. Would it be a good practice to have methods that do not require environment to just use shorter Sink type instead of ZSink? This could be applied to other types

vasilmkd:
I'd say definitely go for it, but maybe as a separate PR.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3167,7 +3167,7 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
    * Constructs a stream from a range of integers (inclusive).
```
nicolasdasriaux: Change comment from
```
Constructs a stream from a range of integers (inclusive)
```
to suggested
```
Constructs a stream from a range of integers (lower bound included, upper bound not included)
```

saraiva132: Good one. It's good to make that clear, I will update

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1227,24 +1227,30 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
...
+   *
+   * Combines elements with same key with supplied function f.
+   */
+  final def collectAllToMapN[K, A](n: Long)(key: A => K)(f: (A, A) => A): Sink[Nothing, Nothing, A, Map[K, A]] =
+    foldLeft[A, Map[K, A]](Map.empty)((curMap, a) => {
+      val k = key(a)
+      curMap
+        .get(k)
+        .fold(
+          if (curMap.size >= n) curMap else curMap.updated(k, a)
```
iravid: We're losing elements here. If we decide not to merge an element into the map, we need to push it back as a leftover.

iravid:
Thank you for picking this up @saraiva132!

streams/shared/src/main/scala/zio/stream/ZSink.scala
vasilmkd: I wonder how I missed this. I thought it was a bit strange that there are no leftovers. :thinking:

saraiva132: Yeah. The leftover was a bit confusing to me. Until I watched your talk on functional scala. Now I understand it! I went DOH when I read your comment. So, question now, How can I accumulate the leftovers? There is no run method that has that ability AFAIK. I would like to also make those reflected in tests. Also, I fail to understand why fold method has a Chunk[A] has the second type of tuple. Why not A and internally those would be accumulated into Chunk[A]? or Chunk[A0] rather? Sorry for the amount of questions. cc @iravid

iravid:
@saraiva132 One last change - can you please split the `range` change to a separate PR? It's a change that I would like to prominently include in the changelog.

saraiva132:
@iravid np then let us merge the other one first and then I will rebase to this one.

iravid:
Can you rebase this @saraiva132?

saraiva132:
@iravid here you go

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1236,23 +1236,34 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
...
-      else if (newMap.size == n) ((newMap, false), Chunk.empty)
-      else ((newMap, true), Chunk.empty)
+  final def collectAllToMapN[K, A](n: Long)(key: A => K)(f: (A, A) => A): Sink[Nothing, A, A, Map[K, A]] = {
+
+    def inner(curMap: Map[K, A], a: A): (Map[K, A], Chunk[A]) = {
+      val k = key(a)
+      curMap
+        .get(k)
+        .fold(
+          if (curMap.size >= n) (curMap, Chunk.single(a))
```
iravid: I think you need to signal that you don't want to continue here in the same way the previous implementation did. Otherwise, this probably won't work right with `ZStream#aggregate`.

Best add a test or two for verifying that behavior.

saraiva132:
I misunderstood how the implementation should be. I think this looks good now. @iravid let me know if more tests are required. Also, let me know if allocating the key and current map in the inner function is acceptable.

Issue #2462, opened 2019-12-10
ZIO Core: Make all effect constructors lazy in all parameters


jdegoes:
All effect constructors (`succeed`, etc.), including those in modules like `Console`, should utilize by-name parameters. The rationale is that people accidentally put side-effects into parameters, and it's better to attempt to make these side-effects lazy than result in confusing behavior.

For pure code, it makes no difference anyway.

```scala
for {
  _ <- blocking(putStrLn(Thread.currentThread().getName))
} yield ()
```

An effect constructor is any method appearing in any companion object that takes simple parameters and returns an effect (`ZIO`, `ZStream`, `ZSink`, `STM`, etc.).

saraiva132:
I can do this

adamgfraser:
Awesome!

saraiva132:
Should a parameter of type `Iterable[ZIO[R1, E, A]]` also be by-name ?

saraiva132:
Effects that get by-name parameters:

- [ ] IO
- [ ] ZIO
- [ ] UIO
- [ ] RIO
- [ ] URIO
- [ ] TASK
- [ ] ZManaged
- [ ] Managed
- [ ] ZStream
- [ ] ZStreamChunk
- [ ] ZSink
- [ ] Pull (ZIO alias)

Methods (Not all effects have all of these methods):

- succeed
- fail
- die/ dieMessage
- effect/effectTotal  
- left/right/some
- fromTry/fromEither/fromOption/ fromFiber.
- halt
- done
- interruptAs
- sleep

Eager variant will add the sufix `Now`. Example `succeedNow`

## Service Modules

-  **Clock**: currentTime and sleep
-  **Console**: putStr and putStrLn
-  **Random**: nextBytes, nextInt, nextLong and nextString
-  **System**: env and property

adamgfraser:
I think this generally seems like a solid plan. A couple of thoughts:

1. We probably don't need `now` variants of all of these methods but only ones that are used in internal combinators with some regularity.
2. I think there are some other combinators that we need to make lazy. `when` in particular has generated several questions.

I would suggest we open a PR with initially a relatively narrow focus, say just `ZIO.scala` as we did with the previous work on `final` methods to see what this looks like in practice and then once we have alignment we can move on to the rest of the codebase.

saraiva132:
[...]

PR #2463, opened 2019-12-11
Make method arguments that are pure values lazy (by-name)

saraiva132:
Initial work towards closing #2462.

I would like to know if I am taking things too far. @jdegoes @adamgfraser 

I do feel I mistook the meaning of effect constructor,


adamgfraser:
@saraiva132 I think I would pull this back a bit. Think of constructors as only methods that exist on the companion object, not the trait, and that don't really do anything more than "lift" the specified value into an effect. So methods like `succeed`, `some`, `fail`, `halt`, and `die` would be constructors that we would want to make lazy, as well as methods like `console.putStrLn`, but `foreachParN` we would not.

saraiva132:
I still need to take a closer look at all of this one, maybe two times. 3 tests were broken for some still unknown reason. After taking a look at a good portion of the codebase I realize there is no fundamental consensus on what is by-name or not. That means that some methods have by-name parameters and others do not, in a seemingly arbitrary fashion, at times. I am trying to normalize it and also clean up some stuff like adding override modifiers or final to methods, while avoiding going overboard.

adamgfraser:
Thanks for diving in and cleaning this up! You are definitely right that we have not been consistent about this in the past. Maybe it would be helpful if you could articulate the principle you are using for deciding which methods should be lazy and potentially a list of them for `ZIO` so we can make sure we are aligned before you do a bunch of work based on that. Then it should be pretty easy to carry that over to other data types.

In particular I noticed that you are making constructors for non-effect types like `Cause` or `Chunk` lazy. These are pure data types so I don't think we need their constructors to be lazy.

saraiva132:
To make things more clear and based on the insights I obtained from what I did so far I would say there are a couple of inconsistencies with the following:

- By name parameters;
- Final methods;
- Overrided methods;
- Final case classes;
- Classes / traits extending AnyVal, Serializable, Product;
- Private definitions in: class / constructor / parameter.

I would like to perhaps get some comments on each and also the rational to why one would make a class final, extends AnyVal, etc etc and perhaps reach a consensus for all this bullet points and clean all of this stuff up, potentially, in a new issue with further PRS along the way.

##Methods that lift pure values to effects

Regarding proposal for **by-name parameters**. It is tricky how far one might take this. This should be any method that takes a pure value and lifts it into an effect. So, in that sense, I do agree that data types like `Exit` and `Cause` should not have any **by-name parameters**. Although, if we follow this rule, then there is some clean-up to do, and I can refer this later on, but there are quite some **by-name parameters** in methods that, if we were to apply this rule, do not make sense.

Effects that should have parameters by-name:
- IO
- ZIO
- UIO
- RIO
- URIO
- TASK
- STM
- ZMANAGED
- MANAGED

Methods:
- Succeed
- Fail
- Die
- Effect
- DieMessage
- Left
- Right
- Some
- fromTry/fromEither/fromOption. **These were already by-name. Why?? Why not halt, for instance? Weird.**

##Methods that proxy pure values to effect lifting method.

@adamgfraser @jdegoes, what about method inside class that take a pure value and return a lifted value by proxying to a ZIO method. 

Example: Promise#succeed

` final def succeed(a: => A): UIO[Boolean] = completeWith(IO.succeed(a))`
or 
` final def succeed(a: A): UIO[Boolean] = completeWith(IO.succeed(a))`

Data types that proxy lifting methods:

- Ref
- RefM
- Promise

##Methods and classes that hold pure values

Last but not least there are data types that hold values that could potentially have side-effects and could be lazily evaluated as well.

Example: Chunk#single

` final def single[A](a: => A): Chunk[A] = Singleton(a)` 
` private case class Singleton[A](a: => A) extends Chunk[A]`

or

`  final def single[A](a: A): Chunk[A] = Singleton(a)` 
`  private case class Singleton[A](a: A) extends Chunk[A]`

Data types that wrap pure values with potential side-effects:

- Chunk
- Cause
- Exit

##Service modules

- **Clock**: `currentTime` and `sleep`
- **Console**: `putStr` and `putStrLn`
- **Random**: `nextBytes`, `nextInt`, `nextLong` and `nextString`
- **System**: `env` and `property`

adamgfraser:
Thanks! This is very helpful. I would say let's scope this PR to "methods that lift pure values to effects" and "service modules".

For "methods that proxy pure values to effect lifting methods" this is definitely something we should look at but let's tackle it separately as part of making sure that combinators on effect traits / classes are appropriately lazy.

For "methods and classes that hold pure values" I don't think there is any value in making these constructors lazy. Even if we made the constructors themselves lazy the data type itself is not lazy so the value would still be evaluated immediately. To change this we would have to change the data types to be lazy (e.g. `final case class Singleton[A](value: () => A) extends Chunk[A]`) which would tank performance. More broadly, these are pure data types so I don't think there can be any expectation that they would manage effects for you or well-defined semantics of what that would mean.

Going back to the first two categories, all of the methods you listed seem like good examples of things that should be by name but is that list comprehensive? For example I didn't see `effectTotal` or `halt` on there.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -92,7 +92,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self.sandbox
       .foldM(
         cause => ZIO.fail(cause.squashWith(f)),
-        ZIO.succeed
+        ZIO.succeed(_)
```
regiskuckaertz: Were these breaking compilation? If not, please revert all those changes to minimise the diff.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -418,7 +418,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * yielding the value of the first effect to succeed with a value.
    * Losers of the race will be interrupted immediately
    */
-  def firstSuccessOf[R1 <: R, E1 >: E, A1 >: A](rest: Iterable[ZIO[R1, E1, A1]]): ZIO[R1, E1, A1] =
+  final def firstSuccessOf[R1 <: R, E1 >: E, A1 >: A](rest: Iterable[ZIO[R1, E1, A1]]): ZIO[R1, E1, A1] =
```
regiskuckaertz: Same here, let's focus on by-name parameters only, as it is already quite controversial.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I think unfortunately these are necessary because a by name function is different than a by value one to the compiler, so we need to explicitly apply the function.

adamgfraser:
The other thing we need to be careful of here is that the bodies of these functions are also lazy. Making the parameter lazy doesn't do anything if the body needs to immediately evaluate the value. For example:

```scala
def succeed[A](value: => A): ZIO[Any, Nothing, A] =
  ZIO.Succeed(value)

// Not actually lazy because value will be immediately evaluated by constructor
```

To make this lazy we essentially have to make `succeed` an alias for `effectTotal`:

```scala
def succeed[A](value: => A): ZIO[Any, Nothing, A] =
  effectTotal(value)
```

Stepping back, I am wondering whether we need to draw a distinction between constructors that are meant to just be "smart constructors" for certain cases of the `ZIO` algebraic data type versus constructors for importing code into `ZIO`. For example, `fromEither` or `fromOption` would be lazy under this conceptualization versus `succeed` would continue to be eager.

I am worried that we have a lot of uses of say `ZIO.succeed` in various internal methods that are just trying to lift a value into an effect and I don't know if we want to make them all create a new `effectTotal` constructor.

This would on the margin increase the risk that unsafe code would not be imported into and managed by `ZIO`, but perhaps we could encourage more use of the `ZIO` apply constructor for this. One thing that might help along these lines is if we changed the signature for the constructor from:

```scala
def apply[A](a: => A): Task[A]
```

to:

```scala
def apply[A](a: => A): ZIO[Any, Nothing, A]
```

The current constructor is a little irregular in that it returns a `ZIO` with error type fixed to throwable whereas `ZIO` is polymorphic in the error type. I realize the justification is that `Throwable` is the error type that any arbitrary Scala code could throw, but `Task.apply` is already available for this and it forces users to use `effectTotal` for importing arbitrary values into a `ZIO`, which is a long name for something we want to encourage as the default behavior.

core/shared/src/main/scala/zio/ZIO.scala
saraiva132: Yeah, makes sense @regiskuckaertz. This will need be properly split into several PRs focusing and addressing different issues

saraiva132:
I believe this is ready for a review @adamgfraser @regiskuckaertz . To go over the changes made:

Effects that get parameterss by-name:

- IO
- ZIO
- UIO
- RIO
- URIO
- TASK
- ZManaged
- Managed
- ZStream
- ZStreamChunk
- ZSink
- Pull

Methods (Not all effects have all of these methods):

- succeed
- fail
- die/ dieMessage
- effect/effectTotal  
- left/right/some
- fromTry/fromEither/fromOption/ fromFiber.
- halt
- done
- interruptAs
- sleep

## Service Modules

-  **Clock**: currentTime and sleep
-  **Console**: putStr and putStrLn
-  **Random**: nextBytes, nextInt, nextLong and nextString
-  **System**: env and property

## Next work

- Tackle STM. Actually, it seems that making, I believe succeed, parameter by-name breaks tests in a non-obvious way. Interestingly. I will tackle that in a different PR to decrease scope. 
- Normalize final functions
- Normalize overrides 
- Normalize private constructors && private fields

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -476,13 +476,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * A more powerful version of `fold` that allows recovering from any kind of failure except interruptions.
    */
-  final def foldCause[B](failure: Cause[E] => B, success: A => B): URIO[R, B] =
+  def foldCause[B](failure: Cause[E] => B, success: A => B): URIO[R, B] =
```
regiskuckaertz: Was this intentional?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -476,13 +476,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
     foldCauseM(new ZIO.MapFn(failure), new ZIO.MapFn(success))
 
   /**
    * A more powerful version of `foldM` that allows recovering from any kind of failure except interruptions.
    */
-  def foldCauseM[R1 <: R, E2, B](failure: Cause[E] => ZIO[R1, E2, B], success: A => ZIO[R1, E2, B]): ZIO[R1, E2, B] =
+  final def foldCauseM[R1 <: R, E2, B](
```
regiskuckaertz: Same here?

core/shared/src/main/scala/zio/console/Console.scala
```diff
@@ -52,7 +52,7 @@ object Console extends Serializable {
       /**
        * Prints a line of text to the console, including a newline character.
        */
-      final def putStrLn(line: String): UIO[Unit] =
+      final def putStrLn(line: => String): UIO[Unit] =
         putStrLn(SConsole.out)(line)
 
       final def putStrLn(stream: PrintStream)(line: String): UIO[Unit] =
```
regiskuckaertz: What about this one?

core/shared/src/main/scala/zio/console/Console.scala
```diff
@@ -41,7 +41,7 @@ object Console extends Serializable {
       /**
        * Prints text to the console.
        */
-      final def putStr(line: String): UIO[Unit] =
+      final def putStr(line: => String): UIO[Unit] =
         putStr(SConsole.out)(line)
 
       final def putStr(stream: PrintStream)(line: String): UIO[Unit] =
```
regiskuckaertz: And this one.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2626,16 +2629,16 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
...
-    def dieMessage(m: String): Pull[Any, Nothing, Nothing]       = UIO.dieMessage(m)
-    def done[E, A](e: Exit[E, A]): Pull[Any, E, A]               = IO.done(e.mapError(Some(_)))
-    def fromPromise[E, A](p: Promise[E, A]): Pull[Any, E, A]     = p.await.mapError(Some(_))
+    def emit[A](a: => A): Pull[Any, Nothing, A]                  = UIO.succeed(a)
+    def fail[E](e: => E): Pull[Any, E, Nothing]                  = IO.fail(Some(e))
+    def halt[E](c: => Cause[E]): Pull[Any, E, Nothing]           = IO.halt(c.map(Some(_)))
+    def die(t: => Throwable): Pull[Any, Nothing, Nothing]        = UIO.die(t)
+    def dieMessage(m: => String): Pull[Any, Nothing, Nothing]    = UIO.dieMessage(m)
+    def done[E, A](e: => Exit[E, A]): Pull[Any, E, A]            = IO.done(e.mapError(Some(_)))
+    def fromPromise[E, A](p: => Promise[E, A]): Pull[Any, E, A]  = p.await.mapError(Some(_))
```
regiskuckaertz: Not sure this one should be lazy.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2626,16 +2629,16 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
...
     def fromEffect[R, E, A](effect: ZIO[R, E, A]): Pull[R, E, A] = effect.mapError(Some(_))
 
-    def fromTake[E, A](take: Take[E, A]): Pull[Any, E, A] =
+    def fromTake[E, A](take: => Take[E, A]): Pull[Any, E, A] =
```
regiskuckaertz: Same here, these are safe imo.

saraiva132: Thanks. You are right

core/shared/src/main/scala/zio/console/Console.scala
saraiva132: I believe this should be by-name.

core/shared/src/main/scala/zio/console/Console.scala
saraiva132: I believe this should be by-name.

core/shared/src/main/scala/zio/ZIO.scala
saraiva132: This was a typo. I will correct.

core/shared/src/main/scala/zio/ZIO.scala
saraiva132: This was a typo. I will correct.

streams/shared/src/main/scala/zio/stream/ZStream.scala
saraiva132: I believe this should be. I do require a second opinion.

adamgfraser:
@jdegoes Looking at the scope of changes here I am wondering if we need two versions of some of these methods: (1) strict for use in internal combinators and possibly by "power users" and (2) lazy for everyone else.

Making everything lazy basically requires wrapping everything in an `effectTotal` constructor. We frequently use methods like `succeed` or `fail` to just lift a value we know is pure into an effect or implement one method in terms of another. And these methods can get called a lot in something like a stream. So I worry that if we just make everything lazy we're setting ourselves up for another large PR to reimplement the internal versions with a strict combinator, though I haven't done any benchmarking on this.

Should we come up with some naming convention (e.g. `succeed` and `succeedNow`) that we can use so we have both lazy methods for everyday users and strict methods for us or users who are implementing their own combinators and want to be able to do the same thing? Or am I off base with the performance concern?

saraiva132:
My conclusion after working on this is exactly this as well. Wrapping things in effectTotal is meaningless and adds overhead. The best approach seems to make the data types Succeed, fail, etc take by-name (Function0) parameters by default and if we want (we probably do) to keep the ability to eagerly evaluate side-effect free code we can have both options.

regiskuckaertz:
> Should we come up with some naming convention

We used to have `succeed` and `succeedLazy`. I keep looking at this PR and it almost feels like a step back, though I really understand the issue and agree usability sometimes trumps economy of features.

adamgfraser:
That's helpful context. I think the challenge is that if we offer two ways to do something some people are going to do something in the "wrong" way and get unexpected results. So we either just offer the lazy way or we need to strongly carrot people towards doing it the safe way.

Maybe in retrospect what was wrong with `succeed` and `succeedLazy` was that the default was off. `succeed` was strict and you had to do something different and longer to get lazy behavior instead of `succeed` being lazy and you have to do something different to get strict behavior.

reibitto:
I'm also of the opinion that the shorter obvious name should be the "safer" one. Which is why if it's decided that there are going to be 2 versions (for performance reasons or whatever it may be), I think I would prefer succeed/succeedNow over succeed/succeedLazy.

regiskuckaertz:
I think you are both right 👍

saraiva132:
Thanks for your input everyone. I will try to make a proposal today. 🙏

I like the now variants. And this makes it worthwhile for the lazy variants to wrap the class construction with effectTotal. I hope this does not penalizes performance too harshly.

saraiva132:
I will tackle this in separate PRs to minimize impact.

PR #2464, opened 2019-12-11
ZIO Test: Bring Back Type Checked Assertion#equalTo

adamgfraser:
We previously moved to universal equality for `Assertion.equalTo` to avoid type inference issues. However, as time has gone on it has become apparent that this makes it very easy for people to write equality assertions that don't make sense, for example comparing an effect to the expected result of the effect.

This PR brings back type checked equality and solves the type inference issues by currying `assert` and introducing new implicit evidence that two types can potentially be equal, which is always available if either type is a subtype of the other. So for example:

```scala
trait Animal
trait Cat extends Animal
trait Dog extends Animal

val animal = new Animal {}
val cat = new Cat {}
cal dog = new Dog {}

assert(animal)(equalTo(cat)) // okay
assert(cat)(equalTo(animal)) // okay
assert(dog)(equalTo(cat))    // does not compile
```

This works on Scala 2. On Dotty equality continues to be universal. I tried to use the equality type class in Dotty but ran into some issues with types getting inferred as `Nothing` when assertions were composed with combinators like `||` and `&&`, so I think we will have to handle that in a subsequent PR.

The old version of `assert` is deprecated and if we merge this I will open an issue to migrate the existing tests to the new format.


test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -109,7 +111,7 @@ package object test extends CompileVariants {
   /**
    * Checks the assertion holds for the given value.
    */
-  final def assert[A](value: => A, assertion: Assertion[A]): TestResult =
+  final def assert[A](value: => A)(assertion: Assertion[A]): TestResult =
```
regiskuckaertz: Coming back to this, it's not part of the type inference issue, is it? I was just wondering if it made the API more usable.

adamgfraser: Actually it is related to the type inference issue. We need to curry `assert` to "lock in" in the type of `A` so that the compiler properly widens the right hand side if it is a subtype of the left hand side, for example `assert(List(1, 2, 3))(equalTo(List.empty))`. And then we need the implicit evidence to prevent it from widening the type to `Any` so nonsensical comparisons don't compile.

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -73,11 +73,12 @@ object AssertionSpec extends ZIOBaseSpec {
...
     } @@ failure,
     test("equalTo must not have type inference issues") {
-      assert(List(1, 2, 3).filter(_ => false), equalTo(List.empty))
+      assert(List(1, 2, 3).filter(_ => false))(equalTo(List.empty))
     },
-    test("equalTo must fail when comparing two unrelated types") {
-      assert(1, equalTo("abc"))
-    } @@ failure,
+    testM("equalTo must not compile when comparing two unrelated types") {
+      val result = typeCheck("assert(1, equalTo(\"abc\"))")
```
kusamakura: Pardon me for commenting here, but shouldn't this use the curried `assert` as well?

test/shared/src/main/scala/zio/test/package.scala
regiskuckaertz: Brilliant!

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
adamgfraser: No pardon necessary, and thank you for the good catch! Fixed now.

jdegoes:
@adamgfraser Awesome work! Addresses a known pain point. 

So this is known to not work in Scala 3, correct?

adamgfraser:
@jdegoes Correct. All code will correctly compile on Dotty but assertions like `assert("Adam)(equalTo(1))` which are technically sound but don't make sense will also compile.

adamgfraser:
@iravid Per your comment, type inference is much better for `hasField` with the new curried `assert` because the first parameter gives the compiler information about the type of the case class. Updated the tests to reflect this.

regiskuckaertz:
@adamgfraser looks like you need to rebase. Just curious, why wouldn't the `given` machinery work with `Refl`?

test/shared/src/main/scala-2.x/zio/test/Eql.scala
```diff
@@ -0,0 +1,21 @@
+package zio.test
```
jdegoes: Apache 2 / copyright license headers.

test/shared/src/main/scala-dotty/zio/test/AssertionVariants.scala
```diff
@@ -0,0 +1,33 @@
...
+package zio.test
+
+import zio.test.Assertion.Render._
+
+trait AssertionVariants {
+
+  /**
+   * Makes a new assertion that requires a value equal the specified value.
+   */
+  final def equalTo[A](expected: A): Assertion[A] =
```
jdegoes: Can we use the new `Eql` in Dotty?

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -127,17 +129,39 @@ package object test extends CompileVariants {
       loop(fragment, FailureDetails(::(AssertionValue(assertion, value), Nil)))
     }
 
+  /**
+   * Checks the assertion holds for the given value.
+   */
+  @deprecated("use curried version", "1.0.0")
```
jdegoes:
```suggestion
  @deprecated("To benefit from much better type inference and type safety, we recommend that you use the curried version of assert, which takes two parameter lists instead of one: assert(value)(assertion)", "1.0.0")
```

test/shared/src/main/scala-dotty/zio/test/AssertionVariants.scala
adamgfraser: Unfortunately not, or at least not right now. We have an issue with the type of assertions composed with combinators such as `||` being inferred as `Nothing` when we use the same technique. See minimized example and link below. This looked like a type checking issue but I wanted to look at it a little more before opening an issue with the Dotty team.

```scala
object Example extends App {
  
  def assert[A](value: A)(assertion: Assertion[A]): Boolean =
    assertion.run(value)
  
  trait Assertion[-A] { self =>
    def run(a: A): Boolean
    def ||[A1 <: A](that: Assertion[A1]): Assertion[A1] =
      a => self.run(a) && that.run(a)
  }
  
  def equalTo[A, B](expected: A)(implicit eq: Eql[A, B]): Assertion[B] =
    _ == expected
  
  def isGreaterThan(a: Int): Assertion[Int] =
    _ > a
  
  // okay
  assert(List(1, 2, 3))(equalTo(List.empty))
  
  // does not compile
  // values of type Int and String cannot be compared with ==  or != 
  // assert("Adam")(equalTo(1))
  
  // does not compile
  // expected Assertion[Int] but found Assertion[Nothing]
  // assert(10)(equalTo(5) || isGreaterThan(9))
}
```

https://scastie.scala-lang.org/p41x4qlvQmiXbxyJKk7K9w

jdegoes: I thought it was the `A1 <: A` that's forcing it to be `Nothing`, but changing that has no effect. Interesting...

We can make "too much" compile but still use `Eql` by adding a type bound:

```scala
  def equalTo[A, B >: A](expected: A)(implicit eq: Eql[A, B]): Assertion[B] =
    _ == expected
```

I'm not sure if using `Eql` would have any practical benefits in the short-term so maybe it's not worth it. Thoughts?

adamgfraser: I noticed the same thing too and initially rejected it because it would allow `assert("Adam")(equalTo(1))` to compile so it didn't seem like it was buying us anything. But thinking about it more I think it would allow us to at least unify the signature of `equalTo` and move it back into `Assertion.scala`, removing the `AssertionVariants` files, which would be nice.

adamgfraser: Ah, with the type bound on Scala 2 comparison's that don't make sense compile, so we can't do that. I don't think we really get any benefit out of using `Eql` in Dotty until we resolve the compilation issue but am fine either way.

jdegoes: OK, no worries

adamgfraser:
Opened lampepfl/dotty#7791 regarding the compilation issue on Dotty.

test-tests/shared/src/test/scala/zio/test/mock/ExpectationSpec.scala
```diff
@@ -301,8 +301,8 @@ object ExpectationSpec extends ZIOBaseSpec {
         equalTo(
           UnmetExpectationsException(
             List(
-              Module.command -> equalTo(2),
-              Module.command -> equalTo(3)
+              Module.command -> (equalTo(2): Assertion[Any]),
```
saraiva132: Why is it necessary to explicitly pass `Assertion[Any]` here? And why `Any`?

saraiva132:
This is going to be really,really, nice to have. Great work @adamgfraser

test-tests/shared/src/test/scala/zio/test/mock/ExpectationSpec.scala
adamgfraser: After looking at this more it appears that a couple of the mock method type signatures were incorrect. The assertion is an assertion about the input, not the output. Good catch!

PR #2467, opened 2019-12-13
Fix bug in ZStream#drainFork

### LINKED ISSUE #2466: StreamDrainFork test is flaky

iravid:
Fixes #2466.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1692,7 +1692,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
         done <- Ref.make(false).toManaged_
         pull = done.get flatMap {
           if (_) Pull.end
-          else as.raceAttempt(p.await.mapError(Some(_)).ensuring(done set true) *> Pull.end)
+          else
+            as.raceAttempt(
+              p.await
+                .mapError(Some(_))
+                .foldCauseM(
+                  c => done.set(true) *> ZIO.halt(c),
+                  _ => done.set(true) *> Pull.end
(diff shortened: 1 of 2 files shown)


iravid:
@vasilmkd Could you have a look?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1692,7 +1692,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
         done <- Ref.make(false).toManaged_
         pull = done.get flatMap {
           if (_) Pull.end
-          else as.raceAttempt(p.await.mapError(Some(_)).ensuring(done set true) *> Pull.end)
+          else
+            as.raceAttempt(
```
regiskuckaertz: Hmm, should you set `done` to `true` if `as` completes successfully?

regiskuckaertz: Ah, I misread the bracketing. Looks great!

iravid: Yeah we only toggle done if the interrupting promise resolves. Thanks for looking!

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1692,7 +1692,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+              p.await
+                .mapError(Some(_))
+                .foldCauseM(
+                  c => done.set(true) *> ZIO.halt(c),
```
vasilmkd: `Pull.halt`

PR #2472, opened 2019-12-14
add ZManaged#preallocateManaged

mschuwalow:
/cc @iravid

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1441,7 +1441,7 @@ object ZManaged {
    */
   final def scope[R]: ZManaged[R, Nothing, Preallocate] =
     ZManaged {
-      Ref.make(Map.empty[Int, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
+      Ref.make(Map.empty[Long, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
         Ref.make(-1L).map(_.update(_ + 1L)).map { nextIndex =>
           Reservation(
             acquire = ZIO.succeed {
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -585,6 +585,23 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
     }
 
   /**
+   * Preallocates the managed resource inside an outer managed, resulting in a ZManaged that reserves and acquires immediately and cannot fail.
+   */
+  final def preallocateManaged: ZManaged[R, E, Managed[Nothing, A]] =
+    ZManaged.finalizerRef[R](_ => ZIO.unit).mapM { finalizer =>
+      ZIO.uninterruptibleMask { restore =>
+        for {
+          env      <- ZIO.environment[R]
+          res      <- reserve
+          _        <- finalizer.set(res.release)
+          resource <- restore(res.acquire)
+        } yield ZManaged.make(ZIO.succeed(resource))(
+          _ => res.release(Exit.Success(resource)).provide(env) *> finalizer.set(_ => ZIO.unit)
+        )
+      }
+    }
+
+  /**
    * Provides the `ZManaged` effect with its required environment, which eliminates
    * its dependency on `R`.
    */
@@ -1422,7 +1439,7 @@ object ZManaged {
   /**
    * Creates a scope in which resources can be safely preallocated.
    */
-  final def preallocateManaged[R, E, A]: ZManaged[R, Nothing, Preallocate] =
+  final def scope[R]: ZManaged[R, Nothing, Preallocate] =
     ZManaged {
       Ref.make(Map.empty[Int, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
         Ref.make(-1).map(_.update(_ + 1)).map { nextIndex =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -572,6 +572,8 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
 
   /**
    * Preallocates the managed resource, resulting in a ZManaged that reserves and acquires immediately and cannot fail.
+   * You should take care that you are not interrupted between running preallocate and actually acquiring the resource
+   * as you might leak otherwise.
    */
   final def preallocate: ZIO[R, E, Managed[Nothing, A]] =
```
adamgfraser: Do we need to expose this as part of the API? It would be safer if we could only preallocate resources in the context of `preallocateManaged` so we couldn't leak them, or at least it would be much more difficult to do so. But maybe that doesn't give you the power to do what you want with this.

mschuwalow: I can see your point. To use this safely you would almost always need to use uninterruptible / ensuring which is not something we would want to expose to users. On the other hand this will be often more convenient than preallocateManaged.

Maybe lets wait for a few more opinions and then decide what to do with it.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -926,6 +945,10 @@ object ZManaged {
       ZManaged.environment.flatMap(f)
   }
 
+  trait Preallocate {
```
iravid: How about:
```scala
trait Scope {
  def allocate
}
```
?

core/shared/src/main/scala/zio/ZManaged.scala
iravid: I'd love to keep `preallocate` around as it's a pretty handy combinator for tricky situations.

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: I think naming this scope is better, Preallocate name is a remnant from previous naming.
I don't know about allocate not being apply. I think this would often become more clumsy:
```scala
val resource: Managed[Nothing, Int] = ???

ZManaged.scope.use { scope =>
  scope.preallocate(resource)
} 
```

iravid: Ok, agreed about `apply` - sounds good to me.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1413,6 +1436,41 @@ object ZManaged {
...
+                  ZIO.uninterruptibleMask { restore =>
+                    for {
+                      env      <- ZIO.environment[R]
+                      res      <- managed.reserve
+                      resource <- restore(res.acquire).onError(err => res.release(Exit.Failure(err)))
+                      index    <- nextIndex
+                      release  = res.release.andThen(_.provide(env))
+                      _        <- finalizers.update(_ + (index -> release))
+                    } yield ZManaged
+                      .make(ZIO.succeed(resource))(_ => release(Exit.Success(resource)) *> finalizers.update(_ - index))
```
```suggestion
                      .make(ZIO.succeed(resource))(_ => release(Exit.Success(resource)).ensuring(finalizers.update(_ - index)))
```

to avoid double execution if a defect occurs?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1413,6 +1436,41 @@ object ZManaged {
    */
   val never: ZManaged[Any, Nothing, Nothing] = ZManaged.fromEffect(ZIO.never)
 
+  /**
+   * Creates a scope in which resources can be safely preallocated.
+   */
+  final def scope[R]: ZManaged[R, Nothing, Preallocate] =
+    ZManaged {
+      Ref.make(Map.empty[Int, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
+        Ref.make(-1).map(_.update(_ + 1)).map { nextIndex =>
```
```suggestion
        Ref.make(-1L).map(_.update(_ + 1L)).map { nextIndex =>
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1413,6 +1436,41 @@ object ZManaged {
    */
   val never: ZManaged[Any, Nothing, Nothing] = ZManaged.fromEffect(ZIO.never)
 
+  /**
+   * Creates a scope in which resources can be safely preallocated.
+   */
+  final def scope[R]: ZManaged[R, Nothing, Preallocate] =
+    ZManaged {
+      Ref.make(Map.empty[Int, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
```
```suggestion
      Ref.make(Map.empty[Long, Exit[Any, Any] => ZIO[R, Nothing, Any]]).flatMap { finalizers =>
```

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: good catch!

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: I hope no one uses this in a way that this becomes necessary :D

iravid: Haha. It's actually not that far fetched if you, for example, have a long running stream that opens and closes files in a high-throughput system.

iravid: I was almost about to suggest `Map[UUID, Finalizer]` but I thought that really is overkill.

iravid:
Ok so just `s/trait Preallocate/trait Scope` and this should be good to go.

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: we could use BigInteger :D

iravid:
Great work @mschuwalow!

PR #2473, opened 2019-12-14
benchmarks' dependencies were updated for supporting Scala 2.13

lvitaly:
- twitter's util-collection has been deleted in v19.2.0 and this dependency was replaced by util-core which is enough for benchmarks compilation (https://github.com/twitter/util/blob/develop/CHANGELOG.rst#1920)
- akka-streams dependency was updated to v2.6.1 and StreamBenchmarks were updated in accordance with it.


benchmarks/src/main/scala/zio/StreamBenchmarks.scala
```diff
@@ -4,7 +4,6 @@ import java.util.concurrent.TimeUnit
 
 import zio.stream._
 import akka.actor.ActorSystem
-import akka.stream.ActorMaterializer
```
softinio: This is not needed anymore?

lvitaly: @softinio yes, this is not needed in Akka v2.6.x.

> A default materializer is now provided out of the box. For the Java API just pass system when running streams, for Scala an implicit materializer is provided if there is an implicit ActorSystem available. This avoids leaking materializers and simplifies most stream use cases somewhat.

https://doc.akka.io/docs/akka/current/project/migration-guide-2.5.x-2.6.x.html#materializer-changes

lvitaly: @softinio but I am not sure about migration to the latest Akka version. Should I rolled Akka version beck to 2.5.26 because of ending support of Scala 2.11? What do you think?
https://doc.akka.io/docs/akka/current/project/migration-guide-2.5.x-2.6.x.html#scala-2-11-no-longer-supported

softinio: Hmmmm ..... On one hand keeping `2.11` support we are consistent as ZIO supports and builds `2.11` and then on the other hand how important is it to run 2.11 benchmarks at the expense of not being able to do this upgrade.

jdegoes:
@lvitaly Looks great! Can you update to master and fix conflicts?

jdegoes:
@lvitaly Looks like 2.11 tests failed, can you take a look?

lvitaly:
@jdegoes all tests passed.

jdegoes:
@lvitaly Thank you!!!

PR #2474, opened 2019-12-14
Hide underlying array in TArray

mijicd:
The array was public val, therefore accessible from outside.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TArray.scala
+++ b/core/shared/src/main/scala/zio/stm/TArray.scala
@@ -35,7 +35,8 @@ final class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal
     find(pf.isDefinedAt).map(_.map(pf))
 
   /**
-   *  Finds the result of applying an STM partial function to the first value in its domain.
+   * Finds the result of applying an transactional partial function to the
+   * first value in its domain.
    */
   final def collectFirstM[E, B](pf: PartialFunction[A, STM[E, B]]): STM[E, Option[B]] =
     find(pf.isDefinedAt).flatMap {
@@ -55,7 +56,7 @@ final class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal
     fold(0)((n, a) => if (p(a)) n + 1 else n)
 
   /**
-   * Count the values in the array matching an STM predicate.
+   * Count the values in the array matching a transactional predicate.
    */
   final def countM[E](p: A => STM[E, Boolean]): STM[E, Int] =
     foldM(0)((n, a) => p(a).map(result => if (result) n + 1 else n))
@@ -66,7 +67,7 @@ final class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal
   final def exists(p: A => Boolean): STM[Nothing, Boolean] = find(p).map(_.isDefined)
 
   /**
-   * Determine if the array contains a value satisfying an STM predicate.
+   * Determine if the array contains a value satisfying a transactional predicate.
    */
   final def existsM[E](p: A => STM[E, Boolean]): STM[E, Boolean] =
     countM(p).map(_ > 0)
@@ -89,13 +90,13 @@ final class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal
     new TArray(array.reverse).find(p)
 
   /**
-   * Find the last element in the array matching an STM predicate.
+   * Find the last element in the array matching a transactional predicate.
    */
   final def findLastM[E](p: A => STM[E, Boolean]): STM[E, Option[A]] =
     new TArray(array.reverse).findM(p)
 
   /**
-   * Find the first element in the array matching an STM predicate.
+   * Find the first element in the array matching a transactional predicate.
    */
   final def findM[E](p: A => STM[E, Boolean]): STM[E, Option[A]] =
     if (array.isEmpty) STM.succeed(None)
@@ -114,14 +115,14 @@ final class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal
     if (array.isEmpty) STM.succeed(None) else array.head.get.map(Some(_))
 
   /**
-   * Atomically folds [[TArray]] with pure function.
+   * Atomically folds using a pure function.
    */
   final def fold[Z](acc: Z)(op: (Z, A) => Z): STM[Nothing, Z] =
     if (array.isEmpty) STM.succeed(acc)
     else array.head.get.flatMap(a => new TArray(array.tail).fold(op(acc, a))(op))
 
   /**
-   * Atomically folds [[TArray]] with STM function.
+   * Atomically folds using a transactional function.
    */
   final def foldM[E, Z](acc: Z)(op: (Z, A) => STM[E, Z]): STM[E, Z] =
     if (array.isEmpty) STM.succeed(acc)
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -17,138 +17,193 @@
 package zio.stm
 
 /** Wraps array of [[TRef]] and adds methods for convenience.
- * Caution: most of methods are not stack-safe.
- * */
-class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
+ *
+ */
+class TArray[A] private (private val array: Array[TRef[A]]) extends AnyVal {
```
adamgfraser: Should we make this and the other `STM` types `final`? Technically classes extending `AnyVal` are already final but could be good to clarify intent. Looks like we do this elsewhere in the code base.

mijicd: Absolutely agreed, done. Regarding the other "leakings", I'll tackle them in a separate PR(s), just to preserve the scope of this one.

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -16,139 +16,193 @@
...
       }
 
-  /** Find the last element in the array matching a predicate. */
-  final def findLast(predicate: A => Boolean): STM[Nothing, Option[A]] =
-    new TArray(array.reverse).find(predicate)
+  /**
+   * Find the last element in the array matching a predicate.
+   */
+  final def findLast(p: A => Boolean): STM[Nothing, Option[A]] =
+    new TArray(array.reverse).find(p)
```
regiskuckaertz: This is counter-intuitive, you would expect subsequent operations to be done on the same array, but here a new one is created, which could yield different results on those operations.

mijicd: This one's actually the old one (diff due to renaming). Technically, it is on the same one (TArray being value class), however, I have similar concerns on a couple more methods (e.g. `existsM`).

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -16,139 +16,193 @@
...
+          else new TArray(array.tail).findM(p)
         }
       }
 
-  /** The first entry of the array, if it exists. */
-  final def firstOption: STM[Nothing, Option[A]] = if (array.isEmpty) STM.succeed(None) else array.head.get.map(Some(_))
+  /**
+   * The first entry of the array, if it exists.
+   */
+  final def firstOption: STM[Nothing, Option[A]] =
```
regiskuckaertz: Note that we could add a `first` that just retries if the array is empty 🤔

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: Edit: I've re-read what you wrote, and yes, this is even more extreme actually.

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: Oh, I believe that was intentionally removed as we didn't want to expose "unsafe" methods (see #2446).

regiskuckaertz: Yes, my point is thanks to the semantics of STM, you can make a safe version of `first` that just parks the computation if the array is empty 😄 WDYT?

mijicd: True that. But the question is should we reflect that behavior to all accesses, e.g. non-existing index?

regiskuckaertz: hmm, good question. I wouldn't, e.g. `array(-1)` would never make sense.

mijicd: Ah, the thing is, the current API doesn't really allow elements to be added, so retrying won't make much sense.

regiskuckaertz: Ah right 👍

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -16,139 +16,193 @@
...
+    if (array.isEmpty) STM.succeed(None) else array.head.get.map(Some(_))
 
-  /** Atomically folds [[TArray]] with pure function. */
+  /**
+   * Atomically folds [[TArray]] with pure function.
```
mschuwalow: typo: a pure function

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -16,139 +16,193 @@
...
-    countM(predicate).map(_ == array.length) //Universal quantification should not be order dependent so the entire array needs to be observed.
+  /**
+   * Atomically evaluate the conjunction of an STM predicate across the members of the array.
+   */
+  final def forallM[E](p: A => STM[E, Boolean]): STM[E, Boolean] =
+    countM(p).map(_ == array.length)
 
-  /** Atomically performs side-effect for each item in array */
+  /**
+   * Atomically performs side-effect for each item in array.
```
mschuwalow: perform a transaction for each item in the array?

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -16,139 +16,193 @@
...
+   */
   final def foreach[E](f: A => STM[E, Unit]): STM[E, Unit] =
-    this.foldM(())((_, a) => f(a))
+    foldM(())((_, a) => f(a))
 
-  /** Get the first index of a specific value in the array or -1 if it does not occur. */
+  /**
+   * Get the first index of a specific value in the array or -1 if it does not occur.
+   */
   final def indexOf(a: A): STM[Nothing, Int] = indexOf(a, 0)
```
mschuwalow: Should we push this failure mode into the error channel?

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -157,55 +211,72 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+    foldM[E, Option[A]](None) { (optAcc, a) =>
+      optAcc match {
+        case Some(acc) => op(acc, a).map(Some(_))
+        case _         => STM.succeed(Some(a))
+      }
+    }
 
-  /** Atomically updates all [[TRef]]s inside this array using pure function. */
+  /**
+   * Atomically updates all elements using pure function.
```
mschuwalow: typo: a pure function

core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -157,55 +211,72 @@ class TArray[A] private (val array: Array[TRef[A]]) extends AnyVal {
...
+   */
   final def update(index: Int, fn: A => A): STM[Nothing, A] =
     if (0 <= index && index < array.length) array(index).update(fn)
     else STM.die(new ArrayIndexOutOfBoundsException(index))
 
-  /** Atomically updates element in the array with given transactional effect. */
+  /**
+   * Atomically updates element in the array with given transactional effect.
+   */
   final def updateM[E](index: Int, fn: A => STM[E, A]): STM[E, A] =
```
mschuwalow: for other parts of zio we have decided to use M suffix only for function involving the zio datatype.
How about we rename these to updateTransactional updateSTM for consistency?

core/shared/src/main/scala/zio/stm/TArray.scala
saraiva132: @mijicd This could also potentially be good first PRs. :D

mijicd: @saraiva132 I'll double-check if it makes sense to do such change, and if so, I'll make some issues!

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: Done.

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: Slightly different, but done.

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: Done.

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: I'd leave it for now. We can deprecate this in the future, assuming there's an agreement about that, although, I find the current API to be more convenient.

core/shared/src/main/scala/zio/stm/TArray.scala
mijicd: I don't think we should as the goal was for this method to be symmetric with the one in `Array`.

mschuwalow: ok, makes sense

core/shared/src/main/scala/zio/stm/TArray.scala
mschuwalow: 👍

PR #2476, opened 2019-12-15
give stm the 'z' treatment

mschuwalow:
/cc @mijicd

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
@@ -25,53 +25,55 @@ object TMapSpec extends ZIOBaseSpec {
   def spec = suite("TMap")(
     suite("factories")(
       testM("apply") {
-        val tx = TMap.make("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3).flatMap(_.toList)
+        val tx = TMap.make("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3).flatMap[Any, Nothing, List[(String, Int)]](_.toList)
         assertM(tx.commit)(hasSameElements(List("a" -> 1, "b" -> 3, "c" -> 2)))
       },
       testM("empty") {
-        val tx = TMap.empty[String, Int].flatMap(_.toList)
+        val tx = TMap.empty[String, Int].flatMap[Any, Nothing, List[(String, Int)]](_.toList)
         assertM(tx.commit)(isEmpty)
       },
       testM("fromIterable") {
-        val tx = TMap.fromIterable(List("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3)).flatMap(_.toList)
+        val tx = TMap
+          .fromIterable(List("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3))
+          .flatMap[Any, Nothing, List[(String, Int)]](_.toList)
         assertM(tx.commit)(hasSameElements(List("a" -> 1, "b" -> 3, "c" -> 2)))
       }
     ),
     suite("lookups")(
       testM("get existing element") {
-        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap(_.get("a"))
+        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap[Any, Nothing, Option[Int]](_.get("a"))
         assertM(tx.commit)(isSome(equalTo(1)))
       },
       testM("get non-existing element") {
-        val tx = TMap.empty[String, Int].flatMap(_.get("a"))
+        val tx = TMap.empty[String, Int].flatMap[Any, Nothing, Option[Int]](_.get("a"))
         assertM(tx.commit)(isNone)
       },
       testM("getOrElse existing element") {
-        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap(_.getOrElse("a", 10))
+        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap[Any, Nothing, Int](_.getOrElse("a", 10))
         assertM(tx.commit)(equalTo(1))
       },
       testM("getOrElse non-existing element") {
-        val tx = TMap.empty[String, Int].flatMap(_.getOrElse("a", 10))
+        val tx = TMap.empty[String, Int].flatMap[Any, Nothing, Int](_.getOrElse("a", 10))
         assertM(tx.commit)(equalTo(10))
       },
       testM("contains existing element") {
-        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap(_.contains("a"))
+        val tx = TMap.make("a" -> 1, "b" -> 2).flatMap[Any, Nothing, Boolean](_.contains("a"))
         assertM(tx.commit)(isTrue)
       },
       testM("contains non-existing element") {
-        val tx = TMap.empty[String, Int].flatMap(_.contains("a"))
+        val tx = TMap.empty[String, Int].flatMap[Any, Nothing, Boolean](_.contains("a"))
         assertM(tx.commit)(isFalse)
       },
       testM("collect all elements") {
-        val tx = TMap.make("a" -> 1, "b" -> 2, "c" -> 3).flatMap(_.toList)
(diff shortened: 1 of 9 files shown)


core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
```diff
@@ -25,53 +25,55 @@ object TMapSpec extends ZIOBaseSpec {
   def spec = suite("TMap")(
     suite("factories")(
       testM("apply") {
-        val tx = TMap.make("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3).flatMap(_.toList)
+        val tx = TMap.make("a" -> 1, "b" -> 2, "c" -> 2, "b" -> 3).flatMap[Any, Nothing, List[(String, Int)]](_.toList)
```
mijicd: This is unfortunate :(

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -463,9 +466,38 @@ object STMSpec extends ZIOBaseSpec {
 
         assertM(chain(10000).run, fails(equalTo(10000)))
       }
+    ),
+    suite("STM environment")(
+      testM("access environment and provide it outside transaction") {
+        STMEnv.make(0).flatMap { env =>
+          ZSTM.accessSTM[STMEnv](_.ref.update(_ + 1)).commit.provide(env) *>
```
mijicd: I would keep the naming, i.e. `accessM` rather than `accessSTM`.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -463,9 +466,38 @@ object STMSpec extends ZIOBaseSpec {
...
+        }
+      }
     )
   )
 
+  trait STMEnv {
+    val ref: TRef[Int]
+  }
+  object STMEnv {
+    def make(i: Int) =
```
mijicd: Please specify return type.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -463,9 +466,38 @@ object STMSpec extends ZIOBaseSpec {
...
+      TRef
+        .make(i)
+        .map { ref0 =>
+          new STMEnv {
+            val ref = ref0
+          }
+        }
+        .commit
```
mijicd: You can use `makeCommit` instead.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -0,0 +1,904 @@
...
+   */
+  final def provide(r: R): ZSTM[Any, E, A] =
+    provideSome(_ => r)
+
+  /**
+   * Provides some of the environment required to run this effect,
+   * leaving the remainder `R0`.
+   */
+  final def provideSome[R0](f: R0 => R): ZSTM[R0, E, A] =
+    new ZSTM((journal, fiberId, aLong, r0) => self.exec(journal, fiberId, aLong, f(r0)))
```
mijicd: Direct exec invocations are problematic.

mschuwalow:
> Very nice work overall. However, I'm not 100% convinced that the environment in a transactional context is useful, and it decreases the API ergonomics. If possible, please provide an example solved (improved) by the feature.

Well, the usecase I had in mind is having stm datastructures in the environment as I did/do with the hyparview pr in keeper.

I have seen similar things done with MonadReader in haskell so I don't think it's an uncommon usecase

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
mschuwalow: There was a similar discussion here https://github.com/zio/zio/pull/2337#discussion_r351704196

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
mschuwalow: This is our good friend https://github.com/scala/bug/issues/11798 showing its ugly head

adamgfraser:
@mschuwalow Not arguing one way or the other but for your use case couldn't you access an `STM` structure in the environment, do a transaction with it, and then commit that transaction without `STM` having an environment type?

mschuwalow:
Problem is that you then lose the ability to compose multiple smaller ones, or you have to pull the entire environment out of R first. Both lead to clumsy code in my opinion:

i.e. here:
```scala
  def handleForwardJoin[T](
    msg: Protocol.ForwardJoin[T]
  )(
    implicit
    ev1: Tagged[Protocol[T]],
    ev2: Tagged[InitialProtocol[T]]
  ) =
    Env.using[T] { env =>
      val accept = for {
        con   <- ZIO.accessM[Transport[T]](_.transport.connect(msg.originalSender))
        reply <- Tagged.write[InitialProtocol[T]](InitialProtocol.ForwardJoinReply(env.myself))
        _     <- con.send(reply)
        _     <- addConnection(msg.originalSender, con)
      } yield ()

      val process = env.activeView.keys.map(ks => (ks.size, msg.ttl.step)).flatMap {
        case (i, _) if i <= 1 =>
          STM.succeed(accept)
        case (_, None) =>
          STM.succeed(accept)
        case (_, Some(ttl)) =>
          for {
            list <- env.activeView.keys.map(_.filterNot(n => n == msg.sender || n == msg.originalSender))
            next <- env.selectOne(list)
            _    <- if (ttl.count == env.cfg.prwl) env.addNodeToPassiveView(msg.originalSender) else STM.unit
          } yield ZIO.foreach(next)(send[T, Protocol[T]](_, msg.copy(sender = env.myself, ttl = ttl)))
      }
      process.commit.flatten
    }
```
I would prefer to not have to wrap everything in Env.using (which is essentially ZIO.environment[Env[T]].map(_.env).flatMap)

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
jdegoes: I don't think there's consistency in the code yet. One could argue for `accessM` (look at `mapM` in STM).

jdegoes:
@mijicd @adamgfraser @mschuwalow 

Originally I could not imagine any application of `R` in STM. But then it was pointed out to me that data structures (queues, maps, caches, etc.) of numerous kinds may exist in the environment. 

Indeed, many data structures will probably be "global" for the scope of a request or even a whole application. While you can currently access the environment at a higher level, and feed it downward to pure STM methods, it would be more idiomatic to support environment in STM, like ZStream, ZQuery, and many other types.

One can see this would lead to the creation of whole libraries that use global modules, e.g. `UserSession` module backed by `TMap`, `LocalCache`, `Metrics`, etc. 

Ultimately I am not strongly opinionated on this but it seems there is some value in accessing environment from STM transactions, especially if type inference issues could be solved, and type aliases introduced to simplify common scenarios.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
mijicd: I'd say let's keep the M suffix everywhere. While it might not be ideal, it is somewhat accepted convention, especially if you look at Haskell counterparts.

mschuwalow:
Note that the type inference issues are the same that occur with regular ref, etc.
As the experience is still very good there, I believe this one will also not be a large problem in practice

mijicd:
Need to specify types is annoying, but it's not unbearable, and as Maxim indicated, not required all the time. The current state looks good to me, so I'm up for merging it and polishing any usability issues on the fly.

jdegoes:
As a followup and prior to 1.0, I'd be in favor of:

```scala
Transaction[A]
ZTransaction[R, E, A]
```

with `STM` type / val for legacy. I wouldn't give it the full ZIO treatment though (separate companion objects), too much work for too little gain.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
som-snytt: The fix is in again for that lint limitation.

PR #2477, opened 2019-12-15
improve code requiring unique identifiers

mschuwalow:
/cc @iravid

### DIFF
--- a/core/shared/src/main/scala/zio/internal/UniqueKey.scala
+++ b/core/shared/src/main/scala/zio/internal/UniqueKey.scala
@@ -1,0 +1,11 @@
+package zio.internal
+
+/**
+ * This can be used whenever an arbitrary number of unique keys needs to be generated as
+ * this will just use memory location for equality.
+ */
+final class UniqueKey
+
+object UniqueKey {
+  def apply(): UniqueKey = new UniqueKey()
+}
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -24,6 +24,7 @@ import zio._
 import zio.clock.Clock
 import zio.duration.Duration
 import zio.stream.ZStream.Pull
+import zio.internal.UniqueKey
 
 /**
  * A `Stream[E, A]` represents an effectful stream that can produce values of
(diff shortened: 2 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1058,9 +1067,9 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
                 // all newly created queues should end immediately
                 _ <- newQueue.set {
                       for {
-                        id    <- nextIndex
                         queue <- Queue.bounded[Take[E1, A1]](1)
                         _     <- queue.offer(endTake)
+                        id    = UniqueKey()
```
jdegoes: Inserting side-effects in pure code is notoriously difficult to reason about. In this case it should be ok (it's in the middle of a flat map and only needs non-determinism).

I don't consider this a good practice in ZIO but of course if there's a good reason for it (performance, lower code volume, etc.), then I'm all for it.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -985,16 +986,25 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
     decide: A => UIO[Int => Boolean]
   ): ZManaged[R, Nothing, List[Queue[Take[E1, A1]]]] =
-    Promise.make[Nothing, A => UIO[Int => Boolean]].toManaged_.flatMap { prom =>
+    Promise.make[Nothing, A => UIO[UniqueKey => Boolean]].toManaged_.flatMap { prom =>
       distributedWithDynamic[E1, A1](maximumLag, (a: A) => prom.await.flatMap(_(a)), _ => ZIO.unit).flatMap { next =>
-        (ZIO.collectAll(List.fill(n)(next.map(_._2))) <* prom.succeed(decide)).toManaged_
+        ZIO.collectAll {
+          Range(0, n).map(id => next.map { case (key, queue) => ((key -> id), queue) })
+        }.flatMap { entries =>
+          val (mappings, queues) = entries.foldRight((Map.empty[UniqueKey, Int], List.empty[Queue[Take[E1, A1]]])) {
```
jdegoes: This looks like more code, not less? But hard to tell with the reformat...

core/shared/src/main/scala/zio/internal/UniqueKey.scala
```diff
@@ -0,0 +1,11 @@
+package zio.internal
+
+/**
+ * This can be used whenever an arbitrary number of unique keys needs to be generated as
+ * this will just use memory location for equality.
+ */
+final class UniqueKey
```
jdegoes: I might make these package private to `zio` so users aren't tempted to use them. And so we don't accidentally leak them anywhere.

streams/shared/src/main/scala/zio/stream/ZStream.scala
mschuwalow: This is not a side effect in that sense. The motivation behind UniqueKey is that we at some points want to keep track of an undefined number of fibers/iterations/etc. This was before done by arbitrarily selection int/long as the key and rolling with it. The problem with that approach is that it is every time an adhoc solution and would lead to very subtle bugs when it actually overflows.

A better key for this purposes is just an arbitrary object that does not override hashcode / equals and therefore uses its memory location as its identity. This allows us to create infinitely many distinct keys (up to the limits of ram) without any additional overhead.

Issue #2232, opened 2019-11-13
ZIO Stream: ZStream#fromIterable Does Not Terminate on Dotty


adamgfraser:
The below test times out on Dotty.

```scala
import zio.stream._
import zio.test._

object StreamSpec
    extends ZIOBaseSpec(
      suite("StreamSpec")(
        testM("fromIterable") {
          for {
            _ <- ZStream.fromIterable(List(1, 2, 3, 4, 5)).runCollect
          } yield assertCompletes
        }
      )
    )
```

serragnoli:
I'll take a look at this issue.

serragnoli:
I haven't had the chance to look into this issue yet, in case someone want to pick it up, feel free to do so.

PR #2480, opened 2019-12-16
WIP Migrate GenSpec

unclebob418:
Migrate GenSpec for #1898 
@adamgfraser almost there! can you please take a look? I've grouped things by some extra `spec`s for some semblance of organisation and refactored to use more assertions (specifically for `checkSample` & `checkShrink`, thanks to your help on discord).

I've found a test that I think can now be removed (fromIterable constructs deterministic generators); can you please confirm we don't need it anymore (`@@ ignore`d at the moment)?

The only bit not working is because of an issue I have an issue relating to the types (or more specifically, type) on `assert`/`assertM`

```scala
final def assert[A](value: => A, assertion: Assertion[A])
```
the `value` needs to be the same type as the `assertion`, which does make sense, except when I create an assertion like 

```scala
assert(List(Some(3), None, Some(2)), contains(isNone) && contains(isSome(anything)))
```
the type of the assertion doesn't match the type of the `List` even though what I'm effectively doing is making an assertion on a list of options.
```
[error]  found   : zio.test.Assertion[Iterable[zio.test.Assertion[Option[Any]]]]
[error]  required: zio.test.Assertion[List[Option[Int]]]
```

Am I just misunderstanding usage or is there a better way to check the types on the assertions? I don't think we really care about the nested assertion part of the type, but I'm not how we could go about separating them


test-tests/shared/src/test/scala/zio/test/GenSpec.scala
```diff
@@ -62,18 +412,128 @@ object GenSpec extends ZIOBaseSpec {
           } yield assert(result.reverse.headOption, isSome(equalTo((m, 0)) || equalTo((0, n))))
         }
       }
-    )
+    ),
+    testM("fromIterable constructs deterministic generators") {
```
adamgfraser: I think we can keep this test but we need to change `zipWith` to `crossWith` and mark it as `scala2Only` due to #2232.

adamgfraser:
@unclebob418 For that use case you need to use `exists` instead of `contains`. `exists` takes an `Assertion[A]` as a parameter whereas `contains` takes an `A` as a parameter. You could think of `contains` as a specialized version of `exists` where the predicate is that the element equals the specified value. When I made that change your examples compiled for me locally.

unclebob418:
thanks @adamgfraser I think that's got the main issues sorted. I've added a bit to the scaladoc for `exists`/`contains` to make make the relationship between them clearer - anything else need changing?

adamgfraser:
@unclebob418 Great! Will review now.

unclebob418:
Excellent :) thanks for your help once again. I'll get started on the next spec shortly.

adamgfraser:
Awesome! Thanks foe all your work on migrating these tests. Hopefully it has been a good learning experience.

unclebob418:
No worries at all! I've learnt heaps from you and just figuring stuff out myself, and it's nice that my work is actually useful to the community rather than just doing exercises for the sake of learning :)

Issue #2005, opened 2019-10-19
Interruptible regions in finalizers cause finalizers to not run after fiber is interrupted


neko-kai:
Example: 
```scala
import zio.console.putStrLn
import zio.duration._
import zio.{URIO, ZIO}

object ZioIncoherentInterruptibleRegionsInFinalizers extends zio.App {
  override def run(args: List[String]): URIO[zio.ZEnv, Int] = {
    ZIO.never.ensuring {
      ZIO.sleep(1.second) *>
      ZIO.unit.interruptible *> // removing this causes print to happen
      putStrLn("Waited 1s") // never printed because interruptible causes immediate interrupt
    }.fork.flatMap {
      ZIO.sleep(1.second) *> _.interrupt
    }.as(0)
  }
}
```

Expected:

Line `Waited 1s` in the console

Actual:

Nothing is printed because interruptible op `ZIO.unit.interruptible` causes the finalizer to be interrupted AGAIN for good by its own `interrupted` status.

/cc @jdegoes @darl

devesh-shetty:
Hi @neko-kai ,
Could I take this up? Thanks!

neko-kai:
@devesh-shetty 
Sure, feel free to!

jdegoes:
@neko-kai Damn, nice find. There mist be a story behind that! 😆

jdegoes:
Hmm, actually isn't this correct behavior? What would you expect to happen? A finalizer is not special, so it is like any ordinary interruptible hole inside an uninterruptible region.

neko-kai:
@jdegoes 
Well, let's see what haskell does 😅 https://repl.it/repls/WhichIrresponsibleCharacterencoding
```
Started
Finalizer: Waiting 1s
Waited 1s
```
The finalizer is not interrupted again – but then that has to do with interruption being an exception –  a one-time event. I would (intuitively) expect the finalizer to become interruptible by _another_ outside signal, but not by the _past_ interruption that caused it to execute.

jdegoes:
@neko-kai To get the behavior you want, you could write:

```scala
ZIO.never.ensuring {
      ZIO.sleep(1.second) *>
      ZIO.unit.interruptible.run *>
      putStrLn("Waited 1s") // never printed because interruptible causes immediate interrupt
    }.fork.flatMap {
      ZIO.sleep(1.second) *> _.interrupt
    }.as(0)
```

[...]

PR #2483, opened 2019-12-16
ZIO: Add bracketFork variants

### LINKED ISSUE #2351: ZIO: Add bracketFork variants
The current `bracket` operation cannot be interrupted during acquisition. This means if a lengthy acquisition is attempted, and the result is no longer needed, then the fiber executing the acquisition will "hang" for quite some time, until the acquisition is completed. It may be desirable in some cases to allow "fake interruption" of acquisition in a new combinator `bracketFork`.

A hypothetical `bracketFork` will allow instantaneous interruption during resource acquisition; but it will not actually interrupt the underlying acquisition, but rather, in a separate fiber, await the acquisition, and then gracefully and immediately release the resource after acquisition succeeds. Thus the fiber executing `bracketFork` will be able to be interrupted right away even if in the middle of a lengthy acquisition operation.

The `bracketFork` variants will have the exact same structure as existing `bracket` combinators. Probably, we can get away with fewer and more common variants.

Existing definition of `bracketExit` is as follows:

```scala
ZIO.uninterruptibleMask[R, E, B](
      restore =>
        acquire.flatMap({ a =>
          restore(use(a)).run.flatMap( { e =>
            release(a, e).foldCauseM(
              cause2 => ZIO.halt(e.fold(_ ++ cause2, _ => cause2)),
              _ => ZIO.done(e)
            )
          })
        })
    )
```

dariusrobson:
Fixes #2351 

Introduces BracketFork variants as well as zio.interruptibleFork.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -647,8 +647,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def interruptibleFork: ZIO[R, E, A] =
     ZIO.uninterruptible {
       for {
-        fiber <- self.interruptible.fork.daemon
-        res   <- fiber.join.interruptible.onInterrupt(fiber.interrupt.fork.daemon)
+        parentFiberId <- ZIO.fiberId
+        fiber         <- self.interruptible.fork.daemon
+        res           <- fiber.join.interruptible.onInterrupt(fiber.interruptAs(parentFiberId).fork.daemon)
       } yield res
     }
 
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -614,10 +614,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * However, the effect and its interruption occurs in a forked fiber
    */
   final def interruptibleFork: ZIO[R, E, A] =
-    for {
-      fiber <- self.fork.daemon
-      res   <- fiber.join.onInterrupt(fiber.interrupt.fork.daemon)
-    } yield res
+    ZIO.uninterruptibleMask { restore =>
+      for {
+        fiber <- restore(self.fork.daemon)
+        res   <- restore(fiber.join).onInterrupt(fiber.interrupt.fork.daemon)
+      } yield res
+    }
 
   /**
    * Switches the interrupt status for this effect. If `true` is used, then the
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -173,6 +173,41 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
         }
     )(use)
 
+  final def bracketFork[R1 <: R, E1 >: E, B](
+    release: A => URIO[R1, Any],
+    use: A => ZIO[R1, E1, B]
+  ): ZIO[R1, E1, B] = ZIO.bracketFork(self, release, use)
+
+  final def bracketFork: ZIO.BracketForkAcquire[R, E, A] = ZIO.bracketFork(self)
+
+  final def bracketFork_[R1 <: R, E1 >: E]: ZIO.BracketForkAcquire_[R1, E1] =
+    new ZIO.BracketForkAcquire_(self)
+
+  final def bracketFork_[R1 <: R, E1 >: E, B](
+    release: URIO[R1, Any],
+    use: ZIO[R1, E1, B]
+  ): ZIO[R1, E1, B] =
+    ZIO.bracketFork(self, (_: A) => release, (_: A) => use)
+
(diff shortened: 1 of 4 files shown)


dariusrobson:
@jdegoes Here is a first go at the bracketFork variants and adding the interruptibleFork method. I might need a hand in trying to get away with fewer and more common variants.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -581,6 +640,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    */
   final def interruptible: ZIO[R, E, A] = interruptStatus(InterruptStatus.Interruptible)
 
+  /**
+   * Returns an effect that when interrupted returns immediately.
+   * However, the effect and its interruption occurs in a forked fiber
+   */
+  final def interruptibleFork: ZIO[R, E, A] =
```
jdegoes: I'd make sure this one has a `ZIO.interruptibleFork` variant for symmetry with `ZIO.interruptible`.

dariusrobson: do you mean add a `ZIO.interruptibleForkMask` ?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -581,6 +640,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+    ZIO.uninterruptible {
+      for {
+        fiber <- self.interruptible.fork.daemon
+        res   <- fiber.join.interruptible.onInterrupt(fiber.interrupt.fork.daemon)
```
regiskuckaertz: Do you need to fork the interruption? Also, it may be necessary to get the fiber id first, and then use `interruptAs(parentId)`

dariusrobson: Does the interruption not need to be forked so that the parent fiber's interruption completes immediately?

neko-kai: @dariusrobson 
Using `.interruptible` precludes code from being usable inside finalizers, see: https://github.com/zio/zio/issues/2005
Usage here can be replaced by `uninterruptibleMask` (`.interruptible.fork` should remain as is though, because that `.interruptible` is in a different fiber.)

jdegoes: This is a `fork` version of `interruptible`, so it should work exactly the same as `interruptible`. If it were implemented using `uninterruptibleMask`, then it would not yield expected results. If we end up deleting `interruptible`, we would delete `interruptibleFork` along with it.

jdegoes: Yes, it is necessary to fork the interruption as per @dariusrobson initial attempt.

neko-kai: @jdegoes 
Sorry, I was confused about the intended semantics here.

>  If we end up deleting `interruptible`, we would delete `interruptibleFork` along with it.

tbh, I'd think that a function to fork off/skip waiting for an inner action's interruption is useful independently from `interruptible` and can be orthogonal to setting a fiber's interruptibility status. I think it may be useful in general to be able to specify *how* interruption finalization will execute, orthogonally from specifying whether it *can* happen, e.g. `.onInterrupt` does not force a fiber to be interruptible, even though it's a no-op in an uninterruptible section.

jdegoes: Ah, good idea to use `interruptAs`. We want to push the parent fiber id into the children interruptions.

jdegoes: I agree with you that such an independent action would be quite nice. 

I think interruption finalizers have to operate the same as others, since internally they are erased (there are no finalizers in ZIO core); also violating linearity of "finalizers" would have grave consequences for reasoning about code. I'm not sure we have much choice there.

But basically in any case where interruption is happening, we need to be able to "choose" whether to wait for that or to fork it. I think there's probably a better way to do that than myriad `fork` variants, but we probably (and unfortunately) aren't going to figure that out pre 1.0...

neko-kai: @jdegoes 
By interruption finalizaters I mean just the usual finalization - when it's triggered by external interruption, i.e. just having a variant of `interruptibleFork` without `interruptible` would allow choosing whether to fork out this finalization or not without affecting the interruptible flag.

dariusrobson:
Thanks @regiskuckaertz - I have changed `interruptibleFork` to use `interruptAs`.
@jdegoes I have added `interruptibleForkMask`.

jdegoes:
@dariusrobson Superb work on this! This gives people the interruptible bracket / acquire they always wanted, without the danger of resource leaks.

core/shared/src/main/scala/zio/ZIO.scala
jdegoes: Yes, I like that.

PR #2484, opened 2019-12-16
Fixes ZStream#fromPull and adds ZStream#repeatPull.

### LINKED ISSUE #2481: Adding ZStream combinator that keeps pulling elements while ignoring state
A while back I saw someone interested in having a combinator in ZStream that would not require you to handle state but keep pulling elements of the stream, so using `unfold` or `paginate` would be awkward. 

@adamgfraser suggested a potential combinator with the name `fromPullRepeat` like:

`def fromPullRepeat[R, E, A](pull: ZIO[R, Option[E], A]): ZStream[R, E, A]`

I guess this would be a nice addition. cc @iravid

saraiva132:
Fixes behaviour of ZStream#fromPull to be the same as fromEffect:
- It either produces one value or no value

Adds ZStream#repeatPull which was the old ZStream#fromPull which:
- Emits values until it fails with None

Adds tests to make sure the behaviour is correct

Closes #2481

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -245,12 +245,12 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
                      )
                      .fork
         bs <- ZStream
-               .repeatPull(consume(stateVar, permits))
+               .repeatEffectOption(consume(stateVar, permits))
                .mapConcatChunk(identity)
                .process
                .ensuringFirst(producer.interrupt.fork)
       } yield bs
-    }.flatMap(ZStream.repeatPull)
+    }.flatMap(ZStream.repeatEffectOption)
   }
 
   /**
@@ -553,7 +553,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
         bs <- consumerStream(stateVar, permits).process
                .ensuringFirst(producer.interruptAs(fiberId).fork)
       } yield bs
-    }.flatMap(ZStream.repeatPull)
+    }.flatMap(ZStream.repeatEffectOption)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3148,6 +3142,14 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
   final def fromQueueWithShutdown[R, E, A](queue: ZQueue[Nothing, Any, R, E, Nothing, A]): ZStream[R, E, A] =
     fromQueue(queue).ensuringFirst(queue.shutdown)
 
+  /**
+   * Creates a stream from a [[Pull]] producing a value of type `A` or an empty Stream
+   */
+  final def fromPull[R, E, A](pull: Pull[R, E, A]): ZStream[R, E, A] =
```
iravid: Possible alternative name: `fromEffectOption`. More descriptive and won't result in silent breakage for anyone upgrading.

e.g.
```suggestion
  final def fromEffectOption[R, E, A](effect: ZIO[R, Option[E], A]): ZStream[R, E, A] =
```

wdyt? @adamgfraser @vasilmkd

vasilmkd: Agreed, we must assume this is already in use.

adamgfraser: It would be nice to have consistency between the repeating and single versions of this and `effect` but your point about breakage is a strong one. One other thought I had is whether users have an understanding of what a `Pull` is. The type alias is extremely useful for us internally but if I wanted to use an `Option` to express stopping would I know what the `Pull` type alias is and that is what I should use? Maybe there is a more descriptive name for both combinators? Just an idea. I’m good with any of these.

iravid: Yes, I agree that we should probably not expose `Pull`.

`fromEffectOption` and `repeatEffectOption` check those marks for me.

iravid:
So @saraiva132 - let's use the following names:
`repeatPull -> repeatEffectOption`
`fromPull -> fromEffectOption`

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Agreed.

saraiva132:
I am changing this but I also noticed that now internal usage (which uses Pull) looks a bit awkward by calling `repeatEffectOption` instead of `repeatPull`

saraiva132:
@iravid here you go :)

PR #2486, opened 2019-12-17
Extract tag regex in CircleCI config

mijicd:
### Summary

- Extracted the tag regex instead of duplicating it.
- Make `release` depend on `mdoc`.


jdegoes:
@mijicd ❤️

PR #2487, opened 2019-12-17
Add ZIO#partitionMPar and partitionMParN

saraiva132:
A take on partitionMPar combinator to partition effects in parallel. 

I would rather do validateMPar and validateFirstMPar on other PRs so I focused on this one for now :)

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -786,22 +786,42 @@ object ZIOSpec extends ZIOBaseSpec {
       } @@ nonFlaky
     ),
     suite("partitionM")(
-      testM("collect only successes") {
+      testM("collects only successes") {
         val in = List.range(0, 10)
         for {
-          res <- ZIO.partitionM(in)(a => ZIO.effect(a))
+          res <- ZIO.partitionM(in)(a => ZIO.succeed(a))
         } yield assert(res._1, isEmpty) && assert(res._2, equalTo(in))
       },
-      testM("collect only failures") {
+      testM("collects only failures") {
         val in = List.fill(10)(0)
         for {
           res <- ZIO.partitionM(in)(a => ZIO.fail(a))
         } yield assert(res._1, equalTo(in)) && assert(res._2, isEmpty)
       },
-      testM("collect failures and successes") {
+      testM("collects failures and successes") {
         val in = List.range(0, 10)
         for {
           res <- ZIO.partitionM(in)(a => if (a % 2 == 0) ZIO.fail(a) else ZIO.succeed(a))
+        } yield assert(res._1, equalTo(List(0, 2, 4, 6, 8))) && assert(res._2, equalTo(List(1, 3, 5, 7, 9)))
+      }
+    ),
+    suite("partitionMPar")(
+      testM("collects a lot of successes") {
+        val in = List.range(0, 1000)
+        for {
+          res <- ZIO.partitionMPar(in)(a => ZIO.succeed(a))
+        } yield assert(res._1, isEmpty) && assert(res._2, equalTo(in))
+      },
+      testM("collects failures") {
+        val in = List.fill(10)(0)
+        for {
+          res <- ZIO.partitionMPar(in)(a => ZIO.fail(a))
+        } yield assert(res._1, equalTo(in)) && assert(res._2, isEmpty)
+      },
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2468,6 +2468,28 @@ private[zio] trait ZIOFunctions extends Serializable {
         )
     }
 
+  /**
+   * Feeds elements of type A to a function f that returns an effect.
+   * Collects all successes and failures, in parallel, in a tupled fashion.
+   *
+   */
+  final def partitionMPar[R, E, A, B](
```
adamgfraser: I think we want to implement this more directly in terms of existing combinators:

```scala
  final def partitionM[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, E, B])(implicit ev: CanFail[E]): ZIO[R, Nothing, (List[E], List[B])] =
    ZIO.foreachPar(in)(f(_).either).map(_.foldRight((List.empty[E], List.empty[B])) { 
      case (Left(e), (es, bs)) => (e :: es, bs)
      case (Right(b), (es, bs)) => (es, b :: bs)
    })
```

I know we went with a slightly different implementation for `partitionM` to avoid the extra allocations but the cost of that is going to be dwarfed by the additional machinery of the queue and promises here.

saraiva132: I had the same issue again. I forgot there was the either combinator.... gosh my approach was over-engineered for nothing. XD

saraiva132:
Thanks for the review and support as always @adamgfraser. 
Question. `partitionMParN` and `partitionMPar` share quite a lot logic. Namely, rightFolding `List[Either[E,B]]` -> `(List[E],List[B])`. 
Is there any value in moving that into a private function. What would one do in such a scenario?

core/shared/src/main/scala/zio/RIO.scala
```diff
@@ -512,6 +512,22 @@ object RIO {
   )(f: A => RIO[R, B])(implicit ev: CanFail[Throwable]): RIO[Nothing, (List[Throwable], List[B])] =
     ZIO.partitionM(in)(f)
 
+  /**
+   * @see See [[zio.ZIO.partitionMPar]]
+   */
+  final def partitionMPar[R, A, B](
+    in: Iterable[A]
+  )(f: A => RIO[R, B])(implicit ev: CanFail[Throwable]): RIO[Nothing, (List[Throwable], List[B])] =
```
jdegoes: It occurs to me we can delete all these implicit `CanFail[Throwable]`, because, well, `Throwable` can fail. 😄  But maybe best as a separate PR since I see the other methods too.

saraiva132: This made me laugh. Thank you for that. I can do that straight after this one.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -786,25 +786,65 @@ object ZIOSpec extends ZIOBaseSpec {
       } @@ nonFlaky
     ),
     suite("partitionM")(
-      testM("collect only successes") {
+      testM("collects only successes") {
         val in = List.range(0, 10)
         for {
-          res <- ZIO.partitionM(in)(a => ZIO.effect(a))
+          res <- ZIO.partitionM(in)(a => ZIO.succeed(a))
```
adamgfraser: I noticed at some point we did a global import of `zio.CanFail.canFail` at the top of this file. Can we please remove that and do imports for specific tests when necessary? Using that import essentially removes the functionality of `CanFail`, so we need to use it sometimes for specific tests but generally we want to keep that functionality so the compiler helps flag for us things that aren't making sense.

adamgfraser:
@saraiva132 If we wanted we could create something like:

```scala
private[zio] object ZIOIterableFunctions { ... }
```

It looks like there is a `partitionMap` function in Scala 2.13 that basically does what we want so we could essentially backport that. Similar functionality is implemented in at least one other place [here](https://github.com/zio/zio/blob/d2cbceda3cb97943fc11853acb3a043579ed824e/test-tests/shared/src/main/scala/zio/test/GenUtils.scala#L82).

saraiva132:
Sounds good @adamgfraser. I will create an issue where that can be further discussed. This does seem to be more generic than ZIO. If partitionMap is in 2.13 then we could perhaps just backport it and refactor the current usages.

Issue #2371, opened 2019-11-30
Optimize parSequence


neko-kai:
As per @Avasil's [benchmarks](https://gist.github.com/pk044/181b32f9af380c9230d8512d14233c2a) (https://github.com/zio/zio/pull/2290#event-2821787971) `ZIO.parSequence` has a long way to go to get good performance. Any contributions to optimize it or related methods are welcome!

vasilmkd:
Possibly a duplicate of #1182, but it needs more visibility. Either way, linking to the previous issue.

horned-sphere:
I've made attempt at this in [this PR](https://github.com/zio/zio/pull/2492). The implementation is fairly ugly but it does work and is faster (I've attached the benchmark comparison). Any feedback on it would be greatly appreciated. 

If you think it has any merit, the change in implementation has broken an assumption of a test case in the stack tracer which I'd like to get some advice on before trying to fix it.

kitlangton:
Should this issue be closed since #2492 got merged?

Issue #2358, opened 2019-11-28
ZIO Execution Traces: Remove "guts" of ZIO from execution traces


jdegoes:
`ZIOFn` is not being consistently used in all new combinators. The code requires a detailed inspection to identify and cleanup all combinators so that ZIO internals do not appear in any execution trace.

If anyone wants to take on this issue, @neko-kai or myself can provide detailed guidance.

jaliss:
@jdegoes I'd like to give this a shot.

jdegoes:
@jaliss 

So here is a possible implementation for `zipRight` (AKA `*>`):

```scala
self.flatMap(_ => that)
```

The problem with this implementation is that ZIO gets stack trace information from closures. And there is a closure in this function (`_ => that`). So ZIO will find that closure and put the trace inside the execution trace of the fiber running the effect. That leads to a lot of lines in an execution trace being taken up by the insides of ZIO.

Worse than that, ZIO won't have any stack trace information on the closure passed to `*>` that was created in user-land code. This closure is the "by name" parameter `that`; Scala will create a `Function0` to wrap it wherever the user invokes the method `*>`, which will have useful information on where in the user's code the function `*>` was called. Ideally, this useful information would appear in execution traces.

To solve both problems, ZIO has "wrapper functions" that let us create things like `_ => that`, but they are tagged in a special way that removes tracing information; in addition, these "wrapper" functions can point to inner closures. So we can point to the `Function0` created for `that`, and add that to an execution trace.

Here is a wrapper class for `ZipLeft`:

```scala
final class ZipLeftFn[R, E, A, B](override val underlying: () => ZIO[R, E, A]) extends ZIOFn1[B, ZIO[R, E, B]] {
    def apply(a: B): ZIO[R, E, B] =
      underlying().as(a)
  }
```

You can see it holds a reference to `underlying`, which tells ZIO where to get the execution trace from; and then because it extends `ZIOFn1`, the `ZIOFn1` itself will not be added to the stack trace.

Does this make sense? 

If so, the next step is to look for ZIO methods that are just creating raw closures, which create unnecessary stack traces, and which do not point to the underlying closures; and then to wrap them with `ZIOFn1` or a new helper class derived from that.

robmwalsh:
[...]

PR #2492, opened 2019-12-18
Attempt to improve the performance of foreachPar for #2371.

horned-sphere:
This isn't quite there but some comments would be great if anyone has the time.

My implementation is a bit ugly but it does work and is somewhat faster. Here's a comparison if the benchmark results (on my laptop).

Current:

`[info] Benchmark                            (count)   Mode  Cnt    Score   Error  Units
[info] ParSequenceBenchmark.zioParSequence      100  thrpt   20  500.436 ± 6.501  ops/s
[info] ParSequenceBenchmark.zioParSequence     1000  thrpt   20   10.674 ± 0.116  ops/s`

Mine:

`[info] Benchmark                             (count)   Mode  Cnt     Score     Error  Units
[info] ParSequenceBenchmark.zioParSequence       100  thrpt   20  4004.473 ± 211.169  ops/s
[info] ParSequenceBenchmark.zioParSequence      1000  thrpt   20   655.739 ±  18.208  ops/`

I currently have a failing test case in zio.StackTracesSpec -> "fiber ancestry example with uploads". This is the only failing test and seems to be because the change in implementation of foreachPar breaks its assumptions. I'm not confident enough of my understanding of the purpose of this test to fix it so any pointers would be greatly appreciated!


saraiva132:
This looks promising!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2098,11 +2098,46 @@ private[zio] trait ZIOFunctions extends Serializable {
...
    * For a sequential version of this method, see `foreach`.
    */
-  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] =
-    as.foldRight[ZIO[R, E, List[B]]](effectTotal(Nil)) { (a, io) =>
-        fn(a).zipWithPar(io)((b, bs) => b :: bs)
-      }
-      .refailWithTrace
+  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] = {
+    val n         = as.size
+    val resultArr = new Array[Any](as.size)
```
vasilmkd: You can use `n` here because you already forced the `Iterable` once.

vasilmkd: You can use `n` here because you already forced the `Iterable` before.

Regarding the `new Array[Any]` I'm wondering if we can demand a `B: ClassTag` in the method definition, in order to avoid the cast at the end. Would the Scala compiler be able to help us here?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2098,11 +2098,46 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+    val resultList: ZIO[R, E, List[B]] = for {
+      parentId <- ZIO.fiberId
+      latch    <- CountdownLatch.make(n)
+      failed   <- Promise.make[Nothing, Int]
+      cause    <- Ref.make[Cause[E]](Cause.empty)
+      fibers <- ZIO.foreach(as.zipWithIndex) {
```
vasilmkd: `zipWithIndex` will create an intermediate collection and together with `ZIO.foreach` there would be two traversals only on this line. I'm wondering if we can use a `view` here (I'm aware that it's not a very popular choice) or the stdlib `Stream/LazyList` (but we will have deprecation issues here).

As an alternative, can we maybe think of a way to force the original `Iterable` to a concrete data type (maybe even a `Vector` for its constant time `size` method) and manually construct the indices that would otherwise require a whole traversal with `zipWithIndex`.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Yes, the size twice was a silly mistake.

I was trying very hard not to alter the signature of the method. Making changes here has wider ranging effects than it might initially appear. A subtly wrong version that I wrote at first managed to lock up the entire build due to it being used in zio-test.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: I think any way of getting rid of the extra traversal would be fine here as it is fairly hidden away. I'll experiment.

If altering the signature is on the table, we could make both the parameter and return types IndexedSeq which would allow constant time size and avoid having to repack into a list at the end. It does make it quite a bit more inflexible though. What do you think?

vasilmkd: It is definitely fine to use impure code in the implementation, as long as it doesn't leak outside or is observable in any way.

I don't know about altering the method signature. Let's try using an IndexedSeq (Vector) inside the method (.toVector) and see where we can go from there.

core/shared/src/main/scala/zio/ZIO.scala
vasilmkd: It's not a mistake. That's why we review code. :)

horned-sphere:
This new version slims down the excessive traversals. I also tried an entirely vector/array based version which can be seen [here](https://github.com/horned-sphere/zio/blob/benchmark/parsequence-alt/core/shared/src/main/scala/zio/ZIO.scala#L2101). I've made this one the official one as it has a slight edge in the benchmarks.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Oddly using a view made it go slower!

The new commit I just pushed reduces the traversals and improves the performance by a small amount. I also tried a completely vector/array based approach as you suggested (which I've put in another branch for comparison). Slightly surprisingly, it comes out slightly slower in the benchmark (though the difference is not large).

vasilmkd: Views aren't exactly know for their amazing performance, so I cannot say that I'm surprised.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2098,11 +2098,55 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] = {
+    val n         = as.size
+    val resultArr = new Array[Any](n)
+    val resultList: ZIO[R, E, List[B]] = for {
+      parentId <- ZIO.fiberId
+      latch    <- CountdownLatch.make(n)
+      failed   <- Promise.make[Nothing, Int]
+      cause    <- Ref.make[Cause[E]](Cause.empty)
+      fibers <- as
+                 .foldLeft[(Int, URIO[R, List[(Int, Fiber[E, B])]])]((0, URIO.succeed(Nil))) { (r, a) =>
```
```suggestion
                 .foldLeft[(Int, URIO[R, List[(Int, Fiber[E, B])]])]((0, URIO.succeed(Nil))) { ((i, acc), a) =>
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2098,11 +2098,55 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+                   r match {
+                     case (i, acc) =>
+                       def handleExit(exit: Exit[E, B]): URIO[R, Any] = exit.foldM(
+                         c => cause.update(cs => cs && c) *> failed.succeed(i),
+                         b => ZIO.effectTotal(resultArr(i) = b)
+                       )
+
+                       val appended = for {
+                         fs <- acc
+                         f <- ZIO
```
```scala
fn(a).foldCauseM(
  c => cause.update(cs => cs && c) *> failed.succeed(i),
  b => ZIO.effectTotal(resultArr(i) = b)
).ensuring(latch.countDown)
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2098,11 +2098,55 @@ private[zio] trait ZIOFunctions extends Serializable {
...
+
+                       (i + 1, appended)
+                   }
+                 }
+                 ._2
+      interrupter = failed.await
+        .flatMap(
+          i =>
+            ZIO.foreach(fibers) {
+              case (j, f) if i != j => f.interruptAs(parentId).unit
```
regiskuckaertz: Given that interrupting a fiber that has already terminated is a noop, you could just write
```scala
ZIO.foreach_(fibers)(_.interruptAs(parentId))
```
and only keep track of the fibers without their index.

regiskuckaertz:
> the change in implementation of foreachPar breaks its assumptions

This is because continuations inside the ZIO combinators should be filtered out from the trace. You will have to wrap your functions in some of the `ZioFn*` to do that. More details [here](https://github.com/zio/zio/issues/2358).

Really great BTW 💪

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Slightly embarrassingly, given how long I've been writing Scala, I didn't realize that was valid syntax...

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: That certainly looks nicer. Is it completely equivalent to what I wrote? (I'm not sure of the subtleties of the difference in semantics of  bracketExit and foldCauseM)

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: It's not guaranteed that it will have terminated when you get to that point (but it will indeed have finished anything that is interruptible anyway). I wrote it like that for clarity of intention. I guess a comment would do just as well for that though.

horned-sphere:
> This is because continuations inside the ZIO combinators should be filtered out from the trace. You will have to wrap your functions in some of the `ZioFn*` to do that. More details [here](https://github.com/zio/zio/issues/2358).

I'd got as far as working out it was something related to that. I'm not 100% sure of how though. I'll experiment till I get something that looks plausible. Thanks to @vasilmkd and @regiskuckaertz for all the help; I think I should be able to polish up a final version now.

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: Yes, in the sense the contract of `ZIO.bracket*(acquire, release, use)` is that `release` is guaranteed to be evaluated if `acquired` succeeded, but since `ZIO.succeed` cannot fail and `handleExit` is just `foldCauseM` in disguise, you can just remove the `bracketExit`.

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: What I meant was that `i == j => f.interruptAs(parentId).unit == ZIO.unit`, as `failed` is set to the index of the first failing fiber, so I think this one is guaranteed.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2219,11 +2219,46 @@ object ZIO {
...
+
+                     (i + 1, appended)
+                 }
+                 ._2
+      interrupter = (failed.await *>
+        ZIO.foreach(fibers)(_.interruptAs(parentId))) // The failed fiber can be safely interrupted.
+      .toManaged_.fork
+      _             <- interrupter.use_(latch.await)
+      combinedCause <- cause.get
+      results <- if (combinedCause.isEmpty) UIO.succeed(resultArr.asInstanceOf[Array[B]].toList)
```
evis: If I'm not mistaken, updates to `resultArr` won't always be visible here, because there is no synchronization on its value. So, there is a race condition.

horned-sphere: The access to resultArr there is guarded by the latch. The completion of the effect on line 2255 guarantees that all of the fibers that were spawned have finished (in one way or another) and so no further changes can be made to the array. Unless there is another way this point could be reached that I'm missing?

evis: From the point of view of Java Memory Model, is it guaranteed, that finishing of these fibers happens-before accessing `resultArr` on L2257? I'm not quite sure about this...

horned-sphere: Thinking about it a bit more, you're absolutely correct. I have a couple of ideas of how to fix it but haven't had a chance to look at it this weekend.

darl: `write to array` HB `latch.countDown`
`latch.countDown` HB `latch.await`
`latch.await` HB `read from array`
is it wrong?

evis: Yeah, seems correct.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -438,6 +438,76 @@ object ZIOSpec extends ZIOBaseSpec {
       }
     ),
     suite("foreachPar")(
+      testM("rins single task") {
```
evis:
```suggestion
      testM("runs single task") {
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2219,11 +2219,46 @@ object ZIO {
...
+      cause    <- Ref.make[Cause[E]](Cause.empty)
+      fibers <- as
+                 .foldLeft[(Int, URIO[R, List[Fiber[E, _]]])]((0, URIO.succeed(Nil))) {
+                   case ((i, acc), a) =>
+                     val task = ZIOFn(fn)(
+                       (x: A) =>
+                         fn(x).traced
+                           .foldCauseM(
+                             c => cause.update(cs => cs && c) *> failed.succeed(()).unit,
+                             b => ZIO.effectTotal(resultArr(i) = b)
```
evis: Could we implement `foreachPar` in terms of `foreachPar_`? The same way as `mergeAllPar` is implemented in #2614, but putting results into Array instead of Ref.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: I changed it to an AtomicReferenceArray which will make the reads and writes to the elements volatile which I _think_ should be adequate.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Yes, that would be entirely reasonable. I'd be more than happy to give up on this PR if you'd like to do it on top of your version.

evis: I'm not sure if I will be able to do it soon :(
I guess, you could do it yourself in this PR too :)

regiskuckaertz:
@evis @horned-sphere using `forEachPar_` makes sense however I missed that at the moment it doesn't accumulate errors, which the previous version did.

horned-sphere:
@regiskuckaertz @evis based on my work on this, the current forEachPar (whilst definitely a lot more elegant than my version) could have it's performance improved significantly with a couple of modifications. I could perhaps make forEachPar_ into a hybrid of its current version and my approach and then implement forEachPar in terms of it?

evis:
Feel free to improve `foreachPar_` performance, sure :)

horned-sphere:
@regiskuckaertz @evis I put the cause accumulation back into foreachPar_, implemented foreachPar in terms of foreachPar_ and improved the performance by cutting down on extraneous fibers.
Benchmark using current master version for foreachPar_:

`
[info] ParSequenceBenchmark.zioParSequence       100  thrpt   20  2642.299 ± 224.553  ops/s

[info] ParSequenceBenchmark.zioParSequence      1000  thrpt   20   382.302 ±  29.994  ops/s
`

Benchmark using this version:

`
[info] ParSequenceBenchmark.zioParSequence       100  thrpt   20  4637.851 ± 304.270  ops/s

[info] ParSequenceBenchmark.zioParSequence      1000  thrpt   20   760.322 ±  29.499  ops/s
`

iravid:
Wow this is really awesome teamwork @horned-sphere @evis!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2231,11 +2233,19 @@ object ZIO {
...
    * For a sequential version of this method, see `foreach`.
    */
-  def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] =
-    as.foldRight[ZIO[R, E, List[B]]](effectTotal(Nil)) { (a, io) =>
-        fn(a).zipWithPar(io)((b, bs) => b :: bs)
-      }
-      .refailWithTrace
+  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] = {
+    val size      = as.size
+    val resultArr = new AtomicReferenceArray[B](size)
```
iravid: @horned-sphere If every fiber is operating in its own index, why do you need an ARA here?

horned-sphere: @iravid The issue isn't with the paralell sets, it's with the sets followed by the gets at the end. As @evis points out above, the Java memory model doesn't guarantee that these writes will be observable to the following reads.

Ideally what I would do is have an array of volatile references, however the JVM only allows for volatile references to arrays. The only ways to perform volatile sets/gets on an array in Java are an atomic array or a VarHandle (JVM >= 9). The ARA is used for volatility not atomicity.

I don't think this would ever be a problem on x86 but on an architecture with weak memory consistency (like arm for example) it would be required

horned-sphere:
> @horned-sphere I think this is good to go. I'd just appreciate if you could add a test for `foreachPar_`'s accumulation of errors.

I'll add that as soon as I get the chance.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2231,11 +2233,19 @@ object ZIO {
...
+
+    val wrappedFn: ZIOFn1[(A, Int), ZIO[R, E, Any]] = ZIOFn(fn) {
+      case (a, i) => fn(a).tap(b => ZIO.effectTotal(resultArr.set(i, b)))
+    }
+
+    foreachPar_(as.toStream.zipWithIndex)(wrappedFn) *>
```
```suggestion
    foreachPar_(as.toStream.zipWithIndex)(wrappedFn).as(
      (0 until size).reverse...
    )
```

horned-sphere: Done in the next commit.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: I've left it as it is for now unless someone has a convincing argument for that reasoning being wrong :).

horned-sphere:
@iravid I added the error accumulation test case and also made it do the interruptions in parallel rather than sequentially.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I think that we are okay without the `AtomicReferenceArray` because we have already completed the `foreachPar_` effect which involves a `Ref` and ` Promise` for coordination so I think by the time we evaluate the continuation we can rely on the effects in `foreachPar_` being completed and visible.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2231,11 +2233,20 @@ object ZIO {
...
-        fn(a).zipWithPar(io)((b, bs) => b :: bs)
+  final def foreachPar[R, E, A, B](as: Iterable[A])(fn: A => ZIO[R, E, B]): ZIO[R, E, List[B]] = {
+    val size      = as.size
+    val resultArr = new AtomicReferenceArray[B](size)
+
+    val wrappedFn: ZIOFn1[(A, Int), ZIO[R, E, Any]] = ZIOFn(fn) {
+      case (a, i) => fn(a).tap(b => ZIO.effectTotal(resultArr.set(i, b)))
+    }
+
+    foreachPar_(as.toStream.zipWithIndex)(wrappedFn).as(
```
adamgfraser: Do we need the `toStream` call here?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2231,11 +2233,20 @@ object ZIO {
...
+      (0 until size).reverse.foldLeft[List[B]](Nil) { (acc, i) =>
```
adamgfraser: Could become `toList` it we change the array to an ordinary `Array`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2259,20 +2270,23 @@ object ZIO {
...
+                     .foldCauseM(c => causes.update(_ && c) *> result.fail(()), _ => {
+                       (succeed.update(_ + 1) >>= { succeed =>
+                         ZIO.when(succeed == size)(result.succeed(()))
+                       }).uninterruptible
+                     })
+                     .fork
+                 }
+        interrupter = result.await
+          .catchAll(_ => ZIO.foreach(fibers)(_.interruptAs(parentId).fork) >>= Fiber.joinAll)
+          .toManaged_
```
adamgfraser: What do we gain by using a `Managed` here? It looks like we are converting it to a `Managed` without a release action and then we immediately use it so seems like we could use a simpler combinator.

iravid: He's forking the `Managed` to get a guarantee on interruption of the fiber when `use_` is done below

iravid:
@adamgfraser How about we merge this and deal with improvements in follow-ups? This PR has been open for quite some time.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: This is in response to a review by @vasilmkd way back up at the top with regards to minimizing traversals of the iterable. I originally resolved it by doing a foldLeft, however after reimplementing foreachPar in terms of foreachPar_ this was no longer possible so I replaced it by converting the interable to a stream (the stream just wraps an iterator so isn't especially expensive). I'm not opposed to getting rid of it.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Yes, that's what I did in the original array based implementation (although it requires a nasty casy from Array[Any] to Array[B] as we don't have a ClassTag in the signature).

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: This is a pattern I've seen elsewhere in the codebase.

horned-sphere: Yes, that is the reason. I introudced this due to a problem I had, early on, when running the benchmark. When the interruption fiber was unbounded, in some proportion of executions it would outlive the fiber running foreachPar in the benchmark. Through some chain of events this led to the interruption fiber being added to Fiber._rootFibers (and not removed). Over the many, many executions in the benchmark that set became huge and the performance took a huge hit. To solve this I decided I wanted a guarantee that the helper fibers could not outlive the execution of the combinator and managing the fiber seemed to be the best way to express that intention.

That no longer seems to happen (I'm guessing there have bene improvements to the fiber lifetime management in the interim) however the principal that the fiber should be guarnateed to not outlive the combinator seems to be good in general. In this case, I think it should be safe to replace that with just joining the fiber rather than using the managed so I can replace it with that.

Apologies for the long winded explanation.

horned-sphere:
@iravid @adamgfraser @regiskuckaertz I really want this PR to go away too but I'd prefer to make it so everyone is happy with it before it gets merged, if that's okay?

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: @iravid Sounds good.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Got it, thanks!

adamgfraser: Couldn't this be simplified by removing the `toManaged` and replacing `use_` with `bracket_` in the line below with `ZIO.unit` as the release action? Or even doing `uninterruptible` instead of `toManaged` and then `*>` instead of `use_` on the line below? I may be missing something but how does a managed without a release action provide us stronger guarantees about the lifetime of the fiber?

horned-sphere: The realease action is added by the call to .fork on the next line (this is forking the managed rather than an effect).

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Got it. The cost of any of this is probably dwarfed by everything else we're doing so fine either way. We could use `iterator` here potentially.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Ah, got it.

horned-sphere: I could make that more explicit by constructing the managed fiber myself rather than doing .use_.fork (or just get rid of the managed block entirely and join the fiber as mentioned above). I _think_ that if the inner effect completes the interrupter thread is guarnateed to complete eventually too so it should be equivalent.

adamgfraser: No I think it is fine. I didn't realize that `Managed#fork` added that logic though it makes sense. I thought we could rely on the fiber being garbage collected when it terminates without having to explicitly interrupt it but it sounds like that wasn't the case at least recently.

adamgfraser:
Looks like there is some kind of build failure on `master` related to formatting. Not sure how that happened. Investigating now.

adamgfraser:
I see it. We had a minor merge conflict. Fixing now.

iravid:
Yikes. Sorry about that @adamgfraser.

PR #2493, opened 2019-12-18
Make Type Signature of ValidateM More Precise

adamgfraser:
We know that if validateM failed there was at least one failure, so we can return a nonempty list to express this in the type signature.

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -640,7 +640,7 @@ object IO {
   /**
    * @see See [[zio.ZIO.validateM]]
    */
-  final def validateM[E, A, B](in: Iterable[A])(f: A => IO[E, B])(implicit ev: CanFail[E]): IO[List[E], List[B]] =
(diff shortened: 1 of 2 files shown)


fanf:
That makes me happy :)

saraiva132:
I considered using NonEmptyList initially but along the way opted not to for an arbitrary reason. Someone mentioned :: might be a bit obscure for some people, would a type alias Nel help? Or do you think it is just fine?

adamgfraser:
@saraiva132 At least right now it is the standard representation in ZIO for a nonempty list. It has the advantage that it is actually a subtype of `scala.collection.immutable.List`, so anything you could do with a List you could do with it, plus being able to pattern match on it and know that it is nonempty which is quite natural with the `::` operator that people are familiar with.

There has been some discussion of having more of a full fledged nonempty list representation but I think the concern has been that it moves us away from the standard library a bit in a way that is not really core to what we are trying to do, plus if you are going to do it right there are a lot of collection operations you would have to reimplement so I don't think it really makes sense.

saraiva132:
Good to get that clarified! Extremely good catch and agreed.

Issue #2500, opened 2019-12-19
ZIO Test: Create new test annotation for repeated tests


jdegoes:
Users would like to know how many times a test was repeated (with, e.g. `nonFlaky` combinator). So we could keep track of those repetitions and render them something like `repeated: 1` (or `repeated: 24` for a suite-level rendering).

To add a test annotation and renderer, the following process must be followed:

1. Add a new TestAnnotation, one which counts repeated tests
2. Access the test annotation from within the `nonFlaky` `TestAspect`
3. Define a new renderer for the TestAnnotation
4. Compose the new renderer with the default test annotation renderer

adamgfraser:
I am working on this.

adamgfraser:
Closed by #2502.

Issue #2499, opened 2019-12-19
ZIO Test: Create new test annotation for retried tests


jdegoes:
Users would like to know how many times a test had to be retried before it succeeded (with, e.g. `flaky` combinator). So we could keep track of those retries and render them something like `retried: 1` (or `retried: 24` for a suite-level rendering).

To add a test annotation and renderer, the following process must be followed:

1. Add a new TestAnnotation, one which counts retried tests
2. Access the test annotation from within the `flaky` `TestAspect`
3. Define a new renderer for the TestAnnotation
4. Compose the new renderer with the default test annotation renderer

adamgfraser:
I am working on this.

adamgfraser:
Closed by #2502.

PR #2502, opened 2019-12-19
Implement Ignored, Repeated, and Retried Test Annotations

### LINKED ISSUE #2498: ZIO Test: Create test annotation for "ignored" tests
Users would like to know which tests were ignored in a given run. So we could keep track of those ignores and render them something like `ignored: 1` (or `ignored: 24` for a suite-level rendering).

To add a test annotation and renderer, the following process must be followed:

1. Add a new TestAnnotation, one which counts ignored tests
2. Access the test annotation from within the `ignore` `TestAspect`
3. Define a new renderer for the TestAnnotation
4. Compose the new renderer with the default test annotation renderer

Am I missing anything, @adamgfraser?

adamgfraser:
Resolves #2498.

A couple of things for us to think about:

1. Right now we are rendering the number of ignored tests but we aren't rendering the labels for the ignored tests. To do that we would need to modify the `DefaultTestRenderer` to render labels for ignored tests with something analogous to the green plus or red minus for passed or failed tests. Not sure if this is a good or a bad thing. On the one hand you might want to know which tests were ignored. On the other hand it makes it very clear where tests were ignored.
2. This works very naturally when we ignore individual tests and then the suite level number of ignored tests is just a natural buildup from that. But if we ignore a whole suite the number of ignored tests in that suite is going to be zero because we are never executing the effect of creating the specs in the suite. I think this is a natural implication of effectual specs but something we should be aware of.

```
[info] + TestAspectSpec
[info] ignored: 8
[info]   + around evaluates tests inside context of Managed
[info]   + dotty applies test aspect only on Dotty
[info]   + dottyOnly runs tests only on Dotty
[info]   + exceptDotty runs tests on all versions except Dotty
[info]   + exceptJS runs tests on all platforms except ScalaJS
[info]   ignored: 1
[info]   ignored: 1
[info]   + failure makes a test pass if the result was a failure
[info]   + failure makes a test pass if it died with a specified failure
[info]   + failure does not make a test pass if it failed with an unexpected exception
[info]   + failure does not make a test pass if the specified failure does not match
[info]   + failure makes tests pass on any assertion failure
[info]   + failure makes tests pass on an expected assertion failure
[info]   + flaky retries a test that fails
[info]   + flaky retries a test that dies
[info]   + flaky retries a test with a limit
[info]   + ifEnv runs a test if environment variable satisfies assertion
[info]   ignored: 1
[info]   ignored: 1
[info]   + ifEnvSet runs a test if environment variable is set
[info]   ignored: 1
[info]   + ifProp runs a test if property satisfies assertion
[info]   ignored: 1
[info]   ignored: 1
[info]   + ifPropSet runs a test if property is set
[info]   ignored: 1
[info]   + js applies test aspect only on ScalaJS
[info]   + jsOnly runs tests only on ScalaJS
[info]   + jvm applies test aspect only on jvm
[info]   + jvmOnly runs tests only on the JVM
[info]   + retry retries failed tests according to a schedule
[info]   + scala2 applies test aspect only on Scala 2
[info]   + scala2Only runs tests only on Scala 2
[info]   + timeout makes tests fail after given duration
[info]   + timeout reports problem with interruption
```

### DIFF
--- a/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
+++ b/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
@@ -71,11 +71,11 @@ object ZTestFrameworkSpec {
           messages.mkString.split("\n").dropRight(1).mkString("\n"),
           List(
             s"${reset("info:")} ${red("- some suite")}",
-            s"${reset("info:")} ${"ignored: 1"}",         
+            s"${reset("info:")} ${"ignored: 1"}",
             s"${reset("info:")}   ${red("- failing test")}",
             s"${reset("info:")}     ${blue("1")} did not satisfy ${cyan("equalTo(2)")}",
             s"${reset("info:")}   ${green("+")} passing test",
-            s"${reset("info:")}   ${"ignored: 1"}",     
+            s"${reset("info:")}   ${"ignored: 1"}"
           ).mkString("\n")
         )
       )
--- a/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
+++ b/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
@@ -70,12 +70,10 @@ object ZTestFrameworkSpec {
           "logged messages",
           messages.mkString.split("\n").dropRight(1).mkString("\n"),
           List(
-            s"${reset("info:")} ${red("- some suite")}",
-            s"${reset("info:")} ${"ignored: 1"}",
+            s"${reset("info:")} ${red("- some suite")} - ignored: 1",
             s"${reset("info:")}   ${red("- failing test")}",
             s"${reset("info:")}     ${blue("1")} did not satisfy ${cyan("equalTo(2)")}",
-            s"${reset("info:")}   ${green("+")} passing test",
-            s"${reset("info:")}   ${"ignored: 1"}"
+            s"${reset("info:")}   ${green("+")} passing test"
           ).mkString("\n")
         )
       )
--- a/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
+++ b/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
@@ -71,9 +71,11 @@ object ZTestFrameworkSpec {
           messages.mkString.split("\n").dropRight(1).mkString("\n"),
           List(
             s"${reset("info:")} ${red("- some suite")}",
+            s"${reset("info:")} ${"ignored: 1"}",         
             s"${reset("info:")}   ${red("- failing test")}",
             s"${reset("info:")}     ${blue("1")} did not satisfy ${cyan("equalTo(2)")}",
-            s"${reset("info:")}   ${green("+")} passing test"
+            s"${reset("info:")}   ${green("+")} passing test",
+            s"${reset("info:")}   ${"ignored: 1"}",     
           ).mkString("\n")
         )
       )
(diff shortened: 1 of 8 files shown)


regiskuckaertz:
I am of the opinion that those numbers are [chartjunk](https://en.wikipedia.org/wiki/Chartjunk), the only meaningful information is the test/suite label: it will both serve as a visual anchor in the rendering, but also as a navigation hotspot should you wish to go to the source of the test (with `ignored: 1` you would have to count the number of tests preceding and then go through the whole spec manually).

adamgfraser:
@regiskuckaertz I hear you. I think to some extent we need to decouple the test annotations from the way they are rendered, at least for now. I think the reason they are not useful right now is that they just appear in the middle of this text output. But what if they were instead part of some rich table where you could filter for only suites that had ignored tests, or sort by suites with the most ignored tests?

We're definitely not there today and we have to continue to provide a good user experience as we are getting there, but I think that is the vision and I am trying to keep an open mind as we get there even though I agree some of this is not great right now as it gets in the middle of rendering the test results.

adamgfraser:
Updated to incorporate #2499.

adamgfraser:
Updated to incorporate #2500.

jdegoes:
I think this would be the ideal output in this case:

```
[info] + TestAspectSpec — ignored: 8
[info]   + around evaluates tests inside context of Managed
[info]   + dotty applies test aspect only on Dotty
[info]   + dottyOnly runs tests only on Dotty
[info]   + exceptDotty runs tests on all versions except Dotty
[info]   + exceptJS runs tests on all platforms except ScalaJS
[info]   + failure makes a test pass if the result was a failure
[info]   + failure makes a test pass if it died with a specified failure
[info]   + failure does not make a test pass if it failed with an unexpected exception
[info]   + failure does not make a test pass if the specified failure does not match
[info]   + failure makes tests pass on any assertion failure
[info]   + failure makes tests pass on an expected assertion failure
[info]   + flaky retries a test that fails
[info]   + flaky retries a test that dies — retried: 4
[info]   + flaky retries a test with a limit
[info]   + ifEnv runs a test if environment variable satisfies assertion
[info]   + ifEnvSet runs a test if environment variable is set
[info]   + ifProp runs a test if property satisfies assertion
[info]   + ifPropSet runs a test if property is set
[info]   + js applies test aspect only on ScalaJS
[info]   + jsOnly runs tests only on ScalaJS
[info]   + jvm applies test aspect only on jvm
[info]   + jvmOnly runs tests only on the JVM
[info]   + retry retries failed tests according to a schedule
[info]   + scala2 applies test aspect only on Scala 2
[info]   + scala2Only runs tests only on Scala 2
[info]   + timeout makes tests fail after given duration
[info]   + timeout reports problem with interruption — time: 10ms
```

where we kind of assume the test annotations are rendered individually in a comma-separated list right after the label, and omit test annotations on ignored tests.

@regiskuckaertz @adamgfraser What do you think about the above? 

Tabular output would make this much nicer, of course. Ultimately we probably do need a table renderer, and to generate structured data rather than String from the renderers, so we can easily output plaintext, markdown, or HTML.

adamgfraser:
@jdegoes I think that looks like a good way to render it in the console. It looks like one implication of that is that renderers need to be able to apply different logic for rendering a suite versus a test (e.g. the `ignored` renderer is rendering the total number of ignored tests for the suite, but the `retried` renderer is not doing that).

jdegoes:
@adamgfraser Oops, that was a mistake. I think `retried` should / would appear in the same way at the suite level. Unless it would be more useful to apply a different rendering?

jdegoes:
Updated faux rendering...

adamgfraser:
@jdegoes I think it is much more regular if suites and tests are rendered the same way so let's try to go with that.

regiskuckaertz:
We also have the full UTF-8 gamut, just in case there's a less intrusive way to denote ignored tests by replacing `+` with, I don't know, something else 😅

adamgfraser:
Updated with new rendering.

jdegoes:
Awesome work!

jdegoes:
`Gen` annotations, when we add them, can probably use `repeat` to count repetitions as well (checks of a given test for random values), rather than having a separate annotation.

adamgfraser:
Yes, the trickier thing there will be what we want to do regarding the distribution of generated values.

PR #2503, opened 2019-12-19
Issue #2341 - added ZSink.drop

### LINKED ISSUE #2341: Add skip to ZSink/Sink
```scala
  // Skips n elements 
  def skip[A](n: Long): ZSink[Any, Unit, A, A, Unit] =
    ZSink.foldUntil[Long, A](0L, n)((s, _) => s + 1).flatMap {
      case m if n != m => ZSink.fail(())
      case _           => ZSink.succeed(())
    }

  // Gets the head of a stream
  def get[A]: ZSink[Any, Unit, A, A, A] =
    ZSink.foldUntil[Option[A], A](None, 1)((_, a) => Some(a)).flatMap {
      case None    => ZSink.fail(())
      case Some(a) => ZSink.succeed(a)
    }
```

I am not totally immersed into your ecosystem, so I did some testing in scalatest:
```scala
  property("test") {
    val rt = new DefaultRuntime {}

    {
      val r = Stream.fromIterable(List(1, 2, 3, 4, 5)).run(for {
        _ <- skip[Int](2)
        r <- get[Int]
      } yield r)
      assert(rt.unsafeRunSync(r).toEither.right.get == 3)
    }

    {
      val r = Stream.fromIterable(List(1, 2, 3)).run(for {
        _ <- skip[Int](2)
        r <- get[Int]
        _ <- skip[Int](1)
      } yield r)
      assert(rt.unsafeRunSync(r).toEither.isLeft)
    }

    {
      val r = Stream.fromIterable(List(1, 2, 3, 4)).run(for {
        _ <- skip[Int](2)
        r <- get[Int]
        _ <- skip[Int](1)
      } yield r)
      assert(rt.unsafeRunSync(r).toEither.right.get == 3)
    }
  }
```

pk044:
Closes #2341 .
It was supposed to be `skip` at first, but I changed it to `drop` to be more consistent with other abstractions (ZStream, ZStreamChunk etc.) - let me know if it's not a good idea and I'll change it :)

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -247,6 +247,7 @@ object SinkSpec extends ZIOBaseSpec {
       suite("drop")(
         testM("happy path") {
           val sink = ZSink.identity[Int].drop(1)
+<<<<<<< HEAD
           assertM(sinkIteration(sink, 1).either, isLeft(equalTo(())))
         },
         testM("init error") {
@@ -260,6 +261,21 @@ object SinkSpec extends ZIOBaseSpec {
         testM("extract error") {
           val sink = extractErrorSink.drop(1)
           assertM(sinkIteration(sink, 1).either, isLeft(equalTo("Ouch")))
+=======
+          assertM(sinkIteration(sink, 1), equalTo(List(())))
+        },
+        testM("init error") {
+          val sink = initErrorSink.drop(3)
+          assertM(sinkIteration(sink, 3).either, isLeft(equalTo("Ouch")))
+        },
+        testM("step error") {
+          val sink = stepErrorSink.drop(3)
+          assertM(sinkIteration(sink, 3).either, isLeft(equalTo("Ouch")))
+        },
+        testM("extract error") {
+          val sink = extractErrorSink.drop(3)
+          assertM(sinkIteration(sink, 3).either, isLeft(equalTo("Ouch")))
+>>>>>>> 8761fecb04052b63f09268bf10864db199793627
         }
       ),
       suite("dropWhile")(
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -244,6 +244,12 @@ object SinkSpec extends ZIOBaseSpec {
           assertM(sinkIteration(sink, "123").either, isLeft(equalTo("Ouch")))
         }
       ),
+      suite("drop")(
+        testM("happy path") {
+          val sink = ZSink.identity[Int].drop(2)
+          assertM(sinkIteration(sink, Stream.fromIterable(List(1,2,3))), equalTo(List(3)))
+        }
+      ),
       suite("dropWhile")(
         testM("happy path") {
           val sink = ZSink.identity[Int].dropWhile(_ < 5)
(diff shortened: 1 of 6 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1062,6 +1062,25 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
...
+        // Cast is redundant but required for Scala 2.11
+        val initial = sink.initial.map((_, 0L)).asInstanceOf[ZIO[R, E, this.State]]
+
+        def step(state: State, a: A) =
+          if (state._2 != n) UIO.succeed((state._1, state._2 + 1))
+          else sink.step(state._1, a).map((_, state._2))
+
+        def extract(state: State) = sink.extract(state._1)
+
+        def cont(state: State) = sink.cont(state._1)
```
vasilmkd: I would think the sink needs to continue for at least `n` iterations and then "switch over" to the underlying sink.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1062,6 +1062,25 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
...
+     */
+    final def drop(n: Long): ZSink[R, E, A0, A, B] =
+      new ZSink[R, E, A0, A, B] {
+        type State = (sink.State, Long)
+
+        // Cast is redundant but required for Scala 2.11
+        val initial = sink.initial.map((_, 0L)).asInstanceOf[ZIO[R, E, this.State]]
+
+        def step(state: State, a: A) =
+          if (state._2 != n) UIO.succeed((state._1, state._2 + 1))
```
vasilmkd: I'd prefer a more explicit `<` check here instead of `!=`.

streams/shared/src/main/scala/zio/stream/ZSink.scala
regiskuckaertz: or when both continue?
```scala
state._2 < n && sink.cont(state._1)
```

vasilmkd:
Not really. This would stop exactly after `n` are dropped. After that point it should be only `sink.cont`.

regiskuckaertz:
Hmm, but then it's as if `n` doesn't count at all, right? _If_ I understand what you mean:
```scala
if (state._2 < n)
  sink.cont(state._1)
else
  sink.cont(state._1)
```

vasilmkd:
The way I understand this combinator is: it demands to drop `n` elements. To me, that means that until `n` elements have passed through the sink, it will not even consult the underlying one. So I would write the `cont` method as follows: `def cont(s: State): Boolean = s._2 < n || sink.cont(s._1)`

streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
```diff
@@ -244,6 +244,24 @@ object SinkSpec extends ZIOBaseSpec {
...
+        testM("init error") {
+          val sink = initErrorSink.drop(1)
+          assertM(sinkIteration(sink, 1).either, isLeft(equalTo("Ouch")))
+        },
+        testM("step error") {
+          val sink = stepErrorSink.drop(1)
+          assertM(sinkIteration(sink, 1).either, isLeft(equalTo("Ouch")))
+        },
+        testM("extract error") {
+          val sink = extractErrorSink.drop(1)
```
regiskuckaertz: Can you please add tests that drop more than one element? As also, one that drops zero element. 🙏

PR #2504, opened 2019-12-19
Clean up partitionMap common behaviour

### LINKED ISSUE #2495: Backport partitionMap from scala 2.13
Currently we are implementing partitionMap ourselves in at least 3 places. [Here](https://github.com/zio/zio/blob/d2cbceda3cb97943fc11853acb3a043579ed824e/test-tests/shared/src/main/scala/zio/test/GenUtils.scala#L82), in ZIO#partitionMPar and ZIO#partitionMParN (#2487).

We can refactor this in terms of List#partitionMap and backport it to previous scala versions.

saraiva132:
Backporting partitionMap was considered but it created more trouble than help. closes #2495

### DIFF
--- a/core/shared/src/main/scala/zio/Exit.scala
+++ b/core/shared/src/main/scala/zio/Exit.scala
@@ -234,10 +234,10 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
 
 object Exit extends Serializable {
 
-  final case class Success[+A](value: A)                   extends Exit[Nothing, A]
-  final case class Failure[+E](cause: _root_.zio.Cause[E]) extends Exit[E, Nothing]
+  final case class Success[+A](value: A)        extends Exit[Nothing, A]
+  final case class Failure[+E](cause: Cause[E]) extends Exit[E, Nothing]
 
-  final def interrupt(id: Fiber.Id): Exit[Nothing, Nothing] = halt(_root_.zio.Cause.interrupt(id))
+  final def interrupt(id: Fiber.Id): Exit[Nothing, Nothing] = halt(Cause.interrupt(id))
 
   final def collectAll[E, A](exits: Iterable[Exit[E, A]]): Option[Exit[E, List[A]]] =
     exits.headOption.map { head =>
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/Exit.scala
```diff
@@ -234,10 +234,10 @@ sealed trait Exit[+E, +A] extends Product with Serializable { self =>
 
 object Exit extends Serializable {
 
-  final case class Success[A](value: A)                   extends Exit[Nothing, A]
-  final case class Failure[E](cause: _root_.zio.Cause[E]) extends Exit[E, Nothing]
+  final case class Success[+A](value: A)        extends Exit[Nothing, A]
+  final case class Failure[+E](cause: Cause[E]) extends Exit[E, Nothing]
```
adamgfraser: Great!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1692,7 +1692,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self &&& that
 }
 
-private[zio] trait ZIOFunctions extends Serializable {
+private[zio] trait ZIOIterableFunctions extends Serializable {
+  def partitionMap[A, A1, A2](iterable: Iterable[A])(f: A => Either[A1, A2]): (List[A1], List[A2]) =
```
adamgfraser: This needs to be `private[zio]` to avoid exposing it. I think I would probably just put the method in the ZIO companion object. Alternatively we could create a new `private[zio]` object in the `zio` package for this but I think it is probably not worth it if it is just going to contain a single method.

test-tests/shared/src/main/scala/zio/test/GenUtils.scala
```diff
@@ -80,8 +80,9 @@ object GenUtils extends DefaultRuntime {
     provideSize(sample(gen).flatMap(effects => ZIO.collectAll(effects.map(_.run))))(size)
 
   def partitionExit[E, A](eas: List[Exit[E, A]]): (List[Failure[E]], List[A]) =
-    eas.foldRight((List.empty[Failure[E]], List.empty[A])) {
-      case (Failure(e), (es, as)) => (Failure(e) :: es, as)
-      case (Success(a), (es, as)) => (es, a :: as)
-    }
+    partitionMap(eas)(_ match {
```
adamgfraser: You can get rid of the extra closure here and just do:

```scala
paritionMap(eas) {
  case Success(a) => ???
  case e @ Failure(_) => ???
}
```

Issue #2343, opened 2019-11-26
Consider making stepChunk not final on ZSink.


sir-wabbit:
```scala
  /**
   * Steps through a chunk of iterations of the sink
   */
  final def stepChunk[A00 >: A0, A1 <: A](state: State, as: Chunk[A1])(
    implicit ev: A1 =:= A00
  ): ZIO[R, E, (State, Chunk[A00])] = {
    val _   = ev
    val len = as.length

    def loop(state: State, i: Int): ZIO[R, E, (State, Chunk[A00])] =
      if (i >= len) UIO.succeed(state -> Chunk.empty)
      else if (self.cont(state)) self.step(state, as(i)).flatMap(loop(_, i + 1))
      else UIO.succeed(state -> as.asInstanceOf[Chunk[A00]].splitAt(i)._2)

    loop(state, 0)
  }
```
could be optimized for some `ZSink`s if it is not declared `final`.

EDIT: I am not even sure HotSpot uses `final` modifiers. I know that Scalac is supposed to be able to use those, but I don't trust Scalac at all :) I would argue that `final` is an anti-pattern except in very low-level highly optimized code.

vasilmkd:
@jdegoes might have some strong feelings on the scalac/jvm comment. I know the ZIO codebase strives for using the `final` modifier anywhere where it is possible to do so. Of course, a PR with these optimizations would be very welcome and I'd be very happy to take a look at.

iravid:
PRs welcome for dropping `final` from stepChunk!

guizmaii:
Closing as the `stepChunk` method doesn't seem to exist anymore

PR #2509, opened 2019-12-20
fix: Normalize final functions.

saraiva132:
I was being bothered by a lot of inconsistencies between final and non final functions. I only covered zio package (core). Afterwards, I extended this to also include vals and classes. 

## The rules followed

1. All methods on classes / traits declared final, by default.

2. No methods on objects declared final.

3. No methods on final classes declared final.

4. All classes inside objects should be defined final.

5. All final classes and potentially classes have their constructors and parameters private.

6. All vals are final. Either in objects or final classes. (Discussed with @regiskuckaertz I have yet to dig deep into the reasoning behind this. Hope he can shed more light)

7. Inside objects or final classes any private val or method does not need to be final. 

8. Inside objects or final classes any package private val or method must be final.

9. If the class has all members final. Make the class final. Remove final annotations from members

**These rules are followed until exceptions are found. But, until exceptions are found (for new code, I believe this should be the standard.**

Let me know your thoughts :D I will still go over everything again for a second round I believe I missed some small cases just to make everything normalized. Also, do let me know, if this is taking things too far.

### DIFF
--- a/core/js/src/main/scala/zio/internal/OneShot.scala
+++ b/core/js/src/main/scala/zio/internal/OneShot.scala
@@ -20,13 +20,13 @@ package zio.internal
  * A variable that can be set a single time. The synchronous,
  * effectful equivalent of `Promise`.
  */
-private[zio] class OneShot[A] private (var value: A) {
+private[zio] final class OneShot[A] private (var value: A) {
 
   /**
    * Sets the variable to the value. The behavior of this function
    * is undefined if the variable has already been set.
    */
-  final def set(v: A): Unit = {
+  def set(v: A): Unit = {
     if (v == null) throw new Error("Defect: OneShot variable cannot be set to null value")
     if (value != null) throw new Error("Defect: OneShot variable being set twice")
     value = v
@@ -35,12 +35,12 @@ private[zio] class OneShot[A] private (var value: A) {
   /**
    * Determines if the variable has been set.
    */
-  final def isSet: Boolean = value != null
+  def isSet: Boolean = value != null
 
   /**
    * Retrieves the value of the variable, blocking if necessary.
    */
-  final def get(): A = {
+  def get(): A = {
     if (value == null) throw new Error("Cannot block for result to be set in JavaScript")
     value
   }
@@ -51,5 +51,5 @@ object OneShot {
   /**
    * Makes a new (unset) variable.
    */
-  final def make[A]: OneShot[A] = new OneShot(null.asInstanceOf[A])
+  def make[A]: OneShot[A] = new OneShot(null.asInstanceOf[A])
 }
--- a/core/js/src/main/scala/zio/internal/PlatformLive.scala
+++ b/core/js/src/main/scala/zio/internal/PlatformLive.scala
@@ -26,9 +26,9 @@ object PlatformLive {
   lazy val Default = Global
   lazy val Global  = fromExecutionContext(ExecutionContext.global)
 
-  final def makeDefault(): Platform = fromExecutionContext(global)
+  def makeDefault(): Platform = fromExecutionContext(global)
 
-  final def fromExecutor(executor0: Executor): Platform =
+  def fromExecutor(executor0: Executor): Platform =
     new Platform {
       val executor = executor0
 
@@ -46,6 +46,6 @@ object PlatformLive {
       val tracing = Tracing(Tracer.Empty, TracingConfig.disabled)
     }
 
-  final def fromExecutionContext(ec: ExecutionContext, yieldOpCount: Int = 2048): Platform =
+  def fromExecutionContext(ec: ExecutionContext, yieldOpCount: Int = 2048): Platform =
     fromExecutor(Executor.fromExecutionContext(yieldOpCount)(ec))
 }
--- a/core/js/src/main/scala/zio/internal/impls/OneElementConcurrentQueue.scala
+++ b/core/js/src/main/scala/zio/internal/impls/OneElementConcurrentQueue.scala
@@ -22,7 +22,7 @@ import java.util.concurrent.atomic.{ AtomicBoolean, AtomicLong, AtomicReference 
 
 import zio.internal.MutableConcurrentQueue
 
-class OneElementConcurrentQueue[A] extends MutableConcurrentQueue[A] with Serializable {
+final class OneElementConcurrentQueue[A] extends MutableConcurrentQueue[A] with Serializable {
   private[this] final val ref = new AtomicReference[AnyRef]()
 
   private[this] final val headCounter   = new AtomicLong(0L)
@@ -33,13 +33,13 @@ class OneElementConcurrentQueue[A] extends MutableConcurrentQueue[A] with Serial
 
   override final val capacity: Int = 1
 
-  override final def dequeuedCount(): Long = headCounter.get()
-  override final def enqueuedCount(): Long = tailCounter.get()
+  override def dequeuedCount(): Long = headCounter.get()
+  override def enqueuedCount(): Long = tailCounter.get()
 
-  override final def isEmpty(): Boolean = ref.get() == null
-  override final def isFull(): Boolean  = !isEmpty()
+  override def isEmpty(): Boolean = ref.get() == null
+  override def isFull(): Boolean  = !isEmpty()
 
-  override final def offer(a: A): Boolean = {
+  override def offer(a: A): Boolean = {
     assert(a != null)
 
     var res     = false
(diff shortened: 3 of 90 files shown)


saraiva132:
Next steps in this endeavour to clean some stuff up is tackle more packages and explicitly tag methods with override modifier when it makes sense

saraiva132:
I wonder, from a perspective of extensibility, if there are drawbacks in making methods final by default.

regiskuckaertz:
Now that the final moodier is redundant in many places there and may lead to warnings in 2.14 or Dotty. Section 2.5.6 of the spec:

| final is redundant for object definitions. Members of final classes or objects are implicitly also final, so the final modifier is generally redundant for them, too. Note, however, that constant value definitions do require an explicit final modifier, even if they are defined in a final class or object. final is permitted for abstract classes but it may not be applied to traits or incomplete members, and it may not be combined in one modifier list with sealed.

saraiva132:
@regiskuckaertz So given that all efforts are towards dotty we can assume 2.14 is not a concern. I am aware of what you mentioned, however, I still believe this normalization has merit given the lack of standards across the codebase. I could have made it a bit more thorough even, I believe. When we migrate to dotty, where the final modifier does not make sense, it can be removed. I believe this to be the correct course of action but I am always ready to stand corrected! Thanks for the input :D

regiskuckaertz:
I'm not the best judge. To me, Scala is already a verbose language and I follow the principle of economy religiously. This project is a learning ground for a lot of people, that's why I'd much prefer to see all the `final` removed except where they are meaningful; it implicitly teaches how to use the language with parsimony. I'm pretty sure we will publish a coding guidelines at some point, it is just a natural consequence of the blooming ecosystem around ZIO.

adamgfraser:
I agree that we want to model best practices here and not have redundant code. I am probably guilty of including `final` in places where it is not necessary. If we are going to go through an exercise of standardizing on this we should definitely be sure we are standardizing on the "right" thing.

I think if we apply that to `ZIO` as a concrete example that would mean methods on the `ZIO` trait should be marked as `final` since it is an algebraic data type and we don't expect methods to be overridden but methods on the `ZIO` companion object should not be marked `final` because that would be redundant. Is that right? Is there any other reason we should be marking methods on objects as `final`?

saraiva132:
I was thinking about this for a bit. @adamgfraser What you mentioned about traits I agree 100%. About objects and final classes, I believe we want the `final` modifier to be the default, which is also what scala 3 enforces and we can remove that when migration is done.

@regiskuckaertz, I hear you and what you says to be honest makes a lot of sense. Can I ask the question the other way then. What is the benefit of not having the final modifier? Do we want any of this to be overridable? I think not. Also, defaulting to final will make it obvious that non-final methods are being overridden somewhere. This means that more information is being conveyed. This can be noticed for instance with `ZIO` `map` and `flatMap` ; Chunk `collect`, `collectM` and so on.

By the way! I hope I am not being too pushy with my arguments! :D I also just want to get this right and one way or the other I think it can be improved we just need to agree on how.

regiskuckaertz:
> I hope I am not being too pushy with my arguments!

Absolutely not, we're all here to push the boundaries with elegance and ROCK 🤘🕺 🎸 

> What is the benefit of not having the final modifier?

The same as when every time you say something, you don't start with "I'm going to say something". Even though it is correct, it is redundant.

> Do we want any of this to be overridable?

You can't override an object, can you?

adamgfraser:
I think if adding a keyword has no semantic significance we shouldn't be doing it, and we definitely shouldn't be doing an exercise to add it to places where it has no effect. I guess you could argue there is some simplicity in saying "everything should be final" but I think only using it on classes and traits that could potentially be extended is relatively straightforward and more correct.

saraiva132:
@adamgfraser You are absolutely correct. That will be the stance I will take. I haven't got time to do any work. These days surrounding Christmas have been chaotic. Objects and final classes will have **no final modifiers** since there is no added meaning to it. Additionally, classes inside objects that are not final, should be.

adamgfraser:
@saraiva132 No worries! Enjoy the holidays!

jdegoes:
I think as conventions going forward:

1. All methods on classes / traits declared final, by default.
2. No methods on objects declared final.
3. No methods on final classes declared final.

Seem reasonable?

Note that (3) may depend on `scalac` / `dotty` generating the correct JVM bytecode.

saraiva132:
@jdegoes I would add a forth one. All classes inside objects should be defined final.

jdegoes:
@saraiva132 Fixed, look good? /cc @adamgfraser

alexknvl:
https://github.com/zio/zio/issues/2343 is relevant I think. There could be situations where `final` methods have very efficient specializations for some subtypes.

adamgfraser:
@alexknvl Agreed. I think these are more guidelines than strict rules. Most of the time in the code base we expect types not to be extended at all or only through certain constructors provided but there are exceptions like `ZSink`.

saraiva132:
@alexknvl. There are also methods in ZIO like map and flatmap that, albeit methods of a trait, are being overridden. Same happens in chunk. There are always exceptions to the rule. Thank you for pointing that out and I appreciate if you have any more examples. Enjoy the festivities

saraiva132:
@jdegoes  @adamgfraser If I may. I would add a couple more:

5. All final classes and potentially classes have their parameters private. 
6. All vals are final. Either in objects or final classes. (Discussed with @regiskuckaertz  I have yet to dig deep into the reasoning behind this. Hope he can shed more light)
7.Should private vals be final? I suppose this one does not make sense.
8. Should private classes have their parameters private? I think we have this in a couple of places but the rational escapes me.

regiskuckaertz:
> All vals are final

`final` is just the hint the compiler needs to inline the RHS for all those `val`s.

jdegoes:
@saraiva132 

All right, so we're at:

1. All methods on classes / traits declared final, by default
2. No methods on objects declared final
3. No methods on final classes declared final
4. All classes inside objects should be defined final
5. All final classes have private constructors & constructor parameters
6. All vals declared final, either in objects or final classes
7. Private methods/vals not declared final since they are already final
8. Package-private vals/methods declared final
9. If a class has all final members, class declared final & final member annotations removed

saraiva132:
@jdegoes Thanks for the cleanup. I was wondering. Should this be reflected in any sort of guidelines in the contributing documentation ( I intend to review that). Or should those guidelines be strictly non code related and this be the guidelines for reviewing. So, implicit.

adamgfraser:
I think having gone to the trouble of coming up with this we should definitely document it somewhere, either in the contributor's guide or potentially in some sort of "ZIO best practices" document so the contributor's guide stays more focused on the "how" of contributing and then refers to that.

A couple of comments on these items:
"5. All final classes have private constructors and constructor parameters" - is this really true? It definitely doesn't apply to case classes. We don't use a ton of classes that aren't case classes but I think often you want one or more of the parameters to be a public `val`.
"6. All vals declared final, either in objects or final classes" - isn't `final` redundant here except for values that can qualify as a constant which is very limited (e.g. literals)?

saraiva132:
@adamgfraser

5 - That is a good point. On my third round I will give particular emphasis to how we are applying private modifiers in class parameters and what is the impact of making every single one of them private.

6 - The reasoning for making them `final` was described by @regiskuckaertz. I am afraid I cannot understand how a val, in the context of an object or final class, is ever not considered a constant.

saraiva132:
I will require to add yet another item guys. With sealed traits:

- Some extend nothing
- Some extends `Serializable`
- Some extend `Serializable with Product`

I was in favor of making them all extend `Serializable with Product`. Any drawbacks ?

alexknvl:
If a `sealed trait` is a parent to a bunch of `case object`s and `case classes`, then it should absolutely extend `Serializable with Product` for good UX. Extending from `Serializable` (even for non-sealed traits) probably won't hurt in any way but may help with Spark. But if a `sealed trait` does not have `case class`es as its children (idk whether there are any such cases in ZIO), extending from `Product` is unnecessary.

So to sum up, IMHO:
* Regular `trait`s should extend from `Serializable` but not from `Product`.
* `sealed trait`s that are ADTs should extend from `Product with Serializable`.
* `sealed trait`s that are not ADTs should extend from `Serializable` but not from `Product`.

saraiva132:
Thanks @alexknvl, for extending this also for regular traits and non ADT sealed trait. I will update my PR description

alexknvl:
On second thought, there might be some `trait`s that should never be captured in a closure and serialized though. It would be a very bad idea to serialize and then deserialize `Runtime` for example. Generally speaking, serializing anything with a mutable state (containing `Ref`s or non-transient `var`s or mutable data structures) is not a good idea. So more caution is necessary when adding `Serializable`. It should be safe for all `sealed trait` ADTs.

saraiva132:
TBH, I was not really thinking on touching any trait. Only ADT sealed trait which are more predictable. So, I am quite comfortable with crossing that one off the list.

saraiva132:
I will tackle this in separate PRs to minimize impact.

Issue #2485, opened 2019-12-16
Fix Broken Links in Documentation and Enable CI


adamgfraser:
I noticed that we have a decent amount of our Scaladoc that is not linking correctly. See [here](https://circleci.com/gh/zio/zio/46387?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for an example. This ticket is to fix all of these so that no more warning are generated and change CI settings so that a warning fails the build to prevent future regressions.

laurynasl-wix:
I've opened a pr for this: #2513

adamgfraser:
Great! Will take a look.

laurynasl-wix:
@adamgfraser The pr was merged, is there something more to do here?

adamgfraser:
@laurynasl-wix No, sorry. Closing now.

adamgfraser:
Closed by #2513.

PR #2513, opened 2019-12-22
Fix broken scaladoc links and fail build on scaladoc warnings

laurynasl-wix:
Adresses #2485. Note that I couldn't generate proper links from `IO, RIO, Task and URIO` to `ZIO.mapN` and `ZIO.mapParN` as it seems that scaladoc doesn't support disambiguating inherited methods (I've opened a pr to scaladoc here https://github.com/scala/scala/pull/8607). In the mean time I replaced those links to links directly to the `ZIO` object to avoid the warnings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ mdoc: &mdoc
     - run:
         name: Generate documentation
         command: |
-          ./sbt coreJVM/doc coreJS/doc streamsJVM/doc streamsJS/doc testJVM/doc testJS/doc
+          ./sbt -Dfatal.warnings=true coreJVM/doc coreJS/doc streamsJVM/doc streamsJS/doc testJVM/doc testJS/doc
           ./sbt ++${SCALA_VERSION}! mdoc
     - <<: *clean_cache
     - <<: *save_cache
--- a/core/shared/src/main/scala/zio/Fiber.scala
+++ b/core/shared/src/main/scala/zio/Fiber.scala
@@ -407,7 +407,7 @@ object Fiber {
    * @param id            The fiber's unique identifier
    * @param interruptors  The set of fibers attempting to interrupt the fiber or its ancestors.
    * @param executor      The [[zio.internal.Executor]] executing this fiber
-   * @param children      The fiber's forked children. This will only be populated if the fiber is supervised (via [[ZIO#supervised]]).
+   * @param children      The fiber's forked children.
    */
   final case class Descriptor(
     id: Fiber.Id,
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -452,19 +452,19 @@ object IO {
     ZIO.lock(executor)(io)
 
   /**
-   *  @see mapN in [[zio.ZIO$ ZIO]]
+   *  @see [[zio.ZIO.mapN[R,E,A,B,C]*]]
    */
   final def mapN[E, A, B, C](io1: IO[E, A], io2: IO[E, B])(f: (A, B) => C): IO[E, C] =
     ZIO.mapN(io1, io2)(f)
 
   /**
-   *  @see mapN in [[zio.ZIO$ ZIO]]
+   *  @see [[zio.ZIO.mapN[R,E,A,B,C,D]*]]
    */
   final def mapN[E, A, B, C, D](io1: IO[E, A], io2: IO[E, B], io3: IO[E, C])(f: (A, B, C) => D): IO[E, D] =
     ZIO.mapN(io1, io2, io3)(f)
 
   /**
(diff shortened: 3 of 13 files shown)


.circleci/config.yml
```diff
@@ -114,7 +114,7 @@ mdoc: &mdoc
     - run:
         name: Generate documentation
         command: |
-          ./sbt coreJVM/doc coreJS/doc streamsJVM/doc streamsJS/doc testJVM/doc testJS/doc
+          ./sbt -Dfatal.warnings=true coreJVM/doc coreJS/doc streamsJVM/doc streamsJS/doc testJVM/doc testJS/doc
```
adamgfraser: Great!

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -407,7 +407,7 @@ object Fiber {
    * @param id            The fiber's unique identifier
    * @param interruptors  The set of fibers attempting to interrupt the fiber or its ancestors.
    * @param executor      The [[zio.internal.Executor]] executing this fiber
-   * @param children      The fiber's forked children. This will only be populated if the fiber is supervised (via [[ZIO#supervised]]).
```
adamgfraser: We should keep this documentation even if we have to take out the link.

core/shared/src/main/scala/zio/IO.scala
```diff
@@ -452,39 +452,39 @@ object IO {
     ZIO.lock(executor)(io)
 
   /**
-   *  @see [[zio.ZIO.mapN]]
+   *  @see mapN in [[zio.ZIO$ ZIO]]
```
adamgfraser: I wonder if we can just move the methods defined on `ZIOFunctions` to the `ZIO` companion object. As far as I can tell we're not getting any benefit from having the trait and it was both solve this problem and slightly simplify the code. I asked about this on Discord so we'll see if anyone else knows why we did it this way.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -404,39 +404,46 @@ object TestAspect extends TimeoutVariants {
   def parallelN(n: Int): TestAspectPoly = executionStrategy(ExecutionStrategy.ParallelN(n))
 
   /**
-   * An aspect that restores a given [[Restorable]]'s state to its starting state after the test is run.
+   * An aspect that restores a given [[zio.test.environment.Restorable Restorable]]'s state to its starting state
```
adamgfraser: Do we need the second `Restorable` here? I thought this would be written as "An aspect that restores a given [[zio.test.environment.Restorable]]'s state to its start state after the test is run.", but I could be wrong. Similar question for the changes below.

core/shared/src/main/scala/zio/Fiber.scala
laurynasl-wix: If I understand correctly, `Fiber`s are always supervised now, no?

test/shared/src/main/scala/zio/test/TestAspect.scala
laurynasl-wix: [[zio.test.environment.Restorable]] is rendered as `zio.test.environment.Restorable` in the scaladoc html, [[zio.test.environment.Restorable Restorable]] is rendered as `Restorable`, which makes the text flow a bit better

core/shared/src/main/scala/zio/IO.scala
laurynasl-wix: Let me know, I can move the methods to the object and fix the links

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: Great point!

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Ah, great.

core/shared/src/main/scala/zio/IO.scala
adamgfraser: Yes, looking in the history it appears that this is a holdover from an older encoding where there were `LowerR` and `UpperE` type bounds that were abstract and then specified in objects extending this trait. So let's move everything to the companion object.

laurynasl-wix: Everything, or just `mapN` and `parMapN`?

adamgfraser: Everything and just delete the trait entirely.

laurynasl-wix: :+1:

laurynasl-wix: Done

PR #2514, opened 2019-12-23
ZIO Test: Don't Exit After Running Tests in Ammonite

### LINKED ISSUE #2510: ZIO Test: Improve Usability with Ammonite
Following up on discussion on Discord, the experience running ZIO Test specs from Ammonite is not great because although the specs can be run using their main methods, on completion the spec will exit the interactive session. This doesn't occur in in SBT because SBT traps that exit and throws a `SecurityException` which we then catch [here](https://github.com/zio/zio/blob/2aca6773ad471af04b986f73bff19100caea31b2/test/shared/src/main/scala/zio/test/RunnableSpec.scala#L58).

I haven't worked with Ammonite much so very open to suggestions here. The constraint is that I think we need to maintain the default behavior as exiting with a failure exit code so that we signal to tools like CI that tests have failed. One idea would be to take a command line argument like `-i` for interactive sessions that prevents the exit behavior. Another would be to see if there is some kind of environment variable we can access to indicate we are in an Ammonite session.

adamgfraser:
Resolves #2510. The Java main class is set to `ammonite.Main` when running an interactive session in Ammonite. So we check for that in `doExit` and don't exit if we are in an Ammonite session to avoid terminating the user's interactive session.

### DIFF
--- a/test/shared/src/main/scala/zio/test/RunnableSpec.scala
+++ b/test/shared/src/main/scala/zio/test/RunnableSpec.scala
@@ -56,6 +56,11 @@ trait RunnableSpec[R, E, L, T, S] extends AbstractRunnableSpec {
   }
 
   private def doExit(exitCode: Int): Unit =
-    try sys.exit(exitCode)
+    try if (!isAmmonite) sys.exit(exitCode)
     catch { case _: SecurityException => }
+
+  private def isAmmonite: Boolean =
+    sys.env.exists {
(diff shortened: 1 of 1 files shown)


adamgfraser:
@regiskuckaertz Yes. Unfortunately the `JAVA_MAIN_CLASS` is followed by an integer identifier that varies between systems and runs (e.g. `JAVA_MAIN_CLASS_19131`) so I don't think we can lookup a partial match in the map directly.

PR #2515, opened 2019-12-23
update ZStream#groupBy docs

mschuwalow:
also remove unused value.


hepin1989:
Support limit substreams

PR #2516, opened 2019-12-23
Lighter memoize with OneShot

regiskuckaertz:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -710,13 +710,17 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
-      _ <- (l.await *> ((self provide r) to p)).fork
-    } yield l.succeed(()) *> p.await
+  final def memoize: URIO[R, IO[E, A]] = {
+    import internal.OneShot
+    (ZIO.environment[R] <*> UIO.effectTotal(OneShot.make[A])).map { case (r, x) =>
+      IO.effectTotal {
+        if (x.isSet)
+          UIO.effectTotal(x.get)
+        else
+          self.tap(a => UIO.effectTotal(x.set(a))).provide(r)
```
iravid: `OneShot#set` throws if called after the value has been set, but we don't have anything guaranteeing it won't be called twice here, if I'm not mistaken.

regiskuckaertz: You mean in sth like `action.memoize.flatMap { a => a <&> a }`? Yeah 🤦‍♂

iravid: Maybe with a `Ref[Either[A, Promise]]` state machine?

regiskuckaertz:
Thought I'd find something more elegant but admittedly failed.

regiskuckaertz:
@iravid Yeah, I switched back to `Ref` but had to create a promise each time, which is ugly. Thanks a lot for reviewing 🤗

Issue #2318, opened 2019-11-24
ZIO Stream: Enable Tests on Dotty


adamgfraser:
Now that we have the test suite for ZIO Core running on Dotty we should try to enable the ZIO Stream test suite as well. We already identified one Dotty specific issue with `ZStream` in #2232 so it okay to mark some tests as `scala2Only` but we should identify all tests that are not passing as Dotty and make sure we have tickets to resolve those issues.

laurynasl-wix:
I did some investigations into enabling the tests, and it seems that there are no issues with the tests or code itself, but there are two issues in dotty `0.21.0-RC1` that are blocking from enabling them now: https://github.com/lampepfl/dotty/issues/7843 (zio.stream.StreamEffect#foldWhileManaged doesn't terminate) and https://github.com/lampepfl/dotty/issues/7840 (problems with type inference)

laurynasl-wix:
One thing I haven't figured out yet is why these tests hang when run without forking.

laurynasl-wix:
Turns out forking was a red herring: the actual issue is with running tests in parallel from sbt. It looks like there's a deadlock with static initializers somewhere (see [thread dump](https://gist.github.com/laurynasl-wix/19267b3d43cc99c173581bb2a1b4c343)), but I couldn't pin it down or create a small reproduction to report to dotty. I see two workarounds for this: replace `val`s in `object`s with `lazy val`s (which would have a negative impact on performance) or run tests sequentially in sbt. It would be great to get to the bottom of this and report to dotty, but I don't know how to go about reproducing a smaller case here

adamgfraser:
Closing as this functionality is already in place.

PR #2519, opened 2019-12-23
[WIP] Enable stream tests on dotty

laurynasl-wix:
Addresses #2318. 
This includes hacks to work around https://github.com/lampepfl/dotty/issues/7840.

Still trying to figure out what are the root causes, as there are quite a few tests that had to be disabled.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ scala_213: &scala_213
   SCALA_VERSION: 2.13.1
 
 scala_dotty: &scala_dotty
-  SCALA_VERSION: 0.21.0-RC1
+  SCALA_VERSION: 0.22.0-bin-20200107-21a5608-NIGHTLY
 
 jdk_8: &jdk_8
   JDK_VERSION: 8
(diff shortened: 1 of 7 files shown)


iravid:
@laurynasl-wix @adamgfraser Great job reporting all those issues to Dotty!! The`var` issue in particular looked nasty.

Looks like both were solved now. Can we use a nightly to avoid the workarounds?

adamgfraser:
@iravid Yes I think so.

laurynasl-wix:
Sorry about the delay, vacations and all that. It seems that most of the issues are solved with latest nightly, but there's a new one: https://github.com/lampepfl/dotty/issues/7877.

adamgfraser:
Thanks for your work on this! It looks like at this point the only changes are to configuration files and test files so shall we merge at this point so we get these tests running on Dotty and then we can follow up when the remaining open issue with Dotty gets resolved?

laurynasl-wix:
@adamgfraser I've updated to the latest nightly and I think this is OK to merge now. There's still one  outstanding issue that blocks from running streams tests in parallel regarding object initialization. I'm still trying to get a smaller repro for that, so that I can report it to dotty, but there's an easy workaround for now.

adamgfraser:
@laurynasl-wix Great!

@iravid You good with this?

PR #2521, opened 2019-12-24
Support dotty in sbt launch script

laurynasl-wix:
This PR add  support for dotty to the `sbt` launcher. Usage is similar to switching scala versions: 
```
./sbt -dotty
```

I found this useful when trying to build and test ZIO with dotty, so I think it might lower the entry bar  for working with dotty a bit. On the flip side, we'd probably need to keep this version in sync to what's in sbt itself and CI config


jdegoes:
🙏

PR #2523, opened 2019-12-25
Support Long index values ZStream#zipWithIndex

vasilmkd:
@iravid and @adamgfraser because I was meddling in your code. 😆


iravid:
Yeah this is a very good change. Stream indices can be arbitrarily large and certainly larger than max int.

vasilmkd:
Can I follow this up with `.drop` and `.take`? WDYT?

@iravid @adamgfraser

iravid:
Sure that makes sense

PR #2524, opened 2019-12-25
ZStream drop and take improvements

vasilmkd:
@iravid 

@adamgfraser I'd like to ask you specifically about the changes I had to make about the widening warnings. Are they sufficient or would you like to go further (accepting `Long` type arguments instead of `Int`).

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1130,7 +1130,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   /**
    * Drops the specified number of elements from this stream.
    */
-  final def drop(n: Int): ZStream[R, E, A] =
+  final def drop(n: Long): ZStream[R, E, A] =
     self.zipWithIndex.filter(_._2 > n - 1).map(_._1)
 
   /**
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2284,10 +2284,10 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       for {
         as      <- self.process
         counter <- Ref.make(0L).toManaged_
-        pull = counter.modify { c =>
-          if (c >= n) (Pull.end, c)
-          else (as, c + 1)
-        }.flatten
+        pull = counter.get.flatMap { c =>
+          if (c >= n) Pull.end
+          else as <* counter.set(c + 1)
+        }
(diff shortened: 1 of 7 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2270,15 +2270,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+  def take(n: Long): ZStream[R, E, A] =
     ZStream {
       for {
         as      <- self.process
-        counter <- Ref.make(0).toManaged_
-        pull = counter.modify { c =>
-          if (c >= n) (Pull.end, c)
-          else (as, c + 1)
-        }.flatten
+        counter <- Ref.make(0L).toManaged_
```
iravid: Was it necessary to change the formulation here?

vasilmkd:
@jdegoes Could you please review this PR? 🙏

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: I think it does if we adopt the stance that failed elements are not counted in, but there needs to be a fallback for when an element fails.
```scala
for {
  as <- self.process
  counter <- Ref.make(n).toManaged_
  pull = ref.get {
    case 0 => Pull.end
    case n => as.foldM({
      case None => Pull.end
      case _ => pull
    },
    a => counter.update(_ - 1) *> Pull.emit(a)
  }
yield pull
```

vasilmkd: This here swallows the error right? I think we do need to propagate it downstream, the user will have enough power to deal with it in the future.

regiskuckaertz: or in the past 🤗 with `stream.either.take(n).absolve`. I also think it's more intuitive that "take the first n" implies the first n _successful_ elements in this context, as errors are only first-class when they are inside the value channel.

vasilmkd: So how does this translate to the "counts (or doesn't) towards the `n` budget" debate?

regiskuckaertz: The version above doesn't count errors in the budget.

mschuwalow: That version would also run into issues if the user provided n is negative.
I agree with the only considering successful elements though.
Excited to see where this is going

regiskuckaertz:
Looks like you need to annotation the type of the stream

vasilmkd:
Happy New Year! @regiskuckaertz

PR #2533, opened 2019-12-27
zio-test: Custom JUnit runner for runnable specs

dkarlinsky:
Annotating your spec with `@RunWith(classOf[ZTestJUnitRunner])` or extending `DefaultSpecWithJUnit` should allow running zio tests under most build tools like bazel, maven, gradle. Such Specs can also be run from Intellij (and probably other IDEs)

* Refactored DefaultTestReporter to allow access to relevant rendering functions.
* Test results are not being mapped correctly to spec structure by Intellij - still trying to figure it out

ide

cc: @adamgfraser

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -241,13 +241,7 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
 lazy val testJunitRunnerJVM = crossProject(JVMPlatform)
   .in(file("test-junit"))
   .settings(stdSettings("zio-test-junit"))
-  .settings(
-    libraryDependencies ++= Seq(
-      "org.portable-scala" %%% "portable-scala-reflect" % "0.1.1",
-      "junit"              % "junit"                    % "4.12"
-    ),
-    mainClass in (Test, run) := Some("zio.test.sbt.TestMain")
-  )
+  .settings(libraryDependencies ++= Seq("junit" % "junit" % "4.12"))
   .dependsOn(test)
   .jvm
 
--- a/build.sbt
+++ b/build.sbt
@@ -86,7 +86,8 @@ lazy val root = project
     stacktracerJS,
     stacktracerJVM,
     testRunnerJS,
-    testRunnerJVM
+    testRunnerJVM,
+    testJunitRunnerJVM
   )
   .enablePlugins(ScalaJSPlugin)
 
@@ -237,6 +238,19 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
   .dependsOn(core)
   .dependsOn(test)
 
+lazy val testJunitRunnerJVM = crossProject(JVMPlatform)
+  .in(file("test-junit"))
+  .settings(stdSettings("zio-test-junit"))
+  .settings(
+    libraryDependencies ++= Seq(
+      "org.portable-scala" %%% "portable-scala-reflect" % "0.1.1",
+      "junit"              % "junit"                    % "4.12"
+    ),
+    mainClass in (Test, run) := Some("zio.test.sbt.TestMain")
+  )
+  .dependsOn(test)
+  .jvm
+
 lazy val testRunnerJVM = testRunner.jvm.settings(dottySettings)
 lazy val testRunnerJS = testRunner.js
   .settings(
@@ -256,8 +270,10 @@ lazy val examples = crossProject(JVMPlatform, JSPlatform)
   .jsSettings(libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC3" % Test)
   .dependsOn(testRunner)
 
-lazy val examplesJS  = examples.js
-lazy val examplesJVM = examples.jvm.settings(dottySettings)
+lazy val examplesJS = examples.js
+lazy val examplesJVM = examples.jvm
+  .settings(dottySettings)
+  .dependsOn(testJunitRunnerJVM)
 
 lazy val isScala211 = Def.setting {
   scalaVersion.value.startsWith("2.11")
--- a/build.sbt
+++ b/build.sbt
@@ -238,12 +238,13 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
   .dependsOn(core)
   .dependsOn(test)
 
-lazy val testJunitRunnerJVM = crossProject(JVMPlatform)
+lazy val testJunitRunner = crossProject(JVMPlatform)
   .in(file("test-junit"))
   .settings(stdSettings("zio-test-junit"))
   .settings(libraryDependencies ++= Seq("junit" % "junit" % "4.12"))
   .dependsOn(test)
-  .jvm
+
+lazy val testJunitRunnerJVM = testJunitRunner.jvm.settings(dottySettings)
 
 lazy val testRunnerJVM = testRunner.jvm.settings(dottySettings)
 lazy val testRunnerJS = testRunner.js
--- a/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
+++ b/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
@@ -14,6 +14,16 @@ import zio.test.TestFailure.{ Assertion, Runtime }
 import zio.test.TestSuccess.{ Ignored, Succeeded }
 import zio.test._
 
+/**
+ * Custom JUnit 4 runner for ZIO Test Specs.<br/>
+ * Any instance of [[zio.test.AbstractRunnableSpec]], that is a class (JUnit won't run objects),
+ * if annotated with `@RunWith(classOf[ZTestJUnitRunner])` can be run by IDEs and build tools that support JUnit.<br/>
+ * Your spec can also extend [[JUnitRunnableSpec]] to inherit the annotation.
+ * In order to expose the structure of the test to JUnit (and the external tools), `getDescription` has to execute Suite level effects.
+ * This means that these effects will be executed twice (first in `getDescription` and then in `run`).
+ * <br/><br/>
+ * Scala.JS is not supported, as JUnit TestFramework for SBT under Scala.JS doesn't support custom runners.
+ */
 class ZTestJUnitRunner(klass: Class[_]) extends Runner with Filterable with DefaultRuntime {
   private val className = klass.getName.stripSuffix("$")
 
--- a/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
+++ b/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
@@ -4,35 +4,20 @@ import org.junit.runner.manipulation.{ Filter, Filterable }
 import org.junit.runner.notification.{ Failure, RunNotifier }
 import org.junit.runner.{ Description, RunWith, Runner }
 import zio.ZIO.effectTotal
+import zio._
 import zio.clock.Clock
 import zio.test.FailureRenderer.FailureMessage.Message
-import zio.test.Spec.{ SuiteCase, TestCase }
+import zio.test.Spec.{ SpecCase, SuiteCase, TestCase }
 import zio.test.TestFailure.{ Assertion, Runtime }
 import zio.test.TestSuccess.{ Ignored, Succeeded }
-import zio.test.{ ExecutedSpec, _ }
-import zio.{ DefaultRuntime, URIO, ZIO }
-
-import scala.util.Try
+import zio.test._
 
 class ZTestJUnitRunner(klass: Class[_]) extends Runner with Filterable with DefaultRuntime {
   private val className = klass.getName.stripSuffix("$")
 
   private lazy val spec: AbstractRunnableSpec = {
-    Try(loadObject())
-      .getOrElse(
-        klass
-          .newInstance()
-          .asInstanceOf[AbstractRunnableSpec]
-      )
-  }
-
-  private def loadObject(): AbstractRunnableSpec = {
-    import org.portablescala.reflect._
-    val fqn = className + "$"
-    Reflect
-      .lookupLoadableModuleClass(fqn, klass.getClassLoader)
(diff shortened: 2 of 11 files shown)


test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
```diff
@@ -0,0 +1,143 @@
...
+              .flatMap(ZIO.foreach(_)(traverse(_, suiteDesc, path :+ label.toString)))
+              .catchAll(_ => ZIO.unit)
+              .unit
+        case TestCase(label, _) =>
+          effectTotal(description.addChild(testDescription(label.toString, path)))
+      }
+    unsafeRun(
+      traverse(filteredSpec, description)
+      // we expect to be able to extract structure without the environment
+        .provide(null.asInstanceOf[spec.Environment])
```
dkarlinsky: @adamgfraser When traversing the structure of the spec, I couldn't find a way to get the environment from the spec's runner. Just providing `null` for now, but this is ugly and may fail if people do rely on the environment for structure... Any ideas?

dkarlinsky: Ok, I found a way to get the environment from `TestExecutor`, but I had promote it to trait and expose the managed environment.

adamgfraser:
Thanks for working on this! This is great. I will review in more detail over the weekend but one initial question. Would it be possible to have the JUnit runner run specs that extend `AbstractRunnableSpec` instead of `DefaultSpecWithJUnit`? I think the ideal user experience would be if everyone could write their tests the same way (e.g. using `DefaultRunnableSpec`) and then you could just run them on different platforms using the appropriate runner versus having to change your code. Not sure if that is possible without forcing a `JUnit` dependency on everyone but we should see if we can do that.

dkarlinsky:
@adamgfraser I see two issues with adding the runner annotation to `AbstractRunnableSpec`:
1. Adding the JUnit dependency as you mentioned.
2. Tests will have to be classes, not objects to run under JUnit.

test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  private lazy val spec: AbstractRunnableSpec = {
+    klass
+      .newInstance()
+      .asInstanceOf[AbstractRunnableSpec]
+  }
+
+  private var filter = Filter.ALL
+
+  lazy val getDescription: Description = {
```
adamgfraser: Is the issue here that JUnit needs a more event based interface and the way ZIO Test is currently structured we run all the tests at once and then render all the tests at once?

Currently we're unsafely running the specs once to get the descriptions and then a second time to run the tests, which could create issues if there are suite level effects such as with `provideManagedShared`. Would it be possible to move to more of a model where we execute all the specs and then report their results? I think this would allow us to avoid the double evaluation problem and reuse more of the existing functionality in ZIO Test.

Separately I do think we want to look into having an interface where it is easier to get notifications in real time about test events instead of waiting for the full output since that is a request we have gotten for ZIO Intellij as well but I think that is outside the scope of this PR.

adamgfraser:
@dkarlinsky Agreed it would have been nice if we could keep the same interface for everything but it doesn't look like JUnit allows that. At least we can say that if you create a spec that works with JUnit that spec can also be run by the SBT test runner.

build.sbt
```diff
@@ -227,6 +228,14 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
   .dependsOn(core)
   .dependsOn(test)
 
+lazy val testJunitRunner = crossProject(JVMPlatform)
+  .in(file("test-junit"))
+  .settings(stdSettings("zio-test-junit"))
+  .settings(libraryDependencies ++= Seq("junit" % "junit" % "4.12"))
+  .dependsOn(test)
+
+lazy val testJunitRunnerJVM = testJunitRunner.jvm.settings(dottySettings)
```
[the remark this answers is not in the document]
dkarlinsky: I'm not how this would work and how to test it...

test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
```diff
@@ -0,0 +1,161 @@
...
+      notifier.fireTestIgnored(testDescription(label.toString, path))
+    }
+  }
+}
+
+private[junit] class TestFailed(message: String, cause: Throwable = null)
+    extends Throwable(message, cause, false, false)
+
+@RunWith(classOf[ZTestJUnitRunner])
+abstract class DefaultSpecWithJUnit extends DefaultRunnableSpec
```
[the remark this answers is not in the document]
dkarlinsky: Sure

test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
dkarlinsky: The leaf tests are not actually evaluated, just the `specs` effects in the suites.
I can actually return an empty `Descriprion` and just report test events as the happen.
(I'm actually reporting start/finish/fail events asynchronously already - look at `instrumentSpec` method).
In this case however, there is no way to report the nested structure of the Spec, the results just come out flat. 
The best I can do is to include the path in the label of each leaf test.

test/shared/src/main/scala/zio/test/TestExecutor.scala
```diff
@@ -16,13 +16,23 @@
...
+import zio.{ Managed, UIO, ZIO }
+
+/**
+ * A `TestExecutor[R, E, L, T, S]` is capable of executing specs containing
+ * tests of type `T`, annotated with labels of type `L`, that require an
+ * environment `R` and may fail with an `E` or succeed with a `S`.
+ */
+trait TestExecutor[+R, E, L, -T, +S] {
+  def run(spec: ZSpec[R, E, L, T], defExec: ExecutionStrategy): UIO[ExecutedSpec[E, L, S]]
+  def environment: Managed[Nothing, R]
```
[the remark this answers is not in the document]
dkarlinsky: I needed the environment for traversing the structure of the Spec in `getDescription`

build.sbt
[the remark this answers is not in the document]
adamgfraser: I'm not an expert on this but it looks like ScalaJS supports JUnit. See [here](https://github.com/scala-js/scala-js/blob/master/junit-plugin/src/main/scala/org/scalajs/junit/plugin/ScalaJSJUnitPlugin.scala). So maybe we could create a testJunitRunnerJS project as well and test that we are able to run a ZIO Test spec using that the same way we could run another JUnit test on Scala.js.

test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
adamgfraser: The problem is the `specs` effects in the suites could have real logic in them. They might create a Kafka client for example and then dispose of it.

Can we have the test executor execute the specs so we get one set of executed specs that don't have any environmental dependency and then traverse that to report everything?

dkarlinsky: Well, I don't think having the spec executed in `getDescription` instead of `run` will be following the JUnit Runner contract (if that's what you mean). IDEs and build tools won't expect this. And also you will loos the live reporting of the test as they run.
The only other option I see to prevent double evaluation is to have empty description and and accept flat spec structure being reported from inside the `run` method.

adamgfraser: Yes, we're running into an issue where specs in JUnit are static (there are just a certain number of tests sitting out there) versus tests in ZIO Test are dynamic (creating tests in a spec is itself an effect).

I see two options:
1. `getDescription` evaluates suite level effects and returns the full structure of the spec. The advantage is you get a description up front with all the suite and test labels. The disadvantage is that you run all suite level effects, like acquiring and releasing shared managed values, when you get the description, which to your point is probably not the contract for getting a description, and then you have to evaluate them again when calling `run`, which is inefficient.
2. `getDescription` just returns the label of the top suite or test and doesn't evaluate anything. I think we could still include information on the full ancestry of tests in events we report for each test by either traversing the spec and recursively passing down a list of ancestors or keeping track of it in a `Ref`. The disadvantage is that calling that `getDescription` method won't be that informative, but it would accurately reflect that we don't have static descriptions beyond the top level label before running the spec and would avoid the double evaluation issue.

I think the second approach is preferable. What do you think?

dkarlinsky: I'm good with 2, but I couldn't find a way to report the full ancestry of the leaves being executed.
I can the report the direct parent though. So what I can do is concatenate the full path in the parent name, so in IntelliJ it will look something like this:
```
SomeSpecWithJUnit
   + suite 1:suite 2
      - test1
      - test2 
   + suite 3:suite 4
      - test3
      - test4 
```

dkarlinsky: Unfortunately, I just discovered that while this works with IDEA JUnit runner, returning empty or fake description doesn't work with Bazel. It seems to expect tests to have accurate descriptions. This may also be the case for Gradle and Maven.

adamgfraser: Sorry I missed this. Can you say a little more about what you mean by "accurate descriptions"? Trying to understand if there is something we can do about that or that effectively means that Option 2 above is not possible.

dkarlinsky: Well, I can describe what is happening in our Bazel environment with Intellj with our customized version of Bazel Intellij plugin, when the runner returns empty description.
When running all tests under target (package) Bazel output reflects the tests run and failures, but IDEA is showing success and doesn't reflect tests run.
When running a single test using a filter like `--test_filter=com.wixpress.infra.zio.test.SimpleFailingTest#` we get  `No tests found matching RegEx[com.wixpress.infra.zio.test.SimpleFailingTest#]`.
It seems that bazel test filter for JUnit expects the class name to appear in the descriptions of leaf tests.

adamgfraser: So you think that if we got the full ancestry to the leaves it would?

dkarlinsky: I've been testing it in our project and if the `getDescription` returns the hierarchical structure of our spec with the spec class name included in the leaves everything works in our environment.
By class name included in the leaves I mean `Description.createTestDescription(className, label)` 

image

adamgfraser: Okay, so how about we do this?

Create a new simple algebraic data type for a description:

```scala
sealed trait SpecDescription extends Product with Serializable {
  // Converts to JUnit's representation of a description
  def toDescription: Description = ???
}
final case class TestDescription(label: String) extends SpecDescription
final case class SuiteDescription(label: String, specs: Vector[SpecDescription]) extends SpecDescription
```

Create an `AtomicReference` to store the description so it is safe for concurrent access, initially populated with just the top level spec name:

```scala
private val description: AtomicReference[SpecDescription] = ???
```

Then `getDescription` becomes:
```scala
`def getDescription = description.get.toDescription
```

Then the final step is in `traverse` we add an additional parameter `ancestors` and add the current label to `ancestors` each time we recurse. This will give us the information we need to update the description as we go.

I think this is all very doable. The upshot though is that `getDescription` will return a different description depending on when it is evaluated (e.g. if it evaluated before any tests are run it will just show the top level label, as suites and tests are created theyl be added to the description). So we need to make sure that works. Otherwise I think we will have to fall back to Option 1 above.

Thoughts?

dkarlinsky: Well at least for bazel, `getDiscription` is called before running the test, and filtering considers the description returned, this will not work, at least not for filtering by class name...
As far as I can tell, this is the contract - `getDescription` is called before `run` and is not expected to change if you called again.
It looks like Option 1 is the way to go, unfortunately.
Perhaps there could be a way to separate the suite effect into "dry run" mode - which only generates the structure and "actual run" mode that also performs setup and cleanup effects and such? 
What if SuiteSpec.specs was `Mode => ZIO[R, E, Vector[A]]` where Mode is either `Run` or `Describe`?

adamgfraser:
Okay, so sounds like we need to go with the first option. Did you get a chance to look at ScalaJS compatibility? Looks like ScalaJS supports JUnit so it would be nice for this to work across platforms? After we get that resolved I will give this one final review and then we should get it merged.

dkarlinsky:
Cool. 
Yeah, looks like ScalaJS indeed has support for running JUnit tests ([JunitTestFramework](https://github.com/scala-js/scala-js/blob/master/junit-runtime/src/main/scala/com/novocode/junit/JUnitFramework.scala)).
I will look into it.

dkarlinsky:
@adamgfraser So I looked into running under ScalaJS, and it looks like the JUnitTestFramework (for sbt) doesn't support custom runners in ScalaJS: 
https://github.com/scala-js/scala-js/blob/master/junit-runtime/src/main/scala/org/junit/runners/BlockJUnit4ClassRunner.scala#L6

adamgfraser:
Got it. Okay, I will review this weekend and then we can get this merged. Can you resolve the conflict?

adamgfraser:
Can you merge upstream changes and document the behavior with regard to suite level effects? Then let's get this merged.

dkarlinsky:
Sure. Where do think is best to document it? Scaladoc of the runner?

adamgfraser:
Yes. At some point we should also add something to the microsite about this but doesn't have to be part of this PR.

PR #2534, opened 2019-12-27
Add foreach_ to STM

mijicd:
While working on TMap optimizations noticed that some operations could benefit from this combinator.


jdegoes:
Looks great!

PR #2536, opened 2019-12-28
Minor cleanup on type annotations.

saraiva132:
@adamgfraser 

To follow up on type annotations on GADTS. I had noticed there were some left. Additionally, I believe the entire mock package in zio-test. `ArgumentExpectation`, `Method`, `Mockable` etc might need some review on its variance. I would love to do that as a learning experience. I did not have the time just yet. Perhaps, I am just plain wrong on that one.

Do you agree with my changes?

### DIFF
--- a/core/shared/src/main/scala/zio/FunctionIO.scala
+++ b/core/shared/src/main/scala/zio/FunctionIO.scala
@@ -206,8 +206,8 @@ object FunctionIO extends Serializable {
     final def unsafeCoerce[E2] = error.asInstanceOf[E2]
   }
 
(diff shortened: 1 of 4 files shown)


adamgfraser:
This looks great! Agreed it would be good to see if we can use variance more in the mocking framework. Do you want to do that here or as part of a separate PR?

saraiva132:
@adamgfraser Let us merge this one. I will be busy travelling and with NYE so I suspect it will take me 2-3 days to tackle it. So might as well close it instead of increasing scope, as I tend to do in previous work.

adamgfraser:
@saraiva132 Sounds good.

PR #2541, opened 2019-12-30
zio-test-mock variance improvements

### LINKED ISSUE #2540: ZIO Test: Mock package variance annotation
**Goals**

I believe the mock package is lacking in terms of type variance. I was playing around with it for a while with the following results:

**Where?**

Files in question from package mock are: `ArgumentExpectation`, `Method` and `Expectation`. Increased correctness by adding contravariant and covariant notation where it made sense.

**Issue found?**

 The compiler can no longer infer types where contravariant notation now exists without type annotation in some tests.

**Test  example**

```
testSpec("||")(
        Module.singleParam(equalTo(10) || isWithin(1, 5)) returns valueF(input => s"foo $input"),
        Module.>.singleParam(3),
        equalTo("foo 3")
      )
    )
```

```
[error]zio/test-tests/shared/src/test/scala/zio/test/mock/ExpectationSpec.scala:266:80: type mismatch;
[error]  found   : Any* => String
[error]  required: Seq[Any] => String
[error]         Module.singleParam(equalTo(10) || isWithin(1, 5)) returns valueF(input => s"foo $input"),
[error]                                                                                ^
[info] Any* => String <: Seq[Any] => String?
[info] false
```

**What is the problem then?**

The compiler thinks `Any* => String <: Seq[Any] => String?` is not sound. Compiler is always right :D
 
This forces me to add types in all places that use methods from `mock.Expectation` object like `valueF`, `valueM`, `failureF` and `failureM`. Which is not what we want.

**Conclusion**

I wanted add variance correctness while keeping 100% type inference or at least, not make it worse. If this is not possible, I will need to give up for now.

Any ideas / techniques to try / evidences to give the compiler?

saraiva132:
closes #2540

### DIFF
--- a/test/shared/src/main/scala/zio/test/mock/Method.scala
+++ b/test/shared/src/main/scala/zio/test/mock/Method.scala
@@ -24,7 +24,7 @@ import zio.test.Assertion
  * A `Model[M, I, A]` represents a capability of module `M` that takes an
  * input `I` and returns an effect that may produce a single `A`.
  */
-trait Method[-M, -I, +A] { self =>
+trait Method[-M, I, +A] { self =>
 
   /**
    * Provides the `Assertion` on method arguments `I` to produce `ArgumentExpectation`.
@@ -32,7 +32,7 @@ trait Method[-M, -I, +A] { self =>
    * Available only for methods that do take arguments.
    */
   @silent("parameter value ev in method apply is never used")
-  def apply[I1 <: I, A1 >: A](assertion: Assertion[I1])(implicit ev: I1 =!= Unit): ArgumentExpectation[M, I1, A1] =
+  def apply[A1 >: A](assertion: Assertion[I])(implicit ev: I =!= Unit): ArgumentExpectation[M, I, A1] =
     ArgumentExpectation(self, assertion)
 
   /**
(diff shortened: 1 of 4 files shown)


test/shared/src/main/scala/zio/test/mock/Method.scala
```diff
@@ -24,15 +24,15 @@ import zio.test.Assertion
...
+trait Method[-M, I, +A] { self =>
 
   /**
    * Provides the `Assertion` on method arguments `I` to produce `ArgumentExpectation`.
    *
    * Available only for methods that do take arguments.
    */
   @silent("parameter value ev in method apply is never used")
-  def apply(assertion: Assertion[I])(implicit ev: I =!= Unit): ArgumentExpectation[M, I, A] =
+  def apply[A1 >: A](assertion: Assertion[I])(implicit ev: I =!= Unit): ArgumentExpectation[M, I, A1] =
```
adamgfraser: I don't think the type parameter is necessary here since `A` doesn't appear anywhere in the method signature.

saraiva132: You are right.

PR #2545, opened 2019-12-31
ZIO Test: Migrate All Tests To Use Curried Assert

### LINKED ISSUE #2520: ZIO Test: Migrate Assertions to New Curried Syntax
In #2464 we changed `assert` and `assertM` to use curried syntax and deprecated the non-curried versions. We should change all usages in our own internal tests to use the new curried version. Because of the number of occurrences we may want to see if we can use a tool such as a Regex or Scalafix to automate this. This would have the advantage that users could also use the tool to update their own tests after we do the next release candidate.

adamgfraser:
Resolves #2520. I implemented a simple Scalafix rule to migrate all of our internal tests to use the new curried versions of `assert` and `assertM`. Do we have a place for this? We should publish a version of this when we do the next release so our users can upgrade their tests as well.

```scala
package fix

import scalafix.v1._
import scala.meta._

class CurriedAssert extends SemanticRule("CurriedAssert") {
  override def fix(implicit doc: SemanticDocument): Patch = {
    doc.tree.collect {
      case t @ Term.Apply(assert @ Term.Name("assert"), List(value, assertion)) =>
        Patch.replaceTree(t, assert + "(" + value + ")(" + assertion + ")")
      case t @ Term.Apply(assertM @ Term.Name("assertM"), List(value, assertion)) =>
        Patch.replaceTree(t, assertM + "(" + value + ")(" + assertion + ")")
      case t =>
        Patch.empty
    }.asPatch
  }
}
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,7 +119,6 @@ lazy val coreTests = crossProject(JSPlatform, JVMPlatform)
   .settings(skip in publish := true)
   .settings(Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat)
   .enablePlugins(BuildInfoPlugin)
-  .settings(scalacOptions ++= { if (isDotty.value) Seq() else Seq("-P:silencer:globalFilters=assert") })
 
 lazy val coreTestsJVM = coreTests.jvm
   .settings(dottySettings)
@@ -153,7 +152,6 @@ lazy val streamsTests = crossProject(JSPlatform, JVMPlatform)
   .settings(skip in publish := true)
   .settings(Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.AllLibraryJars)
   .enablePlugins(BuildInfoPlugin)
-  .settings(scalacOptions ++= { if (isDotty.value) Seq() else Seq("-P:silencer:globalFilters=assert") })
 
 lazy val streamsTestsJVM = streamsTests.jvm.dependsOn(coreTestsJVM % "test->compile")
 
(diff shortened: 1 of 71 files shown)


build.sbt
```diff
@@ -119,7 +119,6 @@ lazy val coreTests = crossProject(JSPlatform, JVMPlatform)
   .settings(skip in publish := true)
   .settings(Compile / classLoaderLayeringStrategy := ClassLoaderLayeringStrategy.Flat)
   .enablePlugins(BuildInfoPlugin)
-  .settings(scalacOptions ++= { if (isDotty.value) Seq() else Seq("-P:silencer:globalFilters=assert") })
```
saraiva132: What is the rational behind this removal?

jdegoes:
@adamgfraser Absolutely love a Scalafix migration! If we get into the habit of making these we can make breaking upgrades relatively painless (during new major versions anyway).

I'm not sure where the best place to put this is to make it accessible.

However, I've seen Scala Steward can [automatically apply migrations](https://github.com/fthomas/scala-steward/blob/master/docs/scalafix-migrations.md) if it is made aware of them. That could be the start of something interesting.

build.sbt
adamgfraser: I added this when we implemented the curried assert because otherwise you would get thousands of deprecation warnings for all the uses of of the old form of assert in our tests and the build would fail with fatal warnings enabled. Now that we have all the tests migrated we want those deprecation warnings so usage of the old form doesn't pass CI.

adamgfraser:
@jdegoes I'm going to merge this now given the scope of the changes and will follow up regarding making the migration rule accessible and making Scala Steward aware of it.

adamgfraser:
@jdegoes I looked into the migration rule more. The first thing we have to do is make the migration rule accessible. This can be done either by publishing it as a Gist or in a Github repository (either as part of this repo or a separate `zio/scalafix` repo). Then we can submit a PR to Scala Steward to make Scala Steward aware of it.

I think we should do it on Github instead of as a Gist so that we have better functionality for reviews and PRs. I think I would have a slight inclination towards just doing it as part of this repository so over time different ZIO projects can have their own rules in their own repositories versus one large shared repository for Scalafix rules but am happy to go either way.

PR #2550, opened 2019-12-31
Remove final defs, final vals, clean final class definitions, etc.

saraiva132:
Continuation of #2509. Finishes core package. With all sub-packages. On final defs, classes, privates, final classes, etc.


adamgfraser:
Fantastic!

PR #2553, opened 2020-01-01
Clean module zio-test. Stream. zio-test-sbt

saraiva132:
Remove/add  final defs, final vals, clean final class definitions, etc. 

Continuation of #2509. Finishes Clean module zio-test. Stream. zio-test-sbt

Almost done :)


test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
```diff
@@ -27,7 +27,7 @@ import zio.scheduler.{ Scheduler, SchedulerLive }
 import zio.test.Annotations
 import zio.test.Sized
 
-case class TestEnvironment(
+final case class TestEnvironment(
```
adamgfraser: I don't think we want to do this. Users need to be able to extend the `TestEnvironment` to provide their own environment types (e.g. `TestEnvironment with Kafka`). We should change `TestEnvironment` to a trait or a type alias to make it easier for users to do this but we can do that in a separate PR.

test/shared/src/main/scala/zio/test/DefaultTestRunner.scala
```diff
@@ -1,27 +0,0 @@
...
- */
-
-package zio.test
-
-import zio.test.environment._
-
-/**
- * A `Runner` that provides a default testable environment.
- */
-object DefaultTestRunner
```
adamgfraser: Let's move this to the ZIO Test package object as `val defaultTestRunner = TestRunner(TestExecutor.managed(testEnvironmentManaged))` and then we can refer to it in `DefaultRunnableSpec`.

test/js/src/main/scala/zio/test/environment/TestEnvironment.scala
```diff
@@ -47,7 +47,7 @@ case class TestEnvironment(
   /**
    * Maps all test implementations in the test environment individually.
    */
-  final def mapAll(
+  def mapAll(
```
adamgfraser: We should keep these methods `final`.

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -20,7 +20,7 @@ import scala.deprecated
 
 import zio.duration.Duration
 import zio.stream.{ ZSink, ZStream }
-import zio.test.environment.{ TestClock, TestConsole, TestRandom, TestSystem }
+import zio.test.environment.{ testEnvironmentManaged, TestClock, TestConsole, TestEnvironment, TestRandom, TestSystem }
```
adamgfraser: Maybe wildcard import at this point?

saraiva132: I am not a big fan of wildcard. This sounds like something to add to scalafix, if it is doable. I typically put wildcard when it passes the line limit. Scalafmt can also just put one import per line.

adamgfraser: Sounds good.

PR #2556, opened 2020-01-02
Build failed on master due to flaky test

saraiva132:
bracketFork release called on interrupt in separate fiber

Build failed is [48168](https://circleci.com/gh/zio/zio/48168?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).


adamgfraser:
Do we have an open issue for this? We should be sure we have an open issue for all flaky tests before marking them as flaky so we can track them. Also let’s please include a link to the failing build so other people can see any relevant details about the failures when diagnosing.

saraiva132:
@adamgfraser  I was not aware we required an issue since the flaky annotation is a flag on itself. The build, you are correct though. Makes sense though.

PR #2558, opened 2020-01-02
ZIO Test: Make TestEnvironment a Trait

adamgfraser:
`TestEnvironment` is currently a case class which is a little awkward because we want users to be able to extend it to provide their own environmental dependencies (e.g. `TestEnvironment with Kafka`). This PR changes it to be a trait so it is easier to mix in.


jdegoes:
Basically everything in the ZIO pattern (Console, Clock, etc.) is there for a reason, which may not be obvious but eventually proves useful.

Although hopefully we'll be obsoleting the pattern and make it a lot easier to build these things...

PR #2566, opened 2020-01-03
ZIO Test: Implement Polymorphic Generators

adamgfraser:
Following up on a request from @afsalthaj for ZIO Config, this PR implements basic support for polymorphic generators for GADTs where each generated value can be of a different type but we make sure that the types line up appropriately. Would appreciate any feedback on how we can make this more idiomatic and ergonomic.

Copying @leigh-perry.


test/shared/src/main/scala/zio/test/TypeWith.scala
```diff
@@ -0,0 +1,113 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test
+
+/**
```
mschuwalow: These docs are amazing

jdegoes:
This is really helpful! The only thing I might suggest to consider is tweaking the interface a bit. Does the user need the ability to assemble their own gens? Or can we do that for them? e.g.:

```scala
trait GenPoly {
  type T
  val genT: Gen[Random with Sized, T]
}
...
def genPoly: Gen[Random with Sized, GenPoly]
```

Then we "pre-bake" the gen with int, string, function, etc.

Thoughts?

Note that polymorphic code technically can be tested with any concrete type and if it works for that concrete type it should work for any other concrete type. That's why many tests hard code to int, string, etc. Of course, users might make more mistakes if they know what the type is, and maybe they don't want to think about what type it is, or maybe they want to ensure no reflection is being used, in which case this kind of machinery can be quite useful.

adamgfraser:
I really like that idea! I think it makes it a lot more ergonomic to use and feels more "ZIO".

The thing I think we need to think through a little bit is how we balance that against flexibility in two areas.

First, sometimes the user may need to generate types with more constraints than just a `Gen` existing. For example, they may want to generate types that have an `Ordering` or an instance of some other type class defined for them. One way we could do this that would I think preserve the benefits of your proposal would be to do something like:

```scala
  // ZIO Test provides this
  trait GenPoly {
    type T
    val genT: Gen[Random with Sized, T]
  }

  // User mixes it in to a type with their own constraints
  // Maybe we provide for limited number of standard library type classes like Ordering and Numeric
  trait GenOrderingPoly extends GenPoly { 
    val ordT: Ordering[T]
  }

  object GenOrderingPoly {
    def apply[A](gen: Gen[Random with Sized, A], ord: Ordering[A]): GenOrderingPoly =
      new GenOrderingPoly {
        type T = A
        val genT = gen
        val ordT = ord
      }
  }
```

The other area where users may want flexibility is in the specific types that are generated. Should `genPoly` include all primitive types? Integral versus floating point numeric types, which can sometimes cause issues in testing because of loss of precision but sometimes we want to test for? Collection types? 

But I agree that there is value in having some defaults and users can always create their own so I think we probably just try to come up with a few that make sense.

What do you think?

jdegoes:
I'd say it wouldn't be a mistake to include up to 200 types, maybe in a deterministic Gen. That seems overkill but some sampling from primitives, collections, functions, seems reasonable.

You're right about the "generate with constraint". I think that's at odds with providing "baked in" types. But possibly one could unify with `GenPoly[Constraint[_]]`, where `genPoly` would return maybe `genPoly[Ordering]` (or some type alias `type Std[A] = Ordering[A] with ...` that includes other standard type classes).

adamgfraser:
We could also provide a couple of "baked in" generators. Like we could provide `genPoly`, `genOrderedPoly`, genNumericPoly`, `genIntegralPoly`, and `genFractionalPoly`. We also need to be careful about including things like functions where equality is not well-defined and have few enough collections that you don't have a risk of a combinatorial explosion of nested collections. Let me take a turn on this and we can review again.

adamgfraser:

jdegoes:
@adamgfraser So where do we want to take this???

adamgfraser:
I think this is in pretty good shape.

The main question in my mind is whether we want to maintain the current structure where we have a hierarchy of `GenPoly`, `GenOrderedPoly`, `GenNumericPoly`, `GenIntegralPoly`, and `GenFractionalPoly` There is a bit of boilerplate on our end in having these different traits and aliasing different generators so you have all the generators of instances of the appropriate type in each companion object, but that is relatively manageable on our end and it allows the user to specify more precisely their constraints to generate values for particular cases.

On the margin we could probably add some more instances that we generate but we already have pretty good coverage.

My sense is that we should try to finalize something we're comfortable with and get something out there so people can use it and we can iterate on it.

jdegoes:
@adamgfraser Sounds good! Thanks for your work on this and the subsequent iteration. It looks great now!

PR #2568, opened 2020-01-03
Fix #2567: Return scalas specific duration type

### LINKED ISSUE #2567: Let `asScala` on `zio.duration` return specific sub type
The interop method `asScala` on `zio.duration.Duration` currently returns the requested instances as their common super type of `scala.concurrent.Duration` on conversion from zio to scala. Although the syntax on the `scala.concurrent.Duration` already captures finiteness and correctly reflects this on creation through the given syntax. However, when primarily using zio's duration this information gets lost in translation and we have to fiddle with the less specific supertype instead:
```
object test {
  import zio.duration._
  val x: scala.concurrent.duration.Duration = 200.millis.asScala
}
```
Through a small change in return types this type information can be preserved and makes this code fully typed and more specific without breaking changes. After that change this code compiles and preserves the finiteness of the captured literal syntax:
```
object test {
  import zio.duration._
  val x: scala.concurrent.duration.FiniteDuration = 200.millis.asScala
}
```

isaias-b:


jdegoes:
@isaias-b Thank you for this contribution, and congratulations on your first pull request!

giphy

isaias-b:
:bowing_man: ty, glad it went through :smiley:

PR #2569, opened 2020-01-03
Correct ZIO.effect(StdIn.readLine()) success type at docs/overview/creating_effects.md

alesavin:


alesavin:
> Thanks for the fix! Do you know why this was able to pass CI? It seems like it should have generated a warning for a discarded non-unit value that would have failed the build.

This is example at markdown documentation, not a real code, this things not covered by tests typically

adamgfraser:
The `mdoc` task should be checking that these code examples compile but we can investigate that separately. Thanks again and congratulations on your first contribution to ZIO!

PR #2560, opened 2020-01-02
Scalafix introduction with a naive configuration for sorting imports

saraiva132:
Closes #2512 

In context of #2549 

I need some feedback on the configuration to use. This could be a first step towards introducing scalafix. Next steps would be to add custom rules to tailor our needs.

Adds scalafix
Adds scalafix standard rules + sorting imports
Adds scalafix to CI

Adds commandAlias fix

PR #2572, opened 2020-01-04
Update welcome tasks and command alias

saraiva132:
Continuation of #2560

After some thought about this. What about, instead of merging fmt and fix, provide a very useful task that is intuitive that would be run before shipping code? (creating a PR)

To be honest, I could also not come up with a intuitive name to merge fix and fmt.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,7 @@ inThisBuild(
 
 ThisBuild / publishTo := sonatypePublishToBundle.value
 
+addCommandAlias("build", "clean; fix; fmt; testJVM")
 addCommandAlias("fix", "all compile:scalafix test:scalafix")
 addCommandAlias(
   "fixCheck",
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ inThisBuild(
 
 ThisBuild / publishTo := sonatypePublishToBundle.value
 
-addCommandAlias("build", "clean; fix; fmt; testJVM")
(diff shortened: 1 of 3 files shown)


adamgfraser:
How about adding a command alias `prepare` that does `fmt` and `fix`? I think the distinction between what Scalafix and Scalafmt does is somewhat artificial. Obviously the tooling has evolved a certain way where they are different tools but fundamentally both of them are supposed to "clean up" your code in some way that makes it nicer while preserving the meaning of your program. So I think we want to abstract over them in most cases (we can always keep the existing command alias if people want to run one but not the other.

I tend to think that we don't want to force people to do `clean` and `testJVM` every time they want to run `fmt` and `fix`. At least in my workflow a lot of the time I am making some final tweaks and just want to make sure everything is properly formatted but don't necessarily need to rerun the tests, much less recompile and run the entire test suite for all projects.

Can we also update the contributors guide as part of this PR?

saraiva132:
@adamgfraser thanks for the advice on command. I thought a run-it-all command would be good. Where basically sometimes at the end for some reason you might want to have something that will ensure enverything works, from scratch. I guess that was a bit too much.

Let me know if the contributor guide change makses sense.

adamgfraser:
Looks good! We can add `build` as well, I just want to make sure contributors have the option to run `fmt` and `fix` without doing everything else.

jdegoes:
@saraiva132 Superb!

PR #2575, opened 2020-01-04
ZIO Test: Implement Gen#crossAll and Gen#zipAll

### LINKED ISSUE #2573: ZIO Test: Implement Gen#sequence
Proposed signature:

```scala
def sequence[R, A](gens: List[Gen[R, A]]): Gen[R, List[A]]
```

We should consider if we want to have `sequence` and `sequencePar` variations implemented in terms of `crossWith` versus `zipWith` as well as `collectAll` variants and possibly `foreach` and `traverse` variants.

I will take the ticket.

adamgfraser:
Resolves #2573. I thought that a stream analogy was more appropriate here and more consistent with our other combinators for `Gen`. The signatures are:

```scala
def crossAll[R, A](gens: Iterable[Gen[R, A]]): Gen[R, List[A]]
def zipAll[R, A](gens: Iterable[Gen[R, A]]): Gen[R, List[A]]
```

`crossAll` returns the cartesian product of an utterable of generators and could be thought of as a version of `crossN` that requires all the generators to be of the same type but allows the number of generators to vary. `zipAll` pulls elements pairwise from each generator and could be thought of as a variant on `ZipN` in the same way.

We should make sure we are comfortable with these names since we should probably align them with similar combinators on `ZStream` if they are added at some point. Open to other suggestions.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/GenSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/GenSpec.scala
@@ -443,6 +443,42 @@ object GenSpec extends ZIOBaseSpec {
     },
     testM("runHead") {
       assertM(Gen.int(-10, 10).runHead)(isSome(isWithin(-10, 10)))
+    },
+    testM("crossAll") {
+      val gen = Gen.crossAll(
+        List(
+          Gen.fromIterable(List(1, 2)),
+          Gen.fromIterable(List(3)),
+          Gen.fromIterable(List(4, 5))
+        )
+      )
+      assertM(gen.runCollect)(
+        equalTo(
+          List(
+            List(1, 3, 4),
+            List(1, 3, 5),
+            List(2, 3, 4),
+            List(2, 3, 5)
+          )
+        )
+      )
+    },
+    testM("zipAll") {
+      val gen = Gen.zipAll(
+        List(
+          Gen.fromIterable(List(1, 2)),
+          Gen.fromIterable(List(3)),
+          Gen.fromIterable(List(4, 5))
+        )
+      )
+      assertM(gen.runCollect)(
+        equalTo(
+          List(
+            List(1, 3, 4),
+            List(2, 3, 5)
+          )
+        )
+      )
     }
   )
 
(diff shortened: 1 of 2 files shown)


test-tests/shared/src/test/scala/zio/test/GenSpec.scala
```diff
@@ -443,6 +443,42 @@ object GenSpec extends ZIOBaseSpec {
...
+          Gen.fromIterable(List(1, 2)),
+          Gen.fromIterable(List(3)),
+          Gen.fromIterable(List(4, 5))
+        )
+      )
+      assertM(gen.runCollect)(
+        equalTo(
+          List(
+            List(1, 3, 4),
+            List(2, 3, 5)
```
iravid: I'm surprised this one is not `List(2, 5)` because I assumed the `List(3)` would be exhausted. Does `Gen.fromIterable` create an infinite `Gen` that cycles the given iterable?

adamgfraser: `zipWith` repeats pulling from the shorter stream if necessary until the longer stream is exhausted.  In this specific case if we're combining multiple generators we could just return a shorter list, but in general if we are combining `Gen[A]` and `Gen[B]` to produce `Gen[(A, B)]` we can't do that because we need both an `A` and a `B`.

This is important because we model random generators as generators of a single effectual value, so if we combine a random generator with a deterministic generator we want to return a new generator with the size of the original deterministic generator that pulls each value from the deterministic generator, pulling a value from the random generator each time.

For combining deterministic generators both approaches would work I think but this also ends up being fairly useful because you make sure that you sample the entire space of the generators you are combining.

I tried to make this clear in the documentation, for example for `zipAll`, but let me know if there is something you would change.

```
  /**
   * Zips the specified generators together pairwise. The new generator will
   * generate elements as long as any generator is generating elements, running
   * the other generators multiple times if necessary.
   */
```

iravid: Ah interesting. That reasoning makes sense to me 👍 . Thanks for expanding on that.

PR #2580, opened 2020-01-05
ZIO Stream : ZStream[R,E,A] to Managed[R, Nothing, Iterator[Either[E,A]]]

ahoy-jon:
As talked in zio-dev, trying to add a conversion from `Stream` to `Managed[Iterator]`.

I am not sure yet where that code should go, either an opensource extension or into ZStream directly.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -19,6 +19,8 @@ package zio.stream
 import java.io.{ IOException, InputStream }
 import java.{ util => ju }
 
+import scala.annotation.tailrec
+
 import com.github.ghik.silencer.silent
 
 import zio._
@@ -2473,6 +2475,72 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
         }
       }
     } yield javaStream
+
+  /**
+   * Converts this stream into a `scala.collection.Iterator` wrapped in a [[ZManaged]].
+   * The returned iterator will only be valid within the scope of the ZManaged.
+   */
+  def toIterator: ZManaged[R, Nothing, Iterator[Either[E, A]]] =
+    for {
+      pull    <- this.process
+      runtime <- ZIO.runtime[R].toManaged_
+    } yield {
+      new Iterator[Either[E, A]] {
+        /*
+         * Internal state of a Iterator pulling from a ZStream
+         *
+         * It starts as Running
+         *
+         * when Running  , on pull (hasNext), pull the ZStream and switch to Closed or Value
+         * when Value    , on consume (next), return the Value and switch to Running
+         * when Closed   , on pull (hasNext), stays Closed
+         */
+        sealed trait State
+        case object Running                   extends State
+        sealed trait ValueOrClosed            extends State
+        case object Closed                    extends ValueOrClosed
+        case class Value(value: Either[E, A]) extends ValueOrClosed
+
+        var state: State = Running
+
+        private def pool(): Unit =
+          state = runtime.unsafeRun(
+            pull
+              .fold({
+                case None    => Closed
+                case Some(e) => Value(Left(e))
+              }, x => Value(Right(x)))
+          )
+
+        private def setRunning(): Unit =
+          state = Running
+
+        @tailrec
+        override def hasNext: Boolean =
+          state match {
+            case Closed   => false
+            case Value(_) => true
+            case Running =>
+              pool()
+              hasNext
+          }
+
+        @tailrec
+        override def next(): Either[E, A] =
+          state match {
+            case Closed => throw new NoSuchElementException("next on empty iterator")
+
+            case Value(a) =>
+              setRunning()
+              a
+
+            //Should not happen, next() has to be called after a successful hasNext() (Iterator spec)
+            case Running =>
+              pool()
+              next()
+          }
+      }
+    }
 
   /**
    * Throttles elements of type A according to the given bandwidth parameters using the token bucket
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -19,8 +19,6 @@ package zio.stream
 import java.io.{ IOException, InputStream }
 import java.{ util => ju }
 
-import scala.annotation.tailrec
-
 import com.github.ghik.silencer.silent
 
 import zio._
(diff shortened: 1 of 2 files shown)


ahoy-jon:
I didn't know `sbt fix`existed, it's so nice. 😃

adamgfraser:
It just got added thanks to @saraiva132.

ahoy-jon:
It took time to make it right, I need to learn how to do the checks on my machine directly.

@adamgfraser what do you think about this PR?

Should I remove the type `ZIterator` and work on a other version ?
```
ZStream[R,E,A] => ZManaged[R, Nothing, (Iterator[A], Promise[E,Unit])]

//The promise will allow works at the end of the Iterator
```

adamgfraser:
@ahoy-jon Sorry for the delay in reviewing this. I think this is interesting and something that would be great to have. A couple of initial thoughts:

1. We're moving to have element level errors on `ZStream`, so I think the signature might become `Managed[R, Nothing, Iterator[Either[E, A]]`.
2. We should think about what the contract is for finalizers. Do you have to do something special with regard to them or do the stream's finalizers get invoked when you pull the last element (maybe with an option for early release)?
3. I wonder whether we can simplify this considerably. For example, could this just convert the stream to a queue and have the iterator's `next` method just unsafely run pulling from the queue. That may be an oversimplification but I think there may be a way to do this using more existing combinators. Related to this I think we probably don't want to expose the `ZIterator` type and have that be an implementation detail, and potentially it can be inlined if we can simplify enough.

ahoy-jon:
@adamgfraser I am having delays on my side too!

1. Deal! Pull-rebased on master and updated.
2. To me, there are no finalizers. The Iterator is wrapped into a ZManaged, early finalization should stay outside. However, there is an extra argument to add an operation when the last element is being pulled.
```scala
  def fromStream[R, E, A](
    zStream: ZStream[R, E, A],
    afterLastPull: URIO[R, Any] = ZIO.unit
  ): ZManaged[R, Nothing, Iterator[Either[E, A]]]
```
3. I don't think a `ZQueue` would be simpler, next and hasNext seems closer to the Stream semantic. Maybe I didn't find a way to do it, on a ZQueue, we would need to do a :
```scala
ZIO.race(awaitShutdown.as(Closed), take.map(x => Value(x))
```
Using `ZStream.process` looks better.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3302,6 +3304,91 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
   def fromJavaIteratorManaged[R, E, A](iterator: ZManaged[R, E, ju.Iterator[A]]): ZStream[R, E, A] =
     managed(iterator).flatMap(StreamEffect.fromJavaIterator)
 
+  /**
+   * Creates a managed (impure) Iterator from a Stream
+   */
+  def toIterator[R, E, A](
+    stream: ZStream[R, E, A],
+    afterLastPull: URIO[R, Any] = ZIO.unit
```
adamgfraser: What if we get rid of the concept of `afterLastPull` here? We are already creating this in the context of a `ZManaged`, so once we exit the scope of the `ZManaged` the stream's finalizers will automatically be executed. If the caller wants to take some further action when the stream is empty that can do it at that point.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3302,6 +3304,91 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
...
+    case object Closed                    extends ValueOrClosed
+    case class Value(value: Either[E, A]) extends ValueOrClosed
+
+    for {
+      pull    <- stream.process
+      runtime <- ZIO.runtime[R].toManaged_
+    } yield {
+      new Iterator[Either[E, A]] {
+
+        var state: State = Running
```
adamgfraser: This is definitely not safe for concurrent access right now. I think that is probably okay as I think the general contractor of an `Iterator` does not include being safe for concurrent access but we should document that.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3302,6 +3304,91 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
...
+                case _      => ZIO.unit
+              })
+          )
+
+        @tailrec
+        override def hasNext: Boolean =
+          state match {
+            case Closed => false
+
+            //Should not happens, unless hasNext is called twice in a row.
```
adamgfraser: I think we have to assume that `hasNext` and `next` may be called in arbitrary sequences here.

streams/shared/src/main/scala/zio/stream/ZStream.scala
ahoy-jon: ~We can, but! The initial use case need it to be able to unsafely turn a  `ZManaged[_,_,Iterator[A]]` into a `ZIO[_,_,Iterator[A]]`.~

I just found another way to do it! Let's remove it.

ahoy-jon:
I tried to introduce a `RefM` for the state. "Concurrently unsafe" is great for now! 😃

```scala
  def toIterator[R, E, A](
    stream: ZStream[R, E, A]
  ): ZManaged[R, Nothing, Iterator[Either[E, A]]] = {
    /*
     * Internal state of a Iterator pulling from a ZStream
     *
     * It starts as Running
     *
     * when Running  , on pull (hasNext), pull the ZStream and switch to Closed or Value
     * when Value    , on consume (next), return the Value and switch to Running
     * when Closed   , on pull (hasNext), stays Closed
     */
    sealed trait State
    case object Running                   extends State
    sealed trait ValueOrClosed            extends State
    case object Closed                    extends ValueOrClosed
    case class Value(value: Either[E, A]) extends ValueOrClosed

    for {
      state   <- RefM.make[State](Running).toManaged_
      pull    <- stream.process
      runtime <- ZIO.runtime[R].toManaged_
    } yield {

      val pool: URIO[R, ValueOrClosed] =
        pull.fold({
          case None    => Closed
          case Some(e) => Value(Left(e))
        }, x => Value(Right(x)))

      val _next: RIO[R, Either[E, A]] = {
        def loop: State => URIO[R, (Either[Throwable, Either[E, A]], State)] = {
          case Closed   => UIO(Left(new NoSuchElementException("next on empty iterator")) -> Closed)
          case Value(x) => UIO(Right(x) -> Running)
          case Running  => pool >>= loop
        }

        state
          .modify(loop)
          .absolve
      }

      val _hasNext: URIO[R, Boolean] = {
        def loop: State => URIO[R, (Boolean, State)] = {
          case Closed   => UIO(false -> Closed)
          case x: Value => UIO(true -> x)
          case Running  => pool >>= loop
        }
        state.modify(loop)
      }

      new Iterator[Either[E, A]] {
        override def hasNext: Boolean     = runtime.unsafeRun(_hasNext)
        override def next(): Either[E, A] = runtime.unsafeRun(_next)
      }
    }
  }

```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3302,6 +3304,76 @@ object ZStream extends ZStreamPlatformSpecificConstructors with Serializable {
   def fromJavaIteratorManaged[R, E, A](iterator: ZManaged[R, E, ju.Iterator[A]]): ZStream[R, E, A] =
     managed(iterator).flatMap(StreamEffect.fromJavaIterator)
 
+  /**
+   * Creates a managed (impure) Iterator from a Stream
+   * /!\ *Unsafe* for concurrent access. /!\
+   */
+  def toIterator[R, E, A](
```
adamgfraser: Why don't we just make this a method on the `ZStream` trait instead of the companion object so you can just do `stream.toIterator`? This would be analogous to the existing `toInputStream` method. There is also some good Scaladoc on that one about the iterator only being valid within the scope of the managed that we could borrow from.

streams-tests/jvm/src/test/scala/zio/stream/ToIteratorSpec.scala
```diff
@@ -0,0 +1,32 @@
+package zio.stream
+
+import zio.ZIOBaseSpec
+import zio._
+import zio.test.Assertion.equalTo
+import zio.test._
+
+object ToIteratorSpec extends ZIOBaseSpec {
```
adamgfraser: Why don't we just move this into `StreamSpec` at this point since it is not that large to avoid creating another file? May also want to inline `testMM` since we are only using it one place and just call `.use(ZIO.succeed)` at the end

streams/shared/src/main/scala/zio/stream/ZStream.scala
ahoy-jon: I didn't like the visibility of it, it is so impure. I will do it! 👍

ahoy-jon:
@adamgfraser about `ZIO.succeed`, doesn't work, the compiler need more info : 

```
[error] /Users/jon/Projects/zio/streams-tests/jvm/src/test/scala/zio/stream/StreamSpec.scala:1982:62: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [A](=> A) => zio.UIO[A]
[error]     (which expands to)  [A](=> A) => zio.ZIO[Any,Nothing,A]
[error]  required: zio.test.package.TestResult => zio.ZIO[zio.stream.StreamSpec.Environment,Any,zio.test.TestResult]
```
It would work if : (without the lazy `=> A`)
```scala
  def succeed[A](a: A) = effectTotal(a)
```

I used instead `ZIO.succeedNow`.

adamgfraser:
Yes, that's perfect. `ZIO.succeedNow` is better anyway for internal combinators.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2474,6 +2476,72 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+  /**
+   * Converts this stream into a `scala.collection.Iterator` wrapped in a [[ZManaged]].
+   * The returned iterator will only be valid within the scope of the ZManaged.
+   */
+  def toIterator: ZManaged[R, Nothing, Iterator[Either[E, A]]] =
+    for {
+      pull    <- this.process
+      runtime <- ZIO.runtime[R].toManaged_
+    } yield {
+      new Iterator[Either[E, A]] {
```
adamgfraser: This is looking really good now. I think the only thing left is the internal state machine feels a little heavy, especially when we aren't managing concurrent state. Could we just implement the iterator as something like this?

```scala
new Iterator[Either[E, A]] {
  var nextTake: Take[E, A] = null
  def unsafeTake(): Unit =
    nextTake = runtime.unsafeRun(Take.fromPull(pull))
  def hasNext: Boolean = {
    if (nextTake == null) {
      unsafeTake()
    }
    !nextTake.isEnd
  }
  def next(): Either[E, A] = {
    if (nextTake == null) {
      unsafeTake()
    }
    val take = nextTake match {
      case Take.End      => throw new NoSuchElementException("next on empty iterator")
      case Take.Fail(e)  => Left(e.failureOrCause.fold(identity, c => throw FiberFailure(c)))
      case Take.Value(a) => Right(a)
    }
    nextTake = null
    take
  }
}
```

ahoy-jon: Indeed, so much better !

PR #2581, opened 2020-01-05
ZIO Test: Implement Generator for UUIDs

### LINKED ISSUE #2576: ZIO Test: Implement Gen#UUID
It is a relatively common type so we should have a generator for it. Also, without one users may lift `UUID.randomUUID` into a `Gen` which means that the generator will use Java's random number generator instead of our purely functional random number generator and thus tests will not be fully deterministic.

I will take the ticket.

adamgfraser:
Resolves #2576.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -238,8 +238,8 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
       mostSigBits  <- Gen.anyLong.noShrink
       leastSigBits <- Gen.anyLong.noShrink
     } yield new UUID(
-      (mostSigBits & ~0x0000f000) | 0x00004000,
-      (leastSigBits & ~(0xc0000000L << 32)) | (0x80000000L << 32)
+      (mostSigBits & ~0x0000F000) | 0x00004000,
+      (leastSigBits & ~(0xC0000000L << 32)) | (0x80000000L << 32)
     )
 
   /**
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -221,6 +229,19 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
...
+   * A generator of universally unique identifiers. The returned generator will
+   * not have any shrinking.
+   */
+  val anyUUID: Gen[Random, UUID] =
+    for {
+      mostSigBits  <- Gen.anyLong.noShrink
+      leastSigBits <- Gen.anyLong.noShrink
+    } yield new UUID(
+      (mostSigBits & ~0x0000f000) | 0x00004000,
+      (leastSigBits & ~(0xc0000000.toLong << 32)) | (0x80000000.toLong << 32)
```
regiskuckaertz: You can avoid the conversion here
```suggestion
      (leastSigBits & ~(0xc0000000L << 32)) | (0x80000000L << 32)
```

adamgfraser: Great suggestion!

NeQuissimus:
Why did you not just wrap `UUID.randomUUID`?

adamgfraser:
@NeQuissimus `UUID.randomUUID` uses `java.util.Random` internally. So if we wrapped it our generators wouldn't be fully deterministic (e.g. if I set the random set to a specified value and generated a `UUID` I wouldn't get the same value each time).

NeQuissimus:
Oh, yeah. Makes sense, thx :-)

*deleting my custom uuid gen* 😜

PR #2582, opened 2020-01-05
ZIO Test: Implement bounded generators

### LINKED ISSUE #2578: ZIO Test: Generator for bounded strings
Lots of time is necessary to generate bounded texts, there are already few solutions:

```
Gen.int(N, M).flatMap(Gen.stringN(_)(Gen.alphaNumericChar))
```
Quite long and unfriendly for beginners

Or using `Sized` which is even more unfriendly since needs fiddling with `provideSome` afte `flatMapping` on `Sized.makeService`

Or by `Gen.anyString.map(_.take(M))` which is not great either

I think it would be useful to have something like `Gen.boundedString(N,M)` that will generate strings of min N,max M size, with a a default for char generator.

adamgfraser:
Resolves #2578. Open to feedback on naming. I couldn't really think of any other combinators in ZIO that had both upper and lower bounds like this.


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -227,6 +234,12 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
   val boolean: Gen[Random, Boolean] =
     elements(false, true)
 
+  /**
+   * A generator whose size falls within the specified bounds.
+   */
+  def bounded[R <: Random, A](min: Int, max: Int)(f: Int => Gen[R, A]): Gen[R, A] =
```
regiskuckaertz: Another name I've seen in the wild is `range`, but this is as good as any.

jdegoes:
@adamgfraser Looks great! Just a few merge conflicts to fixup

adamgfraser:
@jdegoes Fixed now.

PR #2583, opened 2020-01-06
migrate TimeVariantsSpec into GenSpec

unclebob418:
#1898 `TimeVariantsSpec` has been moved into `GenSpec` & common methods moved into `GenUtils`

@adamgfraser can you please take a look?

Looks like IntelliJ lied to me about the "Safe Delete" of `GenUtils` so I have re-created it.

I had to create a new implicit Ordering for `LocalDateTime`. Should this be moved somewhere else? I didn't want to put it in `GenUtils` in case someone accidentally imports it.

Oddly this wasn't an issue for `OffsetDateTime`. Do you know why this is?


adamgfraser:
@unclebob418 Not sure why that is happening. It looks like both of the implicit exist.

```scala
import java.time.LocalDateTime
import java.time.OffsetDateTime

implicitly[Ordering[LocalDateTime]]
implicitly[Ordering[OffsetDateTime]]
```

https://scastie.scala-lang.org/80YZ1FS7TCOkzBgMgsK0eQ

unclebob418:
When I comment out the implicit I get:
```
[info] Compiling 1 Scala source to C:\Users\rober\IdeaProjects\zio\test-tests\jvm\target\scala-2.12\test-classes ...
[error] C:\Users\rober\IdeaProjects\zio\test-tests\shared\src\test\scala\zio\test\GenSpec.scala:215:77: No implicit Ordering defined for java.time.LocalDateTime.
[error] Error occurred in an application involving default arguments.
[error]         checkSample(Gen.localDateTime(min, max))(forall(isGreaterThanEqualTo(min) && isLessThanEqualTo(max)))
[error]                                                                             ^
[error] C:\Users\rober\IdeaProjects\zio\test-tests\shared\src\test\scala\zio\test\GenSpec.scala:215:103: No implicit Ordering defined for java.time.LocalDateTime.
[error] Error occurred in an application involving default arguments.
[error]         checkSample(Gen.localDateTime(min, max))(forall(isGreaterThanEqualTo(min) && isLessThanEqualTo(max)))
```

adamgfraser:
Okay, will take a look tomorrow.

unclebob418:
Thanks!

test-tests/shared/src/test/scala/zio/test/GenZIOSpec.scala
```diff
@@ -38,4 +38,5 @@ object GenZIOSpec extends ZIOBaseSpec {
       }
     }
   )
+
```
adamgfraser: Extra line here?

test-tests/shared/src/test/scala/zio/test/GenSpec.scala
```diff
@@ -1,16 +1,20 @@
...
-import zio.stream.ZStream
 import zio.test.Assertion._
+import zio.test.GenUtils._
 import zio.test.TestAspect.scala2Only
-import zio.test.environment.TestRandom
 import zio.test.{ check => Check, checkN => CheckN }
-import zio.{ Managed, UIO, ZIO }
 
 object GenSpec extends ZIOBaseSpec {
+  implicit val localDateTimeOrdering: Ordering[LocalDateTime] = _ compareTo _
```
adamgfraser: Looked into this a bit more. It looks like this implicit is actually available on Scala 2.13.1 but not on 2.12.10. Specifically, they use a slightly different encoding to derive an `Ordering` for any type for which `Comparable` is defined (which includes `LocalDateTime`). In Scala 2.13:

```scala
type AsComparable[A] = A => Comparable[_ >: A]

implicit def ordered[A](implicit asComparable: AsComparable[A]): Ordering[A] = new Ordering[A] {
  def compare(x: A, y: A): Int = asComparable(x).compareTo(y)
}
```

In Scala 2.12:

```scala
implicit def ordered[A <% Comparable[A]]: Ordering[A] = new Ordering[A] {
  def compare(x: A, y: A): Int = x compareTo y
}
```

For some reason the view bound in Scala 2.12 doesn't work, possibly because `A` is constrained to be the same type versus a super type as in Scala 2.13 and `LocalDateTime` actually implements `Comparable[ChronoLocalDateTime`.

We could potentially create a compatibility trait that would provide this instance but given that it is only used in this file I don't think it is really necessary.

unclebob418: Thanks - I guess it's a nice improvement for 2.13 :) 

I did find the 2.12 definition and guessed the reason it wasn't working was because the A type didn't line up. It looks like OffsetDateTime directly implements `compareTo` while LocalDateTime overrides `compareTo` on in `ChronoLocalDateTime`, but since `ChronoLocalDateTime` implements `compareTo` I would have thought it should still be considered a `Comparable`.

I'll throw that type signature (`type AsWhatever[A] = A => Whatever[_ >: A]`)  in my bag of tricks - looks handy for these sort of circumstances!

PR #2586, opened 2020-01-06
Port tapBoth & tapError to ZManaged 

zach-albia:
Helping out with #2077.

Added `ZManaged#tapBoth` & `ZManaged#tapError`, along with tests. Also added tests for `ZManaged#tap`.


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -715,6 +715,20 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
   def tap[R1 <: R, E1 >: E](f: A => ZManaged[R1, E1, Any]): ZManaged[R1, E1, A] =
     flatMap(a => f(a).as(a))
 
+  /**
+   * Returns an effect that effectfully peeks at the failure or success of the acquired resource.
+   */
+  final def tapBoth[R1 <: R, E1 >: E](f: E => ZManaged[R1, E1, Any])(g: A => ZManaged[R1, E1, Any])(
+    implicit ev: CanFail[E] // curried for better type inference
```
iravid: This isn't curried on ZIO - would be great if we can keep things consistent.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -715,6 +715,20 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
+  ): ZManaged[R1, E1, A] =
+    flatMap(a => g(a) *> tapError(f))
+
+  /**
+   * Returns an effect that effectfully peeks at the failure of the acquired resource.
+   */
+  final def tapError[R1 <: R, E1 >: E](f: E => ZManaged[R1, E1, Any])(implicit ev: CanFail[E]): ZManaged[R1, E1, A] =
+    flatMap(a => flip.tap(f).flip.as(a))
```
iravid: Pretty sure you need to implement this using `foldM`, otherwise `f` will not be executed when `this` fails.

core/shared/src/main/scala/zio/ZManaged.scala
iravid: Also, I think you need to implement this using `foldM`, otherwise `g` will not be executed on error.

zach-albia: Would this be an ok implementation then?
```scala
flatMap(a => foldM(f(_).as(a), a => g(a).as(a)))
```

iravid: Unfortunately not - `flatMap` will only execute the provided function if the `this` instance succeeds.

You need something like:
```
foldM(
  e => f(e) *> ZManaged.fail(e),
  a => g(a).as(a)
)
```

zach-albia: Just asking for my info, is `tap` written as: `flatMap(a => f(a).as(a))` OK because we're not peeking failures?

iravid: Exactly right.

PR #2588, opened 2020-01-06
docs(testing) fix typo, should be ++

benetis:
Copy pasting noticed a small typo which prevents sbt from building on _https://zio.dev/docs/usecases/usecases_testing_ . This a PR to fix it

### DIFF
--- a/docs/usecases/testing.md
+++ b/docs/usecases/testing.md
@@ -10,7 +10,7 @@ libraryDependencies ++= Seq(
   "dev.zio" %% "zio-test"     % zioVersion % "test",
   "dev.zio" %% "zio-test-sbt" % zioVersion % "test"
 )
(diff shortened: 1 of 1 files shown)


adamgfraser:
I don't think this is right. From the SBT documentation [here](https://www.scala-sbt.org/1.x/docs/Testing.html):

```scala
testFrameworks += new TestFramework("custom.framework.ClassName")
```

Also the projects I have seen use the same pattern (for example in ZIO Kafka [here](https://github.com/zio/zio-kafka/blob/aff57de81fee7f1a665f7b9eb26b44346e1cc2f5/build.sbt#L81)) and in fact fail to compile with `++=` instead of `+=`. Are you using multiple test frameworks? I think if you wanted to add multiple test frameworks then you would use `++= Seq(...)`.

benetis:
Maybe I am just sleep deprived, but: 

image

Issue #2549, opened 2019-12-31
Update contributing guidelines with coding rules and recommendations.


saraiva132:
Triggered by #2509 and #2463

Closes  #190, #335, #2462, #2512 

## Possible formats

- https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines
- https://docs.scala-lang.org/style/
- https://github.com/databricks/scala-style-guide

## Scalafix project inspiration

- https://github.com/scalaz/scalazzi
- https://github.com/alejandrohdezma/sensitive-strings
- https://github.com/scalatest/autofix
- https://github.com/pdalpra/scalalint

## Solution

Some rules will be able to be automated and others will remain as general guidelines

- Documentation with expectations about how things should look.

- Scalafix CI and plugin automation to make sure the rules are followed by existing and future code.

### Extensions:

1. value classes must be final and extend AnyVal.
2. method extension classes must be final and extend AnyVal
3. Sealed traits that are ADTs should extend from Product with Serializable.
4. Regular traits and sealed trait that do not form ADTS should extend from Serializable but not from Product.
5. Multiple parameter classes that are ADT should be case class and not extend AnyVal
6. Traits should always extends serializable

### Final definitions

1. All methods on classes / traits declared final, by default
2. No methods on objects declared final
3. No methods on final classes declared final
4. All classes inside objects should be defined final
5. All final classes have private constructors & constructor parameters
6. All vals declared final, either in objects or final classes, if they are constant expressions and without type annotations.
7. Private methods/vals not declared final since they are already final
8. Package-private vals/methods declared final

### Refactor

1. If a class has all final members, class declared final & final member annotations removed
2. use least powerful type alias

### Naming of parameters

1. partial function are called `pf`
2. evidences are called `ev`
3. promise are called `p` (unless in its own class methods, in that case it is called `that`)
4. functions are called `fn`, `fn1`, unless they bear specific meaning: `use`, `release`, ..
4. consider _ method having more meaningful names
5. iterable are called `in`
6. when a parameter type equals own (in a method of a trait) call it `that`
7. please, PLEASE, only use  by name parameters if it makes sense. mind the Function[0] extra allocation and loss of clean syntax
8. folds initial value are called `zero`

### Misc

1. GADTS should have type annotation.
2. Type alias should have type annotation.
[...]

PR #2589, opened 2020-01-06
Add coding guidelines documentation.

saraiva132:
#2549 Draft. Let me know what you think guys. I need a lot more stuff!!! I love this

The plan was never to create strictness on contributors but to help them. For instance, for naming variables, something that I struggle with, if someone is not sure, they could either check the rest of the codebase or check the coding guidelines, which are just that, guidelines.

Additionally, this would serve as a base to get inspiration to produce scalafix rules so that these process could be automated.

The idea, is to improve the codebase, the contribution experience and the ZIO library as a whole.

**The only thing to be automated would be rules based on the scala lang docs (when to use modifiers, extends of certain things, etc) itself removing the need for the contributor to know its details. Improving speed, reducing review iterations and improving code quality.**

### DIFF
--- a/docs/about/coding_guidelines.md
+++ b/docs/about/coding_guidelines.md
@@ -4,7 +4,7 @@ title:  "ZIO Coding Guidelines"
 ---
 
 These are coding guidelines strictly for ZIO contributors for ZIO projects and 
-not general conventions to be applied by the Scala community.
+not general conventions to be applied by the Scala community at large.
 
 Additionally, bear in mind that, although we try to enforce these rules to the 
 best of our ability, both via automated rules (scalafix) and strict reviewing 
@@ -17,15 +17,21 @@ refer to them once in a while when in doubt.
 
 ### Defining classes and traits
 
-1. Value classes must be final and extend `AnyVal`;
+1. Value classes must be final and extend `AnyVal`. 
+This is done to avoid allocating runtime objects; 
 
 2. Method extension classes must be final and extend `AnyVal`;
 
-3. Sealed traits that are ADTs (Algebraic data types) should extend `Product` and `Serializable`;
+3. Avoid overloading standard interfaces. When creating services avoid using the same names as well known standard interfaces.
+Example: Instead of having a service `Random` with methods `nextLong(n)` and `nextInt(n)` consider choosing something like 
+`nextLongBounded(n)` and `nextIntBounded(n)`.
 
-4. Regular traits and sealed trait that do not form ADTs should extend `Serializable` but not `Product`;
+4. Sealed traits that are ADTs (Algebraic data types) should extend `Product` and `Serializable`. 
+This is done to help the compiler infer types;
 
-5. Traits should always extends `Serializable`. (i.e. `ZIO`).
+5. Regular traits and sealed trait that do not form ADTs should extend `Serializable` but not `Product`;
+
+6. Traits should always extend `Serializable`. (i.e. `ZIO`).
 
 ### Final and private modifiers 
 
@@ -35,20 +41,18 @@ refer to them once in a while when in doubt.
 
 3. No methods on final classes declared `final`, because they are `final` by default;
 
-4. All classes inside objects should be defined `final`;
+4. All classes inside objects should be defined `final`, because otherwise they could still be extended;
 
 5. In general, classes that are not case classes have their constructors & constructor parameters private. 
    Typically it is not good practice to expose constructors and constructor parameters but exceptions apply (i.e. `Assertion` and `TestAnnotation`);
 
 6. All `vals` declared `final`, even in objects or `final classes`, if they are constant expressions and without type annotations;
 
-7. Private methods and `vals` are not declared `final` since they are already final;
-
-8. Package-private `vals` and methods should be declared `final`.
+7. Package-private `vals` and methods should be declared `final`.
 
 ### Refactoring
 
-1. If a class has all `final` members, the class should be declared `final` and `final` member annotations should be removed except constant expressions;
+1. If a class has all its members `final`, the class should be declared `final` and `final` member annotations should be removed except constant expressions;
 
 2. All type annotations should use the least powerful type alias. This means, that, let us say, a `ZIO` effect that has 
    no dependencies but throws an arbitrary error, should be defined as `IO`.
@@ -79,7 +83,7 @@ Naming expectations can be helpful in understanding the role of certain paramete
 7. Be mindful of using by-name parameters. Mind the `Function[0]` extra allocation and loss of clean syntax when invoking the method.
    Loss of syntax means that instead of being able to do something like `f.flatMap(ZIO.success)` you require to explicitly do `f.flatMap(ZIO.success(_))`;
    
-8. Fold or fold variants initial value are called `zero`.
+8. Fold or fold variants initial values are called `zero`.
 
 ### Understanding naming of methods
 
@@ -118,7 +122,8 @@ The following rules are good to have in mind when adding new `types`, `traits` o
 
 ### Method alphabetization
 
-In general the following rules should be applied regarding method alphabetization. To fix forward references of values we recommend the programmer to make them lazy (`lazy val`).
+In general the following rules should be applied regarding method alphabetization. 
+To fix forward references of values we recommend the programmer to make them lazy (`lazy val`).
 Operators are any methods that only have non-letter characters (i.e. `<*>` , `<>`, `*>`).
 
 1. Public abstract defs / vals listed first, and alphabetized, with operators appearing before names.
@@ -126,3 +131,19 @@ Operators are any methods that only have non-letter characters (i.e. `<*>` , `<>
 2. Public concrete defs / vals listed second, and alphabetized, with operators appearing before names.
 
 3. Private implementation details listed third, and alphabetized, with operators appearing before names.
+
+### Scala documentation
+
+It is strongly recommended to use scala doc links when referring to other members. 
+This both makes it easier for users to navigate the documentation and enforces that the references are accurate.
+A good example of this are `ZIO` type aliases that are extremely pervasive in the codebase: `UIO`, `Task`, `RIO`, `URIO` and `UIO`.
+To make it easy for developers to see the implementation scala doc links are used, for example:
+
+```
+  /**
+   * @see See [[zio.ZIO.absolve]]
+   */
+  def absolve[R, A](v: RIO[R, Either[Throwable, A]]): RIO[R, A] =
+    ZIO.absolve(v)
+```
+
--- a/docs/about/coding_guidelines.md
+++ b/docs/about/coding_guidelines.md
@@ -18,51 +18,112 @@ refer to them once in a while when in doubt.
 
 ### Defining classes and traits
 
-1. Value classes must be final and extend AnyVal.
-2. Method extension classes must be final and extend `AnyVal`.
-3. Sealed traits that are ADTs should extend `Product` and `Serializable`.
-4. Regular traits and sealed trait that do not form ADTS should extend `Serializable` but not `Product`.
-5. Multiple parameter classes that are ADT should be case class and not extend AnyVal
-6. Traits should always extends `Serializable`. (i.e. `ZIO`)
+1. Value classes must be final and extend AnyVal;
+
+2. Method extension classes must be final and extend `AnyVal`;
+
+3. Sealed traits that are ADTs should extend `Product` and `Serializable`;
+
+4. Regular traits and sealed trait that do not form ADTS should extend `Serializable` but not `Product`;
+
+5. Traits should always extends `Serializable`. (i.e. `ZIO`).
 
 ### Final and private modifiers 
 
-1. All methods on classes / traits are declared `final`, by default.
-2. No methods on objects declared `final`.
-3. No methods on final classes declared `final`.
-4. All classes inside objects should be defined `final`.
-5. All final classes have private constructors & constructor parameters.
-6. All `vals` declared `final`, either in objects or `final classes`, if they are constant expressions and without type annotations.
-7. Private `methods` and `vals` are not declared `final` since they are already final.
-8. Package-private `vals` and `methods` declared `final`.
+1. All methods on classes / traits are declared `final`, by default;
+
+2. No methods on objects declared `final`, because they are `final` by default;
+
+3. No methods on final classes declared `final`, because they are `final` by default;
+
+4. All classes inside objects should be defined `final`;
+
+5. In general, classes that are not case classes have their constructors & constructor parameters private. 
+   Typically it is not good practice to expose constructors and constructor parameters but exceptions apply (i.e. `Assertion` and `TestAnnotation`);
+
+6. All `vals` declared `final`, even in objects or `final classes`, if they are constant expressions and without type annotations;
+
+7. Private methods and `vals` are not declared `final` since they are already final;
+
+8. Package-private `vals` and methods should be declared `final`.
 
 ### Refactoring
 
-1. If a class has all `final` members, the class should be declared `final` and `final` member annotations should be removed except constant expressions.
-2. All type annotations should use the least powerful type alias. This means, that, let us say, a `ZIO` effect that has no dependencies but throws an arbitrary error, should be defined as `IO`.
+1. If a class has all `final` members, the class should be declared `final` and `final` member annotations should be removed except constant expressions;
 
-### Naming of parameters in methods and classes
+2. All type annotations should use the least powerful type alias. This means, that, let us say, a `ZIO` effect that has 
+   no dependencies but throws an arbitrary error, should be defined as `IO`.
 
-1. Partial functions are called `pf`
-2. Evidences are called `ev`
-3. Promise are called `p` (unless in its own class methods, in that case it is called `that`)
-4. Functions are called `fn`, `fn1`, unless they bear specific meaning: `use`, `release`, ..
-4. Consider _ method having more meaningful names
-5. Iterable are called `in`
-6. When a parameter type equals own (in a method of a trait) call it `that`
-7. Never use by-name parameter unless you really know what you are doing and it makes sense. Mind the `Function[0]` extra allocation and loss of clean syntax when invoking the method.
-8. Folds initial value are called `zero`
+### Understanding naming of parameters or values
 
-### Misc
+ZIO often but does not always uses the following naming conventions. This guide can help you understand where the names come from. 
+Naming expectations can be helpful in understanding the role of certain parameters without even glancing at its type when reading code, class or method signatures.
 
-1. GADTS should always have type annotation.
-2. Type alias should always have type annotation.
-3. Methods that lift pure values to effects are dangerous. Dangerous in the sense that they can potentially have dangerous side-effects. 
-Such methods should have a default lazy variant and an eager variant for advanced users that are aware they absolutely do not have side-effects in their code, 
-having slight gains in performance. The lazy variant should have a normal name (succeed, fail, lift, potato) and the eager variant should have a now suffix 
-(succeedNow, liftNow, potatoNow) which makes it clear of its eager behaviour.      
+1. Partial functions have a shortened name `pf`;
 
-### Code structure
+2. In ZIO implicit parameters are often used as compiler evidences;
+   These evidences help you, as a developer, prove something to the compiler (at compile time) and they have the ability to add constraints to a method;
+   They are typically called `ev` if there is only one. Or `ev1`, `ev2`... if more than one;
+   
+3. Promises are called `p` (unless in its own class methods, in that case it is called `that`, like point 6 defines);
 
-1. Method alphabetization - add lazy `vals` to fix forward references.
-2. Sorting of imports                            
+4. Functions are called `fn`, `fn1`, unless they bear specific meaning: `use`, `release`;
+
+4. ZIO effects are called `f`, unless they bear specific meaning like partially providing environment: `r0`;
+
+4. Consider methods ending with _ having more meaningful names;
+
+5. Iterable are called `in`;
+
+6. When a parameter type equals own (in a method of a trait) call it `that`;
+
+7. Be mindful of using by-name parameters. Mind the `Function[0]` extra allocation and loss of clean syntax when invoking the method.
+   Loss of syntax means that instead of being able to do something like `f.flatMap(ZIO.success)` you require to explicitly do `f.flatMap(ZIO.success(_))`;
+   
+8. Fold or fold variants initial value are called `zero`.
+
+### Understanding naming of methods
+
+ZIO goes to great lengths to define method names that are intuitive to the library user. Naming is hard. 
+This section will attempt to provide some guidelines and examples to document, guide and explain naming of methods in ZIO.
+
+1. Methods that lift pure values to effects are dangerous. Dangerous in the sense that they can potentially have dangerous side-effects. 
+   Such methods should have a default lazy variant and an eager variant for advanced users that are aware they absolutely do not have side-effects in their code, 
+   having slight gains in performance. The lazy variant should have a normal name (succeed, fail, die, lift) and the eager variant should have a now suffix 
+   (succeedNow, failNow, dieNow, liftNow) which makes it clear of its eager behaviour;
+
+2. Methods that have the form of `List#zip` are called `zip`, and have an alias called `<*>`. The parallel version, if applicable, has the name `zipPar`, with an alias called `<&>`;
+
+3. Methods that are intended to capture side-effects, convert them into functional effects, should be prefixed by effect*. For example, `ZIO.effect`;
+
+4. The dual of zip, which is trying either a left or right side, producing an Either of the result, should be called `orElseEither`, with alias `<+>`. 
+   The simplified variant where both left and right have the same type should be called `orElse`, with alias `<>`;
+    
+5. Constructors for a data type `X` that are based on another data type `Y` should be placed in the companion object `X` and named `fromY`. 
+   For example, `ZIO.fromOption`, `ZStream.fromEffect`;
+   
+6. Parallel versions of methods should be named the same, but with a `Par` suffix. Parallel versions with a bound on parallelism should use a `ParN` suffix;
+
+7. `Foreach` should be used as the default traverse operation, with `traverse` retained as an alias for programmers with an FP background. For example, `ZIO.foreach`.
+   
+
+### Type annotations
+
+ZIO goes to great lengths to take advantage of the scala compiler in varied ways. Type variance is one of them. 
+The following rules are good to have in mind when adding new `types`, `traits` or `classes` that have either covariant or contravariant types.
+
+1. Generalized ADTs should always have type annotation. (i.e. `final case class Fail[+E](value: E) extends Cause[E]`);
+   
+2. Type alias should always have type annotation. Much like in Generalized ADTs defining type aliases should have carry the type annotations 
+   (i.e. `type IO[+E, +A] = ZIO[Any, E, A]`).
+  
+
+### Method alphabetization
+
+In general the following rules should be applied regarding method alphabetization. To fix forward references we recommend values to be made lazy (`lazy val`).
+
+1. Public abstract defs / vals listed first, and alphabetized, with operators appearing before names.
+
+2. Public concrete defs / vals listed second, and alphabetized, with operators appearing before names.
+
+3. Private implementation details listed third, and alphabetized, with operators appearing before names.
--- a/docs/about/coding_guidelines.md
+++ b/docs/about/coding_guidelines.md
@@ -1,0 +1,65 @@
+---
+id: about_coding_guidelines
+title:  "ZIO Coding Guidelines"
+---
+
+These are coding guidelines strictly for ZIO contributors for ZIO projects and 
+not general conventions to be applied by the Scala community.
+
+Additionally, bear in mind that, although we try to enforce these rules to the 
+best of our ability, both via automated rules (scalafix) and strict reviewing 
+processes, it is both possible to find existing code that does not comply to 
+these rules. If that is the case, we would be extremely grateful if you could 
+make a contribution, by proving a fix to said issue.
+
+Last, but not least, these rules are continuously evolving and as such, 
+refer to them once in a while when in doubt.
+
+
+### Extensions:
+
+1. value classes must be final and extend AnyVal.
+2. method extension classes must be final and extend AnyVal
+3. Sealed traits that are ADTs should extend from Product with Serializable.
+4. Regular traits and sealed trait that do not form ADTS should extend from Serializable but not from Product.
+5. Multiple parameter classes that are ADT should be case class and not extend AnyVal
+6. Traits should always extends serializable
+
+### Final definitions
+
+1. All methods on classes / traits declared final, by default
+2. No methods on objects declared final
+3. No methods on final classes declared final
+4. All classes inside objects should be defined final
+5. All final classes have private constructors & constructor parameters
+6. All vals declared final, either in objects or final classes, if they are constant expressions and without type annotations.
+7. Private methods/vals not declared final since they are already final
+8. Package-private vals/methods declared final
+
+### Refactor
+
+1. If a class has all final members, class declared final & final member annotations removed
+2. use least powerful type alias
+
+### Naming of parameters
+
+1. partial function are called `pf`
+2. evidences are called `ev`
+3. promise are called `p` (unless in its own class methods, in that case it is called `that`)
+4. functions are called `fn`, `fn1`, unless they bear specific meaning: `use`, `release`, ..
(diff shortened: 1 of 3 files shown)


docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+6. All `vals` declared `final`, either in objects or `final classes`, if they are constant expressions and without type annotations.
+7. Private `methods` and `vals` are not declared `final` since they are already final.
+8. Package-private `vals` and `methods` declared `final`.
+
+### Refactoring
+
+1. If a class has all `final` members, the class should be declared `final` and `final` member annotations should be removed except constant expressions.
+2. All type annotations should use the least powerful type alias. This means, that, let us say, a `ZIO` effect that has no dependencies but throws an arbitrary error, should be defined as `IO`.
+
+### Naming of parameters in methods and classes
```
iravid: Sorry, why do we need this section?

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+refer to them once in a while when in doubt.
+
+
+### Defining classes and traits
+
+1. Value classes must be final and extend AnyVal.
+2. Method extension classes must be final and extend `AnyVal`.
+3. Sealed traits that are ADTs should extend `Product` and `Serializable`.
+4. Regular traits and sealed trait that do not form ADTS should extend `Serializable` but not `Product`.
+5. Multiple parameter classes that are ADT should be case class and not extend AnyVal
```
iravid: I'm not sure what this rule means, but you can't extend `AnyVal` on a class with more than one constructor parameter anyway, so that part is redundant.

docs/about/coding_guidelines.md
saraiva132: Perhaps we do not. Perhaps refactoring is just a dumb name. I admit I will go through everything with a lot more details. Put links to where such rules are based on in the first place. 

Maybe it was too soon to open a PR, this is not ready for a review, it is more for people to take a look, if interested.

iravid: Ok, fair enough. I'm in general 👎 on exerting this much control on the code people contribute.

docs/about/coding_guidelines.md
saraiva132: True, the compiler will not allow you to do that.

docs/about/coding_guidelines.md
saraiva132: Most of this I was thinking on enforcing by applying scalafix rules, so, automate it. I just wanted to document it as well. 

I do not want to make things too strict by any means, but create expectations for users reading the code.

docs/about/contributing.md
```diff
@@ -17,7 +17,7 @@ To begin contributing, please follow these steps:
 
 ### Get The Project
 
-If you don't already have one, sign up for a free [GitHub Account](https://github.com/join?source=header-home).
+If you don not  already have one, sign up for a free [GitHub Account](https://github.com/join?source=header-home).
```
ghostdogpr:
```suggestion
If you do not  already have one, sign up for a free [GitHub Account](https://github.com/join?source=header-home).
```

docs/about/coding_guidelines.md
iravid: This would have made sense if it were a commercial team deciding unilaterally that this is how code written in the team should be written. But this is not a commercial team, contributors have different backgrounds and skill levels.

If a reviewer feels a name better suits a method parameter, they can comment on the PR - no need to preemptively specify how the code should be written.

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
+---
```
ghostdogpr: You also need to add this file to https://github.com/zio/zio/blob/master/website/sidebars.json

docs/about/coding_guidelines.md
saraiva132: What you say makes sense. I will revise my approach on this. My intentions were perhaps too focused on the reader and not the writer.

docs/about/contributing.md
saraiva132: I noticed the typo ! :(

docs/about/coding_guidelines.md
adamgfraser: Yes, I would suggest we drop the section on naming. I think we have enough things that are more objective and have a clearer technical rationale that we should focus the time of contributors and reviewers on those.

saraiva132: Because I feel I am passing the message I want people to strictly follow these rules and, that was not my intention. I do not want to put any burden on contributors. My intent is the opposite, to make their life easier.

docs/about/coding_guidelines.md
adamgfraser: Yes, I think if something is just impossible to do we don't need it here. This is for things that people could do that we don't want them to do.

docs/about/coding_guidelines.md
adamgfraser: I think that is a great attitude. One thing we might want to consider along those lines is to have a brief description of the rationale behind each guideline after it (e.g. "alphabetizing makes it easy for users / contributors to find methods in a large file and avoids wasting time thinking about how to organize things"). That could be a good discipline for us to make sure we have a good justification for these and if people understand it they are more likely to feel this is something helping them versus something they have to do.

softinio:
Applying scalafix in the way it is being proposed in my opinion will lead to higher Circleci costs and slower builds to me it feels like we are trying to fix something that is not broken. This is an open source project.

Also we are adding complexity to a developers flow where as in my opinion we need to make it simpler to contribute more. 

Striving for perfectionism has a lot drawbacks that needs to be considered to find the right balance so hope no one minds my two cents on the subject.

saraiva132:
I will refrain from proceeding with this while I think about it a bit more.

I think there is enormous merit to having some rules automated with Scalafix. We are not adding complexity from developers but removing it by allowing them to focus on the problem at hand without needing to worry about the small details of Scala language. 

Much like when you want to focus on your code and have it formatted for you. 

I am going to focus on writing my first Scalafix times soon. Just came back to work so time is not great. 

All feedback is deeply appreciated since this started as a personal opinion. 

The naming conventions were never supposed to be automated. Perhaps I should have made clear what was meant to be automated.

saraiva132:
@iravid @softinio @adamgfraser @regiskuckaertz 

Please do see from my the updates on my descriptions if there is merit to this. If not, I can close it. If there is, I can revise and update it. Appreciate your feedback as always.

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+2. Evidences are called `ev`
+3. Promise are called `p` (unless in its own class methods, in that case it is called `that`)
+4. Functions are called `fn`, `fn1`, unless they bear specific meaning: `use`, `release`, ..
+4. Consider _ method having more meaningful names
+5. Iterable are called `in`
+6. When a parameter type equals own (in a method of a trait) call it `that`
+7. Never use by-name parameter unless you really know what you are doing and it makes sense. Mind the `Function[0]` extra allocation and loss of clean syntax when invoking the method.
+8. Folds initial value are called `zero`
+
+### Misc
```
regiskuckaertz: I think this is too ad hoc and is best left out. We will review PRs and apply judgment.

adamgfraser:
@softinio You raise a lot of good points.

I think there are two somewhat separable issues here:
1. What should be in this coding guidelines document, if we have it at all?
2. What should we use Scalafix for, if we use it at all?

**Coding Guidelines**

On the first one, I think there is value in having a document like this in some form both to codify our own understanding of the best way to do certain things and to communicate this to contributors. For example, there was a bunch of conversation before about when adding `final` to a method adds anything and I think we want to capture those conclusions somewhere instead of just having them sitting around in the comments to an old PR. Similarly, I have observed that reviewers including myself almost always ask contributors to alphabetize methods so I think there is some value in saying that somewhere so contributors know that.

At the same time, I think it  has to be written in a friendly and approachable way, focus on things where there is a clear technical "right answer", and recognize that there is a point of diminishing returns in the ability of contributors and reviewers to follow.

Concretely, I would suggest deleting the sections "Naming of parameters and classes" and "Misc" and trying to do a revision to make all of this a little friendlier (e.g. brief explanations of justifications, more "do X" versus "don't do Y", less "must", etc...), and then we can review another version. @saraiva132 You can do that if you want or I am happy to take a turn on this.

**Use of Scalafix**

So far the only thing we are using Scalafix for is sorting imports and at least my perception is that has gone relatively well. I think that was an issue that was identified by @NeQuissimus, it makes it easier to see where imports are coming from, and the feedback I have seen so far from contributors is it is nice to have a tool that automatically does that for you so you don't have to worry about that yourself. We have a command `prepare` that automatically runs that and `fmt` (maybe we should rename to `lint`?) so if users do that before pushing I don't think there should be that much impact on CI time. But we can also definitely revisit if you don't think that makes sense.

As far as further rules I think we don't really know. If there are really rules that make things easier and make us have to worry about fewer things there could be value there. For example I would love to never have to worry again about whether something should extend `Serializable`. I don't know if we can actually implement a rule for that but I think that would be interesting. If it is just changing how a parameter is named I 100% agree there is not value there.

Does that make sense? This is definitely something we should have consensus on.

NeQuissimus:
My 2 cents:

We have found at $WORK that guidelines are just ignored in the long run (people love them initially). Everything needs to be automated and checked during CI. If cannot be checked automatically, the feedback loop to a PR is too long and just becomes annoying and unproductive (nobody wants to review a PR with 35 "this parameters should be named differently" comments).

From my POV, whatever cannot be done with Scalafix/Scalafmt is just churn for contributors.

FWIW, here is the Scalafmt config we have been happy with at $WORK

```
align.ifWhileOpenParen = false
align.openParenCallSite = false
align.openParenDefnSite = false
align.tokens.add = ["*>", "<-"]
continuationIndent.defnSite = 2
continuationIndent.extendSite = 2
danglingParentheses = true
docstrings = ScalaDoc
importSelectors = singleLine
includeCurlyBraceInSelectChains = false
maxColumn = 120
newlines.alwaysBeforeElseAfterCurlyIf = false
newlines.alwaysBeforeTopLevelStatements = true
optIn.selfAnnotationNewline = false
optIn.configStyleArguments = false
rewrite.rules = ["PreferCurlyFors", "RedundantBraces", "RedundantParens", "SortModifiers"]
spaces.afterTripleEquals = true
spaces.inImportCurlyBraces = true
version=2.0.1
```

Our Scalafix setup is essentially [Scaluzzi](https://github.com/vovapolu/scaluzzi) + [Scalazzi](https://github.com/scalaz/scalazzi), which would be way too tight for ZIO proper.

adamgfraser:
@NeQuissimus Are you using any Scalafix rules? Any feedback on what has worked and what hasn't?

NeQuissimus:
Here is our config:

```
λ cat .scalafix.conf -p
# Copyright: 2018 Sam Halliday
# License: https://www.apache.org/licenses/LICENSE-2.0

# This file enforces the scalazzi safe subset of scala.
#
# For further information, see https://github.com/scalaz/scalazzi

rules = [
  Disable
  DisableSyntax
  ExplicitResultTypes
  LeakingImplicitClassVal
  MissingFinal
  RemoveUnused
  SortImports
]

SortImports.blocks = [
  "java",
  "scala",
  "cats",
  "com.comcast.ip4s",
  "com.github.ghik",
  "com.maxmind.geoip2",
  "com.microsoft.azure",
  "com.typesafe.config",
  "doobie",
  "eu.timepit.refined",
  "fastparse",
  "github.gphat.censorinus",
  "io.circe",
  "machinist",
  "monocle",
  "nequi",
  "net.schmizz.sshj",
  "okhttp3",
  "org.apache.commons",
  "org.apache.kafka",
  "org.asynchttpclient",
  "org.deeplearning4j",
  "org.http4s",
  "org.nd4j",
  "org.parboiled2",
  "org.slf4j",
  "org.tensorflow",
  "pureconfig",
  "redis",
  "scalaprops",
  "scalaz",
  "shapeless",
  "sttp",
  "software.amazon.awssdk",
  "sourcecode",
  "utest",
  "zio",
  "*",
  "COMPANY"
]

Disable.symbols = [
  {
    regex = "^\\Qscala/collection/mutable\\E.*$"
    message = "Java collections have better performance, which is what I assume you were trying to do"
  }
  {
    regex = "^\\Qscala/collection/parallel\\E.*$"
    message = "parallel collections are low performance and block the thread, prefer scalaz.ApplyN"
  }
  {
    regex = "^\\Qscala/math/Big\\E.*$"
    message = "scala arbitrary precision numbers are broken: https://github.com/scala/bug/issues/9670"
  }
  {
    regex = [
      "^\\Qjava/io\\E.*$"
      "^\\Qscala/io\\E.*$"
    ]
    message = "legacy blocking API, prefer java.nio"
  }
  {
    regex = "^\\Qjava/net/URL#\\E.*$"
    message = "URL talks to the network for equality, prefer URI"
  }
  {
    regex = {
      includes = [
        # overrides not fully implemented: https://github.com/scalacenter/scalafix/pull/634
        "^.*\\Q#asInstanceOf().\\E$"
        "^.*\\Q#equals().\\E$"
        "^.*\\Q#hashCode().\\E$"
        "^.*\\Q#toString().\\E$" # doesn't catch string interpolators...
        # more to add here...
      ]
    }
    message = "prefer scalaz.{Equal, Show, Liskov, etc}"
  }

  {
    symbol = "scala/Enumeration"
    message = "prefer a sealed abstract class"
  }

  {
    regex = {
      includes = [
        "^\\Qscala/util/Either.LeftProjection#get().\\E$"
        "^\\Qscala/util/Either.RightProjection#get().\\E$"
        "^\\Qscala/util/Try#get().\\E$"
        "^\\Qscala/Option#get().\\E$"
        "^\\Qscala/collection/IterableLike#head().\\E$"
      ]
    }
    message = "not a total function"
  }
]

Disable.ifSynthetic = [
  "java/io/Serializable"
  # "scala/Any" # Way too much trouble with ZIO environments
  "scala/Product"

  # local type inference + covariant types fires this
  # "scala/Nothing"

  # when upstream broke noImplicitConversion and we don't agree that their
  # implicits are worth the mental burden.
  "scala/Option.option2Iterable"
  "scala/Predef.any2stringadd"

  # I don't understand why these are synthetic
  # https://github.com/scalacenter/scalafix/issues/703
  {
    regex = {
      includes = [
        "^\\Qscala/collection/MapLike#apply().\\E$"
        "^\\Qscala/collection/LinearSeqOptimized#apply().\\E$"
      ]
    }
    message = "not a total function"
  }
]

Disable.unlessInside = [
  {
    safeBlocks = [
      "zio.FunctionIO.effect",
      "zio.FunctionIO.effectTotal",
      "zio.Task",
      "zio.Task.effect",
      "zio.Task.effectAsync",
      "zio.Task.effectAsyncInterrupt",
      "zio.Task.effectAsyncMaybe",
      "zio.Task.effectTotal",
      "zio.IO",
      "zio.IO.effect",
      "zio.IO.effectAsync",
      "zio.IO.effectAsyncInterrupt",
      "zio.IO.effectAsyncMaybe",
      "zio.IO.effectTotal",
      "zio.RIO",
      "zio.RIO.effect",
      "zio.RIO.effectAsync",
      "zio.RIO.effectAsyncInterrupt",
      "zio.RIO.effectAsyncMaybe",
      "zio.RIO.effectTotal",
      "zio.UIO",
      "zio.UIO.effectAsync",
      "zio.UIO.effectAsyncInterrupt",
      "zio.UIO.effectAsyncMaybe",
      "zio.UIO.effectTotal",
      "zio.ZIO",
      "zio.ZIOFunctions.effect",
      "zio.ZIOFunctions.effectAsync",
      "zio.ZIOFunctions.effectAsyncInterrupt",
      "zio.ZIOFunctions.effectAsyncMaybe",
      "zio.ZIOFunctions.effectTotal",
      "zio.stream.Stream.effectAsync",
      "zio.stream.Stream.effectAsyncInterrupt",
      "zio.stream.Stream.effectAsyncMaybe",
      "zio.stream.ZStream.effectAsync",
      "zio.stream.ZStream.effectAsyncInterrupt",
      "zio.stream.ZStream.effectAsyncMaybe",
      "zio.blocking.Blocking.Service.effectBlocking",
      "zio.blocking.Blocking.Service.effectBlockingCancelable"
    ]
    symbols = [
      {
        # This takes the approach of banning everything and then blessing things
        # we trust. It would be entirely feasible to do it the other way and
        # list the explicit list of things that we need to go through IO, but it
        # is less safe. We can have another rule if something in an "excludes"
        # glob catches unsafe things.
        #
        # If something is referentially transparent but not total, instead of
        # adding to this excludes list, add it to the includes list of the
        # scalaz.Maybe.attempt safeBlock. If it can return `null`, add it to a
        # scalaz.Maybe.fromNullable safeBlock.
        regex = {
          includes = [
            "^\\Qjava/\\E.*$"
            #"^.*$"
          ]
          excludes = [
            "^fommil.*$"
            "^scalaz.*$"
            "^\\Qjava/lang/IllegalStateException#\\E.*$"
            "^\\Qjava/lang/Exception#\\E.*$"
            "^\\Qjava/lang/String#\\E.*$"
            "^\\Qjava/lang/Void#\\E.*$"
            "^\\Qjava/time/Duration#\\E.*$"
            "^\\Qjava/time/Instant#\\E.*$"
            "^\\Qjava/time/ZonedDateTime#\\E.*$"
            "^\\Qjava/time/temporal/ChronoUnit#\\E.*$"
            "^\\Qjava/net/URI#\\E.*$"
            "^\\Qjava/net/URLEncoder#\\E.*$"
            "^\\Qjava/net/URLDecoder#\\E.*$"
            "^\\Qjava/time/Instant#\\E.*$"
            "^\\Qjava/time/ZonedDateTime#\\E.*$"
            "^\\Qjava/time/temporal/ChronoUnit#\\E.*$"
            "^\\Qjava/util/NoSuchElementException#\\E.*$"
            "^\\Qjava/util/concurrent/TimeUnit#\\E.*$"
            "^\\Qjava/util/concurrent/TimeoutException#\\E.*$"
            "^\\Qjava/util/UUID#\\E.*$"
            "^\\Qjava/nio/file/Path#\\E.*$"
            "^\\Qjava/nio/file/Paths#\\E.*$"
            # primitives are not supported: https://github.com/scalameta/scalameta/issues/1488

            # WORKAROUND https://github.com/scalacenter/scalafix/issues/715
            "^\\Qjava/\\E$"
            "^\\Qjava/lang/\\E$"
            "^\\Qjava/time/\\E$"
            "^\\Qjava/net\\E$"
          ]
        }
        message = "Untrusted third party library must be called from IO, or blessed in scalafix.conf"
      }
    ]
  }
  {
    safeBlocks = [
      "scalaz/`\/`.fromTryCatchNonFatal",
      "scalaz/`\/`.fromTryCatchThrowable",
      "scalaz/Maybe.attempt",
      "scalaz/Maybe.fromTryCatchNonFatal",
      "scalaz/Maybe.fromTryCatchThrowable"
    ]
    symbols = [
      {
        regex = {
          includes = [
            # should live in a common file so we don't need to duplicate
            "^\\Qjava/net/URLEncoder#\\E.*$"
            "^\\Qjava/net/URLDecoder#\\E.*$"
          ]
          excludes = [
          ]
        }
        message = "Deterministic method is not total, must be called via Maybe.attempt, \/.fromTryCatchNonFatal, etc."
      }
    ]
  }
]

DisableSyntax.keywords = [
  var
  null
  return
  throw
  while
]
DisableSyntax.noSemicolons = true
DisableSyntax.noXml = true
DisableSyntax.noCovariantTypes = true
DisableSyntax.noContravariantTypes = true
DisableSyntax.noValInAbstract = true
DisableSyntax.noImplicitObject = true
DisableSyntax.noImplicitConversion = true

ExplicitResultTypes {
  unsafeShortenNames = true

  fatalWarnings = true

  # these apply to non-implicits
  memberKind = [Def, Val]
  memberVisibility = [Public, Protected]

  # turn to the max...
  skipSimpleDefinitions = false
  skipLocalImplicits = false
}
```

NeQuissimus:
We are ZIO and ZIO Test users, so rules like disabling `Any` had to go :D

saraiva132:
I can't help but say that is an impressive configuration.

NeQuissimus:
It's mostly a Sam Halliday effort, I just adjusted it for `SortImports` and ZIO.

docs/about/coding_guidelines.md
jdegoes: Right, I'd delete 5 for this reason. 👍 

The other ones look great!

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+2. Method extension classes must be final and extend `AnyVal`.
+3. Sealed traits that are ADTs should extend `Product` and `Serializable`.
+4. Regular traits and sealed trait that do not form ADTS should extend `Serializable` but not `Product`.
+5. Multiple parameter classes that are ADT should be case class and not extend AnyVal
+6. Traits should always extends `Serializable`. (i.e. `ZIO`)
+
+### Final and private modifiers 
+
+1. All methods on classes / traits are declared `final`, by default.
+2. No methods on objects declared `final`.
```
jdegoes:
```suggestion
2. No methods on objects declared `final`, because they are `final` by default.
```

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+3. No methods on final classes declared `final`.
```
jdegoes:
```suggestion
3. No methods on final classes declared `final`, because they are already `final`.
```

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+4. All classes inside objects should be defined `final`.
+5. All final classes have private constructors & constructor parameters.
```
jdegoes: Is this really true? We have a lot of case classes that don't follow this pattern.

Maybe: `All final classes that are not case classes have private constructors & constructor parameters`. That I think would be mostly true.

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+6. All `vals` declared `final`, either in objects or `final classes`, if they are constant expressions and without type annotations.
```
jdegoes:
```suggestion
6. All `vals` declared `final`, even in objects or `final classes`, if they are constant expressions and without type annotations.
```

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+7. Private `methods` and `vals` are not declared `final` since they are already final.
```
jdegoes:
```suggestion
7. Private methods and `vals` are not declared `final` since they are already final.
```

I am not sure if we want to mention `private[this]`. I think it is being deleted from Dotty.

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+8. Package-private `vals` and `methods` declared `final`.
```
jdegoes:
```suggestion
8. Package-private `vals` and methods declared `final`.
```

docs/about/coding_guidelines.md
jdegoes: I do think justifying everything is a good idea. That way we're teaching to think, not just teaching to follow.

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Refactoring
+
+1. If a class has all `final` members, the class should be declared `final` and `final` member annotations should be removed except constant expressions.
+2. All type annotations should use the least powerful type alias. This means, that, let us say, a `ZIO` effect that has no dependencies but throws an arbitrary error, should be defined as `IO`.
+
+### Naming of parameters in methods and classes
+
+1. Partial functions are called `pf`
```
jdegoes: I think you could preserve this section if you changed it to:

```
### Understanding Naming

ZIO often but does not always uses the following naming conventions. This guide can help you understand where the names come from.
```

Then just explain why, e.g. implicits are sometimes called `ev`

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+4. Consider _ method having more meaningful names
+5. Iterable are called `in`
+6. When a parameter type equals own (in a method of a trait) call it `that`
+7. Never use by-name parameter unless you really know what you are doing and it makes sense. Mind the `Function[0]` extra allocation and loss of clean syntax when invoking the method.
+8. Folds initial value are called `zero`
+
+### Misc
+
+1. GADTS should always have type annotation.
+2. Type alias should always have type annotation.
```
jdegoes: What does this mean?

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+3. Methods that lift pure values to effects are dangerous. Dangerous in the sense that they can potentially have dangerous side-effects. 
+Such methods should have a default lazy variant and an eager variant for advanced users that are aware they absolutely do not have side-effects in their code, 
+having slight gains in performance. The lazy variant should have a normal name (succeed, fail, lift, potato) and the eager variant should have a now suffix 
+(succeedNow, liftNow, potatoNow) which makes it clear of its eager behaviour.
```
jdegoes: I like this bit on `succeedNow`, etc., but I think we need a whole section on naming.

It could have things like this in it:

1. Methods that have the form of `List#zip` are called `zip`, and have an alias called `<*>`. The parallel version, if applicable, has the name `zipPar`, with an alias called `<&>`.
2. Methods that are intended to capture side-effects, convert them into functional effects, should be prefixed by `effect*`. For example, `ZIO.effect`.
3. The dual of zip, which is trying _either_ a left or right side, producing an `Either` of the result, should be called `orElseEither`, with alias `<+>`. The simplified variant where both left and right have the same type should be called `orElse`, with alias `<>`.
4. Constructors for a data type `X` that are based on another data type `Y` should be placed in the companion object `X` and named `fromY`. For example, `ZIO.fromOption`, `ZStream.fromEffect`.
5. Parallel versions of methods should be named the same, but with a `Par` suffix. Parallel versions with a bound on parallelism should use a `ParN` suffix.
6. `foreach` should be used as the default traverse operation, with `traverse` retained as an alias for programmers with an FP background. For example, `ZIO.foreach`.


Etc. I can probably think of at least 10 more. But we don't have to get them all into the first PR. 😄

docs/about/coding_guidelines.md
```diff
@@ -0,0 +1,68 @@
...
+1. GADTS should always have type annotation.
+2. Type alias should always have type annotation.
+3. Methods that lift pure values to effects are dangerous. Dangerous in the sense that they can potentially have dangerous side-effects. 
+Such methods should have a default lazy variant and an eager variant for advanced users that are aware they absolutely do not have side-effects in their code, 
+having slight gains in performance. The lazy variant should have a normal name (succeed, fail, lift, potato) and the eager variant should have a now suffix 
+(succeedNow, liftNow, potatoNow) which makes it clear of its eager behaviour.      
+
+### Code structure
+
+1. Method alphabetization - add lazy `vals` to fix forward references.
```
jdegoes: Would be good to justify these and spell them out in a bit more detail.

For example, generally we have:

1. Public abstract defs / vals listed first, and alphabetized, with operators appearing before names.
2. Public concrete defs / vals listed second, and alphabetized, with operators appearing before names.
3. Private implementation details listed third, and alphabetized, with operators appearing before names.

Would be nice to enforce using Scalafix ultimately.

docs/about/coding_guidelines.md
adamgfraser: I think even then there are a lot of exceptions. For example, `Assertion` and `TestAnnotation` both have constructor parameters that are not private.

docs/about/coding_guidelines.md
adamgfraser: I think it means they should have variance annotation. Like `type IO[+E, +A] = ZIO[Any, E, A]` rather than `type IO[E, A] = ZIO[Any, E, A]`.

docs/about/coding_guidelines.md
saraiva132: @jdegoes Is there any rule of thumb regarding when to make something a val or a def? Something that could be somehow formalized in a sentence. 

It comes to mind when a colleague approached me to ask about clock live instance:

`val nanoTime: UIO[Long] = IO.effectTotal(System.nanoTime)`

```
def currentDateTime: ZIO[Any, Nothing, OffsetDateTime] = 
     for {
          millis <- currentTime(TimeUnit.MILLISECONDS)
          zone   <- ZIO.effectTotal(ZoneId.systemDefault)
        } yield OffsetDateTime.ofInstant(Instant.ofEpochMilli(millis), zone)
```

The explanation I could come up with was that nanoTime was just a pure value (effect) whereas currentDateTime was composing several. But does that really matter? it is still a value and it takes no parameters. 

Additionally, I considered whether it mattered how flexible `nanoTime` and `currentDateTime` should be when thinking of overriding them.

docs/about/coding_guidelines.md
saraiva132: This looks great to add!

docs/about/coding_guidelines.md
saraiva132: I clarified this

docs/about/coding_guidelines.md
saraiva132: Ultimately I am hoping to pick up kind of an epic where I focus on scalafix rules (dream of zio-lint) but we will see how that works

saraiva132:
@jdegoes I made a review of the entire document. I think it still needs a bit more structure. 

There are two things that I see being mentioned at some point:
- Scalafix rules being applied as we add them.
- A documented and structured way of our label policy (maybe we already have this), make it trivial for people to pick up first issues or even perhaps some sort of mentor approach. For the contributing document.

docs/about/coding_guidelines.md
adamgfraser: @saraiva132 A general guideline would be that if something has no parameters and is purely functional it should be a `val` instead of a `def`. In your example above I don't think there is a good reason for that to be a `def` instead of a `val`. Reasons to not do that could include avoiding early initialization issues, preventing forward reference errors, or if although the value is pure it is computationally expensive and may not be needed.

saraiva132: @adamfraser agreed. I was thinking similarly. In general, to avoid forward references a lazy val would be preferred, I believe. Regarding the computationally expensive since these are only descriptions of programs and nothing is executed I think using 
values might be a good practice also as a mental model ( people using effects as values) . I start to see some things that could potentially be made into Scala fix rule,

adamgfraser:
We should add something about using Scaladoc links when referring to other members. This both makes it easier for users to navigate the documentation and enforces that the references are accurate.

saraiva132:
@adamfraser documentation section makes sense

jdegoes:
@saraiva132 How close is this to being ready for merging?

jdegoes:
@saraiva132 Bumping for status. What else remains to be done here? Thanks for all your work codifying many of the informal guidelines we've been following.

saraiva132:
@jdegoes Sorry for the delay!! Let me add a documentation section today and I think for a first version it is good!

regiskuckaertz:
A colleague just sent me this, which is kind of related: https://nrinaudo.github.io/scala-best-practices/adts/final_case_objects.html

🤦‍♂

ioleo:
@saraiva132 I havent been following PRs lately and just noticed this one. Great effort! One thing I'd love to see is each recommendation followed by a short answer (or reference) to `why?`. Some recommendations may feel obvious, but not all, and probably not to everyone. From my own experience with such rule sets I tend to forget the exact wording (or even what the exact rule was) quite quickly, but if I understand the underlaying principle (the "why") I can always recall from there what the solution was.

adamgfraser:
@saraiva132 Following up on #3305 can you add a point that overloading method should be avoided, and in particular methods on services should NOT be overloaded, since overloaded methods cannot be mocked with a natural syntax?

If you have some time it would be great to do a turn on this and get it merged.

saraiva132:
@jdegoes @adamgfraser I have finally made changes and addressed comments. I am really sorry for the delay. I had to take a time out from OSS due to personal and health issues. I am slowly starting to come back to it. 🙇 🙇

adamgfraser:
@saraiva132 No worries at all! Thanks for adding all this great content. Can you sign the CLA and then we can get this merged?

saraiva132:
I have changed user and signed CLA but it appears it does not change status. Looking at details I also get: `You have signed the CLA for zio/zio`

adamgfraser:
@saraiva132 Can you look at the instructions [here](https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user)?

saraiva132:
@adamgfraser I did. If you check my last commit is identified

jdegoes:
@saraiva132 Thank you! Much needed and appreciated addition. ❤️

Issue #1499, opened 2019-08-26
ZIO.foreachParN_ is causing JVM StackOverflow


abomm:
I am getting a java.lang.StackOverflowError when using ZIO.foreachParN_ or ZIO.foreachPar_.

Please run this test code, on ZIO 1.0.0-RC11-1
If you run the main with args of Array("10000", "1","1") which means build a map of 10K keys and random elements, it generates the error.

https://gist.github.com/abomm/6e3420e93958119fe312e8e5b7dbdee9

Error:
```bash
java.lang.StackOverflowError
	at scala.collection.mutable.HashTable$$anon$1.next(HashTable.scala:212)
	at scala.collection.mutable.HashMap$$anon$3.next(HashMap.scala:165)
	at zio.ZIOFunctions.loop$4(ZIO.scala:1841)
	at zio.ZIOFunctions.loop$4(ZIO.scala:1841)
	at zio.ZIOFunctions.loop$4(ZIO.scala:1841)
```

The offending ZIO code looks something like this:
```scala
        _ ← ZIO.foreachPar_(m.keys) { k ⇒
             UIO {
               m.update(k, m(k).sorted)
               m
             }
           }

```
When this map has >10,000 keys we get blow the stack.

jdegoes:
We will fix this for 1.0!

LGLO:
I'd like to take this!
I'll try to come with an idea tomorrow in the morning and discuss it with someone bright.
If doesn't go well I'll release by 3pm tomorrow.

LGLO:
OFC if someone else wants this but was shy to announce, ping me and issue is yours :D

dzajkowski:
just a couple of notes:
- it's not trivial to replicate, I needed to change .jvmopts: `-Xss1M`
- looks like suspending the loop fixes the failing case:
```scala
  final def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, _]): ZIO[R, E, Unit] =
    ZIO
      .succeed(as.iterator)
      .flatMap { i =>
        def loop(a: A): ZIO[R, E, Unit] =
          if (i.hasNext) f(a).zipWithPar(ZIO.effectSuspendTotal(loop(i.next)))((_, _) => ())
          else f(a).unit
        if (i.hasNext) loop(i.next)
        else ZIO.unit
      }
      .refailWithTrace
```

LGLO:
@hanny24 is taking care of it as a part of #1182

[...]

PR #2590, opened 2020-01-08
#1499 Make foreachPar_ stack-safe

evis:
After this change we can pass large collections to `foreachPar_` methods
without crashing the whole application. Though, traversing through large
collections is very slow because of the issue #1182. For example, this
effect runs for more than a minute:

```scala
UIO.foreachParN_(n = 10)(0 to 10000)(ZIO.succeed)
```

So, now it seems impossible to write regression test which doesn't
exceed 1 minute timeout.

`ZIO.foreachParN_` also become stack-safe, as it calls `foreachPar_`.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2232,14 +2232,8 @@ object ZIO {
    * For a sequential version of this method, see `foreach_`.
    */
   def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] =
-    ZIO
-      .effectTotal(as.iterator)
-      .flatMap { i =>
-        def loop(a: A): ZIO[R, E, Unit] =
-          if (i.hasNext) f(a).zipWithPar(loop(i.next))((_, _) => ())
-          else f(a).unit
-        if (i.hasNext) loop(i.next)
-        else ZIO.unit
+    as.foldLeft(unit: ZIO[R, E, Unit]) { (acc, a) =>
+        acc.zipParLeft(f(a))
       }
       .refailWithTrace
 
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1182,11 +1182,8 @@ object ZManaged {
    * For a sequential version of this method, see `foreach_`.
    */
   def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZManaged[R, E, Any]): ZManaged[R, E, Unit] =
-    ZManaged.succeed(as.iterator).flatMap { i =>
-      def loop: ZManaged[R, E, Unit] =
-        if (i.hasNext) f(i.next).zipWithPar(loop)((_, _) => ())
-        else ZManaged.unit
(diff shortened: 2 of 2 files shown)


evis:
@adamgfraser could you review please?

adamgfraser:
@evis If we just want to be able to write a test that doesn't time out we could extend `DefaultRunnableSpec` instead of `ZIOBaseSpec` and take however long we want with the tests. However, that is probably not a great use of CI resources and is also symptomatic of the larger performance problem you identified.

What do you think about expanding the scope of this to try to optimize `foreachPar_`? There is other optimization work ongoing for parallel collection operations but we should be able to have a somewhat faster operation for this one because we don't need to maintain any intermediate data structure with the results. That would kill birds two birds with one stone.

Otherwise we can merge this now and we just need to remember to come back to this to reimplement it in terms of other combinators once we have the other parallel collection methods optimized.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2232,14 +2232,8 @@ object ZIO {
...
   def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] =
-    ZIO
-      .effectTotal(as.iterator)
-      .flatMap { i =>
-        def loop(a: A): ZIO[R, E, Unit] =
-          if (i.hasNext) f(a).zipWithPar(loop(i.next))((_, _) => ())
-          else f(a).unit
-        if (i.hasNext) loop(i.next)
-        else ZIO.unit
+    as.foldLeft(unit: ZIO[R, E, Unit]) { (acc, a) =>
```
regiskuckaertz: It looks like you could avoid the `foldLeft` just by making `loop` tail recursive:
```scala
@tailrec def loop(a: A, acc: ZIO[R, E, Any]): ZIO[R, E, Any] =
  if (i.hasNext)
    loop(i.next, acc <&> f(a))
  else
    acc

if (i.hasNext) loop(i.next, ZIO.unit).unit else ZIO.unit
```

regiskuckaertz: woops, small mistake, I corrected.

evis: In my opinion, version with `foldLeft` is a lot simpler. Also, consiming one more stack frame doesn't seem like a big deal.

Why should we avoid `foldLeft` here?

evis:
> What do you think about expanding the scope of this to try to optimize `foreachPar_`?

I'd like to merge this PR as is, because it's really scary to have a chance of StackOverflowError in production. Then I will try to develop an optimization in separate PR.

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: I may be wrong but `foldLeft` will box primitive types no matter what, whereas using a while loop won't.

evis: Usually, they will be already boxed anyway inside `as: Iterable[A]`. So, `foldLeft` shouldn't do any boxing again in that case, at the first glance.

regiskuckaertz:
Fair enough. Let’s get rid of the stack safety issue 👍

PR #2593, opened 2020-01-09
refactor: simplify unit-related code

evis:
1. Replace `succeed(())` with `unit`
2. Change `ZIO.unit` type ascription to `UIO`


adamgfraser:
Great cleanup!

PR #2594, opened 2020-01-10
WIP: A simpler and value-composable encoding for environmental effects

jdegoes:


core/shared/src/main/scala/zio/Has.scala
```diff
@@ -0,0 +1,231 @@
...
+ * as shorthands for common services, e.g.:
+ *
+ * {{{
+ * type Console = Has[ConsoleService]
+ * }}}
+ *
+ * Currently, to be portable across all platforms, all services added to an
+ * environment must be monomorphic. Parameterized services are not supported.
+ */
+final class Has[A] private (private val map: Map[TagType, scala.Any], var cache: Map[TagType, scala.Any] = Map())
```
adamgfraser: `private var cache`?

jdegoes: 😅 👍

build.sbt
```diff
@@ -93,6 +93,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
 lazy val coreJVM = core.jvm
   .settings(dottySettings)
   .settings(replSettings)
+  .settings(
+    libraryDependencies +=
+      "org.scala-lang" % "scala-reflect" % scalaVersion.value
```
vasilmkd: I see this is only used on the jvm but I think we need to be careful if we ever consider GraalVM native-image to be a viable "platform" for ZIO as it supports _very limited_ reflection features.

jdegoes: The only thing we need is `TypeTag` (or `ClassTag` as a poor man's substitute) which is a compile-time feature.

core/jvm/src/main/scala-2.x/zio/ScalaSpecific.scala
```diff
@@ -16,13 +16,20 @@
...
 
-import zio.internal.{ Scheduler => IScheduler }
+private[zio] object ScalaSpecific {
+  import scala.reflect.runtime.universe._
 
-// This cannot extend Scheduler.Service[Scheduler] because of Scala 2.11 support
-package object scheduler {
-  val schedulerService: ZIO[Scheduler, Nothing, Scheduler.Service[Any]] =
-    ZIO.access(_.scheduler)
+  type Tagged[A] = TypeTag[A]
```
neko-kai: @jdegoes @adamgfraser 
Please do not use TypeTags, they don't work!

1. `scala-reflect` is not thread-safe at all, see - https://github.com/scala/bug/issues/10766, under any kind of concurrency, it will break, in our case even surrounding every use with a lock didn't remove all random failures. You may experience any kind of random data corruption and false positives anywhere, e.g. `Set[Int]` may be printed as `Set[String]` on a bad day (that happened)
2. `equals` is pointer equality, if it works here – it's only because of scala-reflect's caching. But there are many ways to obtain equal types with different identities that will fail `equals`. You must use `=:=` – but `=:=` may be noticeably slow.
3. TypeTags have a large impact on compile-times. Each type tag splices about 25 lines of mutable initialization logic – it's not even a macro call, it's intrinsic, but it manages to be massively slower than LightTypeTag due to just the size of the tree. (LightTypeTag splices just two string literals)
4. `scala-reflect` takes multiple seconds to initialize
5. Doesn't work on JS
6. Doesn't work on dotty, but dotty's alternative `tasty-reflect` needs the entire compiler on the classpath, takes even longer to initialize.

In Izumi we ditched scala-reflect and made `LightTypeTag` – and an entire Universe of problems went away (pun intended), they solve all of the above problems except dotty support – for now. They support subtype checks and refinement unpacking used here in `taggedIsSubtype` & `taggedGetHasServices`. (although I think these are a bad idea – a list of types is better collected with a macro at compile time, that way a simpler version of type tags without inheritance data can be used).
They also support construction from partially-known types, so the following will work:

```scala
import izumi.fundamentals.reflection.tags.TagKK

def getSvc[F[_, _]: TagKK]: URIO[Has[Service[F]], Service[F]] = ZIO.access(_.get[Service[F]])
```

I'm surprise  because I thought you knew about this library since you read the [blog post](https://blog.7mind.io/lightweight-reflection.html) before, but anyway, this is the exact problem that it's created to address, so I think it's a good idea to use it.

[1] Tag API – https://github.com/7mind/izumi/blob/develop/fundamentals/fundamentals-reflection/src/main/scala/izumi/fundamentals/reflection/Tags.scala
[2] LightTypeTag API – https://github.com/7mind/izumi/blob/develop/fundamentals/fundamentals-reflection/src/main/scala/izumi/fundamentals/reflection/macrortti/LightTypeTag.scala

jdegoes: Would you be willing to donate and help integrate LightTypeTag for Scala 2.x?

ioleo:
Great work everyone involved :heart: 
We still need to rewrite `docs/howto/howto_use_module_pattern.md`.
I'll try to play around with this encoding this week.

core/jvm/src/main/scala-2.x/zio/ScalaSpecific.scala
neko-kai: @jdegoes
Yeah, if we extract it into `zio/izumi-reflect`, then zio core can depend on it - is that what you have in mind?

jdegoes: If we "move" it into ZIO then we can strip it down to the bare minimum of what we need, which could reduce sizes. Could that work?

neko-kai: @jdegoes 
You mean copypaste an altered version directly into this repository? That would mean having to maintain two slightly different copies of the same codebase and would double the maintenance effort, so I don't think that can work very well. Re: sizes, LTT jar is currently 1 MB, most of which is macros, vs. scala-reflect's 3.5 MB – and Scala.js eliminates all unreachable code, so I would expect only a fraction of that to make into output JS.

jdegoes: All right, let's pull the whole thing into ZIO and publish as a separate JAR. That work?

PR #2595, opened 2020-01-10
Clean up welcomeMessage

reibitto:
This is mainly to get rid of the fancy unicode bullet point. I noticed it's rendering as a garbled mess for me on Windows. Better to keep it simple than to deal with those kind of issues.

Also added subItem for cleaner rendering to prevent lines from getting really long.


adamgfraser:
Great! Looks like it just needs a format.

reibitto:
Oops... that bites me every time.

PR #2602, opened 2020-01-11
WIP: migrate SampleSpec

unclebob418:
last spec from #1898 :)

this is working, but it's a pretty minimal migration. The assertions could possibly be cleaned up a little, let me know if you think it's worthwhile.

There's still a bit of work to do to clean up the old test framework; I'm not entirely sure how that all fits together though.


unclebob418:
@adamgfraser can you please review?

PR #2604, opened 2020-01-11
Optimize foreachPar_: can handle large input fast #1182

evis:
Parallelize effects manually, using fibers.

```
[info] Benchmark                                 Mode  Cnt   Score   Error  Units
[info] ForeachParN_Benchmark.foreachParN_       thrpt    5  17.603 ± 0.584  ops/s
[info] ForeachParN_Benchmark.naiveForeachParN_  thrpt    5   0.056 ± 0.005  ops/s
```


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2242,12 +2242,36 @@ object ZIO {
...
+  def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] = {
+    val size = as.size
+    for {
+      parentId <- ZIO.fiberId
+      result   <- Promise.make[E, Unit]
+      succeed  <- Ref.make(0)
+      _ <- ZIO.traverse_(as) {
+            f(_).fork >>= { fiber =>
+              result.await.catchAll(_ => fiber.interruptAs(parentId)).fork *>
+                fiber.join.foldM(result.fail, _ => {
```
regiskuckaertz: This will block until the fiber is done, effectively running each effect in sequence.

evis:
Benchmark results with bugfix are even better:

```
[info] Benchmark                                 Mode  Cnt   Score   Error  Units
[info] ForeachParN_Benchmark.foreachParN_       thrpt    5  27.325 ± 0.386  ops/s
```

iravid:
@evis The benchmark is for `foreachPar_`, right? Could you rename it and the methods in it?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2242,12 +2242,37 @@ object ZIO {
...
-      .refailWithTrace
+  def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] = {
+    val size = as.size
+    for {
+      parentId <- ZIO.fiberId
+      result   <- Promise.make[E, Unit]
+      succeed  <- Ref.make(0)
+      _ <- ZIO.traverse_(as) {
+            f(_)
+              .foldM(result.fail, _ => {
```
iravid: Currently defects in a fiber will not cause the other fibers to be terminated. So let's do this:
```suggestion
              .foldCauseM(result.halt, _ => {
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2242,12 +2242,37 @@ object ZIO {
...
+                succeed.update(_ + 1) >>= { succeed =>
+                  ZIO.when(succeed == size)(result.succeed(()))
+                }
+              })
+              .fork >>= { fiber =>
+              result.await.catchAll(_ => fiber.interruptAs(parentId)).fork
```
iravid: And here we need `catchAllCause`

regiskuckaertz:
wow 😮 massive improvement 💪 🕺

adamgfraser:
Fantastic!

PR #2607, opened 2020-01-11
ZIO STM: add when & whenM (#2606)

### LINKED ISSUE #2606: STM: Add when & whenM
I think it would be useful for api parity with ZIO

KadekM:
Closes #2606

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
@@ -414,6 +414,34 @@ object ZSTMSpec extends ZIOBaseSpec {
         } yield assert(result)(equalTo(2 -> 2))
       }
     ),
+    suite("when combinators")(
+      testM("when true") {
+        for {
+          ref    <- TRef.make(false).commit
+          result <- (STM.when(true)(ref.set(true)) *> ref.get).commit
+        } yield assert(result)(equalTo(true))
+      },
+      testM("when false") {
+        for {
+          ref    <- TRef.make(false).commit
+          result <- (STM.when(false)(ref.set(true)) *> ref.get).commit
+        } yield assert(result)(equalTo(false))
+      },
+      testM("whenM true") {
+        for {
+          ref    <- TRef.make(0).commit
+          isZero = ref.get.map(_ == 0)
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -157,6 +157,16 @@ final class ZSTM[-R, +E, +A] private[stm] (
   def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): ZSTM[R, Nothing, A1] =
     fold(_ => a, identity)
 
+  /**
+   * The moral equivalent of `if (p) exp`
+   */
+  def when[R1 <: R, E1 >: E](b: Boolean): ZSTM[R1, E1, Unit] = ZSTM.when(b)(self)
```
adamgfraser: We don't need the additional type parameters here since they aren't used anywhere.

KadekM: I wondered, but we also have them in https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZIO.scala#L1559 and other places... should I also remove them there?

KadekM:
@adamgfraser alphabetized and added tests

core/shared/src/main/scala/zio/stm/ZSTM.scala
KadekM: removed

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -414,6 +414,21 @@ object ZSTMSpec extends ZIOBaseSpec {
         } yield assert(result)(equalTo(2 -> 2))
       }
     ),
+    suite("when combinators")(
```
adamgfraser: Can we also test that the transaction is not executed if the condition is false?

KadekM: I've added them

KadekM:
> Looks great. Will merge when it goes green. Also generally no need to force push. We squash commits when we merge and it generally makes it easier to review with the commits unless there is a specific reason you want to force it. Thanks for your contribution!

ah ok, will keep that in mind next time, generally I dont like to see 20 commits per 5lines of change :dancer: , but if you squash thats fine

PR #2609, opened 2020-01-12
Don't Shift To Blocking Executor If Already There

adamgfraser:
Following up on a suggestion from @jdegoes on Discord, adds logic to `blocking` to avoid shifting to the blocking executor if the effect is already run there.

Could we push this logic higher up into `lock` so we always check if we are already running on the executor we are being asked to lock to and if so don't do anything? I think if `lock` is the exclusive way effects are shifted between executors in ZIO and inner locks take priority then this is safe to do.

### DIFF
--- a/core/jvm/src/main/scala/zio/blocking/Blocking.scala
+++ b/core/jvm/src/main/scala/zio/blocking/Blocking.scala
@@ -66,7 +66,11 @@ object Blocking extends Serializable {
      * Locks the specified effect to the blocking thread pool.
      */
     def blocking[R1 <: R, E, A](zio: ZIO[R1, E, A]): ZIO[R1, E, A] =
-      blockingExecutor.flatMap(exec => zio.lock(exec))
+      for {
+        executor <- blockingExecutor
+        here     <- ZIO.effectTotal(executor.here)
+        a        <- if (here) zio else zio.lock(executor)
+      } yield a
 
     /**
      * Imports a synchronous effect that does blocking IO into a pure value.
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/zio/blocking/Blocking.scala
```diff
@@ -66,7 +66,11 @@ object Blocking extends Serializable {
      * Locks the specified effect to the blocking thread pool.
      */
     def blocking[R1 <: R, E, A](zio: ZIO[R1, E, A]): ZIO[R1, E, A] =
-      blockingExecutor.flatMap(exec => zio.lock(exec))
+      for {
+        executor <- blockingExecutor
+        here     <- ZIO.effectTotal(executor.here)
```
jdegoes: Unfortunately, `here` is not implemented in any executor. I suggest we do this inside `FiberContext`. When doing `Lock`, we can check to see if the provided Executor is the same (`eq`) as the one on the stock. If it is, don't do anything. Otherwise, do the normal thing.

That will fix it for all thread pools, as long as the executors are static. Then we need to make sure ZIO isn't created new executors every time a `def` is called or anything, and then we should be good to go.

adamgfraser: Excellent point. :smiley:

I updated to do this in `FiberContext` as you suggested. I looked at the constructors for `Executor` and `Blocking` and they appear to generate a single executor so object equality should work. I also manually verified that this works correctly by adding a debug statement to `FiberContext`, though unfortunately it is hard to create a regression test for this.

PR #2610, opened 2020-01-12
add toMap to TMap

KadekM:


core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -158,6 +158,12 @@ final class TMap[K, V] private (
   def toList: STM[Nothing, List[(K, V)]] =
     fold(List.empty[(K, V)])((acc, kv) => kv :: acc)
 
+  /**
+   * Collects all bindings into a map.
+   */
+  def toMap: STM[Nothing, Map[K, V]] =
+    fold(Map.empty[K, V])((acc, kv) => acc + (kv._1 -> kv._2))
```
regiskuckaertz: Looks like you can simplify:
```suggestion
    fold(Map.empty[K, V])(_ + _)
```

KadekM: thanks!

KadekM:
> Looks great! Could you add a couple of tests please?

Of course

KadekM:
@adamgfraser  I've added a case for `toMap` and also for `toList`.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
```diff
@@ -132,6 +132,26 @@ object TMapSpec extends ZIOBaseSpec {
...
     suite("transformations")(
+      testM("toList") {
+        val elems = List("a" -> 1, "b" -> 2)
+        val tx =
+          for {
+            tmap <- TMap.make(elems: _*)
+            list <- tmap.toList
+          } yield list
+
+        assertM(tx.commit)(hasSameElements(elems))
```
adamgfraser: Don't we want to assert here and below that the two collections are equal, not just that they have the same elements? It doesn't really matter for the map but we definitely want the lists to be the same (i.e. same elements in the same order) and the two maps should be equal as well.

KadekM: @adamgfraser Is it requirement for TMap to preserve order? I thought it's only it's implementation detail.

adamgfraser: It isn’t a requirement to preserve order since you are correct that it is an implementation detail but for the same reason the definition of equality for maps is that they have the same bindings regardless of order.

KadekM: >  we definitely want the lists to be the same (i.e. same elements in the same order)

If I assert for equivalence here, if `tmap.toList` returns in different order (which it can, since TMap doesn't has no such invariant), the test would fail.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
```diff
@@ -132,6 +132,26 @@ object TMapSpec extends ZIOBaseSpec {
       }
     ),
     suite("transformations")(
+      testM("toList") {
+        val elems = List("a" -> 1, "b" -> 2)
+        val tx =
+          for {
+            tmap <- TMap.make(elems: _*)
```
adamgfraser: You could use `TMap.fromIterable` here and below to clean things up slightly if you wanted to.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
adamgfraser: You're right. The list won't necessarily be in the same order.

adamgfraser:
Just needs a format.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
KadekM: updated

KadekM:
Formatted.
@adamgfraser thank you for review.

adamgfraser:
You can use `fmt` to format all the source files so you don't have any issue with CI.

adamgfraser:

PR #2611, opened 2020-01-12
Bugfix: complete foreachPar_ on empty input

evis:
Promise was completed only when there were some elements inside passed
iterable.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2253,27 +2253,28 @@ object ZIO {
    *
    * Additionally, interrupts all effects on any failure.
    */
-  def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] = {
-    val size = as.size
-    for {
-      parentId <- ZIO.fiberId
-      result   <- Promise.make[E, Unit]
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2258,6 +2258,7 @@ object ZIO {
     for {
       parentId <- ZIO.fiberId
       result   <- Promise.make[E, Unit]
+      _        <- ZIO.when(as.isEmpty)(result.succeed(()))
```
mschuwalow: How about:

```scala
  def foreachPar_[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] = {
    if (as.isEmpty) ZIO.unit
    else {
      val size = as.size
      for {
        parentId <- ZIO.fiberId
        result   <- Promise.make[E, Unit]
        succeed  <- Ref.make(0)
        _ <- ZIO.traverse_(as) {
          f(_)
              .foldCauseM(result.halt, _ => {
                succeed.update(_ + 1) >>= { succeed =>
                  ZIO.when(succeed == size)(result.succeed(()))
                }
              })
              .fork >>= { fiber =>
              result.await.catchAllCause(_ => fiber.interruptAs(parentId)).fork
            }
          }
        _ <- result.await
      } yield ()
    }
  }
```

Issue #2601, opened 2020-01-11
add a TQueue.empty constructor


sken77:
I found myself needing this. I notice that `TMap` has an empty constructor but `TQueue` doesn't. is there a reason for this? If not, i would gladly add it

adamgfraser:
I believe the issue is that queues are parameterized on a size. Were you looking for a queue with no maximum size? I think that could definitely be a good addition.

sken77:
whats the reason for requiring a capacity? the underlying queue doesnt seem to need one

```
object TQueue {
  final def make[A](capacity: Int): STM[Nothing, TQueue[A]] =
    TRef.make(ScalaQueue.empty[A]).map(ref => new TQueue(capacity, ref))
}
```

adamgfraser:
I think we want to support queues that are both bounded and unbounded. Queues are control flow structures where offering to a queue that doesn’t have capacity suspends so users may want queues with specific sizes, whereas for something like a Map that consideration doesn’t really apply.

sken77:
Also, would be great to add `front/peek`, `back` and `swap`

sken77:
I see. what do you think about the above methods? You think we still want an unbounded Queue? these are 2 separate issues now. but wanted to mention it now that i have your attention

adamgfraser:
Yes, I think we want to support both bounded and unbounded queues. We should be able to do that with one class similar to the standard queue, either by having a unbounded queue just have a size of `Int.MaxValue` or slightly expanding the internal state that is maintains. Also think we want to add more combinators like the one you mentioned.

@mijicd Does that make sense?

mijicd:
@adamgfraser absolutely. @sken77 feel free to tackle both of the issues (constructors and missing methods), and let me know if you need any help.

sken77:
excellent ill get to it

sken77:
this is done

PR #2612, opened 2020-01-12
Expose more methods for Tqueue

sken77:
Add more methods for Tqueue like the ones already added and started to get feedback on what else might be needed or not including a way to create an unbounded queue

https://github.com/zio/zio/issues/2601

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/TQueueSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TQueueSpec.scala
@@ -36,7 +36,7 @@ object TQueueSpec extends ZIOBaseSpec {
     ),
     suite("insertion and removal")(
       testM("offer & take") {
-        val t = for {
+        val tx = for {
           tq    <- TQueue.make[Int](5)
           _     <- tq.offer(1)
           _     <- tq.offer(2)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,75 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+      _.headOption match {
+        case Some(a) => STM.succeed(a)
+        case None    => STM.retry
+      }
+    )
+
+  /**
+   * View the last element inserted into the queue. retries if the queue is empty
+   */
+  def back: STM[Nothing, A] =
```
jdegoes: Maybe call this one `last`?

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,75 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+        case None => STM.retry
+      }
+
+    ref.get.flatMap(go)
+  }
+
+  /**
+   * Splits the queue into a prefix/suffix pair according to a predicate.
+   */
+  def span(f: A => Boolean): STM[Nothing, (TQueue[A], TQueue[A])] =
```
jdegoes: `partition`?

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,75 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+  def groupBy[K](f: A => K): TMap[K, TQueue[A]] = ???
+  //size comparators
+  def longerThan[B](t: TQueue[B]): Boolean = ???
 }
 
 object TQueue {
   def make[A](capacity: Int): STM[Nothing, TQueue[A]] =
     TRef.make(ScalaQueue.empty[A]).map(ref => new TQueue(capacity, ref))
+
+  def makeUnbounded[A]: STM[Nothing, TQueue[A]] = make(Int.MaxValue)
```
jdegoes: `unbounded`

core/shared/src/main/scala/zio/stm/TQueue.scala
jdegoes: This is the one I'm not too sure about. In general, anything in `STM` that returns another structure is probably not that useful. STM is designed for modifying structures, not producing new structures. I think I'd delete this one but keep the others.

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,75 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+  def span(f: A => Boolean): STM[Nothing, (TQueue[A], TQueue[A])] =
+    for {
+      q      <- ref.get
+      (a, b) = q.span(f)
+      aa     <- TRef.make(a)
+      bb     <- TRef.make(b)
+    } yield (new TQueue(capacity, aa), new TQueue(capacity, bb))
+
+  //other possible additions
+  def groupBy[K](f: A => K): TMap[K, TQueue[A]] = ???
```
jdegoes: I'd not keep this one for the same reason: STM is is all about "mutable" structures, not immutable ones. So methods that return copies don't make a whole lot of sense.

sken77:
> 
> 
> Looking good! Just a few comments.
> 
> Also I think it would be safe to expose an STM method like `underlying: TRef[ScalaQueue[A]]` that exposes the underlying queue. That way, if there's something else you want to do, you can always do it.

Oh i added as `STM[Nothing, Tref[...]]`

core-tests/shared/src/test/scala/zio/stm/TQueueSpec.scala
```diff
@@ -0,0 +1,194 @@
...
+          assert(z._2)(equalTo(5))
+      },
+      testM("view the last value") {
+        val t = for {
+          tq   <- TQueue.unbounded[Int]
+          _    <- tq.offerAll(List(1, 2, 3, 4, 5))
+          last <- tq.last
+          size <- tq.size
+        } yield (last, size)
+        for {
```
mijicd: It might be slightly more readable in the following form:

```scala
assertM(t.commit)(equalTo((5,5)))
```

Also, one nitpick, just to align with other tests prefer `tx` to `t`.

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,71 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+  /**
+   * Compares two Queues with the same element type and returns the longest one. If equal it returns the left side.
+   */
+  def longest(t: TQueue[A]): STM[Nothing, TQueue[A]] =
+    (size <*> t.size).map { case (a, b) => if (a >= b) this else t }
+
+  /**
+   * Returns the underlying queue
+   */
+  def underlying: STM[Nothing, TRef[ScalaQueue[A]]] = STM.succeed(ref)
```
mijicd: I'm not sure about this one. While I understand the motivation, even when wrapped in `TRef` it feels like a leak. In case it ever happens that we need to change the internal encoding, this might be an ugly blocker.

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,71 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+        case None => STM.retry
+      }
+
+    ref.get.flatMap(go)
+  }
+
+  /**
+   * Compares two Queues with the same element type and returns the longest one. If equal it returns the left side.
+   */
+  def longest(t: TQueue[A]): STM[Nothing, TQueue[A]] =
```
adamgfraser: Seems a little specialized.

core/shared/src/main/scala/zio/stm/TQueue.scala
jdegoes: It's true. Maybe we should avoid it...

core/shared/src/main/scala/zio/stm/TQueue.scala
jdegoes: Agreed, I think we can do without `longest`.

jdegoes:
@sken77 Two more changes (delete `longest` and `underlying` — sorry about the latter!), and it's good to merge!

adamgfraser:
@jdegoes Aren't they already deleted on the latest version?

jdegoes:
@adamgfraser Oops!

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -54,9 +54,59 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
...
+          else go(as)
+        case None => STM.retry
+      }
+
+    ref.get.flatMap(go)
+  }
 }
 
 object TQueue {
   def make[A](capacity: Int): STM[Nothing, TQueue[A]] =
```
jdegoes: We should call this `bounded` after `Queue.bounded`. If you have the energy to do a followup PR, that'd be great!

jdegoes:
@sken77 Thank you for this contribution! 🎉 

@adamgfraser Thanks for the review!

core/shared/src/main/scala/zio/stm/TQueue.scala
sken77: i'll open the issue so i dont forget!

PR #2614, opened 2020-01-13
Optimize ZIO.mergeAllPar #1182

evis:
Parallelize effects using `foreachPar_` instead of building a tree of
effects with `zipPar`.

```
[info] Benchmark                               Mode  Cnt   Score   Error  Units
[info] MergeAllParBenchmark.mergeAllPar       thrpt    5  11.575 ± 0.344  ops/s
[info] MergeAllParBenchmark.naiveMergeAllPar  thrpt    5   0.038 ± 0.046  ops/s
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1,5 +1,7 @@
 package zio
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import scala.annotation.tailrec
 import scala.util.{ Failure, Success }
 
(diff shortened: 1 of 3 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -722,6 +724,37 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(a)(equalTo(b))
       }
     ),
+    suite("mergeAllPar")(
+      testM("return zero element on empty input") {
+        val zeroElement = 42
+        val nonZero     = 43
+        UIO.mergeAllPar(Nil)(zeroElement)((_, _) => nonZero).map {
+          assert(_)(equalTo(zeroElement))
```
evis: This test will succeed when #2611 will be merged.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2601,11 +2601,20 @@ object ZIO {
...
    * Merges an `Iterable[IO]` to a single IO, working in parallel.
+   *
+   * It's possible to execute side effects inside `f`, as `f` is guaranteed to
+   * be executed at most once for each `in` element.
    */
   def mergeAllPar[R, E, A, B](
     in: Iterable[ZIO[R, E, A]]
   )(zero: B)(f: (B, A) => B): ZIO[R, E, B] =
-    in.foldLeft[ZIO[R, E, B]](succeed[B](zero))((acc, a) => acc.zipPar(a).map(f.tupled)).refailWithTrace
+    RefM.make(zero) >>= { acc =>
```
iravid: Let’s rewrite this with Ref instead.

evis: Then `f` will be retried on high contention, and it will be unsafe to pass side-effecting functions. Are we sure it's OK to change this predicate? Old `mergeAllPar` version doesn't retry `f`.

iravid: Perfectly fine. The ZIO API makes no accommodations for side effecting functions when pure functions are expected.

evis: e.g., this test doesn't pass if rewrite with Ref: https://github.com/zio/zio/pull/2614/files#diff-ce160572c9127744730d5a14dc428162R746-R755

iravid: Right. You should rewrite it without side effects in f.

evis:
@regiskuckaertz @iravid could we merge it please?

PR #2617, opened 2020-01-13
ZIO Test: Implement TestAspect#nonTermination

adamgfraser:
Following up on a question I got, this provides an easy way to require that a test does not terminate within a specified time.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -189,7 +189,7 @@ object TestAspect extends TimeoutVariants {
    * Constructs an aspect that requires a test to not terminate within the
    * specified time.
    */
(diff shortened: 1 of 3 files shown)


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -184,6 +185,22 @@ object TestAspect extends TimeoutVariants {
         effect *> test
     }
 
+  /**
+   * Constructs an aspect that requires a test to not terminate within the
+   * specified time.
+   */
+  def doesNotTerminate(duration: Duration): TestAspect[Nothing, Live[Clock], Nothing, Any, Unit, Unit] =
```
jdegoes: `nonTermination`?

jdegoes:
@adamgfraser Great idea! 👍

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Sure!

PR #2624, opened 2020-01-14
Add algolia search to microsite

softinio:
This is what it looks like (top right): 

https://shared.softinio.com/Screen-Shot-2020-01-14-08-21-43.93.png


Tamriel:
It's not live yet?

adamgfraser:
Right now changes to site only get published when we do a release.

PR #2626, opened 2020-01-14
ZIO Test: Support Automatic Derivation of Generators

### LINKED ISSUE #2570: Should We Have ZIO Test Magnolia Project for Deriving Generators?
I have gotten a couple of requests for support for automatically deriving generators for algebraic data types and I believe @ghostdogpr has implemented a version of this in one of his projects, so I think we should discuss whether it makes sense to have a subproject to provide this functionality to users who want it.

We are already using Magnolia to do derivation for ZIO Config and it would be a separate module so we wouldn't be forcing a dependency on anyone. The main downside I see is that I believe when deriving a generator for an algebraic data type we would need to use some "canonical" generator for types such as `Int`, `String`, `List`, and so on, which to a certain extent goes against the ZIO Test design philosophy of explicit generators and that there is not a "one size fits all" generator for a type. At the same time, this would all be opt in and users could always fall back to the existing combinators in ZIO Test if they wanted more fine grained control of their generators.

What do others think?

adamgfraser:
Resolves #2570.


test-magnolia/shared/src/main/scala/zio/test/magnolia/Derivation.scala
```diff
@@ -0,0 +1,92 @@
...
+  implicit val genUnit: Typeclass[Unit]       = Gen.unit
+  implicit val genUUID: Typeclass[UUID]       = Gen.anyUUID
+
+  implicit def genEither[A, B](implicit ev1: Typeclass[A], ev2: Typeclass[B]): Typeclass[Either[A, B]] =
+    Gen.either(ev1, ev2)
+
+  implicit def genFunction[A, B](implicit ev: Typeclass[B]): Typeclass[A => B] =
+    Gen.function(ev)
+
+  implicit def genList[A](implicit ev: Typeclass[A]): Typeclass[List[A]] =
```
jdegoes: Do we need to add instances for `Vector`, etc.?

zach-albia:
> Separately, I have been thinking about a ZIO Fake project that could add Gen for things like emails, usernames, passwords, company names, addresses, phone numbers, names, ages, website URLs, etc. Could eventually grow to have hundreds of gens in there for common purposes.

Would love to help out with this, with some guidance from people.

test-magnolia-tests/shared/src/test/scala/zio/test/magnolia/DerivationSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+import zio.test.GenUtils._
+import zio.test.Assertion._
+
+import zio.test.magnolia.Derivation._
+
+object DerivationSpec extends DefaultRunnableSpec {
+
+  final case class Person(name: String, age: Int)
+
+  val genPerson: Gen[Random with Sized, Person] = Derivation.gen[Person]
```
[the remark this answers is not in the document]
adamgfraser: Yes, I was thinking about that too. I think the current syntax is a little more idiomatic Magnolia but I like your suggestion. We could also do `DerivingGen[Person]`.

adamgfraser:
@jdegoes Will add documentation. I think it makes sense to have in core for now so it is easy to make changes to ZIO Test to support this functionality in the early stages. At some point we could spin off into its own repository if we wanted. I think ZIO Fake is a great idea!

@zach-albia Would be great to get your help with that! It seems like that could be a great project to get a lot of people involved with because all of those generators would require some individualized thought and double be done independently of each other so people could pick and choose what they wanted to work on.

test-magnolia/shared/src/main/scala/zio/test/magnolia/Derivation.scala
adamgfraser: Yes. I missed that one. I added an instance for `Seq` because we need that to derive variable generators for case classes with variable arguments parameters but I think we also need ones for types like `Iterable`. Will take another pass on this.

test-magnolia-tests/shared/src/test/scala/zio/test/magnolia/DerivationSpec.scala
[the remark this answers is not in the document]
jdegoes: I'm ok with either one `DerivingGen[Person]` or the more "idiomatic" Magnolia way (I wonder how they deal with name clashes if it's idiomatic to name it `Derivation`???).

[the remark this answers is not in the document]
adamgfraser: I think they normally do `DerivationGen` or `GenDerivation` but I like the way we're going a bit more. Will post another version shortly.

adamgfraser:

I ran into an implicit resolution issues because the type class is covariant so if we want to summon an implicit instance for `Seq[Int]` both `List[Int]` and `Vector[Int]` would be more specific and ambiguous instances than the instance for `Seq[Int]`. I tried the usual tricks such as prioritized implicits but was not able to resolve, I think because of a combination of the inherent implicit resolution issues with the macro based machinery in Magnolia.

I resolved by creating an invariant wrapper around the generator. This seems to work well and supports the nice `DeriveGen[Person]` syntax we talked about, but let me know what you think. Also added a bunch of tests.

docs/usecases/testing.md
```diff
@@ -60,6 +61,23 @@ val associativity =
 
 If a property fails, the failure will be automatically shrunk to the smallest failing cases to make it easier for you to diagnose the problem. And shrinking is integrated with the generation of random variables, so you are guaranteed that any shrunk counterexample will meet the conditions of your original generator.
 
+ZIO Test also supports automatic derivation of generators using the ZIO Test Magnolia module:
+
+```scala
```
jdegoes: Possible to use mdoc?

adamgfraser: Yes! On it.

PR #2628, opened 2020-01-15
Implement ZIO#forkDaemon

adamgfraser:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -587,24 +587,35 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
   /**
    * Forks the effect into a new independent fiber, with the specified name.
    */
   final def forkAs(name: String): URIO[R, Fiber[E, A]] =
     (Fiber.fiberName.set(Some(name)) *> self).fork
 
+  /**
+   * Forks the effect into a new daemon fiber, but immediately restores the
+   * fiber's daemon status to the inherited status from whatever region the
+   * effect is composed into. This means that the forked fiber will not
```
```suggestion
   * effect is composed into. This means that the forked fiber will not be
```

jdegoes:
@adamgfraser Love it!

@regiskuckaertz Thanks for the review!

jdegoes:
@adamgfraser Looks like we got a Circle CI failure on merge, can you take a look?

adamgfraser:
@jdegoes Yes, looking now.

adamgfraser:
@jdegoes Looks like it is related to a flaky test for `Stream.groupedWith`. I will open an issue and a PR to mark it as flaky momentarily.

```scala
[info] ZIO Test
[info] - StreamSpec
[info]   - Stream.groupedWithin
[info]     - group before chunk size is reached due to time window
[info]       List(List(1, 2), List(3), List(4, 5)) did not satisfy equalTo(List(List(1, 2), List(3, 4), List(5)))
```

Issue #2629, opened 2020-01-15
ZStream#groupedWithin Test is Flaky


adamgfraser:
@mschuwalow Can you take a look?

```
[info] ZIO Test
[info] - StreamSpec
[info]   - Stream.groupedWithin
[info]     - group before chunk size is reached due to time window
[info]       List(List(1, 2), List(3), List(4, 5)) did not satisfy equalTo(List(List(1, 2), List(3, 4), List(5)))
```

https://circleci.com/gh/zio/zio/50278?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

mschuwalow:
I'll take a look, sorry

adamgfraser:
No problem. It is a tricky one. Can you approve my PR marking the test as flaky so we can resolve the build failure on master?

mschuwalow:
done, I'll push a fixed version in a bit

adamgfraser:
Thanks!

mschuwalow:
@adamgfraser can we close this?

adamgfraser:
Yes.

PR #2631, opened 2020-01-15
update groupedWithin spec

mschuwalow:
#2629 
@adamgfraser 

This seems better than the previous version, but still flaky as there is a race between tap and groupedWithin actually becoming aware of the new element.

"group based on time passed" is also flaky with a low probability and does not cover any behavior the other tests don't, so removed for now


adamgfraser:
Can you merge upstream changes?

mschuwalow:
I'll try to update it early tomorrow

adamgfraser:
Sorry, can you merge upstream changes one more time to trigger the Dotty CI task?

PR #2635, opened 2020-01-16
Optimize ZIO.reduceAllPar #1182

evis:
Parallelize effects using `foreachPar_` via `mergeAllPar` instead of
building a tree of effects with `zipPar`.

```
[info] Benchmark                                 Mode  Cnt   Score    Error  Units
[info] ReduceAllParBenchmark.naiveReduceAllPar  thrpt    5   0.047 ±  0.034  ops/s
[info] ReduceAllParBenchmark.reduceAllPar       thrpt    5  28.159 ± 13.505  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2730,10 +2730,30 @@ object ZIO {
    */
   def reduceAllPar[R, R1 <: R, E, A](a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]])(
     f: (A, A) => A
-  ): ZIO[R1, E, A] =
-    as.foldLeft[ZIO[R1, E, A]](a) { (l, r) =>
-      l.zipPar(r).map(f.tupled)
-    }
+  ): ZIO[R1, E, A] = {
+    def prepend(a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]]): Iterable[ZIO[R1, E, A]] =
+      new Iterable[ZIO[R1, E, A]] {
+        override def iterator: Iterator[ZIO[R1, E, A]] = new Iterator[ZIO[R1, E, A]] {
+          private var started    = false
+          private val asIterator = as.iterator
+
+          override def hasNext: Boolean = !started || asIterator.hasNext
+
+          override def next(): ZIO[R1, E, A] =
+            if (!started) {
+              started = true
+              a
+            } else {
+              asIterator.next()
+            }
+        }
+      }
+
+    val all = prepend(a, as)
+    mergeAllPar(all)(Option.empty[A]) { (acc, elem) =>
+      Some(acc.fold(elem)(f(_, elem)))
+    }.map(_.get)
+  }
 
   /**
    * Replicates the given effect n times.
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2731,22 +2731,9 @@ object ZIO {
   def reduceAllPar[R, R1 <: R, E, A](a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]])(
     f: (A, A) => A
   ): ZIO[R1, E, A] = {
-    def prepend(a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]]): Iterable[ZIO[R1, E, A]] =
-      new Iterable[ZIO[R1, E, A]] {
-        override def iterator: Iterator[ZIO[R1, E, A]] = new Iterator[ZIO[R1, E, A]] {
-          private var started    = false
-          private val asIterator = as.iterator
-
-          override def hasNext: Boolean = !started || asIterator.hasNext
-
-          override def next(): ZIO[R1, E, A] =
-            if (!started) {
-              started = true
-              a
-            } else {
-              asIterator.next()
-            }
-        }
+    def prepend[Z](z: Z, zs: Iterable[Z]): Iterable[Z] =
+      new Iterable[Z] {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2730,10 +2730,30 @@ object ZIO {
...
   def reduceAllPar[R, R1 <: R, E, A](a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]])(
     f: (A, A) => A
-  ): ZIO[R1, E, A] =
-    as.foldLeft[ZIO[R1, E, A]](a) { (l, r) =>
-      l.zipPar(r).map(f.tupled)
-    }
+  ): ZIO[R1, E, A] = {
+    def prepend(a: ZIO[R, E, A], as: Iterable[ZIO[R1, E, A]]): Iterable[ZIO[R1, E, A]] =
+      new Iterable[ZIO[R1, E, A]] {
+        override def iterator: Iterator[ZIO[R1, E, A]] = new Iterator[ZIO[R1, E, A]] {
```
evis: Not sure, if we should locate this method here or somewhere else

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2730,10 +2730,30 @@ object ZIO {
...
+            } else {
+              asIterator.next()
+            }
+        }
+      }
+
+    val all = prepend(a, as)
+    mergeAllPar(all)(Option.empty[A]) { (acc, elem) =>
+      Some(acc.fold(elem)(f(_, elem)))
+    }.map(_.get)
```
evis: Is it better to avoid unnecessary allocations and just deal with null?

```scala
mergeAllPar(all)(null) { (acc, elem) =>
  if (acc eq null) elem else f(acc, elem)
}
```

core/shared/src/main/scala/zio/ZIO.scala
evis: Or maybe, it's OK to just prepend using `Iterable(a) ++ as`

evis:
test_213_jdk8_jvm failed due to `ssh: connect to host github.com port 22: Connection timed out`, could someone restart it please?

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: It looks a lot like `mergeAllPar(as)(a)(f)` 🤔

evis: If we pass `a` as a second argument, then `f` should be of type `(ZIO[R, E, A], A) => ZIO[R, E, A]`. So, doesn't match.

iravid: The allocation overhead is probably miniscule compared to the concurrency overhead, so I wouldn't bother with working with `null` here.

core/shared/src/main/scala/zio/ZIO.scala
iravid: This seems unnecessary. You could do `Iterator(a) ++ as.iterator` to be lazy about this.

core/shared/src/main/scala/zio/ZIO.scala
```suggestion
    }.flatMap(_.fold(a)(UIO.succeed))
```

regiskuckaertz: Sure, I was just questioning the using of `Option`, is all.
```scala
a.map(mergeAllPar(as)(_)(f))
```
perhaps?

evis: If we ever make a bug in this code, then such `flatMap` will silently execute `a` effect twice. Are we sure it's better to rush for purity in this case?

evis: Then `a` isn't run in parallel with `as`. :(

regiskuckaertz: Fair enough, that's too bad but hey the numbers are awesome!

adamgfraser: Yes, I think unfortunately we need this. We could use `zipWithPar` to execute the `a` in parallel but then we don't have a zero value for the merge operation.

adamgfraser: It does seem like it is an internal error of ZIO if `get` fails so I would be inclined to leave it as is.

evis:
@adamgfraser @iravid could you merge it please?

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: @iravid Are you good with this?

adamgfraser:
@evis Just wanted to make sure that @iravid was good with this since we are not incorporating his comment. Otherwise ready to merge anytime from my end.

jdegoes:
@evis Awesome work and love the benchmarks accompanying the Pull Request, demonstrating the improvement in performance.

adamgfraser:
@iravid Do you want to fold over the `Option` and we can die with a message about a ZIO internal error in the empty case?

iravid:
I need to open this combinator in an editor to see why this is awkward to express currently and haven’t had a chance to do that. 

Maybe we should just manually fork `a` and use mergeAllPar for the as?

adamgfraser:
I thought about that. The problem is that we still need a zero value for `mergeAllPar`, which we don't have because the `a` is running in parallel and isn't available yet.

I think conceptually the problem we're running into is we don't have an optimized combinator for the non-empty case. If we had an optimized combinator for the non-empty case, then we could implement the non-empty combinators in terms of it and we could also easily implement the combinators that accept an empty collection, since we can just check if the collection is empty and if so immediately return it or else call the non-empty version of the combinator.

iravid:
Yes, agreed on that.

PR #2647, opened 2020-01-17
Split waiting for errors and waiting for completion in foreachPar_. 

### LINKED ISSUE #2642: Flaky foreachPar_ Test
@horned-sphere Can you take a look?

```
[info] ZIO Test
[info] - ZIOSpec
[info]   - foreachPar_
[info]     - accumulates errors
[info]       Set(1, 2) did not satisfy equalTo(Set(1, 2, 3))
```

https://circleci.com/gh/zio/zio/50708?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

horned-sphere:
Fixes problem where errors are not guaranteed to be accumulated. Fixes #2642


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2300,24 +2300,34 @@ object ZIO {
...
+        causes         <- Ref.make[Cause[E]](Cause.empty)
+        result         <- Promise.make[Nothing, Boolean]
+        failureTrigger <- Promise.make[Unit, Unit]
+        count          <- Ref.make(0)
         fibers <- ZIO.traverse(as) {
                    f(_)
-                     .foldCauseM(c => causes.update(_ && c) *> result.fail(()), _ => {
-                       (succeed.update(_ + 1) >>= { succeed =>
-                         ZIO.when(succeed == size)(result.succeed(()))
+                     .foldCauseM(c => causes.update(_ && c) *> failureTrigger.fail(()), _ => ZIO.unit)
```
adamgfraser: We don't need to add as part of this PR but you could see a `tapCause` combinator being useful here.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2300,24 +2300,34 @@ object ZIO {
...
+                     .ensuring {
+                       (count.update(_ + 1) >>= { done =>
+                         ZIO.when(done == size) {
+                           result.complete(failureTrigger.succeed(()))
+                         }
                        }).uninterruptible
```
adamgfraser: Don't think we need `uninterruptible` here because finalizer in `ensuring` is already uninterruptible.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2300,24 +2300,34 @@ object ZIO {
...
-                     })
+                     }
                      .fork
                  }
-        interrupter = result.await
+        interrupter = failureTrigger.await
           .catchAll(_ => ZIO.foreach(fibers)(_.interruptAs(parentId).fork) >>= Fiber.joinAll)
           .toManaged_
```
adamgfraser: You can use the new `forkManaged` here.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -539,11 +539,21 @@ object ZIOSpec extends ZIOBaseSpec {
     ),
     suite("foreachPar_")(
       testM("accumulates errors") {
-        val failures = ZIO
-          .foreachPar_(1 to 3)(IO.fail(_).uninterruptible)
-          .foldCause(_.failures.toSet, _ => Set.empty)
-        assertM(failures)(equalTo(Set(1, 2, 3)))
-      } @@ flaky,
+        def task(started: Ref[Int], trigger: Promise[Nothing, Unit])(i: Int): IO[Int, Unit] =
```
adamgfraser: Why do we need to change the test here?

adamgfraser: Ah, I see, you are making sure that the effects aren't pre-interrupted. Makes sense.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: In my defence, tapCause didn't exist when I made the PR ;).

I've changed it over as it's fairly trivial and is better.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Good catch. I copy/pasted it into the ensuring without engaging my brain.

core/shared/src/main/scala/zio/ZIO.scala
horned-sphere: Yeah, that does look a bit less mysterious.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
horned-sphere: Yes, sadly the simple version of the test is still flaky as the fibers can be interrupted before they enter the uninterruptible region.

adamgfraser:
Looks great! Will merge when it goes green!

Issue #2271, opened 2019-11-18
Add Additional Methods to Chunk


adamgfraser:
Right now we are a little hit or miss on methods on `Chunk`. For example, we have `drop` and `dropWhile` but we don't have `exists` or `forall`. This can get a little annoying when you are working with `Chunk` and run into one of these situations where you either have to drop back to using a `Seq` or reimplement a method you are used to just having. This ticket is to go through the methods available on `scala.collection.immutable.IndexedSeq`, make a recommendation regarding which of them make sense to implement for `Chunk`, and then once we have agreement implement those methods.

adamgfraser:
Copying @Vilkina.

Vilkina:
I can work on this. Also, we need a benchmark. So I need some help, @adamgfraser :)

saraiva132:
Perhaps list all combinators missing from Chunk similar to how it looks in #2077 ? And reference this issue in future PRs @adamgfraser

pk044:
@Vilkina are you perhaps still working on this? 

i'm up for helping with this issue in case you don't find the time for it :slightly_smiling_face:

Vilkina:
@pk044, you can help if you want. Definitely I don't have time. Thanks

pk044:
@adamgfraser 
what do you think about having these methods in `Chunk`?
```
def find(f: A => Boolean): Option[A]
def head: Option[A]
def last: Option[A]
def indexWhere(f: A => Boolean): Option[Int]
```

adamgfraser:
@pk044 I think those look great. We may want to rename `head` and `last` to `headOption` and `lastOption` to more closely mirror usage in the existing Scala collections API. I realize we use `head` and `last` for `ZStream` but that is an effect with a separate error channel versus this is supposed to be more of a drop in replacement for other data structures so may be more value in sticking closer to the existing API where possible. Similarly would be good to support an optional starting index for `indexWhere`.

pk044:
@adamgfraser do you think there's anything else I can add?

adamgfraser:
@pk044 I think there are a lot more methods we could add. I would take a look at the list of methods available for `scala.collection.immutable.IndexedSeq` [here](https://www.scala-lang.org/api/current/scala/collection/immutable/IndexedSeq.html). That is basically our competition. There are a ton of methods there and obviously some are more or less useful but at some level we need to compete with `Array` and `Vector` in terms of both performance and availability of relevant combinators.

[...]

PR #2648, opened 2020-01-17
Additional methods for Chunk

pk044:
#2271


jdegoes:
@pk044 Wow, this looks great!

I see one failing test (flaky), or maybe bug in Chunk (I didn't check):

```
[info]   - indexWhere
[info]     Test failed after 4 iterations with input: ((Singleton(),Fun( -> true)),-10)
[info]     Original input before shrinking was: ((Singleton(yUhq),Fun(yUhq -> true)),-6)
[info]     0 did not satisfy equalTo(-10)
[info] Done

```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -117,6 +117,83 @@ sealed trait Chunk[+A] { self =>
     case _ => false
   }
 
+  /**
+   * Returns the element that satisfies the predicate.
```
adamgfraser:
```suggestion
   * Returns the first element that satisfies the predicate.
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -117,6 +117,83 @@ sealed trait Chunk[+A] { self =>
...
+    if (isEmpty) None else Some(self(0))
+
+  /**
+   * Returns the last element of this chunk if it exists.
+   */
+  def lastOption: Option[A] =
+    if (isEmpty) None else Some(self(self.length - 1))
+
+  /**
+   * Returns the index for which the given predicate is satisfied.
```
adamgfraser:
```suggestion
   * Returns the first index for which the given predicate is satisfied.
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -117,6 +117,83 @@ sealed trait Chunk[+A] { self =>
...
+   */
+  def indexWhere(f: A => Boolean): Option[Int] =
+    indexWhere(f, 0)
+
+  /**
+   * Returns the index for which the given predicate is satisfied after or at some given index.
```
adamgfraser:
```suggestion
   * Returns the first index for which the given predicate is satisfied after or at some given index.
```

pk044:
> @pk044 Wow, this looks great!
> 
> I see one failing test (flaky), or maybe bug in Chunk (I didn't check):
> 
> ```
> [info]   - indexWhere
> [info]     Test failed after 4 iterations with input: ((Singleton(),Fun( -> true)),-10)
> [info]     Original input before shrinking was: ((Singleton(yUhq),Fun(yUhq -> true)),-6)
> [info]     0 did not satisfy equalTo(-10)
> [info] Done
> ```

interesting. thanks - I'll look into it :)

adamgfraser:
Looks like the flakiness is only occurring on Scala 2.11 on the JVM. May be a version / platform specific issue.

adamgfraser:
It is actually a bug in Scala:

```scala
val as = List("")
as.indexWhere(_ => true, -10) // should be 0 but -10 on Scala 2.11.12
```

https://scastie.scala-lang.org/u0a6DuJ8SVa4FHWOaTTsnQ

@pk044 You can use test aspects or the `TestPlatform` and `TestVersion` functionality to check the version / platform and prevent the test from running on Scala 2.11 JVM.

I will check whether the bug has already been reported and if not file a bug report with Scala team.

adamgfraser:
Ticket opened at https://github.com/scala/bug/issues/11852.

adamgfraser:
Actually I don't think we have quite the functionality to do this in ZIO Test today (we have functionality for Scala 2 versus Dotty but not for specific versions within Scala 2). What if for purposes of this PR we only test for non-negative starting indices and I can work to add something to ZIO Test to address this.

pk044:
@adamgfraser sure :) 

just so we don't forget about this once we are able to exclude this test on 2.11: should I make a GitHub issue or a TODO comment to make the test stop using non-negative indices?

pk044:
@adamgfraser sure :) 

just so we don't forget about this once we are able to exclude this test on 2.11: should I make a GitHub issue or a TODO comment to make the test stop using non-negative indices?

adamgfraser:
Sure. I'm working on right now so should be able to take care of today and fixing this will be a nice test case for the functionality.

adamgfraser:
@pk044 Can you merge upstream changes? I think you are working off a version of master where we disabled Dotty support to get the new environment encoding PR merged so the required CI task is not being triggered.

pk044:
@adamgfraser done! it seems like the tests passed on Dotty

adamgfraser:
@pk044 Great work on this!

PR #2650, opened 2020-01-18
ZIO Test: Improve Version Specific Support

adamgfraser:
Currently version specific support in ZIO Test is limited to Scala 2 and Dotty. However, sometimes it can be necessary to apply different testing logic based on the specific Scala version (e.g. Scala 2.11). This PR adds accessors and test aspects to be able to control the execution of tests based on the specific Scala version and uses this to resolve an issue we had with testing one of the methods on `Chunk` recently.


adamgfraser:
Removed todo as Scala team has indicated they do not intend to do further release of 2.11 that would fix this.

PR #2651, opened 2020-01-18
Implement TQueue#peekOption

adamgfraser:
We previously added `peek` which looks at the first value in the queue without taking it and retries if the queue is empty. It can also be helpful to just look at the first value without retrying, so implements `peekOption` that returns `None` instead of retrying if the queue is empty.

Also alphabetizes methods, renames `make` to `bounded`, and adds Scaladoc.

I noticed that `offerAll` did not seem to respect the capacity of the queue. I changed it to retry if there is not sufficient capacity in the queue for all the elements in the collection, which seemed to be the only sensible way to do it transactionally, but let me know if you have a different thought.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TQueue.scala
+++ b/core/shared/src/main/scala/zio/stm/TQueue.scala
@@ -55,11 +55,11 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
     }
 
   /**
-    * Offers each of the elements in the specified collection to the queue up to
-    * the maximum capacity of the queue, retrying if there is not capacity in
-    * the queue for all of these elements. Returns any remaining elements in the
-    * specified collection.
-    */
+   * Offers each of the elements in the specified collection to the queue up to
+   * the maximum capacity of the queue, retrying if there is not capacity in
+   * the queue for all of these elements. Returns any remaining elements in the
+   * specified collection.
+   */
   def offerAll(as: Iterable[A]): STM[Nothing, Iterable[A]] = {
     val (forQueue, remaining) = as.splitAt(capacity)
     ref.get.flatMap { q =>
--- a/core/shared/src/main/scala/zio/stm/TQueue.scala
+++ b/core/shared/src/main/scala/zio/stm/TQueue.scala
@@ -55,28 +55,17 @@ final class TQueue[A] private (val capacity: Int, ref: TRef[ScalaQueue[A]]) {
     }
 
   /**
-   * Offers each of the elements in the specified collection to the queue,
-   * retrying if there is not capacity in the queue for all of the elements in
-   * the collection.
-   */
-  def offerAll(as: Iterable[A]): STM[Nothing, Unit] =
-    if (as.size > capacity)
-      STM.dieMessage("collection size greater than queue capacity")
-    else
-      ref.get.flatMap { q =>
-        if (as.size <= capacity - q.length) ref.update(_.enqueue(as.toList)).unit
-        else STM.retry
-      }
-
-  /**
-   * Offers each of the elements in the specified collection to the queue up to
-   * the maximum capacity of the queue, retrying if there is not capacity in
-   * the queue for all of these elements. Returns any remaining elements in the
-   * specified collection.
-   */
-  def offerMax(as: Iterable[A]): STM[Nothing, Iterable[A]] = {
+    * Offers each of the elements in the specified collection to the queue up to
+    * the maximum capacity of the queue, retrying if there is not capacity in
+    * the queue for all of these elements. Returns any remaining elements in the
+    * specified collection.
+    */
+  def offerAll(as: Iterable[A]): STM[Nothing, Iterable[A]] = {
     val (forQueue, remaining) = as.splitAt(capacity)
-    offerAll(forQueue) *> STM.succeed(remaining)
+    ref.get.flatMap { q =>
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -18,77 +18,83 @@ package zio.stm
...
+
+  /**
+   * Offers each of the elements in the specified collection to the queue,
+   * retrying if there is not capacity in the queue for all of the elements in
+   * the collection.
+   */
+  def offerAll(as: Iterable[A]): STM[Nothing, Unit] =
+    ref.get.flatMap { q =>
+      if (as.size <= capacity - q.length) ref.update(_.enqueue(as.toList)).unit
+      else STM.retry
```
jdegoes: The other alternative would be placing as many as we can now, and then offeringAll the rest. I think this would result in less surprising behavior (imaging putting 101 elements into a queue with capacity 100—it would never succeed unless you put some of them in and wait to put the rest in).

core/shared/src/main/scala/zio/stm/TQueue.scala
```diff
@@ -18,77 +18,83 @@ package zio.stm
...
       }
     )
 
   /**
-   * Checks if the Queue is empty
+   * Views the next element in the queue without removing it, returning `None`
+   * if the queue is empty.
    */
-  def isEmpty: STM[Nothing, Boolean] = ref.get.map(_.isEmpty)
+  def peekOption: STM[Nothing, Option[A]] =
```
jdegoes: This is useful, and also equivalent to, `peek.map(Some(_)) orElse STM.succeed(None)`.

core/shared/src/main/scala/zio/stm/TQueue.scala
adamgfraser: I thought about the same thing but struggled with how to do it in a transactional context. Here is an implementation that tries to do what you describe and a test for it:

```scala
def offerAll(as: Iterable[A]): STM[Nothing, Unit] =
  if (as.isEmpty)
    STM.unit
  else
    ref.get.flatMap { q =>
      val (forQueue, remaining) = as.splitAt(capacity - q.size)
      if (forQueue.isEmpty) offerAll(remaining)
      else ref.set(q.enqueue(forQueue.toList)) *> offerAll(remaining)
    }

testM("offerAll respects capacity") {
  for {
    tq     <- TQueue.bounded[Int](3).commit
    _      <- tq.offerAll(List(1, 2, 3)).commit.fork
    first  <- STM.collectAll(List(tq.take, tq.take, tq.take)).commit
    second <- STM.collectAll(List(tq.take, tq.take)).commit
  } yield assert(first)(equalTo(List(1, 2, 3))) &&
    assert(second)(equalTo(List(4, 5)))
},
```

The challenge is what we do if the queue is at capacity. In the current version, we immediately try to offer again, which results in us just spinning in a tight loop. If instead we retry at that point we risk not being productive, because if we try to put 101 elements into a queue with capacity 100 we're always going to hit a point where the queue doesn't have capacity at which point we will have to retry and start the whole thing over again.

Conceptually it seems like we need a way to suspend until one of the transactional variables has changed without retrying the transaction. Is there a way to do that or is there something I'm missing here?

core/shared/src/main/scala/zio/stm/TQueue.scala
adamgfraser: Uh oh, our code is too compositional! Better switch to actors! 😉

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -845,10 +845,7 @@ object ZSTM {
    * transactional effect that produces a list of values.
    */
   def collectAll[E, A](i: Iterable[STM[E, A]]): STM[E, List[A]] =
```
adamgfraser: I believe there was a bug here where effects were evaluated in the wrong order. I added a fix for it and a test.

core/shared/src/main/scala/zio/stm/TQueue.scala
jdegoes: Ah right, the retry will rollback and wait... Possibly we could "fix" it like this:

```scala
def offerAll(as: Iterable[A]): STM[Nothing, Iterable[A]]
```

Then the user can decide what to do with the excess.

Though possibly `offerMax` is a better name for such a thing.

Perhaps for `offerAll`, we can just `STM.die` if the user tries to do something impossible, like offering 10 in a 9 element queue?

adamgfraser: Yes, done in the latest version. Is it possible to implement the "suspend for a change but don't retry" functionality?

jdegoes: Not in STM. That would violate transactionality. Of course we could just return IO for this method. Is that better than dying?

adamgfraser: I think people are going to want to compose multiple offers and are just going to do it with something like `collectAll` if we don't provide something which will go back to hanging if the number of elements is greater than the capacity. Maybe we want to just provide `offerMax` and maybe rename it to `offerAll`?

jdegoes: Ok, sounds good!

adamgfraser: All set.

jdegoes:
💪

PR #2655, opened 2020-01-19
Reenable Test On Scala 2.13

adamgfraser:
After this all tests on Scala 2.13 should be running normally.


jdegoes:
@adamgfraser What fixed it??

adamgfraser:
I'm honestly not sure. I noticed that the tests were running now. Trying to go back and figure out what PR fixed it but the solution is as mysterious as the original cause right now.

PR #2660, opened 2020-01-19
ZIO Test: Implement Spy

adamgfraser:
Initial implementation of a `Spy` for stateful testing. A spy wraps an existing live service, maintaining some internal state `S` and allowing assertions to be made based on that internal state and values returned by the live service. Spies will typically delegate to the underlying live service for the actual implementation but can update state based on method calls and values returned and make assertions based on those values.


test-tests/shared/src/test/scala/zio/test/SpySpec.scala
```diff
@@ -0,0 +1,76 @@
...
+    case object Reset     extends Command[Unit]
+
+    val commands = Gen.elements(Increment, Decrement, Get, Reset)
+  }
+
+  val spy: ZLayer[Counter, Nothing, Counter with Spy] =
+    Spy.live[Int, Counter.Service](0) { (counter, state, result) =>
+      new Counter.Service {
+        def increment: UIO[Unit] =
+          counter.increment *> state.update(_ + 1).unit
```
jdegoes: There is a bit of boilerplate here. The ideal way to do this might be something like:

```scala
PartialFunction[(Ref[State], Method[A, B], A, B), UIO[Unit]]
```

That way you get to see the method, the input, the output, and then modify the state as per appropriate to your spy.

Now this can't exactly be written today. We have `Mockable` to go from `Mock => A`, but we don't have something like `Unmockable` (`Spyable`) to go the other way: to deconstruct one or more calls on `A` into the individual methods, with inputs and outputs. But if we had such a thing and if the macros project could generate them automatically, then we might be able to eliminate all the boilerplate, and make it easy to create many different spies for the same interface.

adamgfraser: I revised along the lines you suggested. There is still some boilerplate in implementing the `Spyable` interface but now you only have to do it once for your service and can create many different spies based on that. Also agree it would be very amenable to macros.

I ran into some issues with the input `I` parameter in `Method` being invariant and realized we could make it contravariant by moving a couple of the methods that make assertions based on the input to an implicit syntax class. I think that makes more sense as it reflects a `Method[M, I, A]` modeling a `Function[I, A]`. I think we may also want to consider changing the way we model functions that don't take any inputs from `Method[M, Unit, A]` to `Method[M, Any, A]`. A method that doesn't require any inputs can be run with any input at all and while the unit value is thus one value that would be sufficient there is nothing special about that. This is also more consistent with our treatment of the environment and seems to work better when we are recovering type information by pattern matching against methods.

jdegoes: I agree with you. We should make it `Any` and not `Unit`. 💯 💪

test-tests/shared/src/test/scala/zio/test/mock/SpySpec.scala
```diff
@@ -0,0 +1,95 @@
...
+              case Command.Reset     => environment.get.reset
+            }
+        }
+    }
+
+  val spy: ZLayer[Counter, Nothing, Counter with Spy] =
+    Spy.live[Int, Counter.Service](0) {
+      case (state, result, Invocation(method, _, output)) =>
+        method match {
+          case Command.Increment => state.update(_ + 1).unit
```
jdegoes: This looks super nice now.

test/shared/src/main/scala/zio/test/mock/Spyable.scala
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * `Spyable[A]` models the capability to spy on a service of type `A`.
+ * Implementations must define both a method `environment` to genereate a live
+ * environment given a mock and a method `mock` to generate a mock given a live
+ * environment. Given these definitions, it is possible to convert a service to
+ * a mock, see and modify state based on method calls, inputs, and outputs, and
+ * then convert the mock back to a live environment to spy on the service.
+ */
+trait Spyable[A] extends Mockable[A] { self =>
```
jdegoes: Are there conditions in which we would want to `spy` but not `mock`? If so, we should severe the relationship between Spyable and Mockable, and make them independent.

test/shared/src/main/scala/zio/test/mock/Spyable.scala
```diff
@@ -0,0 +1,56 @@
...
+   * Constructs a mock service from a live environment.
+   */
+  def mock(environment: Has[A]): Mock
+
+  /**
+   * Updates a live environment by spying on it using the specified partial
+   * function, which has the ability to see method calls, inputs, and outputs
+   * and update state based on them.
+   */
+  final def spy(environment: Has[A])(f: PartialFunction[Invocation[A, _, _], UIO[Unit]]): Has[A] = {
```
jdegoes: I like the simpler sig.

test/shared/src/main/scala/zio/test/mock/package.scala
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Constructs a spy that wraps an existing live service. The spy maintains
+     * some internal state `S` with the specified initial value. The update
+     * function `f` takes an existing live service and returns an updated
+     * service that has access to the live service as well as a `Ref` with the
+     * internal spy state and a `Ref` to hold the result of assertions made
+     * about the internal state and values returned.
+     */
+    def live[S, A: Tagged](z: S)(
+      f: PartialFunction[(Ref[S], Ref[TestResult], Invocation[A, _, _]), UIO[Unit]]
```
jdegoes: It occurs to me we can push the `Ref[S]` out of here if we want and make the user create a ref, if desired. Because basically all we're doing is passing along state the user could construct themselves before they call `live`. 

I'm not sure if this is a good idea, however. But it's more "orthogonal". 

Also: on `Ref[TestResult]`, what if we just made it return `UIO[TestResult]`? Would simplify and make easier to understand.

test/shared/src/main/scala/zio/test/mock/Spyable.scala
adamgfraser: The issue is we need to be able to `mock` to implement `spy`. Because to `spy` we need to be able to turn a live service into a mock with `mock`, modify the mock with the specified partial function, and then turn the modified mock back into a live service with `environment` from `Mockable`.

What we could potentially do is have something like:

```scala
trait Mockable {
  def environment[A](mock: Mock): Has[A]
}

trait Unmockable {
  def mock[A](environment: Has[A]): Mock
}

trait Spyable extends Mockable with Unmockable {
  def spy = ???
}
```

That is definitely more orthogonal but it also adds more complexity so I think the question is whether the complexity is worth it, and in particular how much use you can get out of the `Unmockable` interface itself or whatever we call it. With that interface you can turn a live service into a mock and modify the mock, but you're stuck with just having a mock. So if you have a bunch of ready made `Method` values you could feed those to it and get results back, but if you wanted to interact with other systems they are going to be expecting the service being mocked and not the mock type.

jdegoes: You're right, I'd leave it like it is! 👍

test/shared/src/main/scala/zio/test/mock/package.scala
adamgfraser: Could definitely push the `Ref` out if we wanted to. Most of the heavy lifting here is done by `Spyable` that doesn't have the `Ref` for exactly that reason and you can see how it is easy to add in. Another thing we could do is have two variants, a more "fundamental" one that doesn't have the `Ref` and one implemented in terms of it for the "everyday" use case that gives you a `Ref`. That way if you wanted to do something else like use multiple `Ref`s or use a `RefM` it would be easy to do that. We could even have a couple of common baked in implementations (e.g. `Ref[S]`, `RefM[S]`).

I'm a little worried about changing the signature to return `UIO[TestResult]` because that forces the user to make an assertion about every state transition where they want to do anything, whereas in many cases they may just want to update state. For example in the prototype asserts are only being made after `get` calls, all the other calls only update state.

adamgfraser: Thinking about it more I am coming around to your point of view that maybe we just simplify the signature here, potentially to be the same as the one in `Spyable` and then maybe we don't even need `Spy` separate from `Spyable`. The more general `Spyable` signature already gives you the ability to maintain state by creating a `Ref` or other similar type and passing it in. This also gives the user the ability to determine how they want to make assertions.

One advantage of this is right now we have a little bit of awkwardness where the user can maintain state with the `Ref[S]` but can't actually access the `S` outside the function since we aren't parameterizing the `Spy` on `S` because services need to be monomorphic. That also creates the need for the `TestResult` because we need something monomorphic the user can always extract. If we just let the user pass in whatever they want they would have access to that state outside the function and could make whatever assertions they wanted about it.

The downside is the user potentially needs to think a little more about what to do with this function (e.g. you create some state and then you use this effect to update it) versus us kind of serving a `Ref` or two up for them.

adamgfraser: And then we could have more specific combinators implemented in terms of that general one. Like we could have one that takes a live service and gives you back a new service along with a `Ref` which will hold a list of all invocations that have been made on the service up to whenever you access the `Ref`, etc...

jdegoes: How about `UIO[Option[TestResult]]`? Or do we have a `TestResult` the equivalent of `noOp` (looks like not).

jdegoes: @adamgfraser Right!!

test-tests/shared/src/test/scala/zio/test/mock/SpySpec.scala
```diff
@@ -0,0 +1,95 @@
...
+            state.get.flatMap { expected =>
+              result.update(_ && assert(output)(equalTo(expected))).unit
+            }
+          case Command.Reset => state.set(0)
+        }
+    }
+
+  def spec = suite("SpySpec") {
+    testM("counter example") {
+      checkM((Gen.listOf(Command.commands))) { commands =>
```
ioleo: I love this use case :heart:

adamgfraser:
Made some changes to further simplify this. `Spy` goes away so we just have the `Spyable` trait and methods on its companion object. The key method becomes `spyWithRef` which takes a service and returns a new "spied on" version of the service along with a `Ref` that will be updated with all method calls to the service and associated inputs and outputs. This achieves a better decoupling of collection of data from analysis of that data because the spy is solely responsible for returning the data and then the user can do whatever they want with it.

PR #2662, opened 2020-01-20
ZIO Test: Fix Bug in Spec#provideManagedShared

adamgfraser:
Fixes a bug where test level failures could interfere with `provideManagedShared`. Thanks to @dkarlinsky for reporting!

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -343,12 +343,14 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
   final def provideSomeManagedShared[R0, E1 >: E](
     managed: ZManaged[R0, E1, R]
   )(implicit ev: NeedsEnv[R]): Spec[R0, E1, L, T] = {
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -338,7 +348,7 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
         case SuiteCase(label, specs, exec) =>
           specs.flatMap(ZIO.foreach(_)(loop(r))).map(z => Spec.suite(label, ZIO.succeed(z.toVector), exec))
         case TestCase(label, test) =>
-          test.map(t => Spec.test(label, ZIO.succeed(t)))
+          test.run.map(exit => Spec.test(label, ZIO.done(exit)))
```
jdegoes: Could this affect any of the other `provide*` variants?

adamgfraser: This only impacts `shared` variants and all `shared` variants are implemented in terms of `provideManagedShared` so we just need to fix it here.

PR #2666, opened 2020-01-20
Buffer signal bug

vasilmkd:

PR #2674, opened 2020-01-21
Make ZStream#flatMap safe to pull again

vasilmkd:
Found the culprit for #2666.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1289,7 +1289,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       as: Pull[R1, E1, A],
       finalizer: Ref[Exit[_, _] => URIO[R1, _]],
       currPull: Ref[Pull[R1, E1, B]]
-    ): ZIO[R1, Option[E1], B] = {
+    ): Pull[R1, E1, B] = {
       val pullOuter = ZIO.uninterruptibleMask { restore =>
         restore(as).flatMap { a =>
           (for {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1301,20 +1301,19 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
-            e
-          )
-        case None =>
-          (finalizer.get.flatMap(_(Exit.succeed(()))) *>
-            finalizer.set(_ => UIO.unit)).uninterruptible *>
-            pullOuter *>
-            go(as, finalizer, currPull)
+      currPull.get.flatten.catchAllCause { c =>
+        Cause.sequenceCauseOption(c) match {
+          case Some(e) => Pull.halt(e)
```
iravid: Hmm why did we eagerly execute the finalizer here?

vasilmkd: The only thing I didn't think completely through in this PR because I was too eager to put it out is the fact that we lose the error signal through the `Exit`. What do you think we should do?

iravid: Ok, convinvced myself it wasn't needed 👍

vasilmkd: `finalizerRef` for the win!

iravid: Good point regarding the error. Let's add another Ref[Option[Cause[E]]] that is set on error and used in the finalizer?

vasilmkd: No need, `finalizerRef` takes care of it.

vasilmkd: How can we craft a test that confirms this?

vasilmkd: > Good point regarding the error. Let's add another Ref[Option[Cause[E]]] that is set on error and used in the finalizer?

Also on which error? What if there are more than one?

iravid: The `Exit[Any, Any]` required in the finalizer is the one that results from the `ZManaged#use` block. It's a bit hard to see it here, but this should result from the terminal operation consuming the stream.

So the Exit value here really depends if the error was allowed to fail the use block or not. Which is to say - if there were no recovery operators downstream, we should see the error propagate back to the `finalizerRef`.

You can test this by flatMapping elements in a stream to a stream that differentiates between success/failure in its finalizer.

vasilmkd: Exactly, this was my intuition too. I'll add the test.

vasilmkd:
@iravid should be good to go now.

vasilmkd:
Wait until you see what this unlocks. No more racing in `bufferSignal`. :grin:

PR #2677, opened 2020-01-21
ZIO Test: Fix Bug in TestClock

adamgfraser:
Following up on discussion on Discord, `TestClock` does not currently handle the following case correctly:

```scala
testM("TestClock interacts correctly with Scheduled.fixed") {
  for {
    latch     <- Promise.make[Nothing, Unit]
    ref       <- Ref.make(3)
    countdown = ref.update(_ - 1).flatMap(n => latch.succeed(()).when(n == 0))
    _         <- countdown.repeat(Schedule.fixed(2.seconds)).delay(1.second).fork
    _         <- TestClock.adjust(5.seconds)
    _         <- latch.await
  } yield assertCompletes
}
```

This test will timeout instead of passing. The issue is that `Schedule.fixed` calls `nanoTime` internally and uses that as an input to determine how long to sleep for. Since `nanoTime` is already set to 5 seconds by the `TestClock.adjust` call, that leads to corruption internal state in the schedule.

To fix this, we need to slightly change the contract of the `TestClock`. Methods such as `nanoTime` return the fiber time instead of the wall clock time, assuring that calls to `nanoTime` and other methods return a correct time that reflects how much time has actually passed on that fiber. Methods like `setTime` and `adjust` then are solely responsible for modifying the "wall clock time", which determines when pending sleep effects run.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
@@ -95,7 +95,7 @@ object ClockSpec extends ZIOBaseSpec {
       for {
         expected <- UIO.effectTotal(OffsetDateTime.now(ZoneId.of("UTC+9")))
         _        <- setDateTime(expected)
-        _        <- sleep(Duration(expected.toInstant.toEpochMilli, TimeUnit.MILLISECONDS))
+        _        <- sleep(Duration.fromInstant(expected.toInstant))
         actual   <- clock.currentDateTime
       } yield assert(actual.toInstant.toEpochMilli)(equalTo(expected.toInstant.toEpochMilli))
     },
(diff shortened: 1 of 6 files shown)


test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
```diff
@@ -92,24 +95,28 @@ object ClockSpec extends ZIOBaseSpec {
       for {
         expected <- UIO.effectTotal(OffsetDateTime.now(ZoneId.of("UTC+9")))
         _        <- setDateTime(expected)
+        _        <- sleep(Duration(expected.toInstant.toEpochMilli, TimeUnit.MILLISECONDS))
```
jdegoes: Worth adding `Duration.fromInstant` here?

adamgfraser: Yes. Done now.

francisdb:
Thanks for looking into this!

jdegoes:
💪

PR #2680, opened 2020-01-22
Rename ZIO#ensuringExit to ZIO#onExit

adamgfraser:
Minor follow up to confirm name to name of existing combinator with same functionality on `ZManaged`.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -792,8 +792,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     )(_ => self)
 
   /**
-   * Ensures that a clean functions runs, whether this effect succeeds, fails,
-   * or is interrupted.
+   * Ensures that a cleanup functions runs, whether this effect succeeds,
+   * fails, or is interrupted.
    */
   final def onExit[R1 <: R](cleanup: Exit[E, A] => URIO[R1, Any]): ZIO[R1, E, A] =
     ZIO.bracketExit(ZIO.unit)((_, exit: Exit[E, A]) => cleanup(exit))(_ => self)
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -390,6 +390,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
    * specified `finalizer` is guaranteed to begin execution, whether this effect
    * succeeds, fails, or is interrupted.
    *
+   * For use cases that need access to the effect's result, see [[ZIO#onExit]].
+   *
    * Finalizers offer very powerful guarantees, but they are low-level, and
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -796,6 +791,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
         }
     )(_ => self)
 
+  /**
+   * Ensures that a clean functions runs, whether this effect succeeds, fails,
```
vasilmkd: Typo: the cleanup function

vasilmkd:
Wouldn't `ZStream(self.process.onExit(cleanup))` do the trick, regarding the stream implementation. I think it should be ok. Also, do we need an `onExitFirst` on ZIO and ZStream too?

adamgfraser:
So that's what I did initially. The issue is that then our cleanup function need to be `Exit[E, ZIO[R, Option[E], A]]`.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Thanks! Fixed now.

vasilmkd:
I'll look into it.

vasilmkd:
This is good to go, unless you want to add the `onExitFirst` too.

adamgfraser:
I don't think we have the ability to implement `ZIO#onExitFirst` because we're not reifying finalizers the same way we do with `ZManaged`.

vasilmkd:
@iravid @adamgfraser Regarding the `onExit` on `ZStream`. There is no possibility for this combinator on `ZStream` because there is no way to exit the scope of the stream with an error, as we have fixed the error of the `ZStream` to be `E =:= Nothing`.

iravid:
That doesn’t sound right - we should be able to inspect whether the stream ended with success/error.

vasilmkd:
@iravid Not with the current encoding. Unless we have two error channels for the ZManaged, one for entering the scope and one for exiting it.

vasilmkd:
Maybe we need to rethink our scope error...

vasilmkd:
You can check for yourself, even Metals autocomplete confirms it, `self.process.onExit` accepts a `fn: Exit[Nothing, Pull[...]] => ZIO[...]`.

iravid:
@vasilmkd shouldn’t be implemented through the ZManaged layer

vasilmkd:
Doesn't that mean that the exit function gets evaluated on each pull?

vasilmkd:
Actually I got an idea. We could define a custom layer for that that accepts a `cleanup: Exit[E, Unit] => ZIO[...]` or in the future `cleanup: Exit[E, B] => ZIO[...]`. However, that way we'd be reimplementing the finalizer logic from ZManaged?

vasilmkd:
Ah, it's been a taxing day with exams and work. I literally wrote a test for this yesterday. I'll add the combinator now. 😓

PR #2681, opened 2020-01-22
Adds ZManaged missing provide methods

saraiva132:
Using ManagedApp and kinda feel it would be a nice addition. Updated the documentation to be similar to ZIO methods. 

Adds `provideLayer`, `provideManaged`, `provideManagedSome` and `provideSomeM` to `ZManaged`

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -612,17 +612,36 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
    * Provides the `ZManaged` effect with its required environment, which eliminates
    * its dependency on `R`.
    */
-  def provide(r: R)(implicit ev: NeedsEnv[R]): ZManaged[Any, E, A] =
+  def provide(r: R)(implicit ev: NeedsEnv[R]): Managed[E, A] =
     provideSome(_ => r)
+
+  /**
+   * Provides a layer to the `ZManaged`, which translates it to another level.
+   */
+  def provideLayer[E1 >: E, R0, R1 <: Has[_]](layer: ZLayer[R0, E1, R1])(implicit ev: R1 <:< R): ZManaged[R0, E1, A] =
+    layer.translate.value.flatMap(self.provide(_))
+
+  /**
+   * An effectual version of `provide`, useful when the act of provision
+   * requires an effect.
+   */
+  def provideM[E1 >: E](f: ZIO[Any, E1, R])(implicit ev: NeedsEnv[R]): Managed[E1, A] =
+    provideManaged(f.toManaged_)
+
+  /**
+   * Uses the given Managed[E1, R] to the environment required to run this managed effect,
+   * leaving no outstanding environments and returning Managed[E1, A]
+   */
+  def provideManaged[E1 >: E](r0: Managed[E1, R])(implicit ev: NeedsEnv[R]): Managed[E1, A] = provideSomeManaged(r0)
 
   /**
    * Provides some of the environment required to run this effect,
    * leaving the remainder `R0`.
    *
    * {{{
-   * val effect: ZManaged[Console with Logging, Nothing, Unit] = ???
+   * val managed: ZManaged[Console with Logging, Nothing, Unit] = ???
    *
-   * effect.provideSome[Console](env =>
+   * managed.provideSome[Console](env =>
    *   new Console with Logging {
    *     val console = env.console
    *     val logging = new Logging.Service[Any] {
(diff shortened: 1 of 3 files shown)


adamgfraser:
I think this is a great addition. While we're at it do we want to add the rest of the usual cast of characters of `provide` variants? With this PR we would have `provideSomeM` but not `provideM` so I would think we should at least provide that. There is also `provideManaged` and `provideSomeManaged`. Not just if there would be any special issues with providing a managing resource to another managed. There is also the new `provideLayer`.

saraiva132:
Since reservation release does not accept an effect that can throw errors. I had to change things a bit. Before adding more combinators, what do you guys think. @adamgfraser @mschuwalow

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -618,10 +618,42 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
+   * {{{
+   * val effect: ZManaged[Console with Logging, Nothing, Unit] = ???
+   *
+   * val r0: URIO[Console, Console with Logging] = ???
+   *
+   * effect.provideSomeM(r0)
+   * }}}
+   */
+  def provideSomeM[R0, E1 >: E](
+    f: ZIO[R0, E1, R]
```
adamgfraser: I think we should say that the effect cannot fail here.

saraiva132: I thought about that. I thought about going that route. I was wondering if it was too restrictive. But then again, accepting an effect that can fail, with a die hidden under the hood for the release, might lead to even yet more unexpected results, plus, it is binding to a subtype of throwable.

adamgfraser: Actually there is some weirdness here because we are running the effect twice. Why don't we implement this as `f.toManaged_.flatMap(self.provide)`? Then we can also remove the constraint on the error type.

saraiva132:
Today I was learning about `ZLayer` and `Has` and I was wondering how this translate mechanism actually works. `provideLayer` compiles without it which kinda scares me. I might be missing something but by abstracting this behaviour shared by both `provideLayer` into this translate method I could not make it work with by passing the type explicitly  or making the type constraints I wanted . Most likely I am not making any sense in the first place but I would like a second opinion. Thanks

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: Looks great now! Very minor suggestion, maybe call the parameter `r0` instead of `f` for consistency with same method on ZIO?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -612,16 +612,80 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
+   * Provides a layer to the `ZManaged`, which translates it to another level.
+   */
+  def provideLayer[E1 >: E, R0, R1 <: Has[_]](layer: ZLayer[R0, E1, R1])(implicit ev: R1 <:< R): ZManaged[R0, E1, A] =
+    layer.translate.value.flatMap(self.provide(_))
+
+  /**
+   * An effectual version of `provide`, useful when the act of provision
+   * requires an effect.
+   */
+  def provideM[E1 >: E](f: ZIO[Any, E1, R])(implicit ev: NeedsEnv[R]): Managed[E1, A] =
```
adamgfraser: Maybe call parameter `r` instead of `f`?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -612,16 +612,80 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
    */
-  def provide(r: R)(implicit ev: NeedsEnv[R]): ZManaged[Any, E, A] =
+  def provide(r: R)(implicit ev: NeedsEnv[R]): Managed[E, A] =
     provideSome(_ => r)
 
+  /**
+   * Provides a layer to the `ZManaged`, which translates it to another level.
+   */
+  def provideLayer[E1 >: E, R0, R1 <: Has[_]](layer: ZLayer[R0, E1, R1])(implicit ev: R1 <:< R): ZManaged[R0, E1, A] =
+    layer.translate.value.flatMap(self.provide(_))
```
adamgfraser: How about `provideSomeManaged(layer.value.map(ev))`? Then we can revert the changes to `ZIO.scala` and `ZLayer.scala`.

saraiva132: Yeah I do not mind removing the changes in `ZLayer.scala`, was trying to extract common behaviour but although I understood ZLayer that might been unnecessary. I still have the impression provideLayer for `ZIO` and `ZManaged` should look quite similar. Are you positive `provideSomeManaged(layer.value.map(ev))` suffices ?

saraiva132: Looking at it @adamgfraser, your proposal looks quite clean. I was just wondering if this translation from one level to another could be moved to `ZLayer` but TBH I am fine with leaving that aside.

saraiva132: Also, I am unsure whether your proposal would work. Should it not look like something like this, for us to have proper R translation? 
``` 
for {
      r0 <- ZManaged.environment[R0]
      r1 <- layer.value.provide(r0)
} yield ev(r1)).flatMap(self.provide(_)
```

adamgfraser: That isn't necessary because calling `value` on a `ZLayer[R0, E, R]` returns a `ZManaged[R0, E, R]` so it already expresses the ability to partially provide the environment. I checked that this compiles. We could add a test for it if you want.

saraiva132:
@adamgfraser I was not strong enough to write what seemed like useless tests xD

PR #2682, opened 2020-01-22
Add ZStream#onExit and ZStream#onExitFirst

vasilmkd:
Sorry for the confusion earlier.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1994,30 +1994,24 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * or is interrupted, after this stream's finalizers run.
    */
   final def onExit[R1 <: R](cleanup: Exit[E, Any] => ZIO[R1, Nothing, Any]): ZStream[R1, E, A] =
-    ZStream.managed {
-      for {
-        _ <- ZManaged.finalizerRef {
-              case success @ Exit.Success(_) => cleanup(success)
-              case failure @ Exit.Failure(_) => cleanup(failure.asInstanceOf[Exit.Failure[E]])
-            }
-        pull <- self.process
-      } yield pull
-    }.flatMap(ZStream.repeatEffectOption)
+    ZStream {
+      ZManaged.finalizerRef {
+        case success @ Exit.Success(_) => cleanup(success)
+        case failure @ Exit.Failure(_) => cleanup(failure.asInstanceOf[Exit.Failure[E]])
+      } *> self.process
+    }
 
   /**
    * Ensures that the cleanup function runs, whether this stream succeeds, fails,
    * or is interrupted, before this stream's finalizers run.
    */
   final def onExitFirst[R1 <: R](cleanup: Exit[E, Any] => ZIO[R1, Nothing, Any]): ZStream[R1, E, A] =
-    ZStream.managed {
-      for {
-        pull <- self.process
-        _ <- ZManaged.finalizerRef {
-              case success @ Exit.Success(_) => cleanup(success)
-              case failure @ Exit.Failure(_) => cleanup(failure.asInstanceOf[Exit.Failure[E]])
-            }
-      } yield pull
-    }.flatMap(ZStream.repeatEffectOption)
+    ZStream {
+      self.process <* ZManaged.finalizerRef {
+        case success @ Exit.Success(_) => cleanup(success)
+        case failure @ Exit.Failure(_) => cleanup(failure.asInstanceOf[Exit.Failure[E]])
+      }
+    }
 
   /**
    * Switches to the provided stream in case this one fails with a typed error.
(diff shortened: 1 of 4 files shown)


vasilmkd:
@adamgfraser I think the same thing I mentioned yesterday with the hanging specs is happening here too, after the last commit. @iravid

vasilmkd:
@mschuwalow there is a problem with the StreamSpec.Stream.broadcast.Unsubscribe. Could you take a look please? :pray:

iravid:
Hi, please hold on merging until I review.

vasilmkd:
No worries, all reviews are dismissed.

vasilmkd:
I'm still investigating the `broadcast` issue.

mschuwalow:
@vasilmkd are the specs still hanging? Couldn't reproduce locally from your branch

vasilmkd:
Remove the ignore annotation on the test.

vasilmkd:
That's as far as I got with the debugging.

mschuwalow:
Ah sorry, I removed the comment as soon as I saw the ignore

mschuwalow:
I'll take a look. I have a quite strong suspicion what the cause is

vasilmkd:
I'm all ears. Let me know how I can help. I don't quite know that part of the code that well.

mschuwalow:
Ok so it seems to me that the issue is that now the finalizer will only be run if the stream has attempted to pull at least once, as the scope of the zmanaged will only be entered then. The previous behavior was to have the finalizer on the outer zmanaged, ensuring that it runs whether the stream is pulled from or not.

What is the desired behavior for this combinator?

vasilmkd:
That's true, that's the semantics of `ZStream.managed`. Great point. I'll take a look.

vasilmkd:
@mschuwalow Fixed and working beautifully. Thanks.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1985,6 +1989,30 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
     }
   }
 
+  /**
+   * Ensures that the cleanup function runs, whether this stream succeeds, fails,
+   * or is interrupted, after this stream's finalizers run.
+   */
+  final def onExit[R1 <: R](cleanup: Exit[E, Any] => ZIO[R1, Nothing, Any]): ZStream[R1, E, A] =
```
iravid: In the current formulation, this needs to be `Exit[Any, Unit]`. The reason is that `ZManaged.finalizerRef` is actually fed the `Exit` value resulting from `ZManaged#use`. `use` is universally quantified on `A`, therefore you've correctly set `A = Any` here. However, the `E` in the `Exit` is `E1 >: E` - meaning it is any arbitrary supertype of `E`, which means you cannot safely cast it to `E` as is done below.

Another formulation exists: you can modify the returned `Pull` from `self.process` to store the `Exit` in a `Ref[Exit[E, A]]`. Then you can safely feed that to the cleanup function. What we'd be doing here is basically storing the last pulled value every time `self` is pulled.

vasilmkd: 1. I'm not sold that the success branch can have `Unit`, indeed because of the fact that it is universally quantified, which means I'd expect it to be `Any`. But check 3.
2. For the error type, you are correct, I incorrectly assumed that the only error that can happen is the stream pull error `E`.
3. Regarding the other formulation, I thought about it, and I see it like this. If the stream always succeeds and signals success on each pull, the only relevant `Exit` value is the last one, which is the `Exit` value for `Pull.end` (and for example, we explicitly signal this with `Unit` in `flatMap`, so maybe you are right, we should be more explicit about that in the Exit type). On the flip side, if there is an unhandled error, the scope exiting will start to unwind and finish with the `Any` type of the error, which is again, the only relevant `Exit` value. So this tells me that keeping and updating the `Ref[Exit[E, A]]` on each pull is redundant.

So my proposal is the following, we go either with an `Exit[Any, Any]` and give no expectation for the types of error or success, or go with `Exit[Any, Unit]` as you mentioned but manually succeed with a `Unit` in these combinators and anywhere else that is needed to manually run finalizers (e.g. `flatMap`). Honestly, I'd go with the second one if we do decide to eventually go the new encoding way of finishing with a value `B`, which we can switch to in the future. WDYT?

vasilmkd:
@iravid I'll wait for your decision on what the error and success types should be. Maybe it is better to have the error type of the stream, and this combinator can be seen as an alternative error handling combinator. I'm guessing it would be more useful in that regard as well, because the error of the `use` combinator on `ZManaged` can be handled by normal `ZIO` error handling.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: I didn't quite follow what you were saying with 3. I will simplify what I'm saying. There are two possible interpretations for this combinator:
- The cleanup is provided with the `Exit` of processing the stream - that is, the result of `foreach/run`. This is `Exit[Any, Any]` and can be written with the current formulation.
- The cleanup is provided with the `Exit` of the last pull from the stream. This is `Exit[E, A]`, and requires the Ref.

Which one is more useful?

jdegoes:
@vasilmkd @iravid 

This PR has been open for some time. I will close for now, but would like to figure out if and how this should make it into master. Please re-open when ready to proceed!

Thanks for your work on these new combinators. 🙏

PR #2683, opened 2020-01-22
Deprecate Traverse and Sequence

### LINKED ISSUE #2649: Should We Remove Traverse and Sequence Method Aliases?
The `sequence` and `traverse` method aliases feel like a holdover from a time when we had more traditional, less accessible method names (e.g. `redeem` and `absolve`). In other cases we have just gotten rid of the old names but we have kept these around, which creates a bit of an odd inconsistency. It is also not ideal to have two ways to do the exact same thing. And for readers of a codebase using ZIO it could also be confusing if they see `foreach` used in one place and `traverse` used somewhere else and wonder if or how these are different. Finally, it creates a bit of a maintenance version because every different data type that implements `foreach` should ideally implement `traverse` as an alias or users are going to have a less than ideal experience where sometimes this alias is available and sometimes it is not for arbitrary reasons.

I understand the goal is to provide a more familiar interface for users coming from a functional programming background, but these are going to be more sophisticated users anyway so asking them to learn a new name (especially one that is very logical) doesn't seem like too much to ask.

If there is agreement I will take the ticket.

adamgfraser:
Resolves #2649. I don't think we came to any full consensus but deprecating traverse and sequence seems to be the middle ground between doing nothing and deleting them entirely. It also allows people to continue to use them if they want while allowing everyone else to maintain consistency in their code bases.


saraiva132:
`CollectAll` makes perfect sense to me. But actually, I have always seen and got used to `foreach` ignoring the output so it always confused why `ZIO` takes a different turn and uses it in a way that, at least, I had never seen before. I understand that `map` is taken but I do not understand why `sequence` or `traverse` get this much negativity. Things are moving forward so meanings change so I guess I am still stuck to the times when I used `foreach` with `future`. Been a while.

adamgfraser:
Not trying to short circuit discussion on this, just put forward what I think the change would be if we decide to do something here. I personally think this does make sense but it is a collective decision.

Regarding `foreach` in particular, I think the rationale is that it performs an effect for each element in a collection. Also `foreach` in the standard library performs a side effect for each element in a collection, but in a purely functional library we don't want to perform side effects, so performing an effect for each element is the logical equivalent. But your mileage with that explanation may vary.

regiskuckaertz:
Would `mapAll`/`collectAll` appeal to everyone?

regiskuckaertz:
Actually scratch that. There’s enough evidence that foreach works. We agonise over small things for the hope of reaching perfect symmetry, but nothing beats concrete feedback.

jdegoes:
In code in the wild you often see:

```scala
_ <- ZIO.foreach(list) { a => doX(a) }
```

Because many times the result of each iteration is not needed (especially if it's `Unit`). It's as close to Scala's `foreach` as you can get in a pure world, with the added benefit of collecting return values if you care about them (many times you don't).

PR #2690, opened 2020-01-23
Improve docs example: depend on module using with

evis:
1. It's more concise using `with`
2. It's impossible to accidentally misspell name of service and then get surprising errors. E.g.:

```scala
trait NewsletterLive {
  val postmanService: Postman.Service[Any]
  ...
}
...
new NewsletterLive with new PostmanLive {} // val postman is missing!
```


adamgfraser:
@evis I'm going to close this as I think these proposed changes to the documentation have been superseded by the new environment encoding. If you would like to contribute to documentation for that please let me know as there is definitely plenty that needs to be done there.

Issue #2671, opened 2020-01-21
ZIO Test: no equivalence between `*>` and `flatMap` from Expectation


arjun-1:
While using ZIO Test, I found some peculiarity which I cannot wrap my head around :sweat_smile:. Using the `*>` operator on Expecation, I created the following working mock:
```
val mockMomaClient: Expectation[MomaClient, Nothing, Position] = 
  (MockMomaClient.getNodeUrl(equalTo(sessionId)) returns value(
    uri"localhost"
  )) *>
  (MockMomaClient.getFrameCount(anything) returns value(1)) *>
  (MockMomaClient.getPosition(anything) returns value(position))
```
Looking at the definition of `*>` :
```
final def *>[M1 <: M, E1 >: E, B](that: Expectation[M1, E1, B]): Expectation[M1, E1, B] = flatMap(_ => that)
```
This should be equivalent to:
```
val mockMomaClient: Expectation[MomaClient, Nothing, Position] =
  (MockMomaClient.getNodeUrl(equalTo(sessionId)) returns value(
    uri"localhost"
  ))
  .flatMap(_ => MockMomaClient.getFrameCount(anything) returns value(1))
  .flatMap(_ =>
    MockMomaClient.getPosition(anything) returns value(position)
  )
```
In my case however, it is not. A test that previously succeeds, now fails with the error:
```
info] - A main server route
[info]   - returns positions
[info]     Fiber failed.
[info]     ╥
[info]     ╠─An unchecked error was produced.
[info]     ║ java.lang.ClassCastException: class scala.runtime.BoxedUnit cannot be cast to class sttp.model.Uri (scala.runtime.BoxedUnit is in unnamed module of loader sbt.internal.classpath.ClassLoaderCache$Key$CachedClassLoader @171f9268; sttp.model.Uri is in unnamed module of loader sbt.internal.LayeredClassLoader @97a3b59)
[info]     ║    at zio.test.mock.Expectation.$anonfun$managedEnv$9(Expectation.scala:96)
[info]     ║    at zio.internal.FiberContext.evaluateNow(FiberContext.scala:340)
[info]     ║    at zio.internal.FiberContext.$anonfun$fork$2(FiberContext.scala:655)
[info]     ║    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
```

jdegoes:
These should be the same. Actually I'm surprised you can do `flatMap` on Expectation.

[...]

PR #2691, opened 2020-01-23
ZIO Mock: Less powerful expectation

ioleo:
Following up on discussion in #2671 we're dropping `flatMap` and `map` here, as it was only needed for `for-comprehension` syntax. Instead, for larger expectations the following syntax is encouraged:

```scala
import MockConsole._

val mock =
    (putStrLn(equalTo("foo")) returns unit) *>
    (putStrLn(equalTo("bar")) returns unit) *>
    (putStrLn(equalTo("baz")) returns unit) *>
    (getStrLn returns value("acme")) *>
    (putStrLn(equalTo("foo2")) returns unit) *>
    (putStrLn(equalTo("bar2")) returns unit) *>
    (putStrLn(equalTo("baz2")) returns unit) *>
    (getStrLn returns value("acme2"))
```

@jdegoes @adamgfraser

### DIFF
--- a/docs/howto/mock_services.md
+++ b/docs/howto/mock_services.md
@@ -190,7 +190,7 @@ Next we create the live version of the service with the implementation of the ca
 ``` scala mdoc
 import zio.console.Console
 
-val accountObserverLive: ZLayer[Console, Nothing, Has[AccountObserver.Service]] = 
+val accountObserverLive: ZLayer[Console, Nothing, Has[AccountObserver.Service]] =
   ZLayer.fromService[Console.Service, Nothing, Has[AccountObserver.Service]] { console =>
     Has(new AccountObserver.Service {
       def processEvent(event: AccountEvent): UIO[Unit] =
--- a/examples/shared/src/test/scala/zio/examples/test/MockingExampleSpec.scala
+++ b/examples/shared/src/test/scala/zio/examples/test/MockingExampleSpec.scala
@@ -89,6 +89,16 @@ object MockingExampleSpec extends DefaultRunnableSpec {
 
       val result = app.provideLayer(mockEnv)
       assertM(result)(equalTo(42))
+    },
+    testM("failure if unexpected calls") {
+      import MockRandom._
+
+      val app = random.nextInt *> random.nextLong
+      val mockEnv =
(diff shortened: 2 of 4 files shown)


adamgfraser:
This looks great. I wonder if we can go even further here. It looks like we're not even using the power of an applicative here but all we really need is monoidal composition, as illustrated by your removal of the `map` method and the fact that we don't have an implementation of `zipWith`.

Should we change `*>` to be `++`? Normally `*>` is reserved for implementations of `zipWith` whereas this seems more like just concatenation of expectations. This would also make it more accessible to a broader audience since anyone who has used the standard library will be familiar with `++`.

Also, should we hide more of the type parameters in `Method`? Right now `Method` is parameterized on `A` which is basically the result of the "last" method call in the sequence of expected calls. But is there any reason that one is particularly relevant? Or should we just hide that and say a method is just a collection of expected calls returning arbitrary types. Then this really would compose like a monoid.

ioleo:
Regarding the nameing change I have doubts about `++`. I think of it as collection concatenation, but not necessarily implying certain ordering. For `Expectation` the order does matter. It's basically `expect callA andThen callB`. So a right-pointing operator feels like a natural way to communicate this ordering `callA *> callB`.

ioleo:
As for type parameters, they are definately important for the `Call` case, but perhaps `E` and `A` could be removed from the general `Expectation` trait, I will look into that.

adamgfraser:
Definitely understand that perspective. I guess I draw a bit of a different conclusion from what you say. At this point, an `Expectation` is essentially just a list of calls. `++` doesn't necessarily imply a lack of ordering, for example `++` can be used with a `Vector` or an `Iterable` and generally implies a preservation of ordering unless the collection itself is something like a `Set` that explicitly does not have an ordering. I understand that `*>` is what we are using now and we find it familiar, but for people who are coming from another mocking framework they may have never used that operator before, whereas `++` is something almost everyone will be familiar with, and generally means combine two collections of things in left to right order.

I also think that pursuing this "collection of calls" conceptualization, potentially even more explicitly in the implementation, could let us provide a lot more methods out of the box (e.g. "how many calls are expected?", "return the second expected call", etc...).

PR #2695, opened 2020-01-23
Implement Any Layers

### LINKED ISSUE #2694: Add "any" layers for all services in ZEnv
Right now, every service has a `*.live` layer, e.g. `Clock.live`, which is a `ZLayer[In, Error, Out]`.

This is useful to wire together production implementations of services. However, sometimes, in wiring or providing part of an environment, a service needs something, and it doesn't matter if it's live or test or whatever.

To make this just a bit simpler, we can define the following layers inside the companion objects of services:

```scala
object Clock {
  val any: ZLayer[Clock, Nothing, Clock] = ZLayer.environment[Clock]
  ...
}
```

This way, in wiring, people can use `Clock.any ++ System.live ++ ...` (to be read as, _I am building a layer that uses any clock service and the live system service..._), without having to manually use `ZLayer.environment` to construct a "passthrough" layer.

The `*.any` layers can be added to all existing services, including test services; and possibly, searching for `ZLayer.environment` or `ZLayer.identity` will reveal places where this simplification can be utilized.

adamgfraser:
Resolves #2694. I looked for usages of `ZLayer.identity` and `ZLayer.environment` and didn't find many in our code base. @mschuwalow If there are other places where you think we could clean things up based on this let me know and I can include them.

### DIFF
--- a/core/js/src/main/scala/zio/PlatformSpecific.scala
+++ b/core/js/src/main/scala/zio/PlatformSpecific.scala
@@ -26,6 +26,8 @@ private[zio] trait PlatformSpecific {
   type ZEnv = Clock with Console with System with Random with Scheduler
 
   object ZEnv {
+    val any: ZLayer[ZEnv, Nothing, ZEnv] =
+      ZLayer.environment[ZEnv]
     val live: ZLayer.NoDeps[Nothing, ZEnv] =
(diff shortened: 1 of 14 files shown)


jdegoes:
@adamgfraser What do you think, additionally, about renaming `ZLayer.environment` into `ZLayer.requires`?

adamgfraser:
@jdegoes 👍 Will do it as part of this PR.

PR #2700, opened 2020-01-24
Make Effect Constructors Lazy

### LINKED ISSUE #2462: ZIO Core: Make all effect constructors lazy in all parameters

adamgfraser:
Resolves #2462.

Initial work on making all effect constructors on ZIO lazy. Lazy constructors are:

- `die`
- `dieMessage`
- `fail`
- `fromEither` - already lazy
- `fromFiber` - already lazy
- `fromOption` - already lazy
- `fromTry` - already lazy
- `halt`
- `interruptAs`
- `left`
- `lock`
- `provide`
- `require`
- `right`
- `sleep`
- `some`
- `succeed`
- `when` - already lazy
- `whenCase`

Also adds the following strict variants for very commonly used constructors:

- `dieNow`
- `doneNow`
- `failNow`
- `haltNow`
- `succeedNow`

Would be great to get some feedback on this and try to get alignment that we've got treatment of this issue right for ZIO and then I can roll out to other effect types.


iravid:
@adamgfraser Can you make sure all existing usage of `fail/halt/done/succeed/die/etc.` is migrated to the strict versions?

adamgfraser:
@iravid Yes.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -845,27 +845,48 @@ object ZSTM {
...
-      case TExit.Fail(e)    => STM.fail(e)
-      case TExit.Succeed(a) => STM.succeed(a)
+      case TExit.Fail(e)    => STM.failNow(e)
+      case TExit.Succeed(a) => STM.succeedNow(a)
+    }
+
+  /**
+   * Returns a value modelled on provided eagerly evaluated exit status.
+   */
+  def doneNow[E, A](exit: => TExit[E, A]): STM[E, A] =
```
jdegoes: This one is unchanged from `done`?

adamgfraser: Good catch! Will fix in next version.

jdegoes:
At a high-level, looks great! Makes me want to make the `*Now` variants package-private to `zio`, on grounds it's less to explain and less to go wrong... thoughts?

adamgfraser:
Interesting idea. I like the thought. There's definitely a risk that if we don't do something like this everyone will try to do the "most efficient" thing and we end up back in the same situation we are now except with longer names. The question would be whether there are cases where a library or a user wants that extra bit of efficiency and can't get it. But I'm not sure how realistic that is and in most cases there is some kind of escape hatch.

adamgfraser:
The other thing that could be nice about this is it could let us be a little more judicious when we provide `now` variants.

adamgfraser:
This should have everything now.

saraiva132:
@adamgfraser thanks a lot for picking this up since I was not able. You did it super fast as always and with perfection! 😊

jdegoes:
@adamgfraser with or without Now variants?

adamgfraser:
@jdegoes Current version needs a rebase and has `now` variants both exposed publicly and used internally. My plan was to resolve merge conflicts and make the `now` variants package private as you had suggested. Agreed?

jdegoes:
Agreed!

adamgfraser:
@jdegoes This is ready now.

jdegoes:
Awesome work!

PR #2703, opened 2020-01-24
ZStream#Buffer simpler

vasilmkd:


streams/shared/src/main/scala/zio/stream/Take.scala
```diff
@@ -61,7 +61,10 @@ object Take {
   case object End                            extends Take[Nothing, Nothing]
 
   def fromPull[R, E, A](pull: Pull[R, E, A]): ZIO[R, Nothing, Take[E, A]] =
-    pull.fold(_.fold[Take[E, A]](Take.End)(e => Take.Fail(Cause.fail(e))), Take.Value(_))
+    pull.foldCause(
+      Cause.sequenceCauseOption(_).fold[Take[E, A]](Take.End)(Take.Fail(_)),
```
iravid: Ouch, good catch here!

vasilmkd: It's all good.

PR #2704, opened 2020-01-24
ZStream#bufferDropping safe reimplementation

vasilmkd:
Reimplement ZStream#bufferDropping without racing effects and add tests for pull safety.


vasilmkd:
Not only that, but no other element is enqueued until the error has been dequeued.

PR #2715, opened 2020-01-24
Make Fiber a Sum Type

adamgfraser:
Resolves #2706 

This works pretty smoothly. I think the one issue is when we need to access information that is only available for a runtime fiber. Most of the time this doesn't come up because we already have methods on `ZIO` like `descriptor` and `trace`. Where it is noticeable is in producing dumps for individual fibers. We either need a method that lets us access the type of the fiber and perform an effect based on that (implemented here as `typeSpecific`), we need to expose and retain more information about the subtype in combinators, or possibly we could just add another method to `ZIO` that produces a dump of the current fiber.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
@@ -62,11 +62,9 @@ object RTSSpec extends ZIOBaseSpec {
       for {
         promise <- Promise.make[Nothing, Int]
         fiber   <- promise.await.fork
-        dumpStr <- fiber.typeSpecific(
-                    physical => physical.dump.flatMap(_.prettyPrintM),
-                    _ => UIO.dieMessage("synthetic fiber")
-                  )
-        _ <- UIO(println(dumpStr))
+        dump    <- fiber.dump
+        dumpStr <- dump.prettyPrintM
+        _       <- UIO(println(dumpStr))
       } yield assert(dumpStr)(anything)
     },
     testM("interruption causes") {
--- a/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
@@ -62,10 +62,12 @@ object RTSSpec extends ZIOBaseSpec {
       for {
         promise <- Promise.make[Nothing, Int]
         fiber   <- promise.await.fork
-        dump    <- fiber.dump
-        dumpStr <- dump.fold[URIO[Clock, String]](IO.succeed(""))(_.prettyPrintM)
-        _       <- UIO(println(dumpStr))
-      } yield assert(dump)(anything)
+        dumpStr <- fiber.typeSpecific(
+                    physical => physical.dump.flatMap(_.prettyPrintM),
+                    _ => UIO.dieMessage("synthetic fiber")
+                  )
+        _ <- UIO(println(dumpStr))
+      } yield assert(dumpStr)(anything)
     },
     testM("interruption causes") {
       for {
--- a/core/shared/src/main/scala/zio/Fiber.scala
+++ b/core/shared/src/main/scala/zio/Fiber.scala
@@ -103,30 +103,10 @@ trait Fiber[+E, +A] { self =>
   def children: UIO[Iterable[Fiber[Any, Any]]]
 
   /**
-   * Generates a fiber dump, if the fiber is not synthetic.
-   */
-  final def dump: UIO[Option[Fiber.Dump]] =
-    for {
-      name   <- self.getRef(Fiber.fiberName)
-      id     <- self.id
-      status <- self.status
-      trace  <- self.trace
-    } yield for {
-      id    <- id
-      trace <- trace
-    } yield Fiber.Dump(id, name, status, trace)
-
-  /**
    * Gets the value of the fiber ref for this fiber, or the initial value of
    * the fiber ref, if the fiber is not storing the ref.
    */
   def getRef[A](ref: FiberRef[A]): UIO[A]
-
-  /**
-   * The identity of the fiber, if it is a single runtime fiber. Fibers created
-   * from values or composite fibers do not have identities.
-   */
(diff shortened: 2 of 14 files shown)


core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -316,9 +290,17 @@ trait Fiber[+E, +A] { self =>
     ZManaged.make(UIO.succeed(self))(_.interrupt)
 
   /**
-   * The trace of the fiber. Currently only single runtime fibers have traces.
+   * Performs an effect depending on whether this is a runtime or synthetic
+   * fiber.
    */
-  def trace: UIO[Option[ZTrace]]
+  final def typeSpecific[R, E1 >: E, B](
```
jdegoes: We could also call this `foldM`, e.g. after `Either#fold`, since `Fiber` is now a sum type of two possibilities.

Also you really only need `fold`, i.e. fold to a generic `Z`, which will be no less powerful than this one (since that `Z` could also be a ZIO effect).

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -396,13 +376,45 @@ trait Fiber[+E, +A] { self =>
...
     }
 }
 
 object Fiber {
 
+  /**
+   * A runtime fiber that is executing an effect. Runtime fibers ave an
+   * identity and a trace.
+   */
+  trait Runtime[+E, +A] extends Fiber[E, A] { self =>
```
jdegoes: This is so much nicer!! Love the name too.

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: Yes you’re right about making it more generic. I thought about fold but was worried it could be confused with folding over the success and failure cases.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -566,12 +574,10 @@ object Fiber {
       final def await: UIO[Exit[E, A]]                     = IO.succeed(exit)
```
jdegoes: We still have code doing `new Fiber[E, A]`, whereas, we want every fiber to be either `Runtime` or `Synthetic`. So we should change every instance of `new Fiber` in this file to be `new Synthetic`.

Then make sure `Fiber` is sealed in this file, so we won't have to worry about outside code doing the wrong thing.

core-tests/jvm/src/test/scala/zio/RTSSpec.scala
```diff
@@ -62,10 +62,12 @@ object RTSSpec extends ZIOBaseSpec {
       for {
         promise <- Promise.make[Nothing, Int]
         fiber   <- promise.await.fork
-        dump    <- fiber.dump
-        dumpStr <- dump.fold[URIO[Clock, String]](IO.succeed(""))(_.prettyPrintM)
-        _       <- UIO(println(dumpStr))
-      } yield assert(dump)(anything)
+        dumpStr <- fiber.typeSpecific(
```
jdegoes: When you call `fork` on a fiber, you know it's a runtime fiber. So we can change the signature of `fork*` methods to return a `Fiber.Runtime`. Then this code will be simple again.

jdegoes: We should look for other places where we know it's a real fiber, and change them to be `Fiber.Runtime`.

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: Yes. If we want to seal it I think we are going to need to move `FiberContext` into the same file.

core-tests/jvm/src/test/scala/zio/RTSSpec.scala
adamgfraser: Yes. I think we may need to move a couple more implementations down from `Fiber` to `Runtime` and `Synthetic`. For example mapping a fiber would return the same type as the original fiber I believe.

jdegoes: Well, due to covariance, we can override `map` to preserve the subtype of the parent (e.g. in `Runtime#map`, it returns a `Runtime`, but in `Synthetic#map`, it returns a `Synthetic`). But is that actually legit? Technically, a mapped fiber is a synthetic fiber that happens to be 1-to-1 with the original. I think we should preserve the invariant that, essentially, `FiberContext` is the only real `Runtime` fiber, and all the others are `Synthetic`. 

We could always add a method like `childrenRuntime` to grab the children and filter to those that are `Runtime` if that proves helpful.. then if you map and grab the childrenRuntime, you can find the original one again

adamgfraser: Yes, I think that is fair. I was thinking of something like `mapM` where `id` and `trace` currently return the same values as the original fibers but maybe that is not legitimate and being more explicit about this is actually forcing us to clean it up.

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: Used a trick to get around this.

adamgfraser:
Looking really nice now. Every fiber guaranteed to be either a runtime fiber or a synthetic fiber and every operation returns the precise type except for ones that return a collection of fibers.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -396,13 +375,49 @@ trait Fiber[+E, +A] { self =>
...
+      for {
+        name   <- self.getRef(Fiber.fiberName)
+        id     <- self.id
+        status <- self.status
+        trace  <- self.trace
+      } yield Fiber.Dump(id, name, status, trace)
+  }
+
+  private[zio] object Runtime {
+    trait Internal[+E, +A] extends Runtime[E, A]
```
jdegoes: Nice trick! 👍

PR #2717, opened 2020-01-25
Implement ZIO#filter

adamgfraser:


adamgfraser:
Perhaps it would be better to call this filter for consistency with existing methods like foldLeft.

jdegoes:
@adamgfraser Agreed to call this `filter` instead of `filterM`. I think in cases where only the monadic version makes sense, it makes sense to drop the `M`. e.g.: `foreach`, `foldLeft`, `reduceAll`, etc.

PR #2723, opened 2020-01-25
Fix Bugs in ZIO#partitionM and ZIO#validateFirstM

adamgfraser:
`partitionM` evaluated effects in the wrong order. `validateFirstM` accumulated errors in the wrong order. Fixes both and adds regression tests. Also renames combinators to remove `M` suffix` for consistency with other combinators.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2689,23 +2689,17 @@ object ZIO {
    *
    * Collects all successes and failures in a tupled fashion.
    */
-  def partitionM[R, E, A, B](
+  def partition[R, E, A, B](
     in: Iterable[A]
   )(f: A => ZIO[R, E, B])(implicit ev: CanFail[E]): ZIO[R, Nothing, (List[E], List[B])] =
-    ZIO.foldRight(in)(List.empty[E] -> List.empty[B]) {
-      case (a, (es, bs)) =>
-        f(a).fold(
-          e => (e :: es, bs),
-          b => (es, b :: bs)
-        )
-    }
+    ZIO.foreach(in)(f(_).either).map(partitionMap(_)(ZIO.identityFn))
 
   /**
    * Feeds elements of type `A` to a function `f` that returns an effect.
    * Collects all successes and failures in parallel and returns the result as a tuple.
    *
    */
-  def partitionMPar[R, E, A, B](
+  def partitionPar[R, E, A, B](
     in: Iterable[A]
   )(f: A => ZIO[R, E, B])(implicit ev: CanFail[E]): ZIO[R, Nothing, (List[E], List[B])] =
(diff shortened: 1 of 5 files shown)


saraiva132:
Thanks for the cleanup. Is the M suffix being dropped?

jdegoes:
@saraiva132 We decided that if `M` is the only possible variant, it can be dropped.

For example, ZIO would never include a function like `def filter(is: Iterable[A])(f: A => Boolean): Iterable[A]` or like `def foreach(is: Iterable[A])(f: A => Unit): Unit`. Therefore, because the `M` variant is the only possible "useful" variant, it makes sense to drop the `M` just for convenience.

You can see that pattern on: `foreach`, `reduceAll`, `collectAll`, `filter`, etc.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2949,10 +2943,10 @@ object ZIO {
    * This combinator is lossy meaning that if there are errors all successes will be lost.
    * To retain all information please use ZIO#partitionM
```
jdegoes: This Scaladoc mentions `partitionM`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2964,33 +2958,28 @@ object ZIO {
    * This combinator is lossy meaning that if there are errors all successes will be lost.
    * To retain all information please use ZIO#partitionMPar
```
jdegoes: This Scaladoc mentions `partitionMPar`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2689,23 +2689,17 @@ object ZIO {
...
     in: Iterable[A]
   )(f: A => ZIO[R, E, B])(implicit ev: CanFail[E]): ZIO[R, Nothing, (List[E], List[B])] =
-    ZIO.foldRight(in)(List.empty[E] -> List.empty[B]) {
-      case (a, (es, bs)) =>
-        f(a).fold(
-          e => (e :: es, bs),
-          b => (es, b :: bs)
-        )
-    }
+    ZIO.foreach(in)(f(_).either).map(partitionMap(_)(ZIO.identityFn))
```
jdegoes: ❤️  Beautiful.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Done.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Done.

adamgfraser:
@jdegoes Good catch on the documentation. I fixed those and converted them to Scaladoc links so they get checked in the future.

PR #2726, opened 2020-01-26
Introduce ZStream.Control and ZSink.Control

vasilmkd:
I took some liberties with the naming, but it can be negotiated. 😄

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -19,20 +19,18 @@ class ZStream[-R, +E, -I, +B, +A](val process: ZManaged[R, Nothing, ZStream.Cont
 object ZStream {
 
   final case class Control[-R, +E, -I, +B, +A](
-    pull: Pull[R, E, B, A],
-    command: Command[R, E, I]
+    pull: ZIO[R, Either[E, B], A],
+    command: I => ZIO[R, E, Any]
   )
 
-  type Pull[-R, +E, +B, +A] = ZIO[R, Either[E, B], A]
+  object Pull {
+    def end[B](b: => B): IO[Either[Nothing, B], Nothing] = IO.fail(Right(b))
 
-  type Command[-R, +E, -I] = I => ZIO[R, E, Any]
-
-  object Command {
-    val unit: Command[Any, Nothing, Any] = _ => UIO.succeed(())
+    val endUnit: IO[Either[Nothing, Unit], Nothing] = end(())
   }
 
-  object Pull {
-    val end: Pull[Any, Nothing, Unit, Nothing] = IO.fail(Right(()))
+  object Command {
+    val noop: Any => UIO[Unit] = _ => UIO.unit
   }
 
   def apply[R, E, I, B, A](process: ZManaged[R, Nothing, Control[R, E, I, B, A]]): ZStream[R, E, I, B, A] =
@@ -61,6 +59,6 @@ object ZStream {
               } yield a).mapError(Left(_))
           }
         }
-      } yield Control(pull, Command.unit)
+      } yield Control(pull, Command.noop)
     }
 }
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2,18 +2,40 @@ package zio.stream.experimental
...
-  type Pull[-R, +E, -I, +O, +A] = I => ZIO[R, Either[O, E], A]
+
+  final case class Control[-R, +E, -I, +B, +A](
+    pull: Pull[R, E, B, A],
+    command: Command[R, E, I]
+  )
+
+  type Pull[-R, +E, +B, +A] = ZIO[R, Either[E, B], A]
+
+  type Command[-R, +E, -I] = I => ZIO[R, E, Any]
```
jdegoes: I am not sure this type alias adds much (just adds a level of indirection without saving much typing). I think I'd delete it. Or if not delete it, then add the corresponding `type Query` for `ZSink`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2,18 +2,40 @@ package zio.stream.experimental
...
+
+  object Command {
+    val unit: Command[Any, Nothing, Any] = _ => UIO.succeed(())
```
jdegoes: `UIO.unit` will be a LOT faster.

I think I'd call this command `void` or something. Not `unit`, which has a well established meaning in ZIO.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2,18 +2,40 @@ package zio.stream.experimental
...
+  }
+
   object Pull {
-    val end: Pull[Any, Nothing, Any, Unit, Nothing] = _ => IO.fail(Left(()))
+    val end: Pull[Any, Nothing, Unit, Nothing] = IO.fail(Right(()))
```
jdegoes:
```scala
def end[A](a: => A): Pull[Any, Nothing, A, Nothing] = ...

val endUnit = end(())
```

vasilmkd:
Do you want me to go back on the naming?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -27,19 +49,18 @@ object ZStream {
...
-              if (done) IO.fail(Left(()))
-              else
-                (for {
-                  _           <- doneRef.set(true)
-                  reservation <- managed.reserve
-                  _           <- finalizer.set(reservation.release)
-                  a           <- restore(reservation.acquire)
-                } yield a).mapError(Right(_))
-            }
+        pull = ZIO.uninterruptibleMask { restore =>
```
jdegoes: The implementation of `fromEffect` in terms of `managed` is too complicated IMO. The `zio` effect has no finalizers.

I think you should add: `ZIO#once(e: E): UIO[ZIO[R, E, A]]`, which returns a ZIO effect that will execute a single time, and if you call it again, you get the specified error. You could do the same for `ZManaged`.

Then you can implement these two methods simply.

iravid:
I think the naming is appropriate (`Control` for the case class bundling the two actions). However, I suggest that we drop all of the type aliases entirely to keep the naming budget in check.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
vasilmkd: Maybe `noop`?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -19,27 +21,37 @@ class ZStream[-R, +E, -I, +B, +A](val process: ZManaged[R, Nothing, ZStream.Cont
...
+  object Command {
+    val noop: Any => UIO[Unit] = _ => UIO.unit
   }
 
   def apply[R, E, I, B, A](process: ZManaged[R, Nothing, Control[R, E, I, B, A]]): ZStream[R, E, I, B, A] =
     new ZStream(process)
 
   def fromEffect[R, E, A](zio: ZIO[R, E, A]): ZStream[R, E, Any, Unit, A] =
-    managed(zio.toManaged_)
+    ZStream {
```
iravid: Cool. We can extract the technique here to a combinator on `ZIO` in a follow-up.

vasilmkd:
Open the floodgates.

PR #2729, opened 2020-01-26
add ZManaged#scopeIO

mschuwalow:
I don't know whether it's worth having as a combinator, but I it found very useful to have.

/cc @iravid

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1047,11 +1047,11 @@ object ZManaged {
       ZManaged.environment.flatMap(f)
   }
 
-  trait Scope {
+  trait PreallocationScope {
     def apply[R, E, A](managed: ZManaged[R, E, A]): ZIO[R, E, Managed[Nothing, A]]
   }
 
-  trait ScopeIO {
+  trait Scope {
     def apply[R, E, A](managed: ZManaged[R, E, A]): ZIO[R, E, (A, UIO[Any])]
   }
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1539,12 +1543,26 @@ object ZManaged {
...
+    }
+
+  /**
+   * Creates a scope in which resources can be safely allocated into together with a release action.
+   */
+  def scopeIO[R]: ZManaged[R, Nothing, ScopeIO] =
     ZManaged {
       // we abuse the fact that Function1 will use reference equality
-      Ref.make(Set.empty[Exit[Any, Any] => ZIO[R, Nothing, Any]]).map { finalizers =>
+      Ref.make(Set.empty[Exit[Any, Any] => ZIO[Any, Nothing, Any]]).map { finalizers =>
```
```suggestion
      Ref.make(Set.empty[Exit[Any, Any] => UIO[Any]]).map { finalizers =>
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1539,12 +1543,26 @@ object ZManaged {
    * Creates a scope in which resources can be safely preallocated.
    */
   def scope[R]: ZManaged[R, Nothing, Scope] =
```
iravid: How about we rename `scope` to `preallocationScope`, and give the `scope` names to the new combinator?

iravid: This way I wouldn't be entirely lying here: https://twitter.com/iravid_/status/1180569798037708800?s=20 😇

mschuwalow: Nice, will update 😁

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1665,6 +1648,41 @@ object ZManaged {
   def sandbox[R, E, A](v: ZManaged[R, E, A]): ZManaged[R, Cause[E], A] =
     v.sandbox
 
+  /**
+   * Creates a scope in which resources can be safely allocated into together with a release action.
+   */
+  def scope[R]: ZManaged[R, Nothing, Scope] =
```
```suggestion
  val scope: ZManaged[R, Nothing, Scope] =
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1538,35 +1542,14 @@ object ZManaged {
...
-            }
-          },
-          release = exitU =>
-            for {
-              fs    <- finalizers.get
-              exits <- ZIO.foreachPar(fs)(_(exitU).run)
-              _     <- ZIO.done(Exit.collectAllPar(exits).getOrElse(Exit.unit))
-            } yield ()
-        )
+  def preallocationScope[R]: ZManaged[R, Nothing, PreallocationScope] =
```
```suggestion
  val preallocationScope: ZManaged[Any, Nothing, PreallocationScope] =
```

mschuwalow:
Ah sorry, didn't pay attention with the type parameters.
Scope will need to stay a def though or we have to break our alphabetic ordering

iravid:
Ah sure. We can just drop the outer `R` on it then.

PR #2741, opened 2020-01-27
Made foreachPar_ task counting robust against interruption.

### LINKED ISSUE #2739: Nonterminating ZIO#foreachPar Test
May be related to recent optimizations of parallel combinators. Copying @horned-sphere.

```
[info] ZIO Test
[info] - ZIOSpec
[info]   - foreachPar
[info]     - runs a task that fails
[info] Timeout of 10 m exceeded.
```

https://circleci.com/gh/zio/zio/52762?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

horned-sphere:
Fixes #2739.

Also fixes typo error accumulation test case for foreachPar_.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2318,19 +2318,41 @@ object ZIO {
         causes         <- Ref.make[Cause[E]](Cause.empty)
         result         <- Promise.make[Nothing, Boolean]
         failureTrigger <- Promise.make[Unit, Unit]
-        count          <- Ref.make(0)
-        fibers <- ZIO.foreach(as) {
-                   f(_)
-                     .tapCause(c => causes.update(_ && c) *> failureTrigger.fail(()))
-                     .ensuring {
-                       count.update(_ + 1) >>= { done =>
-                         ZIO.when(done == size) {
-                           result.complete(failureTrigger.succeed(()))
-                         }
-                       }
-                     }
-                     .fork
-                 }
+        status         <- Ref.make((0, 0, false))
+
+        startTask = status.modify {
+          case (started, done, failing) =>
+            if (failing) {
+              (false, (started, done, failing))
+            } else {
+              (true, (started + 1, done, failing))
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2318,19 +2318,41 @@ object ZIO {
         causes         <- Ref.make[Cause[E]](Cause.empty)
         result         <- Promise.make[Nothing, Boolean]
         failureTrigger <- Promise.make[Unit, Unit]
-        count          <- Ref.make(0)
-        fibers <- ZIO.foreach(as) {
```
adamgfraser: So if I understand it right the problem here is really `pre-interruption`. Once we get a failure we start interrupting the other tasks and we can "pre-interrupt" them before they start running so the finalizer in `ensuring` is never triggered and we never increment the counter and thus doesn't get to the required value to continue with the computation. Can we use `uninterruptibleMask` to address this?

horned-sphere: There are two routes to problems, one is pre-interruption and the other is interruption during the book keeping tasks surrounding the execution of the function.

My fix comes in two parts:

1. Count the tasks that have started. When any task fails switch the mode so no tasks can start then wait for the completion count to reach the started count rather than the total number of tasks.
2. Run the book keeping uninterrutibly, switching to interruptible for actual task which ensures that the increments of the started and completed tasks always happen, given that the fiber started. The only reason the fibers would not start is that they had been interrupted and so that case is covered by 1.

(I verified that both are actually needed. If either of the fixes is removed the failures tests will hang with varying frequencies).

I looked at uninterruptibleMask for dealing with 2 but the main advantage of it is that it will restore the surrounding interruption state. However we already know the interruption state as we are constructing the entire fiber from scratch. I could replace the bracketing uninterruptible/interruptible with uninterruptibleMask if you thinks it looks clearer (and this would be functionally identical).

adamgfraser: Got it. That makes sense. Thanks!

PR #2743, opened 2020-01-28
Missing ZStream tests and Serializable

vasilmkd:
It turns out I can't even implement a simple constructor without tests. 🤣 

@iravid @jdegoes Could you please enable tests for this branch, I think they will be desperately needed.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -1,0 +1,76 @@
+package zio.stream.experimental
+
+import ZStreamUtils.nPulls
+
+import zio._
+import zio.test.Assertion.{ equalTo, isFalse, isTrue }
+import zio.test._
+
+object ZStreamSpec extends ZIOBaseSpec {
+  def spec = suite("ZStreamSpec")(
+    suite("Combinators")(
+      testM("map") {
+        ZStream
+          .fromEffect(UIO.succeed(1))
+          .map(_.toString)
+          .process
+          .use(nPulls(_, 3))
+          .map(assert(_)(equalTo(List(Right("1"), Left(Right(())), Left(Right(()))))))
+      }
+    ),
+    suite("Constructors")(
+      suite("fromEffect")(
+        testM("success") {
+          ZStream
+            .fromEffect(UIO.succeed(1))
+            .process
+            .use(nPulls(_, 3))
+            .map(assert(_)(equalTo(List(Right(1), Left(Right(())), Left(Right(()))))))
+        },
+        testM("failure") {
+          ZStream
+            .fromEffect(IO.fail("Ouch"))
+            .process
+            .use(nPulls(_, 3))
+            .map(assert(_)(equalTo(List(Left(Left("Ouch")), Left(Right(())), Left(Right(()))))))
+        }
+      ),
+      suite("managed")(
+        testM("success") {
+          for {
+            ref <- Ref.make(false)
+            pulls <- ZStream
+                      .managed(Managed.make(UIO.succeed(1))(_ => ref.set(true)))
+                      .process
+                      .use(nPulls(_, 3))
+            fin <- ref.get
+          } yield assert(fin)(isTrue) && assert(pulls)(equalTo(List(Right(1), Left(Right(())), Left(Right(())))))
+        },
+        testM("acquisition failure") {
+          for {
+            ref <- Ref.make(false)
+            pulls <- ZStream
+                      .managed(Managed.make(IO.fail("Ouch"))(_ => ref.set(true)))
+                      .process
+                      .use(nPulls(_, 3))
+            fin <- ref.get
+          } yield assert(fin)(isFalse) && assert(pulls)(
+            equalTo(List(Left(Left("Ouch")), Left(Right(())), Left(Right(()))))
+          )
+        },
+        testM("inner failure") {
+          for {
+            ref <- Ref.make(false)
+            pulls <- ZStream
+                      .managed(Managed.make(UIO.succeed(1))(_ => ref.set(true)) *> Managed.fail("Ouch"))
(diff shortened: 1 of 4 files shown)


iravid:
What do you mean by "enable tests for this branch"?

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamUtils.scala
```diff
@@ -0,0 +1,13 @@
+package zio.stream.experimental
+
+import zio._
+
+trait ZStreamUtils {
```
iravid: Why both trait and object?

vasilmkd:
I could only see the linter being run. But now I see the tests as well. Weird.

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamUtils.scala
vasilmkd: It was that way in the old utils. Should I change it to just object?

iravid:
> I could only see the linter being run. But now I see the tests as well. Weird.

Yeah, the other workflow steps are dependent on the lint succeeding to save on CI costs.

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamUtils.scala
iravid: Yeah, no need for that trait I think. Probably a leftover of the old inheritance based encoding for ZSpecs.

vasilmkd:
@iravid I'm aware, but on master the UI shows all 11 with 1 running and 10 pending on the linter. This UI only showed 1 check and no other being queued. That's why I thought it was weird. But it seems to be fine now.

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamUtils.scala
vasilmkd: Removed the trait.

vasilmkd:
I'm guessing tests for Sinks should follow the tests for Streams, now that the structure is the same?

vasilmkd:
@jdegoes @iravid Do you mind renaming the `I` type parameter on the sink to `S`, symbolizing more of a state for the sink? I feel it's a bit more descriptive.

jdegoes:
@vasilmkd Maybe `M` would be better, "Message", as it works both for sinks and streams? (input/output message)

vasilmkd:
We'll get on that. The classes need a full scaladoc.

jdegoes:
Btw:

 - sinks that output no state should probably use `Unit`
 - streams that accept no input should probably use `Any` as the input message

And also:

 - streams and sinks that do not terminate with a value should use `Nothing` for `B`
 - streams and sinks that terminate but not with any useful value should use `Unit` for `B`

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+        testM("failure") {
+          ZStream
+            .fromEffect(IO.fail("Ouch"))
+            .process
+            .use(nPulls(_, 3))
+            .map(assert(_)(equalTo(List(Left(Left("Ouch")), Left(Right(())), Left(Right(()))))))
+        }
+      ),
+      suite("managed")(
+        testM("success") {
```
jdegoes: I like these tests, we should try to thoroughly exercise behavior (especially edge cases) now that we have clear semantics for everything.

vasilmkd: I had these tests before, check `StreamPullSafetySpec.scala`. :)

vasilmkd:
> 
> 
> Btw:
> 
>     * sinks that output no state should probably use `Unit`
> 
>     * streams that accept no input should probably use `Any` as the input message
> 
> 
> And also:
> 
>     * streams and sinks that do not terminate with a value should use `Nothing` for `B`
> 
>     * streams and sinks that terminate but not with any useful value should use `Unit` for `B`

I'll double check the existing code. Excellent points.

vasilmkd:
I won't be pushing code any more. Sorry. Feel free to merge.

adamgfraser:
We should use the new `now` variants of effect constructors internally here.

vasilmkd:
I think this branch doesn't have the changes...

adamgfraser:
Okay, no worries. We should just remember to switch them over the next time we rebase this branch on `master`.

iravid:
I'll rebase the base branch after this merge

vasilmkd:
@adamgfraser @iravid Do you mind briefly commenting why that is desirable?

adamgfraser:
@vasilmkd Sure. We moved the default implementation of all effect constructors (e.g. `ZIO.succeed`) to be by name (and if necessary to use combinators such as `effectTotal` to make sure the argument did not get immediately evaluated). The rationale was to improve usability by making sure that if users accidentally performed side effects (e.g. getting the name of the current thread) or threw errors in their code those would be deferred and properly managed by ZIO.

While desirable in general, this potentially has some performance impact in certain cases, so we implemented new `private[zio]` `now` variants that retain the old eager behavior. So in internal combinators where we are no longer at the border with possibly impure code we can use these for potentially slightly better performance. Make sense?

adamgfraser:
It also may be worth looking at the stream versions in particular and seeing if you think there is actually a benefit in having the strict versions or if the implementations of those versions can be further optimized. I did my best but that was a large PR so to some extent my focus was on making the changes consistently and in a value preserving way versus optimizing each use case.

vasilmkd:
Any particular optimizations that we are looking for?

adamgfraser:
It would just be looking at the implementations of the `now` variants and seeing if there is a more efficient way to stuff a value that you know for sure is an existing pure value into a stream, sink, pull, etc...

PR #2747, opened 2020-01-28
Increase keepAlive Time of Default Thread Pools

### LINKED ISSUE #2746: Increase keepAlive time of default thread pools
The default thread pools in ZIO should utilize keep alive times of 60 seconds, rather than their current default of 1 second.

See `zio.internal`, including platform-specific files on JVM.

adamgfraser:
Resolves #2746.

### DIFF
--- a/core/jvm/src/main/scala/zio/blocking/package.scala
+++ b/core/jvm/src/main/scala/zio/blocking/package.scala
@@ -168,7 +168,7 @@ package object blocking {
       Executor.fromThreadPoolExecutor(_ => Int.MaxValue) {
         val corePoolSize  = 0
         val maxPoolSize   = Int.MaxValue
-        val keepAliveTime = 1000L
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/zio/blocking/package.scala
```diff
@@ -168,7 +168,7 @@ package object blocking {
       Executor.fromThreadPoolExecutor(_ => Int.MaxValue) {
         val corePoolSize  = 0
         val maxPoolSize   = Int.MaxValue
-        val keepAliveTime = 1000L
+        val keepAliveTime = 60000L
```
jdegoes: I think maybe we should change `maxPoolSize` to something like 1000.

adamgfraser: Done.

PR #2748, opened 2020-01-28
Improvements to ZIO runtime system

jdegoes:
### Weak automatic supervision

Supervision is now backed by weak sets, which means that child fibers may be freely garbage collected when they cannot resume. This should address memory leaks in scenarios where large numbers of non-terminating fibers are created.

### Structured concurrency

All child fibers are automatically bound to their parent fiber. When the parent fiber exits, the child fiber will be interrupted or disowned, as determined by the `SuperviseMode` it is forked with. The default supervision mode for forked fibers is interruption, which means that by default, when a parent fiber exits, its child fibers will be interrupted. This ensures fibers do not leak.

In order to recapture the previous behavior, one may use `forkDaemon`, or `disown`, which explicitly disowns a child fiber and moves it to a root set of fibers.

### Safer race

Race no longer forces either side to be interruptible. Users may have to perform `left.interruptible.race(right.interruptible)` to acquire the old behavior (which punched holes in uninterruptible regions); or possibly, `left.disconnect.race(right.disconnect)` (if they want the effects to be keep running when interrupted, but in the background).

### New combinators

Several new combinators are added:

 - `disconnect` — disconnects interruption of the effect from its parent. This allows "early interruption" without waiting for finalization. It replaces the need for all `xyzFork` variants, which are now deprecated. It's quite useful with `race`, with or without `interruptible`.
 - `disown` — Called by a parent fiber to disown a child fiber. Relocates the child to the root set.

### New ZIO.never

When `ZIO.never` is forked, it will be garbage collected, so `ZIO.never.onInterrupt(putStrLn("Bye"))` will never execute the finalizer. This can be inconvenient for testing, so `ZIO.infinity` is added which is like `never`, but won't be garbage collected.

### Deletion of Daemon Mode

Daemon mode has been deleted because in practice, code never knows whether grandchildren should be daemons, only whether immediate children should be daemons; and generally, daemon mode is a one-off for a specific forked fiber. 

In place of daemon mode, `effect.forkDaemon` can be used, which is the composition of two other operators: ordinary `fork`, followed by an immediate `disown` of the child fiber.

### Deprecation of *Fork Variations

All `xyzFork` variants are deprecated or removed. For example, `bracketFork`. This is thanks to the new `disconnect` operator which allows for a much more compositional approach to solving this problem. Now if you don't want interruption of something to wait around for completion, just use `effect.disconnect`.


neko-kai:
The `evalOn` changes aren't part of the fix, are they?

jdegoes:
They used to be, I doubt they will end up being necessary. Updated PR incoming today

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1911,16 +1911,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
         permits <- Semaphore.make(n.toLong).toManaged_
         _ <- self.foreachManaged { a =>
               for {
-                latch <- Promise.make[Nothing, Unit]
                 p     <- Promise.make[E1, B]
+                latch <- Promise.make[Nothing, Unit]
                 _     <- out.offer(Pull.fromPromise(p))
-                _     <- (permits.withPermit(latch.succeed(()) *> f(a).to(p))).fork
+                _     <- permits.withPermit(latch.succeed(()) *> f(a).to(p)).fork
```
iravid: The parens were changed slightly here; should be:
```suggestion
                _     <- (permits.withPermit(latch.succeed(())) *> f(a).to(p)).fork
```
The purpose of the latch is to ensure that the forked fiber has started before moving on to the next element.

adamgfraser: Doesn't that change mean that we give back the permit as soon as we complete the latch? Wouldn't that result in higher than intended parallelism in the execution of the `f` effect?

iravid: Yes, but the parallelism is limited by the capacity of the queue.

I actually think the permits are now redundant, but I'll need to look back to recall why they were added.

adamgfraser: Here is the original version:

```scala
(permits.withPermit(latch.succeed(()) *> f(a).to(p))).fork
```

It looks like there is one set of parentheses surrounding everything except `fork` so I think it is value preserving to remove them.

adamgfraser: I'm not sure if the queue is enough because without the semaphore we'll have all these fibers suspending to wait to offer the result of `f` to the queue but I think they will still all be doing `f` in parallel. I'll add a test for this and we can see if we can simplify.

iravid: Ah you're right - the permits were surrounding everything. Good catch.

They won't all be suspending while executing `f`, as they first enqueue a `Pull.fromPromise`, and only then execute `f`. So once `n` have executed, the next one will be suspended on enqueueing the promise.

adamgfraser: Yes I think you're right the semaphore can be eliminated.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1236,9 +1265,30 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
+   * interrupt signal, allowing the earliest possible return.
    */
   final def race[R1 <: R, E1 >: E, A1 >: A](that: ZIO[R1, E1, A1]): ZIO[R1, E1, A1] =
-    raceEither(that).map(_.merge)
+    ZIO.fiberId
+      .flatMap(parentFiberId =>
+        (self raceWith that)(
+          (exit, right) =>
+            exit.foldM[Any, E1, A1](
+              cause => right.join mapErrorCause (cause && _), // TODO: Preserve error?
```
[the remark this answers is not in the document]
jdegoes: I preserved them, I just hate the added overhead. Anyway, I'll delete the TODO.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -744,41 +760,27 @@ private[zio] final class FiberContext[E, A](
...
   private[this] def isInterruptible(): Boolean = interruptStatus.peekOrElse(true)
 
   @inline
-  private[this] def shouldInterrupt(): Boolean =
-    isInterrupted() && isInterruptible()
+  private[this] def isInterrupting(): Boolean = state.get().interrupting
+
+  @inline
+  private[this] final def shouldInterrupt(): Boolean =
+    isInterrupted() && isInterruptible() && !isInterrupting()
```
regiskuckaertz: 😆

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -683,10 +685,20 @@ private[zio] final class FiberContext[E, A](
     if (exitAsync(epoch)) evaluateLater(zio)
   }
 
-  def interruptAs(fiberId: Fiber.Id): UIO[Exit[E, A]] = kill0(fiberId)
+  private def withChildren[A](f: java.util.Set[FiberContext[Any, Any]] => A): A =
```
regiskuckaertz: Is there any perf improvement that could be made by making this function return `Any`?

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: 😆

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: I think it would roughly be the same...

PR #2749, opened 2020-01-28
ztest: single entrypoint

pshirshov:
Trying to address the issues explained in https://github.com/zio/zio/issues/1208#issuecomment-513761904

So, essentially in our case each test is a function PLUS its signature, introspectable in runtime (a sequence of typetags to be precise).
For our workflow we need to collect all the individual tests from all the classes, then decide which components can be reused between the tests (for example, there is no need to re-instantiate a database driver for every test) and  build a "test plan" - a project network for shared components plus individual project networks for every individual test. Then we execute them and report progress along the way.

There are several problems in current zio-test design preventing us from implementing our flow:

1) Every BaseTestTask is independent. It loads an individual class and executes it in an isolated environment. For our workflow we need just ONE Task which loads all the tests and delegates the work to our engine
2) Though ZTestRunner is hardwired with ZTestTask and there is no way to supply custom loading and execution strategies

These are the blockers. There is one more thing which is not critical but kinda annoying:
3) Label is an abstract type and there is no typeclass nor base class for it. So you use .toString instead of a typeclass: searchTerms.exists(term => label.toString.contains(term)). We need to use labels to carry additional test metadata, the best thing we can do is something like 

```
case class IzumiLabel[L](roots: Set[LightTypeTag], label: L) {
def toString = label.toString
}
```

Which I consider to be a very ugly solution

A summary: in order to meet our goal we need
1) An ability to collect *all* the tests from *all* the testclasses at one point (BaseTestTask probably)
2) An ability to set custom execution logic for BaseTestTask (an SBT key would be perfect,  Java SPI may be fine too). Also it would be great to have an ability to use custom test discovery logic in BaseTestTask - this way we may be able to create custom DSLs with ease
3) A typeclass for Labels (not critical but very good to have)

### DIFF
--- a/test-sbt/jvm/src/main/scala/zio/test/sbt/ZTestRunner.scala
+++ b/test-sbt/jvm/src/main/scala/zio/test/sbt/ZTestRunner.scala
@@ -53,24 +53,23 @@ final class ZTestRunner(val args: Array[String], val remoteArgs: Array[String], 
   def tasks(defs: Array[TaskDef]): Array[Task] = {
     val testArgs        = TestArgs.parse(args)
     val tasks           = defs.map(new ZTestTask(_, testClassLoader, sendSummary, testArgs))
-    val entrypointClass = testArgs.rootTask.getOrElse(classOf[ZTestRootTask].getName)
-    val rootTask = getClass.getClassLoader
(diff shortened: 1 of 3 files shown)


pshirshov:
Seems like sbt does not filter "dynamic" tasks. So, I got rid of the "root" task but left the extension point. 

I think in case you accept this we will experiment with custom policies in Izumi then make another P/R once we get a satisfactory result.

PR #2751, opened 2020-01-28
Implement Convenience Methods for Creating a Runtime with an Environment

adamgfraser:
Implements `Runtime.unsafeMake` which builds a runtime from a `ZLayer` and uses it to implement `unsafeDefault` and `unsafeGlobal`, which build runtimes from a `ZLayer` using the default executor and Scala's global execution context, respectively.

I ended up not being happy with the `shutdownHook` solution. It is not consistent on the JVM (e.g. shutdown hooks are not invoked when using SBT until exiting the session) and doesn't provide a way to run finalizers across platforms in a consistent manner. Instead, I had these methods return a tuple of a runtime and an effect that can be run to return the finalizer. This way users can create a runtime at the top of their application, use it throughout, and then specify their own logic about running finalizers (or discard it entirely if there are no finalizers).

### DIFF
--- a/core/native/src/main/scala/zio/internal/PlatformSpecific.scala
+++ b/core/native/src/main/scala/zio/internal/PlatformSpecific.scala
@@ -28,6 +28,12 @@ import zio.internal.tracing.TracingConfig
 private[internal] trait PlatformSpecific {
 
   /**
+   * Adds a shutdown hook that executes the specified action on shutdown.
+   */
+  def addShutdownHook(action: () => Unit): Unit =
+    ()
+
+  /**
    * A Runtime with settings suitable for benchmarks, specifically with Tracing
    * and auto-yielding disabled.
    *
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -167,6 +167,29 @@ trait Runtime[+R] {
 object Runtime {
 
   /**
+   * A runtime that can be shutdown to release resources allocated to it.
+   */
+  trait Managed[+R] extends Runtime[R] {
+
+    /**
+     * Shuts down this runtime and releases resources allocated to it. Once
+     * this runtime has been shut down the behavior of methods on it is
+     * undefined and it should be discarded.
+     */
+    def shutdown(): Unit
+  }
+
+  object Managed {
+
+    def apply[R](r: R, platform0: Platform, shutdown0: () => Unit): Runtime.Managed[R] =
+      new Runtime.Managed[R] {
+        val environment = r
+        val platform    = platform0
+        def shutdown()  = shutdown0()
+      }
+  }
+
+  /**
    * Builds a new runtime given an environment `R` and a [[zio.internal.Platform]].
    */
   def apply[R](r: R, platform0: Platform): Runtime[R] = new Runtime[R] {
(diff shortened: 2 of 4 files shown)


core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -174,4 +173,45 @@ object Runtime {
...
+  def unsafeGlobal[R <: Has[_]](layer: ZLayer.NoDeps[Nothing, R]): (Runtime[R], Exit[Any, Any] => UIO[Any]) =
+    unsafeMake(layer, global.platform)
+
+  /**
+   * Builds a new runtime given an effectual and resourceful environment `R`
+   * and a [[zio.internal.Platform]]. This method is effectful because it
+   * executes acquiring the environment. Returns a finalizer that can be used
+   * to release the environment.
+   */
+  def unsafeMake[R <: Has[_]](
```
regiskuckaertz: We usually mark unsafe methods as package-private; will this be needed outside of the library?

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -174,4 +173,45 @@ object Runtime {
...
+    layer: ZLayer.NoDeps[Nothing, R],
+    platform: Platform
+  ): (Runtime[R], Exit[Any, Any] => UIO[Any]) = {
```
jdegoes: I think that while fully general-purpose, this is too confusing for general consumption (indeed, even I had to think carefully about what `Exit` would be reasonable to pass to the finalizer?). A user can always do this manually if they are "advanced", so this solution should probably be targeted toward mainstream use cases.

I'd probably do it like this:

```scala
/**
 * Unsafely creates a `Runtime` from a `ZLayer` whose resources will be allocated
 * immediately, and not released until the end of the application.
 * 
 * This method is useful for small applications and integrating ZIO with legacy code, but
 * other applications should investigate using `ZIO#provideLayer` directly in their 
 * application entry points.
 */
def unsafeGlobal[R <: Has[_]](layer: ZLayer.NoDeps[Any, R]): Runtime[R] = ???
```

Then if you add to `internal.PlatformSpecific.scala` something like `def addShutdownHook(f: () => Unit): Unit` that does something differently on JVM / JS, and use that inside the implementation of this method to call the finalizer on shutdown.

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -174,4 +173,45 @@ object Runtime {
     val environment = r
     val platform    = platform0
   }
+
+  lazy val default = Runtime((), Platform.default)
+
+  lazy val global = Runtime((), Platform.global)
```
jdegoes: With `unsafeGlobal`, it occurs to me we have 2 different meanings of `Global`: first off, global as in "powered by the global Platform", as in "powered by the global execution context". Then we have `unsafeGlobal`, which is global as in "resources that are global, for the lifetime of the application, and not deallocated until shutdown".

We need to give one of them a different name to avoid confusion.

jdegoes:
> I ended up not being happy with the shutdownHook solution. It is not consistent on the JVM (e.g. shutdown hooks are not invoked when using SBT until exiting the session) and doesn't provide a way to run finalizers across platforms in a consistent manner. Instead, I had these methods return a tuple of a runtime and an effect that can be run to return the finalizer. This way users can create a runtime at the top of their application, use it throughout, and then specify their own logic about running finalizers (or discard it entirely if there are no finalizers).

Just saw this comment.

I don't think users using this feature will particularly care about finalization. That said, perhaps we could return `(Runtime[R], () => Unit)`, or maybe even better, some sort of:

```scala
trait ManagedRuntime[-R] extends Runtime[R] {
  def shutdown(): Unit
}
```

And return that from `unsafeGlobal`? The `Exit[Any, Any] => UIO[Unit]` will be rather difficult to explain IMO, and pulling the runtime out of the tuple is another step we can avoid.

jdegoes:
Maybe have the more powerful one that returns the "shutdown" hook and the default one that just adds the JVM / JS shutdown hook.

core/shared/src/main/scala/zio/Runtime.scala
adamgfraser: Unfortunately this will be needed outside the library. The background here is that with the new environment encoding `Runtime.default` returns a `Runtime[Unit]` instead of a `Runtime[ZEnv]`, which means that if you have a mixed application and you want a runtime to unsafely run effects that use any type of environment you no longer have one "off the shelf". The goal of the PR was to make that use case easier and since all environments are managed now we need to unsafely run them to get a runtime with an environment out.

We could only offer the `unsafeDefault` and `unsafeGlobal` methods above but I don't think that is really any safer and it just forces the user to use `mapPlatform` if they want to use their own executor settings.

adamgfraser:
I really like the `ManagedRuntime` idea. That gives us a nice way to wrap this up into an easier interface and I agree that most of the time there won't even be finalizers. That could also allow us to have a method `shutdownExit` on `ManagedRuntime` that would allow you to provide an exit value to the finalizers if you are a power user and want to. I'm still a little skeptical of the ability to do the automatic finalizers on shutdown in a way that is consistent. Let me work on incorporating the `ManagedRuntime` concept and then we can go from there.

adamgfraser:
Okay, so updated to use `Runtime.Managed`, which cleans things up a lot. I ended up leaving out the `shutdownExit` idea because it did seem a little confusing but we can bring it back.

We still don't have any automatic logic for running finalizers on exit. I could try to add that. We can definitely do on the JVM but I am worried that we are going to have counterintuitive behavior regarding things like when finalizers run in SBT since we're not controlling the whole application that will lead to a lot of questions. I'm not sure we can do anything on ScalaJS so I am inclined to say it is the responsibility of the caller to shutdown if necessary, but happy to go in a different direction.

In this version, `default` means uses the default executor and `global` means uses Scala's global execution context. Both of them have the same behavior regarding you taking on the responsibility for any finalization and getting something that you can reuse across whatever part of your application you want.

jdegoes:
@adamgfraser My feeling is that people will be less surprised if it works in production than if it does not work in SBT (although you are right, they will be surprised it does not work in SBT).

I also see that [Scala.js](https://github.com/scala-js/scala-js-dom/blob/deffe6cc956afd41b7572d0f5b386c4c1b63534b/src/main/scala/org/scalajs/dom/raw/Html.scala) supports onbeforeunload, which would allow us to add a (synchronous) shutdown hook to Javascript as well.

adamgfraser:
@jdegoes Okay, will add. At that point maybe we don't need the managed runtime?

jdegoes:
@adamgfraser I think the `ManagedRuntime` is nice because it gives people the ability to do early termination, or termination inside SBT.

I seriously do hope people don't use this too much though... while nice for upgrading or simple use of the environment, treating resources as if they were global is far from ideal...

adamgfraser:
@jdegoes Totally agree. Will set this up.

adamgfraser:

iravid:
@adamgfraser you could also try testing with “run / fork := true” in SBT. That forks the JVM that is run so finalizers should run properly

core/native/src/main/scala/zio/internal/PlatformSpecific.scala
```diff
@@ -27,6 +27,12 @@ import zio.internal.tracing.TracingConfig
 
 private[internal] trait PlatformSpecific {
 
+  /**
+   * Adds a shutdown hook that executes the specified action on shutdown.
+   */
+  def addShutdownHook(action: () => Unit): Unit =
```
jdegoes: With this added, is it possible to go down to a single `zio.App`, rather than separate versions for JS / JVM?

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -174,4 +200,59 @@ object Runtime {
...
+  /**
+   * Unsafely creates a `Runtime` from Scala's global execution context and a
+   * `ZLayer` whose resources will be allocated immediately, and not released
+   * until the `Runtime` is shut down or the end of the application.
+   *
+   * This method is useful for small applications and integrating ZIO with
+   * legacy code, but other applications should investigate using
+   * [[ZIO.provideLayer]] directly in their application entry points.
+   */
+  def unsafeGlobal[R <: Has[_]](layer: ZLayer.NoDeps[Any, R]): Runtime.Managed[R] =
```
jdegoes: How about we go down to a single method and call it `unsafeFromLayer`, with `platform` as an argument that defaults to `Platform.default`?

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -174,4 +200,59 @@ object Runtime {
...
+   * legacy code, but other applications should investigate using
+   * [[ZIO.provideLayer]] directly in their application entry points.
+   */
+  def unsafeMake[R <: Has[_]](layer: ZLayer.NoDeps[Any, R], platform: Platform): Runtime.Managed[R] = {
+    val runtime = Runtime((), platform)
+    val (environment, shutdown) = runtime.unsafeRun {
+      layer.build.reserve.flatMap {
+        case Reservation(acquire, release) =>
+          Ref.make(true).flatMap { finalize =>
+            val finalizer = () =>
```
jdegoes: To be "ZIO best practice" here we should wrap the inner block in `UIO.effectSuspend` since we are calling `addShutdownHook` there. Won't make a difference here but could be useful for people perusing the code.

core/shared/src/main/scala/zio/Runtime.scala
adamgfraser: Yes I like that a lot. Will make that change.

core/native/src/main/scala/zio/internal/PlatformSpecific.scala
adamgfraser: I think so. We may need a separate method with platform specific exit logic like we had in the bootstrap version of the ZIO Test internal test suite. I will work on this in a subsequent PR if that is alright.

core/shared/src/main/scala/zio/Runtime.scala
adamgfraser: Good point. Will fix.

adamgfraser:
All set.

adamgfraser:
@jdegoes Can I get a review on this?

adamgfraser:
Oh, you mean if you do something like `withReportFailure` you will get back a `Runtime` instead of a `Runtime.Managed`? That's a good point. I can submit a follow up PR to fix that.

PR #2752, opened 2020-01-28
Reimplement ZStream#bufferSliding safely and remove old bufferSignal

vasilmkd:
Out with the old, in with the new.

@iravid

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -765,54 +765,6 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
     }
 
   /**
-   * Creates a stream that passes its elements through a queue that drops elements
-   * (dropping or sliding). It implements a promise that signals abrupt completion
-   * of the stream due to an error or the end of the stream. After the promise has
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -874,7 +826,53 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * @note Prefer capacities that are powers of 2 for better performance.
    */
   final def bufferSliding(capacity: Int): ZStream[R, E, A] =
-    ZStream.managed(ZManaged.make(Queue.sliding[A](capacity))(_.shutdown)).flatMap(bufferSignal)
+    ZStream {
+      for {
+        as    <- self.process
+        queue <- Queue.sliding[(Take[E, A], Promise[Nothing, Unit])](capacity).toManaged(_.shutdown)
+        start <- Promise.make[Nothing, Unit].toManaged_
+        _     <- start.succeed(()).toManaged_
```
regiskuckaertz: I looked through code and it seems we don't do it often enough, but I wonder if there should be `Promise.succeed[Now]` and `Promise.fail[Now]` constructors.

vasilmkd:
Sure, I'll do another pass with strict constructors as well.

PR #2753, opened 2020-01-28
testlabel typeclass

pshirshov:
A part of #2749

### DIFF
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
@@ -36,9 +36,9 @@ abstract class BaseTestTask(
                    }.getOrElse(spec.spec)
                  }
              }
-      summary <- SummaryBuilder.buildSummary(spec)
+      summary <- SummaryBuilder.buildSummary(spec, this.spec.label)
       _       <- sendSummary.provide(summary)
-      events  <- ZTestEvent.from(spec, taskDef.fullyQualifiedName, taskDef.fingerprint)
+      events  <- ZTestEvent.from(spec, taskDef.fullyQualifiedName, taskDef.fingerprint, this.spec.label)
       _       <- ZIO.foreach[Any, Throwable, ZTestEvent, Unit](events)(e => ZIO.effect(eventHandler.handle(e)))
     } yield ()
 
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/ZTestEvent.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/ZTestEvent.scala
@@ -1,6 +1,7 @@
 package zio.test.sbt
 
 import sbt.testing._
+
 import zio.UIO
 import zio.test.{ ExecutedSpec, Spec, TestFailure, TestLabel, TestSuccess }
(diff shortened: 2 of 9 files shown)


adamgfraser:
Have you looked at `TestAnnotations`? Can we use them for this?

pshirshov:
`TestAnnotations` are more like transformers. We need to carry metadata for every test. And anyway, current approach with calling `.toString` on values of an abstract type `L` is very dirty and non-functional. So I think this should make things better anyway.

adamgfraser:
I think you are thinking of test aspects. Test annotations are designed for adding metadata to tests so they seemed like they could be applicable to this use case.

pshirshov:
Nope, in our case metadata is not an optional aspect but an obligatory part of our input. That metadata needs to exist outside of the test's IO - essentially it defines test signature. I had a look at other stuff and I'm sure it wouldn't work for us.

neko-kai:
@adamgfraser 
>  Test annotations are designed for adding metadata to tests so they seemed like they could be applicable to this use case.

No, they're totally unrelated, `TestAnnotation`/`Annotations.Service` allow accumulating _mutable summaries within a running test case_, this is unrelated to attaching immutable metadata to tests, to be inspected _before the test run_. I think the name of `TestAnnotation` is somewhat confusing.

Now static metadata is already addable to tests by specializing the type of `Label` when inheriting `AbstractRunnableSpec`

The `TestLabel[L]` just enables extracting the test name out of `L` for reporting purposes.

```scala
class Requirement(...)
class TestConfiguration(name: String, testRequirements: Requirement, ...)
object TestConfiguration {
  implicit val testLabel: TestLabel[TestConfiguration] = _.name
}

abstract class MyAbstractRunnableSpec extends AbstractRunnableSpec {
  override type Label = TestConfiguration
  override val testLabel = implicitly
  def spec = suite("x")(
    test(TestConfiguration("my-custom-test", Requirement(...)))(assertCompletes),
  )
}
```

adamgfraser:
@neko-kai You're right, test annotations allow you to annotate the results of a test as you are executing it but you want to have this information statically before you start executing tests. The label type is the right vehicle for this, I think. And having a guarantee that there is a meaningful way to render a label to a string would also be valuable for us internally as there are some places we are relying on that, for example in supporting filtering by test label.

In addition to the type class approach, I think one additional way we could achieve this is by just defining a data type:

```scala
final case class Labeled[+A](value: A, label: String)
```

Then specs would contain labels of type `Labeled[L]` instead of `L`. Combinators such as `test` and `testM` would continue to take a string and just use it internally to construct a `Labeled[Unit]` so there would be no change in the API for other users. Whereas for distage you could do something like `Labeled[Set[LightTypeTag]]`.

I think this would be equivalent in power to the type class but potentially more idiomatic ZIO versus traditional functional programming and would be analogous to what we did with `Described` in `ZQuery`.

@jdegoes What do you think?

jdegoes:
I think the functionality is great and long overdue! I like the type-class-less solution proposed by @adamgfraser as it feels more "ZIO idiomatic".

pshirshov:
Okay, closing this then.

adamgfraser:
Okay, I will submit a PR to add a `Labeled` data type as discussed above.

Issue #2737, opened 2020-01-27
Add ZSink.sum


iravid:
Signature:
```scala
object ZSink {
  val sum: Sink[Any, Nothing, Nothing, Long, Long]
}
```

Or possibly use `scala.math.Numeric` for defining the input type.

kolemannix:
I would like to tackle this.

In addition, what about a `count` sink as well that simply counts the elements it receives and emits a `Long`?

iravid:
All yours @kolemannix!

kolemannix:
Trying with this signature
```scala
def sum[A](implicit ev: Numeric[A]): ZSink[Any, Nothing, Nothing, A, A]
```

There are some inconveniences to the caller
```scala
Stream.fromIterable(1 to 10).run(ZSink.sum)
```
yields the unfriendly
```
[error] /zio/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala:1248:60: ambiguous implicit values:
[error]  both object BigIntIsIntegral in object Numeric of type scala.math.Numeric.BigIntIsIntegral.type
[error]  and object ShortIsIntegral in object Numeric of type scala.math.Numeric.ShortIsIntegral.type
[error]  match expected type Numeric[A]
[error]               assertM(Stream.fromIterable(range).run(ZSink.sum))(equalTo(range.sum))
[error]                                                            ^
```
In my experience this happens whenever one tries to be generic over Numeric. Is this satisfactory? It is resolved by specifying `ZSink.sum[Int]`. I don't believe overloading is the answer here, but hardcoding to `Long -> Long` seems a bit narrow.

Being generic also means using a `def`, which means we're instantiating a new `PureSink` on each call, which may or may not matter for performance.

Thoughts?

vasilmkd:
I wouldn't worry about the `def`. It's also ok for the type to be needed to be specified.

iravid:
We can add a shortcut on ZStream that will ease this issue:
```scala
def runSum(implicit A: Numeric[A]): ZIO[R, E, A] = run(Sink.sum[A])
```

regiskuckaertz:
Looks like this is done.

Issue #2736, opened 2020-01-27
Add ZSink.count, ZStream#runCount


iravid:
Signatures:
```scala
object ZSink {
  val count: Sink[Any, Nothing, Nothing, Any, Long]
}

class ZStream[R, E, A] {
  def runCount: ZIO[R, E, Long] = run(ZSink.count)
}
```

kolemannix:
I'll give this one a shot as well

regiskuckaertz:
Looks like this is done.

PR #2756, opened 2020-01-29
Add ZSink#sum and ZSink#count (#2736 #2737)

kolemannix:


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1302,6 +1302,19 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
         p(a).map(if (_) ((a :: as, true), Chunk.empty) else ((as, false), Chunk.single(a)))
     }.map(_._1.reverse)
 
+  /**
+   * Creates a sink which emits the number of elements processed
+   */
+  def count[A]: ZSink[Any, Nothing, Nothing, A, Long] = {
+    new SinkPure[Nothing, Nothing, A, Long] {
```
iravid: To cut down on repetition, let's implement this with `Sink.foldLeft`?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1835,6 +1848,20 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
       def cont(state: State)           = false
     }
 
+  /**
+   * Creates a sink which sums elements, provided they are Numeric
+   */
+  def sum[A](implicit ev: Numeric[A]): ZSink[Any, Nothing, Nothing, A, A] = {
```
iravid: To cut down on repetition, let's implement this with `Sink.foldLeft`?

streams/shared/src/main/scala/zio/stream/ZSink.scala
kolemannix: 👍

streams/shared/src/main/scala/zio/stream/ZSink.scala
kolemannix: 👍

iravid:
Ah, looks like tests are failing on Dotty. Could you check that?

PR #2757, opened 2020-01-29
Change Signature of Ref#update to Return UIO[Unit]

### LINKED ISSUE #2754: Should Ref#update Return Unit?
Right now the signature of `Ref.update` is:

```scala
def update(f: A => A): UIO[A] 
```

In some cases it can be useful to get the updated value back but in many cases it can be suboptimal because we instead need to return a `UIO[Unit]`, requiring calling `unit` after each update operation. For example, from `TestRandom`:

```scala
/**
 * Clears the buffer of booleans.
 */
val clearBooleans: UIO[Unit] =
  bufferState.update(_.copy(booleans = List.empty)).unit
```

From our own code bases, it appears that most usages discard the value, but there are some cases where it is used. So it seems like it would be nice to have combinators for both varieties. I had initially thought of leaving `update` the same and adding `update_` for the variant that returns `UIO[Unit]` even though I don't love the `_` suffixes. @ghostdogpr suggested changing `update` to return `UIO[Unit]` and adding something like `updateAndGet` for the less commonly used value returning variant, which also seems quite nice.

What do other people think?

adamgfraser:
Resolves #2754. `update` goes to returning `UIO[Unit]` and `updateAndGet` is implemented for returning `UIO[A]`. Makes conforming changes to `Ref`, `RefM`, `FiberRef`, and `TRef` includes `update` and `updateSome` variants.

### DIFF
--- a/core/shared/src/main/scala/zio/Ref.scala
+++ b/core/shared/src/main/scala/zio/Ref.scala
@@ -226,29 +226,6 @@ final class Ref[A] private (private val value: AtomicReference[A]) extends AnyVa
 
   /**
    * Atomically modifies the `Ref` with the specified partial function.
-   * If the function is undefined on the current value it returns the old value
-   * without changing it.
-   *
-   * @param pf partial function to atomically modify the `Ref`
-   * @return `UIO[A]` modified value of the `Ref`
-   */
-  def updateAndGetSome(pf: PartialFunction[A, A]): UIO[A] = IO.effectTotal {
-    var loop    = true
-    var next: A = null.asInstanceOf[A]
-
-    while (loop) {
-      val current = value.get
-
-      next = pf.applyOrElse(current, (_: A) => current)
-
-      loop = !value.compareAndSet(current, next)
-    }
-
-    next
-  }
-
-  /**
-   * Atomically modifies the `Ref` with the specified partial function.
    * If the function is undefined on the current value it doesn't change it.
    *
    * @param pf partial function to atomically modify the `Ref`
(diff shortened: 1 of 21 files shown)


core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -117,9 +117,32 @@ final class Ref[A] private (private val value: AtomicReference[A]) extends AnyVa
...
+  /**
+   * Atomically modifies the `Ref` with the specified function and returns the
+   * updated value. This is not implemented in terms of `modify` purely for
+   * performance reasons.
+   *
+   * @param f function to atomically modify the `Ref`
    * @return `UIO[A]` modified value of the `Ref`
    */
-  def update(f: A => A): UIO[A] = IO.effectTotal {
+  def updateAndGet(f: A => A): UIO[A] = IO.effectTotal {
```
iravid: While you're here, do you think it's worth adding `getAndUpdate` too?

adamgfraser: Yes will do.

adamgfraser: It could be nice to have `getAndSet` too.

core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -136,12 +226,13 @@ final class Ref[A] private (private val value: AtomicReference[A]) extends AnyVa
...
    * Atomically modifies the `Ref` with the specified partial function.
-   * if the function is undefined in the current value it returns the old value without changing it.
+   * If the function is undefined on the current value it returns the old value
+   * without changing it.
    *
    * @param pf partial function to atomically modify the `Ref`
    * @return `UIO[A]` modified value of the `Ref`
    */
-  def updateSome(pf: PartialFunction[A, A]): UIO[A] = IO.effectTotal {
+  def updateAndGetSome(pf: PartialFunction[A, A]): UIO[A] = IO.effectTotal {
```
ghostdogpr: How about `updateSomeAndGet`? It feels more logical.

adamgfraser: Sure.

PR #2765, opened 2020-01-30
ZIO Test: Implement Label

adamgfraser:
Implements a label data type to support polymorphic labels while providing a static guarantee that all labels will have a meaningful string rendering.

### DIFF
--- a/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
+++ b/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
@@ -43,13 +43,13 @@ class ZTestJUnitRunner(klass: Class[_]) extends Runner with Filterable with Defa
     ): URIO[R, Unit] =
       spec.caseValue match {
         case SuiteCase(label, specs, _) =>
-          val suiteDesc = Description.createSuiteDescription(label.toString, path.mkString(":"))
+          val suiteDesc = Description.createSuiteDescription(label.render, path.mkString(":"))
           effectTotal(description.addChild(suiteDesc)) *>
             specs
-              .flatMap(ZIO.foreach(_)(traverse(_, suiteDesc, path :+ label.toString)))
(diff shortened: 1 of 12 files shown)


adamgfraser:
Copying @pshirshov.

pshirshov:
I'm okay with it. Though I think typeclass was better - it took  lot less code to implement.

adamgfraser:
@neko-kai @pshirshov Could you please take a look at the new tagging functionality we implemented in #2769? I believe with these changes test annotations can now capture the static metadata you are looking for and provide an alternative to this PR.

neko-kai:
@adamgfraser Removing type information by putting everything into a dynamically typed TestAnnotationMap looks like a straight downgrade to me tbh

adamgfraser:
@neko-kai I think the solution is type safe.

```scala
trait LightTypeTag

val typeTagged: TestAnnotation[Set[LightTypeTag]] =
  ???

def typeTag(
  typeTag: LightTypeTag,
  typeTags: LightTypeTag*
): TestAspectAtLeastR[Annotations] =
  ???
```

You can annotate tests use `typeTag` and access the annotations for each test using `annotations.get(typeTagged)`. You can't call `typeTag` with values that aren't type tags and if you call `get(typeTagged)` you are guaranteed to get back a set of type tags.

neko-kai:
@adamgfraser Unlike with a `type Label` member, you can't _force_ all specs to contain metadata. But yeah, I agree, the default value makes it safe, it just makes it easier to add dependency-less tests, which was possible before anyway and it makes it possible to reuse `zio.test._` syntax for dependency-less tests.

jdegoes:
>  Unlike with a type Label member, you can't force all specs to contain metadata.

For homogeneous sets, lists, etc., it's the same effect. I think you'd need a type-level list to actually see a difference in terms of "static guarantees".

jdegoes:
Closed in favor of #2769, but can re-open later if it is deemed that won't work.

PR #2768, opened 2020-01-31
Revert Change to ZIO#tapError

adamgfraser:
Reverts changes in #2689 to make `tapError` and `tapCause` fail with a `Cause.then` if both the original effect and the tap effect fails, since this can lead to issues with user ergonomics around handling the two errors. The contract goes back to that if the tap effect fails the effect will fail with the result of the tap effect, not the original effect. I do think this is somewhat counterintuitive, particularly for `tapCause`. One possibility would be we remove `tapCause` entirely and say that dealing with fiber failures should be handled more explicitly using combinators such as `foldCauseM`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2564,13 +2564,13 @@ object ZIOSpec extends ZIOBaseSpec {
       }
     ),
     suite("tapCause")(
-      testM("does not lose infomrmation") {
-        val causes = Gen.causes(Gen.anyString, Gen.throwable)
-        checkM(causes, causes) { (c1, c2) =>
-          for {
-            exit <- ZIO.haltNow(c1).tapCause(_ => ZIO.halt(c2)).run
-          } yield assert(exit)(failsCause(equalTo(Cause.Then(c1, c2))))
-        }
+      testM("effectually peeks at the cause of the failure of this effect") {
+        for {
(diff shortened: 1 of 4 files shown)


regiskuckaertz:
What if we changed `Then` and `Both` to something like:
```scala
case class Then[+E](...) extends Cause[::[E]]
```
then combinators that can produce more than one error would say so in the type, as in:
```scala
def tapError[E1 >: E](E => ZIO[R, E1, Any]): ZIO[R, ::[E1], A]
```
And now as a user you can't ignore the fact that there may be more than one error.

iravid:
@adamgfraser 👍 on the tapError change, but can't say I'm sure what's the right thing to do re `tapCause`. I can see arguments both ways:
- Replace the previous error with the current error - consistent with tapError and other combinators;
- Concatenate with `Then` - lossless; caters to potential users of `tapCause` who are most likely expert users / library writers.

I think consistency trumps expressiveness in this case since power users can always fall back to `foldCauseM`.

@regiskuckaertz 
I think that's interesting but have doubts about the ergonomics. Possibly should be deferred to after 1.0?

regiskuckaertz:
@iravid yeah totally. I'll play around with the idea and see where it leads me.

adamgfraser:
@iravid I agree with you, though I do somewhat worry that users who are not power users will accidentally recover from unchecked exceptions with `tapCause`. The other possibility would just be to remove it entirely. It was only added last week and is used once in the codebase. Otherwise I think this is good to go unless we want to get more opinions?

PR #2769, opened 2020-01-31
ZIO Test: Implement Tagged Annotation

adamgfraser:
Allows tagging tests with string tags. Example output:

```
[info]   + monad laws - tagged: "laws", "monad"
[info]     + monad left identity - tagged: "laws", "monad"
[info]     + monad right identity - tagged: "laws", "monad"
[info]     + monad associativity - tagged: "laws", "monad"
```

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -40,14 +40,24 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
     aspect(self.asInstanceOf[ZSpec[R1, E2, L, S]])
 
   /**
+   * Annotates each test in this spec with the specified test annotation.
+   */
+  final def annotate[V](key: TestAnnotation[V], value: V): Spec[R, E, L, T] =
+    transform[R, E, L, T] {
+      case c @ SuiteCase(_, _, _) => c
+      case TestCase(label, test, annotations) =>
+        Spec.TestCase(label, test, annotations.annotate(key, value))
+    }
+
+  /**
    * Returns a new spec with the annotation map at each node.
    */
   final def annotated: Spec[R with Annotations, Annotated[E], L, Annotated[T]] =
     transform[R with Annotations, Annotated[E], L, Annotated[T]] {
       case Spec.SuiteCase(label, specs, exec) =>
         Spec.SuiteCase(label, specs.mapError((_, TestAnnotationMap.empty)), exec)
-      case Spec.TestCase(label, test) =>
-        Spec.TestCase(label, Annotations.withAnnotation(test))
+      case Spec.TestCase(label, test, annotations) =>
+        Spec.TestCase(label, Annotations.withAnnotation(test), annotations)
     }
 
   /**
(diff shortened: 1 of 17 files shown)


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -575,6 +575,12 @@ object TestAspect extends TimeoutVariants {
         }
     }
 
+  /**
+   * Annotates tests with string tags.
+   */
+  def tag(tags: String*): TestAspectAtLeastR[Annotations] =
```
jdegoes:
```suggestion
  def tag(tag: String, tags: String*): TestAspectAtLeastR[Annotations] =
```

jdegoes:
Looks great! Would also be nice to support an option from sbt, like now we have `-t` to run a specific test, maybe we'd have `-tags interruption` to run all tests tagged with `interruption`.

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Great suggestion!

adamgfraser:
I think it may be challenging for us to stop a test from running with test annotations. The annotation is added as part of the effectual test (in this case immediately before the original test effect). So by the time we get to the SBT test runner preparing to execute the tests is just has a monolothic test effect where before executing it there are no annotations and after starting to execute it the test is already in motion.

So I see two ways we could approach this. First, we could have the test aspect that adds the annotation immediately check some global state to see if there is a command line argument about tags and if so ignore the test if it doesn’t contain the right tag. However, I’m not sure how we would do this since the test aspect doesn’t have access to anything other than the spec, which doesn’t have this type of configuration information.

The second way would be to build the tags functionality more on the label PR that we have outstanding. We could modify a label to in addition to having a `render` method have a `tags` field. The label exists statically before we start executing any tests so it would be easy for the test runner to look at that and filter out tests based on it like our existing filtering functionality.

jdegoes:
Hmmm. Let me look. I'm interested in what what sort of changes would make it possible to do this using test annotations... e.g. what if every test stored a test annotation map? Then that could be the initial per test map. Could solve labeling problem too.

adamgfraser:
Right now only executed specs have test annotation maps. We could have all specs include them, but I don't actually think that would be sufficient to solve the problem. Modifying the annotation map is an effect and test annotations don't have the ability to perform effects on their own, they can only modify the effects of the tests.

jdegoes:
@adamgfraser 

I was thinking something like this:


```scala
final case class TestCase[-R, +E, +L, +T](
  label: L, test: ZIO[R, E, T], annotations: TestAnnotationMap) extends SpecCase[R, E, L, T, Nothing]
```

where `annotations` is initially set to `TestAnnotationMap.empty`. But aspects could augment it with other "starting" information, such as tags, prior to a run.

In the normal flow, the test runner would run each test with the `annotations` map, and then, replace the value with the new (effectfully computed) `annotations` map following the run.

Could that work?

adamgfraser:
@jdegoes I think so. Let me work on that.

adamgfraser:

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -111,16 +121,46 @@ final case class Spec[-R, +E, +L, +T](caseValue: SpecCase[R, E, L, T, Spec[R, E,
...
-      case TestCase(label, test) =>
-        if (f(label)) Some(Spec.test(label, test)) else None
+      case TestCase(label, test, annotations) =>
+        if (f(label)) Some(Spec.test(label, test, annotations)) else None
+    }
+  }
+
+  /**
+   * Returns a new spec with only those suites and tests with tags satisfying
+   * the specified predicate. If not tags satisfy the specified predicate then
```
jdegoes:
```suggestion
   * the specified predicate. If no tags satisfy the specified predicate then
```

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -110,7 +110,7 @@ package object test extends CompileVariants {
   /**
    * An `ExecutedSpec` is a spec that has been run to produce test results.
    */
-  type ExecutedSpec[+E, +L, +S] = Spec[Any, Nothing, L, Annotated[ExecutedResult[E, S]]]
+  type ExecutedSpec[+E, +L, +S] = Spec[Any, Nothing, L, ExecutedResult[E, S]]
```
jdegoes: This is a nice simplification.

PR #2770, opened 2020-01-31
Implement ZIO#once

adamgfraser:
Sometimes we want to know that an effect will only be evaluated a single time (e.g. a finalizer that could be subject to early release). `once` takes an effect and returns a new effect that will only be executed once, even if it is evaluated multiple times.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -776,11 +776,53 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     self.lock(Executor.fromExecutionContext(Int.MaxValue)(ec))
 
   /**
-   * Returns an effect that will only be executed once, even if it is evalauted
-   * multiple times.
+   * Returns an effect that, if evaluated, will return the result of this
+   * effect a single time. If the result of this effect has already been
+   * returned the specified fallback value will be returned. Uses the specified
+   * function to determine whether to retry this effect on subsequent
+   * evaluations if the effect fails with an error or is interrupted.
    */
-  final def once: ZIO[Any, Nothing, ZIO[R, E, Unit]] =
-    Ref.make(true).map(ref => self.whenM(ref.getAndSet(false)))
+  final def once[A1 >: A](fallback: => A1): UIO[ZIO[R, E, A1]] =
+    onceWith(fallback)(_ => true)
+
+  /**
+   * Returns an effect that, if evaluated, will return the result of this
+   * effect a single time. If the result of this effect has already been
+   * returned the result of the specified fallback effect will be returned.
+   * If this effect fails with an error or is interrupted subsequent
+   * evaluations will retry the effect.
+   */
+  final def onceM[R1 <: R, E1 >: E, A1 >: A](fallback: ZIO[R1, E1, A1]): UIO[ZIO[R1, E1, A1]] =
+    onceWithM(fallback)(_ => true)
+
+  /**
+   * Returns an effect that, if evaluated, will return the result of this
+   * effect a single time. If the result of this effect has already been
+   * returned the specified fallback value will be returned. If this effect
+   * fails with an error or is interrupted subsequent evaluations will retry
+   * the effect.
+   */
+  final def onceWith[A1 >: A](fallback: => A1)(f: Cause[E] => Boolean): UIO[ZIO[R, E, A1]] =
+    onceWithM(ZIO.succeed(fallback))(f)
+
+  /**
+   * Returns an effect that, if evaluated, will return the result of this
+   * effect a single time. If the result of this effect has already been
+   * returned the result of the specified fallback effect will be returned.
+   * Uses the specified function to determine whether to retry this effect on
+   * subsequent evaluations if the effect fails with an error or is
+   * interrupted.
+   */
+  final def onceWithM[R1 <: R, E1 >: E, A1 >: A](
+    fallback: ZIO[R1, E1, A1]
+  )(f: Cause[E] => Boolean): UIO[ZIO[R1, E1, A1]] =
+    Ref.make(true).map { ref =>
+      ZIO
+        .bracketExit(ref.getAndSet(false))((_, exit: Exit[E, Option[A]]) =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -775,6 +775,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def on(ec: ExecutionContext): ZIO[R, E, A] =
     self.lock(Executor.fromExecutionContext(Int.MaxValue)(ec))
 
+  /**
+   * Returns an effect that will only be executed once, even if it is evalauted
+   * multiple times.
+   */
+  final def once: ZIO[Any, Nothing, ZIO[R, E, Unit]] =
+    Ref.make(true).map(ref => self.whenM(ref.getAndSet(false)))
```
iravid: The semantics here currently are "execute at most once", since errors / interruption will leave the flag set but `self` was not necessarily performed. Maybe we should use `bracketExit` here?

Also, it'd be nice to make this return the value `A` with a fallback value/effect when the ref is already set. This is the prevalent pattern in ZStream.

regiskuckaertz:
Isn't it what `memoize` is doing already?

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I'm not sure what the right semantics are here. My original intention was "guarantee that this effect will only begin execution once" so you could pass the effect around and get the same semantics as if you only evaluated it once (and it potentially was interrupted or failed). Resetting the flag raises the possibility that the "work" will be done multiple times if there are failures or the effect is interrupted after doing significant work. We could definitely generalize by allowing the caller to provide a `Exit[E, A] => Boolean` function and a default value, thought it does complicate the method signature.

adamgfraser:
`memoize` does a little more bookkeeping internally because it has to use two promises to return the lazily computed result, so my original thought was that this would be a more lightweight combinator when you wanted something to be executed only and didn't care about the result but I agree they are similar. I'm happy to close if people don't think this is valuable.

iravid:
@adamgfraser I think this would be pretty useful. If this combinator would allow specifying a fallback value/effect, it'd be sufficiently different from `memoize`.

Possible signatures:
```
def once[A1 >: A](fallback: A1): URIO[R, ZIO[R, E, A1]]
def onceM[R1 <: R, E1 >: E, A1 >: A](fallback: ZIO[R1, E1, A1]): URIO[R1, ZIO[R1, E1, A1]]
```

core/shared/src/main/scala/zio/ZIO.scala
iravid: `onceWith/M` sounds like a great variant :-)

adamgfraser:
@iravid Updated along the lines you suggested. Not sure we are there yet. A couple of observations:

1. I think we can factor out the `fallback` and just have the effect return an `Option[A]` with `Some` on the first completion and `None` thereafter. Then the caller can compose to add their own logic to handle the `None` case. This could let us simplify the method somewhat.
2. Not sure what the right semantics are for concurrent evaluation. If Fiber A starts evaluating it will begin running the effect, then if Fiber B tries to evaluate while Fiber A is still running the effect, Fiber B will get the fallback value. Then if Fiber A fails and Fiber C starts evaluating it will begin running the effect again. Not sure if that is right or if Fiber B should suspend waiting for result of A.
3. Still trying to figure out if we can make the logic about failure handling more composable.

jdegoes:
I hadn't thought about it but the basic version of `once` is actually the same as `memoize`. Hmm. 

I think fallback can be done without new things, e.g. `(a orElse b).memoize`, or `a.memoize(_ orElse b)`, depending on the desired semantic.

adamgfraser:
Yes, originally it was intended to just be a slightly more efficient version of `memoize` when you don't need the result because you can avoid using the two promises. There are a lot of places you want to make sure like a finalizer is only run once but we generally aren't using `memoize` for those. But maybe that is not enough to justify a new combinator.

Definitely agree about being able to remove fallback. At that point is there enough here with the logic for evaluating again in the event of failure or interruption? I find the semantics of it somewhat confusing but am happy to push forward if others think it would be useful or happy to close.

regiskuckaertz:
I feel like we have well over 100 functions in both the trait and object and starting to reach the event horizon after which adding a new one really has to pay off in terms of usability. wdyt?

adamgfraser:
I'm fine with that.

I try to think about whether there is a way to re-use functionality when I am doing something and I noticed a decent number of places where we want to create a finalizer that will only be executed once (e.g. if a resource can be released early) so we create a `Ref` to guard the effect. And seems like we weren't using `memoize` there, maybe because we just weren't thinking about it or maybe because creating two promises versus one `Ref` would be wasteful. So that was the original motivation.

But I agree that we have a ton of combinators and it isn't like a burning need so I'm fine with closing if other people aren't feeling it.

iravid:
This could simplify a bunch of implementations for `ZStream` constructors (`fromEffect`, `succeed`, etc.) but indeed can also be just implemented as a private utility there.

adamgfraser:
If we feel like it could be useful for us I think that potentially indicates that it could be valuable to others as well. We're probably especially heavy users of finalization logic but I think other people are probably going to have to the same thing sometimes. I hear the point about method proliferation but not sure this is the place to draw the line.

regiskuckaertz:
I hear you both, my only worry is the inevitability of a crowded API where the semantic differences between similar combinators with very different names is so small it hurts usability. I'm probably exaggerating as this is an exception.

adamgfraser:
I hear you too. There are definitely areas where we could rationalize more (see #2900) and in general I think it is the right concern, especially now. But I do think the fact that we never use the "similar" method for this in practice is telling.

adamgfraser:
How about we revert this back to the simple `once` variant in the first version of this PR? That is only a single method so I don't think it adds much to the cognitive burden. I just wanted this in #2942 and I don't think we would be comfortable with using `memoize` in its place because of the additional overhead.

jdegoes:
👍

jdegoes:
Maybe we can think about interning memoize so it is fast enough to replace once some day?

adamgfraser:
Would love to work on that!

PR #2774, opened 2020-01-31
Implement ZIO#ifM

adamgfraser:
Just makes it more convenient to execute conditional logic with effects.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2560,6 +2560,12 @@ object ZIO {
   def identity[R]: URIO[R, R] = fromFunction[R, R](ZIO.identityFn[R])
 
   /**
+   * Runs `onTrue` if the result of `b` is `true` and `onFalse` otherwise.
+   */
+  def ifM[R, E](b: ZIO[R, E, Boolean]): ZIO.IfM[R, E] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2559,6 +2559,12 @@ object ZIO {
    */
   def identity[R]: URIO[R, R] = fromFunction[R, R](ZIO.identityFn[R])
 
+  /**
+   * Runs `onTrue` if the result of `b` is `true` and `onFalse` otherwise.
+   */
+  def ifM[R, E](b: ZIO[R, E, Boolean]): ZIO.IfM[R, E] =
+    new ZIO.IfM(b)
```
fsvehla: Just out of curiosity — why is this pattern used vs. inlining `IfM.apply?`

adamgfraser: For improved type inference. If we don't do this then the compiler will fix the type parameters based on the first parameter list in a curried function. So if the `b` effect was something like `ref.get` the compiler would infer `R` as `Any` and `E` as `Nothing`. Then if our actual `onTrue` and `onFalse` effects required some environment or could fail the compiler would complain and we would have to manually specify type parameters and then people would be sad.

Issue #2772, opened 2020-01-31
Support Memoizing Layers


adamgfraser:
We should implement a method to memoize a layer. This is important to allow a layer to be created once and provided to multiple services. This issue is currently impacting ZIO Test because we provide a separate copy of the test environment to each test and the test environment depends on the live environment, so each time we create a new test we create a new copy of the live environment, including allocating resources like a scheduler.

Thanks to @vasilmkd for identifying the issues through profiling of our internal test suites.

I will take the ticket.

vasilmkd:
![zio-timers](https://user-images.githubusercontent.com/7115459/73545815-67e4dd00-443c-11ea-966c-a3da32353b2e.PNG)
![zio-memory-leak](https://user-images.githubusercontent.com/7115459/73545819-6a473700-443c-11ea-9e8d-c4764675d9b0.PNG)
![zio-heap-full](https://user-images.githubusercontent.com/7115459/73545822-6c10fa80-443c-11ea-9155-8e753ccaf202.PNG)

@adamgfraser Feel free to tag anyone here that might be interested in these.

vasilmkd:
I'm also uploading the whole heap (4 Gb), I will update this comment when available.

adamgfraser:
Following up on this, the `default` environments are generally `val` definitions so I don't think this problem is as bad as I feared, though this will still be an important feature for more advanced use cases with layers that do have meaningful costs to acquire and release and need to be shared among multiple dependencies.

jdegoes:
I have some feedback on this. I will write it up.

adamgfraser:
That would be great! I think we have tracked down the cause of the performance problems in our internal test suite to a particular test suite (will have a fix for that tomorrow) so we can focus on the right long term answer here.

jdegoes:
@adamgfraser I too have been thinking about memoization. Not for performance, per se, but rather, to avoid construction of multiple implementations of the same resource.

Right now, if you use the same layer twice, e.g. `foo >>> bar`, `foo >>> baz`, then `foo` resources will be allocated twice. Now a simple `ZLayer#memoize` could let you solve that problem with, `foo.memoize.flatMap { foo => >... }`, but it feels less ergonomic.

Piggybacking on your idea of ZLayers as recipes, it should be possible to add memoize directly, such that `ZLayer#memoize` would return another `ZLayer`. I'll show how shortly...

adamgfraser:
@jdegoes Love it!

jdegoes:
[...]

PR #2775, opened 2020-01-31
Support Memoizing Layers

adamgfraser:
Initial work on resolving #2772. The challenge is not really memoizing a layer, as we can already get a managed out of a layer and memoize that, but the change in architecture in no longer having any "global" environment and the fact that memoization gives us an effect wrapped in an effect. Currently this PR just changes the way the test environment is constructed in ZIO Test to memoize the live environment.

I am wondering if we can build on the idea of layers as "recipes" for environments to allow memoizing layers in a referentially transparent way without creating an additional effect layer. If we prevent you from just getting any managed out of a layer and make `build` equivalent to converting the recipe to the environment in some sense then I think we could potentially make `memoize` just change the "recipe" for building that layer, though that needs to be worked through more.


adamgfraser:
@jdegoes Initial implementation of shared layers is here. This needs some cleanup but seems to be working relatively well, with common dependencies across different layers only creating once and resources being acquired and released in the correct order.

The main issue I see right now is that since we are using layers as the keys in the `MemoMap` we need a way to has them. Right now for this demonstration I am relying on object equality but I don't think that is an acceptable solution. We probably need some kind of concept of labels when creating layers, and then we can have logic for composing those labels. It isn't great to require users to do one more thing to create a layer but relative to the previous ceremony seems manageable. Also potentially provides an easy solution to how to create resources that aren't shared, just give them a different identifier.

jdegoes:
@adamgfraser 

I'll review this today. High-level thoughts:

1. I think piggybacking on object equality/hash code isn't terrible, even if it does weakly violate referential transparency (weakly because for sanely-defined layers, it will have no effect on behavior, only performance, which is ordinarily exempted from referential transparency). The only place it may give unintuitive results is if someone dynamically makes a layer, e.g. `database(770, "localhost") >>> (...) ++ database(770, "localhost") >>> (...)`. But users can solve that problem by factoring out the duplication, e.g. `val db = database(770, "localhost")`. The problem with identifiers is that we can't know two layers with the same identifier actually represent the same layer; whereas we do know that two layers with the same object identity represent the same layer (since it is literally the same object). And when layers are distributed across a 2-3 million LOC application, the odds of accidental collisions is quite high, and could lead to painful and protracted debugging process.
2. I was originally thinking `shared` would be a method on `ZLayer` but after looking at this PR, in which sharing is the default, I think that makes _more_ sense. We can create `ZLayer#fresh` to make a fresh layer, if for some reason it's important to not share. The folks at Septimal Mind have not found a use case for "non-shared" common dependencies, although I can think of some (e.g. some service creates a thread pool and you need two isolated thread pools in different parts of your application for work isolation).

adamgfraser:
@jdegoes Great! I am most of the way through moving `ZLayer` over to the new encoding and will clean this up as part of that so will do that this morning and then your review can be more efficient.

adamgfraser:

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -49,47 +61,79 @@ final case class ZLayer[-RIn, +E, +ROut <: Has[_]](value: ZManaged[RIn, E, ROut]
...
    */
-  def build[RIn2 <: RIn](implicit ev: Any =:= RIn2): Managed[E, ROut] = value.provide(ev(()))
+  def fresh: ZLayer[RIn, E, ROut] =
+    new ZLayer(self.scope)
+
+  /**
+   * Returns an effect that, if evaluated, will return the lazily computed
+   * result of this layer.
+   */
+  def memoize: ZManaged[RIn, Nothing, ZLayer[Any, E, ROut]] =
```
jdegoes: Is this useful anymore? I think we can delete it unless you have a use case in mind.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -115,13 +159,13 @@ object ZLayer {
   /**
    * Constructs a layer that purely depends on the specified service.
    */
-  def fromService[A: Tagged, E, B <: Has[_]](f: A => B): ZLayer[Has[A], E, B] =
+  def fromService[A: Tagged, B <: Has[_]](f: A => B): ZLayer[Has[A], Nothing, B] =
```
jdegoes: Nice cleanups

core/shared/src/main/scala/zio/ZLayer.scala
adamgfraser: Yes I think that is right.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -341,4 +385,63 @@ object ZLayer {
...
+     * Checks the memo map to see if a dependency exists. If it is, immediately
+     * returns it. Otherwise, obtains the dependency, stores it in the memo map,
+     * and adds a finalizer to the outer `Managed`.
+     */
+    final def getOrElseMemoize[E, A, B <: Has[_]](
+      layer: ZLayer[A, E, B],
+      finalizerRef: Ref[Exit[Any, Any] => ZIO[Any, Nothing, Any]]
+    ): ZManaged[A, E, B] =
+      ZManaged {
+        get(layer).flatMap {
```
jdegoes: Is this safe for arbitrary interruption, or would it be useful to add `uninterruptibleMask` to protect critical regions, and allow interruption only some places (e.g. `reservation.acquire`?)?

adamgfraser: So I think we are okay here but it is definitely the right question. My analysis is:

1. First we check if the dependency is in the map and if so just return it. I don't think there is any problem with being interrupted here because we haven't acquired any resources at this point that require finalization.
2. If we don't have the resource yet, we do `layer.scope.flatMap(_.apply(self))` which creates one Managed that describes entering both the inner and outer scopes to create the dependency.
3. Then we call `reserve` on it to access the `acquire` and `release` actions of the managed separately but we still haven't actually acquired any resources.
4. The actual acquisition of resources comes in `reservation.acquire` which is in this reservation we are constructing that will create a new `Managed`, so the acquisition of the dependency and adding it to the memo map will be uninterruptible and if it starts the finalizers are guaranteed to be added to the outer scope, also uninterruptibly, because that is in the release action of the managed.

jdegoes: It's tricky, but I think you're right..

jdegoes:
@adamgfraser 

I don't know how Lint failed in master but didn't in the PR:

<https://github.com/zio/zio/commit/5f99983b18ddc16381f3c1198d599773542eca60>

adamgfraser:
@jdegoes The issue is when we implemented `provideSomeLayer` in the other PR we added a variant to streams that didn't exist on this branch, so when we merged we had a compilation issue because that variant hadn't been updated to the new encoding. Anyway, it is fixed now.

PR #2777, opened 2020-01-31
Implement ZSTM#tap

mijicd:
`tapError` and `tapBoth` to follow in subsequent PRs.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -315,7 +315,7 @@ final class ZSTM[-R, +E, +A] private[stm] (
    * "Peeks" at the success of transactional effect.
    */
   def tap[R1 <: R, E1 >: E](f: A => ZSTM[R1, E1, Any]): ZSTM[R1, E1, A] =
-    flatMap(f) *> self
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -311,6 +311,12 @@ final class ZSTM[-R, +E, +A] private[stm] (
       }
     )
 
+  /**
+   * "Peeks" at the success of transactional effect.
+   */
+  def tap[R1 <: R, E1 >: E](f: A => ZSTM[R1, E1, Any]): ZSTM[R1, E1, A] =
+    flatMap(f) *> self
```
fsvehla: If this would be an effect it would be evalulated twice, right?
Is this OK since it’s STM?

adamgfraser: Good point. I think this should should be `flatMap(a => f(a).as(a))`.

mijicd: Yes, just started writing that one. One sec :)

mijicd: Done, good catch!

PR #2779, opened 2020-02-01
Fix Test Using Excessive Resources

adamgfraser:
@vasilmkd and I have been observing a slowdown in a long running SBT session when running our internal test suites multiple times. We believe we have tracked it down to a particular suite in the ZIO Test test suite. The test is more of an "integration" test that creates a test copy of the test executor and uses it to execute a test suite and test the results. However, the tests are currently being run as property based tests, involving doing this 600 times and seeming to create excessive garbage collection pressure. This is a simple test and one that we haven't had any issues with, so there is no need to do this and we can just switch to a single test for each case.

@vasilmkd Can you please confirm that this fixes the issue?

Thanks to @vasilmkd for all his great work in tracking this down!


vasilmkd:
I call this a great success! Here's the proof.

Screen Shot 2020-02-01 at 15 13 33

vasilmkd:
Feel free to delete the .jvmopts file.

vasilmkd:
Running without .jvmopts. Amazing. I'm so happy.

Screen Shot 2020-02-01 at 15 29 32

vasilmkd:
Excited to see the effect on CI.

PR #2783, opened 2020-02-02
Enable Stream Tests on ScalaJS

### LINKED ISSUE #2069: ZIO Stream: Enable Tests on ScalaJS
Currently all tests for the ZIO Streams project are in the `streams-tests/jvm` folder and thus only being run on the JVM. This ticket is to move all of those tests to the `shared` folder so that they run on both the JVM and ScalaJS and resolve any compilation or test failures that result from running them.

adamgfraser:
Resolves #2069. Moves all streams tests to the `shared` folder so they run on both the JVM and ScalaJS with the exception of the test for `ZSink.fromOutputStream`, which is only implemented on the JVM. I had to disable the following tests on ScalaJS:

1. StreamEffecyAsyncSpec - `effectAsync` uses `unsufeRun` internally which is synchronous and is triggering an error that we can't block on ScalaJS. I think the issue may be that we are doing too many continuations and triggering a yield. We may be able to use an unyielding executor internally in the implementation to resolve.
2. StreamChunkSpec - tests `mapAccum` and `mapAccumM happy path` are failing on ScalaJS with a class cast exception that a number can't be cast to a byte. This appears to be related to how ScalaJS represents numeric values and a value that is small enough to be a byte automatically being treated as a byte. The tests pass if I increase the starting value to a number that is too large to be a byte. See https://www.scala-js.org/doc/semantics.html.
3. StreamSpec - `fromQueue` is timing out on ScalaJS. Not sure what the issue is here.

We may want to merge even with these tests ignored and open separate issues as we are still dramatically increasing test coverage from today. We should also watch CI times. The JS builds are already taking significantly longer than the JVM builds even without any stream tests being run so we may want to use some version specific settings (e.g. reducing the number of tests for property based generators) to keep build times in line.


vasilmkd:
@adamgfraser Very cool, thank you. I'll try to take a look at this these days. It's something that I want to see done. :pray:

adamgfraser:
Closing pending addition of global configurable parameter for number of property based tests.

PR #2789, opened 2020-02-03
Implement orElse Variants

### LINKED ISSUE #2787: ZIO Core: Add orElse variants
```scala
trait ZIO[-R, +E, +A] {
  def orElseFail[E1 >: E](e1: => E1): ZIO[R, E1, A] = ???
  def orElseSucceed[A1 >: A](a1: => A1): ZIO[R, E, A1] = ???
}
```

adamgfraser:
Resolves #2787. I think we can make the type signatures slightly more specific than originally specified, if I am understanding the semantics correctly. `orElseFail` will always replace the original failure with the new one, so `E1` does not have to be a super type of `E`. `orElseFail` will always replace the original failure with a success so the resulting effect cannot fail.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1042,6 +1042,24 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(orElse)(equalTo(true))
       }
     ),
+    suite("orElseFail")(
+      testM("executes this effect and returns its value if it succeeds") {
+        import zio.CanFail.canFail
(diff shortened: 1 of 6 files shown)


jdegoes:
@adamgfraser I was thinking these would be equivalent to:

```scala
a orElse (ZIO.succeed(b))
a orElse (ZIO.fail(b))
```

Hence, `orElseSucceed`, `orElseFail`.

I've found myself wanting both. In particular, I want `orElseFail`, when I have `Unit` as an error and want to unify that with something else; and want `orElseSucceed` when I have a constant value that will work in case the effect fails.

PR #2793, opened 2020-02-03
Short circuit Chunk#exists

vasilmkd:


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -173,7 +173,7 @@ sealed trait Chunk[+A] { self =>
     val len    = self.length
     var exists = false
     var i      = 0
-    while (i < len) {
+    while (!exists && i < len) {
```
regiskuckaertz: 🤦‍♂

vasilmkd: Haha, it's not a big deal really, it happens. :)

PR #2799, opened 2020-02-03
ZIO Core: Add Control Flow Combinators

### LINKED ISSUE #2786: ZIO Core: Add more control flow combinators to avoid the need for explicit recursion
The following combinators might be useful to add to `ZIO`, together with the companion objects for the ZIO type aliases (`Task`, `UIO`, `URIO`, `RIO`, `IO`).

```scala
def loop[R, E, A, S](initial: S, cont: S => Boolean, inc: S => S)(body: S => ZIO[R, E, A]): ZIO[R, E, List[A]] = ???

def loop_[R, E, S](initial: S, cont: S => Boolean, inc: S => S)(body: S => ZIO[R, E, Any]): ZIO[R, E, Unit] = ???
```

A more powerful version would be:

```scala
def iterateM[R, E, S](initial: S, cont: S => Boolean)(body: S => ZIO[R, E, S]): ZIO[R, E, S] = ???
```

#### Example Usage

```scala
ZIO.loop(0, _ < 10, _ + 1) { index =>
  putStrLn(s"Currently at index ${index}")
}
```

adamgfraser:
Resolves #2786.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -862,18 +862,6 @@ object ZSTM {
     suspend(if (p) STM.unit else retry)
 
   /**
-   * The moral equivalent of `if (p) exp`
-   */
-  def when[R, E](b: => Boolean)(stm: ZSTM[R, E, Any]): ZSTM[R, E, Unit] =
-    suspend(if (b) stm.unit else unit)
-
-  /**
-   * The moral equivalent of `if (p) exp` when `p` has side-effects
-   */
-  def whenM[R, E](b: ZSTM[R, E, Boolean])(stm: ZSTM[R, E, Any]): ZSTM[R, E, Unit] =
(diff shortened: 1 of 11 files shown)


jdegoes:
Excellent work!

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -964,6 +952,63 @@ object ZSTM {
...
+  /**
+   * Loops with the specified transactional function purely for its
+   * transactional effects. The moral equivalent of:
+   *
+   * {{{
+   * val s = initial
+   *
+   * while (cont(s)) {
+   *   body(s)
+   *   inc(s)
```
neko-kai: @adamgfraser 
should be `s = inc(s)`, `val s`->`var s` here?

adamgfraser: Yes you're right. Just submitted a PR to fix.

PR #2800, opened 2020-02-03
Implement ZIO#provideSomeLayer

### LINKED ISSUE #2794: ZIO Core: Add provideLayerSome
A hypothetical `provideLayerSome` could split `R` into two parts: the part being provided, and the remainder.

Experiment:

```scala
trait ZLayer[-A, +E, +B]

trait ZIO[-R, +E, +A] { self =>
  // def provideLayerSome[E1 >: E, R1, R2](zlayer: ZLayer[R1, E1, R2])(implicit ev: R1 with R2 <:< R): ZIO[R1, E1, A] = ???
  def provideLayerSome[Remainder]: ProvideLayerSome[R, E, A, Remainder] = new ProvideLayerSome[R, E, A, Remainder](self)
}

class ProvideLayerSome[-R, +E, +A, R1](self: ZIO[R, E, A]) {
  def apply[E1 >: E, R2](zlayer: ZLayer[R1, E1, R2])(implicit ev: R1 with R2 <:< R): ZIO[R1, E1, A] = ???
}

trait Clock
trait Random

lazy val clockLayer: ZLayer[Any, Nothing, Clock] = ???

lazy val zio: ZIO[Clock with Random, Nothing, Unit] = ???

lazy val zio2 = zio.provideLayerSome[Random](clockLayer)

```

adamgfraser:
Initial work to resolve #2794. Right now we need evidence that the remaining environment type `R0` is a subtype of `Has[_]` and that there is a type tag available for `R0`. Not sure if we can get rid of the type tag requirement, though it still seems to work okay. Needs documentation, tests, and implementation for other effect types.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -3289,7 +3289,7 @@ object ZIO {
     def apply[E1 >: E, R1 <: Has[_]](
       layer: ZLayer[R0, E1, R1]
     )(implicit ev: R0 with R1 <:< R, tagged: Tagged[R1]): ZIO[R0, E1, A] =
-      self.provideSome[R0 with R1](ev).provideLayer[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer)
+      self.provideLayer[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer)
   }
 
   implicit final class ZIOWithFilterOps[R, E, A](private val self: ZIO[R, E, A]) extends AnyVal {
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -952,6 +952,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def provideSome[R0](f: R0 => R)(implicit ev: NeedsEnv[R]): ZIO[R0, E, A] =
     ZIO.accessM(r0 => self.provide(f(r0)))
 
+  /**
+   * Splits the environment into two parts, providing one part using the
+   * specified layer and leaving the remainder `R0`.
+   *
+   * {{{
+   * val clockLayer: ZLayer[Any, Nothing, Clock] = ???
+   *
+   * val zio: ZIO[Clock with Random, Nothing, Unit] = ???
+   *
+   * val zio2 = zio.provideSomeLayer[Random](clockLayer)
+   * }}}
+   */
   final def provideSomeLayer[R0 <: Has[_]]: ZIO.ProvideSomeLayer[R0, R, E, A] =
     new ZIO.ProvideSomeLayer[R0, R, E, A](self)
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3199,6 +3202,13 @@ object ZIO {
       zio.provideSome[R2](r2 => r2.add(r1))
   }
 
+  final class ProvideSomeLayer[R0 <: Has[_], -R, +E, +A](private val self: ZIO[R, E, A]) extends AnyVal {
+    def apply[E1 >: E, R1 <: Has[_]](
+      layer: ZLayer[R0, E1, R1]
+    )(implicit ev: R0 with R1 <:< R, tagged: Tagged[R0]): ZIO[R0, E1, A] =
+      self.provideSome[R0 with R1](ev).provideLayer[E1, R0, R0 with R1](layer ++ ZLayer.identity[R0])
```
jdegoes: If you flip the order of the `++`, you can delete the `Tagged`. Only the RHS of `++` needs the `Tagged`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -952,6 +952,9 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def provideSome[R0](f: R0 => R)(implicit ev: NeedsEnv[R]): ZIO[R0, E, A] =
     ZIO.accessM(r0 => self.provide(f(r0)))
 
+  final def provideSomeLayer[R0 <: Has[_]]: ZIO.ProvideSomeLayer[R0, R, E, A] =
```
jdegoes: If this works well, would be nice to add to the usual suspects (`Spec`, `ZManaged`, `ZStream`).

adamgfraser: Yes I agree. One other thing that might be worth thinking about is our nomenclature for `provide` methods. I think we could potentially implement a similar mechanic for, say providing a managed, by converting the managed to a layer and calling `provideSomeLayer`, with the appropriate implicit evidence. So we could conceptually have three varieties of "provide", (1) provide the whole environment, (2) providing part of the environment by providing something that goes from a part to the whole environment (e.g. existing `provideSome`), and (3) providing part of the environment and leaving the remainder (e.g. new `provideSomeLayer`. And you could have variants for each different data type being provided. Maybe `provideSome` could be use for (3) and we find another name for (2)? Or do you think this should exclusively be a mechanic for layers?

jdegoes: I do think it's worth rethinking this right now, before RC18.

One could even get away from `provide` for ZLayer, favoring maybe:

 - `interpret` / `interpretSome` ("interpret from one layer to another")
 - `translate` / `translate` ("translate from one layer to another")
 - `shift` / `shiftSome` ("shift from one layer to another")

Or we could continue running with `provide` and just try to emphasize the primacy of ZLayer.

Now providing method to convert `ZManaged` into a `ZLayer` sounds good and could let us delete the `*Managed` variants without too much fuss.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I tried doing this and then we need `Tagged` for `R1`. Not sure if one is better than the other.

adamgfraser: Maybe it is better to require a tag for `R1`, the environment we are "providing", so we just prune the part we're providing versus what is leftover?

jdegoes: Yes, that's right. The user will generally know the type they are providing, so a tag will already be available for it.

adamgfraser:
Revised to require a tag for part of the environment being provided and to implement for all effect types.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3270,6 +3285,13 @@ object ZIO {
       zio.provideSome[R2](r2 => r2.add(r1))
```
adamgfraser: `provideOne` can be trivially implemented in terms of `provideSomeLayer` now. It isn't being used anywhere. We should either delete if we want layers to be the sole means of doing this or move up to the top level with new naming convention.

jdegoes: Agreed, let's delete it.

jdegoes:
@adamgfraser Looks fantastic! ❤️ This is a real pleasure to use compared to the old `provideSome`. Yes, the user has to manually specify the type of the remainder, but that's about at the limit of what we can achieve, I think.

Issue #2532, opened 2019-12-27
ZIO Test: Improve Distinction Between Test and Suite Level Resources


adamgfraser:
This came up this morning on Discord. The snippet below will acquire the resource four times, instead of three times. The issue is that with suites being effectual, a suite of tests that each require an environment of type `R` also requires an environment of type `R`, so we need to provide the environment once to the suite and then to each of the tests.

I am thinking to really address this we may need to have separate environment types for the suite level environment and the test level environment. Right now we are losing information to some extent because when we combine a collection of tests that each require an environment we automatically assume that the suite requires the same environment, which is not necessarily the case. I don't love that solution so open to other thoughts or suggestions..

```scala
import zio.{ test => _, _ }
import zio.test._
import zio.test.Assertion._

object ExampleSpec extends DefaultRunnableSpec {

  trait Resource {
    def use: Boolean = true
  }
  object Resource {
    def make: Managed[Nothing, Resource] =
      ZIO
        .effectTotal(println("Acquiring resource."))
        .as(new Resource {})
        .toManaged(_ => ZIO.effectTotal(println("Releasing resource.")))
  }

  def spec =
    suite("ExampleSpec")(
      testM("first test") {
        assertM(ZIO.access[Resource](_.use))(isTrue)
      },
      testM("second test") {
        assertM(ZIO.access[Resource](_.use))(isTrue)
      },
      testM("third test") {
        assertM(ZIO.access[Resource](_.use))(isTrue)
      }
    ).provideManaged(Resource.make)
}
```

adamgfraser:
We might be able to get away with an encoding where suites are effectual but don’t themselves have an environmental requirement, they just contain tests that may require an environment. That would give us the ability to execute effects at the suite level to use a managed resource and provide it to tests but keep one type parameter.

jdegoes:
@adamgfraser 

> We might be able to get away with an encoding where suites are effectual but don’t themselves have an environmental requirement

I think this is the right track. Then maybe we allow tests and suites to store a `ZLayer`. Then compose them all together in a logical fashion. Such that people can do `provideLayer(foo.shared)` or something. It'd be awesome if the whole "sharing" thing could be factored out of variants and pushed into `ZLayer`.

I sense #2772 is relevant.

hearnadam:
I think suites can now have layers in ZIO 2.x

PR #2803, opened 2020-02-04
ZIO Test: Implement TestAspect#verify

### LINKED ISSUE #2792: ZIO Test: Add verify TestAspect
It would be useful to have a `TestAspect` that allows verification of "post conditions" after tests are run.

```scala
def verify[R, E, A](condition: ZIO[R, E, TestResult]): TestAspect[...]
```

**Example**

```scala
myTest(...) {
} @@ verify(databaseIsReset)
```

adamgfraser:
Resolves #2792.

The each variant of `verify` is implemented for now. There is some ambiguity of what we should do if both the test and the post-condition fail. Right now if the test fails we just fail immediately and don't evaluate the post condition. If the test passes we then evaluate the post-condition and fail the test if the post-condition fails.

The more serious issue I ran into is with the all variant. To implement this we want to implement a combinator like:

```scala
trait Spec[-R, +E, +T] {
  // Models running all the tests in this spec
  final def collectAll: URIO[R, Spec[Any, E, T]] =
    caseValue match {
      case Spec.SuiteCase(label, specs, exec) =>
        specs.foldCauseM(
          c => ZIO.succeedNow(Spec.suite(label, ZIO.halt(c), exec)),
          ZIO.foreach(_)(_.collectAll).map(_.toVector).map(specs => Spec.suite(label, ZIO.succeedNow(specs), exec))
        )
      case Spec.TestCase(label, test, annotations) =>
        test.run.map(exit => Spec.test(label, ZIO.done(exit), annotations))
    }
}
```

That way we could call `collectAll` on the spec, then evaluate the post-condition, and then if necessary fail the tests in the spec. However, this violates the sandboxing of resources we're trying to achieve. We now have a spec that depends on one `R` instead of each test and suite depending on its own `R`. I'm not sure how we model this with our current representation of specs.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestAspectSpec.scala
@@ -232,6 +232,16 @@ object TestAspectSpec extends ZIOBaseSpec {
         } @@ timeout(10.milliseconds, 1.nanosecond) @@ failure(diesWith(equalTo(interruptionTimeoutFailure)))
         result <- isSuccess(spec.provideLayer(liveClock))
       } yield assert(result)(isTrue)
+    },
+    testM("verify verifies the specified post-condition after each test is run") {
+      for {
+        ref <- Ref.make(false)
+        spec = suite("verify")(
+          testM("first test")(ZIO.succeedNow(assertCompletes)),
+          testM("second test")(ref.set(true).as(assertCompletes))
+        ) @@ sequential @@ verify(assertM(ref.get)(isTrue))
+        result <- isSuccess(spec)
+      } yield assert(result)(isFalse)
     }
   )
 
--- a/test/shared/src/main/scala/zio/test/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/TestAspect.scala
@@ -593,6 +593,15 @@ object TestAspect extends TimeoutVariants {
       }
     }
 
+  /**
+   * Verifies the specified post-condition after each test is run.
+   */
+  def verify[R0, E0](condition: => ZIO[R0, E0, TestResult]): TestAspect[Nothing, R0, E0, Any] =
+    new TestAspect.PerTest[Nothing, R0, E0, Any] {
+      def perTest[R <: R0, E >: E0](test: ZIO[R, TestFailure[E], TestSuccess]): ZIO[R, TestFailure[E], TestSuccess] =
+        test <* ZTest(condition)
+    }
+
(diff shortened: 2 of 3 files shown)


jdegoes:
> Right now if the test fails we just fail immediately and don't evaluate the post condition. If the test passes we then evaluate the post-condition and fail the test if the post-condition fails.

Could also compose them with `&&`, right? Or maybe that's unadvisable due to possibility of verification failing.

> I'm not sure how we model this with our current representation of specs.

The structure of `F[Vector[G[TestResult]]]` implies that `F` comes before `G`. But what we want to do is modify the `TestResult` after all `G` have been run.

It's possible to solve that in a variety of ways (all of them some variation of storing a "postprocess" in the suite) but none of them seems clearly essential to me... maybe if there are other problems that could be solved with this machinery, it would be more obvious what it should be.

adamgfraser:
We could do a variant of `&&`. The `ZIO[R, TestFailure[E], TestSuccess]` is a slightly different representation that we haven't combined before other that composing into suites. Conceptually we have four cases (1) runtime failure, (2) assertion failure, (3) succeeded, (4) ignored. So I think here we would need a "left biased" version (where left is the test and right is the verification) and we would say:

- If both are runtime failures, return a runtime failure and combine the causes
- Otherwise if the left hand said is a runtime failure, just return that
- If they are both assertion failures combine the assertion failures with `&&`
- Otherwise return either failure
- If they are both success / ignore then return whatever the value is for the left hand side.

Not sure if that is better than just returning one or the other.

adamgfraser:
Regarding the resources, as I look at it more I think there are a couple of problems to think about:

1. Right now whenever we want to execute suite level effects we risk losing sandboxing of resources. In particular, if we try to provide part of the environment as a shared environment with `provideManagedShared` and then provide the rest on a per test basis, it doesn't work because `provideManagedShared` needs to use the shared resource in a single effect, collapsing the environment type (it works if you do it the other way around and provide the per test resources first). Also in `filterLabels` and `filterAnnotations` we need to execute perform suite level effects to get down to the test nodes, see if there are nodes satisfying the predicate, and then prune empty suites. In this case it would only be noticeable if you had separate copies of suite level resources for different suites within a larger suite, which I haven't really seen, but also an issue.
2. #2532 is also potentially relevant here. If we have test level resources required we don't have a way to statically know that these are only required at the test level and not the suite level, so me may end up creating resources more times than is necessary.

jdegoes:
We really need to solve this duo (trio?) of issues with one unified and clean solution well-integrated with ZLayer.

jdegoes:
Great work on this! Love the cleanup too.

PR #2805, opened 2020-02-05
ZIO Core: Improve Ergonomics of ZIO#summarized

adamgfraser:
Working with ZIO#summarized is kind of annoying right now. Because the `B` parameter comes first and is in input position it can't be inferred, so we either have to specify all four type parameters of `summarized` or annotate the types of the inputs to `f`, which often otherwise could be very concise. By switching the order of the arguments we get the `B` type inferred from the result type of `summary`. I think this is also more logical conceptually, as first we are specifying the summary effect we want to use, and then we specify how we want to combine its values.

```scala
final def summarized[R1 <: R, E1 >: E, B, C](f: (B, B) => C)(summary: ZIO[R1, E1, B]): ZIO[R1, E1, (C, A)]
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -4,8 +4,6 @@ import scala.annotation.tailrec
 import scala.util.{ Failure, Success }
 
 import zio.Cause._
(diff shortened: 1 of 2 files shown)


jdegoes:
@adamgfraser Awesome!

PR #2807, opened 2020-02-05
Add alias for ZSTM#orElse

mijicd:


jdegoes:
@mijicd Good to see more consistency here!

PR #2808, opened 2020-02-05
Don't allocate either on ZSTM#ignore

mijicd:


jdegoes:
@mijicd Love little optimizations like this. You can make it faster too if you factor out the lambdas into private constants, e.g. `Any => Unit`, and then cast them.

PR #2810, opened 2020-02-05
Implement ZIO#asSomeError

adamgfraser:
When working with optional values it is often convenient to move the possibility of the absence of a value to the error channel to let us work just with the "happy path" using `optional` and `some`. However, one issue we can run into is we then have a `ZIO[R, Option[E], A]` so if we want to compose it with other effects that have an error type `E` we need to lift that error into an option to be able to compose it. This PR adds a method `asSomeError` to do that as well as a `asSome` method that operates on the value channel for symmetry. Variants are included for `ZManaged` and `ZSTM`.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -112,6 +112,18 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     mapError(new ZIO.ConstFn(() => e1))
 
   /**
+   * Maps the success value of this effect to an optional value.
+   */
+  final def asSome: ZIO[R, E, Option[A]] =
+    map(Some(_))
+
+  /**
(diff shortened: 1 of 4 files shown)


jdegoes:
@adamgfraser Looks great! Should it be `asSomeError` or `asErrorSome`? The latter seems a bit more consistent with other things `mapPar` / `mapParN`, etc.

jdegoes:
@adamgfraser Actually I'm not sure about that. It reads rather well like this and I see the ambiguity of the choice...

adamgfraser:
@jdegoes Yeah I was trying to be consistent with other combinators with `error` variants where it seemed like the `error` came last so the two variants would appear together in autocompletion (e.g. `tap` and `tapError`, `flatMap` and `flatMapError`) but in general when we have two "suffixes" I don't think we have a great rule for which comes first.

Issue #2802, opened 2020-02-04
Port Combinators from ZIO to STM


adamgfraser:
There are a variety of combinators on ZIO that don't exist on STM right now. Many of these methods don't make sense for STM (e.g. methods dealing with fibers, interruption, importing side effecting code into ZIO, etc...). But there are many others that would make sense and would be helpful to have (e.g. `flatMapError`).

The first part of this issue is to go through the methods in ZIO and identify which of them would make sense to implement for STM. Then we can use this issue as a master tracker and people can take different methods they want to implement. In general it should be possible to implement these methods in terms of existing STM combinators, so this is a great opportunity to learn more about STM and add some new methods to the public API.

Copying @mijicd.

mijicd:
I will go through this tonight and update the issue with implementable combinators.

adamgfraser:
Awesome!

mijicd:
I've collected all combinators that can be added to (Z)STM. Obviously, we could add more by committing and invoking the respective operation on ZIO, but I'd prefer not to do that now as it would clutter the interface.

Please take a look at the lists, suggest combinators that we could add/remove, and enjoy hacking :).

### ZSTM

- [ ] absolve
- [ ] andThen (with aliases)
- [ ] bimap
- [ ] catchAll
- [ ] catchSome
- [ ] filterOrDie
- [ ] filterOrDieMessage
- [ ] filterOrElse
- [ ] filterOrElse_
- [ ] filterOrFail
- [ ] flatMapError
- [ ] flip
- [ ] flipWith
- [ ] flattenErrorOption
- [ ] get
- [ ] head
- [ ] merge
- [ ] none
- [ ] onFirst
- [ ] onLeft
- [ ] onRight
- [ ] onSecond
- [ ] optional
- [ ] orDie
- [ ] orDieWith
- [ ] left
- [ ] leftOrFail
- [ ] leftOrFailException
- [ ] refineOrDie
- [ ] refineOrDieWith
- [ ] right
- [ ] rightOrFail
- [ ] rightOrFailException
- [ ] reject
- [ ] reject
- [ ] rejectM
- [ ] some
- [ ] someOrFail
- [ ] someOrFailException

### ZSTM companion

- [ ] absolve
- [ ] foldLeft
- [ ] foldRight
- [ ] fromFunction
- [ ] fromFunctionM
- [ ] fromOption
- [ ] left
- [ ] mapN
- [ ] mergeAll
- [ ] none
- [ ] reduceAll
[...]

PR #2812, opened 2020-02-05
Add retryUntil and retryWhile to ZSTM

mijicd:
The reason for adding these 2 combinators is misleading semantics of `filter` and `withFilter` in the current implementation as discussed [here](https://github.com/zio/zio/issues/2802).

### Summary

- `filter` has been renamed to `retryUntil`.
- `withFilter` is dropped.
- `retryWhile` has been added.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
@@ -158,7 +158,7 @@ object ZSTMSpec extends ZIOBaseSpec {
         ) {
           for {
             tvar <- TRef.makeCommit(42)
-            join <- tvar.get.filter(_ == 42).commit
+            join <- tvar.get.retryUntil(_ == 42).commit
             _    <- tvar.set(9).commit
             v    <- tvar.get.commit
           } yield assert(v)(equalTo(9)) && assert(join)(equalTo(42))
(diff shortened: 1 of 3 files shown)


jdegoes:
In for comprehensions, the syntax `if balance > 200` used to work, and of course retried.

One could make an argument that `filter` / `withFilter` (and `collect`) on ZIO should be consistent with STM. Since it's not that well-defined as it is for ZIO.

```scala
getStrLn.filter(isNumber(_))

doHttpRequest.filter(_.statusCode == 200).timeout
```

jdegoes:
That said, I do love `retryUntil`, that should definitely be the default name. The `filter` stuff, however, is nice for `for` comprehensions (if we decide to keep it). And `collect` is potentially useful more because Scala developers know to look for it.

mijicd:
Should we re-introduce `withFilter` and `filter` as aliases for use in for-comprehensions then? Also, would that imply we should look for another name for companion object filter (see [this comment](https://github.com/zio/zio/issues/2802#issuecomment-582538772))?

adamgfraser:
One thing I worry about is that `withFilter` could retry if there is a match error which I think could be quite surprising.

```scala
// will retry forever
val x = (1, 2)
for {
  (a, b, c) <- ZSTM.succeed(x)
yield (a, b, c)
```

jdegoes:
@adamgfraser Right. I use to not be in fan of `WithFilter` for that reason. It's a widely acknowledged Scala wart. But fortunately it only occurs with "destructuring bind", and not with `if` statements embedded in the conditional. Actually the only thing I care about is `if` statements.

adamgfraser:
@jdegoes Yeah, I just worry that failing if you do a match like that is to some extent an "expected risk" given the language but retrying on that is not. But maybe that is just the STM equivalent and we shouldn't optimize for unfortunately known warts.

jdegoes:
@adamgfraser I do think unexpected termination is easier to diagnose than unexpected non-termination. But I think the most productive path may be pressure on Scala to fix this wart...

adamgfraser:
@jdegoes I definitely agree with that. But it is going to be a long time before that is actually available.

jdegoes:
@mijicd That sounds good to me! 👍

mijicd:
@jdegoes done!

PR #2817, opened 2020-02-06
Add bipmap, absolve, doUntil, and doWhile to STM

sken77:
test are next. just checking this are good

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
@@ -5,7 +5,6 @@ import zio.duration._
 import zio.test.Assertion._
 import zio.test.TestAspect.nonFlaky
 import zio.test._
-import zio.test._
 import zio.test.environment.Live
 
 object ZSTMSpec extends ZIOBaseSpec {
@@ -18,6 +17,23 @@ object ZSTMSpec extends ZIOBaseSpec {
       testM("`STM.failed` to make a failed computation and check the value") {
         assertM(STM.failNow("Bye bye World").commit.run)(fails(equalTo("Bye bye World")))
       },
+      suite("`absolve` to convert ")(
+        testM("A successful Right computation into the success channel") {
+          assertM(STM.succeedNow(Right(42)).absolve.commit)(equalTo(42))
+        },
+        testM("A successful Left computation into the error channel") {
+          assertM(STM.succeedNow(Left("oh no!")).absolve.commit)(fails(equalTo("oh no!")))
+        }
+      ),
+      suite("`bimap` when")(
+        testM("having a success value") {
+          import zio.CanFail.canFail
+          assertM(STM.succeedNow(1).bimap(_ => -1, s => s"$s as string").commit)(equalTo("1 as string"))
+        },
+        testM("having a fail value") {
+          assertM(STM.failNow(-1).bimap(s => s"$s as string", _ => -1).commit)(equalTo("-1 as string"))
+        }
+      ),
       suite("`either` to convert")(
         testM("A successful computation into Right(a)") {
           import zio.CanFail.canFail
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -132,6 +139,13 @@ final class ZSTM[-R, +E, +A] private[stm] (
   final def asSomeError: ZSTM[R, Option[E], A] =
     mapError(Some(_))
 
+  /**
+   * Returns an `STM` effect whose failure and success channels have been mapped by
+   * the specified pair of functions, `f` and `g`.
+   */
+  final def bimap[E2, B](f: E => E2, g: A => B)(implicit ev: CanFail[E]): ZSTM[R, E2, B] =
+    mapError(f).map(g)
```
adamgfraser: Can be implemented slightly more efficiently in terms of `foldM`.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -161,6 +175,18 @@ final class ZSTM[-R, +E, +A] private[stm] (
   def commitEither: ZIO[R, E, A] =
     either.commit.absolve
 
+  /**
+   * Repeats this `STM` effect until its result satisfies the specified predicate.
+   */
+  final def doUntil(f: A => Boolean): ZSTM[R, E, A] =
+    flatMap(a => if (!f(a)) doUntil(f) else ZSTM.succeedNow(a))
```
adamgfraser: Could switch the order here if you wanted to avoid the `(!(f(a))`. If the condition is true return immediately, otherwise continue.

adamgfraser:
Also needs a format. You can run `sbt prepare` to lint your code before submitting a PR so that doesn't happen.

core/shared/src/main/scala/zio/stm/ZSTM.scala
mijicd: I'd invert the condition here, just to make it aligned with `doWhile`, but treat this as a nitpick.

sken77:
yep. just checking the implementations look good. ill do the tests next as i said

sken77:
finished https://github.com/zio/zio/issues/2802

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -18,6 +17,37 @@ object ZSTMSpec extends ZIOBaseSpec {
...
+      suite("`bimap` when")(
+        testM("having a success value") {
+          import zio.CanFail.canFail
+          assertM(STM.succeedNow(1).bimap(_ => -1, s => s"$s as string").commit)(equalTo("1 as string"))
+        },
+        testM("having a fail value") {
+          assertM(STM.failNow(-1).bimap(s => s"$s as string", _ => 0).commit.run)(fails(equalTo("-1 as string")))
+        }
+      ),
+      testM("`doWhile` to run effect while it satisfies predicate") {
```
adamgfraser: You can do this one and the one below in a single transaction instead of committing after each step. Generally better to stay within the transactional context if possible.

sken77:
oops. while i was waiting for review i was working on other combinators and commit the file by accident. ill revert it

adamgfraser:
If you want we can expand the scope of this PR to include other combinators. Or we can merge this and you can open another one. Whatever you prefer.

sken77:
> 
> 
> If you want we can expand the scope of this PR to include other combinators. Or we can merge this and you can open another one. Whatever you prefer.

yeah that wasnt intentional. i was doing the others while waiting and commited by accident. ill revert the commit on that file

adamgfraser:
Sounds good.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -35,18 +35,18 @@ object ZSTMSpec extends ZIOBaseSpec {
         }
       ),
       testM("`doWhile` to run effect while it satisfies predicate") {
-        for {
-          a <- TQueue.bounded[Int](5).commit
-          _ <- a.offerAll(List(0, 0, 0, 1, 2)).commit
-          n <- a.take.doWhile(_ == 0).commit
-        } yield assert(n)(equalTo(1))
+        (for {
```
adamgfraser: You can also define a local variable here if you don't want to have to put the parentheses around the for comprehension.

```
val transaction = for { ... }
assertM(transaction.commit)(???)
```

PR #2818, opened 2020-02-06
Make fromServiceM functions not require R <: Has[_]

### LINKED ISSUE #2816: In ZLayer.fromServicesM, R should not have type bounds
In `ZLayer.fromServiceM` functions, `R` has a type bound of `<: Has[_]`. I think that this requirement is unnecessary, as in the definition of `ZLayer[RIn, E, ROut]` - `RIn` is unbounded. This creates a problem when the effect provided fromServiceM is an `IO`, as in that case `R` is `Any`, and is not a sub-type of `Has[_]`.

Let me know if this is unintentional and I can put together a PR.

thesamet:
Fixes #2816


adamgfraser:
Thanks for the fix!

PR #2825, opened 2020-02-06
Add combinators to STM

sken77:
andThen >>>
compose <<<
eventually
catchAll
flatMapError
flattenErrorOption

tests are next

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -112,13 +112,13 @@ final class ZSTM[-R, +E, +A] private[stm] (
   /**
    * Propagates the given environment to self
    */
-  final def >>>[R1 >: A, E1 >: E, B](that: ZSTM[R1, E1, B]): ZSTM[R, E1, B] =
+  def >>>[R1 >: A, E1 >: E, B](that: ZSTM[R1, E1, B]): ZSTM[R, E1, B] =
     flatMap(that.provide)
 
   /**
    * Propagates self environment to that
    */
-  final def <<<[R1, E1 >: E](that: ZSTM[R1, E1, R]): ZSTM[R1, E1, A] =
+  def <<<[R1, E1 >: E](that: ZSTM[R1, E1, R]): ZSTM[R1, E1, A] =
     that >>> self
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -219,12 +252,25 @@ final class ZSTM[-R, +E, +A] private[stm] (
       case TExit.Retry      => ZSTM.retry
     }
 
+  /**
+   * Creates a composite effect that represents this effect followed by another
+   * one that may depend on the error produced by this one.
+   */
+  final def flatMapError[R1 <: R, E2](f: E => ZSTM[R1, Nothing, E2])(implicit ev: CanFail[E]): ZSTM[R1, E2, A] =
+    flipWith(_ flatMap f)
```
adamgfraser: Probably more efficient in terms of `foldM`.

sken77:
test for `eventually` is timing out and im not sure why

adamgfraser:
I think it because `orElse` resets the journal before continuing with the right hand side effect, which is somewhat surprising to me.

@mijicd Is this intended behavior?

mijicd:
Exactly, in case of failure/retry of left, journal will be nuked. That's the intended behavior.

adamgfraser:
Hmmm. I wonder if we should think about that or at least document it more because it is significantly different than in ZIO. In ZIO you could define `orElse(zio)` as `foldM(_ => zio, ZIO.succeed)` whereas the semantic here is quite different. STM does provide different functionality so don't want to shoehorn it into ZIO but it does seem like it could be surprising.

sken77:
It works fine if i use `foldM` instead of `orElse`

adamgfraser:
Yes, that's definitely a good solution for this PR.

sken77:
I'm not sure how would i go about testing `andThen` and `compose` couldnt find an example on `ZIOSpec`

adamgfraser:
Try creating two STM effects that access say an `Int` in the environment and do something with it, then glue them together using `andThen` and `compose`, and test that if you provide them with a value for the environment you get the expected value back.

sken77:

PR #2829, opened 2020-02-07
Add fromOption, fromFunction and fromFunctionM to ZSTM

mijicd:
Part of #2802.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -94,25 +94,25 @@ object STM {
     ZSTM.fromEither(e)
 
   /**
-   * @see See [[zio.stm.ZSTM.fromTry]]
+   * @see See [[zio.stm.ZSTM.fromFunction]]
    */
   def fromFunction[R, A](f: R => A): ZSTM[R, Nothing, A] =
     ZSTM.fromFunction(f)
 
   /**
-   * @see [[zio.stm.ZSTM.fromFunctionM]]
+   * @see See [[zio.stm.ZSTM.fromFunctionM]]
    */
   def fromFunctionM[R, E, A](f: R => STM[E, A]): ZSTM[R, E, A] =
     ZSTM.fromFunctionM(f)
 
   /**
-   * @see [[zio.stm.ZSTM.fromOption]]
+   * @see See [[zio.stm.ZSTM.fromOption]]
    */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -102,13 +102,31 @@ object STM {
...
+   * @see See [[zio.stm.ZSTM.fromEither]]
    */
   def fromEither[E, A](e: => Either[E, A]): STM[E, A] =
     ZSTM.fromEither(e)
 
   /**
-   * Creates an STM effect from a `Try` value.
+   * @see See [[zio.stm.ZSTM.fromFunction]]
+   */
+  def fromFunction[R, A](f: R => A): ZSTM[R, Nothing, A] =
```
adamgfraser: Do we want to constrain the types on these? Normally here and like in `IO.scala` we would constrain the `R` type to be `Any`, which admittedly is not particularly useful but is consistent with elsewhere and honors the contract that a method on the `STM` companion object is going to give you an instance of `STM`.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -684,11 +684,30 @@ object ZSTM {
     }
 
   /**
-   * Creates an STM effect from a `Try` value.
+   * Lifts a function `R => A` into a `ZSTM[R, Nothing, A]`.
+   */
+  def fromFunction[R, A](f: R => A): ZSTM[R, Nothing, A] =
```
adamgfraser: Could also implement this and `fromFunctionM` as `access(f)` and `assessM(f)`.

mijicd: Good point. Should we change that in ZIO as well?

core/shared/src/main/scala/zio/stm/STM.scala
mijicd: Absolutely, fixed.

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: Probably slightly more efficient since we do it in one `Read` operation` instead of a `Read` and a `FlatMap`.

mijicd: Done.

PR #2830, opened 2020-02-07
Delegate STM docs to ZSTM

mijicd:


mijicd:
Exactly my thoughts, we shouldn't allow piling these, as they'll inevitably diverge.

PR #2843, opened 2020-02-08
Mark Fiber Supervision Test as Flaky

adamgfraser:


jdegoes:
Arg. Can't wait for reports to see how bad the situation is getting. 😅

jdegoes:
Maybe we need a combinator, `almostNonFlaky(0.95, 200)`. 😆

adamgfraser:
I'm not sure about the name but I think that would be a valuable combinator. I will work on adding.

PR #2847, opened 2020-02-08
Add onLeft and onRight to ZSTM

mijicd:
Part of #2802

### Summary

- Added `identity`, `|||` and `+++`.
- Added `onLeft` and `onRight`.
- Document undocumented (related) ZIO combinators.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -823,9 +823,19 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       ZIO.descriptorWith(descriptor => if (descriptor.interruptors.nonEmpty) cleanup else ZIO.unit)
     )
 
+  /**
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -823,9 +823,19 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
       ZIO.descriptorWith(descriptor => if (descriptor.interruptors.nonEmpty) cleanup else ZIO.unit)
     )
 
+  /**
+   * Returns this effect if environment is on the left, otherwise returns
+   * whatever is on the right unmodified. Note that the result is lifted
+   * in either.
+   */
   final def onLeft[R1 <: R, C]: ZIO[Either[R1, C], E, Either[A, C]] =
```
adamgfraser: We don't need the `R1` parameter here or in `onRight`..

mijicd:
@adamgfraser good point dropped them from both ZSTM and ZIO.

PR #2848, opened 2020-02-08
Reduce visibility of private STM methods

mijicd:


adamgfraser:
Would we ever want other ZIO libraries to be able to use the more efficient versions if they are implementing STM functionality?

Whatever changes we make here we should probably also make in streams.

mijicd:
@adamgfraser I think it makes sense to have them available for such use, but I'd rather keep them "closer" to their "origin", which is `stm`. WDYT?

adamgfraser:
I don't have a strong point of view. I'm just wondering if say you're writing code for  `zio.actors` and need to implement STM functionality for some reason you could use these if they are `private[zio]` but couldn't if they are `private[stm]`. On the other hand that does increase the visibility of these methods.

mijicd:
That's true. I'll close the PR for now then, in case we decide to strenghten the constraint, we can easily re-open or create a new one.

PR #2849, opened 2020-02-08
Re-implement ignore using unitFn

mijicd:
Dropped lambdas in favor of constant cast as suggested in [this comment](https://github.com/zio/zio/pull/2808#issuecomment-582370210).

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -3239,6 +3239,7 @@ object ZIO {
   def apply[A](a: => A): Task[A] = effect(a)
 
   private val _IdentityFn: Any => Any = (a: Any) => a
+  private val _UnitFn: Any => Unit = (_: Any) => ()
 
   private[zio] def identityFn[A]: A => A = _IdentityFn.asInstanceOf[A => A]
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -670,7 +670,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Returns a new effect that ignores the success or failure of this effect.
    */
-  final def ignore: URIO[R, Unit] = self.foldM(_ => ZIO.unit, _ => ZIO.unit)
+  final def ignore: URIO[R, Unit] = self.fold(ZIO.unitFn, ZIO.unitFn)
```
jdegoes: For ZIO, this can be slightly faster: you can do `foldCause` which involves no wrapping.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3251,6 +3253,8 @@ object ZIO {
         )
     }
 
+  private[zio] def unitFn[A]: A => Unit = _UnitFn.asInstanceOf[A => Unit]
```
adamgfraser: This may be a stupid question on my part but do we need this? Isn't `Any => Unit` already a valid function for any place we want to use this? Or do we save time by doing the cast ourselves instead of the compiler having to realize that?

jdegoes: Ah you're right, we don't need the helper function, just the `val`...

mijicd:
@jdegoes ready for re-review.

mijicd:
Unfortunately, had to revert the change `fold => foldCause` as it ended up swallowing exceptions, which violates:

```scala
testM("not catch throwable") {
  assertM(ZIO.dieNow(ExampleError).ignore.run)(dies(equalTo(ExampleError)))
}
```

PR #2853, opened 2020-02-09
Return More Specific Type from Methods on Managed Runtime

adamgfraser:
Minor follow up on #2751. Makes methods on a managed runtime return a managed runtime instead of just a runtime. This way you can call methods on the managed runtime and retain the ability to shut it down early.

### DIFF
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -39,12 +39,12 @@ trait Runtime[+R] {
   /**
    * Constructs a new `Runtime` by mapping the environment.
    */
-  final def map[R1](f: R => R1): Runtime[R1] = Runtime(f(environment), platform)
+  def map[R1](f: R => R1): Runtime[R1] = Runtime(f(environment), platform)
 
   /**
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -177,6 +177,33 @@ object Runtime {
...
+    override final def withReportFatal(f: Throwable => Nothing): Runtime.Managed[R] =
+      mapPlatform(_.withReportFatal(f))
+
+    override final def withReportFailure(f: Cause[Any] => Unit): Runtime.Managed[R] =
+      mapPlatform(_.withReportFailure(f))
+
+    override final def withTracing(t: Tracing): Runtime.Managed[R] =
+      mapPlatform(_.withTracing(t))
+
+    override final def withTracingConfig(config: TracingConfig): Runtime[R] =
```
iravid:
```suggestion
    override final def withTracingConfig(config: TracingConfig): Runtime.Managed[R] =
```

adamgfraser: Good catch! Fixed now.

PR #2855, opened 2020-02-09
Add ZStream filter combinator

bmarinovic:
Tests cover only one-element Stream. Will change once `fromIterable()` is implemented


bmarinovic:
Sure! Is this ok?
>* @param pred predicate used for deciding whether element should be retained

vasilmkd:
Take a look at the other combinators in the file. Probably needs a @return also. Otherwise feel free to write anything.

vasilmkd:
@iravid I thinks this looks ready for merging.

PR #2857, opened 2020-02-10
Improve Scala Native Support

adamgfraser:
Work in process on improving ZIO support for Scala Native.

Where we currently are is that ZIO compiles on Scala Native but we have not been able to run tests because of the lack of a java time implementation for Scala Native. As a result, we know it compiles but we don't actually have assurance that it runs the way we want it to which is not a great place to be.

I realized that the java time dependency issue is really driven by the `currentDateTime` method on `Clock` and in particular the way the `TestClock` uses a default time zone which results in a runtime error if there is not an implementation of `ZoneId`. We can get around this by making the `zoneId` field in `TestClock.FiberData` lazy so you will fail at runtime if you call `dateTime`, `setDateTime` or `setTimeZone` but otherwise everything will work normally.

Using this technique I was able to get ZIO Stream, ZIO Test, and the test suite for Core compiling on Scala Native. A couple of other issues which seem solvable:

1. There isn't any implementation of java time published for 0.4.2 of Scala Native. We need to have at least stubs to compile. Right now I have just created a few in the test project.
2. Scala Native does not implement `java.util.concurrent.ConcurrentLinkedQueue`. For now I have adapted the ScalaJS implementation. We can probably use a version of that internally to run our tests if necessary and Scala Native can add.
3. Scala Native does not offer a constructor for throwable with all the parameters that we use to control suppression. Open PR is at https://github.com/scala-native/scala-native/pull/1688. In the meantime we can port some code or do a platform specific version that doesn't address that. For now I have just removed the suppression logic in this branch to try to get this to run.
4. `scala-portale-reflect` is not currently available for Scala Native so we can't use the SBT test runner. We will have to run specs with their main method.

With these changes I am able to run a spec by doing something like:

```
sbt
++2.11.12
project coreTestsNative
test:run
```

However, after selecting a test (e.g. `ZLayerSpec`), I get a successful completion but no test output.

Copying @rwhaling and @lolgab from the Scala Native team who has been fantastic about helping troubleshoot this.


jdegoes:
@adamgfraser Status on this one?

adamgfraser:
@jdegoes In working on this have run into several limitations in Scala Native functionality and test framework. At this point I think it is highly unlikely we will be able to merge this before RC18 or even 1.0, though I think we will be able to add it in a binary compatible way after that.

adamgfraser:
Closing this pending resolution of issues with Scala Native test framework.

sideeffffect:
@lolgab added some commits on top of @adamgfraser 's: https://github.com/zio/zio/compare/master...lolgab:native

lolgab:
Hi everyone,
ZIO Scala Native is mainly blocked by https://github.com/scala-native/scala-native/issues/1683 and by Scala Native lacking a java.time implementation used in Clock.
We are thinking to create an external artifact (something like `scala-native-loop`, `scala-native-zio`) with the missing parts and I have to investigate if I can patch a part of Scala Native javalib (`java.lang.Throwable`) to inject the fix for the Throwable suppression present here: https://github.com/scala-native/scala-native/pull/1688.
I play to do it myself but I had not the chance to spend enough time on it.
Is it possible to do this after `1.0.0` in case you release @jdegoes ?
Do you want to maintain the experimental nature, with the Scala Native build, or you want changes to be much more cautious after `1.0.0` is released?
Thank you :)

adamgfraser:
@lolgab I think it is okay to be more experimental with Scala Native even after 1.0 is released as long as it does not impact binary compatibility for JVM and ScalaJS users. At least my point of view at this point is if you are using Scala Native you are an early adopter and would prefer changes for additional functionality, particularly when it is going from things not working to working.

PR #2858, opened 2020-02-10
add onFirst, on Second, &&& alias to STM

sken77:

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -392,7 +392,7 @@ final class ZSTM[-R, +E, +A] private[stm] (
    * passes the effect input `R` along unmodified as the second element
    * of the tuple.
    */
-  def onFirst[R1 <: R, A1 >: A]: ZSTM[R1, E, (A1, R1)] =
+  def onFirst[R1 <: R]: ZSTM[R1, E, (A, R1)] =
     self <*> ZSTM.identity[R1]
 
   /**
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -381,6 +387,14 @@ final class ZSTM[-R, +E, +A] private[stm] (
       case TExit.Retry      => ZSTM.retry
     }
 
+  /**
+   * Propagates the success value to the first element of a tuple, but
+   * passes the effect input `R` along unmodified as the second element
+   * of the tuple.
+   */
+  def onFirst[R1 <: R, A1 >: A]: ZSTM[R1, E, (A1, R1)] =
```
adamgfraser: You don't need the `A` parameter here. I realize it was this way in `ZIO.scala` as well. We can either correct it as part of this PR or open a separate issue.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -397,6 +411,14 @@ final class ZSTM[-R, +E, +A] private[stm] (
   def onRight[C]: ZSTM[Either[C, R], E, Either[C, A]] =
     ZSTM.identity[C] +++ self
 
+  /**
+   * Propagates the success value to the second element of a tuple, but
+   * passes the effect input `R` along unmodified as the first element
+   * of the tuple.
+   */
+  def onSecond[R1 <: R, A1 >: A]: ZSTM[R1, E, (R1, A1)] =
```
adamgfraser: Same comment as above regarding `onFirst`.

PR #2860, opened 2020-02-10
Bring back java interop and add some docs

### LINKED ISSUE #2187: Fold interop-java back into this project
We moved the conversion methods for CompletionStage and friends out to a separate project, but that seems pointless as the Java platform is always there.

The methods there can be added to a platform-specific file in this project.

Anyone know of a reason why we shouldn’t do this?

regiskuckaertz:
Some colleagues have complained multiple times about the difficulty to find Java interop features as well as the lack of documentation.

A few questions/observations:
- [x] ~~I'm creating synthetic fibers, which feels wrong~~: nay
- [x] ~~Can this move directly into the ZIO/Task/etc. companion in the JVM branch instead of via the interop package?~~
- [x] ~~The signature of `fromFutureJava` is a bit weird, why not a by-name parameter?~~
- unrelated but there should be the same interop layer for JavaScript promises so that ZIO can directly interact with the bajillion of js libraries

Resolves #2187


iravid:
Definitely move into the core zio package, into the companion objects, similarly to the scala.concurrent.Future 👍🏻👍🏻

core/jvm/src/main/scala/zio/JavaSpecific.scala
```diff
@@ -0,0 +1,128 @@
...
+    def toZio: Task[A] = ZIO.fromCompletionStage(csUio)
+  }
+
+  implicit class FutureJavaconcurrentOps[A](private val futureUio: UIO[Future[A]]) {
+
+    /** WARNING: this uses the blocking Future#get, consider using `CompletionStage` */
+    def toZio: RIO[Blocking, A] = ZIO.fromFutureJava(futureUio)
+  }
+
+  implicit class FiberObjOps(private val fiberObj: Fiber.type) {
```
neko-kai: mixing these extension methods into the package object works only for Scala 2, in Scala 3 enclosing package object is no longer part of the implicit scope, so to make this work on Scala 3 you may better mix directly into Fiber object:

```scala
trait FiberJavaSpecific {
  implicit class ...
}

object Fiber extends FiberJavaSpecific
```

regiskuckaertz: Thanks for the tip, done 👍

jdegoes:
Awesome work!

> The signature of fromFutureJava is a bit weird, why not a by-name parameter?

Should be. Same for other methods like this. Always by-name.

> unrelated but there should be the same interop layer for JavaScript promises so that ZIO can directly interact with the bajillion of js libraries

Agreed. We may have to split off Java vs Js mixins into the ZIO trait / object.

core/jvm/src/main/scala/zio/FiberPlatformSpecific.scala
```diff
@@ -0,0 +1,96 @@
...
+
+import _root_.java.util.concurrent.{ CompletionStage, Future }
+
+import zio.Fiber.Status
+import zio.blocking.Blocking
+import zio.interop.javaz
+
+private[zio] trait FiberPlatformSpecific {
+
+  implicit class FiberObjOps(private val fiberObj: Fiber.type) {
```
jdegoes: Rather than using extension methods, let's mix directly into the `Fiber` companion object.

core/jvm/src/main/scala/zio/IOPlatformSpecific.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package zio
+
+import _root_.java.util.concurrent.CompletableFuture
+
+import zio.interop.javaz
+
+private[zio] trait IOPlatformSpecific {
+
+  implicit class IOCompletableFutureOps[E, A](private val io: IO[E, A]) {
```
jdegoes: I'd mix this into `ZIO` directly rather than use extension methods. A layer of indirection can allow different mixins on JS versus JVM.

core/jvm/src/main/scala/zio/TaskPlatformSpecific.scala
```diff
@@ -0,0 +1,30 @@
...
+
+package zio
+
+import _root_.java.util.concurrent.{ CompletableFuture }
+
+import zio.interop.javaz
+
+private[zio] trait TaskPlatformSpecific {
+
+  implicit class TaskCompletableFutureOps[A](private val io: Task[A]) {
```
jdegoes: I'd mix this into `ZIO` directly rather than use extension methods. A layer of indirection can allow different mixins on JS versus JVM.

core/jvm/src/main/scala/zio/UIOPlatformSpecific.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package zio
+
+import _root_.java.util.concurrent.{ CompletionStage, Future }
+
+import zio.blocking.Blocking
+
+private[zio] trait UIOPlatformSpecific {
+
+  implicit class CompletionStageJavaconcurrentOps[A](private val csUio: UIO[CompletionStage[A]]) {
```
jdegoes: I'd mix this into `ZIO` directly rather than use extension methods. A layer of indirection can allow different mixins on JS versus JVM.

core/jvm/src/main/scala/zio/UIOPlatformSpecific.scala
```diff
@@ -0,0 +1,35 @@
...
+    def toZio: Task[A] = ZIO.fromCompletionStage(csUio)
+  }
+
+  implicit class FutureJavaconcurrentOps[A](private val futureUio: UIO[Future[A]]) {
```
jdegoes: This is just weird `UIO[Future[A]]`. Rather we should have `ZIO.fromFutureJava` or something.

core/jvm/src/main/scala/zio/ZIOPlatformSpecific.scala
```diff
@@ -0,0 +1,37 @@
...
+import _root_.java.nio.channels.CompletionHandler
+import _root_.java.util.concurrent.{ CompletionStage, Future }
+
+import zio.blocking.Blocking
+import zio.interop.javaz
+
+private[zio] trait ZIOPlatformSpecific {
+
+  implicit class ZioObjJavaconcurrentOps(private val zioObj: ZIO.type) {
+    def withCompletionHandler[T](op: CompletionHandler[T, Any] => Unit): Task[T] =
```
jdegoes: Ditto, prefer no extension methods.

core/jvm/src/main/scala/zio/ZIOPlatformSpecific.scala
```diff
@@ -0,0 +1,37 @@
...
+      javaz.withCompletionHandler(op)
+
+    def fromCompletionStage[A](csUio: UIO[CompletionStage[A]]): Task[A] = javaz.fromCompletionStage(csUio)
+
+    /** WARNING: this uses the blocking Future#get, consider using `fromCompletionStage` */
+    def fromFutureJava[A](futureUio: UIO[Future[A]]): RIO[Blocking, A] = javaz.fromFutureJava(futureUio)
```
jdegoes: Ditto, prefer no extension methods. We just need things like `RIOJVM`, `RIOJS`, etc. traits that are mixed into the companion objects.

core/jvm/src/main/scala/zio/interop/javaz.scala
```diff
@@ -0,0 +1,105 @@
...
+import scala.concurrent.ExecutionException
+
+import _root_.java.nio.channels.CompletionHandler
+import _root_.java.util.concurrent.{ CompletableFuture, CompletionException, CompletionStage, Future }
+
+import zio._
+import zio.blocking.{ blocking, Blocking }
+
+private[zio] object javaz {
+  def withCompletionHandler[T](op: CompletionHandler[T, Any] => Unit): Task[T] =
```
jdegoes: `ZIO.effectAsyncWithCompletionHandler`???

core/jvm/src/main/scala/zio/interop/javaz.scala
```diff
@@ -0,0 +1,105 @@
...
+              }
+              cb(io)
+            }
+          }
+        }
+      }
+    }
+
+  /** WARNING: this uses the blocking Future#get, consider using `fromCompletionStage` */
+  def fromFutureJava[A](futureUio: UIO[Future[A]]): RIO[Blocking, A] =
```
jdegoes: `=> Future[A]`

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -416,7 +416,7 @@ object Fiber {
    * A synthetic fiber that is created from a pure value or that combines
    * existing fibers.
    */
-  sealed trait Synthetic[+E, +A] extends Fiber[E, A] {}
+  private[zio] abstract class Synthetic[+E, +A] extends Fiber[E, A] {}
```
jdegoes: This needs to be sealed, ideally. There is an internal object you can use to extend `Runtime` (from within the `zio` package), and we can do the same trick for `Synthetic`. /cc @adamgfraser

jdegoes:
@regiskuckaertz Looks awesome and quite useful! Just need a bit of integration work and upgrading to "new style" and should be ready to merge.

Thanks for your work on this, and welcome back. 😄

core/shared/src/main/scala/zio/Fiber.scala
regiskuckaertz: Oh, nice 👍

regiskuckaertz:
All done @jdegoes thanks for reviewing 🙏 

> Thanks for your work on this, and welcome back. 😄

🕺

core/jvm/src/main/scala/zio/IOPlatformSpecific.scala
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package zio
+
+import _root_.java.util.concurrent.CompletableFuture
+
+private[zio] trait IOPlatformSpecific {
+
+  def toCompletableFuture[E, A](io: IO[E, A])(implicit ev: E <:< Throwable): UIO[CompletableFuture[A]] =
```
jdegoes: My feeling is that all `toXYZ` methods are more idiomatic when added to the `ZIO` type itself. Also this makes it easier since we only have one of those!

This follows the convention of `ZIO#toFuture`, etc.

core/jvm/src/main/scala/zio/IOPlatformSpecific.scala
```diff
@@ -0,0 +1,29 @@
...
+    toCompletableFutureWith(io)(ev)
+
+  def toCompletableFutureWith[E, A](io: IO[E, A])(f: E => Throwable): UIO[CompletableFuture[A]] =
```
jdegoes: Ditto for this and all other examples. If a method takes an effect as its first parameter, it's a good sign it's better added to the ZIO data type as a method, rather than the companion object.

regiskuckaertz: Yes, gotcha. So the only I know of doing mixins on traits this way is by using self-type. Is that what you had in mind or is there another technique which I'm not aware of? i.e.

1. in the JVM branch:

```scala
trait ZIOPlatformSpecific[-R, +E, +A] { self: ZIO[R, E, A] =>
  def toCompletableFutureWith(f: E => Throwable): URIO[R, CompletableFuture[A]] = ???
}
```

2. in the shared branch:

```scala
sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E, A]
```

jdegoes: Self-type should be fine!

regiskuckaertz:
No worries ✌️ 

cc @adamnfish

PR #2862, opened 2020-02-10
Make Lock Heritable

### LINKED ISSUE #2859: effectBlockingInterrupt does not shift into the blocking pool
While `blocking.blocking` properly shifts the effect into the Blocking pool, `effectBlockingInterrupt` does not.

Wrapping the effect returned by `effectBlockingInterrupt` into `blocking` does not resolve the issue.

I can also reproduce on RC17 with `effectBlocking` (which AFAICT was only renamed in master).

I’ve pushed a branch to reproduce it [here](https://github.com/fsvehla/zio/tree/fsvehla/effect-blocking-investigation)

Here is the minimal failing ZIO test:

```scala
testM("effectBlockingInterrupt shifts into a blocking pool") {
  val threadName = effectBlockingInterrupt(Thread.currentThread.getName())

  assertM(threadName)(containsString("blocking"))
}
```

adamgfraser:
Fixes a bug in `FiberContext` where when a fiber was forked it was forked on the platform's default executor instead of the current active executor. We had a test for this but the test was based on the  fiber's descriptor which said it was running on a certain executor even though it wasn't actually running there.

Resolves #2859.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2639,13 +2639,20 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(v)(equalTo(InterruptStatus.uninterruptible))
       },
       testM("executor is heritable") {
-        val exec = internal.Executor.fromExecutionContext(100)(scala.concurrent.ExecutionContext.Implicits.global)
+        val threadName = ZIO.effectTotal(Thread.currentThread.getName)
+        val executor = internal.Executor.fromExecutionContext(100) {
+          scala.concurrent.ExecutionContext.Implicits.global
+        }
+        val poolName =
+          if (TestPlatform.isJS) "main"
+          else if (TestVersion.is211) "ForkJoinPool"
+          else "scala-execution-context-global"
         val io = for {
-          parentName <- ZIO.effectTotal(Thread.currentThread.getName)
-          childName  <- ZIO.effectTotal(Thread.currentThread.getName).fork.flatMap(_.join)
-        } yield assert(parentName)(containsString("scala-execution-context-global")) &&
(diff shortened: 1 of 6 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -2639,17 +2639,20 @@ object ZIOSpec extends ZIOBaseSpec {
...
-          } yield v.contains(exec)
-
-        assertM(io)(isTrue)
+        val threadName = ZIO.effectTotal(Thread.currentThread.getName)
+        val executor = internal.Executor.fromExecutionContext(100) {
+          scala.concurrent.ExecutionContext.Implicits.global
+        }
+        val poolName =
+          if (TestPlatform.isJS) "main"
+          else if (TestVersion.is211) "ForkJoinPool"
```
jdegoes: Did you try using `Thread.currentThread.getThreadGroup`? May be less brittle than substring testing of the thread name.

adamgfraser: Good idea. Used now.

jdegoes:
💪

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
adamgfraser: Actually `getThreadGroup` doesn't exist on ScalaJS so we need to revert to using the thread names.

jdegoes:
How about adding `getCurrentThreadGroup` to `Platform` companion object? In Javascript, it can just return `""` or whatever? I'm just concerned even a small change in Scala std lib or Java could affect the stability of the test case.

adamgfraser:
Okay.

jdegoes:
@adamgfraser Thank you, looks great and very robust!

PR #2863, opened 2020-02-10
add get to STM

sken77:

I have one question here.

The current code is failing to compile and im wondering if this is the behavior that we want:

` def get[E1 >: E, B](implicit ev1: E1 =:= Nothing, ev2: A <:< Option[B]): ZSTM[R, Unit, B]`

is from the `ZIO` class. If i understand this correctly this is saying that the effect we are doing this call should have signature `ZSTM[R, Nothing, Option[B]]` Im curious why it needs to be Nothing on the error channel. there is a very similar method to this called `someOrFail(e: E)` instead of implementing this call i tried to just call `someOrFail(())` but it seems i get problems on the Error channel. Also, the implementation here is my own and different from ZIO#get since that seems to convert to either first and do more complicated stuff. anyway. any feedback would be greatly appreciated

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
@@ -150,6 +150,22 @@ object ZSTMSpec extends ZIOBaseSpec {
           assertM(STM.failNow(None).flattenErrorOption("default error").commit.run)(fails(equalTo("default error")))
         }
       ),
+      suite("get")(
+        testM("extracts the value from Some") {
+          assertM(STM.succeedNow(Some(1)).get.commit)(equalTo(1))
+        },
+        testM("fails with Unit on None") {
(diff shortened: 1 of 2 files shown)


sken77:
Also, just to add. the test that is currently failing to compile is when we call get on a different effect after it already failed.

adamgfraser:
@sken77 We want the error type to be `Nothing` because we are potentially going to be stuffing a new `Unit` value into the error channel so we need to be able to unify that with whatever is currently in the error channel, which we can't do if we have an arbitrary `E` value already there. We could implement it in terms of `someOrFail` but we need to use `ev` to prove that we satisfy the constraints:

```scala
  /**
    */
   def get[E1 >: E, B](implicit ev1: E1 =:= Nothing, ev2: A <:< Option[B]): ZSTM[R, Unit, B] =
     mapError(ev1).someOrFail(())
```

PR #2864, opened 2020-02-10
ZLayer Improvements

adamgfraser:
Makes a couple of improvements to the usability of `ZLayer`:

- Adds `map` and `mapError` methods on `ZLayer`
- Adds constructors for `TestFailure` to make it easier to lift the error type into a `TestFailure` for providing environments to specs in ZIO Test
- Implements `set` which is like `update` but just lets you set a service to a specified implementation
- Implements a new method `overwrite` on `ZLayer` that lets you overwrite the services output by a layer with the services from the specified layer. For example, you can do `TestEnvironment.live.overwrite(Scheduler.live >>> Clock.live)` to create a version of the test environment that uses the live clock.

I'm having some issues with the last one. I would have thought that I could implement it in terms of some combination of `++` and `+!+` with different orderings of the layers but I wasn't able to get that to work deterministically so right now it is implemented directly similar to other combinators on `ZLayer` such as `++` and `>>>`. Also, right now I have a restriction that the layer being used to overwrite has to have no environmental requirements. I found that when I didn't do that I wasn't able to consistently override the environment. It would be nice to be able to relax that restriction so we could just to `TestEnvironment.live.overwrite(Clock.live)` for example.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -181,6 +181,44 @@ object ZLayerSpec extends ZIOBaseSpec {
         _      <- env.use_(ZIO.unit)
         actual <- ref.get
       } yield assert(actual)(equalTo(expected))
+    },
+    testM("map does not interfere with sharing") {
+      val expected = Vector(
+        "Acquiring Module 1",
+        "Acquiring Module 2",
+        "Acquiring Module 3",
+        "Releasing Module 3",
+        "Releasing Module 2",
+        "Releasing Module 1"
+      )
+      for {
+        ref    <- makeRef
+        layer1 = makeLayer1(ref)
+        layer2 = makeLayer2(ref)
+        layer3 = makeLayer3(ref)
+        env    = ((layer1.map(identity) >>> layer2) ++ (layer1 >>> layer3)).build
+        _      <- env.use_(ZIO.unit)
+        actual <- ref.get
+      } yield assert(actual)(equalTo(expected))
+    },
+    testM("mapError does not interfere with sharing") {
+      val expected = Vector(
+        "Acquiring Module 1",
+        "Acquiring Module 2",
+        "Acquiring Module 3",
+        "Releasing Module 3",
+        "Releasing Module 2",
+        "Releasing Module 1"
+      )
+      for {
+        ref    <- makeRef
+        layer1 = makeLayer1(ref)
+        layer2 = makeLayer2(ref)
+        layer3 = makeLayer3(ref)
+        env    = ((layer1.mapError(identity) >>> layer2) ++ (layer1 >>> layer3)).build
+        _      <- env.use_(ZIO.unit)
+        actual <- ref.get
+      } yield assert(actual)(equalTo(expected))
     }
   )
 }
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -108,8 +108,12 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
    * Returns a layer with its error channel mapped using the specified
    * function.
    */
-  def mapError[E1 >: E](f: E => E1): ZLayer[RIn, E1, ROut] =
-    new ZLayer(scope.map(run => run(_).mapError(f)))
+  def mapError[E1](f: E => E1): ZLayer[RIn, E1, ROut] =
+    new ZLayer(
+      Managed.finalizerRef(_ => UIO.unit).map { finalizerRef => memoMap =>
+        memoMap.getOrElseMemoize(self, finalizerRef).mapError(f)
+      }
+    )
 
   /**
    * Overwrites the services output by this laywr with the services output by
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -99,17 +99,46 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
     new ZLayer(self.scope)
 
   /**
+   * Returns a new layer whose output is mapped by the specified function.
+   */
+  def map[ROut1 >: ROut <: Has[_]](f: ROut => ROut1): ZLayer[RIn, E, ROut1] =
+    self >>> ZLayer.fromEnvironment(f)
+
+  /**
+   * Returns a layer with its error channel mapped using the specified
+   * function.
+   */
+  def mapError[E1 >: E](f: E => E1): ZLayer[RIn, E1, ROut] =
+    new ZLayer(scope.map(run => run(_).mapError(f)))
+
+  /**
+   * Returns a new layer with some of the services output by this layer
+   * replaced by the implementations in the specified layer
+   */
+  def overwrite[E1 >: E, A: Tagged](
+    that: ZLayer.NoDeps[E1, Has[A]]
+  )(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E1, ROut] =
+    new ZLayer(
+      Managed.finalizerRef(_ => UIO.unit).map { finalizerRef => memoMap =>
+        for {
+          l <- memoMap.getOrElseMemoize(self, finalizerRef)
+          r <- memoMap.getOrElseMemoize(that, finalizerRef)
+        } yield l.update[A](_ => r.get[A])
+      }
+    )
+
+  /**
    * Converts a layer that requires no services into a managed runtime, which
    * can be used to execute effects.
    */
-  def toRuntime[RIn2 <: RIn](p: Platform)(implicit ev: Any =:= RIn2): Managed[E, Runtime[ROut]] =
+  def toRuntime(p: Platform)(implicit ev: Any <:< RIn): Managed[E, Runtime[ROut]] =
     build.provide(ev).map(Runtime(_, p))
 
   /**
    * Updates one of the services output by this layer.
    */
   def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
-    new ZLayer(scope.map(run => run(_).map(_.update[A](f))))
+    self >>> ZLayer.fromEnvironment(_.update[A](f))
 }
 
 object ZLayer {
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -112,12 +112,12 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
     new ZLayer(scope.map(run => run(_).mapError(f)))
 
   /**
-   * Returns a new layer with some of the services output by this layer
-   * replaced by the implementations in the specified layer
+   * Overwrites the services output by this laywr with the services output by
+   * the specified layer.
    */
-  def overwrite[E1 >: E, A: Tagged](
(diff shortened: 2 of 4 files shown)


adamgfraser:
Ah, I see the issue here. When we just do `TestEnvironment.live.overwrite(Clock.live)` we overwrite the `Clock` service in the environment. But the clock service depends on a scheduler service, so if that is all we do we end up filling in the scheduler with the test scheduler which gets us a "live" clock that is not really live. That's why we need to overwrite with `Scheduler.live >>> Clock.live`.

adamgfraser:
I wonder if we should change `sleep` to require the `Scheduler` directly or combine the two services. This current situation where they are separate but the key method of `sleep` on clock just refers to the scheduler seems like a recipe for confusion.

adamgfraser:

test/shared/src/main/scala/zio/test/TestFailure.scala
```diff
@@ -23,4 +23,22 @@ sealed trait TestFailure[+E]
...
+  /**
+   * Constructs a new assertion failure with the specified result.
+   */
+  def assertion(result: TestResult): TestFailure[Nothing] =
+    Assertion(result)
+
+  /**
+   * Constructs a new runtime failure with the specified cause.
+   */
+  def die[E](cause: Cause[E]): TestFailure[E] =
```
iravid: For consistency with other data types, should this be die(t: Throwable)?

vasilmkd: Probably `halt` should be used for failing with a `Cause`.

adamgfraser: Yes. Fixed now.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -98,18 +98,47 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
...
+   */
+  def overwrite[E1 >: E, RIn2, A: Tagged](
+    that: ZLayer[RIn2, E1, Has[A]]
+  )(implicit ev: ROut <:< Has[A]): ZLayer[RIn with RIn2, E1, ROut] =
+    new ZLayer(
+      Managed.finalizerRef(_ => UIO.unit).map { finalizerRef => memoMap =>
+        for {
+          l <- memoMap.getOrElseMemoize(self, finalizerRef)
+          r <- memoMap.getOrElseMemoize(that, finalizerRef)
+        } yield l.update[A](_ => r.get[A])
```
jdegoes: I noticed (like you pointed out) this is a specialization fo `++` when `ROut2` is just a single `Has` that is already a part of `ROut`.

The implementation is almost the same, except instead of `union`, `update` is being called to do a targeted replacement of `A`.

I wonder why `union` would not overwrite in this case? In theory, `l union r` should always cause `r` to overwrite any parts already contained by `l`.

Any thoughts on this?

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -98,18 +98,47 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
   def fresh: ZLayer[RIn, E, ROut] =
     new ZLayer(self.scope)
 
+  /**
+   * Returns a new layer whose output is mapped by the specified function.
+   */
+  def map[ROut1 >: ROut <: Has[_]](f: ROut => ROut1): ZLayer[RIn, E, ROut1] =
```
jdegoes: Looks great!

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -98,18 +98,47 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
...
+    self >>> ZLayer.fromEnvironment(f)
+
+  /**
+   * Returns a layer with its error channel mapped using the specified
+   * function.
+   */
+  def mapError[E1 >: E](f: E => E1): ZLayer[RIn, E1, ROut] =
+    new ZLayer(scope.map(run => run(_).mapError(f)))
```
jdegoes: Quite useful. 👍

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -98,18 +98,47 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
...
-  def toRuntime[RIn2 <: RIn](p: Platform)(implicit ev: Any =:= RIn2): Managed[E, Runtime[ROut]] =
+  def toRuntime(p: Platform)(implicit ev: Any <:< RIn): Managed[E, Runtime[ROut]] =
     build.provide(ev).map(Runtime(_, p))
 
   /**
    * Updates one of the services output by this layer.
    */
   def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
-    new ZLayer(scope.map(run => run(_).map(_.update[A](f))))
+    self >>> ZLayer.fromEnvironment(_.update[A](f))
```
jdegoes: Nice cleanup!

test/shared/src/main/scala/zio/test/TestFailure.scala
```diff
@@ -23,4 +23,28 @@ sealed trait TestFailure[+E]
...
+  /**
+   * Constructs a runtime failure that fails with the specified error.
+   */
+  def fail[E](e: E): TestFailure[E] =
+    halt(Cause.fail(e))
+
+  /**
+   * Constructs a runtime failure with the specified cause.
+   */
+  def halt[E](cause: Cause[E]): TestFailure[E] =
```
jdegoes: Nice to see all these new constructors. 👍

jdegoes:
> I wonder if we should change sleep to require the Scheduler directly or combine the two services. This current situation where they are separate but the key method of sleep on clock just refers to the scheduler seems like a recipe for confusion.

Indeed I originally thought we should move `sleep` to `Scheduler` such that `Scheduler` would know only how to schedule things into the future, and `Clock` would know only how to get the current time. Of course, them being separate means you can replace one of them without the other one, which will lead to unpredictable behavior if one of them is mock / testing implementation but the other one is not. Maybe that argues for pushing `sleep` into `Clock` and deleting `Scheduler`.

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
```diff
@@ -181,6 +181,44 @@ object ZLayerSpec extends ZIOBaseSpec {
         _      <- env.use_(ZIO.unit)
         actual <- ref.get
       } yield assert(actual)(equalTo(expected))
+    },
+    testM("map does not interfere with sharing") {
```
jdegoes: Good catch. Maybe there's something we can do to make new combinators more "obviously correct" with respect to sharing??

core/shared/src/main/scala/zio/ZLayer.scala
adamgfraser: I agree. I haven't done a deep dive on this yet. Maybe something to do with the ordering of the underlying map implementations? I can look into this.

adamgfraser:
Regarding the Scheduler yes I think we need to do one of two things:
(1) remove `sleep` method from `Clock`, have `ZIO.sleep` depend on `Scheduler`, and replace `Clock` with `Scheduler` in all the method signatures that use it
(2) delete `Scheduler`

The first answer seems slightly more theoretically "right" but I worry that people are so used to thinking about the `Clock` for sleeping that it will be a disruptive change without much benefit (though maybe if you are going to do it there is no better time than when we are redoing the whole environment encoding).

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
adamgfraser: Yes I was thinking about that too. I think like the implementation of `ZManaged` in terms of `Reservation` the underlying implementation in terms of the `MemoMap` has some unavoidable complexity so we want to get to a point where you can do almost anything in terms of a few basic combinators that make sure the right thing happens. Then the message is "use the combinators unless you really know what you are doing".

You can see that working in terms of methods like `map`, `update` and `set` that are all implemented directly or indirectly in terms of `>>>`. The problem for implementing `mapError` is that we can't change the error type with `>>>` or `++`.

I think we might want to think about getting two two basic combinators, a more powerful version of `>>>` that lets you sequentially combine two layers getting access to either the error or success of the first layer, kind of like `foldM`, and then a parallel combinator like `++` or `+!+` that lets you combine two layers in parallel with a function. I think then you could implement everything else in terms of those and have it be "right".

What do you think?

core/shared/src/main/scala/zio/ZLayer.scala
adamgfraser: It looks like when `l` and `r` are combined with `union` in `++` we get a `Has` instance that has two entries in the map associated with the `Clock`, which is leading to the indeterminacy.

adamgfraser: More updates on this. I think the issue is that in addition to having a string representation type tags have a hash code and the hash code is not always the same for the same type. In particular, it appears that if a type tag is summoned from different compilation units the hash code will be different. So for example if I print `implicitly[Tagged[Has[Clock.Service]]]` from `zio.clock` and `zio.test.environment.EnvironmentSpec` they will have different hash codes. That's why we're getting duplicate entries in the map underlying `Has`. When we do `unionAll` in `Has` we're assuming that we will just overwrite the old implementations of the same service but instead we're adding a second one.

I think we probably want to say that two tags are equal if they have are for the same type so I think we probably either need our own type to wrap tags that hashes them based on their string representation or a specialized map for type tags that hashes them based on their string representation.

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
jdegoes: I like that idea. I also think we should hide the constructor so people can _only_ use the provided methods. I'd rather users come to us and say, "Please add a new operator" then take the chance they get the memo map & finalizers right.

core/shared/src/main/scala/zio/ZLayer.scala
jdegoes: Oh my. Well, Kai did warn us. 😆  

I think we absolutely will replace `TypeTag` but until then we could wrap it in something like:

```scala
case class FixedTypeTag[A](value: TypeTag[A]) {
  override def equals(that: Any): Boolean = 
    that match { case FixedTypeTag(that) => value.toString == that.toString }
  override def hashCode = value.toString.hashCode
}
```

Then `Tagged[A]` becomes `FixedTypeTag[A]` and no ZIO code (outside that package object) needs to know how broken it is. Obviously we gotta fix the problem at the source for RC18.

jdegoes:
> The first answer seems slightly more theoretically "right" but I worry that people are so used to thinking about the Clock for sleeping that it will be a disruptive change without much benefit (though maybe if you are going to do it there is no better time than when we are redoing the whole environment encoding).

I think I agree with deleting `Schedule`. To minimize disruption and also because, in practice, it seems like implementing one requires implementing the other.

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
adamgfraser: I made the constructor private when I implemented the new encoding so it is only us who can get it wrong.

adamgfraser:
Merging now. Will follow up on type tags, scheduler, and simplifications to internal combinators in separate PRs.

jdegoes:
Awesome. Thanks for your work on this!

PR #2874, opened 2020-02-11
Implement Deprecated DefaultRuntime

### LINKED ISSUE #2873: ZIO Core: Add temporary DefaultRuntime to assist with migration
I think we should bring back a backward-compatible `DefaultRuntime` for RC18 and ship it `@deprecated` so that people will know it's going away in 1.0. This will assist with the migration. Also in the docs we can point people to the more official way to do things: with the bootstrap runtime and `provideLayer`.

To implement this ticket:

 - Dig back into RC17 code and look at `DefaultRuntime`
 - Implement new `DefaultRuntime` with the same methods; it will use `Runtime.unsafeFromLayer` in order to create the `ZEnv` environment that is used by all the `unsafeRun` methods
 - Add Scaladoc to describe the new preferred way, which is to use `Runtime.global` and provide layers directly via `ZIO#provideLayer`; explaining that, in general, layers are managed resources and require allocation and deallocation and therefore that scoping their resources is highly recommended as a best practice, but pointing to `Runtime.unsafeFromLayer` as a fallback that can be used when the 'best practice' is too inconvenient

adamgfraser:
Resolves #2873. I renamed the existing `DefaultRuntime` that has `Unit` for the environment type to `BootstrapRuntime`.

### DIFF
--- a/core/shared/src/main/scala/zio/DefaultRuntime.scala
+++ b/core/shared/src/main/scala/zio/DefaultRuntime.scala
@@ -18,12 +18,16 @@ package zio
 
 import zio.internal.Platform
 
-trait DefaultRuntime extends Runtime[Unit] {
-  val environment: Unit = ()
-
-  /**
-   * The platform of the runtime, which provides the essential capabilities
-   * necessary to bootstrap execution of tasks.
-   */
-  val platform: Platform = Platform.default
+@deprecated(
+  "Use `Runtime.global` and provide the required environment directly using" +
+    "`ZIO#provideLayer`. In general, layers are managed resources that " +
+    "require allocation and deallocation and therefore scoping resources to" +
+    "an effect is highly recommended as a best practice. " +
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/zio/DefaultRuntime.scala
```diff
@@ -18,12 +18,16 @@ package zio
...
-trait DefaultRuntime extends Runtime[Unit] {
-  val environment: Unit = ()
-
-  /**
-   * The platform of the runtime, which provides the essential capabilities
-   * necessary to bootstrap execution of tasks.
-   */
-  val platform: Platform = Platform.default
+@deprecated(
+  "Use `Runtime.global` and provide the required environment directly using" +
```
jdegoes:
```suggestion
  "Use `Runtime.default` and provide the required environment directly using" +
```

(My bad)

PR #2875, opened 2020-02-11
ZIO Test: Gracefully Handle Malformed Class Names in Assertion#isSubtype

### LINKED ISSUE #2861: ZIO test: java.lang.InternalError: Malformed class name
I think this is the same error as [this](https://github.com/mockito/mockito-scala/issues/117).

I have an exception defined in something similar as
```
sealed trait Exception
object Exception {
  case class MyException() extends Exception
}
```

When I try, in a test, to do `fails(isSubtype[Exception.MyException](anything))`,  I get an internal error
```
java.lang.InternalError: Malformed class name
	at java.lang.Class.getSimpleName(Class.java:1330)
	at zio.test.Assertion$.isSubtype(Assertion.scala:527)
```

If I redefine my Exception to be
```
sealed trait Exception
case class MyException() extends Exception
```
and my test to use `fails(isSubtype[MyException](anything))`, it works

adamgfraser:
Resolves #2861. We try to call `simpleClassName`. If that fails we fall back to calling `getName`. This was the vast majority of users still get the more readable simple class name. If the user provides us with a class with a degenerate name they just get a longer class name.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Assertion.scala
+++ b/test/shared/src/main/scala/zio/test/Assertion.scala
@@ -111,6 +111,16 @@ object Assertion extends AssertionVariants {
     final case class Infix(left: RenderParam, op: String, right: RenderParam)    extends Render
 
     /**
+     * Creates a string representation of a class name.
+     */
+    def className[A](C: ClassTag[A]): String =
+      try {
+        C.runtimeClass.getSimpleName
+      } catch {
+        case _: Throwable => C.runtimeClass.getName
+      }
+
+    /**
      * Creates a string representation of a field accessor.
      */
     def field(name: String): String =
@@ -585,7 +595,7 @@ object Assertion extends AssertionVariants {
    * }}}
    */
   def isSubtype[A](assertion: Assertion[A])(implicit C: ClassTag[A]): Assertion[Any] =
-    Assertion.assertionRec("isSubtype")(param(C.runtimeClass.getSimpleName))(assertion) { actual =>
+    Assertion.assertionRec("isSubtype")(param(className(C)))(assertion) { actual =>
       if (C.runtimeClass.isAssignableFrom(actual.getClass())) Some(actual.asInstanceOf[A])
       else None
     }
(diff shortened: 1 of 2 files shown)


vasilmkd:
I think I've seen similar code by @neko-kai on the internet. Maybe he can take another look.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -110,6 +110,16 @@ object Assertion extends AssertionVariants {
...
     final case class Infix(left: RenderParam, op: String, right: RenderParam)    extends Render
 
+    /**
+     * Creates a string representation of a class name.
+     */
+    def className[A](C: ClassTag[A]): String =
+      try {
+        C.runtimeClass.getSimpleName
+      } catch {
+        case _: Throwable => C.runtimeClass.getName
```
neko-kai: You could limit the catch to just `InternalError` ([as in the code @vasilmkd was referring to most likely](https://github.com/7mind/izumi/blob/2c7d27fb24fa467866aa08b3b9c3de239606b18a/fundamentals/fundamentals-reflection/src/main/scala/izumi/fundamentals/reflection/Tags.scala#L27))

adamgfraser: Good idea. Will do. Thank you!

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -110,6 +110,17 @@ object Assertion extends AssertionVariants {
...
     final case class Infix(left: RenderParam, op: String, right: RenderParam)    extends Render
 
+    /**
+     * Creates a string representation of a class name.
+     */
+    def className[A](C: ClassTag[A]): String =
+      try {
+        C.runtimeClass.getSimpleName
+      } catch {
+        case t: InternalError if t.getMessage == "Malformed class name" =>
```
regiskuckaertz: It may be worth mentioning the Scala bug here, as this code can probably be dropped at some point in the future.

adamgfraser: Good idea. Done now.

PR #2878, opened 2020-02-11
Handle Infinite Duration Time To Live In ZIO#cached

### LINKED ISSUE #2876: ZIO.cached doesn't work for ttl = Duration.Infinity
```
println(unsafeRun(for {
    ref <- Ref.make(0)
    getAndIncrement = ref.modify(curr => (curr, curr + 1))
    cached <- getAndIncrement.cached(Duration.Infinity)
    a <- cached
    b <- cached
    c <- cached
} yield (a, b, c)))
```
prints:
```
(0, 1, 2)
```
expected output:
```
(0, 0, 0)
```

Also, please note the javadoc on `System.nanoTime()`:

> To compare two nanoTime values
> 
>  long t0 = System.nanoTime();
>  ...
>  long t1 = System.nanoTime();
> one should use t1 - t0 < 0, not t1 < t0, because of the possibility of numerical overflow.

https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--

adamgfraser:
Resolves #2876.


jdegoes:
@adamgfraser Good catch!

PR #2880, opened 2020-02-12
Add ZSTM.foldLeft and ZSTM.foldRight

mijicd:
Part of #2802.


mijicd:
@regiskuckaertz I find that one confusing as well. That one's on ZIO as well.

PR #2883, opened 2020-02-12
Add ZIO#doUntil, doWhile, retryUntil and retryWhile equals and effectful variants

saraiva132:
A couple of combinators missing in ZIO that were present in scheduler that could be added for free.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -378,10 +378,34 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
     repeat(Schedule.doUntil(f))
 
   /**
+   * Repeats this effect until its result is equal to the predicate.
+   */
+  final def doUntilEquals[A1 >: A](a: => A1): ZIO[R, E, A1] =
+    repeat(Schedule.doUntilEquals(a))
+
+  /**
+   * Repeats this effect until its result satisfies the specified effectful predicate.
+   */
+  final def doUntilM(f: A => UIO[Boolean]): ZIO[R, E, A] =
+    repeat(Schedule.doUntilM(f))
+
+  /**
    * Repeats this effect while its result satisfies the specified predicate.
    */
   final def doWhile(f: A => Boolean): ZIO[R, E, A] =
     repeat(Schedule.doWhile(f))
+
+  /**
+   * Repeats this effect for as long as the predicate is equal to its result.
+   */
+  final def doWhileEquals[A1 >: A](a: => A1): ZIO[R, E, A1] =
+    repeat(Schedule.doWhileEquals(a))
+
+  /**
+   * Repeats this effect while its result satisfies the specified effectful predicate.
+   */
+  final def doWhileM(f: A => UIO[Boolean]): ZIO[R, E, A] =
+    repeat(Schedule.doWhileM(f))
 
   /**
    * Returns an effect whose failure and success have been lifted into an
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1340,15 +1364,39 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
+   * Retries this effect until its error satisfies the specified effectful predicate.
+   */
+  final def retryUntilM(f: E => UIO[Boolean])(implicit ev: CanFail[E]): ZIO[R, E, A] =
+    retry(Schedule.doUntilM(f))
+
   /**
    * Retries this effect while its error satisfies the specified predicate.
    */
-  final def retryWhile(f: E => Boolean): ZIO[R, E, A] =
+  final def retryWhile(f: E => Boolean)(implicit ev: CanFail[E]): ZIO[R, E, A] =
```
adamgfraser: Good catch!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1340,15 +1364,39 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
     retry(Schedule.doWhile(f))
 
+  /**
+   * Repeats this effect for as long as the error equals the predicate.
+   */
+  final def retryWhileEquals[E1 >: E](e: => E1)(implicit ev: CanFail[E1]): ZIO[R, E1, A] =
+    retry(Schedule.doWhileEquals(e))
```
adamgfraser: You need to change the parameter to `Schedule.doWhileEquals` to be by name if you want this to actually be lazy.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1340,15 +1364,39 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
    */
-  final def retryUntil(f: E => Boolean): ZIO[R, E, A] =
+  final def retryUntil(f: E => Boolean)(implicit ev: CanFail[E]): ZIO[R, E, A] =
     retry(Schedule.doUntil(f))
 
+  /**
+   * Retries this effect until its error equals the predicate.
+   */
+  final def retryUntilEquals[E1 >: E](e: => E1)(implicit ev: CanFail[E1]): ZIO[R, E1, A] =
+    retry(Schedule.doUntilEquals(e))
```
adamgfraser: Same comment as below regarding `Schedule.doUntilEquals`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1340,15 +1364,39 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   /**
    * Retries this effect until its error satisfies the specified predicate.
    */
-  final def retryUntil(f: E => Boolean): ZIO[R, E, A] =
+  final def retryUntil(f: E => Boolean)(implicit ev: CanFail[E]): ZIO[R, E, A] =
```
adamgfraser: Nice!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -377,12 +377,36 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
   final def doUntil(f: A => Boolean): ZIO[R, E, A] =
     repeat(Schedule.doUntil(f))
 
+  /**
+   * Repeats this effect until its result is equal to the predicate.
+   */
+  final def doUntilEquals[A1 >: A](a: => A1): ZIO[R, E, A1] =
```
mijicd: I believe you can omit the type bound here. Also, as Adam pointed out, `doUntilEquals` accepts parameter by value.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -377,12 +377,36 @@ sealed trait ZIO[-R, +E, +A] extends Serializable { self =>
...
   /**
    * Repeats this effect while its result satisfies the specified predicate.
    */
   final def doWhile(f: A => Boolean): ZIO[R, E, A] =
     repeat(Schedule.doWhile(f))
 
+  /**
+   * Repeats this effect for as long as the predicate is equal to its result.
+   */
+  final def doWhileEquals[A1 >: A](a: => A1): ZIO[R, E, A1] =
```
mijicd: Same here.

core/shared/src/main/scala/zio/ZIO.scala
mijicd: Ah, scratch it, forgot about the variance. Laziness it is :).

PR #2885, opened 2020-02-12
Add fromIterable, fromIterator constructor methods

jkobejs:
I'm aware that design is not final yet, but I think that implementation will not change much with new design.

I see that signature of `fromIterator` in `zio.stream.ZStream` is `def fromIterator[R, E, A](iterator: ZIO[R, E, Iterator[A]]): ZStream[R, E, A] `.  What is reason to have iterator inside `ZIO`?

Also, these methods are currently implemented using `StreamEffect`, I suppose that reason is optimization, can someone point me to some resources or explain it to me quickly?
Will similar optimization be necessary with new encoding?


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -128,4 +128,37 @@ object ZStream extends Serializable {
...
+   * Creates a stream from an iterable collection of values
+   *
+   * @tparam A the iterable value type
+   * @return a stream which emits values from iterable
+   */
+  def fromIterable[A](as: => Iterable[A]): ZStream[Any, Nothing, Any, Unit, A] =
+    ZStream {
+      Managed.effectTotal {
+        Control(
+          IO.succeedNow(as.iterator).flatMap(iter => if (iter.hasNext) IO.succeed(iter.next()) else Pull.endUnit),
```
regiskuckaertz: It looks like you could just write `fromIterator(as.iterator)` for this function.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -128,4 +128,37 @@ object ZStream extends Serializable {
...
+   * Creates a stream from an iterator
+   *
+   * @tparam A the value type of elements produced by iterator
+   * @return a stream which emits values produced by iterator
+   */
+  def fromIterator[A](iterator: => Iterator[A]): ZStream[Any, Nothing, Any, Unit, A] =
+    ZStream {
+      Managed.effectTotal {
+        Control(
+          IO.succeedNow(iterator).flatMap(iter => if (iter.hasNext) IO.succeed(iter.next()) else Pull.endUnit),
```
iravid: Pulling from iterators is effectful, so this should be rewritten as:
```
IO.effectSuspendTotal {
  if (iter.hasNext) UIO(iter.next())
  else Pull.endUnit
}
```

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
iravid: Yes, that'd be preferable. I think the current version does `fromChunk(Chunk.fromIterable(as))` which should be faster in most common cases.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Shouldn't the `IO.succeedNow(iterator)` be better placed outside of `Control`?

iravid: It's inside `Managed.effectTotal`, so I think we could just do...

```
Managed.effectTotal {
  val it = iterator
  Control(
    IO.effectSuspendTotal {
      // ...
    },
    Command.noop
}
```

Requires testing though.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
jkobejs: That makes sense, I'll implement it using chunk.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
jkobejs: Changed argument type of method to `ZIO[R, E, Iterator[A]]` and implemented it in terms of `fromEffect` and `flatMap`

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -454,6 +454,39 @@ object ZStream extends Serializable {
...
+   * @param iterator effect that produces iterator
+   * @return a stream which emits values produced by iterator
+   */
+  def fromIterator[R, E, A](iterator: ZIO[R, E, Iterator[A]]): ZStream[R, E, Any, Unit, A] =
+    fromEffect(iterator).flatMap(iter =>
+      ZStream {
+        Managed.effectTotal {
+          Control(
+            IO.effectSuspendTotal {
+              if (iter.hasNext) UIO.succeed(iter.next()) else Pull.endUnit
```
regiskuckaertz:
```suggestion
              if (iter.hasNext) UIO.succeedNow(iter.next()) else Pull.endUnit
```
☝️ I think this one _must_ be strict

jkobejs:
> Not sure about the wrapping of the iterator in ZIO (usually we use by-name parameters), but otherwise looking great, just a few nits.

I change it to match current signature https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L3331
What do you think, should we change it to use only by-name parameters?

iravid:
Hi @jkobejs. Sorry to keep this open longer - could you reopen against #2934? Closing this one out for now.

jkobejs:
Ofc

PR #2889, opened 2020-02-13
Add buffer, into, toQueue

regiskuckaertz:
I supposed `Take` must remain with a minor modification for the end marker.


regiskuckaertz:
Flaky `validateFirstPar` is unrelated.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -76,6 +172,29 @@ object ZStream extends Serializable {
...
+          val l = c.length
+          val pull = iRef.get.flatMap { i =>
+            if (i >= l)
+              Pull.endUnit
+            else
+              iRef.update(_ + 1) *> Pull.emit(c(i))
+          }
+          Reservation(UIO.succeedNow(Control(pull, Command.noop)), _ => UIO.unit)
+        }
+      }
```
vasilmkd: `Managed.succeed`?

vasilmkd:
Yeah, I supppose `Take` will have to remain as it is a nice way to reify a take from the stream. I use it extensively in the more complicated versions of `buffer`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Isn't it necessary to allocate the `Ref` outside of the reservation?

regiskuckaertz:
I saw that 😵

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
vasilmkd: I don't think so, check the implementation in `StreamEffect.scala`.

regiskuckaertz: ha yeah, great point 🤦‍♂

regiskuckaertz: but it has to be created outside of the `Control`, right? So perhaps `Managed.fromEffect` instead 🤔

vasilmkd: It's basically nitpicking at this point, but maybe this:
```scala
ZStream {
  for {
    ref  <- Ref.make(0).toManaged_
    len  = c.length
    pull = ref.get.flatMap { idx =>
      if (idx >= len) Pull.endUnit
      else ref.set(idx + 1) *> Pull.emit(c(idx))
    }
  } yield Control(pull, Command.noop)
}
```

streams/shared/src/main/scala/zio/stream/experimental/Take.scala
```diff
@@ -0,0 +1,72 @@
...
+      case (_, t @ Take.Fail(_))          => t
+    }
+}
+
+object Take {
+  final case class Fail[+E](value: Cause[E]) extends Take[E, Nothing, Nothing]
+  final case class Value[+A](value: A)       extends Take[Nothing, Nothing, A]
+  final case class End[+B](marker: B)        extends Take[Nothing, B, Nothing]
+
+  def fromPull[R, E, B, A](pull: ZIO[R, Either[E, B], A]): ZIO[R, Nothing, Take[E, B, A]] =
```
vasilmkd: This `fromPull` is old, can you replace it with the one from `master`?

regiskuckaertz: Good catch 🙏  fixed.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: I have a slight preference for my version as it has less roundtrips but happy to be convinced otherwise 😊

regiskuckaertz:
The only thing I'm not sure of is do we need the lazy and eager variants for `Pull.emit|fail|...`? or just the latter is enough?

regiskuckaertz:
Can think about it later.

PR #2891, opened 2020-02-13
Unify buffer code after Queue#offer changes

vasilmkd:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -764,6 +764,53 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+              _     <- ref.set(p).when(added)
+            } yield ()
+
+          case _ =>
+            for {
+              latch <- ref.get
+              _     <- latch.await
+              p     <- Promise.make[Nothing, Unit]
+              _     <- queue.offer((take, p))
+              _     <- ref.set(p)
```
regiskuckaertz: what if offer returns false in this case?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -764,6 +764,53 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
+          case Take.Value(_) =>
+            for {
+              p     <- Promise.make[Nothing, Unit]
+              added <- queue.offer((take, p))
+              _     <- ref.set(p).when(added)
+            } yield ()
+
+          case _ =>
+            for {
+              latch <- ref.get
```
regiskuckaertz: just curious: why is it necessary to wait on error/failure?

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Good question. It will not happen, because this code is blocked until the queue is drained (it is always waiting on the last enqueued promise, and the taker always unblocks promises as it encounters them), so there will always be space in the queue for this offer, no matter its type.

streams/shared/src/main/scala/zio/stream/ZStream.scala
vasilmkd: Continuing my comment from above, we are forcing a complete pause of the pipeline, so that no error is dropped. This way we wait until the whole queue is drained, enqueue the error, and wait until it is dequeued.

vasilmkd: It is undesirable to silently swallow errors.

vasilmkd:
They are not noob questions, it wasn't even completely obvious to me when I first implemented it!

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: aaaah yeah I see, in case the queue is dropping when full, you may lose errors indeed. Great 👍

vasilmkd:
It would have worked the first time, but the sliding `Queue#offer` fix was necessary.

PR #2893, opened 2020-02-13
ZIO Core: Simplify Layer Combinators

adamgfraser:
The last item we had from #2864 was to try to simplify some of the `ZLayer` combinators so that new implementations would be more "obviously right".  I tried to do that here by implementing `zipWith` that lets you take the input of two layers and combine their outputs using an arbitrary function. With this we can reimplement `++`, `+!+`, and could have implemented `overwrite` if we had needed it.

The other function we need to unify the existing combinators is something like `>>>` but that is able to handle errors so we can implement both `>>>` and `mapError` in terms of it. The best I came up with is:

```scala
  def foldM[E1, ROut2 <: Has[_]](failure: ZLayer[E, E1, ROut2], success: ZLayer[ROut, E1, ROut2]): ZLayer[RIn, E1, ROut2] =
    new ZLayer(
      Managed.finalizerRef(_ => UIO.unit).map { finalizerRef => memoMap =>
        memoMap.getOrElseMemoize(self, finalizerRef).foldM(
          e => memoMap.getOrElseMemoize(failure, finalizerRef).provide(e),
          r => memoMap.getOrElseMemoize(success, finalizerRef).provide(r)
        )
      }
    )
```

This lets you compose another layer based on either the error or success values. But it requires you to have a concept of layers that take errors as input, which while a logical extension of the idea of layers as arrows I think is potentially confusing. The name is also not quite right.

Not sure if we can do better or it makes sense to just leave those combinators as is. The constructor is already private so there is no risk of users implementing their own combinators that don't do the right thing with the memo map.

Another combinator I thought about is `eitherWith` that would let you combine two layers and get a layer that took either of their inputs and gave the corresponding output. Not sure how useful that would be in practice.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -61,26 +61,12 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
   def ++[E1 >: E, RIn2, ROut1 >: ROut <: Has[_], ROut2 <: Has[_]](
     that: ZLayer[RIn2, E1, ROut2]
   )(implicit tagged: Tagged[ROut2]): ZLayer[RIn with RIn2, E1, ROut1 with ROut2] =
-    new ZLayer(
-      Managed.finalizerRef(_ => UIO.unit).map { finalizerRef => memoMap =>
-        for {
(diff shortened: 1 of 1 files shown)


jdegoes:
I'd call it `fold` but looks good to me!

adamgfraser:
Okay, will submit a follow up PR for that.

PR #2894, opened 2020-02-13
ZIO Core: Implement ZLayer#fold

adamgfraser:
Implements `ZLayer#fold` and reimplements `>>>` and `mapError` in terms of `fold`. With this change, all combinators on `ZLayer` will be implemented in terms of `fold` and `zipWith`.


jdegoes:
❤️

PR #2896, opened 2020-02-14
ZStream: flatMap (and ZSink.collectAll)

regiskuckaertz:
Had to add `ZSink.collectAll` as it is used in all the tests. Some tests will have to be commented out until more combinators have been migrated, but I don't think this is bad as I'm reusing most of the existing implementation.

cc @vasilmkd

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
@@ -48,11 +48,11 @@ object ZSink extends Serializable {
     new ZSink(process)
 
   /**
-    * Accumulates all incoming elements into a list.
-    *
-    * @tparam A the type of elements
-    * @return a sink accumulating (forever) incoming elements
-    */
+   * Accumulates all incoming elements into a list.
+   *
+   * @tparam A the type of elements
+   * @return a sink accumulating (forever) incoming elements
+   */
   def collectAll[A]: ZSink[Any, Nothing, List[A], A, Nothing] = {
     import scala.collection.mutable.ListBuffer
     ZSink[Any, Nothing, List[A], A, Nothing] {
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
@@ -53,14 +53,12 @@ object ZSink extends Serializable {
    * @tparam A the type of elements
    * @return a sink accumulating (forever) incoming elements
    */
-  def collectAll[A]: ZSink[Any, Nothing, List[A], A, Nothing] = {
-    import scala.collection.mutable.ListBuffer
+  def collectAll[A]: ZSink[Any, Nothing, List[A], A, Nothing] =
     ZSink[Any, Nothing, List[A], A, Nothing] {
       for {
-        buf   <- Ref.make(ListBuffer.empty[A]).toManaged_
-        push  = (a: A) => buf.update(_ :+ a)
-        query = buf.get.map(_.toList)
+        buf   <- Ref.make(List.empty[A]).toManaged_
+        push  = (a: A) => buf.update(a :: _)
+        query = buf.get.map(_.reverse)
       } yield Control(push, query)
     }
-  }
 }
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
@@ -46,4 +46,21 @@ object ZSink extends Serializable {
    */
   def apply[R, E, M, A, B](process: ZManaged[R, E, ZSink.Control[R, E, M, A, B]]): ZSink[R, E, M, A, B] =
     new ZSink(process)
+
+  /**
+    * Accumulates all incoming elements into a list.
+    *
+    * @tparam A the type of elements
+    * @return a sink accumulating (forever) incoming elements
+    */
+  def collectAll[A]: ZSink[Any, Nothing, List[A], A, Nothing] = {
+    import scala.collection.mutable.ListBuffer
+    ZSink[Any, Nothing, List[A], A, Nothing] {
+      for {
+        buf   <- Ref.make(ListBuffer.empty[A]).toManaged_
+        push  = (a: A) => buf.update(_ :+ a)
+        query = buf.get.map(_.toList)
+      } yield Control(push, query)
+    }
+  }
 }
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -34,6 +34,73 @@ class ZStream[-R, +E, -M, +B, +A](
         }
       } yield Control(pull, Command.noop)
     }
+  
+  /**
+   * Returns a stream made of the concatenation in strict order of all the streams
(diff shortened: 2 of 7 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -115,6 +182,36 @@ class ZStream[-R, +E, -M, +B, +A](
       _ <- pull.toManaged_
     } yield ()
 
+  /**
+   * Runs the sink on the stream to produce either the sink's internal state or an error.
+   */
+  def runQuery[R1 <: R, E1 >: E, A1 >: A, B](sink: ZSink[R1, E1, B, A1, Nothing]): ZIO[R1, E1, B] =
```
vasilmkd: I think we can allow any sink here, not just sinks that don't terminate. I'd say we need to change the `Nothing` to `Any`.

regiskuckaertz: Yeah I had that initially, just wondered whether this is something we'd want to impose, but you're right it is clear from the type at we never use the output anyway.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -35,6 +35,73 @@ class ZStream[-R, +E, -M, +B, +A](
...
+            _           <- currCmd.set(Some(control.command))
+          } yield ())
+        }
+      }
+
+      currPull.get.flatMap {
+        case None =>
+          pullOuter.foldCauseM(
+            _.failureOrCause match {
+              case Left(Left(e))  => Pull.failNow(e)
```
iravid: It'd also be nice in the implementation of the above to not use `failureOrCause`, as that potentially "loses" other content in the `Cause`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -35,6 +35,73 @@ class ZStream[-R, +E, -M, +B, +A](
...
+              case Right(c)       => Pull.haltNow(c)
+            },
+            Pull.emitNow(_)
+          )
+      }
+    }
+
+    ZStream {
+      for {
+        currPull  <- Ref.make[Option[ZIO[R1, Either[E1, B1], C]]](None).toManaged_
```
iravid: The previous implementation defaults to `Pull.end` here to avoid the `Option`. Would that work here?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -115,6 +182,36 @@ class ZStream[-R, +E, -M, +B, +A](
...
     } yield ()
 
+  /**
+   * Runs the sink on the stream to produce either the sink's internal state or an error.
+   */
+  def runQuery[R1 <: R, E1 >: E, A1 >: A, B](sink: ZSink[R1, E1, B, A1, Nothing]): ZIO[R1, E1, B] =
+    (self.process <*> sink.process).use {
+      case (command, control) =>
+        def pull: ZIO[R1, E1, B] =
+          command.pull.foldM(
```
iravid: Love this implementation. The new encoding looks to be paying off so far!

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
iravid: It'd be nice to get variants of all the `run` methods that preserve the `ZManaged` for composition. In fact, the ZIO variants can be implemented in terms of those.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -35,6 +35,73 @@ class ZStream[-R, +E, -M, +B, +A](
...
+            _           <- currPull.set(Some(control.pull))
+            _           <- currCmd.set(Some(control.command))
+          } yield ())
+        }
+      }
+
+      currPull.get.flatMap {
+        case None =>
+          pullOuter.foldCauseM(
+            _.failureOrCause match {
```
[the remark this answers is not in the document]
vasilmkd: `sequenceCauseEither`, called it. 🤣

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
vasilmkd: Not sure, and this is a pain point of returning a value at the end, we don't have the value to return. :/

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -46,4 +46,21 @@ object ZSink extends Serializable {
...
+   * Accumulates all incoming elements into a list.
+   *
+   * @tparam A the type of elements
+   * @return a sink accumulating (forever) incoming elements
+   */
+  def collectAll[A]: ZSink[Any, Nothing, List[A], A, Nothing] = {
+    import scala.collection.mutable.ListBuffer
+    ZSink[Any, Nothing, List[A], A, Nothing] {
+      for {
+        buf   <- Ref.make(ListBuffer.empty[A]).toManaged_
```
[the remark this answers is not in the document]
regiskuckaertz: Done; I didn't know that 👍

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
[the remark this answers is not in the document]
regiskuckaertz: Did that 😆

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: We will still be losing information with `Cause.sequenceCauseEither`, as soon as you have an `B` you must drop everything else.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: You can't because there is no `B` to put into `Pull.end`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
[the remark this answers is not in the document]
vasilmkd: I know, I was just joking around.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Yes, so clear!

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Done

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Hmm, in `sequenceOption` it was dropping the information that there had been a `None` along the way, but I'm not sure I can drop the `B` here.

regiskuckaertz:
Right, I think this is ready save for the question mark around `sequenceCauseEither`.

vasilmkd:
Can we press on or do you consider this a blocking issue?

regiskuckaertz:
press on 😁

PR #2899, opened 2020-02-14
Port ZStream#ensuring,take,runDrain; ZStream.bracket,bracketExit; ZSi…

iravid:
…nk.drain

More tests that are commented out. Deferred those so the branch doesn't get huge.

@vasilmkd @regiskuckaertz

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -36,6 +36,26 @@ class ZStream[-R, +E, -M, +B, +A](
     }
 
   /**
+   * Executes the provided finalizer after this stream's finalizers run.
+   *
+   * @tparam R1 the environment required by the finalizer
+   * @param fin the finalizer
+   * @return a stream that executes the provided finalizer in addition to the existing ones
+   */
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -213,22 +267,21 @@ class ZStream[-R, +E, -M, +B, +A](
...
+   * @return a stream with `n` or less elements
    */
-  final def runCollectManaged: ZManaged[R, E, List[A]] = runQueryManaged(ZSink.collectAll[A])
+  def take(n: Long): ZStream[R, E, M, Option[B], A] =
+    ZStream {
+      for {
+        as      <- self.process
+        counter <- Ref.make(0L).toManaged_
+        pull = counter.get.flatMap { c =>
+          if (c >= n) Pull.end(None)
```
regiskuckaertz: The question is do we want the marker to be permanently `Some(b)` in the event the stream is smaller than n elements and downstream keep pulling ad infinitum, or is it fine to eventually switch to `None`?

regiskuckaertz: Ha scratch that, the update will only happen if the pull is successful 🤦‍♂

PR #2903, opened 2020-02-14
add mergeAll to (Z)STM companion object

sken77:

### DIFF
--- a/core/shared/src/main/scala/zio/stm/STM.scala
+++ b/core/shared/src/main/scala/zio/stm/STM.scala
@@ -178,9 +178,9 @@ object STM {
   /**
    * @see See [[zio.stm.ZSTM.mergeAll]]
    */
-  def mergeAll[R, E, A, B](
-    in: Iterable[ZSTM[R, E, A]]
-  )(zero: B)(f: (B, A) => B): ZSTM[R, E, B] = ZSTM.mergeAll(in)(zero)(f)
+  def mergeAll[E, A, B](
+    in: Iterable[ZSTM[Any, E, A]]
+  )(zero: B)(f: (B, A) => B): ZSTM[Any, E, B] = ZSTM.mergeAll(in)(zero)(f)
 
   /**
    * @see See [[zio.stm.ZSTM.partial]]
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1085,6 +1085,14 @@ object ZSTM {
    */
   val none: STM[Nothing, Option[Nothing]] = succeedNow(None)
 
+  /**
+   * Merges an `Iterable[IO]` to a single IO, working sequentially.
```
mijicd: `IO` should be replaced with `ZSTM`.

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -178,9 +178,9 @@ object STM {
   /**
    * @see See [[zio.stm.ZSTM.mergeAll]]
    */
-  def mergeAll[R, E, A, B](
-    in: Iterable[ZSTM[R, E, A]]
-  )(zero: B)(f: (B, A) => B): ZSTM[R, E, B] = ZSTM.mergeAll(in)(zero)(f)
+  def mergeAll[E, A, B](
+    in: Iterable[ZSTM[Any, E, A]]
```
mijicd: `STM[E, A]`

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -178,9 +178,9 @@ object STM {
...
+  )(zero: B)(f: (B, A) => B): ZSTM[Any, E, B] = ZSTM.mergeAll(in)(zero)(f)
```
mijicd: Same here.

sken77: that was sloppy sorry about that

mijicd: No worries :)

sken77:
I changed the implementation of `mergeAll` to use `zipWith` instead of zip

PR #2904, opened 2020-02-15
Migrate ZStream.die|fail|halt|empty|never|range|iterate and Stream#takeWhile

regiskuckaertz:
Refrained from doing `unit` as it builds on `concat` which is a beast that needs to be tackled on its own.

@vasilmkd

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -504,16 +504,13 @@ object ZStream extends Serializable {
    */
   def iterate[A](a: A)(f: A => A): UStream[A] =
     ZStream {
-      Managed.effectTotal {
-        var currA = a
-        Control(
-          ZIO.effectTotal {
-            val ret = currA
-            currA = f(currA)
-            ret
-          },
-          Command.noop
-        )
+      Managed.fromEffect {
+        Ref.make(a).map { currA =>
+          Control(
+            currA.modify(a => f(a) -> a),
+            Command.noop
+          )
+        }
       }
     }
 
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -507,7 +507,7 @@ object ZStream extends Serializable {
       Managed.fromEffect {
         Ref.make(a).map { currA =>
           Control(
-            currA.modify(a => f(a) -> a),
+            currA.modify(a => a -> f(a)),
             Command.noop
           )
         }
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -425,6 +469,34 @@ object ZStream extends Serializable {
...
+  def halt[E](cause: => Cause[E]): Stream[E, Nothing] =
+    fromEffect(ZIO.halt(cause))
+
+  /**
+   * The infinite stream of iterative function application: a, f(a), f(f(a)), f(f(f(a))), ...
+   */
+  def iterate[A](a: A)(f: A => A): UStream[A] =
+    ZStream {
+      Managed.effectTotal {
+        var currA = a
```
iravid: Do we want to encourage this style in the ZStream module? It was previously confined to StreamEffect (for no real reason).

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -454,6 +526,21 @@ object ZStream extends Serializable {
...
+  val never: UStream[Nothing] =
+    ZStream(ZManaged.succeedNow(Control(UIO.never, Command.noop)))
+
+  /**
+   * Constructs a stream from a range of integers (lower bound included, upper bound not included)
+   *
+   * @param min the lower bound
+   * @param max the upper bound
+   */
+  def range(min: Int, max: Int): ZStream[Any, Nothing, Any, Option[Unit], Int] =
```
iravid: Should we set B = Unit/Any here?

iravid (commented):
Looks good! Two questions and this is good to go.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: yeah, I think so, it's annoying right now.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Only perhaps for performance-sensitive combinators? of which this one is not, so I'll switch to a `Ref` instead.

iravid: Well I'd prefer they all be implemented in the most performant way possible. We previously forwarded to `StreamEffect`'s companion object for constructors that don't involve effects but that isn't really necessary.

vasilmkd: `StreamEffect` was a nice place to encapsulate the throwing and catching of signal exceptions and keep it separate from the ZIO code in ZStream.

regiskuckaertz: Ok, I just saw that. I'm fine to revert if you want to.

iravid:
@regiskuckaertz Can you reopen against #2934? 🙏🏻

regiskuckaertz:
Hey! Sorry for real silence, yes will do! This is great 👍🥳

PR #2910, opened 2020-02-15
add reduceAll to (Z)STM companion

sken77:


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1115,6 +1115,14 @@ object ZSTM {
   )(f: A => ZSTM[R, E, B])(implicit ev: CanFail[E]): ZSTM[R, Nothing, (List[E], List[B])] =
     ZSTM.foreach(in)(f(_).either).map(ZIO.partitionMap(_)(ZIO.identityFn))
 
+  /**
+   * Reduces an `Iterable[ZSTM]` to a single `ZSTM`, working sequentially.
+   */
+  def reduceAll[R, R1 <: R, E, A](a: ZSTM[R, E, A], as: Iterable[ZSTM[R1, E, A]])(
+    f: (A, A) => A
+  ): ZSTM[R1, E, A] =
+    as.foldLeft[ZSTM[R1, E, A]](a)((l, r) => l.zipWith(r)(f))
```
adamgfraser: Could even be simplified to `as.foldLeft[ZSTM[R1, E, A]](a)(_.zipWith(_)(f))`. We should make the same simplification using `zipWith` on `ZIO` and `ZManaged` and we could do the same thing with the `mergeAll` and `mergeAllPar` implementations for both data types. Basically any time we do `a.zip(b).map(f)` we could do `a.zipWith(b)(f)`.

sken77: i started doing this but it started to get a bit big. for example. the `reduceAllPar` and `mergeAllPar` methods are also implemented in the manner you are saying with `zipPar... map` instead of `zipWithPar` so that would need to be refactored as well. I can open a separate ticket and do those there.

PR #2916, opened 2020-02-16
ZIO Core: Implement ***

adamgfraser:
Implements `***` combinator for ZIO, ZManaged, and ZSTM analogous to existing method on `Schedule`. Simplifies type signatures for `_1`,`_2`, and `swap`. Alphabetizes symbolic operators. Should we rename `_1` and `_2` to `first` and `second`?


jdegoes:
> Should we rename _1 and _2 to first and second?

👍 from me on that.

PR #2921, opened 2020-02-17
add validate & validateFirst to ZSTM companion

sken77:


sken77:
@adamgfraser any ideas on this error?

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -277,6 +277,22 @@ object STM {
   val unit: STM[Nothing, Unit] =
     ZSTM.unit
 
+  /**
+   * @see See [[zio.stm.ZSTM.validate]]
+   */
+  def validate[R, E, A, B](
```
mijicd: There is no `R` in `STM`.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -1045,6 +1045,50 @@ object ZSTMSpec extends ZIOBaseSpec {
         assertM(tx.commit)(equalTo(1))
       }
     ),
+    suite("ZSTM validate")(
+      testM("returns all errors if never valid") {
+        val in  = List.fill(10)(0)
+        val res = ZSTM.validate(in)(a => ZSTM.failNow(a))
```
mijicd: Replace with calls to STM, such that no environment is needed.

core/shared/src/main/scala/zio/stm/STM.scala
sken77: does, partition above also needs then?

sken77: oh i see the commit now

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -279,6 +279,22 @@ object STM {
...
+  )(f: A => STM[E, B])(implicit ev: CanFail[E]): STM[::[E], List[B]] =
+    ZSTM.validate(in)(f)
+
+  /**
+   * @see See [[zio.stm.ZSTM.validateFirst]]
+   */
+  def validateFirst[E, A, B](
+    in: Iterable[A]
+  )(f: A => STM[E, B])(implicit ev: CanFail[E]): STM[List[E], B] =
+    ZSTM.foreach(in)(f(_).flip).flip
```
adamgfraser: Don't we want this to just be an alias for the method on `ZSTM`?

sken77: we do. you can slap me for that one

Issue #2784, opened 2020-02-02
ZIO Test: Supporting pluggable diffing


jdegoes:
It's probably out of scope for ZIO Test to directly support diffing without external dependencies (how does MUnit do it?). Nonetheless, the design of ZIO Test should at minimum support _pluggable_ diffing, and we could prove that out with an integration module for `diffx`.

adamgfraser:
They do it internally. See [here](https://github.com/scalameta/munit/tree/master/munit/shared/src/main/scala/munit/internal/difflib) We may want to consider doing the same to provide the best possible experience with zero dependencies. It looks like they use the Myers diff algorithm, which is the same one used by Github. It would be a bit of a project to add but I don't think it would be too bad. This would do string diffing and wouldn't have special logic for case classes like diffX has. Perhaps string based diffing could be the out of the box solution for ZIO Test. That really covers the "missing case" in assertions where you've zeroed in on the specific element but it is still a large string. And then separately we could potentially support integrations with other diffing modules.

jdegoes:
👍 for built in string diffing (easy/lightweight).

dkarlinsky:
I would like to give this a shot. 
@adamgfraser are we talking only about string diffing or also structured things like case classes?

adamgfraser:
@dkarlinsky That would be great! Initial scope is just string diffing.

dkarlinsky:
Cool @adamgfraser. 
So it seems we would need to have the diff or the the two values to be diffed somehow included in `AssertionValue`. The actual value is there already, so if `Assertion` could optionally have the expected value, we would be able to create the diff in `DefaultTestReporter`, as part of `renderFragment`, for example.
So `equalTo` assertion will have the expected value and the test reporter will have a differ that initially only handles string-string diffs.
WDYT?

adamgfraser:
@dkarlinsky I think that makes sense.

dkarlinsky:
So I copied the MUnit difflib into zio-test and hooked it up to `renderFragment`.
This is the output I'm currently getting:
![image](https://user-images.githubusercontent.com/1577551/74238797-317e4c00-4cdf-11ea-824b-8a4fc41aeced.png)
It looks like this algorithm is geared towards multi line inputs, and highlights changes at line granularity. 
I think that a lot of string equality assertions in tests are done on single line strings for which this presentation is not ideal IMO....

adamgfraser:
[...]

PR #2923, opened 2020-02-17
#2784 - adding Myers based diff to rendered string equality assertion failures

dkarlinsky:


dkarlinsky:
@adamgfraser I haven't ironed all the kinks yet but this is an example of diff rendering I currently have:
image
Where the test looks like this:
```scala
test("multiline strings not equal") {
      assert("foo bar\nbazz buzzard")(Assertion.equalTo("bar foo\nbazard"))
    },
```

adamgfraser:
@dkarlinsky Wow! This looks amazing. Will review in detail tomorrow.

dkarlinsky:
@adamgfraser I still have some issues / open questions on how to format multiline diffs, especially when CR/LF char is the only char added/removed...
I'm thinking of perhaps escaping `\n` and `\r` in the result but also adding the line break for better human readability...

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -27,7 +27,8 @@ import zio.{ Cause, Exit, ZIO }
  */
 final class Assertion[-A] private (
   val render: Assertion.Render,
-  val run: (=> A) => AssertResult
+  val run: (=> A) => AssertResult,
+  val expected: Option[Any] = None
```
adamgfraser: Really I think we want to say that the expected value has to be an `A` but we can't easily do that here because `A` is contravariant. What if we had `Assertion` take a `renderDiff: Option[A => String]` parameter (maybe returning a more structured data type than a string). Then `A` would be in contravariant position.

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
```diff
@@ -23,11 +23,13 @@ trait AssertionVariants {
   /**
    * Makes a new assertion that requires a value equal the specified value.
    */
-  final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] =
-    Assertion.assertion("equalTo")(param(expected)) { actual =>
+  final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] = {
```
adamgfraser: We should make sure we replicate any changes here across the other version specific file.

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
```diff
@@ -23,11 +23,13 @@ trait AssertionVariants {
...
+    val assertion: Assertion[B] = Assertion.assertion("equalTo")(param(expected)) { actual =>
       (actual, expected) match {
         case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
         case (left, right)                     => left == right
       }
     }
+    assertion.withExpected(Some(expected))
```
adamgfraser: Do we want to add the expectation for every equality assertion or only those involving strings? This gets a little to the "when do we want to diff?" question.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -22,10 +22,11 @@ import scala.io.AnsiColor
```
adamgfraser: It would be nice to push more of the logic in this file into the `diff` package so the test reporter is really just handing over an actual and expected value and getting back what it should render.

test/shared/src/main/scala/zio/test/diff/Delta.scala
```diff
@@ -0,0 +1,28 @@
+package zio.test.diff
+
+import zio.test.diff.Delta.TYPE.{ CHANGE, DELETE, INSERT }
+import zio.test.diff.Delta._
+
+private[diff] case class Delta[T](`type`: TYPE, original: Chunk[T], revised: Chunk[T])
```
adamgfraser: Maybe make this more Scala idiomatic. Like `diffType` instead of `type` so you don't need to use the back ticks, make `Type` becomes `DiffType` and then the case objects can go to `Change`, `Delete`, etc...

test/shared/src/main/scala/zio/test/Assertion.scala
dkarlinsky: I guess `renderDiff` could return `Vector[DiffElement]`, or some case class wrapping it.
In this case `Assertion.equalTo`  (or `assertEqualString`) will be the one deciding on diff implementation, not `DefaultTestReporter`

adamgfraser: We could do `diffOption: Option(A => Differ => String)` to defer the implementation of the differ to the test reporter.

jdegoes:
@dkarlinsky Would love to see this one make it in! You think you'll be able to get back to it sometime soon?

dkarlinsky:
@jdegoes yes, have pushed bunch of commits addressing @adamgfraser 's comments.
Now working on making the CI pass

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
dkarlinsky: @adamgfraser see my latest commits.
It is still string only by default but you can customise it by passing a different `Diffing` to `equalTo`

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
dkarlinsky: I would love to split `DefaultTestReporter` into to parts, where one translates test results into a set of simple (ish) domain model representing notions like assertion failures and diffs but without any markup stuff like lines, offsets and ANSI colors.
This would be useful for JUnitRunner and IDE integration @hmemcpy is working on, I think. 
And rendering with colors and offsets would be done by a different class.

But I think this should probably be a separate PR

dkarlinsky: I did move `FailureRenderer`, `MessageMarkup` and `DiffRenderer` out into their own files.

test/shared/src/main/scala/zio/test/FailureRenderer.scala
```diff
@@ -0,0 +1,190 @@
...
+import zio.test.mock.MockException.{
+  InvalidArgumentsException,
+  InvalidMethodException,
+  UnexpectedCallExpection,
+  UnmetExpectationsException
+}
+import zio.test.mock.{ Method, MockException }
+import zio.{ Cause, UIO }
+
+object FailureRenderer {
```
adamgfraser: Can this be `private[test]`?

test/shared/src/main/scala/zio/test/MessageMarkup.scala
```diff
@@ -0,0 +1,83 @@
+package zio.test
+
+import scala.io.AnsiColor
+
+object MessageMarkup {
```
adamgfraser: Can this be `private[test]`?

adamgfraser:
@dkarlinsky This looks great! It is a big PR and I'm sure there is more follow up we can do, especially around the `DefaultTestReporter` as you suggested, but I think we could get this in now. My main question at this point is whether we should diff all strings or only strings over a certain size or if you use a combinator that says you want to diff?

dkarlinsky:
@adamgfraser we could make the default `Diffing` to only apply to strings longer than say 20 characters. 
And maybe provide a simple way to construct one with a different limit. 
WDYT?

hmemcpy:
This is great! I started a PR #2975 where I separated the `DefaultTestReporter` and introduced a base trait `TestRenderer` to support different outputs. I'm currently on vacation, so I won't be able to proceed with it until next week, so please go ahead and make the changes you need, I'll rework it.

My main need in that PR is to introduce the ability to support a custom renderer via command line arguments, so I can implement one externally for the IntelliJ test runner (without having to reimplement the command line runner logic).

adamgfraser:
@dkarlinsky That sounds reasonable to me. I don't really have a strong opinion here, I just think it is something we should consider. There is value in having uniformity but at least personally I have been annoyed at times when I am comparing simple strings and get a whole diff from other test frameworks, especially when it is part of a larger test where the difference is the obvious part.

@jdegoes What do you think?

dkarlinsky:
@adamgfraser even if the diff is redundant due to the strings being short, you still also get the message: `str1 didn't satisfy equalTo(str2)` as before. So you can look at that and ignore the diff.

adamgfraser:
@dkarlinsky Can you resolve the conflict and then we will merge?

@jdegoes Had an idea that we could look at the percent of differences to the total length of the string, so if the strings are close we show the diff, if they are nothing alike we just show the two strings. That seems pretty good to me but we can follow up on that later.

PR #2924, opened 2020-02-17
Add back test for effectBlockingInterrupt

fsvehla:
I’ve noticed that `effectBlockingInterrupt` has an `ignore`d test.
I think I came up with a test that is not flaky—it ran locally for some time without issues.

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/blocking/BlockingSpec.scala
@@ -2,9 +2,10 @@ package zio.blocking
 
 import java.util.concurrent.atomic.AtomicBoolean
 
+import zio.Promise
+import zio.ZIO
 import zio.duration._
 import zio.test.Assertion._
-import zio.test.TestAspect.ignore
 import zio.test._
 import zio.{ UIO, ZIOBaseSpec }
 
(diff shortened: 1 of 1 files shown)


fsvehla:
I’m really happy that the commit message didn’t jinx the build

adamgfraser:
Nice! Will take a look.

mlangc:
I've looked into this, and I think that the test is actually fine - starting something, and then cancelling it with a timeout of 0 should always work. If I'm guessing correctly, the flakiness comes from a race condition in `zio.blocking.Blocking.Service#effectBlockingInterrupt`: If `ensuring(interruptThread *> awaitInterruption)` is executed before the fiber that executes the actual effect has started, `awaitInterruption` will wait forever, because no one will ever set the barrier. In my opinion we should fix the implementation, not tweak the test.

adamgfraser:
We also have the issue that this was potentially not running on the correct thread poo before (see https://github.com/zio/zio/pull/2862) so there are several moving pieces here. But completely agree that looking at both the implementation and the test needs to be in scope here.

adamgfraser:
Looking at this more I think @mlangc is correct. @fsvehla Your test demonstrates the problem but we need to push this logic into `effectBlockingInterrupt` so the user can be confident that a blocking effect can be safely interrupted without having to invoke external coordination mechanisms. Do you want to take a crack at this?

fsvehla:
Yes!

jdegoes:
@fsvehla What's the status of this one?? /cc @adamgfraser

fsvehla:
I’m down with a cold. This looks like something that should be fixed before the next RC, so it would be good if someone else could pick it up

jdegoes:
@fsvehla No worries. Please get well soon!

PR #2927, opened 2020-02-18
ZSink for delimited framing of chunked streams.

quelgar:
Both the idea and naming stolen from [Akka Streams](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Framing$.html).

Originally if the maximum frame size was hit, I had it just emit the chunk, essentially acting as if it found a delimiter. But then I thought the Akka behaviour of failing the stream in that case might be better. I've used `IllegalInputException` to indicate failure, any thoughts on something better? Maybe `Unit` works?


jdegoes:
@iravid You available to take a look at this?

@quelgar Thank you for this contribution!

streams-tests/jvm/src/test/scala/zio/stream/FramingSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      val sink = Framing.delimiter(Chunk(-1), 3)
+      val in   = Stream(Chunk(1, 2), Chunk(3, 4, -1, 5))
+      assertM(in.aggregate(sink).runCollect.run)(
+        fails(isSubtype[IllegalArgumentException](anything))
+      )
+    },
+    testM("succeeds if maximum frame length hit exactly") {
+      val sink = Framing.delimiter(Chunk(-1), 4)
+      val in   = Stream(Chunk(1, 2), Chunk(3, 4, -1, 5))
+      assertM(in.aggregate(sink).runCollect.run)(
```
regiskuckaertz: Isn’t this test supposed to succeed?

quelgar: Good catch, thanks.

quelgar:
> Isn’t this similar to the `split*` sinks?

Yes, it's very similar to `splitLinesChunk`, but works with any chunk element type.

jdegoes:
@quelgar Looks like it's nearly ready to merge, so once you make the changes @iravid recommended, let me know and we'll get this merged in. Thanks for your work on this!

quelgar:
Ok, rename is done.

quelgar:
Got some JS failures, which seem unrelated. I don't have permissions to rerun.

adamgfraser:
Definitely unrelated. Rerunning.

iravid:
Great work @quelgar. Thank you for contributing this!

PR #2930, opened 2020-02-18
ZIO Core: Acquire Layers in Parallel

adamgfraser:
Upgrades layers so that resources that do not depend on each other will automatically be acquired in parallel.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -428,9 +428,7 @@ object ZLayer {
      */
     def make: UIO[MemoMap] =
       RefM
-        .make[Map[ZLayer[Nothing, Any, Has[_]], (Promise[_, _], Exit[Any, Any] => ZIO[Any, Nothing, Any], Ref[Int])]](
-          Map.empty
-        )
+        .make[Map[ZLayer[Nothing, Any, Has[_]], Reservation[Any, Any, Any]]](Map.empty)
         .map { ref =>
           new MemoMap { self =>
             final def getOrElseMemoize[E, A, B <: Has[_]](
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -413,57 +410,76 @@ object ZLayer {
...
      */
     def make: UIO[MemoMap] =
-      Ref.make[Map[ZLayer[Nothing, Any, Has[_]], Any]](Map.empty).map { ref =>
-        new MemoMap {
-          def get[E, A, B <: Has[_]](layer: ZLayer[A, E, B]): UIO[Option[B]] =
-            ref.get.map(_.get(layer).asInstanceOf[Option[B]])
-          def memoize[E, A, B <: Has[_]](layer: ZLayer[A, E, B], b: B): UIO[Unit] =
-            ref.update(_ + (layer -> b))
+      RefM
+        .make[Map[ZLayer[Nothing, Any, Has[_]], (Promise[_, _], Exit[Any, Any] => ZIO[Any, Nothing, Any], Ref[Int])]](
```
iravid: A case class could be nicer for the value type.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -413,57 +410,76 @@ object ZLayer {
...
+                      }
+                    case None =>
+                      for {
+                        a           <- ZIO.environment[A]
+                        promise     <- Promise.make[E, B]
+                        observers   <- Ref.make(0)
+                        reservation <- layer.scope.flatMap(_.apply(self)).reserve
+                      } yield (
+                        Reservation(
+                          observers.update(_ + 1) *>
```
iravid: I'm wondering if you should maybe use `uninterruptibleMask` to make sure the `observers` ref is always updated?

core/shared/src/main/scala/zio/ZLayer.scala
adamgfraser: Good idea!

core/shared/src/main/scala/zio/ZLayer.scala
adamgfraser: I think we are okay because `observers` is always updated in either the `acquire` or `release` action of a `Managed`.

PR #2931, opened 2020-02-19
ZIO Stream: Lift A Runnable Schedule Into A Stream

adamgfraser:
Sometimes we want to work with schedules independent of the idea of repeating or retrying. If a `Schedule` has any for its input type we can think of it as "runnable" and generate a stream that repeatedly extracts and updates the state of the schedules. This could be useful for example if we want an exponential schedule but want to retain the durations output by the schedule.


iravid:
I like this. Immediate usecase is `stream <& ZStream.fromSchedule(sched)` for metering a stream.

Can we replace `ZStream#scheduleWith` with the above implementation? The current implementation is probably more efficient but that's probably negligible.

iravid:
Ah, sorry, we can't replace `ZStream#scheduleWith` as that one uses the stream's elements as schedule inputs, so it's a bit more general.

PR #2932, opened 2020-02-19
Increase the number of generators to 6 in check* methods

erdeszt:


test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -226,37 +226,89 @@ package object test extends CompileVariants {
...
+    test: (A, B, C, D, E, F) => TestResult
+  ): ZIO[R, Nothing, TestResult] =
+    check(rv1 <*> rv2 <*> rv3 <*> rv4 <*> rv5 <*> rv6)(reassociate(test))
+
   /**
    * Checks the effectual test passes for "sufficient" numbers of samples from
    * the given random variable.
    */
-  def checkM[R, R1 <: R, E, A](rv: Gen[R, A])(test: A => ZIO[R1, E, TestResult]): ZIO[R1, E, TestResult] =
+  def checkM[R, R1 <: R, Env, A](rv: Gen[R, A])(test: A => ZIO[R1, Env, TestResult]): ZIO[R1, Env, TestResult] =
```
adamgfraser: This is confusing because `E` is the error type, not the environment type. I would suggest just leaving `E` as is, since it is very well known that it stands for the error type, and skipping to `F` for generic type parameters. Hopefully we don't have to add enough parameters to get to `R`. 😃

erdeszt: Sorry I mixed up the error and the environment. I will change it back to `E` and use your suggestions for the extra argument names.

erdeszt: I've changed `Env` back to `E` and skipped to `F` where there was a clash. I've kept `E` where the error is `Nothing`. Let me know if I should change those as well.

erdeszt:
@adamgfraser I don't know how to interpret the failing test. Doesn't seem to be related to my change but I don't have enough experience with zio to be able to tell for sure.

adamgfraser:
@erdeszt It definitely isn't you. It is an occasional issue we have been having with ScalaJS in CI that we haven't been able to pinpoint the cause of yet. I just triggered the tests to run again.

test/shared/src/main/scala/zio/test/package.scala
adamgfraser: Yes, I think it is slightly more consistent with for example the treatment of `mapN` in `UIO.scala` to skip `E` for a generic parameter since it typically has that other meaning though admittedly it is a very minor point. With that change should be good to go.

erdeszt: Done. I've kept `E` in `reassociate` because it was `E` already before my PR and it also doesn't use the `ZIO` type.

PR #2933, opened 2020-02-19
ZIO Core: Make ZManaged#zipWithPar Robust To Interruption

adamgfraser:


adamgfraser:
@iravid I still need to do some more work on ZLayer but would be good to nail down `zipWithPar`. It is stable now but the one thing I don't like is we are executing the finalizers sequentially instead of in parallel. Do you understand why that is creating an issue?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1064,27 +1064,40 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
           acquire = {
             val left = ZIO.uninterruptibleMask { restore =>
-              reserve
-                .flatMap(res => finalizers.update(fs => res.release :: fs).as(res))
-                .flatMap(res => restore(res.acquire))
+              self.reserve.flatMap { reservation =>
+                ref.updateSomeAndGet {
+                  case Right(finalizers) => Right(reservation.release :: finalizers)
+                }.flatMap {
+                  case Left(exit) => restore(reservation.acquire).ensuring(reservation.release(exit))
```
iravid: Very tricky indeed. Nice catch. Happy to see it's a relatively esoteric scenario.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1064,27 +1064,40 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-            for {
-              fs    <- finalizers.get
-              exits <- ZIO.foreachPar(fs)(_(exitU).run)
-              _     <- ZIO.doneNow(Exit.collectAllPar(exits).getOrElse(Exit.unit))
-            } yield ()
+          release = exit =>
+            ref.getAndSet(Left(exit)).flatMap {
+              case Right(finalizers) =>
+                for {
+                  exits <- ZIO.foreach(finalizers)(_(exit).run)
```
iravid: I don't know why `foreachPar` is causing trouble, but I am assuming it is the same issue we had with `race` in finalizers after interruption. `foreachPar_` (which `foreachPar` is based on) has some interruptible sections.

PR #2936, opened 2020-02-20
ZIO Core: Align ZLayer Constructors

### LINKED ISSUE #2929: ZLayer succeed, fromEffect, and fromManaged inconsistent?
```ZLayer.fromEffect``` and ```ZLayer.fromManaged``` accept a IO/Managed where ```A <: Has[_]```
whereas ```ZLayer.succeed``` just accepts ```A``` (and the implementation wraps this in ```Has(a)```). 

Client code to create a Layer for effectful/managed resources requires wrapping in ```Has(_)``` but this is not the case for values. Is this inconsistent? (and should it be addressed?)

```scala
def fromEffect[E, A <: Has[_]](zio: IO[E, A]): ZLayer.NoDeps[E, A]

def fromManaged[E, A <: Has[_]](m: Managed[E, A]): ZLayer.NoDeps[E, A]

def succeed[A: Tagged](a: => A): ZLayer.NoDeps[Nothing, Has[A]]```

adamgfraser:
First step at resolves #2929. Changes `succeed` to lift an `A <: Has[_]` into a layer.

As is this is a victory for consistency but not concision. It would be nice to have versions of `succeed`, `fromEffect`, and `fromManaged` that lift a `A` with implicit evidence `Tagged[A]` into a layer for the simple case where we have an existing single service and just want to construct layer from it. `fromService`, `fromServiceM`, and `fromServiceManaged` seem most logical for that but overlap with existing `fromServices` methods.

Does anyone else have ideas for naming, or renaming the existing methods?

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3165,6 +3165,6 @@ object ZIOSpec extends ZIOBaseSpec {
 
   object Logging {
     trait Service
-    val live: ZLayer[Any, Nothing, Logging] = ZLayer.succeed(new Logging.Service {})
+    val live: ZLayer[Any, Nothing, Logging] = ZLayer.succeed(Has(new Logging.Service {}))
   }
 }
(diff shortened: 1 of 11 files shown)


adamgfraser:
Superseded by #2937.

PR #2937, opened 2020-02-20
ZIO Core: Improve Ergonomics of Layers

adamgfraser:
Adds `toLayer` methods on `ZIO` and `ZManaged`. Adds `fromAcquireRelease` constructor on `ZLayer`. Deprecates `provideManaged` and `provideSomeManaged`.

Open items:
- Should we deprecate more `provide` variants?
- Should we rename `provideSome` to `contramap`?
- Other methods that would be helpful?
- I noticed some of the `ZLayer` constructors require an effect that doesn't have an environmental dependency. I think we could relax that unless we think think it adds clarity.
- Should we rename `fromEnvironment` to `fromFunction`? It basically lifts a function into a Layer "arrow". I found that a bit more intuitive.
- When should constructors take `Tagged` instead of `Has`. `fromAcquireRelease` for example seems like it would be significantly more ergonomic for the common case if it took `Tagged` but there is a risk of inconsistency and proliferation of methods here.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -189,8 +189,8 @@ object ZLayerSpec extends ZIOBaseSpec {
       testM("layers can be acquired in parallel") {
         for {
           promise <- Promise.make[Nothing, Unit]
-          layer1  = ZLayer.fromManaged(Managed.make(ZIO.never)(_ => ZIO.unit))
-          layer2  = ZLayer.fromManaged(Managed.make(promise.succeed(()).map(Has(_)))(_ => ZIO.unit))
+          layer1  = ZLayer.fromManagedMany(Managed.make(ZIO.never)(_ => ZIO.unit))
+          layer2  = ZLayer.fromManagedMany(Managed.make(promise.succeed(()).map(Has(_)))(_ => ZIO.unit))
           env     = (layer1 ++ layer2).build
           _       <- env.use_(ZIO.unit).fork
           _       <- promise.await
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -61,14 +61,14 @@ object ZLayerSpec extends ZIOBaseSpec {
   def spec =
     suite("ZLayerSpec")(
       testM("Size of >>> (1)") {
-        val layer = ZLayer.succeed(1) >>> ZLayer.fromService((i: Int) => Has(i.toString))
+        val layer = ZLayer.succeed(1) >>> ZLayer.fromServiceMany((i: Int) => Has(i.toString))
 
         testSize(layer, 1)
       },
       testM("Size of >>> (2)") {
         val layer = ZLayer.succeed(1) >>>
-          (ZLayer.fromService((i: Int) => Has(i.toString)) ++
-            ZLayer.fromService((i: Int) => Has(i % 2 == 0)))
+          (ZLayer.fromServiceMany((i: Int) => Has(i.toString)) ++
+            ZLayer.fromServiceMany((i: Int) => Has(i % 2 == 0)))
 
         testSize(layer, 2)
       },
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -61,14 +61,14 @@ object ZLayerSpec extends ZIOBaseSpec {
   def spec =
     suite("ZLayerSpec")(
       testM("Size of >>> (1)") {
-        val layer = ZLayer.succeed(1) >>> ZLayer.fromServiceMany((i: Int) => Has(i.toString))
+        val layer = ZLayer.succeed(1) >>> ZLayer.fromService((i: Int) => i.toString)
 
         testSize(layer, 1)
       },
       testM("Size of >>> (2)") {
         val layer = ZLayer.succeed(1) >>>
-          (ZLayer.fromServiceMany((i: Int) => Has(i.toString)) ++
-            ZLayer.fromServiceMany((i: Int) => Has(i % 2 == 0)))
+          (ZLayer.fromService((i: Int) => i.toString) ++
+            ZLayer.fromService((i: Int) => i % 2 == 0))
 
         testSize(layer, 2)
       },
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2100,6 +2100,7 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * An effectual version of `provide`, useful when the act of provision
    * requires an effect.
    */
+  @deprecated("use provideLayer", "1.0.0")
   final def provideM[E1 >: E](r: ZIO[Any, E1, R])(implicit ev: NeedsEnv[R]): Stream[E1, A] =
     provideSomeM(r)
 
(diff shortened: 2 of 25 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2155,12 +2161,18 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * leaving the remainder `R0`.
    */
   final def provideSomeM[R0, E1 >: E](env: ZIO[R0, E1, R])(implicit ev: NeedsEnv[R]): ZStream[R0, E1, A] =
-    provideSomeManaged(env.toManaged_)
+    ZStream.managed {
```
jdegoes: I think we can deprecate `provideSomeM` as it's the same as `provideLayer(zio.toLayer)`, right?

adamgfraser: Yes. The only issue is what if the `R` type is not a subtype of `Has` so you can't call `toLayer`. But it could always be replaced with `env.flatMap(zio.provideSome)`.

jdegoes:
> Should we deprecate more provide variants?

Yes. I think we can deprecate `provideM`, `provideSomeM`, etc., basically, anything that is a simple `toLayer` away.

> Should we rename provideSome to contramap?

Probably, just to make it esoteric so fewer people try to use it. Although we could think of generally better names for `contramap`, if we are using `contramap` elsewhere (and I think we are), we could just go with that.

> I noticed some of the ZLayer constructors require an effect that doesn't have an environmental dependency. I think we could relax that unless we think think it adds clarity.

As long as it infers well, it makes sense to relax the requirement.

> Should we rename fromEnvironment to fromFunction? It basically lifts a function into a Layer "arrow". I found that a bit more intuitive.

Yes, good idea.

> When should constructors take Tagged instead of Has. fromAcquireRelease for example seems like it would be significantly more ergonomic for the common case if it took Tagged but there is a risk of inconsistency and proliferation of methods here.

We could use `Many` postfix to indicate you can produce more than one service, e.g. `fromAcquireReleaseMany` being the `Has` variant. Could apply to other methods too.

The common case is more important than fringe cases, as long as we have a general-purpose one or two, we don't need to provide generalized versions of everything IMO.

adamgfraser:
Sounds good. Let me take another turn on this.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Only the A parameter needs to be, not R, right?

adamgfraser: The `A` type of the `ZLayer` needs to be which would be the `R` type of the `ZIO` that the layer is being provided to.

adamgfraser:
One issue with `contramap` is that some data type that we may want to provide an environment to (e.g. `Schedule`, `ZSink`) already have `contrmap` methods defined for other parameters. We could change it to something like `contramapEnvironment` or just leave as `provideSome` for now.

adamgfraser:
I deprecated `provideM` and `provideSomeM`.

I also added `Many` variants of several of the constructors. I think that is a nice solution. One slight inconsistency we have is that when there are normal and `Many` variants the normal variant returns a value that is not in a `Has`, whereas when there is only one variant the normal variant returns the value in the `Has`. I'm wondering if we should just go through and give all the `ZLayer` constructors the `Many` treatment.

I also added normal and `Many` variants for `toLayer` and added a `asService` method on ZIO and Managed that just maps the success value to a Has if you need that.

jdegoes:
> I'm wondering if we should just go through and give all the ZLayer constructors the Many treatment.

This sounds preferable to inconsistency. Agreed?

> We could change it to something like contramapEnvironment or just leave as provideSome for now.

OK, why don't we leave it to `provideSome` since that's become synonymous with environment. 👍

adamgfraser:
Agreed on both counts. Will make that change now.

adamgfraser:
I added the `Many` variants. One other thing I was thinking about is right now we don't constrain the input services type to be a subtype of `Has`. This obviously isn't strictly necessary as we are not requiring it today, but it creates the risk that people will accidentally create layers that are not really useful in that they don't compose with any other layers and aren't even of the right type where they could ever compose. I am wondering if we should require that input services are a subtype of `Has` as well.

adamgfraser:
Thinking about that more I don't think it works because `Any` isn't a subtype of `Has`. Not sure if there is another natural way we could express a lack of environmental requirements in that framework.

jdegoes:
> Thinking about that more I don't think it works because Any isn't a subtype of Has. Not sure if there is another natural way we could express a lack of environmental requirements in that framework.

In an earlier model, I made the input be `<: Has[_]`, but it was super painful because of `Any`. I had to have a special `Has.Any` that was just truly awful. The code became much cleaner when I dropped that constraint.

If anything, I think I'd go the other way, drop the output constraint on `<: Has[_]`, and require it only for `++` and `+!+`, with a generic `zipWith` not even having that requirement.

Then you can allow things existing in intermediate forms, where they're not quite ready to be composable layers yet. Or you can imagine other people using type-level techniques to compose `ZLayer` (not that we should support that, of course, that would be on them).

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
```diff
@@ -61,14 +61,14 @@ object ZLayerSpec extends ZIOBaseSpec {
   def spec =
     suite("ZLayerSpec")(
       testM("Size of >>> (1)") {
-        val layer = ZLayer.succeed(1) >>> ZLayer.fromService((i: Int) => Has(i.toString))
+        val layer = ZLayer.succeed(1) >>> ZLayer.fromServiceMany((i: Int) => Has(i.toString))
```
jdegoes: This one and the other ones can be simplified to `fromService` since they don't produce many.

test-tests/shared/src/test/scala/zio/test/SpecSpec.scala
```diff
@@ -17,34 +17,31 @@ object SpecSpec extends ZIOBaseSpec {
...
         for {
-          ref <- Ref.make(true)
-          _ <- execute {
-                spec.provideManagedShared {
-                  ZManaged.make(ref.set(false).as(ref))(_ => ZIO.unit)
-                } @@ ifEnvSet("foo")
-              }
+          ref    <- Ref.make(true)
+          layer  = ZLayer.fromEffect(ref.set(false).as(ref))
+          _      <- execute(spec.provideCustomLayerShared(layer) @@ ifEnvSet("foo"))
```
jdegoes: Nice cleanup!

test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -342,41 +342,63 @@ final case class Spec[-R, +E, +T](caseValue: SpecCase[R, E, T, Spec[R, E, T]]) {
...
    */
+  @deprecated("use provideLayer", "1.0.0")
   final def provideManaged[E1 >: E](managed: Managed[E1, R])(implicit ev: NeedsEnv[R]): Spec[Any, E1, T] =
     provideSomeManaged(managed)
 
   /**
    * Uses the specified `Managed` once to provide all tests in this spec with
    * a shared version of their required environment.
    */
+  @deprecated("use provideLayerShared", "1.0.0")
```
jdegoes: It's really nice to see all these deprecations. The endless variations were confusing.

test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -453,12 +453,12 @@ package object test extends CompileVariants {
 
     def live(size: Int): ZLayer.NoDeps[Nothing, Sized] =
       ZLayer.fromEffect(FiberRef.make(size).map { fiberRef =>
-        Has(new Sized.Service {
+        new Sized.Service {
```
jdegoes: Nice cleanup and emphasizes this was the right default choice!

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
adamgfraser: Good catch! Fixed now.

adamgfraser:
Good point about the `Has` requirement. You can see how it is sometimes a hassle having to convert to `Has` before creating a layer versus even just being able to create it and immediately put the value in a `Has`. I can make that change in a follow up PR.

I think the one thing we lose by going in that direction is if we want to do more static analysis I think we are in a situation where we have the edges of the dependency graph but don't have information on the nodes they are supposed to point to in order to do a lot with them.

test/shared/src/main/scala/zio/test/Spec.scala
adamgfraser: Totally agree! Especially for `Spec`.

jdegoes:
> I think the one thing we lose by going in that direction is if we want to do more static analysis I think we are in a situation where we have the edges of the dependency graph but don't have information on the nodes they are supposed to point to in order to do a lot with them.

I think every time we captured `zipWith` in a graph, we'd capture the combining function as well. That might be enough. Even now, if we stay fixed to `Has`, we'll have to capture the `Tagged` of the right side, when we capture the zip with in a data structure; because we have to later (during `build`) do all the appends. The question is: is it more or less useful to store an append function versus storing a `Tagged`?

PR #2942, opened 2020-02-20
ZIO Core: Move FinalizerRef to ZManaged

adamgfraser:
@iravid I confirmed that `scope` has the same issue as we have been seeing. I moved `FinalizerRef` to `ZManaged`, implemented `scope` and `zipWithPar` in terms of that, and moved several other combinators to using that.

There are a couple of combinators like `switchable` and `ZStream#catchAllCause` that use more complex patterns involving removing finalizers that can't easily be expressed in terms of this. I renamed the existing `finalizerRef` method to `finalizerRefInternal` and made it `private[zio]`. I think it is tricky to use correctly given that the finalizer may be run before child fibers have registered their finalizers but obviously open to feedback on that.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -630,14 +630,14 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
    * Preallocates the managed resource inside an outer managed, resulting in a ZManaged that reserves and acquires immediately and cannot fail.
    */
   def preallocateManaged: ZManaged[R, E, Managed[Nothing, A]] =
-    ZManaged.finalizerRef[R](_ => ZIO.unit).mapM { finalizer =>
+    ZManaged.finalizerRef[R](_ => ZIO.unit).mapM { finalizers =>
       ZIO.uninterruptibleMask { restore =>
         for {
           env      <- ZIO.environment[R]
           res      <- reserve
-          _        <- finalizer.add(res.release)
+          _        <- finalizers.add(res.release)
           resource <- restore(res.acquire)
-        } yield ZManaged.make(ZIO.succeedNow(resource))(_ => finalizer.run(Exit.succeed(resource)).provide(env))
+        } yield ZManaged.make(ZIO.succeedNow(resource))(_ => finalizers.runAll(Exit.succeed(resource)).provide(env))
       }
     }
 
@@ -1111,7 +1111,7 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
             }
             left.zipWithPar(right)(f0)
           },
-          release = finalizers.run
+          release = finalizers.runAll
         )
       }
     }
@@ -1149,10 +1149,16 @@ object ZManaged {
     def add(finalizer: Exit[Any, Any] => URIO[R, Any]): URIO[R, Unit]
 
     /**
+     * Runs the specified finalizer and removes it from the finalizers
+     * associated with this scope.
+     */
+    def run(finalizer: Exit[Any, Any] => URIO[R, Any])(exit: Exit[Any, Any]): URIO[R, Unit]
+
+    /**
      * Runs the finalizers associated with this scope. After this any
      * finalizers added to this scope will be run immediately.
      */
-    def run(exit: Exit[Any, Any]): URIO[R, Unit]
+    def runAll(exit: Exit[Any, Any]): URIO[R, Unit]
   }
 
   object FinalizerRef {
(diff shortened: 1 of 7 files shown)


adamgfraser:
@iravid I realized there was another issue with `zipWithPar`. With the `FinalizerRef` state machine we are now making sure that all finalizers get executed at some point. But since the wind down of the parent and child fibers are occurring at the same time we have another issue that the finalizer could be executed by the parent fiber and simultaneously or after that the child fiber could execute the acquire action.

I added further coordination to address this. It may be possible to simplify but this at least seems correct and the interruption tests for `zipWithPar` and `ZLayer` are now passing for several hundred thousand iterations for me locally.

adamgfraser:
@iravid Could you take a look at this when you get the chance? We're seeing some flakiness on the ZLayer resource safety tests that I think this addresses.

iravid:
Yes, it's on my list. Sorry.

adamgfraser:
No worries!

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1811,20 +1889,13 @@ object ZManaged {
                     res      <- managed.reserve
                     resource <- restore(res.acquire).onError(err => res.release(Exit.Failure(err)))
                     release  = res.release.andThen(_.provide(env))
-                    _        <- finalizers.update(_ + release)
-                    done <- Ref
-                             .make(release(Exit.Success(resource)).ensuring(finalizers.update(_ - release)))
-                             .map(_.modify((_, ZIO.unit)).flatten)
+                    _        <- finalizer.add(release)
+                    done     = finalizer.run(Exit.succeed(resource))
```
iravid: I think this is slightly wrong; the previous formulation would only run the release for the current resource. This formulation runs *all* finalizers.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1120,6 +1135,52 @@ object ZManaged {
...
+    def run(exit: Exit[Any, Any]): URIO[R, Unit]
+  }
+
+  object FinalizerRef {
+
+    /**
+     * Constructs a new `FinalizerRef` with the specified initial finalizer.
+     */
+    def make[R](initial: Exit[Any, Any] => URIO[R, Any]): UIO[FinalizerRef[R]] =
+      Ref.make[Either[Exit[Any, Any], Exit[Any, Any] => URIO[R, Any]]](Right(initial)).map { ref =>
```
iravid: Would it be possible for the `Right` case here to be a `Set[Exit[Any, Any] => URIO[R, Any]]`? We can then have, in addition to `run`, `runPar` and combinators for manipulating the finalizers (`remove`, `replace`, etc). At that point, I don't think we'd need `finalizerRefInternal` anymore.

Needless to say, this is really nice!

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1062,45 +1060,62 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
    * in parallel, combining their results with the specified `f` function. If
    * either side fails, then the other side will be interrupted.
    */
-  def zipWithPar[R1 <: R, E1 >: E, A1, A2](that: ZManaged[R1, E1, A1])(f0: (A, A1) => A2): ZManaged[R1, E1, A2] =
+  def zipWithPar[R1 <: R, E1 >: E, A1, A2](that: ZManaged[R1, E1, A1])(f0: (A, A1) => A2): ZManaged[R1, E1, A2] = {
+
+    sealed trait State
+    case object Running   extends State
```
iravid: Can't get the state machine more explicit than this!

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1062,45 +1060,62 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-                  case _          => restore(reservation.acquire)
-                }
-              }
+            val left = ZIO.uninterruptibleMask {
+              restore =>
+                for {
+                  ref   <- Ref.make[State](Running)
+                  latch <- Promise.make[Nothing, Unit]
+                  res   <- self.reservation
+                  acquire = ZIO.ifM(ref.getAndSet(Acquiring).map(_ == Running))(
```
iravid: Great catches all around here. The explicit state machine makes the mechanism very clear 👍

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: Thanks!

I think we could definitely make this a set. I'm a little worried about safety of some of the combinators though. `runPar` should be fine though it seems like at least right now it leads to finalizers not running in certain situations. I think `remove` or `replace` may not be safe because even if you remove or replace a finalizer it may not have gotten attached yet so you may not actually be able to remove the finalizer. Perhaps we could maintain a separate set of "deactivated" finalizers and change `add` to check the deactivated finalizers.

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: Good point. Fixed now!

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: I added a combinator to run a specified finalizer, safely removing it from the `FinalizerRef` and used this to fix the issue with `scope`. Still thinking about how to safely add more combinators to `FinalizerRef` to implement the remaining methods that use `finalizerRefInternal` (`ZManaged.switchable`, `ZStream.catchAllCause`, and `Stream.flatMap`) but may make sense to merge this if we think it is correct and follow up on that.

adamgfraser:
All set.

core/shared/src/main/scala/zio/ZManaged.scala
iravid: Yes, makes sense. Will review again today and merge.

PR #2946, opened 2020-02-21
Only count marked fibers in test

### LINKED ISSUE #2842: Flaky Supervise Fibers Test
```
[info] - ZIOSpec
[info]   - RTS concurrency correctness
[info]     - supervise fibers
[info]       3 did not satisfy equalTo(2)
```

https://circleci.com/gh/zio/zio/55390

mlangc:
By using a marker for the test fibers, we make sure that we don't sometimes count fibers that are left over from other tests.
Fixes #2842

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -10,7 +10,7 @@ import zio.duration._
 import zio.internal.Platform
 import zio.random.Random
 import zio.test.Assertion._
-import zio.test.TestAspect.{ flaky, jvm, nonFlaky, scala2Only }
+import zio.test.TestAspect.{ flaky, forked, jvm, nonFlaky, scala2Only }
 import zio.test._
 import zio.test.environment.{ Live, TestClock }
 
@@ -2270,7 +2270,7 @@ object ZIOSpec extends ZIOBaseSpec {
           } yield value
 
         assertM(Live.live(io))(equalTo(2))
-      } @@ flaky,
+      } @@ forked,
       testM("race of fail with success") {
         val io = IO.failNow(42).race(IO.succeedNow(24)).either
         assertM(io)(isRight(equalTo(24)))


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -2247,20 +2247,30 @@ object ZIOSpec extends ZIOBaseSpec {
...
+                  fs.foldLeft(waitForFiberMarkersToBeSet) { (acc, f) =>
+                    acc *> ZIO.whenM(isMarked(f))(f.interrupt.ignore *> counter.update(_ + 1))
+                  }
                 }
             value <- counter.get
           } yield value
 
         assertM(Live.live(io))(equalTo(2))
-      } @@ flaky,
+      },
```
jdegoes: `@@ forked` can do the same thing, I think. I fixed this in my PR too, in a slightly different way.

mlangc: If this issue is already fixed your PR, should we drop this one then?

jdegoes: No, but I would simplify to `@@ forked` if that does the job. Nice work!

adamgfraser: Or if `forked` doesn't work we should think about why it doesn't work since this is the kind of problem it is supposed to solve for us.

mlangc: Hmm, I can't find `TestAspect.forked` in current master. Is it somewhere else, or maybe part of another PR?

adamgfraser: That's so funny. I think we both talked about adding it and ended up not when we implemented the test aspects to restore the test environment. Just submitted a PR with it at #2954.

adamgfraser: Well it is on master now. 😃

mlangc: Thanks... `@@ forked` seem to do the trick indeed :-)

PR #2950, opened 2020-02-21
ZIO Core: Support Parameterized Services

adamgfraser:
Lifts the restriction on parameterized services. Note that parameterized services are not currently supported on Dotty.


neko-kai:
@adamgfraser Note, I noticed the `Tagged`  case class still contains some vestigal remains of old logic, e.g. `.toString` in `equals` & `hashCode` – this may actually cause false negatives, because there are cases when, using highly polymorphic tags as in
```scala
def getHKTService[F[_, _]: TagKK, E: Tag] = access(_.get[Service[F[E, ?]]])
getHKTService[IO, Throwable]
```
Can cause the output `Tag` to have different numerical indexes for lambda parameters than when the same tag is created monomorphically:
```scala
access(_.get[Service[IO[Throwable, ?]]])
```
– they will still be equal & have the same hashcode, but different `toString`'s.

We may also think about re-exporting polymorphic tags (`TagK`, `TagKK` etc.) in `zio` package object to allow people to write polymorphic code with higher kinds.

adamgfraser:
@neko-kai Great point! It seems like a quick fix would just be to delete the `hashCode` and ` toString` implementations in `Tagged` so it is a pure wrapper, and then we could potentially eliminate along with a unification of the platform specific files. Does that make sense to you?

neko-kai:
@adamgfraser Yep, for now! I think at some point we may start publishing shims (or maybe a full impl right away) for Tag for dotty, then we could eliminate all cross-files for this from ZIO

adamgfraser:
@neko-kai I tried removing the `equals` and `hashCode` methods on `Tagged` and found that the test "Siblings can be updated independently` in `HasSpec.scala` was failing. Could you take a look?

neko-kai:
@adamgfraser Tagged should contain `TagType`, not `TaggedType[A]` - that'll work. Equals is undefined for `Tag[A]` right now, only for `LightTypeTag`

adamgfraser:
@neko-kai Thank you!

jdegoes:
@adamgfraser This is good work! I'm a bit reluctant to recommend parametric services knowing they will result in runtime exceptions on Dotty. If anything, this PR means we will need to accelerate efforts to support Dotty. /cc @neko-kai

adamgfraser:
@jdegoes Should we add back the warnings? I think it is going to be hard to hold back the demand for parameterized services. The ZIO Kafka team already looking to use them.

pshirshov:
@jdegoes : you may help us to port our macro to dotty. We've checked, there are enough APIs for the port, so it's a question of some fair amount of mostly mechanical work.

PR #2951, opened 2020-02-21
Fix flaky test

### LINKED ISSUE #2096: Flaky ZIO ForkAll Test
```scala
[info] ZIO Test
[info] - ZIO
[info]   - forkAll
[info]     - propagates defects
```

https://circleci.com/gh/zio/zio/36839?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

mlangc:
`ZIO.forkAll`, when being executed with `List(ZIO.dieNow(boom))` will
internally execute `nilIO.zipWithPar(boomIO)((as, a) => a :: as)`, where
`nilIO` is a simple ZIO value that results in `Nil`, and `boomIO` a
simple ZIO value that results in `ZIO.dieNow(boom)`. In rare cases,
`boomIO` completes before `nilIO` and then interrupts `nilIO`. The
result of the test is then not just a `Die`, but also an interruption.

Fixes #2096

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2604,11 +2604,22 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * composite fiber that produces a list of their results, in order.
    */
   def forkAll[R, E, A](as: Iterable[ZIO[R, E, A]]): URIO[R, Fiber.Synthetic[E, List[A]]] =
-    as.foldRight[URIO[R, Fiber.Synthetic[E, List[A]]]](succeedNow(Fiber.succeed(Nil))) { (aIO, asFiberIO) =>
-      asFiberIO.zipWith(aIO.fork) {
-        case (asFiber, aFiber) =>
-          asFiber.zipWith(aFiber)((as, a) => a :: as)
-      }
+    as.toList match {
+      case Nil => succeedNow(Fiber.succeed(Nil))
+
+      case a :: Nil => a.fork.map(_.map(List(_)))
+
+      case as =>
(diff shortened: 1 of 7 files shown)


adamgfraser:
@mlangc Excellent analysis of the situation.

It seems counterintuitive that you could get an interruption exception by forking a single effect. I wonder if we need to change the logic in `forkAll` to address this. If the list of effects is empty we can just immediately return an empty list. If it is non-empty we can fork and combine those fibers.

What do you think?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2604,11 +2604,22 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * composite fiber that produces a list of their results, in order.
    */
   def forkAll[R, E, A](as: Iterable[ZIO[R, E, A]]): URIO[R, Fiber.Synthetic[E, List[A]]] =
-    as.foldRight[URIO[R, Fiber.Synthetic[E, List[A]]]](succeedNow(Fiber.succeed(Nil))) { (aIO, asFiberIO) =>
-      asFiberIO.zipWith(aIO.fork) {
-        case (asFiber, aFiber) =>
-          asFiber.zipWith(aFiber)((as, a) => a :: as)
-      }
+    as.toList match {
```
adamgfraser: Can we use `as.reduceRightOption(???).getOrElse(ZIO.succeedNow(Fiber.succeed(Nil)))` to clean this up?

mlangc: I've just refactored the code as you proposed.

adamgfraser: Looks great!

PR #2958, opened 2020-02-23
Remove AsError and Fallback Methods

### LINKED ISSUE #2900: Rationalize Redundant Methods
I noticed that with some of our recent additions we have a couple of methods that do the same thing:

```scala
  /**
   * Maps the error value of this effect to the specified constant value.
   */
  final def asError[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
    mapError(new ZIO.ConstFn(() => e1))

  /**
   * Executes this effect and returns its value, if it succeeds, but
   * otherwise fails with the specified error.
   */
  final def orElseFail[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
    orElse(ZIO.failNow(e1))
```

```scala
  /**
   * Executes this effect and returns its value, if it succeeds, but otherwise
   * returns the specified value.
   */
  final def fallback[A1 >: A](a: => A1)(implicit ev: CanFail[E]): ZIO[R, Nothing, A1] =
    fold(_ => a, identity)

  /**
   * Executes this effect and returns its value, if it succeeds, but
   * otherwise suceeds with the specified value.
   */
  final def orElseSucceed[A1 >: A](a1: => A1)(implicit ev: CanFail[E]): URIO[R, A1] =
    orElse(ZIO.succeedNow(a1))
```

Does anyone have a preference for which versions we should keep?

adamgfraser:
Resolves #2900.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -203,12 +203,6 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def as[B](b: => B): ZIO[R, E, B] = self.flatMap(new ZIO.ConstZIOFn(() => b))
 
   /**
-   * Maps the error value of this effect to the specified constant value.
-   */
-  final def asError[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
-    mapError(new ZIO.ConstFn(() => e1))
-
-  /**
    * Maps the success value of this effect to a service.
    */
   final def asService[A1 >: A](implicit tagged: Tagged[A1]): ZIO[R, E, Has[A1]] =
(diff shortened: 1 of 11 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -202,12 +202,6 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    */
   final def as[B](b: => B): ZIO[R, E, B] = self.flatMap(new ZIO.ConstZIOFn(() => b))
 
-  /**
-   * Maps the error value of this effect to the specified constant value.
-   */
-  final def asError[E1](e1: => E1)(implicit ev: CanFail[E]): ZIO[R, E1, A] =
```
jdegoes: Should we deprecate for RC18 and then axe it in 1.0?

adamgfraser: Good idea! Fixed now.

PR #2959, opened 2020-02-23
ZIO Core: Fix Exponential Schedule

adamgfraser:
The first delay in `Schedule.exponential(10.milliseconds)` should be 10 milliseconds, not 20 milliseconds.


jdegoes:
Good catch!

PR #2963, opened 2020-02-23
Update Schedule Documentation

adamgfraser:
`schedule && Schedule.elapsed.whileOut(_ < time)` is not a great way to place an overall time limit on a schedule with the new encoding because `elapsed` will return immediately with the current elapsed time versus the elapsed time after the other schedule that has done sleeping, so you generally end up with an additional repetition of the schedule versus what you want. Updates the documentation to not use this example and instead put an overall timeout on the effect.

### DIFF
--- a/docs/datatypes/schedule.md
+++ b/docs/datatypes/schedule.md
@@ -100,8 +100,7 @@ val expCapped = Schedule.exponential(100.milliseconds) || Schedule.spaced(1.seco
 Stops retrying after a specified amount of time has elapsed:
 
 ```scala mdoc:silent
-val action = clock.nanoTime.flatMap(n => console.putStrLn(n.toString)) *> ZIO.fail("fail")
-action.retry(expCapped).timeout(30.seconds)
+val expMaxElapsed = (Schedule.exponential(10.milliseconds) >>> Schedule.elapsed).whileOutput(_ < 30.seconds)
 ```
 
 Retry only when a specific exception occurs:
--- a/docs/datatypes/schedule.md
+++ b/docs/datatypes/schedule.md
@@ -100,7 +100,8 @@ val expCapped = Schedule.exponential(100.milliseconds) || Schedule.spaced(1.seco
 Stops retrying after a specified amount of time has elapsed:
 
 ```scala mdoc:silent
-val expMaxElapsed = Schedule.exponential(10.milliseconds) && Schedule.elapsed.whileOutput(_ < 30.seconds)
+val action = clock.nanoTime.flatMap(n => console.putStrLn(n.toString)) *> ZIO.fail("fail")
+action.retry(expCapped).timeout(30.seconds)
 ```
 
 Retry only when a specific exception occurs:


jdegoes:
I'd suggest what we do is delete `Schedule.elapsed` and add it as a method on `Schedule`, e.g. `schedule.elapsed.whileOutput(...)`. Or maybe something else.

Another alternative is having a sequential version of `&&` (`**`???) which would allow this to function as expected in the older encoding.

jdegoes:
I am not sure I would recommend `timeout`, because then you lose composability: you can't do anything except "timeout at the end of the world".

adamgfraser:
I agree we could use an additional operator here. I don't think sequential composition does quite what we need here. If we do that then we execute `Schedule.spaced` once and delay 10 seconds, then we evaluate `Schedule.delayed.whileOutput(...)` and get a delay of 10 seconds so decide not to continue, but then when we try to execute the composed schedule the next time we do it sequentially so we don't get the fail fast behavior but instead need the next 10 second delay before we execute `delayed` again and terminate the schedule.

Another idea I had related is if we could separate the concept of racing from the concept of proceeding if either schedule says continue. Then we could have an operator that says race the two schedules but only continue when both say continue. That would potentially allow us to implement something like the `timeout` operation in terms of schedules.

adamgfraser:
@jdegoes  What about something like this?

```scala
final def **[R1 <: R, A1 <: A, C](that: Schedule[R1, A1, C]): Schedule[R1, A1, (B, C)] =
  new Schedule[R1, A1, (B, C)] {
    type Loser = Option[Either[Fiber[Unit, self.State], Fiber[Unit, that.State]]]
    type State = (self.State, that.State, Loser)

    def race(
      left: ZIO[R, Unit, self.State],
      right: ZIO[R1, Unit, that.State],
      s1: self.State,
      s2: that.State
    ): ZIO[R1, Unit, State] =
      left.raceWith(right)(
        (exit, fiber) =>
          exit.fold(
            c => fiber.interrupt *> ZIO.haltNow(c),
            s1 => (ZIO.succeedNow((s1, s2, Some(Right(fiber)))))
          ),
        (exit, fiber) =>
          exit.fold(
            c => fiber.interrupt *> ZIO.haltNow(c),
            s2 => (ZIO.succeedNow((s1, s2, Some(Left(fiber)))))
          )
      )

    val initial = self.initial.zipWith(that.initial)((_, _, None))
    val extract = (a, s) => (self.extract(a, s._1), that.extract(a, s._2))
    val update = { case (a, (s1, s2, loser)) =>
      loser match {
        case None               => race(self.update(a, s1), that.update(a, s2), s1, s2)
        case Some(Left(fiber))  => race(fiber.join, that.update(a, s2), s1, s2)
        case Some(Right(fiber)) => race(self.update(a, s1), fiber.join, s1, s2)
      }
    }
  }
```

Then you could get a schedule that would recur once at 3 seconds and then immediately terminate like so:

```scala
Schedule.spaced(10.seconds) && (Schedule.spaced(3.seconds) && Schedule.recurs(1))
```

mschuwalow:
Wouldn't ```(schedule >>> Schedule.elapsed).whileOut(_ < time)``` behave the way we want?

jdegoes:
@adamgfraser 

> Another idea I had related is if we could separate the concept of racing from the concept of proceeding if either schedule says continue. Then we could have an operator that says race the two schedules but only continue when both say continue.

I like that idea if it would be made to work.

Something like:

```scala
def update(a: A, s: S): ZIO[R, Nothing, (B, S)]
def continue(s: S): IO[Unit, Unit]
```

Not exactly sure.

adamgfraser:
@mschuwalow That is a great point. I tried something similar but had the `whileOutput` being called on `elapsed` instead of the composed schedule. I will update accordingly. We could potentially explore further changes discussed above but not sure it makes sense to make the API more complex if we don't need to.

adamgfraser:
@jdegoes I think this is ready to go now.

PR #2964, opened 2020-02-24
Command-line args support for running the standalone specs

hmemcpy:
By default, the spec is a command-line application that can be executed directly. It currently does not perform any filtering and runs the entire spec.
This PR considers the arguments that may be passed and applies the filtering, similar to the sbt runner.

```
> java .... MySpec -t myTest -t myTest2 ...
```

This is a required step to improve the UX from tests runners such as IntelliJ (via the zio-intellij plugin), allowing running individual tests.

### DIFF
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
@@ -26,7 +26,7 @@ abstract class BaseTestTask(
 
   protected def run(eventHandler: EventHandler) =
     for {
-      spec    <- specInstance.runner.run(FilteredSpec(specInstance.spec, args))
+      spec    <- specInstance.runSpec(FilteredSpec(specInstance.spec, args))
       summary <- SummaryBuilder.buildSummary(spec)
       _       <- sendSummary.provide(summary)
       events  <- ZTestEvent.from(spec, taskDef.fullyQualifiedName, taskDef.fingerprint)
--- a/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
+++ b/test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
@@ -4,7 +4,7 @@ import sbt.testing.{ EventHandler, Logger, Task, TaskDef }
 
 import zio.UIO
 import zio.clock.Clock
-import zio.test.{ AbstractRunnableSpec, SummaryBuilder, TestArgs, TestLogger }
+import zio.test.{ AbstractRunnableSpec, FilteredSpec, SummaryBuilder, TestArgs, TestLogger }
 import zio.{ Runtime, ZIO, ZLayer }
 
 abstract class BaseTestTask(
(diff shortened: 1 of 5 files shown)


test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
```diff
@@ -26,30 +25,7 @@ abstract class BaseTestTask(
...
-                 spec.runner.run {
-                   spec.spec.filterTags { tag =>
-                     testSearchTerms.exists(term => tag == term)
-                   }.flatMap(_.filterLabels { label =>
-                       tagSearchTerms.exists(term => label.contains(term))
-                     })
-                     .getOrElse(spec.spec)
-                 }
-             }
+      spec    <- specInstance.runner.run(FilteredSpec(specInstance.spec, args))
```
hmemcpy: Is there a difference between running the spec directly with `spec.run` or passing it to the `runner`?

hmemcpy: Whatever it is, if I run the (unfiltered) spec with `specInstance.runner.run`, it causes `StackTracesSpec` to fail, for some reason... The only difference, as far as I can tell, is this skips weaving of the aspects (which is currently just the timeout?). Unless there's something I'm missing, @adamgfraser?

adamgfraser: @hmemcpy I think that should be the only difference. `StackTracesSpec` can be a little finicky because it depends on the exact trace that is rendered so if you do something that causes an additional step to be included in the trace the test can fail.  There is a debug option in that test suite if you want to figure it out but I think you want to weave in the test aspects, right?

hmemcpy: @adamgfraser I think so... in the original code it seems, though, that if there's no filter, the entire spec is run with all aspects applied (1):
```
case (Nil, Nil) => spec.run    <-- (1)
               case (testSearchTerms, Nil) =>
                 spec.runner.run {  <-- (2)
                   spec.spec.filterLabels { label =>
                     testSearchTerms.exists(term => label.contains(term))
                   }.getOrElse(spec.spec)
                 }
                 ...
```

however, if there are filtered items, it runs it through the runner directly, bypassing the aspect weaving (2).

After my change, it always runs with aspects applied, regardless of whether it was filtered or not. All tests are passing, so I don't think that's a problem.

hmemcpy:
I'm exploring how I can provide an external test reporter, but this PR is good to go!

test-sbt/shared/src/main/scala/zio/test/sbt/BaseTestTask.scala
hmemcpy: @adamgfraser Wow, this actually was a bug! Filtered items did not have aspects applied to them. I inadvertently fixed it by always applying the aspects!

If you currently run e.g. `coreTestsJVM/testOnly *.StackTracesSpec -- -t foreachTrace` with the master branch, you'll get test failures!

adamgfraser:
@hmemcpy Great work on this!

hmemcpy:
Yay! Thanks a lot. I'll pick your brain about the other thing 🎉

PR #2965, opened 2020-02-24
ZIO Test: Implement Diagnose Test Aspect

adamgfraser:


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -162,6 +161,24 @@ object TestAspect extends TimeoutVariants {
...
 
+  /**
+   * An aspect that runs each test on a separate fiber and prints a fiber dump
+   * if the test fails or has not termianted within the specified duration.
+   */
+  def diagnose(duration: Duration): TestAspectAtLeastR[Live] =
+    new PerTest.AtLeastR[Live] {
+      def perTest[R <: Live, E](test: ZIO[R, TestFailure[E], TestSuccess]): ZIO[R, TestFailure[E], TestSuccess] = {
+        def dump[E, A](fiber: Fiber.Runtime[E, A]): ZIO[Live, Nothing, Unit] =
+          fiber.dump.flatMap(_.prettyPrintM).flatMap(s => Live.live(console.putStrLn(s)))
```
iravid: Could be nice to prefix this with the test’s label, if possible.

adamgfraser: Good idea! Will add.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -162,6 +161,33 @@ object TestAspect extends TimeoutVariants {
...
+   */
+  def diagnose(duration: Duration): TestAspectAtLeastR[Live] =
+    new TestAspectAtLeastR[Live] {
+      def some[R <: Live, E](predicate: String => Boolean, spec: ZSpec[R, E]): ZSpec[R, E] = {
+        def diagnose[R <: Live, E](
+          label: String,
+          test: ZIO[R, TestFailure[E], TestSuccess]
+        ): ZIO[R, TestFailure[E], TestSuccess] =
+          test.fork.flatMap { fiber =>
+            Live.live(clock.sleep(duration)) *> fiber.poll.flatMap {
```
darl: What happens if test completes before `duration` elapsed?

adamgfraser: Good point. We should race the test with the sleep fiber. Then if the test completes first we can just print the fiber dump if the test failed and interrupt the sleep fiber. If the sleep fiber completes first we can print the fiber dump and then join back the test fiber. Thank you!

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -162,6 +161,33 @@ object TestAspect extends TimeoutVariants {
...
+          test: ZIO[R, TestFailure[E], TestSuccess]
+        ): ZIO[R, TestFailure[E], TestSuccess] =
+          test.fork.flatMap { fiber =>
+            fiber.join.raceWith(Live.live(ZIO.sleep(duration)))(
+              (exit, sleepFiber) => dump(label, fiber).when(!exit.succeeded) *> sleepFiber.interrupt *> ZIO.done(exit),
+              (_, _) => dump(label, fiber) *> fiber.join
+            )
+          }
+        def dump[E, A](label: String, fiber: Fiber.Runtime[E, A]): ZIO[Live, Nothing, Unit] =
+          fiber.dump.flatMap(_.prettyPrintM).flatMap(s => Live.live(console.putStrLn(s"$label: $s")))
```
jdegoes: We should do this for all children of the fiber, too!

adamgfraser: Ah. I thought that `dump` already did that. I added a method to the `Fiber` companion object `dumpAll` to do that and implemented the existing `dump` method that dumps all fibers in terms of that.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -592,10 +592,17 @@ object Fiber extends FiberPlatformSpecific {
...
    */
-  val dump: UIO[Iterable[Dump]] = UIO.effectSuspendTotal {
+  val dump: UIO[Iterable[Dump]] =
+    dumpAll(_rootFibers.asScala: @silent("JavaConverters"))
+
+  /**
+   * Collects a complete dump of the specified fibers and all children of the
+   * fibers.
+   */
+  def dumpAll(fibers: => Iterable[Fiber.Runtime[_, _]]): UIO[Iterable[Dump]] = UIO.effectSuspendTotal {
```
jdegoes: Maybe the old one should be called `dumpAll`, and this one, `dumpOnly` or just `dump`?

adamgfraser: Yes. I like `dumpAll` for dump everything from the root fibers and just `dump` for this.

adamgfraser: Updated now.

PR #2971, opened 2020-02-25
ZIO Test: Fix Subtype Assertion

### LINKED ISSUE #2970: ZIO Test: Subtype Assertion Does Not Work with Primitive Types
The following test fails:

```scala
test("isSubtype") {
  assert(1)(isSubtype[Int](anything))
}
```

The reason appears to be that the `runtimeClass` of the `ClassTag` is `int` but the class of the actual value is `java.lang.Integer`. We probably need some special logic to handle these types.

Thanks to @sken77 for reporting!

adamgfraser:
Resolves #2970. Class tags have an unapply method that contains special logic for handling primitive types we can use.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Assertion.scala
+++ b/test/shared/src/main/scala/zio/test/Assertion.scala
@@ -597,10 +597,7 @@ object Assertion extends AssertionVariants {
    * }}}
    */
   def isSubtype[A](assertion: Assertion[A])(implicit C: ClassTag[A]): Assertion[Any] =
-    Assertion.assertionRec("isSubtype")(param(className(C)))(assertion) { actual =>
-      if (C.runtimeClass.isAssignableFrom(actual.getClass())) Some(actual.asInstanceOf[A])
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -597,10 +597,7 @@ object Assertion extends AssertionVariants {
    * }}}
    */
   def isSubtype[A](assertion: Assertion[A])(implicit C: ClassTag[A]): Assertion[Any] =
-    Assertion.assertionRec("isSubtype")(param(className(C)))(assertion) { actual =>
-      if (C.runtimeClass.isAssignableFrom(actual.getClass())) Some(actual.asInstanceOf[A])
-      else None
-    }
+    Assertion.assertionRec("isSubtype")(param(className(C)))(assertion)(C.unapply(_))
```
iravid: TIL about ClassTag.unapply!

PR #2972, opened 2020-02-25
ZIO Test: Improve Support for Setting and Getting Random Seeds

adamgfraser:
Changes default `TestRandom` instance to `deterministic`. Adds a new method on `TestRandom` `getSeed` to get the random seed. Adds test aspects `nondeterministic` to set a random seed and `setSeed` to set a specified seed.

I think this gives us pretty much everything we want in terms of being able to set the seed. There is a little more work to do on getting the seed. Right now the user can get the seed with `TestRandom.getSeed` but I think the ideal behavior would be that if any test fails where `TestRandom` was invoked the seed is displayed. I think the easiest way to do this would be to make the `TestRandom` layer depend on the `Annotations` layer so that internally we can write annotations. Then we could maintain a `Ref` to track whether the test random has ever been invoked and if it is the first time we write to the annotations map.

However, I think this is also going to require some augmentations to annotations. Right now annotation renderers don't have access to whether a test passed or failed. I think we would only want to display this information for failed tests. Also, I'm not sure monoid composition makes sense here. The seed associated with a particular test is very meaningful, the seed of seeds or the first seed for some subset of tests in a suite does not seem particularly meaningful.

Alternatively we could just have the `TestExecutor` extract the seed from the annotations map directly and pass it to the `TestReporter`, but that seems less scalable.


jdegoes:
Maybe we could add an "annotation type" sum type:

 — `Diagnostic` (or `Debug`)
 — `Standard`

and diagnostic annotations would be printed only in the event of failures?

I would define the monoid append to wipe out information. So suites have no information on seeds.

adamgfraser:
Wouldn’t obliterating violate the law that a plus empty equals a?

jdegoes:
> Wouldn’t obliterating violate the law that a plus empty equals a?

Depends on the empty. But for this one it would be `Option[Seed]` (with empty = `None`) or something so combining two of those could always yield `None`.

jdegoes:
^^^ Need to look at that

adamgfraser:
Yeah, isn’t that the opposite? If None is the empty value then empty plus something else would have to be the other thing, not empty. Practically I think using optional values could work well and could even solve the problem of successful tests because at the end of successful tests we could just obliterate the value but seems like no our composition may be too limiting.

jdegoes:
> Yeah, isn’t that the opposite? If None is the empty value then empty plus something else would have to be the other thing, not empty.

You're right, it's not a monoid, but it's a semigroup, at least. In any case, we can encode it with annotation.

adamgfraser:
Agreed.

jdegoes:
@adamgfraser Good to merge as soon as conflicts are fixed!

jdegoes:
I'm not sure what to do on the rendering of seeds. I think possibly the renderers should access the _whole_ `TestAnnotationMap` for the test / suite. So they can peek at other values. Then we just have to make sure we have enough default annotations that the renderers can make good decisions (e.g. a post-test annotation saying whether a test succeeded, failed, or was ignored).

This doesn't make test annotations any more powerful (which is good), but gives rendering more flexibility (which is probably fine since it's "non-critical" just reporting data to the user).

But I'm not sure on it. There may be better approaches.

adamgfraser:
On it.

adamgfraser:
@jdegoes Fixed.

PR #2975, opened 2020-02-26
Support for custom test renderers

hmemcpy:
The `DefaultTestReporter` currently has two responsibilities: rendering the report and printing it out to the console. This makes it a bit harder to introduce a different renderer to support custom output.

This PR separates the responsibilities of the test reporter: it introduces a new interface, `TestRenderer`, which can render a given executedSpec. The `DefaultTestReporter` now accepts a renderer as an argument and uses it to output the report.

In addition, introduced a new flag `-renderer` to pass the fqn type name of the custom renderer, instantiating it with reflection, falling back to the default renderer.

This change will allow me to supply a custom test renderer that outputs IntelliJ-specific format, and opens the API for other implementations, such as the JUnit format (needed for certain CI servers).


test/shared/src/main/scala/zio/test/SummaryBuilder.scala
```diff
@@ -16,7 +16,7 @@ object SummaryBuilder {
                  case _                          => false
                }
       failures <- extractFailures(executedSpec)
-      rendered <- ZIO.foreach(failures)(DefaultTestReporter.render(_, TestAnnotationRenderer.silent))
+      rendered <- ZIO.foreach(failures)(renderer.render(_, TestAnnotationRenderer.silent))
```
hmemcpy: This previously coupled the `SummaryBuilder` to the `DefaultTestReporter`. I now pass the renderer from the outside as an argument, but wonder if passing it via `R` can be a better idea? WDYT?

test/shared/src/main/scala/zio/test/TestRunner.scala
```diff
@@ -29,12 +29,14 @@ import zio.internal.Platform
...
   platform: Platform = Platform.makeDefault().withReportFailure(_ => ()),
-  reporter: TestReporter[E] = DefaultTestReporter(TestAnnotationRenderer.default),
-  bootstrap: ZLayer.NoDeps[Nothing, TestLogger with Clock] = ((Console.live >>> TestLogger.fromConsole) ++ Clock.live)
+  renderer: TestRenderer[E] = DefaultTestRenderer(),
+  bootstrap: ZLayer.NoDeps[Nothing, TestLogger with Clock] = (Console.live >>> TestLogger.fromConsole) ++ Clock.live
 ) { self =>
 
   lazy val runtime = Runtime((), platform)
 
+  lazy val reporter = DefaultTestReporter(renderer, TestAnnotationRenderer.default)
```
hmemcpy: The reporter is now just passes rendered results to the logger provided by `TestLogger`. If someone wants to change it, I suppose this can be done by providing a different bootstrap layer.

test/shared/src/main/scala/zio/test/SummaryBuilder.scala
adamgfraser: I think generally passing it as a parameter is simpler if we don't need the power of the environment to avoid plumbing the dependency through a bunch of different levels. Seems pretty straightforward here.

test/shared/src/main/scala/zio/test/TestRunner.scala
```diff
@@ -29,12 +29,14 @@ import zio.internal.Platform
 final case class TestRunner[R <: Has[_], E](
   executor: TestExecutor[R, E],
   platform: Platform = Platform.makeDefault().withReportFailure(_ => ()),
-  reporter: TestReporter[E] = DefaultTestReporter(TestAnnotationRenderer.default),
```
adamgfraser: It seems like we are losing the ability to supply different test annotation renderers here right now. Maybe you pass both a test renderer and a test annotation renderer? Or maybe you pass a test reporter and you can easily construct a test reporter from a test renderer and an annotation renderer?

jdegoes:
@hmemcpy I will close this for now since it's been inactive, but please re-open when you're ready to proceed! 

Supporting IntelliJ IDEA-style rendering and JUnit-style rendering would be phenomenal for tool integration and I think is critical to our overall adoption strategy. Just need to figure out how to do that cleanly and generically. /cc @adamgfraser

PR #2976, opened 2020-02-27
Added signal for fiber dump

### LINKED ISSUE #2263: Add a signal handler for dumping the fiber stack in zio.App
Using `sun.misc.{Signal, SignalHandler}`, add a signal hander for `SIGUSR1` that dumps the current fiber stack to stderr.

Those classes will work for the JVM, but not for ScalaJS. Something equivalent needs to be found for ScalaJS (and it should only be registered if running under node.js).

The fiber stack can be retrieved using `Fiber.dump` and `Dump#prettyPrintM`.

softinio:
Resolves #2263 

# Testing

### Local Test results

https://gist.github.com/softinio/fe20ee01dd2c85314ca26bb27b11237a

I also ran website local to see all links and page working which they did: 

https://shared.softinio.com/Screen-Shot-2020-03-14-18-13-33.01.png

https://shared.softinio.com/Screen-Shot-2020-03-14-18-02-34.26.png


adamgfraser:
@softinio I would start by adding a `addInterruptionHook` method to `zio.internal.PlatformSpecific.scala` analogous to the existing `addShutdownHook` method. Then you can add methods in `Platform` and `Runtime` that let the user register an effect to be performed on interruption. We could then implement convenience methods to let the user configure dump on interrupt on those and automatically do it for `App`, `ManagedApp`.

I think because this is at the very edge of our program it is going to be very hard to test other than by implementing it and verifying that it works as expected when used.

core/jvm/src/main/scala/zio/Signals.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import sun.misc.{Signal, SignalHandler}
+import zio.console._
+
+object Signals {
+  def fiberDump: SignalHandler =
+    Signal.handle(new Signal("USR1"), (sig: Signal) => {
+      for {
+      dumps <- Fiber.dump
+      _ <- URIO.foreach(dumps)(_.prettyPrintM.flatMap(console.putStrLn))
```
adamgfraser: We may want to consolidate all the dumps into one string and print that to make sure that nothing else being printed concurrently is printed in between the dumps.

core/jvm/src/main/scala/zio/Signals.scala
```diff
@@ -0,0 +1,15 @@
+package zio
+
+import sun.misc.{Signal, SignalHandler}
+import zio.console._
+
+object Signals {
+  def fiberDump: SignalHandler =
+    Signal.handle(new Signal("USR1"), (sig: Signal) => {
+      for {
```
iravid: For this to actually print, you'll need the runtime here and an `unsafeRun` on the resulting effect.

jdegoes:
@softinio Not that you need more to do but how's this one coming along? 😆

softinio:
@jdegoes done. Thanks for your patience.

@adamgfraser can you review again? Let me know if the docs I updated makes sense or if I can improve. Thanks so much for your help earlier.

softinio:
Updated PR description.

core/jvm/src/main/scala/zio/internal/PlatformSpecific.scala
```diff
@@ -38,6 +40,13 @@ private[internal] trait PlatformSpecific {
       }
     }
 
+  /**
+   * Adds a signal handler that executes the specified action on receiving a signal.
+   */
+  def addSignalHandler(signal: String)(handler: () => Unit): Unit = {
```
adamgfraser: This is a pretty unsafe method as we saw earlier because if a signal handler has already been installed for the signal this will throw. What if we catch the error here and change the function to return a boolean indicating either true if we were able to successfully install the signal hander or false if another signal handler was already installed?

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -36,6 +37,31 @@ trait Runtime[+R] {
    */
   val platform: Platform
 
+  /**
+   * Add generic signal handler
+   *
+   * @param signal to listen for
+   * @param handler to run on receiving signal
+   */
+  def addSignalHandler(signal: String)(handler: URIO[R, Unit]): Unit =
```
adamgfraser: I'm wondering if we should move these methods to `Platform` and then provide aliases in `Runtime`, similar to the existing treatment of methods like `withReportFailure`.

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -36,6 +37,31 @@ trait Runtime[+R] {
...
+    Platform.addSignalHandler(signal)(() => unsafeRun(handler))
+
+  /**
+   * Use SIGUSR2 signal to interrupt runtime
+   *
+   * @param handler to run when SIGUSR2 is received
+   */
+  def addInterruptHandler(handler: URIO[R, Unit]): Unit =
```
adamgfraser: Maybe we should change these names to `withSignalHander`, `withInterruptHandler`, and `withFiberDumpOnInterrupt` to mirror the existing methods.

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -36,6 +37,31 @@ trait Runtime[+R] {
...
+    addSignalHandler("USR2")(handler)
+
+  /**
+   * Print Fiber Dump of all fibers when interrupt signal received
+   *
+   * @param ev is the console environment
+   */
+  def printFiberDumpOnInterrupt(implicit ev: R <:< Console): Unit =
```
adamgfraser: We may want to just use `println` here instead of `console.putStrLn`, which would allow us to simplify this signature and make this always available. We're already doing this at the edge of our program where we are unsafely running this and since it is for debugging purposes I think arguably we would want to print to the standard output even if there is a different console implementation (e.g. a test console). Users who want to do something fancier can always fall back on the more general combinators we created.

docs/howto/fiber_dump.md
```diff
@@ -0,0 +1,95 @@
...
+title:  "Sending Interrupt Signal to get fiber dump"
+---
+
+You can send a `SIGUSR2` signal from your os to interrupt your app and a get a fiber dump of all fibers.
+
+You need to add the hook by including `printFiberDumpOnInterrupt` method in your runtime environment.
+
+## Example application 
+
+```scala
```
adamgfraser: `mdoc`?

softinio:
@adamgfraser Pushed my changes for your PR comments but haven't tested it yet. Going to test tonight and will let you know. 

Thanks.

adamgfraser:
Can you fix the documentation?

core/js/src/main/scala/zio/internal/PlatformSpecific.scala
```diff
@@ -35,6 +35,13 @@ private[internal] trait PlatformSpecific {
     val _ = action
   }
 
+  /**
+   * Adds a signal handler that executes the specified action on receiving a signal.
+   */
+  @silent("never used")
+  def addSignalHandler(signal: String)(handler: () => Unit): Boolean =
+    true
```
adamgfraser: Should this actually return `false`? I thought the meaning of the return value was `true` if the signal handler was successfully installed and `false` otherwise. Here the signal handler was not successfully installed (admittedly because we don't currently support installing signal handlers on ScalaJS at all). So I think `false` is most accurate in communicating to the caller that the signal handler they tried to install is never going to run.

core/jvm/src/main/scala/zio/internal/PlatformSpecific.scala
```diff
@@ -38,6 +40,19 @@ private[internal] trait PlatformSpecific {
...
+  /**
+   * Adds a signal handler that executes the specified action on receiving a signal.
+   */
+  def addSignalHandler(signal: String)(handler: () => Unit): Boolean =
+    try {
+      val _ = Signal.handle(new Signal(signal), _ => handler())
+      true
+    } catch {
+      case e: IllegalArgumentException =>
+        println(s"Failed to install signal: $e")
```
adamgfraser: I think we don't want to print this and instead the caller can decide how to handle the failure to install the signal.

core/native/src/main/scala/zio/internal/PlatformSpecific.scala
```diff
@@ -33,6 +33,11 @@ private[internal] trait PlatformSpecific {
   def addShutdownHook(action: () => Unit): Unit =
     ()
 
+  /**
+   * Adds a signal handler that executes the specified action on receiving a signal.
+   */
+  def addSignalHandler(signal: String)(handler: () => Unit): Boolean = true
```
adamgfraser: See comment above regarding return value.

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -77,17 +78,48 @@ trait Platform { self =>
    */
   def reportFailure(cause: Cause[Any]): Unit
 
+  /**
+   * Print Fiber Dump of all fibers when interrupt signal received
+   *
```
adamgfraser: Extra line?

docs/howto/fiber_dump.md
```diff
@@ -0,0 +1,95 @@
+---
+id: howto_fiber_dump
+title:  "Sending Interrupt Signal to get fiber dump"
+---
+
+You can send a `SIGUSR2` signal from your os to interrupt your app and a get a fiber dump of all fibers.
```
adamgfraser: Great documentation! Is there a way to get this to work when you do CTRL-C in SBT? That would be a really nice feature. Also may be good to mention current lack of support on Scala JS and Scala Native.

docs/resources/resources.md
```diff
@@ -88,4 +88,3 @@ _These articles reflect the state of ZIO at the time of their publication. The c
 ## Support
 
 - [Offical Discord Server](https://discord.gg/2ccFBr4)
-- [Official Gitter Channel](https://gitter.im/ZIO/Core)
```
adamgfraser: Yes!

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -77,17 +78,48 @@ trait Platform { self =>
...
+   */
+  def withFiberDumpOnInterrupt: Platform =
```
iravid: Is the terminology here (and other places) right? We're not dumping the fibers on *interruption*, just on receiving the signal.

softinio: Yes its when we receive a signal. Maybe I should rename to: 

```suggestion
  def withFiberDumpOnSignal: Platform =
```

adamgfraser: Yes, there is a little bit of ambiguity here because we want to say we will do the fiber dump when we receive this specific `USR2` signal. So `onSignal` is a little vague and `OnInterrupt` is confusing with ZIO interruption. Potentially we could do `onInterruptSignal` or `onInterruptionSignal` or otherwise `onSignal` is probably fine and I doubt too many people are going to be adding other signal handlers.

iravid: Yeah that'd be great :+1: There are a bunch of other places that could use the correction - on `Runtime` and on the docs

softinio: yup I'll make the changes everywhere thanks for feedback. Agreed!

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -77,17 +78,47 @@ trait Platform { self =>
...
       override def reportFailure(cause: Cause[Any]): Unit = f(cause)
     }
+
+  /**
+   * Add generic signal handler
+   *
+   * @param signal to listen for
+   * @param handler to run on receiving signal
+   */
+  def withSignalHandler(signal: String)(handler: Unit): Boolean =
```
adamgfraser, line 111: I think we want the handler here to be a function `() => Any`.

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -77,17 +78,47 @@ trait Platform { self =>
...
+        ()
+      }
+    }
+
+  /**
+   * Use SIGUSR2 signal to interrupt runtime
+   *
+   * @param handler to run when SIGUSR2 is received
+   */
+  def withInterruptHandler(handler: Unit): Boolean =
```
adamgfraser, line 97: `withInterruptSignalHandler`?

core/shared/src/main/scala/zio/internal/Platform.scala
adamgfraser: `onSignal`?

core/shared/src/main/scala/zio/internal/Platform.scala
adamgfraser: `onInterruptSignal`?

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -77,17 +78,47 @@ trait Platform { self =>
    */
   def reportFailure(cause: Cause[Any]): Unit
 
+  /**
+   * Print Fiber Dump of all fibers when interrupt signal received
+   */
+  def onInterruptSignal: Platform =
```
adamgfraser, line 84: We can't expose this here because we need a `Runtime` that can unsafely run the effect. Instead we will implement this on `Runtime` and call `onInterruptSignal`, using the runtime to unsafely run the effect in the handler.

adamgfraser:
@softinio Just wanted to check in on this. Would be great to get it in before 1.0.

softinio:
@adamgfraser agreed! Apologies for the delay! Bandwidth deprived :-) 

WIll try to get it done this week.

adamgfraser:
Sounds good. Depending on when the fiber supervision changes land we may need to update for that. Maybe we could pair on that if you want?

jdegoes:
I think we are going to have to move this into ZIO ZMX. Because there is no more fiber graph in this project as of the auto-supervision changes. Will close here.

@softinio Thanks for your work on this and ZIO ZMX!

PR #2977, opened 2020-02-27
Implemented Task.fromOption

wi101:
sorry I used option.get  🙈 but it will be catched! 🤷‍♀


adamgfraser:
I think this is definitely good functionality to have.

The only issue I see is that in every other case except for the `apply` methods when we have a method on the companion object it is an alias for the method on `ZIO` and has the same signature narrowed as appropriate. In contrast, here we have:

```scala
object ZIO {

  /**
   * Lifts an `Option` into a `ZIO`.
   */
  def fromOption[A](v: => Option[A]): IO[Unit, A] =
    effectTotal(v).flatMap(_.fold[IO[Unit, A]](failNow(()))(succeedNow))
}

object Task {

  /**
    * Wraps the value of the given option into a Task, if the option is empty the task fails with NoSuchElementException
    */
   final def fromOption[A](v: => Option[A]): Task[A] = Task(v.get)

}
```

Where `Throwable` is not a subtype of `Unit`. I think having this type of consistency is important but this is still a valuable method so I would suggest we give it a different name and then we can define it on `ZIO` and alias it on the companion objects as appropriate. Perhaps `getOrFail`?

wi101:
you're right,  I would add `getOrFail` to RIO too 
but I would remove `fromOption` from RIO I saw it now it returns `IO[Unit, A] `

core/shared/src/main/scala/zio/RIO.scala
```diff
@@ -418,14 +418,17 @@ object RIO {
   def fromFutureInterrupt[A](make: ExecutionContext => scala.concurrent.Future[A]): Task[A] =
     ZIO.fromFutureInterrupt(make)
 
-  def fromOption[A](v: => Option[A]): IO[Unit, A] = ZIO.fromOption(v)
```
adamgfraser: Good catch!

core/shared/src/main/scala/zio/Task.scala
```diff
@@ -421,6 +421,11 @@ object Task extends TaskPlatformSpecific {
   def fromTry[A](value: => scala.util.Try[A]): Task[A] =
     ZIO.fromTry(value)
 
+  /**
+   * Wraps the value of the given option into a Task, if the option is empty the task fails with NoSuchElementException
+   */
+  final def getOrFail[A](v: => Option[A]): Task[A] = Task(v.get)
```
adamgfraser: I would move this implementation to the ZIO companion object and just alias it here.

Also, I think it will be slightly more efficient if you do something like `ZIO.effectSuspendTotal(v.fold(ZIO.failNow(new NoSuchElementException("None.get")))(ZIO.succeedNow))`. There is some cost in throwing an exception and creating the stack trace and we don't need that since we have our own trace. Also I think we want to make sure we die if any errors are thrown other than just the fact that the option is empty.

core/shared/src/main/scala/zio/RIO.scala
```diff
@@ -418,14 +418,17 @@ object RIO {
...
-
   /**
    * @see See [[zio.ZIO.fromTry]]
    */
   def fromTry[A](value: => scala.util.Try[A]): Task[A] =
     ZIO.fromTry(value)
 
+  /**
+   * @see See [[zio.Task.getOrFail]]
```
adamgfraser: Would refer to the documentation on the ZIO companion object here.

core/shared/src/main/scala/zio/RIO.scala
```diff
@@ -418,14 +418,17 @@ object RIO {
...
+   */
+  final def getOrFail[A](v: => Option[A]): Task[A] = Task.getOrFail(v)
```
adamgfraser: I think we could provide this method on the `IO` companion object as well.

PR #2979, opened 2020-02-28
ZIO Test: TestClock Improvements

adamgfraser:
A couple of incremental improvements to the TestClock:

1. Currently when the TestClock is restored as part of combinators like `nonFlaky` the flag of whether the warning that sleep is being used but the TestClock is not being adjusted gets restored as well. This can result in the warning being rendered multiple times if the clock is not being adjusted. This PR fixes this so the warning will only be rendered once per test.
2. Adds a method `runAll` that sets the clock time to infinity so all pending effects run and any new scheduled effects run immediately.
3. Some simplifications to the internal representation of TestClock.


jdegoes:
Awesome work. Would be nice to have a test aspect to call runAll too!

adamgfraser:
Good idea! I will submit another PR with improvements to the TestConsole in a minute and will fold the test aspect into that.

PR #2980, opened 2020-02-28
ZIO Test: TestConsole Improvements

adamgfraser:
One of the pain points of working with ZIO Test is that console output created using `console.putStrLn` is not displayed to the console so for debugging purposes users have to resort to `UIO(println(line))` which is not normally best practice and not what they would be using in their code. This PR changes the TestConsole to by default display console output to standard output in addition to writing it to the buffer. For most applications that are not console applications this should be strictly an improvement. For cases where the console output clutters the display test aspects `debug` and `silent` can be used to turn this on and off. Also adds test aspect `noDelay` for setting the TestClock so all effects run without delay.

### DIFF
--- a/test/shared/src/main/scala/zio/test/environment/package.scala
+++ b/test/shared/src/main/scala/zio/test/environment/package.scala
@@ -613,7 +613,7 @@ package object environment extends PlatformSpecific {
        * being written to the output buffer.
        */
       def debug[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
-        debugState.locally(false)(zio)
+        debugState.locally(true)(zio)
 
       /**
        * Writes the specified sequence of strings to the input buffer. The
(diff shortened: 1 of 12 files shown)


test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -580,13 +580,20 @@ package object environment extends PlatformSpecific {
...
+      def feedLines(lines: String*): UIO[Unit]
+      def output: UIO[Vector[String]]
+      def silent: UIO[Unit]
     }
 
-    case class Test(consoleState: Ref[TestConsole.Data]) extends Console.Service with TestConsole.Service {
+    case class Test(
+      consoleState: Ref[TestConsole.Data],
+      live: Live.Service,
+      debugState: Ref[Boolean]
```
jdegoes: FiberRef?

adamgfraser: Good idea! Now everything is scoped!

PR #2982, opened 2020-02-28
ZIO Core: Add Platform Flags

adamgfraser:
It can be useful to know at runtime what platform we are on, for example so we can implement details of the fiber supervision model to reflect the platform. This PR adds flags identifying the platform available internally to ZIO similar to the ones already exposed in ZIO Test.


adamgfraser:
The only reason I didn't is because those are exposed publicly and I think that is useful for more complex scenarios in testing where you don't want to just ignore a test on a platform but assert something different for example. Whereas for ZIO itself I think we just want to use them internally.

PR #2984, opened 2020-02-29
Document That Mutable Value Should Not Be Used Inside Ref

adamgfraser:


core/shared/src/main/scala/zio/Ref.scala
```diff
@@ -30,6 +30,12 @@ import java.util.concurrent.atomic.AtomicReference
  *   _   <- console.putStrLn("Value = " + v) // Value = 5
  * } yield ()
  * }}}
+ *
+ * NOTE: WHile `Ref` provides the functional equivalent of a mutable reference,
```
ritschwumm: typo: "WHile"

adamgfraser: Good catch! Will fix tonight.

PR #2987, opened 2020-03-01
Fix lazy constructors on streams

pierangeloc:
being `ZStream.fromInputStream` lazy, it causes infinite loop (and heap exhaustion) when invoked with an inputstream coming from a `def`.
The same happens for `fromIterator`. Fixed by capturing their value in the inner expression. Also made tests passing `def` values rather than `val`


adamgfraser:
@pierangeloc Can you merge upstream changes?

pierangeloc:
> @pierangeloc Can you merge upstream changes?

@adamgfraser rebased on latest master. Ran `test` and `fmt` locally so CI should be happy now

jdegoes:
@pierangeloc Thank you!

PR #2995, opened 2020-03-02
Fix small errors in `ZIO.sandbox` scaladocs

felher:
The old version of the scaladoc for `ZIO.sandbox` did not compile and if updated to compile, it didn't work because it matched on `Cause.Die` directly, but the cause is wrapped within a `Traced`. The following changes were made:

* Let `caught` return an `IO[DomainError, Unit]` instead of `UIO[Unit]`
* Map the error with `_.untraced` to remove the outer layer and make the pattern matching work.
* Change the `IO.succeed(0)` to `IO.unit`, because some folks will get an `discarded non-Unit value` warning otherwise.


jdegoes:
@felher Thank you! These examples really _do_ matter. 

Also: congratulations on your first ZIO Pull request!

giphy

PR #2999, opened 2020-03-02
Use Tagged In Test Annotations

### LINKED ISSUE #2998: ZIO Test: Use lightweight type tag instead of ClassTag in TestAnnotation

adamgfraser:
Resolves #2998. Should we replace `ClassTag` with `Tagged` everywhere it is used?


jdegoes:
Almost certainly, given `ClassTag` permits no polymorphism...

adamgfraser:
Okay, I will do that after the release notes.

PR #3000, opened 2020-03-03
ZIO Mock - composable mocks

### LINKED ISSUE #2692: ZIO Mock: mocking composed environments
Given a program that depends on more modules, eg:

```scala
val app =
  for {
    n <- random.nextInt
    _ <- console.putStrLn(s"$n")
    m <- random.nextLong
    s <- console.getStrLn
  } yield s"$s : ${n + m.toInt}"
```

We can construct individual mocks for `Console` and `Random` expecting each to be called twice, then convert them to mock implementations and provide to the program. 

```scala
import MockConsole._
import MockRandom._

val mockRandom = (nextInt._1 returns value(1)) *> (nextLong._0 returns value(2L))
val mockConsole = (putStrLn(equalTo("1") returns unit) *> (getStrLn returns value("foo"))

val combinedMock: ZLayer.NoDeps[Nothing, Console with Random] =
  mockRandom ++ mockConsole
```

However the relative order in which the metods were called will only be checked within each service's scope.

So the mocks will check that `nextLong` was called after `nextInt`, and that `getStrLn` was called after `putStrLn`, but will not check if `nextLong` was called after or before `putStrLn`.

So the following program would also pass the test:

```scala
val app =
  for {
    n <- random.nextInt
    m <- random.nextLong
    _ <- console.putStrLn(s"$n")
    s <- console.getStrLn
  } yield s"$s : ${n + m.toInt}"
```

Sometimes that is OK, and we don't care about the total order of calls, as long as the relative order is kept, but sometimes it might be relevant (especially when testing business processes) to make sure some calls happen before others in the "larger" environment.

The construction of such expectations would be straightforward and is already possible:

```scala
import MockConsole._
import MockRandom._

val largerEnvMock = (
  (nextInt._1 returns value(1)) *>
  (putStrLn(equalTo("1") returns unit) *>
  (nextLong._0 returns value(2L)) *>
  (getStrLn returns value("foo"))
)  
```

What we're missing is a way to auto-derrive `Mockable[A with B]` given `Mockable[A]` and `Mockable[B]`. Now a mockable is simply a single function `Mock => Has[A]`.

```scala
implicit def compose[A, B]()(implicit m1: Mockable[A], m2: Mockable[B], Tagged[Has[B]]): Mockable[A with B] =
  (mock: Mock) => {
    val env1: Has[A] = m1.environment(mock)
    val env2: Has[B] = m2.environment(mock)
    val env3: Has[A] with Has[B] = env1 union env2

    // this does not compile
    // it gets `Has[A] with Has[B]` but expects `Has[Has[A] with Has[B]]`
    env3
  }
```

I suspect if we changed the `Mockable` a little we could actually get that auto-derrivation working and have this super nice feature in ZIO Mock.

Both approaches could be composed, eg given a program that depends on 3 modules `A`, `B` and `C`, we could create a mock for `A with B` and seperately a mock for `C` and combine them using `++` just like we did in the first example, which would give us a mocked environment for `Has[A] with Has[B] with Has[C]` where relative order of calls on `A` and `B` is checked, while on `C` we just check that the expected calls happened, not careing when (in relation to A and B calls).

@jdegoes @adamgfraser I'm putting this out as I'm not sure when I'll have the time to follow this up, so if anyone has time, feel free to take this.

ioleo:
Fixes #2692

This is a major ZIO Mock upgrade focusing on:
- composable expectations, similar to `Assertion`
- introducing cross-module mocks with auto-derrivation of `Mockable` for composed environment

Things that were previously possible, but syntax has changed:
- checking all expectations *in sequential order*
  **before**: `A *> B *> C *> D`
  **now**: `A andThen B andThen C andThen D` or `A ++ B ++ C ++ D`

Things that work now which were previously not possible:
- checking all expectations *in any order* `A and B and C and D` or `A && B && C && D`
- expecting only one of `A or B or C or D` or `A || B || C || D`
- repeated expectations `(A or B) repeats (3 to 5)`
- combining all of above to express more possible valid scenarios eg. `((A && B) ++ ((C || D) atMost 3) ++ E ++ (F atLeast 2)`
- checking relative order of calls between multiple services

```scala
import zio.{ clock, console, random, system }
import zio.clock.Clock
import zio.console.Console
import zio.duration._
import zio.random.Random
import zio.system.System
import zio.test.{ suite, ZIOBaseSpec }
import zio.test.Assertion._
import zio.test.mock.Expectation._

object ExampleSpec extends ZIOBaseSpec {

  def spec = suite("ExampleSpec")(
    suite("mocking composed environments")(
      testM("Random with Clock with System with Console") {
        val cmd1 = MockRandom.nextInt._1 returns value(42)
        val cmd2 = MockClock.sleep(equalTo(42.seconds)) returns unit
        val cmd3 = MockSystem.property(equalTo("foo")) returns value(None)
        val cmd4 = MockConsole.putStrLn(equalTo("None")) returns unit
        val composite = (cmd1 ++ cmd2 ++ cmd3 ++ cmd4)

        val app =
          for {
            n <- random.nextInt
            _ <- clock.sleep(n.seconds)
            v <- system.property("foo")
            _ <- console.putStrLn(v.toString)
          } yield ()

        val result = composite.build.use(app.provide _)
        assertM(result)(check)
      }
    )
  )
}
```

## Follow-up tasks

- [ ] write a complete example for the docs, showcasing all the features
- [x] investigate [weird test issue](https://github.com/zio/zio/pull/3000/files#diff-08c38641279d835bb350377d1300ca28R316)
- [x] auto-derrivation of `Mockable` for dotty
- [x] consult / confirm chocies for method names (eg `++` as alias for `andThen`)
- [x] fix issue with scala-reflect dependency


test-tests/shared/src/test/scala/zio/test/mock/MockSpec.scala
```diff
@@ -0,0 +1,350 @@
...
+            hasField(
+              "expectation",
+              _.expectation,
+              isSubtype[E0](
+                hasField[E0, L](
+                  "children",
+                  _.children,
+                  isSubtype[L](
+                    /**
+                     * No idea why this takes so long... @adamgfraser could you take a look?
```
ioleo: @adamgfraser Please have a look here.

test-tests/shared/src/test/scala/zio/test/mock/MockComposedEnv.scala
```diff
@@ -0,0 +1,53 @@
...
+            _ <- clock.sleep(n.seconds)
+            v <- system.property("foo")
+            _ <- console.putStrLn(v.toString)
+          } yield ()
+
+        testSpecComposed[Random with Clock with System with Console, Throwable, Unit](
+          "Random with Clock with System with Console"
+        )(composite, program, isUnit)
+      }
+    ) @@ exceptDotty
```
jdegoes: Why doesn't this one work on Dotty?

test/shared/src/main/scala-2.x/zio/test/mock/MockableDerrivation.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.test.mock
+
+import zio.Has
+
+trait MockableDerrivation {
```
jdegoes: `MockableDerivation`

test/shared/src/main/scala-2.x/zio/test/mock/Macros.scala
```diff
@@ -0,0 +1,70 @@
...
+
+package zio.test.mock
+
+import scala.reflect.macros.blackbox.Context
+
+import zio.Has
+
+private[test] object Macros {
+
+  def summonMockable_impl[R <: Has[_]: c.WeakTypeTag](c: Context): c.Expr[Mockable[R]] = {
```
jdegoes: This is the first instance of using macros in Core... if we're going to do this for 2.x we should make sure there's a corresponding path forward for Dotty.

test/shared/src/main/scala-dotty/zio/test/mock/MockableDerrivation.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package zio.test.mock
+
+import zio.Has
+
+trait MockableDerrivation {
+
+  /**
+   * Derrives an instance of `Mockable[R]` given a mockable for each
```
jdegoes:
```suggestion
   * Derives an instance of `Mockable[R]` given a mockable for each
```

test/shared/src/main/scala-2.x/zio/test/mock/Macros.scala
adamgfraser: We actually use a them already to implement the `assertCompiles` functionality, though that is fairly specialized and Dotty provides specific support for that.

test-tests/shared/src/test/scala/zio/test/mock/MockComposedEnv.scala
ioleo: The composed environment auto-derrivation is done via macro.

test/shared/src/main/scala-2.x/zio/test/mock/Macros.scala
ioleo: Well, the `typeChecks` assertion was already using macros in `zio.test` (and this one is also in `zio.test`) and there was really no other way (implicit induction won't work here unfortunately), thats why I decided to introduce it.

For now the effect on Dotty is -> you'd have to create the composed `Mockable` instanced on your own. The `MockableDerivation` is lowest in the implicit search, so if you do provide it (by any other means) the disfunctional/missing macro in Dotty won't affect you.

If you do nothing, currently Dotty auto-derrivation will throw compile-time error.

ioleo: I have not yet went deeply into Dotty, but AFAIK some features of meta-programming for Dotty are not yet available.

jdegoes:
@ioleo The docs are all failing. Can you update those?

docs/howto/mock_services.md
```diff
@@ -246,27 +246,21 @@ object AccountObserverSpec extends DefaultRunnableSpec {
 
 ## Mocking multiple collaborators
 
-In some cases we have more than one collaborating service being called. In such situations we need to build our expectations separately for each
-service and then combine them into single environment:
+In some cases we have more than one collaborating service being called. You can create mocks for rich environments, a mockable for it will be automatically derrived if you have implicit mockables for each part of composed environment in scope.
```
jdegoes:
```suggestion
In some cases we have more than one collaborating service being called. You can create mocks for rich environments, a mockable for it will be automatically derived if you have implicit mockables for each part of composed environment in scope.
```

ioleo: Yes, fixing failing CI tests one by one now.

jdegoes:
@ioleo The functionality of extending expectation to `and` and `or`, as well as adding `repeat` is extremely useful and something I've wanted to do for a while now.

I'll let others do a more detailed review but excited for how this is shaping up!

jdegoes:
@ioleo If we want to get this into RC18 then the tests will need to be passing ASAP.

ioleo:
@jdegoes Working on it. Yes I'd like to get it into RC18.

test-tests/shared/src/test/scala/zio/test/mock/MockSpec.scala
adamgfraser: @ioleo Looking at this now.

adamgfraser: There is definitely something very wrong here. It looks like `Assertion#isSubtype` is being called 450,000 times in this test.

adamgfraser: Could this be related to the invocations search? I am running tests with multiple nested subtype assertions outside of the mocking framework and am not seeing anything unusual.

ioleo: Not really, this is an assertion on the result of the call. The Mock framework ends (in this case) with a `ZIO.die(UnsatisfiedExpectationsException(...))` and all I'm trying to do here is to check, that the structure of the failed expectation is as expected.

adamgfraser: Narrowing on the issue. Not sure on timing to address.

jdegoes:
@ioleo Testing these locally will lead to a much faster debug cycle... and also save us $$$. 😉

test-tests/shared/src/test/scala/zio/test/mock/MockSpec.scala
adamgfraser: Okay, I have a fix. Give me a minute to push.

ioleo:
@jdegoes @adamgfraser Unfortunately I gotta go now and probably won't be able to return to this until tomorrow. I've moved some files into `scala-2.x` due to problems on dotty as a quick fix, but I believe with current config `MockComposedEnvSpec` is not being run at all (my intention was to compile&run it under 2.x only).

The JS builds also fail due to exceeded memory, but I can't make any connection what is the underlaying issue.

jdegoes:
@ioleo It's ok to take your time, we will still get this in for ZIO 1.0.

ioleo:
Finally fixed all the issues. Haveing missed the window for RC18, I plan to add a bit more into this PR:
* [ ] more/better docs/examples
* [ ] move `@mockable` macro for method tags generation from `zio-macros` into `ZIO Test`

jdegoes:
@ioleo Can you fix conflicts? Would like to get in for the Monday release.

ioleo:
@jdegoes I'm on it.

ioleo:
@jdegoes I've been thinking some more about this and came to conclusion that we can replace the implicits completely by ZLayer. I've got a working implementation ready, only missing two bits:
- [ ] removed Spyable, could not adapt existing implementation to ZLayer, gotta give it some more thinking to find the right encoding for this
- [ ] half-baked `@mockable` macro, in my opinion, it does not make sense to merge this without the accompanying macro

jdegoes:
Compilation errors. Can you fix? Would like to get this merged in ASAP!

ioleo:
@jdegoes Hmm weird, I did run full suite in 2.11 and did not have this issue.

```
[error] /home/circleci/project/test-tests/shared/src/test/scala/zio/test/mock/ExpectationSpec.scala:17:50: illegal cyclic reference involving object MockRuntime
[error]       test("A and B")(assert(A and B)(equalTo(And(A :: B :: Nil, ModuleMock.Mock)))),
[error]                                                  ^
[error] /home/circleci/project/test-tests/shared/src/test/scala/zio/test/mock/ExpectationSpec.scala:18:45: Symbol 'type zio.test.mock.internal.MockRuntime.Service' is missing from the classpath.
```

I think it's some cacheing on CI bug. :thinking:

ioleo:
Regarding the implicit-free `Spy` implementation it's possible with slightly diffrent (a bit awkward?) API.
The general idea is that `Method` tag has a reference to it's spy/mock machinery. As we combine expectations, we always start with a `Method` tag, eg `MockConsole.GetStrLn ...`, therefore we can leverage the fact that we can update ZLayer and every time we combine Expectations, we also combine the underlaying `MockRuntime -> R` and `R -> MockRuntime` mappings (encoded as Layers).

`Method -> ZLayer[MockRuntime, Nothing, R] -> Layer[Nothing, R]`

We could do the same if for Spyies, if we start with some object that holds a reference to related spy/mock pair. 

```scala
case class Invocation[R <: Has[_], I, A](f: PartialFunction[(I, A), UIO[Unit]])(val spy: ZLayer[R, Nothing, MockRuntime], val mock: ZLayer[MockRuntine, Nothing, R])
def spy[R <: Has[_]](cases: Invocation[R, _, _]*): Layer[Nothing, R] =
// here we foldLeft on cases and as we go we combine spy/mock for individual services into spy/mock for whole R
// then we use it
```

The API would be less flexible, as for example to do:
"count all invocations, never mind which method" you'd have to list all:
```scala
spy(
  invocation for PutStrLn,
  invocation for PutStr,
  invocation for GetStrLn
)
```

while in previous encoding this could be done by one simple partialFunction.

#### Alternatives

Alternatives include: have users explicitly build the "spy" instance, by API like:
`def spy(SpyConsole ++ SpyClock)(env: R)(f: PartialFunction ...)`

Or revert to implicits, but that has other drawbacks:
* either we pack imlicits into `main` (eg. to be auto-discoverable implicits for console must be in `Console` companion object), which breaks the seperation of prod code/test code
* or we force the users to import implicits, at which point there is no point of using implicits at all (if you must manually point the compiler, better have an API require explicitly passing spy/mock layers)

docs/howto/mock_services.md
```diff
@@ -78,126 +78,169 @@ ZIO Test provides a framework for mocking your modules.
...
-helpers (within `>` object) in the service companion object:
-
-```scala mdoc:invisible
-trait AccountEvent
-```
+We'll be assuming you've read [How to use modules and layers][doc-use-modules-and-layers] guide. In the main sources we define the _service_, a module alias and _capability accessors_. In test sources we're defining the _mock object_ which will hold _capability tags_ and mock layer machinery.
 
 ```scala mdoc:silent
 import zio.test.mock._
+import zio.test.mock.internal.MockRuntime
```
adamgfraser: Feels like this `MockRuntime` is pretty pervasive to be in an internal package. Also can we come up with an lias so we don't have to write `Has[MockRuntime]` everywhere? Maybe use the traditional pattern where type `MockRuntime = Has[MockRuntime.Service]`?

docs/howto/mock_services.md
```diff
@@ -246,27 +289,20 @@ object AccountObserverSpec extends DefaultRunnableSpec {
 
 ## Mocking multiple collaborators
 
-In some cases we have more than one collaborating service being called. In such situations we need to build our expectations separately for each
-service and then combine them into single environment:
+In some cases we have more than one collaborating service being called. You can create mocks for rich environments, a mockable for it will be automatically derived if you have implicit mockables for each part of composed environment in scope.
```
adamgfraser: This needs to be updated, right?

test/shared/src/main/scala-2.x/zio/test/mock/MockableMacro.scala
```diff
@@ -0,0 +1,169 @@
...
+package zio.test.mock
+
+import scala.reflect.macros.whitebox.Context
+
+import com.github.ghik.silencer.silent
+
+/**
+ * Generates method tags for a service into annotated trait/object.
+ */
+private[mock] object MockableMacro {
```
adamgfraser: I wonder if either the macro functionality or the mocking functionality should be its own project.

test/shared/src/main/scala/zio/test/DefaultTestReporter.scala
```diff
@@ -52,12 +47,12 @@ object DefaultTestReporter {
             specs <- executedSpecs
             failures <- UIO.foreach(specs)(_.exists {
                          case Spec.TestCase(_, test, _) => test.map(_.isLeft);
-                         case _                         => UIO.succeedNow(false)
+                         case _                         => UIO.succeed(false)
```
adamgfraser: Why this change?

jdegoes:
@ioleo I am a fan of "no implicit imports". The need to import implicits is a big red flag and a huge usability obstacle. But I haven't yet taken a close enough look at this PR (I'll @adamgfraser do that) to have any suggestions on that front.

adamgfraser:
So it seems like we should be able to do something to preserve the spying functionality.

Conceptually, the interesting idea of the mocking framework to me is defining a translation from methods to data types that describe those methods. In the old encoding you were then capturing those data types into a sequential structure and now you are capturing them in a more sophisticated structure.

What the spying functionality added on top of that was a translation from the data types back to the methods. That way you actually had an isomorphism between them and you could map from the method call to the data type, do something based on the data type, and then map back.

It seems like we should be able to do that with the new encoding. Maybe we need a new service that is like one of the test services where it implements the existing service and also provides new methods. Then in addition to implementing going from methods to data types it would implement going from data types to methods.

adamgfraser:
For example here is an implementation of a spy under the current encoding;

```scala
implicit val spyableCounter: Spyable[Counter.Service] =
  new Spyable[Counter.Service] {
    def environment(mock: Mock): Has[Counter.Service] =
      Has {
        new Counter.Service {
          def increment: UIO[Unit] = mock(Command.Increment)
          def decrement: UIO[Unit] = mock(Command.Decrement)
          def get: UIO[Int]        = mock(Command.Get)
          def reset: UIO[Unit]     = mock(Command.Reset)
        }
      }
    def mock(environment: Has[Counter.Service]): Mock =
      new Mock {
        def invoke[R0, E0, A0, M0, I0](method: Method[M0, I0, A0], input: I0): ZIO[R0, E0, A0] =
          method match {
            case Command.Increment => environment.get.increment
            case Command.Decrement => environment.get.decrement
            case Command.Get       => environment.get.get
            case Command.Reset     => environment.get.reset
          }
      }
  }
```

You can see that in your current implementation of the mock layer you are essentially doing the first half of that:

```scala
val Mock: ZLayer[Has[MockRuntime], Nothing, Example] =
   ZLayer.fromService(mock =>
     new Example.Service {
       val static                                 = mock(Static)
       def zeroArgs                               = mock(ZeroArgs)
       def zeroArgsWithParens()                   = mock(ZeroArgsWithParens)
       def singleArg(arg1: Int)                   = mock(SingleArg, arg1)
       def multiArgs(arg1: Int, arg2: Long)       = mock(MultiArgs, arg1, arg2)
       def multiParamLists(arg1: Int)(arg2: Long) = mock(MultiParamLists, arg1, arg2)
       def command(arg1: Int)                     = mock(Command, arg1)
       def overloaded(arg1: Int)                  = mock(Overloaded._0, arg1)
       def overloaded(arg1: Long)                 = mock(Overloaded._1, arg1)
     }
   )
 ```

We need a service that does the second half of that to spy.

File: docs/howto/mock_services.md (hunk 7 above)
Comment (ioleo) [reply docs/howto/mock_services.md]: Renamed to `Mock` and moved one package up. Instead old `Mock` became `MockFactory` and moved to internal. Regarding the alias - that was my initial approach, but then I had those weird cyclic dependency errors in 2.11, which seemed to be caused by the package object/alias to `MockRuntime`.

Comment (adamgfraser) [reply docs/howto/mock_services.md]: Yeah, take a look at where we put the object and alias in ZIO Core and ZIO Test for the test implementations to see the best we came up with to work around this.

File: test/shared/src/main/scala-2.x/zio/test/mock/MockableMacro.scala (hunk 9 above)
Comment (ioleo) [reply test/shared/src/main/scala-2.x/zio/test/mock/MockableMacro.scala]: The macros are closly tried to the implementation details and whenever we change ZIO Mocks, the Mockable macro must be updated aswell, that is why I've moved this into `zio/zio`. We could extract this into seperate subproject though. Since most of `zio-macros` (delegate family) got obsoleted with `ZLayer` and `Has` we are planning to move also `@accessible` into `zio/zio` (maintaining a seperate repository for just one macro feels like too much overhead).

Comment (adamgfraser) [reply test/shared/src/main/scala-2.x/zio/test/mock/MockableMacro.scala]: Yes, if we do anything agree it would be in the same repository. Would be like the ZIO Test Magnolia project. That would solve the problem of updating things at the same time while potentially being more modular.

File: test/shared/src/main/scala/zio/test/DefaultTestReporter.scala (hunk 10 above)
Comment (ioleo) [reply test/shared/src/main/scala/zio/test/DefaultTestReporter.scala]: My mistake when rebaseing. Reverting.

**Comment by ioleo** [comment]
@adamgfraser Yes, agreed that the mock part is "half of spying". The thing is, we have no way to get the correct `Spy` from a type `R <: Has[_]`, as there is no way in Scala (without macros/reflection) to deconstruct this type into its parts.

So, the previous spy API `def spy[R <: Has[_]](environment: R)(f: PartialFunction[..., ...])` will not work for composed environments. 

With this new mocking approach I am under the hood constructing the layer for composed environment, every time you add another Expectation.Call (which we build from `Method[_, _, _]` tags).

image

With spying however, the order is reversed. We're starting with a composed `R <: Has[_]` environment and want to deconstruct it into parts for which we then can create mocks (with our spying logic) and compose them back into an `R`. So when we do:

```scala
def spy(env: R) {
  case Invocation(MockConsole.PutStrLn, _, _) => ???
  case Invocation(MockClock.Nanotime, _, _) => ???
  case Invocation(MockSystem.Property, _, _) => ???
}
```

there is no way AFAIK (if there is, I'd love to learn it!) to "get out" the method tags from this pattern matching. Therefore we can't hide (in the tags) the layers for individual parts of the final environment, like I did with Mocks.

docs/howto/mock_services.md
```diff
@@ -78,126 +78,166 @@ ZIO Test provides a framework for mocking your modules.
...
   }
+}
+
+// test sources
+object ExampleMock {
 
-  object Service {
-    object processEvent extends Method[Service, AccountEvent, Unit]
+  sealed trait Tag[I, A] extends Method[Example, I, A] {
+    val mock: ZLayer[Has[Mock], Nothing, Example] = ???
```
adamgfraser: Can we do `type Mock = Has[Mock.Service]` like we do with all the other services?

docs/howto/mock_services.md
```diff
@@ -78,126 +78,166 @@ ZIO Test provides a framework for mocking your modules.
...
   }
 
-  // autogenerated `object Service { ... }`
-  // autogenerated `object > extends Service[AccountObserver] { ... }`
-  // autogenerated `implicit val mockable: Mockable[AccountObserver] = ...`
+  def processEvent(event: AccountEvent) =
+    ZIO.accessM[AccountObserver](_.get.processEvent(event))
 }
+
+@Mockable[AccountObserver.Service]
```
adamgfraser: Any way to avoid having to specify the type parameter here?

ioleo: No. In previous approach we did not have to do that becouse we were annotating the companion object (in which the Service lied), however that was bad practice, as it was introducing `test` dependencies into your `src/main` code.

Here, we are generating the Mock machinery into a completely unrelated object and we need to somehow inform the macro for which service we are generating tags.

adamgfraser:
@ioleo Sorry, didn't mean to delete your comment. Was deleting my own since you answered it on Discord and I think you were responding at the same time.

I have two thoughts on the spying functionality.

First, maybe the user needs to specify the spying logic at the level of the individual service. You implement a spy for the console service and then you compose that spied on service into a larger environment with all interactions with the environment that use that service spied on.

Second, in your proposed signature could we just always write any method, input, and output to a `Ref` and just return the results of the `Ref` to the user and the user could handle as they wanted? Seems like that could be more ergonomic.

ioleo:
@adamgfraser I tried again and again to use the `object Module { trait Service }` + package object with alias `type Module = Has[Module.Service]` pattern, but I kept bumping into:

```
[error] /home/user/Projects/zio/test-tests/shared/src/test/scala/zio/test/mock/module/ModuleMock.scala:46:46: illegal cyclic reference involving type Proxy
[error]   private[test] lazy val envBuilder: URLayer[Proxy, Module] =
```

I tried renaming, moveing to another package, clearing cache, adding `-Ybreak-cycles` scalac option but this kept failing on `2.11`. So I finally give up and decided to shorten the name to `Proxy` and use `Has[Proxy]` where applicable.

I'll add this to my personal "tech dept" list to investigate later, perhaps with fresh perspective, for now I'd like to move on and get this merged.

Also in latest commit rebased to current master and added some finishing refactors.

adamgfraser:
Okay. If I have time I will see if I can get it working later today and then let's get this merged.

PR #3009, opened 2020-03-03
Add a note to disambiguate the `foreach` method's name

vil1:
Since the name `foreach` is already widely used in Scala's standard library with a slightly different meaning, I thought it would be valuable (especially for the people whose first contact with Scala will be the ZIO library) to add a comment stressing out that fact.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2484,6 +2484,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * returns the results in a new `List[B]`.
    *
    * For a parallel version of this method, see `foreachPar`.
+   *
+   * NOTICE: this has a different meaning than the `foreach` method one would find elsewhere
+   * in Scala's standard library. On standard collections, `foreach` applies a side-effecting
+   * function on the elements of the collection and returns `Unit`. Here, the `foreach` method
+   * applies the `f` function to every elements of `in` and aggregates the results in a single
+   * `ZIO` context. In other libraries, this behaviour is most often called `traverse`.
    */
   def foreach[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, E, B]): ZIO[R, E, List[B]] =
     in.foldRight[ZIO[R, E, List[B]]](effectTotal(Nil))((a, io) => f(a).zipWith(io)((b, bs) => b :: bs))
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2489,7 +2489,8 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * in Scala's standard library. On standard collections, `foreach` applies a side-effecting
    * function on the elements of the collection and returns `Unit`. Here, the `foreach` method
    * applies the `f` function to every elements of `in` and aggregates the results in a single
-   * `ZIO` context. In other libraries, this behaviour is most often called `traverse`.
+   * `ZIO` context. 
+   * ADVANCED: In other typeclass-based FP libraries, this behaviour is most often called `traverse`.
    */
   def foreach[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, E, B]): ZIO[R, E, List[B]] =
     in.foldRight[ZIO[R, E, List[B]]](effectTotal(Nil))((a, io) => f(a).zipWith(io)((b, bs) => b :: bs))


fanf:
I really, really like the proposed evolution! It allows curious people to learn about what the classical naming for the method is, it's properties, etc! It's fabulous, because it allows people to gain access to a 30 year old knowledge base, and build up on the shoulders of past giants. And for people who don't care and just want to finish the task that the boss gave them as fast as possible, well they won't read it anyhow. 

I would like to propose a small modification: find a common word/tag that could be used everywhere in method doc to signal that that part of the doc is targetted to user who want to go a bit further than just using said method. These section would be able to talk about:

- classical name in the literrature of the concept (traverse, sequence obviously),
- provided properties (like composition, monoid law, etc)
- underlying concept that makes the construction sound (error type based on semi ring comes to mind), 
- links to external doc/article/etc to go even further. 

It would really help willing novices to go further and know where to look. 

On twitter, @hmemcpy proposed "ADVANCED" and the term seems good: it enconpasses both the fact that you don't need to read it and the fact that the section can contains mathematical/CT concepts.

vil1:
@fanf, you want me to change `NOTICE` to `ADVANCED`, it that it?

fanf:
yes, first step. 
And then, in other pr, use `ADVANCED` everywhere to spread knowledge :)

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2484,6 +2484,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * returns the results in a new `List[B]`.
    *
    * For a parallel version of this method, see `foreachPar`.
+   *
+   * NOTICE: this has a different meaning than the `foreach` method one would find elsewhere
+   * in Scala's standard library. On standard collections, `foreach` applies a side-effecting
+   * function on the elements of the collection and returns `Unit`. Here, the `foreach` method
+   * applies the `f` function to every elements of `in` and aggregates the results in a single
+   * `ZIO` context. In other libraries, this behaviour is most often called `traverse`.
```
vil1:
```suggestion
   * `ZIO` context. 
   * ADVANCED: In other typeclass-based FP libraries, this behaviour is most often called `traverse`.
```

adamgfraser:
I wonder if we should put this in the documentation on the website instead of in the Scaladoc. This is great content but there is some opportunity cost to the Scaladoc being very long, especially when it is not directly related to how to use the code. If we had a section on the website for this we could really go into it and I think not have that constraint.

core/shared/src/main/scala/zio/ZIO.scala
vil1: @fanf something like that 👆?

fanf:
@adamgfraser I don't know, I really like java std lib long explanation that you find in thread pool and the like. They deals with general topic, reference external papers and implementation choice, etc. It's extremelly enlightning to read them, and I would definitly never have read them if they weren't in the javadoc. None theless, I can see your point, and perhaps we should try to keep it minimal in scaladoc, with a pointer to extended documentation on the site. That would be the best of both world: 

- knowledge colocated with the code, so people get discover easely when the code that there is more to learn, 
- extended information in a doc-like place, updatable independly from sources, and also usable outside of a dev env. 

Something like that:

```
   * NOTICE: this has a different meaning than the `foreach` method one would find elsewhere
   * in Scala's standard library. In other typeclass-based FP libraries, this behaviour is most often called `traverse`. To learn more about these difference, go to: https://zio.dev/docs/overview/overview_basic_operations
```
(well, `foreach` is not in basic operation, not sure where it is introduced before parallelism?)

jdegoes:
@vil1 Thanks for the pull request!

A few of my own thoughts on the issue:

> NOTICE: this has a different meaning than the `foreach` method one would find elsewhere in Scala's standard library. On standard collections, `foreach` applies a side-effecting function on the elements of the collection and returns `Unit`. 

`ZIO.foreach` _also_ applies an effectful function on the elements of a collection. 

There are only 2 differences between `ZIO.foreach` and standard `foreach`:

1. `ZIO.foreach` uses functional effects, whereas standard `foreach` uses side-effects (that's an expected difference and uniform throughout ZIO, not unique to `foreach`).
2. `ZIO.foreach` collects the return values of the function, whereas standard `foreach` discards them. 

For (2), it's common to ignore the collected values most of the time. When they are not ignored, imperative programming, it's an established convention to use a mutable list and push results onto the list during the iteration, for subsequent processing outside the iteration.

Users commonly write code like this:

```scala
def uploadAllFiles(files: List[File]): Task[Any]
  ZIO.foreach(files) { file =>
    uploadFile(file)
  }

def askQuestionsGetAnswers(questions: List[String]) = 
  ZIO.foreach(questions) { question =>
    for {
      _      <- putStrLn(question)
      answer <- getStrLn
    } yield answer
  }
```

etc. This is the kind of code they write in the wild without coaching.

> ADVANCED: In other typeclass-based FP libraries, this behaviour is most often called `traverse`

This is perhaps not precise enough. `traverse` is bound by law to return the same structure it accepts as input; so if you feed it a `Tree[A]`, you get back a `Tree[B]`, with the same structure, the only difference being that each `A` is effectfully processed into a `B` using the provided applicatively effectful function. This guarantee is _not_ maintained by `ZIO.foreach`, which always returns a `List` as output and only accepts Scala collections as input.

A more accurate comment would be:

> ADVANCED: The generalization of ZIO.foreach to arbitrary traversable data structures and arbitrary lax monoidal functors is often called `traverse` in type-class-based functional programming libraries.

However, who does this comment help? A functional programmer who has found `ZIO.foreach` will already know this from the type signature (in other words, it's too late to help them if they found `ZIO.foreach` already). And from my experience, non-functional programmers don't care.

In addition, one could make the argument that every method should point programmers to additional resources, e.g. `map` should have a comment:

> ADVANCED: The `map` function models a covariant endofunctor in SCAL and is often called `fmap` in type-class based functional programming libraries.

Etc. There are literally dozens of places where one could insert "FP docs" into ZIO, which would be mildly interesting for functional programmers but not really essential for them; and not useful for anyone who comes to the library for async/concurrent machinery.

I think @adamgfraser's suggestion makes more sense: put this into the website, because then functional programmers have a chance of finding it. You can imagine a whole:

### ZIO for Cats Programmers
### ZIO for Scalaz Programmers
### ZIO for Haskell Programmers

which goes through and explain things for functional programmers: `ZIO.foreach` is a specialization of `traverse` to collections and ZIO effects; `Cause[E]` is a free semiring with an additive and multiplicative identity; etc.

jdegoes:
Lastly: this strikes me as a solution in search of a problem. No users of ZIO have ever complained of being confused between Scala `foreach` and ZIO `foreach` (to me). That's not to say they don't exist, but I teach this frequently and interact on all the forums and I don't recall any confusion between Scala `foreach` and ZIO `foreach`; and, moreover, the only complaints I hear about the names are for people who are already familiar with `traverse` / `sequence`.

fanf:
Some more of my thought on the issue, and then I will let it go. 
@jdegoes: the core of my comment above was to allow people to learn from ZIO and expend their knowledge to new horizons - that by definition they don't know exist yet. There is people who want to understand the underlying structures and concept when they use a function/framework (and actually, Spring success was built around that: it was a mine of information about IoC, DI, what and how they choose to do it, etc and people who were learning were excited about all they learn and in the same time it was solving pain point, so they can't stop talking about it). These people could look at the scala doc and see "ho ! That's nice ! I was using that thing like I was used to do b/c of my past experience, but I see there is other way to do it. Why are they telling it's more powerfull ?". 
The point is really *not* to have zio use `traverse`. And it is also not to explain ZIO to cats/haskell user - while to be clear, I'm not at all arguing that it wouldn't be nice and usefull and a wonderfull source of hints, an awesome "theory to code: how to implement effect management, modules included, in scala and fulfill Martin grand vision of FP in the small, OO in the large").

I'm sensitive to that "showing the path" point because without frameworks and people to show me better ways, I would not have learnt as much, I wouldn't even know that these constructs existed and why/how they were built - or it would have taken me many more years, or, as I'm not that smart, I would have remaind with broken things with no idea why they were broken.
For full story, I build something similar to `traverse` when I started scala. For total irony, I called it `sequence`. It had bugs on some cases (you can easily imagine which: imagine how a Java dev approch `traverse` as you said with mutable states and lists, and without chechink any rules or having the least idea about what an applicative functor is). It was really great to learn why my code wasn't working, that is was a classical problem with well documented solutions, and that other people had worked hard to provide generic solutions - and it was nice to learn that even if a lib provided said solution. It made me a better developper. 

`ZIO` was born along with a massive coaching and as plateform which lifted hundreds of developpers to functionnal programming. That spirit was amazing, a massive wave of learning that fuelled a generation of new contributors, and they learned not only by direct coaching, but also because they where able to take advantage from litterature, other framework, and other giant's shoulders. It was nice, and it is that spirit of letting people who don't know most things -yet - the opportunity to learn with a guiding hand, a breacrump of words and concepts that I tried (obviously badly) to advertise in above comment.

Until now, I didn't want to let it go, whatever @hmemcpy adviced me. Sorry if it was perceived as purist not understanding that real life devs don't care of such things. It was not. (Look at rudder code: does it look like code produced by FP ivory tower wizard? Not even remotly. There is vars and returns, and all kind of lovecraftian horrors in it). It was intented to be the voice of a not so long ago java dev who need guidance from smarter and more knowledgable people to get things done, and who was at home with `ZIO` ecosystem spirit.

But I had missed the point where somewhere in the path, `ZIO` guiding spirit morphed from that "lifting people to better themself"'s adventure into an utilitaristic framework that pragmatically care for a clearly identified, narrow market target, the _java devs with imperative only background and who don't care to learn new concepts nor become better at their work but by using magic framework_ (as far as I understand by your comments). "lifting people to FP" is burried deep. I can't arg with that, as I'm not in your shoes, and goals evolve with life. I'm a bit sad that it seems to be an exclusive `or` in your mind.

Oh, and for the record: your comment was *deeply* interesting, and it clearly shows that you are totally able to know what could go in these "to go further" sections, and I'm sure *loads* of people who read it learnt something, especially the one that are not that good in FP, and who had no ideas of the all the subtilities that you talk about - me the first. And the "me of 10y ago" would have loved it even more - yes, the java devs coding in scala, just starting to see that better alternatives were possible. That past me would have love "ZIO for imperative java dev who are willing to learn why it works so well, what are the rules and the benefits". And my dev culture (ie lazyness) would have loved to have pointers to its relevant parts directly from the code. 

Now I let it go @hmemcpy.

Cheers!

fanf:
(yes, I know, let it go, let it go: but I wanted to highlight the amazingly extra wonderful job @adamgfraser is doing on discord and everywhere actually, also everytime, to help hundreds of people, whatever their problem - thank you so much Adam for that!)

jdegoes:
Since there's no consensus on this and we don't have any end-user reports of problems in this area, let's wait for more user-level feedback before we dive in here. I do think the "ZIO for Haskell/etc" guides is a good idea and should be written up separately.

jdegoes:
@fanf Btw, FP education is very important to me still, and I will do a lot for that this year. But my number one strategy for promoting FP is no longer education, but rather, library development. I hope that puts us all in a place where there's more demand than ever for FP education.

PR #3010, opened 2020-03-03
Fix Website

### LINKED ISSUE #3008: Zlayer docs missing from microsite due to misconfiguration
New pages added to docs need to be added to the side bar config here: 

https://github.com/zio/zio/blob/master/website/sidebars.json#L68

Noticed this page was not:

https://github.com/zio/zio/blob/master/docs/howto/use_modules_and_layers.md

Be good to do an audit of the docs to make sure nothing else missing.

adamgfraser:
Resolves #3008.

I removed the existing "how to use the module pattern" page in this PR in favor of the new one that uses layers. That page overlaps substantially with the layers page and does not reflect the current best practice for the environment. I think if we are going to get people to use the new environment encoding we need to be consistent in all our own documents so I think we should remove this for now. If we want to add back an updated version in the future we can definitely do that but for the website to be a live document that is putting our best foot forward I think we should remove for now.


adamgfraser:
@softinio What do you think about merging this and then publishing the website again?

softinio:
> @softinio What do you think about merging this and then publishing the website again?

I really was hoping you wouldn't ask this though waited in anticipation :-) 

You need to a new release to update the docs. maybe a `v1.0.0-RC18-1` or `v1.0.0-RC19`

There is a manual way though painful.

adamgfraser:
Hmmm. Yes publishing again doesn’t sound good. Is there a way we can decouple that? It seems like especially with 1.0 coming up the code is going to be relatively stable but I think there is a lot more we could do to improve the documentation so it would be nice to make that closer to real time.

softinio:
> Hmmm. Yes publishing again doesn’t sound good. Is there a way we can decouple that? It seems like especially with 1.0 coming up the code is going to be relatively stable but I think there is a lot more we could do to improve the documentation so it would be nice to make that closer to real time.

@adamgfraser for sure decoupling an option and has been discussed in the passed. I think we even have an old issue for it. Also as discussed in passed I would like a snapshot documentation site where it gets auto published on merge to master same as snapshot does. 

Decoupling has drawbacks too as for example our build auto updates the docs with sbt instructions to with correct version to use based on the tag you did release with. Decoupling will become a manual effort. 

Better solution is probably leave coupling there but have an efficient way to do manual doc updates too. Maybe a cli tool. 

For now I say do a new release before anything significant gets merged into master. As same day as RC18 its fine and only thing will be is the doc updated so it doesn't matter if people use what we released earlier or now. DO IT :-)

PR #3015, opened 2020-03-04
Services refactor

vasilmkd:
@adamgfraser @jdegoes Do you think this is a useful change for code organization purposes? I feel like the further nesting of the `Service` is not beneficial and the package itself provides enough disambiguation.


jdegoes:
I'd like a `ClockService` that is adjacent to `Clock = Has[ClockService]`. That'd allow wildcard imports without clashes (`Service` is very susceptible to clashes).

However, whether that's enough of an improvement to warrant breaking the existing `XYZ.Service` convention, I'm not sure...

adamgfraser:
I'm inclined to think we need a little stability after the change to the environment encoding.

vasilmkd:
Thanks. Closing.

PR #3016, opened 2020-03-04
Change laziness of when methods.

quelgar:
Two small changes to `ZIO.when` and `ZIO.whenM`.

The first is straight-forward: in most cases of `when`, the boolean parameter is strictly evaluated, so it should be passed by value.

The second is more debatable: changed the effect parameter to be lazy, so the effect value is only constructed when the test is true, as would be the case for the `if` equivalent. The current strict construction of the effect value can be surprising.

In my case, I was dealing with a non-typesafe API and I had the code constructing the effect value perform a cast that was only valid if the test was true. Took me a while to understand why I was getting `ClassCastException`. This can be fixed with `ZIO.effectSuspendTotal`, but it's not that obvious.

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -679,7 +679,7 @@ object IO {
   /**
    * @see See [[zio.ZIO.when]]
    */
-  def when[E](b: => Boolean)(io: IO[E, Any]): IO[E, Unit] =
+  def when[E](b: Boolean)(io: => IO[E, Any]): IO[E, Unit] =
     ZIO.when(b)(io)
 
   /**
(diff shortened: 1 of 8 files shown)


neko-kai:
`when` was passed by value before, then changed to lazy. So I doubt this change would be accepted...

jdegoes:
@quelgar I would definitely take the _additional_ laziness on the effect variations. But I would not be in favor of the eager versions for the `Boolean`, because they will help in some cases but hurt in others (and for consistency I'd rather all effect constructors be lazy to keep as many fatal errors inside ZIO as possible).

jdegoes:
Something we might be able to do is try / catch `ClassCastException` somewhere and turn it into a more useful error if you could provide a reproducer for that.

quelgar:
> @quelgar I would definitely take the _additional_ laziness on the effect variations. But I would not be in favor of the eager versions for the `Boolean`, because they will help in some cases but hurt in others (and for consistency I'd rather all effect constructors be lazy to keep as many fatal errors inside ZIO as possible).

Ok, so the `Boolean` is lazy for consistency with the cases where it actually matters? But in the `ZIO.when` case it's not _really_ lazy, is it? It's always evaluated immediately anyway.

Idle thought: suspend the whole thing?

```scala
def when[R, E](b: => Boolean)(zio: => ZIO[R, E, Any]): ZIO[R, E, Unit] =
    ZIO.effectSuspendTotal(if (b) zio.unit else unit)
```

PR #3018, opened 2020-03-04
Fix `catchSome` docstring

ruurtjan:
The current example throws a compile error: `class FileNotFoundException is not a case class, nor does it have a valid unapply/unapplySeq member`. For Java classes like the `FileNotFoundException`, the syntax shown in https://zio.dev/docs/overview/overview_handling_errors#catching-some-errors should be used.


ruurtjan:
Ha, thanks!

PR #3031, opened 2020-03-05
Add tests and documentation for TSemaphore

calvinlfer:
@FrancisToth and I noticed that there weren't any tests for TSemaphore whilst documenting it so we added some :)

Part of https://github.com/zio/zio/issues/2122

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/TSemaphoreSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TSemaphoreSpec.scala
@@ -1,0 +1,94 @@
+package zio.stm
+
+import zio.ZIOBaseSpec
+import zio.random.Random
+import zio.test.Assertion._
+import zio.test._
+
+object TSemaphoreSpec extends ZIOBaseSpec {
+  override def spec = suite("TSemaphore")(
+    suite("factories")(
+      testM("make") {
+        checkM(Gen.long(1L, Int.MaxValue)) { expected =>
+          val actual = for {
+            sem <- TSemaphore.make(expected)
+            cap <- sem.available
+          } yield cap
+
+          assertM(actual.commit)(equalTo(expected))
+        }
+      }
+    ),
+    suite("acquire and release")(
+      testM("acquiring and releasing a permit should not change the availability") {
+        checkM(Gen.long(1L, Int.MaxValue)) { expected =>
+          val actual = for {
+            sem <- TSemaphore.make(expected)
+            _   <- sem.acquire *> sem.release
+            cap <- sem.available
+          } yield cap
+          assertM(actual.commit)(equalTo(expected))
+        }
+      },
+      testM("used capacity must be equal to the # of acquires minus # of releases") {
+        checkM(usedCapacityGen) {
+          case (capacity, acquire, release) =>
+            val actual = for {
+              sem <- TSemaphore.make(capacity)
+              _   <- repeat(sem.acquire)(acquire) *> repeat(sem.release)(release)
+              cap <- sem.available
+            } yield cap
+
+            val usedCapacity = acquire - release
+            assertM(actual.commit)(equalTo(capacity - usedCapacity))
+        }
+      },
+      testM("acquireN/releaseN(n) is acquire/release repeated N times") {
+        checkM(Gen.long(1, 100)) {
+          capacity =>
+            def acquireRelease(
+              sem: TSemaphore
+            )(acq: Long => STM[Nothing, Unit])(rel: Long => STM[Nothing, Unit]): STM[Nothing, (Long, Long)] =
+              for {
+                _            <- acq(capacity)
+                usedCapacity <- sem.available
+                _            <- rel(capacity)
+                freeCapacity <- sem.available
+              } yield (usedCapacity, freeCapacity)
+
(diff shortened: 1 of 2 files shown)


calvinlfer:
something is flaky on 2.13: 
```bash
org.scalajs.testcommon.RPCCore$ClosedException: org.scalajs.jsenv.ComJSEnv$ComClosedException: JSCom has been closed
```

A re-run might fix it

adamgfraser:
Yes, it appears to be some kind of CI issue. We are investigating but in the meantime I am rerunning.

docs/datatypes/tsemaphore.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+## Acquire a permit
+Acquiring a permit reduces the number of remaining permits that the `TSemaphore` contains. Acquiring a permit is done 
+when a user wants to access a common shared resource:
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val tSemaphoreAcq: UIO[TSemaphore] = (for {
```
adamgfraser: This might be a little more readable if you pulled the transaction and committing it into separate expressions so you didn't have the awkward `(for` syntax:

```scala
val transaction = ???
transaction.commit
```

docs/datatypes/tsemaphore.md
```diff
@@ -0,0 +1,99 @@
...
+  tSem <- TSemaphore.make(2L)
+  _    <- tSem.acquire
+} yield tSem).commit
+```
+Note that if you try to acquire a permit when there are no more remaining permits in the semaphore then execution will be blocked until a permit is ready to be acquired.
```
adamgfraser: I know it is always the case but maybe emphasize that semantic blocking doesn't actually block any threads and won't loop again until one of the the number of permits changes.

docs/datatypes/tsemaphore.md
```diff
@@ -0,0 +1,99 @@
...
+
+val tSemaphoreWithoutPermit: IO[Nothing, Unit] = 
+  (for {
+    sem <- TSemaphore.make(1L)
+    _   <- sem.acquire
+    a   <- yourSTMAction
+    _   <- sem.release
+  } yield a).commit
+```
+You can simply use `withPermit` instead:
```
adamgfraser: Maybe emphasize more that outside of advanced use cases this is generally best practice because it makes sure that we release the permits when we are done.

docs/datatypes/tsemaphore.md
```diff
@@ -0,0 +1,99 @@
...
+## Execute an arbitrary STM action with automatic acquire and release
+You can choose to execute any arbitrary STM action that requires acquiring and releasing permit on TSemaphore as part
+of the same transaction. Rather than doing:
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+def yourSTMAction: STM[Nothing, Unit] = STM.unit
+
+val tSemaphoreWithoutPermit: IO[Nothing, Unit] =
```
adamgfraser: `UIO`?

docs/datatypes/tsemaphore.md
```diff
@@ -0,0 +1,99 @@
...
+    a   <- yourSTMAction
+    _   <- sem.release
+  } yield a).commit
+```
+You can simply use `withPermit` instead:
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val tSemaphoreWithPermit: IO[Nothing, Unit] = (for {
```
adamgfraser: `UIO`?

core-tests/shared/src/test/scala/zio/stm/TSemaphoreSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+            sem    <- TSemaphore.make(expected)
+            _      <- sem.withPermit(STM.unit)
+            actual <- sem.available
+          } yield assert(actual)(equalTo(expected))
+        }
+      }
+    )
+  )
+
+  private def repeat[E, A](stm: STM[E, A])(n: Long): STM[E, A] = n match {
```
adamgfraser: Not as part of this PR but we should think about how to implement functionality like this in `STM` itself. In ZIO it is powered by `Schedule` which we have issues with using because it can execute arbitrary effects but maybe we could have a `TSchedule` structure more similar to the old encoding of `Schedule` that we could use for logic like this.

docs/datatypes/tsemaphore.md
calvinlfer: done :-)

docs/datatypes/tsemaphore.md
calvinlfer: done :-), have added a note at the end of the code sample

docs/datatypes/tsemaphore.md
calvinlfer: oops, changed this all to STM so this no longer applies

docs/datatypes/tsemaphore.md
calvinlfer: oops, changed this all to STM so this no longer applies

calvinlfer:
@adamgfraser thanks for the quick turnaround, we really appreciate it 😄

Issue #3011, opened 2020-03-03
Can't combine Duration.Finite with a Duration


CptnKirk:
Regression from RC17 -> RC18

It's not obvious how to combine a `Duration.Finite` and a `Duration` from `.timed`.

As a result a `mapAccum(Duration.Zero){...}` from RC17 breaks in RC18.

adamgfraser:
Let me take a look.

adamgfraser:
@CptnKirk I don't think there is an issue in combining `Duration.Finite` and `Duration`. You can just do it with `+` assuming you want to add them. I think the issue is that `Duration.Zero` returns a `Duration.Finite` so when you do `mapAccum(Duration.Zero)(???)` the inferred type of the state being accumulated is `Duration.Finite` instead of `Duration` and so `+` doesn't return the right type (since adding two finite duration could potentially return an infinite duration given overflow).

You can resolve by using a type ascription (e.g. `mapAccum(Duration.Zero: Duration)`) or by using a type annotation. Note that this isn't really a ZIO specific issue. You have the problem in Scala in general if you try to do something like `as.foldRight(Nil)((_ :: _)`.

We could potentially solve by having `Duration.Zero` return `Duration` instead of `Duration.Finite` but we are throwing away information and there are other cases where we want to know that the duration is finite. We could also add a smart constructor `val zero: Duration = Duration.Zero` but that still only addresses this specific case (e.g. `Duration.fromNanos` also returns a finite duration for example).

CptnKirk:
I see.  I was confused by the error message and pointer at the `+`
```
[error]  found   : zio.duration.Duration
[error]  required: zio.duration.Duration.Finite
[error]      .mapAccum(Duration.Zero){case (time, el) => (time + el._3, (el, time + el._3))}
[error]                                                        ^
```

Adding type ascription did solve my problem.  However, this isn't very ergonomic.  In fact it seems that combining Finite values results in a Duration.  This is also unexpected, as two finite values can't combine to form an infinite (only a larger finite value).  You mentioned something about an overflow, however I'd expect that this would be modeled differently.  Or you could back the scalar value of the duration implementation with a BigInt or some non-overflowable type.

I'd also expect the return values of `.timed` to be Finite, since any returning timed value must be a discrete finite value.  Having `.timed` return non-Finite Durations seems to communicate information that isn't possible. 

If timers returned Finite durations and combining Finite durations also yielded Finite durations, then we could fold/aggregate naturally without ascriptions.

[...]

PR #3032, opened 2020-03-05
Fix Regression in Ergonomics of Working with Durations

adamgfraser:
#2567 caused a regression in the ergonomics of working with durations as reported in #3011.

My analysis of the situation is that we are currently suffering from being halfway between two different encoding.

Prior to #2567 the encoding was more like a traditional algebraic data type where we hid the underlying type (`Finite` or `Infinite`) as much as possible. This generally worked well but resulted in us not being able to honor the roundtrip property that converting a Scala finite duration to a ZIO duration and back would preserve type information.

#2567 fixed that but created an issue where most of our constructors return `Finite` but because of the possibility of overflow almost all operators return just `Duration`, which causes type inference problems when, for example, accumulating a collection of durations with a fold.

This PR addresses by implementing an approach that all constructors on `Duration` should return a `Duration`, while constructors on `Finite` can return a finite duration. Thus, for most users who want the better type inference doing something like `durations.foldLeft(Duration.Zero)(_ + _)` will work like it did on RC17. Users who want to construct durations returning the more specific subtype can use the `apply` method on `Finite`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/duration/DurationSpec.scala
@@ -23,7 +23,7 @@ object DurationSpec extends ZIOBaseSpec {
         assert(Duration.fromNanos(1) * -1.0)(equalTo(Duration.Zero: Duration))
       },
       test("Its stdlib representation is correct and matches type") {
-        val duration: ScalaFiniteDuration = Duration.fromNanos(1234L).asScala
+        val duration: ScalaFiniteDuration = Duration.Finite(1234L).asScala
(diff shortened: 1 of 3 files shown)


quelgar:
There is a downside to this.

Works:

```scala
import scala.concurrent.duration._

def foo(x: FiniteDuration) = ???

foo(5.seconds)
```

Does not work:

```scala
import zio.duration._

def foo(x: scala.concurrent.duration.FiniteDuration) = ???

foo(5.seconds.asScala)
```

Can't see a way to get the best of both worlds, alas.

jdegoes:
I think we are still figuring out whether to turn `zio.duration.Duration` into just `java.time.Duration`..

PR #3033, opened 2020-03-05
ZIO Core: Implement ZLayer#memoize

adamgfraser:


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -110,6 +110,13 @@ final class ZLayer[-RIn, +E, +ROut <: Has[_]] private (
   def mapError[E1](f: E => E1): ZLayer[RIn, E1, ROut] =
     fold(ZLayer.fromFunctionManyM(f andThen ZIO.failNow), ZLayer.identity)
 
+  /**
+   * Returns a managed effect that, if evaluated, will return the lazily
+   * computed result of this layer.
+   */
+  def memoize: ZManaged[RIn, Nothing, ZLayer[Any, E, ROut]] =
```
jdegoes: Do you think we could keep `RIn` inside the `ZLayer`?

adamgfraser: I think so. Do you think we would want to make the same change to the `memoize` methods in general? I was hoping to be able to implement this in terms of `ZManaged#memoize` so we could keep all the resource management logic there but we could have a custom implementation.

jdegoes: It's just an intuition for `ZLayer`, because if we don't do that, you can no longer use `>>>` and `++` for the inner `ZLayer`, So the memoization would force you to some how supply the layer dependencies not through ZLayer but through the outer ZManaged.

adamgfraser: Yes. Okay, let me work on this.

adamgfraser: @jdegoes I refactored `ZIO#memoize` and `ZManaged#memoize` to require the environment for the inner effect instead of the outer effect. I think this creates a nice consistency across these methods, `ZLayer#memoize`, and `ZIO#once`. It is also more general than the previous signature because you can always access the environment and provide it to the effect you are memoizing to get the old signature, but you can't go the other way around. Are you good with this?

@iravid If you have a chance could you take a look at the `ZManaged#memoize` implementation and of course if you think this makes sense?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -479,30 +479,23 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-
-          val release2 = (_: Exit[_, _]) =>
-            ref.updateSome {
-              case Some((res, e)) => res.release(e).as(None)
-            }
-
-          Reservation(acquire2, release2)
-        }
-      }
+  def memoize: ZManaged[Any, Nothing, ZManaged[R, E, A]] =
```
iravid: Big improvement!

jdegoes:
Looking good!

PR #3034, opened 2020-03-05
Added Hello World working example 

tampler:


adamgfraser:
@tampler I think with recent changes to include the `ZEnv` in the default runtime again we don't need this anymore. What do you think?

tampler:
Alright, I don't monitor all those changes in real time. If this is not actual, should I close it?

adamgfraser:
I think so. See #3055. The behavior is going to be the same as in RC17.

tampler:
not actual anymore

PR #3040, opened 2020-03-06
Fixes #3013

### LINKED ISSUE #3013: [Bug] Runtime#unsafeRunToFuture is broken in RC-18
`Runtime#unsafeRunToFuture` is broken in RC-18 release as part of interruption rework in https://github.com/zio/zio/commit/681063606b157b7f98f51c664fc7e7a4c1a82918. I did a full bisect of the recent commits and that one was confirmed to be the cause.

Here's a quick reproducer:

```
import java.time.{LocalDate, LocalDateTime}

import zio.clock.Clock
import zio.console.{Console, putStrLn}
import zio.duration._
import zio.{Schedule, Task, ZIO}

object BrokenUnsafeRunToFutureReproducer extends zio.App {
  override def run(args: List[String]): ZIO[zio.ZEnv, Nothing, Int] = {
    val effect = ZIO.effectSuspendTotal(putStrLn(s"Current time is ${LocalDateTime.now}"))
      .delay(1.second)

    val toFutureAndBackToEffect = for {
      rt <- ZIO.runtime[Console with Clock]
      _ <- Task.fromFuture { _ =>
        rt.unsafeRunToFuture(effect)
      }
    } yield ()

    toFutureAndBackToEffect
      .repeat(Schedule.fixed(1.second))
      .orDie
      .as(0)
  }
}
```

Which will terminate with:

```
Fiber failed.
java.util.concurrent.ExecutionException: Boxed Exception
	at scala.concurrent.impl.Promise$.scala$concurrent$impl$Promise$$resolve(Promise.scala:99)
	at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:242)
	at scala.concurrent.Promise.complete(Promise.scala:57)
	at scala.concurrent.Promise.complete$(Promise.scala:56)
	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:104)
	at scala.concurrent.Promise.failure(Promise.scala:109)
	at scala.concurrent.Promise.failure$(Promise.scala:109)
	at scala.concurrent.impl.Promise$DefaultPromise.failure(Promise.scala:104)
	at zio.Fiber.$anonfun$toFutureWith$2(Fiber.scala:297)
	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:383)
	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:679)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.InterruptedException: Interrupted by fibers: #15
	at zio.Cause.$anonfun$squashWith$1(Cause.scala:348)
	at scala.Option.orElse(Option.scala:477)
	at zio.Cause.squashWith(Cause.scala:345)
	at zio.Cause.squashWith$(Cause.scala:343)
	at zio.Cause$Internal$Interrupt.squashWith(Cause.scala:648)
	at zio.Cause.squashWithTrace(Cause.scala:370)
	at zio.Cause.squashWithTrace$(Cause.scala:369)
	at zio.Cause$Internal$Interrupt.squashWithTrace(Cause.scala:648)
	... 6 more
Caused by: zio.Cause$FiberTrace: Fiber failed.
An interrupt was produced by #15.
No ZIO Trace available.
```

It's somewhat non-deterministic (hence the `repeat`).

jdegoes:
Fixes #3013

### DIFF
--- a/core-tests/shared/src/test/scala/zio/CancelableFutureSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/CancelableFutureSpec.scala
@@ -1,56 +1,83 @@
 package zio
 
+import zio.clock.Clock
+import zio.console._
+import zio.duration._
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
+import zio.test.environment._
 
 object CancelableFutureSpec extends ZIOBaseSpec {
+  def roundtrip[R, A](zio: RIO[R, A]): RIO[R, A] =
+    for {
+      future <- zio.toFuture
+      a      <- RIO.fromFuture(_ => future)
+    } yield a
 
-  def spec = suite("CancelableFutureSpec")(
-    testM("interrupts the underlying task on cancel 2") {
-      for {
-        p  <- Promise.make[Nothing, Unit]
-        p2 <- Promise.make[Nothing, Int]
-        f <- (p.succeed(()) *> IO.never)
-              .onInterrupt(p2.succeed(42))
-              .fork
-        _    <- p.await
-        _    <- f.interrupt
-        test <- p2.await
-      } yield assert(test)(equalTo(42))
-    } @@ nonFlaky,
-    testM("interrupts the underlying task on cancel") {
-      for {
-        p  <- Promise.make[Nothing, Unit]
-        p2 <- Promise.make[Nothing, Int]
-        f <- (p.succeed(()) *> IO.never)
-              .onInterrupt(p2.succeed(42))
-              .toFuture
-        _    <- p.await
-        _    <- ZIO.fromFuture(_ => f.cancel())
-        test <- p2.await
-      } yield assert(test)(equalTo(42))
-    } @@ nonFlaky,
-    testM("cancel returns the exit reason") {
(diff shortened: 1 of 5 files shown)


adamgfraser:
This looks good. The only thing I noticed is that if I run the original reproduction example from #3013 and then terminate the program the message continues to get displayed to the console every second. I would have expected the effect to be terminated at that point.

jdegoes:
@adamgfraser I added a test that's closer and it still passes.

However, it originally failed when I used `nonFlaky` with the original test, but that's due to the fact that the original test runs forever, and even if you add a timeout, if you repeat with `nonFlaky`, it will still timeout.

If you can provide a failing testcase that doesn't timeout for legit reason, I'll take a closer look!

jdegoes:
I broke it!

PR #3042, opened 2020-03-06
Add a reference towards `foreach_` in `foreach` doc

fanf:
Add a comment similar to the one for `foreachPar` for `foreach_` in relevant `foreach` scala doc. 
The goal is to make `foreach_` more discoverable, its intent clear, and make it clear that the `_` doesn't mean that you should not use it. 

This is especially important since most people comming from an imperative background will likelly use patterns where their loops doesn't return anything usefull, and so they should default to `foreach_` for noticeable performance improvement.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2301,6 +2301,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * returns the results in a new `List[B]`.
    *
    * For a parallel version of this method, see `foreachPar`.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2341,6 +2343,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * returns the results in a new `Chunk[B]`.
    *
    * For a parallel version of this method, see `foreachPar`.
+   * If you do not need the results, see `foreach_` for a more efficient implementation.
```
adamgfraser: Don't think we need here since this is the `foreach_` method.

PR #3043, opened 2020-03-06
Delete Now Variants

adamgfraser:
Deletes all now variants except `succeedNow`. Adds public versions of `succeedNow` variants in `zio.internal` and `zio.stream.internal` packages.


adamgfraser:
@jdegoes This is ready to go.

jdegoes:
Later we'll have to update the benchmarks to use lazy `fail` (etc.) for the other data types, otherwise it's not 1-to-1.

PR #3045, opened 2020-03-06
ZIO Stream: Simplify Signature of ZSink#succeed

adamgfraser:
Signature going from:

```scala
def succeed[A, B](b: => B): ZSink[Any, Nothing, A, A, B]
```

to:

```scala
def succeed[A](a: => A): ZSink[Any, Nothing, Nothing, Any, A]
```

In other words, the sink can accept any input and will never emit any leftovers but only the value provided. This is slightly different than the current semantic, which seems to be it will accept inputs of a specified type and return those same inputs as leftovers but every test passes with no changes so the leftovers don't seem to be something we were using and this seems to be a more specific contract and allows us to remove the totally unbound type parameter in the current implementation, allowing us to avoid manually specifying type parameters in several locations.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1835,13 +1835,13 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
   /**
    * Creates a single-value sink from a value.
    */
-  def succeed[A, B](b: => B): ZSink[Any, Nothing, A, A, B] =
-    new SinkPure[Nothing, A, A, B] {
-      type State = Chunk[A]
-      val initialPure                  = Chunk.empty
-      def stepPure(state: State, a: A) = state ++ Chunk(a)
-      def extractPure(state: State)    = Right((b, state))
-      def cont(state: State)           = false
+  def succeed[A](a: => A): ZSink[Any, Nothing, Nothing, Any, A] =
+    new SinkPure[Nothing, Nothing, Any, A] {
+      type State = Chunk[Any]
+      val initialPure                    = Chunk.empty
+      def stepPure(state: State, a: Any) = state ++ Chunk(a)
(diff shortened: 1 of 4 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1835,13 +1835,13 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
...
-      val initialPure                  = Chunk.empty
-      def stepPure(state: State, a: A) = state ++ Chunk(a)
-      def extractPure(state: State)    = Right((b, state))
-      def cont(state: State)           = false
+  def succeed[A](a: => A): ZSink[Any, Nothing, Nothing, Any, A] =
+    new SinkPure[Nothing, Nothing, Any, A] {
+      type State = Chunk[Any]
+      val initialPure                    = Chunk.empty
+      def stepPure(state: State, a: Any) = state ++ Chunk(a)
+      def extractPure(state: State)      = Right((a, initialPure))
```
iravid: Because we don’t have laws, I can’t say right away if we are relying on this remainder-preserving behavior. 

This is basically a safety net for a caller not observing cont = false. I recall that there are several places where we don’t check cont between calling initial and the first step; those might lose elements when ZSink.succeed is composed through flatMap with another sink.

adamgfraser: Okay. Shall we close?

iravid: If you have the cycles, we can go through combinators and ZSink callers and check if we do any unwanted steps.

But that’s a pretty big chunk of effort that’s probably better put into the new encoding :-)

adamgfraser: That's fine. There is plenty of other work to do. Just noticed this when I was cleaning by the `Now` variants.

PR #3051, opened 2020-03-07
Rename ZLayer.succeed to ZLayer.make

mikail-khan:


test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
```diff
@@ -28,6 +28,7 @@ class ZTestJUnitRunner(klass: Class[_]) extends Runner with Filterable with Boot
 
   private lazy val spec: AbstractRunnableSpec = {
     klass
+      .getDeclaredConstructor()
```
mikail-khan: Not sure why this was not failing the CI builds because it tests fail to compile on my machine locally w/o this change.

adamgfraser:
One other thing to consider is just that this is another change. I'm personally somewhat neutral on this just because in almost every other context in ZIO `make` implies something else, basically that it would return the type wrapped in an effect constructor. We've just spent the last few days teaching people how to use layers and the existing name. It is still early so not the end of the world to change it but feels like if we are going to change we collectively think the new name is a significant improvement.

mikail-khan:
> One other thing to consider is just that this is another change. I'm personally somewhat neutral on this just because in almost every other context in ZIO `make` implies something else, basically that it would return the type wrapped in an effect constructor. We've just spent the last few days teaching people how to use layers and the existing name. It is still early so not the end of the world to change it but feels like if we are going to change we collectively think the new name is a significant improvement.

mikail-khan:
> One other thing to consider is just that this is another change. I'm personally somewhat neutral on this just because in almost every other context in ZIO `make` implies something else, basically that it would return the type wrapped in an effect constructor. We've just spent the last few days teaching people how to use layers and the existing name. It is still early so not the end of the world to change it but feels like if we are going to change we collectively think the new name is a significant improvement.

For me, `ZLayer.succeed` implied that there exists `ZLayer.fail` analogous to `ZIO.succeed` and `ZIO.fail` and that isn't the case.

adamgfraser:
Well I think that actually points in the direction of `succeed` being an appropriate name. `ZLayer.fail` is a perfectly sensible combinator. We haven't written it to date because it hasn't proved necessary but it is a perfectly logical combinator. Layers can fail and we could create a layer that fails, as well as recovering from failures in layers (see `ZLayer#fold`).

mikail-khan:
> Well I think that actually points in the direction of `succeed` being an appropriate name. `ZLayer.fail` is a perfectly sensible combinator. We haven't written it to date because it hasn't proved necessary but it is a perfectly logical combinator. Layers can fail and we could create a layer that fails, as well as recovering from failures in layers (see `ZLayer#fold`).

In that case, the name as is makes sense. I can see how construction of a layer could fail using one of the `fromFunction...` combinators.

jdegoes:
Agreed on the pain of a rename at this stage, and given that `fail` could make sense too, maybe that's an argument to keep it `succeed`...

adamgfraser:
@mikail-khan I think at this point we should keep the existing naming. Are you comfortable closing this?

mikail-khan:
> @mikail-khan I think at this point we should keep the existing naming. Are you comfortable closing this?

sure @adamgfraser :+1:

PR #3053, opened 2020-03-07
Weak roots

jdegoes:
1. Switch to use weak roots
2. Add `ZIO.adopt`
3. Generalize some callbacks so you can return anything you want, without being forced to return `Unit`
4. Fix a flaky test
5. Polish some scaladoc
6. Rename `squashWithTrace` to `squashTraceWith`

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1838,7 +1838,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   /**
    * Returns an effect that adopts the specified fiber as a child of the fiber
    * running this effect. Note that adoption will succeed only if the specified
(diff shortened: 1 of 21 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1824,6 +1835,19 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def accessM[R]: ZIO.AccessMPartiallyApplied[R] =
     new ZIO.AccessMPartiallyApplied[R]
 
+  /**
+   * Returns an effect that adopts the specified fiber as a child of the fiber
+   * running this effect. Note that adoption will succeed only if the specified
+   * fiber is not a child fo any other fiber.
```
AdamPaynter: Typo: "fo"

PR #3054, opened 2020-03-07
Add Higher Arity Variants of ZLayer and Has Combinators

adamgfraser:
Implements higher arity variants of all relevant `ZLayer` and `Has` combinators up to arity 22. Removes unused `upcast` methods. Removes remaining documentation referring to restriction to monomorphic services.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -13,3 +13,8 @@ spaces {
 optIn.annotationNewlines = true
 
 rewrite.rules = [SortImports, RedundantBraces]
+
+project.excludeFilters = [
+  "core/shared/src/main/scala/zio/Has.scala",
+  "core/shared/src/main/scala/zio/ZLayer.scala"
+]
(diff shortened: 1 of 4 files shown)


jdegoes:
😱 😆

adamgfraser:
It looks like we have an issue on Dotty here. It looks like compilation on Dotty is timing out. I have retried in CI and can also reproduce the issue locally. Compilation doesn't fail, it looks like it just doesn't terminate or at least takes a really long time. I will raise an issue with the Dotty team. We may need to put the higher arity methods in a separate trait anyway so we can put them in a version specific folder for now.

jdegoes:
@adamgfraser I think we could just drop `MustNotHave[A, B]`. It's more like a warning to use `update` instead of `add` if it's statically known the environment already contains the service.

adamgfraser:
@jdegoes Sounds good. Will do.

PR #3055, opened 2020-03-07
Provide ZEnv in Default Runtime

adamgfraser:
With the changes we made to expose the services for the default environment types we can bundle these services together and include them in the default runtime to provide the same experience as before. I think this will make it significantly easier to work with mixed applications and in particular it sounds like something like this is necessary for `cats-interop`.


jdegoes:
Looks good! The `zio.App` don't need to `provideLayer` anymore

PR #3058, opened 2020-03-08
ZIO Core: Derive Tags for Polymorphic Services

adamgfraser:
Adds support for deriving tags for polymorphic services based on tags for the concrete types. This should reduce the amount of implicit evidence required when working with polymorphic services.


core/shared/src/main/scala-2.x/zio/Tagged.scala
```diff
@@ -26,6 +28,664 @@ final case class Tagged[A](tag: TaggedType[A]) {
...
 
 object Tagged {
-  implicit def tagged[A](implicit tag: TaggedType[A]): Tagged[A] =
-    Tagged(tag)
+
+  implicit def tagged[A: TaggedType](implicit a: TaggedType[A]): Tagged[A] =
+    Tagged(a)
+
+  @silent("is never used")
+  implicit def taggedF[F[_], A0](implicit tag: TaggedTypeF[F], a0: Tagged[A0]): Tagged[F[A0]] = {
```
neko-kai: @adamgfraser 
All these implicits are entirely redundant, they just repackage the general-purpose combining logic, that's already implemented for all kinds by a macro, for specific cases, which is nonsensical. I think the correct solution here is to alias `type TaggedType[A] = Tag[A]`, then every single implicit in `TaggedType` can be removed.

core/shared/src/main/scala-2.x/zio/VersionSpecific.scala
```diff
@@ -1,12 +1,36 @@
...
 private[zio] trait VersionSpecific {
 
   type TaggedType[A] = Tag[A]
-  type TagType       = LightTypeTag
+  type TaggedTypeF[F[_]] = TagK[F]
+  type TaggedTypeF2[F[_, _]] = TagKK[F]
+  type TaggedTypeF3[F[_, _, _]] = TagK3[F]
+  type TaggedTypeF4[F[_, _, _, _]] = HKTag[ { type Arg[A0, A1, A2, A3] = F[A0, A1, A2, A3] } ]
+  type TaggedTypeF5[F[_, _, _, _, _]] = HKTag[ { type Arg[A0, A1, A2, A3, A4] = F[A0, A1, A2, A3, A4] } ]
+  type TaggedTypeF6[F[_, _, _, _, _, _]] = HKTag[ { type Arg[A0, A1, A2, A3, A4, A5] = F[A0, A1, A2, A3, A4, A5] } ]
```
neko-kai: I think it's not necessary to avoid syntax sugar here, the internals of HKTag aren't interesting for someone reading zio code, so you could write them in a shorter way:
```scala
type TaggedTypeF5[F[_, _, _, _, _]] = Tag.auto.T[F]
type TaggedTypeF6[F[_, _, _, _, _, _]] = Tag.auto.T[F]
```

core/shared/src/main/scala-2.x/zio/Tagged.scala
adamgfraser: @neko-kai You are 100% right we are just repackaging the existing derivation functionality. I think this may be necessary given need to support Dotty as well but open to other suggestions. I think the challenge is that we want the user to be able to work with just the `Tagged` abstraction (whatever that is for the specific version) to abstract over version specific differences.

I tried doing `type Tagged[A] = TaggedType[A]` to just get the existing implicit derivation machinery working for us. The challenge I ran into there is that when we put these values in the `Has` map we need to use the definition of `equals` and `hashCode` from the `LightTypeTag` instead of the `Tag`, so I think we still need some kind of wrapper and then to re-export the implicit derivation as done here.

Again, very much agree with the sentiment and open to other solutions but that is how I got to the current approach.

core/shared/src/main/scala-2.x/zio/VersionSpecific.scala
adamgfraser: Thanks! Will do.

core/shared/src/main/scala-2.x/zio/Tagged.scala
neko-kai: @adamgfraser 
> I think this may be necessary given need to support Dotty as well but open to other suggestions.

This won't help current support, because you can't combine ClassTags anyway, and there's no way to express a higher-kinded ClassTag. For dotty version of Tag, Tag will be kind-polymorphic `Tag[_ <: AnyKind]`, so even the aliases for different kinds will become redundant. 

> The challenge I ran into there is that when we put these values in the Has map we need to use the definition of equals and hashCode from the LightTypeTag instead of the Tag

I don't understand the current distinction between `Tagged` & `TaggedType` - since both of them are parameterized, and both of them are searched for implicitly. But you could define a third alias `type TagDataType = LightTypeTag`/`type TagDataType = Class[_]` and use that as a key in `Has`. (You can also add `equals`/`hashCode` to izumi-reflect Tag too)

core-tests/shared/src/test/scala/zio/TaggedSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+object TaggedSpec extends ZIOBaseSpec {
+
+  def spec = suite("TaggedSpec")(
+    testM("tags can be derived for polymorphic services") {
+      val result = typeCheck {
+        """
+            trait Producer[R, K, V]
+
+            def test[R: Tagged, K: Tagged, V: Tagged]: Boolean = {
+              val _ = implicitly[Tagged[Producer[R, K, V]]]
```
iravid: 😍

core/shared/src/main/scala-2.x/zio/Tagged.scala
neko-kai: Regarding Dotty support, @ivan71kmayshan27 contributed a macro for constructing `LightTypeTag` from Dotty's quoted.Type in https://github.com/7mind/dotty-typetag-research, so the basic infrastructure is in place and it's getting closer.

adamgfraser: Can we define `equals` and `hashCode` for `Tag` to delegate to the `LightTypeTag`? I think that would be ideal and allow us to eliminate all of this and just have `Tagged[A] = `Tag[A]`.

adamgfraser: @neko-kai `Tagged` is really just to wrap the underlying type so we can provide our own implementation of `equals` and `hashCode`, getting around issues with it not being consistent for `ClassTag` and right now making sure we delegate to the `LightTypeTag`.

neko-kai: @adamgfraser Sure, could you make a PR?

adamgfraser: @neko-kai Sure!

neko-kai: @adamgfraser 
> Tagged is really just to wrap the underlying type so we can provide our own implementation of equals and hashCode, getting around issues with it not being consistent for ClassTag and right now making sure we delegate to the LightTypeTag.

It was necessary before, because equality was inconsistent for `scala-reflect`'s `TypeTag`, but it's no longer necessary, for Dotty the JVM guarantees that instances of `java.lang.Class[_]` are unique singletons and can be compared referentially (`equals` & `hashCode` implemented with object identity will work correctly because Classes are unique). I believe it's not necessary to compare `.toString`'s of ClassTag, instead of comparing the underlying `Class[_]` by reference. I believe there is no longer any reason to have `Tagged`, instead this will work just fine:

```scala
// dotty
type Tagged[A] = ClassTag[A]
private[zio] type TaggedType = Class[_]

def tagValue(tagged: Tagged[A]): TaggedType = tagged.runtimeClass
```

&

```scala
// scala 2
type Tagged[A] = Tag[A]
private[zio] type TaggedType = LightTypeTag

def tagValue(tagged: Tagged[A]): TaggedType = tagged.tag
```

where

```scala
final class Has[A] private (
  private val map: Map[TaggedType, scala.Any],
  private var cache: Map[TaggedType, scala.Any] = Map()
)
```

neko-kai: @adamgfraser Note, using LightTypeTag as the key directly will also reduce the storage overhead a bit, because `Tag` also carries a `val closestClass: Class[_]` which i believe is redundant for the purpose (but I'm not sure, given that you asked how to get a Class from LTT, if you use it somewhere, it's fine to keep Tag there)

adamgfraser: @neko-kai Thank you! Will try that now.

core/shared/src/main/scala-2.x/zio/Tagged.scala
```diff
@@ -1,691 +0,0 @@
-/*
```
neko-kai: Nice! 👍

core/shared/src/main/scala-2.x/zio/VersionSpecific.scala
```diff
@@ -1,42 +1,19 @@
...
+import izreflect.fundamentals.reflection.Tags.Tag
 import izreflect.fundamentals.reflection.macrortti.{ LightTypeTag, LightTypeTagRef }
 
 private[zio] trait VersionSpecific {
 
-  type TaggedType[A] = Tag[A]
-  type TaggedTypeF[F[_]] = TagK[F]
-  type TaggedTypeF2[F[_, _]] = TagKK[F]
-  type TaggedTypeF3[F[_, _, _]] = TagK3[F]
-  type TaggedTypeF4[F[_, _, _, _]] = HKTag[ { type Arg[A0, A1, A2, A3] = F[A0, A1, A2, A3] } ]
```
neko-kai: @adamgfraser There should probably be some way for the user to summon these things when they work with highly polymorphic code.
I think we should just alias the tags in izumi-reflect here:

```scala
type TagK[F[_]] = Tags.TagK[F]
type TagKK[F[_, _]] = Tags.TagKK[F]
type TagK3[F[_, _, _]] = Tags.TagK3[F]
```

And maybe synchronize the naming too and just switch to `Tag` everywhere:

```scala
@deprecated("Use zio.Tag[A]")
type Tagged[A] = Tags.Tag[A]
type Tag[A] = Tags.Tag[A]
```

```scala
@deprecated("Use zio.Tag[A]")
type Tagged[A] = ClassTag[A]
type Tag[A] = ClassTag[A]
```

adamgfraser: @neko-kai Added back the type aliases. I left it as `Tagged` to avoid breaking one more thing and kept the `TaggedF` because I think it is slightly more accessible to users without a functional programming background. Just my personal opinion and happy to change if others have a different point of view but figured would get the code out there first.

neko-kai: @adamgfraser I think users without background would not have an intuition regarding `F` vs `K` - neither actually tells anything to someone unfamiliar, and if they do have cats background they might find `K` more intuitive since `K` suffix in cats means 'higher-kinded' (`MonoidK`, `SemigroupK`, `FunctorK` etc.)

neko-kai:
@adamgfraser @jdegoes Does anything stop this from being merged currently?

adamgfraser:
@neko-kai I don't think so.

PR #3059, opened 2020-03-08
Improve Documentation for ZLayer#fromServicesMany

### LINKED ISSUE #3038: ZIO Core: ZLayer.fromServicesMany should be removed
It is confusing what the difference is between ZLayer.fromServices and ZLayer.fromServicesMany is and when you should use one or the other.

Would be good to consider removing ZLayer.fromServicesMany entirely.

adamgfraser:
Resolves #3038.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -900,7 +900,7 @@ object ZLayer {
     fromServiceManyM[A, Any, Nothing, B](a => ZIO.succeedNow(f(a)))
 
   /**
-   * Constructs a layer that purely depends on the specified service, which
+   * Constructs a layer that purely depends on the specified services, which
    * must return one or more services. For the more common variant that returns
    * a single service see `fromServices`.
    */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -893,14 +893,16 @@ object ZLayer {
...
+   * must return one or more services. For the more common variant that returns
+   * a single service see `fromService`.
    */
   def fromServiceMany[A: Tagged, B <: Has[_]](f: A => B): ZLayer[Has[A], Nothing, B] =
     fromServiceManyM[A, Any, Nothing, B](a => ZIO.succeedNow(f(a)))
 
   /**
-   * Constructs a layer that purely depends on the specified services, which
-   * must return one or more services.
+   * Constructs a layer that purely depends on the specified service, which
```
jdegoes: I would use plural for the "services" (n > 1) cases.

adamgfraser: Good catch! Fixed now.

PR #3067, opened 2020-03-08
Implement ZRef#Derived

adamgfraser:
Implements `ZRef#Derived` to support transforming the type of the `Ref` while maintaining atomicity guarantees.

### DIFF
--- a/core/shared/src/main/scala/zio/ZRef.scala
+++ b/core/shared/src/main/scala/zio/ZRef.scala
@@ -15,182 +15,107 @@
  */
 
 package zio
+
 import java.util.concurrent.atomic.AtomicReference
 
-/**
- * A mutable atomic reference for the `IO` monad. This is the `IO` equivalent of
- * a volatile `var`, augmented with atomic operations, which make it useful as a
- * reasonably efficient (if low-level) concurrency primitive.
- *
- * {{{
- * for {
- *   ref <- Ref.make(2)
- *   v   <- ref.update(_ + 3)
- *   _   <- console.putStrLn("Value = " + v) // Value = 5
- * } yield ()
- * }}}
- *
- * NOTE: While `Ref` provides the functional equivalent of a mutable reference,
- * the value inside the `Ref` should be immutable. For performance reasons
- * `Ref` is implemented in terms of compare and swap operations rather than
- * synchronization. These operations are not safe for mutable values that do
- * not support concurrent access.
- */
-trait ZRef[+EA, +EB, -A, +B] extends Serializable { self =>
+sealed trait ZRef[+EA, +EB, -A, +B] extends Serializable { self =>
 
-  /**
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -15,182 +15,107 @@
...
+  final def collect[C](pf: PartialFunction[B, C]): ZRef[EA, Option[EB], A, C] =
+    fold(identity, Some(_), Right(_), pf.lift(_).fold[Either[Option[EB], C]](Left(None))(Right(_)))
+
+  final def contramap[C](f: C => A): ZRef[EA, EB, C, B] =
+    contramapEither(c => Right(f(c)))
+
+  final def contramapEither[EC >: EA, C](f: C => Either[EC, A]): ZRef[EC, EB, C, B] =
+    dimapEither(f, Right(_))
+
+  final def dimap[C, D](f: C => A, g: B => D): ZRef[EA, EB, C, D] =
```
jdegoes: `mapBoth` / `mapBothError`? Maybe a little strained because one is contramapping but then again, dimap doesn't draw any such distinction.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -15,182 +15,107 @@
...
+  final def map[C](f: B => C): ZRef[EA, EB, A, C] =
+    mapEither(b => Right(f(b)))
 
+  final def mapEither[EC >: EB, C](f: B => Either[EC, C]): ZRef[EA, EC, A, C] =
+    dimapEither(Right(_), f)
+
+  final def unifyError[E](ea: EA => E, eb: EB => E): ZRef[E, E, A, B] =
+    dimapError(ea, eb)
+
+  final def unifyValue[C](ca: C => A, bc: B => C): ZRef[EA, EB, C, C] =
```
jdegoes: All these look like a good start. I'd also add:

```scala
readOnly
writeOnly
```

So it's easy to hand off a "read only" or "write only" view downstream.

core/shared/src/main/scala/zio/ZRef.scala
adamgfraser: We can. I thought `dimap` was the name we used for this function, for example on `Schedule` and `Sink`?

core/shared/src/main/scala/zio/ZRef.scala
adamgfraser: Done. Let me know if there are any others you can think of. It is very easy to add them.

core/shared/src/main/scala/zio/ZRef.scala
jdegoes: Ah, ok, in that case, let's stick with `dimap`.

PR #3080, opened 2020-03-09
Use Not Combinator on Dotty

### LINKED ISSUE #3068: Trick used to support implicit search negation cannot be used with Dotty
https://github.com/zio/zio/tree/master/core/shared/src/main/scala-2.12%2B/zio contains a bunch of definitions all following the same pattern:
```scala
implicit def neq[A, B]: A =!= B = new =!=[A, B] {}

@implicitAmbiguous("Cannot prove that ${A} =!= ${A}")
implicit def neqAmbig1[A]: A =!= A = unexpected
implicit def neqAmbig2[A]: A =!= A = unexpected
```

This trick is intentionally not supported by dotty, see part 4 of http://dotty.epfl.ch/docs/reference/changed-features/implicit-resolution.html, instead one should use [`scala.implicits.Not`](https://github.com/lampepfl/dotty/blob/master/library/src/scala/implicits/Not.scala).

While the code as-is will compile with dotty, it will not have the desired behavior and in fact might lead to blowing up compile times as experienced in https://github.com/lampepfl/dotty/issues/8463

adamgfraser:
Resolves #3068.


adamgfraser:
Copying @smarter.

core/shared/src/main/scala-dotty/zio/NeedsEnv.scala
```diff
@@ -16,13 +16,22 @@
...
+ * A value of type `NeedsEnv[R]` provides implicit evidence that an effect with
+ * environment type `R` needs an environment, that is, that `R` is not equal to
+ * `Any`.
+ */
+@implicitNotFound(
+  "This operation assumes that your effect requires an environment. " +
+    "However, your effect has Any for the environment type, which means it " +
+    "has no requirement, so there is no need to provide the environment."
+)
+sealed trait NeedsEnv[+R]
```
smarter: This trait + object + implicit def combo looks pretty involved, instead I think you could do:
```scala
type NeedsEnv[+R] = Not[R =:= Any]
```
(this can be at the top-level since dotty supports top-level definitions, or wrapped in a package object)

adamgfraser: @smarter If we do that is there a way to provide a custom error message when the implicit is not found like we do today? I think that is quite important.

smarter: `@implicitNotFound` works on a type alias: https://scastie.scala-lang.org/DSpKypbqRT6mXAjM7ly8YA (hmm, looks like we should disable the automatic import suggestions when implicitNotFound is used).

adamgfraser: Okay great!

adamgfraser: So playing with this more I don't know if it is going to work because of variance. We need the variance on `CanFail` and `NeedsEnv` so they work with all of our data types and methods that use variance extensively. But if it is just a type alias we need to have the same variance as `Not` which is invariant.

smarter: Can you give an example where you need it to be variant?

PR #3081, opened 2020-03-10
Implement ZIO#unrefine

### LINKED ISSUE #3063: Combinators to recover from defects
I have an application which is a client-facing HTTP API. This application communicates with an external service which may occasionally be unreachable.

Whenever this happens and kill the fiber with `IO.die(ServiceFooNotAvailable`: I don't want this error to appear in my "domain errors" since my business logic can't do anything about it.

I do need however to translate this error to something intelligible for my HTTP clients, so at the "end of the world" I want to transform this defect into a failure and deal with it.

The best way I could find (and discussing on Discord it seems I didn't miss more elegant ways to do it) is something like:

```scala
def myMethod: IO[String, Int] = IO.die(ServiceFooNotAvailable)

// ... later ...

myMethod.catchSomeCause {
  case Cause.Traced(Cause.Die(_: ServiceFooNotAvailable), _) => IO.succeed(42)
}
```

or (probably better)

```scala
myMethod.catchSomeCause {
  case c if c.dieOption.exists(_.isInstanceOf[ServiceFooNotAvailable]) => IO.succeed(42)
}
```

Long story short, it would be nicer to have a direct combinator to recover from a dead fiber. Something along the lines of

```scala
def catchSomeDieCause[R1 <: R, E1 >: E, A1 >: A](pf: PartialFunction[Throwable, ZIO[R1, E1, A1]]): ZIO[R1, E1, A1]
```

adamgfraser:
Resolves #3063. Implements inverse of `refineOrDie` variants:

```scala
final def unrefine[E1 >: E](pf: PartialFunction[Throwable, E1]): ZIO[R, E1, A]

final def unrefineTo[E1 >: E: ClassTag]: ZIO[R, E1, A]

final def unrefineWith[E1](pf: PartialFunction[Throwable, E1])(f: E => E1): ZIO[R, E1, A]
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2769,6 +2769,53 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(result)(isNone)
       }
     ),
+    suite("unrefine")(
+      testM("converts some fiber failures into errors") {
+        val s    = "division by zero"
+        val zio1 = ZIO.die(new IllegalArgumentException(s))
+        val zio2 = zio1.unrefine { case e: IllegalArgumentException => e.getMessage }
+        assertM(zio2.run)(fails(equalTo(s)))
+      },
+      testM("leaves the rest") {
+        val t    = new IllegalArgumentException("division by zero")
+        val zio1 = ZIO.die(t)
+        val zio2 = zio1.unrefine { case e: NumberFormatException => e.getMessage }
+        assertM(zio2.run)(dies(equalTo(t)))
+      }
+    ),
+    suite("unrefineTo")(
+      testM("converts some fiber failures into errors") {
+        val t    = new IllegalArgumentException("division by zero")
+        val zio1 = ZIO.die(t)
+        val zio2 = zio1.unrefineTo[IllegalArgumentException]
+        assertM(zio2.run)(fails(equalTo(t)))
+      },
+      testM("leaves the rest") {
+        val t    = new IllegalArgumentException("division by zero")
+        val zio1 = ZIO.die(t)
+        val zio2 = zio1.unrefineTo[NumberFormatException]
+        assertM(zio2.run)(dies(equalTo(t)))
+      }
+    ),
+    suite("unrefineWith")(
+      testM("converts some fiber failures into errors") {
+        val s    = "division by zero"
+        val zio1 = ZIO.die(new IllegalArgumentException(s))
+        val zio2 = zio1.unrefineWith { case e: IllegalArgumentException => Option(e.getMessage) }(_ => None)
+        assertM(zio2.run)(fails(isSome(equalTo(s))))
+      },
+      testM("leaves the rest") {
+        val t    = new IllegalArgumentException("division by zero")
+        val zio1 = ZIO.die(t)
+        val zio2 = zio1.unrefineWith { case e: NumberFormatException => Option(e.getMessage) }(_ => None)
+        assertM(zio2.run)(dies(equalTo(t)))
+      },
+      testM("uses the specified function to convert the `E` into an `E1`") {
+        val zio1 = ZIO.fail("fail")
+        val zio2 = zio1.unrefineWith { case e: IllegalArgumentException => Option(e.getMessage) }(_ => None)
+        assertM(zio2.run)(fails(isNone))
+      }
+    ),
     suite("unsandbox")(
       testM("unwraps exception") {
         val failure: IO[Cause[Exception], String] = IO.fail(fail(new Exception("fail")))
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -410,7 +410,7 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
       }
       .reverse
 
-  private def find[Z](f: PartialFunction[Cause[E], Z]): Option[Z] = {
+  final def find[Z](f: PartialFunction[Cause[E], Z]): Option[Z] = {
     @tailrec
     def loop(cause: Cause[E], stack: List[Cause[E]]): Option[Z] =
       f.lift(cause) match {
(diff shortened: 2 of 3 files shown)


adamgfraser:
Copying @gabro.

gabro:
That was fast, thanks for paging me!

I’m a bit confused by the names: refine means going from Throwable to E, right?

If so I would expect “unrefine” to go from E to Throwable.

This operation seems to be the opposite of “die” hence my name proposal in the issue to go with “revive” or something along those lines. That’s justified by my understanding is that this operation reverts the death of a fiber, but maybe I’m getting this wrong?

What do you think?

gabro:
Ah, another thing: this seems to work only going from a defect to a failure.

I would have also expected to be able to go from a defect to a success. Could that be fit in by returning a ZIO directly? Or do you think there are more appropriate combinators for doing that?

regiskuckaertz:
rescue 😅 I think it should be symmetric to refine and so I would avoid “unrefining” to a success.

regiskuckaertz:
A “law” could be io.refine(f).unrefine(f-1) == io

gabro:
Ok, but then shouldn't the resulting error be a Throwable in order to preserve the symmetry?

regiskuckaertz:
Refine narrows the error channel and everything that doesn’t fit is turned into a Throwable that goes in the (invisible) defect channel. Unrefine broadens the error channel and embeds some defects into it. In truth there is not broadening/narrowing as E and E1 are unrelated, but that’s the idea.

If you look at the types next to each other it becomes clearer:
```scala
final def unrefineWith[E1](pf: PartialFunction[Throwable, E1])(f: E => E1): ZIO[R, E1, A]
final def refineOrDieWith[E1](pf: PartialFunction[E, E1])(f: E => Throwable)(implicit ev: CanFail[E]): ZIO[R, E1, A]
```

adamgfraser:
I don't have a particularly strong view on the naming. I wanted to get the code out there and then we could figure that out. It does "unrefine" the error type in some sense as @regiskuckaertz described because you are mapping both `E` values and some `Throwable` values into `E1` so in some sense `E1` is "bigger" than `E`. That being said, there is nothing that says the inverse operation has to have the same name.

As far as going from a die throwable directly to success it is a perfectly well defined operation and can be implemented either in terms of `unrefineWith` or `catchSomeCause`. I think the existing combinators are nice because of the symmetry with `refineOrDie` but we could potentially have that as well. I think our biggest challenge would just be coming up with a name. 😂

gabro:
Hehe, as always naming is the most difficult thing.

My biggest gripe with unrefine is this:

```scala
def refineOrDie[E1](pf: PartialFunction[E, E1])(implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZIO[R, E1, A]
def unrefine[E1 >: E](pf: PartialFunction[Throwable, E1]): ZIO[R, E1, A]
```

both transformations go from `Throwable` to something more refined (`E1`) so I struggle to see how one is the inverse of the other.

> As far as going from a die throwable directly to success it is a perfectly well defined operation and can be implemented either in terms of `unrefineWith` or `catchSomeCause`

@adamgfraser how would one use `unrefineWith` to recover from a death fiber to a success? I see it takes a transformation from `Throwable` to `E1` and from `E1` to `E`, so I can never really return a `IO.success`, can I? Using `catchSomeCause` would work, but isn't the whole point of these new combinators to provide a simpler way to recover from dead fibers (as opposed to handle all possible `Cause`s)?

adamgfraser:
@gabro `refineOrDie` is going from some errors to fiber failures and thus making the error type smaller. `unrefine` is going from some fiber failures back to errors and thus making the error type larger. For example:

```scala
val unrefined: IO[Throwable, Nothing] = ???

val refined: IO[NumberFormatException, Nothing] =
  unrefined.refineOrDie { case t: NumberFormatException => t }

val unrefinedAgain: IO[Throwable, Nothing] =
  refined.unrefine { case t: Throwable => t }
```

You would recover from a fiber failure by using `unrefine` and then another error handling combinator. For example:

```scala
val died: UIO[Nothing] = ???

val recovered: UIO[IllegalArgumentException] =
  died.unrefineTo[IllegalArgumentException].flip
```

To your point it might be nice to have a single combinator that does that for you.

jdegoes:
> Ah, another thing: this seems to work only going from a defect to a failure.

That's true, but we already have lots of combinators to go from failure to success (`catchAll`, etc.). This is the most orthogonal way to allow a user to decide that some fatal error is relevant to the business: because once a defect is pulled into the error channel, it can be handled via the ordinary error combinators.

I think what you're looking for is something like `catchSomeDefect`, which would be the composition of `unrefine` and `catchSome`: where the partial function would be applied to defects and the catch some would be applied to the subset matching.

Something like:

```scala
effect.catchSomeDefect {
  case e : SecurityException => backup()
}
```

which would be equivalent to:

```scala
effect.unrefineTo[SecurityException].catchSome { case e : SecurityException => backup() }
```

Does that seem about right?

jdegoes:
Merged but let's continue to discuss naming and also `catchSomeDefect` or similar...

adamgfraser:
That makes sense to me. I will submit follow up PR to implement `catchSomeDefect`.

gabro:
@adamgfraser looking at your example I keep thinking we're conflating the two behaviors of `refineOrDie`. I see two cases:

- the refinement succeeds: the result is a failure
- the refinement fails: the result is a defect (a dead fiber)

What the proposed operator does is to recover from the second scenario.
`unrefine` suggests we're reverting the first, but that isn't true, is it?

```scala
val x = IO.fail(new Exception("failure")).refineOrDie {
  case _ => "some other error"
} // the refinement succeeded 

x.unrefine {
  case _ => "this will never run, since the fiber did not die"
}
```

Without looking at the types, I would expect `unrefine` to just broaden the error (which it does, since `E1 :> E`) but there's nothing in the name that suggests the "death recovering" behavior. That's why I think it would be much more user friendly if the name included something referring to "reviving" the fiber.

---

@jdegoes yes, `catchSomeDefect` would serve the purpose! The use case would be recovering from a defect without going through an intermediate failure (which I would have to do if I wanted to do it via `unrefine`)

--

Anyway, I 'm just a user and I think I've voiced my feedbacks enough. I trust your judgment as maintainers so I won't throw a fit about naming :-)

Thank you all for the quick feedback and for working on this so quickly!

jdegoes:
Neither refine nor unrefine alter failure behavior, only failure mode:

 - refine pinches off a piece of `E`, and converts it into a fatal error
 - unrefine pinches off a piece of a fatal error, and converts it into `E`

That is to say, if an effect fails, then after refine or unrefine, it will still fail; and if an effect succeeds, then after refine or unrefine, it will still succeed; only the manner in which it fails will be altered by these two methods.

Although they are symmetrical and do opposite things, they don't directly address your original use case. I wrote up #3085 for this use case.

PR #3086, opened 2020-03-10
ZIO Core: Implement ZIO#catchSomeDefect

### LINKED ISSUE #3085: ZIO Core: catchSomeDefect
A hypothetical `catchSomeDefect` would allow responding to a subset of defects:

```scala
sealed trait ZIO[-R, +E, +A] {
  def catchSomeDefect[R1 <: R, E1 >: E, A1 >: A](f: PartialFunction[Throwable, ZIO[R1, E1, A1]]): ZIO[R1, E1, A1]
}
```

/cc @gabro

adamgfraser:
Resolves #3085. Also implements `catchAllDefect` for symmetry with existing combinators. Copying @gabro.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -325,6 +325,11 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * {{{
    * effect.catchSomeDefect(_ => backup())
    * }}}
+   *
+   * '''WARNING''': There is no sensible way to recover from defects. This
+   * method should be used only at the boundary between ZIO and an external
+   * system, to transmit information on a defect for diagnostic or explanatory
+   * purposes.
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -319,6 +319,16 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def catchAllCause[R1 <: R, E2, A1 >: A](h: Cause[E] => ZIO[R1, E2, A1]): ZIO[R1, E2, A1] =
     self.foldCauseM[R1, E2, A1](h, new ZIO.SucceedFn(h))
 
+  /**
+   * Recovers from all defects with provided function.
+   *
```
jdegoes:
```suggestion

   * '''WARNING''': There is no sensible way to recover from defects. This method should be used only at the boundary between ZIO and an external system, to transmit information on a defect for diagnostic or explanatory purposes.
```

Here and below.

adamgfraser: Done.

PR #3096, opened 2020-03-11
Make the argument of ZIO.when by-name.

vvlevykin:
This is needed to make it the proper moral equivalent of `if (p) exp` as is written in the doc.

Otherwise, for example, this code fails with `java.lang.ArithmeticException: / by zero`:
```scala
val x = 0
IO.when(x > 0) {
  val y = 5 / x
  IO.succeed(y)
}
```

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -685,7 +685,7 @@ object IO {
   /**
    * @see See [[zio.ZIO.when]]
    */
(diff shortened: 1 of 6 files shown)


luis3m:
@vvlevykin I think it would also make sense to change `whenM` as well

adamgfraser:
This example involves performing side effects outside of any effect constructor. If you move `5 / x` into `ZIO.succeed` this code runs correctly. I'm all for addressing common errors but we already make effect constructors by name to address this and don't think we can manage performing side effects outside effect constructors in any consistent way.

Issue #3082, opened 2020-03-10
Improve performance of `TReentrantLock`


unclebob418:
Following from @KamalKang's work to benchmark performance in #3005 , we have some work to do to try and close the gap. I'll be working on this with support from @mijicd

PR #3099, opened 2020-03-11
Improve performance of `TReentrantLock`

unclebob418:
#3082

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TReentrantLock.scala
+++ b/core/shared/src/main/scala/zio/stm/TReentrantLock.scala
@@ -18,7 +18,8 @@ package zio.stm
 
 import TReentrantLock._
 
-import zio.{ Fiber, Managed, UIO, UManaged }
+import zio.stm.ZSTM.internal.TExit
+import zio.{ Fiber, Managed, UManaged }
 
 /**
  * A `TReentrantLock` is a reentrant read/write lock. Multiple readers may all
@@ -36,33 +37,35 @@ import zio.{ Fiber, Managed, UIO, UManaged }
  * allows this structure to be composed into more complicated concurrent
  * structures that are consumed from effectful code.
  */
-final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
+final class TReentrantLock private (data: TRef[LockState]) {
+
+  private val tExitOne = TExit.Succeed(1)
 
   /**
    * Acquires a read lock. The transaction will suspend until no other fiber
-   * is holding a write lock. Succeeds with the number of read locks held by
-   * this fiber.
+   * is holding a write lock. Succeeds with the number of read locks held by this fiber.
    */
-  lazy val acquireRead: USTM[Int] =
-    STM.fiberId.flatMap(fiberId => adjustRead(fiberId, 1))
+  lazy val acquireRead: USTM[Int] = adjustRead(1)
 
   /**
    * Acquires a write lock. The transaction will suspend until no other
    * fibers are holding read or write locks. Succeeds with the number of
    * write locks held by this fiber.
    */
-  lazy val acquireWrite: USTM[Int] =
-    for {
-      fiberId <- STM.fiberId
-      w <- data.get.collect {
-            case Left(readLock) if (readLock.noOtherHolder(fiberId)) =>
-              WriteLock(1, readLock.readLocks(fiberId), fiberId)
+  lazy val acquireWrite: USTM[Int] = new ZSTM((journal, fiberId, _, _) =>
+    data.unsafeGet(journal) match {
 
-            case Right(WriteLock(n, m, `fiberId`)) =>
-              WriteLock(n + 1, m, fiberId)
-          }
-      _ <- data.set(Right(w))
-    } yield w.writeLocks
+      case readLock: ReadLock if readLock.noOtherHolder(fiberId) =>
+        data.unsafeSet(journal, WriteLock(1, readLock.readLocks(fiberId), fiberId))
+        tExitOne
+
+      case WriteLock(n, m, `fiberId`) =>
+        data.unsafeSet(journal, WriteLock(n + 1, m, fiberId))
+        TExit.Succeed(n + 1)
+
+      case _ => TExit.Retry
+    }
+  )
 
   /**
    * Just a convenience method for applications that only need reentrant locks,
@@ -85,40 +88,47 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
     Managed.make(acquireRead.commit)(_ => releaseRead.commit)
 
   /**
-   * Retrieves the number of acquired read locks.
+   * Retrieves the total number of acquired read locks.
    */
-  def readLocks: USTM[Int] = data.get.map(_.fold(_.readLocks, _.readLocks))
+  def readLocks: USTM[Int] = data.get.map(_.readLocks)
+
+  /**
+   * Retrieves the number of acquired read locks for this fiber.
+   */
+  def fiberReadLocks: USTM[Int] =
+    new ZSTM((journal, fiberId, _, _) => TExit.Succeed(data.unsafeGet(journal).readLocks(fiberId)))
+
+  /**
+   * Retrieves the number of acquired write locks for this fiber.
+   */
+  def fiberWriteLocks: USTM[Int] =
+    new ZSTM((journal, fiberId, _, _) => TExit.Succeed(data.unsafeGet(journal).writeLocks(fiberId)))
 
   /**
    * Determines if any fiber has a read lock.
    */
-  def readLocked: USTM[Boolean] = readLocks.map(_ > 0)
+  def readLocked: USTM[Boolean] = data.get.map(_.readLocks > 0)
 
   /**
    * Releases a read lock held by this fiber. Succeeds with the outstanding
    * number of read locks held by this fiber.
    */
-  lazy val releaseRead: USTM[Int] =
-    STM.fiberId.flatMap(fiberId => adjustRead(fiberId, -1))
+  lazy val releaseRead: USTM[Int] = adjustRead(-1)
 
   /**
    * Releases a write lock held by this fiber. Succeeds with the outstanding
    * number of write locks held by this fiber.
    */
-  lazy val releaseWrite: USTM[Int] =
-    STM.fiberId.flatMap(fiberId =>
-      data.modify {
-        case Right(WriteLock(1, m, `fiberId`)) =>
-          0 -> Left(ReadLock(fiberId, m))
-
-        case Right(WriteLock(n, m, `fiberId`)) if n > 1 =>
-          val newCount = n - 1
-
-          newCount -> Right(WriteLock(newCount, m, fiberId))
-
-        case s => die(s"Defect: Fiber ${fiberId} releasing write lock it does not hold: ${s}")
-      }
-    )
+  lazy val releaseWrite: USTM[Int] = new ZSTM((journal, fiberId, _, _) => {
+    val res = data.unsafeGet(journal) match {
+      case WriteLock(1, m, `fiberId`) => ReadLock(fiberId, m)
+      case WriteLock(n, m, `fiberId`) if n > 1 =>
+        WriteLock(n - 1, m, fiberId)
+      case s => die(s"Defect: Fiber ${fiberId} releasing write lock it does not hold: ${s}")
+    }
+    data.unsafeSet(journal, res)
+    TExit.Succeed(res.writeLocks(fiberId))
+  })
 
   /**
    * Obtains a write lock in a managed context.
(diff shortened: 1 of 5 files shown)


unclebob418:
adjustRead was returning the sum of all read locks in the map, but the only places it's used both say it returns the number of locks held by the current fiber. replaced summing an iterable of values in a map with a single map lookup

unclebob418:
@mijicd I've taken a reasonably detailed look at the code (just within `TReentrantLock`, haven't looked at `TRef` at all). The `adjustRead` cahnges mentioned above seems to have improved performance a bit, but not as much as I originally thought. I'm not really trusting the benchmarks. I was getting some inconsistent results so decided to up the warm-up and measurement runs to 20 each, and ran it 3 times in a row overnight with nothing else open. As you can see below, the fluctuation in some of the tests is significantly more than the reported error. Any ideas what may be causing this? I have a theory that sometimes contention spirals out of control and may in rare cases significantly slow down throughput (potentially for the rest of the run), but have no idea how I'd test that.
image

Apart from the 'fix' I've done with adjustRead, the only other idea I've come up with is to return unit from all the acquire/release functions. This seems to improve performance -  the question is, do we need the number of read/write locks when we create/release them? I imagine the usual way of using the `TReentrantLock` is to use `writeLock` and `readLock` and not actually care how many read/write locks there are, only that you have one. If you do care for some reason you could still get the value fairly easily.

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -136,6 +136,9 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
    */
   def writeLocks: STM[Nothing, Int] = data.get.map(_.fold(_ => 0, _.writeLocks))
 
+  /**
```
mijicd: Maybe not put the scaladoc on private method.

mijicd:
My intuition always is to run 15+ iterations for both phases, it can help with reducing the noise. Regarding the signature change, I don't think it's unreasonable to use `Unit` (although penalty will be paid in obtaining the count).

unclebob418:
@jdegoes what are your thoughts on swapping the return type to `Unit`? I seems to improve performance in most (but not all) cases (this is showing totals of 3 runs of 20 iterations each overnight with nothing else open).
image
@mijicd is working on some other magic at a deeper level which will hopefully help a lot more.

mijicd:
@jdegoes The magic @unclebob418 is referring to was the trick with locks we spoke about. My hunch here is that `collect` is being dreadful due to retries, but I have no proof about that. Perhaps benchmarking it separately might confirm that (or not)?

jdegoes:
I am fine swapping return types to `Any` to avoid flatMapping to `Unit` overhead.

@mijicd Yes, we should create a ticket for dealing with "retry storms".

Honestly optimizing this lock is going to come down to two things:

1. Minimizing allocations in the happy path, which will require careful inspection and translation from one form to another.
2. Core STM performance work, which has degraded a bit in recent times due to necessary features like trampolining.

mijicd:
I'm already on retry storming. I'd say point 1 should be handled here, together with changing the signatures as you suggested.

unclebob418:
@mijicd @jdegoes I've done some more work on this using the unsafe get and set that @mijicd recently added. It gives a nice performance boost over and above swapping to return `unit` (and I could do it without flatMapping to `unit`). The unsafe changes also provide cheaper access to the number of locks. I guess the question is how much performance are we willing to sacrifice for a nicer (and more performant) experience when a user cares about the number of locks?

two alternatives:
1. provide both options (which would require code duplication)
1. the middle ground where we provide the number of write locks but not read locks (`WriteLock` count is available for free - it's the `ReadLock` count that costs). I don't like the inconsistency here.

benchmark results:
image

unclebob418:
@KamalKang any thoughts on above? do you use the lock counts?

jdegoes:
@unclebob418 How do these compare from baseline, pre-optimization?

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -98,32 +108,28 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
...
+  lazy val releaseWrite: USTM[Unit] = new ZSTM((journal, fiberId, _, _) => {
+    val res = data.unsafeGet(journal) match {
+      case Right(WriteLock(1, m, `fiberId`)) => Left(ReadLock(fiberId, m))
+      case Right(WriteLock(n, m, `fiberId`)) if n > 1 =>
+        val newCount = n - 1
+        Right(WriteLock(newCount, m, fiberId))
+      case s => die(s"Defect: Fiber ${fiberId} releasing write lock it does not hold: ${s}")
+    }
+    data.unsafeSet(journal, res)
+    TExit.Succeed(())
```
jdegoes, line 126: I would add `TExit.unit` and make sure all code uses that. No need to (endlessly re-)allocate constants

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -98,32 +108,28 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
...
-          val newCount = n - 1
-
-          newCount -> Right(WriteLock(newCount, m, fiberId))
-
-        case s => die(s"Defect: Fiber ${fiberId} releasing write lock it does not hold: ${s}")
-      }
-    )
+  lazy val releaseWrite: USTM[Unit] = new ZSTM((journal, fiberId, _, _) => {
+    val res = data.unsafeGet(journal) match {
+      case Right(WriteLock(1, m, `fiberId`)) => Left(ReadLock(fiberId, m))
```
jdegoes, line 119: There's an optimization you can do here, if you want: right now we represent the state as "either". But we can fuse that sum type into a new type `LockState`, which will be either `WriteLock` or `ReadLock`. This will eliminate 1 layer of wrapping and a bunch of needless allocations.

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -136,15 +142,29 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
...
+   * Adjusts the number of read locks. Succeeds with the number of read locks held by this fiber.
+   */
+  private def adjustRead(delta: Int): USTM[Unit] =
+    new ZSTM((journal, fiberId, _, _) =>
+      data.unsafeGet(journal) match {
+
+        case Left(readLock) =>
+          val res = readLock.adjust(fiberId, delta)
+          data.unsafeSet(journal, Left(res))
+          TExit.Succeed(())
```
jdegoes, line 155: `TExit.unit`

core/shared/src/main/scala/zio/stm/TReentrantLock.scala
```diff
@@ -136,15 +142,29 @@ final class TReentrantLock private (data: TRef[Either[ReadLock, WriteLock]]) {
...
+
+        case Right(WriteLock(w, r, `fiberId`)) =>
+          val newTotal = r + delta
+          if (newTotal < 0)
+            die(s"Defect: Fiber ${fiberId} releasing read locks it does not hold, newTotal: $newTotal")
+          else
+            data.unsafeSet(journal, Right(WriteLock(w, newTotal, fiberId)))
+          TExit.Succeed(())
```
jdegoes, line 163: `TExit.unit`

unclebob418:
@jdegoes that's the blue bars, probably should have called it baseline instead of benchmark but .. I didn't :) Reran that when I rebased. It's a significant percentage but still nothing compared to the java reentrant lock (which I also added a benchmark for) or stamped lock.

jdegoes:
@unclebob418 Are you open to improving the benchmark in this pull request?

There are some problems with it:

1. First, `writeLock` and `readLock` are used directly, which utilize `ZManaged`. That gives you safety, ensuring you release what you acquire, but it also adds overhead, and Java's locks don't have this feature, so it's not a fair comparison. So instead of using functions that use ZManaged, we should acquire and release manually.
2. Second, for purely functional code, it does not make sense to benchmark one operation in isolation and have JMH re-run it a lot. That's because for JMH to execute a purely functional test, it must call `unsafeRun`. This function creates a fiber (with a lot of resources) and uses a thread, so you end up measuring the overhead of fibers, and not the overhead of the operation itself.

To solve these problems: both Java and ZIO implementations should read/write lock manually and explicitly, in sequence; and these should be repeated a lot inside the benchmark methods (e.g. 10,000 times). This is separate from JMH iterations.

For the STM one, you can stay inside the STM monad, and just commit the repetition. 

Do these changes make sense?

unclebob418:
@jdegoes I should be able to manage that. Once I've done that I'll compare unit/int return types with the new benchmark and we can make a call on if we're willing to sacrifice a bit of performance for better ergonomics. I don't really want to rewrite all the tests unless we're committed to that.

unclebob418:
Gut feel is that with fiber overhead reduced in the benchmarks, the disparity will be greater, but better to measure these things!

mijicd:
@unclebob418 you can check out tmap benchmarks where we added the amortisation @jdegoes mentioned for all  single-element operations.

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
```diff
@@ -12,8 +12,8 @@ import zio._
 @State(Scope.Group)
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
-@Measurement(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 10)
-@Warmup(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 10)
+@Measurement(iterations = 5, timeUnit = TimeUnit.SECONDS, time = 5)
```
mijicd, line 15: I wouldn't decrease these.

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
```diff
@@ -176,28 +175,29 @@ class TReentrantLockBenchmark {
...
   def stampedLockWrite(): Unit = {
     val stamp = stampedLock.writeLock()
     doWork()
     stampedLock.unlockWrite(stamp)
-  }
+  }*/
+
+  @CompilerControl(CompilerControl.Mode.DONT_INLINE)
+  def reentrantLockWrite(): Unit =
+    for (_ <- calls) {
```
mijicd, line 187: I'd use `calls.foreach`, it's less awkward :).

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
```diff
@@ -141,29 +142,27 @@ class TReentrantLockBenchmark {
...
-    val io = for {
-      lock <- zioLock
-      _    <- lock.writeLock.use(_ => doWorkM())
-    } yield ()
-
-    unsafeRun(io)
-  }
+  def zioLockRead(): Unit =
+    unsafeRun(
+      ZIO.foreach_(calls)(_ => zioLock.flatMap(lock => lock.acquireRead.commit *> doWorkM() *> lock.releaseRead.commit))
```
jdegoes, line 154: This will measure the overhead of `foreach`. Instead, try:

```scala
def repeat(n: Int, effect: USTM[Any]): USTM[Any] = 
  if (n <= 0) effect else effect.flatMap(_ => effect)
```

Then use that in several places:

```scala
unsafeRun {
  repeat(CallCount)(lock.acquireRead.flatMap(_ => lock.releaseRead)).commit
}
```

Note: everything is important here. `zioLock` is actually a _recipe_ to create a lock, not the actual lock itself. So in some setup outside the actual benchmark, you've got to do: `zioLock = unsafeRun(ReentrantLock.make)`, where `zioLock` is a `var`. That way it will be _already created_ when the benchmark runs.

Similar changes elsewhere.

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
```diff
@@ -176,28 +175,29 @@ class TReentrantLockBenchmark {
...
+      reentrantLock.lock()
+      doWork()
+      reentrantLock.unlock()
+    }
 
   @CompilerControl(CompilerControl.Mode.DONT_INLINE)
+  def reentrantLockRead(): Unit =
+    for (_ <- calls) {
+      reentrantLock.lock()
+      doWork()
```
jdegoes, line 197: I would delete `doWork` and `doWorkM`, we don't want to measure the cost of doing any work, even a function call. We want to measure only the lock/unlock overhead.

unclebob418: won't this have the effect of significantly reducing contention? the locks will be for such a short time that they'll rarely conflict

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
unclebob418: `(lock.acquireRead.flatMap(_ => lock.releaseRead)).commit`
doesn't acquiring and releasing a read before committing it result in the lock always being in its starting state (i.e. no locks) as far as any other transactions are concerned? A lock isn't acquired until it's committed, and it's not released until it's committed again. the transaction overhead is part of the cost of using this lock, so I think it should be included in the benchmark.

jdegoes: Ah, you're right, we need to commit them separately!

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
jdegoes: OK, you're right about that. Let's do `ZIO.yieldTo` for `doWorkM`

For `doWork`, we are going to have to do a sleep, but the question is how long? If we want it to be fair, it would be the same "size" as `doWorkM` (which involves submitting work to a thread pool, and having a thread pick up and execute that work).

unclebob418:
I'm not going to have time for the rest of the week to work on this, so how about we merge these changes (still a significant performance improvement) and we can revisit the benchmarks when I have time.

created #3469 for benchmark improvements.

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
unclebob418: I did fiddle with this a bit, main issue is even the smallest sleep (1 ms) kills so many iterations that we probably need to run the  benchmark longer to get an accurate measure. I think we probably need to do something crazy like actually update some data structure. Something like the zio "compute pi" exercise - both the readers and writers have to do something to give some level of contention (updating the state or computing in vs out for the readers), but that will still be less than 1 ms, and it's a somewhat legitimate use case; the whole reason for having locks is to prevent concurrent updates in this sort of situation (I'm always suss on benchmarks that don't do anything somewhat useful in the real world).

benchmarks/src/main/scala/zio/stm/TReentrantLockBenchmark.scala
unclebob418: just as a quick and dirty check speed check - I'll bump them back up at the end of the other pr

mijicd:
Can you post the results before and after the change, just for the reference? Besides that, the changes you made look good, and I agree with the idea to tweak the benchmark in https://github.com/zio/zio/pull/3469.

unclebob418:
Sure, here's before and after
image

and in text
[info] Benchmark                                                            Mode  Cnt        Score       Error  Units      Run        Change
[info] TReentrantLockBenchmark.ZioLockBasic                                thrpt   20   522139.251   11491.685  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockBasic:zioLockReadGroup               thrpt   20   230120.090    3995.350  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockBasic:zioLockWriteGroup              thrpt   20   292019.160    8303.931  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockHighContention                       thrpt   20   629125.356   22723.223  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockHighContention:zioLockReadGroup3     thrpt   20   304528.013   11244.138  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockHighContention:zioLockWriteGroup3    thrpt   20   324597.343   11495.718  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockLowContention                        thrpt   20   600956.357   25310.179  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockLowContention:zioLockReadGroup1      thrpt   20   475434.587   18639.291  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockLowContention:zioLockWriteGroup1     thrpt   20   125521.770    6680.379  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockMediumContention                     thrpt   20   628972.304   10211.665  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockMediumContention:zioLockReadGroup2   thrpt   20   409846.666    7833.818  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockMediumContention:zioLockWriteGroup2  thrpt   20   219125.638    2442.258  ops/s        1        before
[info] TReentrantLockBenchmark.ZioLockBasic                                thrpt   20   599344.626   25511.630  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockBasic:zioLockReadGroup               thrpt   20   299051.943   13191.109  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockBasic:zioLockWriteGroup              thrpt   20   300292.684   12327.459  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockHighContention                       thrpt   20   715508.996    9512.984  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockHighContention:zioLockReadGroup3     thrpt   20   360064.931    4667.825  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockHighContention:zioLockWriteGroup3    thrpt   20   355444.065    5042.625  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockLowContention                        thrpt   20   726110.785   20294.538  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockLowContention:zioLockReadGroup1      thrpt   20   581371.937   15294.006  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockLowContention:zioLockWriteGroup1     thrpt   20   144738.848    5064.187  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockMediumContention                     thrpt   20   697996.173   15483.251  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockMediumContention:zioLockReadGroup2   thrpt   20   468654.310    9765.954  ops/s        2         after
[info] TReentrantLockBenchmark.ZioLockMediumContention:zioLockWriteGroup2  thrpt   20   229341.862    5751.922  ops/s        2         after

unclebob418:
@jdegoes @mijicd can we please merge this?

mijicd:
@unclebob418 yes :)

PR #3102, opened 2020-03-11
ZIO Core: Remove Has Constraint on ZLayer

adamgfraser:
Removes the constraint that `ROut <: Has[_]` in `ZLayer`. This constraint is actually only required for a couple of operations (`++` and `update`) so moves the type constraint to those methods directly. Of course to have a `ZLayer` that composes in the normal ways you need `ROut <: Has[_]` but this allows more flexibility in combinators to create intermediate layers that potentially do not satisfy this constraint as long as the final layer does.


jdegoes:
LOVE IT!!!!

PR #3104, opened 2020-03-11
Attempt to Resolve CI Issues on ScalaJS

adamgfraser:


adamgfraser:
Well at least we're getting a different error. 😃

jdegoes:
From what I know, a bigger container won't help: more memory must be allocated to node.js vi command line options or maybe Env variables

adamgfraser:
@jdegoes I don't know much about this but I believe we are doing that too with the new `jsSettings` which I took from the settings suggested in your comment to the issue. The other thing that is interesting is that with these changes the build appears to be consistently failing versus previously the build was flaky, for what I thought would have been innocuous changes.

jdegoes:
> The other thing that is interesting is that with these changes the build appears to be consistently failing versus previously the build was flaky, for what I thought would have been innocuous changes.

Well, I suppose that's a good sign: we're at least messing with the right settings?

jdegoes:
Found [this](https://support.circleci.com/hc/en-us/articles/360009208393-How-can-I-increase-the-max-memory-for-Node-) which also recommends increasing `--max-old-space-size`

`xlarge` has 8 GB so we could try the combination `xlarge` with `8000`, at least for diagnostic purposes.

adamgfraser:
Trying now.

adamgfraser:
All ScalaJS builds still failing.

jdegoes:
@adamgfraser 

Looks like they are failing in a new way:

```
- ZIOSpec
[info]   - raceAll
[info]     - returns last failure
[info] Timeout of 1 m exceeded.
[info]     - returns success when it happens after failure
[info] Timeout of 1 m exceeded.
[info]   - RTS finalizers
[info]     - bracket regression 1
[info] Timeout of 1 m exceeded.
[info]   - RTS asynchronous correctness
[info]     - deep effectAsyncM doesn't block threads
[info] Timeout of 1 m exceeded.
```

jdegoes:
This could indicate, perhaps, that we have tests which are flaky in javascript, and sometimes these tests result in out of memory errors; that when we "fix", we see the actual failures. At least that's one hypothesis we should keep in mind as we dig...

adamgfraser:
Seems like this is working now, though I don't think we ever fully identified the underlying cause.

PR #3105, opened 2020-03-11
clock.currentDateTime reflects exceptions in type signature (#3052)

### LINKED ISSUE #3052: ZIO Clock: Failure to reflect exception in type
Generated by `clock.currentDateTime`, which can apparently throw:

```
java.time.zone.ZoneRulesException: Unknown time-zone ID: America/New_York
```

The fix is to reflect the possibility of failure in the return type. I am not sure if `ZoneRulesException` is the only exception, or if there are others that can be thrown, but they should be reflected in the type and captured via `Task(...).refineOrDie { ... }`.

jorge-vasquez-2301:
Closes #3052. I've modified `clock.currentDateTime` to reflect the fact that an exception can be thrown, for that I've audited all Java methods we call in `clock/package.scala` to see if they throw exceptions or not, so we can properly reflect those in types:

- `System.currentTimeMillis`: https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#currentTimeMillis-- This doesn't throw exceptions.
- `System.nanoTime`: https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime-- This doesn't throw exceptions.
- `ZoneId.systemDefault`: https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#systemDefault-- It can throw a `DateTimeException` if the converted zone ID has an invalid format, or a `ZoneRulesException` if the converted zone region ID cannot be found. And, it turns out `ZoneRulesException` is a subtype of `DateTimeException`.
- `Instant.ofEpochMilli`: https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#ofEpochMilli-long- It can throw a `DateTimeException` if the instant exceeds the maximum or minimum instant
- `OffsetDateTime.ofInstant`: https://docs.oracle.com/javase/8/docs/api/java/time/OffsetDateTime.html#ofInstant-java.time.Instant-java.time.ZoneId- It can throw a `DateTimeException` if the result exceeds the supported range

As you can see, all of this methods throw a `DateTimeException`, so now the type signature of `clock.currentDateTime` is `IO[DateTimeException, OffsetDateTime]`.

### DIFF
--- a/core/shared/src/main/scala/zio/clock/package.scala
+++ b/core/shared/src/main/scala/zio/clock/package.scala
@@ -16,18 +16,19 @@
 
 package zio
 
-import java.time.{ Instant, OffsetDateTime, ZoneId }
+import java.time.{ DateTimeException, Instant, OffsetDateTime, ZoneId }
 import java.util.concurrent.TimeUnit
 
 import zio.duration.Duration
 
 package object clock {
+
   type Clock = Has[Clock.Service]
 
   object Clock extends PlatformSpecific with Serializable {
     trait Service extends Serializable {
       def currentTime(unit: TimeUnit): UIO[Long]
-      def currentDateTime: UIO[OffsetDateTime]
+      def currentDateTime: IO[DateTimeException, OffsetDateTime]
       def nanoTime: UIO[Long]
       def sleep(duration: Duration): UIO[Unit]
     }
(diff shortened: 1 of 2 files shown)


jorge-vasquez-2301:
Thank you @adamgfraser!

adamgfraser:
@jorge-vasquez-2301 Looks like the documentation needs to be updated to reflect the change.

jorge-vasquez-2301:
Yeah, I've just realized that @adamgfraser, thanks! I'll make the adjustments

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -22,12 +22,14 @@ import java.util.concurrent.TimeUnit
 import zio.duration.Duration
 
 package object clock {
+
+  import java.time.DateTimeException
```
adamgfraser: Very minor but can we move this to the imports at the top of the file with the other imports from `java.time`?

jorge-vasquez-2301: Ohhh yeah, sorry, didn't realize that

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -45,11 +47,13 @@ package object clock {
             Left(UIO.effectTotal(canceler()))
           }
 
-        def currentDateTime: ZIO[Any, Nothing, OffsetDateTime] =
-          for {
-            millis <- currentTime(TimeUnit.MILLISECONDS)
-            zone   <- ZIO.effectTotal(ZoneId.systemDefault)
-          } yield OffsetDateTime.ofInstant(Instant.ofEpochMilli(millis), zone)
+        def currentDateTime: IO[DateTimeException, OffsetDateTime] =
+          (for {
```
adamgfraser: Could define an intermediate value here if you wanted to avoid the awkward parentheses around the for comprehension.

jorge-vasquez-2301: Yeah, makes sense, I'll do that

adamgfraser:

jorge-vasquez-2301:
Thanks very much @adamgfraser!

PR #3109, opened 2020-03-12
Port reject/rejectM to ZManaged

vpavkin:
Ticks reject/rejectM in #2077

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -712,6 +712,24 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
     catchAll(e => pf.lift(e).fold[ZManaged[R, E1, A]](ZManaged.die(f(e)))(ZManaged.fail(_)))
 
   /**
+   * Fail with the returned value if the `PartialFunction` matches, otherwise
+   * continue with our held value.
+   */
+  def reject[R1 <: R, E1 >: E](pf: PartialFunction[A, E1]): ZManaged[R1, E1, A] =
+    rejectM(pf.andThen(ZManaged.fail(_)))
+
+  /**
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -711,6 +711,24 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
   )(f: E => Throwable)(implicit ev: CanFail[E]): ZManaged[R, E1, A] =
     catchAll(e => pf.lift(e).fold[ZManaged[R, E1, A]](ZManaged.die(f(e)))(ZManaged.fail(_)))
 
+  /**
+   * Fail with the returned value if the `PartialFunction` matches, otherwise
+   * continue with our held value.
+   */
+  def reject[R1 <: R, E1 >: E](pf: PartialFunction[A, E1]): ZManaged[R1, E1, A] =
```
adamgfraser: You can delete the `R1` type parameter here since you aren't using it anywhere. The partial function doesn't use an environment so the environmental requirements of the new effect will be the same as those of the original effect.

adamgfraser: Looks like the same issue is present in the implementations for `ZIO` and `ZSTM`. You want to fix there as well?

vpavkin: Yeah, I shouldn't have blindly copied that parameter :man_facepalming: .
Good catch.
WRT to ZIO and ZSTM, this might break explicitly annotated user calls, are we ok with it?

adamgfraser: No worries! That is why we do these reviews. Yes we are okay with breaking compatibility at this point. I highly doubt there is significant usage with annotations, users will be glad to not have to specify an extra type parameter if they do have to, and now is the time to make these changes before we release 1.0 and have to ensure binary compatibility. Good question though!

PR #3110, opened 2020-03-12
Add tags to tests/suites (#3029)

### LINKED ISSUE #3029: Add tags to tests / suites to make it easier to run related tests together
Currently ZIO Test lets us attach tags to tests and suites, but we're not using that feature.

It would be useful to add tags that allow us to group related tests. For example, all tests of interruption could utilize the tag "interruption", whether they are testing ZIO, ZStream, ZManaged, etc. Similarly, error handling tests could utilize the tag "errors". And so forth.

jorge-vasquez-2301:
Closes #3029

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOBaseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOBaseSpec.scala
@@ -1,8 +1,45 @@
 package zio
 
+import scala.annotation.tailrec
+
 import zio.duration._
 import zio.test._
 
 trait ZIOBaseSpec extends DefaultRunnableSpec {
   override def aspects = List(TestAspect.timeout(60.seconds))
+
+  sealed trait ZIOTag {
+    val value: String
+    val subTags: List[ZIOTag] = Nil
+  }
+  object ZIOTag {
+    case object errors extends ZIOTag { override val value = "errors" }
+    case object future extends ZIOTag { override val value = "future" }
+    case object interop extends ZIOTag {
+      override val value                 = "interop"
+      override val subTags: List[ZIOTag] = List(future)
+    }
+    case object interruption extends ZIOTag { override val value = "interruption" }
+    case object regression   extends ZIOTag { override val value = "regression"   }
+    case object supervision  extends ZIOTag { override val value = "supervision"  }
+  }
+
+  def zioTag(zioTag: ZIOTag, zioTags: ZIOTag*): TestAspectPoly = {
+    val tags = zioTags.map(_.value) ++ getSubTags(zioTag) ++ zioTags.flatMap(getSubTags)
+    TestAspect.tag(zioTag.value, tags.distinct: _*)
+  }
+
+  private def getSubTags(zioTag: ZIOTag): List[String] = {
+    @tailrec
+    def loop(currentZioTag: ZIOTag, remainingZioTags: List[ZIOTag], result: List[String]): List[String] =
+      (currentZioTag.subTags, remainingZioTags) match {
+        case (Nil, Nil)      => currentZioTag.value :: result
+        case (Nil, t :: ts)  => loop(t, ts, currentZioTag.value :: result)
+        case (st :: sts, ts) => loop(st, sts ++ ts, currentZioTag.value :: result)
+      }
+    zioTag.subTags match {
+      case t :: ts => loop(t, ts, Nil)
+      case Nil     => Nil
+    }
+  }
 }
--- a/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
@@ -12,6 +12,8 @@ import zio.test.environment.Live
 
 object RTSSpec extends ZIOBaseSpec {
 
+  import ZIOTag._
+
   def spec = suite("Blocking specs (to be migrated to ZIOSpecJvm)")(
     testM("blocking caches threads") {
       import zio.blocking.Blocking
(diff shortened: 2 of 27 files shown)


jorge-vasquez-2301:
@jdegoes @adamgfraser I've added tags to some of the tests in `ZIOSpec`, and before I continue I would like to get some feedback from you guys, do you think the tags I'm creating/applying make sense? Maybe I'm creating too many categories or am I in the right path?

adamgfraser:
@jorge-vasquez-2301 First, I love what you did with the ZIO Tags! That is really nice.

My suggestion would be that we approach this incrementally. There are thousands of tests and a lot of different ways to categorize so I tend to think that a good PR would make incremental progress on this and not be a comprehensive solution.

What if we initially focus on a couple "high priority" tags and try to define those tags and tag everything related to them and then we can expand out to other takes or potentially merge with some tags done and pursue others separately.

I think if you look at the existing string tags so far (e.g. interruption, supervision, interop, future) they are probably a good starting point and plenty of work to do initially.

One other idea that I think might be good to explore with the ZIO Tags is some notion of sub typing. For example, right now we have "future" and "interop" tags. I would think "future" is a subset of "interop" in that "future" is one specific other protocol that we want to interop with. So it could be nice if we can express that so we can just tag a test with future and it will run whether we run all tests tagged future or interop.

Does that make sense?

jorge-vasquez-2301:
@adamgfraser OK, I think I get the idea, I'm thinking about a solution like this for `ZIOTag`, what do you think?
```scala
  sealed trait ZIOTag {
    val value: String
    val subTags: List[ZIOTag] = Nil
  }
  object ZIOTag {
    case object ErrorHandling extends ZIOTag { override val value = "error-handling" }
    case object Future        extends ZIOTag { override val value = "future"         }
    case object Interop extends ZIOTag {
      override val value                 = "interop"
      override val subTags: List[ZIOTag] = List(Future)
    }
    case object Interruption extends ZIOTag { override val value = "interruption" }
    case object Regression   extends ZIOTag { override val value = "regression"   }
    case object Supervision  extends ZIOTag { override val value = "supervision"  }
  }

  def zioTag(zioTag: ZIOTag, zioTags: ZIOTag*): TestAspectPoly = {
    val tags = zioTags.map(_.value) ++ getSubTags(zioTag) ++ zioTags.flatMap(getSubTags)
    TestAspect.tag(zioTag.value, tags.distinct: _*)
  }

  private def getSubTags(zioTag: ZIOTag): List[String] = {
    @tailrec
    def loop(currentZioTag: ZIOTag, remainingZioTags: List[ZIOTag], result: List[String]): List[String] =
      (currentZioTag.subTags, remainingZioTags) match {
        case (Nil, Nil)      => currentZioTag.value :: result
        case (Nil, t :: ts)  => loop(t, ts, currentZioTag.value :: result)
        case (st :: sts, ts) => loop(st, sts ++ ts, currentZioTag.value :: result)
      }
    zioTag.subTags match {
      case t :: ts => loop(t, ts, Nil)
      case Nil     => Nil
    }
  }
```

core-tests/shared/src/test/scala/zio/ZIOBaseSpec.scala
```diff
@@ -5,4 +5,36 @@ import zio.test._
 
 trait ZIOBaseSpec extends DefaultRunnableSpec {
   override def aspects = List(TestAspect.timeout(60.seconds))
+
+  sealed trait ZIOTag {
+    val value: String
+  }
+  object ZIOTag {
+    case object Caching extends ZIOTag { override val value = "caching" }
```
jdegoes: From the perspective of a user who will be running these, I suggest each tag be 1 word long, maximum. And shortest form of a word, e.g. `ErrorHandling` => `Errors`, `Memoization` => `Memo`, `ResourceHandling` => `Resource`, etc.

adamgfraser:
@jorge-vasquez-2301 Yes. I think you could simplify by using a case class.

core-tests/shared/src/test/scala/zio/ZIOBaseSpec.scala
jorge-vasquez-2301: Yeah, makes sense, thanks for the suggestion @jdegoes!

core-tests/shared/src/test/scala/zio/ZIOBaseSpec.scala
```diff
@@ -1,8 +1,45 @@
...
 trait ZIOBaseSpec extends DefaultRunnableSpec {
   override def aspects = List(TestAspect.timeout(60.seconds))
+
+  sealed trait ZIOTag {
+    val value: String
+    val subTags: List[ZIOTag] = Nil
+  }
+  object ZIOTag {
+    case object Errors        extends ZIOTag { override val value = "errors" }
+    case object FutureInterop extends ZIOTag { override val value = "future" }
```
jorge-vasquez-2301: I had to use two words for this tag: `FutureInterop`, I couldn't make it just `Future` because it would clash with scala `Future`, do you have any ideas for making this just one word @adamgfraser @jdegoes?

jorge-vasquez-2301:
@jdegoes @adamgfraser could you please take a look now to this PR? I've added tags to several tests, and want to know your opinion and if the tags I've added are sufficient for now

jorge-vasquez-2301:
@jdegoes Just updated the PR according to your suggestions 🙂

jdegoes:
@jorge-vasquez-2301 Thank you for your contribution!

jorge-vasquez-2301:
@jdegoes @adamgfraser I'm happy to help! 😄

PR #3111, opened 2020-03-12
ZIO Test: Implement Laws

adamgfraser:
Initial work on implementing laws. I didn't like that the previous signature of laws was `Laws[Caps[_], R, A]`, which implies that the laws are parameterized on a particular type (e.g. they are equality laws for integers, rather than equality laws in general). In this PR the signature goes to:

```scala
sealed trait Laws[-Caps[_], -R] {
  def run[R1 <: R, A](caps: Caps[A], gen: Gen[R1, A]): ZIO[R1, Nothing, TestResult]
}
```

So a set of laws has to be able to return a test result for any type `A` that has the required capabilities and a generator. This requires that laws be polymorphic and only use information reflected in the capabilities and allows us to define laws for equality in general instead of for a particular type. I think we could recover monomorphic laws to the extent we want them by defining a capability as an isomorphism between an abstract type and a concrete type, possibly under the hood.

I think the result of this is very nice but it does require a bit more work to define the laws because the laws are now universally quantified subject to the capabilities constraint so defining a law requires implement a method in a trait. I've tried to make this as ergonomic as possible and I think it is not bad but this would be an area for further work.


test/shared/src/main/scala/zio/test/Laws.scala
```diff
@@ -0,0 +1,71 @@
...
+    def apply[Caps[_], R](left0: Laws[Caps, R], right0: Laws[Caps, R]): Laws[Caps, R] =
+      new Both[Caps, R] {
+        val left  = left0
+        val right = right0
+      }
+  }
+
+  abstract class Law[-Caps[_]](label: String) extends Laws[Caps, Any] { self =>
+
+    def apply[A](caps: Caps[A], a1: A, a2: A, a3: A): TestResult
```
jdegoes:
```suggestion
    def apply[A](a1: A, a2: A, a3: A)(implicit caps: Caps[A]): TestResult
```

test/shared/src/main/scala/zio/test/Laws.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  object Both {
+    def apply[Caps[_], R](left0: Laws[Caps, R], right0: Laws[Caps, R]): Laws[Caps, R] =
+      new Both[Caps, R] {
+        val left  = left0
+        val right = right0
+      }
+  }
+
+  abstract class Law[-Caps[_]](label: String) extends Laws[Caps, Any] { self =>
```
jdegoes:
```suggestion
  abstract class Law3[-Caps[_]](label: String) extends Laws[Caps, Any] { self =>
```

For "law of 3 values"? Suggesting `Law1`, `Law2`, etc.

test/shared/src/main/scala/zio/test/Laws.scala
```diff
@@ -0,0 +1,71 @@
...
+ * return a test result. Laws can be combined using `+` to produce a set of
+ * laws that require both sets of laws to be satisfied.
+ */
+sealed trait Laws[-Caps[_], -R] { self =>
+
+  /**
+   * Test that values of type `A` satisfy the laws using the specified
+   * generator.
+   */
+  def run[R1 <: R, A](caps: Caps[A], gen: Gen[R1, A]): ZIO[R1, Nothing, TestResult]
```
jdegoes:
```suggestion
  def run[R1 <: R, A](gen: Gen[R1, A])(implicit caps: Caps[A]): ZIO[R1, Nothing, TestResult]
```

jdegoes: Or possibly even:

```suggestion
  def run[R1 <: R, A: Caps](gen: Gen[R1, A]): ZIO[R1, Nothing, TestResult]
```

test/shared/src/main/scala/zio/test/Laws.scala
jdegoes: Or possibly:

```suggestion
    def apply[A: Caps](a1: A, a2: A, a3: A): TestResult
```

test/shared/src/main/scala/zio/test/Laws.scala
```diff
@@ -0,0 +1,71 @@
...
+        val right = right0
+      }
+  }
+
+  abstract class Law[-Caps[_]](label: String) extends Laws[Caps, Any] { self =>
+
+    def apply[A](caps: Caps[A], a1: A, a2: A, a3: A): TestResult
+
+    final def run[R, A](caps: Caps[A], gen: Gen[R, A]): ZIO[R, Nothing, TestResult] =
+      check(gen, gen, gen)(apply(caps, _, _, _).map(_.label(label)))
```
jdegoes: In the event of a failure, the user would like to know: what are the values of the parameters which triggered the failure?

test-tests/shared/src/test/scala/zio/test/LawsSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+    def apply[A](caps: Equal[A], a1: A, a2: A, a3: A) = {
+      implicit val eq: Equal[A] = caps
+      assert((a1 === a2) ==> (a2 === a1))(isTrue)
+    }
+  }
+
+  val transitivityLaw = new Laws.Law[Equal]("transitivityLaw") {
+    def apply[A](caps: Equal[A], a1: A, a2: A, a3: A) = {
+      implicit val eq: Equal[A] = caps
+      assert(((a1 === a2) && (a2 === a3)) ==> (a1 === a3))(isTrue)
```
jdegoes: If all the laws are going to be `assert(x)(isTrue)`, then it makes sense for them to return boolean values. It doesn't lead to helpful error messages (see also: ScalaCheck!), but printing out the "failure combination" will help a lot with that.

On the other hand, I feel like "downstream", it could make sense to introduce a ZIO Test DSL where `(a === a2)` actually generates a `TestResult` given `Equal[A]`. Or maybe use a different operator, e.g. Cats uses `<->`.

test-tests/shared/src/test/scala/zio/test/LawsSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  val equalLaws = reflexiveLaw + symmetryLaw + transitivityLaw
```
jdegoes: This is really nice to see. ❤️

jdegoes: It gives me the following idea:

```scala
trait ZLawful[-Caps[_], -R] {
  def laws: Laws[Caps, R]
}
trait Lawful[-Caps[_]] extends ZLawful[Caps, Gen with Random]
...
trait Equal[-A] { ... }
object Equal extends Lawful[Equal] {
  def laws = ...
}
...
checkAllLaws(Equal)(gen.anyInt)
checkAllLaws(Equal, Hash, Ord)(gen.anyInt)
...
```

test/shared/src/main/scala/zio/test/Laws.scala
adamgfraser: Wasn't sure how far we wanted to go with the implicits but agree this makes sense and will make writing laws more ergonomic.

test/shared/src/main/scala/zio/test/Laws.scala
adamgfraser: Yes.

test/shared/src/main/scala/zio/test/Laws.scala
adamgfraser: Yes.

test/shared/src/main/scala/zio/test/Laws.scala
adamgfraser: I think we already get this out of the box from the test framework. For example, here is the output if I misstate the reflexive law and say that any two values are always equal instead of the same value is always equal to itself:

```scala
[info] - LawsSpec
[info]   - equalLaws
[info]     - int
[info]       Test failed after 1 iteration with input: ((0,1),0)
[info]       Original input before shrinking was: ((-1295463240,421138995),-1691075023)
[info]       false did not satisfy (isTrue() ?? "reflexiveLaw")
```

test-tests/shared/src/test/scala/zio/test/LawsSpec.scala
adamgfraser: Yes. This was just an example. In this case we can rewrite these to assert that the left hand side is equal to the right hand side so we get more helpful error messages. In general as you say we need an assertion that uses `Equal`. Question would be where that should live. I like the idea of calling it `===` but am worried that may create ambiguity or confusion with the `===` syntax for `Equal` so we may need to do something like `<->`.

test-tests/shared/src/test/scala/zio/test/LawsSpec.scala
adamgfraser: Yes!

jdegoes:
Here's overall my thoughts on laws: they will almost only ever be created by library authors. 

However, they will be used by a lot more people than that, to test custom instances, especially if the process of using and composing them is seamless and pain-free.

So it's more important to bias toward the latter group...

adamgfraser:
Great! I think with this encoding testing and composing them can be extremely ergonomic, especially, and even for library authors it can be quite easy with the changes you suggested above. I will work on incorporating these and will have another version shortly.

adamgfraser:

adamgfraser:
Yes. The other thing we should think about in regard to that is we may need more generators, because in some cases we need to be able to generate for example both an `A` and an `F[A]` for testing laws. Perhaps we can deal with that by requiring a `Gen[R, A]` and an inject function `A => F[A]`. I really want to avoid these situations where you have laws that require twenty different generators.

PR #3122, opened 2020-03-13
Move macros into main repository

ioleo:
@softinio @mijicd Please validate the build config.


build.sbt
```diff
@@ -305,10 +309,10 @@ lazy val examples = crossProject(JVMPlatform, JSPlatform)
   .in(file("examples"))
   .settings(stdSettings("examples"))
   .settings(crossProjectSettings)
-  .settings(macroSettings)
+  .settings(macroExpansionSettings)
   .settings(testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework"))
   .jsSettings(libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC5" % Test)
-  .dependsOn(testRunner)
+  .dependsOn(macros, testRunner)
```
adamgfraser: Do we need this? I would like to avoid making ZIO Test depend on ZIO Macros and instead have it be an optional feature that users can include if they want? Also it doesn't seem to be a dependency on other platforms right now which seems to indicate it is not necessary.

ioleo: It's not ZIO Test - it's examples. I agree with you on macros being optional.

adamgfraser: Ah sorry I misread.

jdegoes:
I think all of our macro stuff should be in one module (including generating `Gen`).

ioleo:
@jdegoes I though about to splitting this into two modules:
* `zio-macros-core`
* `zio-macros-test` 

and move the `@Mockable` to the macros-test module. So far we published under the same names, but the latest version was `0.6.2`, so if we start publishing now under the same version as ZIO Core there will be no clashes.

I'm not familiar with the `Gen` macro. What other macros we have/plan to have? I'd like to clean this up before 1.0.

jdegoes:
OTHO perhaps we should have no separate macro package and just include them in the 2.x line of ZIO Test?

ioleo:
The thing is `Accessible` is useful in `src/main` code where you don't want to have a dependency on ZIO Test. `@Mockable` is perfectly fine in ZIO Test, so the only reason to move it to `zio-macros-test` would be to seperate macros from the rest of the projects. That still leaves the `typeCheck` backed by macro in ZIO Test.

adamgfraser:
Seems like we have `accessible` that needs to be in some not ZIO Test related project.

Then we have `mockable` and `Gen` derivation which is powered by Magnolia.

I'm open to thoughts on what would create the best user experience. My initial thought is to possibly split out all the mocking functionality into its own project, including macros. At least historically one of our selling points has been having a very light dependency footprint. So we would have:

`zio-test`
`zio-test-mock`
`zio-test-magnolia`

But there are a lot of ways to cut this and open to others.

ioleo:
@adamgfraser Part of mocking functionality is tight integration with ZIO Test. Basically the mocked services die with package private `MockException`, which receives special handling by DefaultTestRepoerter. To seperate those we'd have to refactor the test reporter into more pluggable design.

adamgfraser:
@ioleo Yes I think the amount of mock specific code in there is a design smell and separation would enforce a positive refactoring there, which would also support a cleaner separation with JUnit and Intellij test runners.

PR #3125, opened 2020-03-13
ZIO Core: Disconnect Children in Uninterruptible Race

adamgfraser:
Automatically disconnects children if `race` is performed in an uninterruptible region. I think there was a bug with the prior implementation of `disconnect`:

```scala
for {
  id    <- ZIO.fiberId
  fiber <- restore(self).forkDaemon
  a     <- restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)
} yield a
```

The problem is that we call `restore` on `fiber.join`, which means if we are in an uninterruptible region joining the fiber is itself uninterruptible so so we suspend when we attempt to interrupt the join until the child fiber completes executing (possibly never). I think we want to change that line to be `fiber.join.interruptible`. This leaves the underlying effect uninterruptible if it was in an uninterruptible region but allows us to continue with the effect in the background.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2686,24 +2686,6 @@ object ZIOSpec extends ZIOBaseSpec {
 
         assertM(Live.live(io))(isTrue)
       },
-      testM("does not make effect interruptible") {
-        for {
-          latch1 <- Promise.make[Nothing, Unit]
-          latch2 <- Promise.make[Nothing, Unit]
-          fiber  <- latch1.succeed(()).ensuring(latch2.succeed(()).disconnect).fork
-          _      <- latch1.await
-          _      <- fiber.interrupt
-          _      <- latch2.await
-        } yield assertCompletes
-      } @@ nonFlaky,
-      testM("allows interruption to return immediately even in an uninterruptible region") {
-        ZIO.uninterruptible {
-          for {
-            fiber <- ZIO.infinity.disconnect.fork
-            _     <- fiber.interrupt
-          } yield assertCompletes
-        }
-      },
       testM("cause reflects interruption") {
         val io =
           for {
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2268,6 +2268,10 @@ object ZIOSpec extends ZIOBaseSpec {
         val io = IO.effectTotal(42).race(IO.never)
         assertM(io)(equalTo(42))
       },
+      testM("race in uninterruptible region") {
+        val effect = (ZIO.unit.race(ZIO.infinity)).uninterruptible
+        assertM(effect)(isUnit)
+      },
       testM("firstSuccessOf of values") {
         val io = IO.firstSuccessOf(IO.fail(0), List(IO.succeedNow(100))).either
         assertM(io)(isRight(equalTo(100)))
@@ -2681,6 +2685,24 @@ object ZIOSpec extends ZIOBaseSpec {
           } yield test
 
         assertM(Live.live(io))(isTrue)
+      },
+      testM("does not make effect interruptible") {
+        for {
+          latch1 <- Promise.make[Nothing, Unit]
+          latch2 <- Promise.make[Nothing, Unit]
+          fiber  <- latch1.succeed(()).ensuring(latch2.succeed(()).disconnect).fork
+          _      <- latch1.await
+          _      <- fiber.interrupt
+          _      <- latch2.await
+        } yield assertCompletes
+      } @@ nonFlaky,
+      testM("allows interruption to return immediately even in an uninterruptible region") {
+        ZIO.uninterruptible {
+          for {
+            fiber <- ZIO.infinity.disconnect.fork
+            _     <- fiber.interrupt
+          } yield assertCompletes
+        }
       },
       testM("cause reflects interruption") {
         val io =
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -430,7 +430,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
       for {
         id    <- ZIO.fiberId
         fiber <- restore(self).forkDaemon
-        a     <- fiber.join.interruptible.onInterrupt(fiber.interruptAs(id).forkDaemon)
+        a     <- restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)
       } yield a
     )
 
@@ -1153,8 +1153,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
     ZIO.descriptorWith { descriptor =>
       val parentFiberId   = descriptor.id
       val uninterruptible = descriptor.interruptStatus.isUninterruptible
-      val left            = if (uninterruptible) self.disconnect else self
-      val right           = if (uninterruptible) that.disconnect else that
+      val left =
+        if (uninterruptible) self.uninterruptible.disconnect.interruptible
+        else self
+      val right =
+        if (uninterruptible) that.uninterruptible.disconnect.interruptible
+        else that
       (left raceWith right)(
         (exit, right) =>
           exit.foldM[Any, E1, A1](
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -3291,6 +3291,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     def apply[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
       zio.interruptStatus(flag)
 
+    /**
+     * Returns a new effect that, if the parent region is uninterruptible, can
+     * be interrupted in the background instantaneously. If the parent region
+     * is interruptible, then the effect can be interrupted normally, in the
+     * foreground.
+     */
     def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
       if (flag == InterruptStatus.Uninterruptible) zio.uninterruptible.disconnect.interruptible
       else zio.interruptStatus(flag)
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -430,7 +430,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
       for {
         id    <- ZIO.fiberId
         fiber <- restore(self).forkDaemon
-        a     <- restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3298,7 +3300,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
 
   final class TimeoutTo[R, E, A, B](self: ZIO[R, E, A], b: B) {
     def apply[B1 >: B](f: A => B1)(duration: Duration): ZIO[R with Clock, E, B1] =
-      (self map f) raceFirst (ZIO.sleep(duration).interruptible as b)
+      (self map f) raceFirst (ZIO.sleep(duration) as b)
```
neko-kai: I think there's no reason not to make it interruptible, even with disconnect - disconnect will still leak the fiber.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -430,7 +430,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
       for {
         id    <- ZIO.fiberId
         fiber <- restore(self).forkDaemon
-        a     <- restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)
+        a     <- fiber.join.interruptible.onInterrupt(fiber.interruptAs(id).forkDaemon)
```
neko-kai: I don't think this is intuitive, this means that `x.disconnect` automatically implies `x.interruptible` for the current fiber - i.e. any critical section can be terminated if there's disconnect. I think the act of disconnecting the interruption of x should be separate from making sure this interruption can actually happen.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1150,22 +1150,24 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
-          (exit, left) =>
-            exit.foldM[Any, E1, A1](
-              cause => left.join mapErrorCause (_ && cause),
-              a => (left interruptAs parentFiberId) as a
-            )
-        )
+    ZIO.descriptorWith { descriptor =>
+      val parentFiberId   = descriptor.id
+      val uninterruptible = descriptor.interruptStatus.isUninterruptible
+      val left            = if (uninterruptible) self.disconnect else self
```
neko-kai: You can use `self.uninterruptible.disconnect.interruptible` here, without changing interruptibility for all uses of disconnect

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Yes, you're right. Will change.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: I don't think that is right because we are still doing `restore(self)` on the underlying effect. Without this change I don't think `disconnect` has its intended effect of allowing you to continue without waiting for interruption to complete if called within an uninterruptible region.

neko-kai: @adamgfraser I think it's logical that within an uninterruptible region `disconnect` becomes almost a no-op – because the outer effect can't receive interruptions, the underlying effect can't be interrupted in normal circumstances. _Forcing_ it to receive interruptions compromises the entire uninterruptible region of the outer fiber, I don't think that would make sense. And you can use `self.uninterruptible.disconnect.interruptible` to force the outer effect's `join` to be interruptible, while leaving the inner effect uninterruptible.

Note, _even_ with this change if the underlying effect is interrupted, but not the outer effect, the outer effect will still wait until the fiber stops completely before returning from `fiber.join` - observers aren't woken up until the fiber finishes completely.

So, to guarantee that interruption of the inner effect results in immediate resumption of the outer effect, we must use a Promise + onInterrupt in the inner effect – `fiber.join.interruptible` does not achieve this, AND it exposes the entire region to interruption which is avoidable with Promises.

adamgfraser: @neko-kai I added some tests to try to demonstrate what I think is the correct behavior.

I see the impact of `disconnect` in an uninterruptible region as quite significant, and essentially guaranteeing that interruption of a disconnected effect will return immediately regardless of whether the region is interruptible or uninterruptible.

This can be done in a way where we are only ensuring that the interruption can return immediately, the underlying effect itself is still guaranteed to execute if it is in an uninterruptible region.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -2682,6 +2686,24 @@ object ZIOSpec extends ZIOBaseSpec {
 
         assertM(Live.live(io))(isTrue)
       },
+      testM("does not make effect interruptible") {
```
neko-kai: What does this test? There is no uninterruptible region, all the effects in this test are interruptible.

There's a test for the behavior I'm talking about:

```scala
for {
    p <- Promise.make[String, Unit]
    fiber <- ZIO.never.disconnect.onInterrupt(p.fail("Interrupted even when inside uninterruptible region")).uninterruptible.fork
    res <- fiber.interrupt.timeout(5.seconds)
  } yield res.isEmpty // interruption shouldn't succeed
```

Here the changes to `disconnect` will force the fiber's `.onInterrupt` to happen despite the entire fiber being uninterruptible.

adamgfraser: The finalizer in `ensuring` will be uninterruptible. If I replace `disconnect` with `interruptible` there the test will fail.

neko-kai: @adamgfraser The effect inside `.disconnect` will be uninterruptible inded, but everything outside, inside the `ensuring` will still be interrupted. Try with `ZIO.unit.disconnect *> latch2.succeed(())` - this should fail

adamgfraser: @neko-kai Yes I see your point. I am wondering whether there is a way to address both of these points. I think you're right that we shouldn't change the interruptibility of the outer effect. But the fact that we can't just use `disconnect` in situations like this and need to then make the disconnection interruptible seems problematic to me. But I may just be wrong there. 😃

neko-kai: @adamgfraser I think it's ok to use `.disconnect.interruptible.ensuring(<rest-of-the-finalizer>)` explicitly in these situations, when you really need to disconnect early inside a finalizer. There's no way to react to interruption, without being interrupted - that means there's no way to be disconnected without being interrupted too. This tradeoff is fine by me, I think it's reasonable, I have done a lot of arguing to make sure that interruption is irreversible and intend to keep doing it. Everyone who knows what they're doing can achieve the  desired behavior explicitly, but we shouldn't open interruptibility holes in innocuous looking functions because this can seriously mess up with people's ability to understand what happens to their code.

adamgfraser: @neko-kai Yes. Updated to incorporate your feedback. Thank you for pushing here.

adamgfraser:
Updated to just disconnect the children in `race` and leave `disconnect` unchanged.

jdegoes:
Please don't merge until I can review!

adamgfraser:
@jdegoes Will not do anything without your approval!

jdegoes:
I think this problem should be solved by modifying:

```scala
final class InterruptStatusRestore(private val flag: zio.InterruptStatus) extends AnyVal {
  def apply[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
    zio.interruptStatus(flag)

  def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] = 
    if (flag == InterruptStatus.Uninterruptible) zio.disconnect.interruptible
    else zio.interruptStatus(flag)
}
```

Then you can use:

```scala
ZIO.uninterruptibleMask { interruptible =>
  ... interruptible.force(zio) ...
}
```

Other comments inline.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1150,22 +1150,24 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
-            exit.foldM[Any, E1, A1](
-              cause => left.join mapErrorCause (_ && cause),
-              a => (left interruptAs parentFiberId) as a
-            )
-        )
+    ZIO.descriptorWith { descriptor =>
+      val parentFiberId = descriptor.id
+      def maybeDisconnect[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
+        if (descriptor.interruptStatus.isInterruptible) zio
+        else zio.uninterruptible.disconnect.interruptible
```
jdegoes: Why `zio.uninterruptible.disconnect.interruptible`? It should be sufficient (and more performant) to do:

```
zio.disconnect.interruptible
```

neko-kai: @jdegoes So that the underlying `zio` itself remains uninterruptible and can't be sent interrupts to - otherwise it could still be interrupted indirectly by supervision etc.

jdegoes: When it's disconnected there's no way to interrupt it aside from having a reference to the fiber.

neko-kai: @jdegoes 
It will be interrupted inside `disconnect`:
```scala
  final def disconnect: ZIO[R, E, A] =
    ZIO.uninterruptibleMask(restore =>
      for {
        id    <- ZIO.fiberId
        fiber <- restore(self).forkDaemon
        a     <- restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)
      } yield a
    )
```

With `zio.disconnect.interruptible` the following will print "interrupted uninterruptible never":

```scala
ZIO.sleep(5.seconds).race(ZIO.never.onInterrupt(putStrLn("interrupted uninterruptible never")))
```

[Proof](https://scastie.scala-lang.org/VBPVYBOnS2ycowo7s6Sx7Q)

Because the right hand side will be interrupted in `restore(fiber.join).onInterrupt(fiber.interruptAs(id).forkDaemon)`, i.e. omitting `zio.uninterruptible` makes all races interruptible as in RC17 – which I'm probably all for, but doesn't seem like what you intend.

jdegoes: Ah right, sorry, I should know that. 😆

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3289,6 +3290,10 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   final class InterruptStatusRestore(private val flag: zio.InterruptStatus) extends AnyVal {
     def apply[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
       zio.interruptStatus(flag)
+
+    def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
```
jdegoes:
```suggestion
    /**
     * Returns a new effect that, if the parent region is uninterruptible, can be interrupted in the background
     * instantaneously. If the parent region is interruptible, then the effect can be interrupted normally,
     * in the foreground.
     */
    def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
```

PR #3128, opened 2020-03-14
Add ZStream.fromIterator variants

regiskuckaertz:


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2113,7 +2206,7 @@ object ZStream {
             if (done) ZIO.fail(None)
             else
               (for {
-                reservation <- managed.reserve
+                reservation <- managed.reserve.onError(_ => doneRef.set(true))
```
iravid: Good catch! We should also add that on the `reservation.acquire`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2027,6 +2046,41 @@ object ZStream {
       } yield pull
     }
 
+  /**
+   * Creates a stream from a [[java.io.InputStream]]
+   */
+  def fromInputStream(
```
iravid: Can you port the tests for `fromInputStream` as well?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2027,6 +2046,41 @@ object ZStream {
...
+        done       <- Ref.make(false).toManaged_
+        buf        <- Ref.make(Array.ofDim[Byte](chunkSize)).toManaged_
+        capturedIs <- Managed.effectTotal(is)
+        pull = {
+          def go: IO[Option[IOException], Chunk[Byte]] = done.get.flatMap {
+            if (_) Pull.end
+            else
+              for {
+                bufArray  <- buf.get
+                bytesRead <- IO.effect(capturedIs.read(bufArray)).refineToOrDie[IOException].mapError(Some(_))
```
iravid:
```suggestion
                bytesRead <- blocking.effectBlockingInterruptible(capturedIs.read(bufArray)).refineToOrDie[IOException].mapError(Some(_))
```

iravid:
I'll follow up on my comments in `feature/zconduit`. Thanks @regiskuckaertz!

PR #3132, opened 2020-03-14
ZStream: collectAll, run, effectAsync*

regiskuckaertz:
~~Not sure why but `effectAsyncM` does not compile with a weird type mismatch error 🤔~~

Had to change the callback in `effectAsyncM` to produce an action that cannot fail. Though I could revert that and use it to define the value of `done`


streams/shared/src/main/scala/zio/stream/experimental/Take.scala
```diff
@@ -16,61 +16,104 @@
...
+import zio.{ Cause, Chunk, ZIO }
 
 /**
- * A `Take[E, B, A]` represents a single `take` from a queue modeling a stream of
+ * A `Take[E, A]` represents a single `take` from a queue modeling a stream of
  * values. A `Take` may be a failure cause `Cause[E]`, an element value `A`
- * or an end-of-stream marker `B`.
+ * or an end-of-stream marker.
  */
-sealed trait Take[+E, +B, +A] extends Product with Serializable { self =>
```
regiskuckaertz: TBH I'm not sure `Take` is still useful so far, it was just weird in the `effectAsync` that I could just write `unsafeRun(output.offer(k))`. But I could do that, what do you think @iravid ?

iravid: Let's get rid of it and use `Exit[Option[E], Chunk[A]]` instead.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1996,6 +2012,128 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+      } yield pull
+    }
+
+  /**
+   * Creates a stream from an asynchronous callback that can be called multiple times
+   * The registration of the callback itself returns an effect. The optionality of the
+   * error type `E` can be used to signal the end of the stream, by setting it to `None`.
+   */
+  def effectAsyncM[R, E, A](
+    register: (ZIO[R, Option[E], Chunk[A]] => Unit) => ZIO[R, Nothing, Any],
```
iravid: Right, you can work around the `ZManaged[R, Nothing` problem and revert this back to `ZIO[R, E, Any]` by formulating this as:
```scala
managed {
  // everything inside ZStream { ...}
}.flatMap(repeatEffectChunkOption(_))
```

regiskuckaertz: Done

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1388,15 +1388,31 @@ abstract class ZStream[-R, +E, +O](
...
+  def run[R1 <: R, E1 >: E, O1 >: O, B](sink: ZSink[R1, E1, O1, B]): ZIO[R1, E1, B] =
+    (process <*> sink.push).use {
+      case (pull, push) =>
+        def go: ZIO[R1, E1, B] = pull.foldCauseM(
+          Cause
+            .sequenceCauseOption(_)
+            .fold(
+              push(None).foldCauseM(
+                Cause.sequenceCauseEither(_).fold(IO.halt(_), IO.succeedNow),
+                _ => IO.dieMessage("empty stream / empty sinks")
```
regiskuckaertz: Couldn't find a way around this one, when the stream sends a stop signal and the sink does not produce a Z, it's clearly a bug.

iravid: Yeah. That was the whole premise of SignalFunction, no?

regiskuckaertz: Yes. I think this will be the only place where this happens though.

iravid: I think it's a reasonable price to pay to not have to carry around SignalFunction :-)

Won't it happen with all sinks though?

regiskuckaertz: Hmm, I'm not so sure. I wrote `collectAll` and `fromEffect` with two different styles: in the first, I keep track of the state of the machine and will never process elements once it's done, while in the latter I trust that the stream will tell me when things are over. I think the latter is better, it simplifies implementations by a lot. Still an implicit contract, but now all the responsibility is in `ZStream`.

iravid:
@regiskuckaertz Any hints about the ScalaJS failures?

iravid:
Merged in 6207b2fd

regiskuckaertz:
@iravid sorry I got sidetracked with the `bufferSliding` test. Great! as discussed on Discord, I'll see if somehow we can use `ZIO.effectAsync*` under the hood.

PR #3136, opened 2020-03-14
ZIO Mock handle polymorphic methods

### LINKED ISSUE #2978: Mockable - unable to mock methods with type parameters
Assuming we have a method on a module similar to this:

`def put(key: String, value: String): UIO[Unit]`

we can easily create a capability tag and mockable implementation:

`object put extends Method[Any, (String, String), Unit]`

However this doesn't work if we change the method to:

`def putK[K: Show](key: K, value: String): UIO[Unit]`

I was able to create a capability tag using a class:

`class putK[K: Show] extends Method[Any, (K, String), Unit]`

However I then always get an expectation failure

Please see the attached test case. Is there a workaround for this? Did I misunderstand something?
[zio-testcase.zip](https://github.com/zio/zio/files/4266470/zio-testcase.zip)

ioleo:
This is my inital take on polymorphic methods. I have not had enough time to thoroughly explore if there are no undesired side effects, so I'm putting this out and will come back to this tomorrow.

What is still missing is updateing the `@mockable` macro to generate proper tags for polymorphic methods, but the initial tests (see `MockSpec`) look promising.

Fixes #2978.


ioleo:
Rebased, fix for dotty and added more test coverage.

Interesting findings: needed to hint the compiler with value type in fully polymorphic case, but only in the `combined` test (when combining multiple polymorphic expectations), otherwise it dies with ClassCastException (as it infers `Nothing` and tries to cast the value to it).

I'm not sure if we can do much about it, but I guess at least we could catch the ClassCastException at higher level and turn it into a nice error message telling the developer that he should add type hints for the compiler.

ioleo:
Almost ready. Finishing touches on `@mockable` macro. Lessons learned: polymorphism and typechecking are tricky.

ioleo:
It seems like the generated output of macro is correct, however for some reason the compiler fails to recognize the mocked methods do implement the service interface. See #zio-macros on discord for details.

ioleo:
@adamgfraser @jdegoes all set and ready! Just one problem with dotty:
the ClassTag inference is a bug in the compiler [dotty#4742](https://github.com/lampepfl/dotty/issues/4742), but we need to make it compile somehow, I wonder how can we resolve that? @neko-kai ideas? perhaps we could also have izumi Tags for dotty?

docs/howto/mock_services.md
```diff
@@ -252,24 +251,77 @@ For each built-in ZIO service you will find their mockable counterparts in `zio.
...
 
+We can combine our expectation to build complex scenarios using combinators defined in `zio.test.mock.Expectation`:
+
+- `andThen` (alias `++`) Compose two expectations, producing a new expectation to **satisfy both sequentially**.
+- `and` (alias `&&`) Compose two expectations, producing a new expectation to **satisfy both in any order**.
+- `or` (alias `||`) Compose two expectations, producing a new expectation to **satisfy only one of them**.
+- `repeated` Repeat expectation within given bounds, produces a new expectation to **satisfy itself sequentially given number of times**.
+- `atLeast` Lower-bounded variant of `repeated`, produces a new expectation to satisfy **itself sequentially at least given number of times**.
+- `atMost` Upper-bounded variant of `repeated`, produces a new expectation to satisfy **itself sequentially at most given number of times**.
+- `optional` Aliast for `atMost(1)`, produces a new expectation to satisfy **itself at most once**.
```
ioleo: @ self => fix typo in `Aliast => Alias`

test/shared/src/main/scala/zio/test/mock/Result.scala
```diff
@@ -19,15 +19,20 @@ package zio.test.mock
...
   val io: I => IO[E, A]
 }
 
-object ReturnExpectation {
+object Result {
 
-  private[mock] final case class Succeed[-I, +A](io: I => UIO[A])      extends ReturnExpectation[I, Nothing, A]
-  private[mock] final case class Fail[-I, +E](io: I => IO[E, Nothing]) extends ReturnExpectation[I, E, Nothing]
+  protected[mock] final case class Succeed[-I, +A](io: I => UIO[A]) extends Result[I, Nothing, A] { self =>
+    def of[E]: Result[I, E, A] = self.asInstanceOf[Result[I, E, A]]
```
ioleo: @ self => do we need this? revisit mixed example, for all other cases it seems this is not necessary

test/shared/src/main/scala/zio/test/mock/Result.scala
```diff
@@ -19,15 +19,20 @@ package zio.test.mock
...
+  }
+
+  protected[mock] final case class Fail[-I, +E](io: I => IO[E, Nothing]) extends Result[I, E, Nothing] { self =>
+    def of[A]: Result[I, E, A] = self.asInstanceOf[Result[I, E, A]]
```
ioleo: @ self => same as above

test/shared/src/main/scala/zio/test/mock/internal/ProxyFactory.scala
```diff
@@ -164,11 +172,18 @@ object ProxyFactory {
                         if (updatedChild.saturated) resetTree(updatedChild)
                         else updatedChild
 
+                      def inRange(value: Int): Boolean =
+                        if (range.end != -1) range contains value
+                        else {
+                          val fakeUnboundedRange = range.start to Int.MaxValue by range.step
```
ioleo: @ self => we could use Long internally instead of Int for the counter and have a formula like `(value - range.start) % range.step == 0` to check if its satisfied, therefore moveing the limit from Int.MaxValue to Long.MaxValue for open-ended repetition upper bounds, however letting the test repeat `Int.MaxValue` times is already an overkill, I can't imagine ATM a scenario where a higher limit is useful

neko-kai:
@ioleo 
> @neko-kai ideas? perhaps we could also have izumi Tags for dotty?

Yes, we could, the effort is underway, you can contribute to https://github.com/7mind/dotty-typetag-research - but there are also upstream blockers currently, missing APIs in tasty-reflect: https://github.com/lampepfl/dotty/issues/8514

> Just one problem with dotty: the ClassTag inference is a bug in the compiler dotty#4742,

The ticket states it's fixed on master, why not try to switch to current latest nightly build?

ioleo:
Aparently in latest nightly its not fixed.

neko-kai:
@ioleo 
Can you reopen the issue (https://github.com/lampepfl/dotty/issues/4742) in dotty and/or sound alarms in dotty gitter? Seems like ClassTag[Nothing] is pretty important to have. Also can you try to workaround by declaring an explicit `implicit val classtagNothing: ClassTag[Nothing] = ...` in the files that don't compile for dotty?

neko-kai:
You could also try replacing ClassTag with a custom summoner with a fallback for Nothing:

```scala
final class ClassTagWrapper[A](cls: ClassTag[A])

object ClassTagWrapper extends LowPrio {
  implicit def fromClassTag[T: ClassTag] = new ClassTagWrapper(scala.reflect.classTag[T])
}
trait LowPrio {
  implicit val clsNothing: ClassTagWrapper[Nothing] = ...
}

**Comment by ioleo** [comment]
@neko-kai I just noticed there are newer nightlies published under `dotty-compiler_0.24`. Testing.

**Comment by neko-kai** [comment]
Regarding dotty blockers for izumi-reflect, you may upvote or write a comment on these issues to bring attention for them to the dotty team:

https://github.com/lampepfl/dotty/issues/8514
https://github.com/lampepfl/dotty/issues/8520
https://github.com/lampepfl/dotty/issues/8521

regarding `ClassTag[Nothing]` we can also just summon @smarter here :) From what I gather the change in the linked issues it *seems* like the change is by design, but it also clearly breaks downstream users...

**Comment by smarter** [comment]
> regarding ClassTag[Nothing] we can also just summon @smarter here 

`ClassTag[Nothing]` is bad because it means you can instantiate an `Array[Nothing]` which breaks soundness given the rules we have for erasing arrays: https://github.com/lampepfl/dotty/issues/1730. So I'm very glad that it's gone. What are you using it for anyway?

**Comment by smarter** [comment]
Also regarding the tasty-reflect issues you mentioned: please open PRs to add the stuff you want to the tasty-reflect API, there's only one person working on tasty-reflect who cannot do all the work himself. It just requires figuring out what method in the compiler does what you want, then adding a forwarder to that method in https://github.com/lampepfl/dotty/blob/master/library/src/scala/tasty/Reflection.scala

**Comment by ioleo** [comment]
@smart-ass We are using ClassTag to overcome type erasure of generic parameters.

Our effect type `ZIO[R, E, A]` has 3 type parameters that correspond to:
* `R` dependencies/environment required to run the effect
* `E` type of error in case of failure
* `A` type of value in case of success

Now, if an effect **can't fail** (it either succeeds or computes forever) we represent that as type `ZIO[R, Nothing, A]`, becouse you can't construct a value of type `Nothing`.

**Comment by ioleo** [comment]
@neko-kai Regarding the "issue fixed in master" upon rereading the thread I see Martin was talking about the original issue, which is diffrent case from ours:

```scala
def foo[T <: String: ClassTag](f: T => Int) = 1
def bar(f: String => Int) = foo(f)
                                  ^
                                  No ClassTag available for Nothing
```

smarter:
> becouse you can't construct a value of type Nothing.

And since you can't construct a value of type `Nothing` there is no legitimate JVM class that corresponds to that type, so it doesn't make sense for `ClassTag[Nothing]` to exist.

ioleo:
@smarter Yes, but the reason we're asking for `ClassTag[Nothing]` is becouse we are asking for `ClassTag` for generic type `E`, eg. when constructing a `Method` tag (see _Method.scala_ in this PR), and we need it to differentiate between diffrent instances of the same polymorphic method, without it erasure kicks in and we're screwed (at runtime).

smarter:
I see you're using `taggedIsSubtype` which is defined for dotty as:
```scala
  private[zio] def taggedIsSubtype[A, B](left: TagType, right: TagType): Boolean =
    right.runtimeClass.isAssignableFrom(left.runtimeClass)
```
Assignability of runtime classes is a separate concept from subtyping, so even if there was a `ClassTag[Nothing]`, this would never do what you want. If you want accurate subtype checks at runtime you need to run the whole compiler at runtime which is extremely slow. I don't know what you're trying to do, but I would caution against going further in this direction, I think it's a dead-end.

core/shared/src/main/scala-dotty/zio/VersionSpecific.scala
```diff
@@ -1,15 +1,29 @@
...
+  // Workaround for disabled summon for ClassTag[Nothing]
+  // due to https://github.com/lampepfl/dotty/issues/1730
+  final class ClassTagBox[A](val cls: ClassTag[A])
+
+  object ClassTagBox extends LowPrio {
+    implicit def fromClassTag[T](implicit tag: ClassTag[T]): ClassTagBox[T] = new ClassTagBox(tag)
+  }
+
+  abstract class LowPrio {
+    implicit val classTagNothing: ClassTagBox[Nothing] = new ClassTagBox(ClassTag(classOf[Nothing]))
```
smarter: Maybe next we'll disallow `classOf[Nothing]` to prevent this kind of misuse.

ioleo:
Updated code providing manually `ClassTag[Nothing]` to make the code compile, however this just revealed that `PolyMockSpec` tests fail on dotty. I'll see tomorrow if I can do something about it, also I'll see if I can help with issues @neko-kai pointed and unblock izumi tags for dotty.

core/shared/src/main/scala-dotty/zio/VersionSpecific.scala
smarter: To be 100% clear, you should not use `classOf[Nothing]`:
- It's not going to work properly (`taggedIsSubtype(taggedTagType(classTagNothing), anyOtherTag)` will always return false, when it should always return true).
- It leads to the same kind of soundness hole we had previously, so we'll have to disallow it and your code will stop compiling.

neko-kai: @smarter This wouldn't make much sense, in Scala 2 `classOf` cannot fail, even for class-less types like abstract type members. There is _always_ an available erasure for any concrete type.

ioleo: I'm not sure why you insist this is misuse. `Nothing` is just a special case we're not particuralily interested in, it just has to be supported becouse it is used all over ZIO as a marker that an effect can't fail. I'll try to give it some more thought tomorrow, or maybe you could explain in more details why this is wrong?

smarter: > This wouldn't make much sense, in Scala 2 classOf cannot fail, even for class-less types like abstract type members. 

That doesn't seem to be true, I tried 2.13.1:
```scala
scala> def foo[T] = classOf[T]
                            ^
       error: class type required but T found

scala> class Foo { type A; classOf[A] }
                                   ^
       error: class type required but Foo.this.A found
```


> I'll try to give it some more thought tomorrow, or maybe you could explain in more details why this is wrong

`classOf[Nothing]` returns `scala.runtime.Nothing$` a dummy class that is never instantiated and is not a subclass or superclass of anything else. You told me you needed this ClassTag for Method, I looked at Method and saw:
```scala
      taggedIsSubtype(self.inputTag, that.inputTag) &&
      taggedIsSubtype(self.errorTag, that.errorTag) &&
      taggedIsSubtype(self.outputTag, that.outputTag)
```
and the definition of `taggedIsSubtype` is:
```scala
  private[zio] def taggedIsSubtype[A, B](left: TagType, right: TagType): Boolean =
    right.isAssignableFrom(left)
```
**This will never return true when left is classOf[Nothing], even though Nothing is a subtype of every other type** so this is wrong (it is wrong in many other ways too, there are many situations where subtyping and subclassing are different).

neko-kai: @smarter 
> That doesn't seem to be true, I tried 2.13.1:

It succeeds for all stable paths. (except type projections)

> classOf[Nothing] returns scala.runtime.Nothing$ a dummy class that is never instantiated

It's used in Scala 2 in erasure of methods returning Nothing. Also `.asInstanceOf[Nothing]` will try to cast to that class - it's not total compiler fiction, Nothing does refer to a material class currently.

```
def erasure: Nothing = ???
public scala.runtime.Nothing$ erasure()
```

Whenever working with java reflection it still makes sense to e.g. check method return type against `classOf[Nothing]`, as long as the latter exists in bytecode, so I disagree that it would make any sense to remove it.
For `ClassTag`, `Array` can use an opaque type over ClassTag that disallows bottoms without changing ClassTag or breaking binary compatibility, I'll leave it for you to judge whether that's worth the indirection, but I do think that a soundness problem with `Array` should lead to changes in `Array`, not to changes in `ClassTag`, a thing that's only tangentially related to `Array` and makes sense for as long as `scala.runtime.Nothing$` actually exists in bytecode.

> This will never return true when left is classOf[Nothing], even though Nothing is a subtype of every other type so this is wrong (it is wrong in many other ways too, there are many situations where subtyping and subclassing are different).

This is true, but there are also other cases where `isAssignableFrom` will fail when in Scala it would succeed either at runtime or at compile-time (`int`vs.`Int`vs.`Integer`). (@ioleo Clearly you'll need to add special cases in `taggedIsSubtype` for this to work, yep)

smarter: > It succeeds for all stable paths. (except type projections)

I don't understand what you mean by that.

> Whenever working with java reflection it still makes sense to e.g. check method return type against classOf[Nothing], as long as the latter exists in bytecode, so I disagree that it would make any sense to remove it.

In general, you can't rely on `classOf` to give you a type that matches exactly the result type of a method, for example a method that returns `(Int, Int)` in source code will return `scala.Tuple2$mcII$sp` in bytecode, but `classOf[(Int, Int)` returns `scala.Tuple2`.

> I do think that a soundness problem with Array should lead to changes in Array, not to changes in ClassTag, a thing that's only tangentially related to Array

At this point, I think that making array is the only justified usage of ClassTag, it's possible to use it to avoid unchecked warning, but this is actually unsound and will have to be removed: https://github.com/lampepfl/dotty/issues/7554. Otherwise it seems to be mostly used incorrectly (like in this PR, where it is mistaken as a way to do subtype checks at runtime).

> This is true, but there are also other cases where isAssignableFrom will fail when in Scala it would succeed either at runtime or at compile-time (intvs.Intvs.Integer

Exactly, you can't do subtype checks at runtime using ClassTag at all, so it doesn't matter if ClassTag[Nothing] exists or not, this is fundamentally the wrong approach.

> Clearly you'll need to add special cases in taggedIsSubtype for this to work

No amount of special cases that you add will ever make this correct in general.

neko-kai: @smarter 
> I don't understand what you mean by that.

I was wrong, `classOf` only works for class types and intersections, 

> Exactly, you can't do subtype checks at runtime using ClassTag at all, so it doesn't matter if ClassTag[Nothing] exists or not, this is fundamentally the wrong approach.

> At this point, I think that making array is the only justified usage of ClassTag, it's possible to use it to avoid unchecked warning

Fine, but the ecosystem uses ClassTag for a lot besides that, e.g. printing the class name is `pureconfig` - `ClassTag[Nothing]` is completely correct for printing.

> No amount of special cases that you add will ever make this correct in general.

It can make it correct for a narrow amount of well known cases - such as unparameterized concrete class types, which is all that's necesasry until [`izumi-reflect`](https://github.com/zio/izumi-reflect) for dotty is ready and will be able handle a larger amount of known cases.

smarter: > Fine, but the ecosystem uses ClassTag for a lot besides that

I think the ecosystem will have to switch to other typeclasses depending on their usecase. Either making their own (like Typeable in shapeless: https://github.com/milessabin/shapeless/blob/shapeless-3/core/src/main/scala/shapeless/typeable.scala, or reusing the stuff in dotty like [scala.quoted.Type](http://dotty.epfl.ch/docs/reference/metaprogramming/macros.html) (currently only usable from inside an inline def, but maybe that limitation can be lifted for globally accessible types).

ioleo:
@smarter Thank you for explaining in details :pray: I've updated Tagged implementation for dotty such that `ClassTag[Nothing]` is never used. I see it already improved a lot, we're down to one failing test. @neko-kai could you review Tagged changes?

ioleo:
So indeed, as @smarter pointed:
> In general, you can't rely on classOf to give you a type that matches exactly the result type of a method, for example a method that returns (Int, Int) in source code will return scala.Tuple2$mcII$sp in bytecode, but classOf[(Int, Int)] returns scala.Tuple2.
The failing suite deals with Tuples and therefore `isAssignableFrom` indeed diverges from `<:<` relationship. I had quick look around `ClassTag` API, but none of the other methods available can help to "fix".

To progress from here I'm marking this test as ignored on dotty (and add appropriate comment). Since there are not many dotty users yet (and dotty is itself not stable) I think it's OK for now to merge it "as is", knowing that for some cases it's broken on dotty.

Meanwhile I'll try to get involved with dotty and submit PRs for missing tasty API parts that are blocking izumi tags support.

PR #3142, opened 2020-03-15
Add assertions to zio-test

Michaelt293:
Added the assertions `isPositive`, `nonNegative`, `isNegative`, `nonPositive`, `isZero`, `equalToOneOf`, `containsNoneOf`, `containsOneOf`, `containsAtLeastOneOf` and `hasAllElements` to zio-test.

Scalafmt is currently giving the following error for `Assertion.scala` so unable to format the code (not caused by the changes in this PR) -
```
Parse error: <input>:31: error: this type can't be used in a constructor call ) extends ((=> A) => AssertResult) { self => ^
```

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/TestUtils.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestUtils.scala
@@ -23,7 +23,7 @@ object TestUtils {
     }
   }
 
-  def isSuccess[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
+  def isSucceeded[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
     val execSpec = execute(spec)
     forAllTests(execSpec) {
       case Right(TestSuccess.Succeeded(_)) => true
--- a/test-tests/shared/src/test/scala/zio/test/TestUtils.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestUtils.scala
@@ -23,7 +23,7 @@ object TestUtils {
     }
   }
 
-  def isSucceeded[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
+  def succeeded[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
     val execSpec = execute(spec)
     forAllTests(execSpec) {
       case Right(TestSuccess.Succeeded(_)) => true
--- a/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
+++ b/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
@@ -20,14 +20,17 @@ import zio.test.Assertion.Render._
 
 trait AssertionVariants {
 
+  final def equals[A, B](a: A, b: B)(implicit eql: Eql[A, B]): Boolean =
+    (a, b) match {
+      case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
+      case (left, right)                     => left == right
+    }
+
   /**
    * Makes a new assertion that requires a value equal the specified value.
    */
   final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] =
     Assertion.assertion("equalTo")(param(expected)) { actual =>
-      (actual, expected) match {
-        case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
-        case (left, right)                     => left == right
-      }
+      equals(expected, actual)
     }
 }
--- a/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
+++ b/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
@@ -20,8 +20,8 @@ import zio.test.Assertion.Render._
 
 trait AssertionVariants {
 
-  final def equals[A, B](a: A, b: B)(implicit eql: Eql[A, B]): Boolean =
-    (a, b) match {
+  final def equals[A, B](b: B, a: A)(implicit eql: Eql[A, B]): Boolean =
+    (b, a) match {
       case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
       case (left, right)                     => left == right
     }
@@ -30,5 +30,5 @@ trait AssertionVariants {
    * Makes a new assertion that requires a value equal the specified value.
    */
   final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] =
-    Assertion.assertion("equalTo")(param(expected))(actual => equals(expected, actual))
+    Assertion.assertion("equalTo")(param(expected))(actual => equals(actual, expected))
 }
--- a/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
+++ b/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
@@ -20,15 +20,14 @@ import zio.test.Assertion.Render._
 
 trait AssertionVariants {
 
-  final def equals[A, B](b: B, a: A)(implicit eql: Eql[A, B]): Boolean =
-    (b, a) match {
-      case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
-      case (left, right)                     => left == right
-    }
-
   /**
    * Makes a new assertion that requires a value equal the specified value.
    */
   final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] =
-    Assertion.assertion("equalTo")(param(expected))(actual => equals(actual, expected))
+    Assertion.assertion("equalTo")(param(expected)) { actual =>
+      (actual, expected) match {
+        case (left: Array[_], right: Array[_]) => left.sameElements[Any](right)
+        case (left, right)                     => left == right
+      }
+    }
 }
--- a/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
+++ b/test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
@@ -30,7 +30,5 @@ trait AssertionVariants {
    * Makes a new assertion that requires a value equal the specified value.
    */
   final def equalTo[A, B](expected: A)(implicit eql: Eql[A, B]): Assertion[B] =
-    Assertion.assertion("equalTo")(param(expected)) { actual =>
-      equals(expected, actual)
-    }
+    Assertion.assertion("equalTo")(param(expected))(actual => equals(expected, actual))
 }
--- a/test/shared/src/main/scala/zio/test/Assertion.scala
+++ b/test/shared/src/main/scala/zio/test/Assertion.scala
@@ -387,11 +387,11 @@ object Assertion extends AssertionVariants {
     )
 
   /**
-   * Makes a new assertion that requires an Iterable to have all of the elements
+   * Makes a new assertion that requires an Iterable to have the same distinct elements
    * as the other Iterable, though not necessarily in the same order
    */
-  def hasAllOf[A](other: Iterable[A]): Assertion[Iterable[A]] =
-    Assertion.assertion("hasAllOf")(param(other))(actual => actual.toSet == other.toSet)
+  def hasSameElementsDistinct[A](other: Iterable[A]): Assertion[Iterable[A]] =
+    Assertion.assertion("hasSameElementsDistinct")(param(other))(actual => actual.toSet == other.toSet)
 
   /**
    * Makes a new assertion that requires a sequence to contain an element
--- a/test/shared/src/main/scala/zio/test/Assertion.scala
+++ b/test/shared/src/main/scala/zio/test/Assertion.scala
@@ -452,8 +452,8 @@ object Assertion extends AssertionVariants {
     }
 
   /**
-   * Makes a new assertion that requires a Map to has the specified key
-   * With value matching the specified assertion.
+   * Makes a new assertion that requires a Map to have the specified key
+   * with value satisfying the specified assertion.
    */
   def hasKey[K, V](key: K, assertion: Assertion[V]): Assertion[Map[K, V]] =
     Assertion.assertionRec("hasKey")(param(key))(assertion)(_.get(key))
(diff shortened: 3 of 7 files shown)


test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -282,13 +282,94 @@ object Assertion extends AssertionVariants {
...
    * satisfying an assertion.
    */
   def contains[A](element: A): Assertion[Iterable[A]] =
     Assertion.assertion("contains")(param(element))(_.exists(_ == element))
 
+  /**
+   * Makes a new assertion that requires an Iterable contain none of the
+   * specified elements.
+   */
+  def containsNoneOf[A](value: A, other: A, remaining: A*): Assertion[Iterable[A]] = {
```
adamgfraser: Maybe we could have a `hasIntersection` assertion that lets you assert that the intersection of two collections satisfies a particular assertion. Then you could implement these different variants in terms of that. I think I would also just have the parameter be a collection instead of using variable arguments.

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
```diff
@@ -20,14 +20,15 @@ import zio.test.Assertion.Render._
 
 trait AssertionVariants {
 
+  final def equals[A, B](b: B, a: A)(implicit eql: Eql[A, B]): Boolean =
```
adamgfraser: Does this actually give you any type safety? I think the compiler is just going to widen the types here.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -282,13 +282,94 @@ object Assertion extends AssertionVariants {
...
+  /**
+   * Makes a new assertion that requires a numeric value is zero.
+   */
+  def isZero[A](implicit num: Numeric[A]): Assertion[A] =
+    Assertion.assertion("isZero")()(_ == num.zero)
+
+  /**
+   * Makes a new assertion that requires a value to be equal to one of the specified values.
+   */
+  def equalToOneOf[A, B](value: A, other: A, remaining: A*)(implicit eql: Eql[A, B]): Assertion[B] = {
```
adamgfraser: Maybe `isOneOf`? I would make this take a collection instead of using variable arguments.

Michaelt293: Good idea. For the collection type, I think `Set` would be a good idea to avoid passing in duplicate elements, i.e. `isOneOf(List(1, 1, 2))`.

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
Michaelt293: I believe the main reason for defining `equals` was so that `equalTo` and `equalToOneOf` had the same behaviour when testing equality of arrays.

test/shared/src/main/scala/zio/test/Assertion.scala
Michaelt293: I like this idea. I will go ahead with this approach.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -437,7 +518,35 @@ object Assertion extends AssertionVariants {
...
+      val actualSeq = actual.toSeq
+      val otherSeq  = other.toSeq
+
+      otherSeq.diff(actualSeq).isEmpty
+    }
+
+  /**
+   * Makes a new assertion that requires a Seq is sorted.
+   */
+  def isSorted[A](implicit ord: Ordering[A]): Assertion[Seq[A]] =
```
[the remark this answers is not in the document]
Michaelt293: Agree, will go ahead and write as a recursive loop. Similarly, I was thinking that another assertion `isDistinct` might also be another good addition.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -282,13 +282,94 @@ object Assertion extends AssertionVariants {
...
-   * element. See [[Assertion.exists]] if you want to require an iterable to contain an element
+   * Makes a new assertion that requires a numeric value is positive.
+   */
+  def isPositive[A](implicit num: Numeric[A]): Assertion[A] =
+    Assertion.assertion("isPositive")()(actual => num.gt(actual, num.zero))
+
+  /**
+   * Makes a new assertion that requires a numeric value is non negative.
+   */
+  def nonNegative[A](implicit num: Numeric[A]): Assertion[A] =
```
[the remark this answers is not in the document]
Michaelt293: I initially wrote these numeric assertions in this way. I changed them so that the name in the error message could be more specific. If this is not a concern, I will go with this approach.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -437,7 +518,19 @@ object Assertion extends AssertionVariants {
     }
 
   /**
-   * Makes a new assertion that requires the size of an iterable be satisfied
+   * Makes a new assertion that requires an Iterable to have all the elements
+   * as the specified Iterable, though not necessarily in the same order
+   */
+  def hasAllElements[A](other: Iterable[A]): Assertion[Iterable[A]] =
```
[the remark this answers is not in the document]
Michaelt293: With this assertion, the other iterable should be a subset of the tested iterable.

test/shared/src/main/scala-2.x/zio/test/AssertionVariants.scala
adamgfraser: If you just want to do that then I would suggest defining a helper method that takes a single type parameter. Using the multiple type parameters and the `Eql` is trying to do some more type level machinery for the `equalTo` assertion in particular so if we just want to handle arrays we can simplify substantially.

test/shared/src/main/scala/zio/test/Assertion.scala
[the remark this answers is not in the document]
adamgfraser: Yes that would be great!

test/shared/src/main/scala/zio/test/Assertion.scala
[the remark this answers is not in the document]
adamgfraser: One thing we could explore here is adding a `relabel` combinator that lets you replace the existing label with a new one. Then we could implement these assertions in terms of other ones and still have the more specific names.

test/shared/src/main/scala/zio/test/Assertion.scala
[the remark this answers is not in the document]
adamgfraser: I see. Maybe `hasSubset`? I think this could be implemented in terms of the intersection operator as well, basically asserting that the intersection of the actual and `other` is equal to `other`.

test/shared/src/main/scala/zio/test/Assertion.scala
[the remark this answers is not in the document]
Michaelt293: I like the idea to introduce a new combination to allow the label to be replaced. I will experiment with this idea.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -398,6 +405,20 @@ object Assertion extends AssertionVariants {
       }
     }
 
+  /**
+   * Makes a new assertion that requires an Iterable contain at least one of the
+   * specified elements.
+   */
+  def hasAtLeastOneOf[A](values: Set[A]): Assertion[Iterable[A]] =
```
adamgfraser: Since `intersection` is defined for types other than `Set` should we change the `values` type to `Iterable` here?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -482,6 +557,22 @@ object Assertion extends AssertionVariants {
   def isFalse: Assertion[Boolean] =
     Assertion.assertion("isFalse")()(!_)
 
+  /**
+   * Makes a new assertion that requires a Failure value satisfying a specified
+   * assertion.
+   */
+  def isFailure(assertion: Assertion[Throwable]): Assertion[Try[Any]] =
```
adamgfraser: Great!

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -563,13 +678,45 @@ object Assertion extends AssertionVariants {
...
+   */
+  val isSome: Assertion[Option[Any]] =
+    isSome(anything)
+
+  /**
+   * Makes a new assertion that requires a Seq is sorted.
+   */
+  def isSorted[A](implicit ord: Ordering[A]): Assertion[Seq[A]] = {
+    @scala.annotation.tailrec
+    def loop(seq: Seq[A]): Boolean = seq match {
```
adamgfraser: May want to use iterator here since some collections don't support efficient restructuring like this. That could also allow us to expand input type to `Iterable`. Some iterables like sets or maps don't have a well defined ordering but others like sorted sets and maps do, so could be nice to let users test this. What do you think?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -600,12 +763,30 @@ object Assertion extends AssertionVariants {
...
    * Makes a new assertion that requires a given string to match the specified regular expression.
    */
   def matchesRegex(regex: String): Assertion[String] =
     Assertion.assertion("matchesRegex")(param(regex))(_.matches(regex))
 
+  /**
+   * Makes a new assertion that requires a numeric value is non negative.
+   */
+  def nonNegative[A](implicit num: Numeric[A]): Assertion[A] =
+    Assertion.assertion("nonNegative")()(actual => num.gteq(actual, num.zero))
```
adamgfraser: `isGreaterThanEqualTo(0)`?

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -541,18 +650,24 @@ object Assertion extends AssertionVariants {
...
   /**
    * Makes a new assertion that requires a null value.
    */
   val isNull: Assertion[Any] =
     Assertion.assertion("isNull")()(_ == null)
 
+  /**
+   * Makes a new assertion that requires a value to be equal to one of the specified values.
+   */
+  def isOneOf[A, B](values: Set[A])(implicit eql: Eql[A, B]): Assertion[B] =
```
adamgfraser: Maybe this can take an `Iterable`? Not sure what the harm is if the user duplicates elements and may just be more flexible for the caller. Also would suggest removing the `eql` here as is significantly complicated the type signature and don't think it buys us additional type safety here. If you're worried about arrays we could add an additional helper method for that.

test/shared/src/main/scala/zio/test/Assertion.scala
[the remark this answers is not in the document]
Michaelt293: I have now added `hasSubset` using `hasIntersection`.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -437,7 +484,14 @@ object Assertion extends AssertionVariants {
     }
 
   /**
-   * Makes a new assertion that requires the size of an iterable be satisfied
+   * Makes a new assertion that requires an Iterable to have the same unique elements
+   * as the specified Iterable, though not necessarily in the same order
+   */
+  def hasSameUniqueElements[A](values: Set[A]): Assertion[Iterable[A]] =
```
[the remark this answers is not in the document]
Michaelt293: Changed as suggested. Now uses an `Iterable` instead of a `Set`.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -464,6 +525,20 @@ object Assertion extends AssertionVariants {
   ): Assertion[Sum] =
     Assertion.assertionRec("isCase")(param(termName), param(unapply(termName)), param(assertion))(assertion)(term(_))
 
+  /**
+   * Makes a new assertion that requires a Seq is distinct.
+   */
+  val isDistinct: Assertion[Seq[Any]] = {
```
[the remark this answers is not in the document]
Michaelt293: Has now been changed to accept an `Iterable`.

test-tests/shared/src/test/scala/zio/test/TestUtils.scala
```diff
@@ -23,7 +23,7 @@ object TestUtils {
     }
   }
 
-  def isSuccess[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
+  def isSucceeded[E](spec: ZSpec[environment.TestEnvironment, E]): ZIO[Any, Nothing, Boolean] = {
```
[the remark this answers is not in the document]
Michaelt293: I have changed `isSucceeded` to `succeeded` but ran into a naming collision when trying to use `ignored`.

test/shared/src/main/scala/zio/test/Assertion.scala
Michaelt293: All assertions based on the intersection of two iterables have now been changed to accept an `Iterable`.

test/shared/src/main/scala/zio/test/Assertion.scala
Michaelt293: I have re-factored this code to use an iterator.

test/shared/src/main/scala/zio/test/Assertion.scala
Michaelt293: This has been changed to accept an `Iterable`. The implicit parameter was left by mistake and has now been removed.

test/shared/src/main/scala/zio/test/Assertion.scala
```diff
@@ -385,6 +386,13 @@ object Assertion extends AssertionVariants {
       BoolAlgebraM.success
     )
 
+  /**
+   * Makes a new assertion that requires an Iterable to have all of the elements
+   * as the other Iterable, though not necessarily in the same order
+   */
+  def hasAllOf[A](other: Iterable[A]): Assertion[Iterable[A]] =
```
adamgfraser: I'm still a little confused on the contract for this. Is it just supposed to be a version of `hasSameElements` that ignores the number of times an element appears? I think we may also need to think about the name as I confused it with `hasSubset`.

Michaelt293: Yes, that is correct. Perhaps we can use the name `hasSameDistinctElements `? The name is a bit longer but I think it is more descriptive. The other option is that we simply remove this assertion if there is going to be confusion on the behaviour of these assertions.

adamgfraser: How about `hasSameElementsDistinct`? That will be more discoverable by IDEs.

Michaelt293: I have changed the assertion to `hasSameElementsDistinct`. Please let me know if there's anything I need to do before this can be merged into master.

PR #3144, opened 2020-03-15
Add stream buffer* methods

regiskuckaertz:


regiskuckaertz:
hmm, in the sliding test the queue goes below capacity 🤔 I consistently get the 9 and 17 dropped off the queue and never making it to the consumer side.

vasilmkd:
I'm still puzzled why this is the case...

regiskuckaertz:
Will look into it in more detail this week-end ✌️

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -183,6 +183,73 @@ abstract class ZStream[-R, +E, +O](
       } yield pull
     }
 
+  private final def bufferSignal[E1 >: E, O1 >: O](
```
vasilmkd: Sorry, @regiskuckaertz, but how does this code deal with resuming after error? I believe you have removed the mechanism for that. Doesn't this just stop after it encounters the first error?

regiskuckaertz: I was looking at `upstream` and it cannot fail from the inside, whereas in downstream, the `pull` intercepts any error and set the flag `done` to true. After that, `pull` will always yield `Pull.End`. Is my reasoning correct? I'll add safety tests to make sure of it! But this was a test to figure out why bufferSliding is behaving differently, and it seems that downstream is pulling the end signal which blows up the test with an error of `None` as it should. But I wonder why it is pulling that end signal in the first place 🤔

vasilmkd: I'm talking about the `Promise` for each element that paused the pipeline until the error has been handled downstream.

regiskuckaertz: Ha, yes that's the thing I missed. Thanks : 🙏

vasilmkd:
I see we haven't copied the `StreamPullSafetySpec` tests. We need them too.

regiskuckaertz:
There was nothing wrong in the `bufferSliding`, the problem was in `takeWhile` (used in `range`) that didn't signal the end of the stream with a failure but with an empty chunk _then_ a failure.

regiskuckaertz:
that empty chunk was making the head of the queue drop the 9. or 17.

regiskuckaertz:
@iravid hiya - hope you're all well! can you please review when you've got some time?

iravid (approved):
Looks great @regiskuckaertz. Thanks for picking this up!

PR #3145, opened 2020-03-15
Add *WithTrace variants for error handling combinators

### LINKED ISSUE #2323: Error traces are not accessible via standard combinators
I want to log error traces for better diagnostics, but combinators like `catchAll`/`catchSome`/`tapError`/`foldM`/etc strip traces off and their `Cause`-aware cousins like `catchSomeCause` catch many things besides errors.

Ideally, I would like to have something along the lines
```scala
final case class ErrorWithTrace[E](error: E, trace: Option[ZTrace])
object Error {
  def unapply[E](e: ErrorWithTrace[E]): Option[E] = Some(e.error)
}

foo.catchSomeWithTrace {
  case e @ Error(MyErrorExtractor(...)) => log(e, "message")
}
```

mvv:
Fixes #2323 by providing `*WithTrace` variants of error handling combinators. The `Cause.failureWithTraceOption` should probably be extended to handle at least `Traced(Meta(...Meta(Fail(e), ...)), trace)` cases, maybe more (does `Traced(Traced(Fail(e), trace2), trace1)` make sense?). Would appreciate some input on this.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -77,7 +77,7 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
    * Returns the `E` associated with the first `Fail` in this `Cause` if one
    * exists, along with its (optional) trace.
    */
-  def failureWithTraceOption: Option[(E, Option[ZTrace])] =
+  def failureTraceOption: Option[(E, Option[ZTrace])] =
     find {
       case Traced(Fail(e), trace) => (e, Some(trace))
       case Fail(e)                => (e, None)
@@ -98,7 +98,7 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
    * if there are no checked errors return the rest of the `Cause`
    * that is known to contain only `Die` or `Interrupt` causes.
    * */
-  final def failureWithTraceOrCause: Either[(E, Option[ZTrace]), Cause[Nothing]] = failureWithTraceOption match {
+  final def failureTraceOrCause: Either[(E, Option[ZTrace]), Cause[Nothing]] = failureTraceOption match {
     case Some(errorAndTrace) => Left(errorAndTrace)
     case None                => Right(self.asInstanceOf[Cause[Nothing]]) // no E inside this cause, can safely cast
   }
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -74,6 +74,16 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
     find { case Fail(e) => e }
 
   /**
+   * Returns the `E` associated with the first `Fail` in this `Cause` if one
+   * exists, along with its (optional) trace.
+   */
+  def failureWithTraceOption: Option[(E, Option[ZTrace])] =
+    find {
+      case Traced(Fail(e), trace) => (e, Some(trace))
+      case Fail(e)                => (e, None)
+    }
+
+  /**
    * Retrieve the first checked error on the `Left` if available,
    * if there are no checked errors return the rest of the `Cause`
    * that is known to contain only `Die` or `Interrupt` causes.
@@ -81,6 +91,16 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
   final def failureOrCause: Either[E, Cause[Nothing]] = failureOption match {
     case Some(error) => Left(error)
     case None        => Right(self.asInstanceOf[Cause[Nothing]]) // no E inside this cause, can safely cast
+  }
+
+  /**
+   * Retrieve the first checked error and its trace on the `Left` if available,
+   * if there are no checked errors return the rest of the `Cause`
+   * that is known to contain only `Die` or `Interrupt` causes.
+   * */
+  final def failureWithTraceOrCause: Either[(E, Option[ZTrace]), Cause[Nothing]] = failureWithTraceOption match {
+    case Some(errorAndTrace) => Left(errorAndTrace)
+    case None                => Right(self.asInstanceOf[Cause[Nothing]]) // no E inside this cause, can safely cast
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mvv:
@adamgfraser just `tapCause` won't do, because trace dumping is often conditional and tied to recovery, e.g.
```scala
  foo.catchSomeWithTrace {
    case (e: BarError, trace) =>
      log.warn(trace, s"Failed to foo with $e, falling back") *> recover
  }
```

An alternative to new combinators would be having a `Cause` extractor like
```scala
object Failure {
  def unapply[E](cause: Cause[E]): Option[E] = cause.failureOption
  // Or
  def unapply[E](cause: Cause[E]): Option[(E, Option[ZTrace])] = cause.failureWithTraceOption  
}
```
so that you could write
```scala
  foo.catchAllCause {
    case c @ Failure(e: BarError) =>
      log.warn(c, s"Failed to foo with $e, falling back") *> recover
    case c @ Failure(e) =>
      log.error(c, s"Failed to foo: $e") *> ZIO.halt(c)
  }
```
Not sure what is the proper trace to log here, is it `c.traces` or the trace tied to the `e` we have extracted (in that case the second `Failure.unapply` variant should be used)?

mvv:
@adamgfraser There is also a problem with replacing `catchAll` with `catchAllCause`. Let's say we have a
```scala
foo.catchAll {
  case e: BarError => recover1: UIO[A]
  case e: BazError => recover2: UIO[A]
}: UIO[A]
```
and now want to add logging. We can't use `catchSomeCause`, because the resulting expression won't have the desired `UIO[A]` type, but the same goes for `catchAllCause` with the extra fallback `case`:
```scala
foo.catchAllCause {
  case c @ Failure(e: BarError) => log(c, ...) *> recover1: UIO[A]
  case c @ Failure(e: BazError) => log(c, ...) *> recover2: UIO[A]
  case c => ZIO.halt(c): IO[E, Nothing] // !!!
}: IO[E, Nothing]
```
Something like
```scala
trait Failure[+E] {
  def value: E
  def fullCause: Cause[E]
} 
def catchAllFailure[E1, A1 >: A](f: Failure[E] => IO[E1, A1]): IO[E1, A1] 
```
could help solve it

adamgfraser:
@mvv You make good points.

What if we do something like:

```scala
trait ZIO[-R, +E, +A] {
  final def withTrace: ZIO[R, (E, Option[ZTrace]), A]
}
```

Basically, this "surfaces" the trace into the error channel. I am coming around to the view that the trace should be only the trace associated with the "primary" error. Then you could use all the normal error handling combinators to deal with errors and log the trace.

mvv:
@adamgfraser `withTrace` is clunky vs `catchSome`:
```scala
foo
  .withTrace
  .catchSome {
    case ... => recover // Have to be IO[E1 >: (E, Option[ZTrace]), A]
  }
  .mapError(_._1) // Have to map it back
```
Also, now that I think about it, `Tuple2[E, Option[ZTrace]]` is kinda awkward to refail/rehalt without loss of info.

adamgfraser:
@mvv Yes, there is an inverse operation that submerges the trace back into the cause. Potentially we could bundle the two together into something like:

```scala
/**
 * Performs the specified effect with the trace of the "primary" cause
 * surfaced to the error channel.
 */
def withTrace[R1, E1, A1](f: ZIO[R, (E, Option[ZTrace]), A] => ZIO[R1, E1, A1]): ZIO[R1, E1, A1] =
  f(self.catchAllCause(_.failureWithTraceOrCause.fold(ZIO.fail(_), ZIO.halt(_))))
```

That addresses your second point but not your first one since we still need our `f` to return an `E` instead of a `(E, Option[ZTrace])` which doesn't work great with `catchSome` since we need a fallback to convert the unmatched cases back to an `E`. We could address that by implementing some more specialized combinators in terms of this. For example:

```scala
def catchSomeWithTrace[R1 <: R, E1 >: E, A1 >: A](pf: PartialFunction[(E, Option[ZTrace]), ZIO[R1, E1, A1]]): ZIO[R1, E1, A1] =
  withTrace { effect =>
    effect.catchAll(pf.applyOrElse(_, (e: (E, Option[ZTrace])) => ZIO.fail(e._1)))
  }
```

mvv:
@adamgfraser So we are back to the original combinators, just with different implementation? Or do you want to skip `catchAllWithTrace`/`tapErrorWithTrace`/`foldWithTrace` and just have `withTrace` and `catchSomeWithTrace`? I don't think those two deliver the same ergonomics, e.g.
```scala
foo.tapErrorWithTrace { case (e, trace) => log(trace, s"Got error: $e" }
```
would become something like
```scala
foo.withTrace(_.tapError { case (e, trace) => log(trace, s"Got error: $e") })
```
Not much longer, but more nesting. We could also pick a less verbose name for `tapErrorWithTrace`, say, `tapFailure`, which would further increase the readability.

adamgfraser:
@mvv Yes, thanks for bearing with me on this. I've been trying to figure out the "primitive" we were missing here that we could implement other things in terms of. You've sold me on needing some specialized combinators that can hopefully be implemented in terms of this.

One thing we could do to cut down on the method length would be to do `Trace` instead of `WithTrace`. So like it would be `catchSomeTrace`.


I'm fine with having some more specialized combinators. Maybe we keep `catchAllTrace`, `catchSomeTrace`, and `tapErrorTrace` and drop the `foldM` variant?

mvv:
@adamgfraser Made the rename. I left `foldTraceM`, because I personally use `foldM` quite a bit (mainly in `Pull`s, but in general to avoid the extra allocation of `either.flatMap`). I could write something like `foo.catchAllTrace(ZIO.fail).foldM(...)`, but that's wordy.

adamgfraser:
@mvv Shall we add `foldTrace` for consistency with `foldTraceM`?

mvv:
@adamgfraser Can't really think of a use case for accessing the trace in pure function. Same goes for `mapError`, for example. Some combinators work with errors as values and others work with them as failures, that's my mental model for this. `flatMapError` and `forkWithErrorHandler` seem to be in the latter category, but they are rarely used.

adamgfraser:
Okay. I have a feeling we are going to have someone ask us for it at some point in the future but we can deal with that when the time comes. 😃

adamgfraser:
Thanks for contributing!

PR #3149, opened 2020-03-15
Implement Random#between

### LINKED ISSUE #3146: Random range functions
I feel Random might benefit from range variants for generating random numbers, like `nextInt(low, high)` (or `nextIntRange(low, high)` if you'd rather not use overloading).

Reason being for convenience considering that it's a very common operation. And to avoid mistakes like https://github.com/zio/zio/issues/3141. At a glance it's not easy to tell there's a bug there.

adamgfraser:
Resolves #3146. These methods were added to `scala.util.Random` in 2.13 to generate numeric values within a specified range. Adding them to `Random` service maintains parity between our implementation of the standard services and the ones in the Scala standard library and should also provide an easier way for people to generate values in a range.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/RandomSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/RandomSpec.scala
@@ -1,0 +1,72 @@
+package zio
+
+import zio.random.Random
+import zio.test.Assertion._
+import zio.test._
+import zio.test.environment.Live
+
+object RandomSpec extends ZIOBaseSpec {
+
+  def spec = suite("RandomSpec")(
+    testM("between generates doubles in specified range") {
+      checkM(genDoubles) {
+        case (min, max) =>
+          for {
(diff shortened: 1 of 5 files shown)


luis3m:
@adamgfraser in the `betweenWith` definitions, isn't `difference` suppose to be always greater than zero given the condition above?

adamgfraser:
@luis3m We need it to deal with overflow. For example, say `max` is `Int.MaxValue` and min is `-10`. `max` is greater than `min` in that case, but if we do `max - min` we get `-2147483639`.

Conceptually the problem is that `nextInt(n: Int)` can only generate a space of integers up to `Int.MaxValue`. But the total space of integers is up to ~2X that. So if the space of integers that the caller is requesting is larger than that we need to fall back on the normal `nextInt` method and then retry until we get a value in range (which on average should only take two times).

PR #3157, opened 2020-03-17
Add mapDependency to map over a ZLayer with Has

vpavkin:
This `mapDependency` (naming is up for suggestions) extension simplifies maps over `ZLayer` instances with `Has[_]` outputs.

An example would be a need "zoom" into a config layer to create a more specific layer with a config subset for specific component. Currently it's requires some boilerplate:

```scala
val configLayer = TypesafeConfig.fromDefaultLoader(appConfigDesc)
val dbConfigLayer = configLayer.map(c => Has(c.get.dbConfig))
val apiConfigLayer = configLayer.map(c => Has(c.get.apiConfig))
```
With this patch in the above code would look like:
```scala
val configLayer = TypesafeConfig.fromDefaultLoader(appConfigDesc)
val dbConfigLayer = configLayer.mapDependency(_.dbConfig)
val apiConfigLayer = configLayer.mapDependency(_.apiConfig)
```

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -155,6 +155,24 @@ object ZLayer {
   @deprecated("use Layer", "1.0.0")
   type NoDeps[+E, +B] = ZLayer[Any, E, B]
 
+  trait HasSingleService[A] {
+    type Out
+  }
+
+  object HasSingleService {
+    type Aux[A, Out1] = HasSingleService[A] {
+      type Out = Out1
+    }
+
+    implicit def proof[A]: Aux[Has[A], A] = new HasSingleService[Has[A]] {
+      type Out = A
+    }
+
+    implicit def multipleAmbig[A, B]: Aux[Has[A] with B, A] = new HasSingleService[Has[A] with B] {
+      type Out = A
+    }
+  }
+
   /**
    * Constructs a layer from a managed resource.
    */
@@ -2058,14 +2076,14 @@ object ZLayer {
      * Proves that this layer has a single `Has[_]` output,
      * which allows more operations to be done lawfully, for example map a value inside `Has[_]`.
      */
-    def singleService[A](implicit ev: ROut =:= Has[A]): SingleServiceOutputLayerOps[RIn, E, A] =
-      new SingleServiceOutputLayerOps[RIn, E, A](self.asInstanceOf[ZLayer[RIn, E, Has[A]]])
+    def singleService(implicit ev: HasSingleService[ROut]): SingleServiceOutputLayerOps[RIn, E, ev.Out] =
+      new SingleServiceOutputLayerOps[RIn, E, ev.Out](self.asInstanceOf[ZLayer[RIn, E, Has[ev.Out]]])
   }
 
   final class SingleServiceOutputLayerOps[RIn, E, A](private val self: ZLayer[RIn, E, Has[A]]) extends AnyVal {
 
     /**
-     * Maps this layer's single Has[A] output into a single Has[B] output.
+     * Maps this layer's single Has[A] output into a single Has[B] outpuцетаt.
      */
     def map[B: Tagged](f: A => B)(implicit A: Tagged[A]): ZLayer[RIn, E, Has[B]] =
       self.map(x => Has(f(x.get)))
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -2053,13 +2053,21 @@ object ZLayer {
      */
     def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
       self >>> ZLayer.fromFunctionMany(_.update[A](f))
+
+    /**
+     * Proves that this layer has a single `Has[_]` output,
+     * which allows more operations to be done lawfully, for example map a value inside `Has[_]`.
+     */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2055,6 +2055,14 @@ object ZLayer {
       self >>> ZLayer.fromFunctionMany(_.update[A](f))
   }
 
+  implicit final class ZLayerHasROutOps1[RIn, E, A](private val self: ZLayer[RIn, E, Has[A]]) extends AnyVal {
+    /**
+     * Maps layer output which is wrapped in `Has`
+     */
+    def mapDependency[B: Tagged](f: A => B)(implicit A: Tagged[A]): ZLayer[RIn, E, Has[B]] =
```
adamgfraser: We could potentially call this `mapService`. The issue here is that this only works for layers that output a single service. If we try to call it on layers that output more than one service we can only map over the "first" service and silently discard the rest, which violates the principle that the order of services within `Has` doesn't matter.

vpavkin: I see, makes sense. I'll dig into what can be done here to support multiple services, thanks for a good point!

adamgfraser: Definitely. I think it is going to be very challenging to do this. We would need to define a method on `Has` that would take an existing service `A`, remove it from the map, and add a new service `B`. We can definitely add and remove services from the map, but I think it may not be possible in Scala's type system to describe the type of an intersection with one element removed.

vpavkin: Right, just bumped into that. There's another option I'm exploring now - restrict this function to only be used with a single service layer, so that nothing is ever silently lost.

vpavkin: So it turned out as quite challenging to implement it in an ergonomic way :)
The best I could do so far is to make it look like this:

```scala
val layer = ???
layer.singleService.map(x => ???)
```
Here `.singleService` will only compile if layer has a single `Has[_]` output.
I think we might want other operations on layers with single outputs in future, so this "narrowing" step might be useful.

Or maybe this is an overkill, WDYT?

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2053,6 +2071,22 @@ object ZLayer {
      */
     def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
       self >>> ZLayer.fromFunctionMany(_.update[A](f))
+
+    /**
+     * Proves that this layer has a single `Has[_]` output,
+     * which allows more operations to be done lawfully, for example map a value inside `Has[_]`.
+     */
+    def singleService(implicit ev: HasSingleService[ROut]): SingleServiceOutputLayerOps[RIn, E, ev.Out] =
```
neko-kai: I feel like HasSingleService is redundant, does it work if replaced by
```

def singleService[A](implicit ev: ROut =:= Has[A]): SingleServiceOutputLayerOps[RIn, E, A] =
```
?

core/shared/src/main/scala/zio/ZLayer.scala
neko-kai: `RowLacks` is certainly expressible for a finite collection, or with a macro. If Has is to be extended with more operations we'd definitely need it sooner rather than later...

core/shared/src/main/scala/zio/ZLayer.scala
vpavkin: I tried it at first, and it works, but you have to annotate the call, like this:
```
layer.singleServie[Int].map(...)
```
It might be okay to be explicit what single service you want to get, but to me it seems redundant.
So this whole typeclass with a type member thing is to avoid this annotation.

core/shared/src/main/scala/zio/ZLayer.scala
vpavkin: @neko-kai I feel this one was intended for some other thread ;)

neko-kai: @vpavkin no, this refers to

> but I think it may not be possible in Scala's type system to describe the type of an intersection with one element removed.

Adam's sentence in messages above

core/shared/src/main/scala/zio/ZLayer.scala
neko-kai: You're right, very odd, but `=:=` doesn't work without ascription

vpavkin:
@adamgfraser @neko-kai So we now should decide our next step:) This is not a huge deal since the original purpose was to just have some sugar in a rather specific case.

@neko-kai says it should be possible to make it work in a much more generic way, but I don't feel competent to give it a shot ATM. It's probably a long-term goal anyway. So we can either merge this PR as is, or drop it as a too ad-hoc one. I'm fine with both ways.

adamgfraser:
Very interesting solution! You have summarized the situation quite well.

Regarding adding functionality to remove types in general from `Has` I tend to think that given how close we are to 1.0 and the focus on resolving remaining outstanding issues that is out of scope for us for now and is something to put on the requested features list of the next version of `ZLayer`.

In light of that, I tend to think that this PR adds significant complexity to make one specific case a bit more ergonomic and we are better off omitting for now.

Note that now that we have removed the restriction that `ROut <: Has[_]` you may be able to do what you want without any of this by deferring lifting your service into a `Has` until you actually need to compose it with other services. Then all of your method calls will work normally with the existing `map`.

vpavkin:
@adamgfraser agreed :) Thanks for guidance in the process and neat final tips.

PR #3163, opened 2020-03-17
JS Promise interop

### LINKED ISSUE #2881: Add a JS-specific interop layer for promises
Same as Java interop but for [`scala.scalajs.js.Promise`](https://www.scala-js.org/api/scalajs-library/1.0.0-M2/scala/scalajs/js/Promise.html). Methods can be added to ZIO trait and companion objects via the `*PlatformSpecific` traits in the `js` branch.

vasilmkd:
Resolves #2881.

In line with the Scala.js recommendations: https://www.scala-js.org/api/scalajs-library/1.0.1/scala/scalajs/js/Promise.html

@regiskuckaertz as original issue author.

### DIFF
--- a/core/js/src/main/scala/zio/ZIOPlatformSpecific.scala
+++ b/core/js/src/main/scala/zio/ZIOPlatformSpecific.scala
@@ -32,7 +32,7 @@ private[zio] trait ZIOPlatformSpecific[-R, +E, +A] { self: ZIO[R, E, A] =>
    * error type with `f`.
    */
   def toPromiseJSWith(f: E => Throwable): URIO[R, JSPromise[A]] =
-    self.mapError(f).fold(JSPromise.reject, JSPromise.resolve[A](_))
+    self.foldCause(c => JSPromise.reject(c.squashWith(f)), JSPromise.resolve[A](_))
 }
 
 private[zio] trait ZIOCompanionPlatformSpecific { self: ZIO.type =>
--- a/core/js/src/main/scala/zio/ZIOPlatformSpecific.scala
+++ b/core/js/src/main/scala/zio/ZIOPlatformSpecific.scala
@@ -16,6 +16,40 @@
 
 package zio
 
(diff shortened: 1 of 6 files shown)


regiskuckaertz:
Amazing ❤️ I looked into the code on Scala JS, the `toFuture` does not seem like it is doing too much: https://github.com/scala-js/scala-js/blob/bbf3629faedb59a260d39f961bc75b31ff3821a9/library/src/main/scala/scala/scalajs/js/Thenable.scala#L54-L70

Do you have bandwidth to convert this using ZIO's own Promise?

vasilmkd:
Thanks. Will do.

vasilmkd:
@regiskuckaertz How about now, I think it's even more straight to the point.

vasilmkd:
@regiskuckaertz Could you please take another look? Thanks.

core/js/src/main/scala/zio/ZIOPlatformSpecific.scala
```diff
@@ -16,6 +16,40 @@
...
+   */
+  def toPromiseJS(implicit ev: E <:< Throwable): URIO[R, JSPromise[A]] =
+    toPromiseJSWith(ev)
+
+  /**
+   * Converts the current `ZIO` to a Scala.js promise and maps the
+   * error type with `f`.
+   */
+  def toPromiseJSWith(f: E => Throwable): URIO[R, JSPromise[A]] =
+    self.mapError(f).fold(JSPromise.reject, JSPromise.resolve[A](_))
```
regiskuckaertz: I suggest to take possible defects in as well:
```suggestion
    self.foldCause(c => JSPromise.reject(c.squashWith(f)), JSPromise.resolve[A](_))
```

regiskuckaertz:
Sorry for the lack of feedback. Thanks again for doing this, there's one minor nit then it looks great to me 👍

vasilmkd:
No problem, I committed your suggestion.

Issue #3187, opened 2020-03-22
Improve type inference for ZSink


sir-wabbit:
POC example:
```scala
final case class ZIO[-R, +E, +A](run: R => Either[E, A]) {
  def map[B](f: A => B): ZIO[R, E, B] = ???
  def flatMap[R1 <: R, E1 >: E, B](f: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] = ???
}

trait ZSink[-R, +E, +A0, -A, B] { self =>
  def run(a: A): ZIO[R, E, (A0, B)]
  
  def zip[R1 <: R, E1 >: E, A01 >: A0, A1 >: A0 <: A, B1]
  (that: ZSink[R1, E1, A01, A1, B1]): ZSink[R1, E1, A01, A1, (B, B1)] =
    new ZSink[R1, E1, A01, A1, (B, B1)] {
      def run(a: A1): ZIO[R1, E1, (A01, (B, B1))] = {
        self.run(a).flatMap { case (a0, b) =>
          that.run(a) // just testing that it can eat A's
          
          that.run(a0).map { case (a01, b1) =>
            // testing that it can also have leftovers of A0
            val _ = a0 : A01
            
            (a01, (b, b1))
          }
        }
      }
    }
    
  def flatMap[R1 <: R, E1 >: E, A01 >: A0, A1 >: A0 <: A, C](
    f: B => ZSink[R1, E1, A01, A1, C]
  ): ZSink[R1, E1, A01, A1, C] = 
    new ZSink[R1, E1, A01, A1, C] {
      def run(a: A1): ZIO[R1, E1, (A01, C)] = 
        self.run(a).flatMap { case (a0, b) =>
          f(b).run(a0)
        }
    }
}
```

As you can see, it is possible to implement `ZSink`'s methods without relying on `=:=` proofs, which would inevitably result in bad type inference (AFAIU). I am currently investigating whether all methods on `ZSink` can be fixed in a similar manner.

alexknvl:
This method is particularly suspect:
```scala
/**
 * A `Sink[E, A0, A, B]` consumes values of type `A`, ultimately producing
 * either an error of type `E`, or a value of type `B` together with a remainder
 * of type `A0`.
 *
 * Sinks form monads and combine in the usual ways.
 */
trait ZSink[-R, +E, +A0, -A, +B] extends Serializable { self =>

  final def stepChunk[A00 >: A0, A1 <: A](state: State, as: Chunk[A1])(
[...]

PR #3188, opened 2020-03-22
Remove equality proofs from ZSink.

alexknvl:
See #3187. It is still not 100% clear to me whether type inference is improved by these changes, but:

- the number of `asInstanceOf` inside `ZSink` is reduced
- all tests passed locally
- all signatures are strictly more general (subtyping relations instead of type equalities)


alexknvl:
Looks like I am hitting type-inference bugs in Dotty.

alexknvl:
I will revert the changes that do not work in Dotty.

PR #3189, opened 2020-03-22
Simpler ZSink.succeed signature.

alexknvl:
```scala
// BEFORE
def succeed[A, B](b: => B): ZSink[Any, Nothing, A, A, B]
// AFTER
def succeed[B](b: => B): ZSink[Any, Nothing, Nothing, Any, B]
```

This is also in line with:
```scala
def fromEffect[R, E, B](b: => ZIO[R, E, B]): ZSink[R, E, Nothing, Any, B]
```

And the new behavior ensures:
```scala
ZSink.succeed(x) == ZSink.fromEffect(UIO.succeed(x))
```

Semi-related to #3187

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/SinkSpec.scala
@@ -94,7 +94,7 @@ object SinkSpec extends ZIOBaseSpec {
           assertM(sinkIteration(sink, 1).either)(isLeft(equalTo("Ouch")))
         } @@ zioTag(errors),
         testM("interaction with succeed") {
-          val sink = ZSink.succeed[Int, Int](5).collectAll
+          val sink = ZSink.succeed(5).collectAll
           for {
             init <- sink.initial
             s <- sink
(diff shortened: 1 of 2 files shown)


iravid:
This was the signature for succeed at some point, but we had to change it because of the interaction with other combinators. 

Let's see if the tests pass!

alexknvl:
It looks like the error is caused by the fact that `collectAll` is defined as a syntax extension over `ZSink[R, E, A, A, B]` and `ZSink.succeed : ZSink[Any, Nothing, Nothing, Any, B]` doesn't automatically upcast to that signature in Dotty. I will see if `collectAll` can be defined without syntax extensions. `test_212_jdk8_js` failure appears to be spurious.

adamgfraser:
Isn't the problem with this that it doesn't preserve the remainders and we're not sure if we are relying on that behavior somewhere in a way that the tests aren't capturing? See discussion in #3045.

iravid:
Yes, that is exactly it. Thank you Adam for the reminder. `ZStream#aggregateAsync` for example does not check `cont` between `initial` and `step`.

PR #3192, opened 2020-03-22
Improve fiber dump output

### LINKED ISSUE #2981: ZIO Core: Improve Rendering of Fiber Dumps
There is a lot of great information in fiber dumps about what fibers currently exist, who their parents are, and what they are doing to help diagnose deadlocks or other concurrency issues. However, right now the visualize presentation is not as good as it could be in clearly showing which fibers are parents of another fiber versus siblings. This ticket is to improve the `Fiber.prettyPrintM` to print more useful and easily understandable fiber dumps.

simpadjo:
Closes #2981

Added fiber hierarchy tree to the fiber dump output.

@bogdanmanate @adamgfraser could you please help with the testcase?
How do I create a non-trivial fiber hierarchy?
To create this output I cheated and dumped Fiber.roots.

Fiber dump output example:

Screenshot from 2020-03-22 15-38-02

### DIFF
--- a/core-tests/shared/src/test/scala/zio/FiberSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberSpec.scala
@@ -120,28 +120,42 @@ object FiberSpec extends ZIOBaseSpec {
         assertM(Fiber.collectAll(fibers).join)(anything)
       }
     ) @@ sequential,
-    suite("prettyPrint improvements")(
-      testM("fiber with one child") {
+    suite("fiber dump tree")(
+      testM("fiber dump tree shape sanity check") {
+        val expectedLinePrefixes =
+          """#+---"parent"
+            #    +---"parent"
+            #    |   +---"parent"
+            #    |   |   +---"child"
+            #    |   |       +---"grand"
+            #    |   +---"parent"
+            #    |   |   +---"child"
+            #    |   |       +---"grand"
+            #    |   +---"parent"
+            #    |   |   +---"child"
+            #    |   |       +---"grand""""
+            .stripMargin('#')
+            .split("\n")
+            .map(_.replace('|', ' '))
         for {
-          fiberRef <- FiberRef.make(0, math.max)
-          child    <- fiberRef.update(_ + 1).fork
-          dump     <- Fiber.dumpStr(child)
-          childId  <- child.id
-          _        <- child.join
+          blocker <- Promise.make[Nothing, Unit]
+          branchyFiber <- ZIO
+                           .foreachPar(1 to 3) { _ =>
+                             ZIO.foreachPar(1 to 3)(_ =>
+                               blocker.await
+                                 .forkAs(s"grand")
+                                 .flatMap(_.join)
+                                 .forkAs("child")
+                                 .flatMap(_.join)
+                             )
+                           }
+                           .forkAs("parent")
+          dumpStr <- Fiber.dumpStr(branchyFiber)
         } yield {
-          assert(dump.contains(s"+---#${childId.seqNumber} Status: Done"))(equalTo(true))
-        }
-      },
-      testM("fiber with multiple children") {
-        for {
-          parent   <- (ZIO.infinity.forkAs("child2") *> ZIO.infinity).forkAs("child1")
-          childId  <- parent.id
-          childSeq = childId.seqNumber
-          children <- parent.children
-          dumpStr  <- Fiber.dumpStr(parent)
-        } yield {
-          assert(dumpStr.contains(s"#${childSeq} Status"))(equalTo(true))
-          assert(children.size)(equalTo(1))
+          val linePrefixes = dumpStr.split("\n").toSeq.take(200).map(_.replace('|', ' '))
+          expectedLinePrefixes.toSeq
+            .map(exp => assert(linePrefixes)(exists(containsString(exp))))
+            .reduce(_ && _)
         }
       }
     )
--- a/core-tests/shared/src/test/scala/zio/FiberSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberSpec.scala
@@ -1,6 +1,7 @@
 package zio
 
 import zio.LatchOps._
+import zio.internal.FiberRenderer
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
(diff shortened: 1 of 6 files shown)


simpadjo:
@adamgfraser removing [WIP] label. 
@bogdanmanate created couple of testcases, I think they could be added as a follow-up PR.

bogdanmanate:
@adamgfraser and @simpadjo hier is the branch with the tests. https://github.com/bogdanmanate/zio/tree/pretty-dump-with-tests
Maybe we should merge it to this PR. What do you think ?

simpadjo:
@bogdanmanate not sure how to do it, I'm bad at git magic. Also keep in mind that it your commit is based on commit by "evgenii", not "simpadjo" and "evgenii" didn't sign CLA.

bogdanmanate:
@simpadjo I think I can do this. Please give me read/write access to you fork

adamgfraser:
Can you fix the failing tests?

bogdanmanate:
@adamgfraser I will have a look at the tests in the afternoon.

simpadjo:
@adamgfraser I think I'm done. Could you please take a look?

Screenshot from 2020-03-26 22-29-24

simpadjo:
@adamgfraser yeah, totally agree

core-tests/shared/src/test/scala/zio/FiberSpec.scala
```diff
@@ -119,7 +120,71 @@ object FiberSpec extends ZIOBaseSpec {
       testM("collectAll") {
         assertM(Fiber.collectAll(fibers).join)(anything)
       }
-    ) @@ sequential
+    ) @@ sequential,
+    suite("fiber dump tree")(
+      zio.test.test("render fiber hierarchy tree") {
```
simpadjo: If I call just `test` instead of `zio.test.test` I get an error:
/zio/core-tests/shared/src/test/scala/zio/FiberSpec.scala:125:7: package zio.test is not a value

simpadjo:
@adamgfraser done. Although have a small problem in a test

core-tests/shared/src/test/scala/zio/FiberSpec.scala
adamgfraser: Yes. This is an unfortunate side effect of the fact that `test` is both the package name and a method within the package. Fortunately this is only a problem within the `zio` package itself and I don't think there is a way to resolve other than renaming the package or the method, which I don't think we want to do, so we have to live with it. You handled it correctly.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -472,43 +477,16 @@ object Fiber extends FiberPlatformSpecific {
...
+    def prettyPrintM: UIO[String] = FiberRenderer.prettyPrintM(this)
+  }
 
-      s"""
-         |${name}#${fiberId.seqNumber} (${lifeMsg}) ${waitMsg}
-         |   Status: ${statMsg}
-         |${trace.prettyPrint}
-         |""".stripMargin
-    }
+  final case class Tree(
```
adamgfraser: There is a lot of overlap between this and the existing `Dump` data type. Could we just augment `Dump` to include a collection of children?

simpadjo: It bugs me as well but OTOH I can see few downsides: 
1) fiber tree feels more lightweight than entire dump and ability to collect just the tree w/o messing with stacktraces feels beneficial (uninformed IMHO)
2) Data stored in `Dump` objects become (more) denormalized
3) If implemented naively - collecting subtrees few times (once for each transitive parent) have memory and performance implications. Could be important given that we usually collect such information when the application is already not in the best shape.

adamgfraser: What if `Dump` when to including an `Option[ZTrace]` instead of a `ZTrace`, so methods that created dumps could decide whether to capture tracing information or not? Then dumps could just be fiber trees if we wanted. I think if we changed to this representation generating a dump would just mean generating a dump for each of the parents (since the children would be included in the parent dumps) which would involve the same work as today.

simpadjo: Do you mean that dump will look like that: 
Dump(fiberId: Fiber.Id, fiberName: Option[String], status: Status, children: Iterable[Dump],trace: Option[ZTrace]) and 

`def dump(fibers: Fiber.Runtime[_, _]*): UIO[Iterable[Dump]]` will return `n` dumps for `n` fibers (so become useless) and to render all stacktraces we will traverse the tree?

adamgfraser: Yeah, I'm thinking we would have:

```scala
final case class Dump(
  fiberId: Fiber.Id,
  fiberName: fiberName: Option[String],
  status: Status,
  children: Iterable[Dump],
  trace: Option[ZTrace]
)
```

Then `dump` would retain its signature but its meaning would change somewhat:

```scala
def dump(fibers: Fiber.Runtime[_, _]*): UIO[Iterable[Dump]]
```

Previously, this would return a "flat" structure. For example, if we did a dump of one fiber that had five children we would get an iterable that had five elements, one for each fiber. In the new encoding it would return an iterable with a single element and that element would have four children.

Then a render could be implemented as:

```scala
def render(dumps: Iterable[Dump]): String = ???
```

And the render could have access to the full structure of the fiber tree in deciding how to render it.

simpadjo:
@adamgfraser could you please take a look again?

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -403,13 +403,15 @@ object Fiber extends FiberPlatformSpecific {
...
-        id     <- self.id
-        status <- self.status
-        trace  <- self.trace
-      } yield Fiber.Dump(id, name, status, trace)
+        name       <- self.getRef(Fiber.fiberName)
+        id         <- self.id
+        status     <- self.status
+        trace      <- if (withTrace) self.trace.map(Some(_)) else UIO(None)
+        ch         <- self.children
+        childDumps <- ZIO.foreach(ch)(c => c.asInstanceOf[Fiber.Runtime[_, _]].dump(withTrace))
```
adamgfraser, line 413: All the children should be runtime fibers. You can change the type of `children` in `Fiber.Descriptor` to `Fiber.Runtime[Any, Any]` and the return type of `children` in `FiberContext` to `Fiber.Runtime[Any, Any]` to remove the need for this cast.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -403,13 +403,15 @@ object Fiber extends FiberPlatformSpecific {
     /**
      * Generates a fiber dump.
      */
-    final def dump: UIO[Fiber.Dump] =
+    final def dump(withTrace: Boolean): UIO[Fiber.Dump] =
```
adamgfraser, line 406: Do we expect generating a dump without tracing to be a common use case? We always capture traces today and don't seem to have had any issues with that.

If we do want this as an option I would suggest we change it to:

```scala
final def dumpWith(trace: Boolean): UIO[Fiber.Dump] = ???

val dump: UIO[Fiber.Dump] =
  dumpWith(true)
```

That way we have both the flexibility and the simple `dump` signature for the common case.

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -614,46 +584,24 @@ object Fiber extends FiberPlatformSpecific {
    * TODO: Switch to "streaming lazy" version.
    */
   @silent("JavaConverters")
-  val dumpAll: UIO[Iterable[Dump]] =
+  def dumpAll(withTrace: Boolean): UIO[Iterable[Dump]] =
```
adamgfraser, line 587: Same comment as above.

core/shared/src/main/scala/zio/Fiber.scala
simpadjo: I will add `dumpWith` to make this functionality accessible and and leave only one version in other places.

simpadjo: @adamgfraser done!

simpadjo:
@adamgfraser thank you!

adamgfraser:
@simpadjo Thank you! That was a great idea you had about rending the fiber tree first before the tracing information.

PR #3193, opened 2020-03-22
Fiber.Status.Suspended: make asyncTrace Option instead of List

simpadjo:
This field always consists of 0 or 1 elements


simpadjo:
Sorry, messed up 2 different PRs

PR #3196, opened 2020-03-22
ZStream error refinement.

alexknvl:


jdegoes:
Good to see these missing functions added. Thank you! 🙏

PR #3206, opened 2020-03-23
ZIO Core: Implement Cause#stripDefects

adamgfraser:
Sometimes we want to remove all defects of a particular type from a cause, just like we remove failures with `stripFailures`. This PR implements `stripDefects` to do this.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -370,6 +370,35 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
     attachTrace(squashWith(f))
 
   /**
+   * Remove all `Die` causes satisfying the specified predicate, returning
+   * `Some` with the remaining causes or `None` if there are no remaining
+   * causes.
+   */
+  final def stripDefects(f: Throwable => Boolean): Option[Cause[E]] =
+    self match {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -369,6 +369,35 @@ sealed trait Cause[+E] extends Product with Serializable { self =>
   final def squashTraceWith(f: E => Throwable): Throwable =
     attachTrace(squashWith(f))
 
+  /**
+   * Remove all `Die` causes satisfying the specified predicate, returning
+   * `Some` with the remaining causes or `None` if there are no remaining
+   * causes.
+   */
+  final def stripDefects(f: Throwable => Boolean): Option[Cause[E]] =
```
jdegoes: This almost looks like a partial function that produces no information other than partiality, e.g.:

```scala
def stripSomeDefects(f: PartialFunction[Throwable, Any]): Option[Cause[E]] = ...
...
cause.stripSomeDefects { case e : IOException => () }
```

adamgfraser: I was thinking the same thing. I was a little hesitant about using a partial function and not returning a value but it is equivalent and has better ergonomics. Updated accordingly.

PR #3207, opened 2020-03-23
Add `Schedule.durations`

jeremyrsmith:
Add `Schedule.durations` which should take a list of `Duration`s and make a schedule that delays each time based on the given durations, and then stops.

### DIFF
--- a/core/shared/src/main/scala/zio/Schedule.scala
+++ b/core/shared/src/main/scala/zio/Schedule.scala
@@ -735,6 +735,13 @@ object Schedule {
     elapsed.untilOutput(_ >= duration)
 
   /**
+   * A schedule that will recur with the specified durations, and then stop.
+   * Returns the total elapsed time.
+   */
+  def durations(durations: Duration*): Schedule[Clock, Any, Duration] =
+    unfold(0)(_ + 1).whileOutput(_ < durations.size).map(i => durations(i)).addDelay(d => d).fold(Duration.Zero)(_ + _)
+
+  /**
    * A schedule that recurs forever without delay. Returns the elapsed time
    * since the schedule began.
    */


core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -734,6 +734,13 @@ object Schedule {
   def duration(duration: Duration): Schedule[Clock, Any, Duration] =
     elapsed.untilOutput(_ >= duration)
 
+  /**
+   * A schedule that will recur with the specified durations, and then stop.
+   * Returns the total elapsed time.
+   */
+  def durations(durations: Duration*): Schedule[Clock, Any, Duration] =
```
adamgfraser: This is going to die with an `IndexOutOfBounds` right now because the effect is going to get evaluated once after the `whileOutput` returns `false`. Would definitely suggest we add some tests for this.

I think the fact that we can't implement this easily in terms of `unfoldM` is a bit of a code smell. We don't necessarily have to do as part of this PR but we may want to allow that to terminate instead of allows continuing forever. Then we could very easily implement in terms of that.

Much more minor but I wonder if we want to call `fromiterabloe` and have it take an `Iterable[Duration]` instead of variable arguments.

jeremyrsmith: Really, even though it does `whileOutput(_ < durations.size)`?

jeremyrsmith: What's a good place to add a test for this, so at least the test can fail 😄 I couldn't find one for `Schedule.duration` so I punted

adamgfraser: Here is a simple reproducer:

```scala
package zio

import zio.duration._

object Example extends App {

  def run(args: List[String]): ZIO[ZEnv, Nothing, Int] =
    effect.as(0)

  val schedule = Schedule.durations(1.second, 2.seconds, 3.seconds, 4.seconds)

  val effect = for {
    currentTime <- clock.nanoTime
    _ <- clock.nanoTime
          .map(_ - currentTime)
          .flatMap(n => console.putStrLn(n.toString))
          .repeat(Schedule.durations(1.second, 2.seconds, 3.seconds, 4.seconds))
  } yield ()
}
```

adamgfraser: In `ScheduleSpec.scala` if you look at the tests for something like "fibonacci delay" or "linear delay" you should be able to do something like that.

adamgfraser: We could change `Schedule.apply` to allow the `update` function to fail with `Unit`. Then we could change the signature of `foldM` to allow `f` to fail with `Unit`. Both of these changes would be compatible with all existing code since we're just accepting a wider type. Then we could implement this in terms of `delayed` and `foldM` similar to `fibonacci`.

We still may want to change the signature of `unfold` and `unfoldM` to return an `Option` and rename the existing variants to `iterate` and `iterateM`.

fanf: Also, it is not clear if the proposed method has `fixed` or `spaced` semantic, i.e if the duration of action is part of delay or not. It should be added in comment, or function name should keep a parallel with spaced/fixed.

Tests seems very mandatory for that method, I found it's semantic to be tricky, and user will want to see these example to understand how it behaves for real. 
For example, it's not self-evident that you still have a first occurence at time = 0 because of semantic of `ZIO.repeat` and that you should treat the first delay in a special way if you want to execute the action after a delay corresponding to the first item of the seq. In the same idea, it should be clear what happens with the last item, and that we don't either get an `OutOfBound` exception or that the last delay is ignored.

jeremyrsmith: Ah, I see what you mean. I agree `unfoldM` is currently the wrong way to do this. Thanks for looking!

PR #3210, opened 2020-03-24
Prevent retry storms on ZSTM

mijicd:
The current implementation of ZSTM suffers from "retry storms." For example, short transactions can force long-running ones to retry. 

To verify this behavior, I've set the following benchmark:

- make a `TRef` containing a list of 10000 elements
- have multiple short transactions updating only the head of the list
- have one transaction increment all elements
- race them

Running this benchmark on the current implementation produces the following result:

```
Benchmark                             Mode  Cnt  Score     Error  Units
STMRetryBenchmark.mixedTransactions  thrpt   15  0.125   ± 0.219  ops/s
```

The solution to this problem is to count the number of retries, and obtain a global lock once this count breaches some predefined limit. Acquiring a global lock should allow a long-running transaction to complete. To avoid paying the high locking "price," the number of max retries should be chosen from `[10, 100]`. Based on the results of my trial benchmarks, I've decided to set it to 10. 

With this change in place, the benchmark mentioned above produced the following result:

```
Benchmark                             Mode  Cnt  Score     Error  Units
STMRetryBenchmark.mixedTransactions  thrpt   15  425.153   ± 4.321  ops/s
```

### DIFF
--- a/benchmarks/src/main/scala/zio/stm/STMFlatMapBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/stm/STMFlatMapBenchmark.scala
@@ -9,6 +9,9 @@ import zio._
 @State(Scope.Thread)
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
+@Measurement(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 10)
+@Warmup(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 10)
+@Fork(1)
 class STMFlatMapBenchmark {
   @Param(Array("20"))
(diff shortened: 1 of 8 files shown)


adamgfraser:
Does there need to be some concept of how long a lock can last? With this change if one transaction is extremely long it can prevent all other transactions from executing indefinitely. This seems problematic because a local issue in one transaction can prevent the entire system of STM transactions involving this reference from being responsive.

mijicd:
@adamgfraser I have to think about that one. It obviously "prioritizes" the ones that were facing numerous retries, I'm not sure whether we can cut them off safely.

jdegoes:
@adamgfraser We should open a ticket for that /cc @mijicd

Issue #2991, opened 2020-03-01
Utilize `blockingOn` pervasively in FiberContext


jdegoes:
When async effects are constructed, they may specify the fiber ids of fibers they are blocking on. This is utilized in Promise but not utilized in FiberContext or many other places.

To provide more diagnostic information, this feature should be utilized whenever possible in FiberContext methods.

runtologist:
I'm working on this.

PR #3215, opened 2020-03-24
Utilize `blockingOn` pervasively in FiberContext

runtologist:
implements #2991

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -225,27 +225,34 @@ private[zio] final class FiberContext[E, A](
     val left  = fork[EL, A](race.left.asInstanceOf[IO[EL, A]])
     val right = fork[ER, B](race.right.asInstanceOf[IO[ER, B]])
 
-    ZIO
-      .effectAsync[R, E, C] { cb =>
-        val leftRegister = left.register0 {
-          case exit0: Exit.Success[Exit[EL, A]] =>
-            complete[EL, ER, A, B](left, right, race.leftWins, exit0.value, raceIndicator, cb)
-          case exit: Exit.Failure[_] => complete(left, right, race.leftWins, exit, raceIndicator, cb)
-        }
+    for {
+      lid <- left.id
+      rid <- right.id
+      r <- ZIO
+            .effectAsync[R, E, C](
+              { cb =>
+                val leftRegister = left.register0 {
+                  case exit0: Exit.Success[Exit[EL, A]] =>
+                    complete[EL, ER, A, B](left, right, race.leftWins, exit0.value, raceIndicator, cb)
+                  case exit: Exit.Failure[_] => complete(left, right, race.leftWins, exit, raceIndicator, cb)
+                }
 
-        if (leftRegister ne null)
-          complete(left, right, race.leftWins, leftRegister, raceIndicator, cb)
-        else {
-          val rightRegister = right.register0 {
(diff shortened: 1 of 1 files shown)


runtologist:
I'm not sure whether/how to write a test for this. It seems the `blockingOn` functionality is not covered with tests at the moment?

core-tests/shared/src/test/scala/zio/FiberSpec.scala
```diff
@@ -119,7 +119,29 @@ object FiberSpec extends ZIOBaseSpec {
...
+        for {
+          p    <- Promise.make[Nothing, Unit]
+          f1   <- p.await.fork
+          f1id <- f1.id
+          f2   <- f1.await.fork
+          dump <- f2.dump
+        } yield assert(dump.status)(
+          isSubtype[Fiber.Status.Suspended](hasField("blockingOn", _.blockingOn, equalTo(List(f1id))))
+        )
+      } @@ eventually,
```
jdegoes: I would generally use `flaky` instead of `eventually`.

Can you use latches to remove the flakiness from this one and the below one?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -225,27 +225,34 @@ private[zio] final class FiberContext[E, A](
...
-        if (leftRegister ne null)
-          complete(left, right, race.leftWins, leftRegister, raceIndicator, cb)
-        else {
-          val rightRegister = right.register0 {
-            case exit0: Exit.Success[Exit[_, _]] =>
-              complete(right, left, race.rightWins, exit0.value, raceIndicator, cb)
-            case exit: Exit.Failure[_] => complete(right, left, race.rightWins, exit, raceIndicator, cb)
-          }
+    for {
+      lid <- left.id
```
jdegoes: This will slow down the implementation quite a lot. Instead, you can just make the `fiberId` accessible (without being wrapped in an effect) by making it a protected val in the `FiberContext` constructor.

core-tests/shared/src/test/scala/zio/FiberSpec.scala
runtologist: I have added latches. Strictly speaking there could still be a race condition, but the probability is low and I was not able to cause one.

core/shared/src/main/scala/zio/internal/FiberContext.scala
runtologist: Good idea. Done.

core-tests/shared/src/test/scala/zio/FiberSpec.scala
runtologist: ... until it ran on CI. 🤣

runtologist: Using latches is difficult here: As I test blocking behavior, I must release the latch right before starting to block, so there could still be a race condition. One way of making the behavior deterministic is using a single threaded `Executor`. That makes the test platform specific, which I do not like either. I've resorted to retrying while the `Fiber.Status` still is `Running`. What do you think?

jdegoes: If the test must be flaky, then I would probably use `flaky` to indicate this.

runtologist: 👍

jdegoes:
@runtologist This ready for review?

runtologist:
👍  All conflicts resolved.

PR #3216, opened 2020-03-24
Improve ZQueue performance

mijicd:
Benchmarking the current implementation of ZQueue produced the following results:

```
Benchmark                           Mode  Cnt     Score     Error  Units
QueueParallelBenchmark.fs2Queue    thrpt   45  1204.274 ±  60.423  ops/s
QueueParallelBenchmark.monixQueue  thrpt   45  5081.348 ±  63.149  ops/s
QueueParallelBenchmark.zioQueue    thrpt   45  1584.057 ± 366.933  ops/s
QueueParallelBenchmark.zioTQueue   thrpt   45   200.517 ±  27.726  ops/s

Benchmark                             Mode  Cnt      Score    Error  Units
QueueSequentialBenchmark.fs2Queue    thrpt   45   3315.031 ± 17.993  ops/s
QueueSequentialBenchmark.monixQueue  thrpt   45  15968.575 ± 11.872  ops/s
QueueSequentialBenchmark.zioQueue    thrpt   45   1596.692 ± 16.479  ops/s
QueueSequentialBenchmark.zioTQueue   thrpt   45    341.544 ±  0.716  ops/s

Benchmark                               Mode  Cnt     Score     Error  Units
QueueBackPressureBenchmark.fs2Queue    thrpt   45   494.531 ±   3.835  ops/s
QueueBackPressureBenchmark.monixQueue  thrpt   45  3035.008 ±  39.692  ops/s
QueueBackPressureBenchmark.zioQueue    thrpt   45   552.961 ± 178.826  ops/s
QueueBackPressureBenchmark.zioTQueue   thrpt   45   422.290 ±   0.839  ops/s
```

I've identified the following problems:

- frequent invocation of expensive `checkShutdownState`
- multiple layers of effect nesting
- generic offer
- tail map calls (for-comprehension)

Running benchmarks mentioned above using the updated implementation produced the following results:

```
Benchmark                             Mode  Cnt     Score     Error  Units
QueueParallelBenchmark.zioQueue      thrpt   45  3363.840 ±  42.224  ops/s
QueueSequentialBenchmark.zioQueue    thrpt   45  4571.890 ±  30.653  ops/s
QueueBackPressureBenchmark.zioQueue  thrpt   45  1189.765 ±  12.754  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/zio/ZQueue.scala
+++ b/core/shared/src/main/scala/zio/ZQueue.scala
@@ -325,7 +325,7 @@ object ZQueue {
       p.unsafeDone(IO.succeedNow(a))
 
     sealed trait Strategy[A] {
-      def handleSurplus(as: List[A], queue: MutableConcurrentQueue[A], checkShutdownState: UIO[Unit]): UIO[Boolean]
+      def handleSurplus(as: List[A], queue: MutableConcurrentQueue[A], isShutdown: AtomicBoolean): UIO[Boolean]
 
       def unsafeOnQueueEmptySpace(queue: MutableConcurrentQueue[A]): Unit
 
@@ -338,7 +338,7 @@ object ZQueue {
       def handleSurplus(
         as: List[A],
         queue: MutableConcurrentQueue[A],
-        checkShutdownState: UIO[Unit]
+        isShutdown: AtomicBoolean
       ): UIO[Boolean] = {
         @tailrec
         def unsafeSlidingOffer(as: List[A]): Unit =
@@ -365,7 +365,7 @@ object ZQueue {
       def handleSurplus(
         as: List[A],
         queue: MutableConcurrentQueue[A],
-        checkShutdownState: UIO[Unit]
+        isShutdown: AtomicBoolean
       ): UIO[Boolean] = IO.succeedNow(false)
 
       def unsafeOnQueueEmptySpace(queue: MutableConcurrentQueue[A]): Unit = ()
@@ -389,7 +389,7 @@ object ZQueue {
       def handleSurplus(
         as: List[A],
         queue: MutableConcurrentQueue[A],
-        checkShutdownState: UIO[Unit]
+        isShutdown: AtomicBoolean
       ): UIO[Boolean] =
         UIO.effectSuspendTotal {
           @tailrec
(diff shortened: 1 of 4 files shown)


mijicd:
True, but we are making a fair trade-off (pun intended :)).

core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -484,104 +483,118 @@ object ZQueue {
...
+
+          if (noRemaining) IO.succeedNow(true)
+          else {
+            // not enough takers, offer to the queue
+            val surplus = unsafeOfferAll(queue, List(a))
+            unsafeCompleteTakers()
+
+            if (surplus.isEmpty) IO.succeedNow(true)
+            else
+              strategy.handleSurplus(surplus, queue, shutdownFlag) <* IO.effectTotal(unsafeCompleteTakers())
```
jdegoes: This could be made faster by pushing the `unsafeCompleteTakers()` into `handleSurplus`, perhaps. But you have 2 `FlatMap` from `<*` and another `effectTotal` wrapper around `unsafeCompleteTakers()`.

core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -484,104 +483,118 @@ object ZQueue {
 
     val capacity: Int = queue.capacity
 
-    def offer(a: A): UIO[Boolean] = offerAll(List(a))
+    def offer(a: A): UIO[Boolean] =
+      UIO.effectSuspendTotal {
```
jdegoes: We could optimize `effectSuspendTotal` primitive to do peekahead and immediately evaluate any `EffectTotal` / `Succeed` embedded inside there (such as the `succeedNow` in the below happy path).

jdegoes: Actually that probably wouldn't make a difference, scratch that.

core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -484,104 +483,118 @@ object ZQueue {
...
+        if (shutdownFlag.get) ZIO.interrupt
+        else {
+          val noRemaining =
+            unsafePollN(takers, 1) match {
```
jdegoes: May be worth drilling down into `unsafePollN` and optimizing it. For example, right now it is building this `List` every time you call it.

core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -484,104 +483,118 @@ object ZQueue {
...
+
+          if (remaining.isEmpty) IO.succeedNow(true)
+          else {
+            // not enough takers, offer to the queue
+            val surplus = unsafeOfferAll(queue, remaining.toList)
+            unsafeCompleteTakers()
+
+            if (surplus.isEmpty) IO.succeedNow(true)
+            else
+              strategy.handleSurplus(surplus, queue, shutdownFlag) <* IO.effectTotal(unsafeCompleteTakers())
```
jdegoes: Lots of overhead here as in the other case.

jdegoes:
Most (if not quite all) of the ZIO overhead has been squashed, next step is optimizing the `unsafe*` methods.

core/shared/src/main/scala/zio/ZQueue.scala
mijicd: Will do, probably the same specialization can be done with offering a single element.

PR #3221, opened 2020-03-25
Implement Schedule#fromDurations

### LINKED ISSUE #3205: Missing API for `ZSchedule` which takes a list of durations
Sometime, we want to exexcute a given action after specific (user provided) intervals of time, and it would be handy to have a pair of API for that: one matching `fixed` semantic, and one matching `spaced` one. 

I imagine a combinator taking a list of durations for parameter, and building the corresponding schedule so that: 
```
// start action for each of the provided times: 
val s1 = ZSchedule.startAt(0 second :: 2 minutes :: 5 minutes :: Nil)

// start action at 0s, then 2min after end of first action, then 5 minutes after end of second action
val s2 = ZSchedule.startSpacedBy(0 second :: 2 minutes :: 5 minutes :: Nil)
```
(I will let you choose better names

adamgfraser:
Resolves #3205.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -211,6 +211,12 @@ object ScheduleSpec extends ZIOBaseSpec {
             List.fill(5)(())
           )
         )(equalTo(List(0, 1, 4, 13, 40).map(i => (i * 100).millis)))
+      },
+      testM("fromDurations") {
+        val schedule = Schedule.fromDurations(4.seconds, 7.seconds, 12.seconds, 19.seconds)
+        val expected = List(0.seconds, 4.seconds, 11.seconds, 23.seconds, 42.seconds)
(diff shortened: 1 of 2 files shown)


adamgfraser:
@mijicd I'm seeing some test failures here in `StreamBufferSpec - Stream.bufferUnbounded - buffer the Stream`. These don't appear to be related to this PR as it only adds new methods to the API of `Schedule` and doesn't use them anywhere. Could this be related to the recent optimizations to `ZQueue`?

adamgfraser:
Looks like the same failures are on `master`.

mijicd:
It might be the case, but I haven't realized these locally or in PR build. Let me see.

mijicd:
Identified and fixed, running benchmarks to double-check. PR coming soon :).

mijicd:
Should be fixed in #3222

adamgfraser:
Copying @jeremyrsmith and @fanf.

PR #3223, opened 2020-03-25
Document how to use `provideSome` with `Has`

MaT1g3R:


adamgfraser:
This will work if `R <: Has[_]` but in that case this can also typically be done more ergonomically using `provideSomeLayer`. Otherwise users will need to continue to use the old pattern of defining a new trait that extends both modules and implementing each of the services.

MaT1g3R:
It's not obvious from documentation on how to use `provideSomeLayer` like so:
```scala
type MyContext = Has[MyContext.Context]
val myContext: MyContext = Has(???)
val eff1: ZIO[Blocking with MyContext, Nothing, Unit] = ???
val eff2: ZIO[Blocking, Nothing, Unit] = eff1.provideSome[Blocking](_ ++ myContext)
```

I could add the old documentation back in addition to the changes I made if that's desired

adamgfraser:
@MaT1g3R You can do `provideSomeLayer[Blocking](ZLayer.succeedMany(myContext))` or `provideSomeLayer[Blocking](ZLayer.succeed(myContext))` if `myContext` is not already a subtype of `Has[_]`. Yes, I think it would be good to to point people to `provideSomeLayer` as the preferred way to do this with your environment as a subtype of `Has` and then this being the way to do it you aren't using the new environment encoding for some reason.

MaT1g3R:
Is there a reason to prefer `provideSomeLayer` over  `provideSome`  with the `++` operator? The latter seems more concise to me.

adamgfraser:
I think if you already have a service it is slightly more ergonomic but ideally we want to carrot people to using layers as the idiomatic way to compose services. Because for a simple service that isn't effectual you may just have the service sitting around but if the service is effectual or requires finalization it will have to be embedded in a layer.

MaT1g3R:
I see, that makes sense for services.

We are using the `R` type in our code base (an HTTP service) to propagate information for a given HTTP request. (for example the unique ID we generate for a given request for logging purposes, or the user who made the request) It looks like something like this:

```scala
final case class LoggedInUser(user: User)
final case class TraceID(id: UUID)
type UserContext = Has[LoggedInUser]
type TraceContext = Has[TraceID]
type AppContext = TraceContext with UserContext

val getUser: ZIO[TraceContext, Throwable, User] = ???
val getData: ZIO[AppContext, Throwable, Data] = ???
val handleRequest: ZIO[TraceContext, Throwable, Data] = for {
    user <- getUser
    data <- getData.provideSome[TraceContext](_ ++ Has(user))
} yield data
```

I'm assuming there are other users of zio that does something like the above as well. Is there a reason to use layers for this case?

adamgfraser:
@MaT1g3R What you are doing is completely fine and equivalent to putting it into a layer. The one potential advantage of putting it into a layer is that it could be easier to refactor if you end up needing to use other services that do require effects or finalization. Similarly from an educational perspective I think there is some benefit in teaching people one way to work with environmental services that is powerful enough for everything they need to do.

I think you understand the different considerations pretty well so do you want to take another revision of this and then we can get it merged?

MaT1g3R:
Sounds good, I will revise the documentation after work hours for me

adamgfraser:
Awesome! Thank you for contributing!

PR #3231, opened 2020-03-27
ZIO Test: rename TestAspect failure -> failing

ioleo:
Due to name clash we can't import both `TestAspect._` and `Expectation._`.
Since `fails` alternative is already taken by `Assertion._` I believe the most semantically sound fix is to rename `TestAspect.failure` to `TestAspect.failing`. It also happens to be same length.

@adamgfraser


jdegoes:
I do think we should try to be internally name-clash free as much as possible as it results in a much better experience.

adamgfraser:
I agree.

PR #3232, opened 2020-03-27
Tighten Constraint on Has#get

adamgfraser:


jdegoes:
I think the caching code can now be simplified too, right? That might allow us to delete one of the tagged methods.

adamgfraser:
Updated to remove subtype check in cache. The one other place it is used in `Has` is in `prune`, which I think is now safe to remove. Unfortunately `taggedIsSubtype` is used in the mocking framework now so we can't remove entirely.

ioleo:
@adamgfraser Once ZIO Test is refactored to be more modular I'll work on extracting ZIO Mock to a seperate subproject and then iff `taggedIsSubtype` is only used in ZIO Mock we'll move it there.

PR #3239, opened 2020-03-29
DO NOT MERGE - ZIO Core: Add ifM Variants

### LINKED ISSUE #3235: ZIO Core: Add ZIO.if
In an ideal world I would be able to write code like this: 

```scala
def getChanges: Task[List[Change]] = 
  for {
    numberOfChanges <- checkForChanges
    _ <- logger.debug(s"Number of changes is: $numberOfChanges")
    results <- Task.if(numberOfChanges > 0)(getChanges)(Task(List()))
  } yield results   
```

And it would match up nicely with ZIO.when.

### LINKED ISSUE #3236: ZIO Core: Clarity over ZIO.when/if
Given the following code .. (2) and (3) compile and (1) and (4) don't.
This is confusing as it is perfectly normal to have an if statement without the else.

```scala
for {
  _ <- ZIO.ifM(hasChanges)(doSomething)  ... (1)
  _ <- ZIO.ifM(hasChanges)(doSomething)(doSomethingElse) ... (2)
  _ <- ZIO.whenM(hasChanges)(doSomething) ... (3)
  _ <- ZIO.whenM(hasChanges)(doSomething)(doSomethingElse) ... (4)
}
```
One way of reconciling this is to only support "when" at the end e.g. doSomething.when(..) and add support for (1) to "if". Or just make "if" a synonym for "when".

adamgfraser:
Resolves #3235. Resolves #3236.

We can't use `if` as a method name because it is a reserved keyword in Scala. To work around this, I overloaded `ifM` so it can accept either a pure of effectual condition. Overloading a method name is not ideal but given that we can't use the normal name I think this is the best alternative.

I also added variants to allow using `ifM` with a `onTrue` clause without an `onFalse` clause. This is duplicative of existing functionality provided by `when` and `whenM` so we could potentially remove these but given user expectations about the `if` control structure I think there is some value here.

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -461,13 +461,13 @@ object IO {
   def identity: IO[Nothing, Any] = ZIO.identity
 
   /**
-   * @see [[zio.ZIO.ifM]]
+   * @see [[zio.ZIO.ifM(b:=>Boolean)*]]
    */
-  def ifM[E](b: => Boolean): ZIO.IfM[Any, E] =
+  def ifM(b: => Boolean): ZIO.IfM[Any, Nothing] =
     ZIO.ifM(b)
 
   /**
-   * @see [[zio.ZIO.ifM]]
+   * @see [[zio.ZIO.ifM[R,E]*]]
    */
   def ifM[E](b: IO[E, Boolean]): ZIO.IfM[Any, E] =
     ZIO.ifM(b)
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -463,8 +463,14 @@ object IO {
   /**
    * @see [[zio.ZIO.ifM]]
    */
+  def ifM[E](b: => Boolean): ZIO.IfM[Any, E] =
+    ZIO.ifM(b)
+
+  /**
+   * @see [[zio.ZIO.ifM]]
+   */
   def ifM[E](b: IO[E, Boolean]): ZIO.IfM[Any, E] =
-    new ZIO.IfM(b)
+    ZIO.ifM(b)
 
   /**
    * @see See [[zio.ZIO.interrupt]]
--- a/core/shared/src/main/scala/zio/Managed.scala
+++ b/core/shared/src/main/scala/zio/Managed.scala
@@ -181,13 +181,13 @@ object Managed {
     ZManaged.halt(cause)
 
   /**
-   * See [[zio.ZManaged.ifM]]
+   * See [[zio.ZManaged.ifM(b:=>Boolean)*]]
    */
-  def ifM[E](b: => Boolean): ZManaged.IfM[Any, E] =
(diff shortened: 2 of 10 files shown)


ioleo:
@adamgfraser Maybe instead of overloading `ifM` we rename the combinator to `iff`?

adamgfraser:
@ioleo I'm a little hesitant because as you say "iff" has a very specific meaning of "if and only if" which doesn't really make sense here as it would require the left hand side and right hand side to be booleans or something else that has a sense of being true or false and typically would also return true or false.

nadenf:
Is Task.ifElse a possible option ?

If not it would seem to be better to just standardise on when/whenM since it would at least be consistent.

nadenf:
Also random thought .. possibly keep "when" and replace ifM when a ternary :) 

def hasChanges: UIO[Boolean]
val result = hasChanges ? doA, doB

adamgfraser:
I don't think we should change the meaning of `when`. It is a well-defined combinator that just executes one effect conditionally. It has been around for a long time and we are close to doing our final release.

At this point I'm a little unsure what the problem we are trying to solve here is.

#3236 was about `ifM` versus `when` / `whenM`. I think the distinction is `ifM` is used if you are choosing between two alternatives, `when` is used to execute one action conditionally. We could potentially keep the change here to allow `ifM` to be used with a single clause as well but that is really just duplicating `when`.

#3235 was about a non-effectual version of `ifM` but thinking about it more I think that is just the `if` control structure. Going back to the motivating example of being able to write something like:

```scala
def getChanges: Task[List[Change]] = 
  for {
    numberOfChanges <- checkForChanges
    _ <- logger.debug(s"Number of changes is: $numberOfChanges")
    results <- Task.if(numberOfChanges > 0)(getChanges)(Task(List()))
  } yield results   
```

This can already be done today as:

```scala
def getChanges: Task[List[Change]] = 
  for {
    numberOfChanges <- checkForChanges
    _ <- logger.debug(s"Number of changes is: $numberOfChanges")
    results <- if (numberOfChanges > 0) getChanges else Task(List())
  } yield results   
```

So I am inclined to close this as well as the referenced issued at this point.

ioleo:
@adamgfraser I think it would still be nice to have this combinator, as writeing the full `if () else {}` block is very verbose. Perhaps `ifElse` like @nadenf proposed?

adamgfraser:
@ioleo I don't think it is more verbose. In the example above using the built in control structure is actually one character shorter:

```scala
results <- Task.if(numberOfChanges > 0)(getChanges)(Task(List()))
results <- if (numberOfChanges > 0) getChanges else Task(List())
```

adamgfraser:
What do we want to do here?

I think with the recent changes to add `unless` and `unlessM` we don't need this anymore. At this point if you want to execute one effect on either an effectual or non-effectual condition being true or false you can use `when`, `whenM`, `unless`, or `unlessM`. If you want to execute two effects alternatively and the condition is effectual you can use `ZIO.ifM(cond)(onTrue, onFalse)`. If the condition is not effectual you can do `if (cond) onTrue else onFalse`.

Issue #3238, opened 2020-03-29
Optimize Chunk#++


jdegoes:
`zio.Chunk`

Currently, chunk concatenation is lazy, but does not utilize a balanced tree. Balancing is necessary to achieve efficient performance in the presence of operations on the concatenated tree.

See [here](https://en.wikipedia.org/wiki/Rope_(data_structure))

ahoy-jon:
after #3241
I could help someone do it.

ahoy-jon:
I take it ( #3311 )

adamgfraser:
Awesome!

ahoy-jon:
Current implementation on master is not working for a large chunk:
```scala
val chunk = (1 to 32768).foldLeft(Chunk(0))(_ + _)
chunk.map(_ + 1)
```
Is raising a stackoverflow.

Issue #2554, opened 2020-01-01
Add support for non-empty list with ZIO.foreach (or ZIO.collectAll and similar operations)


afsalthaj:
As of now, passing a nonEmptyList (which can be `scala.collection.immutable.::[A]`) to `ZIO.foreach` (or `ZIO.collectAll` and similar operations if any) returns a `List` which would mean it could be empty. However, in practice, passing a non-empty list to such operations can never yield an empty list. 

Hence, it is a good idea if ZIO could support a `collectAll` or `foreach` operation that can operate on `::` and returns a `::`

afsalthaj:
If we are happy with this issue, I can take it up myself.

adamgfraser:
To a certain extent this is raising the issue again of us not having a `Traverse` type class or another way to abstract over various collection methods. We already have another PR to add variants of these for `Option`, we want to add variants for `Chunk`, and potentially over time we will want the same thing for other methods that operate on collections like `partitionM`.

In the short term I agree the straightforward solution is just to add variants for this but may be worthwhile to think about if there is something else we can do here.

saraiva132:
@afsalthaj @adamgfraser Indeed. I think there needs to be some careful thinking on our APIs that have iterables as input can be improved. For instance #2451 

I am unsure what is the direction to go but, just the fact we are accepting mutable data structures as input is enough to raise a red flag.

afsalthaj:
@adamgfraser 
One of the thoughts that I have at this point is why would we even want to call certain zio functions (`zio.foreach`, `zio.collectAll`) with a empty collection at all, while we know its difficult to impose non-emptiness with the collection hierarchy. Correct me if I am wrong, that, let's make some functions support only non-emptiness?

From a user's perspective, it is difficult to form a non-empty version of the iterable to form, however, it will still be acceptable, given the fact we are trying to avoid computations that are invalid with the help of types.

Your view on `Traversable` type class is very interesting.  Having variants of `::` and `Chunk` could be abstracted out as single functions.

However, `Option` gets into the same question of why `foreach` on `None`. `ZIO.fromOption(opt).flatMap(ZIO.effect(...))` could be an alternate ?

adamgfraser:
[...]

PR #3241, opened 2020-03-30
Prototype for NonEmptyChunk

ahoy-jon:
Following a discussion on Discord with @adamgfraser 
This PR is to add a `NonEmptyChunk[A]` to ZIO.

A `Chunk[A]` would be either an `Empty` or a `NonEmptyChunk[A]`.

Some methods are being overloaded in the `trait NonEmptyChunk[A]`, some are moving to an `Ops`.

Design decisions :
* When possible, return a `NonEmptyChunk[A]`
* All the subtype of `Chunk[A]`, except `Empty`, extend `NonEmptyChunk[A]`
* Minimize the change to existing code
* Complete compatibility with Scala 2.11 is optional (overloaded method are not working well in 2.11)

So far it seems possible! 😃


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -831,7 +831,7 @@ object ZSink extends ZSinkPlatformSpecificConstructors with Serializable {
         type State = CollectAllNState
         case class CollectAllNState(s: sink.State, bs: List[B], n: Int, leftover: Chunk[A], dirty: Boolean)
 
-        val initial = sink.initial.map(CollectAllNState(_, List(), 0, Chunk(), false))
+        val initial = sink.initial.map(CollectAllNState(_, List(), 0, Chunk.empty, false))
```
ahoy-jon: this change is not needed, `Chunk()` works fine.

ahoy-jon:
Related issues/PR : 
* #2554 
* #2493

ahoy-jon:
@adamgfraser if you can take a look to check it is going in the right direction!

On my list of things to add : 
- there are a couple of tests to add (reduce, flatMap, ... )
- performance tests

core/shared/src/main/scala/zio/package.scala
```diff
@@ -42,6 +42,8 @@ package object zio extends EitherCompat with PlatformSpecific with VersionSpecif
   type Ref[A]      = ZRef[Nothing, Nothing, A, A]
   type ERef[+E, A] = ZRef[E, E, A, A]
 
+  type NonEmptyChunk[A] = Chunk.NonEmpty[A]
```
adamgfraser: Covariance here?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2453,6 +2467,16 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
     in.mapM(f)
 
+  /**
+   * Applies the function `f` to each element of the `NonEmptyChunk[A]` and
+   * returns the results in a new `NonEmptyChunk[B]`.
+   *
+   * For a parallel version of this method, see `foreachPar`.
+   * If you do not need the results, see `foreach_` for a more efficient implementation.
+   */
+  final def foreach[R, E, A, B](in: NonEmptyChunk[A])(f: A => ZIO[R, E, B]): ZIO[R, E, NonEmptyChunk[B]] =
```
adamgfraser: We should eventually add variants for `foreachParN` and `collectAllParN` as well as aliases in the companion objects.

streams-tests/jvm/src/test/scala/zio/stream/ChunkSpec.scala
```diff
@@ -277,5 +277,11 @@ object ChunkSpec extends ZIOBaseSpec {
       assert(Chunk(1, 2, 3).zipAllWith(Chunk(3, 2))(_ => 0, _ => 0)(_ + _))(equalTo(Chunk(4, 4, 0))) &&
       assert(Chunk(1, 2).zipAllWith(Chunk(3, 2, 1))(_ => 0, _ => 0)(_ + _))(equalTo(Chunk(4, 4, 0)))
     }
+    /*test("heterogeneous") {
```
adamgfraser: Delete per discussion in #3242.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -658,7 +658,7 @@ trait ZSink[-R, +E, +A0, -A, +B] extends Serializable { self =>
                 if (self.cont(s2)) UIO.succeedNow(Left(s2))
                 else self.extract(s2).map(Right(_))
               },
-            { case (b, leftover) => UIO.succeedNow(Right((b, leftover ++ Chunk.single(a)))) }
```
adamgfraser: I don't think these change are needed.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -34,20 +36,14 @@ sealed trait Chunk[+A] { self =>
    */
   val length: Int
 
-  /**
-   * Returns the concatenation of this chunk with the specified chunk.
-   */
-  final def ++[A1 >: A](that: Chunk[A1]): Chunk[A1] =
```
adamgfraser: Can we keep this on `Chunk` and override it in `NonEmpty` instead of using the implicit syntax class? I think that will be more discoverable and looks like it improves type inference in some cases (e.g. `++` in `Sink.scala` below).

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -783,21 +662,33 @@ sealed trait Chunk[+A] { self =>
 }
 
 object Chunk {
+  //For priority on specialized methods
+  implicit class ChunkOps[A](private val self: Chunk[A]) extends AnyVal {
```
adamgfraser: I think this can go away per comment above.

jdegoes:
Organization-wise: think this should be `Chunk.NonEmpty` to mirror `Chunk.Empty`.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -916,7 +807,232 @@ object Chunk {
...
+      }
+
+      dest.map {
+        case (state, array) => (state, Chunk.Arr(array))
+      }
+    }
+
+    final def ++[A1 >: A](that: Chunk[A1]): NonEmpty[A1] = Chunk.concat(self, that)
+
+    final def head: A        = self(0)
```
jdegoes: `head` / `tail` really only make sense with linked list structures.

Personally, I'd rename `head` => `first`, and then omit `tail`. Why omit tail? Because it's not a performant operation. If you recursively "uncons" from a chunk, you'll build up n-levels of allocation and indirection by the end—severe performance penalty.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -916,7 +807,232 @@ object Chunk {
...
+    final def tail: Chunk[A] = self.slice(1, length - 1)
+
+    final def last: A = self(length - 1)
```
jdegoes: This one is fine and would mirror `first`

streams/shared/src/main/scala/zio/stream/ZSink.scala
ahoy-jon: So far they are, I will check again an investigate.

core/shared/src/main/scala/zio/Chunk.scala
ahoy-jon: It is not possible to keep in on `Chunk` without losing some of the new features around concatenation.

```
**Chunk ++ NonEmpty => NonEmpty**
NonEmpty ++ Chunk => NonEmpty
NonEmpty ++ NonEmpty => NonEmpty
```

If I put them back
```scala 
scala> Chunk(1,2,3) ++ Chunk(1)
<console>:25: error: ambiguous reference to overloaded definition,
both method ++ in trait NonEmpty of type [A1 >: Int](that: zio.Chunk[A1])zio.Chunk.NonEmpty[A1]
and  method ++ in trait Chunk of type [A1 >: Int](that: zio.Chunk.NonEmpty[A1])zio.Chunk.NonEmpty[A1]
match argument types (zio.Chunk.NonEmpty[Int])
       Chunk(1,2,3) ++ Chunk(1)
```

This still works
```
scala> Chunk.empty ++ Chunk(1)
res1: zio.Chunk.NonEmpty[Int] = Singleton(1)
```

If `def ++[A1 >: A](that: NonEmpty[A1]): NonEmpty[A1]`  is removed, then it works again.

core/shared/src/main/scala/zio/ZIO.scala
ahoy-jon: To be sure, do you mean moving to `object Chunk`?

core/shared/src/main/scala/zio/Chunk.scala
ahoy-jon: I wish it would go away as well! 😃(maybe losing the features is fine)

streams/shared/src/main/scala/zio/stream/ZSink.scala
ahoy-jon: I am this error : 
```scala
[info] Compiling 1 Scala source to /Users/jon/Projects/zio/streams/jvm/target/scala-2.12/classes ...
[error] /Users/jon/Projects/zio/streams/shared/src/main/scala/zio/stream/ZSink.scala:661:86: type mismatch;
[error]  found   : zio.Chunk.NonEmpty[A1]
[error]  required: zio.Chunk.NonEmpty[A00]
[error]             { case (b, leftover) => UIO.succeedNow(Right((b, leftover ++ Chunk.single(a)))) }
[error]     
```

A00 and A1 are the same here, I am taking a deeper look.

adamgfraser: Yeah, I saw that error too when I was playing with it. Changing the `++` encoding resolved it.

core/shared/src/main/scala/zio/Chunk.scala
ahoy-jon: Solved it adding covariance in ChunkOps
```scala

implicit class ChunkOps[+A](private val self: Chunk[A]) extends AnyVal {
    /**
     * Returns the concatenation of this chunk with the specified chunk.
     */
    final def ++[A1 >: A](that: Chunk[A1]): Chunk[A1] = concat(self, that)

    /**
     * Returns the concatenation of this chunk with the specified chunk.
     */
    final def ++[A1 >: A](that: NonEmpty[A1]): NonEmpty[A1] = concat(self, that)
  }
```

Unchanging the code on Sink.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: No. In `ZIO.scala` there are `foreachParN` variants and `collectAllParN` variants. So we should implement versions for `NonEmptyChunk` in that file. Then all of these methods are aliased in `UIO.scala`, `URIO.scala`, `Task.scala`, etc...

core/shared/src/main/scala/zio/Chunk.scala
ahoy-jon: ChunkOps is removed.

core/shared/src/main/scala/zio/ZIO.scala
ahoy-jon: ok, I see, they don't exist yet for `Chunk[A]`:
```scala
scala> ZIO.foreachParN(5)(Chunk())
<console>:25: error: type mismatch;
 found   : zio.Chunk[Nothing]
 required: Iterable[?]
       ZIO.foreachParN(5)(Chunk())
```

I will add both versions of them.

ahoy-jon:
@jdegoes @adamgfraser Thanks you for the code review!

I have a lot to clean before the next review.
- 0 asInstanceOf
- reorder the methods in NonEmpty
- ...

As for the performance, it is similar to the previous version (thanks @Vilkina for quick help).

- Old Version
```scala
[info] Benchmark                     (size)  Mode  Cnt      Score      Error  Units
[info] MixedChunkBenchmarks.find       1000  avgt    5     64.854 ±    0.452  ns/op
[info] MixedChunkBenchmarks.flatMap    1000  avgt    5  80139.933 ± 3931.659  ns/op
[info] MixedChunkBenchmarks.fold       1000  avgt    5  22705.831 ± 1053.015  ns/op
[info] MixedChunkBenchmarks.foldM      1000  avgt    5  31908.680 ± 3762.721  ns/op
[info] MixedChunkBenchmarks.map        1000  avgt    5  26212.104 ± 2510.529  ns/op
[info] MixedChunkBenchmarks.mapM       1000  avgt    5  28448.603 ± 3103.181  ns/op
zio/master
```

- New Version
```scala
[info] Benchmark                     (size)  Mode  Cnt      Score      Error  Units
[info] MixedChunkBenchmarks.find       1000  avgt    5     61.773 ±    0.425  ns/op
[info] MixedChunkBenchmarks.flatMap    1000  avgt    5  80019.958 ±  201.439  ns/op
[info] MixedChunkBenchmarks.fold       1000  avgt    5  24802.410 ± 2749.465  ns/op
[info] MixedChunkBenchmarks.foldM      1000  avgt    5  29257.296 ± 1952.077  ns/op
[info] MixedChunkBenchmarks.map        1000  avgt    5  24998.218 ±  865.577  ns/op
[info] MixedChunkBenchmarks.mapM       1000  avgt    5  29055.468 ± 1118.978  ns/op
nonEmptyChunk
```

I don't find the difference conclusive, so more benchmarks!

core/shared/src/main/scala/zio/ZIO.scala
ahoy-jon: I did the alias for all the methods in UIO, Task, ..., ...

For `foreachParN` on Chunk and NonEmpty that will go in another PR.

adamgfraser: Agreed.

bernit77:
working with @ahoy-jon  , here are the performance test on my desktop 
with the command ```./sbt 'benchmarks/jmh:run  -i 15 -wi 15 -f1 -t2 ".*ChunkBenchmarks.*"'```
- nonEmptyChunk
```scala
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Benchmark                     (size)  Mode  Cnt       Score      Error  Units
[info] MixedChunkBenchmarks.find       1000  avgt   15      81,038 ±    0,483  ns/op
[info] MixedChunkBenchmarks.flatMap    1000  avgt   15  110688,064 ± 1006,285  ns/op
[info] MixedChunkBenchmarks.fold       1000  avgt   15   36156,638 ±  178,179  ns/op
[info] MixedChunkBenchmarks.foldM      1000  avgt   15   38390,781 ±  242,665  ns/op
[info] MixedChunkBenchmarks.map        1000  avgt   15   36968,630 ±  247,083  ns/op
[info] MixedChunkBenchmarks.mapM       1000  avgt   15   38547,848 ±  195,323  ns/op
[success] Total time: 1885 s (31:25), completed 2 avr. 2020 11:09:29
ahoy-jon/master
```
- master
```scala
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Benchmark                     (size)  Mode  Cnt       Score      Error  Units
[info] MixedChunkBenchmarks.find       1000  avgt   15      81,806 ±    1,835  ns/op
[info] MixedChunkBenchmarks.flatMap    1000  avgt   15  114643,764 ± 1404,866  ns/op
[info] MixedChunkBenchmarks.fold       1000  avgt   15   27936,776 ±  318,577  ns/op
[info] MixedChunkBenchmarks.foldM      1000  avgt   15   34992,621 ±  274,165  ns/op
[info] MixedChunkBenchmarks.map        1000  avgt   15   34528,973 ±  173,703  ns/op
[info] MixedChunkBenchmarks.mapM       1000  avgt   15   37150,904 ±  246,418  ns/op
[success] Total time: 1865 s (31:05), completed 2 avr. 2020 11:45:50
zio/ef148c12
```

ahoy-jon:
Thanks!

I will take a look at `fold`.

ahoy-jon:
I remove the code around slice, that could be a part of the effort in #3238

The code bring improvements but is at risk of raising StackOverflow errors with the current `concat` unbalancing.

💡Maybe we should upgrade Arr to integrate Slice, like: 
```scala
private final case class Arr[A](private val array: Array[A], offset: Int, l: Int) extends NonEmpty[A] {
```

- with the dispatch on slice with concat (39 nodes)
```scala
[info] Concat
[info] | Concat
[info] | | Concat
[info] | | | Concat
[info] | | | | Concat
[info] | | | | | Concat
[info] | | | | | | Concat
[info] | | | | | | | Slice 0 150
[info] | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | Concat
[info] | | | | | | | | Concat
[info] | | | | | | | | | Slice 150 50
[info] | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | | Concat
[info] | | | | | | | | | | Slice 200 50
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | | | Slice 250 150
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | Slice 400 50
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | Slice 450 50
[info] | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | Concat
[info] | | | | | | Slice 500 100
[info] | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | Slice 600 150
[info] | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | Concat
[info] | | | | | Slice 750 50
[info] | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | Slice 800 50
[info] | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | Singleton(802)
[info] | | Concat
[info] | | | Slice 752 48
[info] | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | Slice 800 50
[info] | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | Singleton(1000)
```
- without (64 nodes)
```scala
[info] Concat
[info] | Concat
[info] | | Concat
[info] | | | Concat
[info] | | | | Concat
[info] | | | | | Concat
[info] | | | | | | Concat
[info] | | | | | | | Slice 0 150
[info] | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | Concat
[info] | | | | | | | | Concat
[info] | | | | | | | | | Slice 150 50
[info] | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | | Concat
[info] | | | | | | | | | | Slice 200 50
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | | | Slice 250 150
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | Slice 0 50
[info] | | | | | | | | | Concat
[info] | | | | | | | | | | Slice 400 100
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | | | Slice 500 100
[info] | | | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | Slice 50 50
[info] | | | | | | | Concat
[info] | | | | | | | | Slice 400 100
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | Slice 500 100
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | Concat
[info] | | | | | | Slice 100 100
[info] | | | | | | | Concat
[info] | | | | | | | | Slice 400 100
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | Slice 500 100
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | Slice 0 150
[info] | | | | | | | Concat
[info] | | | | | | | | Slice 600 150
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | | Slice 750 50
[info] | | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | Concat
[info] | | | | | Slice 150 50
[info] | | | | | | Concat
[info] | | | | | | | Slice 600 150
[info] | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | | Slice 750 50
[info] | | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | Slice 800 50
[info] | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | Singleton(802)
[info] | | Slice 2 98
[info] | | | Concat
[info] | | | | Slice 150 50
[info] | | | | | Concat
[info] | | | | | | Slice 600 150
[info] | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | | | Slice 750 50
[info] | | | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | | | | Slice 800 50
[info] | | | | | Arr(1,2,3,4,5,6,7,8,9,10,11,12 ...
[info] | Singleton(1000)
```


(some code for the tree)
```scala

  final def tree[A](chunk: Chunk[A]): String = {

    def tree(chunk: Chunk[A], depth: Int): NonEmpty[String] = {
      val prefix = "| " * depth
      chunk match {
        case Concat(left, right)  => single(prefix + "Concat") ++ tree(left, depth + 1) ++ tree(right, depth + 1)
        case Slice(ch, offset, l) => single(prefix + s"Slice $offset $l") ++ tree(ch, depth + 1)
        case _ =>
          val str = {
            val str = chunk.toString
            val l   = 30
            if (str.length < l) str else str.take(l) + " ..."
          }
          single(prefix + str)
      }
    }

    tree(chunk, 0).mkString("\n")
  }
```

bernit77:
With the new version and :
```
sbt 'benchmarks/jmh:run -i 15 -wi 15 -f1 -t 2 “zio.chunks.*”'
```
```scala
[info] Benchmark                         (size)  Mode  Cnt      Score      Error  Units
[info] ArrayBenchmarks.find                1000  avgt   15     26,942 ±    0,818  ns/op
[info] ArrayBenchmarks.findOptimized       1000  avgt   15      5,427 ±    0,262  ns/op
[info] ArrayBenchmarks.flatMap             1000  avgt   15  22083,641 ±  208,726  ns/op
[info] ArrayBenchmarks.flatMapOptimized    1000  avgt   15  10755,606 ±  384,065  ns/op
[info] ArrayBenchmarks.fold                1000  avgt   15   9594,572 ±  297,517  ns/op
[info] ArrayBenchmarks.foldOptimized       1000  avgt   15    268,918 ±   18,812  ns/op
[info] ArrayBenchmarks.map                 1000  avgt   15  12291,378 ±   44,157  ns/op
[info] ArrayBenchmarks.mapOptimized        1000  avgt   15    608,422 ±    8,144  ns/op
[info] ChunkArrayBenchmarks.find           1000  avgt   15     26,264 ±    0,451  ns/op
[info] ChunkArrayBenchmarks.flatMap        1000  avgt   15  81095,690 ± 1168,859  ns/op
[info] ChunkArrayBenchmarks.fold           1000  avgt   15   6769,394 ±   20,036  ns/op
[info] ChunkArrayBenchmarks.foldM          1000  avgt   15  16551,973 ±   65,901  ns/op
[info] ChunkArrayBenchmarks.map            1000  avgt   15   6748,881 ±   24,788  ns/op
[info] ChunkArrayBenchmarks.mapM           1000  avgt   15  14987,576 ± 3682,394  ns/op
[info] MixedChunkBenchmarks.find           1000  avgt   15     53,163 ±    2,275  ns/op
[info] MixedChunkBenchmarks.flatMap        1000  avgt   15  81339,852 ± 1468,492  ns/op
[info] MixedChunkBenchmarks.fold           1000  avgt   15  26671,136 ±  966,332  ns/op
[info] MixedChunkBenchmarks.foldM          1000  avgt   15  33207,425 ± 1557,041  ns/op
[info] MixedChunkBenchmarks.map            1000  avgt   15  26464,614 ± 1327,751  ns/op
[info] MixedChunkBenchmarks.mapM           1000  avgt   15  31967,890 ± 1931,565  ns/op
[success] Total time: 6060 s (01:41:00), completed 2 avr. 2020 17:20:05
ahoy-jon/master
```

adamgfraser:
Does supporting nonempty chunks make this worth? My suggestion would be that we focus this PR on adding nonempty chunks to the API and make as few changes to the underlying implementation as possible and then we can pursue optimizations in separate PRs.

ahoy-jon:
on more (from @bernit77 )
```scala
info] Benchmark                         (size)  Mode  Cnt      Score     Error  Units
[info] ArrayBenchmarks.find                1000  avgt   15     26,443 ±   0,272  ns/op
[info] ArrayBenchmarks.findOptimized       1000  avgt   15      5,510 ±   0,095  ns/op
[info] ArrayBenchmarks.flatMap             1000  avgt   15  22281,623 ± 469,810  ns/op
[info] ArrayBenchmarks.flatMapOptimized    1000  avgt   15   9977,730 ± 290,703  ns/op
[info] ArrayBenchmarks.fold                1000  avgt   15   8852,843 ±  28,939  ns/op
[info] ArrayBenchmarks.foldOptimized       1000  avgt   15    235,250 ±   0,378  ns/op
[info] ArrayBenchmarks.map                 1000  avgt   15   6710,605 ±  23,257  ns/op
[info] ArrayBenchmarks.mapOptimized        1000  avgt   15    398,848 ±   0,503  ns/op
[info] ChunkArrayBenchmarks.find           1000  avgt   15     14,331 ±   0,076  ns/op
[info] ChunkArrayBenchmarks.flatMap        1000  avgt   15  43430,912 ± 294,322  ns/op
[info] ChunkArrayBenchmarks.fold           1000  avgt   15   4136,066 ±  15,416  ns/op
[info] ChunkArrayBenchmarks.foldM          1000  avgt   15   8944,188 ±  57,128  ns/op
[info] ChunkArrayBenchmarks.map            1000  avgt   15   4341,903 ±  22,704  ns/op
[info] ChunkArrayBenchmarks.mapM           1000  avgt   15  11449,996 ± 152,716  ns/op
[info] MixedChunkBenchmarks.find           1000  avgt   15     46,374 ±   0,189  ns/op
[info] MixedChunkBenchmarks.flatMap        1000  avgt   15  63203,012 ± 307,207  ns/op
[info] MixedChunkBenchmarks.fold           1000  avgt   15  20928,559 ± 420,732  ns/op
[info] MixedChunkBenchmarks.foldM          1000  avgt   15  25200,845 ± 132,856  ns/op
[info] MixedChunkBenchmarks.map            1000  avgt   15  22759,742 ± 187,125  ns/op
[info] MixedChunkBenchmarks.mapM           1000  avgt   15  26708,820 ±  98,812  ns/op
[success] Total time: 6024 s (01:40:24), completed 2 avr. 2020 19:24:07
```

ahoy-jon:
Sure, I removed most of the not necessary changes, the rest will come to another PR to have a better +, ++, slice.

If you see anything not needed (like [this](https://github.com/zio/zio/pull/3241/files#diff-290e96e6c64027a32c36d751a35d65deR928-R938)), don't hesitate to review it.

Some of the array initializations have been changed : 
- before 
```scala
var dest = null.asInstanceOf[Array[B]]
var i = 0
...
 if(dest == null) {
  implicit val B: ClassTag[B] = Chunk.Tags.fromValue(b)
  dest    = Array.ofDim[B](len)
} 
```
- after
```scala
val b     = f(first)
implicit val B: ClassTag[B] = Chunk.Tags.fromValue(b)
val dest    = Array.ofDim[B](len)

dest(0) = b
var i = 1
```

That can be changed back.

adamgfraser:
Sounds good. Will take a look tonight.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -498,21 +439,34 @@ sealed trait Chunk[+A] { self =>
   final def splitAt(n: Int): (Chunk[A], Chunk[A]) =
     (take(n), drop(n))
 
+  /**
+   * Takes `n` elements of the chunk, starting from the specified index `start`.
+   * in.slice(start,n) == in.drop(start).take(n)
+   */
+  final def slice(start: Int, n: Int): Chunk[A] = {
```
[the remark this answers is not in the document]
ahoy-jon: Not as an API, slice could have an Empty element before. I am moving it.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1061,6 +922,21 @@ object Chunk {
       drop(i)
     }
 
+    /**
+     * Returns the first element that satisfies the predicate.
+     */
+    override def find(f: A => Boolean): Option[A] = {
```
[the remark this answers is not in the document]
ahoy-jon: Just performance improvement on `Arr.find`, moving to another PR.

core/shared/src/main/scala/zio/Chunk.scala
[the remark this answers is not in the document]
ahoy-jon: moved here : 
https://github.com/zio/zio/pull/3241/files#diff-290e96e6c64027a32c36d751a35d65deR1078-R1093

streams-tests/jvm/src/test/scala/zio/stream/ChunkSpec.scala
```diff
@@ -276,6 +276,26 @@ object ChunkSpec extends ZIOBaseSpec {
...
+    },
+    test("concat slice") {
+      val in = Chunk(1, 2, 3) ++ Chunk(4, 5, 6)
+      assert(in.slice(0, 3))(equalTo(Chunk(1, 2, 3))) &&
+      assert(in.slice(1, 3))(equalTo(Chunk(2, 3, 4))) &&
+      assert(in.slice(-1, 6))(equalTo(in)) &&
+      assert(in.slice(0, 10))(equalTo(in)) &&
+      assert(in.slice(0, -1))(equalTo(Chunk.empty))
+    },
+    testM("flatMap Non Empty") {
```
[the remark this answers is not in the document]
ahoy-jon: https://github.com/zio/zio/pull/3241/files#diff-e70b1123afcb5d5449a81bc97a79a0d1R292-R306

streams-tests/jvm/src/test/scala/zio/stream/ChunkSpec.scala
```diff
@@ -276,6 +276,48 @@ object ChunkSpec extends ZIOBaseSpec {
...
+        val c: NonEmptyChunk[Int]        = c_ + 0
+        val f: Int => NonEmptyChunk[Int] = f_.andThen(_ + 0)
+
+        val in: NonEmptyChunk[Int] = c.flatMap(f)
+        val expected: Seq[Int]     = c.toSeq.flatMap(f.andThen(_.toSeq))
+
+        assert(in.toSeq)(equalTo(expected))
+      }
+    },
+    testM("nonEmptyChunk subtype") {
```
adamgfraser, line 292: If we just want to check that the subtype is preserved we probably don't need to generate a bunch of random data. We could do something like:
```scala
def assertNonEmpty[A](value: => NonEmptyChunk[A]): TestResult = {
  lazy val _ = value
  assertCompletes
}

lazy val nonEmptyChunk: NonEmptyChunk[A] = ???

suite("NonEmptyChunk subtype preservation)(
  zio.test.test("map")(nonEmptyChunk.map(identity))
  zio.test.test("flatMap")(nonEmptyChunk.flatMap(ZIO.succeed(_)))
  // All other combinators defined on subtype and supertype
}
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -75,14 +74,7 @@ sealed trait Chunk[+A] { self =>
 
     if (n <= 0) self
     else if (n >= len) Chunk.empty
-    else
```
adamgfraser, line 78: I thought we were going to keep the old implementations in this PR and do `slice` along with `++`?

ahoy-jon: `private def slice(chunk, start, size)` is a smart constructor for `Slice`, it would make sense to have it since `Slice` changed a bit.

Anyway, went back to the old implementation.

adamgfraser:
Looks great! Can you fix the warning and resolve conflicts so we can merge?

streams-tests/jvm/src/test/scala/zio/stream/ChunkSpec.scala
ahoy-jon: Did some changes in that direction. 
I didn't name the tests in a suite, those are compile-time checks.

https://github.com/zio/zio/pull/3241/files#diff-e70b1123afcb5d5449a81bc97a79a0d1R292-R326

ahoy-jon:
Sure, I will do that as soon as possible!

adamgfraser:
Hmmm, my idea for `++` didn't work. Maybe we should add another combinator like `concatNonEmpty` for that use case?

ahoy-jon:
Yes! A couple of potential names :
- `concatNonEmpty`
- `appendNonEmpty`
- `++!`
- ...
What is your choice? 😄

ahoy-jon:
@jdegoes  Just tried it, look promising but didn't work yet. I will update Sink a bit and check if there are other parts that don't pass. 

```scala
[error] /Users/jon/Projects/zio/streams/shared/src/main/scala/zio/stream/ZSink.scala:671:86: type mismatch;
[error]  found   : zio.NonEmptyChunk[A1]
[error]     (which expands to)  zio.Chunk.NonEmpty[A1]
[error]  required: zio.NonEmptyChunk[A00]
[error]     (which expands to)  zio.Chunk.NonEmpty[A00]
[error]             { case (c, leftover) => UIO.succeedNow(Right((c, leftover ++ Chunk.single(a)))) }
```

```scala
diff --git a/core/shared/src/main/scala/zio/Chunk.scala b/core/shared/src/main/scala/zio/Chunk.scala
index c63fbcdd..bb1822bb 100644
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -19,7 +19,8 @@ package zio
 import java.nio._
 
 import scala.collection.mutable.Builder
-import scala.reflect.{ classTag, ClassTag }
+import scala.language.higherKinds
+import scala.reflect.{ClassTag, classTag}
 
 /**
  * A `Chunk[A]` represents a chunk of values of type `A`. Chunks are designed
@@ -38,10 +39,13 @@ sealed trait Chunk[+A] { self =>
    */
   val length: Int
 
+
+  type Concat[A1] <: Chunk[A1]
   /**
    * Returns the concatenation of this chunk with the specified chunk.
    */
-  def ++[A1 >: A](that: Chunk[A1]): Chunk[A1] = Chunk.concat(self, that)
+  def ++[A1 >: A](that: Chunk[A1]): Concat[A1]
+  final def ++[A1 >: A](that: NonEmptyChunk[A1]):NonEmptyChunk[A1] = Chunk.concat(self,that)
 
   /**
@@ -1205,10 +1209,18 @@ object Chunk {
      * Zips this chunk with the specified chunk using the specified combiner.
      */
     override def zipWith[B, C](that: Chunk[B])(f: (Nothing, B) => C): Chunk[C] = Empty
+
+    override type Concat[A1] = Chunk[A1]
+
+    /**
+     * Returns the concatenation of this chunk with the specified chunk.
+     */
+    override def ++[A1](that: Chunk[A1]): Chunk[A1] = that
   }
 
   sealed trait NonEmpty[+A] extends Chunk[A] { self =>
 
+
     final def first: A = self(0)
     final def last: A  = self(length - 1)
 
@@ -1223,6 +1235,8 @@ object Chunk {
       res
     }
 
+    override type Concat[A1] = NonEmpty[A1]
+
     /**
      * Returns the concatenation of this chunk with the specified chunk.
      */
```

ahoy-jon:
There are more errors after on simpler cases:
```
[error] /Users/jon/Projects/zio/core-tests/jvm/src/test/scala/zio/SerializableSpec.scala:157:35: ambiguous reference to overloaded definition,
[error] both method ++ in trait NonEmpty of type [A1 >: Int](that: zio.Chunk[A1])zio.NonEmptyChunk[A1]
[error] and  method ++ in trait Chunk of type [A1 >: Int](that: zio.NonEmptyChunk[A1])zio.NonEmptyChunk[A1]
[error] match argument types (zio.NonEmptyChunk[Int])
[error]       val chunk = Chunk.single(1) ++ Chunk.single(2)
[error]                                   ^
[error] /Users/jon/Projects/zio/core-tests/shared/src/test/scala/zio/ChunkSpec.scala:11:83: ambiguous reference to overloaded definition,
[error] both method ++ in trait NonEmpty of type [A1 >: Int](that: zio.Chunk[A1])zio.NonEmptyChunk[A1]
[error] and  method ++ in trait Chunk of type [A1 >: Int](that: zio.NonEmptyChunk[A1])zio.NonEmptyChunk[A1]
[error] match argument types (zio.NonEmptyChunk[Int])
[error]         val chunk = Chunk.empty ++ Chunk.fromArray(Array(1, 2)) ++ Chunk(3, 4, 5) ++ Chunk.single(6)
[error] 
```

ahoy-jon:
The initial solution with still compile:
```scala
  implicit class ChunkOps[A](val self: Chunk[A]) extends AnyVal {
    def ++[A1 >: A](chunk: Chunk[A1]): Chunk[A1]                         = concat(self, chunk)
    def ++[A1 >: A](nonEmptyChunk: NonEmptyChunk[A1]): NonEmptyChunk[A1] = concat(self, chunk)
  }
```
But : 
- it decrease the visibility 
- don't compile in ZSink without some code change (`xxx ++ single(a)` =>  `xxx + a`)

adamgfraser:
@jdegoes The issue we are actually having is with getting `chunk ++ nonEmpty` to work.

@ahoy-jon That should work equally well if you move both of those to the trait itself instead of an implicit syntax class so I don't think the first point is an issue. I'm looking at the issue with `ZSink` now.

PR #3247, opened 2020-03-31
ZIO Mock support for impure methods, ZStream and ZSink

### LINKED ISSUE #2734: ZIO Mock: support methods that accept streams and/or produce streams
Currently ZIO Mock supports methods that gets parameters and return an effect that produces a single value. The ask is to provide support for methods that produce/consume streams.

ioleo:
Fixes #2734

This PR introduces support for ZStream/ZSink and non-ZIO methods (regular functions/vals) as well as many refactorings/renames as the additional use cases crystalized the design to be more clean.

To be completed it still needs:
- [x] add support for ZStream/ZSink and non-ZIO methods in `@mockable` macro
- [x] add support for ZStream/ZSink and non-ZIO methods in `@accessible` macro
- [ ] ~add test coverage for more advanced use cases composing Streams with Effects~
- [ ] ~add more advanced examples of mocks composing Streams with Effects~

@adamgfraser @jdegoes Once these are completed I'd say ZIO Mock is `1.0` ready. I'll try to complete these as soon as possible. Opening the PR earlier so you can track progress and review the approach.

Out of the scope of this PR is support for polymorphic ZStreams/ZSinks. It will be tricky to get right, so I'll tackle it seperately and would not like it to be a blocker for this PR.


ioleo:
Updated `@mockable` macro.

ioleo:
As soon as CI finishes this is ready to be reviewed & merged @adamgfraser .

ioleo:
@adamgfraser can we rerun https://app.circleci.com/pipelines/github/zio/zio/6519/workflows/3cf22b04-08a3-4c31-9752-bb78704e3270/jobs/67033 ?
it works locally, I suspect this is some flaky test

test/shared/src/main/scala/zio/test/mock/Capability.scala
```diff
@@ -21,17 +21,20 @@ import java.util.UUID
...
- * To represent polymorphic capabilities you must use one of lazy `Method.Poly` types which
+ * To represent polymorphic capabilities you must use one of lazy `Capability.Poly` types which
  * allow you to delay the declaration of some types to call site.
+ *
+ * To construct capability tags you should start by creating a `Mock[R]` and extend publicly
+ * available `Effect`, `Method`, `Sink` or `Stream` type members.
  */
-abstract class Method[R <: Has[_]: Tagged, I: Tagged, E: Tagged, A: Tagged](val compose: URLayer[Has[Proxy], R])
-    extends Method.Base[R] { self =>
+protected[mock] abstract class Capability[R <: Has[_]: Tagged, I: Tagged, E: Tagged, A: Tagged](val mock: Mock[R])
```
adamgfraser, line 36: Can we add variance here or no because of the tags?

test/shared/src/main/scala/zio/test/mock/Mock.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import zio.internal.Executor
+import zio.stream.{ ZSink, ZStream }
+import zio.test.TestPlatform
+import zio.{ Has, Runtime, Tagged, URIO, URLayer, ZIO }
+
+/**
+ * A `Mock[R]` represents a mockable environenment `R`.
+ */
+abstract class Mock[R <: Has[_]: Tagged] { self =>
```
adamgfraser, line 27: Covariant?

docs/howto/mock_services.md
```diff
@@ -136,29 +144,39 @@ For overloaded methods we nest a list of numbered objects, each representing sub
...
+        def zeroArgs                               = proxy(ZeroArgs)
+        def zeroArgsWithParens()                   = proxy(ZeroArgsWithParens)
+        def singleArg(arg1: Int)                   = proxy(SingleArg, arg1)
+        def multiArgs(arg1: Int, arg2: Long)       = proxy(MultiArgs, arg1, arg2)
+        def multiParamLists(arg1: Int)(arg2: Long) = proxy(MultiParamLists, arg1, arg2)
+        def command(arg1: Int)                     = proxy(Command, arg1)
+        def overloaded(arg1: Int)                  = proxy(Overloaded._0, arg1)
+        def overloaded(arg1: Long)                 = proxy(Overloaded._1, arg1)
+        def function(arg1: Int)                    = rts.unsafeRunTask(proxy(Function, arg1))
+        def sink(a: Int)                           = rts.unsafeRun(proxy(Sink, a).catchAll(error => UIO(ZSink.fail(error))))
```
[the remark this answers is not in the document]
ioleo: The internal mock machinery returns an `IO[E, A]` where `A <: ZStream` or `A <: ZSink`.
For ZStream `E = Nothing`, becouse any errors happening in creation of the stream are deferred to the first `take` (moved into streams error channel `ZStream[_, E, _]`). For `ZSink` any errors happening upon creation are caught immidietly resulting in a failed Sink, so we need to manually move the error from our internal mock `IO[E, _]` to a failed ZSink.

docs/howto/mock_services.md
```diff
@@ -97,34 +98,41 @@ object Example {
...
 ```
 
-A _capability tag_ is just a value which extends the `zio.test.mock.Method[R <: Has[_], I, E, A]` type constructor, where:
+A _capability tag_ is just a value which extends the `zio.test.mock.Capability[R <: Has[_], I, E, A]` type constructor, where:
 - `R` is the type of environment the method belongs to
 - `I` is the type of methods input arguments
 - `E` is the type of error it can fail with
 - `A` is the type of return value it can produce
 
+The `Capability` type is not publicly available, instead you have to extend `Mock` dependent types `Effect`, `Method`, `Sink` or `Stream`.
```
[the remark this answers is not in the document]
ioleo: Yes, we should have all ZIO data types covered. I don't see a clean way to abstract over this ATM, but will definately think about it in future.

test/shared/src/main/scala/zio/test/mock/Capability.scala
ioleo: We can't. Variance will break things.

docs/howto/mock_services.md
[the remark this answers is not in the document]
adamgfraser: Could we do something like `io.foldCauseM(e => ZStream.fail(e), stream)`? Maybe I'm not understanding.

test/shared/src/main/scala/zio/test/mock/Mock.scala
ioleo: It would probably break things, but have not tested. I don't think it would bring any value though, at least in current implementation.

PR #3256, opened 2020-04-01
 ZIO Test: Use Suite Execution Strategy in ProvideSomeLayerShared

adamgfraser:
In `provideSomeLayerShared` we need to create one effect that models execution of all tests in the spec to provide the layer to. This PR updates the implementation to use the execution strategy for each suite in the composed effect.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Spec.scala
+++ b/test/shared/src/main/scala/zio/test/Spec.scala
@@ -189,7 +189,7 @@ final case class Spec[-R, +E, +T](caseValue: SpecCase[R, E, T, Spec[R, E, T]]) {
       case SuiteCase(label, specs, exec) =>
         specs.foldCauseM(
           c => f(SuiteCase(label, ZIO.halt(c), exec)),
-          Spec
+          ZIO
             .foreachExec(_)(exec.getOrElse(defExec))(_.foldM(defExec)(f))
             .flatMap(z => f(SuiteCase(label, ZIO.succeedNow(z.toVector), exec)))
         )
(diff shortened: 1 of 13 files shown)


test/shared/src/main/scala/zio/test/Spec.scala
```diff
@@ -562,8 +547,22 @@ object Spec {
...
     )(implicit ev1: R0 with R1 <:< R, ev2: NeedsEnv[R], tagged: Tagged[R1]): Spec[R0, E1, T] =
-      self.provideLayerShared[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer)
+      self.provideLayerShared[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer, defExec)
   }
+
+  /**
+   * Applies the function `f` to each element of the `Iterable[A]` and returns
+   * the result in a new `List[B]` using the specified execution strategy.
+   */
+  private def foreachExec[R, E, A, B](
```
adamgfraser, line 560: This could be worth upstreaming to ZIO Core. It provides a nice way of abstracting over execution strategies.

jdegoes: I agree, I like it.

jdegoes:
@adamgfraser I like this as-is. But it does strike me there's an alternate API we should consider:

```scala
sealed trait ExecutionStrategy[+A]
...
foreachExec(Parallel(List(1, 2, 3, 4))) { element =>
  ...
}
```

adamgfraser:
@jdegoes Yes. I think that might not work as well when used with combinators that are methods on a trait versus a companion object. For example in `Spec#foreachExec` it seems like we would need to move that from the `Spec` trait to the companion object so we could take `ExecutionStrategy[Spec[R, E, T]]` as an argument, which would be a little less than ideal.

Conceptually it seems like it might also be valuable sometimes to be able to talk about an execution strategy separate from any concrete collection (e.g. tell an application whether it is supposed to do whatever it does sequentially or in parallel). This could be handled by doing `ExecutionStrategy[Unit]` but still seems like there are separate ideas here.

jdegoes:
@adamgfraser Sounds good to me!

PR #3257, opened 2020-04-02
Implemented lens-like ZRef focusing on state components

Odomontois:
This is a proposal for more general than `fold` transformer for `ZRef` named `focusFold` with lens-like interface.
Such a transformer can focus mutable access on individual components of product types, such as case classes and collections.
Changes overview:
- added abstract method `focusFold`, which can analyze old read value during an update.
- added third variant of `ZRef` named `Focused` which has worse performance in general than previous `Derived` . 
- `Derived` renamed to `Mapped` and now extends `Focused` to reduce code repetition

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZRefSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZRefSpec.scala
@@ -267,6 +267,147 @@ object ZRefSpec extends ZIOBaseSpec {
           value     <- ref.get
         } yield assert(value)(equalTo(update))
       }
+    ),
+    suite("Focused")(
+      testM("atomicity") {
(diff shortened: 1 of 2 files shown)


adamgfraser:
Ah funny I have been working on something similar. Will post my current version as well.

jdegoes:
@Odomontois @adamgfraser Great minds think alike. 😆 

Glad to see this development finally happening. it was the main reason for the `ZRef` refactoring.

I'll let you two figure out how to merge these ideas and do a review when you're ready.

jdegoes:
Closed in favor of the hybrid #3258. @Odomontois Thanks for your work on this one!

PR #3258, opened 2020-04-02
Implement Optics for ZRef

adamgfraser:
Adds support for optics to `ZRef`. Implements a new method `foldS` that allows access to the state in the modification function. This allows us to implement `accessField` and `accessCase`, which allow applying a lens or a prism to a `ZRef`, respectively. Also includes initial work on encoding of traversals.

Copying @Odomontois.


core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -320,33 +340,68 @@ object ZRef extends Serializable {
...
 
     final def setAsync(a: A): IO[EA, Unit] =
-      setEither(a).fold(ZIO.fail(_), value.setAsync)
+      value.modify { s =>
+        setEither(a)(s) match {
+          case Left(e)  => (Left(e), s)
+          case Right(s) => (Right(()), s)
+        }
+      }.absolve
+  }
```
Odomontois, line 380: This was the main reason to add another type for `foldS` result in my PR.
These implementations are clearly more complex than older one, which can be significant in the 
And `setAsync` loses ability to delay modification.
So I felt like it's good to keep a somewhat performant version for simple mappings.

adamgfraser: Yes could definitely add this.

core/shared/src/main/scala/zio/package.scala
```diff
@@ -42,6 +42,16 @@ package object zio extends EitherCompat with PlatformSpecific with VersionSpecif
   type Ref[A]      = ZRef[Nothing, Nothing, A, A]
   type ERef[+E, A] = ZRef[E, E, A, A]
 
+  type ZLens[-S, +T, +A, -B]      = (S => A, B => S => T)
```
adamgfraser, line 45: We could be more polymorphic here, allowing each of these optics to have a get error and a set error here, but then we would need some constructors that make it easy for users to lift simple optics into these more polymorphic types.

core/shared/src/main/scala/zio/package.scala
```diff
@@ -42,6 +42,16 @@ package object zio extends EitherCompat with PlatformSpecific with VersionSpecif
...
+  type ZOptional[-S, +T, +A, -B]  = (S => Option[A], B => S => T)
+  type ZPrism[-S, +T, +A, -B]     = (S => Option[A], B => T)
+  type ZTraversal[-S, +T, +A, -B] = (S => List[A], List[B] => S => Option[T])
```
adamgfraser, line 48: I'm encoding a traversal right now as a pair of functions `S => List[A]` and `List[B] => S => Option[T]`. The set operation can fail if the number of elements being set doesn't match the expected number of elements to be set in the structure.

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -320,33 +340,73 @@ object ZRef extends Serializable {
...
+        b => optional._1(b).fold[Either[Unit, D]](Left(()))(Right(_))
+      )
+  }
+
+  implicit class PrismSyntax[E, A, B](private val self: ZRef[E, Unit, A, B]) extends AnyVal {
+    def accessCase[C, D](prism: ZPrism[B, A, D, C]): ZRef[E, Unit, C, D] =
+      self.fold(identity, identity, c => Right(prism._2(c)), b => prism._1(b).fold[Either[Unit, D]](Left(()))(Right(_)))
+  }
+
+  implicit class TraversalSyntax[A, B](private val self: ZRef[Unit, Unit, A, B]) extends AnyVal {
```
adamgfraser, line 402: This one in particular feels like it could benefit from a more polymorphic encoding of the optics.

core/shared/src/main/scala/zio/package.scala
adamgfraser: This could also allow us to support some other types of optics such as partial isomorphisms from ZIO Config.

core/shared/src/main/scala/zio/ZRef.scala
adamgfraser: Updated to add subtype to optimize `set` and `setAsync` for derived `ZRef` instances where possible.

adamgfraser:
@Odomontois Would you take a look at this when you get the chance? I revised to incorporate your point about retaining the performance of the original version where possible and made it more polymorphic to support a wide range of optics.

adamgfraser:
Added support for using optics with transactional references.

adamgfraser:
Added documentation.

core/shared/src/main/scala/zio/package.scala
```diff
@@ -42,8 +42,39 @@ package object zio extends EitherCompat with PlatformSpecific with VersionSpecif
   type Ref[A]      = ZRef[Nothing, Nothing, A, A]
   type ERef[+E, A] = ZRef[E, E, A, A]
 
+  type ZOptic[-SG, -SS, -B, +EG, +ES, +A, +T] = (SG => Either[EG, A], B => SS => Either[ES, T])
```
jdegoes, line 45:
```suggestion
  /**
    * A `ZOptic` is able to get and set a piece of a whole, possibly failing. In the most general possible case,
    * the set / get types are distinct, and setting may fail with a different error than getting.
    *
    * See more specific type aliases for concrete examples of what optics can be used to do.
    */
  type ZOptic[-GetWhole, -SetWholeBefore, -SetPiece, +GetError, +SetError, +GetPiece, +SetWholeAfter] = 
  (GetWhole => Either[GetError, GetPiece], SetPiece => SetWholeBefore => Either[SetError, SetWholeAfter])
```

?

😆

core/shared/src/main/scala/zio/ZRef.scala
```diff
@@ -57,6 +54,23 @@ sealed trait ZRef[+EA, +EB, -A, +B] extends Serializable { self =>
     bc: B => Either[ED, D]
   ): ZRef[EC, ED, C, D]
 
+  /**
+   * Folds over the error and value types of the `ZRef`, allowing access to
+   * the state in transforming the `set` value. This is a more powerful version
+   * of `fold` but requires unifying the error types.
+   */
+  def foldS[EC >: EB, ED, C, D](
```
jdegoes, line 62: `foldAll`?

bmeesters:
Hello @adamgfraser , as a random bystander I am just curious on why the optics have been pulled out. Seemed pretty cool!

adamgfraser:
@bmeesters We are going to provide them as part of a separate `zio-optics` project. This will allow us to continue to evolve the API for this functionality after we guarantee binary compatibility for 1.0, fully explore what we can do with this without risk of bloating the core ZIO project, and allow users who want to use other optics implementations to do that. Look for more on this soon!

PR #3259, opened 2020-04-02
core: Add `unless/unlessM` as dual of `when/whenM`

sirthias:
in the spirit of `filter/filterNot`, `doUntil/doWhile`, `retryUntil/retryWhile` and the like.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2873,6 +2873,49 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert(result)(isNone)
       }
     ),
+    suite("unless")(
+      testM("executes correct branch only") {
+        for {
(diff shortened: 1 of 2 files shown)


sirthias:
`unlessM` is especially useful since negating the condition for use with `when` is more clunky on the "lifted" level than it is on the value level:

```scala
val zio: ZIO[..., ..., ...] = ...

val cond: Boolean = ...
val when = zio.when(cond)
val unless0 = zio.when(!cond) // negation is easy
val unless1 = zio.unless(cond) // maybe clearer

val condM: UIO[Boolean] = ...
val whenM = zio.whenM(cond)
val unlessM0 = zio.whenM(cond.map(!_)) // negation is more clunky
val unlessM1 = zio.unlessM(cond) // clearer and more concise
```

adamgfraser:
Yes. There is a little danger here that we have another different set of ways to do very similar things but I am inclined to think this makes sense.

PR #3266, opened 2020-04-02
Introduce ZTRef

### LINKED ISSUE #3260: Give TRef the polymorphic treatment
Given that `ZRef` is only used to update immutable data, and given that much immutable data that is accessed concurrently is ideally stored in `TRef` structures inside the environment, it can make a lot of sense to port the polymorphism of `ZRef` into `TRef`.

Of course, we could have `ZTRef` and `TRef`, where `TRef` would be the less polymorphic version, similar to the relationship between `ZRef` and `Ref`. So the polymorphism would be opt-in.

This ticket will involve creating a version of `TRef` (proposed name: `ZTRef`) that is nearly API identical to `ZRef`, and then making a type / companion object alias for `TRef`.

vasilmkd:
Resolves #3260.

@jdegoes @adamgfraser

### DIFF
--- a/benchmarks/src/main/scala/zio/stm/STMRetryBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/stm/STMRetryBenchmark.scala
@@ -24,7 +24,7 @@ class STMRetryBenchmark {
   @Setup(Level.Trial)
   def setup(): Unit = {
     val data       = (1 to Size).toList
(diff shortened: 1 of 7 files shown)


vasilmkd:
@adamgfraser Thanks, it's easy when there's a lot of great code lying around. Scala 2.11 peculiarities should be fixed now.

PR #3269, opened 2020-04-03
add andTo and >+> to ZLayer

felher:
Here is my attempt to add the `>+>` operator to the ZLayer ops which allows for the linear chaining of topologically sorted layers.

Please let me know if we need tests for this. I'm happy to provide them but wasn't sure we need them, given that it's really only an alias for `++` and `>>>`, which are both covered by tests already.

Also, if I went overboard with the documentation, please let me know and I'll reduce it.

One last thing: If we want to settle on another name, I'm happy to make that change, too.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -2041,37 +2041,9 @@ object ZLayer {
       self.zipWithPar(that)(_.unionAll[ROut2](_))
 
     /**
-      * A named alias for `++`.
-      */
-    def and[E1 >: E, RIn2, ROut1 >: ROut, ROut2 <: Has[_]](
-      that: ZLayer[RIn2, E1, ROut2]
-    )(implicit tagged: Tagged[ROut2]): ZLayer[RIn with RIn2, E1, ROut1 with ROut2] =
-      self ++ that
-
-    /**
-     * Updates one of the services output by this layer.
-     */
-    def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
-      self >>> ZLayer.fromFunctionMany(_.update[A](f))
-
-    /**
-     * Returns a new layer that keeps the output of this layer and also adds the output of the other layer,
-     * given that this layer produces everything the other layer requires.
-     * This allows for the linear accumulation of layers:
-     * {{{
-     * def baker       : URLayer[Any                       , Baker]       = ???
-     * def ingredients : URLayer[Any                       , Ingredients] = ???
-     * def oven        : URLayer[Any                       , Oven]        = ???
-     * def dough       : URLayer[Baker with Ingredients    , Dough]       = ???
-     * def cake        : URLayer[Baker with Oven with Dough, Cake]        = ???
-     *
-     * def all =
-     *   baker >+>       // provides: Baker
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2053,6 +2053,38 @@ object ZLayer {
...
+     *
+     * def all =
+     *   baker >+>       // provides: Baker
+     *   ingredients >+> // provides: Baker & Ingredients
+     *   oven >+>        // provides: Baker & Ingredients & Oven
+     *   dough >+>       // provides: Baker & Ingredients & Oven & Dough
+     *   cake            // provides: Baker & Ingredients & Oven & Dough & Cake
+     * }}}
+     */
+    def >+>[E1 >: E, RIn2 >: ROut, ROut1 >: ROut, ROut2 <: Has[_]](
```
adamgfraser, line 2076: These two methods should have the same signature.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2053,6 +2053,38 @@ object ZLayer {
...
+      that: ZLayer[RIn2, E1, ROut2]
+    )(implicit tagged: Tagged[ROut2]): ZLayer[RIn, E1, ROut1 with ROut2] =
+      self ++ (self >>> that)
+
+    /**
+     * A named alias for `>+>`.
+     */
+    def andTo[E1 >: E, RIn2 >: ROut, ROut2 <: Has[_]](
```
adamgfraser, line 2084: Alphabetize with symbolic operators before alphanumeric methods.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2053,6 +2053,38 @@ object ZLayer {
      */
     def update[A: Tagged](f: A => A)(implicit ev: ROut <:< Has[A]): ZLayer[RIn, E, ROut] =
       self >>> ZLayer.fromFunctionMany(_.update[A](f))
+
+    /**
+     * Returns a new layer that keeps the output of this layer and also adds the output of the other layer,
```
adamgfraser, line 2058: I would reduce this.

adamgfraser: Maybe:
```scala
/**
   * Feeds the output services of this layer into the input of the specified
   * layer, resulting in a new layer with the inputs of this layer, and the
   * outputs of both this layer and the specified layer.
   */
```

core/shared/src/main/scala/zio/ZLayer.scala
felher: Done :+1:

core/shared/src/main/scala/zio/ZLayer.scala
felher: Done. Thanks for reviewing :)

mkotsbak: The cake example could be useful to include somwhere else, like at website in page https://zio.dev/docs/howto/howto_use_layers I don't think the documentation there is complete enough yet.

adamgfraser: Yes, I think we could use a section there discussing passed through versus hidden dependencies.

adamgfraser: I am working on now.

PR #3281, opened 2020-04-04
ZStream.aggregate, ZTransducer.collectAllWhile

### LINKED ISSUE #3290: New Stream encoding: very basic transducing doesn't work
I tried to write basic examples and failed so far:

```
testM("trans1"){
        val x = ZStream[Int](1,2,3).transduce[Any, Nothing, Int, Chunk[Int]](idSink.toTransducer).runCollect.map(l => l.flatMap(_.toSeq))
        assertM(x)(equalTo(List(1,2,3)))
      } @@ timeout(3.seconds),
testM("trans2"){
        val x = ZStream[Int](1,2,3).transduce[Any, Nothing, Int, Int](ZTransducer.chunkN(10)).runCollect
        assertM(x)(equalTo(List(1,2,3)))
} @@ timeout(3.seconds)

//returns incoming chunks untouched
def idSink[I]: ZSink[Any, Nothing, I, Chunk[I]] =
    ZSink(ZManaged.succeed({
      case Some(ch) => {
        //println("-- " + ch)
        Push.emit(ch)
      }
      case None => {
        //println("--None")
        Push.more
      }
    }))
```

Both tests fail due to timeout.
In the first one `idSink` continuously receives `None` input.

Pinging @regiskuckaertz and @iravid

regiskuckaertz:
Closes #3290 

ZTransducer is missing map and flatMap before we can uncomment all the tests, I started working on it but prefer to do a PR for that only.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -642,6 +642,35 @@ sealed trait Chunk[+A] { self =>
   }
 
   /**
+   * Effectfully maps the elements of this chunk.
+   */
+  final def mapManaged[R, E, B](f: A => ZManaged[R, E, B]): ZManaged[R, E, Chunk[B]] = {
+    val len                             = self.length
+    var array: ZManaged[R, E, Array[B]] = ZManaged.succeedNow(null.asInstanceOf[Array[B]])
+    var i                               = 0
+
+    while (i < len) {
+      val j = i
+      array = array.zipWith(f(self(j))) { (array, b) =>
+        val array2 = if (array == null) {
+          implicit val B: ClassTag[B] = Chunk.Tags.fromValue(b)
+          Array.ofDim[B](len)
+        } else array
+
+        array2(j) = b
+        array2
+      }
+
+      i += 1
+    }
+
+    array.map(array =>
+      if (array == null) Chunk.empty
+      else Chunk.fromArray(array)
+    )
+  }
+
+  /**
    * Effectfully maps the elements of this chunk in parallel.
    */
   final def mapMPar[R, E, B](f: A => ZIO[R, E, B]): ZIO[R, E, Chunk[B]] = {
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -46,7 +46,7 @@ object ZStreamSpec extends ZIOBaseSpec {
       ),
       suite("aggregate")(
         testM("aggregate") {
-          val s = ZStream('1', '2', ',', '3', '4')
+          val s      = ZStream('1', '2', ',', '3', '4')
           val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt)
 
           assertM(s.aggregate(parser).runCollect)(equalTo(List(12, 34)))
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -102,8 +102,8 @@ object ZStreamSpec extends ZIOBaseSpec {
         // },
         testM("propagate managed error") {
           val fail = "I'm such a failure!"
-          val t    = ZManaged.fail(fail)
-          assertM(ZStream(1, 2, 3).aggregateManaged(t).runCollect.either)(isLeft(equalTo(fail)))
+          val t    = ZTransducer.fail(fail)
+          assertM(ZStream(1, 2, 3).aggregate(t).runCollect.either)(isLeft(equalTo(fail)))
         }
       ),
       suite("bracket")(
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -45,12 +45,12 @@ object ZStreamSpec extends ZIOBaseSpec {
         })
       ),
       suite("aggregate")(
-        // testM("aggregate") {
-        //   val s = ZStream('1', '2', ',', '3', '4')
-        //   val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt) <* ZTransducer.collectAllWhile[Char](_ == ',')
+        testM("aggregate") {
+          val s = ZStream('1', '2', ',', '3', '4')
+          val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt)
 
-        //   assertM(s.aggregate(parser).runCollect)(equalTo(List(12, 34)))
-        // },
+          assertM(s.aggregate(parser).runCollect)(equalTo(List(12, 34)))
+        },
         // testM("no remainder") {
         //   val t = ZTransducer.fold(100)(_ % 2 == 0)((s, a: Int) => (s + a, Chunk[Int]()))
         //   assertM(ZStream(1, 2, 3, 4).aggregate(t).runCollect)(equalTo(List(101, 105, 104)))
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -46,9 +46,8 @@ object ZStreamSpec extends ZIOBaseSpec {
       ),
       suite("aggregate")(
         // testM("aggregate") {
-        //   val s = Stream('1', '2', ',', '3', '4')
-        //   val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt) <* ZSink
-        //     .collectAllWhile[Char](_ == ',')
+        //   val s = ZStream('1', '2', ',', '3', '4')
+        //   val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt) <* ZTransducer.collectAllWhile[Char](_ == ',')
 
         //   assertM(s.aggregate(parser).runCollect)(equalTo(List(12, 34)))
         // },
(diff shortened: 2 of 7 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,46 @@ abstract class ZStream[-R, +E, +O](
...
+   * Applies an aggregator to the stream, which converts one or more elements
+   * of type `A` into elements of type `B`.
+   */
+  def aggregate[R1 <: R, E1 >: E, O1 >: O, P](sink: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] =
+    aggregateManaged(ZManaged.succeedNow(sink))
+
+  /**
+   * Applies a managed transducer to the stream, converting elements of type `A` into elements of type `B`.
+   */
+  final def aggregateManaged[R1 <: R, E1 >: E, O1 >: O, P](
```
iravid, line 95: I think we can just do the `aggregate` variant - a transducer already has a `ZManaged` built into it.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -87,40 +87,28 @@ abstract class ZStream[-R, +E, +O](
...
+        pull <- self.process
+        push <- sink.push
+        done <- ZRef.makeManaged(false)
+        run = {
+          def go: ZIO[R1, Option[E1], Chunk[P]] = done.get.flatMap {
+            if (_)
+              push(None)
+            else
+              pull
+                .foldM(
```
iravid, line 101: Maybe just `foldCauseM(Cause.sequenceCauseOption(...), ...)` here instead of `foldM.catchAllCause`?

regiskuckaertz: Hmm, I don't think you can remove the `catchAllCause` as it is necessary to handle the "next" signal from the transducer and pull again, right? I switched to `foldM` to avoid the redundancy in case of a defect.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,34 @@ abstract class ZStream[-R, +E, +O](
...
+  def aggregate[R1 <: R, E1 >: E, O1 >: O, P](sink: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] =
+    ZStream {
+      for {
+        pull <- self.process
+        push <- sink.push
+        done <- ZRef.makeManaged(false)
+        run = {
+          def go: ZIO[R1, Option[E1], Chunk[P]] = done.get.flatMap {
+            if (_)
+              push(None)
```
regiskuckaertz, line 98: I'm still not convinced this is correct because in the transducer failing with `None` means give me more, whereas failing with `None` in the stream means this is the end.

iravid: Oh wait I think I am confused. Transducer failing with `None` should mean `end`, not give me more. A transducer can just send an empty Chunk if it doesn't have anything to send.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
iravid: Right! Good point. It does look like you can replace `catchAllCause` with `catchAll` though, because you're just forwarding the cause with `Pull.halt` on the non-typed-failure case.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: Ha. Ok so that changes everything 😆 give me a sec.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,37 @@ abstract class ZStream[-R, +E, +O](
...
+  def aggregate[R1 <: R, E1 >: E, O1 >: O, P](sink: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] =
+    ZStream {
+      for {
+        pull <- self.process
+        push <- sink.push
+        done <- ZRef.makeManaged(false)
+        run = {
+          def go: ZIO[R1, Option[E1], Chunk[P]] = done.get.flatMap {
+            if (_)
+              push(None).repeat(Schedule.doWhile(_.isEmpty))
```
regiskuckaertz, line 98: What do you think of this @iravid ?

streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
```diff
@@ -44,6 +44,68 @@ object ZStreamSpec extends ZIOBaseSpec {
           } yield assert(res1)(fails(equalTo(s))) && assert(res2)(fails(equalTo(s)))
         })
       ),
+      suite("aggregate")(
+        testM("aggregate") {
+          val s      = ZStream('1', '2', ',', '3', '4')
+          val parser = ZTransducer.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt)
```
regiskuckaertz, line 50: The current version of the transducer for this test is:
```
val parser = ZSink.collectAllWhile[Char](_.isDigit).map(_.mkString.toInt) <* ZSink
          .collectAllWhile[Char](_ == ',')
```
Nice simplification. (Also, ZTransducers may not form monads after all)

simpadjo:
`Stream#aggregate` used to be implemented as a special case of `aggregateManaged`. 
Will `aggregateManaged` exist or managed transducers make no sense?

regiskuckaertz:
@simpadjo transducers are manager resources so it would be redundant

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -62,22 +65,55 @@ object ZTransducer {
...
     }
 
+  def collectAllWhile[I](p: I => Boolean): ZTransducer[Any, Nothing, I, List[I]] =
+    ZTransducer {
+      for {
+        buffered <- ZRef.makeManaged[Chunk[I]](Chunk.empty)
+        push = { (in: Option[Chunk[I]]) =>
+          buffered.modify { buf0 =>
+            val buf = in.foldLeft(buf0)(_ ++ _)
+            if (buf.isEmpty)
```
iravid, line 94: This doesn’t look right - shouldn’t we check if `in` was originally `None`?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
iravid: It’s fine. A direct loop would be more performant but this should probably be an uncommon code path.

iravid (commented):
Looks good! One small comment about collectAllWhile.

@simpadjo could you check your transducer tests with the identity sink against the implementation here?

simpadjo:
@iravid The second test from https://github.com/zio/zio/issues/3290 is passing now, cool.
The first is still red but I guess it's a separate issue with `toTransducer`.

iravid:
Awesome!

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: if it is then `buf == buf0` and if it isn't empty then my guess is we still have to go through the remainder. Is that correct?

iravid: Yes, that sounds right. Except that if it is `None`, we want to flush the remainder anyway (even if we haven't seen an `I` that satisfies `p`).

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,37 @@ abstract class ZStream[-R, +E, +O](
...
+            else
+              pull
+                .foldM(
+                  _.fold(done.set(true) *> push(None))(Pull.fail(_)),
+                  os => push(Some(os))
+                )
+                .foldCauseM(
+                  Cause.sequenceCauseOption(_).fold(go)(Pull.halt(_)),
```
iravid, line 106: The `fold(go)` here looks suspicious, because we can get here both from `push(Some(os))` and from `push(None)`.

- From `push(Some(os))` - this looks like a protocol violation, right? Do we allow transducers to end on their own accord? I'm not clear on this 🧐
- From `push(None)` - this will recurse again if the transducer ended in response to the stream ending. Ah, it'll end on the next iteration because of `done set true`. Ok, I'd prefer to `ZIO.fail(None)` directly here but not critical 👍 

So let's just clarify the first point. What do you think?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -652,6 +652,35 @@ sealed trait Chunk[+A] { self =>
     )
   }
 
+  /**
+   * Effectfully maps the elements of this chunk.
+   */
+  final def mapManaged[R, E, B](f: A => ZManaged[R, E, B]): ZManaged[R, E, Chunk[B]] = {
```
iravid, line 658: Do we need this one? It's not used in this PR.

iravid: Oh sorry I see the forwarder added in `ZManaged`.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
regiskuckaertz: I think this also happens with sinks, the way they signal end is by failing which, thanks to ZIO being well-typed, you have to account for everywhere. That's another instance of what we were trying to fix with `SignalFunction`. Not sure there is a solution other than writing laws and throwing all the transducers at them.

regiskuckaertz: I could refactor so that we die in case of a protocol error, but somehow that feels imperative. Note that having two different paths for the `None` and `Some` cases won't change the fact that the transducer may fail.

iravid: So I was thinking about this for a bit and perhaps it’s useful to allow transducers to signal end? E.g. that allows to create transducers that can short-circuit a stream when a predicate is violated.

regiskuckaertz: Can you give an example? With `collectAllWhile`, I took the position that the transducer would just drop all elements violating the predicate, but still keep on accepting the rest of the stream. Is there another class of transducers where that wouldn't be possible?

iravid (approved):
Let’s get this in and refine the ZTransducer contract more later. Great work!

PR #3283, opened 2020-04-04
Varargs support in accessible

senia-psm:
Support for varargs (`def m(arg: Int*)`) in accessible macro.


ioleo:
Great addition, thank you! :pray: 
Could you perhaps also add varargs support to `@mockable` macro?

Regarding the implementation, I don't think we have access to internal `isRepeatedParamType`, so I'm OK with copying.

senia-psm:
@ioleo it's in my nearest plans. Right before updating `zio-intellij` support for both. Though I'd prefer to do it in another PR.

senia-psm:
@ioleo see #3292 for varargs support in `mockable`.

PR #3291, opened 2020-04-05
Add d4s as library that has direct ZIO support.

VladPodilnyk:


VladPodilnyk:
> Thanks for adding!
😊🙂

PR #3293, opened 2020-04-05
ZIO Stream: Remove More Extraneous Type Parameters

adamgfraser:
Did a pass through stream combinators to remove other unnecessary type parameters.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1448,28 +1448,28 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
   /**
    * Executes a pure fold over the stream of values - reduces all elements in the stream to a value of type `S`.
    */
-  final def fold[A1 >: A, S](s: S)(f: (S, A1) => S): ZIO[R, E, S] =
-    foldWhileManagedM[R, E, A1, S](s)(_ => true)((s, a) => ZIO.succeedNow(f(s, a))).use(ZIO.succeedNow)
+  final def fold[S](s: S)(f: (S, A) => S): ZIO[R, E, S] =
+    foldWhileManagedM(s)(_ => true)((s, a) => ZIO.succeedNow(f(s, a))).use(ZIO.succeedNow)
 
   /**
    * Executes an effectful fold over the stream of values.
    */
-  final def foldM[R1 <: R, E1 >: E, A1 >: A, S](s: S)(f: (S, A1) => ZIO[R1, E1, S]): ZIO[R1, E1, S] =
-    foldWhileManagedM[R1, E1, A1, S](s)(_ => true)(f).use(ZIO.succeedNow)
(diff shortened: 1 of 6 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1773,26 +1771,26 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
    * Enqueues elements of this stream into a queue. Stream failure and ending will also be
    * signalled.
    */
-  final def into[R1 <: R, E1 >: E, A1 >: A](
-    queue: ZQueue[R1, Nothing, Nothing, Any, Take[E1, A1], Any]
```
regiskuckaertz, line 1777: Surely this one could stay, right? (though the result type can be `ZIO[R1, E, Unit]`)

adamgfraser: The method is staying. I don't think the additional type parameters add any flexibility and could cause type inference issues in some situations. We have a `ZStream[R, E, A]` and we can send it into any `ZQueue` that can accept values of type `Take[E, A]`. Because the input type accepted by the `Queue` is contravariant we can still send values to a `Queue` that can accept a supertype of `E` or `A`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1494,18 +1494,16 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
-    foldWhileManagedM[R1, E1, A1, S](s)(cont)(f).use(ZIO.succeedNow)
+  final def foldWhileM[R1 <: R, E1 >: E, S](s: S)(cont: S => Boolean)(f: (S, A) => ZIO[R1, E1, S]): ZIO[R1, E1, S] =
+    foldWhileManagedM[R1, E1, S](s)(cont)(f).use(ZIO.succeedNow)
 
   /**
    * Executes a pure fold over the stream of values.
    * Returns a Managed value that represents the scope of the stream.
    * Stops the fold early when the condition is not fulfilled.
    */
-  def foldWhileManaged[A1 >: A, S](s: S)(cont: S => Boolean)(f: (S, A1) => S): ZManaged[R, E, S] =
```
iravid, line 1507: The fold* changes are counter-intuitive (because you'd think, well, this function can use any step function that takes a super-type of `A`), but the inference improvements are clear so I'm 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: I think that due to the covariance of `E` and `A` on `Take` and the covariance of the out type of `ZQueue`, these are redundant. Is this accurate @adamgfraser?

iravid: Oops missed the comment from Adam. 👍

PR #3302, opened 2020-04-06
Add ZStream.aggregateAsync

regiskuckaertz:
I used a reentrant lock instead of a semaphore, but there may be a performance cost so happy to change that. Also, nice simplification over [the existing one](https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L103-L266).

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -94,16 +94,12 @@ abstract class ZStream[-R, +E, +O](
    * Any sink can be used here, but see [[Sink.foldWeightedM]] and [[Sink.foldUntilM]] for
    * sinks that cover the common usecases.
    */
-  final def aggregateAsync[R1 <: R, E1 >: E, O1 >: O, P](transducer: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] = {
-    import zio.Semaphore
-
+  final def aggregateAsync[R1 <: R, E1 >: E, O1 >: O, P](transducer: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] =
     ZStream {
       for {
         pull   <- self.process
         push   <- transducer.push
-        lock   <- Semaphore.make(1).toManaged_
-        bucket <- ZRef.makeManaged[Take[E1, P]](Take.End)
-        start  <- Promise.make[Nothing, Unit].toManaged_
+        bucket <- ZQueue.bounded[Take[E1, P]](1).toManaged(_.shutdown)
         done   <- ZRef.makeManaged(false)
         produce = {
           def finish: ZIO[R1, Nothing, Boolean] =
@@ -112,33 +108,33 @@ abstract class ZStream[-R, +E, +O](
               .foldCauseM(
                 Cause
                   .sequenceCauseOption(_)
-                  .fold(bucket.set(Take.End))(c => bucket.set(Exit.halt(c.map(Some(_))))),
-                ps => bucket.set(Exit.succeed(ps))
+                  .fold(bucket.offer(Take.End))(c => bucket.offer(Exit.halt(c.map(Some(_))))),
+                ps => bucket.offer(Exit.succeed(ps))
               )
               .as(false)
 
           def maybeDrain(os: Chunk[O]): ZIO[R1, Nothing, Boolean] =
             push(Some(os))
               .foldCauseM(
-                Cause.sequenceCauseOption(_).fold(IO.unit)(c => bucket.set(Exit.halt(c.map(Some(_))))),
-                ps => bucket.set(Exit.succeed(ps))
+                Cause.sequenceCauseOption(_).fold(IO.unit)(c => bucket.offer(Exit.halt(c.map(Some(_)))).unit),
+                ps => bucket.offer(Exit.succeed(ps))
               )
               .as(true)
 
           val go: ZIO[R1, Nothing, Boolean] =
             pull.foldCauseM(
-              Cause.sequenceCauseOption(_).fold(finish)(c => bucket.set(Exit.halt(c.map(Some(_)))).as(true)),
+              Cause.sequenceCauseOption(_).fold(finish)(c => bucket.offer(Exit.halt(c.map(Some(_)))).as(true)),
               maybeDrain
             )
 
-          lock.withPermit(start.succeed(()) *> go).repeat(Schedule.doWhile(identity))
+          go.repeat(Schedule.doWhile(identity))
         }
 
         consume = {
           lazy val go: ZIO[R1, Option[E1], Chunk[P]] =
-            lock.withPermit(bucket.get.flatMap(take => done.set(true).when(take == Take.End) *> Pull.fromTake(take)))
+            bucket.take.flatMap(take => done.set(true).when(take == Take.End) *> Pull.fromTake(take))
 
-          start.await *> done.get.flatMap {
+          done.get.flatMap {
             if (_)
               Pull.end
             else
@@ -147,7 +143,6 @@ abstract class ZStream[-R, +E, +O](
         }
       } yield produce.fork *> consume
     }
-  }
 
   /**
    * Maps the success values of this stream to the specified constant value.
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -103,7 +103,7 @@ abstract class ZStream[-R, +E, +O](
         push   <- transducer.push
         lock   <- Semaphore.make(1).toManaged_
         bucket <- ZRef.makeManaged[Take[E1, P]](Take.End)
-        start <- Promise.make[Nothing, Unit].toManaged_
+        start  <- Promise.make[Nothing, Unit].toManaged_
         done   <- ZRef.makeManaged(false)
         produce = {
           def finish: ZIO[R1, Nothing, Boolean] =
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -95,14 +95,15 @@ abstract class ZStream[-R, +E, +O](
    * sinks that cover the common usecases.
    */
   final def aggregateAsync[R1 <: R, E1 >: E, O1 >: O, P](transducer: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] = {
-    import zio.stm.TReentrantLock
+    import zio.Semaphore
 
     ZStream {
       for {
         pull   <- self.process
         push   <- transducer.push
-        lock   <- TReentrantLock.makeCommit.toManaged_
-        bucket <- ZRef.makeManaged[Option[Take[E1, P]]](None)
+        lock   <- Semaphore.make(1).toManaged_
+        bucket <- ZRef.makeManaged[Take[E1, P]](Take.End)
+        start <- Promise.make[Nothing, Unit].toManaged_
         done   <- ZRef.makeManaged(false)
         produce = {
           def finish: ZIO[R1, Nothing, Boolean] =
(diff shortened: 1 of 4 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,73 @@ abstract class ZStream[-R, +E, +O](
...
+                  .sequenceCauseOption(_)
+                  .fold(bucket.set(Some(Take.End)))(c => bucket.set(Some(Exit.halt(c.map(Some(_)))))),
+                ps => bucket.set(Some(Exit.succeed(ps)))
+              )
+              .as(false)
+
+          def maybeDrain(os: Chunk[O]): ZIO[R1, Nothing, Boolean] =
+            push(Some(os))
+              .foldCauseM(
+                Cause.sequenceCauseOption(_).fold(IO.unit)(c => bucket.set(Some(Exit.halt(c.map(Some(_)))))),
```
regiskuckaertz, line 121: Not sure what to do if the stream is not done but the transducer signals the end?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,79 @@ abstract class ZStream[-R, +E, +O](
...
+              if (ps.nonEmpty)
+                ZIO.succeedNow(true)
+              else
+                pull.foldCauseM(
+                  Cause.sequenceCauseOption(_).fold(finish)(c => bucket.set(Some(Exit.halt(c.map(Some(_))))).as(true)),
+                  maybeDrain
+                )
+            }
+
+          lock.writeLock.use(_ => go).repeat(Schedule.doWhile(identity))
```
regiskuckaertz, line 138: Hmm, thinking about it more, this will loop if downstream is slow, perhaps I should just switch to a bounded queue of 1 instead.

iravid: I'm fairly sure I went through that thought process when writing the original version and you have to use `Ref+Promise+Semaphore` for the state machine transitions.

regiskuckaertz: Done.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,73 @@ abstract class ZStream[-R, +E, +O](
...
+   *
+   * This operator divides the stream into two asynchronous "islands". Operators upstream
+   * of this operator run on one fiber, while downstream operators run on another. Whenever
+   * the downstream fiber is busy processing elements, the upstream fiber will feed elements
+   * into the sink until it signals completion.
+   *
+   * Any sink can be used here, but see [[Sink.foldWeightedM]] and [[Sink.foldUntilM]] for
+   * sinks that cover the common usecases.
+   */
+  final def aggregateAsync[R1 <: R, E1 >: E, O1 >: O, P](transducer: ZTransducer[R1, E1, O1, P]): ZStream[R1, E1, P] = {
```
iravid, line 97: So I think this implementation needs a bit more work to achieve the previous guarantees. Here are the main problem points:
1. The consumer will keep spinning and taking data from the bucket even if the producer hasn't produced anything. You basically have to take the mechanism you created with `start` and thread it throughout the whole process. This can be done by the producer storing a `Promise` inside the `Ref` that the consumer can await to know when there's data available in the bucket.
2. The producer has no backpressuring mechanism and will overwrite data that hasn't been taken by the consumer (line 124). This can be solved by attaching a `Promise` to the bucket that will be resolved by the consumer when it  took the produced result.
3. I actually think the Semaphore is now redundant because you can do all the bucket swaps atomically.
4. Looking at your implementation, it may be worth exploring the queue option! I think the current encoding removes some of the constraints the previous implementation had.

regiskuckaertz: 🤦 I don't know why I keep thinking there is going to be a strict order of takes. Will address this.

regiskuckaertz: I think a bounded queue of 1 is exactly the semantic I'm looking for. Like an `IVar` that can be emptied and filled multiple times.

regiskuckaertz:
Ok, updated. I think it fits perfectly, and now that I start looking at `aggregateAsyncWithin`, it may be possible to express the former in terms of the latter.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -82,6 +82,68 @@ abstract class ZStream[-R, +E, +O](
...
+                  .fold(bucket.offer(Take.End))(c => bucket.offer(Exit.halt(c.map(Some(_))))),
+                ps => bucket.offer(Exit.succeed(ps))
+              )
+              .as(false)
+
+          def maybeDrain(os: Chunk[O]): ZIO[R1, Nothing, Boolean] =
+            push(Some(os))
+              .foldCauseM(
+                Cause.sequenceCauseOption(_).fold(IO.unit)(c => bucket.offer(Exit.halt(c.map(Some(_)))).unit),
+                ps => bucket.offer(Exit.succeed(ps))
```
iravid, line 120: Another observation - this implementation allows the operator to get started on processing the next chunk of data before the previous chunk was taken from the queue. Arguable whether this is desirable or not, but it’s different from the previous implementation.

iravid:
Sorry for the multiple rounds of comments! It’s a pretty tricky operator and the re-implementation is proves challenging to completely review :-)

regiskuckaertz:
@iravid Absolutely not! Might be worth line up the set of properties that these combinators expose.

>  this implementation allows the operator to get started on processing the next chunk of data before the previous chunk was taken from the queue

🤔 isn't it the whole purpose of being async, that producers and consumers can work independently?

iravid:
> 🤔 isn't it the whole purpose of being async, that producers and consumers can work independently?

It is, but it is also about limiting “lookahead”. Currently we will be buffering n + 1 chunks in memory, with n being the queue size. This makes reasoning about the memory usage a bit harder.

regiskuckaertz:
That makes sense. I'd be interested to learn more about the framework you are using to reason about space-time cost in streams, do you have any resource to share?

> The desired behavior here is to flush out the buffered data whenever mapM pulls the upstream - regardless of whether chunkN already buffered 50 elements.

But then doesn't this mean that every transducer is basically a "best effort" transformation? And it becomes impossible to express "I _really_ do want 50 elements"?

iravid:
> That makes sense. I'd be interested to learn more about the framework you are using to reason about space-time cost in streams, do you have any resource to share?
Nothing formal 🤷🏼‍♂️
> 
> But then doesn't this mean that every transducer is basically a "best effort" transformation? And it becomes impossible to express "I _really_ do want 50 elements"?

Yes, for this combinator. We’re trying to balance latency and throughput. For the “really strictly 50 elements”, that’s “aggregate”. For the “50 elements but up to a timeout”, that’s aggregateAsyncWithin.

regiskuckaertz:
so yeah, I think `push(None)` fits well in this canvas. I will tackle this after house cleaning 😅

iravid:
Awesome. 

One last thought (this is also relevant to our discussion in #3281): the contract for transducers should be that `push(None)` means "flush everything you have". E.g., the law should be `assertM(push(None) *> push(None))(equalTo(Chunk.empty))`.

Given this, we can decide that transducers should never end - e.g. they should be `ZIO[R, E, Chunk[O]]`. What do you think? We can make this change after these two branches are merged.

regiskuckaertz:
Let's say the last chunk of a stream is `Chunk(1,2,3,10,4,5,6,10,7,8,9)` and the transducer is `collectAllWhile(_ < 10)`: sending `None` should then process the whole chunk in one go? But then if it was the second to last chunk of the stream, the transducer would output three small chunks?

iravid:
Ah that's a good point - transducers that have buffered more data than would fit in a single output chunk.

Another example would be `chunkN(10)` where the last chunk of the stream had 100 elements.This is actually a bigger deficiency than the last chunk, because in this case we're pulling upstream even though we have data buffered in the transducer.

Maybe this is telling us that we need a sum type on the transducer's output:
- `NeedMoreInput(Chunk[I])` - here's the last chunk I have, I need more data to continue
- `HaveOutput(Chunk[I])` - here's some data I have, and I have more to provide

iravid:
Another possibility is that transducers like `chunkN` and `collectAllWhile` should have an output type of `Chunk[A] / List[A]`. So you can always output the individual elements in a single chunk. In your example it'd always be `Chunk(List(1, 2, 3), List(4, 5, 6), List(7, 8, 9))` regardless if it's the last chunk or not.

Maybe that's simpler.

regiskuckaertz:
I like the idea of a sum type, actually when I started writing `aggregate` and found that it was impossible for a stream to know when a transducer was empty, I played with outputting `(Boolean, Chunk[O])` instead. What about elements that belong to different frames upstream but should be part of the same one downstream then? If you have `Chunk(1,2,3,10,4,5,6), Chunk(7,8,9)`, I built `collectAllWhile` to yield `Chunk(List(1,2,3)),Chunk(List(4,5,6,7,8,9))`

iravid:
So I think the sum type isn’t actually needed because you can always output multiple elements in the chunk.

The implementation we have now for collectAllWhile is fine; if I understand correctly, it would do the following:
```
Push(Some(Chunk(1, 2, 3, 10, 4, 5, 6)) => Chunk(1, 2, 3)
Push(Some(Chunk(7, 8, 9, 10))) => Chunk(4, 5, 6, 7, 8, 9)

Push(Some(Chunk(1, 2, 3, 10, 4, 5, 6)) => Chunk(1, 2, 3)
Push(None) => Chunk(4, 5, 6)
```

If these two sequences are correct, then we’re good. Dropping the `Option` from the transducer’s failure type still appears to make sense. I’ll experiment with that now.

regiskuckaertz:
Correct. Oh but I totally forgot: failing with None is how I currently signal the internal buffer is empty, and an empty chunk signals “I need more”.

iravid:
That should be fine, because the `aggregate/run` operators can now signal `push(None)` once and move on, instead of waiting for a `Fail(None)`.

regiskuckaertz:
@iravid can we also say something about concurrent calls to `push`, i.e. it would be nice if we could consider `push(None)` as an Out-of-Band sort of thing. Let's say the the producer is busy aggregating (so in `BatchMiddle` in the existing implementation) and the consumer is pulling, it would be nice if we could just write (simplifying):
```scala
val strand1 = pull.flatMap(os => push(Some(os)))
val strand2 = notifyProducer.await *> push(None)

(strand1 race strand2).repeat(Schedule.doWhile(_.isEmpty).run.flatMap(bucket.offer)
```

iravid:
Concurrent calls to pull/push is something we haven't considered yet. I don't know if it's smart to introduce those as part of the protocol because that means all transducers will now have to account for concurrency, instead of specific concurrent operators (like aggregateAsync and possibly others) accounting for that.

regiskuckaertz:
If you merge your PR I'll be able to do some cleanup (notably remove that spurious `succeed(true)`). In this latest change, pulling will force the producer to flush the current element, but I also noticed that I didn't take into account the fact that the transducer might output an empty chunk, in which case there is no other choice that to keep pulling.

iravid:
@regiskuckaertz Merged it - will take a look in a bit 👍

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -110,6 +110,68 @@ abstract class ZStream[-R, +E, +O](
...
+              if (_)
+                drain(None)(iterate)
+              else
+                pull.foldCauseM(
+                  Cause.sequenceCauseOption(_).fold(drain(None)(finish))(c => bucket.offer(Exit.halt(c.map(Some(_))))),
+                  os => drain(Some(os))(iterate)
+                )
+            }
+
+          go.repeat(Schedule.doWhile(identity))
```
iravid, line 158: All of the methods will always end with `true`, so this will never stop. Should we do `as(false)` after enqueueing `Take.End`?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -110,6 +110,68 @@ abstract class ZStream[-R, +E, +O](
...
+        pull   <- self.process
+        push   <- transducer.push
+        bucket <- ZQueue.bounded[Take[E1, P]](1).toManaged(_.shutdown)
+        flush  <- ZRef.makeManaged(false)
+        done   <- ZRef.makeManaged(false)
+        produce = {
+          // Upstream is done, we need to flush the transducer. If the output is
+          // empty, we send the end signal downstream.
+          def finish(ps: Chunk[P]): URIO[R1, Boolean] =
+            bucket.offer(if (ps.isEmpty) Take.End else Exit.succeed(ps)) as false
```
iravid, line 137: We should probably only do the `as false` if we enqueued `Take.End` because if we exit the loop after `Exit.succeed`, we won't enqueue the `Take.End`.

There's a simplification though: the contract for transducers says that `push(None)` empties everything in the transducer, so you don't need to push again afterwards. You can do `bucket.offerAll(List(Exit.succeed(ps), Take.End)) as false`, no?

regiskuckaertz: Yeah I just saw that 🤦 I think it works that way indeed 👍

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -110,6 +110,68 @@ abstract class ZStream[-R, +E, +O](
...
+            flush.set(true) *> bucket.take.flatMap(take => done.set(true).when(take == Take.End) *> Pull.fromTake(take))
+
+          done.get.flatMap {
+            if (_)
+              Pull.end
+            else
+              go
+          }
+        }
+      } yield produce.fork *> consume
```
iravid, line 172: Alright, last comment and we can merge! Can we move the `produce.fork` into the `ZManaged` layer as `_ <- produce.forkManaged`?

regiskuckaertz:
Nice 🎉 I wanted to tackle the one with Schedule but I think I'll first tackle a few more transducers.

iravid:
That'd be great. We need those to unlock some of the tests.

PR #3303, opened 2020-04-06
Port ZSink fold, foldleft

simpadjo:
Bring back `contFn` to `ZSink#fold`
Add `foldLeft`
Port existing tests that are still relevant
Add couple of new tests

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
@@ -364,7 +364,7 @@ object ZSink {
         if (shouldStart) {
           for {
             state <- Ref.make(z).toManaged_
-            push: zio.stream.experimental.ZSink.Push[R, E, I, S] = (is: Option[Chunk[I]]) =>
+            push = (is: Option[Chunk[I]]) =>
               is match {
                 case None => state.get.flatMap(Push.emit)
                 case Some(is) => {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -356,29 +356,52 @@ object ZSink {
...
-          is match {
-            case None => state.get.flatMap(s => ZIO.fail(Right(s)))
-            case Some(is) =>
-              state.get
-                .flatMap(is.foldM(_)(f))
-                .flatMap(state.set)
-                .mapError(Left(_))
-          }
-      } yield push
+      ZManaged.succeed(contFn(z)).flatMap { shouldStart =>
```
regiskuckaertz, line 363: Looks like there is no effect to capture here:
```scala
if (contFn(z))
  for { ... }
else
  ZManaged.succeed(_ => Push.emit(z))
```

simpadjo: Thank you! 
I did like that to ensure laziness of condition evaluation. 
Seems that we care about it

iravid, line 363: How about the simpler:
```scala
if (!contFn(z)) ZManaged.succeed(_ => Push.emit(z))
else for { ... } yield push
```
?

simpadjo: I though that we want it to be lazily evaluated

iravid: I don't think suspending that is particularly important, so let's go with the simpler version.

PR #3306, opened 2020-04-06
Add `ZStream#haltWhen(IO)` and `ZStream#interruptWhen(IO)`

jeremyrsmith:


jeremyrsmith:
Ping @iravid

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1677,6 +1677,17 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       } yield pull
     }
 
+  /**
+   * Halts the evaluation of this stream when the provided IO completes. The given IO
+   * will be evaluated as part of this stream, and its success will be discarded.
+   *
+   * If the IO completes with a failure, the stream will emit that failure.
+   */
+  final def haltWhen[E1 >: E](io: IO[E1, Unit]): ZStream[R, E1, A] =
```
jeremyrsmith, line 1686: There is probably a better way to do this one

iravid: Hmm. Why don't we implement `haltWhen(Promise)` in terms of this combinator? I think it should be simpler.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1767,6 +1780,15 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
       } yield pull
     }
 
+  /**
+   * Interrupts the evaluation of this stream when the provided promise resolves. This
+   * combinator will also interrupt any in-progress element being pulled from upstream.
+   *
+   * If the promise completes with a failure, the stream will emit that failure.
+   */
+  final def interruptWhen[E1 >: E](p: Promise[E1, _]): ZStream[R, E1, A] =
```
iravid, line 1789: Could you change the signature while you're here?
```suggestion
  final def interruptWhen[E1 >: E](p: Promise[E1, Any]): ZStream[R, E1, A] =
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
jeremyrsmith: Agreed, just didn't want to disrupt the existing logic (which depends on `poll`) too much.

iravid: Oh, I forgot it does that. In that case your implementation may be the simpler one. Let me know if so and happy to merge as is.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jeremyrsmith: Not sure this would have the desired outcome – `Promise` is invariant, so you'd have to exactly pass a `Promise[E1, Any]`. The existential is unfortunate here, but it doesn't appear to be avoidable unless `Promise` is made covariant.

The `IO` variants could certainly take `IO[E, Any]`, but now that I think about it, that's basically tantamount to a value discard, which is usually considered a wart. So I changed them to take `IO[E, Unit]` (it's easy to explicitly discard the value by just calling `.unit` on the argument)

streams/shared/src/main/scala/zio/stream/ZStream.scala
jeremyrsmith: I *think* an alternate implementation of `haltWhen(IO)` could be like this, and then `haltWhen(Promise)` could be implemented in terms of `haltWhen(p.await)`. But I'm probably missing some subtlety.

```scala
  final def haltWhen[E1 >: E](io: IO[E1, Unit]): ZStream[R, E1, A] =
    ZStream {
      for {
        as    <- self.process
        runIO <- (io.asSomeError *> Pull.end).forkManaged
      } yield as.raceFirst(runIO.join)
    }
```

jeremyrsmith: Actually, this implementation fails the test because it doesn't explicitly check the IO *first*, which the test expects because it is expecting an already-successful `ZStream` to fail if the `IO` has failed. Not sure if that's more an issue with the expectation or the implementation (or maybe that's the subtlety I'm missing 😁 )

jeremyrsmith:
Thinking about this more, there are two subtly different possible behaviors for these variants, given that they take an effect.

One possibility is that they evaluate the effect on each pull (racing with the underlying pull). This would be surprising for something like `stream.haltWhen(ZIO.sleep(10.seconds))` – you'd think this would halt the stream after 10 seconds, but really it would halt only when any given pull took more than 10 seconds.

The other possibility is that they fork the given effect just before the first pull, and race each pull against that fiber. I can't think of when this would be surprising, but I'm sure it also has cases where it's not what's expected.

Currently the `interruptWhen` behavior is the former, and the `haltWhen` behavior is effectively the latter (since it effectively forks the `IO` to a new `Promise` and then calls `haltWhen(Promise)`). I'll change them both to be the latter behavior explicitly, because it's more intuitive to me and it at least needs to be consistent.

This will make the IO variants look very similar to the Promise variants, except that they'll poll a fiber instead of a promise. I'm hesitant to change the existing `Promise` variants to delegate to the new `IO` variants, though, since these will now involve allocating a fiber while the `Promise` variants don't.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1741,6 +1761,21 @@ class ZStream[-R, +E, +A] private[stream] (private[stream] val structure: ZStrea
...
   }
 
+  /**
+   * Interrupts the evaluation of this stream when the provided IO completes. The given
+   * IO will be forked as part of this stream, and its success will be discarded. This
+   * combinator will also interrupt any in-progress element being pulled from upstream.
+   *
+   * If the IO completes with a failure, the stream will emit that failure.
+   */
+  final def interruptWhen[E1 >: E](io: IO[E1, Unit]): ZStream[R, E1, A] =
```
iravid, line 1771: This can be `[R1 <: R, E1 >: E](zio: ZIO[R1, E1, Any])`.

It is actually idiomatic in ZIO to use `Any` when the result is not required - that saves the callers an extra call to `map`, resulting in (slightly) more efficient and concise code :-)

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Good catch! The implementation in this revision looks good. Let’s just generalize it to `[R1 <: R, E1 >: E](zio: ZIO[R1, E1, Any])`.

iravid:
> Thinking about this more, there are two subtly different possible behaviors for these variants, given that they take an effect.
> 
> 
> 
> One possibility is that they evaluate the effect on each pull (racing with the underlying pull). This would be surprising for something like `stream.haltWhen(ZIO.sleep(10.seconds))` – you'd think this would halt the stream after 10 seconds, but really it would halt only when any given pull took more than 10 seconds.
> 
> 
> 
> The other possibility is that they fork the given effect just before the first pull, and race each pull against that fiber. I can't think of when this would be surprising, but I'm sure it also has cases where it's not what's expected.

This is definitely the desired behavior. Apologies if I’ve missed this in the first review.

> 
> 
> 
> Currently the `interruptWhen` behavior is the former, and the `haltWhen` behavior is effectively the latter (since it effectively forks the `IO` to a new `Promise` and then calls `haltWhen(Promise)`). I'll change them both to be the latter behavior explicitly, because it's more intuitive to me and it at least needs to be consistent.
> 
> 
> 
> This will make the IO variants look very similar to the Promise variants, except that they'll poll a fiber instead of a promise. I'm hesitant to change the existing `Promise` variants to delegate to the new `IO` variants, though, since these will now involve allocating a fiber while the `Promise` variants don't.

I’d be fine with forwarding to the IO variants using `Promise.await` but we can deal with that in a future follow-up. I’ll probably do that in the ZConduit branch.

Thanks for thinking this through! These are very useful combinators.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jeremyrsmith: Understood – done!

jeremyrsmith:
OK, made those changes and removed the broken scaladoc link (I could not figure out the incantation to get a link to the overloaded `interruptWhen` method. Oh well)

streams-tests/jvm/src/test/scala/zio/stream/StreamInterruptWhenSpec.scala
```diff
@@ -10,31 +10,61 @@ object StreamInterruptWhenSpec extends ZIOBaseSpec {
...
+            _      <- started.await *> halt.succeed(())
+            _      <- fiber.await
+            result <- interrupted.get
+          } yield assert(result)(isTrue)
+        },
+        testM("propagates errors") {
+          for {
+            halt <- Promise.make[String, Nothing]
+            _    <- halt.fail("Fail")
+            result <- ZStream.fromIterable(scala.Stream.from(0))
```
jeremyrsmith, line 62: @iravid just want to call this out – I had to change this from `ZStream(1)` because if the ZStream is already successful and is small, then there is no chance to observe the completion of `halt.await` (even though it's already done) and the resulting stream doesn't fail like it should. I'm not sure if that's an issue or not, in real life (what does it mean to halt/interrupt an already successful stream?)

jeremyrsmith:
Oh, FFS :angry: guess I can't use `scala.Stream` either.

jeremyrsmith:
@iravid finally got through the gauntlet of scalafmt and test flakiness. Please see my comment re: testing against strict (`ZStream.apply`-based) streams, I think the behavior there is a judgement call.

PR #3309, opened 2020-04-07
Fix ++(nonEmpty) on Chunk

ahoy-jon:
following discussion #3241

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -41,7 +41,7 @@ sealed trait Chunk[+A] { self =>
   /**
    * Returns the concatenation of this chunk with the specified chunk.
    */
-  def ++[A1 >: A](that: Chunk[A1]): Chunk[A1] = Chunk.concat(self, that)
+  def appendNonEmpty[A1 >: A](nonEmptyChunk: NonEmptyChunk[A1]): NonEmptyChunk[A1] = Chunk.concat(self, nonEmptyChunk)
 
   /**
    * Appends an element to the chunk
@@ -596,6 +596,11 @@ sealed trait Chunk[+A] { self =>
 }
 
 object Chunk {
+
+  implicit class ChunkOps[A](private val self: Chunk[A]) extends AnyVal {
+    def ++[A1 >: A](chunk: Chunk[A1]): Chunk[A1]                         = concat(self, chunk)
+    def ++[A1 >: A](nonEmptyChunk: NonEmptyChunk[A1]): NonEmptyChunk[A1] = concat(self, nonEmptyChunk)
+  }
 
   /**
    * Returns the empty chunk.
@@ -1205,6 +1210,13 @@ object Chunk {
      * Zips this chunk with the specified chunk using the specified combiner.
      */
     override def zipWith[B, C](that: Chunk[B])(f: (Nothing, B) => C): Chunk[C] = Empty
+
+    //override type Concat[A1] = Chunk[A1]
+
+    /**
+     * Returns the concatenation of this chunk with the specified chunk.
+     */
+    def ++[A1](that: Chunk[A1]): Chunk[A1] = that
   }
 
   sealed trait NonEmpty[+A] extends Chunk[A] { self =>
@@ -1226,7 +1238,7 @@ object Chunk {
     /**
      * Returns the concatenation of this chunk with the specified chunk.
      */
-    final override def ++[A1 >: A](that: Chunk[A1]): NonEmptyChunk[A1] = Chunk.concat(self, that)
+    final def ++[A1 >: A](that: Chunk[A1]): NonEmptyChunk[A1] = Chunk.concat(self, that)
 
     /**
      * Materializes a chunk into a chunk backed by an array. This method can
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -658,7 +658,7 @@ trait ZSink[-R, +E, +A0, -A, +B] extends Serializable { self =>
                 if (self.cont(s2)) UIO.succeedNow(Left(s2))
                 else self.extract(s2).map(Right(_))
               },
-            { case (b, leftover) => UIO.succeedNow(Right((b, leftover ++ Chunk.single(a)))) }
```
adamgfraser, line 661: You can apply `ev` here to resolve the type inference issues.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -41,7 +41,7 @@ sealed trait Chunk[+A] { self =>
   /**
    * Returns the concatenation of this chunk with the specified chunk.
    */
-  def ++[A1 >: A](that: Chunk[A1]): Chunk[A1] = Chunk.concat(self, that)
+  def appendNonEmpty[A1 >: A](nonEmptyChunk: NonEmptyChunk[A1]): NonEmptyChunk[A1] = Chunk.concat(self, nonEmptyChunk)
```
adamgfraser, line 44: I think we can delete now?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1205,6 +1210,13 @@ object Chunk {
      * Zips this chunk with the specified chunk using the specified combiner.
      */
     override def zipWith[B, C](that: Chunk[B])(f: (Nothing, B) => C): Chunk[C] = Empty
+
+    //override type Concat[A1] = Chunk[A1]
```
adamgfraser, line 1214: Delete.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1205,6 +1210,13 @@ object Chunk {
...
+
+    /**
+     * Returns the concatenation of this chunk with the specified chunk.
+     */
+    def ++[A1](that: Chunk[A1]): Chunk[A1] = that
```
adamgfraser, line 1219: I think there is some duplication between this and the implicit syntax class we can get rid of.

ahoy-jon:
@adamgfraser thanks for the review! changing it.

Following https://github.com/zio/zio/pull/3241#issuecomment-610138035

it would work if we don't define : 
```scala
trait NonEmpty[A] {
  def ++[A1 >: A](that: Chunk[A1]): NonEmptyChunk[A1] 
}
```

adamgfraser:
@ahoy-jon Yes. It looks like if we only define both `++` variants on the implicit syntax and then the `++` variant on the `NonEmpty` subtype that works. Not sure if we can simplify any more.

streams/shared/src/main/scala/zio/stream/ZSink.scala
ahoy-jon: Ok!

```scala
leftover ++ Chunk.single(ev(a))
```

adamgfraser: Yes. I think the constraints in some of these sink combinators could possibly be simplified further but that is a problem for another day.

adamgfraser:
I think we can get away with just defining the variant that takes a non-empty chunk in the implicit syntax class and leave everything else on the trait.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -597,6 +597,11 @@ sealed trait Chunk[+A] { self =>
 
 object Chunk {
 
+  implicit class ChunkOps[A](private val self: Chunk[A]) extends AnyVal {
```
adamgfraser, line 595: Don't need the `A1` type parameter here since `A` in `ChunkOps` is invariant.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -597,6 +592,19 @@ sealed trait Chunk[+A] { self =>
...
+
+    /**
+     * Returns the concatenation of this chunk with the specified chunk.
+     */
+    def ++[A1 >: A](chunk: Chunk[A1]): Chunk[A1] = concat(self, chunk)
```
adamgfraser, line 600: Can move this variant back to the trait.

ahoy-jon:
<_feelings_>
@adamgfraser 🤪this technique was the initial PR!

I was actually fine with letting it go, and add a ```appendNonEmpty`!
</_feelings_>

Thanks for the reviews.

About `ZSink`, I have modifications coming to improve the performance of `+`, once it is done, I will update `ZSink`.

core/shared/src/main/scala/zio/Chunk.scala
ahoy-jon: Not sure about that, it doesn't create compilation error in ZIO, but when with use subtyping, doesn't work.

```scala
trait A
trait B extends A
val x: Chunk[B] = Chunk.empty
//x: zio.Chunk[B] = Empty$()
x ++ (new A {})
<console>:28: error: overloaded method value ++ with alternatives:
  (nonEmptyChunk: zio.NonEmptyChunk[B])zio.NonEmptyChunk[B] <and>
  (chunk: zio.Chunk[B])zio.Chunk[B]
 cannot be applied to (A)
       x ++ (new A {})
```

Issue #3301, opened 2020-04-05
Improving performance on Chunk


ahoy-jon:
Chunk could be improved in different ways.

- `Concat` could be specialized and improved with:
```scala
case class ConcatArr[A](body: Array[Arr[A]], tail:Array[Arr[A]]) extends NonEmpty[A]
case class Concat[A](body: Array[NonEmpty[A]], tail:Array[NonEmpty[A]]) extends NonEmpty[A]
```
This change could help a lot with #3233 .
The tail of those structures would be used to hold the element just added to the end of the structure.

For example:
```scala
val c1 = Arr(Array(1,2,3))
val c2 = c1 + 4
//ConcatArr(body = Array(c),  tail = Array(4))
val c3 = c2 + 5
//ConcatArr(body = Array(c), tail = Array(4,5))
val c4 = c2 + 6 + 7 + 8 + 9 + 10 + 11
//ConcatArr(body = Array(c, Arr(Array(4, 5, 6, 7, 8, 9, 10, 11)), tail = Array())
```

- `Slice` could be removed and pushed to `Arr` and `VectorChunk`
```scala
case class Arr[A](array: Array[A], offset: Int, l: Int) extends NonEmpty[A]
case class VectorChunk[A](private val vector: Vector[A], offset: Int, l: Int) extends NonEmpty[A]
```

- `materialize` could be optimized, keep Concat and group substructures.

- most of the complex components should have private "smart constructors".
```scala
private def arr[A](array: Array[A]): Arr[A] =
    Arr(array, ClassTag(array.getClass.getComponentType))

private case class Arr[A](private val array: Array[A], ct: ClassTag[A]) extends NonEmpty[A] {
```

What do you think about those changes?

Vilkina:
I'll take that ticket

ahoy-jon:
Interesting resources : 
* https://github.com/scala/scala/pull/8534
* https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/PersistentVector.java

jdegoes:
1. Materialize should not be changed, it's purpose is to force the chunk to be strict (no laziness), so making it lazy which just defeat the point of its existence.
2. How would `Concat` specialization improve anything?
3. Removing `Slice` could be useful. "Inlining" it into `Concat` (concat would have to force it down into the left and right children, discarding, if necessary, left or right when they are entirely sliced out), `Vector`, and `Array` chunks would create more flattening and reduce nesting.

[...]

PR #3311, opened 2020-04-07
WIP, Balancing on ++

ahoy-jon:
Solving #3238 

Here is a proposition to balance concatenation on `Chunk`.

Todos: 
- [ ] Performance check
- [ ] use a stack to avoid recursive calls.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -754,15 +754,19 @@ object Chunk {
   /**
    * Returns the concatenation of this chunk with the specified chunk.
    */
-  final def concat[A](l: NonEmpty[A], r: NonEmpty[A]): NonEmptyChunk[A] = Concat(l, r)
-
-  /**
-   * Returns the concatenation of this chunk with the specified chunk.
-   */
   final def concat[A](l: NonEmpty[A], r: Chunk[A]): NonEmptyChunk[A] = r match {
     case Empty           => l
     case ne: NonEmpty[A] => concat(l, ne)
   }
+
+  /**
+   * Returns the concatenation of this chunk with the specified chunk.
+   */
+  final def concat[A](l: NonEmpty[A], r: NonEmpty[A]): NonEmptyChunk[A] =
+    l match {
+      case Concat(ll, lr) if lr.size < ll.size => Concat(ll, concat(lr, r))
+      case _                                   => Concat(l, r)
+    }
 
   /**
    * Returns the `ClassTag` for the element type of the chunk.
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -764,8 +764,8 @@ object Chunk {
    */
   final def concat[A](l: NonEmpty[A], r: NonEmpty[A]): NonEmptyChunk[A] =
     l match {
-      case Concat(ll, lr) if lr.size < ll.size => Concat(ll, concat(lr, r))
-      case _                                   => Concat(l, r)
+      case Concat(ll, lr) if ll.size >= lr.size + r.size => Concat(ll, concat(lr, r))
+      case _                                             => Concat(l, r)
     }
 
   /**


adamgfraser:
I think we may need to do more to support specific use cases such as repeated concatenation of single elements and explicit rebalancing. [this](https://citeseer.ist.psu.edu/viewdoc/download?doi=10.1.1.14.9450&rep=rep1&type=pdf) article has some good discussion of these issues. There are a bunch of other good resources out there about implementing a rope data structure assuming that is what we want to do.

ahoy-jon:
Indeed, I will read the article. 

I was thinking of pushing the design further to support better ropes:
https://github.com/zio/zio/issues/3301

Currently, with this new implementation, the concatenation cost might be around `1 + (log(N) / 2)` instead of 1. Not great for repeated concatenations.

The goal of this PR is just to update `concat` without changing the underlying datastructure.

The goal is to have good performance on :
```scala
@Setup
val chunks = (0 until size).map(Chunk.single)

chunks.reduce(_ ++ _)
```

jdegoes:
@ahoy-jon Definitely would love a balanced version. Will close this for now until you have a chance to work on it some more, but feel free to re-open anytime, and thanks for your work on this!

ahoy-jon:
@jdegoes The current implementation is working
https://gist.github.com/ahoy-jon/4be89db816a38e2bba02f65c2b1f3a7e

However, I need your inputs on #3301 before pushing it further.

jdegoes:
@ahoy-jon 

The current implementation is quite performant for `Concat` only.

In order to see the need for balancing, it is necessary to benchmark repeated random access (or chunk size computation) on a chunk formed through a great number of `Concat` nodes (e.g. 1000 `Chunk` nodes).

A balanced tree will minimize the vertical height of the tree, to maximize the speed of getting to the right element (or computing the size). A lop-sided tree will necessitate lots of indirect lookups, causing performance to be quite poor compared to a balanced tree.

Now as an alternative to balancing, we could consider some internal materialization: perhaps we compute `materialize` inside `Chunk` on any operation that would be slow, if the tree is sufficiently deep; then switch over to calling methods on the materialized chunk. This would, however, be slower for mixed workloads (read/write interleaved repeatedly); a faster approach for mixed workloads would be balancing the tree.

ahoy-jon:
I should have said the current proposed algorithm, it is doing anything needed.

Currently, on my machine, it's a local recursive version, with a rewrite on N<8 : 
```scala
  final def +[A1 >: A](a: A1): NonEmptyChunk[A1] =
    self match {
      case Chunk.Empty => Chunk.single(a)
      case ne: NonEmptyChunk[A] =>
        val lefts: Array[NonEmptyChunk[A]] = Array.ofDim[NonEmptyChunk[A]](16)
        var n: Int                         = 0

        var point: NonEmptyChunk[A] = ne

        var continue = true
        var rewrite  = false

        while (continue) {
          point match {
            case _ if point.length + 1 <= 8 =>
              continue = false
              rewrite = true
            case Chunk.Concat(ll, lr) if n < 16 && ll.size >= lr.size + 1 =>
              lefts(n) = ll
              n += 1
              point = lr
            case _ =>
              continue = false
          }
        }

        val node: NonEmptyChunk[A1] = if (rewrite) {
          val len              = point.length + 1
          val ct: ClassTag[A1] = Chunk.classTagOf(point)
          val dest             = Array.ofDim[A1](len)(ct)
          point.toArray(0, dest)
          dest(len - 1) = a
          new Chunk.Arr(dest, ct)
        } else {
          Chunk.Concat(point, Chunk.single(a))
        }

        //foldLeft
        var res = node
        var i   = n - 1
        while (i > 0) {
          res = Chunk.Concat(lefts(i), res)
          i -= 1
        }
        res
    }
```

The benchmark is on up to 32k concats, and map access (which is not optimized on chunk yet).

The proposition is lacking a performance optimization on the right to simulate a tail. ( #3301 )

jdegoes:
@ahoy-jon 

For `add` performance, the relevant comparison is `Vector`, which has the fastest `add` of Scala collection types.

jdegoes:
@ahoy-jon 

I'd recommend at least trying a "hybrid" approach like this:

```scala
import java.util.concurrent.atomic.AtomicInteger
import scala.reflect.ClassTag

sealed trait Chunk[+A] { self =>
  def + [A1 >: A](a1: A1): Chunk[A1] = {
    val buffer = Array.ofDim[AnyRef](32)
    
    new Chunk.Add[A1](self, Array.ofDim(32), 1, new AtomicInteger(0))
  }
}
object Chunk {
  final class Arr[A](array: Array[A]) extends Chunk[A]
  final class Add[A](start: Chunk[A], buffer: Array[AnyRef], bufferUsed: Int, chain: AtomicInteger) extends Chunk[A] { self =>
    override def + [A1 >: A](a1: A1): Chunk[A1] = {
      if (bufferUsed == buffer.length) {
        // TODO: Buffer full, time to push over and begin new buffer.
        ???
      } else if (chain.compareAndSet(bufferUsed, bufferUsed + 1)) {
        buffer(bufferUsed)
        new Chunk.Add[A1](self, buffer, bufferUsed + 1, chain)  
      } else {
        // TODO: Cannot reuse the buffer, because someone else used it.
        //       Make new buffer.
        ???
      }
      
    }
  }
  
  def empty: Chunk[Nothing] = new Arr[Nothing](Array.empty[Nothing])
}
```

The idea being that you can safely push things onto a mutable array as long as you're the only one doing it.

ahoy-jon:
I reduced the scope of the PR and moved the work to #3358.

ahoy-jon:
@jdegoes thanks a lot for the advice on the hybrid approach!
I was looking at 
https://github.com/clojure/clojure/blob/30a36cbe0ef936e57ddba238b7fa6d58ee1cbdce/src/jvm/clojure/lang/PersistentVector.java#L25

I thought it was too crazy for ZIO! 😃

PR #3318, opened 2020-04-08
Implement BitChunk

### LINKED ISSUE #3234: Create BitChunk
A BitChunk is a chunk of bits, whose length method returns the number of bits in the chunk, and whose apply / get / etc. methods all operate on the bits.

A `Chunk[X]` should have a `toBitChunk` method, for all `X` such that `X` can be easily viewed as a sequence of bits—most notably, `Chunk[Byte]`.

The methods on `BitChunk` should be similar to the methods on `Chunk`, but of course, specialized to bits. Any attempt to "map" a `BitChunk` should return an ordinary chunk—if we even have map-like functions.

adamgfraser:
Initial work to resolve #3234.


core/shared/src/main/scala/zio/BitChunk.scala
```diff
@@ -0,0 +1,59 @@
...
+    builder.toString
+  }
+
+  def toBinaryString: String =
+    mkString("", "", "")
+
+  override def toString: String =
+    mkString
+
+  private def apply(n: Int): Boolean =
```
jdegoes, line 51: Probably should expose this.

jdegoes:
@adamgfraser Looking good! Did you want to merge this in now or wait till it has monomorphic versions of Chunk methods?

adamgfraser:
@jdegoes I will add more methods. Just wanted to make sure I was on the right track.

jdegoes:
@adamgfraser Yes, looks good to me!

jdegoes:
@adamgfraser 

After working on ZIO Codec, I think we need to make this:

```scala
object Chunk {
  ...
  final class Bit(...) extends Chunk[Boolean] {
    /* all bit chunk methods here */
}
```

Then potentially we can specialize on construction of `Chunk[Boolean]` and of course have manual constructors from `Chunk[Byte]`.

adamgfraser:
@jdegoes On it.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -835,6 +835,7 @@ object Chunk {
     case x: Singleton[A]   => x.classTag
     case x: Slice[A]       => x.classTag
     case x: VectorChunk[A] => x.classTag
+    case _: BitChunk       => ClassTag.Boolean.asInstanceOf[ClassTag[A]]
```
jdegoes, line 838: This is technically correct, I think. Although in any case where we look at this class tag and are tempted to use it (to create a new array), we should probably be special casing and instead creating a bit chunk... so hopefully we never use the class tag.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1192,11 +1195,69 @@ object Chunk {
...
   }
 
+  final class BitChunk private (private val bytes: Chunk[Byte]) extends NonEmpty[Boolean] { self =>
+
+    def apply(n: Int): Boolean =
+      (bytes(n >> 3) & (1 << (7 - (n & 7)))) != 0
+
+    def foreach(f: Boolean => Any): Unit = {
+      var i = 0
+      while (i < length) {
```
jdegoes, line 1210: Eventually this can be optimized so you iterate over the bytes in `bytes` and call `f` 8 times for each byte.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1192,11 +1195,69 @@ object Chunk {
 
     override def apply(n: Int): A = vector(n)
 
-    override def foreach(f: A => Unit): Unit = vector.foreach(f)
+    override def foreach(f: A => Any): Unit = vector.foreach(f)
 
     override def toArray[A1 >: A](n: Int, dest: Array[A1]): Unit = { val _ = vector.copyToArray(dest, n, length) }
   }
 
+  final class BitChunk private (private val bytes: Chunk[Byte]) extends NonEmpty[Boolean] { self =>
```
jdegoes, line 1203: For all subsetting operations, we aren't going to want to create a true subset. So it could make sense to think about introducing a `minBitIndex` and `maxBitIndex` to represent a range within `bytes` that are valid.

Or we could just implement methods and then go back later and optimize this case.

Imagine, e.g., taking 14 bits from a bit chunk: you want to get back something that's still a bit chunk, such that subsequent operations on the bit chunk will be efficient. If instead you got back `Chunk.Range`, then you'd end up materializing the bit chunk into `Array[Boolean]` before long. Of course we can try to fix that elsewhere, but then all the "lazy" operations need to override everything and special case bit chunk... it may be easier instead to merely specialize all operations on bit chunk to return another bit chunk (if possible).

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1192,11 +1195,69 @@ object Chunk {
...
+    def toBinaryString: String =
+      mkString("", "", "")
+
+    override def toString: String =
+      mkString
+  }
+
+  object BitChunk {
+
+    def fromByteChunk(bytes: Chunk[Byte]): BitChunk =
```
jdegoes, line 1257: We can add this to `Chunk`, as `Chunk#asBits`, when there exists implicit evidence that `A <:< Byte`.

adamgfraser: One thing we're running a bit into here is the new `NonEmptyChunk` concept. Right now `BitChunk` is extending `NonEmpty` which I think technically means we shouldn't be able to convert a `Chunk[Byte]` to a `BitChunk` unless we know statically that it is nonempty (or have to return an `Option` to express the fact that it might not be). But it seems like that is the last thing we want to do in performance sensitive code.

jdegoes: I think it's ok to make that return `Option`, because the `Some` allocation overhead is amortized over the length of the bit chunk, which is hopefully quite large.

adamgfraser: Okay great. We may also have to think about the contract around emptiness with operations. If we want to allow creating subsets and still return a `BitChunk` then I I think we would have to accept the possibility that there could be an empty `BitChunk`.

jdegoes: Hmmm. Would it be better for `BitChunk` to _not_ extend `NonEmpty`?

adamgfraser: I think it would be better for `BitChunk`. But then you kind of have an implementation leak where any other code within `Chunk` that wants to deal with empty versus nonempty chunks now has three cases to deal with. A chunk isn't either empty or nonempty. It is empty, nonempty, or this other bit chunk thing that could be one or the other.

jdegoes: Ok, how about this: `BitChunk` extends `NonEmpty`, but in any case where we don't know if it's empty, we return only `Chunk`, and force the use to pattern match (or test) if they care about non-emptiness.

In this world, we would guard the constructor to `BitChunk` to make sure that it can only be invoked if there is at least one bit in it.

adamgfraser: I think that generally works. It may depend on how often users are going to want to do operations that reduce the number of bytes. If you want to do a bunch of those and lose type information every time so you can't chain operations it could get annoying.

jdegoes: That's true, although even if you don't know it's a bit chunk, the bit chunk operations will still be optimized, so e.g. `bitChunk.take(5).take(3)` will still give you a BitChunk, assuming there were enough bits. So you could check at the end of a long series of operations, if you needed access to bit chunk specific methods.

Furthermore, it's possible the "bit chunk specific methods" could be exposed on chunk with evidence that `A <:< Boolean`, in which case they're not bit chunk specific anymore (just optimized quite nicely on bit chunks!), but rather boolean specific, so as you manipulate the size, you still get back something on which you can apply the bit chunk specific methods.

adamgfraser: It would be nice if we could do that because there is also a little weirdness here that right now you aren't allowed to know anything about the internal implementation of chunk, just whether or not it is empty, and now we're exposing this one really specialized case. So if we could pull it back to methods assuming a chunk is of a certain type that are then optimized under the hood would be very nice.

jdegoes: Sounds good to me, let's try to make it work!

adamgfraser: Minimal implementation in the current version. Seems to work well and definitely nicer for the Chunk API.  The main downside I see is that it is easy to accidentally do operations like mapping over a chunk that turn it back into a chunk backed by an array.

jdegoes: (1) If we override all methods in BitChunk, we can make sure that once you have a bit chunk, you stay with a bit chunk.

(2) Now making sure that if you start with something else, and end up with Chunk[Boolean], that it is backed by BitChunk, that would be harder, because it would mean that in any case where we create an `Array[Boolean]`, we need to actually change the logic and construct an `Array[Byte]` and set the bits inside there appropriately (maybe with a facade, a `BitChunkBuilder` that wraps an `Array[Byte]` and lets you change bits before building).

We don't have to do any of this work in the PR of course. We can split out tickets for (1) and (2) separately.

PR #3321, opened 2020-04-08
ZIO Test: Implement Generators for Chunks

adamgfraser:
Implements generators for chunks analogous to existing generators for other collection types. Specializes generators for non-empty lists and chunks to return more specific type. Also implements `Chunk.fill`, `Assertion.hasSizeChunk`, and `Assertion.isEmptyChunk`.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -665,6 +665,21 @@ object Chunk {
         arr(des)
     }
 
+  def fill[A](n: Int)(elem: => A): Chunk[A] =
+    if (n <= 0) Chunk.empty
+    else {
+      val first                     = elem
+      implicit val tag: ClassTag[A] = Tags.fromValue(first)
+      val array                     = Array.ofDim[A](n)
+      array(0) = first
+      var i = 1
+      while (i < n) {
+        array(i) = elem
+        i += 1
+      }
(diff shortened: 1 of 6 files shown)


iravid:
Perhaps as generalized `small/medium/large` combinators

adamgfraser:
We already have generalized `small`, `medium`, and `large` combinators in `Gen` so you can do `small(chunkOf(anyInt))`.

I think the other consideration that is relevant is that these generators are not trying to generate different internal representations of chunks. We just take a bunch of values and stick them into a chunk, versus trying to construct chunks from different combinations of subtypes.

I think this is the right behavior for users of Chunk, since they should get the same behavior regardless of whether a chunk is represented as a single array or a concatenation of two arrays, for example. And this is consistent with the treatment of other collection generators like `Vector`. But for the implementer of chunk it is valuable to have generators of different combinations of subtypes to verify that is actually true.

So what I would suggest is we could do the following:
1. Move `ChunkSpec` along with the existing generators in `ChunkUtils` to `core-tests`. This file and these generators would be responsible for verifying the correctness of `Chunk` constructors and combinators.
2. Move all other generators in stream tests to use new implementation of chunk generators implemented here. This should be a very simple change basically just changing the implementation of `smallChunks`, etc...  to be as described above.

iravid:
Sounds good and agreed on the action items!

Issue #3268, opened 2020-04-03
Improve ZIO Mock documentation


ioleo:
I have a feeling that the mocking documentation could be improved. I could use help from someone who will approach it with a fresh look to proofread whats already available and propose improvements either in form of PR or at least some comments that might help me identify parts that are unclear.

ioleo:
Also document support for `by-name` parameters #3314

pierangeloc:
I'm having a look and will come with some suggestions

hearnadam:
zio-mock is no longer.

PR #3328, opened 2020-04-10
Comment expectation creation code in mocking docs

vpavkin:
I have a small addition in the scope of #3268.
Overall it feels like a very solid doc to my eye. I had one bump though while reading - "Setting up expectations" part starts with a piece of code I couldn't understand at all. Of course it's explained in detail afterwards, but it forced me to go back and revisit the snippet again.

I propose a more self-explaining sample so that following text is based on some initial understanding already.


ioleo:
Thank you :pray: 

@adamgfraser

PR #3332, opened 2020-04-10
Unify ChunkSpec

### LINKED ISSUE #3324: Unify ChunkSpec
Currently tests for Chunk are split up between two different `ChukSpec.scala` files in the `coreTests` and `streamTests` projects. We should unify all chunk tests in the file in `coreTests` and also move the chunk generators in `ChunkUtils.scala`, which explicitly generate all the different subtypes of chunks, to that file.

Then we should update the chunk generators used in all the stream tests to use the new chunk generators implemented in ZIO Test.

vasilmkd:
Resolve #3324.


streams-tests/jvm/src/test/scala/zio/stream/StreamChunkSpec.scala
```diff
@@ -14,6 +14,18 @@ object StreamChunkSpec extends ZIOBaseSpec {
 
   import ZIOTag._
 
+  def tinyChunks[R <: Random, A](a: Gen[R, A]): Gen[R with Sized, Chunk[A]] =
+    Gen.chunkOfBounded(0, 3)(a)
+
+  def smallChunks[R <: Random, A](a: Gen[R, A]): Gen[R with Sized, Chunk[A]] =
+    Gen.small(Gen.chunkOfBounded(0, _)(a))
```
adamgfraser, line 21: If you're just bounding it at 0 you can just do `Gen.small(Gen.chunkOfN(_)(a))`.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -1,40 +1,398 @@
...
+  val intGen = Gen.int(-10, 10)
+
+  def toBoolFn[R <: Random, A]: Gen[R, A => Boolean] =
+    Gen.function(Gen.boolean)
+
+  def tinyChunks[R <: Random, A](a: Gen[R, A]): Gen[R, Chunk[A]] =
+    Gen.chunkOfBounded(0, 3)(a)
+
+  def smallChunks[R <: Random, A](a: Gen[R, A]): Gen[R with Sized, Chunk[A]] =
+    Gen.small(Gen.chunkOfBounded(0, _)(a))
```
adamgfraser, line 23: `Gen.small` generates a random value from a small size distribution. `Gen.chunkOfBounded` generates values between the minimum and the maximum. So by combining them in this way I think you are biasing the size downward a little bit. I you just do `Gen.small(Gen.chunkOfN(_)(a))` you will get the "small" distribution of sizes.

vasilmkd:
@adamgfraser I got the impression that `chunkOfN` creates chunks of size exactly `N`. I'm more than happy to improve it. Can you please confirm that `chunkOfN` is what we want? Sorry for the inconvenience. Thanks.

adamgfraser:
@vasilmkd You're correct that `Gen.chunkOfN` generates chunks of exactly the specified size. But `Gen.small` returns a distribution of size values itself. So if you take each sample from the distribution of size values from `small` and feed it into `chunkOfN` you will get chunks with a distribution of sizes.

vasilmkd:
Awesome, thanks for the explanation. Change incoming.

Issue #2720, opened 2020-01-25
Port ZStream combinators and constructors to the new encoding


iravid:
Submit PRs to `feature/zconduit`. Reply below to claim an item.

Constructors:
- [x] absolve 
- [x] access 
- [x] accessM 
- [x] accessStream 
- [x] apply(A*)
- [x] bracket 
- [x] bracketExit 
- [x] crossN 
- [x] die 
- [x] dieMessage 
- [x] effectAsync 
- [x] effectAsyncInterrupt 
- [x] effectAsyncM 
- [x] effectAsyncMaybe 
- [x] environment 
- [x] fail 
- [x] finalizer 
- [x] flatten 
- [x] flattenPar 
- [x] flattenParUnbounded 
- [x] fromChunk 
- [x] fromEffect 
- [x] fromEffectOption 
- [x] fromInputStream 
- [x] fromIterable 
- [x] fromIterator 
- [x] fromIteratorManaged 
- [x] fromJavaIterator 
- [x] fromJavaIteratorManaged 
- [x] fromQueue 
- [x] fromQueueWithShutdown 
- [x] fromTQueue 
- [x] halt 
- [x] iterate 
- [x] managed 
- [x] mergeAll 
- [x] mergeAllUnbounded 
- [x] paginate 
- [x] paginateM 
- [x] range 
- [x] repeatEffect 
- [x] repeatEffectOption 
- [x] repeatEffectWith 
- [x] succeed 
- [x] unfold 
- [x] unfoldM 
- [x] unwrap 
- [x] unwrapManaged 
- [x] zipN 
- [x] empty 
- [x] never 
- [x] unit 

Combinators:
- [x] &> 
- [x] *> 
- [x] ++ 
- [x] <& 
- [x] <&> 
- [x] <* 
- [x] <*> 
- [x] absolve 
- [x] aggregate 
- [x] aggregateAsync
- [x] aggregateAsyncWithin 
- [x] aggregateAsyncWithinEither
- [ ] ~aggregateManaged~
- [x] as 
- [x] bimap 
- [x] broadcast 
- [x] broadcastDynamic 
- [x] broadcastedQueues
- [x] broadcastedQueuesDynamic
- [x] buffer 
- [x] bufferDropping 
- [x] bufferSignal 
- [x] bufferSliding 
- [x] bufferUnbounded 
- [x] catchAll 
- [x] catchAllCause
- [x] chunkN - port as transducer
- [x] collect 
- [x] collectM 
[...]

Issue #2719, opened 2020-01-25
Port ZSink constructors and combinators to the new encoding


iravid:
See current design in https://github.com/zio/zio/issues/2551#issuecomment-578415241.

Submit PRs to `feature/zconduit`. Reply below to claim an item.

Constructors:    
- [ ] ~await~ ported as `ZSink.head`
- [x] collectAll 
- [x] collectAllN - port as transducer
- [x] collectAllToMap 
- [x] collectAllToSet 
- [x] collectAllToMapN - port as transducer
- [x] collectAllToSetN - port as transducer
- [x] collectAllWhile - port as transducer
- [x] collectAllWhileM - port as transducer
- [x] die 
- [x] dieMessage 
- [x] drain 
- [x] fail
- [x] fold
- [x] foldLeft 
- [x] foldLeftM 
- [x] foldM
- [x] foldUntil - port as transducer
- [x] foldUntilM - port as transducer
- [x] foldWeighted - port as transducer
- [x] foldWeightedDecompose - port as transducer
- [x] foldWeightedDecomposeM - port as transducer
- [x] foldWeightedM - port as transducer 
- [x] fromEffect
- [x] fromFunction - will need a different signature (`A => Either[Z, Unit]`)
- [x] fromFunctionM - will need a different signature (`A => ZIO[R, Either[E, Z], Unit]`)
- [x] halt
- [x] head 
- [ ] ~identity~ - ported as `head`
- [x] ignoreWhile - port as transducer
- [x] ignoreWhileM - port as transducer
- [x] last 
- [ ] ~pull1~ requires leftovers to operate properly, don't port
- [ ] ~read1~ requires leftovers to operate properly, don't port
- [ ] splitOn - port as transducer
- [x] succeed
- [x] splitLines - port as transducer
- [ ] ~splitLinesChunk~ chunks are built-in so this will be ported as `splitLines`
- [x] utf8DecodeChunk - port as transducer

Combinators:
- [x] &> 
- [x] *> 
- [x] <& 
- [x] <&> 
- [x] <* 
- [x] <*> 
- [ ] <| 
- [x] as 
- [ ] asError 
- [x] contramap
- [x] contramapM
- [x] dimap
- [x] flatMap
- [x] map
- [x] mapError
- [x] mapM
- [ ] orElse
- [ ] provideSome
- [x] race 
- [x] raceBoth
- [ ] tapInput 
- [ ] tapOutput 
- [ ] timed
- [ ] unit 
- [x] zip 
[...]

Issue #2551, opened 2020-01-01
Upgrading the design of ZSink


vasilmkd:
### Background
`ZSink`'s state machine is defined in terms of its four abstract methods: `initial`, `step`, `extract` and `cont`. In particular, there is an implicit contract in place (the way all `ZSink` constructors and combinators are implemented) that:

Starting from the `initial` state, a `ZSink` is to be `step`ped over and over until the `cont` method returns `false`, at which point an `extraction` is generally safe (assuming no external effect errors) to be performed, after which the state of the `ZSink` can be safely discarded and the process should be started from the beginning.

This is the contract that _most_ of the `ZStream` combinators that interface with `ZSink` expect and respect.

### Always-continuing `ZSink`s
Until now, the binary nature of `ZSink#cont` has been generally accepted and worked around in the implementation of the `ZSink` and `ZStream` combinators. However, there have been instances in the past where this has proven problematic. I'll illustrate with an example.

There are certain `ZSink`s like `ZSink.collectAll[A]` which are "always-continuing" (`cont = true`), in the sense that they will keep on accumulating input elements in a `List[A]` and only return it upon extraction. As you can probably guess, the only way to use this sink is to disregard the value of `cont` and extract it when it is desired or convenient.

### `ZStream#aggregateAsync`
Then there are the asynchronous aggregation combinators which have a wonderfully complex implementation, because of their desire to provide one particular feature: balance between aggregation throughput and aggregation latency. To achieve this, the state of the sink is shared between two fibers, with the "downstream" fiber always attempting an extraction whenever it gets a chance (again, completely disregarding the value of `cont`). Of course, this works for sinks that are happy to always continue with the aggregation and can be extracted at will, but all in all, it is a violation of the `ZSink` contract and should not work in the general case, especially when building effectful sinks.

### Proposal
Havin `cont` return an ADT that denotes the different modes that the sink can be in:
1. Requires more elements before extraction ("hard" `cont = true`).
2. Can accept new elements but is otherwise ready for extraction ("soft" `cont = true`).
3. Doesn't accept any new elements and must be extracted ("hard" `cont = false`).

That way any combinator that gets a value as outlined in `2.` could either continue to feed the sink, or extract it, depending on the desired behavior.

As always, your feedback is greatly appreciated.

iravid:
Great analysis. We have been putting off far too many questions about the contract of ZSink. What you've listed is the current established usage but we've not documented or verified any of the contract elements. I constantly asked myself when writing combinators:
[...]

PR #3341, opened 2020-04-11
Change transducer encoding to force them to be infinite

iravid:


mschuwalow:
Looks good!

Sadly I was out of the loop for a while. Is there anywhere where I can get up to speed on the current design direction of ZStream?

iravid:
@mschuwalow Hey! I will do a summary soon of the current design. You can also look at the ZTransducer/ZSink classes; they have some notes on them. ZStream is mostly unchanged except it now operates on chunks by default.

regiskuckaertz:
@mschuwalow Great to have you back! The tickets are also a great source of history: https://github.com/zio/zio/issues/2551, https://github.com/zio/zio/issues/2720, https://github.com/zio/zio/issues/2719

mschuwalow:
Thanks, I'll take a look!

PR #3342, opened 2020-04-12
Make Chunk Extend IndexedSeq

adamgfraser:
Makes Chunk extend `IndexedSeq` from the Scala standard library. This will improve interoperability with other code and pave the way for us to unify the `Iterable` and `Chunk` versions of a variety of combinators.

Because of changes to Scala's collection library in 2.13, this is implemented in terms of a new version specific trait `ChunkLike` that `Chunk` extends. `ChunkLike` extends the appropriate traits for each versions and implements the corresponding builder interface so that code in `Chunk` can continue to be written on a cross-version basis.

Potential issues:

- I think we are going to have to give up on `empty ++ nonempty`  returning a nonempty chunk because we now have another `++` variant on `IndexedSeq`. We could address by picking another operator for combining two chunks that doesn't clash to get back the current behavior.
- A few methods had to be move to `ChunkLike` due to version specific differences. In some cases Scala 2.13 provides a signature that matches ours because it doesn't include `CanBuildFrom` but Scala 2.12 doesn't. So we need to `override` in one case but not the other. In other cases methods are defined as `final` in `IndexedSeq` in Scala 2.13 (e.g. `size` is defined in terms of `length`) so we can't override them.
- `ChunkBuiler` implementation could be optimized further. It is already backed by an `ArrayBuilder` that has specialized implementations for primitives but we could override the `++=` method for increased efficiency.
- `Chunk#fold` has to be renamed `Chunk#foldLeft`. I think this is fine and it probably should have been this way all along but it is a breaking change.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -192,7 +192,7 @@ object ChunkSpec extends ZIOBaseSpec {
       trait Dog extends Animal
 
       val vector   = Vector(new Cat {}, new Dog {}, new Animal {})
-      val actual   = Chunk.fromIterable(vector).map(a => a)
+      val actual   = Chunk.fromIterable(vector).map[Animal](identity)
       val expected = Chunk.fromArray(vector.toArray)
 
       assert(actual)(equalTo(expected))
@@ -201,7 +201,7 @@ object ChunkSpec extends ZIOBaseSpec {
       assert(Chunk.empty.toArray[String])(equalTo(Array.empty[String]))
     },
     zio.test.test("to Array for an empty Chunk using filter") {
-      assert(Chunk(1).filter(_ == 2).map(_.toString).toArray[String])(equalTo(Array.empty[String]))
+      assert(Chunk(1).filter(_ == 2).map[String](_.toString).toArray[String])(equalTo(Array.empty[String]))
     },
     testM("toArray with elements of type String") {
       check(mediumChunks(Gen.alphaNumericString))(c => assert(c.toArray.toList)(equalTo(c.toList)))
@@ -313,7 +313,7 @@ object ChunkSpec extends ZIOBaseSpec {
       // foreach should not throw
       c.foreach(_ => ())
 
-      assert(c.filter(_ => false).map(_ * 2).length)(equalTo(0))
+      assert(c.filter(_ => false).map[Int](_ * 2).length)(equalTo(0))
     },
     zio.test.test("toArrayOnConcatOfSlice") {
       val onlyOdd: Int => Boolean = _ % 2 != 0
@@ -367,8 +367,8 @@ object ChunkSpec extends ZIOBaseSpec {
         Chunk.succeed(x),
         nonEmptyChunk ++ chunk,
         // chunk ++ nonEmptyChunk,
-        nonEmptyChunk.flatMap(i => Chunk(i)),
-        nonEmptyChunk.map(identity(_)),
+        nonEmptyChunk.flatMap((i: Int) => Chunk(i)),
+        nonEmptyChunk.map[Int](identity),
         nonEmptyChunk.zipAllWith(Chunk(0))(l => (l, l), r => (r, r))((l, r) => (l, r)),
         nonEmptyChunk.zipWithIndex,
         nonEmptyChunk.zipWithIndexFrom(0)
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -60,7 +60,7 @@ object ChunkSpec extends ZIOBaseSpec {
     testM("apply") {
       check(chunkWithIndex(Gen.unit)) {
         case (chunk, i) =>
-          assert(chunk.apply(i))(equalTo(chunk.toSeq.apply(i)))
+          assert(chunk.apply(i))(equalTo(chunk.toList.apply(i)))
       }
     },
     testM("corresponds") {
@@ -68,7 +68,7 @@ object ChunkSpec extends ZIOBaseSpec {
       val genFunction = Gen.function[Random with Sized, (Int, Int), Boolean](Gen.boolean).map(Function.untupled(_))
       check(genChunk, genChunk, genFunction) { (as, bs, f) =>
         val actual   = as.corresponds(bs)(f)
-        val expected = as.toSeq.corresponds(bs.toSeq)(f)
+        val expected = as.toList.corresponds(bs.toList)(f)
         assert(actual)(equalTo(expected))
       }
     },
@@ -81,18 +81,18 @@ object ChunkSpec extends ZIOBaseSpec {
       }
     },
     testM("length") {
-      check(largeChunks(intGen))(chunk => assert(chunk.length)(equalTo(chunk.toSeq.length)))
+      check(largeChunks(intGen))(chunk => assert(chunk.length)(equalTo(chunk.toList.length)))
     },
     testM("equality") {
       check(mediumChunks(intGen), mediumChunks(intGen)) { (c1, c2) =>
-        assert(c1.equals(c2))(equalTo(c1.toSeq.equals(c2.toSeq)))
+        assert(c1.equals(c2))(equalTo(c1.toList.equals(c2.toList)))
       }
     },
     zio.test.test("inequality") {
       assert(Chunk(1, 2, 3, 4, 5))(Assertion.not(equalTo(Chunk(1, 2, 3, 4, 5, 6))))
     },
     testM("materialize") {
-      check(mediumChunks(intGen))(c => assert(c.materialize.toSeq)(equalTo(c.toSeq)))
+      check(mediumChunks(intGen))(c => assert(c.materialize.toList)(equalTo(c.toList)))
     },
     testM("foldLeft") {
       check(Gen.alphaNumericString, Gen.function2(Gen.alphaNumericString), smallChunks(Gen.alphaNumericString)) {
@@ -112,7 +112,7 @@ object ChunkSpec extends ZIOBaseSpec {
     ),
     testM("map") {
       val fn = Gen.function[Random with Sized, Int, Int](intGen)
-      check(smallChunks(intGen), fn)((c, f) => assert(c.map(f).toSeq)(equalTo(c.toSeq.map(f))))
+      check(smallChunks(intGen), fn)((c, f) => assert(c.map(f).toList)(equalTo(c.toList.map(f))))
     },
     suite("mapM")(
       testM("mapM happy path")(checkM(mediumChunks(intGen), Gen.function(Gen.boolean)) { (chunk, f) =>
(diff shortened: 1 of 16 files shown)


core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -313,7 +313,7 @@ object ChunkSpec extends ZIOBaseSpec {
       // foreach should not throw
       c.foreach(_ => ())
 
-      assert(c.filter(_ => false).map(_ * 2).length)(equalTo(0))
+      assert(c.filter(_ => false).map[Int](_ * 2).length)(equalTo(0))
```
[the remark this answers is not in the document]
adamgfraser: The issue is that we now have two overloaded versions of `map` on Scala 2.11 and 2.12:

```scala
// From Chunk
def map[B](f: A => B): Chunk[B]

// from IndexedSeq
def map[B, That](f: A => B)(implicit bf: CanBuildFrom[List[A], B, That]): That
```

So in the presence of that the Scala compiler, especially on 2.11 has a harder time inferring these types and complains that the type of an anonymous function should be fully known (even though it seems like it should be inferable here).

I think to address we need to consider moving methods like this to `ChunkLike` and only having one implementation on Scala 2.11 / Scala 2.12 that uses the signature from `IndexedSeq` but has an efficient implementation.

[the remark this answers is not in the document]
iravid: Ah, I see. Will we able to retain the efficient implementation given that signature? We'd have to go through the CBF and won't be able to use an array directly.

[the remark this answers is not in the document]
adamgfraser: That's the question. I think we can. We can match on the CBF to get back a `ChunkBuilder` and add methods on the `ChunkBuilder` to support an efficient implementation in that case.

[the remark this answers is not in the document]
iravid: Ah yeah. That sounds viable!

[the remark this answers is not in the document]
jdegoes: Worth writing up in tickets so we don't forget!

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -60,15 +60,15 @@ object ChunkSpec extends ZIOBaseSpec {
     testM("apply") {
       check(chunkWithIndex(Gen.unit)) {
         case (chunk, i) =>
-          assert(chunk.apply(i))(equalTo(chunk.toSeq.apply(i)))
+          assert(chunk.apply(i))(equalTo(chunk.toList.apply(i)))
```
jdegoes, line 63: We can omit the whole `.toList.`and use the `apply` directly on Chunk.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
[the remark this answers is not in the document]
adamgfraser: I am incorporating a solution in this PR. Will push an updated version shortly.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
jdegoes: Ah never mind, it's testing that it's the same as List implementation.

adamgfraser: I think the goal in this test was to make sure the apply method itself is accurate, and so we are comparing the `apply` implementation in `Chunk` with the known correct value from accessing the same index in the list. I suppose maybe it would be better to check against `Vector` here.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -382,17 +382,17 @@ object ChunkSpec extends ZIOBaseSpec {
...
-      trait A
-      trait B extends A
+    //   trait A
+    //   trait B extends A
 
-      val empty: Chunk[B] = Chunk.empty
+    //   val empty: Chunk[B] = Chunk.empty
 
-      val _: NonEmptyChunk[A] = empty ++ Chunk(new A {})
+    //   val _: NonEmptyChunk[A] = empty ++ Chunk(new A {})
```
jdegoes, line 393: Dead code in light of the changes to `++`? Here and above.

adamgfraser: Yes, I think we are going to have to add a new operator with a different name, something like `concatNonEmpty`.

adamgfraser:
This is ready for another review. I added a `ChunkCanBuildFrom` subtype of `CanBuildFrom`. We can pattern against this to prove that the target type `That` is a `Chunk`, allowing us to use all of our existing efficient implementations of `Chunk` combinators.

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
```diff
@@ -0,0 +1,187 @@
...
+    val len                    = self.length
+    var chunks: List[Chunk[B]] = Nil
+
+    var i               = 0
+    var total           = 0
+    var B0: ClassTag[B] = null.asInstanceOf[ClassTag[B]]
+    while (i < len) {
+      val chunk = f(self(i)) match {
+        case chunk: Chunk[B] => chunk
+        case other           => Chunk.fromIterable(other.toList)
```
jdegoes, line 138: This `toList` will be detrimental to performance. Probably best to do `Chunk.fromArray(other.toArray)` if we need to do that.

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
```diff
@@ -0,0 +1,187 @@
...
+      Chunk.arr(dest)
+    }
+  }
+
+  /**
+   * Constructs a new `ChunkBuilder`. This operation allocates mutable state
+   * and is not referentially transparent. It is provided for compatibility
+   * with Scala's collection library and should not be used for other purposes.
+   */
+  override protected[this] def newBuilder: ChunkBuilder[A] =
```
jdegoes, line 175: By convention, side-effecting methods should have nullary parameter list: `def newBuilder()`.

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
adamgfraser: Will change.

core/shared/src/main/scala-2.13+/ChunkBuilder.scala
```diff
@@ -0,0 +1,63 @@
...
+ * the approach in `Chunk`.
+ */
+private[zio] trait ChunkBuilder[A] extends Builder[A, Chunk[A]]
+
+private[zio] object ChunkBuilder {
+
+  /**
+   * Constructs a new `ChunkBuilder`.
+   */
+  def make[A]: ChunkBuilder[A] =
```
jdegoes, line 41:
```suggestion
  def make[A](capacity: Int = 10): ChunkBuilder[A] =
```

I saw you could call `ensureSize` on `ArrayBuilder` if you sub class it. Which would allow "pre-allocation" in cases where it's known about how many things will be added.

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
adamgfraser: Yes. The signature of the `newBuilder` method we are overriding doesn't have the nullify parameter list. Let me see if I can add and still have it be recognized as a valid override.

core/shared/src/main/scala-2.13+/ChunkLike.scala
```diff
@@ -0,0 +1,132 @@
...
+    val len                    = self.length
+    var chunks: List[Chunk[B]] = Nil
+
+    var i               = 0
+    var total           = 0
+    var B0: ClassTag[B] = null.asInstanceOf[ClassTag[B]]
+    while (i < len) {
+      val chunk = f(self(i)) match {
+        case chunk: Chunk[B] => chunk
+        case other           => Chunk.fromIterable(other.iterator.to(List))
```
jdegoes, line 64: `Chunk.fromArray(...toArray)` if possible.

If not possible maybe we should directly make a `Chunk.fromIterator` or something since going "through" List will not be efficient.

core/shared/src/main/scala-2.13+/ChunkLike.scala
```diff
@@ -0,0 +1,132 @@
...
+   * Returns the empty `Chunk`.
+   */
+  def empty[A]: Chunk[A] =
+    Chunk.empty
+
+  /**
+   * Constructs a `Chunk` from the specified `IterableOnce`.
+   */
+  def from[A](source: IterableOnce[A]): Chunk[A] =
+    Chunk.fromIterable(source.iterator.to(Iterable))
```
jdegoes, line 123: Same here.

core/shared/src/main/scala-2.13+/ChunkBuilder.scala
adamgfraser: Yes, I think we want to override the `sizeHint` method. Because these methods will mostly be called by Scala collection library combinators that are outside our control.

project/BuildHelper.scala
```diff
@@ -259,7 +271,8 @@ object BuildHelper {
               Seq(file(sourceDirectory.value.getPath + "/main/scala-2.12+")),
               CrossType.Full.sharedSrcDir(baseDirectory.value, "main").toList.map(f => file(f.getPath + "-2.12+")),
               CrossType.Full.sharedSrcDir(baseDirectory.value, "test").toList.map(f => file(f.getPath + "-2.12+")),
-              CrossType.Full.sharedSrcDir(baseDirectory.value, "main").toList.map(f => file(f.getPath + "-dotty"))
+              CrossType.Full.sharedSrcDir(baseDirectory.value, "main").toList.map(f => file(f.getPath + "-dotty")),
+              CrossType.Full.sharedSrcDir(baseDirectory.value, "main").toList.map(f => file(f.getPath + "-2.13+"))
```
jdegoes, line 275: Sbt's not going to like this. 😆

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
adamgfraser: We can't call `toArray` here because we don't have a `ClassTag` at this point. May make sense to refactor this so we just iterate over the original collection and either get the class tag from the `Chunk` if it is a chunk or otherwise from the first value of the collection using `Tag.fromValue`.

core/shared/src/main/scala-2.13+/ChunkLike.scala
adamgfraser: I think adding that method is a good idea.

project/BuildHelper.scala
adamgfraser: I was surprised this worked as smoothly as it did!

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
jdegoes: Right. I think if we can only iterate over it once, then the best we can do is use the chunk builder to turn the "iterable once" into a chunk.

OTOH if we match against a few other cases, e.g. Vector, List, etc., we know what collection type we are dealing with and can handle it more specially (e.g. Vector we can preallocate, List cannot be pre-allocated since it's O(n) to determine the size, etc.).

But open question how much of that performance work to do here. I'm fine merging this in now, we can create tickets for performance work.

adamgfraser: I've got an idea about how we can do this efficiently. Will do a turn on it now.

core/shared/src/main/scala-2.11-2.12/ChunkLike.scala
adamgfraser: Yes unfortunately adding the nullary parameter list generate a compiler error so I think we will have to leave as is.

adamgfraser:
Done.

PR #3346, opened 2020-04-12
Add test for Schedule#onDecision and documentation for scheduling

marktoujiline:
While using `Schedule`, I found that getting `onDecision` to work correctly was a bit of a challenge. There were no usages of it in the repo to use as a reference. With some help, I was able to get it to work.

I'd like to add a test for the function since none exists as well as add some additional documentation. Hopefully the combination of the two will provide enough information for anyone else interested in using the very helpful `Schedule#onDecision` method.

Many of the Use Cases pages are blank, so I'm not sure what the original vision for them was. I operated under the assumption that "use cases" means practical applications.

Another question/concern I have is about the use of `putStr` in the `ScheduleSpec` file. Ideally, I'd be able to avoid depending on it, but I couldn't think of how else I could track a `URIO` as required by `onDecision`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -3,8 +3,10 @@ package zio
 import scala.concurrent.Future
 
 import zio.clock.Clock
+import zio.console.putStr
 import zio.duration._
 import zio.test.Assertion._
+import zio.test.environment.TestConsole
 import zio.test.environment.{ TestClock, TestRandom }
 import zio.test.{ assert, assertM, suite, testM, TestResult }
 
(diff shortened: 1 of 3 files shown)


core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
```diff
@@ -319,6 +321,23 @@ object ScheduleSpec extends ZIOBaseSpec {
           (Schedule.stop || (Schedule.spaced(2.seconds) && Schedule.stop)) >>> testElapsed
         )(List.fill(5)(()))
       )(equalTo(List(Duration.Zero)))
+    },
+    testM("perform log for each recurrence of effect") {
```
adamgfraser, line 325: Thanks for contributing! You can create a `Ref`, then update the `Ref` from within `onDecision`, and get the `Ref` at the end of your test to verify that this method behaves correctly.

adamgfraser:

PR #3349, opened 2020-04-13
ZTransducer folds

regiskuckaertz:
fold/foldM
foldUntil/foldUntilM
foldWeighted/foldWeightedM
foldWeightedDecompose/foldWeightedDecomposeM
collectAllN/collectAllWhile/collectAllWhileM


streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +125,250 @@ object ZTransducer {
   def fromPush[R, E, I, O](push: Option[Chunk[I]] => ZIO[R, E, Chunk[O]]): ZTransducer[R, E, I, O] =
     ZTransducer(Managed.succeed(push))
 
+  /**
+   * Creates a transducer by folding over a structure of type `S`.
+   */
+  def fold[I, O](z: O)(contFn: O => Boolean)(f: (O, I) => (O, Chunk[I])): ZTransducer[Any, Nothing, I, O] =
```
iravid, line 131: Why does the step function return `Chunk[I]` in addition to `O`? Is this for leftovers? If so, it isn't necessary; the operator can check the `contFn` after every application of `(O, I) => O` to see if it should continue applying `f` to the chunk.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +125,250 @@ object ZTransducer {
...
+    ZTransducer {
+      final case class FoldState(started: Boolean, result: O, buffer: Chunk[I])
```
iravid, line 133: Why do we need to buffer `I` values? Conceptually the operator can fold incoming chunks using `f` and whenever `contFn == false`, the resulting `O` is added to the output chunk. Multiple `O` values can be emitted in a single chunk. This way you'd only need to buffer an `O` value for which `contFn` was always true.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Nice, I blindly copied the existing one. There's another place where remainders point their nose, it's in `foldDecompose`. AFAICS, in the existing implementation, the result of `decompose(input)` is added to the leftovers, but since there is no such thing here, I have no other choice than to add it to the buffer. The danger is that decomposing is called ad nauseam.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Ah, right, I'll address that.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Perhaps we could borrow from the comonoid stuff and change decompose to `A => Option[(A, A)]`

regiskuckaertz: I guess the empty chunk can be None here.

regiskuckaertz: Meaning, if it returns the empty chunk, I have no other choice but to cross the maximum, otherwise I drop an element from the stream.

regiskuckaertz:
I removed the buffer in foldWeightDecompose too, and it slightly changes the behaviour: it will keep decomposing until either it is not possible anymore (at which point there is no choice but to cross max) or we reach another branch of `go`.

iravid:
Awesome! Will review once I'm off work 👍🏻

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+            } else if (total == max)
+              go(in.drop(1), os0 + f(state.result, i), initial)
+            else
+              go(in.drop(1), os0, FoldWeightedState(true, f(state.result, i), total))
+        }
+
+      ZRef.makeManaged(initial).map { state =>
+        {
+          case Some(in) => state.modify(go(in, Chunk.empty, _))
+          case None     => state.getAndSet(initial).map(s => if (s.started) Chunk.single(s.result) else Chunk.empty)
```
iravid, line 278: Same comment re started

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+
+      @tailrec def go(in: Chunk[I], os0: Chunk[O], state: FoldWeightedState): (Chunk[O], FoldWeightedState) =
+        in.headOption match {
+          case None => os0 -> state
+          case Some(i) =>
+            val total = state.cost + costFn(i)
+            if (total > max) {
+              val is = decompose(i)
+              if (is.isEmpty)
+                go(in.drop(1), os0 + f(state.result, i), initial)
```
iravid, line 264: We can use `Chunk#foldWhile` here too.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+              else if (is.length == 1)
+                go(in.drop(1), os0 + f(state.result, is(0)), initial)
```
iravid, line 266: We should avoid incrementally building chunks like this (at least until we have tree rebalancing in place). Let's try and build the entire output chunk at once?

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+    costFn: I => ZIO[R, E, Long],
+    max: Long,
+    decompose: I => ZIO[R, E, Chunk[I]]
+  )(f: (O, I) => ZIO[R, E, O]): ZTransducer[R, E, I, O] =
+    ZTransducer {
+      final case class FoldWeightedState(started: Boolean, result: O, cost: Long)
+
+      val initial = FoldWeightedState(false, z, 0)
+
+      def go(in: Chunk[I], os: Chunk[O], state: FoldWeightedState): ZIO[R, E, (FoldWeightedState, Chunk[O])] =
```
iravid, line 316: Same comments as in `foldWeightedDecompose`

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -90,27 +94,27 @@ object ZTransducer {
...
-                  if (!p(el)) (newBuffer.reverse :: outBuffer, List())
-                  else (outBuffer, el :: newBuffer)
-              }
-
-              (Chunk.fromIterable(outBuffer.reverse.filter(_.nonEmpty)), Chunk.fromIterable(newBuffer.reverse))
-            }
-        }
-      }
-    }
+    fold[I, (List[I], Boolean)]((Nil, true))(_._2) {
```
iravid, line 107: Love this btw!

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+            if (contFn(o))
+              go(in.drop(1), os0, FoldState(true, o))
+            else
+              go(in.drop(1), os0 + o, initial)
+        }
+
+      ZRef.makeManaged(initial).map { state =>
+        {
+          case Some(in) => state.modify(go(in, Chunk.empty, _))
+          case None     => state.getAndSet(initial).map(s => if (s.started) Chunk.single(s.result) else Chunk.empty)
```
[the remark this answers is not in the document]
regiskuckaertz: Yes, though when it comes to responding to `push(None)`, you have to account for the possibility that the transducer is in the middle of a fold, because if it isn't you cannot produce `Chunk(z)` but `Chunk.empty`.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +128,223 @@ object ZTransducer {
...
+
+      val initial = FoldState(false, z)
+
+      @tailrec def go(in: Chunk[I], os0: Chunk[O], state: FoldState): (Chunk[O], FoldState) =
+        in.headOption match {
+          case None => os0 -> state
+          case Some(i) =>
+            val o = f(state.result, i)
+            if (contFn(o))
+              go(in.drop(1), os0, FoldState(true, o))
```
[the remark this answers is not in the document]
regiskuckaertz: Yeah, let me work this out 👍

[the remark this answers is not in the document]
regiskuckaertz: I think `fold` would be best then for a single pass.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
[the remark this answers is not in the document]
iravid: You can always emit `Chunk(state.z)` if the initial `FoldState` contains the `z` passed to the function, no?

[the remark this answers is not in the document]
regiskuckaertz: So it's fine the transducer outputs `Chunk(z)` ad infinitum?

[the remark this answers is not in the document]
regiskuckaertz: I guess that makes sense, as that is the behaviour on an empty input. Thanks, will do that!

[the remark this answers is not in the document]
regiskuckaertz: Not so sure after all. I updated foldDecompose just for an example, but it breaks foldUntil because now when the stream pushes None the zero will be output and there is no way to know whether it is a valid output calculated from some previous runs, or it is because it was reset already.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Sadly not, because `decompose` will spread the current input and it has to be inserted in place. I could change `go` to take a `List[Chunk[I]]` as an input instead, that would save the `is ++ in.drop(1)`.

iravid: Ah right. This suggests something like a stack into which you push the current chunk being iterated upon. Sounds a bit overkill for this stage of development though.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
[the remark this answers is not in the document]
iravid: Can you write out the inputs for this case?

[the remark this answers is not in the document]
regiskuckaertz: Yep, the test:
```scala
ZStream[Long](1, 1, 1, 1, 1, 1)
  .aggregate(ZTransducer.foldUntil(0L, 3)(_ + (_: Long)))
  .runCollect
```
yields `List(3,3,0)` instead of `List(3,3)`

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Forget it, I just found a way to do it single-pass 🤦

regiskuckaertz: ... though I'd suggest to change `decompose: A => Option[(A, A)]`

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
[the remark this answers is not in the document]
iravid: Cool. That's a great example. I agree that the result is not satisfactory in this case.

For `ZStream.empty` though, it still seems that we should emit the initial state. What do you think? I'll check what other libraries are doing in this case.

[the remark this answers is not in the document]
iravid: We should also make sure that we're consistent with `ZStream#mapAccum` as a transducer fold is basically a scan.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
iravid: Is that a must? It's inefficient for cases in which you can decompose to more than two elements.

regiskuckaertz: not a must no, check out this version: no `drop`, no `++`, but there is a recursive fold.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: I've been looking at the rebalancing work from a distance. I don't really understand why we have `Singleton`, `Slice` and `Vector` chunks. Isn't locality of reference more important than space economy on tiny chunks?

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
iravid: Looks good! The danger here is that it isn't stacksafe, so decompositions that loop will blow that stack. That's a feature in some sense :-)

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
iravid: So Slice is there to make slicing operations (like take, drop, etc.) O(1) and zero-copy; similarly Vector is there because it's fine to keep as-is and avoid copying. Singleton is probably redundant because an array would occupy just an additional word in memory.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
regiskuckaertz: Yes 😆

iravid:
Gotta head to bed. I'll review again tomorrow 💪🏻

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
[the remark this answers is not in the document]
regiskuckaertz: > For ZStream.empty though, it still seems that we should emit the initial state. What do you think? I'll check what other libraries are doing in this case.

I agree. I can just replace the current state with `Option[O]`, set it to `Some(z)` initially and reset to `None` after each emit.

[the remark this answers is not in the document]
iravid: @regiskuckaertz Sounds perfect.

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +127,215 @@ object ZTransducer {
...
+
+      def go(in: Chunk[I], os0: Chunk[O], state: FoldWeightedState): (Chunk[O], FoldWeightedState) =
+        in.fold[(Chunk[O], FoldWeightedState)](os0 -> state) {
+          case ((os0, state), i) =>
+            val total = state.cost + costFn(i)
+            if (total > max) {
+              val is = decompose(i)
+              if (is.isEmpty)
+                (os0 + f(state.result, i)) -> initial
+              else if (is.length == 1)
```
iravid, line 257: I like this escape hatch! Worth adding a note to the scaladoc 👍🏻

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -124,6 +127,215 @@ object ZTransducer {
...
+                (os0 + f(state.result, is(0))) -> initial
+              else
+                go(is, os0, state)
```
iravid, line 260: Lovely ❤️

regiskuckaertz:
Excellent! Onto `aggregateAsyncWithin` now 🤓

PR #3352, opened 2020-04-13
Port ZSink#race

simpadjo:
Not quite happy with the size of my implementation :(

Some irrelevant changes are just results of running `sbt prepare`


simpadjo:
Fails on 2.13 due to irrelevant issues.

Ping @iravid

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -215,6 +215,97 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
+              onlyOne
+                .active(in)
+                .foldM({
+                  case Left(err) => onlyOne.halt(err)
+                  case Right(z)  => onlyOne.emit(z)
+                }, _ => Push.more)
+            }
+            case None => {
+              p1(in).either.zipPar(p2(in).either).flatMap {
+                case (res1, res2) => {
```
iravid, line 270: image

simpadjo: Thanks god there are only 2 sinks to race!
I tried to come with a state structure that could handle 2 updates one-by-one.
But after few attempts just threw a pragmatic hadouken.

simpadjo:
@iravid I started porting `zipPar` as well and understood that it's gonna be very hadoukenish as well. 
`zipPar` and `race` are very similar and that's good.
I feel lucky and think that it's possible to implement them all at once.
So I 99% will rewrite this code. 
OTOH it would be nice to have everything ported ASAP. 
Please decide yourself if it worth merging.

iravid:
@simpadjo Happy to wait with this if you have zipPar in the pipeline as well.

simpadjo:
@iravid looks better now. Haven't yet implemented `zipWithPar`but now it will be very easy

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -215,6 +216,70 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
+  ): ZSink[R1, E1, I1, Z1] =
+    self.raceBoth(that).map(_.merge)
+
+  /**
+   * Runs both sinks in parallel on the input, returning the result from the
+   * one that finishes successfully first.
+   */
+  final def raceBoth[R1 <: R, E1 >: E, A0, I1 <: I, Z1](
+    that: ZSink[R1, E1, I1, Z1]
+  ): ZSink[R1, E1, I1, Either[Z, Z1]] = {
```
regiskuckaertz, line 242: Can't we use the `ZIO#race*` combinators in some fashion?
```scala
for {
  p1 <- self.push
  p2 <- that.push
  push = (in: Option[Chunk[I1]]) => p1(in).raceEither(p2(in))
} yield push
```
☝️ this is probably too good to be true, but just for the sake of the argument.

simpadjo: Generally we need results from both sides because the first result may not yield a value.

`raceWith` in theory could in theory save from waiting for one value at the very last step...

`zipPar` has been used in the old implementation so chances to solve it with a one-liner are low.

regiskuckaertz: gotcha. yeah I'm pretty sure the new encoding is more amenable to these idioms than the previous sink state machinery. Perhaps:
```scala
p1(in).raceWith(p2(in))(
  (exit1, fiber2) => exit1.foldM(
    Cause.sequenceCauseOption(_).fold(fiber2.join)(ZIO.halt(_)),
    a => ZIO.succeedNow(Left(a))
  ),
  (exit2, fiber1) => ...
)
```
Just a thought though.

iravid: `raceWith` will definitely save a whole lot of boilerplate on this combinator. Worth a try!

simpadjo: @iravid it is already based on `zipPar`

iravid: Yes - I saw that - it is worth trying to utilize raceWith there as the logic you've implemented is very similar.

simpadjo:
After discussions with @iravid semantics of `raceWith` changed a bit. 
Now it returns the result from the fastest sink (no matter if it succeeded of failed). 
Not sure if it's right or not but the implementation is much shorter.

@iravid probably it should be added to the release notes draft of to the ticket

iravid:
@simpadjo Agreed about the release notes 👍🏻 We'll need both implementations eventually, but I'd rather tick these off quickly so we can get the big branch merged.

Will review now, thank you for your patience!

PR #3356, opened 2020-04-13
ZIO Test: Improve Ergonomics of Laws for Higher Kinded Types

adamgfraser:
Minor improvements to support for laws for parameterized types based on initial experience working with them:

1. It is nice for the user not to have to create their own `GenF` for standard types, so provides those.
2. Adds a few more type aliases.
3. Adds the law constructor for one `F[A]` and two functions `A => B` and `B => C` for contravariant and invariant types. We still have a little problem of needing more law constructors for different shapes of inputs, but the user experience is so much better when we provide them that I think we just try to do it for reasonable combinations that come up.


test/shared/src/main/scala/zio/test/laws/ZLawsF.scala
```diff
@@ -196,6 +196,16 @@ object ZLawsF {
       final def run[R1 <: R, F[-_]: CapsF, A: Caps](genF: GenF[R1, F], gen: Gen[R1, A]): ZIO[R1, Nothing, TestResult] =
         checkM(genF(gen), genF(gen), genF(gen))(apply(_, _, _).map(_.map(_.label(label))))
     }
+
+    /**
+     * Constructs a law from a pure function taking three parameters.
+     */
+    abstract class Law3Function[-CapsF[_[-_]], -Caps[_]](label: String) extends Contravariant[CapsF, Caps, Any] {
```
jdegoes, line 203: This is always a function compose law, so maybe `LawCompose` or `LawFunctionCompose`?

test/shared/src/main/scala/zio/test/laws/ZLawsF.scala
```diff
@@ -280,5 +290,14 @@ object ZLawsF {
       final def run[R1 <: R, F[_]: CapsF, A: Caps](genF: GenF[R1, F], gen: Gen[R1, A]): ZIO[R1, Nothing, TestResult] =
         checkM(genF(gen), genF(gen), genF(gen))(apply(_, _, _).map(_.map(_.label(label))))
     }
+
+    /**
+     * Constructs a law from a pure function taking three parameters.
+     */
+    abstract class Law3Function[-CapsF[_[_]], -Caps[_]](label: String) extends Invariant[CapsF, Caps, Any] { self =>
```
jdegoes, line 297: For invariant, the likely form this law would take would be `<=>` instead of `=>`. So I'm not sure whether to put that in here or to defer it downstream. Thinking about it, nothing stops us from including these type classes in ZIO Prelude, so users can do things like `new zio.prelude.Invaraint.ComposeLaw`.

test/shared/src/main/scala/zio/test/laws/ZLawsF.scala
jdegoes: The other possibility is to push it into ZIO Prelude as a class in the companion object of the associated type class (Covariant).

test/shared/src/main/scala/zio/test/laws/ZLawsF.scala
adamgfraser: Yes, for this one in particular we have to be quite careful in the way we generate the pair of functions so they form an equivalence relation.

test/shared/src/main/scala/zio/test/laws/ZLawsF.scala
adamgfraser: I like the renaming. Pushing it to ZIO Prelude is interesting. The disadvantage there would be that it is less ergonomic if someone else wants to implement their own law with the same shape, or even if we want to do so somewhere else in ZIO Prelude. But it does solve the problem that we don't necessarily know at the ZIO Test love the shape of all the laws we will want to test.

adamgfraser:
Renamed covariant and contravariant versions to `ComposeLaw` and removed the invariant version.

jdegoes:
@adamgfraser Good to merge when conflicts are fixed!

adamgfraser:
@jdegoes All set.

adamgfraser:
@jdegoes Could I get a review on this? I added a `FlattenLaw` and more importantly unsealed the law traits so that downstream libraries can implement their own laws when the "shapes" of laws they need don't match the ones already provided.

jdegoes:
@adamgfraser I like that direction (opening up the trait). If placed in the companion object of the corresponding type class, it should be easy for users to reuse these laws and know where to look for them. 👍

PR #3358, opened 2020-04-13
Balanced Concat

ahoy-jon:
following #3311 , fixing #3238 

This PR reimplement `+` and `concat` on Chunk, without changing the substructure.
On both methods is compute an insertion point in the tree using a local loop.

On `+`:
* it materializes the subtree before moving it when the `length` is equal to `8`.
* it uses a tail (the right-hand side of the root) to accelerate the `add`.

### DIFF
--- a/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
@@ -1,0 +1,54 @@
+package zio.chunks
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations.{ Benchmark, BenchmarkMode, Mode, OutputTimeUnit, Param, Scope, Setup, State }
+
+import zio.{ Chunk, NonEmptyChunk }
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class ChunkConcatBenchmarks {
+
+  @Param(Array("1", "8", "64", "512", "4096", "32768"))
+  var size: Int = _
+
+  var chunk: NonEmptyChunk[Int]     = _
+  var chunk_old: NonEmptyChunk[Int] = _
+  var array: Array[Int]             = _
+  var vector: Vector[Int]           = _
+
+  @Setup
+  def setup(): Unit = {
+    chunk = chunkAdd
+    chunk_old = chunkAdd_old
+    array = arrayAdd
+    vector = vectorAdd
+  }
+
+  @Benchmark
+  def chunkAdd = (1 to size).foldLeft(Chunk(0))(_ + _)
+
+  @Benchmark
+  def chunkAdd_old = (1 to size).foldLeft(Chunk(0))(Chunk.addOld)
+
+  @Benchmark
+  def arrayAdd = (1 to size).foldLeft(Array(0))(_ :+ _)
+
+  @Benchmark
+  def vectorAdd = (1 to size).foldLeft(Vector(0))(_ :+ _)
+
+  def transform(x: Int): Int = x + 1
+
+  @Benchmark
+  def chunkMap = chunk.map(transform)
+
+  @Benchmark
+  def chunkMap_old = chunk_old.map(transform)
+
+  @Benchmark
+  def arrayMap = array.map(transform)
+
+  @Benchmark
+  def vectorMap = vector.map(transform)
+}
--- a/benchmarks/src/main/scala/zio/chunks/ChunkPlusBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkPlusBenchmarks.scala
@@ -8,7 +8,7 @@ import zio.{ Chunk, NonEmptyChunk }
 @State(Scope.Thread)
 @BenchmarkMode(Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
-class ChunkConcatBenchmarks {
+class ChunkPlusBenchmarks {
 
   @Param(Array("1", "8", "64", "512", "4096", "32768"))
   var size: Int = _
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -393,6 +393,9 @@ object ChunkSpec extends ZIOBaseSpec {
       val _: NonEmptyChunk[A] = empty concatNonEmpty Chunk(new A {})
 
       assertCompletes
-    } @@ TestAspect.ignore
+    } @@ TestAspect.ignore,
+    zio.test.test("+ on large number of elements") {
+      assert[Seq[Int]]((1 to 32767).foldLeft(Chunk(0))(_ + _))(equalTo(0 until 32768))
+    }
   )
 }
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -394,10 +394,13 @@ object ChunkSpec extends ZIOBaseSpec {
 
       assertCompletes
     } @@ TestAspect.ignore, {
-      val max = 32768
-      val out = 1 to max
+      //numbers fit in JS short representation
+      val min = 128
+      val max = 32767
 
-      val maxDepth = 32 - Integer.numberOfLeadingZeros(out.size)
+      val out = min to max
+
+      val maxDepth = 2 + 31 - Integer.numberOfLeadingZeros(out.size)
 
       def check(label: String, ints_ : => Chunk[Int]) =
         zio.test.test(label)({
@@ -407,9 +410,9 @@ object ChunkSpec extends ZIOBaseSpec {
         })
 
       suite("balanced concat/+ ")(
-        check("+", (2 to max).foldLeft(Chunk(1))(_ + _)),
-        check("concat on the left", (2 to max).foldLeft(Chunk(1))((c, i) => Chunk.concat(c, Chunk.single(i)))),
-        check("concat on the right", (1 until max).foldRight(Chunk(max))((i, c) => Chunk.concat(Chunk.single(i), c)))
+        check("+", (min + 1 to max).foldLeft(Chunk(min))(_ + _)),
+        check("concat on the left", (min + 1 to max).foldLeft(Chunk(min))((c, i) => Chunk.concat(c, Chunk.single(i)))),
+        check("concat on the right", (min until max).foldRight(Chunk(max))((i, c) => Chunk.concat(Chunk.single(i), c)))
       )
     }
   )
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -393,21 +393,23 @@ object ChunkSpec extends ZIOBaseSpec {
       val _: NonEmptyChunk[A] = empty concatNonEmpty Chunk(new A {})
 
       assertCompletes
-    } @@ TestAspect.ignore,
-    zio.test.test("+ on large number of elements") {
-      val max  = 32767
-      val ints = (1 to max).foldLeft(Chunk(0))(_ + _)
-      assert[Seq[Int]](ints)(equalTo(0 to 32767))
-    },
-    zio.test.test("concat on the left") {
-      val max  = 32767
-      val ints = (1 to max).map(Chunk.single).foldLeft(Chunk(0))(Chunk.concat)
-      assert[Seq[Int]](ints)(equalTo(0 to max))
-    },
-    zio.test.test("concat on the right") {
-      val max  = 32767
-      val ints = (1 to max).map(Chunk.single).foldLeft(Chunk(0))((acc, e) => Chunk.concat(e, acc))
-      assert[Seq[Int]](ints)(equalTo(max.to(0, -1)))
+    } @@ TestAspect.ignore, {
+      val max = 32768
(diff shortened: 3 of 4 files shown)


ahoy-jon:
Finished the work in the scope (no change to the structure). @adamgfraser / @jdegoes if you could review it!

jdegoes:
Can you post benchmarks? This looks like it will significantly slow down concat versus the old implementation. It would be nice to see the tradeoffs.

iravid:
Wouldn't it make more sense to track the subtree depth as a constructor parameter in `Concat` and only do the rebalancing operation when the depth is `> n` for some arbitrary `n`?

ahoy-jon:
@iravid I am really not sure it is needed, the length is already available and a good metric for the depth if we assume the structures are already balanced and we don't want to do a lot more allocations.

eg. 
```scala
val chunck = Concat(Concat(Single(1), Single(2)), Concat(Single(3), Single(4))) 
chunk.size = 4
chunk.depth = 3
2 + 31 - Integer.numberOfLeadingZeros(4 - 1) == 3

val chunck = Concat(Concat(Concat(Single(1), Single(2)), Concat(Single(3), Single(4))), Single(5))
chunk.size = 5
chunk.depth = 4
2 + 31 - Integer.numberOfLeadingZeros(5 - 1) == 4
//...
```

At the end it gives already trees like that :
```
┳┳┳┳┳━  0
┃┃┃┃┗━  1
┃┃┃┗┳━  2
┃┃┃ ┗━  3
┃┃┗┳┳━  4
┃┃ ┃┗━  5
┃┃ ┗┳━  6
┃┃  ┗━  7
┃┗┳┳┳━  8
┃ ┃┃┗━  9
┃ ┃┗┳━ 10
┃ ┃ ┗━ 11
┃ ┗┳┳━ 12
┃  ┃┗━ 13
┃  ┗┳━ 14
┃   ┗━ 15
┗┳━━━━ 16
 ┗━━━━ 17
```

(I need to update the test to add the '-1')

If you want to try things out with the algorithm, we have this toy exercise:
```scala
package io.univalence.exercice

import org.scalatest.funsuite.AnyFunSuite

import scala.annotation.tailrec

sealed trait TreeSeq[+A] {

  lazy val size: Int = this match {
    case Leaf(_)             => 1
    case Branch(left, right) => left.size + right.size
  }

  def +[A1 >: A](a1: A1): TreeSeq[A1] = this ++ Leaf(a1)

  def ++[A1 >: A](tree: TreeSeq[A1]): TreeSeq[A1] = {
    this match {
      case Branch(l, r) if l.size >= r.size + tree.size =>
        Branch(l, r ++ tree)
      case _ => Branch(this, tree)
    }
  }

  def map[B](f: A => B): TreeSeq[B] = this match {
    case Leaf(a)      => Leaf(f(a))
    case Branch(l, r) => Branch(l.map(f), r.map(f))
  }

  def depth: Int = this match {
    case Leaf(_)      => 1
    case Branch(l, r) => Math.max(l.depth, r.depth) + 1
  }

  def pprintSimple: String = {
    def pprintSimple(treeSeq: TreeSeq[A], prefix: String): String =
      treeSeq match {
        case Leaf(a) => prefix + "Leaf(" + a + ")"
        case Branch(l, r) =>
          (prefix + "Branch") + "\n" +
            pprintSimple(l, prefix + "  ") + "\n" +
            pprintSimple(r, prefix + "  ")
      }

    pprintSimple(this, "")
  }

  def pprint: String = {

    def mapCap(vec: Vector[(String, A)])(onFirst: String,
                                         e: String): Vector[(String, A)] = {
      vec.zipWithIndex.map({
        case ((p, v), 0) => onFirst + p -> v
        case ((p, v), _) => e + p -> v
      })
    }

    def prefix(self: TreeSeq[A]): Vector[(String, A)] =
      self match {
        case Leaf(a) => Vector(("━", a))
        case Branch(l, r) =>
          mapCap(prefix(l))("┳", "┃") ++
            mapCap(prefix(r))("┗", " ")
      }

    val lines = prefix(this)
    val size = lines.map(_._1.length).max
    val stringSize = lines.map(_._2.toString.length).max

    lines
      .map({
        case (p, v) =>
          val rep = v.toString
          val c = size - p.length
          val c2 = 1 + stringSize - rep.length
          p + "━" * c + " " * c2 + rep
      })
      .mkString("\n")

  }

  def fold[B](leaf: A => B)(branch: (B, B) => B): B =
    this match {
      case Leaf(a)      => leaf(a)
      case Branch(l, r) => branch(l.fold(leaf)(branch), r.fold(leaf)(branch))
    }

  def toVector: Vector[A] = {

    @tailrec
    def loop(res: Vector[A],
             point: TreeSeq[A],
             rest: List[TreeSeq[A]]): Vector[A] = {
      point match {
        case Leaf(a) =>
          rest match {
            case Nil     => res :+ a
            case x :: xs => loop(res :+ a, x, xs)
          }
        case Branch(l, r) => loop(res, l, r :: rest)
      }
    }

    loop(Vector.empty, this, Nil)
  }

}

case class Branch[A](left: TreeSeq[A], right: TreeSeq[A]) extends TreeSeq[A]

case class Leaf[A](value: A) extends TreeSeq[A]

object TreeSeq {

  def single[A](value: A): TreeSeq[A] = Leaf(value)

  def apply[A](head: A, tail: A*): TreeSeq[A] = {
    tail.foldLeft(single(head))(_ + _)
  }
}

class TreeSeqTest extends AnyFunSuite {

  test("print") {

    (1 to 17).foldLeft(TreeSeq.single(0))((x, e) => {

      val z = x + e

      println(z.pprintSimple)

      z

    })

  }

  test("all") {

    assert(TreeSeq.single(1).pprint == "━ 1")
    val s12: Branch[Int] = Branch(Leaf(1), Leaf(2))
    assert(
      s12.pprint ==
        """┳━ 1
          |┗━ 2""".stripMargin
    )
    assert(
      Branch(s12, Branch(s12, s12)).pprint ==
        """┳┳━━ 1
          |┃┗━━ 2
          |┗┳┳━ 1
          | ┃┗━ 2
          | ┗┳━ 1
          |  ┗━ 2""".stripMargin
    )

    //Optimized balance

    assert(TreeSeq(1) == Leaf(1))
    assert(TreeSeq(1, 2) == s12)
    assert(TreeSeq(1, 2, 3) == Branch(s12, Leaf(3)))

    val t1234 = TreeSeq(1, 2, 3, 4)

    assert(t1234 == Branch(s12, Branch(Leaf(3), Leaf(4))))
    assert(TreeSeq(1, 2, 3, 4, 5) == Branch(t1234, Leaf(5)))
    assert(TreeSeq(1, 2, 3, 4, 5, 6) == Branch(t1234, TreeSeq(5, 6)))

    val t1to8 = TreeSeq(1, 2, 3, 4, 5, 6, 7, 8)
    assert(t1to8 == Branch(t1234, TreeSeq(5, 6, 7, 8)))
  }
}
```
Happy to discuss about it on discord! ( ahoy-jon#3887 )

ahoy-jon:
> do the rebalancing operation when the height is `> n` for some arbitrary `n`?

Sorry, did respond completely, Currently, concat is not doing any rewriting (materialize and build a new chunk).

what I understand is, if you have : 
```
C(C(C(1, 2), C(3 , 4)), 5) ++ C(6, 7)
```
you want
```
C(C(C(C(1, 2), C(3 , 4)), 5), C(6, 7))
```
instead of
```
C(C(C(1, 2), C(3 , 4)), C(5, C(6, 7)))
```

it is possible indeed to do that if you keep the lenght information somewhere or change the substructure, and having something like : 
if (2^depth / length >  XXX or someother condition) then rebalance. 
it could reduce the number of allocations.

ahoy-jon:
@jdegoes 
quick benchmarks, will run some more later on : 
```scala
[info] Benchmark                         (size)  Mode  Cnt         Score   Error  Units
[info] ChunkPlusBenchmarks.chunkAdd           1  avgt    2        55.622          ns/op
[info] ChunkPlusBenchmarks.chunkAdd           8  avgt    2      1512.212          ns/op
[info] ChunkPlusBenchmarks.chunkAdd          64  avgt    2     17324.647          ns/op
[info] ChunkPlusBenchmarks.chunkAdd         512  avgt    2     99276.312          ns/op
[info] ChunkPlusBenchmarks.chunkAdd        4096  avgt    2    818831.413          ns/op
[info] ChunkPlusBenchmarks.chunkAdd       32768  avgt    2   6549688.275          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old       1  avgt    2        97.993          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old       8  avgt    2       458.822          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old      64  avgt    2      3370.762          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old     512  avgt    2     26932.775          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old    4096  avgt    2    219055.941          ns/op
[info] ChunkPlusBenchmarks.chunkAdd_old   32768  avgt    2   1788192.144          ns/op
[info] ChunkPlusBenchmarks.chunkMap           1  avgt    2        62.032          ns/op
[info] ChunkPlusBenchmarks.chunkMap           8  avgt    2       108.610          ns/op
[info] ChunkPlusBenchmarks.chunkMap          64  avgt    2      1152.556          ns/op
[info] ChunkPlusBenchmarks.chunkMap         512  avgt    2     14543.794          ns/op
[info] ChunkPlusBenchmarks.chunkMap        4096  avgt    2    140451.777          ns/op
[info] ChunkPlusBenchmarks.chunkMap       32768  avgt    2   1407131.407          ns/op
[info] ChunkPlusBenchmarks.chunkMap_old       1  avgt    2        65.812          ns/op
[info] ChunkPlusBenchmarks.chunkMap_old       8  avgt    2       157.346          ns/op
[info] ChunkPlusBenchmarks.chunkMap_old      64  avgt    2      4721.452          ns/op
[info] ChunkPlusBenchmarks.chunkMap_old     512  avgt    2    314749.492          ns/op
[info] ChunkPlusBenchmarks.chunkMap_old    4096  avgt    2  26107095.110          ns/op
[success] Total time: 945 s (15:45), completed 14 Apr 2020, 18:32:23
```
Conclusions : 
* add is sometimes 4 times slower
* map is sometimes 200 times faster (usually around 4 times faster)

iravid:
I'm suggesting something like this:
```scala
case class Concat[A](l: Chunk[A], r: Chunk[A], depth: Int)
```

With `Chunk.concat` doing:
```
(l, r) match {
  case (l: Chunk.Concat, r: Chunk.Concat) => Chunk.Concat(l, r, l.depth max r.depth)
}
```
and so forth. This way we don't pay the constant cost of materialization on every concatenation.

iravid:
For the benchmarks, it is also important to measure the cost of `++`.

ahoy-jon:
@iravid 
```scala
[info] Benchmark                              (time)  Mode  Cnt          Score   Error  Units
[info] ChunkConcatBenchmarks.chunkConcat           1  avgt    2         83.652          ns/op
[info] ChunkConcatBenchmarks.chunkConcat           8  avgt    2        360.550          ns/op
[info] ChunkConcatBenchmarks.chunkConcat          64  avgt    2       4141.551          ns/op
[info] ChunkConcatBenchmarks.chunkConcat         512  avgt    2      34786.031          ns/op
[info] ChunkConcatBenchmarks.chunkConcat        4096  avgt    2     377925.411          ns/op
[info] ChunkConcatBenchmarks.chunkConcat_old       1  avgt    2         72.908          ns/op
[info] ChunkConcatBenchmarks.chunkConcat_old       8  avgt    2        176.056          ns/op
[info] ChunkConcatBenchmarks.chunkConcat_old      64  avgt    2        845.286          ns/op
[info] ChunkConcatBenchmarks.chunkConcat_old     512  avgt    2       7243.888          ns/op
[info] ChunkConcatBenchmarks.chunkConcat_old    4096  avgt    2      56198.870          ns/op
[info] ChunkConcatBenchmarks.chunkMap              1  avgt    2        435.025          ns/op
[info] ChunkConcatBenchmarks.chunkMap              8  avgt    2       3107.088          ns/op
[info] ChunkConcatBenchmarks.chunkMap             64  avgt    2     123754.174          ns/op
[info] ChunkConcatBenchmarks.chunkMap            512  avgt    2     330328.082          ns/op
[info] ChunkConcatBenchmarks.chunkMap           4096  avgt    2    2846512.480          ns/op
[info] ChunkConcatBenchmarks.chunkMap_old          1  avgt    2        426.849          ns/op
[info] ChunkConcatBenchmarks.chunkMap_old          8  avgt    2       3199.780          ns/op
[info] ChunkConcatBenchmarks.chunkMap_old         64  avgt    2      85820.007          ns/op
[info] ChunkConcatBenchmarks.chunkMap_old        512  avgt    2    5376173.577          ns/op
[info] ChunkConcatBenchmarks.chunkMap_old       4096  avgt    2  416480322.500          ns/op
[success] Total time: 828 s (13:48), completed 14 Apr 2020, 20:56:51
```

ahoy-jon:
@iravid indeed, the proposition is not performant at all.

ahoy-jon:
@vasilmkd is proposing (in #3371) to use a [`Conc-tree_list`](https://en.wikipedia.org/wiki/Conc-tree_list) which would follow @iravid on the request on adding the level of depth to the structure.


Adding the level could postpone the balancing and improve the performance on the repeated concatenation : 
```scala
private final case class Concat[A](l: NonEmpty[A], r: NonEmpty[A]) extends NonEmpty[A] { self =>

    implicit val classTag: ClassTag[A] = classTagOf(l)

    override val length: Int = l.length + r.length

    override val level: Int = Math.max(l.level, r.level) + 1

```

adamgfraser:
@ahoy-jon Are you still working on this?

ahoy-jon:
@adamgfraser Not at all, maybe I will go back to it in a couple of weeks. 
I will ping you on Discord before.

PR #3367, opened 2020-04-14
Add support for setting Random seed

reibitto:
A question in Discord brought up the fact that there currently doesn't seem to be a way to set the seed for Random.

This might be a little tricky in that Random is a "default" service that is already provided for you, but you could always override that with something like `provideLayer(random.Random.withSeed(5L))`.

Also, I'm not sure what the naming conventions would be in this situation. Some common names seem to be `default`, `live`, and `make`. And whether `live()` should take the params or not.

I went with keeping `live` the same and adding `withSeed` and `make`. But whatever is the most consistent is fine with me.

### DIFF
--- a/core/shared/src/main/scala/zio/random/package.scala
+++ b/core/shared/src/main/scala/zio/random/package.scala
@@ -20,52 +20,50 @@ package object random {
       def nextLongBounded(n: Long): UIO[Long]
       def nextPrintableChar: UIO[Char]
       def nextString(length: Int): UIO[String]
+      def setSeed(seed: Long): UIO[Unit]
       def shuffle[A](list: List[A]): UIO[List[A]]
     }
 
-    import scala.util.{ Random => SRandom }
-
     object Service {
-      val live: Service = make(new SRandom())
+      val live: Service = new Service {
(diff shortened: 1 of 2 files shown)


adamgfraser:
Scala 2.13 exposes a `setSeed` method on `Scala.util.Random`. Why don't we just make it a method on the `Random` service and call that?

reibitto:
Looks like `setSeed` has been there prior to 2.13. Yeah, I could just expose that. You want me to replace the methods I added then?

With a mutable `setSeed`, there might be cases where you need to be careful about calling setSeed early enough in the lifecycle (like if you have a Layer that depends on Random perhaps?) so that you don't generate a few values with the default seed by mistake. That's the only potential downside I can think of.

adamgfraser:
I think so. You could always call `setSeed` immediately after creating your layer.

reibitto:
Alright, makes sense. I reverted the previous changes and added setSeed.

reibitto:
Oops, I missed `MockRandom`. I've yet to use `MockRandom`, but I added `setSeed` to that too following the pattern in that file. Not sure if it needs anything else.

PR #3370, opened 2020-04-14
drop implicit CanFail constraint from stm orElse combinator

swachter:


jdegoes:
@swachter Excellent work!

PR #3374, opened 2020-04-15
Improve Support for NonEmptyChunk

### LINKED ISSUE #3357: Improve Support for NonEmptyChunk
While overall `Chunk` took a step forward with extending `IndexedSeq`, our support for the `NonEmptyChunk` concept took a step back with that since method names we would like to use are now already defined by `IndexedSeq` so right now we have the somewhat awkward `mapNonEmpty`, `flatMapNonEmpty`, and `concatNonEmpty` to preserve the information that a `Chunk` is nonempty.

I wonder if we should consider an alternative encoding:

```scala
final class private NonEmptyChunk[+A](private val chunk: Chunk[A]) {

  // methods that are guaranteed to return a nonempty chunk or are only safe on a nonempty chunk implemented here
}

object NonEmptyChunk {

  // smart constructors preserve guarantee that chunk is initially nonempty

  // can always "throw away" information and widen to chunk that may or may not be empty
  implicit def asChunk[A](nonEmpty: NonEmptyChunk[A]): Chunk[A] =
    nonEmpty.chunk
}
```

This would require a little more code on our part but I think it could have two advantages:

1. Improve ergonomics of `NonEmptyChunk` -We would be able to have methods like `map`, `flatMap`, and `++` that would do exactly what we would expect
2. Simplify `Chunk` implementation - If we did this we would not have to preserve the static type information of empty versus nonempty within `Chunk` combinators, which I think could simplify some things given we already have a lot of complexity we are dealing with.

adamgfraser:
Resolves #3357.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -181,7 +181,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
     }
 
     if (j == 0) Chunk.Empty
-    else Chunk.Slice(Chunk.arr(dest), 0, j)
+    else Chunk.Slice(Chunk.Arr(dest), 0, j)
   }
 
   /**
@@ -214,7 +214,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
     dest.map {
       case (array, arrLen) =>
         if (arrLen == 0) Chunk.empty
-        else Chunk.Slice(Chunk.arr(array), 0, arrLen)
+        else Chunk.Slice(Chunk.Arr(array), 0, arrLen)
     }
   }
 
@@ -389,7 +389,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
 
     s ->
       (if (dest == null) Chunk.empty
-       else Chunk.arr(dest))
+       else Chunk.Arr(dest))
   }
 
   /**
@@ -515,7 +515,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   def materialize[A1 >: A]: Chunk[A1] =
     self.toArrayOption[A1] match {
       case None        => Chunk.Empty
-      case Some(array) => Chunk.arr(array)
+      case Some(array) => Chunk.Arr(array)
     }
 
   /**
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -36,26 +36,26 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   /**
    * Appends an element to the chunk
    */
-  final def +[A1 >: A](a: A1): NonEmptyChunk[A1] =
-    self match {
-      case Chunk.Empty          => Chunk.single(a)
-      case ne: NonEmptyChunk[A] => Chunk.concat(ne, Chunk.single(a))
-    }
+  final def +[A1 >: A](a: A1): Chunk[A1] =
```
mschuwalow, line 39: this could still return NonEmptyChunk

adamgfraser: We could but there is potentially a little extra overhead in creating the `NonEmptyChunk` which I thought you might not want if you were doing repeated appends and were doing high performance stuff where you didn't care about whether it was empty or not (e.g. you're just calling `foreach` on it). So right now you kind of stay in the `Chunk` world unless you explicitly go to the `NonEmptyChunk` world, and once you're there we preserve that information whenever possible. But we could definitely change as you suggest.

core/shared/src/main/scala/zio/NonEmptyChunk.scala
```diff
@@ -0,0 +1,222 @@
...
+import zio.NonEmptyChunk._
+
+/**
+ * A `NonEmptyChunk` is a `Chunk` that is guaranteed to contain at least one
+ * element. As a result, operations which would not be safe when performed on
+ * `Chunk`, such as `head` or `reduce`, are safe when performed on
+ * `NonEmptyChunk`. Operations on `NonEmptyChunk` which could potentially
+ * return an empty chunk will return a `Chunk` instead.
+ */
+final class NonEmptyChunk[+A] private (private val chunk: Chunk[A]) { self =>
```
mschuwalow, line 30: Could we make this a value class?

adamgfraser: Potentially. The issue there is that then it is the same class at runtime as `Chunk` so methods that have overloads for `Chunk` and `NonEmptyChunk` like `foreach` cause compiler errors. We could possibly resolve by adding a dummy parameter to them.

PR #3376, opened 2020-04-15
Optimize BitChunk#foreach

adamgfraser:


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1091,9 +1091,14 @@ object Chunk {
       (bytes(n >> 3) & (1 << (7 - (n & 7)))) != 0
 
     override def foreach[A](f: Boolean => A): Unit = {
-      var i = 0
-      while (i < length) {
-        f(apply(i))
+      var i    = minBitIndex
```
jdegoes, line 1094: This can be much faster if you iterate over the bytes. Then inside the while loop, call `f` 8 times, once for each bit.

Outside the loop you can do "initial" and "trailing" bits, those that don't fall entirely on a byte boundary.

adamgfraser:

PR #3377, opened 2020-04-15
Add Diagnostic Code to Has

adamgfraser:


adamgfraser:
Interesting. The mocking framework appears to fail with this.

@ioleo We're trying to add some diagnostic code to `Has` that throws a runtime exception if the `Has` map does not actually contain the services that are statically known to be contained within it. This helped us diagnose a runtime failure that occurred in ZLayer due to a Scala compiler error. We had thought that this was a property that should always be true and its failure would be a defect, but it appears that adding this is causing widespread failures in the mocking framework. Do you have an understanding of why this would be the case and whether it is an issue in the mocking framework or whether our expectation here is wrong somehow?

jdegoes:
@adamgfraser Looks good to me. There was a type safety issue with mocking casts??

adamgfraser:
@jdegoes I believe some extraneous statically known types were being created internally in the process of building the mock layers, so I believe there was no actual unsoundness in user facing code but it has been fixed now.

PR #3378, opened 2020-04-15
TReentrantLock documentation

calvinlfer:
References https://github.com/zio/zio/issues/2122 in order to complete the TReentrantLock work

### DIFF
--- a/docs/datatypes/treentrantlock.md
+++ b/docs/datatypes/treentrantlock.md
@@ -1,0 +1,150 @@
+---
+id: datatypes_treentrant_lock
+title:  "TReentrantLock"
+---
+
+A TReentrantLock allows safe concurrent access to some mutable state efficiently, allowing multiple fibers to read the 
+state (because that is safe to do) but only one fiber to modify the state (to prevent data corruption). Also, even though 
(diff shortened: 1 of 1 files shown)


calvinlfer:
Sorry, I realized that some of the documentation was not compiling due to additional `commit`s, didn't have type signatures and needed imports so just fixed it

adamgfraser:
Could also be good to provide a high level summary of why you would want to use this. A `TReentrantLock` allows safe concurrent access to some mutable state in an efficient way, allowing multiple fibers to read the state (because that is safe to do) but only one fiber to modify the state (to prevent data corruption). Also even though the `TReentrantLock` is implemented using `STM` the reads and writes can be committed, allowing this to be used as a building block for solutions that expose purely `ZIO` effects and internally allow locking on more than one piece of state in a simple way (because of STM).

calvinlfer:
Nice! I will incorporate this into the documentation

calvinlfer:
Incorporated @adamgfraser :)

jdegoes:
@calvinlfer Awesome work. Thank you!

PR #3391, opened 2020-04-16
Implement ZRefM

### LINKED ISSUE #3350: Support Reference That Writes Values to a Stream
I think we can implement this by giving `RefM` the `ZRef` treatment, which would allow use to add combinators like `mapM`, contramapM`, `tapInput`, and `tapOutput` to `RefM`. Then we could add a constructor for `RefM` that returns a ` RefM` and a `ZQueue` with the reference set to automatically publish changes to the queue. Then in ZIO Stream we could expose a version of this combinator that just lifts the resulting queue into a stream.

I am working on this.

Copying @jdegoes, @iravid, @jeremyrsmith.

### LINKED ISSUE #3373: FiberRefs do not propagate into RefM.modify
`RefM.make` spawns a fiber during construction, so `FiberRef`s do not correctly propagate inside of `RefM.modify(...)`:

```scala
for {
  ref1 <- FiberRef.make(0)
  ref2 <- RefM.make(())
  _ <- ref1.set(1)
  _ <- ref2.modify { _ =>
    ref1.get.flatMap(x => putStrLn(x.toString))
  }
} yield ()
```

I expected it to print `1`, but it actually prints `0`. I don't know what the correct behavior should be, but I think this is a bit counter-intuitive.

adamgfraser:
Resolves #3350. Resolves #3373. 

Changes encoding of `RefM` to be backed by a `Semaphore` instead of a `Queue`. Implements `ZRefM`, a polymorphic version of `RefM` that supports effectual transformations of values being written and read. Uses this functionality to implement `signalRef` combinator that creates a new `RefM` and immediately adds a tap that writes all values written to a `Queue` or a `Stream`.

Copying @jeremyrsmith.


core/shared/src/main/scala/zio/RefM.scala
```diff
@@ -1,234 +1,22 @@
...
-    n: Int = 1000,
-    onDefect: Cause[Any] => UIO[Unit] = _ => IO.unit
-  ): UIO[RefM[A]] =
-    for {
-      ref   <- Ref.make(a)
-      queue <- Queue.bounded[Bundle[_, A, _]](n)
-      _     <- queue.take.flatMap(b => ref.get.flatMap(a => b.run(a, ref, onDefect))).forever.forkDaemon
-    } yield new RefM[A](ref, queue)
-
+  def signalRef[A](a: A): UIO[(RefM[A], Queue[A])] =
```
iravid, line 20: A few suggestions:
- The queue should be a `Dequeue[A]` - not sure if we added this here or in `feature/zconduit`, but it's ZQueue for which the input is Nothing.
- Perhaps we could call this `subscriptionRef`, and have it return something like:
```scala
trait SubscriptionRefM[A] extends RefM[A] { // or ZRef[... Any, A]
  def await: UIO[A] = queue.take
}

def subscriptionRef[A](a: A): UIO[SubscriptionRefM[A]]
```
- For the variant in `ZStream`, we could have another subtype that extends `SubscriptionRefM` with `def changes = ZStream.fromQueue(queue)`.

adamgfraser: @iravid I updated to return a `Dequeue` and to change the name. I actually had that type signature in an earlier version and thought the type signature made it a little intimidating, but `Dequeue` is a great addition!

I don't think it is as straightforward to extend `RefM` because it is a sealed trait and we rely on that for the exhaustiveness of pattern matching in the implementation. Also I would think if we exposed a `SubscriptionRef` trait we would want the different combinators to return a `SubscriptionRef` so would have to override them.

This goes a little bit to whether we think of `Ref` as a data type or an interface. Right now `Ref` and `RefM` are very much data types so it isn't really possible to extend them, but you could potentially see them more as interfaces, potentially with `RefM` and `TRef` extending `Ref`. That could allow you to substitute different `Ref` implementations or extend them. Not sure if that makes sense.

iravid: Right, forgot about that. What do you think about this for the streams module:
```
case class SubscriptionRef(ref: RefM[A], changes: ZStream[A])
```
?

iravid: That can also have FunctorFilter methods later on.

adamgfraser: I think that is super straightforward. Wraps it up into a nice pair. Very easy to use. And the user can easily access the underlying `Ref` if they want to, for example to swap into a section of their code. Makes me wonder if we should do something similar for core.

adamgfraser: Updated to add `SubscriptionRef` and change name of combinator on `RefM` to `dequeueRef`.

streams/shared/src/main/scala/zio/stream/SubscriptionRef.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package zio.stream
+
+import zio.RefM
+
+/**
+ * A `SubscriptionRef[A]` contains a `RefM[A]` and a `Stream` that will emit
+ * every change to the `RefM`.
+ */
+final case class SubscriptionRef[A](ref: RefM[A], changes: Stream[Nothing, A])
```
jdegoes, line 25: Should the constructor be private so we can guarantee it is created in the appropriate way?

adamgfraser: Good idea.

iravid: Might as well also make it a class so it doesn't have `copy`.

adamgfraser: Already done. I'm moving the constructors there as well and adding one that lets you make it in the context of a managed.

streams/shared/src/main/scala/zio/stream/SubscriptionRef.scala
```diff
@@ -16,10 +16,28 @@
...
+  def make[A](a: A): UIO[SubscriptionRef[A]] =
+    RefM.dequeueRef(a).map {
+      case (ref, queue) => new SubscriptionRef(ref, ZStream.fromQueue(queue))
+    }
+
+  /**
+   * Creates a new `SubscriptionRef` with the specified value in the context of
+   * a `Managed`.
+   */
+  def makeManaged[A](a: A): UManaged[SubscriptionRef[A]] =
```
jdegoes, line 41: Personally I would delete this since the subscription ref is not tied to the returned `Managed`, and the user can always call `toManaged_` themselves.

adamgfraser: Done.

adamgfraser: We could also tie them together by shutting down the queue in the `release` action of the `Managed`.

PR #3396, opened 2020-04-17
ZSink.collectAllToMap/Set, ZTransducer.collectAllTo{Set,Map}N

iravid:


streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -276,17 +276,33 @@ object ZSink {
...
+   * using the `f` function.
+   */
+  def collectAllToMap[A, K](key: A => K)(f: (A, A) => A): ZSink[Any, Nothing, A, Map[K, A]] =
+    foldLeftChunks(Map[K, A]()) { (acc, as) =>
+      as.fold(acc) { (acc, a) =>
+        val k = key(a)
+
+        acc.updated(
+          k,
+          // Avoiding `get/getOrElse` here to avoid an Option allocation
```
regiskuckaertz, line 293: Nice 👍

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -94,10 +100,66 @@ object ZTransducer {
...
   def collectAllN[I](n: Long): ZTransducer[Any, Nothing, I, List[I]] =
-    foldUntil[I, List[I]](Nil, n)((list, element) => element :: list).map(_.reverse).filter(_.nonEmpty)
+    foldUntil[I, List[I]](Nil, n)((list, element) => element :: list).map(_.reverse)
+
+  /**
+   * Creates a transducer accumulating incoming values into maps of up to `n` keys. Elements
+   * are mapped to keys using the function `key`; elements mapped to the same key will
+   * be merged with the function `f`.
+   */
+  def collectAllToMapN[K, I](n: Long)(key: I => K)(f: (I, I) => I): ZTransducer[Any, Nothing, I, Map[K, I]] =
```
regiskuckaertz, line 113: Hmm, if `costFn` had access to `O` in `foldWeighted`, we could probably write this in terms of it.

iravid: Yeah. That's what I thought as well. Do you think that'd be a good addition? A `costFn` of the form `(I, O) => Long` could allow for more sophisticated forms of cost calculation.

regiskuckaertz: Yes, definitely. The most powerful version can then power all other combinators. ZIO is full of the same kind of thing.

streams-tests/shared/src/test/scala/zio/stream/experimental/ZTransducerSpec.scala
```diff
@@ -38,7 +38,7 @@ object ZTransducerSpec extends ZIOBaseSpec {
         },
         testM("empty list") {
           val sink = ZTransducer.collectAllN[Int](0)
-          assertM(sink.push.use(_(None)))(equalTo(Chunk.empty))
+          assertM(sink.push.use(_(None)))(equalTo(Chunk(List())))
```
regiskuckaertz, line 41: Ha great, this is proof `fold` works correctly 😇

iravid: Exactly!

streams/shared/src/main/scala/zio/stream/experimental/ZTransducer.scala
```diff
@@ -266,31 +307,40 @@ object ZTransducer {
...
+            val total = state.cost + costFn(state.result, i)
+
             if (total > max) {
               val is = decompose(i)
-              if (is.isEmpty)
-                (os0 + f(state.result, i), initial, false)
-              else if (is.length == 1)
-                (os0 + f(state.result, is(0)), initial, false)
+
+              if (is.length <= 1 && !dirty)
```
regiskuckaertz, line 319: I see, yeah that makes sense, great catch🕵️‍♂️ 💪

PR #3397, opened 2020-04-17
Add ZStream.aggregateAsyncWithin

regiskuckaertz:
Just opening for review now as I fix the timeout. Nice overall win when @iravid realised `aggregateAsync` could be expressed in terms of `aggregateAsyncWithin`, a sign the logic is solid.


regiskuckaertz:
The race condition you discovered, that was the reason why the tests were failing 😅

regiskuckaertz:
(adding `grouped` and `groupedWithin`)

iravid:
Merged 👍🏻

PR #3401, opened 2020-04-18
Consistent naming for ensuring* methods

vitaliis:
The goal of this change is to improve usability of methods performing effects before and after finalizers.
This can be achieved by using consistent naming scheme. 
It is all about convenience of usage and making this functionality easy to discover and use.
- [x] propose a naming scheme
- [ ] check test names and adjust them accordingly
- [ ] sort methods in file (by name)

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
+++ b/core-tests/jvm/src/test/scala/zio/FiberRefSpecJvm.scala
@@ -28,7 +28,7 @@ object FiberRefSpecJvm extends ZIOBaseSpec {
 
         _      <- fiberRef.set(update1)
         thread <- UIO(new Thread(unsafelyGetSetGet))
-        _      <- UIO(thread.start()).ensuring(UIO(thread.join()))
+        _      <- UIO(thread.start()).ensuring_(UIO(thread.join()))
 
         value0                   <- fiberRef.get
         values                   <- UIO(resRef.get())
--- a/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RTSSpec.scala
@@ -37,7 +37,7 @@ object RTSSpec extends ZIOBaseSpec {
         done  <- Ref.make(false)
         start <- IO.succeed(internal.OneShot.make[Unit])
         fiber <- blocking.effectBlockingInterrupt { start.set(()); Thread.sleep(60L * 60L * 1000L) }
-                  .ensuring(done.set(true))
+                  .ensuring_(done.set(true))
                   .fork
         _     <- IO.succeed(start.get())
         res   <- fiber.interrupt
@@ -130,7 +130,7 @@ object RTSSpec extends ZIOBaseSpec {
       def test =
         IO.effect(if (c.incrementAndGet() <= 1) throw new RuntimeException("x"))
           .forever
-          .ensuring(IO.unit)
+          .ensuring_(IO.unit)
           .either
           .forever
 
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -109,7 +109,7 @@ object ScheduleSpec extends ZIOBaseSpec {
         for {
           p          <- Promise.make[Nothing, Unit]
           r          <- Ref.make(0)
-          _          <- r.update(_ + 2).repeat(Schedule.recurs(2)).ensuring(p.succeed(()))
+          _          <- r.update(_ + 2).repeat(Schedule.recurs(2)).ensuring_(p.succeed(()))
           v          <- r.get
           finalizerV <- p.poll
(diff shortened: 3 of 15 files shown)


vitaliis:
There was a problem with associated email address. Hence force push.

jdegoes:
@vitaliis Thank you!

@iravid I like this name change, any objections?

iravid:
I’m generally -1 on renames at this point but I don’t mind entertaining the idea.

Should we also change the existing `ZManaged#ensuring` to `ensuring_` and `ZManaged#onExit` to `ensuring`, for consistency? That means ZManaged would be inconsistent with ZIO and ZStream.

vitaliis:
I believe it improves usability. It will make it easier for people to find those methods as they are using the same names for the concepts people already know. And not having different ones that people have to learn.

> Should we also change the existing ZManaged#ensuring to ensuring_ and ZManaged#onExit to ensuring, for consistency?

I considered this, but decided to start with a small change and see where it goes. I can do this change (onExit => ensuring, ...) also in this MR.

> That means ZManaged would be inconsistent with ZIO and ZStream.

I though about this as well. Something like ensuringBeforeClose or similar for streams. But I am not expert on streams. So please advice, guys, if we need to adjust anything there.

iravid:
> I believe it improves usability. It will make it easier for people to find those methods as they are using the same names for the concepts people already know. And not having different ones that people have to learn.

I'm onboard with that, but I don't think the current changes are a particularly better tradeoff. In particular, following up on the consistency issues (`ZStream#ensuring, ZManaged#onExit`) creates an inconsistency with ZIO.

So to summarize - I am unconvinced by this change.

vitaliis:
> following up on the consistency issues (ZStream#ensuring, ZManaged#onExit) creates an inconsistency with ZIO.

I would like to understand that. It must be possible to work something out. We already have a solution for `ZManaged#onExit`.

vitaliis:
And to bring one more general point. I would be skeptical about such renames after 1.0 is out. 

I believe it is better to pull naming straight before the 1.0 release, when we are still in the RC phase.

vitaliis:
I played with streams on the weekend. And also took a look at ZIO. @iravid , you have a point. With all the finalizers, onExit and ensuring... I will try to work something out.

vitaliis:
Perfect solution IMO would be to use polymorphism and use the same plain name for functions with different parameters. But there is a caveat to it. Completely same name results in the following error at multiple places:
```
/core-tests/shared/src/test/scala/zio/ZIOSpec.scala:1167:34: missing parameter type for expanded function
[error] The argument types of an anonymous function must be fully known. (SLS 8.5)
[error] Expected type was: ?
[error]           _ <- ZIO.unit.ensuring {
```
That is probably the reason we have `.use` and `.use_`.

I also considered: `ensuringWith`, `ensuringOnExit`, `ensuringAs`.

But the cleanest and the most consistent with the current state solution seems to be `.ensuring` and `.ensuring_`.

vitaliis:
@iravid please take a look at the latest version

iravid:
@vitaliis Thank you for your work on this! Appreciate the time and thought you've spent on coming up with a naming scheme.

I'm indifferent to the naming and don't think it provides a clear benefit compared to the disruption of renaming core methods on ZIO at this point, so I will remain 👎 on this change.

vitaliis:
I have no problems closing this MR. I just believe it makes the whole family of ensuing methods easier to discover and use (without having to know that onExit is in fact kind of ensuring and what first exactly means in this context).

It seems that API evolved with time and was done by different people, which is fine for a project with so many parts. I think we can have a better API which is easy to use across data types and has consistent naming. It will result in a reduced cognitive load and complexity by a bit. And that is a good thing, in my opinion.
Changes to the core API, I think they are acceptable before a major version is released and has names (at least for some time) written in stone.

Any other opinions?

adamgfraser:
I'm also somewhat negative on the change. I'm not sure that `onExit` and `ensuring` need to have the same prefix. Obviously they are related in that they both deal with finalizers but there is also a clear difference that one just performs an arbitrary effect whereas the other gives access to the exit value.

In cases like this there is some advantage to a common prefix but there is also some advantage to a descriptive name, which I think we have to a reasonable extent. The current alternative also uses the `_` suffix which I would say is disfavored in general and is a somewhat atypical usage (normally it would denote a variant that returns `Unit` as opposed to a value.

I'm not opposed to changes if we think the new version is substantially better but right now it seems like a slight negative.

vitaliis:
Ok. What actually brought me to this change was `ensuringFirst` on `ZManaged`. It all started by changing it and then unwinding all the related methods. What I still would like to see is the change from `ensuringFirst` to something more meaningful. First, `First` is not even a noun. And, second, it is not immediately clear to which part does it relate. Any objections of changing it to `ensuringBeforeRelease`?

adamgfraser:
I'm slightly negative. It is longer and I'm not sure it is clearer in that I would conceptualize it not as being executed before the release action but as part of the release action with this being the first finalizer executed. But that's just my opinion and happy to defer to others.

PR #3403, opened 2020-04-18
Add AssertionM for effectfull assertions.

senia-psm:
Add `result` field to `AssertionValue` to avoid test reevaluation.
Protect `AssertionValue.assertion` to avoid test reevaluation.

Use `lazy val tryValue = Try(value)` to avoid exception reevaluation:
`lazy val` stores only successful case.

This `PR` reverts #3307 in a sense, but without effects in `Assertion` class. So we can use `Assertion` in pure computations.

### DIFF
--- a/test/shared/src/main/scala/zio/test/FailureDetails.scala
+++ b/test/shared/src/main/scala/zio/test/FailureDetails.scala
@@ -24,7 +24,7 @@ final case class FailureDetails(assertion: ::[AssertionValue], gen: Option[GenFa
   def label(string: String): FailureDetails =
     FailureDetails(
       assertion match {
-        case h :: t => ::(AssertionValue(h.assertion.label(string), h.value), t)
+        case h :: t => ::(h.label(string), t)
       },
       gen
     )
--- a/test-magnolia-tests/shared/src/test/scala/zio/test/magnolia/DeriveGenSpec.scala
+++ b/test-magnolia-tests/shared/src/test/scala/zio/test/magnolia/DeriveGenSpec.scala
@@ -27,17 +27,17 @@ object DeriveGenSpec extends DefaultRunnableSpec {
 
   sealed trait NonEmptyList[+A] { self =>
     def foldLeft[S](s: S)(f: (S, A) => S): S =
-     self match {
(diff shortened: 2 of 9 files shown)


adamgfraser:
@senia-psm Will review tomorrow!

test/shared/src/main/scala/zio/test/FailureDetails.scala
```diff
@@ -19,12 +19,13 @@ package zio.test
 /**
  * `FailureDetails` keeps track of details relevant to failures.
  */
-final case class FailureDetails(assertion: ::[AssertionValue], gen: Option[GenFailureDetails] = None) {
```
adamgfraser, line 22: We're losing the static knowledge here that there is at least one failure inside `FailureDetails`. Can we revert this change?

senia-psm: Done.
IMHO there should be a custom `NonEmptyList` instead of invariant `::`. As I know in scala 3 it's not a enum branch without its own type only due to compatibility reasons.

senia-psm:
> I think there is probably a little more clean up we can do but let's get this in and we can figure that out later. 

@adamgfraser any hints about this cleanup? I might try to do it in a separate review.

test/shared/src/main/scala/zio/test/FailureDetails.scala
adamgfraser: I agree and there will be a ZIO version of one in ZIO Prelude and `::` is actually covariant in Scala 2.13, but we've got to work with the tools we've got here.

adamgfraser:
@senia-psm The main other thing I was thinking is we could alias the `Render` type and companion object in `Assertion` like this:

```scala
object Assertion {
  type Render = AssertionM.Render
  val Render = AssertionM.Render
}
```

That way users who are just using pure assertions don't need to know anything about effectual assertions or add an additional import to create their own rendering in places like `LawsSpec` or `LawsFSpec`.

Maybe we just do it now.

senia-psm:
@adamgfraser done.

PR #3404, opened 2020-04-18
Implement ZIO#mapEffect

### LINKED ISSUE #3402: Safer map alternative
Using methods that can throw exceptions inside of `map` such as `foo.map(_.unsafeOpThatThrows)` will result in a defect. This often bites new users.

Currently there is `foo.flatMap(x => Task(x.unsafeOpThatThrows))` or `foo.map(_.unsafeOpThatThrows).absorb` (the latter isn't _quite_ equivalent because it casts too wide of a net). The idea is to have an alternate like `mapSafe` (or `mapAbsorb` or whatever a good name for it would be).

Just the fact that it would come up in autocomplete as you type map could remind you, "Wait, is my map call safe?"

This safe map would need evidence `E =:= Throwable`.

adamgfraser:
Resolves #3402.


adamgfraser:
Added a `mapPartial`variant for STM.

PR #3406, opened 2020-04-18
Port ZSink#zipPar

simpadjo:


streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -253,6 +253,111 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
 
+  /**
+   * Runs both sinks in parallel on the input and combines the results
+   * using the provided function.
+   */
+  final def zipWithPar[R1 <: R, E1 >: E, I1 <: I, Z1, Z2](
+    that: ZSink[R1, E1, I1, Z1]
+  )(f: (Z, Z1) => Z2): ZSink[R1, E1, I1, Z2] = {
+
+    type State = (Option[Z], Option[Z1])
```
iravid, line 264: Change this to:
```
sealed trait State
case object BothRunning extends State
case class LeftDone(z: Z) extends State
case class RightDone(z: Z1) extends State
```

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -253,6 +253,111 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
+
+    def applyLeft(s: State, res: Either[Either[E, Z], Unit]): ZIO[R1, Either[E1, Z2], State] =
```
iravid, line 266: You can call the `applyLeft/Right` functions only when `Either[E, Z/Z2]` is received to save the outer either.

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -253,6 +253,111 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
+      }
+
+    ZSink(for {
+      ref <- ZRef.make[State]((None, None)).toManaged_
+      p1  <- self.push
+      p2  <- that.push
+      push: Push[R1, E1, I1, Z2] = {
+        in =>
+          ref.get.flatMap { state =>
+            val newState: ZIO[R1, Either[E1, Z2], State] = {
```
iravid, line 301: This will be simpler with the `State` trait I suggested above.

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -253,6 +253,112 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
+      ref <- ZRef.make[State](BothRunning).toManaged_
+      p1  <- self.push
+      p2  <- that.push
+      push: Push[R1, E1, I1, Z2] = {
+        in =>
+          ref.get.flatMap {
+            state =>
+              val newState: ZIO[R1, Either[E1, Z2], State] = {
+                state match {
+                  case BothRunning => {
```
iravid, line 279: I have a hunch this case could be simplified, but not super important.

PR #3408, opened 2020-04-19
Add rightOrFailWith and leftOrFailWith and error transform docs

Fristi:
I had some functions already which were doing this already, would be nice if we can lift these boxes inside a ZIO effect without too much folding and such :)

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -994,6 +994,14 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(UIO(Right(12)).leftOrFail(ExampleError).flip)(equalTo(ExampleError))
       } @@ zioTag(errors)
     ),
+    suite("leftOrFailWith")(
+      testM("on Left value") {
+        assertM(Task(Left(42)).leftOrFailWith((x: Throwable) => x))(equalTo(42))
+      },
+      testM("on Right value") {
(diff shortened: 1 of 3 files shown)


adamgfraser:
Thanks for contributing!

I think several of these are very similar to existing combinators. In particular:

- `fromEitherM(v)` <-> `v.absolve`
- `ofEither(v, ifLeft)` <-> `fromEither(v).mapError(ifLeft)`
- `ofEitherM(v, ifLeft)` <-> `v.flatMap(fromEither(_).mapError(ifLeft))`
- `ofOption(v, ifEmpty)` <-> `fromOption(v).orElseFail(ifEmpty)`
- `ofOptionM(v, ifEmpty)` <-> `v.someOrFail(ifEmpty)`

I think most of these are pretty straightforward and at most take two method calls when you essentially want to do two things: lifting a value into an effect an transform the error value. The one that is not quite as nice is the replacement for your `ofEitherM` method. `rightOrFail` almost does what we need but not quite because it just lets us specify a new error value and not transform the existing error value. Maybe we add a `rightOrFailWith` variant for that?

adamgfraser:
@Fristi Are you interested in working on the `leftOrFailWith` and `rightOrFailWith` variants?

Fristi:
Yes, let's see if I can squeeze some time on it this week

adamgfraser:
Sounds good. Let me know if there is anything I can do to help.

Fristi:
I added those combinators and some helpful documentation I think :) Let me know what you think

Fristi:
Great! Looking forward to have it in the next release

adamgfraser:
@Fristi Thanks for your contribution on this!

PR #3409, opened 2020-04-19
Add some current time variants

Fristi:
I had some functions already which were doing this already, would be nice if we can have these in the core package

### DIFF
--- a/core/shared/src/main/scala/zio/clock/package.scala
+++ b/core/shared/src/main/scala/zio/clock/package.scala
@@ -16,7 +16,7 @@
 
 package zio
 
-import java.time.{ DateTimeException, Instant, OffsetDateTime, ZoneId }
+import java.time.{ DateTimeException, Instant, LocalDate, LocalTime, OffsetDateTime, ZoneId, ZonedDateTime }
 import java.util.concurrent.TimeUnit
 
 import zio.duration.Duration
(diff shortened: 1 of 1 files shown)


adamgfraser:
@Fristi Just wanted to check in on whether you are still interested in working on this. If so is there anything I can do to help?

Fristi:
Yes, I'll check the `envOrElseWith` pattern and see if I can come with up something this week

adamgfraser:
Just so you don't have to implement the same logic twice in the `Clock` and the `TestClock` since one you had it to the service anything that extends the service now needs to implement those methods.

Fristi:
That would be smart! However `TestClock` is not providing any means to get the currentTime right? How would you apply it to `TestClock`?  `TestClock` is only for time travelling it seems :)

adamgfraser:
No if you look at the [implementation](https://github.com/zio/zio/blob/e27a7ee701a7e8ba3f34cbf8f08742179a3c9296/test/shared/src/main/scala/zio/test/environment/package.scala#L273) it extends `Clock.Service` with `TestClock.Service`.

fsvehla:
I think that the variants that return `Instant` can be unexceptional if they are not implemented in terms of `currentDateTime`

Fristi:
I think it's a good suggestion, but would `Clock` be still the right place to supply these functions?

jdegoes:
@Fristi Great idea! Some of these were already added. However, we welcome more following the same pattern as established in the existing code base (e.g. naming the methods after the corresponding `java.time` method). Now that we depend heavily on java.time, there's essentially no reason to have all the variants. Please re-open if you get a chance to work on this again!

PR #3414, opened 2020-04-19
Add alternative combinator to ZSTM

mijicd:
### Problem

The current `orElse` implementation does not retry left once provided alternative retries but stays "stuck" retrying only the alternative. 

Since our STM implementation is inspired by the one in Haskell, I made a small program to double-check the expected behavior.

> The example is taken from a fantastic STM workshop by Michael Snoyman, which you can check out [here](https://github.com/snoyberg/why-you-should-use-stm).

```haskell
#!/usr/bin/env stack
-- stack --resolver lts-13.21 script
import Control.Applicative (Alternative (..))
import Control.Concurrent (threadDelay)
import Control.Concurrent.Async
import Control.Concurrent.STM
import Control.Exception
import Control.Monad (forever)

data Boom = Boom deriving Show

instance Exception Boom

-- | Add 5 to the variable every 100ms
payAlice :: TVar Int -> IO ()
payAlice aliceVar = forever $ do
  atomically $ do
    alice <- readTVar aliceVar
    writeTVar aliceVar $! alice + 5
  threadDelay 100000 -- 100ms, threadDelay takes nanoseconds

-- | Transfer 40 from Alice to Bob.
transfer
  :: TVar Int -- ^ Alice
  -> TVar Int -- ^ Bob
  -> IO ()
transfer aliceVar bobVar = atomically $ do
  let amount = 40
  alice <- readTVar aliceVar

  -- each of the following variants fail the entire transaction
  -- throwSTM Boom <|> check (alice >= amount)
  -- check (alice >= amount) <|> throwSTM Boom

  -- retries both (2nd retry is obviously redundant but it was failing test for ZSTM)
  check (alice >= amount) <|> retry

  writeTVar aliceVar $ alice - amount
  bob <- readTVar bobVar
  writeTVar bobVar $ bob + amount

main :: IO ()
main = do
  aliceVar <- newTVarIO 0
  bobVar <- newTVarIO 0
  race_ (payAlice aliceVar) (transfer aliceVar bobVar)
  alice <- atomically $ readTVar aliceVar
  bob <- atomically $ readTVar bobVar
  putStrLn $ "Alice has: " ++ show alice
  putStrLn $ "Bob has: " ++ show bob
```

Experimenting with alternative combinations, I've noticed that our implementation "swallows" failures.


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -545,14 +545,14 @@ final class ZSTM[-R, +E, +A] private[stm] (
   /**
    * Named alias for `<>`.
    */
-  def orElse[R1 <: R, E1, A1 >: A](that: => ZSTM[R1, E1, A1]): ZSTM[R1, E1, A1] =
+  def orElse[R1 <: R, E1 >: E, A1 >: A](that: => ZSTM[R1, E1, A1]): ZSTM[R1, E1, A1] =
```
jdegoes, line 548: We have a problem with this: ZIO#orElse will try the right if the left fails. That's the "error swallowing" behavior you mention.

Maybe what this is saying is we need two combinators: one for doing "retry" fallback, and one for doing "error" fallback. What we used to have was one combinator for doing both. Or maybe doing two things at once is sufficiently common to warrant conflating them?

mijicd: This is something I briefly discussed with Adam. The current behavior of `orElse` (alternative) is wrong according to the cited paper and Haskell's implementation. The only thing I wouldn't do is leave it incorrect. Other than that, we are going to be asymmetric with something.

mijicd:
@jdegoes @adamgfraser As discussed over discord, I've reverted the old `orElse` behavior, and introduced another combinator called `alternative` (with symbolic alias `<|>`) that will have Haskell's semantic.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -154,6 +161,37 @@ final class ZSTM[-R, +E, +A] private[stm] (
   def absolve[E1 >: E, B](implicit ev: A <:< Either[E1, B]): ZSTM[R, E1, B] =
     ZSTM.absolve(self.map(ev))
 
+  /**
+   * Named alias for `<|>`.
+   */
+  def alternative[R1 <: R, E1 >: E, A1 >: A](that: => ZSTM[R1, E1, A1]): ZSTM[R1, E1, A1] =
```
jdegoes, line 167:
```suggestion
  def orTry[R1 <: R, E1 >: E, A1 >: A](that: => ZSTM[R1, E1, A1]): ZSTM[R1, E1, A1] =
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -603,7 +641,7 @@ final class ZSTM[-R, +E, +A] private[stm] (
    * Tries this effect first, and if it fails or retries, succeeds with the specified
    * value.
    */
-  def orElseSucceed[A1 >: A](a1: => A1): URSTM[R, A1] =
+  def orElseSucceed[A1 >: A](a1: => A1): ZSTM[R, E, A1] =
```
jdegoes, line 644: How did `E` sneak in here?

jdegoes:
```suggestion
  def orElseSucceed[A1 >: A](a1: => A1): URSTM[R, A1] =
```

mijicd: Copy paste 101.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -128,6 +128,13 @@ final class ZSTM[-R, +E, +A] private[stm] (
   def <>[R1 <: R, E1, A1 >: A](that: => ZSTM[R1, E1, A1]): ZSTM[R1, E1, A1] =
     orElse(that)
 
+  /**
+   * Tries this effect first, and if it fails, tries the other effect. This is
```
jdegoes, line 132:
```suggestion
   * Tries this effect first, and if it enters retry, then it tries the other effect. This is
```

docs/datatypes/stm.md
```diff
@@ -107,17 +107,17 @@ There are many other variants of the `STM.retry` combinator like `STM.check` so
...
 
 ```scala mdoc:silent
 def transferMoneyFailFast(from: TRef[Long], to: TRef[Long], amount: Long): STM[String, Long] =
-    transferMoneyNoMatterWhat(from, to, amount) orElse STM.fail("Sender does not have enough of money")
+    transferMoneyNoMatterWhat(from, to, amount) orTry STM.fail("Sender does not have enough of money")
 ```
 
-This will cause the transfer to fail immediately if the sender does not have money because of the semantics of `orElse`.
\ No newline at end of file
+This will cause the transfer to fail immediately if the sender does not have money because of the semantics of `orElse`.
```
jdegoes, line 123: This one still mentions `orElse` instead of `orRetry`

mijicd: Fixed.

PR #3416, opened 2020-04-20
Implement Fast Single Element Append for Chunk

### LINKED ISSUE #3371: Implement fast single-element append for Chunk
A prototype of the proposed approach:

```scala
sealed trait Chunk[+A] { self =>
  def + [A1 >: A](a1: A1): Chunk[A1]
}
object Chunk {
  val BufferSize = 100

  final class Arr[A](val array: Array[A]) extends Chunk[A] { self =>
    def + [A1 >: A](a1: A1): Chunk[A1] = {
      val buffer = Array.ofDim[AnyRef](BufferSize)

      buffer(0) = a1.asInstanceOf[AnyRef]
      
      new Chunk.Add[A1](self, buffer, 1, new AtomicInteger(1))
    }
  }
  final class Concat[A](left: Chunk[A], right: Chunk[A]) extends Chunk[A] {
    def + [A1 >: A](a1: A1): Chunk[A1] = new Concat(left, right + a1)
  }
  final class Add[A](start: Chunk[A], buffer: Array[AnyRef], bufferUsed: Int, chain: AtomicInteger) extends Chunk[A] { self =>
    def + [A1 >: A](a1: A1): Chunk[A1] = {
      if ((bufferUsed < buffer.length) && chain.compareAndSet(bufferUsed, bufferUsed + 1)) {
        buffer(bufferUsed) = a1.asInstanceOf[AnyRef]

        new Chunk.Add[A1](start, buffer, bufferUsed + 1, chain)  
      } else {
        // Cannot reuse the buffer, because someone else used it or it's full.
        val buffer = Array.ofDim[AnyRef](BufferSize)

        buffer(0) = a1.asInstanceOf[AnyRef]
      
        new Chunk.Add[A1](self, buffer, 1, new AtomicInteger(1))
      }
      
    }
  }

  def apply[A: ClassTag](as: A*): Chunk[A] = 
    new Arr(as.toArray)
  
  def empty: Chunk[Nothing] = new Arr[Nothing](Array.empty[Nothing])
}

object Main {
  def benchmark[A](name: String, iterations: Long = 10000, ops: Long = 10000)(a: A)(f: A => A): Unit = {
    System.gc()

    val start = System.currentTimeMillis()

    var j = 0 
    while (j < iterations) {
      var i = 0 
      var currentA = a 
      while (i < ops) {
        currentA = f(currentA)

        i = i + 1 
      }
      j = j + 1
    }

    val end = System.currentTimeMillis()

    println(name + ": " + iterations.toDouble * ops.toDouble / ((end - start).toDouble / 1000.0) + " ops/second")
  }

  def main(args: Array[String]): Unit = {
    benchmark("vector")(Vector(1))(_ :+ 2)

    benchmark("chunk")(Chunk(1))(_ + 2)
  }
}
```

There may be bugs in the approach (needs testing) but currently benchmarking at 3x the performance of Vector's add.

/cc @ahoy-jon

adamgfraser:
Initial work to resolve #3371. I am also seeing dramatically faster performance than `Vector`.


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -873,6 +875,36 @@ object Chunk {
...
+        buffer(0) = a1.asInstanceOf[AnyRef]
+        Add(start, buffer, bufferUsed, chain)
+      }
+
+    def apply(n: Int): A =
+      if (n < start.length) start(n) else buffer(n - start.length).asInstanceOf[A]
+
+    protected[zio] def toArray[A1 >: A](n: Int, dest: Array[A1]): Unit = {
+      start.toArray(n, dest)
+      val _ = buffer.asInstanceOf[Array[A]].copyToArray(dest, n + start.length, bufferUsed)
```
mijicd, line 904: Doesn't `copyToArray` return `Unit`?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -873,6 +875,36 @@ object Chunk {
       case _: BitChunk       => ClassTag.Boolean.asInstanceOf[ClassTag[A]]
     }
 
+  private val bufferSize: Int =
```
mijicd, line 878: Please rename to `BufferSize`.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: On Scala 2.13 it returns an integer with the number of elements written to the array.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Done.

benchmarks/src/main/scala/zio/chunks/ChunkAddBenchmark.scala
```diff
@@ -0,0 +1,50 @@
...
+  var chunk: Chunk[Int]   = _
+  var vector: Vector[Int] = _
+
+  @Param(Array("10000"))
+  var size: Int = _
+
+  @Setup(Level.Trial)
+  def setup() =
+    chunk = Chunk(1)
+  vector = Vector(1)
```
mijicd, line 23: You can even drop the setup and initialize them above.

core/shared/src/main/scala/zio/Chunk.scala
mijicd: Ah, I see, my bad.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -36,7 +37,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   /**
    * Appends an element to the chunk
    */
-  final def +[A1 >: A](a: A1): Chunk[A1] =
+  def +[A1 >: A](a: A1): Chunk[A1] =
```
jdegoes, line 40: Should we call this `:+` for consistency with `Vector`?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -873,6 +875,39 @@ object Chunk {
       case _: BitChunk       => ClassTag.Boolean.asInstanceOf[ClassTag[A]]
     }
 
+  /**
+   * The maximum number of elements in the buffer for fast append.
+   */
+  private val BufferSize: Int =
```
jdegoes, line 881: This one is going to be tricky. I'm not sure the right number to choose. I'd think somewhere between `32` and `128`. Higher numbers equal faster performance for repeated adds, while lower numbers yield higher waste for infrequent adds.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -873,6 +875,39 @@ object Chunk {
...
+    100
+
+  private final case class Add[A](start: Chunk[A], buffer: Array[AnyRef], bufferUsed: Int, chain: AtomicInteger)
```
jdegoes, line 884: Maybe our `buffer` should be `ChunkBuilder` since we are trying to make that smart and efficient? We'd need a way to check `bufferUsed` and capacity, and we'd need to use `sizeHint` (and make sure it's working properly).

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -57,6 +57,53 @@ object ChunkSpec extends ZIOBaseSpec {
         assert(chunk.size)(equalTo(chunk.length))
       }
     ),
+    suite("add")(
```
jdegoes, line 60: We should probably test `Chunk.empty + 2` (given that empty is an edge case in Chunk).

core/shared/src/main/scala/zio/Chunk.scala
jdegoes, line 884: Maybe good to call this `AppendN`, to pave the way for `PrependN` (Vector `+:`). Then we can support efficient prepends and appends.

benchmarks/src/main/scala/zio/chunks/ChunkAddBenchmark.scala
adamgfraser: Done.

adamgfraser:
Benchmarks:

```scala
[info] Benchmark                     (size)   Mode  Cnt     Score     Error  Units
[info] ChunkAddBenchmarks.chunkAdd    10000  thrpt   25  2400.011 ± 128.596  ops/s
[info] ChunkAddBenchmarks.vectorAdd   10000  thrpt   25  1144.855 ±  32.112  ops/s
```

So more than twice as fast as `Vector` for repeated appends.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Yes I think so. That will also set us up to do the same thing for prepend.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Yes let me work on that.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
adamgfraser: The `chunkOf` generator generates a good number of empty chunks so I think we are getting quite a bit of coverage for this but can add another test.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Maybe we start with 64 and we can always adjust? I think we probably need some aggregate benchmark for a varied set of tasks to really pick the "right" number here and may also make sense to get in some of the other improvements we're making before doing that degree of optimization.

core-tests/shared/src/test/scala/zio/ChunkSpec.scala
jdegoes: Ah, no need then! Just was at the top of my mind.

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: Ok, awesome. Please just double check `ChunkBuilder` doesn't lead to a performance regression. Wouldn't want to lose these awesome numbers!

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: Makes sense to me! 👍

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: Will do!

adamgfraser: Thinking about this more I'm not sure this works. We need to be able to support actual collection operations on `AppendN` chunks which could require looking up values that are in the buffer. The `Builder` interface doesn't support that, that is the difference between the builder of a collection and a collection. We could potentially keep track of the size of the builder but if we also want to allow looking up the elements of the builder I think we are just creating another collection and at that point we might as well stick with an array.

jdegoes: You're right. I mean we'd need access to the underlying array with a guarantee that `clear` isn't called... and even then, that would only work up until we specialize bit chunks.

The main thing we miss with the current approach is primitives (which we always box right now, even if unnecessary).

adamgfraser: We could specialize this.

core/shared/src/main/scala/zio/Chunk.scala
adamgfraser: One other thing we could do here is have some heuristic for the buffer size. Like we could start out with a buffer size of `1` and double the buffer size every time something is "too big" so we are learning and optimizing for how the user is working with the chunk.

jdegoes: My feeling is that it will not perform that well compared to just choosing a static number.

I remember some research showing the vast majority of collections are "small collections". The constant doubling (from 1 to 2, 2 to 4, etc.) means a lot of copying on your way to a small size.

adamgfraser: Yes. We could also start at an intermediate size like say `8` and double from there but also fine as is and can always optimize later.

jdegoes:
@adamgfraser Good to merge when conflicts are fixed! Excellent work.

adamgfraser:
@jdegoes This is ready to go.

adamgfraser:
Shall we merge this before the next release?

jdegoes:
@adamgfraser Good to merge when conflicts are fixed!

adamgfraser:
@jdegoes All set.

PR #3419, opened 2020-04-20
Back up Chunk[Boolean] by Chunk.BitChunks 

whysoserious:


core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
```diff
@@ -79,28 +81,52 @@ private[zio] object ChunkBuilder {
...
-      Chunk.fromArray(arrayBuilder.result())
+    def clear(): Unit = {
+      bytes.clear()
+      maxBitIndex = 0
+    }
+    def result(): Chunk[SBoolean] = {
+      val chunk = Chunk.fromArray(bytes.toArray)
+      BitChunk(chunk, 0, maxBitIndex)
+    }
     override def ++=(as: TraversableOnce[SBoolean]): this.type = {
```
adamgfraser, line 116: I don't think this implementation is safe because we only know that it is safe to traverse `as` once but we are traversing it twice right now (once for `as.size` and once for `as.foreach`). We may just need to delegate to doing `+=` with each `Boolean` here. Unfortunately if we don't have the ability to traverse more than once there isn't a whole lot else we can do than process the elements one at a time unless we want to materialize them into their own collection which is kind of repeating the world of the `ChunkBuilder`.

core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
```diff
@@ -79,28 +81,52 @@ private[zio] object ChunkBuilder {
    * values.
    */
   final class Boolean extends ChunkBuilder[SBoolean] { self =>
-    private val arrayBuilder: ArrayBuilder[SBoolean] = {
-      new ArrayBuilder.ofBoolean
+
+    private val bytes: ArrayBuffer[SByte] = ArrayBuffer()
```
adamgfraser, line 85: It looks like `ArrayBuffer` is not specialized so this will box primitive values. Can we instead use an `ArrayBuilder.ofByte` or even a `ChunkBuilder.Byte`? I think this would require maintaining a `var` with the last "unfinished" byte that we manipulate until we are done and add to the `ChunkBuilder` either when it is done or when `result` is called.

whysoserious:
👀

core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
whysoserious: > 
> 
> It would also be good to add some tests here, for example that when we use the new `ChunkBuilder` to build a chunk from a list of random booleans and turn it back into a list that we get back the list unchanged.

It looks like we already have such test:
https://github.com/whysoserious/zio/blob/master/core-tests/shared/src/test/scala/zio/ChunkBuilderSpec.scala#L17-L23

Btw, that's great that we are using property based tests in the project. They already helped me to catch a few subtle bugs in my impl.

whysoserious:
It looks like we already have such test:
https://github.com/whysoserious/zio/blob/master/core-tests/shared/src/test/scala/zio/ChunkBuilderSpec.scala#L17-L23

Btw, that's great that we are using property based tests in the project. They already helped me to catch a few subtle bugs in my impl.

whysoserious:
>It would also be good to add some tests here, for example that when we use the new ChunkBuilder to build a chunk from a list of random booleans and turn it back into a list that we get back the list unchanged.

It looks like we already have such test:
https://github.com/whysoserious/zio/blob/master/core-tests/shared/src/test/scala/zio/ChunkBuilderSpec.scala#L17-L23

Btw, that's great that we are using property based tests in the project. They already helped me to catch a few subtle bugs in my impl.

adamgfraser:
Ah, good point.

whysoserious:
@adamgfraser  it's ready for re-review

adamgfraser:
Thanks for your work on this and congratulations on your first contribution to ZIO!

PR #3420, opened 2020-04-20
ZIO Test: Support Laws for Capabilities Parameterized on Two Types

adamgfraser:
This adds support for laws describing capabilities parameterized on two types, for example `Equivalence[A, B]` as opposed to `Equal[A]`. This is done by implementing a new `ZLaws2` trait that is parameterized on both a capability that takes two type parameters, such as `Equivalence`, as well as possibly capabilities that the left or right value must have.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/laws/Laws2Spec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/laws/Laws2Spec.scala
@@ -4,6 +4,8 @@ import zio.test._
 
 object Laws2Spec extends ZIOBaseSpec {
 
+  type AnyF[_] = Any
+
   def equalTo[A: Equal](expected: A): Assertion[A] =
     Assertion.assertion("equalTo")(Assertion.Render.param(expected))(_ === expected)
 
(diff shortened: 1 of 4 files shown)


test-tests/shared/src/test/scala/zio/test/laws/Laws2Spec.scala
```diff
@@ -0,0 +1,74 @@
...
+  implicit class EqualSyntax[A](private val self: A) extends AnyVal {
+    def ===(that: A)(implicit eq: Equal[A]): Boolean =
+      eq.equal(self, that)
+  }
+
+  final case class Equivalence[A, B](to: A => B, from: B => A)
+
+  object Equivalence extends Lawful2[Equivalence, Equal, Equal] {
+
+    val leftIdentity = new Laws2.Law1Left[Equivalence, Equal, Equal]("leftIdentity") {
```
adamgfraser, line 38: @jdegoes We are requiring a little more than we need to here, since technically the left law only needs an `Equal` instance for the left type and vice versa for the right law. Conceptually it would be nice to have something like the identity or the empty capability that provides no functionality and is always available. This would be the equivalent of `Any` for the environment except that `Any` is of the wrong kind.

I tried something like:

```scala
trait Identity[A]

object Identity {
  implicit def identity[A]: Identity[A] =
    new Identity[A] {}
}
```

But the problem is Scala can't unify that with other capabilities so it ends up leaking into the type signatures which is not ideal. Is there a way to express that? The type `AnyF[_]` such that for all `F[_]`, `F[x] <: AnyF[x]`?

jdegoes:
```scala
type AnyF[_] = Any
```

adamgfraser: That was easy! 😆

jdegoes:
I think we still have to decide if specific laws like `Left` and `Right` will be useful generally, or if they will only be useful in the context of a downstream library.

jdegoes:
Great work and looking forward to testing this out in Prelude!

PR #3422, opened 2020-04-21
Implement ZIO#service

adamgfraser:
Following up on good suggestion from @ghostdogpr implements `getService` which is like `environment` but does the `get` for you. So if you wanted the `Clock.Service` you could do `ZIO.getService[Clock.Service]` and get a `ZIO[Clock, Nothing, Clock.Service]`. Still needs variants for other type aliases and other ZIO effect types assuming we like this.

### DIFF
--- a/core/shared/src/main/scala/zio/RIO.scala
+++ b/core/shared/src/main/scala/zio/RIO.scala
@@ -506,6 +506,31 @@ object RIO {
   final def getOrFail[A](v: => Option[A]): Task[A] = ZIO.getOrFail(v)
 
   /**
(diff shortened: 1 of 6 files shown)


hmemcpy:
Not a huge fan of the get* prefix, but overall +1 :)

adamgfraser:
@hmemcpy You would prefer just `service`?

hmemcpy:
I think it would be more consistent, yes.

adamgfraser:
I'm not opposed to it. I think the original line of thinking was starting with something like `accessService` and then shortening it to `getService`, but I can see the symmetry of `service` with `environment`.

PR #3424, opened 2020-04-21
ZIO Test: Make TestClock Run Effects In Order

adamgfraser:
Implements `TestClock#tick`, which runs the first scheduled effect, semantically blocking until the effect completes execution or is suspended. It also semantically blocks until there is at least one scheduled effect. With this, tests like the following work:

```scala
testM("tick") {
  for {
    ref    <- Ref.make(false)
    _      <- ref.set(true).delay(1.second).fork
    _      <- TestClock.tick
    result <- ref.get
  } yield assert(result)(isTrue)
} @@ nonFlaky,
```

Notice that no external coordination is needed here. I think that with this we can reimplement `adjust` as essentially "tick until there are no more scheduled effects that should run by this time, and then advance the time further if necessary to the user specified time". This should allow us to support the various use cases of adjusting the time with a single combinator and simplify the API. Still need to refactor existing combinators to use this and verify their correctness but wanted to share current direction.


adamgfraser:
To summarize the changes here:

**Setting the Time Runs Effects in Order**

`adjust` and `setTime` no long just run everything that is scheduled to occur on or before the specified time. Instead they step through the queue of scheduled effects, reactivating the first one, allowing it to run until it completes or is suspended again, and then running the next scheduled effect until there are no more effects scheduled to run on or before the new clock time. In each step the clock time is adjusted to the time that effect was scheduled to run until after the last one the clock is adjusted to the user specified time, if greater. 

This should significantly reduce the need to external coordination in verifying conditions after adjusting the time since you will now have a guarantee that scheduled effects have been performed after calling `adjust` without having to use a promise or a queue.

**Getting the Time Returns the Clock Time**

`fiberTime` is now exclusively used for tracking how much time has passed on a fiber and determining when a suspended effect should be scheduled to be resumed. All methods that get the time now return the clock time so setting the time and then immediately getting the time will always return the set value. More complex schedules that access the time still work correctly because of the way we step through the scheduled effects when resuming them described above. `adjust` has been deleted.`

**Tick Added**

`tick` combinator added to run the next scheduled effect. This inspired these changes but it is actually not necessary at this point given the changes to adjusting the time. Could be useful in some situations if you want to run the next effect and don't know when it was scheduled for but I think we could also remove if we wanted. We did not have before and I think the demand for it really grew out of some of the deficiencies with the prior way of adjusting the time that have now hopefully been addressed.

**AwaitScheduled Added**

Because we have these combinators that step through the scheduled effects, it can be important to make sure that an effect has been scheduled to kick off the process, and there can be a race condition here if we fork an effect and then immediately try to adjust the time. Adds new combinators `awaitScheduled` and `awaitScheduledN` to semantically block until one or more effects have been placed in the queue. This can be useful in some situations immediately before adjusting the time.

adamgfraser:
@fanf @runtologist You have been two of the heavier uses of the `TestClock` functionality. If you have a chance would appreciate if you could try out this version and let me know what you think.

runtologist:
Awesome. I'll take a look. That might take until early next week.

jdegoes:
This sounds too good to be true! Any loopholes or edge cases you can think of?

 I think you're right that we don't need `tick` at this point.

adamgfraser:
The only one I can think of is that we are relying on the accuracy of the blocking on information when we introspect the fiber graph. If the blocking on information is not accurate we could conclude that a fiber is indefinitely suspended, and thus resume the fiber executing the next effect, when there is actually some other fiber running that will cause the first fiber to resume.

I haven't observed this so far and I think it is a good forcing function for us to capture this information accurately, but that is the potential issue I can think of.

One other idea is that we could potentially use functionality like this to warn the user about deadlocks in normal code. For example, if the run loop has not executed an instruction within a specified period we could introspect on the fiber graph, determine if it is indefinitely suspended, and then issue a warning.

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -335,83 +338,156 @@ package object environment extends PlatformSpecific {
...
+       * but is not advancing the `TestClock`.
+       */
       private[TestClock] val warningDone: UIO[Unit] =
         warningState.updateSome[Any, Nothing] {
           case WarningData.Start          => ZIO.succeedNow(WarningData.done)
           case WarningData.Pending(fiber) => fiber.interrupt.as(WarningData.done)
         }
 
+      /**
+       * Polls untl the specified fiber completes execution or is indefinitely
```
runtologist, line 420:
```suggestion
       * Polls until the specified fiber completes execution or is indefinitely
```

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -335,83 +338,156 @@ package object environment extends PlatformSpecific {
...
+            case Some(sleep) if sleep.duration <= duration =>
+              clockState.update(_.copy(duration = sleep.duration)) *>
+                scheduled.take.map(_.copy(duration = duration)).asSome
+            case _ =>
+              clockState.update(_.copy(duration = duration)) *> STM.none
+          }
+        }.commit.flatMap {
+          case None => UIO.unit
+          case Some(sleep) =>
+            sleep.promise.succeed(()) *>
```
runtologist, line 453: Would it make sense to `ZIO.yieldNow` here? In most cases the `awaitSuspend` will return `false` the first time, as the sleeping Fiber will not have finished. On a single threaded environment that will happen every time.

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -335,83 +338,156 @@ package object environment extends PlatformSpecific {
...
         }
 
+      /**
+       * Polls untl the specified fiber completes execution or is indefinitely
+       * suspended, meaning that the fiber and all the fibers it is blocking on
+       * and all the fibers they are blocking on recursively are suspended.
+       */
+      private def awaitSuspended(fiberId: Fiber.Id): UIO[Unit] =
+        live.provide {
+          suspended(fiberId).repeat {
```
runtologist, line 426: As `Fiber.get` can be a rather expensive operation if we have many Fibers, would it make sense to split this logic into a first step, which just gets the Fiber by id and a second which checks for suspension? Suspended could be modified to take a Fiber instead of just an Id.

jdegoes:
> The only one I can think of is that we are relying on the accuracy of the blocking on information when we introspect the fiber graph. If the blocking on information is not accurate we could conclude that a fiber is indefinitely suspended, and thus resume the fiber executing the next effect, when there is actually some other fiber running that will cause the first fiber to resume.

The `blockingOn` is basically a hint. It's built into FiberContext, and that information should be accurate because it's only used when one fiber waits on another (for interrupt/join/await).

Then separately, it's used in Promise, and there, it's a "best guess": the code assumes (by default, anyway), that the code which created the promise will complete the promise. That's generally accurate, but it doesn't have to be so.

In worst case, the fiber id could be wrong (because a promise is handed elsewhere); or some code could use an async constructor directly for fiber-to-fiber communication, and fail to specify the blocking on hint.

Ah, we should make sure STM uses it properly too...

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -757,11 +761,25 @@ object Fiber extends FiberPlatformSpecific {
...
-    else false
+  private[zio] def get(fiberId: Fiber.Id): UIO[Option[Fiber.Runtime[Any, Any]]] = {
+
+    def loop(fiber: Fiber.Runtime[Any, Any]): UIO[Option[Fiber.Runtime[Any, Any]]] =
+      fiber.id.flatMap { id =>
+        if (id == fiberId) UIO.succeedNow(Some(fiber))
+        else fiber.children.flatMap(ZIO.collectFirst(_)(loop))
+      }
+
+    roots.flatMap(ZIO.collectFirst(_)(loop))
```
jdegoes, line 775: What happens if there are a million fibers? Or is the idea this is only intended to be used in a test with a small number of fibers?

adamgfraser: I think most of the time there are only a small number of effects in a test that is using time but yes there is some inefficiency here. One of the things I was wondering about is whether we could create a weak map from fiber IDs to fibers when we create them so we could just look up the fiber ID in the map instead of having to do a search for it. But there is potentially a performance implication from that. We could also potentially memoize the results so only have to search for a given fiber once.

test/shared/src/main/scala/zio/test/environment/package.scala
adamgfraser: Yes I think some kind of memoization of the fibers we get definitely makes sense if we can't just look them up directly.

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -335,83 +338,156 @@ package object environment extends PlatformSpecific {
...
+       */
+      private def suspended(fiberId: Fiber.Id): UIO[Boolean] = {
+
+        def loop(fiberId: Fiber.Id, suspended: Set[Fiber.Id]): UIO[Boolean] =
+          Fiber.get(fiberId).flatMap {
+            case None => UIO.succeedNow(true)
+            case Some(fiber) =>
+              fiber.status.flatMap {
+                case Fiber.Status.Suspended(_, _, _, blockingOn, _) =>
+                  ZIO.forall(blockingOn.filterNot(suspended))(loop(_, suspended + fiberId))
```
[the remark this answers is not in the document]
adamgfraser: Yes/

test/shared/src/main/scala/zio/test/environment/package.scala
adamgfraser: Sure.

adamgfraser:
Updated with a few improvements:

- Defined a total ordering on fiber identifiers based on their timestamp and sequence number. Since the children of a fiber will have "larger" identifiers we never have to search the children of a fiber that has an identifier larger than the one we are searching for.
- Memoized getting fibers by identifier.
- Optimized introspection to see if a fiber is indefinitely suspended.
- Addressed a potential issue with inconsistent state. When we check if a fiber is indefinitely suspended we can't ever synchronize on more than one fiber. So we could have a situation where we look at fiber A and say it is suspended on fiber B, then we go to fiber B and see that it is suspended, but in between fiber B has resumed fiber A. To address this I added a method `freeze` that takes a (potentially not fully consistent) snapshot of the blockers of a fiber. Then we can take two snapshots sequentially and conclude that we really are suspended if in both snapshots the fiber is indefinitely suspended and both snapshots are identical.

adamgfraser:
Further changes based on feedback from @fanf.

The issue we had is if we have a forked fiber that has been sleeping for 10 seconds, and then on the main fiber we set the time to 12 seconds and then trigger an effect on the forked fiber that sleeps for another 10 seconds, that effect should resume at 22 seconds (the time we set plus the sleep duration) not 20 seconds (the fiber time plus the sleep duration).

Addressing that essentially leads to removing the concept of fiber time. From a plus side, this definitely adds something in terms of simplicity because now there is only a single concept of time. On the negative side, it makes the `awaitScheduled` operator much more important because if we do `sleep(1.second).fork *> adjust(1.second)` there is a race condition where if the adjust resolves first then the sleep doesn't resume until 2 seconds have passed and we don't terminate. The `awaitScheduled` operator is a relatively straightforward way of dealing with this but from an API perspective it would be nice if the user didn't have to do that.

I've tried playing with a couple of things to bundle the await and the adjustment together but haven't found anything I am totally satisfied with since sometimes we don't want to await at all (e..g if we're just setting the time and checking that the new time is equal to the set time), and sometimes we want to wait on multiple effects being scheduled with `awaitScheduledN`. So open to other ideas on this.

jdegoes:
@adamgfraser Happy to review in detail next week! Glad we waited for some feedback on this one.

adamgfraser:
@jdegoes Yes, let's definitely spend some time on this. I am losing confidence somewhat in the approach here, at least in the current implementation. I am seeing subtle flakiness in the tests here, which I think is partially related to out capturing of the `blockingOn` information in `FiberContext` not quite being 100% but also I don't know if we can be confident that this information will always be accurate, which is quite critical for this implementation at this point.

I am wondering if a better approach is to capture a reference to the main fiber within the test (maybe there is a method on `TestClock` to do this). Then instead of dealing with this `blockingOn` information we could just require that all the fibers forked by that fiber are suspended or done. We would have to figure out how to capture information about daemon fibers there. Then when we call adjust we could wait for the system to enter a suspended state, run the first effect, and repeat until done. I think this would also remove the need to do the `awaitScheduled`.

jdegoes:
@adamgfraser That's an interesting idea. Perhaps we could add the "test main fiber" as an annotation prior to the tests being run?

As for capturing information about daemon fibers, maybe that's a sign that the `ZIO.adoptAll` combinator could be useful, as marking the "adoption point" for disowned fibers?

adamgfraser:
@jdegoes I think we would want the "main fiber" for this particular test (we want to wait for all fibers involved in this test being suspended, it is okay if other fibers are still doing work on other tests) but yes maybe we could capture it with an annotation so the user didn't have to do it.

I think the trick with the daemon fibers is capturing the information without changing the semantics of the program under test. It seems like ideally we would like something that says "any time this fiber or its children fork a fiber, do this".

adamgfraser:
@jdegoes This approach seems quite promising. I am not seeing any of the flakiness that I did with the previous approach. It also creates a very nice API because the user never has to manually suspend before adjusting the clock.

Still needs cleanup, approach for handling deamon fibers, and approach for handling `adjust` being called from something other than the main fiber in a test, but this is already handling all the cases in our tests and I think there are some tests that can be simplified substantially.

jdegoes:
@adamgfraser Awesome. Well let me know when it's cleaned up and we'll merge! Also I can do `ZIO.adoptAll` as part of my PR.

runtologist:
Beautiful. This approach immediately feels right when reading the tests. :)

adamgfraser:

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -401,17 +342,100 @@ package object environment extends PlatformSpecific {
...
+              promise.succeed(()) *>
+                ZIO.yieldNow *>
+                run(_ => end)
+          }
+
+      /**
+       * Returns whether all descendants of this fiber are done or suspended.
+       */
+      private lazy val suspended: UIO[Boolean] =
+        freeze.zipWith(freeze)(_ == _).orElseSucceed(false)
```
jdegoes, line 427: Is this just checking to see you get 2 successive reads with the same statuses? i.e. "stable statuses"?

adamgfraser: Yes.

jdegoes: So there's a bit of a race condition here. What happens if it says that it is suspended, when it is not in fact suspended? Is there other logic to handle that case?

adamgfraser: Yes, we're relying on the two successive snapshots to address this, that if it isn't really done between the first snapshot and the second snapshot some fiber will have changed its status. This seems to be working in practice but I agree it isn't bulletproof. Not sure what the best solution is for this. I tried adding a delay between the snapshots but it seemed to cause issues on Scala.js. Maybe we yield between snapshots?

jdegoes: Given the thread pools have FIFO delays, `yieldTo` would indeed be better. Although that would probably cause issues on Scala.js too (may be worth discovering what those are as they may point to another issue).

We should also maybe make a ticket for this race issue so we don't forget about it. I'm trying to think of what sort of tools would be necessary here.

adamgfraser: `yield` seemed to work for me locally on Scala.js. Running CI now.

adamgfraser: Okay, seems to be working. Hopefully this hardens it a little.

jdegoes: Approved, and sorry for the delay! I think we should create a ticket for this one because it will eventually come back to haunt us...

PR #3426, opened 2020-04-21
Explore Using Implicit Conversion from Chunk to IndexedSeq

adamgfraser:
There has been some discussion about providing an implicit conversion from `Chunk` to `IndexedSeq` instead of using subtyping. This PR is still rough but tries to explore the trade-offs:

**Disadvantages of Implicit Conversion**

- Type inference is definitely worse in certain situations. A `Chunk[A]` is viewable as an `IndexedSeq[A]` but a `ZIO[R, E, Chunk[A]]` is not viewable as a `ZIO[R, E, IndexedSeq[A]]`.

- The signature for a generic `foreach` method would need to change from something like:

```scala
 def foreach0[CC[x] <: Iterable[x], R, E, A, B, That](
    in: CC[A]
  )(f: A => ZIO[R, E, B])(implicit bf: CanBuildFrom[CC[A], B, That]): ZIO[R, E, That] =
    ???
```

to:

```scala
def foreach0[CC[x], R, E, A, B, That](
  in: CC[A]
)(f: A => ZIO[R, E, B])(implicit ev: CC[A] => Iterable[A], bf: CanBuildFrom[CC[A], B, That]): ZIO[R, E, That] =
  ???
```
Again probably would have negative type inference implications in some situations.

- Assertions for iterables no longer work for chunks. Possible adding a similar implicit evidence parameter to assertions could address this but these additional implicit evidence parameters are likely to detract from type inference in some situations when type inference with assertions is already not perfect.

**Advantages of Implicit Conversion**

- Chunk could have a purely functional interface - I'm not sure if we can really hold to this though. The whole point of chunk is for efficient low level operations so I worry we will have to expose operations like `foreach` anyway. We're kind of cheating by making it package private when my understanding is other ZIO libraries are using these methods as well.
- We aren't constrained by other design limitations of Scala collections - I'm not sure how important having more than 32 million elements is but we could have more.
- Internal implementation is probably easier since we don't have to deal with differences in `IndexedSeq` and friends implementation between Scala versions.


jdegoes:
> Chunk could have a purely functional interface - I'm not sure if we can really hold to this though. 

I think you are right here, and we have to expose at least `apply(i: Int)`. Yes, it's unsafe, yes, it's dangerous, but it's also required for the performance-critical code that Chunk enables.

> We aren't constrained by other design limitations of Scala collections - I'm not sure how important having more than 32 million elements is but we could have more.

I think it's only useful for bit chunks, and even then, it's not essential (just a nice to have). 

> The signature for a generic foreach method would need to change from something like:

How about `<%` ("view bounds")? Could they help?

> Type inference is definitely worse in certain situations. A Chunk[A] is viewable as an IndexedSeq[A] but a ZIO[R, E, Chunk[A]] is not viewable as a ZIO[R, E, IndexedSeq[A]].

This is worth paying attention to. In this case, `ZIO[Chunk[...]]` is unlikely to be an input to a function (since an effect combinator is unlikely to take something with that signature; more likely the chunk itself would be an input), so the lack of implicit conversion is probably ok. 

But it would be sad to negatively affect inference...

iravid:
> How about <% ("view bounds")? Could they help?
Deprecated in Dotty: https://scastie.scala-lang.org/2w4aA5tBSZ25hTpzGjMieQ

I don't think it's particularly important to keep the purely functional interface. `foreach` and `apply` are very important for performance-sensitive code.

Just wondering, is overloading `apply` for long values possible? Looks viable: https://scastie.scala-lang.org/1V9URAnYQtS6IxIHtIRRXg

jdegoes:
@iravid The problem is the length / size of the collection. You are capped to the size of an int, which means you have 260 MB (give or take) to store a bit chunk. You can never have a bit chunk bigger than that because you can't talk about its size. 

For a byte chunk, on the other hand, you can consume 2.1 GB before you can no longer describe the size of the chunk. Which is clearly well beyond the realm of usable for most applications.

jdegoes:
> I don't think it's particularly important to keep the purely functional interface. foreach and apply are very important for performance-sensitive code. I am in favor of keeping the current extension to IndexedSeq. Far more overall benefits.

I'm open to either way on this. There's no perfect solution, it's all about tradeoffs.

iravid:
It may be a fair tradeoff to require refining a `Chunk[Boolean]` to a `BitChunk` in order to work with it optimally. So you'd have a `def actuallyBiggerThanIntSize: Long` just on the subtype. :-)

adamgfraser:
Aren't view bounds being removed in Dotty?

iravid:
@adamgfraser Yup: https://scastie.scala-lang.org/2w4aA5tBSZ25hTpzGjMieQ

jdegoes:
> It may be a fair tradeoff to require refining a Chunk[Boolean] to a BitChunk in order to work with it optimally. So you'd have a def actuallyBiggerThanIntSize: Long just on the subtype. :-)

It's a big violation of abstraction though (LSP), since now you have to know the subtype to determine if a method in the parent is actually going to do what it's supposed to do.

I think we can give up "big bit chunks" (which I'm fine with doing) or give up `IndexedSeq`, but we have to give up one...

jdegoes:
> Aren't view bounds being removed in Dotty?

Yes but it's easy to fake them with implicit `<:<`.

adamgfraser:
Yes, that is similar to what I did with the generic implementation of `foreach` above. I think that will definitely work in some places. The area where it doesn't seem to work well is with assertions, where including that type of implicit evidence requirement leads to types being inferred as `Nothing` when composing assertions using `&&`. I've tried a couple of other approaches like creating a new `ChunkLike` type that has implicit conversions from `Iterable` and `Chunk` and haven't been able to get something to work. But assertions shouldn't be driving the bus here and we can always pursue that in a follow up PR.

adamgfraser:

adamgfraser:
Closing as I think we are okay with not having chunks with more than `Int.MaxValue` elements.

Issue #3427, opened 2020-04-21
Better ZLayer bounds error messages


tabdulradi:
Hello 👋 I've been using ZLayer at work for a while now. One of my biggest problems is the bound related error messages. For example, sometimes I mix up `provideCustomLayer` and `provideSomeLayer` like:
```scala
def foo: ZLayer[Foo] = ???
def x: ZIO[Foo with Bar with ZEnv, Nothing, Unit] = ???
x.provideCustomLayer(foo)
```
 which shows cryptic error message like:
> Cannot prove that zio.ZEnv with Foo <:< Foo with Bar with zio.ZEnv.

My proposal is to customise the error message by using type aliases and implicitNotFound. I've pushed a proof of concept here https://github.com/zio/zio/pull/3428
The same snippet on my branch produces this error message instead: 
>scala> x.provideCustomLayer(foo)
<console>:29: error: provideCustomLayer expect all layers except those provided by ZIO (ZEnv) instead it got Foo. If you intend to provide some of layers only then use provideSomeLayer instead
       x.provideCustomLayer(foo)
                           ^

jdegoes:
@tabdulradi Love it!

axaluss:
There is also the layer composition error messages which can be very long lists of things that are expected vs what was given. 
Is there a way to create a delta list of things that are missing?

deabreu:
Hey is this a dead issue? I've been facing the same problems.

jdegoes:
cc @kitlangton for insight on if / how this might be possible

kitlangton:
The original case is handled by the layer macros (compiling it today gives the following output—provideCustom no longer exists of course, as the default services don't use the environment). But I'd be good to know what situation @deabreu's running into, maybe there are additional cases that are troublesome.

```scala
──── ZLAYER ERROR ────────────────────────────────────────────────────

 Please provide layers for the following 2 types:

   1. course.cohort4.Cool.Bar
   2. course.cohort4.Cool.ZEnv
      
──────────────────────────────────────────────────────────────────────
```

adamgfraser:
Yes, I think this problem is solved by using automatic layer construction.

@deabreu Is that what you are doing? If so and you are still running into problems we can look into it but otherwise I think the solution is just to use automatic layer construction.

deabreu:
Hello all (@jdegoes, @kitlangton and @adamgfraser). I'm afraid this is on a different compilation step.

[...]

PR #3428, opened 2020-04-21
POC: Better implicit errors

tabdulradi:
Check https://github.com/zio/zio/issues/3427 for details

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -27,6 +27,11 @@ import zio.internal.tracing.{ ZIOFn, ZIOFn1, ZIOFn2 }
 import zio.internal.{ Executor, Platform }
 import zio.{ TracingStatus => TracingS }
 
+object Evidences {
+  @implicitNotFound("provideCustomLayer expects providing all environment layers (${R}) except those already provided by ZIO (ZEnv), but it only got ${R1}.\nPlease provide the missing layer(s) or instead use provideSomeLayer if you actually intend to only provide some of layers.") 
+  type ProvideCustomLayer[-R1, +R] = ZEnv with R1 <:< R
+}
+
 /**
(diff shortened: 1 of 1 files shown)


adamgfraser:
I think this is a fantastic idea! We already took some steps in this direction in #3319 but would be great to do more here. I think you can provide a custom implicit message along with the parameters [see (here](https://www.scala-lang.org/api/current/scala/annotation/implicitNotFound.html)) so maybe we don't need the type aliases?

tabdulradi:
oh cool! I didn't know you can can customise messages per param. That would simplify things ALOT. 👍

adamgfraser:
@tabdulradi Just wanted to check in on this. I think this would be a great feature to get in and probably needs to be done sooner rather than later if we are going to get it in for 1.0. Anything I can do to help?

tabdulradi:
Hey @adamgfraser Thanks for the nudge. I rebased the PR, and improved the error message. Then [attempted to use the trick you mentioned](https://github.com/zio/zio/pull/3428/commits/a2a36f5beea0951c523d1be7721d4f5b625bfbfe), but unfortunately seems it works on Scala 2.13 only.

tabdulradi:
I removed the interpolation that's annoying the compiler. Now it compiles but without any effect, the user still gets "Cannot prove bla bla ...". I assume when we cross-compile, users on 2.13 should see the (no-interpolation) message but not sure it.

adamgfraser:
@tabdulradi Yes, I think unfortunately since the `@implicitNotFound` annotation only works in parameter position on Scala 2.13 we need to revert to the encoding where we require implicit evidence of a trait.

jdegoes:
@tabdulradi I am a big fan of this. Although now, we have to be careful to do this in a binary backward compatible way, which I suspect will be trickier. I'd love if all the implicits had custom error messages, including `<:<` which is used a lot with `Has`. 

I will close this for now, but feel free to re-open anytime if you get a chance to continue this work!

PR #3431, opened 2020-04-22
Add specialized accessor operations to Chunk

whysoserious:


core-tests/shared/src/test/scala/zio/ChunkSpec.scala
```diff
@@ -30,7 +30,7 @@ object ChunkSpec extends ZIOBaseSpec {
 
   def chunkWithIndex[R <: Random, A](a: Gen[R, A]): Gen[R with Sized, (Chunk[A], Int)] =
     for {
-      chunk <- Gen.chunkOfBounded(0, 100)(a)
+      chunk <- Gen.chunkOfBounded(1, 100)(a)
```
adamgfraser, line 33: Good catch!

PR #2479, opened 2019-12-16
WIP: Scala.js upgrade to v1.0.0 (#2392)

lvitaly:

PR #3432, opened 2020-04-22
Update Magnolia

nightscape:
There is a bug in Magnolia versions < 0.15 which causes derivation for case classes with default values to fail:
https://github.com/propensive/magnolia/issues/227
The update should fix it.


ghostdogpr:
Magnolia is only available for Scala.js 1.0+, so you will need this merged first: https://github.com/zio/zio/pull/2479

adamgfraser:
@nightscape Thanks! I wonder why this wasn't getting picked up by Scala Steward?

ghostdogpr:
@adamgfraser for the reason above 😄 Scala Steward is smart!

adamgfraser:
@ghostdogpr Ah good point. Well fortunately it looks like we aren't far away on that one.

adamgfraser:
@ghostdogpr Indeed!

PR #3439, opened 2020-04-23
Relax Type Constraints on ProvideLayer

adamgfraser:
We don't actually need the output type of the layer to be a subtype of `Has` for `provideLayer`, only for the "fancier" `provideSomeLayer` and `provideCustomLayer` operators that make use of `++`. This makes it easier for people who want `ZLayer` for the resource management but don't want to use `Has`. In particular in ZIO Test `provideLayer` is the way to provide a resource to a test, so this lets you use that functionality without wrapping your resource in a `Has` if you don't want to.


jdegoes:
Awesome!

PR #3444, opened 2020-04-24
Properly Suspend Effects in ZIO#foreachPar

### LINKED ISSUE #3429: Bug in nested parallel operations ?
```scala
val op = random.nextInt.tap(s => console.putStrLn(s"random: $s"))
val col = ZIO.collectAllPar(List(op, op, op)).tap(s => console.putStrLn(s"List: $s"))
val zip = col.zipPar(col).tap(s => console.putStrLn(s"Lists $s"))
val prog = zip.map(t => t._1 == t._2)
Runtime.default.unsafeRun(prog.provideLayer(ZEnv.live))
```

Sometimes it will produce 6 different random values but two lists from zip are identical.

adamgfraser:
Resolves #3429.


jdegoes:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #3445, opened 2020-04-24
Update to izumi-reflect 1.0.0-M1, use izumi-reflect on dotty

neko-kai:
Type signatures were added in SpecSpec and mock modules due to https://github.com/lampepfl/dotty/issues/8764

Also had to add a workaround for inference in PolyMockSpec, inference changed when `ClassTag` was substituted for `Tag` - causing spurious inference of `Any`, the only explanation i can think of is Tag's parameter being `<: AnyKind` causes this.

/cc @smarter


neko-kai:
@ioleo Also i moved `@mockable` macro to consider only methods with `Tag` as tagged: https://github.com/zio/zio/pull/3445/commits/d6e91c837b146bd024ce9ed5eab67140ae89248f (`ClassTag` is not used by `Tag`'s derivation macro, it will not help assemble the tag for polymorphic .get)

PR #3451, opened 2020-04-25
ZTransducer.dropWhile

regiskuckaertz:


iravid (approved):

iravid:
Can you fix the conflicts @regiskuckaertz?

regiskuckaertz:
All set @iravid I commented out the test as it failed. My assumption is that the failing element should just bypass the aggregation step, but it didn't. There may be a bit more work needed there. What do you think?

iravid:
The error handling combinators currently short-circuit the stream. Is that a test from the previous version?

iravid (approved):

regiskuckaertz:
No, I wrote it thinking about pull safety.

iravid:
Ok, we'll get there soon!

PR #3452, opened 2020-04-25
Use short forms where it makes sense

reibitto:
I ran zio-intellij inspections across the project to see what would come up.

There are some I didn't change because they were specifically testing things like `*>`. I also didn't touch the tracing tests because I assume there might be subtle differences if I use different methods. Not sure.

I also didn't change certain things (like using ZIO.when) as micro optimizations. I figure this is the same reason why some longer forms are used throughout the ZIO codebase.

Changes like using `ZIO.some` and so on should actually be faster than the old code because internally it uses `succeedNow`.


adamgfraser:
Looks great! I think in some cases it may just be because we didn't have the shorter versions at the time some of this code was written or different contributors weren't aware of it.

PR #3453, opened 2020-04-25
Use ChunkBuilder more prominently in Chunk

### LINKED ISSUE #3393: Migrate all lazy array creation code in Chunk to use ChunkBuilder instead
Currently, many `Chunk` operations like `map` allocate arrays, lazily creating them until they know what their correct element type should be.

This creates a lot of obfuscation in the code base and makes it difficult to create type-specific behavior around chunk creation. 

Since we have `ChunkBuilder` now, it could be a good idea to migrate this manual lazy creation of arrays to use `ChunkBuilder` instead, which has the same logic inside it.

To do this efficiently, each operation like `map` should create a new chunk builder, and then call `sizeHint` to set the size of the expected output chunk. Then it can add the elements into the chunk builder, and finally convert to the return chunk type.

mschuwalow:
resolves #3393

some benchmarks:

old
Benchmark                     (size)  Mode  Cnt      Score     Error  Units
MixedChunkBenchmarks.filterM    1000  avgt   25  91126.072 ± 262.119  ns/op
MixedChunkBenchmarks.fold       1000  avgt   25  22232.029 ±  67.771  ns/op
MixedChunkBenchmarks.mapM       1000  avgt   25  37386.854 ±  72.429  ns/op

new
Benchmark                     (size)  Mode  Cnt      Score      Error  Units
MixedChunkBenchmarks.filterM    1000  avgt   25  89863.537 ± 4217.411  ns/op
MixedChunkBenchmarks.fold       1000  avgt   25  22145.585 ±  115.194  ns/op
MixedChunkBenchmarks.mapM       1000  avgt   25  38005.838 ±  934.068  ns/op

### DIFF
--- a/core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
+++ b/core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
@@ -43,7 +43,7 @@ private[zio] object ChunkBuilder {
   /**
    * Constructs a generic `ChunkBuilder`.
    */
-  def make[A]: ChunkBuilder[A] =
+  def make[A](): ChunkBuilder[A] =
     new ChunkBuilder[A] {
       var arrayBuilder: ArrayBuilder[A] = null
       var size: SInt                    = -1
--- a/core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
+++ b/core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
@@ -43,7 +43,7 @@ private[zio] object ChunkBuilder {
   /**
    * Constructs a generic `ChunkBuilder`.
    */
-  def make[A](): ChunkBuilder[A] =
+  def make[A]: ChunkBuilder[A] =
     new ChunkBuilder[A] {
       var arrayBuilder: ArrayBuilder[A] = null
       var size: SInt                    = -1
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala-2.11-2.12/ChunkBuilder.scala
```diff
@@ -43,7 +43,7 @@ private[zio] object ChunkBuilder {
   /**
    * Constructs a generic `ChunkBuilder`.
    */
-  def make[A](): ChunkBuilder[A] =
+  def make[A]: ChunkBuilder[A] =
```
jdegoes, line 46: We should keep the parens. It is a Scala convention for side effecting nullary methods to use parens to distinguish them from vals at the call-site.

mschuwalow: I was thinking so as well, but noticed that the standard library doesn't use parens in for example ArrayBuilder#make.
Discussion why they decided so is here:
https://github.com/scala/collection-strawman/issues/520

Should we still stick with parens?

jdegoes: I think it was a mistake to not use parens in `ArrayBuilder.make` and that they would change that decision now.

mschuwalow: alright, updated

PR #3456, opened 2020-04-26
ZIO Test: Improve Management of Resources in Specs

adamgfraser:
Right now specs are effectual but because the effect is a `ZIO` it has a fixed "scope" so when we do `provideLayerShared` we need to create one effect that conceptually describes "running" the entire spec so that we can provide the layer to it. This creates issues because we have to prematurely decide what execution strategy to use and some combinators don't work after `provideLayerShared` since we have "already run" the spec by then.

This PR changes specs to use a `ZManaged` so we can expand the scope dynamically. This way we can just describe providing a resource to the spec and only at the time we execute the spec keep the scope open while we execute all the tests in the suite.


iravid:
As t -> ♾, all `ZIO` values become `ZManaged` values 😂

jdegoes:
:shipit:

adamgfraser:
@jdegoes I think this is ready to go.

PR #3466, opened 2020-04-26
Implement ZManaged#release

adamgfraser:
Following up on discussion on Discord, implements `Managed#release`, which is equivalent to `managed.use(ZIO.succeed)` and lets you go back from `ZManaged` to a `ZIO` by describing running the acquire and release actions and returning the result. This is only safe if the result is valid outside the scope in the same way that the user needs to not leak the resource in the `use` action of `bracket` but can be useful if we have already converted the resource into a result in the context of the managed.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -725,14 +725,6 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R, 
     new ZManaged.ProvideSomeLayer[R0, R, E, A](self)
 
   /**
-   * Runs the acquire and release actions and returns the result of this
-   * managed effect. Note that this is only safe if the result of this managed
-   * effect is valid outside its scope.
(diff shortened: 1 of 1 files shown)


iravid:
@adamgfraser I misunderstood your description of the combinator. I thought you were going for a constructor that could be used as: `a *> ZManaged.release *> b`.

Not sure if `release` is the best name for what you have here; maybe `useUnwrap`/`useAndGet`?

adamgfraser:
Yes agree isn't the right name. Don't know if I love the others but also don't have something better at the moment.

iravid:
Definitely should be prefixed with `use` to indicate what it's doing, right?

adamgfraser:
I think so. I could also see something like `run` as in run it to a ZIO like with the `run` variants on `ZStream` but we already have a different meaning for that. I could see overloading `use` with no arguments. `useNow`?

iravid:
`useNow` 👍🏻

adamgfraser:
Done.

PR #3469, opened 2020-04-26
Reentrant lock benchmark improvements

robmwalsh:
follow on from #3099 for benchmark improvements

and no, I'm not going to leave it at 1 iteration :)


adamgfraser:
@unclebob418 What do we need to do to get this over the finish line?

unclebob418:
@adamgfraser sorry, got a bit side tracked with zio-sql. [Here](https://github.com/zio/zio/pull/3099#discussion_r415411775) was my last comment from the previous PR. Do you think this is a reasonable idea? Ideally we actually use the reentry capabilities at least to the extent that java supports them to do this (e.g. read first, then upgrade to write for the writers)

jdegoes:
@robmwalsh I know this was an experiment and WIP, so I will close for now, but feel free to re-open anytime!

PR #3477, opened 2020-04-28
Add ZIO#filterNot and ZSTM#filterNot

### LINKED ISSUE #3476: How about adding `ZIO.filterNot`?
```scala
ZIO.filter(list)(predicate(_).map(!_))
```
can be simplified like:
```scala
ZIO.filterNot(list)(predicate)
```

taeguk:
Closes #3476

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2467,7 +2467,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
 
   /**
    * Filters the collection using the specified effectual predicate, removing
-   * all elements that satisfies the predicate.
+   * all elements that satisfy the predicate.
    */
   def filterNot[R, E, A](as: Iterable[A])(f: A => ZIO[R, E, Boolean]): ZIO[R, E, List[A]] =
     filter(as)(f(_).map(!_))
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2465,6 +2465,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
       f(a).zipWith(zio)((p, as) => if (p) a :: as else as)
     }
 
+  /**
+   * Filters the collection using the specified effectual predicate, removing
+   * all elements that satisfies the predicate.
```
ghostdogpr, line 2470: minor:
```suggestion
   * all elements that satisfy the predicate.
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1044,6 +1044,13 @@ object ZSTM {
       f(a).zipWith(zio)((p, as) => if (p) a :: as else as)
     }
 
+  /**
+   * Filters the collection using the specified effectual predicate, removing
+   * all elements that satisfies the predicate.
```
ghostdogpr, line 1049: minor:
```suggestion
   * all elements that satisfy the predicate.
```

core/shared/src/main/scala/zio/ZIO.scala
taeguk: Thanks :)

PR #3479, opened 2020-04-28
docs: remove unused (and shadowed) alias for Has.add

vitaliis:
Every time I was reading this part of documentation, I was confused by the aforementioned operator.

First: '+' operator doesn't exist on `ZLayer`
Second: using '+' results into `String`, which is most probably not what we want

### DIFF
--- a/core/shared/src/main/scala/zio/Has.scala
+++ b/core/shared/src/main/scala/zio/Has.scala
@@ -97,7 +97,6 @@ object Has {
   }
 
   implicit final class HasSyntax[Self <: Has[_]](private val self: Self) extends AnyVal {
-    def +[B](b: B)(implicit tag: Tagged[B]): Self with Has[B] = self.add(b)
 
     def ++[B <: Has[_]](that: B)(implicit tagged: Tagged[B]): Self with B = self.union[B](that)
 
--- a/docs/howto/use_modules_and_layers.md
+++ b/docs/howto/use_modules_and_layers.md
@@ -107,7 +107,7 @@ We encountered two new data types `Has` and `ZLayer`, let's get familiar with th
 
 ### The `Has` data type
 
-`Has[A]` represents a dependency on a service of type `A`. Two `Has[_]` can be combined _horizontally_ through `+` and `++` operators, as in
+`Has[A]` represents a dependency on a service of type `A`. Two `Has[_]` can be combined _horizontally_ through `++` operator, as in
 
 ```scala mdoc:invisible
 object Repo {


ghostdogpr:
Hmm, there is one no? https://github.com/zio/zio/blob/a6fcdce1c7efe0e640f2d09296bda5ccf9d44de6/core/shared/src/main/scala/zio/Has.scala#L100

ghostdogpr:
The sentence is about `Has`, which has a `+` operator: https://github.com/zio/zio/blob/a6fcdce1c7efe0e640f2d09296bda5ccf9d44de6/core/shared/src/main/scala/zio/Has.scala#L100

neko-kai:
@ghostdogpr @vitaliis @adamgfraser 
The `+` operator is not resolved because `any2stringadd` beats it, because Predef is treated as if imported as a wildcard `import Predef._` - so it beats out everything. 

So, this definition might as well not exist.

The behavior is different on Dotty - `+` is resolved, but only becaause `any2stringadd` is intentionally [blocked out](https://github.com/lampepfl/dotty/blob/f26808e59da8d4e05cd54d496e684d00888cddd1/compiler/src/dotty/tools/dotc/typer/ImportInfo.scala#L32) (I think this is undocumented, although maybe i missed it. Opened a [ticket](https://github.com/lampepfl/dotty/issues/8830))

adamgfraser:
Indeed. Shall we delete `+`? It is a symbolic alias for `add` so we're not losing any functionality and right now it creates the potential for a lot of confusion.

neko-kai:
Yeah, I think it should be deleted. Although I would also add `++` as a symbolic alias to `add`, since `Has` is 'collection-like'.

adamgfraser:
We already have `++` / `union` for combining two `Has` values. `+` / `add` was for combining one `Has` value with another value that was not inside a `Has`.

vitaliis:
Yep, `++` is currently not possible because `the conflicting method ++ was defined at line 100:9`.

We could rename it to `:+` in case we want to keep the symbolic alias. This would be analogous to a `List`:
```
@ val l = List(3,4,5) 
l: List[Int] = List(3, 4, 5)

@ l :+ 7 
res2: List[Int] = List(3, 4, 5, 7)
```
Alternative: I also like the explicit self-describing name like `addService`. Although it might be too bulky for this DSL, it might serve as a good documentation piece directly in code, which makes the api more descriptive.

vitaliis:
I'll change it to `:+` collection-like alias. 

In case it has problems or there are arguments against it, we can remove it.

adamgfraser:
Personally I would be inclined to delete it. It isn't used a single time in our entire codebase which I think is telling. Given it is rarely used and we have a pretty short method name in `add` I think we can just use that.

vitaliis:
Yeah, also haven't seen a use case until now. So I am also fine with deleting.

PR #3483, opened 2020-04-29
Explore Initial Encoding of ZLayer

adamgfraser:
Explores moving to an initial encoding of `ZLayer` to allow more static analysis of the dependency graph. It goes to an algebraic data type as follows:

```scala
sealed trait ZLayer[-RIn, +E, +ROut]

object ZLayer {
  private final case class Fold[RIn, E, E1, ROut, ROut1](self: ZLayer[RIn, E, ROut], failure: ZLayer[(RIn, Cause[E]), E1, ROut1], success: ZLayer[ROut, E1, ROut1]) extends ZLayer[RIn, E1, ROut1]
  private final case class Managed[-RIn, +E, +ROut](self: ZManaged[RIn, E, ROut]) extends ZLayer[RIn, E, ROut]
  private final case class ZipWithPar[-RIn, +E, ROut, ROut2, ROut3](self: ZLayer[RIn, E, ROut], that: ZLayer[RIn, E, ROut2], f: (ROut, ROut2) => ROut3) extends ZLayer[RIn, E, ROut3]
}
```

Everything is working except I don't think we can support the `unwrap` method with this encoding, at least if we want to retain any ability to do static analysis of what is inside the, and we have to think about the meaning of the `fresh` combinator.

I added a `size` method to demonstrate the ability to do some static analysis. I think we may want to consider including some type of identifier in layers. It seems like one of the more useful things we could do here is render the dependency graph. Right now we can capture the high level structure but don't really have any meaningful values to label the nodes of the dependency graph with.


iravid:
Very nice. The algebra here is fairly small so no downside in the size of the interpreter, and ZLayer usually won't be on the hot path of most applications. Looks like a pretty good tradeoff for being able to introspect the dependency graph.

jdegoes:
> Everything is working except I don't think we can support the unwrap method with this encoding, at least if we want to retain any ability to do static analysis of what is inside the, and we have to think about the meaning of the fresh combinator.

Yes, although if you add unwrap, it could be another term: it just means in some cases (where `Unwrap` is used), you cannot fully analyze the layer without performing an effect.

> I think we may want to consider including some type of identifier in layers. 

I think you're right. Maybe something like ZIO Test:

```scala
myLayer ?? "Postgres product db"
```

> It seems like one of the more useful things we could do here is render the dependency graph. 

👍 

Big 👍 on this from me. This is going to be very useful and will also allow us more power with optimization.

As for `fresh`, we could implement it as `Fresh` unary operator that triggers a rebuild of that layer; if we want to keep existing semantics. Or we can reinvestigate them.

adamgfraser:
Implemented `fresh` as a new instruction that says "ignore the provided memoMap when building this layer and build it with an empty one". I think this achieves nice semantics where `fresh` ensures a layer will be built again but if components inside the layer are themselves reused they will only be rebuilt once. For example `(layer1 ++ layer1) ++ (layer1 ++ layer1).fresh` should build the layer exactly twice. This also avoids having to bring in any concept of identifiers for memoization so that can be a completely orthogonal concept that is used for rendering.

jdegoes:
@adamgfraser Superb!

adamgfraser:
@jdegoes Great! Will pursue the rendering in a follow up PR.

PR #3484, opened 2020-04-29
Implement assertElements

wi101:
assertElements helps to test a list of nested data and check each field with different assertions, 

example
 ```scala

case class Person(name: String, age: Int)
case class User(name: String, birthYear: Int)
...
val l1: List[User] = ???
val l2: List[Person] = ???
assert(l1)(hasSize(equalTo(l2))) && 
assertElements(l1.zip(l2)){case (user, person) =>
assert(user.name)(equalTo(person.name)) && 
assert(user.birthYear)(equalTo(currentYear - person.age)) }
``` 

sorry for the old commits (I couldn't squash them 🙈 )

### CODE REVIEW

**Comment by adamgfraser** [comment]
@wi101 Isn't this the same functionality as `Assertion#forall`?

**Comment by wi101** [comment]
@adamgfraser Not really, `assertElements` uses each element for a given assertion. Or is there a better way?

**Comment by adamgfraser** [comment]
@wi101 Ah, I see. You can use `BoolAlgebra.foreach` for this:

```scala
import zio.test._
import zio.test.Assertion._

object Example {

  val currentYear = 2020

  case class Person(name: String, age: Int)
  case class User(name: String, birthYear: Int)

  lazy val l1: List[User] = ???
  lazy val l2: List[Person] = ???

  val sameSize = assert(l1)(hasSize(equalTo(l2.size)))
  val sameElements = BoolAlgebra.foreach(l1.zip(l2)) { case (user, person) =>
    assert(user.name)(equalTo(person.name)) && 
    assert(user.birthYear)(equalTo(currentYear - person.age))
  }
  sameSize && sameElements.getOrElse(assertCompletes)
}
```

wi101:
Thanks, I tried it, the problem is that `foreach` requires a `BoolAlgebra` it seems that i need something like this:
```scala
if (assertion.isSuccess)
  BoolAlgebra.success(true) 
else  
 BoolAlgebra.fail("wrong")
```

adamgfraser:
I think maybe there is an issue with other version changes? The example I posted works on `master`.

wi101:
Alright, I think this method is not needed, I will close this PR.

PR #3510, opened 2020-05-01
Redesign ZManaged around an embedded finalizer map

iravid:


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1201,142 +1081,170 @@ object ZManaged {
...
+   * it is the result of executing [[ZResource#use]] or an arbitrary value
+   * passed into [[ReleaseMap#release]].
    */
-  trait FinalizerRef[R] {
+  type Finalizer = Exit[Any, Any] => UIO[Any]
+  object Finalizer {
+    val noop: Finalizer = _ => UIO.unit
+  }
+
+  final class IfM[R, E](private val b: ZManaged[R, E, Boolean]) extends AnyVal {
```
regiskuckaertz, line 1093:
```scala
ifM (expr) {
  consequentM
} elseM {
  alternateM
}
```
😃

iravid: Lol. Happy to satisfy your DSL desires in a follow up 😂

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1201,142 +1081,170 @@ object ZManaged {
...
+    def apply[E1 >: E, R1 <: Has[_]](
+      layer: ZLayer[R0, E1, R1]
+    )(implicit ev1: R0 with R1 <:< R, ev2: NeedsEnv[R], tagged: Tagged[R1]): ZManaged[R0, E1, A] =
+      self.provideLayer[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer)
+  }
+
+  /**
+   * A key used for identifying finalizers in a [[ReleaseMap]].
+   */
+  case class ReleaseKey(l: Long) extends AnyVal {
```
regiskuckaertz, line 1115: Could use [`UniqueKey`](https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/internal/UniqueKey.scala) instead

iravid: Currently I can't because I rely on Long inside `ReleaseKey` for ordering. But I would like to embed the ordering into the data structure holding the finalizers.

iravid:
@ioleo I'm fairly confident that ZLayer is constructed correctly now, but the mocking specs are still not passing. Any chance you could look into this?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -299,25 +296,8 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-            for {
-              fs    <- finalizers.get
-              exits <- ZIO.foreach(fs)(_(exitU).run)
-              _     <- ZIO.done(Exit.collectAll(exits).getOrElse(Exit.unit))
-            } yield ()
-        )
-      }
-    }
+  def flatMap[R1 <: R, E1 >: E, B](f: A => ZManaged[R1, E1, B]): ZManaged[R1, E1, B] =
+    ZManaged(self.zio.flatMap { case (_, a) => f(a).zio })
```
regiskuckaertz, line 300: Isn't the finaliser of the first effect important to keep here?

iravid: That's a good question. The finalizers returned from the underlying `ZIO` are only used for early release.

I'm wondering if the correct thing to do is only allow the resulting ZIO to be released early, or should an early release of the resulting ZIO should also release the flatMapped ZIO.

Will think about this. Thank you for pointing this out!

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: Doe we maybe want to make it an abstract path-dependant type on ReleaseMap to make sure that 1. users can't do anything with it except use it as a key into the map 2. make sure they can't mix the keys

iravid: Yes, good idea! Will do that.

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1201,142 +1101,214 @@ object ZManaged {
...
-     * Constructs a new `FinalizerRef` with the specified initial finalizer.
+     * Creates a new ReleaseMap.
      */
-    def make[R](initial: Exit[Any, Any] => URIO[R, Any]): UIO[FinalizerRef[R]] = {
+    def make: UIO[ReleaseMap] = {
+      sealed abstract class State
+      final case class Exited(nextKey: ReleaseKey, exit: Exit[Any, Any], done: Map[ReleaseKey, Finalizer]) extends State
+      final case class Running(nextKey: ReleaseKey, finalizers: Map[ReleaseKey, Finalizer])                extends State
+
+      Ref.make[State](Running(ReleaseKey.initial, Map())).map { ref =>
```
mschuwalow, line 1211: What about using a TreeSeqMap or other insertion order preserving data structure here?

iravid: That looks really nice. I wasn't aware of that.

But from what I see, it's only available for 2.13 :-(

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1481,25 +1462,25 @@ object ZManaged {
...
-    ZManaged {
-      for {
-        finalizers <- FinalizerRef.make(initial)
-        reservation = Reservation(
-          acquire = ZIO.succeedNow(finalizers),
-          release = finalizers.runAll
-        )
-      } yield reservation
-    }
+  // def finalizerRef[R](
```
mschuwalow, line 1472: I think it makes sense to keep this one for when you don't need the full power of ReleaseMap

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: I faced a similar problem in keeper some time ago. I was thinking / planning to reimplement Java's LinkedHashMap (or some alternative) as a low-level STM data structure. Might also be a good fit here

iravid: Well, if you'd be up for that in the future, it's going to be very useful as we're planning on baking `ReleaseMap` (in some form) into the `FiberContext`. So a more efficient implementation would be awesome!

iravid: For this PR, I think the current scheme is sufficient, and we can improve on it later on 💪🏻

core/shared/src/main/scala/zio/ZManaged.scala
iravid: Ok, good point. The latest revision of `ZManaged.finalizerRef` actually gave you a `FinalizerRef` trait which is almost identical to `ReleaseMap`. In fact, `ReleaseMap` was based on it :-)

I think I will dial down `finalizerRef` to return something that is like a combination of `switchable` and `Ref[Finalizer]`.

core/shared/src/main/scala/zio/ZManaged.scala
iravid: @mschuwalow So I ended up switching to `scala.collection.immutable.LongMap`. Cut down a nice chunk of logic and also keeps iteration order for us 💪🏻

core/shared/src/main/scala/zio/ZManaged.scala
iravid: Ok, it's definitely right to compose the early release finalizers as well. Thanks @regiskuckaertz.

iravid: Ha, I even did that without thinking for `ZManaged.foreach`.

regiskuckaertz: nice 👌

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3129,6 +3127,21 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+   *
+   * It's unsafe to execute side effects inside `f`, as `f` may be executed
+   * more than once for some of `in` elements during effect execution.
+   */
+  def mergeAllParN[R, E, A, B](n: Int)(
+    in: Iterable[ZIO[R, E, A]]
+  )(zero: B)(f: (B, A) => B): ZIO[R, E, B] =
+    Ref.make(zero) >>= { acc =>
+      foreachParN_(n)(in) {
+        Predef.identity(_) >>= { a => acc.update(f(_, a)) }
```
regiskuckaertz, line 3141: Do you need the `Predef.identity`? 😮

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3217,6 +3230,21 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     mergeAllPar(all)(Option.empty[A])((acc, elem) => Some(acc.fold(elem)(f(_, elem)))).map(_.get)
   }
 
+  /**
+   * Reduces an `Iterable[IO]` to a single `IO`, working in up to `n` fibers in parallel.
```
regiskuckaertz, line 3234: Perhaps worth mentioning that `f` must be commutative and associative?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -934,49 +904,36 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
+                             (result, sleeper) => sleeper.interrupt *> ZIO.done(result.map(tp => Right(tp._2))),
+                             (_, resultFiber) => UIO.succeed(Left(resultFiber))
+                           )
+                           .provide(r)
+                       }
+          a <- raceResult match {
+                case Right(value) => UIO.succeed(Some(value))
+                case Left(fiber) =>
+                  ZIO.fiberId.flatMap { id =>
+                    fiber.interrupt
```
regiskuckaertz, line 928: I think you could pass id here too right?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1125,62 +1076,18 @@ final class ZManaged[-R, +E, +A] private (reservation: ZIO[R, E, Reservation[R,
...
-                  _  <- finalizers.add(release)
-                  a1 <- acquire
-                } yield a1
-            }
-            left.zipWithPar(right)(f0)
-          },
-          release = finalizers.runAll
-        )
-      }
+  def zipWithPar[R1 <: R, E1 >: E, A1, A2](that: ZManaged[R1, E1, A1])(f: (A, A1) => A2): ZManaged[R1, E1, A2] =
```
regiskuckaertz, line 1079: nice simplification 😲

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1201,142 +1108,218 @@ object ZManaged {
...
+      // into the map using keys that will build it in reverse. That way, when we do the final iteration,
+      // the finalizers are already in correct order.
+      val initialKey: Long = -1L
+      def next(l: Long) =
+        if (l == 0L) throw new RuntimeException("ReleaseMap wrapped around")
+        else if (l == Long.MinValue) Long.MaxValue
+        else l - 1
+
+      sealed abstract class State
+      final case class Exited(nextKey: Long, exit: Exit[Any, Any], done: LongMap[Finalizer]) extends State
```
regiskuckaertz, line 1226: do you need `done` at all?

regiskuckaertz:
This is amazing

core/shared/src/main/scala/zio/ZIO.scala
iravid: It’s shadowed by ZIO.identity 🤷🏼‍♂️

iravid: Oh sorry I misread. This is silly. Will fix.

I admit to just copying the mergeAllPar implementation 😂

core/shared/src/main/scala/zio/ZManaged.scala
iravid: `Fiber#interrupt` automatically uses the current fiber's ID as the interruptor.

core/shared/src/main/scala/zio/ZManaged.scala
iravid: No, good riddance!

core/shared/src/main/scala/zio/ZIO.scala
iravid: Yes 👍

core/shared/src/main/scala/zio/ZManaged.scala
mschuwalow: very cool stuff!

PR #3511, opened 2020-05-02
Fix doc: creating effects succeed method description

YingLiu4203:
Both the `ZIO.succeed` and `Task.succeed` take a by-name parameter `(a: => A)`. The parameter is not eagerly constructed.

### DIFF
--- a/docs/overview/creating_effects.md
+++ b/docs/overview/creating_effects.md
@@ -23,7 +23,7 @@ You can also use methods in the companion objects of the `ZIO` type aliases:
 val s2: Task[Int] = Task.succeed(42)
 ```
 
-The `succeed` method is eager, which means the value passed to `succeed` will be constructed _before_ the method is invoked. Although this is the most common way to construct a successful effect, you can achieve lazy construction using the `ZIO.effectTotal` method:
+The `succeed` method takes a by-name parameter. Although this is the most common way to construct a successful effect, you can achieve lazy construction using the `ZIO.effectTotal` method:
 
 ```scala mdoc:silent
 lazy val bigList = (0 to 1000000).toList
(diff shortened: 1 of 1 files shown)


docs/overview/creating_effects.md
```diff
@@ -23,7 +23,7 @@ You can also use methods in the companion objects of the `ZIO` type aliases:
 val s2: Task[Int] = Task.succeed(42)
 ```
 
-The `succeed` method is eager, which means the value passed to `succeed` will be constructed _before_ the method is invoked. Although this is the most common way to construct a successful effect, you can achieve lazy construction using the `ZIO.effectTotal` method:
+The `succeed` method takes a by-name parameter. It is the same as the `ZIO.effectTotal` method:
```
adamgfraser, line 26: How about something like?

"The `succeed` method takes a by-name parameter to make sure that any accidental side effects from constructing the value can be properly managed by the ZIO Runtime. However, `succeed` is intended for values which do not have any side effects. If you know that your value does have side effects consider using `ZIO.effectTotal` for clarity."

YingLiu4203: That's a wonderful description that I didn't know before. 

A clarification: by-name parameter is not lazy -- it is like `def`, executed every time when it is used.

YingLiu4203: However, the following sample code doesn't match the correct description -- there is no side effect in `mkString`. We should come with a better sample.

```scala mdoc:silent
lazy val bigList = (0 to 1000000).toList
lazy val bigString = bigList.map(_.toString).mkString("\n")

val s3 = ZIO.effectTotal(bigString)
```

YingLiu4203: changed description and code sample in the following commit

jdegoes:
@YingLiu4203 Thank you, and congratulations on your first contribution!

giphy

Issue #3440, opened 2020-04-23
ZStream: add intersperse


mvillafuertem:
Hi, maybe this option is interesting for users

This is present in Akka Streams

https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/intersperse.html

```
Source(1 to 4).map(_.toString).intersperse("[", ", ", "]").runWith(Sink.foreach(print))
```

It's great when you have a streaming api, and always you would send a valid json

```
Source(jsonTweets).intersperse("[", ", ", "]").runWith(Sink.foreach(print))
```

Thanks!

fagossa:
I would like to take this one

fagossa:
@iravid I got a WIP PR in here https://github.com/zio/zio/pull/3514 . I still got some issues because I'm quite new to the base code.

iravid:
Closed in #3514

PR #3514, opened 2020-05-02
Implement ZStream#intersperse

fagossa:
This is an implementation for `intersperse` (https://github.com/zio/zio/issues/3440).

For example:

```scala
Stream(1, 2, 3, 4).map(_.toString).intersperse("@")
//  "1", "@", "2", "@", "3", "@", "4"
```

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1527,6 +1527,25 @@ abstract class ZStream[-R, +E, +O](
   }
 
   /**
+   * Intersperse stream with provided element similar to <code>List.mkString</code>.
+   */
+  final def intersperse[O1 >: O](middle: O1): ZStream[R, E, O1] =
+    ZStream {
+      for {
+        state  <- ZRef.makeManaged(false)
+        chunks <- self.process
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
Let's make this a transducer instead, right? Then it's just a fold that inserts the element every even one:
```scala
state.modify { flag =>
  os.foldRight(Nil -> flag) { 
    case (o, (out, true)) => (a :: o :: out) -> false
    case (o, (out, false)) => (o :: out) -> true
  }
}
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1526,6 +1526,25 @@ abstract class ZStream[-R, +E, +O](
     }
   }
 
+  /**
+   * Intersperse stream with provided element similar to <code>List.mkString</code>.
+   */
+  final def intersperse[O1 >: O](middle: O1): ZStream[R, E, O1] =
```
iravid, line 1532: @fagossa Could we add a variant that also adds start/end markers? This could simply be `ZStream(start) ++ intersperse(middle) ++ ZStream(end)`.

fagossa: done!

PR #3515, opened 2020-05-02
Rewrite Chunk#dropWhileM via ChunkBuilder

heraklos:
Closes : https://github.com/zio/zio/issues/3494

### Changes in Code
- Rewrite Chunk#dropWhileM via ChunkBuilder
- Change predicate parameter for Chunk#dropWhileM test case to verify not dropping matched elements if there is already unmatched element before

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -164,35 +164,26 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   }
 
   def dropWhileM[R, E](p: A => ZIO[R, E, Boolean]): ZIO[R, E, Chunk[A]] = {
-    val len                                       = self.length
-    var dest: ZIO[R, E, (Boolean, Int, Array[A])] = UIO.succeedNow((true, 0, null.asInstanceOf[Array[A]]))
+    val len     = self.length
+    val builder = ChunkBuilder.make[A]()
+    builder.sizeHint(len)
+    var dest: ZIO[R, E, (Boolean, ChunkBuilder[A])] = IO.succeedNow((true, builder))
 
     var i = 0
     while (i < len) {
       val j = i
       dest = dest.flatMap {
-        case (dropping, skip, array) =>
+        case (dropping, builder) =>
           val a = self(j)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -164,35 +164,26 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   }
 
   def dropWhileM[R, E](p: A => ZIO[R, E, Boolean]): ZIO[R, E, Chunk[A]] = {
-    val len                                       = self.length
-    var dest: ZIO[R, E, (Boolean, Int, Array[A])] = UIO.succeedNow((true, 0, null.asInstanceOf[Array[A]]))
+    val len     = self.length
```
jdegoes, line 167: This whole function body needs to be encased in `ZIO.effectSuspendTotal`. Right now, it is not referentially transparent, which will cause bugs.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -164,35 +164,26 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
...
+    val builder = ChunkBuilder.make[A]()
+    builder.sizeHint(len)
+    var dest: ZIO[R, E, (Boolean, ChunkBuilder[A])] = IO.succeedNow((true, builder))
```
jdegoes, line 170: There is no need to pass the `builder` into the fold, because the builder is a constant outside the loop and can be referred to in the functions below.

heraklos:
@jdegoes  
Thank you so much for your review! I'll fix it.

core/shared/src/main/scala/zio/Chunk.scala
heraklos: Thanks! I got it.
Should I also encase the function bodies of `mapAccumM` in `ZIO.effectSuspendTotal`?

jdegoes: Yes please, the construction of a builder is impure, and those methods must be wrapped in `effectSuspendTotal`. Thank you so much! 🙏

jdegoes:
@heraklos Thank you for your contribution! Excellent work.

Issue #3495, opened 2020-04-30
Add ZTransducer arrow operations


iravid:
ZTransducer is an effectful chunk processing function, and thus forms an arrow. We should add the following combinators on it:
- [x] zipping: `(ZTransducer[A, B], ZTransducer[A, C]) => ZTransducer[A, (B, C)]`
- [ ] ~racing: `(ZTransducer[A, B], ZTransducer[A, C]) => ZTransducer[A, Either[B, C]]`~ looks problematic to implement currently
- [ ] both: `(ZTransducer[A, C], ZTransducer[B, D]) => ZTransducer[(A, C), (B, D)]`
- [ ] choose: `(ZTransducer[A, C], ZTransducer[B, D]) => ZTransducer[Either[A, C], Either[B, D]]`

Part of this issue would be to choose the appropriate names and see whether there are more generalized versions of these (e.g. for choose, a version where the types can unify through subtyping without `Either`)

We should also check for similarities between these implementations and `ZStream#zip/merge`, and figure out if we should move the `ZStream` implementations here.

simpadjo:
In `race`, do we remove the loser from the competition?

Consider:
`t1` output:
 ---B1-----------B2----------
`t2` output:
------------C1---------C2-----

should `t1.race(t2)` emit  
----B1----C1---B2--C2-----

or
----B1-----------B2----------
?

simpadjo:
No one is hungry of good first issues so let me take it. I'll start with `zip`

iravid:
@simpadjo Yes, we should remove the loser from the competition. But! We must not interrupt the loser. That's currently against the ZStream protocol. We should just wait for it to finish, continuing to feed data to the winner in the meantime and making sure to discard its result when/if it does finish.

fagossa:
I would like to take `ZTransducer#both`

simpadjo:
`zip` is done, picking `race` now

simpadjo:
@iravid If implemented naively, result of `race` depends on chunking strategy.
If both sides emitted a value when being fed a huge chunk, there is no way to determine who was the fastest. `choose` will suffer from the same issue.

We can either relax guarantees or feed upstream chunks with elements one-by-one.
What do you think?

simpadjo:
Also taking `choose` since it can be easily implemented in terms of `race`

simpadjo:
@fagossa Hey! Seems that `both` could be implemented using `zip` (like `choose` using `race`).
[...]

PR #3516, opened 2020-05-02
Fix `ZStream#zip`

### LINKED ISSUE #3513: ZStream#zip loses data
Testcase:
```
testM("zip"){
        val s1 = ZStream.fromChunk(Chunk[Int](1,2,3,4,5))
        val s2 = ZStream.fromChunks(Chunk[String]("a"), Chunk[String]("b"), Chunk[String]("c"))
        assertM(s1.zip(s2).runCollect)(equalTo(List((1, "a"), (2, "b"), (3, "c"))))
      }
```
>  List((1,a), (2,b)) did not satisfy equalTo(List((1,a), (2,b), (3,c)))

Implementation doesn't handle properly the case when 2 streams terminate at different moments.

simpadjo:
Closes #3513

A bit messy, improvements are welcome.
There is a chance that it will be deuglified it as a part of #3495 (In case I find out how to unify `ZStream#zip` and `Ztransducer#zip`)

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -2078,22 +2078,18 @@ object ZStreamSpec extends ZIOBaseSpec {
           })
         ),
         suite("zipWith")(
+          testM("zip doesn't pull too much when one of the streams is done") {
+            val l = ZStream.fromChunks(Chunk(1, 2), Chunk(3, 4), Chunk(5)) ++ ZStream.fail("Nothing to see here")
+            val r = ZStream.fromChunks(Chunk("a", "b"), Chunk("c"))
+            assertM(l.zip(r).runCollect)(equalTo(List((1, "a"), (2, "b"), (3, "c"))))
+          },
           testM("zip equivalence with Chunk#zipWith") {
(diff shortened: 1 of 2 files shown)


iravid:
Thanks @simpadjo. The code looks good overall. I don't have the cycles to do a deeper review here so happy to merge this as is and maybe follow up later if I have comments.

Can you fix the Dotty compilation?

simpadjo:
@iravid thank you for the review. Fixed.
Btw, now JS tests run for 30+ minutes. Is it normal?

iravid:
Yes, they've slowed down considerably. I'm not sure why that is but haven't had a chance to delve deeper aside from disabling some slow tests on Scala.js.

iravid:
@simpadjo Thanks for fixing, great catch once again!

adamgfraser:
I think the key driver is we are running the streams tests on Scala.js now. Previously all the stream tests were running only on the JVM. So we've significantly increased the workload (especially with streams using a lot of property based tests) on a platform that way always slower to begin with. We definitely need to do something about this. I think reducing the number of times property based tests are evaluated would be a good start and I can work on that.

iravid:
Oh, that's right. I hadn't realized that. Good point about the number of iterations for the property checks!

Issue #3489, opened 2020-04-30
ZStream.fromQueue creates a singleton chunk stream


iravid:
Follow-up to #3079.

`ZStream.fromQueue` creates a stream of singleton chunks because we have no way of taking multiple elements from a queue and blocking until at least some of the takes are satisfied. Using `takeAll.repeatWhile(_.size == 0)` would create a busy-waiting loop.

We would need to implement `takeBetween(min: Int, max: Int)` methods on Queue which block until at least `min` elements and at most `max` elements can be satisfied. ZStream.fromQueue can use that to grab everything the queue can offer at every pull.

cc @mijicd

mijicd:
I'm working on something similar, as discussed over chat. Tomorrow's the day off, so I'll be back on this.

iravid:
Thanks for putting in the time to do this @mijicd! Just to clarify, this isn’t urgent as it is a performance optimization.

PR #3520, opened 2020-05-03
Add ZQueue.takeBetween

mijicd:
This PR is related to #3489.

I have implemented `takeBetween` via repeated invocations of `takeUpTo`. While this approach isn't the optimal one, it is the simplest one. The "optimal" solution would probably require some form of "smart" takers, i.e., promises aware of requested `min` and `max` values, which would require significant refactoring of queue internals.

Besides efficiency, this implementation might suffer from interruption-related issues.

### DIFF
--- a/core/shared/src/main/scala/zio/ZQueue.scala
+++ b/core/shared/src/main/scala/zio/ZQueue.scala
@@ -109,7 +109,7 @@ trait ZQueue[-RA, -RB, +EA, +EB, -A, +B] extends Serializable { self =>
    * collected.
    */
   final def takeBetween(min: Int, max: Int): ZIO[RB, EB, List[B]] =
-    if (max < min) ZIO.succeedNow(Nil)
+    if (max < min) UIO.succeedNow(Nil)
     else
       takeUpTo(max).flatMap { bs =>
         val remaining = min - bs.length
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -103,6 +103,30 @@ trait ZQueue[-RA, -RB, +EA, +EB, -A, +B] extends Serializable { self =>
...
+    def loop(remaining: Int, collected: Int, acc: List[B]): ZIO[RB, EB, List[B]] =
+      takeUpTo(remaining).flatMap { data =>
+        val len          = data.length
+        val newCollected = collected + len
+        val newAcc       = acc ++ data
+
+        if (newCollected >= min)
+          ZIO.succeedNow(newAcc)
+        else
+          loop(remaining - len, newCollected, newAcc)
```
iravid, line 121: I believe this is a busy-loop, no?

Maybe you could repeat `take` for `min` times then do `takeUpTo(max - min)`?

iravid: Well I admit both of these implementations would suffer from performance so maybe it isn't the greatest idea to expose either of them. Not sure.

mijicd: Both are spinners, yes. I could benchmark them, but I don't think we're going to get any substantial improvements there.

jdegoes: I think a busy loop is not acceptable. If you `takeUpTo` and you still need to take more, I would `take` until you get to the minimum.

Or if you want to get fancy, you could `take` the first one, and then `takeBetween` the ones after that (recursively), such that it would at least _try_ to `takeUpTo` for remaining elements after something was added to the queue.

mijicd: Re-implemented as `takeUpTo + (min - size) * take`.

PR #3521, opened 2020-05-03
Add ZStream.debounce

### LINKED ISSUE #2565: Add debounce support
I brought this up on Discord, but there are more design decisions to be made than I initially expected so having an issue to sort everything out for organization's sake is probably a good idea.

So basically what I was initially was looking for is debounce on ZIO itself. Something like:

```scala
def debounce(waitTime: Duration): URIO[Clock, ZIO[R, E, Option[A]]]
```

But then it seems like you'd also want versions on the ZIO companion object too:

```scala
def debounce1[R, E, A, T](f: T => ZIO[R, E, A], waitTime: Duration): URIO[Clock, T => ZIO[R, E, Option[A]]]
```

and `debounce2`, `debounce3`, etc... which is kind of awkward. Plus a ZIO returning another ZIO isn't very ergonomic in the first place.

Maybe it's better to forget that route and focus just on the ZIO Streams case? Adding debounce on streams generalizes better I think.

There is one additional issue where there are multiple interpretations of what `debounce` should do. For example, lodash and a few other libraries have a "leading" and "trailing" option. I found some diagrams which might be helpful visualizing this:
https://medium.com/@ellenaua/throttle-debounce-behavior-lodash-6bcae1494e03

Do we want to support all these? If not, which one should be the default implementation?

Any other ideas or things to think about?

reibitto:
Resolves #2565

This is my first attempt. I'm not sure if this is a reasonable implementation.

A few things about this:
- I added ZTransducer `headOption` and `lastOption`. Does it makes sense to expose these? I used `lastOption` in `debounce`.
- "Debounce" has multiple interpretations depending on how you want to handle the "leading" and "trailing" cases. I mentioned that in #2565. In any case, I went with what seems like the most common intepretation (`{leading: false, trailing: true}` in lodash terms). This seemed like the most useful default to me. If we want to support `leading = true` I guess it's just a matter of peeling the head element, returning that immediately and then `debounce` the rest as normal?
- Writing concise tests for this is hard. I believe I'm running into the same race condition problems as I did with `groupedWithin`. In the end I just took @mschuwalow's solution in #2631 and used it for `debounce` too. But it might be better to generalize that piece of code.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -13,6 +13,7 @@ import zio.test.Assertion._
 import zio.test.TestAspect.flaky
 import zio.test._
 import zio.test.environment.TestClock
+import zio.clock.Clock
 
 object ZStreamSpec extends ZIOBaseSpec {
   import ZIOTag._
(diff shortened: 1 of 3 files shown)


streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -1345,7 +1386,7 @@ object ZStreamSpec extends ZIOBaseSpec {
           assertM(ZStream(1, 2, 3, 4).grouped(2).runCollect)(equalTo(List(List(1, 2), List(3, 4))))
         ),
         suite("groupedWithin")(
-          testM("999") {
+          testM("group in multiple buckets after waitTime passed") {
```
iravid, line 1389: 😅

reibitto:
Not sure what that 1 failure was about. Maybe unrelated flakiness? I added `@@ TestAspect.nonFlaky` locally just to see and it passes fine. 🤔 

Anyway, I added more tests and refactored the already-existing race condition prevention code to be more generic. I re-used that in multiple tests rather than copy/pasting that everywhere.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -741,6 +741,13 @@ abstract class ZStream[-R, +E, +O](
   final def crossRight[R1 <: R, E1 >: E, O2](that: ZStream[R1, E1, O2]): ZStream[R1, E1, O2] =
     (self crossWith that)((_, o2) => o2)
 
+  /**
+   * Only emits elements after `waitTime` has passed. If another element in the source stream is
+   * produced before `waitTime` has passed, the previous element is will be dropped.
```
luis3m, line 746: I guess that last `is` is a typo

reibitto: Oops. Thanks for catching that.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2454,4 +2477,36 @@ object ZStreamSpec extends ZIOBaseSpec {
         }
       )
     ) @@ TestAspect.timed
+
+  trait ChunkCoordination[A] {
```
iravid, line 2481: I love this!!

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -208,6 +209,49 @@ object ZStreamSpec extends ZIOBaseSpec {
...
+            assertWithChunkCoordination(List(Chunk(1, 2), Chunk(3, 4), Chunk.single(5))) { c =>
+              val stream = ZStream
+                .fromQueue(c.queue)
+                .collectWhileSuccess
+                .flattenChunks
+                .debounce(1.second)
+                .tap(_ => c.proceed)
+
+              assertM(for {
+                f      <- stream.runCollect.fork
```
iravid, line 223: If all testing patterns converge to "fork the stream, offer-advance-next a bunch of times then collect the results", it may be simpler to initialize the stepper with a `List[Assertion[A]]` that is tested against the incoming chunks.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -208,6 +209,49 @@ object ZStreamSpec extends ZIOBaseSpec {
             assertM(ZStream(1, 2, 3).aggregate(t).runCollect.either)(isLeft(equalTo(fail)))
           }
         ),
+        suite("debounce")(
+          testM("drop earlier elements within waitTime") {
+            assertWithChunkCoordination(List(Chunk(1, 2), Chunk(3, 4), Chunk.single(5))) { c =>
+              val stream = ZStream
+                .fromQueue(c.queue)
+                .collectWhileSuccess
```
iravid, line 217: We can extract this bit out into the stepper - have it take a function `ZStream[A] => ZStream[B]` that is applied to `ZStream.fromQueue` and then fed into `.tap(verifyAssertions)`.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
iravid: We can circle back to this when someone actually needs it, but I can see this being useful in some non-test-uses :-)

PR #3526, opened 2020-05-04
Revert debounce

reibitto:
Reverting the debounce implementation since it isn't quite right (needs to reset `waitTime` when new elements arrive). I'll submit another PR with the new implementation. It's not possible to do this with ZTransducer.

I also renamed ZTransducer headOption and lastOption to be consistent with ZSink.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -539,7 +539,7 @@ object ZTransducer {
   /**
    * Creates a transducer that returns the first element of a chunk, if it exists.
    */
-  def headOption[O]: ZTransducer[Any, Nothing, O, Option[O]] =
+  def head[O]: ZTransducer[Any, Nothing, O, Option[O]] =
     foldLeft[O, Option[O]](Option.empty[O]) {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -539,7 +539,7 @@ object ZTransducer {
   /**
    * Creates a transducer that returns the first element of a chunk, if it exists.
    */
-  def headOption[O]: ZTransducer[Any, Nothing, O, Option[O]] =
+  def head[O]: ZTransducer[Any, Nothing, O, Option[O]] =
```
iravid, line 542: Another thing I realized is that this transducer will emit the very first element of the stream; not the first element of every chunk.

Same for `last`.

iravid: These are useful transducers so let's just fix the scaladoc to reflect what they do.

reibitto: Ah, yeah, that's true. Not sure why I messed that up when even one of my debounce tests was explicitly checking for that. 😅

I updated the wording.

PR #3530, opened 2020-05-04
Better compile errors for pattern guards in for-comprehension

simpadjo:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3497,6 +3498,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
       self.provideLayer[E1, R0, R0 with R1](ZLayer.identity[R0] ++ layer)
   }
 
+  @implicitNotFound(
```
jdegoes, line 3501: We discovered you can "Inline" this annotation to avoid the need for a new `CanFilter` class.

jdegoes:
@simpadjo Great idea! One comment on how to do it without the new class, then good to merge!

core/shared/src/main/scala/zio/ZIO.scala
simpadjo: Like that?
```
object Evidences {
  @implicitNotFound("provideCustomLayer expect all layers except those provided by ZIO (ZEnv) instead it got $R1. If you intend to provide some of layers only then use provideSomeLayer instead")
  type ProvideCustomLayer[-R1, +R] = ZEnv with R1 <:< R
}
```

jdegoes: Even more! Inline where the function is:

```scala
@implicitNotFound("") implicit ev: ...
```

Try that!

neko-kai: Note that this applies to 2.13+ only (which may be fine anyway)

simpadjo: Yeah, I tried it and it doesn't work on default scala version. 
Is there any runtime overhead? If no - I'd go with a bulky solution.
If yes - we can put extensions in version-specific directories

simpadjo: @jdegoes @neko-kai I made implementation version-specific. Have no idea what I'm doing but it compiles!

adamgfraser:
I'm inclined to think with the version specific files this is more complexity and maintenance burden at this point. What if we just do something like?

```scala
@implicitNotFound("Pattern guards are only supported when the error type is a supertype of NoSuchElementException. However, your effect has ${E} for the error type.")
sealed trait CanFilter[+E] {
  def apply(t: NoSuchElementException): E
}

object CanFilter {
  implicit def canFilter[E >: NoSuchElementException]: CanFilter[E] =
    new CanFilter[E] {
      def apply(t: NoSuchElementException): E = t
    }
}
```

simpadjo:
@adamgfraser @jdegoes Let's agree on something or flip a coin )

jdegoes:
@simpadjo I'm on board with @adamgfraser's suggestion! 🍀

simpadjo:
Applied suggestions by @adamgfraser . Looks much cleaner now!
Also lots of confusion comes from using type annotations and tuple deconstruction.
Would be great to warn about it specifically, but don't know how to formulate it shortly ~~and w/o dissing the compiler too much~~.
Js build is failing due to unrelated problems on master.

simpadjo:
@adamgfraser I rebased it. Let's merge it then?

PR #3535, opened 2020-05-04
Fix #3531 Document auto-composition of ZIO Layers using DIStage

### LINKED ISSUE #3531: Auto-composition of ZIO Layers
Not sure if that is possible, but I want to propose an idea.
It would be really nice if one could provide a bunch of required layers and zio would automatically figure out how to wire them coherently together.

Right now you have to provide all the required layers to other layers using `>>>`. You do this for layers that depend on other layers that depend on other layers, until at most there is ZEnv left. So you can provide a list of "resolved" layers to the app.

Maybe it is possible through the marcos or in some other way to build that graph of dependencies and satisfy dependencies automatically step by step.
I can imaging having a map of lazy values (keys are resulting layer types and values are layers themselves). Then traversing this map multiple times until all the dependencies are satisfied and we can build a coherent environment. Starting from services that don't have requirements, building them first. And then feeding them where they are required.
Having it in this "dynamic way" can enable us building more precise error messeges where exactly which dependencies are missing. Instead of unwinding a big error message. For which I built a tool because error can become really huge and incomprehensible. But that is a different topic.

neko-kai:


docs/howto/use_modules_and_layers.md
```diff
@@ -391,4 +391,36 @@ lazy val all: ZLayer[Any, Nothing, Baker with Ingredients with Oven with Dough w
...
-And if you do build your dependency graph more explicitly you can be confident that layers used in multiple parts of the dependency graph will only be created once due to memoization and sharing.
\ No newline at end of file
+And if you do build your dependency graph more explicitly you can be confident that layers used in multiple parts of the dependency graph will only be created once due to memoization and sharing.
+
+# Auto-wiring ZIO Layers
+
+For big enough programs, explicitly sequencing `ZLayer`s
+https://izumi.7mind.io/latest/release/doc/distage
+
+`ModuleDef` is a set of derivation rules - it declares what `ZLayer` should be used to create a given Service.
```
vitaliis, line 401:
```suggestion
`ModuleDef` contains a set of derivation rules. It defines relation between a given Service and a `ZLayer` definition for creating it.
```

vitaliis:
@neko-kai thanks for adding this to the documentation! I see it is still a draft, but I may already leave couple of comments.

vitaliis:
I would also add a link to the `fromHas` binding: https://izumi.7mind.io/latest/release/doc/distage/basics.html#zio-has-bindings

jdegoes:
@neko-kai Would love to get this in! Please update and re-open when ready.

PR #3541, opened 2020-05-05
Add Operators

### LINKED ISSUE #3538: Add <+> operator for `orElseEither`
Every data type that has `orElseEither` should get `<+>` as a symbolic alias for the named method.

This is the dual of `<*>` (`zip`).

### LINKED ISSUE #3539: Add <|> operator for `raceEither`
Every data type that has `raceEither` should get `<|>` as a symbolic alias for the named method.

This is the dual of `<&>` (`zipPar`).

adamgfraser:
Resolves #3538. Resolves #3539. Note that I think there was an error with the type signature of `Fiber#orElse` where it was requiring the error type of the right hand side to be a super type of the error type of the left hand side instead of allowing it to be independent.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/FiberSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberSpec.scala
@@ -32,7 +32,7 @@ object FiberSpec extends ZIOBaseSpec {
           } yield assert(value)(equalTo(update))
         },
         testM("`orElse`") {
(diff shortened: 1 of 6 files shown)


jdegoes:
This is great! The other thing you could do would be to add `<>` as `orElse`. It's done that way in ZIO but not elsewhere.

adamgfraser:
Will do.

Issue #3482, opened 2020-04-29
window and watermark function in ZIO Streams


michalwarecki:
Hi guys,

Recently I've been dealing with ZIO Streams and in my use case I found two missing functionalities in ZIO Streams:
- watermark - removing incoming events older than provided duration,
- window - grouping events by sliding time window.

I have some time to implement such contribution but I need your input/review.
The first thing is answer to question: are such functions even needed in ZIO Streams?
The second is to review design of both functions:

`def window(windowDuration: Duration, tick: Duration, timestamp: A => Instant, buffer: Int = 16): ZStream[R with Clock, E, List[A]]`

- `windowDuration` - sliding window duration,
- `tick` - how frequent consumer will receive grouped events,
- `timestamp` - time in UTC representing each event (question: should we use Instant or Long as a timestamp?),
- `buffer` - event buffer

`def watermark(delayThreshold: Duration, timestamp: A => Instant): ZStream[R with Clock, E, A]`

- `delayThreshold` - watermark duration (events older than `currentTime - delayThreshold` will be dropped),
- `timestamp` - time representing each event

simpadjo:
Hey,
I think having something like this would be great.
Couple of comments: 
1) A major pull request with a complete redesign zio-streams is about to be merged. Makes sense to wait a bit with coding.
2) Would be useful be able to use current time as an event timestamp as well

michalwarecki:
Thanks for the comment @simpadjo !

1. I see the changes are already merged so I'll start prototyping today.
2. Do you mean that by default it should be current time?

simpadjo:
Hard to say how final API would look like. I'm talking about usecases:
1) extract time from a stream element
2) use a "natural" timestamp

There could be few ways to achieve this including:
a) changing timestamp timestamp type to  `A => ZIO[R, E, Instant]`
b) asking a caller to zip the stream with the current time themself

iravid:
Hello @michalwarecki, sorry I did not chime in yet. I admit I find it hard to see a family of combinators around watermarks directly in `ZStream` as that is a fairly domain-specific usecase, whereas most ZStream combinators are very generic.

For example, #3542 is a monomorphic version of `filterM`.

But I am open to being convinced. Pinging other contributors for some more opinions: @vasilmkd @regiskuckaertz @mschuwalow

As a side-note, we will be picking up the development of https://github.com/zio/zio-analytics soon(tm). That will include full-fledged support for watermarked streams.

vasilmkd:
[...]

PR #3542, opened 2020-05-05
Watermark implementation

michalwarecki:
Hi,

Regarding #3482 I've started with much simpler function which is watermark. I've created draft implementation of the feature. Let me know what do you think about that and what could be improved/changed.


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1435,6 +1435,34 @@ abstract class ZStream[-R, +E, +O](
...
+  def watermarkM[R1 <: R, E1 >: E](
+    delayThreshold: Duration,
+    timestamp: O => ZIO[R1, E1, java.time.Instant]
+  ): ZStream[R1 with Clock, E1, O] =
+    filterM(event =>
+      for {
+        currentTime        <- clock.currentTime(ju.concurrent.TimeUnit.MILLISECONDS)
+        eventTimestamp     <- timestamp(event)
+        watermarkPredicate <- ZIO.succeedNow(eventTimestamp.toEpochMilli >= (currentTime - delayThreshold.toMillis))
+      } yield watermarkPredicate
```
simpadjo, line 1463: Looks like it could be simplified by  `yield (eventTimestamp.toEpochMilli >= (currentTime - delayThreshold.toMillis))`

michalwarecki: thanks, I fixed it

iravid:
Closing following the discussion in #3482.

PR #3548, opened 2020-05-07
Move `chunkN` from Transducer to Stream

### LINKED ISSUE #3522: ZTransducer#chunkN might emit empty chunks
...and explicitly says it :)
```
buffered
                .modify(buf => (if (buf.isEmpty) Push.emit(Chunk.empty) else Push.emit(buf)) -> Chunk.empty)
                .flatten
```

simpadjo:
Closes #3522


simpadjo:
@iravid thank you! Could you please merge?

PR #3551, opened 2020-05-07
Add Transducer#zip

simpadjo:
Part of #3495

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -89,7 +89,7 @@ abstract class ZTransducer[-R, +E, -I, +O](
                         stream.internal.Utils.zipChunks(left, right, f)
                     }
               (emit, newExcess) = res
-              _                 <- ref.set(in.fold(Left(Chunk.empty): Either[Chunk[O], Chunk[O2]])(newExcess))
+              _                 <- ref.set(in.fold(Left(Chunk.empty): Either[Chunk[O], Chunk[O2]])(_ => newExcess))
             } yield {
               emit
             }
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -89,7 +89,7 @@ abstract class ZTransducer[-R, +E, -I, +O](
                         stream.internal.Utils.zipChunks(left, right, f)
                     }
               (emit, newExcess) = res
-              _                 <- if (in.isDefined) ref.set(newExcess) else ZIO.unit
+              _                 <- ref.set(in.fold(Left(Chunk.empty): Either[Chunk[O], Chunk[O2]])(newExcess))
             } yield {
               emit
             }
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -32,6 +32,76 @@ abstract class ZTransducer[-R, +E, -I, +O](
...
+        p1  <- self.push
+        p2  <- that.push
+        push = (in: Option[Chunk[I1]]) => {
+          ref.get.flatMap { excess =>
+            for {
+              res <- p1(in).zipWithPar(p2(in)) {
+                      case (leftUpd, rightUpd) => {
+                        val (left, right) = excess.fold(l => (l ++ leftUpd, rightUpd), r => (leftUpd, r ++ rightUpd))
+                        stream.internal.Utils.zipChunks(left, right, f)
+                      }
```
iravid, line 90: Sorry to nitpick. Can you get rid of the redundant braces? Here and in the yields below:
```suggestion
                      case (leftUpd, rightUpd) =>
                        val (left, right) = excess.fold(l => (l ++ leftUpd, rightUpd), r => (leftUpd, r ++ rightUpd))
                        stream.internal.Utils.zipChunks(left, right, f)
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -32,6 +32,76 @@ abstract class ZTransducer[-R, +E, -I, +O](
...
+                    }
+              (emit, newExcess) = res
+              _                 <- if (in.isDefined) ref.set(newExcess) else ZIO.unit
```
iravid, line 93: Prefer Option#fold or just straight up pattern matching over the use of `Option#isDefined`. Beyond that, for this case, we should take care to empty the state reference when the `Option` is a `None`:
```suggestion
              _                 <- ref.set(in.fold(Left(Chunk.empty): Either[Chunk[O], Chunk[O2]])(newExcess))
```

simpadjo: Hm, I thought that if `None` is received, subsequent invocations would use a fresh `push` instance. Could you please clarify that?

iravid: No, transducers are not restarted during stream processing.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -32,6 +32,75 @@ abstract class ZTransducer[-R, +E, -I, +O](
...
+    type State = Either[Chunk[O], Chunk[O2]]
+    ZTransducer {
+      for {
+        ref <- ZRef.make[State](Left(Chunk.empty)).toManaged_
+        p1  <- self.push
+        p2  <- that.push
+        push = (in: Option[Chunk[I1]]) => {
+          ref.get.flatMap { excess =>
+            for {
+              res <- p1(in).zipWithPar(p2(in)) {
```
regiskuckaertz, line 86: With this version it is possible that one slow stream will block a fast stream, as it should, but chunks will accumulate in memory potentially until collapse. I think it's worth pulling from one stream only when there's an excess, perhaps with a heuristic on the size of the excess.

regiskuckaertz: Hmm, no, I misread. Ha, I take it back 😅

PR #3554, opened 2020-05-07
Add Schedule noDelay and run

### LINKED ISSUE #3468: Simulate a Schedule
I think it would be helpful to be able to take any arbitrary Schedule and "simulate" it to get its output. For example, say I have:

```scala
val retryPolicy = Schedule.exponential(1.second) <* Schedule.recurs(5)
```

I'm imagining a function that you could use to answer questions like, "What would the output of this schedule be given this input?" Another important thing this function would do is give you the answer without actually sleeping (I think you might need a Clock like zio-test's TestClock to do this).

A function like this would open up possibilities to use `Schedule` in more situations than you can now (such as in distributed systems).

To give one real-world example:

I have multiple message queue consumers that processeses messages and will retry them later if they fail (possibly on a completely different consumer node next time). Each time the message comes through again, there's a `ReceiveCount` header. If the `ReceiveCount` is 4, I want to know the 4th value in the retryPolicy Schedule so that I can tell the message queue "pop it back in the queue to show up after X time has passed".

It would be cool to leverage ZIO's Schedule to create any kind of schedule that I can imagine rather than write my own schedule implementation just for this specific purpose.

reibitto:
Resolves #3468

This skips any sleep calls so that we can simulate a schedule and get its output instantaneously.

One thing I was considering was making `input` a LazyList so that passing in infinite input was possible if you wanted to (in that case you would solely rely on your Schedule's end condition). Not sure whether that's needed or not though.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -115,6 +115,20 @@ object ScheduleSpec extends ZIOBaseSpec {
         } yield assert(v)(equalTo(6)) && assert(finalizerV.isDefined)(equalTo(true))
       }
     ),
+    suite("Simulate a schedule")(
+      testM("without timing out") {
+        val schedule  = Schedule.exponential(1.minute)
+        val scheduled = schedule.simulate(List.fill(5)(()))
+        val expected  = List(1.minute, 2.minute, 4.minute, 8.minute, 16.minute)
+        assertM(scheduled)(equalTo(expected))
+      },
+      testM("respect Schedule.recurs even if more input is provided than needed") {
+        val schedule  = Schedule.recurs(2) && Schedule.exponential(1.minute)
+        val scheduled = schedule.simulate(1 to 10)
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
```diff
@@ -115,6 +115,20 @@ object ScheduleSpec extends ZIOBaseSpec {
         } yield assert(v)(equalTo(6)) && assert(finalizerV.isDefined)(equalTo(true))
       }
     ),
+    suite("Simulate a schedule")(
+      testM("without timing out") {
+        val schedule  = Schedule.exponential(1.minute)
+        val scheduled = schedule.simulate(List.fill(5)(()))
+        val expected  = List(1.minute, 2.minute, 4.minute, 8.minute, 16.minute)
+        assertM(scheduled)(equalTo(expected))
+      },
```
simpadjo, line 124: Maybe explicitly add `@@ timeout(1.seconds)` ?

reibitto:
That's a really good idea. I like the idea of separating these into 2 methods.

I made the changes and re-used the `run` method in all of ScheduleSpec.

PR #3556, opened 2020-05-07
Remove ZConduit

### LINKED ISSUE #3547: Delete ZConduit
Important before release: delete ZConduit (for now, at least) and verify the methods defined on it are available on ZStream/ZTransducer/ZSink.

iravid:
Resolves #3547 

Might be re-added later if we find a good way to use it.


regiskuckaertz:
Oooh yeah it makes sense.

PR #3557, opened 2020-05-08
add ZTransducer#race

simpadjo:
Part of #3495

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -171,6 +171,19 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
     }
   }
 
+  /**
+   * {@link #raceBoth}, but doesn't restrict input type of `that`.
+   */
+  def choose[R1 <: R, E1 >: E, I1, O1](
+    that: ZTransducer[R1, E1, I1, O1]
+  ): ZTransducer[R1, E1, Either[I, I1], Either[O, O1]] = {
+    val lefts: ZTransducer[R1, E1, Either[I, I1], I] =
+      ZTransducer.identity[Either[I, I1]].filter(_.isLeft).map(_.left.get)
+    val rights: ZTransducer[R1, E1, Either[I, I1], I1] =
+      ZTransducer.identity[Either[I, I1]].filter(_.isRight).map(_.right.get)
+    (lefts >>> self).raceBoth(rights >>> that)
+  }
+
 }
 
 object ZTransducer {
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -171,22 +171,6 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
       }
     }
   }
-
-  /**
-   * {@link #raceBoth}, but doesn't restrict input type of `that`.
-   */
-  def choose[R1 <: R, E1 >: E, I1, O1](
-    that: ZTransducer[R1, E1, I1, O1]
-  ): ZTransducer[R1, E1, Either[I, I1], Either[O, O1]] = {
-    def impossible = throw new RuntimeException("impossible")
-    val lefts: ZTransducer[R1, E1, Either[I, I1], I] =
-      ZTransducer.identity[Either[I, I1]].filter(_.isLeft).map(_.left.getOrElse(impossible))
-    @com.github.ghik.silencer.silent("deprecated")
-    val rights: ZTransducer[R1, E1, Either[I, I1], I1] =
-      ZTransducer.identity[Either[I, I1]].filter(_.isRight).map(_.right.getOrElse(impossible))
-    (lefts >>> self).raceBoth(rights >>> that)
-  }
-
 }
 
 object ZTransducer {
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -1,7 +1,7 @@
 package zio.stream
 
 import scala.collection.mutable
-import zio.{ Chunk, _ }
+import zio._
 
 // Contract notes for transducers:
 // - When a None is received, the transducer must flush all of its internal state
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -98,6 +97,95 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+    }
+  }
+
+  /**
+   * {@link #raceBoth}, but doesn't restrict input type of `that`.
+   */
+  def choose[R1 <: R, E1 >: E, I1, O1](
+    that: ZTransducer[R1, E1, I1, O1]
+  ): ZTransducer[R1, E1, Either[I, I1], Either[O, O1]] = {
+    def impossible = throw new RuntimeException("impossible")
```
simpadjo, line 180: I'm sorry :(

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -98,6 +97,95 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+    val lefts: ZTransducer[R1, E1, Either[I, I1], I] =
+      ZTransducer.identity[Either[I, I1]].filter(_.isLeft).map(_.left.getOrElse(impossible))
+    @com.github.ghik.silencer.silent("deprecated")
```
simpadjo, line 183: Code with `Either` is hard to cross-compile w/o warnings

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -98,6 +97,95 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+          newState._2
+        }
+      }
+    }
+  }
+
+  /**
+   * {@link #raceBoth}, but doesn't restrict input type of `that`.
+   */
+  def choose[R1 <: R, E1 >: E, I1, O1](
```
simpadjo, line 177: `choose` is very similar to `race`.
There were 3 options:
1) implement `race` in terms of `choose` w performance overhead for `race`
2) implement `choose` in terms of `race` w [bigger] performance overhead for `choose`
3) copypaste the implementation.

I went with 2.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -98,6 +98,95 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+    that: ZTransducer[R1, E1, I1, O1]
+  ): ZTransducer[R1, E1, Either[I, I1], Either[O, O1]] = {
+    def impossible = throw new RuntimeException("impossible")
+    val lefts: ZTransducer[R1, E1, Either[I, I1], I] =
+      ZTransducer.identity[Either[I, I1]].filter(_.isLeft).map(_.left.getOrElse(impossible))
+    @com.github.ghik.silencer.silent("deprecated")
+    val rights: ZTransducer[R1, E1, Either[I, I1], I1] =
+      ZTransducer.identity[Either[I, I1]].filter(_.isRight).map(_.right.getOrElse(impossible))
```
iravid, line 186: Please implement `ZTransducer#collect` so you won't have to use the partial methods here.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -98,6 +98,95 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+
+    ZTransducer {
+      for {
+        ref <- ZRef.make[State](State(0, 0)).toManaged_
+        p1  <- self.push
+        p2  <- that.push
+      } yield { (in: Option[Chunk[I1]]) =>
+        for {
+          st       <- ref.get
+          newState <- p1(in).zipWithPar(p2(in)) { case (lefts, rights) => updateState(st, lefts, rights) }
```
iravid, line 166: This will both transducers in parallel and only emit when both of them are done. I think that's a bit misleading for a `race` operation, no?

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
iravid: `choose` is not a racing operation, but a branching operation. It says, push `Left` values to the left transducer and push `Right` values to the right transducer. The implementation should split the incoming chunks into chunks of consecutive `Left` or `Right` values, sequentially push those to the appropriate transducer and re-assemble the resulting chunk.

So I think you won't need to use `race` here, unless I'm missing something.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
simpadjo: Yeah. Good catch. I should add a testcase with a never-emitting transducer.

simpadjo: I added a test where one of transducers never emits and this test passes.
Here I wait for both sides to process a single chunk (resulting in potentially empty outputs).
It doesn't wait for both sides to actually emit a value.

Or do you mean we should go further and interrupt the last push of the slowest transducer?

iravid: @simpadjo your test uses `filter` which emits empty chunks. Try using mapM with ZIO.never.

simpadjo: Racing with such transducers can't work, as far as I understand. 
Transducers process the input sequentially. So if we don't want to wait for the result of `never`, me must stage incoming chunks. But this would lead to space leaks. 
Such optimization would work for the very last push tho. 
If you think that this is beneficial, I would be happy to explore.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
simpadjo: I removed `choose` from this PR since it's getting complicated. Will push it later separately

iravid:
@simpadjo I spent some time thinking about this as well and I think you're right - it is impossible to implement without either always waiting for both transducers to finish (which isn't really racing...) or interrupting the losing transducer which isn't safe yet.

Before we can implement this combinator, we should first explore if we can make our transducers interruption safe. This means that their state must not be corrupted if we interrupt a call to `push`. It's fine to lose the elements (because they got interrupted), but we need to decide what to do about the state. I have a hunch this hovers between "impossible without STM" and "very bad for performance", so I wouldn't get my hopes up.

What do you think?

iravid:
Btw I think `choose` is still possible to implement. That one doesn't require racing.

simpadjo:
@iravid perhaps valid usecases for `race` can be covered by `sink1.race(sink2).toTransducer`

PR #3560, opened 2020-05-08
Add someOrElse to ZSTM, ZIO and ZManaged

valenterry:
This PR adds `.someOrElse` which behaves similar to `.getOrElse` of `Option` but in the context of `ZSTM`, `ZIO` or `ZManaged`

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -770,6 +770,12 @@ final class ZSTM[-R, +E, +A] private[stm] (
     )
 
   /**
+   * Extracts the optional value, or returns the given 'default'.
+   */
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -769,6 +769,12 @@ final class ZSTM[-R, +E, +A] private[stm] (
       _.fold[ZSTM[R, Option[E], B]](ZSTM.fail(Option.empty[E]))(ZSTM.succeedNow)
     )
 
+  /**
+   * Extracts the optional value, or returns the given 'default'.
+   */
+  def someOrElse[B, B1 <: B](default: => B1)(implicit ev: A <:< Option[B]): ZSTM[R, E, B] =
```
adamgfraser, line 775: I don't think you need the `B1` type parameter here.

PR #3561, opened 2020-05-08
WIP: more simple `zio.App`

### LINKED ISSUE #2153: Provide a simple method to run a program to completion without explicitly folding
In most examples, the program is run at the entry point like `program.fold(_ => 1, _ => 0)`. This gets repetitive, and also provides a source of confusion for Scala developers not familiar/comfortable with folding.

Perhaps a simpler method, like `execute`, that would fold the program with 1 and 0 as defaults? Perhaps even a `execute(errorCode: Int, exitCode: Int)` overload?

simpadjo:
closes #2153

People (me including) complain about messing with exit codes and environments.
Here is yet another attempt to solve it.

As far as I understand, problem with approaches in the linked ticket is that we don't want to pollute `ZIO` api with helper methods which don't make sense outside of the main class.

### DIFF
--- a/core/jvm/src/main/scala/zio/App.scala
+++ b/core/jvm/src/main/scala/zio/App.scala
@@ -37,7 +37,7 @@ package zio
  * }
  * }}}
  */
-trait App extends BootstrapRuntime {
+trait App extends BootstrapRuntime with AppOps {
 
   /**
    * The main function of the application, which will be passed the command-line
--- a/core/jvm/src/main/scala/zio/AppExample.scala
+++ b/core/jvm/src/main/scala/zio/AppExample.scala
@@ -1,0 +1,9 @@
+package zio
+
+object AppExample extends App {
+  override def run(args: List[String]): ZIO[Any, Nothing, Int] = {
+    ZIO.succeed("xxx").tap(e => {
+      zio.console.putStrLn(e)
+    }).createValueForBusiness
+  }
+}
--- a/core/shared/src/main/scala/zio/AppOps.scala
+++ b/core/shared/src/main/scala/zio/AppOps.scala
@@ -16,17 +16,17 @@
 
 package zio
 
-trait App extends BootstrapRuntime with AppOps {
+import scala.language.implicitConversions
 
-  /**
-   * The main function of the application, which will be passed the command-line
-   * arguments to the program.
-   */
-  def run(args: List[String]): ZIO[ZEnv, Nothing, Int]
+trait AppOps {
+  self: BootstrapRuntime =>
 
-  /**
-   * The Scala main function, intended to be called only by the Scala runtime.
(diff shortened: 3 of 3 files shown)


adamgfraser:
I'm not sure this helps us. It seems like there are two potential issues here, the environment and the exit code.

Regarding the environment, I think we have to explain the environment at some point and having it be part of the signature of `run` is a relatively good way to introduce it in a simplified way. Using this other implicit syntax method that does it just feels like it makes it even more mysterious.

On the exit code, maybe we could simplify to just doing the folding to `1` or `0` ourselves depending on if anyone is using other exit codes but I do think there is some educational value in getting people started with some error handling early.

I would suggest that the real problem here is the lack of a good "Hello World" example in the documentation. If you go to the microsite we don't get to actually running a program until the "Running Effects" section of the overview. I think if the very first section was "getting started" and walked people through a simple "Hello World" app that would go a long way toward solving this problem.

simpadjo:
@adamgfraser Hi! Thanks for the comment! Since this PR haven't attracted any interest, my view on usability is not shared by the community. Let me still answer on your concerns.

1. I'm not proposing changing the signature of `main`. Just adding an extension method to `ZIO`. So a user is in 1-click distance from understanding how the magic works. 

2. Teaching is important, but I prefer to learn 1 thing at a time. 
Zio has at least 2 concepts that requires understanding and mastering for an average dev: effects-as-values and Reader monad.
I'm using Zio mostly because of bifunctor IO, and currently prefer passing arguments to functions directly. Have yet to dig into `ZLayer`. 
Perhaps I'm wrong and I should be using `R` extensively, but that's the point of FP: I don't have to be right to build working programs.

adamgfraser:
@simpadjo Definitely agree that you shouldn't be forced to use the environment extensively, but to some extent at least basic use of it is going to be inevitable (e.g. in how using any methods like `delay` adds a `Clock` dependency). At least in my experience you can say "this is the capabilities that you get to use in your effect, for App default implementations of all the core services are provided for you" and get people going initially.

simpadjo:
Ok, should then the linked ticket be closed?

jdegoes:
How about something like:

```scala
def asExitCode: ZIO[R, Nothing, Int] = 
  self.foldCauseM(cause => UIO(println(cause.prettyPrint)) as 1, _ => 0)
```

So you could do:

```scala
def run(args: List[String]) = myApp.asExitCode
```

Doesn't really deal with the environment part, but ever since ZLayer, it is my impression (possibly incorrect) that the ZIO ecosystem is moving to use it more and more (though it definitely should not be required).

adamgfraser:
Sounds great!

simpadjo:
Closing in favor of https://github.com/zio/zio/pull/3583

Issue #3493, opened 2020-04-30
Implement additional variants for ZStream#timeout


iravid:
The only variant implemented currently fails the stream with an interruption. This is arguably unexpected because we could have just ended the stream.

We should implement variants of timeout that mirror the ones on ZIO:
- `timeout(Duration)` should end the stream when a pull times out
- `timeoutErr(e: E)` should fail the stream with `E` when a pull times out
- `timeoutTo(ZStream)` should switch over to another stream when a pull times out

simpadjo:
@iravid close?

iravid:
Closed in #3562

PR #3562, opened 2020-05-09
Add ZStream timeout variants

luis3m:
@iravid as suggested here https://github.com/zio/zio/issues/3493

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2528,16 +2528,66 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     }
 
   /**
-   * Interrupts the stream if it does not produce a value after d duration.
+   * Ends the stream if it does not produce a value after d duration.
    */
   final def timeout(d: Duration): ZStream[R with Clock, E, O] =
     ZStream[R with Clock, E, O] {
+      for {
+        timeout <- Ref.make(false).toManaged_
+        next    <- self.process
+        pull = timeout.get.flatMap {
+          if (_) Pull.end
+          else
+            next.timeout(d).flatMap {
+              case Some(a) => Pull.emit(a)
+              case None    => timeout.set(true) *> Pull.end
+            }
+        }
+      } yield pull
+    }
+
+  /**
+   * Fails the stream with given error if it does not produce a value after d duration.
+   */
+  final def timeoutError[E1 >: E](e: E1)(d: Duration): ZStream[R with Clock, E1, O] =
+    timeoutErrorCause(Cause.fail(e))(d)
+
+  /**
+   * Halts the stream with given cause if it does not produce a value after d duration.
+   */
+  final def timeoutErrorCause[E1 >: E](cause: Cause[E1])(d: Duration): ZStream[R with Clock, E1, O] =
+    ZStream[R with Clock, E1, O] {
       self.process.map { next =>
         next.timeout(d).flatMap {
-          case Some(a) => ZIO.succeedNow(a)
-          case None    => ZIO.interrupt
+          case Some(a) => Pull.emit(a)
+          case None    => Pull.halt(cause)
         }
       }
+    }
+
+  /**
+   * Switches the stream if it does not produce a value after d duration.
+   */
+  final def timeoutTo[R1 <: R, E1 >: E, O2 >: O](
+    d: Duration
+  )(that: ZStream[R1, E1, O2]): ZStream[R1 with Clock, E1, O2] =
+    ZStream[R1 with Clock, E1, O2] {
+      for {
+        currStream   <- Ref.make[ZIO[R1, Option[E1], Chunk[O2]]](Pull.end).toManaged_
+        switchStream <- ZManaged.switchable[R1, Nothing, ZIO[R1, Option[E1], Chunk[O2]]]
+        switched     <- Ref.make(false).toManaged_
+        _            <- switchStream(self.process).flatMap(currStream.set).toManaged_
+        pull = {
+          val effect       = currStream.get.flatten
+          lazy val switch  = switched.set(true) *> switchStream(that.process).flatMap(currStream.set) *> go
+          lazy val timeout = effect.timeout(d).flatMap(_.map(Pull.emit).getOrElse(switch))
+
+          lazy val go: ZIO[R1 with Clock, Option[E1], Chunk[O2]] =
+            switched.get.flatMap(if (_) effect else timeout)
+
+          go
+        }
+      } yield pull
     }
 
   /**
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -2145,18 +2145,84 @@ object ZStreamSpec extends ZIOBaseSpec {
                 .runCollect
             )(equalTo(List(1)))
           },
-          testM("should interrupt stream") {
+          testM("should end stream") {
             assertM(
               ZStream
                 .range(0, 5)
                 .tap(_ => ZIO.sleep(Duration.Infinity))
                 .timeout(Duration.Zero)
-                .runDrain
-                .sandbox
-                .ignore
-                .map(_ => true)
-            )(isTrue)
-          } @@ zioTag(interruption)
+                .runCollect
+            )(isEmpty)
+          }
+        ),
+        testM("timeoutError") {
+          assertM(
+            ZStream
+              .range(0, 5)
+              .tap(_ => ZIO.sleep(Duration.Infinity))
+              .timeoutError(false)(Duration.Zero)
+              .runDrain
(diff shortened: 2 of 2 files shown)


luis3m:
Will update, just realized it will apply the timeout to the new stream as well.

luis3m:
Done

luis3m:
Once again, tests are passing on top of https://github.com/zio/zio/pull/3424

luis3m:
Probably renaming `timeoutErr` to `timeoutFail` makes sense to keep consistency with ZIO

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,58 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   * Ends the stream if it does not produce a value after d duration.
    */
   final def timeout(d: Duration): ZStream[R with Clock, E, O] =
     ZStream[R with Clock, E, O] {
       self.process.map { next =>
         next.timeout(d).flatMap {
-          case Some(a) => ZIO.succeedNow(a)
-          case None    => ZIO.interrupt
+          case Some(a) => Pull.emit(a)
+          case None    => Pull.end
```
iravid, line 2482: We should add a Ref here that remembers that this stream has ended due to timeout.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,58 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+
+  /**
+   * Switches the stream if it does not produce a value after d duration.
+   */
+  final def timeoutTo[R1 <: R, E1 >: E, O2 >: O](
+    d: Duration
+  )(that: ZStream[R1, E1, O2]): ZStream[R1 with Clock, E1, O2] =
+    ZStream[R1 with Clock, E1, O2] {
+      for {
+        currStream   <- Ref.make[ZIO[R1, Option[E1], Chunk[O2]]](Pull.end).toManaged_
```
iravid, line 2508: Is it possible to use catchAllCause for this combinator? Implementation looks fairly similar.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,58 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+          case Some(a) => Pull.emit(a)
+          case None    => Pull.end
         }
       }
     }
 
+  /**
+   * Ends the stream with given error if it does not produce a value after d duration.
+   */
+  final def timeoutErr[E1 >: E](e: E1)(d: Duration): ZStream[R with Clock, E1, O] =
```
iravid, line 2490: How about timeoutError and timeoutErrorCause?

luis3m: So that `timeoutErrorCause` uses `ZIO.die`?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: I will take a look

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @iravid done, working on the other suggestions

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @iravid not sure what should be the behavior here, something like:

```scala
ZStream[R with Clock, E, O] {
  for {
    timeout <- Ref.make(false).toManaged_
    next    <- self.process
    pull = timeout.get.flatMap {
      if (_) Pull.end
      else
        next.timeout(d).flatMap {
          case Some(a) => Pull.emit(a)
          case None    => timeout.set(true) *> Pull.end
        }
    }
  } yield pull
}
```
?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Given that `timeoutErrorCause` has been implemented, the following implementation comes to mind:

```scala
final def timeoutTo[R1 <: R, E1 >: E, O2 >: O](
    d: Duration
)(that: ZStream[R1, E1, O2]): ZStream[R1 with Clock, E1, O2] = {
  object StreamTimeout extends Throwable
  self.timeoutErrorCause(Cause.die(StreamTimeout))(d).catchSomeCause { case Cause.Die(StreamTimeout) => that }
}
```

It requires implementing `catchSomeCause` though which I can add (maybe in a separate PR). @iravid what do you think?

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Yes, just like that.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,71 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-          case Some(a) => ZIO.succeedNow(a)
-          case None    => ZIO.interrupt
+          case Some(a) => Pull.emit(a)
+          case None    => Pull.end
         }
       }
     }
 
+  /**
+   * Ends the stream with given error if it does not produce a value after d duration.
```
iravid, line 2488:
```suggestion
   * Fails the stream with given error if it does not produce a value after d duration.
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,71 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   */
+  final def timeoutError[E1 >: E](e: E1)(d: Duration): ZStream[R with Clock, E1, O] =
+    ZStream[R with Clock, E1, O] {
+      self.process.map { next =>
+        next.timeout(d).flatMap {
+          case Some(a) => Pull.emit(a)
+          case None    => Pull.fail(e)
+        }
+      }
+    }
```
iravid, line 2498:
```suggestion
timeoutErrorCause(Cause.fail(e))(d)
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2472,18 +2472,71 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+        switchStream <- ZManaged.switchable[R1, Nothing, ZIO[R1, Option[E1], Chunk[O2]]]
+        switched     <- Ref.make(false).toManaged_
+        _            <- switchStream(self.process).flatMap(currStream.set).toManaged_
+        pull = {
+          def go: ZIO[R1 with Clock, Option[E1], Chunk[O2]] =
+            for {
+              switched_ <- switched.get
+              effect    = currStream.get.flatten
+              switch    = switched.set(true) *> switchStream(that.process).flatMap(currStream.set) *> go
+              timeout   = effect.timeout(d).flatMap(_.map(Pull.emit).getOrElse(switch))
```
iravid, line 2531: To save a (small) amount of allocations and simplify `go` a bit, these could become `val`s outside of `go`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Sounds good for a follow up if you have the cycles later on. The current implementation looks good for this PR! Thanks!

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Right, thanks. Done.

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Done, I had to make those lazy to avoid the `forward reference` compile error.

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Cool, I will submit a separate PR later with `catchSomeCause` and the refactor of `timeoutTo`

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: :+1:

luis3m:
Rebased on top of #3424, tests should pass now

luis3m:
@iravid I addressed all the suggestions and rebased on top of the latest TestClock changes

Issue #3442, opened 2020-04-24
Reconsider Use of PartialFunction[A, ZIO[R, E, B]] Signature


adamgfraser:
Currently we use the signature `PartialFunction[A, ZIO[R, E, B]]` for the effectual version of functions that take a partial function. However, I think that may not be quite the right signature. As I understand it, the contract there would be that given a value of type `A`, the partial function can immediately and purely tell us whether it is defined at a given input, but if we want to compute the value of the partial function for that input we need to evaluate the effect. But I think most of the time when we think of effectual partial functions we need to evaluate effects to even know whether the function is defined for an input. I would suggestion we consider changing to `A => ZIO[R, E, Option[B]]`.

jdegoes:
I'd suggest:

```scala
Z => ZIO[R, Option[E], A]
```

We could add some syntax to those (`orElse`).

adamgfraser:
Good idea.

PR #3568, opened 2020-05-10
WIP: Add combinators collectSome and collectSomePar

ghidei:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2252,6 +2252,20 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   )(f: PartialFunction[A, U]): ZIO[R, E, List[U]] =
     ZIO.collectAllParN(n)(in).map(_.collect(f))
 
+  /**
+   * Evaluate each effect in the structure from left to right, collecting the
+   * the successful values and discarding the empty cases. For a parallel version, see `collectSomePar`.
+   */
+  def collectSome[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
+    foreach(in)(f).optional.map(_.fold(List.empty[B])(ZIO.identityFn))
```
adamgfraser, line 2260: I don't think this is going to do what you want it to. Right now any missing value is going to cause `foreach` to fail. To get your intended semantics you need to do something like `foreach(in)(a => f(a).optional).map(_.flatten)`. Would be good to add some tests to verify the correct behavior here.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2252,6 +2252,20 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   )(f: PartialFunction[A, U]): ZIO[R, E, List[U]] =
     ZIO.collectAllParN(n)(in).map(_.collect(f))
 
+  /**
+   * Evaluate each effect in the structure from left to right, collecting the
+   * the successful values and discarding the empty cases. For a parallel version, see `collectSomePar`.
+   */
+  def collectSome[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
```
adamgfraser, line 2259: I would be inclined to name this `collect`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2252,6 +2252,20 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+    foreach(in)(f).optional.map(_.fold(List.empty[B])(ZIO.identityFn))
+
+  /**
+   * Evaluate each effect in the structure from left to right, collecting the
+   * the successful values and discarding the empty cases.
+   */
+  def collectSomePar[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
```
adamgfraser, line 2266: Could also add a `ParN` variant.

core/shared/src/main/scala/zio/URIO.scala
```diff
@@ -198,6 +198,18 @@ object URIO {
   def collectAllWithParN[R, A, B](n: Int)(as: Iterable[URIO[R, A]])(f: PartialFunction[A, B]): URIO[R, List[B]] =
     ZIO.collectAllWithParN(n)(as)(f)
 
+  /**
+   * @see See [[zio.ZIO.collectSome]]
+   */
+  def collectSome[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
```
adamgfraser, line 204: For the aliases on the companion object the contract is that it should give us the more specific type. So for example since this is defined on `URIO` we should drop the `E` type parameter, `f` should be `A => URIO[R, Option[Nothing], B]` and the return type should be `URIO[R, List[B]]`.

core/shared/src/main/scala/zio/ZIO.scala
ghidei: Just a heads up - if we change the name to `collect` then `collectPar` and `collectParN` are not going to be defined at the same place. The `collectAll*` functions will come in between.

adamgfraser: I think that is fine.

ghidei:
By the way, is this PR related to what we're trying to do: https://github.com/zio/zio/issues/3442?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2252,6 +2259,22 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   )(f: PartialFunction[A, U]): ZIO[R, E, List[U]] =
     ZIO.collectAllParN(n)(in).map(_.collect(f))
 
+  /**
+   * Evaluate each effect in the structure from left to right, collecting the
```
adamgfraser, line 2263:
```suggestion
   * Evaluate each effect in the structure in parallel, collecting the
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2252,6 +2259,22 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+   * the successful values and discarding the empty cases.
+   */
+  def collectPar[R, E, A, B](in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
+    foreachPar(in)(a => f(a).optional).map(_.flatten)
+
+  /**
+   * Evaluate each effect in the structure from left to right, collecting the
```
adamgfraser, line 2270:
```suggestion
   * Evaluate each effect in the structure in parallel, collecting the
```

ghidei:
@adamgfraser I've just added the combinators for `ZManaged` and `ZSTM`. I saw that there were some aliases for them, but they didn't really have any companion objects the same way the `ZIO` datatype has.

ghidei:
@adamgfraser Friendly reminder :)

Issue #3492, opened 2020-04-30
ZStream#toIterator does not use chunks directly


iravid:
`toIterator` currently uses `BufferedPull` to pull elements. This creates iterators that unnecessarily invoke `runtime.unsafeRun` to pull individual elements.

Instead, the iterator should store chunks in its mutable state and return elements from the current chunk without invoking `runtime.unsafeRun`.

iravid:
Fixed in #3569.

PR #3569, opened 2020-05-10
Modify ZStream#toIterator to use chunks

luis3m:
As suggested here https://github.com/zio/zio/issues/3492

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2535,14 +2535,15 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    */
   def toIterator: ZManaged[R, Nothing, Iterator[Either[E, O]]] =
     for {
-      pull    <- this.process.mapM(BufferedPull.make(_))
       runtime <- ZIO.runtime[R].toManaged_
+      pull    <- process
     } yield {
       new Iterator[Either[E, O]] {
+        var nextTake: Take[E, O] = null
+        var chunkIdx: Int        = 0
 
-        var nextTake: Exit[Option[E], O] = null
         def unsafeTake(): Unit =
-          nextTake = runtime.unsafeRunSync(pull.pullElement)
+          nextTake = runtime.unsafeRunSync(pull)
 
         def hasNext: Boolean = {
           if (nextTake == null) {
@@ -2555,7 +2556,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
                 case Left(None) => false
                 case _          => true
               }
-            case _ => true
+            case Exit.Success(chunk) => !chunk.isEmpty
           }
         }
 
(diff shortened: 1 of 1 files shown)


luis3m:
Previously had the following but I think current solution should be faster:
```scala
case Exit.Success(chunk) =>
  val (first, tail) = chunk.splitAt(1)
  val value         = first.headOption.toRight(throw new NoSuchElementException("next on empty iterator"))
  nextChunk = if (tail.isEmpty) null else Exit.Success(tail)
  value
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2535,14 +2535,15 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    */
   def toIterator: ZManaged[R, Nothing, Iterator[Either[E, O]]] =
     for {
-      pull    <- this.process.mapM(BufferedPull.make(_))
       runtime <- ZIO.runtime[R].toManaged_
+      pull    <- process
     } yield {
       new Iterator[Either[E, O]] {
+        var nextTake: Take[E, O] = null
```
simpadjo, line 2542: What do you think about storing
```
var chunkIterator: Iterator[O]
var error: E   
```
instead? 
This could save some CPU cycles on a random access to the chunk and on unwrapping `Take` multiple times/

luis3m: @simpadjo wouldn't this suggestion ignore failures though? or you mean `Iterator[Either[E, O]]`?

luis3m: @simpadjo this is what comes to mind:

```scala
new Iterator[Either[E, O]] {
  var nextTake: Exit[Option[E], Iterator[O]] = null

  def unsafeTake(): Unit =
    nextTake = runtime.unsafeRunSync(pull.map(_.toIterator))

  def hasNext: Boolean = {
    if (nextTake == null) {
      unsafeTake()
    }

    nextTake match {
      case Exit.Failure(cause) =>
        cause.failureOrCause match {
          case Left(None) => false
          case _          => true
        }
      case Exit.Success(it) => it.hasNext
    }
  }

  def next(): Either[E, O] = {
    if (nextTake == null) {
      unsafeTake()
    }

    nextTake match {
      case Exit.Failure(cause) =>
        cause.failureOrCause match {
          case Left(None)    => throw new NoSuchElementException("next on empty iterator")
          case Left(Some(e)) => Left(e)
          case Right(c)      => throw FiberFailure(c)
        }
      case Exit.Success(it) =>
        val value = it.next()

        if (!it.hasNext) {
          nextTake = null
        }

        Right(value)
    }
  }
}
```
Track of failures won't be lost.

simpadjo: You could store it in a separate `var` but nevermind: seems that low-level manipulations can be avoided altogether

```
      def unfoldPull: Iterator[Either[E, O]] = {
        runtime.unsafeRunSync(pull) match {
          case Exit.Success(chunk) => chunk.iterator.map(Right(_)) ++ unfoldPull
          case Exit.Failure(cause) =>  cause.failureOrCause match {
            case Left(None)    => Iterator.empty
            case Left(Some(e)) => Iterator.single(Left(e))
            case Right(c)      => throw FiberFailure(c)
          }
        }
      }
```

luis3m: @simpadjo nice! just updated with your solution

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2535,49 +2535,21 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-
-          nextTake match {
-            case Exit.Failure(cause) =>
-              cause.failureOrCause match {
-                case Left(None) => false
-                case _          => true
-              }
-            case _ => true
-          }
+      def unfoldPull: Iterator[Either[E, O]] =
```
iravid, line 2541: Wow this is *really* nice. Great work @simpadjo and @luis3m.

PR #3574, opened 2020-05-11
Add ZIO#debounce

luis3m:
Add debounce function for ZIO which guarantees the effect will be executed just once after the given duration regardless of how many times the produced effect is evaluated.

For example, if a debounced effect with two seconds is evaluated five times within a second then two seconds later after the last evaluation the effect will be executed, taking a total of three seconds. Based on https://medium.com/@jamischarles/what-is-debouncing-2505c0648ff1

Current implementation will run first evaluation and fail None any other during debounce (though will adjust sleep time regardless). It's not exactly equivalent to the implementation of the post because that one clears previous evaluations and keep the latest one. Doing so would make this combinator heavier or at least in the implementation I had in mind so decided to go with this lighter approach.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -159,6 +159,32 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert((a, b, c))(equalTo((0, 0, 0)))
       }
     ),
+    suite("debounce")(
+      testM("should execute once in parallel computations") {
+        for {
+          ref     <- Ref.make[Int](0)
+          io      <- ref.update(_ + 1).debounce(1.second)
+          fiber   <- ZIO.tupledPar(io.either, io.either, io.either).fork
+          _       <- TestClock.advance(2.seconds)
+          tuple   <- fiber.join
+          success = tuple.productIterator.collect { case Right(a) => a }.length
+          count   <- ref.get
+        } yield assert(count)(equalTo(1)) && assert(success)(equalTo(1))
+      },
+      testM("should reset sleep time on every evaluation") {
+        for {
+          io      <- ZIO.unit.debounce(1.second)
+          fiber1  <- io.either.fork
+          fiber2  <- (clock.sleep(100.millis) *> io.either).fork
+          fiber3  <- (clock.sleep(400.millis) *> io.either).fork
+          _       <- TestClock.advance(1400.millis)
+          result1 <- fiber1.join
+          result2 <- fiber2.join
+          result3 <- fiber3.join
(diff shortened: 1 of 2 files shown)


luis3m:
JS Tests passing on top of https://github.com/zio/zio/pull/3424

jdegoes:
Is this different than `ZIO#cached`?

luis3m:
@jdegoes yes, `cached` will cache the value for a given amount of time whereas `debounce` does not cache anything at all. It just ensure the effect will be executed once after X duration.

I can see it being useful along with ZStreams or event handlers in Scalajs. Look at the scroll handler example in the link above. https://redd.one/blog/debounce-vs-throttle this post has an interactive button which can help understand its behavior.

reibitto:
I was also looking for this. I have some context about what I was trying to with a streams version in #2565

luis3m:
Perhaps making `debounce` provide a fiber straight away makes sense as @reibitto's solution

reibitto:
Another thing I was thinking about is whether to use `forkDaemon` or not. In my implementation I went with `fork` but return the fiber so that the user could always choose to `join` the very last one if it being canceled is not the desired behavior. But I'm not sure what the right default is.

adamgfraser:
I would have thought that the signature would be as below and that if there are multiple calls they will each just suspend until the value is eventually computed and then return that value.

```scala
final def debounce(duration: Duration): UIO[ZIO[R with Clock, E, A]]
```

luis3m:
@adamgfraser I will take a look at this tomorrow

luis3m:
@adamgfraser I'm looking into this now, I will probably update the solution. With `UIO[ZIO[R with Clock, E, A]]` you will have to halt the IO which is why current solution has an option as error.

Failure `None` means the IO has been interrupted by a newer call within the specified duration. Any suggestions?

adamgfraser:
My thought was that the semantics would be that if there is a newer call you just suspend and return the result when it is computed. That seems to reflect the idea that a "debounced" has this semantic of not being executed until a certain amount of time has passed from subsequent calls but is otherwise identical to the original effect, and would mirror the signatures of `memoize` and `once`.

I think to do that you would need a `Ref` to keep track of whether the actual effect has started execution and another to hold a fiber that is waiting for the delay and then executing the effect. Then every time the effect gets invoked you see if actual execution has started. If so await the fiber that is currently doing it. If not interrupt that fiber and fork a new one to do it with the appropriate delay.

jdegoes:
@luis3m This sounds quite promising (thank you for the idea) and I like @adamgfraser's proposed signature and semantics for this. 

As this PR is out of date, I will close for now, but feel free to re-open anytime with updates!

Issue #3387, opened 2020-04-16
Add specialized chunks for every primitive type


jdegoes:
In order to facilitate box-less operations, we can add chunks for different primitive types. Technically, we should support not just `Array` but also NIO buffers (byte buffer, etc). 

For organizational purposes, these should probably be placed inside an object inside the companion object of Chunk, e.g.:

```scala
object Chunk {
  object Byte {
    final class ByteChunkArr(array: Array[Byte]) extends Chunk[A] {
      override def byte(index: Long): Byte = array(index)
    }
    final class ByteChunkBuf(buffer: ByteBuffer) extends Chunk[A] {
      ..
    }
  }
}
```

Specialized chunk subtypes will have efficient and box-free implementations of primitive-specialized methods.

See #3386

vasilmkd:
With `Chunk` now being an `IndexedSeq` what do we stand to gain from this change (which was turned down ~6 months ago due to too much code duplication and a burden on maintainability)?

jdegoes:
Any specialized methods we add (which are not inside `IndexedSeq`) do not need to box.

michalwarecki:
I can take it

iravid:
Resolved in #3576.

PR #3576, opened 2020-05-12
Add specialized chunks for every primitive type

michalwarecki:
Hi,

I continue exploring zio internals and picked up #3387 . I started with specialized chunks for arrays and I'm not sure sure if nio buffers implementation is needed as described in the #3387 .
All of `from*Buffer` implementations are converting nio buffers into arrays so it looks like using specialized chunks for arrays solves nio buffers as well. CC @jdegoes 
I've also added `from*Array` methods to avoid array type check in `fromArray` but I'm not sure if that is preferred implementation.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -21,14 +21,6 @@ import java.nio._
 import scala.collection.mutable.Builder
 import scala.reflect.{ classTag, ClassTag }
 
-import zio.Chunk.Byte.ByteChunkArr
-import zio.Chunk.Char.CharChunkArr
-import zio.Chunk.Double.DoubleChunkArr
-import zio.Chunk.Float.FloatChunkArr
-import zio.Chunk.Int.IntChunkArr
-import zio.Chunk.Long.LongChunkArr
-import zio.Chunk.Short.ShortChunkArr
-
 /**
  * A `Chunk[A]` represents a chunk of values of type `A`. Chunks are designed
  * are usually backed by arrays, but expose a purely functional, safe interface
@@ -547,7 +539,7 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
   def materialize[A1 >: A]: Chunk[A1] =
     self.toArrayOption[A1] match {
       case None        => Chunk.Empty
-      case Some(array) => Chunk.Arr(array)
+      case Some(array) => Chunk.fromArray(array)
     }
 
   /**
@@ -761,49 +753,19 @@ object Chunk {
    * Returns a chunk backed by an array.
    */
   def fromArray[A](array: Array[A]): Chunk[A] =
-    if (array.isEmpty) Empty else Arr(array)
-
-  /**
-   * Returns a chunk backed by an byte array.
-   */
-  def fromByteArray(array: Array[Byte]): Chunk[Byte] =
-    if (array.isEmpty) Empty else ByteChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an int array.
-   */
-  def fromIntArray(array: Array[Int]): Chunk[Int] =
-    if (array.isEmpty) Empty else IntChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an char array.
-   */
-  def fromCharArray(array: Array[Char]): Chunk[Char] =
-    if (array.isEmpty) Empty else CharChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an double array.
-   */
-  def fromDoubleArray(array: Array[Double]): Chunk[Double] =
-    if (array.isEmpty) Empty else DoubleChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an float array.
-   */
-  def fromFloatArray(array: Array[Float]): Chunk[Float] =
-    if (array.isEmpty) Empty else FloatChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an long array.
-   */
-  def fromLongArray(array: Array[Long]): Chunk[Long] =
-    if (array.isEmpty) Empty else LongChunkArr(array)
-
-  /**
-   * Returns a chunk backed by an short array.
-   */
-  def fromShortArray(array: Array[Short]): Chunk[Short] =
-    if (array.isEmpty) Empty else ShortChunkArr(array)
+    (if (array.isEmpty) Empty
+     else
+       array.asInstanceOf[AnyRef] match {
+         case x: Array[AnyRef]  => RefChunkArr(x)
+         case x: Array[Int]     => IntChunkArr(x)
+         case x: Array[Double]  => DoubleChunkArr(x)
+         case x: Array[Long]    => LongChunkArr(x)
+         case x: Array[Float]   => FloatChunkArr(x)
+         case x: Array[Char]    => CharChunkArr(x)
+         case x: Array[Byte]    => ByteChunkArr(x)
+         case x: Array[Short]   => ShortChunkArr(x)
+         case x: Array[Boolean] => BooleanChunkArr(x)
+       }).asInstanceOf[Chunk[A]]
 
   /**
    * Returns a chunk backed by a [[java.nio.ByteBuffer]].
@@ -813,7 +775,7 @@ object Chunk {
     val pos  = buffer.position()
     buffer.get(dest)
     buffer.position(pos)
-    Chunk.fromByteArray(dest)
+    Chunk.fromArray(dest)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -882,16 +932,21 @@ object Chunk {
...
+      case x: Singleton[A]    => x.classTag
+      case x: Slice[A]        => x.classTag
+      case x: VectorChunk[A]  => x.classTag
+      case _: BitChunk        => ClassTag.Boolean.asInstanceOf[ClassTag[A]]
     }
 
-  private final case class Arr[A](private val array: Array[A]) extends Chunk[A] with Serializable { self =>
+  private final case class Arr[A](array: Array[A]) extends ArraySupport[A]
+
+  private[zio] trait ArraySupport[A] extends Chunk[A] with Serializable { self =>
```
iravid, line 947: To make this simpler, instead of a mixin trait, let's make `Arr[A]` an abstract class and have all the primitive subtypes inherit it. In addition, we can add a `RefChunkArr` subtype for `A <: AnyRef` types, which will subsume the role of `Arr[A]` today.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1322,4 +1377,52 @@ object Chunk {
     private val CharClass       = classTag[Char]
     private val CharClassBox    = classTag[java.lang.Character]
   }
+
+  object Byte {
```
iravid, line 1381: Is there a reason for all the subtypes to live in these `Byte/Char/Int` objects?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -755,6 +763,48 @@ object Chunk {
   def fromArray[A](array: Array[A]): Chunk[A] =
```
iravid, line 763: Let's make this method do the right thing by pattern matching on the array. You can see an example for this in `WrappedArray.make`.

michalwarecki:
Thanks for the review! Let's check if that version is OK and I then I can add some tests to `fromArray`

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -891,7 +903,9 @@ object Chunk {
       case _: BitChunk       => ClassTag.Boolean.asInstanceOf[ClassTag[A]]
     }
 
-  private final case class Arr[A](private val array: Array[A]) extends Chunk[A] with Serializable { self =>
+  private[zio] abstract class Arr[A] extends Chunk[A] with Serializable { self =>
```
iravid, line 906:
```suggestion
  private[zio] sealed abstract class Arr[A] extends Chunk[A] with Serializable { self =>
```

iravid:
Looks good. Here's the relevant bytecode; appears to not be boxing:
```scala
object ChunkTest {
  def test() = {
    val bytes = Chunk.fromArray(Array[Byte](1, 2, 3))

    var i = 0
    var x: Byte = -1
    while (i < bytes.length) {
      val b = bytes.byte(i)
      x = b
      i += 1
    }
  }
}
```

Compiles to:
```java
  public void test();
    Code:
       0: getstatic     #23                 // Field zio/Chunk$.MODULE$:Lzio/Chunk$;
       3: getstatic     #28                 // Field scala/Array$.MODULE$:Lscala/Array$;
       6: getstatic     #33                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
       9: iconst_3
      10: newarray       byte
      12: dup
      13: iconst_0
      14: iconst_1
      15: bastore
      16: dup
      17: iconst_1
      18: iconst_2
      19: bastore
      20: dup
      21: iconst_2
      22: iconst_3
      23: bastore
      24: invokevirtual #37                 // Method scala/Predef$.wrapByteArray:([B)Lscala/collection/mutable/WrappedArray;
      27: getstatic     #42                 // Field scala/reflect/ClassTag$.MODULE$:Lscala/reflect/ClassTag$;
      30: invokevirtual #46                 // Method scala/reflect/ClassTag$.Byte:()Lscala/reflect/ClassTag;
      33: invokevirtual #50                 // Method scala/Array$.apply:(Lscala/collection/Seq;Lscala/reflect/ClassTag;)Ljava/lang/Object;
      36: invokevirtual #54                 // Method zio/Chunk$.fromArray:(Ljava/lang/Object;)Lzio/Chunk;
      39: astore_1
      40: iconst_0
      41: istore_2
      42: iconst_m1
      43: istore_3
      44: iload_2
      45: aload_1
      46: invokeinterface #60,  1           // InterfaceMethod zio/Chunk.length:()I
      51: if_icmpge     79
      54: aload_1
      55: iload_2
      56: getstatic     #33                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
      59: invokevirtual #64                 // Method scala/Predef$.$conforms:()Lscala/Predef$$less$colon$less;
      62: invokeinterface #68,  3           // InterfaceMethod zio/Chunk.byte:(ILscala/Predef$$less$colon$less;)B
      67: istore        4
      69: iload         4
      71: istore_3
      72: iload_2
      73: iconst_1
      74: iadd
      75: istore_2
      76: goto          44
      79: return
```

And the corresponding `Chunk.ByteChunkArr#byte` compiles to:
```java
  public byte byte(int, scala.Predef$$less$colon$less<java.lang.Object, java.lang.Object>);
    Code:
       0: aload_0
       1: invokevirtual #31                 // Method array:()[B
       4: iload_1
       5: baload
       6: ireturn
```

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1322,4 +1336,38 @@ object Chunk {
     private val CharClass       = classTag[Char]
     private val CharClassBox    = classTag[java.lang.Character]
   }
+
+  final case class RefChunkArr[A <: AnyRef](array: Array[A]) extends Arr[A]
```
iravid, line 1340: Last naming request - can we call the chunks `Refs, Bytes, Ints` etc. for brevity?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -753,7 +753,19 @@ object Chunk {
    * Returns a chunk backed by an array.
    */
   def fromArray[A](array: Array[A]): Chunk[A] =
-    if (array.isEmpty) Empty else Arr(array)
+    (if (array.isEmpty) Empty
+     else
+       array.asInstanceOf[AnyRef] match {
+         case x: Array[AnyRef]  => AnyRefArray(x)
```
jdegoes, line 759: Have you tested to make sure Scala is doing the right thing here w.r.t type erasure? Ordinarily you can't match on, e.g., `List[Int]` because of type erasure. There is a means to do this for arrays, because they store their type inside them, but I'm not 100% sure if Scala utilizes this to make this type casing sane. Worth testing to be sure.

Or to avoid you could match on `Array#elemTag` which gives you the tag of the element type.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1322,4 +1336,38 @@ object Chunk {
...
+  final case class DoubleArray(array: Array[Double]) extends Arr[Double] {
+    override def double(index: Int)(implicit ev: Double <:< Double): Double = array(index)
+  }
+
+  final case class FloatArray(array: Array[Float]) extends Arr[Float] {
+    override def float(index: Int)(implicit ev: Float <:< Float): Float = array(index)
+  }
+
+  final case class ShortArray(array: Array[Short]) extends Arr[Short] {
+    override def short(index: Int)(implicit ev: Short <:< Short): Short = array(index)
```
jdegoes, line 1367: Really nice to see all these box-free accessors!! ❤️

joprice: Is there a note anywhere on this technique for avoiding boxing? I don't understand what role the implicit is serving here.

core/shared/src/main/scala/zio/Chunk.scala
iravid: I’ve tested in the REPL that a byte array becomes a ByteArray

iravid: The pattern matching works here because an Array[Int] is a distinct type at runtime, whereas List[Int] is just a List[AnyRef] at runtime

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: The implicit acts as proof in the parent type that it is safe to call this method in the child type. It's a way of exposing the specialized method in the parent.

core/shared/src/main/scala/zio/Chunk.scala
jdegoes: Ok, so Scalac has baked this in. 👍

PR #3580, opened 2020-05-13
Add ZStream#debounce

luis3m:
- Both JVM & JS tests passing on top of https://github.com/zio/zio/pull/3424 as it guarantees to run the effects in the correct order.

- Took two tests from https://github.com/zio/zio/pull/3521 (cc @reibitto) and added three more.

- It does reset time

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2599,54 +2599,44 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     }
 
   final def debounce(d: Duration): ZStream[R with Clock, E, O] = {
-    type Loser = Either[Fiber[Nothing, Take[E, O]], Fiber[Nothing, Take[E, O]]]
-
     sealed abstract class State
     case object NotStarted                                extends State
     case class Previous(fiber: Fiber[Nothing, Chunk[O]])  extends State
-    case class Current(fiber: Fiber[Nothing, Take[E, O]]) extends State
+    case class Current(fiber: Fiber[Option[E], Chunk[O]]) extends State
     case object Done                                      extends State
-
-    def race(
-      previous: ZIO[R, Option[E], Chunk[O]],
-      current: ZIO[R, Option[E], Chunk[O]]
-    ): URIO[R, (Take[E, O], Loser)] =
-      previous.run.raceWith(current.run)(
-        (exit, current) => ZIO.done(exit).map(a => (a, Right(current))),
-        (exit, previous) => ZIO.done(exit).map(b => (b, Left(previous)))
-      )
-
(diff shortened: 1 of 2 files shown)


luis3m:
These two failures seem to be unrelated

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2469,6 +2469,60 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+    ZStream[R with Clock, E, O] {
+      for {
+        chunks <- self.process
+        ref    <- Ref.make[Option[Fiber[E, Chunk[O]]]](None).toManaged_
+        pull = {
+          ref.get.flatMap {
+            case None =>
+              chunks >>= store(ref)
+            case Some(previous) =>
+              val prev = previous.join.run.map(_.mapError(Some(_)))
```
iravid, line 2497:
```suggestion
              val prev = previous.await.map(_.mapError(Some(_)))
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2469,6 +2469,60 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+                          fiber  <- loser.merge.join
+                          done   <- ZIO.done(fiber)
+                          _      <- ref.set(None)
+                          result <- Pull.emit(done)
+                        } yield result
+                    },
+                    chunk =>
+                      loser.fold(
+                        _.interrupt *> store(ref)(chunk),
+                        _.interrupt *> ref.set(None) *> Pull.emit(chunk)
```
iravid, line 2516: We can't interrupt an ongoing pull. That is likely to lead to corruption of internal operator state. We have to keep the ongoing pull around.

luis3m: @iravid solved

luis3m:
1) Refactored a bit
2) Some extra annotations to fix Dotty compilation issues

luis3m:
I can remove the top level type parameters by deferring them to `State` but I'm not sure it's worth doing
```scala
sealed abstract class State[+E, +O]
case object NotStarted                                      extends State[Nothing, Nothing]
case class Previous[O](fiber: Fiber[Nothing, Chunk[O]])     extends State[Nothing, O]
case class Current[E, O](fiber: Fiber[Option[E], Chunk[O]]) extends State[E, O]
case object Done                                            extends State[Nothing, Nothing]
```

reibitto:
@iravid @luis3m Hmm, I think there's one corner case still left with the handling of the end of a stream. I expect the following to result in `List(3)`, but right now it's an empty list:

```scala
ZStream(1, 2, 3).fixed[Any](500.millis).debounce(1.second).runCollect
```

iravid:
Ah ok. That should be fixed. Can you open a follow up issue?

reibitto:
Sure thing.

PR #3581, opened 2020-05-13
Add Fork Operator to FiberRef

### LINKED ISSUE #3579: Augment FiberRef constructor with "fork" operator
Currently, to construct a fiber ref, one must supply:

```scala
initial: A
join: (A, A) => A = (_, r) => r
```

To be fully general, we should support an additional operation, "fork", which will be called when the value of the fiber ref is "forked" into a child:

```scala
initial: A
fork: A => A = identity(_)
join: (A, A) => A = (_, r) => r
```

The default, of course, is just the identity function.

adamgfraser:
Resolves #3579.

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -576,8 +576,8 @@ private[zio] final class FiberContext[E, A](
                   case ZIO.Tags.FiberRefNew =>
                     val zio = curZio.asInstanceOf[ZIO.FiberRefNew[Any]]
 
-                    val fiberRef = new FiberRef[Any](zio.initialValue, zio.combine)
-                    fiberRefLocals.put(fiberRef, zio.initialValue)
+                    val fiberRef = new FiberRef[Any](zio.initial, zio.onFork, zio.onJoin)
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -576,8 +576,8 @@ private[zio] final class FiberContext[E, A](
                   case ZIO.Tags.FiberRefNew =>
                     val zio = curZio.asInstanceOf[ZIO.FiberRefNew[Any]]
 
-                    val fiberRef = new FiberRef[Any](zio.initialValue, zio.combine)
-                    fiberRefLocals.put(fiberRef, zio.initialValue)
+                    val fiberRef = new FiberRef[Any](zio.initial, zio.onFork, zio.onJoin)
+                    fiberRefLocals.put(fiberRef, zio.onFork(zio.initial))
```
jdegoes, line 580: Wouldn't the right place to put this be where a child fiber accesses a parent fiber's ref for the first time?

This is calling 'fork' whenever a new fiber ref is created. But we would like it just to be called during "handoff" of a fiber ref from parent to child.

adamgfraser: Yes. Fixed now.

PR #3583, opened 2020-05-14
Add App ExitCode

luis3m:
Addresses #2153 (cc @adamgfraser @jdegoes)

### DIFF
--- a/core/shared/src/main/scala/zio/ExitCode.scala
+++ b/core/shared/src/main/scala/zio/ExitCode.scala
@@ -16,18 +16,9 @@
 
 package zio
 
-trait ManagedApp extends BootstrapRuntime { ma =>
+final case class ExitCode(code: Int)
 
-  /**
-   * The main function of the application, which will be passed the command-line
-   * arguments to the program and has to return an `ZManaged` with the errors fully handled.
-   */
-  def run(args: List[String]): ZManaged[ZEnv, Nothing, ExitCode]
-
-  private val app = new App {
-    override def run(args: List[String]): ZIO[ZEnv, Nothing, ExitCode] =
-      ma.run(args).use(exit => ZIO.effectTotal(exit))
-  }
-
-  final def main(args: Array[String]): Unit = app.main(args)
+object ExitCode {
+  val success = ExitCode(0)
+  val failure = ExitCode(1)
 }
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -516,8 +516,11 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Maps this effect to the default exit codes.
    */
-  final def exitCode: URIO[R, ExitCode] =
-    self.foldCauseM(cause => UIO(println(cause.prettyPrint)) as ExitCode.failure, _ => ZIO.succeedNow(ExitCode.success))
+  final def exitCode: URIO[R with console.Console, ExitCode] =
+    self.foldCauseM(
+      cause => console.putStrLn(cause.prettyPrint) as ExitCode.failure,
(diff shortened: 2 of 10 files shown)


core/shared/src/main/scala/zio/ExitCode.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+case class ExitCode(code: Int) extends AnyVal
```
jdegoes, line 19:
```suggestion
final case class ExitCode(code: Int)
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -209,6 +209,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def asSomeError: ZIO[R, Option[E], A] =
     mapError(Some(_))
 
+  /**
+   * Maps this effect to the default exit codes.
+   */
+  final def asExitCode: URIO[R, ExitCode] =
```
jdegoes, line 215:
```suggestion
  final def exitCode: URIO[R, ExitCode] =
```

core/shared/src/main/scala/zio/ExitCode.scala
luis3m: @jdegoes any particular concern with the value class?

core/shared/src/main/scala/zio/ZIO.scala
luis3m: Done

core/shared/src/main/scala/zio/ExitCode.scala
jdegoes: This is used one time at the exit of the application, so there's no need to use AnyVal, and using AnyVal won't actually work because of the way this is returned from a lambda.

luis3m: Right, just read the cases where it actually allocates. It's indeed useless here.

luis3m: Done

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -513,6 +513,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def either(implicit ev: CanFail[E]): URIO[R, Either[E, A]] =
     self.foldM(ZIO.succeedLeft, ZIO.succeedRight)
 
+  /**
+   * Maps this effect to the default exit codes.
+   */
+  final def exitCode: URIO[R, ExitCode] =
+    self.foldCauseM(cause => UIO(println(cause.prettyPrint)) as ExitCode.failure, _ => ZIO.succeedNow(ExitCode.success))
```
iravid, line 520: Let’s use putStrLn instead of println here.

luis3m: @iravid I took this combinator from #2153, I suspect John suggested `println` to avoid adding a dependency on Console at the end of your program. OTOH, App uses `ZEnv` anyway so even if you write `program.provideSomeLayer(ZEnv.live ++ customLayers).exitCode` it won't crash as Console would be its only missing dependency which the default runtime will provide.

I will update in a few minutes (cc: @jdegoes)

luis3m: @iravid done

jdegoes:
@luis3m Thank you for this wonderful contribution! Much better than `Int`.

PR #3586, opened 2020-05-14
Harden TestClock Implementation

adamgfraser:
In the implementation of `TestClock` we are relying on taking two consecutive "snapshots" of the status of descendent fibers and concluding that an effect is indefinitely suspended if all fibers are done / suspended in each snapshot and the snapshots are identical. In general this seems to be working quite well but we are seeing rare flakiness that seems to be related to situations where the two snapshots are the same but we actually aren't done yet. I am trying to resolve by adding a 1 millisecond delay between the snapshots.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/environment/ClockSpec.scala
@@ -22,7 +22,7 @@ object ClockSpec extends ZIOBaseSpec {
         _      <- adjust(11.hours)
         result <- ref.get
       } yield assert(result)(isTrue)
-    } @@ nonFlaky,
+    } @@ forked @@ nonFlaky,
     testM("sleep delays effect until time is adjusted") {
       for {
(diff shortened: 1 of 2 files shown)


adamgfraser:
Observing some strange behavior on Scala.js that I am trying to figure out. If I replace the definition of `suspended` with the following:

```scala
      /**
       * Returns whether all descendants of this fiber are done or suspended.
       */
      private lazy val suspended: UIO[Boolean] = {
        val zio = for {
          first <- freeze
          _     <- UIO(println("Going to sleep"))
          _     <- ZIO.sleep(1.milliseconds).provideLayer(Clock.live)
          _     <- UIO(println("Done sleeping"))
          last  <- freeze
        } yield first == last
        zio.orElseSucceed(false)
      }
```

"Going to sleep" is printed to the console but "Done sleeping" is never printed. Also, this issue doesn't occur if I run just the test at issue for "cached" in `ZIOSpec.scala` but if I run the entire test suite it fails.

Issue #3575, opened 2020-05-11
Why does collectAll use List?


borissmidt:
I was looking into the code and i i found it a bit weird to find out that a collectAll on a stream collects everything into a List and then reverts it (which looks really inefficient for large streams). I understand that you want to be 100% immutable but a list buffer wouldn't have leaked its state and might have been more efficient?

```
  def collectAll[A]: ZSink[Any, Nothing, Nothing, A, List[A]] =
    foldLeft[A, List[A]](List.empty[A])((as, a) => a :: as).map(_.reverse)
```
Also don't you return a Seq and then you are still able to change the collection to something else if it gives higher performance and make binary compatibility easier?

borissmidt:
Maybe an offtopic question,
Why is there no `collect` that allows me to just do a filter and map like Monix/scalaZ stream? its just unexpected that it throws when a match doesn't match. I use a collect often to 'flatmap' a stream with options to remove the None values.

jdegoes:
We want to switch to using `Chunk` as the default data type. 

Any interest in doing a PR? 😉

borissmidt:
Yea why not i'll have a try.

iravid:
Replacing the list with Chunk and ChunkBuilder there would be great 👍🏻

borissmidt:
I also see a simular collect all in the scheduler class, i assume that it is best to replace any place where there is a `foldleft` with a list in it?

mijicd:
@borissmidt bonus points if you propagate this change throughout the lib :). Do we want to replace the list-based methods or expose additional chunk-based methods? We can swap their implementation to use the chunk-based ones, and perhaps mark them as deprecated.

borissmidt:
Yes i will do it but on a part by part basis otherwise the PR becomes really big. I had a different branch where i started with most list parts. But then the scope became too big.
I was even hitting the engine itself.

borissmidt:
> @borissmidt bonus points if you propagate this change throughout the lib :).
> 
> Do we want to replace the list-based methods or expose additional chunk-based methods? We can swap their implementation to use the chunk-based ones, and perhaps mark them as deprecated.

In my opinion the default methods should give you the highest performance. 
i.e. if chunk is faster than List than vanilla collectAll should return that.
Then you could add the collectAllList. So it is a more verbose to use. 

Or better yet add a collectAllSeq then it is compatible with most existing scala code unless chunk already is a Seq?

[...]

PR #3587, opened 2020-05-14
ZSink.collectAll and Zstream.runAll use chunk

borissmidt:
Change ZSink.collectAll to return Chunk
Change Zstream.runAll to return Chunk

I think i left one todo for the ZIO.foreach that returns a list as well but it would make this PR too big.

There was on inconvenience with the current test system i only had 1 compilation error left 
but because it didn't check all test cases i had to compile many times with only a single error reported.

#3575

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
@@ -7,7 +7,7 @@ import zio._
 import zio.stream.SinkUtils.{ findSink, sinkRaceLaw }
 import zio.stream.ZStreamGen._
 import zio.test.Assertion.{ equalTo, isTrue, succeeds }
-import zio.test._
+import zio.test.{ assertM, _ }
 
 object ZSinkSpec extends ZIOBaseSpec {
   def spec = suite("ZSinkSpec")(
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -176,33 +176,6 @@ object ZStreamSpec extends ZIOBaseSpec {
                 .runCollect
             )(equalTo(Chunk(1 + 2 + 3)))
           },
-          // testM("managed") {
-          //   final class TestSink(ref: Ref[Int]) extends ZSink[Any, Throwable, Int, Int, List[Int]] {
-          //     type State = (List[Int], Boolean)
-
-          //     def extract(state: State) = UIO.succeedNow((state._1, Chunk.empty))
-
-          //     def initial = UIO.succeedNow((Nil, true))
-
-          //     def step(state: State, a: Int) =
-          //       for {
-          //         i <- ref.get
-          //         _ <- if (i != 1000) IO.fail(new IllegalStateException(i.toString)) else IO.unit
-          //       } yield (List(a, a), false)
-
-          //     def cont(state: State) = state._2
-          //   }
-
-          //   val stream = ZStream(1, 2, 3, 4)
-
-          //   for {
-          //     resource <- Ref.make(0)
-          //     sink     = ZManaged.make(resource.set(1000).as(new TestSink(resource)))(_ => resource.set(2000))
-          //     result   <- stream.aggregateManaged(sink).runCollect
-          //     i        <- resource.get
-          //     _        <- if (i != 2000) IO.fail(new IllegalStateException(i.toString)) else IO.unit
-          //   } yield assert(result)(equalTo(Chunk(List(1, 1), List(2, 2), List(3, 3), List(4, 4))))
-          // },
           testM("propagate managed error") {
             val fail = "I'm such a failure!"
             val t    = ZTransducer.fail(fail)
@@ -332,10 +305,12 @@ object ZStreamSpec extends ZIOBaseSpec {
         suite("bracket")(
           testM("bracket")(
             for {
-              done           <- Ref.make(false)
-              iteratorStream = ZStream.bracket(UIO(0 to 2))(_ => done.set(true)).flatMap(ZStream.fromIterable(_))
-              result         <- iteratorStream.runCollect
-              released       <- done.get
+              done <- Ref.make(false)
+              iteratorStream = ZStream
+                .bracket(UIO(0 to 2))(_ => done.set(true))
+                .flatMap(ZStream.fromIterable(_))
+              result   <- iteratorStream.runCollect
+              released <- done.get
             } yield assert(result)(equalTo(Chunk(0, 1, 2))) && assert(released)(isTrue)
           ),
           testM("bracket short circuits")(
@@ -351,10 +326,11 @@ object ZStreamSpec extends ZIOBaseSpec {
           ),
           testM("no acquisition when short circuiting")(
             for {
-              acquired       <- Ref.make(false)
-              iteratorStream = (ZStream(1) ++ ZStream.bracket(acquired.set(true))(_ => UIO.unit)).take(0)
-              _              <- iteratorStream.runDrain
-              result         <- acquired.get
+              acquired <- Ref.make(false)
+              iteratorStream = (ZStream(1) ++ ZStream.bracket(acquired.set(true))(_ => UIO.unit))
+                .take(0)
+              _      <- iteratorStream.runDrain
+              result <- acquired.get
             } yield assert(result)(isFalse)
           ),
           testM("releases when there are defects") {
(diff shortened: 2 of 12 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -481,10 +481,10 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
     new ZSink(push) {}
 
   /**
-   * A sink that collects all of its inputs into a list.
+   * A sink that collects all of its inputs into a chunk.
    */
-  def collectAll[A]: ZSink[Any, Nothing, A, List[A]] =
-    foldLeftChunks(Chunk[A]())(_ ++ (_: Chunk[A])).map(_.toList)
+  def collectAll[A]: ZSink[Any, Nothing, A, Chunk[A]] =
+    foldLeftChunks(Chunk[A]())(_ ++ (_: Chunk[A]))
```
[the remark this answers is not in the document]
borissmidt: Good catch, i'll see ik there is any other place with a similar mistake.

borissmidt:
> Taking a step back here, @borissmidt I'd like to see a benchmark of this first.
> 
> Building a list (and yes, reversing it) is very fast in Scala because the appends are O(1) and the reversal can happen with a pre-allocated buffer.
> 
> According to this: https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#construction-performance the only thing that is faster than prepending to a list is pre-allocating an array and filling it up, but we can't do that because we don't know up-front what is the size of the stream.
> 
> Another thing to consider when collecting to a chunk is the size limit: a ChunkBuilder will accumulate elements into an array, so this has a hard size limit at `2^31`. Not sure if this is a showstopper or not.

Maybe it is a better idea to change it to Seq then as the return type then you can change it on to something that would give the best performance without breaking binary compatibility? 
 
Since i can imagine that there are other better structures than an array that has to be copied every so often and a list that has a lot of indirection. Also in case of reversing it has to iterate over the whole list?

How could i best benchmark this? 

looking at the scala 2.13 source for the reverse it doesn't use a pre-allocated buffer so you need to iterate the full list.
```
  final override def reverse: List[A] = {
    var result: List[A] = Nil
    var these = this
    while (!these.isEmpty) {
      result = these.head :: result
      these = these.tail
    }
    result
  }
```

borissmidt:
I've updated the scala-bench project to use builders and will run them overnight.
It could be used to benchmark the `chunk` collection?

https://github.com/borissmidt/scala-bench

iravid:
@borissmidt I think it could be easier to just add another test case to `ChunkBenchmarks.scala` in this project.

borissmidt:
[scala-bench.zip](https://github.com/zio/zio/files/4640007/scala-bench.zip)
I was just curios my self to see if the array builder is faster than the normal list add operation and indeed it is. results are the average runtime in nano seconds of iteration of a run.

iravid:
Can you post the results here?

iravid:
You were testing list prepend, not append, right?

borissmidt:
Yes list prepend! the :: operation. you can view the code and run it yourself i've posted it.

iravid:
Right. I'm looking at the results of `ListReverse` and `ArrayBuilder` which is what we'd be using in practice. At 250k elements, the `List` becomes faster.

I think this is good evidence that `ChunkBuilder` and `Chunk` are good for the `runCollect` usecase!

borissmidt:
I think the added advantage with the chunk is that it's part of the
library. So you can improve it.

Maybe you could make it a linked list of arrays of at most 128k elements :)

simpadjo:
ZSink is expected to be the final stage of stream processing. If a user decided to collect a huge stream into a List/Chunk, it opted-in for having abnormally large objects in its program, and will face problems anyway. 
I don't think we should spend too much of an effort optimizing for this case.

@borissmidt you can get rid of linting errors by running `sbt prepare` (see https://github.com/zio/zio/blob/master/docs/about/contributing.md)

borissmidt:
There is still the open decision to keep the old Api and add the chunk as separate function calls.
Or replace them with chunk as i've done now.

borissmidt:
`"size can be modified locally" at src/test/scala/zio/test/GenSpec.scala:607` fails and i don't really understand why this is related to the changes i had made.

iravid:
@adamgfraser Can you peek at the test failure here and give us a hint?

adamgfraser:
@iravid Yes let me take a look.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -481,10 +500,11 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
     new ZSink(push) {}
 
   /**
-   * A sink that collects all of its inputs into a list.
+   * A sink that collects all of its inputs into a chunk.
    */
-  def collectAll[A]: ZSink[Any, Nothing, A, List[A]] =
-    foldLeftChunks(Chunk[A]())(_ ++ (_: Chunk[A])).map(_.toList)
+  def collectAll[A]: ZSink[Any, Nothing, A, Chunk[A]] =
+    foldLeftChunks[A, ChunkBuilder[A]](ChunkBuilder.make[A]())((b, chunk) => b ++= chunk)
```
adamgfraser, line 506: I believe the issue is that `ChunkBuilder.make()` is not wrapped in an effect constructor and is violating referential transparency.

borissmidt:
Ah so i'm sharing a reference between a mutable array because its passed by
value and not as a thunk, that sounds logical to me. I'll fix it

borissmidt:
I fixed, it but i'm not sure if it is a clean way of doing it, i tried t make a 'lazy chunkbuilder' that creates a new instance on the first += operation. But it wasn't possible with the typing of the return type.
An alternative could have been using Option. But then i wonder what the impact of the extra allocations would have been.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -481,10 +500,25 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
...
   /**
-   * A sink that collects all of its inputs into a list.
+   * A sink that collects all of its inputs into a chunk.
    */
-  def collectAll[A]: ZSink[Any, Nothing, A, List[A]] =
-    foldLeftChunks(Chunk[A]())(_ ++ (_: Chunk[A])).map(_.toList)
+  def collectAll[A]: ZSink[Any, Nothing, A, Chunk[A]] =
+    //Cannot make a builder that initializes itself on the first operation.
+    // falling back to null instead of an option/None combination
+    foldLeftChunks[A, ChunkBuilder[A]](null: ChunkBuilder[A])((b, chunk) =>
```
adamgfraser, line 508: Can we just wrap the normal `ChunkBuilder.make()` in `ZIO.effectTotal`? That is going to be the easiest way to solve this?

borissmidt: Then you are stuck with the `Sink[Any, Nohting,A, UIO[Chunkbuilder[A]]]` where you cannot get the Chunkbuilder out of the UIO

iravid: @borissmidt
```scala
ZSink {
  for {
    builder <- UIO(ChunkBuilder.make()).toManaged_
    foldingSink = foldLeftChunks(...)
    push <- foldingSink.push
  } yield push
}
```

streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
```diff
@@ -25,6 +25,17 @@ object ZSinkSpec extends ZIOBaseSpec {
             .run(ZSink.collectAllToMap((_: Int) % 3)(_ + _))
         )(equalTo(Map[Int, Int](0 -> 18, 1 -> 12, 2 -> 15)))
       ),
+//      testM("collectAllToMap referential transparancy") {
```
iravid, line 28: Can we decide on the fate of this test? Either it stays uncommented or it gets deleted :-)

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -160,21 +160,21 @@ object ZStreamSpec extends ZIOBaseSpec {
...
           },
           testM("with a sink that always signals more") {
             assertM(
               ZStream(1, 2, 3)
                 .aggregate(ZTransducer.fold(0)(_ => true)(_ + _))
                 .runCollect
-            )(equalTo(List(1 + 2 + 3)))
+            )(equalTo(Chunk(1 + 2 + 3)))
           },
           // testM("managed") {
```
iravid, line 179: Do you mind deleting the commented lines while you're here? 🙏🏻

borissmidt:
I think it should be ok now? i've added the missing merge conflicts.

streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
borissmidt: I'm just not well versed in the zio test libary yet to easily implement it. I need more experience first.

borissmidt:
I ran the fmt and fix command but there are no more changes detected. so i don't know why the fmtCheck still fails. Is there any reason why fix and fmt aren't ran during the build?

iravid:
Use the `prepare` command in sbt. I think the contributor guide mentions that.

The CI doesn't commit to Github, so it doesn't format the code there.

borissmidt:
Aha you have to do `sbt clean prepare` otherwise it has chached the formatting and will not do anything. I've more experience with gradle.

borissmidt:
Oke i fixed all issues

borissmidt:
@iravid Could this be merged before there is another merge conflict?

iravid:
Thank you @borissmidt!

PR #3589, opened 2020-05-15
Add ZStream right, some & orElse variants

luis3m:
Add useful transformations for error & successful channel just like ZIO.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1958,8 +1958,42 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    *
    * See also [[ZStream#catchAll]].
    */
-  def orElse[R1 <: R, E2, O1 >: O](that: => ZStream[R1, E2, O1])(implicit ev: CanFail[E]): ZStream[R1, E2, O1] =
+  final def orElse[R1 <: R, E2, O1 >: O](that: => ZStream[R1, E2, O1])(implicit ev: CanFail[E]): ZStream[R1, E2, O1] =
     catchAll(_ => that)
+
+  /**
+   * Switches to the provided stream in case this one fails with a typed error.
+   *
+   * See also [[ZStream#catchAll]].
+   */
+  final def orElseEither[R1 <: R, E2, O2](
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2133,6 +2167,21 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+    self.mapError(Some(_)).rightOrFail(None)
+
+  /**
+   * Fails with fiven error 'e' if value is `Left`.
+   */
+  final def rightOrFail[O1, O2, E1 >: E](e: => E1)(implicit ev: O <:< Either[O1, O2]): ZStream[R, E1, O2] =
+    self.flatMap(ev(_) match {
+      case Right(value) => ZStream.succeed(value)
+      case Left(_)      => ZStream.fail(e)
+    })
```
iravid, line 2183: Would it be possible to use `mapM` here? It'd be vastly more efficient as:
```suggestion
    self.mapM(ev(_) match {
      case Right(value) => ZIO.succeed(value)
      case Left(_)      => ZIO.fail(e)
    })
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2325,6 +2374,27 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+    map(_.getOrElse(default))
+
+  /**
+   * Extracts the optional value, or fails with the given error 'e'.
+   */
+  final def someOrFail[O2, E1 >: E](e: => E1)(implicit ev: O <:< Option[O2]): ZStream[R, E1, O2] =
+    self.flatMap(ev(_) match {
+      case Some(value) => ZStream.succeed(value)
+      case None        => ZStream.fail(e)
+    })
```
iravid, line 2396: Similarly to `rightOrFail` - could we use mapM here?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Let me take a look at that

luis3m:
@iravid as suggested, replaced `flatMap` with `mapM`

PR #3594, opened 2020-05-15
add ZTransducer#choose

simpadjo:
=========75%===   of #3495

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -1,5 +1,7 @@
 package zio.stream
 
+import java.util.UUID
+
 import scala.io.Source
 
 import zio._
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +222,12 @@ object ZTransducer {
       case None     => ZIO.succeedNow(Chunk.empty)
     }
 
+  /**
+   * Applies the partial function to the stream
+   */
+  def collect[I, O](f: I => Option[O]): ZTransducer[Any, Nothing, I, O] =
```
iravid, line 228:
```suggestion
  def collect[I, O](f: PartialFunction[I, O]): ZTransducer[Any, Nothing, I, O] =
```

simpadjo:
JS builds are failing but they are failing on master as well
@iravid btw, I no longer can see the logs of failing builds

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -120,6 +120,37 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+          if (ch.isEmpty) {
+            (Some(Chunk.empty), Some(Chunk.empty))
+          } else {
+            val leftBuilder: ChunkBuilder[I]   = ChunkBuilder.make()
+            val rightBuilder: ChunkBuilder[I1] = ChunkBuilder.make()
+            ch.foreach {
+              case Left(v)  => leftBuilder += v
+              case Right(v) => rightBuilder += v
+            }
+            (Some(leftBuilder.result()), Some(rightBuilder.result()))
```
iravid, line 144: Given the input `Chunk(Left(1), Right(1), Left(1), Right(1))`, this will produce `Chunk(Left(1), Left(1), Right(1), Right(1))`. That is- it reorders the elements.

This implementation should create chunks of consecutive left/right elements from `in` and feed them in order to `p1` and `p2`.

simpadjo: I'm not sure if this requirement could be formulated as a chunking-independent algebraic property. For me it seems that 2 transducers live in parallel timelines and we are could demand only internal output ordering.
Let me think about it a bit more.

regiskuckaertz: Seems to me this should hold at least:
```scala
stream.transduce(id choose id).runCollect == stream.runCollect
```
then probably this too:
```scala
stream.transduce(f choose g).transduce(h choose i).runCollect ==
  stream.transduce((f >>> h) choose (g >>> i)).runCollect
```

simpadjo: @regiskuckaertz great findings!
The second law being rewritten as `(f choose g) >>> (h choose i) == (f >>> h) choose (g >>> i)` looks even cooler

regiskuckaertz: Oh yeah nice 😍

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
```diff
@@ -26,6 +28,69 @@ object ZTransducerSpec extends ZIOBaseSpec {
...
+        testM("best-effort order preservation") {
+          val t = Transducer.identity[Int].choose(Transducer.identity[UUID])
+          checkM(Gen.listOf(Gen.chunkOf(Gen.either(Gen.anyInt, Gen.anyUUID)))) { chunks =>
+            val content = chunks.flatMap(_.toList)
+            val stream  = ZStream.fromChunks(chunks: _*)
+            for {
+              res <- stream.transduce(t).runCollect
+            } yield {
+              assert(res)(equalTo(content))
+            }
```
iravid, line 58:
```suggestion
stream.transduce(t).runCollect.map(assert(_)(equalTo(content)))
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -120,6 +121,51 @@ abstract class ZTransducer[-R, +E, -I, +O](val push: ZManaged[R, Nothing, Option
...
+      def impossible: Nothing = sys.error("impossible")
+      in.headOption match {
+        case Some(Left(_)) =>
+          val (init, rest) = in.span(_.isLeft)
+          groupSides(rest, acc += Left(Some(init.map(_.left.getOrElse(impossible)))))
+        case Some(Right(_)) =>
+          val (init, rest) = in.span(_.isRight)
+          groupSides(rest, acc += Right(Some(init.map(_.right.getOrElse(impossible)))))
+        case None => acc.result()
+      }
```
iravid, line 147: Deconstructing chunks like this is highly inefficient. Let's convert this to a fold or a plain `while` loop.

simpadjo:
@iravid I rebased it. Is it still needed?

jdegoes:
@simpadjo Thank you for your work on this. If you decide to pick this up again at any point in the future, please feel free to re-open.

Issue #3550, opened 2020-05-07
Tests: add generators and/or laws for testing stream chunking invariance


simpadjo:
Implementation code of streams/transducers/sinks deals with chunks a lot.
However, results of `ZStream#run` in 99% of cases must not depend on how the steam is chunked.
So it would be useful to be able to verify this property.

A testcase for inspiration:

```
testM("zip equivalence with Chunk#zipWith") {
            checkM(
              tinyListOf(Gen.chunkOf(Gen.anyInt)),
              tinyListOf(Gen.chunkOf(Gen.anyInt))
            ) { (l, r) =>
              val expected = Chunk.fromIterable(l).flatten.zip(Chunk.fromIterable(r).flatten)
              assertM(ZStream.fromChunks(l: _*).zip(ZStream.fromChunks(r: _*)).runCollect)(equalTo(expected.toList))
            }
          }
```

Api could look like that:
`dontCareAboutChunking(content: Iterable[A])(prop: Stream[A] => AssertionM): AssertionM`

This ticket includes
1) writing a `Gen` that generates partitionings of a given `Iterable`
2) actually writing a combinator

luis3m:
@simpadjo I just realized there's a `ZStreamGen`. Aren't `streamOfInts`, `pureStreamOfInts` and `streamGen` enough?

simpadjo:
@luis3m I expect that you'll need something to generate different chunkings from a given content.
Seems that splitting an `Iterable` by repeatedly applying `streamOfInt` would result in suboptimal test coverage.
Anyway, you can just go through stream tests, try to improve them and write what you need.

hearnadam:
Not pursued. Feel free to reopen.

PR #3596, opened 2020-05-16
Add ZStream#zipWithNext & ZStream#zipWithPrevious

luis3m:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2969,6 +2969,50 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+
+  /**
+  * Zips each element with the next element if present.
+  */
+  final def zipWithNext: ZStream[R, E, (O, Option[O])] =
+    ZStream {
+      for {
+        chunks <- self.process
+        ref    <- Ref.make[Option[O]](None).toManaged_
+        done   <- Ref.make(false).toManaged_
```
simpadjo, line 2981: The stream always terminates at the same moment as the original stream. 
Do we need `done` here?

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2415,6 +2415,34 @@ object ZStreamSpec extends ZIOBaseSpec {
             _     <- TestClock.setTime(210.milliseconds)
             value <- fiber.join
           } yield assert(value)(equalTo(List(0 -> 0, 0 -> 1, 1 -> 1, 1 -> 2)))
+        },
+        testM("zipWithNext") {
+          for {
+            result0 <- ZStream(1, 2, 3).zipWithNext.runCollect
```
simpadjo, line 2421: Each of 3 functions have a property that determines them
e.g. running `zipWithNext` is equivalent to running the original stream and performing `list.zip(list.tail)` afterwards.

Can it be turned into property-based tests?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @simpadjo yes and no. Your question made me really I can just set `ref` to `None` instead of having an extra reference for termination. OTOH, without `done` or `*> ref.set(None)` it would get stuck in an infinite loop. Updating in a few minutes.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: @simpadjo true that, will update these.

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @simpadjo done

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: @simpadjo done

iravid:
Hey @luis3m, did you convert the tests to property checks as @simpadjo suggested? Not seeing those changes.

luis3m:
@iravid I somehow missed that part, will update in a few.

luis3m:
@iravid added the new properties

simpadjo:
@luis3m could you please make test input randomly generated? 
Here is an example: https://github.com/zio/zio/blob/master/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala#L65

luis3m:
@simpadjo done

simpadjo:
Generating streams using `ZStream.fromChunks` gives better coverage (perhaps there should be an utility method for it). `ZStream.fromIterable` consists of a single chunk.

luis3m:
@simpadjo done (not sure what you mean by the utility method, something like `streamOf`?)

simpadjo:
Yes, something like `(Gen[A], Gen[Int]) => Gen[Stream[A]]`. Maybe `Gen[Int]` could be just hardcoded.
If you want to work on it, look at this issue as well: https://github.com/zio/zio/issues/3550

luis3m:
@simpadjo cool, I will probably work on this at some point this week

PR #3599, opened 2020-05-16
optimize ZStream.toInputStream

### LINKED ISSUE #3491: ZStream#toInputStream creates a suboptimal InputStream
The created InputStream only supports reading a single byte, whereas we could directly support reading entire arrays (the `read(Array[Byte])` variant) and checking whether a read can be completed without blocking (the `available` method).

simpadjo:
closes #3491

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -2291,7 +2291,56 @@ object ZStreamSpec extends ZIOBaseSpec {
             } yield assert(result)(equalTo(List(1, 2, 4, 5)))
           }
         ),
-        inputStreamSuite,
+        suite("toInputStream")(
+          testM("read one-by-one") {
+            checkM(tinyListOf(Gen.chunkOf(Gen.anyByte))) { chunks =>
+              val content = chunks.flatMap(_.toList)
+              ZStream.fromChunks(chunks: _*).toInputStream.use[Any, Throwable, TestResult] { is =>
+                ZIO.succeedNow(
+                  assert(Iterator.continually(is.read()).takeWhile(_ != -1).map(_.toByte).toList)(equalTo(content))
+                )
+              }
+            }
+          },
+          testM("read in batches") {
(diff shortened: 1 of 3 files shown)


streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2774,4 +2760,52 @@ object ZStreamSpec extends ZIOBaseSpec {
       testResult <- assertion(chunkCoordination)
     } yield testResult
 
+  private def inputStreamSuite = {
+    val bytes: List[Byte] = (-10 to 10).map(_.toByte).toList
+    val stream            = ZStream.fromIterable(bytes)
+
+    def testStream(assertion: java.io.InputStream => TestResult): ZIO[Any, Nothing, TestResult] =
+      ZIO
+        .foreach(List(1, 2, 5, 30)) { chunkSize =>
```
iravid, line 2769: Would it make sense to use a generator here?

streams/shared/src/main/scala/zio/stream/internal/ZInputStream.scala
```diff
@@ -0,0 +1,95 @@
+package zio.stream.internal
+
+import scala.annotation.tailrec
+
+import zio.Runtime
+import zio.{ Chunk, Exit, FiberFailure, ZIO }
+
+private[zio] class ZInputStream(chunks: Iterator[Chunk[Byte]]) extends java.io.InputStream {
+  private var current: Iterator[Byte]      = Iterator.empty
```
iravid, line 9: Can we get rid of this and keep the current chunk and offsets in a var? This will box the bytes.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
simpadjo: Added a generator and ended-up removing all my hand-written machinery.
Thanks!

simpadjo:
@iravid thanks for kind words!
CI fails due to unrelated js stuff. Could you please merge?

PR #3600, opened 2020-05-17
Simplify ZStream#catchAllCause

luis3m:
- Remove two atomic references
- Abstract over usage of stream

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -468,17 +468,17 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * stream is uninterruptible.
    */
   final def catchAllCause[R1 <: R, E2, O1 >: O](f: Cause[E] => ZStream[R1, E2, O1]): ZStream[R1, E2, O1] = {
-    sealed abstract class State
-    case object NotStarted                                              extends State
-    case class Self(finalizer: ZManaged.Finalizer, pull: Pull[R, E, O]) extends State
-    case class Other(pull: Pull[R1, E2, O1])                            extends State
+    sealed abstract class State[+E]
+    case object NotStarted                                                   extends State[Nothing]
+    case class Self[E](finalizer: ZManaged.Finalizer, pull: Pull[R1, E, O1]) extends State[E]
+    case class Other(pull: Pull[R1, E2, O1])                                 extends State[Nothing]
 
     ZStream {
       for {
         finalizers <- ZManaged.releaseMap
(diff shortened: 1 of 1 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -468,55 +468,32 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-            Cause.sequenceCauseOption(e) match {
-              case None    => Pull.end
-              case Some(c) => next(c)
-            }
-          }
+          def use[R, E0, O](stream: ZStream[R, E0, O])(state: (ZManaged.Finalizer, Pull[R, E0, O]) => State[E]) =
+            stream.process.zio
+              .provideSome[R]((_, finalizers))
+              .flatMap { case (release, pull) => ref.set(state(release, pull)) *> pull }
+              .uninterruptible
```
iravid, line 485: This was incorrectly marked as `uninterruptible` before. Good thing we're changing it :-)

Should be:
```suggestion
ZIO.uninterruptibleMask { restore =>
  restore(stream.process.zio)
    .provideSome[R]((_, finalizers))
    .flatMap { case (release, pull) => ref.set(state(release, pull)) *> restore(pull) }
}
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -468,55 +468,32 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-
-            case State.Self(release) =>
-              selfPull.get.flatten.catchAllCause(switch(release, _))
-
-            case State.Other =>
-              otherPull.get.flatten
+          def switch(release: ZManaged.Finalizer, e: Cause[Option[E]]): Pull[R1, E2, O1] =
+            Cause
+              .sequenceCauseOption(e)
+              .map(cause => release(Exit.fail(e)) *> use(f(cause))((_, pull) => Other(pull)))
```
iravid, line 490: This needs to be uninterruptible on the finalizer:
```suggestion
              .map(cause => release(Exit.fail(e)).uninterruptible *> use(f(cause))((_, pull) => Other(pull)))
```

It doesn't have to be entirely uninterruptible because if it gets interrupted before running `release`, the release map will still release it when exiting the ZManaged scope.

iravid:
Great simplifications here!

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Done, thanks!

luis3m:
@iravid done, thanks for the guidance!

PR #3607, opened 2020-05-17
Contributors guide: ask to squash commits

simpadjo:
Commits are squashed before merging to master. But resulting commit message sometimes consists of ~20 lines and it causes bad UX when browsing the code in IDE.

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -199,6 +199,14 @@ Code reviews can sometimes take a few days, because open source projects are lar
 
 Once you receive a review, you will probably have to go back and make minor changes that improve your contribution and make it follow existing conventions in the code base. This is normal, even for experienced contributors, and the rigorous reviews help ensure ZIO's code base stays high quality.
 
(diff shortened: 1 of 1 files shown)


simpadjo:
If you agree with the idea but don't like poor English - feel free to reword!

mijicd:
Commit by commit history is valuable. The UX problem you're mentioning can be addressed by nuking the squashed commit message, but I wouldn't just "sacrifice" the commit history from the pull request / branch itself.

simpadjo:
@mijicd I'm talking about cases like:

"Add feature"
"More"
"format"
"apply review suggestion"
"format"
"fix test"
...

mijicd:
@simpadjo these are plain bad, I agree, and can be addressed via rebasing (either your way or interactive which might be easier). However, the text suggests squashing commits every time you end up with more than one on branch :).

simpadjo:
@mijicd that's true. Although new contributors usually work on features that can be pushed atomically. So it's fine as a rule of thumb, I think.

Issue #3602, opened 2020-05-17
ZStream.debounce emit at end of stream


reibitto:
I expect the following to result in `List(3)`, but right now it's an empty list:

```scala
ZStream(1, 2, 3).fixed[Any](500.millis).debounce(1.second).runCollect
```

I think the end-of-stream case isn't being handled. @luis3m Do you know if that's the case?

luis3m:
@reibitto I will check this out in a few hours

iravid:
Thank you @luis3m!

luis3m:
@iravid @reibitto by just adding `if (chunk.isEmpty) Pull.empty else previous.interrupt *> store(chunk)` when the current fiber beats the previous one it works. I'm not sure why it's getting empty chunks though

PR #3611, opened 2020-05-17
Fix debounce to properly handle empty chunks

luis3m:
Resolves https://github.com/zio/zio/issues/3602


reibitto:
I think that fixes that particular issue. 👍 

However, I think there's something else:

```scala
ZStream("a", "b", "c").debounce(1.second).runCollect
```

This returns all the elements when I'd expect it to only return `c`. I think if you receive a chunk with more than 1 element you only want to return the last element from it, right?

luis3m:
@reibitto yes, I mentioned this particular difference in the previous PR I think. I can change this.

iravid:
@luis3m Do you want to address @reibitto's comment here or in a follow-up?

luis3m:
@iravid preferably a follow-up

PR #3617, opened 2020-05-18
Add ZIO#validateWith

luis3m:
Based on https://discord.com/channels/629491597070827530/630498701860929559/711700647455490141

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1896,7 +1896,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * specified combiner function. Combines the causes in case both effect fail.
    */
   final def validateWith[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] =
-    self.run.zipWith(that.run)(_ zip _).flatMap(ZIO.done(_)).map(f.tupled)
+    self.run.zipWith(that.run)(_.zipWith(_)(f, _ ++ _)).flatMap(ZIO.done(_))
 
   /**
    * Returns an effect that executes both this effect and the specified effect,
@@ -1904,7 +1904,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * both sides fail, then the cause will be combined.
    */
   final def validateWithPar[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] =
-    self.run.zipWithPar(that.run)(_ zip _).flatMap(ZIO.done(_)).map(f.tupled)
+    self.run.zipWithPar(that.run)(_.zipWith(_)(f, _ && _)).flatMap(ZIO.done(_))
 
   /**
    * The moral equivalent of `if (p) exp`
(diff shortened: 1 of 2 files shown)


adamgfraser:
I wonder if there is a family of combinators here that is more around "validation". It seems like the purpose of this combinator is really to collect all the potential error in a computation rather than failing fast, similar to the `Validation` data types in functional programming libraries and the existing `ZIO.validate` methods. The relationship between this and `ZIO.validate` is really analogous to the relationship between `zip` and `foreach`.

Maybe we call these `validateWith` and `validateWithPar`?

luis3m:
@adamgfraser updated

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1877,6 +1877,35 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
+   */
+  final def validatePar[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[R1, E1, (A, B)] =
+    validateWithPar(that)((_, _))
+
+  /**
+   * Sequentially zips this effect with the specified effect using the
+   * specified combiner function. Combines the causes in case both effect fail.
+   */
+  final def validateWith[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] =
+    self.run.zipWith(that.run)(_ zip _).flatMap(ZIO.done(_)).map(f.tupled)
```
adamgfraser, line 1899: We can save a couple of allocations here by pattern matching on the exit values in `zipWith` / `zipWithPar` and combining them into the desired type instead of doing the subsequent `flatMap` and `map`. We should be sure to combine the failures using `Cause.Then` in the sequential case and `Cause.Both` in the parallel case.

luis3m:
```scala
final def validateWith[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] =
    self.run.flatMap { exit =>
      that.run.flatMap { (exit, _) match {
        case (Exit.Success(a), Exit.Success(b)) => ZIO.succeedNow(f(a, b))
        case (Exit.Failure(l), Exit.Failure(r)) => ZIO.halt(Cause.Then(l, r))
        case (Exit.Failure(cause), _)           => ZIO.halt(cause)
        case (_, Exit.Failure(cause))           => ZIO.halt(cause)
      }}
    }

  final def validateWithPar[R1 <: R, E1 >: E, B, C](that: ZIO[R1, E1, B])(f: (A, B) => C): ZIO[R1, E1, C] =
    self.run.zipWithPar(that.run) {
      case (Exit.Success(a), Exit.Success(b)) => ZIO.succeedNow(f(a, b))
      case (Exit.Failure(l), Exit.Failure(r)) => ZIO.halt(Cause.Both(l, r))
      case (Exit.Failure(cause), _)           => ZIO.halt(cause)
      case (_, Exit.Failure(cause))           => ZIO.halt(cause)
    }.flatten
```
@adamgfraser like this?

adamgfraser: Yes though I think we can clean it up even more if we use `zipWith` on `Exit`. So we could do:

```scala
self.run.zipWith(that.run)(_.zipWith(_)(f, _ ++ _)).map(ZIO.done(_))
self.run.zipWithPar(that.run)(_.zipWith(_)(f, _ && _)).map(ZIO.done(_))
```

luis3m: Right, much better. Done. @adamgfraser

Issue #3598, opened 2020-05-16
ZStream#repeatEffectWith: allow schedule to rely on the effect value


simpadjo:
[Good first issue]

Current signature of the method is 

`def repeatEffectWith[R, E, A](fa: ZIO[R, E, A], schedule: Schedule[R, Any, _]): ZStream[R, E, A]`.

It could be 
`def repeatEffectWith[R, E, A](fa: ZIO[R, E, A], schedule: Schedule[R, A, _]): ZStream[R, E, A]`
instead.

For example, this would allow to ask an external source for data, check whether returned data is empty and if it is - add a delay before the next try.

Could be achieved by rewriting this method in terms of `ZStream.unfoldM`.

iravid:
This would be really nice. There are a bunch of other operators on ZStream that take schedules with an `Any` input. We should re-examine those too.

PR #3618, opened 2020-05-18
Allow schedule on ZStream.repeatEffectWith rely on effect value

luis3m:
Resolves https://github.com/zio/zio/issues/3598

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -2804,18 +2804,30 @@ object ZStreamSpec extends ZIOBaseSpec {
             } yield assert(res)(equalTo(List(1, 2, 3, 4)))
           )
         ),
-        testM("repeatEffectWith")(
-          for {
-            ref <- Ref.make[List[Int]](Nil)
-            fiber <- ZStream
-                      .repeatEffectWith(ref.update(1 :: _), Schedule.spaced(10.millis))
-                      .take(2)
-                      .runDrain
-                      .fork
-            _      <- TestClock.adjust(50.millis)
-            _      <- fiber.join
-            result <- ref.get
-          } yield assert(result)(equalTo(List(1, 1)))
+        suite("repeatEffectWith")(
+          testM("succeed")(
+            for {
+              ref <- Ref.make[List[Int]](Nil)
+              fiber <- ZStream
+                        .repeatEffectWith(ref.update(1 :: _), Schedule.spaced(10.millis))
+                        .take(2)
+                        .runDrain
+                        .fork
+              _      <- TestClock.adjust(50.millis)
+              _      <- fiber.join
+              result <- ref.get
+            } yield assert(result)(equalTo(List(1, 1)))
+          ),
+          testM("allow schedule rely on effect value")(checkM(Gen.int(1, 100)) { (length: Int) =>
+            var x        = length
+            val effect   = UIO(if (x <= 0) 0 else { val y = x; x -= 1; y })
+            val schedule = Schedule.identity.addDelay((i: Int) => (if (i == 0) 1 else 0).seconds).forever
+            for {
+              fiber  <- ZStream.repeatEffectWith(effect, schedule).take(length + 1L).runCollect.fork
+              _      <- TestClock.adjust(1.second)
+              result <- fiber.join
+            } yield assert(result)(equalTo((length to 0 by -1).toList))
+          })
         ),
         testM("unfold") {
           assertM(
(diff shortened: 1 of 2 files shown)


simpadjo:
Hey! Semantics of  `repeatWith` has changed. It used to update the schedule at the end of every chunk and now it updates it only if the stream was nonempty. No surprise, since any stream can produce `Any` but not every stream can produce `A`.

I think that `repeatEffectWith` is a better place to make changes. Unlike `repeatWith`, here we know that a substream emitted exactly one value and can pass the value to the schedule w/o losing any information about the stream.

iravid:
Maybe repeatWith’s schedule can take a Chunk[O] as input?

simpadjo:
I thought about it as well, but FWIW it exposes chunking to high level API.

luis3m:
@simpadjo @iravid done

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3479,11 +3479,10 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
    * Creates a stream from an effect producing a value of type `A` which repeats using the specified schedule
    */
-  def repeatEffectWith[R, E, A](
-    fa: ZIO[R, E, A],
-    schedule: Schedule[R, Any, _]
-  ): ZStream[R, E, A] =
-    fromEffect(fa).repeat(schedule)
+  def repeatEffectWith[R, E, A](effect: ZIO[R, E, A], schedule: Schedule[R, A, _]): ZStream[R, E, A] =
+    fromEffect(schedule.initial).flatMap { initial =>
+      unfoldM(initial)(state => effect.flatMap(value => schedule.update(value, state).option.map(_.map((value, _)))))
```
simpadjo, line 3484: Replacing `.option` with `.fold` will kill 2 `.map`s

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2804,18 +2804,30 @@ object ZStreamSpec extends ZIOBaseSpec {
...
+                        .take(2)
+                        .runDrain
+                        .fork
+              _      <- TestClock.adjust(50.millis)
+              _      <- fiber.join
+              result <- ref.get
+            } yield assert(result)(equalTo(List(1, 1)))
+          ),
+          testM("allow schedule rely on effect value")(checkM(Gen.int(1, 100)) { (length: Int) =>
+            var x        = length
```
simpadjo, line 2822: If you want to countdown from `length` to `0` you can, for example, allocate a `Ref[Int]` and use `ref.getAndUpdate`.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2804,18 +2804,30 @@ object ZStreamSpec extends ZIOBaseSpec {
...
+            val effect   = UIO(if (x <= 0) 0 else { val y = x; x -= 1; y })
+            val schedule = Schedule.identity.addDelay((i: Int) => (if (i == 0) 1 else 0).seconds).forever
+            for {
```
simpadjo, line 2825: As far as I can see, this test doesn't check timings despite it uses Clock.
I feel that we can achieve the same coverage with a simpler testcase.
What about, e.g.:
1) take an effect that counts down from 100 to 0, but crashes on 51
2) pick a schedule that stops when it sees `50`
3) check that it collected (100..50)
4) optionally the schedule might as well crash on elements below 49

simpadjo:
Looks good! But let's try to simplify the test.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: Right, should've thought about that

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: @simpadjo let me give that a try

luis3m: @simpadjo thanks for the suggestions, done.

luis3m:
@simpadjo test has been simplified as suggested

simpadjo:
@luis3m great work!

The last comment: if PR description was "Closes #3598" instead of "Addresses #3598", merging this PR would authomatiacally close the issue. It's good to keep the issue list clean.

luis3m:
@simpadjo oh, I didn't know that. Good to know and thanks :)

PR #3631, opened 2020-05-18
Add withFilter in ZStream

paulpdaniels:
Adds `withFilter` to `ZStream` so that we can use guard clauses in for-comprehensions


luis3m:
As a note, for better or worse the following code would filter None out:
```scala
for {
  value            <- stream0
  Some(other) <- stream1
} yield (value, other)
```

paulpdaniels:
@luis3m expected though right? I assume that doesn't even compile now?

luis3m:
@paulpdaniels I think that behavior makes sense here, yes.

PR #3663, opened 2020-05-20
fix ZSink#collectAllWhileWith

### LINKED ISSUE #3657: ZSink#collectAllWhileWith always returns zero value
RC18-2:
```
@ Stream(1,2,3,5).run(ZSink.sum[Int].collectAllWhileWith(_ => true)(-1)(_ + _)).unsafeRun  
res30: Int = 10
```

RC19-2:
```
@ Stream(1,2,3,5).run(ZSink.sum[Int].collectAllWhileWith(-1)(_ => true)(_ + _)).unsafeRun 
res19: Int = -1
```

simpadjo:
Closes #3657

Not sure how useful this combinator is with the new encoding.
- Its implementation is shorter than the declaration
- it uses ZSink as a Transducer without a huge disclaimer
- (consequently) it has no meaningful laws

FWIW, here is the fix


simpadjo:
sorry, found a bug

PR #3664, opened 2020-05-20
Add ZStream constructor from an OutputStream writer

joroKr21:
This is useful for interop with Java APIs that can write to
`OutputStream`.

I open this so I don't forget.
I will come back for tests later, but in the mean time open for feedback and bikeshedding.

### DIFF
--- a/streams/jvm/src/main/scala/zio/stream/platform.scala
+++ b/streams/jvm/src/main/scala/zio/stream/platform.scala
@@ -1,9 +1,8 @@
 package zio.stream
 
-import java.io.{ IOException, InputStream, OutputStream, Reader }
+import java.io._
 import java.net.InetSocketAddress
-import java.nio.channels.FileChannel
-import java.nio.channels.{ AsynchronousServerSocketChannel, AsynchronousSocketChannel, CompletionHandler }
+import java.nio.channels.{ AsynchronousServerSocketChannel, AsynchronousSocketChannel, CompletionHandler, FileChannel }
 import java.nio.file.StandardOpenOption._
 import java.nio.file.{ OpenOption, Path }
 import java.nio.{ Buffer, ByteBuffer }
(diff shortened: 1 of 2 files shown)


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -213,6 +213,36 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
       .managed(is)
       .flatMap(fromInputStream(_, chunkSize))
 
+  /**
+   * Creates a stream from a callback that writes to [[java.io.OutputStream]].
+   * Note: the input stream will be closed after the `write` is done.
+   */
+  def fromOutputStream(
```
iravid, line 220: Most methods that capture effects in ZIO are called `effect<something>`. Should we call this `effectOutputStream`? Or maybe `fromOutputStreamWriter`?

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -213,6 +213,36 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+    write: OutputStream => Unit,
+    chunkSize: Int = ZStream.DefaultChunkSize
+  ): ZStream[Blocking, Throwable, Byte] = {
+    def from(in: InputStream, out: OutputStream, err: Promise[Throwable, None.type]) = {
+      val readIn = fromInputStream(in, chunkSize).ensuring(ZIO.effectTotal(in.close()))
+      val writeOut = ZStream.fromEffect {
+        blocking
+          .effectBlocking(write(out))
```
iravid, line 228: Would it make sense to use `blocking.effectBlockingInterrupt` to be able to interrupt this?

streams/jvm/src/main/scala/zio/stream/platform.scala
joroKr21: I think I like `fromOutputStreamWriter` more (because of the symmetry with `fromInputStream`) - but I doubt we will be able to come up with a good name in general. So as long as it has `OutputStream` in it (for auto completion) I am fine.

streams/jvm/src/main/scala/zio/stream/platform.scala
joroKr21: Oh yeah probably, I didn't know about this 👍

streams/jvm/src/main/scala/zio/stream/platform.scala
iravid: `fromOutputStreamWriter` it is!

joroKr21:
Finally came back to this 😄 
Fortunately still pre 1.0.0

joroKr21:
@iravid you forgot to merge I think - rebased now

iravid:
@joroKr21 Can you run `sbt prepare` on the rebased branch?

joroKr21:
> @joroKr21 Can you run sbt prepare on the rebased branch?

What the hell, it runs out of memory...

joroKr21:
Done now

joroKr21:
> What the hell, it runs out of memory...

Indeed looks like scalafix 0.9.18 has a terrible performance regression: scalacenter/sbt-scalafix#142

joroKr21:
@adamgfraser I can't reproduce the flaky test locally - does it happen only on a certain JVM?

adamgfraser:
@joroKr21 Hmmm. Did you remove the `flaky` annotation and add `nonFlaky`? With `nonFlaky(1000)` I was able to reproduce it pretty quickly. I'm on `2.13.5`.

joroKr21:
> Did you remove the flaky annotation and add nonFlaky? With nonFlaky(1000) I was able to reproduce it pretty quickly. I'm on 2.13.5.

Yes, that's what I tried. I'm on Mac, usually running GraalVM 11, but I tried also AdoptJDK 11 and 16 - does it happen on Java 8?

adamgfraser:
I'm not sure what the difference is. I'm on Mac, JDK 11. This test, right?

```
[info] - ZStream JVM
[info]   - Constructors
[info]     - fromOutputStreamWriter
[info]       - is not affected by closing the output stream
```

adamgfraser:
@joroKr21 I think #5027 may be related to this.

joroKr21:
@adamgfraser the only difference between this and other tests of the same method is that the output stream is closed by the write function (which can happen when you can't control it, i.e. calling some library that takes an output stream and closes it). That function is already wrapped in blocking so I'm at a loss here.

PR #3665, opened 2020-05-20
fix ZSink#collectAllWhileWith (second attempt)

### LINKED ISSUE #3657: ZSink#collectAllWhileWith always returns zero value

simpadjo:
Closes #3657

Not sure how useful this combinator is with the new encoding.
It uses ZSink as a Transducer without a huge disclaimer, (consequently) it has no meaningful laws

FWIW, here is the fix

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -78,27 +78,13 @@ abstract class ZSink[-R, +E, -I, +Z] private (
    * using the stepping function `f`.
    */
   def collectAllWhileWith[S](z: S)(p: Z => Boolean)(f: (S, Z) => S): ZSink[R, E, I, S] =
-    ZSink {
-      Push.restartable(push).flatMap {
-        case (push, restart) =>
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -78,27 +78,13 @@ abstract class ZSink[-R, +E, -I, +Z] private (
...
-                    state
-                      .updateAndGet(f(_, z))
-                      .flatMap(s =>
-                        if (p(z)) restart
-                        else ZIO.fail(Right(s))
-                      )
-                }
-            }
-          }
+    self.toTransducer >>> ZSink
```
simpadjo, line 81: Could be cleaner if we had 
`takeWhile(p: A => Boolean): Transducer[A, Option[A]]`

PR #3670, opened 2020-05-20
Refactor ZSink#untilOutputM

simpadjo:


simpadjo:
This refactoring no longer works after adding leftovers to sinks. Putting it on hold.
But I believe that another world is possible, so not closing it.

simpadjo:
No, this is wrong now

PR #3673, opened 2020-05-21
FiberPools

iravid:
Converted ZStream#mapMPar so far. Will do `ZStream#flatMapPar/Switch` in a follow-up.

### DIFF
--- a/core/shared/src/main/scala/zio/FiberPool.scala
+++ b/core/shared/src/main/scala/zio/FiberPool.scala
@@ -1,6 +1,6 @@
 package zio
 
-import FiberPool.{ ShutdownType, State }
+import FiberPool.{ Shutdown, State }
 
 import scala.collection.immutable.{ Queue => IQueue }
 
@@ -78,14 +78,14 @@ class FiberPool(state: Ref[State], workerLimit: Long, defectHandler: Cause[Nothi
     }
 
   /**
-   * Shuts down the fiber pool. Behavior is dependent on the specified [[ShutdownType]].
+   * Shuts down the fiber pool. Behavior is dependent on the specified [[Shutdown]].
    */
-  def shutdown(shutdownType: ShutdownType): UIO[Any] =
+  def shutdown(shutdownType: Shutdown): UIO[Any] =
     state.modify { s =>
       shutdownType match {
-        case ShutdownType.ShutdownNow =>
+        case Shutdown.Immediate =>
           (Fiber.interruptAll(s.workers), s.copy(shutdown = Some(shutdownType)))
-        case ShutdownType.DrainRunning | ShutdownType.DrainPending =>
+        case Shutdown.DrainRunning | Shutdown.DrainPending =>
           (
             ZIO.foreach(s.freeWorkers)(_.apply(ZIO.interrupt)) *> Fiber.awaitAll(s.workers),
             s.copy(shutdown = Some(shutdownType))
@@ -97,7 +97,7 @@ class FiberPool(state: Ref[State], workerLimit: Long, defectHandler: Cause[Nothi
     (task *> workerLoop).forkDaemon.tap { worker =>
       state.modify { s =>
         s.shutdown match {
-          case Some(ShutdownType.ShutdownNow) => (worker.interrupt, s)
+          case Some(Shutdown.Immediate) => (worker.interrupt, s)
           case _                              => (UIO.unit, s.copy(workers = worker :: s.workers))
         }
       }.flatten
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Submits a new task to the pool. This method will semantically block until a fiber is available
+   * to start the task. Defects caused by the task will be handled by the pool's defect handler.
+   *
+   * If the pool is already shutdown, this method will interrupt its caller.
+   */
+  def submit[R](task: URIO[R, Any]): URIO[R, Any] =
+    ZIO.environment[R].flatMap { r =>
+      Promise.make[Nothing, Unit].flatMap { taskReceived =>
```
jdegoes, line 28: Good point to specify fiber id when you make the promise?

core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+        val wrappedTask = task.catchAllCause(defectHandler(_).sandbox.ignore).provide(r)
```
jdegoes, line 29: You can omit `.ignore` for better performance.

core/shared/src/main/scala/zio/FiberPool.scala
iravid: This is the same situation where we don't know which fiber is going to resolve it, because it's going to be resolved by one of the fibers in the pool. 🤷🏽‍♂️

core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+                      freeWorkers,
+                      pendingTasks.enqueue(taskReceived.succeed(()).as(wrappedTask)),
+                      shutdownType,
+                      workers
+                    )
+                  )
+                }
+              case _ => (ZIO.interrupt, s)
+            }
+        }.flatten
```
jdegoes, line 75: What happen if it is interrupted after the state change, but before executing the effect produced by the state change?

core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+    }
+
+  /**
+   * Shuts down the fiber pool. Behavior is dependent on the specified [[Shutdown]].
+   */
+  def shutdown(shutdownType: Shutdown): UIO[Any] =
```
jdegoes, line 83: We should make sure we can replicate this to `Queue#shutdown` so it's not just a one-off data type.

jdegoes: On returning `Any`, I think it's best to:

 - Allow users to return `Any` in callbacks, etc.
 - Always return `Unit` ourselves, unless for justifiable performance reasons

Because a `Unit` type is clearer as an output from a ZIO method.

core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+    state.modify { s =>
+      shutdownType match {
+        case Shutdown.Immediate =>
+          (Fiber.interruptAll(s.workers), s.copy(shutdown = Some(shutdownType)))
+        case Shutdown.DrainRunning | Shutdown.DrainPending =>
+          (
+            ZIO.foreach(s.freeWorkers)(_.apply(ZIO.interrupt)) *> Fiber.awaitAll(s.workers),
```
jdegoes, line 90: `foreach_ ` is much faster if you don't need the results.

core/shared/src/main/scala/zio/FiberPool.scala
```diff
@@ -0,0 +1,184 @@
...
+            s.copy(shutdown = Some(shutdownType))
+          )
+      }
+    }.flatten
```
jdegoes, line 94: As before, for all of these, I'd ask: what happens if it's interrupted after the state change, but before the consequent effect is executed (or during its execution)? In many cases it may be benign, of course.

core/shared/src/main/scala/zio/FiberPool.scala
iravid: Thanks for pointing that out. That will lead to lost workers. Will fix that.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1819,19 +1819,18 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   final def mapMPar[R1 <: R, E1 >: E, O2](n: Int)(f: O => ZIO[R1, E1, O2]): ZStream[R1, E1, O2] =
     ZStream[R1, E1, O2] {
       for {
-        out     <- Queue.bounded[ZIO[R1, Option[E1], O2]](n).toManaged(_.shutdown)
-        permits <- Semaphore.make(n.toLong).toManaged_
+        out  <- Queue.bounded[ZIO[R1, Option[E1], O2]](n).toManaged(_.shutdown)
+        pool <- FiberPool.make(n.toLong).toManaged_
```
jdegoes, line 1823: This looks way nicer! ❤️

iravid:
Will come back to this in a few days. Needs a bit of investigation because something is still off.

PR #3676, opened 2020-05-21
[WIP] experiments with non-empty chunks in ZConduit

simpadjo:


simpadjo:
@iravid almost there, just couple of things to clean up. Could you confirm that it worth pursuing?

iravid:
The changes look fairly minimal so it looks promising. I'd like to see `ZTransducer/ZSink.wrapPush` disappear. Is that possible?

simpadjo:
I'm afraid no. ZStream and ZTransducer output is potentially empty, but downstream consumer accepts only non-empty input. Someone has to gatekeep empty chunks.

iravid:
Right, so I'd be interested if that can be confined to ZStream: `ZStream#transduce` and `ZStream#run` should use the `pullNonEmpty` trick to always pull non-empty chunks.

simpadjo:
Good idea. 
We need wrapping for Stream-Trans, Stream-Sink, Trans-Trans and Trans-Sink compositions.
`pullNonEmpty` is a local method that is already used twice. Let's promote it to utility method and it will do the job for Stream-Trans and Stream-Sink cases.

simpadjo:
@iravid Ok, could you please take a look once again?
I looks not as appealing as I thought when I started.
I removed wrappers for transducers and sinks (I personally think they would make the code cleaner tho). 
There is some cruft in `splitLines` test, but it's addressed here: https://github.com/zio/zio/pull/3694

jdegoes:
@iravid Please re-open if necessary, will close for now!

PR #3677, opened 2020-05-21
#3649 - Add ZStream.fromTcpSocketServer

### LINKED ISSUE #3649: Add ZStream.fromTcpSocketServer
A stream that binds a listening TCP socket and emits a stream of connections wrapped in ZManaged. Something like:
```
def fromSocketServer(<params>): ZStream[R, Throwable, ZManaged[R, Throwable, Connection]]
```

This ticket also involves defining the `Connection` representation. At a minimum, it should sport the following methods:
```
def read: ZStream[Any, Throwable, Byte]
def write: ZSink[Any, Throwable, Byte, Unit]
```

regis-leray:
closes #3649


iravid:
@regis-leray Nice work. This is looking good.

Can we switch to using `AsynchronousServerSocketChannel` here? See https://stackoverflow.com/questions/22177722/java-nio-non-blocking-channels-vs-asynchronouschannels for reasoning.

regis-leray:
@iravid thanks for the the tips asynchronous, i tested locally it works. missing test and it will be ready to merge

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,81 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+                       override def failed(exc: Throwable, attachment: Void): Unit = callback(ZIO.fail(exc))
+                     }
+                   )
+                 }
+                 .map(c => Option(c -> s))
+             }
+
+    } yield Managed.make(ZIO.succeed(conn))(c => ZIO.effectTotal(c.close()))
+
+  private[zio] class Connection(socket: AsynchronousSocketChannel) {
```
iravid, line 278: This should probably be public.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,81 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
     ZStream.fromJavaIteratorTotal(stream.iterator())
+
+  /**
+   * Create a stream of accepted connection from server socket
+   * Emit socket `Connection` from which you can read / write and ensure it is closed after it is used
+   */
+  def fromSocketServer(
+    port: Int,
+    host: Option[String] = None
+  ): ZStream[Blocking, Throwable, Managed[Throwable, Connection]] =
```
iravid, line 250: We're not using `blocking` anywhere, but it's possible that we should: we need to audit the methods we're calling on the channel and socket and see which ones might block. `AsynchronousServerSocketChannel.open/bind` look suspicious, for example, so we might have to switch to the blocking pool there.

The methods that take a completion handler look safe though.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,81 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+              new CompletionHandler[Integer, Void] {
+                override def completed(result: Integer, attachment: Void): Unit = callback(ZIO.succeed(result.toInt))
+
+                override def failed(error: Throwable, attachment: Void): Unit = callback(ZIO.fail(error))
+              }
+            )
+          }
+      }
+
+    private[zio] def close(): Unit = socket.close()
```
iravid, line 317: `UIO[Unit]` please! Can be public too.

streams/jvm/src/main/scala/zio/stream/platform.scala
iravid: Let's give this a companion object with a make function that returns a `ZManaged`.

streams/jvm/src/main/scala/zio/stream/platform.scala
regis-leray: i forgot to cleanup the method. Good catch `open/bind` should be wrap in blocking pool.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,85 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+    for {
+      server <- ZStream.managed(ZManaged.fromAutoCloseable(blocking.effectBlocking {
+                 AsynchronousServerSocketChannel
+                   .open()
+                   .bind(
+                     host.fold(new InetSocketAddress(port))(new InetSocketAddress(_, port))
+                   )
+               }))
+
+      conn <- ZStream.unfoldM(server) { s =>
```
iravid, line 260: This can just be `ZStream.repeatEffect(IO.effectAsync(...))`.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,85 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+              }
+            )
+          }
+      }
+
+    def close(): UIO[Unit] = ZIO.effectTotal(socket.close())
+  }
+
+  object Connection {
+    def managed(conn: Connection): UManaged[Connection] = Managed.make(ZIO.succeed(conn))(_.close())
```
iravid, line 321:
```suggestion
    def make(conn: Connection): UManaged[Connection] = Managed.make(ZIO.succeed(conn))(_.close())
```

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -236,4 +239,84 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
     ZStream.fromJavaIteratorTotal(stream.iterator())
+
+  /**
+   * Create a stream of accepted connection from server socket
+   * Emit socket `Connection` from which you can read / write and ensure it is closed after it is used
+   */
+  def fromSocketServer(
+    port: Int,
+    host: Option[String] = None
+  ): ZStream[Blocking, Throwable, UManaged[Connection]] =
```
mschuwalow, line 250: What do you think about using ZManaged#scope to pull the manageds into the stream itself?
This leads to a far more ergonomic signature and is still resource safe as all connections will be closed when the stream is closed.

Somethinghttps://github.com/zio/zio-keeper/blob/master/keeper/src/main/scala/zio/keeper/transport/tcp.scala like this

iravid: Yes I think that'd be good. `ZStream[Connection]` is indeed nicer than `ZStream[Managed[Connection]]`. What do you think @regis-leray?

regis-leray: 👍

iravid:
We just need a test or two after this @regis-leray and this should be good to go.

regis-leray:
@iravid currently we are supporting 
- ReadConnection  which is Stream[Throwable, Byte] by calling `read()`
- WriteConnection  which is  Sink[Throwable, Byte, Int] by calling `write()`

It can be challenging for user to know what connection as been accepted, like this you can filter and only treat specific request

```scala
sealed trait Connection
  
object Connection{
   class ConnectionRead extends Connection {
      def read: Stream[Throwable, Byte] =
   }

  class ConnectionWrite(socket: AsynchronousSocketChannel) extends Connection {
      def write: Sink[Throwable, Byte, Int] = ???
  }
}
```

what do you think ?

regis-leray:
@iravid sorry for the delay. I provided some test.

regis-leray:
@iravid i added some javadoc

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -302,4 +304,107 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
+                   new CompletionHandler[AsynchronousSocketChannel, Void]() {
+                     self =>
+                     override def completed(socket: AsynchronousSocketChannel, attachment: Void): Unit =
+                       callback(ZIO.succeed(Connection.make(socket)))
+
+                     override def failed(exc: Throwable, attachment: Void): Unit = callback(ZIO.fail(exc))
+                   }
+                 )
+               }
+             } >>= (ZStream.managed(_))
```
iravid, line 338: This is subtly wrong, because it means that as soon as you progress to the next connection, the previous connection is closed. Instead, let's use `ZManaged.scope` here:
```scala
for {
  server <- ZStream.managed(...)
  registerConnection <- ZStream.managed(ZManaged.scope)
  conn <- ZStream.repeatEffect {
                 IO.effectAsync(...).flatMap(managedConn => registerConnection(managedConn).map(_._2))
                }
} yield conn
```

regis-leray: Thanks for the tip ! i was not aware of this `ZManaged.scope`

iravid:
Great work @regis-leray, thank you!

PR #3678, opened 2020-05-21
Fix ZStream.fromInputStream

iravid:
cc @reibitto

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -767,6 +767,8 @@ object Chunk {
 
   /**
    * Returns a chunk backed by an array.
+   *
+   * WARNING: The array must not be mutated after creating the chunk.
    */
   def fromArray[A](array: Array[A]): Chunk[A] =
     (if (array.isEmpty) Empty
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -1,5 +1,6 @@
 package zio.stream
 
+import java.io.ByteArrayInputStream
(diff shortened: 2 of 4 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -767,6 +767,8 @@ object Chunk {
 
   /**
    * Returns a chunk backed by an array.
+   *
+   * WARNING: The array must not be mutated after creating the chunk.
```
reibitto, line 771: I wonder if it would make sense to have a naming convention to serve the same purpose as this warning. `Chunk.fromArray` could copy by default and then the alternate "unsafe" one would be the fast version. Depends if there is already a precedent for that in the ZIO codebase.

Also when seeing this, I also thought about if there might be a use for a `Chunk.fromArray[A](array: Array[A], start: Int, len: Int)` version.

simpadjo:
Maybe I'm too late for the party, but here is my take:

```
val x: ZStream[Blocking, IOException, Byte] = ZStream.fromEffect(
            blocking.effectBlocking({
              val buf = Array.ofDim[Byte](chunkSize)
              val res = is.read(buf)
              if(res < 0) {
                Chunk.empty
              } else{
                val arr = if(res < chunkSize){
                  buf.take(res)
                } else buf
                Chunk.fromArray(arr)
              }
            }).refineToOrDie[IOException]).takeWhile(_.nonEmpty).flattenChunks
```

core/shared/src/main/scala/zio/Chunk.scala
iravid: @reibitto I like the slicing version! Could be a nice PR if you have the cycles.

Re the naming: I think a warning is sufficient because we're not going to add the other variant. The caller can always defensively copy if required.

PR #3684, opened 2020-05-22
NonEmptyChunk as newtype

simpadjo:
Thanks @adamgfraser for a newtype template.

I'm not very happy with the implementation:
-  `import NonEmptyChunk._` is required to call any method
- I had to move some methods of `Chunk` into `ChunkSyntax` extension.
And decision which method should be moved and which should stay was based on fixing compile errors rather than reasoning.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -1,6 +1,6 @@
 package zio
 
-import NonEmptyChunk._
+import NonemptyChunkModule._
 
 import zio.random.Random
 import zio.test.Assertion._
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -1,10 +1,11 @@
 package zio
 
+import NonEmptyChunk._
+
 import zio.random.Random
 import zio.test.Assertion._
 import zio.test.TestAspect.exceptScala211
 import zio.test._
-import NonEmptyChunk._
 
 object ChunkSpec extends ZIOBaseSpec {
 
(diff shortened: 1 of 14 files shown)


adamgfraser:
```scala
package zio.experimental

import zio.Chunk

object Experiment {

  private[zio] trait NonEmpty[F[+_]] {
    type Type[+A]
    def wrap[A](chunk: F[A]): Type[A]
    def unwrap[A](newtype: Type[A]): F[A]
  }

  object NonEmptyChunk {

    private[zio] val newtype: NonEmpty[Chunk] =
      new NonEmpty[Chunk] {
        type Type[+A] = Chunk[A]
        def wrap[A](chunk: Chunk[A]): Type[A]     = chunk
        def unwrap[A](newtype: Type[A]): Chunk[A] = newtype
      }

    private def wrap[A](chunk: Chunk[A]): NonEmptyChunk[A] =
      newtype.wrap(chunk)

    def unwrap[A](chunk: NonEmptyChunk[A]): Chunk[A] =
      newtype.unwrap(chunk)

    def apply[A](a: A, as: Chunk[A]): NonEmptyChunk[A] =
      wrap(Chunk(a) ++ as)

  }

  type NonEmptyChunk[+A] = NonEmptyChunk.newtype.Type[A]

  implicit class NonEmptyChunkSyntax[+A](private val self: NonEmptyChunk[A]) extends AnyVal {
    def first: A = NonEmptyChunk.unwrap(self)(0)
  }
}

object Test extends scala.App {
  val as = Experiment.NonEmptyChunk(1, Chunk(2, 3, 4))
  val result = as.first
  println(result)
}
```

I think you need to define the implicit syntax class in the same scope as you define the type. See example above

simpadjo:
Sorry, doesn't work so far. I put everything directly under `object NonemptyChunkModule`.
But when used from the outside, extension methods are not seen w/o `import NonemptyChunkModule._`

PR #3687, opened 2020-05-22
add NonEmptyChunk.fold

simpadjo:
and minor improvements in type signatures

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -553,6 +553,12 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
     }
 
   /**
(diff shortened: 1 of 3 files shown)


iravid:
@adamgfraser can you chime in on the name?

adamgfraser:
It seems strange. Normally `fold` would be a method on a data type and here it is not a method on a data type but the data type we are "folding" is `Chunk` instead of `NonEmptyChunk` and we are folding over the hypothetical empty and nonempty cases instead of the elements as we normally would for a collection.

Maybe make it a method on `Chunk` and call it `nonEmptyOrElse`?

simpadjo:
Agree!

simpadjo:
@adamgfraser @iravid done

PR #3689, opened 2020-05-22
Modify ZStream#debounce to take chunk last element

luis3m:
Follow-up https://github.com/zio/zio/pull/3611 (cc @iravid @reibitto)


reibitto:
Looks like this solves the problem to me. 👍

PR #3690, opened 2020-05-22
Fix ZTransducer.splitLines

### LINKED ISSUE #3659: ZTransducer.splitLines fails with OutOfMemoryError
I encounter an `OutOfMemoryError` when using `ZTransducer.splitLines`. However I can read the full file content into a String using Apache Commons `FileUtils.readFileToString`.

I would expect a streaming library to not consume more heap space when processing a file line by line than the file length.
The file has a size of ~107 MB (110,499.590 Bytes) with 1568 lines. The largest line is a String of ~224,000 characters.
JVM InitialHeapSize is set to 512MB, MaxHeapSize to 4GB (shouldn't matter since FileUtils can read the file to String).

```scala
for {
      file <- args.headOption
        .map(filePath => IO(Paths.get(filePath)))
        .getOrElse(IO.fail("No file specified"))

      inputStream <- IO(Files.newInputStream(file))
      //_ <- console.putStrLn("reading file to memory using FileUtils.readFileToString...")
      //s = FileUtils.readFileToString(file.toFile, Charset.forName("UTF-8"))
      //_ <- console.putStrLn(s"lenght of file in memeory: ${s.length}")
      stream = ZStream.fromInputStream(inputStream)
        .transduce(ZTransducer.utf8Decode)
        .transduce[ZEnv, IOException, String](ZTransducer.splitLines)
      _ <- stream.runDrain
} yield ()
```

```
java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3332)
    at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
    at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:448)
    at java.lang.StringBuilder.append(StringBuilder.java:136)
    at zio.stream.ZTransducer$.$anonfun$splitLines$6(ZTransducer.scala:635)
    at zio.stream.ZTransducer$.$anonfun$splitLines$6$adapted(ZTransducer.scala:634)
    at zio.stream.ZTransducer$$$Lambda$205/1628294067.apply(Unknown Source)
    at zio.Chunk$Singleton.foreach(Chunk.scala:1146)
    at zio.Chunk$Concat.foreach(Chunk.scala:1124)
    at zio.stream.ZTransducer$.$anonfun$splitLines$4(ZTransducer.scala:634)
```

I'm willing to fix this bug myself, with some help.
From the implementation I'd suggest to also replace String concat in `ZTransducer.splitLines` with a `StringBuilder` stored as local mutable state. However I'm not sure if there is a restriction on using local state in `ZTransducers`.

iravid:
Fixes #3659.

We were already pulling the leftover into the `carry`, so concatenating it to the chunk was redundant.

I was going to write a smarter generator to catch this, but ran out of time.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -362,6 +362,14 @@ object ZTransducerSpec extends ZIOBaseSpec {
             )(equalTo("bc"))
           }
         },
+        testM("handles leftovers 2") {
+          assertM(
+            ZStream
+              .fromChunks(Chunk("aa", "bb"), Chunk("\nbbc\n", "ddb", "bd"), Chunk("abc", "\n"), Chunk("abc"))
+              .transduce(ZTransducer.splitLines)
+              .runCollect
+          )(equalTo(List("aabb", "bbc", "ddbbdabc", "abc")))
+        },
         testM("aggregates chunks") {
           ZTransducer.splitLines.push.use { push =>
             for {
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -631,7 +631,7 @@ object ZTransducer {
                 var inCRLF = wasSplitCRLF
                 var carry  = leftover getOrElse ""
 
-                (Chunk.fromIterable(leftover) ++ strings).foreach { string =>
+                strings.foreach { string =>
                   val concat = carry + string
 
                   if (concat.length() > 0) {
@@ -775,8 +775,8 @@ object ZTransducer {
 
               val (toConvert, newLeftovers) = concat.splitAt(computeSplit(concat))
 
-              if (toConvert.isEmpty) (Chunk.empty, newLeftovers)
-              else (Chunk.single(new String(toConvert.toArray[Byte], "UTF-8")), newLeftovers)
+              if (toConvert.isEmpty) (Chunk.empty, newLeftovers.materialize)
+              else (Chunk.single(new String(toConvert.toArray[Byte], "UTF-8")), newLeftovers.materialize)
             }
         }
       }


iravid:
@simpadjo Could you review this?

simpadjo:
@iravid LGTM!
I tried to write a property-based test for it. That's what I came up with so far:
```

testM("pbt"){
          val breakGen: Gen[Random, String] = Gen.elements("'\r\n", "\n", "", "", "")
          val stringWithBreaks: Gen[Random with Sized, String] = Gen.string(Gen.printableChar).zipWith(breakGen)(_ ++ _)
          val g: Gen[Random with Sized, Chunk[String]] = Gen.listOf(stringWithBreaks).map(Chunk.fromIterable(_))
          val gen = Gen.listOf(g).map(Chunk.fromIterable(_))
          checkM(gen) { lines =>
            ZStream.fromChunks(lines).flattenChunks.transduce(ZTransducer.splitLines).runCollect.map(res => {
              println(s"${res.length}  " + res)
              val input = lines.map(_.mkString).mkString
              val expected = Source.fromString(input).getLines().toList
              assert(res)(equalTo(expected))
            })
          }
        }
```

But it doesn't catch this particular issue. Maybe you could improve the generator.

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
```diff
@@ -362,6 +362,14 @@ object ZTransducerSpec extends ZIOBaseSpec {
             )(equalTo("bc"))
           }
         },
+        testM("handles leftovers 2") {
+          assertM(
+            ZStream
+              .fromChunks(Chunk("aa", "bb"), Chunk("\nbbc\n", "ddb", "bd"), Chunk("abc", "\n"), Chunk("abc"))
+              .transduce(ZTransducer.splitLines)
+              .runCollect
+          )(equalTo(List("aabb", "bbc", "ddbbdabc", "abc")))
```
regiskuckaertz, line 371:
```suggestion
          )(equalTo(List("aabb", "theguardian", "ddbbdabc", "abc")))
```
🤣

iravid: 🤣

iravid:
@simpadjo I will follow up with a better generator. If you want to pick that up, the idea is to:
- Generate a list of strings - keep that around as `target`
- Randomly intersperse "\n", "\r\n" or ""
- `.mkString` on the list
- Generate a random list of indices on which to split the string
- Use `.grouped(n)` with a random chunk size to generate the chunks
- Create a stream from the resulting list of chunks, `transduce(splitLines).runCollect` and compare that to `target`

simpadjo:
@iravid not planning to work on it right now. Btw, logic of chunking is related to this ticket https://github.com/zio/zio/issues/3550

Issue #3682, opened 2020-05-21
Replace the implementation of ZStream#merge with one that doesn't race effects


iravid:
We're currently avoiding interruption of pulls in ZStream because most operators aren't interruption safe. While this is obviously bad and we should fix it at some point, our alternative strategy of using `raceWith` and keeping around the losing fiber can cause memory leaks over time (see e.g. #3679 and scalapb/zio-grpc#63 which I suspect are about this).

For this ticket, we need to replace the implementation of `merge` which currently races two pulls and emits the one that succeeded, with an implementation that runs both streams in the background and pipes them into a queue. The front of the stream is then a fiber that pulls from that queue.

This is pretty similar to `ZStream.mergeAll(left, right)` but we need a separate implementation to maintain the termination strategies.

iravid:
cc @luis3m if you're interested!

luis3m:
@iravid cool, I will take this one

PR #3691, opened 2020-05-23
Modify ZStream#mergeWith implementation to use a queue

luis3m:
Closes https://github.com/zio/zio/issues/3682

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1991,7 +1991,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
       import TerminationStrategy.{ Left => L, Right => R, Either => E }
 
       for {
-        queue   <- Queue.unbounded[Take[E1, O3]].toManaged(_.shutdown)
+        handoff <- ZStream.Handoff.make[Take[E1, O3]].toManaged_
         done    <- RefM.makeManaged[Option[Boolean]](None)
         chunksL <- self.process
         chunksR <- that.process
@@ -2006,11 +2006,11 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
                     case (state @ Some(true), _) =>
                       ZIO.succeedNow((false, state))
                     case (state, Right(chunk)) =>
-                      queue.offer(Take.chunk(chunk)).as((true, state))
+                      handoff.offer(Take.chunk(chunk)).as((true, state))
                     case (_, Left(Some(cause))) =>
-                      queue.offer(Take.halt(cause)).as((false, Some(true)))
+                      handoff.offer(Take.halt(cause)).as((false, Some(true)))
                     case (option, Left(None)) if terminate || option.isDefined =>
-                      queue.offer(Take.end).as((false, Some(true)))
+                      handoff.offer(Take.end).as((false, Some(true)))
                     case (None, Left(None)) =>
                       ZIO.succeedNow((false, Some(false)))
                   }
@@ -2020,7 +2020,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
         }
         _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).forkManaged
         _ <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).forkManaged
-      } yield queue.take.flatMap(_.done)
+      } yield handoff.take.flatMap(_.done)
     }
 
   /**
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2016,9 +2016,9 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
             }
             .repeat(Schedule.doWhileEquals(true))
         }
-        fiberL <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).forkManaged
-        fiberR <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).forkManaged
-      } yield queue.take.flatMap(_.done).tapError(_ => fiberL.interrupt *> fiberR.interrupt)
+        _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).forkManaged
+        _ <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).forkManaged
+      } yield queue.take.flatMap(_.done)
     }
 
   /**
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1992,7 +1992,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
 
       for {
         handoff <- ZStream.Handoff.make[Take[E1, O3]].toManaged_
-        done    <- RefM.makeManaged[Option[Boolean]](None)
+        done    <- Ref.makeManaged[Option[Boolean]](None)
         chunksL <- self.process
         chunksR <- that.process
         handler = (pull: Pull[R1, E1, O3], terminate: Boolean) =>
@@ -2004,17 +2004,17 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
                 done.modify {
                   (_, exit.fold(c => Left(Cause.sequenceCauseOption(c)), Right(_))) match {
                     case (state @ Some(true), _) =>
-                      ZIO.succeedNow((false, state))
+                      ZIO.succeedNow(false) -> state
                     case (state, Right(chunk)) =>
-                      handoff.offer(Take.chunk(chunk)).as((true, state))
+                      handoff.offer(Take.chunk(chunk)).as(true) -> state
                     case (_, Left(Some(cause))) =>
-                      handoff.offer(Take.halt(cause)).as((false, Some(true)))
+                      handoff.offer(Take.halt(cause)).as(false) -> Some(true)
                     case (option, Left(None)) if terminate || option.isDefined =>
-                      handoff.offer(Take.end).as((false, Some(true)))
+                      handoff.offer(Take.end).as(false) -> Some(true)
                     case (None, Left(None)) =>
-                      ZIO.succeedNow((false, Some(false)))
+                      ZIO.succeedNow(false) -> Some(false)
                   }
-                }
+                }.flatten
               }
           }.repeat(Schedule.doWhileEquals(true))
         _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).fork.toManaged_
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1992,7 +1992,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
 
       for {
         handoff <- ZStream.Handoff.make[Take[E1, O3]].toManaged_
-        done    <- Ref.makeManaged[Option[Boolean]](None)
+        done    <- RefM.makeManaged[Option[Boolean]](None)
         chunksL <- self.process
         chunksR <- that.process
         handler = (pull: Pull[R1, E1, O3], terminate: Boolean) =>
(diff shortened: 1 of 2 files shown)


luis3m:
I will update at some point tomorrow to fix Dotty compilation issue

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+                  case (None, Left(None)) =>
+                    ZIO.succeedNow((false, Some(false)))
+                }
+              }
+            }
+            .repeat(Schedule.doWhileEquals(true))
         }
+        fiberL <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).forkManaged
+        fiberR <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).forkManaged
+      } yield queue.take.flatMap(IO.done(_)).tapError(_ => fiberL.interrupt *> fiberR.interrupt)
```
regiskuckaertz, line 2037: I don't think fibers should be interrupted on error since a stream can resume after producing an error. But if the stream is interrupted from the outside that's another story. However since upstream can terminate, why not query the `done` flag then:
```suggestion
      } yield done.get.flatMap(done =>
        if (done.contains(true)) Take.end 
        else queue.take.flatMap(IO.done(_)).onInterrupt(_ => fiberL.interrupt *> fiberR.interrupt))
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-              )
-          }
+      for {
+        queue   <- Queue.unbounded[Take[E1, O3]].toManaged(_.shutdown)
+        done    <- RefM.make[Option[Boolean]](None).toManaged_
+        chunksL <- self.process
+        chunksR <- that.process
+        handler = (pull: Pull[R1, E1, O3], terminate: Boolean) => {
+          pull
+            .foldCause(c => Left(Cause.sequenceCauseOption(c)), Right(_))
```
regiskuckaertz, line 2016: It looks like you could avoid some allocation by just using `run` here and doing a fold over the result in the modify section below
```suggestion
            .run
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                      termination[E, O](leftDone, true, l),
-                      termination[E1, O2](true, rightDone, r)
-                    )
-                },
-                chunk => ZIO.succeedNow(Exit.succeed((chunk, (leftDone, rightDone, Some(loser)))))
-              )
-          }
+      for {
+        queue   <- Queue.unbounded[Take[E1, O3]].toManaged(_.shutdown)
+        done    <- RefM.make[Option[Boolean]](None).toManaged_
```
regiskuckaertz, line 2011: I think you can get by with a `Ref` here.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+        chunksL <- self.process
+        chunksR <- that.process
+        handler = (pull: Pull[R1, E1, O3], terminate: Boolean) => {
+          pull
```
regiskuckaertz, line 2015: I think you should check `done` before pulling to avoid pulling more than necessary (e.g. when the other fiber is done and the termination strategy is `Either`).

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+            .foldCause(c => Left(Cause.sequenceCauseOption(c)), Right(_))
+            .flatMap { either =>
+              done.modify {
```
regiskuckaertz, line 2018: Inside here, there is no effect that needs to be weaved in the modification of the ref to determine its next value, that is why a `Ref` will do.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,40 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                    loser.fold(
-                      termination[E, O](leftDone, true, l),
-                      termination[E1, O2](true, rightDone, r)
-                    )
-                },
-                chunk => ZIO.succeedNow(Exit.succeed((chunk, (leftDone, rightDone, Some(loser)))))
-              )
-          }
+      for {
+        queue   <- Queue.unbounded[Take[E1, O3]].toManaged(_.shutdown)
```
regiskuckaertz, line 2010: Do we want to leave this unbounded? If upstream is faster than downstream, it may eat up memory. Perhaps use the same strategy as in aggregateAsync.

iravid: Definitely needs to be bounded. 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: The `onInterrupt` is actually not needed either, because the fibers are forked with `forkManaged`, so that will take care of the interruption when the ZManaged scope ends.

luis3m: @regiskuckaertz wouldn't that change throw away elements in the queue that happened before the end of the stream and haven't been pulled yet?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz `ZStream.Handoff` you mean?

luis3m: @regiskuckaertz updated

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz I couldn't make tests pass with `get` and `set` instead of `modify`

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: ditto

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz done, thanks

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Well, replaced the queue with a Handoff anyway

regiskuckaertz: @luis3m right you are, I guess you could `poll` the queue in the consequent instead of just `Take.end` 🤔

luis3m:
Taking a look why `zipWithLatest` fails

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,42 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                    loser.fold(
-                      termination[E, O](leftDone, true, l),
-                      termination[E1, O2](true, rightDone, r)
-                    )
-                },
-                chunk => ZIO.succeedNow(Exit.succeed((chunk, (leftDone, rightDone, Some(loser)))))
-              )
-          }
+      for {
+        handoff <- ZStream.Handoff.make[Take[E1, O3]].toManaged_
```
iravid, line 2010: This is actually an interesting approach we could take for all concurrent combinators. It's easier to reason about because it doesn't introduce any implicit buffers. If the user wants a buffer, they can easily add one with `.buffer`.

@regiskuckaertz

luis3m:
@iravid by changing `. forkManaged` with `.toManaged_.fork` I got `zipWithLatest` passing as it was hanging. Not sure whether I need to add back interruption for fibers now.

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: You can still use modify if the left projection of the pair is an effect, e.g.
```scala
pull.run.flatMap { exit =>
  done.modify { done =>
    handoff.offer(exit) -> Some(true)
  }
}
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: Yes! and Handoff is quite nice as it acts as a semaphore between the two fibers as well.

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz interesting, I tried this out but all related tests are timing out. I suspect because it's changing the state after offering to Handoff

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz thanks to your other suggestion it isn't a queue anymore 😅

streams/shared/src/main/scala/zio/stream/ZStream.scala
regiskuckaertz: whoops I forgot the flatten 😅 amended

luis3m: @regiskuckaertz right, I should've figure it out. Updating.

luis3m: @regiskuckaertz done, thank you

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2002,58 +2002,41 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+                      handoff.offer(Exit.fail(None)).as(false) -> Some(true)
+                    case (None, Left(None)) =>
+                      ZIO.succeedNow(false) -> Some(false)
+                  }
+                }.flatten
+              }
+          }.repeat(Schedule.doWhileEquals(true))
+        _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).fork.toManaged_
+        _ <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).fork.toManaged_
+      } yield handoff.take.flatMap(IO.done(_))
```
regiskuckaertz, line 2038: hmm, we may have to add a `poll` to `Handoff`. Say both fibers are done and downstream is pulling, this will block ad infinitum; the contract for streams is that they should just keep on emitting `Take.end`. What do you think?

regiskuckaertz:
Great work! I think it's almost there, there's that end of stream invariant to sort out but you may have a better idea than I did.

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @regiskuckaertz good point, will make it `poll` when done is true.

luis3m:
Solved conflicts

luis3m:
@regiskuckaertz I'd been trying to figure out why the following code wasn't working.
```scala
for {
  done  <- done.get
  take   <- if (done.contains(true)) handoff.poll.some.orElseFail(None) else handoff.take
  result <- take.done
} yield result
```
By moving back to `RefM` I got it working as it will change the state after it offers to the queue, not before as the current implementation does.

luis3m:
1) Updated with requested change about `poll`
2) Added a test case which fails with `Ref`. Passing with `RefM`

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1986,53 +1986,47 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+                    case (option, Left(None)) if terminate || option.isDefined =>
+                      handoff.offer(Take.end).as((false, Some(true)))
+                    case (None, Left(None)) =>
+                      ZIO.succeedNow((false, Some(false)))
+                  }
+                }
+              }
+          }.repeat(Schedule.doWhileEquals(true))
+        _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).fork.toManaged_
+        _ <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).fork.toManaged_
```
iravid, line 2021:
```suggestion
        _ <- handler(chunksL.map(_.map(l)), List(L, E).contains(strategy)).forkManaged
        _ <- handler(chunksR.map(_.map(r)), List(R, E).contains(strategy)).forkManaged
```

Tests pass with this change, and we get the benefit of knowing that the running streams are definitely interrupted before their ZManaged finalizers run.

iravid: Ah `zipWithLatest` is failing. I'll check why...

luis3m: @iravid yes, I had that one failing with `forkManaged`

luis3m: Pulling was hanging for some reason. Both fibers were stuck and the test was timing out

iravid: Looks like a bug in Handoff, I think. Digging into it.

luis3m: `zipWithLatest` isn't hanging on JS and it does hang on the JVM even with an unbounded queue.

luis3m: I think `Handoff` (or `Queue`) are unrelated here. It's hanging on `pull.run` (or at least was when I tried yesterday before making more suggested changes) which hasn't offered anything yet at that point to the queue/handoff.

iravid: Ah it might be something about the TestClock and the fact that `forkManaged` uses `forkDaemon`.

iravid: Hmm no. Changing forkDaemon in ZManaged#fork back to fork doesn't fix it either.

iravid: Well it's definitely something with the TestClock. The test hangs in the schedule's sleep.

luis3m: Are you looking into it or is this something Adam should check?

iravid: Yes, found the problem. We're just using it wrong. The `adjust` call happens concurrently with the fibers sleeping in the stream. Checking how best to fix it now. Basically the same thing as `assertWithChunkCoordination` fixes.

iravid: Oh I must have not saved the file between running the tests. Changing `forkDaemon -> fork` in `ZManaged#fork` does fix the problem.

Now I'm pretty sure it's because the TestClock doesn't see the producer fibers suspending.

For now @luis3m let's change this from `.fork.toManaged_` to `.fork.interruptible.toManaged(_.interrupt)`. This works correctly.

luis3m: @iravid done

PR #3694, opened 2020-05-24
make Transducer.splitLines test blackbox

simpadjo:


simpadjo:
Errors are from flaky tests

simpadjo:
@iravid if you like it, `utf8decode` and maybe `splitOn` tests could be simplified the same way.

PR #3696, opened 2020-05-24
ZIO Test: Close Resources Earlier In ProvideLayerShared

adamgfraser:
Right now we are keeping resources open too long with `provideLayerShared`. Basically we are building one giant `ZManaged` for all the suites and specs in a test and then running them so the result is we are closing all the shared resources at the end of the spec instead of at the end of the suite they are provided to. To address this I added a `release` combinator on `ZManaged` that conceptually "closes" the scope.

This works but looking at this the `only` combinator is quite problematic. It requires us to traverse the entire spec, which means acquiring all the resources in any nested suites, to determine whether there is any test annotated with `only`. For example if a spec is to be run sequentially we need to acquire and release the resources for suites one by one but we also need to potentially examine all of the suites and acquire their resources to determine which ones we should run. I have removed it for now though it is a very nice combinator and I would like to figure out a way to support it.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -773,6 +773,15 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
     }
 
   /**
+   * Runs all the finalizers associated with this scope. This is useful to
+   * conceptually "close" a scope when composing multiple managed effects.
+   * Note that this is only safe if the result of this managed effect is valid
+   * outside its scope.
+   */
+  def release: ZManaged[R, E, A] =
+    ZManaged.fromEffect(useNow)
+
+  /**
    * Retries with the specified retry policy.
    * Retries are done following the failure of the original `io` (up to a fixed maximum with
    * `once` or `recurs` for example), so that that `io.retry(Schedule.once)` means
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -772,6 +772,15 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
         .applyOrElse[A, ZManaged[R1, E1, A]](v, ZManaged.succeedNow)
     }
 
+  /**
+   * Runs all the finalizers associated with this scope. This is useful to
+   * conceptually "close" a scope when composing multiple managed effects.
+   * Note that this is only safe if the result of this managed effect is valid
+   * outside its scope.
+   */
+  def release: ZManaged[R, E, A] =
```
iravid, line 781: I think there's an issue here in that `a *> b.release *> c` won't work, as after `b.release`, the ReleaseMap is already closed.

The workaround may be for this combinator to run `self` on a nested `ReleaseMap`.

adamgfraser: Added a test for this. Seems to be working as expected:

```scala
testM("closes the scope") {
  val expected = Chunk("acquiring a", "acquiring b", "releasing b", "acquiring c", "releasing c", "releasing a")
  for {
    ref     <- Ref.make[Chunk[String]](Chunk.empty)
    a       = Managed.make(ref.update(_ + "acquiring a"))(_ => ref.update(_ + "releasing a"))
    b       = Managed.make(ref.update(_ + "acquiring b"))(_ => ref.update(_ + "releasing b"))
    c       = Managed.make(ref.update(_ + "acquiring c"))(_ => ref.update(_ + "releasing c"))
    managed = a *> b.release *> c
    _       <- managed.useNow
    log     <- ref.get
  } yield assert(log)(equalTo(expected))
}
```

Issue #3646, opened 2020-05-19
Add ZStream.fromFile


iravid:
Signature TBD. Open to suggestions.

This should use Java NIO and support:
- reading complete file
- reading range
- tailing the file

Specifying the desired chunk size for every read operation should also be supported.

The returned stream should be a stream of bytes that opens the file, emits the bytes and closes the file when the stream ends.

luis3m:
@iravid I would like to give this one a try but can't promise a solution as I'm not familiar with NIO. Anyway, I'm wondering how the tailing part would work, tail the file for as long the stream is open? Would that be a separate function?

I would also need clarification about reading range if possible :)

iravid:
@luis3m Could be nice to unify the signatures with a `ReadMode = Complete | Range(Long, Long) | Tail` config

mtjon:
@iravid Are you still working on this since merging the initial 'complete file' version in #3697 ?

I would be interested in picking this up if noone is working on it.

Question: 
I see you use `nio.FileChannel` in your complete file implementation. What's the intent for the partial reader? Should the range refer to a startpoint and length in the stream or a line range in the file?
Should the other variants also work based on FileChannels or would the `Files` class from NIO be OK to use?

iravid:
Hey @mtjon, I'm not working on the other variants currently. If @luis3m doesn't mind it's a great ticket to work on.

For the range option, it's an offset and length. All the variants should re-use the same filechannel for efficiency rather than use `Files` which would open/close the file on every pull.

luis3m:
@mtjon feel free to take this one :)

mtjon:
Hadn't replied yet, but I'm on it (in my off hours).  ✌️

ogresto:
@mtjon is it in progress?

mtjon:
@ogresto got caught up in a project and honestly kind of forgot about it. I'd like to take a look at it properly soon. But if you think you know how to solve it, feel free to take the issue from me.

cherkaskyb:
@iravid @ogresto looking for my first contribution, can I have a crack at this one?

adamgfraser:
Closing as this has already been implemented.

PR #3697, opened 2020-05-25
Add ZStream.fromFile

iravid:
Refs #3646

Just the simple whole-file version currently.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
@@ -1,11 +1,12 @@
 package zio.stream
 
+import java.nio.file.{ Files, NoSuchFileException, Paths }
+
 import scala.concurrent.ExecutionContext.global
 
-import zio.ZQueueSpecUtil.waitForSize
+import zio._
 import zio.test.Assertion._
-import zio.test.{ testM, _ }
-import zio.{ Chunk, _ }
+import zio.test._
 
 object ZStreamPlatformSpecificSpec extends ZIOBaseSpec {
   def spec = suite("ZStream JVM")(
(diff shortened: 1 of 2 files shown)


streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
```diff
@@ -174,22 +175,16 @@ object ZStreamPlatformSpecificSpec extends ZIOBaseSpec {
...
-                    )
-                    .map(_.reverse)
-                    .fork
-          _     <- waitForSize(queue, -1)
-          _     <- queue.shutdown
-          items <- fiber.join
-        } yield assert(items)(equalTo(c.toSeq.toList))
-      })
+      testM("fromFile") {
+        val data = List.fill(76)("a").mkString
```
simpadjo, line 179: We are are pulling the same data every time. Since we are dealing with mutable buffers, it makes the test less powerful.
Maybe `data = (1 to 100).mkString` ?

streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
```diff
@@ -174,22 +175,16 @@ object ZStreamPlatformSpecificSpec extends ZIOBaseSpec {
...
+
```
simpadjo, line 180: Also would be great to have a negative test. E.g. reading from a missing file

streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
iravid: Good point!

iravid:
The fromQueue test somehow ended up in the platform-specific file, whereas it exists in the cross-platform tests :-)

luis3m:
LGTM as well

PR #3699, opened 2020-05-25
ZIO Test: Make ExecutedSpec Its Own Data Type

adamgfraser:
Makes `ExecutedSpec` its own data type instead of being an alias for `Spec`. This way we can say that executed specs contain actual results instead of effects that produce results so we can simplify code related to rendering test results. This should also allow us to be more flexible in refactoring `Spec` since we won't have to refactor test reporting for any changes we make. Still needs some cleanup.


jdegoes:
@adamgfraser I love this. However, we also talked about removing effects from specs, and instead placing layers in them. 

Aren't these alternatives in some ways?

adamgfraser:
@jdegoes I think it depends on whether we can unify specs in a way that places layers in them and doesn't use effects anywhere.

It seems like to do this we would need something like:

```scala
final case class Spec[+R, +E, +T](caseValue: SpecCase[R, E, T, Spec[R, E, T]]

object Spec {

  trait SpecCase[-R, +E, +T, +A[-_]]

  final case class SuiteCase[-R, +E, A[-_], S](label: String, specs: Vector[A[S]], layer: Layers[R, E, S]) extends SpecCase[R, E, Nothing, A]

  final case class TestCase[+T](label: String, test: T, annotations: TestAnnotationMap) extends SpecCase[R, E, T, Nothing]
}

// Layers is data type that reifies a set of layers with subtypes for composition, shared layers, and per test layers.
```

It seems like we need a higher kind for the recursive case to express the way the layer needs to provide the environment required for the recursive case, which is not great.

So my thinking had been to get this PR in to handle the `ExecutedSpec`, then work on moving `ZSpec` to the layers encoding, and then we could see if there was an opportunity to unify, though I'm also open to trying to do everything at once.

jdegoes:
Can't believe I forgot about this! 😱

adamgfraser:
No worries! Just resolved conflicts but also fine to deal with this after 1.0.

PR #3701, opened 2020-05-25
ZStream#toInputStream laziness

simpadjo:


iravid:
So I think this means we can revert the previous change to `toInputStream`?

simpadjo:
No, that one is still needed. This one just suspends the first chunk, the previous one catches errors

iravid:
If I'm not mistaken, the change there only catches errors thrown by pulling the first chunk, no?

simpadjo:
No, it catches all errors (turns them into `E` instead of dying). `"Preserves errors in the middle"` testcase verifies this.

iravid:
@simpadjo Your tests are passing even when that change is reverted. The reason is that the `Task` is done running by the time you're inside `use`, so it does not affect the actual runtime of the InputStream.

simpadjo:
:thinking: not sure I totally get it so far. Do you mean that coverage is weak (and it's still possible to `die`)  or that the first patch is indeed redundant?

iravid:
No, the coverage is fine. It's just that now that `ZInputStream` is constructed lazily (due to the `Iterator.empty ++` tweak), `ZInputStream.fromPull` will not throw when given a failed stream.

I can expand more later if needed.

simpadjo:
No worries, just revert the redundant commit.

simpadjo:
@iravid reverted the previous fix as you suggested

PR #3712, opened 2020-05-27
Add ZSink.fromFile functionality (#3647)

### LINKED ISSUE #3647: Add ZSink.fromFile
A ZSink that reads chunks of bytes and writes them to a file.

Should support starting to write at a certain offset in the file.

lvitaly:
Resolves #3647 issue.


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -29,6 +32,37 @@ trait ZSinkPlatformSpecificConstructors { self: ZSink.type =>
...
+                    blocking.effectBlockingInterrupt(FileChannel.open(path, options.asJava).position(position)).orDie
+                  )(chan => blocking.effectBlocking(chan.close()).orDie)
+        push = (is: Option[Chunk[Byte]]) =>
+          is match {
+            case None => state.get.flatMap(Push.emit)
+            case Some(byteChunk) =>
+              for {
+                bytesWritten <- state.get
+                justWritten <- blocking.effectBlockingInterrupt {
+                                Files.write(path, byteChunk.toArray)
```
iravid, line 58: This is not needed, right? 😅

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -29,6 +32,37 @@ trait ZSinkPlatformSpecificConstructors { self: ZSink.type =>
...
+                                channel.write(ByteBuffer.wrap(byteChunk.toArray))
+                              }.mapError(Left(_))
+                more <- state.set(bytesWritten + justWritten) *> Push.more
```
iravid, line 61: You could save the first `state.get` by doing `state.update(_ + justWritten)`.

streams/jvm/src/main/scala/zio/stream/platform.scala
lvitaly: Ups 🤭 My bad! Removing

streams/jvm/src/main/scala/zio/stream/platform.scala
lvitaly: Got it! 👌

streams/jvm/src/main/scala-2.13/zio/stream/ZSinkFromFileConstructor.scala
```diff
@@ -0,0 +1,43 @@
...
+  final def fromFile(
+    path: => Path,
+    position: Long = 0L,
+    options: Set[OpenOption] = Set(WRITE, TRUNCATE_EXISTING, CREATE)
+  ): ZSink[Blocking, Throwable, Byte, Long] =
+    ZSink {
+      for {
+        state <- Ref.make(0L).toManaged_
+        channel <- ZManaged.make(
+                    blocking.effectBlockingInterrupt(FileChannel.open(path, options.asJava).position(position)).orDie
```
iravid, line 28: Would it be possible to just create the set manually with a loop instead of duplicating the entire file?

lvitaly: Yep. It would be better than playing around sbt and crossversion.

iravid:
Great work @lvitaly. Thank you for contributing this!

PR #3725, opened 2020-05-29
Align Methods on Blocking Service with Accessors

### LINKED ISSUE #3724: Some helper functions in zio.blocking._ do not exist on service trait
In particular, `effectBlockingIO` is not defined in the service, but exists as a helper. The trait should have all methods because it will be used directly in layers.

adamgfraser:
Resolves #3724. Adds `blockingExecutor` accessor method and `effectBlockingIO` method to service. With this the same methods are available both ways.


jdegoes:
💪🙏

PR #3727, opened 2020-05-29
Added chunk combinators for streams 

ajaychandran:
This is useful for communication endpoints (network, device) that have a limit on the number of bytes that can be transmitted.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -79,14 +79,6 @@ object ZTransducerSpec extends ZIOBaseSpec {
       )
     ),
     suite("Constructors")(
-      suite("chunkChunks")(
-        testM("chunks individual chunks") {
-          val push = ZTransducer.chunkChunks[Byte].push
-          checkM(Gen.chunkOf(Gen.anyByte))(bytes =>
-            push.use(_.apply(Some(bytes)).map(chunk => assert(chunk.flatten)(equalTo(bytes))))
-          )
-        }
-      ),
       suite("chunkLimit")(
         testM("limits chunks by size") {
           val max  = 64
@@ -129,6 +121,14 @@ object ZTransducerSpec extends ZIOBaseSpec {
                 .runCollect
             )(equalTo(bytes.flatten.padTo(count + (if (pad == 0) 0 else size - pad), 0)))
           }
+        }
+      ),
+      suite("chunks")(
+        testM("chunks individual chunks") {
+          val push = ZTransducer.chunks[Byte].push
+          checkM(Gen.chunkOf(Gen.anyByte))(bytes =>
+            push.use(_.apply(Some(bytes)).map(chunk => assert(chunk.flatten)(equalTo(bytes))))
+          )
         }
       ),
       suite("collectAllN")(
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -79,56 +79,63 @@ object ZTransducerSpec extends ZIOBaseSpec {
       )
     ),
     suite("Constructors")(
-      suite("chunkLimit")(
-        testM("limits chunks by size") {
-          val max  = 64
-          val push = ZTransducer.chunkLimit[Byte](max).push
-          checkM(Gen.chunkOf(Gen.chunkOf(Gen.anyByte)))(bytes =>
-            push.use(
-              _.apply(Some(bytes)).map(chunk =>
-                assert(chunk.forall(_.length <= max))(equalTo(true)) && assert(chunk.flatten)(equalTo(bytes.flatten))
-              )
+      suite("chunk*")(
+        suite("chunkLimit")(
+          testM("limits chunks by size") {
+            checkM(tinyChunkOf(Gen.chunkOf(Gen.anyByte)), Gen.int(1, 100))((bytes, max) =>
+              ZTransducer
+                .chunkLimit[Byte](max)
+                .push
+                .use(
+                  _.apply(Some(bytes))
+                    .map(chunk =>
+                      assert(chunk.forall(_.length <= max))(equalTo(true)) &&
+                        assert(chunk.flatten)(equalTo(bytes.flatten))
+                    )
+                )
             )
-          )
-        }
-      ),
-      suite("chunkN")(
-        testM("emits chunks of a fixed size") {
-          val size = 4
-          val push = ZTransducer.chunkN[Any, Nothing, Int](size).push
-          checkM(Gen.chunkOf(Gen.chunkOf(Gen.anyInt)))(bytes =>
-            push.use(
-              _.apply(Some(bytes)).map { chunk =>
+          }
+        ),
+        suite("chunkN")(
+          testM("emits chunks of a fixed size") {
+            checkM(tinyChunkOf(Gen.chunkOf(Gen.anyByte)), Gen.int(1, 100)) {
+              case (bytes, size) =>
                 val bs = bytes.flatten
-                assert(chunk.forall(_.length == size))(equalTo(true)) &&
-                assert(chunk.flatten)(equalTo(bs.take(bs.length / size * size)))
-              }
-            )
-          )
-        },
-        testM("pads last chunk") {
-          val size = 4
-          val pad  = 0
-          checkM(Gen.chunkOf(Gen.chunkOf(Gen.anyInt))) { bytes =>
-            val bs = bytes.flatten
-            val ds = bs.length % size
-            assertM(
+                ZTransducer
+                  .chunkN[Byte](size)
+                  .push
+                  .use(
+                    _.apply(Some(bytes))
+                      .map(chunk =>
+                        assert(chunk.forall(_.length == size))(equalTo(true)) &&
+                          assert(chunk.flatten)(equalTo(bs.take(bs.length / size * size)))
+                      )
+                  )
+            }
+          },
+          testM("pads last chunk") {
+            val pad = 0.toByte
+            checkM(tinyChunkOf(Gen.chunkOf(Gen.anyByte)), Gen.int(1, 100))((bytes, size) =>
               ZStream
-                .fromChunks(bytes)
-                .aggregate(ZTransducer.chunkN(size, (chunk: Chunk[Int]) => ZIO.succeedNow(chunk.padTo(size, pad))))
+                .fromChunks(bytes: _*)
+                .chunkN(size, pad)
                 .flattenChunks
                 .runCollect
-            )(equalTo(bs.padTo(bs.length + (if (ds == 0) 0 else size - ds), 0)))
+                .map(chunk => assert(chunk)(equalTo(bytes.flatten.padTo(chunk.length, pad))))
+            )
           }
-        }
-      ),
-      suite("chunks")(
-        testM("chunks individual chunks") {
-          val push = ZTransducer.chunks[Byte].push
-          checkM(Gen.chunkOf(Gen.anyByte))(bytes =>
-            push.use(_.apply(Some(bytes)).map(chunk => assert(chunk.flatten)(equalTo(bytes))))
-          )
-        }
+        ),
+        suite("chunks")(
+          testM("chunks individual chunks") {
+            val push = ZTransducer.chunks[Byte].push
+            checkM(Gen.chunkOf(Gen.anyByte))(bytes =>
+              push.use(
+                _.apply(Some(bytes))
+                  .map(chunk => assert(chunk.headOption)(equalTo(Some(bytes).filter(_.nonEmpty))))
+              )
+            )
+          }
+        )
       ),
       suite("collectAllN")(
         testM("happy path") {
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -95,31 +95,30 @@ object ZTransducerSpec extends ZIOBaseSpec {
       suite("chunkN")(
         testM("emits chunks of a fixed size") {
           val size = 4
+          val push = ZTransducer.chunkN[Any, Nothing, Int](size).push
           checkM(Gen.chunkOf(Gen.chunkOf(Gen.anyInt)))(bytes =>
-            ZTransducer
-              .chunkN[Int](size)
-              .push
-              .use(
-                _.apply(Some(bytes)).map { chunk =>
-                  val bs = bytes.flatten
-                  assert(chunk.forall(_.length == size))(equalTo(true)) &&
-                  assert(chunk.flatten)(equalTo(bs.take(bs.length / size * size)))
-                }
-              )
+            push.use(
+              _.apply(Some(bytes)).map { chunk =>
+                val bs = bytes.flatten
+                assert(chunk.forall(_.length == size))(equalTo(true)) &&
+                assert(chunk.flatten)(equalTo(bs.take(bs.length / size * size)))
+              }
+            )
           )
         },
         testM("pads last chunk") {
           val size = 4
+          val pad  = 0
           checkM(Gen.chunkOf(Gen.chunkOf(Gen.anyInt))) { bytes =>
-            val count = bytes.foldLeft(0)(_ + _.length)
-            val pad   = count % size
+            val bs = bytes.flatten
+            val ds = bs.length % size
             assertM(
               ZStream
                 .fromChunks(bytes)
-                .aggregate(ZTransducer.chunkN[Int](size, _.padTo(size, 0)))
+                .aggregate(ZTransducer.chunkN(size, (chunk: Chunk[Int]) => ZIO.succeedNow(chunk.padTo(size, pad))))
                 .flattenChunks
                 .runCollect
-            )(equalTo(bytes.flatten.padTo(count + (if (pad == 0) 0 else size - pad), 0)))
+            )(equalTo(bs.padTo(bs.length + (if (ds == 0) 0 else size - ds), 0)))
           }
         }
       ),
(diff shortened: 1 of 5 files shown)


ajaychandran:
Thanks for the info @simpadjo .

Closing the PR.

simpadjo:
@ajaychandran actually the idea is quite nice, it could've saved the original transducer.

ajaychandran:
Thanks @simpadjo.
I have reopened the PR.

Actually, `groupChunks` is different from `chunkN`. The former will pass any chunk that is under the limit whereas the latter would do this for the last chunk (going by the example in #3522).

Communication endpoints have a limit on the number of bytes that can be sent. Most messages will be under the limit and should be written immediately (and not wait for the limit to be reached).

ajaychandran:
Perhaps, the name should be changed to `limitChunks` or `throttleChunks`?

ajaychandran:
@simpadjo 
The 2.13 and Dotty builds are failing with errors related to changes in the Scala Collections API.
I think this would require adding a method like `fromIterator` somewhere in the platform specific directories. Would you agree?

```sbt
[info] Compiling 10 Scala sources to /home/circleci/project/streams/jvm/target/scala-2.13/classes ...
 
[error] /home/circleci/project/streams/shared/src/main/scala/zio/stream/ZTransducer.scala:600:98: trait Chunk takes type parameters
 
[error]     identity[A].mapChunks(chunk => if (chunk.length <= limit) chunk else chunk.grouped(limit).to[Chunk].flatten)
 ```

**Comment by ajaychandran** [comment]
Or we could go with this,

```scala
def groupChunks[A](limit: Int): ZTransducer[Any, Nothing, A, A] =
  identity[A].mapChunks(chunk =>
    if (chunk.length <= limit) chunk else Chunk.fromArray(chunk.grouped(limit).toArray).flatten
  )
```

ajaychandran:
The original implementation was incorrect.
The revised implementation breaks for the following input:
```scala
Chunk(0), Chunk(1, 2, 3), Chunk.empty, Chunk(4, 5, 6, 7, 8, 9)
```
The last chunk is emitted unchanged.

To fix this, `throttleChunks` will have to be changed to:
```scala
def throttleChunks[A](limit: Int): ZTransducer[Any, Nothing, A, Chunk[A]] = {

  @scala.annotation.tailrec
  def go(z: Chunk[Chunk[A]], chunk: Chunk[A]): Chunk[Chunk[A]] =
    if (chunk.length <= limit) z + chunk else go(z + chunk.take(limit), chunk.drop(limit))

  ZTransducer.identity[A].mapChunks(go(Chunk.empty, _))
}

```

simpadjo:
@iravid could you please say your word about adding this combinator?

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -727,6 +727,23 @@ object ZTransducer {
       }
     }
 
+  /**
+   * Creates a transducer that limits byte chunks to the given size.
+   */
+  def throttleBytes(limit: Int): ZTransducer[Any, Nothing, Byte, Chunk[Byte]] =
+    throttleChunks[Byte](limit)
```
iravid, line 734: We can drop this one as it's a monomorphic version of the other variant.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -727,6 +727,23 @@ object ZTransducer {
...
+  /**
+   * Creates a transducer that limits chunks to the given size.
+   */
+  def throttleChunks[A](limit: Int): ZTransducer[Any, Nothing, A, Chunk[A]] = {
+    @scala.annotation.tailrec
+    def go(z: Chunk[Chunk[A]], chunk: Chunk[A]): Chunk[Chunk[A]] =
+      if (chunk.isEmpty) z else go(z + chunk.take(limit), chunk.drop(limit))
+
+    ZTransducer.identity[A].mapChunks(go(Chunk.empty, _))
+  }
```
iravid, line 745: Let's move this to a combinator on `ZStream` so it doesn't have the chunks exposed. Combinators that alter the chunking structure are not a great fit for transducers.

Also, we need a better name: `chunkUpTo`, `chunkMaxN`, something like that.

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
```diff
@@ -443,6 +443,16 @@ object ZTransducerSpec extends ZIOBaseSpec {
...
+      suite("throttleChunks")(
+        testM("limits chunks by size") {
+          assertM(
+            Stream
+              .fromChunks(Chunk.empty, Chunk(0), Chunk(1, 2, 3), Chunk(4, 5, 6, 7, 8, 9))
+              .aggregate(ZTransducer.throttleChunks(2))
+              .runCollect
+          )(equalTo(Chunk(Chunk(0), Chunk(1, 2), Chunk(3), Chunk(4, 5), Chunk(6, 7), Chunk(8, 9))))
+        }
+      ),
```
iravid, line 455: We need a property check here.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Consider an endpoint defined as
```scala
type Endpoint[-R, +E] = ZTransducer[R, E, Byte, Int]

def endpoint[R, E](limit: Int): Endpoint[R, E] =
  throttleChunks[Byte](limit).mapM(write)

def write(chunk: Chunk[Byte]): UIO[Int] = ???
```
Moving the combinator to `ZStream` would require enforcing the limit upstream.
I agree that exposing the chunks is not ideal but that is unavoidable in this case.
Perhaps we could make an exception? 

Is the name `limitChunks` okay?

iravid: `chunkLimit` would be great; the idea is to be completion friendly in the sense that all chunking-related methods can be found by typing `stream.chunk...`.

> Moving the combinator to ZStream would require enforcing the limit upstream. I agree that exposing the chunks is not ideal but that is unavoidable in this case.
Perhaps we could make an exception?

It's probably going to end up looking pretty similar to your current code, so give it a try and let me know if you need a hand.

ajaychandran: This is how the stream combinator would look:
```scala
def chunkLimit(max: Int): ZStream[R, E, Chunk[O]] =
  mapChunks(Chunk.single).mapConcatChunk {chunk =>
    @scala.annotation.tailrec
    def go(z: Chunk[Chunk[O]], chunk: Chunk[O]): Chunk[Chunk[O]] =
      if (chunk.isEmpty) z else go(z + chunk.take(max), chunk.drop(max))
    go(Chunk.empty, chunk)
  }
```
Do you want the combinator in `ZTransducer` to be removed?

ajaychandran: Or, the combinator in `ZTransducer` could be defined as:
```scala
def chunkLimit[A](max: Int): ZTransducer[Any, Nothing, Chunk[A], Chunk[A]]
```

ajaychandran:
@iravid 
What do you think of these combinators for `ZTransducer`?
```scala
  /**
   * Creates a transducer that chunks individual chunks.
   */
  def chunked[A]: ZTransducer[Any, Nothing, A, Chunk[A]] =
    ZTransducer.identity[A].mapChunks(chunk => if (chunk.isEmpty) Chunk.empty else Chunk.single(chunk))

  /**
   * Creates a transducer that limits chunks to the given size.
   */
  def chunkLimit[A](max: Int): ZTransducer[Any, Nothing, Chunk[A], Chunk[A]] = {
    def go(chunk: Chunk[A]): Chunk[Chunk[A]] =
      if (chunk.isEmpty) Chunk.empty
      else if (chunk.length <= max) Chunk.single(chunk)
      else chunk.take(max) +: go(chunk.drop(max))

    ZTransducer.identity[Chunk[A]].mapChunks(_.flatMap(go))
  }
```

`ZStream.chunkLimit` can then be defined as
```scala
aggregate(ZTransducer.chunked >>> ZTransducer.chunkLimit(n)).mapConcatChunk(identity)
```

ajaychandran:
@iravid Please review.

It is now possible to define the following combinator on `ZStream`:
```scala

def aggregateChunks[R1 <: R, E1 >: E, O1](transducer: ZTransducer[R1, E1, Chunk[O], O1]): ZStream[R1, E1, O1] =
  aggregate(ZTransducer.chunkChunks[O] >>> transducer)
```
The would (marginally) improve ergonomics at call site.
I will submit a separate PR, if this is desired.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Removed

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
ajaychandran: Done.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: The combinator has been revised. Please review.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +192,61 @@ object ZTransducer {
       case None     => ZIO.succeedNow(Chunk.empty)
     }
 
+  /**
+   * Creates a transducer that chunks individual chunks.
+   */
+  def chunkChunks[A]: ZTransducer[Any, Nothing, A, Chunk[A]] =
```
iravid, line 198:
```suggestion
  def chunks[A]: ZTransducer[Any, Nothing, A, Chunk[A]] =
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +192,61 @@ object ZTransducer {
...
+
+    ZTransducer.identity[Chunk[A]].mapChunks(go(_, Chunk.empty))
+  }
+
+  /**
+   * Creates a transducer that emits chunks of a fixed `size`.
+   * The `pad` function is called on the last chunk if it does not have sufficient elements.
+   */
+  def chunkN[A](size: Int, pad: Chunk[A] => Chunk[A]): ZTransducer[Any, Nothing, Chunk[A], Chunk[A]] =
+    apply(ZRefM.make[Chunk[A]](Chunk.empty).toManaged_.map { ref =>
```
iravid, line 226:
```suggestion
    apply(ZRef.make[Chunk[A]](Chunk.empty).toManaged_.map { ref =>
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +192,61 @@ object ZTransducer {
...
+      @scala.annotation.tailrec
+      def go(as: Chunk[A], in: Chunk[Chunk[A]], out: Chunk[Chunk[A]]): (Chunk[Chunk[A]], Chunk[A]) =
+        if (in.isEmpty)
+          if (as.length < size) (out, as) else go(as.drop(size), in, out + as.take(size))
+        else if (as.length < size) go(as ++ in.head, in.tail, out)
```
iravid, line 231: Incrementally building up chunks with `++` can create very inefficient chunks, because we currently don't balance the `Concat` tree. Could you convert this to a loop that builds up the chunk using a `ChunkBuilder`?

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +192,61 @@ object ZTransducer {
...
+      }
+
+    ZTransducer.identity[Chunk[A]].mapChunks(go(_, Chunk.empty))
+  }
+
+  /**
+   * Creates a transducer that emits chunks of a fixed `size`.
+   * The `pad` function is called on the last chunk if it does not have sufficient elements.
+   */
+  def chunkN[A](size: Int, pad: Chunk[A] => Chunk[A]): ZTransducer[Any, Nothing, Chunk[A], Chunk[A]] =
```
iravid, line 225: Nice touch, I like the pad! Let's make this default to identity:
```suggestion
  def chunkN[A](size: Int, pad: Chunk[A] => Chunk[A] = c => c): ZTransducer[Any, Nothing, Chunk[A], Chunk[A]] =
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -192,6 +192,61 @@ object ZTransducer {
...
+    @scala.annotation.tailrec
+    def make(in: Chunk[A], out: Chunk[Chunk[A]]): Chunk[Chunk[A]] =
+      if (in.length <= max) out + in else make(in.drop(max), out + in.take(max))
+
+    @scala.annotation.tailrec
+    def go(in: Chunk[Chunk[A]], out: Chunk[Chunk[A]]): Chunk[Chunk[A]] =
+      (in.length: @scala.annotation.switch) match {
+        case 0                          => out
+        case 1 if in.head.length <= max => out ++ in
+        case _                          => go(in.tail, make(in.head, out))
```
iravid, line 215: Same here - recursively decomposing chunks is not the best access pattern. Straight up iteration is much more efficient.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: @iravid The name `chunks` could be misunderstood as wrapping individual elements in a chunk.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Will do

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Will do.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
iravid: It could, but with proper scaladoc, it won't :-)

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Done

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
ajaychandran: Done

ajaychandran:
@iravid This is ready for review.

> If you could add the ZStream versions that'd be great too.

I will submit a separate PR for stream combinators.

ajaychandran:
@iravid Do not merge this yet, I think the pad function needs to return an effect.
Also, I will add the stream combinators.

ajaychandran:
@iravid Please review.

The dotty job failed with `Too long with no output (exceeded 10m0s): context deadline exceeded`.
There was a compilation warning but that should not have caused this.

Also, the tests for the new comibinators are an order of magnitude slower on JS. Hope this isn't an issue.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -559,47 +565,34 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-          }, ch => emitOrAccumulate(buffer ++ ch, false, ref, pull))
-      } else {
-        val (chunk, leftover) = buffer.splitAt(n)
-        ref.set(State(leftover, done)) *> Pull.emit(chunk)
-      }
+  /**
+   * Re-chunks the chunks of the stream into non-empty chunks of a fixed `size`.
+   * The last chunk may have less than `size` elements.
+   */
+  def chunkN(size: Int): ZStream[R, E, Chunk[O]] =
```
iravid, line 577: Sorry, missed this: the signature here changed. Should stay `ZStream[R, E, O]`. Same for `chunkLimit` which is newly added but should also use `ZStream[R, E, O]`.

ajaychandran: @iravid Wouldn't the rationale for chunking be to subsequently operate on chunks?
The current signature allows for
```scala
byteStream.chunkN(4).map(_.toByteBuffer.readInt)
```
If we change the signature, we would have to write this as
```scala
byteStream.chunkN(4).mapChunks(_.toByteBuffer.readInt)
```

iravid: Ah I see where the confusion stems from. 

No, the purpose of the chunking combinators on ZStream is to modify the underlying chunking structure. For most cases, it is unnecessary to expose those chunks because all operators work on chunks automatically.

The chunks can be operated on as you suggested with `mapChunks`, or by exposing them using the `ZStream` version of your `ZTransducer.chunks` transducer that we'll add soon.

iravid: Ah yes. `flattenChunks` is too eager. Let's fix that in a follow-up using `BufferedPull` instead of `mapConcatChunk`.

ajaychandran: @iravid 
Just sharing my thoughts.

> No, the purpose of the chunking combinators on ZStream is to modify the underlying chunking structure. For most cases, it is unnecessary to expose those chunks because all operators work on chunks automatically.

I agree with the first part. And the new signatures make this explicit!

Are there any combinators, other than `chunkN`, that alter the chunking structure?

Is there a usecase that alters the chunking structure and then subsequently operates on individual elements? I don't consider the above testcase a valid use (it just requires input arriving in chunks).

>The chunks can be operated on as you suggested with mapChunks, or by exposing them using the ZStream version of your ZTransducer.chunks transducer that we'll add soon.

Yes, but this penalizes users who would want to operate on chunks with a redundant `un-chunk + re-chunk`.

ajaychandran: Alternatively, we can remove the `chunk*` combinators on `ZStream` and let users construct the argument for `aggregateChunks`.

For my usecase, I only need `chunkLimit` on `ZTransducer`. It really doesn't make sense on `ZStream`, IMO.

ajaychandran:
Please hold merging of this PR.

I am working on a separate PR with an alternate encoding for streams that would invalidate this request.

ajaychandran:
So, its unlikely that the new PR will be finsihed or even included in 1.x.

We can proceed with merging.
@iravid 
> Ah yes. flattenChunks is too eager. Let's fix that in a follow-up using BufferedPull instead of mapConcatChunk.

I assume that there is nothing pending for this PR.

jdegoes:
@ajaychandran @iravid This has grown stale. Feel free to re-open if you can update / address feedback!

PR #3730, opened 2020-05-30
Fix ZStream#intersperse

### LINKED ISSUE #3729: ZStream#intersperse breaks with usage of Stream.repeatEffect
`ZStream#intersperse` breaks with usage of `Stream.repeatEffect`.

Test case:
```scala
testM("intersperse several from repeat effect") {
  Stream
    .repeatEffect(ZIO.succeed(42))
    .map(_.toString)
    .intersperse("@")
    .take(4)
    .runCollect
    .map(result => assert(result)(equalTo(Chunk("42", "@", "42", "@"))))
}
```
Failure: `Chunk(42,42,42,42) did not satisfy equalTo(Chunk(42,@,42,@))`

Note: Moving `ZStream#take` beforehand `ZStream#intersperse` yields the same failure.

_Edit:_ Correct failure message

iravid:
Resolves #3729

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1692,11 +1692,21 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
         chunks <- self.process
         pull = chunks.flatMap { os =>
           state.modify { flag =>
-            os.foldRight(List.empty[O1] -> flag) {
-              case (o, (Nil, curr)) => List(o)              -> !curr
-              case (o, (out, curr)) => (o :: middle :: out) -> !curr
+            val builder = ChunkBuilder.make[O1]()
+
+            val flagResult = os.foldLeft(flag) { (flag, o) =>
+              if (!flag) {
+                builder += o
+                !flag
+              } else {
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1692,11 +1692,21 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+            val builder = ChunkBuilder.make[O1]()
+
+            val flagResult = os.foldLeft(flag) { (flag, o) =>
+              if (!flag) {
+                builder += o
+                !flag
+              } else {
+                builder += middle
+                builder += o
+                flag
```
regiskuckaertz, line 1704: Hmm 🧐 true once, true forever

luis3m: Noticed that, but it makes sense. You only need to avoid adding middle the first time

iravid: Ah let me simplify that.

luis3m: Maybe adding a flag var and changing the foldLeft for a foreach

PR #3731, opened 2020-05-30
Prevent ZStream.repeatEffectChunkOption from evaluating after end

iravid:
@regiskuckaertz @luis3m if any of you are free for another review 🙏🏻


luis3m:
Looks good to me. You already changed `tapError` which was my only observation when I saw those changes in the other PR

iravid:
Thanks @luis3m 🙏🏻 I missed what you wrote about that. Was it the fact that it alters the `Cause`?

luis3m:
Actually never wrote it because you said you were going to remove those changes on Discord. It seems you overwritten the changes on the other PR so I can't confirm for sure but I think it was using `tapError` as `mapError`.

PR #3738, opened 2020-06-01
Add provideCustom

ahoy-jon:
to improve parity between provide and provideLayer, this PR add :

```scala
  /**
   * Provides the part of the environment that is not part of the `ZEnv`,
   * leaving an effect that only depends on the `ZEnv`.
   *
   * {{{
   * val logging: Logging = ???
   *
   * val zio: ZIO[ZEnv with Logging, Nothing, Unit] = ???
   *
   * val zio2 = zio.provideCustom(logging)
   * }}}
   */
  final def provideCustom[R1 <: Has[_]](
    r1: R1
  )(implicit ev: zio.ZEnv with R1 <:< R, tagged: Tag[R1]): ZIO[zio.ZEnv, E, A] =
    provideSome[zio.ZEnv] { zenv =>
      val all = zenv ++ r1
      ev(all)
    }
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1442,6 +1442,14 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(zio2)(anything)
       }
     ),
+    suite("provideCustom")(
+      testM("provides the part ot the environment that is not part of the `ZEnv`") {
+        val logging: Logging                           = Has(new Logging.Service {})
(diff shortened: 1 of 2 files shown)


adamgfraser:
I have some concerns about this. After we added layers we deleted a bunch of provide variants and tried to simplify things to just `provide` and `provideSome` for people who didn't want to use `ZLayer` / `Has` and `provideLayer` variants for everything else. I'm not sure it is worth adding additional variants for this, though open to other opinions.

ahoy-jon:
I agree with you, and it is not working with dotty at the moment!

I am closing it. If someone thinks it is worth it, that can be reopened.

PR #3741, opened 2020-06-01
Replace ZIO[R, E, A] by aliases if possible + update documentation

ahoy-jon:
Specializing types on `zio.App`.


adamgfraser:
I wonder if there is pedagogical value in not using the type alias here? I usually use `App` early on in a "Hello, World!" example in ZIO and find it helpful to talk through how this is an effect that can use the `ZEnv`, can't fail, and returns an exit code. This is obviously the same but the type alias is one more layer of indirection when you're already introducing a lot of new concepts. Maybe it doesn't matter because you could always not use the type alias in educational materials but just a thought.

ahoy-jon:
Good point!

I would take another path to improve the pedagogical aspects. 
I would use comments (and a different ordering) in `package.scala`
```scala
  type Task[+A]     = ZIO[Any, Throwable, A] // Returns `A`, can fails with `Throwable`, depends on nothing (`Any`).
  type RIO[-R, +A]  = ZIO[R, Throwable, A]   // Like `Task`, but depends on `R`.
  type UIO[+A]      = ZIO[Any, Nothing, A]   // Always returns A, never fails (`Nothing`), depends on nothing (`Any`).
  type URIO[-R, +A] = ZIO[R, Nothing, A]     // Like `UIO`, always returns `A`, never fails (`Nothing`), depends on `R`.
  type IO[+E, +A]   = ZIO[Any, E, A]         // Returns `A`, can fails with `E` , depends on nothing (`Any`)
```

A lot of learning use navigation in the source (go to declaration, then back navigation).

As for `URIO[ZEnv, ExitCode]` indeed it is a bit harder at the beginning, but it may help get in ZIO faster and improve thought structure in the long term. (ie, an application depends on `ZEnv`, always returns an `ExitCode`). As you see fit!

luis3m:
I'd suggest to change the docs as well in case of moving forward with these changes.

ahoy-jon:
Sure! Probably a bigger change. I let you guys discuss it if you know which direction it should go. 

I was just doing some ZIO blogging and that thing got in the way! 😃

Outside of this PR (I don't have any hard opinion on it): 
* I will try to find better places in ZIO were the type could use an alias. 
* try to work on a scalafix solution to improve code quality by replacing with aliases on an existing codebase.

adamgfraser:
Okay, let's get this in. I agree we should update the documentation as well and it would be good to add Scaladoc for the type aliases. Do you want to do here or shall we do in a follow up PR?

adamgfraser:
Can ZIO Initellij be used for this? Copying @hmemcpy.

hmemcpy:
So what is being proposed here? Specialize all `ZIO[R, E, A]` to aliases? The plugin currently doesn't have this specific functionality as a bulk operation, but I believe this is something I could add!

adamgfraser:
I think that would be idea. Might be good to have some setting for which ones to rewrite. Like I think the ZIO type aliases are well known so I see the value there but I would wonder if we would want to rewrite all the `Managed` or `STM` aliases given the additional indirection and lower user awareness of those type aliases.

ahoy-jon:
@hmemcpy would love to use the plugin for that or ScalaFix. (thank you for the plugin, using it)

For information, currently, I am using regexps in IntelliJ to find and replace all occurrences and replace them when needed in the code and in the documentation.

adamgfraser:
Can you fix the test failures?

ahoy-jon:
@adamgfraser sorry, sure will do, I thought It was just some timeout at the beginning. It's back to normal.

btw, @bernit77 is not able to sign the `cla`, do you know why?

adamgfraser:
@ahoy-jon Thanks! We're experiencing an issue with the CLA Assistant functionality. I have opened a ticket with them.

PR #3747, opened 2020-06-02
Implement Fast Single Element Prepend for Chunk

adamgfraser:
```scala
[info] Benchmark                             (size)   Mode  Cnt     Score    Error  Units
[info] ChunkPrependBenchmarks.chunkPrepend    10000  thrpt   25  7788.576 ± 35.149  ops/s
[info] ChunkPrependBenchmarks.vectorPrepend   10000  thrpt   25  4383.454 ± 14.474  ops/s
```


iravid:
Love it! Why the change from `+` to `+:`?

adamgfraser:
I thought if we are having `+:` for prepend then for symmetry we should also have `:+` for append. Also, now that `Chunk` extends `IndexedSeq`, `:+` and `+:` are already methods brought in from the standard library so seems a little weird to have a different symbolic alias to do the same thing and it is now consistent with other Scala collections.

adamgfraser:
Yes. Done now.

PR #3739, opened 2020-06-01
add resurrect and resurrectWith

ahoy-jon:
Symmetrical to orDie and orDieWith:
```scala
ZIO.fail(new Exception("msg")).orDie.resurrect
```

PR #3750, opened 2020-06-03
Remove `resurrectWith`

ahoy-jon:
Sorry, I should have thought of it before. 
Following #3739, change to a partial function on `resurrectWith`.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1062,8 +1062,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    *   val f3: IO[String, Unit] = f2.resurrectWith(_.getMessage)
    * }}}
    */
-  final def resurrectWith[E1 >: E](f: Throwable => E1): ZIO[R, E1, A] =
-    self.sandbox.mapError(_.failureOrCause.map(c => f(c.squash)).merge)
+  final def resurrectWith[E1 >: E](pf: PartialFunction[Throwable, E1]): ZIO[R, E1, A] =
+    self.sandbox
+      .mapError({
+        case cause if cause.failed => cause
+        case cause                 => cause.defects.headOption.collect(pf).map(Cause.fail).getOrElse(cause)
+      })
+      .unsandbox
 
   /**
    * Executes this effect and returns its value, if it succeeds, but
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1039,7 +1039,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * the specified function to convert the `E` into a `Throwable`.
    */
   final def orDieWith(f: E => Throwable)(implicit ev: CanFail[E]): URIO[R, A] =
-    (self mapError f) catchAll (IO.die(_))
+    self.foldM(e => ZIO.die(f(e)), ZIO.succeedNow)
 
   /**
    * Unearth the unchecked failure of the effect. (opposite of `orDie`)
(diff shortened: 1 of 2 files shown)


adamgfraser:
Isn't this just `unrefineWith`?

ahoy-jon:
@adamgfraser indeed, good catch. It is `unrefined`, just pushed a commit to reuse it.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1039,7 +1039,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * the specified function to convert the `E` into a `Throwable`.
    */
   final def orDieWith(f: E => Throwable)(implicit ev: CanFail[E]): URIO[R, A] =
-    (self mapError f) catchAll (IO.die(_))
+    self.foldM(e => ZIO.die(f(e)), ZIO.succeedNow)
```
ahoy-jon, line 1042: @adamgfraser not sure about this one, it looked better with fewer calls.

adamgfraser: @ahoy-jon Looks good.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1050,7 +1050,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * }}}
    */
   final def resurrect(implicit ev1: E <:< Throwable): RIO[R, A] =
-    self.sandbox.mapError(_.squash)
+    self.foldCauseM(cause => ZIO.fail(cause.squash), ZIO.succeedNow)
```
adamgfraser, line 1053: `unrefineWith { case e => e }(ev1)`

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1062,8 +1062,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    *   val f3: IO[String, Unit] = f2.resurrectWith(_.getMessage)
    * }}}
    */
-  final def resurrectWith[E1 >: E](f: Throwable => E1): ZIO[R, E1, A] =
-    self.sandbox.mapError(_.failureOrCause.map(c => f(c.squash)).merge)
+  final def resurrectWith[E1 >: E](pf: PartialFunction[Throwable, E1]): ZIO[R, E1, A] =
```
adamgfraser, line 1065: I think if we already have an existing combinator that does the same thing we should remove this.

ahoy-jon: Not sure about this, there is `refineWith` and `orDieWith`. The idea was to improve parity/symmetry.

core/shared/src/main/scala/zio/ZIO.scala
ahoy-jon: Not sure, `unrefineWith` due to the partial is doing a `find` and a recursive `fold`, instead of a direct call to `foldCauseM`.

adamgfraser: That's fine.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: In a data type like ZIO that has a lot of structure there are going to be multiple symmetries but I don't think we can have different methods that do the same thing. I think the distinction is that `orDie` / `orDieWith` take all errors and convert them to fiber failures whereas `refine` variants take some errors and convert them to fiber failures. Similarly `resurrect` takes all unchecked errors and converts them to typed errors whereas `unrefineWith` only does that with some errors. So I think that is the appropriate naming.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1050,7 +1050,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * }}}
    */
   final def resurrect(implicit ev1: E <:< Throwable): RIO[R, A] =
-    self.sandbox.mapError(_.squash)
+    self.unrefineWith({ case e => e })(ev1)
```
adamgfraser, line 1053: Fine to do it the other way here if you think that is more efficient.

ahoy-jon: Using `foldCauseM` is like taping directly into `new ZIO.Fold` so might be hard to beat that!
However, reusing existing combinators improves the discovery of those!

PR #3754, opened 2020-06-04
Implement Balanced Concatenation for Chunk

### LINKED ISSUE #3238: Optimize Chunk#++

### LINKED ISSUE #3753: Chunk.toArray StackOverflow, ZIO-RC19
```
Java.lang.StackOverflowError at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at zio.Chunk$Concat.toArray(Chunk.scala:1128) at 
zio.Chunk$Concat.toArray(Chunk.scala:1128) at
```
and so on

adamgfraser:
Resolves #3238. Resolves #3753.

Implements balanced concatenation for Chunk based on the algorithm used in Conc-Trees. This also addresses the stack overflow issue.

### DIFF
--- a/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
@@ -1,0 +1,54 @@
+package zio.chunks
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+
+import zio.Chunk
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class ChunkConcatBenchmarks {
+
+  val chunk: Chunk[Int] = Chunk.empty
+
+  @Param(Array("10000"))
+  var size: Int = _
+
+  def concatBalanced(n: Int): Chunk[Int] =
+    if (n == 0) Chunk.empty
+    else if (n == 1) Chunk(1)
+    else concatBalanced(n / 2) ++ concatBalanced(n / 2)
+
+  @Benchmark
+  def leftConcat(): Chunk[Int] = {
+    var i       = 1
+    var current = chunk
+
+    while (i < size) {
+      current = Chunk(i) ++ current
+      i += 1
+    }
+
+    current
+  }
+
+  @Benchmark
+  def rightConcat(): Chunk[Int] = {
+    var i       = 1
+    var current = chunk
+
+    while (i < size) {
+      current = current ++ Chunk(i)
+      i += 1
+    }
+
+    current
+  }
+
+  @Benchmark
+  def balancedConcat(): Chunk[Int] = {
+    concatBalanced(size)
+  }
+}
--- a/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkConcatBenchmarks.scala
@@ -48,7 +48,6 @@ class ChunkConcatBenchmarks {
   }
 
   @Benchmark
-  def balancedConcat(): Chunk[Int] = {
+  def balancedConcat(): Chunk[Int] =
     concatBalanced(size)
-  }
 }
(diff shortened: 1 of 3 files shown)


ahoy-jon:
@adamgfraser Maybe I am completely wrong, but you may not need depth to do that, the size difference between left and right is enough to choose where to land the new piece in the tree.

If you adapt this code to concat it works.
https://github.com/zio/zio/pull/3311#issuecomment-613015881

ahoy-jon:
I went back to the code I wrote at the time, with the perf test, this was doing a decent job and perfect balancing on the repeated concatenation.
```
final def concat[A](l: NonEmpty[A], r: NonEmpty[A]): NonEmptyChunk[A] =
    l match {
      case Concat(ll, lr) if ll.size >= lr.size + r.size => Concat(ll, concat(lr, r))
      case _                                             => Concat(l, r)
    }
```
This one is the same version without the recursion
```
def concat[A](l: NonEmpty[A], r: NonEmpty[A]): NonEmptyChunk[A] =
    if (l.length == r.length) Concat(l, r)
    else if (l.length > r.length) {

      val lefts: Array[NonEmptyChunk[A]] = Array.ofDim[NonEmptyChunk[A]](16)
      var n: Int                         = 0

      var point: NonEmptyChunk[A] = l

      var continue = true

      while (continue && n < 16) {
        point match {
          case _ if point.length <= r.length =>
            continue = false
          case Concat(ll, lr) if ll.length >= lr.length + r.length =>
            lefts(n) = ll
            n += 1
            point = lr
          case _ =>
            continue = false
        }
      }

      var res = Concat(point, r)
      var i   = n - 1

      while (i >= 0) {
        res = Chunk.Concat(lefts(i), res)
        i -= 1
      }

      res

    } else {
      val rights = Array.ofDim[NonEmptyChunk[A]](16)
      var n: Int = 0

      var point: NonEmptyChunk[A] = r

      var continue = true

      while (continue && n < 16) {
        point match {
          case _ if point.length <= l.length =>
            continue = false
          case Concat(rl, rr) if rr.length >= rl.length + l.length =>
            rights(n) = rr
            n += 1
            point = rl
          case _ =>
            continue = false
        }
      }

      var res = Concat(l, point)
      var i   = n - 1

      while (i >= 0) {
        res = Chunk.Concat(res, rights(i))
        i -= 1
      }

      res
    }
```

The issue was, due to the limited structure of the `Chunk` (with `Concat` having only 2 possibles children), the performance did not improve much (but both fixed the StackOverflow issues).

With your work on repeated appends and pretends (thanks 🙏 ), the performance might not be visible on regular use-cases.

iravid:
Looks great @adamgfraser. Can you add a repeated concatenation benchmark to see the perf. hit?

adamgfraser:
Yes. It depends a lot on usage pattern. For balanced concatenation it doesn't make much of a difference. For degenerate cases where you repeatedly concatenate a single element it is quite a hit. Maybe when we concatenate we could check the length of the chunks and treat it as an append or prepend if one of them has a length of one to avoid the very worst case?

```
[info] Benchmark                             (size)   Mode  Cnt     Score    Error  Units
[info] ChunkConcatBenchmarks.balancedConcat   10000  thrpt   25  1576.387 ±  6.589  ops/s
[info] ChunkConcatBenchmarks.leftConcat       10000  thrpt   25  1199.527 ± 16.542  ops/s
[info] ChunkConcatBenchmarks.rightConcat      10000  thrpt   25  1210.338 ± 12.808  ops/s
```

```
[info] Benchmark                             (size)   Mode  Cnt     Score    Error  Units
[info] ChunkConcatBenchmarks.balancedConcat   10000  thrpt   25  1469.000 ± 16.430  ops/s
[info] ChunkConcatBenchmarks.leftConcat       10000  thrpt   25   431.142 ±  2.654  ops/s
[info] ChunkConcatBenchmarks.rightConcat      10000  thrpt   25   397.721 ±  3.231  ops/s
```

PR #3755, opened 2020-06-04
Make fromCompletionStage Referentially Transparent

adamgfraser:
Fixes a bug where `fromCompletionStage` was not referentially transparent because we were capturing a `lazy val` not suspended in an effect. One change in semantics to note. Previously if the `thunk` threw an exception that would be translated into a `Die` failure. I think since we are returning a `Task` we should translate it into a `Fail` failure. Let me know what you think.

Copying @reibitto.


reibitto:
Thanks! This looks good to me. Interestingly enough the workaround we're using in the meantime is also wrapping the thunk in an effect. I personally like that change too. 👍

PR #3756, opened 2020-06-04
Implement TPriorityQueue

adamgfraser:
Simple `TPriorityQueue` implementation extracted from some of my previous work involving the `TestClock`.

Copying @phderome.


core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -0,0 +1,123 @@
...
+
+package zio.stm
+
+import scala.collection.immutable.SortedMap
+
+/**
+ * A simple `TPriorityQueue` implementation. A `TPriorityQueue` contains values
+ * of type `V`. Each value is associated with a key of type `K` that an
+ * `Ordering` is defined on. Unlike a `TQueue`, `take` returns the highest
+ * priority value (the value that is first in the specified ordering) as
```
phderome, line 25: s/first/largest

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -0,0 +1,123 @@
...
+    ZSTM.foreach_(values) { case (key, value) => offer(key, value) }
+
+  /**
+   * Peeks at the first value in the queue without removing it, retrying until
+   * a value is in the queue.
+   */
+  def peek: USTM[V] =
+    tref.get.flatMap { map =>
+      map.headOption match {
+        case None                  => ZSTM.retry
```
phderome, line 55: with ZSTM, we have have the advantage of not blocking here, yes?

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -0,0 +1,123 @@
...
+        case None                                    => ZSTM.none
+        case Some((key, value :: Nil))               => tref.update(_ - key).as(Some(value))
+        case Some((key, value :: values :: values1)) => tref.update(_ + (key -> ::(values, values1))).as(Some(value))
+      }
+    }
+
+  /**
+   * Collects all values into a list.
+   */
+  def toList: USTM[List[V]] =
```
phderome, line 112: just curious as to why providing a list collection is preferable over a vector.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * A simple `TPriorityQueue` implementation. A `TPriorityQueue` contains values
+ * of type `V`. Each value is associated with a key of type `K` that an
+ * `Ordering` is defined on. Unlike a `TQueue`, `take` returns the highest
+ * priority value (the value that is first in the specified ordering) as
+ * opposed to the first value offered to the queue. The ordering that elements
+ * with the same priority will be taken from the queue is not guaranteed.
+ */
+final class TPriorityQueue[K, V] private (private val tref: TRef[SortedMap[K, ::[V]]]) extends AnyVal {
```
mijicd, line 29: We can do the similar one with TMap.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
mijicd: @phderome yup,suspended until changes occur.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+
+  /**
+   * Peeks at the first value in the queue without removing it, returning
+   * `None` if there is not a value in the queue.
+   */
+  def peekOption: USTM[Option[V]] =
+    tref.get.map { map =>
+      map.headOption match {
```
mijicd, line 66: You can go with map here.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
mijicd: Actually, I'd go with chunks here from the beginning.

jdegoes:
@adamgfraser I might back with `TMap[Int, A], TRef[Int]`, or similar, on the theory that if `TMap` is sufficiently optimized, it will permit less contention than a single `Ref` that contains the whole queue. /cc @mijicd

mijicd:
@jdegoes Based on the last benchmarks we have they're comparable, but we might have some "cut-offs" with `TMap`. However, in the end it's up to numbers, so I definitely propose a follow-up to benchmark the implementation.

adamgfraser:
Initial version with `TMap`. This moves to a single type parameter which I think is definitely an improvement. I do worry about the number of STM operations we are using in the implementation. With the `TMap` we are potentially getting more parallelism because different fibers could be modifying different parts of the data structure but we are making non-local changes to maintain the heap structure and having to go through multiple STM operations instead of just being able to modify a single underlying data structure so I am not sure we come out ahead.

jdegoes:
@adamgfraser I agree the overhead of this will be extremely high, and a `TRef` with sorted set is likely to be much, much faster. But it could be done in followup.

PR #3761, opened 2020-06-05
[WIP] Experimental stream encoding

ajaychandran:
This PR is an attempt to test the hypothesis,

_chunking has to be a part of the internal structure of every stream for efficiency._

**Analysis**
The hypothesis is built on the assumption that input values arrive in chunks.
1. When the assumption is false, performance should be adversely impacted due to wrapping with `Chunk.single`.
2. In many usecases, the input is sourced in chunks but transformed into a single value (like an HTTP server endpoint decoding bytes to a request). At every subsequent step in the pipeline, the same performance penalty would have to be paid.

_Even when the hypothesis is true, it does not mean that efficiency cannot be achieved without internal chunking._

**New design**
This PR introduces a new design for streams, in the `zio.streams.experimental` package.

The major change, relevant to the hypothesis, is the removal of the internal chunking structure.
```scala
type Pull[-R, +E, +O] = ZIO[R, Option[E], O]

final class ZStream[-R, +E, +O] private (val pull: URManaged[R, Pull[R, E, O]])
```

The new encoding also decouples the forward propagation of values, during stream processing, from the backward propagation of the final result.
```scala
type Push[-R, +E, -I, +Z] = I => ZIO[R, Either[E, Z], Unit]
type Step[-R, +E, -I, +O] = I => ZIO[R, E, O]

final class ZSink[-R, +E, -I, +Z] private (val process: URManaged[R, (Push[R, E, I, Z], ZIO[R, E, Z])])
final class ZTransducer[-R, +E, -I, +O] private (val process: URManaged[R, (Step[R, E, I, O], ZIO[R, E, Option[O]])])
```

 A `Chunk` is treated as a value and can appear in signatures like
```scala
val stream: ZStream[Any, Nothing, Chunk[Int]] = ZStream(Chunk(1 to 10: _*), Chunk(11 to 20: _*))
```

**Initial results**
The new encoding offers new possibilities when it comes to composition.
```scala

val rawBytes: ZStream[Any, Nothing, Chunk[Byte]] = ???

val program: ZIO[Console with system.System, Nothing, Unit] =
  rawBytes
    .run(ZTransducer.utf8Decode >>> ZTransducer.newLines >>> ZSink.putStrLn.chunked)

// signatures
// utf8Decode: ZTransducer[Any, Nothing, Chunk[Byte], String]
// newLines: ZTransducer[system.System, Nothing, String, Chunk[String]]
// putStrLn: ZSink[console.Console, Nothing, String, Unit]

```
Note the use of the `chunked` combinator on `ZSink.putStrLn`.
_The shape of an existing downstream process is being modified to align with the upstream process (and this is achieved with no intermediate chunking)!_ 

**Future work**
1. Implement more combinators to validate the new design.
2. Benchmark old and new encodings to test the hypothesis.

If someone could point out any flaws in the design or provide an example of a combinator that cannot be implemented without internal chunking, it would be greatly appreciated. Thanks.


ajaychandran:
**Update**
The new encoding has been simplified to
```scala
type Pull[-R, +E, +O] = ZIO[R, Option[E], O]

final class ZStream[-R, +E, +O] private (val process: URManaged[R, Pull[R, E, O]])

final class ZTransducer[-R, +E, -I, +O] private (val process: URManaged[R, (I => Pull[R, E, O], Pull[R, E, O])])

final class ZSink[-R, +E, -I, +Z] private (val process: URManaged[R, (I => Pull[R, E, Any], ZIO[R, E, Z])])
```
This change uses `Pull` in downstream procesess to signal early termination.

This should open more pathways for composition. Watch this space!

iravid:
Hi @ajaychandran. Since you asked for my review, here are my thoughts.

As I mentioned on your gist, embedding chunks was an explicit decision in the latest iteration on ZStream/ZTransducer/ZSink. Without working on chunks, achieving competitive performance on streaming operations (for example, decoding an HTTP request) is hopeless.

Making chunks optional means basic operators like `ZStream#take`, `ZStream#zipWithIndex` and so forth will not be chunk aware. We had `ZStreamChunk` previously, a newtype over streams of chunks that implemented chunk-aware versions of all the operators. However, that meant that we had to implement two versions of every combinator which is obviously a lot of work for little gain.

To me, chunks are a non-negotiable aspect of the design; they also make it much easier to keep primitives unboxed.

On the contents of this PR: this design is pretty similar to the previous design of ZStream. The decoupling of result/push of transducers/sinks was also suggested in one of the design discussions. See here: https://github.com/zio/zio/issues/2551. It was eventually discarded because it leads to some weird aspects (ordering issues between calling push/pull).

Thank you for writing this up!

regiskuckaertz:
I'm having deja vu 🤔

ajaychandran:
@iravid Thank you for the comments and info.

> As I mentioned on your gist, embedding chunks was an explicit decision in the latest iteration on ZStream/ZTransducer/ZSink. Without working on chunks, achieving competitive performance on streaming operations (for example, decoding an HTTP request) is hopeless.

An internal chunked encoding will be performant only if the **input arrives in chunks**. Only the user has knowledge of this.

```scala
// current encoding - internal chunking
val httpBytes: ZStream[Any, Nothing, Byte] = ???
val httpRequests: ZStream[Any. Nothing, Request] = httpBytes.mapChunksM(deserialize)

// new encoding - external chunking
val httpBytes: ZStream[Any, Nothing, Chunk[Byte]] = ???
val httpRequests: ZStream[Any. Nothing, Request] = httpBytes.mapM(derserialize)
```
When a chunk of bytes is de-serialized into a request, it would have to be wrapped in `Chunk.single`. And this would have to be done at every step, in the pipeline, where a single value is produced.

> Making chunks optional means basic operators like ZStream#take, ZStream#zipWithIndex and so forth will not be chunk aware. We had ZStreamChunk previously, a newtype over streams of chunks that implemented chunk-aware versions of all the operators. However, that meant that we had to implement two versions of every combinator which is obviously a lot of work for little gain.

In the new encoding, there is a chunk aware combinator (`chunked`) on `ZSink` and `ZTransducer`. I will attempt an encoding that allows for overriding this combinator.

> The decoupling of result/push of transducers/sinks was also suggested in one of the design discussions. See here: #2551. It was eventually discarded because it leads to some weird aspects (ordering issues between calling push/pull).

I could not find a reference to the ordering issues. Do you recall the combinator? Is it `aggregateAsync`?

Thanks again!

ajaychandran:
**Update**
`ZSink` and `ZTransducer` implementations can now overridde `chunked` and `forall` combinators to provide more efficient versions. See `ZSink.collect*` and `ZTransducer.newLines` for examples.

The `forall` combinator allows downstream processes to operate on `Iterable`s (no conversion to `Chunk` required).
```scala
  val upstream: ZStream[Any, Nothing, List[Int]] = ZStream(List(1, 2), List(3, 4))
  val downstream: ZSink[Any, Nothing, Iterable[Int], Chunk[Int]] = ZSink.collect[Int].forall
  val program: UIO[Chunk[Int]] = upstreamList.run(downstreamList)
```

With this change, it is possible for stream processing to be efficient irrespective of whether the input arrives in chunks or one at a time.

ajaychandran:
**Update**
1. Refined internal encoding of downstream processes.
2. Removed `.foreach` operator (to limit the scope of this PR).
3. Added initial benchmarks.

ajaychandran:
### Benchmarks

**Command**
```sbt
benchmarks/jmh:run -i 3 -wi 3 -f1 -t1 zio.stream.StreamExperiments*
```

**Results**
The initial results look quite promising.
1. When the input arrives one element at a time, the new encoding offers a **5x improvement** in throughput.
```sbt
[info] Benchmark                               (count)   Mode  Cnt     Score     Error  Units
[info] StreamExperiments.newFilterMapSum         10000  thrpt    3   800.628 ±  11.056  ops/s
[info] StreamExperiments.zioFilterMapSum         10000  thrpt    3   120.586 ±  27.334  ops/s
```
2. When the input arrives in chunks, the new encoding offers parity with the current encoding  (considering the margin of error).
```sbt
[info] Benchmark                               (count)   Mode  Cnt     Score     Error  Units
[info] StreamExperiments.newChunkFilterMapSum    10000  thrpt    3  2362.118 ± 596.665  ops/s
[info] StreamExperiments.zioChunkFilterMapSum    10000  thrpt    3  1726.903 ±  30.985  ops/s
```

**Analysis**
The new encoding avoids two kinds of allocations.
1. When input arrives one at a time, each element does not have to be wrapped in `Chunk.single`.
2. The downstream processes in `ZSink` and `ZTransducer` separate the "push" operation (forward propagation of values) from the "read" operation (backward propagation of result). This approach, yet to be fully validated, avoids the wrapping of elements/chunks in an `Option`. 

Additionally, the `chunked` operator is overridden for pure transducers and sinks.

jdegoes:
I think I would rather pay a small overhead on wrapping "big elements" (like requests) than not have chunking baked in, especially since we are producing chunked constructors everywhere now (almost everything is always chunked).

Thank you for your work investigating this and exploring other ideas in the space!

ajaychandran:
@jdegoes This PR is an ongoing investigation. Was it closed due to lack of activity?

PR #3764, opened 2020-06-05
Add ZStream#runManaged

### LINKED ISSUE #3758: Add ZStream#runManaged
`ZStream#run[B]` produces a value and immediately closes allocated resources.
Sometimes a user wants to do something with `B` before finalizing the resources.
So `ZStream` has a bunch of `*managed` methods e.g. `foreachChunkWhileManaged` to lift limitations of `run`.
What we need:
1) add `ZStream#runManaged(sink: ZSink[B])` which returns `ZManaged[R, E, B]`
2) `ZStream.run` will just delegate to `runManaged`
3) all `ZStream#*Managed` methods  go away

simpadjo:
Closes #3758

After adding `ZStream#runManaged` all other `ZStream#.....Managed` methods become redundant. 
~~I left the following methods~~
~~1) that spark joy~~
~~2) that are used internally~~
~~3) that have tests~~

I kept all existing methods, made them delegating to appropriate sink

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1148,25 +1148,26 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * Consumes all elements of the stream, passing them to the specified callback.
    */
   final def foreach[R1 <: R, E1 >: E](f: O => ZIO[R1, E1, Any]): ZIO[R1, E1, Unit] =
-    foreachWhile(f.andThen(_.as(true)))
+    run(ZSink.foreach(f))
 
   /**
    * Consumes all elements of the stream, passing them to the specified callback.
    */
   final def foreachChunk[R1 <: R, E1 >: E](f: Chunk[O] => ZIO[R1, E1, Any]): ZIO[R1, E1, Unit] =
-    foreachChunkWhile(f.andThen(_.as(true)))
+    run(ZSink.foreachChunk(f))
 
   /**
    * Like [[ZStream#foreachChunk]], but returns a `ZManaged` so the finalization order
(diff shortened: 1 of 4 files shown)


simpadjo:
@iravid dotty error is irrelevant

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2292,7 +2294,10 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * Runs the sink on the stream to produce either the sink's result or an error.
    */
   def run[R1 <: R, E1 >: E, B](sink: ZSink[R1, E1, O, Any, B]): ZIO[R1, E1, B] =
-    (process <*> sink.push).use {
+    runManaged(sink).useNow
+
+  def runManaged[R1 <: R, E1 >: E, B](sink: ZSink[R1, E1, O, Any, B]): ZManaged[R1, E1, B] =
+    (process <*> sink.push).flatMap {
```
iravid, line 2300:
```suggestion
    (process <*> sink.push).mapM {
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2308,7 +2313,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
               .foldCauseM(c => Cause.sequenceCauseEither(c.map(_._1)).fold(IO.halt(_), ZIO.succeedNow), _ => go)
         )
 
-        go
+        go.toManaged_
```
iravid, line 2316:
```suggestion
        go
```

simpadjo: :+1:

Issue #3740, opened 2020-06-01
Fiber failure with ZStream.aggregateAsyncWithin and ZStream.interruptWhen


svroonland:
In this infinite stream, `interruptWhen(ZIO.never)` is used, yet the stream fails with an interruption error. This appears to be present in RC19-2 and RC20 (https://scastie.scala-lang.org/ywITkPfqR3e627YCKEC9uA and https://scastie.scala-lang.org/0fDYgZ5jSBO3gT84LQiwYA)

svroonland:
I tried to change `interruptWhen` to use `raceAll` instead of `raceFirst`. That gives even stranger behavior, with the stream appearing to be stuck. See example here, no more "aggregate async done" messages are being printed after a short while.

https://scastie.scala-lang.org/5WAVeooJTfeT5kxfMrNNdA

luis3m:
@svroonland it's the same issue. The thing is, `raceAll` interrupts the other fibers as soon as it gets its first success. `as` is being interrupted therefore it waits for the promise to succeed and that's why it hangs.

```scala
as.tapCause(c => UIO(println(c.prettyPrint)))
```
Add this to `myInterruptWhen` and you'll get an interruption error.

iravid:
Hey @luis3m. Do you have any insight as to what's going on here?

luis3m:
@iravid sort of, I'm still trying to understand why it's happening, I will give an update tomorrow. It's very likely I will need help understanding the issue but can tell that disowning `producerWaiting` solves the issue.

luis3m:
Thinking a bit deeper I realized that this should also happen with other combinators such as debounce as it also stores a fiber to await later and I was right, `debounce` does not work with `interruptWhen` (or any other combinator racing fibers I guess):

https://scastie.scala-lang.org/hjcUbjl5QfSTHKkC5MTIzw

luis3m:
I feel I'm missing something basic here but my understanding is that it's not working because parent fibers terminate children which is why these fibers saved for later are being interrupted but I fail to understand why `interruptWhen` is needed to recreate the issue.

svroonland:
I played around with this and I think I also understand now, thanks to @luis3m. 

The forked fiber for `waitForProducer` created by `updateSchedule.raceWith(waitForProducer)` is a child fiber of the `runIO.join.raceFirst(as)` in `interruptWhen`, which also forks `as`. So there's:
[...]

PR #3765, opened 2020-06-05
Fix ZStream#debounce interruption issue

luis3m:
Relates to #3740

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2702,7 +2702,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
         pull = {
           def store(chunk: Chunk[O2]): URIO[Clock, Chunk[O2]] =
             chunk.lastOption
-              .map(last => clock.sleep(d).as(last).fork.flatMap(f => ref.set(Previous(f))))
+              .map(last => clock.sleep(d).as(last).forkDaemon.flatMap(f => ref.set(Previous(f))))
               .getOrElse(ref.set(NotStarted))
               .as(Chunk.empty)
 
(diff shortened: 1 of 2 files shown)


luis3m:
Actually, I think only the `cause.interrupted` is needed here. The other two cases require the fiber to finish. Will update in a few minutes.

luis3m:
Done

luis3m:
I will change `tapError` with `onInterrupt`

luis3m:
Done

luis3m:
@iravid this is ready for review

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2734,7 +2734,13 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
             case Done =>
               Pull.end
-          }
+          }.onInterrupt(
+            ref.get.flatMap {
+              case Previous(fiber) => fiber.interrupt
+              case Current(fiber)  => fiber.interrupt
+              case _               => ZIO.unit
+            }
+          )
```
iravid, line 2743: Sorry for this PR dragging out. It's probably better for this finalizer to be attached to the construction of the `Ref`; this way the fiber is interrupted even if the stream is interrupted "in-between" pulls. What do you think?

luis3m: @iravid I think you mean:
```scala
Ref.make[State](NotStarted).toManaged {
  _.get.flatMap {
    case Previous(fiber) => fiber.interrupt
    case Current(fiber)  => fiber.interrupt
    case _               => ZIO.unit
  }
}
```
Don't you?

iravid: Yes, exactly

luis3m: @iravid makes sense, done. Also added a new test case for the fiber interruption.

PR #3786, opened 2020-06-09
Make ZStream#either not terminate stream on error

luis3m:
- Based on https://discord.com/channels/629491597070827530/630498701860929559/719935596415418420
- I'm aware `collectWhileM { case x => ZIO.fail(e) }.either` does not play very well.

Thoughts? I suspect this wasn't a slip.


luis3m:
This would probably cause unexpected behaviors :thinking:, I'm inclined to close this PR but would like an expert opinion.

simpadjo:
In general it's not safe to pull again from a failed stream.
It's similar to not calling `next` on a failed iterator.

luis3m:
@simpadjo alright, I'm going to close this one then

PR #3788, opened 2020-06-09
add ZSink.managed

### LINKED ISSUE #3651: Add ZSink.managed
```scala
def managed[R, E, I, A, Z](m: R, E, A)(push: Chunk[I] => ZIO[R, E, Unit])
```

simpadjo:
closes #3651 

1) Add `ZSink.managed`
2) Utilize it for `ZSink.fromOutputStreamManaged`
3) Also refactor `ZSink.fromFile`
4) Reify `ZSink.foreachChunk` signature

@regis-leray sorry to steal this ticket from you, I need this at work :)
@lvitaly - pinging you as the author of `ZSink#fromFile`

### DIFF
--- a/streams/jvm/src/main/scala/zio/stream/platform.scala
+++ b/streams/jvm/src/main/scala/zio/stream/platform.scala
@@ -22,14 +22,26 @@ trait ZSinkPlatformSpecificConstructors { self: ZSink.type =>
    */
   final def fromOutputStream(
     os: OutputStream
-  ): ZSink[Blocking, IOException, Byte, Nothing, Long] =
-    ZSink.foldLeftChunksM(0L) { (bytesWritten, byteChunk: Chunk[Byte]) =>
-      blocking.effectBlockingInterrupt {
-        val bytes = byteChunk.toArray
-        os.write(bytes)
-        bytesWritten + bytes.length
-      }.refineOrDie {
-        case e: IOException => e
+  ): ZSink[Blocking, IOException, Byte, Byte, Long] = fromOutputStreamManaged(ZManaged.succeedNow(os))
+
+  /**
+   * Uses the provided `OutputStream` resource to create a [[ZSink]] that consumes byte chunks
+   * and writes them to the `OutputStream`. The sink will yield the count of bytes written.
+   *
+   * The `OutputStream` will be automatically closed after the stream is finished or an error occurred.
+   */
+  final def fromOutputStreamManaged(
+    os: ZManaged[Blocking, IOException, OutputStream]
+  ): ZSink[Blocking, IOException, Byte, Byte, Long] =
+    ZSink.managed(os) { out =>
+      ZSink.foldLeftChunksM(0L) { (bytesWritten, byteChunk: Chunk[Byte]) =>
+        blocking.effectBlockingInterrupt {
+          val bytes = byteChunk.toArray
+          out.write(bytes)
+          bytesWritten + bytes.length
+        }.refineOrDie {
+          case e: IOException => e
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -824,6 +824,9 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
       } yield push
     }
 
+  def managed[R, E, I, A, L, Z](resource: ZManaged[R, E, A])(fn: A => ZSink[R, E, I, L, Z]): ZSink[R, E, I, L, Z] =
```
regis-leray, line 827: this signature is much simpler 👍

simpadjo: yeah, I started with the original signature from the ticket, then tried to use it in `fromFile` and it immediately folded. Nested resources are too useful.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -41,37 +53,31 @@ trait ZSinkPlatformSpecificConstructors { self: ZSink.type =>
...
+    )(chan => blocking.effectBlocking(chan.close()).orDie)
+
+    val writer = ZSink.managed(managedChannel) { chan =>
+      ZSink.foreachChunk[Blocking, Throwable, Byte](byteChunk =>
+        blocking.effectBlockingInterrupt {
+          chan.write(ByteBuffer.wrap(byteChunk.toArray))
+        }
+      )
     }
+    writer &> ZSink.count
```
iravid, line 79: Very nice @simpadjo

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -745,8 +745,8 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
   /**
    * A sink that executes the provided effectful function for every chunk fed to it.
    */
-  def foreachChunk[R, E, I](f: Chunk[I] => ZIO[R, E, Any]): ZSink[R, E, I, I, Unit] =
-    ZSink.fromPush[R, E, I, I, Unit] {
+  def foreachChunk[R, E, I](f: Chunk[I] => ZIO[R, E, Any]): ZSink[R, E, I, Nothing, Unit] =
```
iravid, line 748: 👍

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -824,6 +824,13 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
       } yield push
     }
 
+  /**
+   * A sink that depends on another managed value
+   * `resource` will be finalized after the processing.
+   */
+  def managed[R, E, I, A, L, Z](resource: ZManaged[R, E, A])(fn: A => ZSink[R, E, I, L, Z]): ZSink[R, E, I, L, Z] =
+    ZSink(resource.fold[ZSink[R, E, I, L, Z]](err => ZSink.fail[E, I](err).dropLeftover, m => fn(m)).flatMap(_.push))
```
iravid, line 832: Why are the leftovers dropped in the error case?

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -41,37 +53,31 @@ trait ZSinkPlatformSpecificConstructors { self: ZSink.type =>
...
+          FileChannel
+            .open(
+              path,
+              options.foldLeft(new ju.HashSet[OpenOption]()) { (acc, op) =>
+                acc.add(op); acc
+              } // for avoiding usage of different Java collection converters for different scala versions
+            )
+            .position(position)
+        )
+        .orDie
```
iravid, line 69: I seem to have missed this `orDie` here. Why is it required?

streams/shared/src/main/scala/zio/stream/ZSink.scala
simpadjo: `ZSink.fail` returns leftovers of type `I` and we require `L`.
I don't know how to properly treat allocation errors w/o loosing too much of ergonomics.

streams/jvm/src/main/scala/zio/stream/platform.scala
simpadjo: It doesn't (I removed it).
Actually, we could reify `Throwable` to `IOException`, but I'm not sure how it would work in case of interruption.

streams/shared/src/main/scala/zio/stream/ZSink.scala
simpadjo: Alternatively we could require ` ZManaged[R, Nothing, A]` but I expect most of the resources to be IO-related.

iravid: Let’s require ZSink[R, E, I, I, Z] here. Dropping the leftovers is a bad choice.

adamgfraser:
@simpadjo Can you fix the compilation errors on Dotty?

PR #3798, opened 2020-06-12
Optimize TPriorityQueue

adamgfraser:


core-tests/shared/src/test/scala/zio/stm/TPriorityQueueSpec.scala
```diff
@@ -15,6 +15,24 @@ object TPriorityQueueSpec extends ZIOBaseSpec {
         } yield values
         assertM(transaction.commit)(equalTo(vs.sorted))
       }
+    },
+    testM("removeIf") {
+      val elems = List(2, 4, 6, 3, 5, 6)
```
phderome, line 20: you could possibly choose the even numbers out of order here to show evidence that the priority matters.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -16,137 +16,105 @@
...
   def take: USTM[A] =
-    takeOption.collect { case Some(a) => a }
+    takeOption.get.orElse(STM.retry)
 
   /**
    * Takes all values from the queue.
    */
   def takeAll: USTM[List[A]] =
-    map.size.flatMap(n => STM.collectAll(STM.replicate(n)(take)))
+    ref.modify(map => (map.flatMap { case (k, v) => List.fill(v)(k) }.toList, map.empty))
```
phderome, line 87: ha, v is the count reference of how many times k occurs.

core-tests/shared/src/test/scala/zio/stm/TPriorityQueueSpec.scala
adamgfraser: Changed it to a property based test to cover that.

adamgfraser:
@phderome Thank you for taking a look!

PR #3803, opened 2020-06-13
Make ZStream#buffer buffer single values

### LINKED ISSUE #3766: ZStream.buffer buffers chunks instead of single values
For example, `someZStream.buffer(100)` will buffer up to 100 chunks instead of 100 elements.

luis3m:
Closes #3766

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -381,13 +381,41 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     distributedWithDynamic(maximumLag, _ => decider, _ => ZIO.unit).map(_.map(_._2))
   }
 
+  private final def bufferElement[E1 >: E, O1 >: O](queue: Queue[Exit[Option[E1], O1]]): URManaged[R, Pull[R, E1, O1]] =
+    for {
+      done <- Ref.makeManaged(false)
+      os   <- self.process.mapM(BufferedPull.make(_))
+      _    <- os.pullElement.run.flatMap(e => queue.offer(e).as(e)).doWhile(_.succeeded).forkManaged
+      pull = done.get.flatMap {
+        if (_) Pull.end
+        else
+          queue.take.flatMap(exit => ZIO.done(exit.map(Chunk.single))).catchSome {
+            case None => done.set(true) *> Pull.end
+          }
+      }
+    } yield pull
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * up to `capacity` elements in a queue.
+   *
+   * @note Prefer capacities that are powers of 2 for better performance.
+   */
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -463,18 +491,46 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * up to `capacity` elements in a sliding queue.
+   *
+   * @note Prefer capacities that are powers of 2 for better performance.
+   */
+  final def bufferSliding(capacity: Int): ZStream[R, E, O] =
+    ZStream {
+      for {
+        queue <- Queue.sliding[Exit[Option[E], O]](capacity).toManaged(_.shutdown)
+        pull  <- bufferElement(queue)
```
iravid, line 530: You'll need to replicate the `bufferSignal` logic for the `bufferSliding` and `bufferDropping` cases; your `bufferElement` implementation is good for the plain backpressuring queue case, but when the queue is sliding or dropping, it is likely to lose error or end signals.

We should also replicate the testcases so we are guarding against these issues.

luis3m: Good point, will update

luis3m: Hmm, isn't this just a problem for bufferDroping though? In case of bufferSliding, the producer will stop as soon as one error is found therefore it won't be lost. Am I missing something? @iravid

iravid:
Happy to help on this one too @luis3m

jdegoes:
@luis3m This is a great feature addition. Please re-open the pull request when you have a bit more time to push it through, and thanks for your work to date on it!

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: @iravid I'm about to reopen this pull request, I'm now wondering how the `bufferElement` implementation won't work for `bufferSliding` and `bufferDropping` if the producer stops as soon as there's an error encountered. How the error or end signal will be lost?

iravid: For bufferDropping, consider the case where the consumer is slow and the queue is full. When the producer tries to offer an end/error signal, it gets dropped. The producer then stops. The consumer eventually drains the queue and hangs because the queue is empty and nothing is producing to it anymore.

bufferSliding is a different case. In that case, if the consumer is slow, we will be dropping an element that is present in the queue in favour of an error/end signal. This is arguable but I think that's undesirable.

Either way, I think just slapping a BufferedPull on the existing implementation of `bufferSignal` could do the trick with a minimal amount of work.

iravid: And thank you for following up!

PR #3804, opened 2020-06-13
Add ZSink.toQueue

### LINKED ISSUE #3735: Add ZSink.toQueue
There is `ZStream#into` operator that allows for writing elements of a stream into a queue.
Would be better to have this functionality as a `ZSink` instead because sinks can be composed.

This sink could also take a boolean parameter like `shutdownWhenFinished` to decide if the queue should be terminated after the stream is done.

I also feel that it's possible to remove `Take` from the queue type w/o loss of functionality.

luis3m:
Closes #3735

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -838,4 +838,28 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
    */
   def sum[A](implicit A: Numeric[A]): ZSink[Any, Nothing, A, Nothing, A] =
     foldLeft(A.zero)(A.plus)
+
+  /**
+   * A sink that offers values to a given queue. Shuts down the queue if specified.
+   */
+  def toQueue[R, I](
+    queue: ZQueue[R, Nothing, Nothing, Any, Chunk[I], Any],
+    shutdown: Boolean
+  ): ZSink[R, Nothing, I, I, Unit] =
+    ZSink {
+      for {
+        state <- Ref.make[Chunk[I]](Chunk.empty).toManaged_
+        push = (is: Option[Chunk[I]]) =>
+          is match {
+            case Some(ch) =>
+              queue.offer(ch) *> state.update(_ ++ ch) *> Push.more
+            case None =>
+              for {
+                _     <- queue.shutdown.when(shutdown)
+                chunk <- state.get
+                value <- Push.emit((), chunk)
+              } yield value
+          }
+      } yield push
+    }
 }
--- a/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
@@ -148,6 +148,24 @@ object ZSinkSpec extends ZIOBaseSpec {
             equalTo(("ok", Chunk(1, 2, 3)))
           )
         }
+      ),
+      suite("toQueue")(
+        testM("fills a queue") {
+          assertM(
+            for {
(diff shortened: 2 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -838,4 +838,28 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
...
+  /**
+   * A sink that offers values to a given queue. Shuts down the queue if specified.
+   */
+  def toQueue[R, I](
+    queue: ZQueue[R, Nothing, Nothing, Any, Chunk[I], Any],
+    shutdown: Boolean
+  ): ZSink[R, Nothing, I, I, Unit] =
+    ZSink {
+      for {
+        state <- Ref.make[Chunk[I]](Chunk.empty).toManaged_
```
iravid, line 851: I don't understand the semantics here. Why do we need to accumulate all chunks offered to the queue and pass them on as a leftover? This uses unbounded memory and seems pretty unsafe to me.

Beyond that, I wouldn't expect a queue sink to result in any leftovers so the effort sounds a bit redundant; we could probably use `Nothing` for leftovers here and consider anything offered to the queue to be consumed.

What do you think @luis3m?

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -838,4 +838,28 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
    */
   def sum[A](implicit A: Numeric[A]): ZSink[Any, Nothing, A, Nothing, A] =
     foldLeft(A.zero)(A.plus)
+
+  /**
+   * A sink that offers values to a given queue. Shuts down the queue if specified.
+   */
+  def toQueue[R, I](
```
iravid, line 845:
```suggestion
  def fromQueue[R, I](
```

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -838,4 +838,28 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
    */
   def sum[A](implicit A: Numeric[A]): ZSink[Any, Nothing, A, Nothing, A] =
     foldLeft(A.zero)(A.plus)
+
+  /**
+   * A sink that offers values to a given queue. Shuts down the queue if specified.
```
iravid, line 843:
```suggestion
   * A sink that offers values to a given queue. Shuts down the queue on stream completion if specified.
```

simpadjo:
I have a feeling that `ZSink.foreach` + something like `.mapM`  could be useful here

streams/shared/src/main/scala/zio/stream/ZSink.scala
simpadjo: Then tests for this functions must be under `suite("nstructors"){}`

streams/shared/src/main/scala/zio/stream/ZSink.scala
luis3m: @iravid Sorry, I'm not familiar with ZSink 😅

So you're saying the leftover part is unnecessary? I thought of leftovers based on the comment that ZSink are composable.

iravid:
Hey @luis3m, need any help getting this over the finish line?

luis3m:
@iravid I forgot about these, I can work on this over the weekend.

iravid:
Thank you!

jdegoes:
@luis3m It sounds like direction for this pull request is clear and just a bit more work necessary to get it in. Please re-open if you have the time to work more on the feature! 🙏

PR #3805, opened 2020-06-13
Error leak inside `effectAsyncM`

tusharmath:
Adding a failing test here for a weird behavior that I bumped into.

```scala
val program =  ZIO
                      .effectAsyncM[Any, Throwable, String](register =>
                        for {
                          _ <- UIO(register(UIO("SUCCEED")))
                          _ <- ZIO.fail(new Throwable("FAIL"))
                        } yield ()
                      )
```

IMHO, the power of ZIO is that there are no leaks in terms of resources and exceptions. In the above `program`, it the failure happening inside the callback is lost magically.

For me, it took me a lot of time to debug that an error was being silently ignored in my program.

---

If the issue seems reasonable, I could help out with a proposed solution to the PR also.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3278,7 +3278,21 @@ object ZIOSpec extends ZIOBaseSpec {
         val effect: Task[Unit] = ZIO.fail(error).unit.orDie.resurrect
         assertM(effect.either)(isLeft(equalTo(error)))
       }
-    )
+    ),
+    suite("effectAsyncM") {
+      testM("should not leak internal errors") {
+        val program =  ZIO.effectAsyncM[Any, Throwable, String](register =>
+            for {
+              _ <- UIO(register(UIO("SUCCEED")))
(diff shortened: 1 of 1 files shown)


iravid:
There's an inherent race condition in effectAsyncM, in which the callback could complete before the failure from the callback registration.

But a failure in the callback registration has to be reflected in the overall effect, otherwise it's a deadlock waiting to happen. E.g., you're running effectAsyncM, there's a failure in the callback registration and nothing's going to call `cb` anymore. What now?

tusharmath:
I think the callback and the registration should both complete successfully, only then should the returned effect complete. Run the registration and callback using `zipPar` ?

jdegoes:
@iravid What is the problem with the failure of the registration propagating back to the stream?

i.e. we can wait for both the registration to complete and for the "first value" to be passed to the callback, and if either fails, we fail the stream with the failure. It would delay the first element but not by much.

@tusharmath Is that what you had in mind?

jdegoes:
I will close this for now because it is failing, but let's continue to discuss.

tusharmath:
Yes @jdegoes.

PR #3807, opened 2020-06-13
Fixed #3806: Update AccessibleMacro to properly handle ZManaged

### LINKED ISSUE #3806: accessible macro does not properly handle ZManaged
When a module has a member with a return type of `ZManaged[R, E, A]`, I think the expected accessor generated by `@accessible` would be `ZManaged[R with Has[Service], E, A]`, but the current implementation seems to fall back to treating it as a method and generates an accessor with return type `ZIO[Has[Service], Throwable, ZManaged[R, E, A]]` which is more difficult to work with and doesn't match the API provided by the other ZIO classes.

I have a PR in the works to address this and should be able to submit it soon.

kentfrazier:
First off, this is my first PR to zio, so please let me know if I am not following proper procedures or if there is anything else I should have done to be a better contributor. I am also only a week or so into using zio, so feedback about better ways to accomplish my goals is also very welcome.

When a module has a member with a return type of `ZManaged[R, E, A]`, I think the expected accessor generated by @accessible would be `ZManaged[R with Has[Service], E, A]`, but the current implementation seems to fall back to treating it as a method and generates an accessor with return type `ZIO[Has[Service], Throwable, ZManaged[R, E, A]]` which is more difficult to work with and doesn't match the API provided by the other ZIO classes.

This PR addresses the issue by adding in specific handling for `ZManaged` in `AccessibleMacro`, in a way that mostly mirrors the way `ZIO` itself is handled (though `ZManaged.accessM` doesn't compose the same way, so I had to use a slightly different approach there.) I was mostly just following the example of the surrounding code, which was fairly easy to follow, especially for macro code. I added tests to verify the new behavior and make sure that it had coverage similar to the level of the effect types.

Let me know if anything looks improper and I am happy to fix it. Thanks!

### DIFF
--- a/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
+++ b/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
@@ -52,6 +52,7 @@ private[macros] class AccessibleMacro(val c: Context) {
     sealed trait Capability
     object Capability {
       case class Effect(r: Tree, e: Tree, a: Tree)                 extends Capability
+      case class Managed(r: Tree, e: Tree, a: Tree)                extends Capability
       case class Method(a: Tree)                                   extends Capability
       case class Sink(r: Tree, e: Tree, a: Tree, l: Tree, b: Tree) extends Capability
       case class Stream(r: Tree, e: Tree, a: Tree)                 extends Capability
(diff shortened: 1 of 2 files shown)


adamgfraser:
@kentfrazier Thanks for contributing! Sounds right to me. @ioleo Can you take a look?

ioleo (approved):
Looks good to me :+1: 
@kentfrazier Thank you for contributing!

kentfrazier:
@adamgfraser any progress on this? It looks like it got approved, but hoping these changes will make it into the next release. Thanks!

adamgfraser (approved):

adamgfraser:
@kentfrazier Thanks for the reminder! Merged now.

PR #3813, opened 2020-06-15
Add a new `fromOptionError` operator

paulpdaniels:
Adds a new `fromOptionError` operator which has better composability with the existing `some`/`optional`/`asSomeError` combinators.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -406,7 +406,7 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
     }
 
   /**
-   * Unwraps the optional success of this effect, but can fail with unit value.
+   * Unwraps the optional success of this effect, but can fail with None value.
    */
   def get[B](implicit ev1: E <:< Nothing, ev2: A <:< Option[B]): ZManaged[R, Option[Nothing], B] =
     ZManaged.absolve(mapError(ev1)(CanFail).map(ev2(_).toRight(None)))
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -408,8 +408,8 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
   /**
    * Unwraps the optional success of this effect, but can fail with unit value.
    */
-  def get[B](implicit ev1: E <:< Nothing, ev2: A <:< Option[B]): ZManaged[R, Unit, B] =
-    ZManaged.absolve(mapError(ev1)(CanFail).map(ev2(_).toRight(())))
+  def get[B](implicit ev1: E <:< Nothing, ev2: A <:< Option[B]): ZManaged[R, Option[Nothing], B] =
+    ZManaged.absolve(mapError(ev1)(CanFail).map(ev2(_).toRight(None)))
 
   /**
    * Returns a new effect that ignores the success or failure of this effect.
--- a/docs/overview/creating_effects.md
+++ b/docs/overview/creating_effects.md
@@ -58,7 +58,7 @@ Scala's standard library contains a number of data types that can be converted i
 An `Option` can be converted into a ZIO effect using `ZIO.fromOption`:
 
 ```scala mdoc:silent
-val zoption: IO[Unit, Int] = ZIO.fromOption(Some(2))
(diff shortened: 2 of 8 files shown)


adamgfraser:
I think this is definitely interesting. Have you explored using `Option[Nothing]` instead of `Unit` more broadly for the error type? The two types are isomorphic but agree this is more composable so wonder if this is the more practical signature, though it is pretty late for API changes.

iravid:
I find this vastly more useful than the current fromOption, FWIW.

adamgfraser:
Yes. I wonder if we just move everything over.

paulpdaniels:
> Yes. I wonder if we just move everything over.

I'm ok with moving everything over to this. I use this extension locally in most of my projects because I wanted a drop in replacement for `OptionT`. I don't think I have any current instances of `fromOption` because they always require `mapError` or `asError` directly after them to make them work with other operators.

Also I'd prefer to move, because the naming of this operator is somewhat confusing, it would be better for it to just be the defacto.

iravid:
Agreed

docs/overview/creating_effects.md
```diff
@@ -58,7 +58,7 @@ Scala's standard library contains a number of data types that can be converted i
 An `Option` can be converted into a ZIO effect using `ZIO.fromOption`:
 
 ```scala mdoc:silent
-val zoption: IO[Unit, Int] = ZIO.fromOption(Some(2))
+val zoption: IO[Option[Nothing], Int] = ZIO.fromOption(Some(2))
 ```
 
 The error type of the resulting effect is `Unit`, because the `None` case of `Option` provides no information on why the value is not there. You can change the `Unit` into a more specific error type using `ZIO#mapError`:
```
AdamPaynter, line 64: Does this paragraph need to be updated in lieu of the error channel now being defined as `Option[Nothing]` instead of `Unit`?

jdegoes:
Agree to use `Option[Nothing]` instead of `Unit` for `fromOption` and related cases. 👍

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -408,8 +408,8 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
   /**
    * Unwraps the optional success of this effect, but can fail with unit value.
    */
-  def get[B](implicit ev1: E <:< Nothing, ev2: A <:< Option[B]): ZManaged[R, Unit, B] =
-    ZManaged.absolve(mapError(ev1)(CanFail).map(ev2(_).toRight(())))
+  def get[B](implicit ev1: E <:< Nothing, ev2: A <:< Option[B]): ZManaged[R, Option[Nothing], B] =
```
jdegoes, line 411: Would be good to update the comment for this one!

docs/overview/creating_effects.md
```diff
@@ -58,15 +58,30 @@ Scala's standard library contains a number of data types that can be converted i
 An `Option` can be converted into a ZIO effect using `ZIO.fromOption`:
 
 ```scala mdoc:silent
-val zoption: IO[Unit, Int] = ZIO.fromOption(Some(2))
+val zoption: IO[Option[Nothing], Int] = ZIO.fromOption(Some(2))
 ```
 
-The error type of the resulting effect is `Unit`, because the `None` case of `Option` provides no information on why the value is not there. You can change the `Unit` into a more specific error type using `ZIO#mapError`:
+The error type of the resulting effect is `Option[Nothing]`, because the `None` case of `Option` provides no information on why the value is not there. You can change the `Option[Nothing]` into a more specific error type using `ZIO#mapError`:
```
jdegoes, line 64:
```suggestion
The error type of the resulting effect is `Option[Nothing]`, which provides no information on why the value is not there. You can change the `Option[Nothing]` into a more specific error type using `ZIO#mapError`:
```

docs/overview/creating_effects.md
```diff
@@ -58,15 +58,30 @@ Scala's standard library contains a number of data types that can be converted i
...
 
 ```scala mdoc:silent
 val zoption2: IO[String, Int] = zoption.mapError(_ => "It wasn't there!")
 ```
 
+You can also readily compose it with other operators while preserving the optional nature of the result (similar to an `OptionT`)
+
+```scala mdoc:silent
+val maybeId: IO[Option[Nothing], String] = ZIO.fromOption(Some("abc123"))
```
jdegoes, line 73: Really nice example!

jdegoes:
@paulpdaniels Made a few comments but looks really good! Let me know when it's ready & tests are passing. Thank you for your work on this!

paulpdaniels:
@jdegoes comments addressed!

PR #3814, opened 2020-06-15
Add Documentation for TPriorityQueue

### LINKED ISSUE #3308: Interrupt w/ bracket doesn't work.
```scala
import zio._
import zio.duration._
import zio.console.putStrLn

...

    val rt = Runtime.default
    val mw = (putStrLn("+") *> ZIO.sleep(1.seconds)).forever
    val io = for {
      r <- ZIO.bracket(mw.fork)(_.interrupt) { _ =>
        (putStrLn("*") *> ZIO.sleep(1.seconds)).repeat(Schedule.recurs(5))
      }
    } yield r

    println(rt.unsafeRun(io))
```

Never terminates.

### LINKED ISSUE #3808: Add to zio.dev Data Types doc for TPriorityQueue
May need to be tied to next RC release.

adamgfraser:
Resolves #3808. Copying @phderome.

### DIFF
--- a/docs/datatypes/index.md
+++ b/docs/datatypes/index.md
@@ -18,7 +18,7 @@ ZIO contains a small number of data types that can help you solve complex proble
  - **[Chunk](chunk.md)** — ZIO `Chunk`: Fast, Pure Alternative to Arrays
  - **[TArray](tarray.md)** - A `TArray[A]` is an array of mutable references that can participate in transactions.
  - **[TMap](tmap.md)** - A `TMap[A]` is a mutable map that can participate in transactions.
- - **[TPriorityQueue](tpriorityqueue.md)** - A `TPriorityQueue[A]` is a mutabe priority queue that can participate in transactions.
+ - **[TPriorityQueue](tpriorityqueue.md)** - A `TPriorityQueue[A]` is a mutable priority queue that can participate in transactions.
  - **[TPromise](tpromise.md)** - A `TPromise` is a mutable reference that can be set exactly once and can participate in transactions.
  - **[TQueue](tqueue.md)** - A `TQueue` is a mutable queue that can participate in transactions.
  - **[TRef](tref.md)** - A `TRef` is a mutable reference to an immutable value that can participate in transactions.
--- a/docs/datatypes/index.md
+++ b/docs/datatypes/index.md
@@ -18,6 +18,7 @@ ZIO contains a small number of data types that can help you solve complex proble
  - **[Chunk](chunk.md)** — ZIO `Chunk`: Fast, Pure Alternative to Arrays
  - **[TArray](tarray.md)** - A `TArray[A]` is an array of mutable references that can participate in transactions.
  - **[TMap](tmap.md)** - A `TMap[A]` is a mutable map that can participate in transactions.
+ - **[TPriorityQueue](tpriorityqueue.md)** - A `TPriorityQueue[A]` is a mutabe priority queue that can participate in transactions.
  - **[TPromise](tpromise.md)** - A `TPromise` is a mutable reference that can be set exactly once and can participate in transactions.
  - **[TQueue](tqueue.md)** - A `TQueue` is a mutable queue that can participate in transactions.
  - **[TRef](tref.md)** - A `TRef` is a mutable reference to an immutable value that can participate in transactions.
--- a/docs/datatypes/tpriorityqueue.md
+++ b/docs/datatypes/tpriorityqueue.md
@@ -1,0 +1,70 @@
+---
+id: datatypes_tpriorityqueue
+title:  "TPriorityQueue"
+---
+
+A `TPriorityQueue[A]` is a mutable queue that can participate in STM transactions. A `TPriorityQueue` contains values of type `A` that an `Ordering` is defined on. Unlike a `TQueue`, `take` returns the highest priority value (the value that is first in the specified ordering) as opposed to the first value offered to the queue. The ordering that elements with the same priority will be taken from the queue is not guaranteed.
+
+## Creating a TPriorityQueue
+
+You can create an empty `TPriorityQueue` using the `empty` constructor:
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val minQueue: STM[Nothing, TPriorityQueue[Int]] =
+  TPriorityQueue.empty
+```
+
+Notice that a `TPriorityQueue` is created with an implicit `Ordering`. By default, `take` will return the value that is first in the specified ordering. For example, in a queue of events ordered by time the earliest event would be taken first. If you want a different behavior you can use a custom `Ordering`.
+
+```scala mdoc:silent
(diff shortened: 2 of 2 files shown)


docs/datatypes/index.md
```diff
@@ -18,6 +18,7 @@ ZIO contains a small number of data types that can help you solve complex proble
  - **[Chunk](chunk.md)** — ZIO `Chunk`: Fast, Pure Alternative to Arrays
  - **[TArray](tarray.md)** - A `TArray[A]` is an array of mutable references that can participate in transactions.
  - **[TMap](tmap.md)** - A `TMap[A]` is a mutable map that can participate in transactions.
+ - **[TPriorityQueue](tpriorityqueue.md)** - A `TPriorityQueue[A]` is a mutabe priority queue that can participate in transactions.
```
phderome, line 21:
```suggestion
 - **[TPriorityQueue](tpriorityqueue.md)** - A `TPriorityQueue[A]` is a mutable priority queue that can participate in transactions.
```

adamgfraser: @phderome Thank you! fixed now.

docs/datatypes/tpriorityqueue.md
```diff
@@ -0,0 +1,70 @@
+---
+id: datatypes_tpriorityqueue
+title:  "TPriorityQueue"
+---
+
+A `TPriorityQueue[A]` is a mutable queue that can participate in STM transactions. A `TPriorityQueue` contains values of type `A` that an `Ordering` is defined on. Unlike a `TQueue`, `take` returns the highest priority value (the value that is first in the specified ordering) as opposed to the first value offered to the queue. The ordering that elements with the same priority will be taken from the queue is not guaranteed.
```
phderome, line 6:
```suggestion
A `TPriorityQueue[A]` is a mutable queue that can participate in STM transactions. A `TPriorityQueue` contains values of type `A` for which an `Ordering` is defined. Unlike a `TQueue`, `take` returns the highest priority value (the value that is first in the specified ordering) as opposed to the first value offered to the queue. The ordering of elements sharing the same priority when taken from the queue is not guaranteed.
```

docs/datatypes/tpriorityqueue.md
```diff
@@ -0,0 +1,70 @@
...
+
+```scala mdoc:silent
+import zio._
+import zio.stm._
+
+val minQueue: STM[Nothing, TPriorityQueue[Int]] =
+  TPriorityQueue.empty
+```
+
+Notice that a `TPriorityQueue` is created with an implicit `Ordering`. By default, `take` will return the value that is first in the specified ordering. For example, in a queue of events ordered by time the earliest event would be taken first. If you want a different behavior you can use a custom `Ordering`.
```
phderome, line 20:
```suggestion
Notice that a `TPriorityQueue` is created with an implicit `Ordering`. By default, `take` will return the value that is first in the specified ordering. For example, in a queue of events ordered by time the **last** event would be taken first. If you want a different behavior you can use a custom `Ordering`.
```

docs/datatypes/tpriorityqueue.md
```diff
@@ -0,0 +1,70 @@
...
+```scala mdoc:silent
+val sorted: STM[Nothing, Chunk[Int]] =
+  for {
+    queue  <- TPriorityQueue.empty[Int]
+    _      <- queue.offerAll(List(2, 4, 6, 3, 5, 6))
+    sorted <- queue.takeAll
+  } yield sorted
+```
+
+You can also use `takeOption` method to take the first value from the queue if it exists without suspending or the `poll` method to observe the first element of the queue if it exists without removing it from the queue.
```
phderome, line 54: is not it `peek` rather than `poll`? Peek is the standard term for such an operation.

docs/datatypes/tpriorityqueue.md
adamgfraser: This isn't correct. The earliest duration event will be taken first.

docs/datatypes/tpriorityqueue.md
adamgfraser: Yes you're right. Good catch!

docs/datatypes/tpriorityqueue.md
adamgfraser: Incorporated.

adamgfraser:
This is great! Thanks for taking a look. Incorporated all of your suggestions except the comment about the ordering of taking events which I don't think is correct. If you have an example that is showing that could you please share?

docs/datatypes/tpriorityqueue.md
phderome: I swear I tried this out. Does not a Priority Queue on Longs take highest numbers first and thus, epochs with highest numbers (latest) taken first?

adamgfraser: `scala.collection.mutable.PriorityQueue` takes the "last" element in the ordering first, so if you put in `List(2, 4, 6, 3, 5, 6)` and then took all the values you would get `List(6, 6, 5, 4, 3, 2)`. `TPriorityQueue` as currently implemented takes the "first" element in the ordering first, so It would return `List(2, 3, 4, 5, 6, 6)`. I think that is the more intuitive behavior but admittedly it is somewhat arbitrary and the difference is less than ideal so we could certainly change.

phderome: https://scastie.scala-lang.org/GXNM6EUOTTKHXRejowPHig

phderome: but when people think of a priority queue, they think of servicing elements with **highest** (largest) priority value.

phderome: wikipedia: In computer science, a priority queue is an abstract data type similar to regular queue or stack data structure in which each element additionally has a "priority" associated with it. In a priority queue, an element with **high** priority is served before an element with low priority**.**

phderome: I admit the standard understanding of priority or ordering forces users to do something different for a typical arrival time view of things.

adamgfraser: Yeah, I think the question is which element should have the highest "priority". For example the Haskell implementation [here](https://hackage.haskell.org/package/stm-data-collection-0.1.0.0/candidate/docs/Data-STM-PriorityQueue-Class.html) also returns the lowest value with the "smallest" key when you take an element. So I think both choices are equally valid it is just which one you want to use. Taking the "smallest" value first seems to be more useful for some applications like events and is consistent with sorting, but the "largest value first" is maybe more consistent with "highest priority equals highest value" and with `scala.collection.mutable.Queue`.

@mijicd Do you have an opinion?

mijicd: I would stay aligned with haskell's API, even though I understand what Phil's saying above. We should be able to pick an ordering on creation if I'm not mistaken.

PR #3815, opened 2020-06-15
Implement TArray#size

### LINKED ISSUE #3552: Add size for TArray and TMap
TArray and TMap don't have `size` method.

adamgfraser:
Resolves #3552.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TArray.scala
+++ b/core/shared/src/main/scala/zio/stm/TArray.scala
@@ -331,6 +331,12 @@ final class TArray[A] private[stm] (private[stm] val array: Array[TRef[A]]) exte
     }
 
   /**
+   * Returns the size of the array.
+   */
+  def size: Int =
+    array.size
+
+  /**
    * Collects all elements into a list.
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/TArray.scala
```diff
@@ -330,6 +330,12 @@ final class TArray[A] private[stm] (private[stm] val array: Array[TRef[A]]) exte
       }
     }
 
+  /**
+   * Returns the size of the array.
+   */
+  def size: Int =
```
mijicd, line 336: Should we lift this value to STM? It would be consistent with the API, although this version has better performance as it removes one mapping.

adamgfraser: I thought about it or adding a note. It seems strange because the size is a constant so it doesn't require an effect to evaluate but don't feel strongly.

mijicd: Let's keep it this way then.

Issue #3775, opened 2020-06-06
Leftovers of Zsink leftovers


simpadjo:
This  PR https://github.com/zio/zio/pull/3722 need few follow-ups:
 
1) `ZSink#toTransducer`
In case of error, sink doesn't push leftovers to the next incarnation
@iravid suggests that it should
I'm not sure about it: 

 iravid yesterday Member
> This is a weak spot. We have no contract that says that a transducer cannot be called after failing (and we shouldn't have such a contract). That means we're dropping leftovers inputs after failing.
> How about placing these leftovers in a Ref that will be read on every push?

simpadjo

> I don't disagree with you, although I can't come up with a testcase so far.
> Seems that once you called stream.transduce(t) you can't see beyond the first error.

~~2) `ZSink#collectAllWhileWith` must be reimplemented~~ DONE
3) `SinkUtils`: `sinkRaceLaw` and `zipParLaw` must check leftovers as well
~~4) `ZStream#foldWhileManagedM` must check the `cont` more often (addressed here: https://github.com/zio/zio/pull/3764)~~ DONE

PR #3817, opened 2020-06-15
Improve ZSink laws

simpadjo:
Part of #3775 

1) Check leftver properties for `ZSink.zipPar` and `ZSink.race`

Collateral changes:
2) Hide `.exposeLeftover`. It is useful for testing but not lawful (drops leftovers in case of error).
3) Add `ZSink.either`

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -177,6 +177,19 @@ abstract class ZSink[-R, +E, -I, +L, +Z] private (
     contramapChunksM(f).mapM(g)
 
   /**
+   * Catches potential error into `Either`
+   */
+  def either: ZSink[R, Nothing, I, L, Either[E, Z]] = ZSink {
+    self.push.map(p =>
+      in => {
+        p(in).catchAll {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -176,6 +176,19 @@ abstract class ZSink[-R, +E, -I, +L, +Z] private (
   ): ZSink[R1, E1, I2, L, Z2] =
     contramapChunksM(f).mapM(g)
 
+  /**
+   * Catches potential error into `Either`
+   */
+  def either: ZSink[R, Nothing, I, L, Either[E, Z]] = ZSink {
```
iravid, line 182: Can this be implemented in terms of ZSink#foldM?

simpadjo: I tried `foldM(e => ZSink.succeed(Left(e)), z => ZSink.succeed(Right(z)))`.
Unfortunately, leftovers typed don't match.

jdegoes:
@iravid Please re-open / take over as necessary, and @simpadjo, thanks for your work on this!

PR #3819, opened 2020-06-16
fix ZStream.grouped

### LINKED ISSUE #3818: ZStream.grouped can return empty Lists
This behavior changed after the ZTransducer changes:

```scala
ZStream.fromIterable(List.empty[Int]).grouped(5).mapM { x =>
  // This block will run with `x` being an empty list
  ...
}
```

I found this a bit surprising. Prior to RC19, that block would not run since the stream is empty and there is nothing to map.

simpadjo:
Closes #3818 
This fix looks like a dirty hack but this is fine.
`grouped` is implemented using Transducer. 
Empty output is a valid output of the transducer (it has to emit something).
So we have to filter-out empty chunks explicitly.


iravid:
I’m not in front of the code so I don’t understand why would collectAllN emit empty elements. It can just emit an empty chunk with no elements.

But happy to merge this to get into the next release.

PR #3824, opened 2020-06-16
ZIO Test: Increase Polling Interval

adamgfraser:
Increases the polling interval for the `TestClock` to wait for all fibers to enter into a suspended state. Hopefully this will address the flakiness we're seeing for some of the tests involving the time.

### DIFF
--- a/test/shared/src/main/scala/zio/test/environment/package.scala
+++ b/test/shared/src/main/scala/zio/test/environment/package.scala
@@ -360,7 +360,7 @@ package object environment extends PlatformSpecific {
       private lazy val awaitSuspended: UIO[Unit] =
         live.provide {
           suspended.repeat {
-            Schedule.doUntilEquals(true) && Schedule.fixed(1.millisecond)
+            Schedule.doUntilEquals(true) && Schedule.fixed(5.milliseconds)
           }
         }.unit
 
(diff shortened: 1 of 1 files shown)


test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -360,7 +360,7 @@ package object environment extends PlatformSpecific {
       private lazy val awaitSuspended: UIO[Unit] =
         live.provide {
           suspended.repeat {
-            Schedule.doUntilEquals(true) && Schedule.fixed(1.millisecond)
+            Schedule.doUntilEquals(true) && Schedule.fixed(5.milliseconds)
```
jdegoes, line 363: Hard to judge if this will work (but master is failing now).

What do we need to solve this for "real"?

adamgfraser: I agree. As far as I can tell the CI failure was unrelated to this so I think given the flakiness is rare (we're only seeing it occasionally even when many of the tests at issue already have `nonFlaky`) I think this has at least a chance to work.

To fix it "for real" I think we need a way to supervise a set of fibers and wait for them all to be done or suspended in a way that is concurrency safe (e.g. it can't be the case that both are suspended but one is "about to be woken up", or we observe them at slightly different times so one is suspended but then starts running again by the time we observe the other). That seems like a hard problem so thought it made sense at least to try the brute force approach.

PR #3825, opened 2020-06-16
inflate and gunzip transducers

### LINKED ISSUE #3879: Add ZTransducer.inflate, ZTransducer.gunzip

LGLO:
Resolves #3879

Ready for review
- [x] docs
- [x] same and better exceptions in inflate and ungzip
- [x] property test for random input compressed with Java, decompressed by this code
- [x] tests for gzipped headers with extra, file name, file comment, with/without CRC16 of header

deflate and gzip are out of scope of this PR, decompression only


simpadjo:
Would be great to have property-based tests:
1) `doSomething(bytes) == doSomethingUsingJavaLib(bytes)`
2) `undoSomethingUsingJavaLib(doSomething(bytes)) == bytes`

LGLO:
There is only decompression implemented in this PR, so I think I could add `decompress(compressWithJava(input)) == input` property tests only, it's equivalent to
`decompressWithJava(compressWithJava(input)) == decompress(compressWithJava(input))`.

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
```diff
@@ -0,0 +1,184 @@
+package zio.stream.compression
+
+import java.util.zip.{ CRC32, DataFormatException, Inflater }
+import java.{ util => ju }
+
+import scala.annotation.tailrec
+
+import zio._
+
+private[compression] class Gunzipper private (var state: Gunzipper.State) {
```
iravid, line 10: I suggest making the functionality in this file mutable and ZIO-less. It'd be much faster this way and you're only doing synchronous effects and error handling, so you can just replace that with exceptions.

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+  def close: UIO[Unit] = state.close
+}
+
+private[compression] object Gunzipper {
+
+  private val fixedHeaderLength = 10
+
+  sealed trait State {
```
iravid, line 25: I love how you organized the steps here. I'm not knowledgeable about the GZIP encoding, so I won't review everything here, but it looks great from the maintenance angle.

Hopefully ZIO Codec will make such parsers easier to write 💪🏻

streams/jvm/src/main/scala/zio/stream/compression/package.scala
```diff
@@ -0,0 +1,89 @@
...
+    def makeInflater(
+      bufferSize: Int
+    ): ZManaged[Any, Nothing, Option[zio.Chunk[Byte]] => ZIO[Any, CompressionException, Chunk[Byte]]] =
+      ZManaged
+        .make(ZIO.effectTotal((new Array[Byte](bufferSize), new Inflater(noWrap)))) {
+          case (_, inflater) => ZIO.effectTotal(inflater.end())
+        }
+        .map {
+          case (buffer, inflater) => {
+            case None => ZIO.succeed(Chunk.empty)
```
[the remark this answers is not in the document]
LGLO: `pullOutput`, which I'll rename to `pullAllOutput` does it, when after reading some part of data it is `.finished()`. What is more if there are multiple delfated parts inside one stream, it would work. 
However it wouldn't work if not whole deflated part was transduced (`Inflater` wouldn't be `.finished()`). Thanks for catching this! I'll add test for this case and fix!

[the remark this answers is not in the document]
LGLO: @iravid actually I don't know if I can test it at ZStream level, should I write the test on level of calling `push`?

iravid:
Yeah that’s fine. There’s an abstraction in the transducer tests that lets you manually drive the transducer.

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
```diff
@@ -0,0 +1,181 @@
...
+  private var state: State = new ParseHeaderStep(Array.emptyByteArray, new CRC32)
+
+  def reset(): Unit = {
+    state.close()
+    state = new ParseHeaderStep(Array.emptyByteArray, new CRC32)
+  }
+
+  def close(): Unit = state.close()
+
+  def onChunk(c: Chunk[Byte]): ZIO[Any, CompressionException, Chunk[Byte]] =
```
iravid, line 25: In terms of code organization, I'd be fine with this just being `: Chunk[Byte]` and wrapping the exceptions into the `CompressionException` in `compression.gunzip`.

streams/jvm/src/main/scala/zio/stream/compression/package.scala
```diff
@@ -0,0 +1,104 @@
...
+      bufferSize: Int
+    ): ZManaged[Any, Nothing, Option[zio.Chunk[Byte]] => ZIO[Any, CompressionException, Chunk[Byte]]] =
+      ZManaged
+        .make(ZIO.effectTotal((new Array[Byte](bufferSize), new Inflater(noWrap)))) {
+          case (_, inflater) => ZIO.effectTotal(inflater.end())
+        }
+        .map {
+          case (buffer, inflater) => {
+            case None =>
+              ZIO.succeed {
```
iravid, line 33: Technically this should be `ZIO.effectTotal` (`ZIO.succeed` is lazy and just forwards to `effectTotal` but it's nicer for maintainability to use `effectTotal`).

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
```diff
@@ -0,0 +1,181 @@
...
+    commentsToSkip: Int
+  ): Gunzipper.State =
+    if (parseExtra) new ParseExtraStep(acc, crc32, checkCrc16, commentsToSkip)
+    else if (commentsToSkip > 0) new SkipCommentsStep(checkCrc16, crc32, commentsToSkip)
+    else if (checkCrc16) new CheckCrc16Step(Array.emptyByteArray, crc32.getValue())
+    else new Decompress()
+
+  private class ParseHeaderStep(acc: Array[Byte], crc32: CRC32) extends State {
+
+    //TODO: If whole input is shorther than fixed header, not output is produced and no error is singaled. Is it ok?
```
iravid, line 49: Sounds like we should throw an exception if the input does not contain a valid header. Is that possible to do?

iravid:
Oh, one last request: can we move the `deflate` and `gunzip` transducer constructors to a platform-specific trait mixed into `ZTransducer`?

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
LGLO: Should I change:
```scala
case None =>
              ZIO.succeed {
                gunzipper.reset()
                Chunk.empty
              }
```
to
```scala
case None =>
              ZIO.succeed {
                if(gunzipper.isFinished()){
                  gunzipper.reset()
                  Chunk.empty
                } else {
                  <fail because stream wasn't complete gzipped>
                }
              }
``` ?

LGLO:
I'll move this to trait and mix in to platform :+1: 
I'd like to add test for valid and invalid headers with various properties

streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala
er1c: I would prefer the error and not the empty chunk :)  Is there an "auto decompress" path that decompresses if the header exists, and leaves the bytes intact if not?

LGLO: If we would like to have such path, then invalid header would have to fallback to clear text stream. Because there is no criterion to distinguish between invalid header and no header.

iravid: @LGLO To your question: Yes, you should add that `else` branch, but you should use `ZIO.effect` instead.

LGLO: Thanks!
I didn't left this work here. Hopefully I could spend some time today in the evening and next week.

iravid: Thank you @LGLO!

LGLO:
Docs are added. Tests against more involved headers are added (and some bugs fixed as well). Removed WIP.

iravid:
@LGLO Fantastic work and coverage. Thank you!!

PR #3828, opened 2020-06-17
Use ChunkBuilder in Chunk.fromIterable

### LINKED ISSUE #3822: Chunk.fromIterable may lose the first element
Since an `Iterable` can be mutable and:
```scala
  /**
   * Returns a chunk backed by an iterable.
   */
  def fromIterable[A](it: Iterable[A]): Chunk[A] =
    it match {
      case chunk: Chunk[A]              => chunk
      case iterable if iterable.isEmpty => Empty
      case vector: Vector[A]            => VectorChunk(vector)
      case iterable =>
        val first                   = iterable.head
        implicit val A: ClassTag[A] = Tags.fromValue(first)
        fromArray(it.toArray)
    }
```

in the last case, `iterable.head` will mutate the original iterable, resulting in a `Chunk` without that head element.

I got bit by this when working with Datastax java driver for Cassandra where the `curentPage()` method in `AsyncResultSet` returns a mutable java iterable.

Example:

```scala
import scala.jdk.CollectionConverters.IterableHasAsScala

val session: CqlSession = ???
val statement: String   = ???

ZIO
  .fromCompletionStage(session.executeAsync(statement))
  .map(asyncResultSet => Chunk.fromIterable(asyncResultSet.currentPage().asScala))
```

palanga:
Closes #3822

### DIFF
--- a/benchmarks/src/main/scala/zio/chunks/ChunkFromIterableBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkFromIterableBenchmarks.scala
@@ -1,0 +1,49 @@
+package zio.chunks
+
+import java.util.concurrent.TimeUnit
+
+import scala.reflect.ClassTag
+
+import org.openjdk.jmh.annotations._
+
+import zio.Chunk
+import zio.Chunk.{ fromArray, Tags }
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class ChunkFromIterableBenchmarks {
+
+  var array: Array[Int] = _
+  var list: List[Int]   = _
+
+  @Param(Array("1000"))
+  var size: Int = _
+
+  @Setup(Level.Trial)
+  def setup() = {
+    array = (1 to size).toArray
+    list = array.toList
+  }
+
+  @Benchmark
+  def fromIterableArray(): Chunk[Int] = Chunk.fromIterable(array)
+
+  @Benchmark
+  def fromIterableArrayUnsafe(): Chunk[Int] = fromIterableUnsafe(array)
+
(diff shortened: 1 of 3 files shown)


palanga:
Benchmark later

adamgfraser:
@palanga Looks great! Let me know when this is ready.

palanga:
Sorry for the delay guys. I'm on vacation.
These are my results:

```
[info] Benchmark                                            (size)  Mode  Cnt      Score      Error  Units
[info] ChunkFromIterableBenchmarks.fromIterableArray          1000  avgt    3  11868,502 ± 1803,243  ns/op
[info] ChunkFromIterableBenchmarks.fromIterableArrayUnsafe    1000  avgt    3     70,446 ±    6,668  ns/op
[info] ChunkFromIterableBenchmarks.fromIterableList           1000  avgt    3   6817,559 ±  561,745  ns/op
[info] ChunkFromIterableBenchmarks.fromIterableListUnsafe     1000  avgt    3  20033,511 ± 7449,947  ns/op
```

* Building a Chunk from an array is way worst (no surprises) but we have `fromArray` for that case. Is that ok ?
* Building a Chunk from a list is faster !

Let me know what you think.

adamgfraser:
@palanga I think we want to avoid gotchas like that if we can. Why don't we just add another case to match on the collection and determine if it is a `WrappedArray`. If so we can just wrap the array up in a chunk and return it immediately. Separately we can also probably make `ChunkBuilder` smarter about this but it is not going to be as fast as just returning the array immediately because the `ChunkBuilder` still has to support potentially adding more elements.

jdegoes:
@palanga Can you add another type case to check to see if it's an `Array`, and use `fromArray` in that case rather than defaulting to the chunk builder? That would be awesome and then we can get this PR merged in.

I will close for now but please re-open if you get a chance to work on this!

PR #3829, opened 2020-06-17
Scopes5

jdegoes:


AdamPaynter:
Just out of curiosity, what was the goal of this effort? I see it has *something* to do with forking and supervision.

jdegoes:
@AdamPaynter 

To make this work:

```scala
left.fork zipPar right
```

without auto-supervision immediately killing the fiber executing `left`.

phderome:
> @AdamPaynter
> 
> To make this work:
> 
> ```scala
> left.fork zipPar right
> ```
> 
> without auto-supervision immediately killing the fiber executing `left`.

discussed in Zionomicon book chapter 8.

PR #3830, opened 2020-06-18
add someOrElseM method to ZIO, ZManaged and ZSTM

vladimirkl:
This PR adds new `someOrElseM` method  to `ZIO`, `ZManaged` and `ZSTM` - similar to `someOrElse` but with effect as default. I found this helper quite useful when working with `some` / `asSomeError` / `optional` combinators. Of course this can be achieved  with `flatMap` doing pattern matching or folding, but new helper looks cleaner.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2881,6 +2881,17 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(ZIO.fail(ExampleError).someOrElse(42).run)(fails(equalTo(ExampleError)))
       } @@ zioTag(errors)
     ),
+    suite("someOrElseM")(
+      testM("extracts the value from Some") {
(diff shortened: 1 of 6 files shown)


ghostdogpr:
Can't you do `.some orElse` already?

vladimirkl:
Unfortunately `some.orElse` won't help because it runs `orElse` effect either on `None` value or  error. `someOrElseM` is similar to `someOrElse` - so it runs default effect only is value is `None` and doesn't hide error.

For example, given the following effects:
` def f1: IO[String, Option[String]] = IO.fail("err1")`
`def f2: IO[String, String]  = IO.succeed("val1")`

`f1.some.orElse(f2)` will return `"val1"` but `f1.someOrElseM(f2)` will fail with `"err1"`.

ghostdogpr:
You’re right 👍

vladimirkl:
@ghostdogpr Would you mind to approve this PR? Some CI checks still fail but this is not related to PR.

ghostdogpr:
Can you rebase to the latest master? I believe the flaky tests have been fixed. Thanks!

PR #3844, opened 2020-06-21
Spelling

jsoref:
Misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

It reported: https://github.com/jsoref/zio/commit/6cb11ce67bfb7a62246107513e604d09a254b090#commitcomment-40056494

And it validated that the changes in this PR made it happy: https://github.com/jsoref/zio/commit/5fb861bab27ded23721bd630cd566bbb5b7c401e

Note: this PR does not include the action, if you're interested, that can be offered separately.


adamgfraser:
@jsoref Fantastic! Can you merge upstream changes? That should fix the test failures.

simpadjo:
Nice improvement! But the tool wrongly corrects Wiem's surname. Could you exclude it from the analysis?

jsoref:
Are you sure? It's based on:
> Wiem Zine El Abidine (@wi101)
from the code of conduct.

simpadjo:
Hm, different sources spell differently: e.g. https://www.linkedin.com/in/wiem-zine-el-abidine/?originalSubdomain=de , https://www.signifytechnology.com/blog/2020/06/zio-from-home-by-wiem-zine-elabidine-scala-in-the-city-conference .

jsoref:
This is more or less the logic I went through that got me here:
* The conference item is most likely to have a spelling error.
* Linkedin probably has a 3 word limit which is going to force some compaction. Given the way these words compact, merging `el` with the last thing is probably traditional.
* GitHub is freeform which should allow someone to express their preferred name.
* Given that we're in GitHub, it seemed favoring that spelling was reasonable.
* Especially given that this spelling is in one of the files which seems most canonical, it seemed like the right choice.

They are active this month, so if you want to wait for them to confirm, I'm fine with that.

To be clear: the tool only identifies misspellings, it doesn't make suggestions for fixing them. Finding a correction is left as an exercise to the person operating it. And of course, it's important to confirm that the corrections are correct. In this case, solving this one item was by far the longest leg in generating this PR. -- Sometimes I call out my thoughts in my PR, but this was done late one night and I was too tired.

To the original point: it's totally possible to exclude a given name (or a bunch of variants) from analysis. And I can if that's really desirable. I suspect that's the wrong choice here. Assuming this is a single person, it's usually best to use their current preferred name throughout, except in specific cases where there's an absolute need to reference a past name.

adamgfraser:
@jsoref Can you merge upstream changes? Then let's get this in.

jsoref:
If you like, I'd be happy to make a PR for the action so that you could try to keep spelling errors out...

PR #3861, opened 2020-06-23
Implement Weak Map

### LINKED ISSUE #3851: Threads getting blocked with RC21
While upgrading ZIO Query to RC21, we observed some tests started getting timeouts. Using snapshots we confirmed this started to appear from the Scopes PR.

Here's a view of the threads during one of the failing tests, before Scopes:
Screen_Shot_2020-06-21_at_3 09 36_PM
After Scopes:
Screen_Shot_2020-06-21_at_3 09 45_PM

We can see the "blocked" time is spent in the following function:
Screen_Shot_2020-06-21_at_3 12 55_PM

On top of that, I put a breakpoint in `unsafeEnsure` to see what was going on and the finalizer list had a size over a million, after a few seconds running the test.

It can be reproduced by using [this branch](https://github.com/zio/zio-query/pull/65) and running "zipPar does not prevent batching" (use `nonFlaky(10000)` to make the test run long enough).

adamgfraser:
Work to resolve #3851.

Implements a weak concurrent map to track finalizers without blocking while still allowing them to be garbage collected. I think this needs a little more work to prevent a potential race condition between adding a finalizer / releasing the finalizers and we should be able to remove synchronization in more places but wanted to get something out there.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,27 +45,27 @@ addCommandAlias(
 addCommandAlias("compileNative", ";coreNative/compile")
 addCommandAlias(
   "testJVM",
-  ";coreTestsJVM/test;stacktracerJVM/test;streamsTestsJVM/test;testTestsJVM/test;testMagnoliaTestsJVM/test;testRunnerJVM/test:run;examplesJVM/test:compile;benchmarks/test:compile;macrosJVM/test"
+  ";coreTestsJVM/testOnly zio.query.ZQuerySpec"
 )
 addCommandAlias(
   "testJVMNoBenchmarks",
-  ";coreTestsJVM/test;stacktracerJVM/test;streamsTestsJVM/test;testTestsJVM/test;testMagnoliaTestsJVM/test;testRunnerJVM/test:run;examplesJVM/test:compile"
+  ";coreTestsJVM/testOnly zio.query.ZQuerySpec"
 )
 addCommandAlias(
   "testJVMDotty",
-  ";coreTestsJVM/test;stacktracerJVM/test:compile;streamsTestsJVM/test;testTestsJVM/test;testRunnerJVM/test:run;examplesJVM/test:compile"
+  ";coreTestsJVM/testOnly zio.query.ZQuerySpec"
 )
 addCommandAlias(
   "testJVM211",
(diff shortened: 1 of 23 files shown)


core/jvm/src/main/scala/zio/internal/WeakMap.scala
```diff
@@ -0,0 +1,242 @@
...
+   * equal to the hashcode of the key.
+   */
+  private final class WeakKey[K](key: K, queue: ReferenceQueue[_ >: K]) extends WeakReference[K](key, queue) { self =>
+    override def equals(that: Any): Boolean =
+      that match {
+        case weakKey: WeakKey[_] => self.get == weakKey.get
+        case _                   => false
+      }
+    override val hashCode: Int =
+      key.hashCode
```
darl, line 190: will it hold a strong reference to `key`?

adamgfraser: No because we eagerly compute the hash code when the `WeakKey` is created so we don't capture a reference after that. We need that hash code so we can look up the entry and remove it from the `ConcurrentHashMap` later in `garbageCollect()`. You can see if working if you run an example like [this](https://github.com/zio/zio/issues/3851#issuecomment-647207026) and add a debug `println(finalizers(mode).size)` statement in `unsafeEnsure` in `ZScope`. I will see if I can add a test for this.

build.sbt
```diff
@@ -45,27 +45,27 @@ addCommandAlias(
 addCommandAlias("compileNative", ";coreNative/compile")
 addCommandAlias(
   "testJVM",
-  ";coreTestsJVM/test;stacktracerJVM/test;streamsTestsJVM/test;testTestsJVM/test;testMagnoliaTestsJVM/test;testRunnerJVM/test:run;examplesJVM/test:compile;benchmarks/test:compile;macrosJVM/test"
+  ";coreTestsJVM/testOnly zio.query.ZQuerySpec"
```
mijicd, line 48: What happened here? :worried:

adamgfraser:
Superseded by #3884. There may still be some opportunities for optimization with the finalizer map but I think it makes sense to pursue as part of separate PR once we have more feedback on current version.

PR #3866, opened 2020-06-24
Create Child Scope in Transplant

adamgfraser:
Explores an alternative approach to resolving #3851. The problem there is basically that when we transplant in zipPar we are moving everything up to the parent scope in `zipPar` so when we build very large parallel programs we are creating this one piece of shared state that is accessed by all the fibers. Instead we can create a new child scope in transplant, attach the child scope to the parent scope, and graft the forked fibers in zipPar onto that. This builds a tree like structure of scopes instead of one giant scope. At least for the ZQuery use case this seems to perform very well and it passes all of our tests. However, it seems to create a memory leak with infinite zipPar processes like the ones in the issue. If we could resolve that it seems like this could be a good way to go.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,9 +18,6 @@ jdk_8: &jdk_8
 jdk_11: &jdk_11
   JDK_VERSION: 11
 
-machine_resource: &machine_resource
-  resource_class: large
-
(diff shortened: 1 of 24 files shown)


adamgfraser:
Superseded by #3884.

PR #3867, opened 2020-06-25
Add toReader for ZStream

### LINKED ISSUE #3873: Add ZStream.toReader
Interop with java.io.Reader

er1c:
Resolves #3873

Is there a good way to create tests for this?

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -1,6 +1,7 @@
 package zio.stream
 
 import java.io.ByteArrayInputStream
+import java.io.IOException
 import java.util.concurrent.TimeUnit
 
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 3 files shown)


simpadjo:
Good addition!
I see that implementation is inspired by `.toIterator`. Maybe it is possible to just wrap `Iterator[Char]` into a `Reader` and piggy-back `.toIterator`?

Regarding tests, you can take a look at how `ZStream.toIterator` is tested.

er1c:
@simpadjo took the implementation idea from `ZInputStream` I'll look into the toIterator tests, thanks!

semi-related, what's the special sbt foo to run the code formatter check locally?

simpadjo:
Yes, probably `Reader` has more capabilities than Iterator, so makes sense to implement it like that.
Then you could look at `ZStream.toInputStream` tests: https://github.com/zio/zio/blob/master/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala#L2706

simpadjo:
The best way to format the code is `sbt prepare` (see Contributors Guide).

er1c:
Just found the `toInputStream` tests: https://github.com/zio/zio/blob/b5f18309a7883144b44d09f69a6d8a4a63c827df/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala#L2706 adding it here for my reference.

er1c:
Tests added

PR #3881, opened 2020-06-26
Interrupt tasks on failure in ZStream#mapMPar

iravid:


adamgfraser:
@iravid Seeing some flakiness in the test for this [here](https://app.circleci.com/pipelines/github/zio/zio/8130/workflows/55a401b6-80ae-4b30-8411-9a2ee01050a7/jobs/82222/steps). This is on the branch that removes finalizers when a fiber terminates so it could be related to that but since we haven't observed any other issues outside of Scala.js right now my hypothesis would be the issue is with this.

PR #3884, opened 2020-06-26
Remove Finalizer when Fibers Terminate

adamgfraser:
After further investigation, it appears that the issue we are having with ZIO Query relates to excessive memory consumption. That is why the performance is so much worse in CI than it is locally and why changing the machine in CI from `medium` to `large` has such a performance impact. It appears that although the finalizers can be garbage collected the fact that we are creating such a large data structure with all of these finalizers (~1M in this case) and then have to copy them to an array and sort them as part of releasing them creates excessive memory pressure.

This PR removes the finalizer from the parent scope when the fiber terminates. This seems to dramatically reduce the memory consumption. There may be further optimizations to do to avoid locks but this seems to address the immediate issue.

### DIFF
--- a/core/shared/src/main/scala/zio/ZScope.scala
+++ b/core/shared/src/main/scala/zio/ZScope.scala
@@ -45,7 +45,7 @@ sealed trait ZScope[+A] { self =>
    * is closed. The returned effect will succeed with `true` if the finalizer
    * will not be run by this scope, and `false` otherwise.
    */
-  def deny(key: => ZScope.Key): UIO[Boolean]
+  def deny(key: => ZScope.Key): UIO[Boolean] = UIO(unsafeDeny(key))
 
   /**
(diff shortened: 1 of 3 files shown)


iravid:
Definitely worth doing RC21-1 with this yeah?

adamgfraser:
Yes. Have we merged any PRs with binary incompatible changes since RC21?

iravid:
No, I just went over the changes and they're all binaryu compatible.

adamgfraser:
Yes. Just did the same. Hopefully this will resolve the issues we have been seeing with flakiness on Scala.js in CI. Let's get this in, see if there are any other outstanding PRs we want to get merged, and then get another release out.

jdegoes:
@adamgfraser Awesome work on this! 🙏 🎉

PR #3885, opened 2020-06-26
Implement ZTransducer.foldUntil using ZTransducer.fold

### LINKED ISSUE #3860: ZStream.groupedWithin doesn't groups batch when got exact items count
Hello!

I've met strange behaviour of the `ZStream.groupedWithin` which can significantly affect performance of an application when it's passing batches of messages of fixed size. Here is the reproducer (works both on RC-20, RC-21 and SNAPSHOT):

```scala
import zio._
import zio.duration._
import zio.stream.ZStream

object TestApp extends App {

  def run(args: List[String]): ZIO[ZEnv, Nothing, ExitCode] =
    for {
      queue <- Queue.unbounded[Int]
      stream = ZStream.fromQueue(queue)
      _ <- stream
        .tap(i => ZIO.effectTotal(println(java.time.LocalDateTime.now() + " got: " + i)))
        .groupedWithin(4, 10.seconds)
        .mapM(batch =>
          ZIO
            .effectTotal(
              println(java.time.LocalDateTime.now() + s" got batch [${batch.size}]: " + batch.mkString("[", ", ", "]"))
            )
            .as(true)
        )
        .runDrain
        .fork

      _ <- ZIO.foreach(1 to 4)(i => queue.offer(i))
      _ <- ZIO.sleep(15.second)

    } yield ExitCode.success
}
```

It produces the following output:
```
2020-06-23T21:57:59.711925 got: 1
2020-06-23T21:57:59.866961 got: 2
2020-06-23T21:58:00.107421 got: 3
2020-06-23T21:58:00.188661 got: 4
2020-06-23T21:58:10.422722 got batch [4]: [1, 2, 3, 4]
```

One can see that we are able to compose a full batch after a millis from start, but actually we got batch only at the end of a group window (after 10 seconds).

Is it a bug or an expected behaviour?

Thank you!

iravid:
The previous implementation used foldWeighted, which prevented
transducers based on it (like collectAllN) from emitting immediately
when N elements were accumulated.

Resolves #3860.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -74,6 +74,11 @@ object ZTransducerSpec extends ZIOBaseSpec {
         },
         testM("doesn't emit empty trailing chunks") {
           assertM(run(ZTransducer.collectAllN[Int](3), List(Chunk(1, 2, 3))))(equalTo(Chunk(List(1, 2, 3))))
+        },
+        testM("emits chunks when exactly N elements received") {
+          ZTransducer.collectAllN[Int](4).push.use { push =>
+            push(Some(Chunk(1, 2, 3, 4))).map(result => assert(result)(equalTo(Chunk(List(1, 2, 3, 4)))))
+          }
         }
       ),
       suite("collectAllToMapN")(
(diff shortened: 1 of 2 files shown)


iravid:
@simpadjo appreciate your review here

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
```diff
@@ -74,6 +74,11 @@ object ZTransducerSpec extends ZIOBaseSpec {
         },
         testM("doesn't emit empty trailing chunks") {
           assertM(run(ZTransducer.collectAllN[Int](3), List(Chunk(1, 2, 3))))(equalTo(Chunk(List(1, 2, 3))))
+        },
+        testM("emits chunks when exactly N elements received") {
```
simpadjo, line 78: The test could be blackboxified: something like
`(ZStream(1,2,3,4) ++ ZStream.fromEffect(ZIO.never)).transduce(t).take(1)`

iravid: Thanks! I actually prefer to test the transducer directly in this case as it is a bit of an edge case.

PR #3886, opened 2020-06-26
Replace List with Chunk in ZTransducer.collectAllN

svranesevic:
Replace usage of `List` in favor of a more performant `Chunk` throughout ZTransducer.collectAllN


simpadjo:
Yes, it's very tempting to get rid of lists. Unfortunately, `Chunk :+ ` is not as fast. 
You can get around it by using `ChunkBuilder` or even `ArrayBuilder` as an accumulator.

svranesevic:
Thought about going the `ChunkBuilder` route, nice addition there would be giving the builder a size hint in `collectAllN`.

iravid:
Hey @svranesevic! Thank you for following up on this.

A ChunkBuilder would be best here indeed. Appending elements to a chunk is fast, but not as fast as building with a ChunkBuilder.

svranesevic:
@iravid @simpadjo Thank you for the feedback, on it!

iravid:
Thanks @svranesevic, this revision looks great. The only thing we need to do is make sure the creation of the `ChunkBuilder` is suspended because it is mutable. You can see how that is done in `ZSink.collectAll` for an example.

svranesevic:
Feedback needed whether https://github.com/zio/zio/pull/3886/commits/beca969e6e9c62bf3dcc14ce1c9670ccfa1bed50 is good enough or I should rewrite ZTransducer.{collectAllWhile|collectAllWhileM} along the lines of https://github.com/zio/zio/pull/3886/commits/161616e417f444d561a1cbe173d25992056a4351

svranesevic:
> Feedback needed whether [beca969](https://github.com/zio/zio/commit/beca969e6e9c62bf3dcc14ce1c9670ccfa1bed50) is good enough or I should rewrite ZTransducer.{collectAllWhile|collectAllWhileM} along the lines of [161616e](https://github.com/zio/zio/commit/161616e417f444d561a1cbe173d25992056a4351)

Follow-up, benchmark wise https://github.com/zio/zio/pull/3886/commits/109b4596adda714c2d3ba4bf36efe37e66e6cd54 looks like  the way to go.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -124,10 +124,38 @@ object ZTransducer {
...
   /**
-   * Creates a transducer accumulating incoming values into lists of maximum size `n`.
+   * Creates a transducer accumulating incoming values into chunks of maximum size `n`.
    */
-  def collectAllN[I](n: Long): ZTransducer[Any, Nothing, I, List[I]] =
-    foldUntil[I, List[I]](Nil, n)((list, element) => element :: list).map(_.reverse).filter(_.nonEmpty)
+  def collectAllN[I](n: Long): ZTransducer[Any, Nothing, I, Chunk[I]] =
+    ZTransducer {
+
+      def go(in: Chunk[I], builder: ChunkBuilder[I], size: Long): (ChunkBuilder[Chunk[I]], ChunkBuilder[I], Long) =
```
simpadjo, line 132: You can append `in` to `builder` all at once instead of appending elements one by one. It is faster and would save some lines of code as well

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -124,10 +124,38 @@ object ZTransducer {
...
+            stateRef
+              .getAndSet((ChunkBuilder.make(n.toInt), 0))
+              .map {
+                case (chunkBuilder, chunkSize) =>
+                  if (chunkSize == 0) Chunk.empty
+                  else Chunk(chunkBuilder.result())
+              }
+
+          case Some(in) =>
+            stateRef.modify {
```
simpadjo, line 151: `ChunkBuilder` could be reused after emitting a chunk, if I'm not wrong

simpadjo:
@svranesevic I have couple of ideas how to simplify implementation a bit. Let me explore it today or tomorrow.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
svranesevic: It could be reused and I almost went for it, but after giving it a thought I'd argue that it's reusability is implementation details as ChunkBuilder is not [`ReusableBuilder`](https://www.scala-lang.org/api/current/scala/collection/mutable/ReusableBuilder.html) as of right now.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
svranesevic: Good idea, I was wrapping my mind around transducer internals while writing this :sweat_smile:
In that approach we need to handle case when appending the `in` would produce Chunk that has more elements than `N` by splitting it accordingly and carrying over leftover. I will check `Chunk#splitAt` but afaik it should be better than current impl.

svranesevic:
> @svranesevic I have couple of ideas how to simplify implementation a bit. Let me explore it today or tomorrow.

Sure thing, looking forward to it!

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
simpadjo: `ChunkBuilder` has a public  `clear` method, so yes

simpadjo:
@svranesevic how about finalizing `collectAllN` and then rewriting other transducers in a separate PR?
They also need some polishing so let's keep the PR scope reasonable. Optimizing `collectAllN` and introducing `sizeHint` is already a good addition!

Regarding `collectAllN`, I suggest 
- not reallocating `ChunkBuilder`, use just one and `clean` it
- update `ChunkBuilder` in a bulk way

svranesevic:
> @svranesevic how about finalizing `collectAllN` and then rewriting other transducers in a separate PR?
> They also need some polishing so let's keep the PR scope reasonable. Optimizing `collectAllN` and introducing `sizeHint` is already a good addition!
> 
> Regarding `collectAllN`, I suggest
> 
> * not reallocating `ChunkBuilder`, use just one and `clean` it
> * update `ChunkBuilder` in a bulk way

I agree with scoping down the PR to `collectAllN` :+1: 
Regarding the suggestions:
 - I hold to my disagreement on reusing the ChunkBuilder as it is [`Builder`](https://www.scala-lang.org/api/current/scala/collection/mutable/Builder.html) for which, by the abstraction, behavior is undefined after calling the `Builder#result`. Reason being that we should rely on abstractions not on implementation details, so here we either bump up `ChunkBuilder` to be [`ReusableBuilder`](https://www.scala-lang.org/api/current/scala/collection/mutable/ReusableBuilder.html) or @iravid and/or more folks hop in and vote for the reusing the ChunkBuilder as it is.
- "update `ChunkBuilder` in a bulk way" - :100:

simpadjo:
I guess the main reason why `ChunkBuilder` is not a `ReusableBuilder` is that there are not many people who even heard about this trait :)

iravid:
Admittedly reusing the ChunkBuilder is probably unsafe with streams of `Any` due to the use of classtags. But that's probably not a valid usecase.

In any case, let's scope this down to `collectAllN` like @simpadjo suggested, keep the implementation without reusing the builder and switch to bulk additions to the builder. Then we can merge 💪

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -124,10 +124,41 @@ object ZTransducer {
     }
 
   /**
-   * Creates a transducer accumulating incoming values into lists of maximum size `n`.
+   * Creates a transducer accumulating incoming values into chunks of maximum size `n`.
    */
-  def collectAllN[I](n: Long): ZTransducer[Any, Nothing, I, List[I]] =
-    foldUntil[I, List[I]](Nil, n)((list, element) => element :: list).map(_.reverse).filter(_.nonEmpty)
+  def collectAllN[I](n: Int): ZTransducer[Any, Nothing, I, Chunk[I]] =
```
svranesevic, line 129: Implementation detail that bubbled up due to the Chunk's size being `Int`, might be a deal breaker for this change?

Affects https://github.com/zio/zio/pull/3886/files#diff-e2c16046e05624c39e46b9e4c2870b9cR1596 and https://github.com/zio/zio/pull/3886/files#diff-e2c16046e05624c39e46b9e4c2870b9cR1603.

iravid: This is a fair compromise, `n` is fine as an `Int` here.

streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
```diff
@@ -67,17 +67,19 @@ object ZTransducerSpec extends ZIOBaseSpec {
     suite("Constructors")(
       suite("collectAllN")(
         testM("happy path") {
-          assertM(run(ZTransducer.collectAllN[Int](3), List(Chunk(1, 2, 3, 4))))(equalTo(Chunk(List(1, 2, 3), List(4))))
+          assertM(run(ZTransducer.collectAllN[Int](3), List(Chunk(1, 2, 3, 4))))(
```
iravid, line 70: @svranesevic Last request and then this is good to merge: the implementation is a bit more involved now, so can we add a property check? Just equivalence with `List#grouped` should be sufficient.

iravid:
Hey @svranesevic, need any help getting this over the finish line?

svranesevic:
Hey @iravid, haven't had time lately to wrap this up, sorry about it, will do today!

svranesevic:
Ping @iravid

iravid:
Looks good @svranesevic! Thanks!

PR #3891, opened 2020-06-27
Align ZStream.repeatEffectWith with ZIO#repeat

### LINKED ISSUE #3863: ZStream.tick does not run the action first before waiting the interval
`ZStream.tick(duration)` relies on `ZStream.spaced(duration)`, which is defined like this in the scaladoc :
`|action|-----interval-----|action|-----interval-----|action|`

But it seems that it is rather behaving like this  : `-----interval-----|action|-----interval-----|action|`

Example : 

```scala
object TestTick extends App {

  val tick = ZStream.tick(1.second).map(_ => "tick")
  val tack = ZStream.tick(10.second).map(_ => "tack")
  zio.Runtime.default.unsafeRun(
    tick.merge(tack).foreach(t => putStrLn(t))
  )
}
```

If you run this, you will not see "tack" directly but only after 10 seconds : 

```
tick
tick
tick
tick
tick
tick
tick
tick
tick
tack <-
```

iravid:
Repetitions should be performed in addition to the first time.

Resolves #3863.


svroonland:
I just noticed this PR, it's probably good to consider this together with #3959.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3697,11 +3697,12 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
    * Creates a stream from an effect producing a value of type `A` which repeats using the specified schedule
    */
   def repeatEffectWith[R, E, A](effect: ZIO[R, E, A], schedule: Schedule[R, A, _]): ZStream[R, E, A] =
-    fromEffect(schedule.initial).flatMap { initial =>
-      unfoldM(initial) { state =>
-        effect.flatMap(value => schedule.update(value, state).fold(_ => None, state => Some((value, state))))
+    ZStream.fromEffect(effect) ++
+      fromEffect(schedule.initial).flatMap { initial =>
```
svroonland, line 3701: Doesn't this do `effect -> effect -> delay` instead of `effect -> delay -> effect` ?

iravid:
Superseded by #3959.

Issue #3882, opened 2020-06-26
OOM using bufferSliding


jsfwa:
**ZIO version**: 1.0.0-RC21

Hi guys, simple reproducible OOM via `bufferSliding`:
```scala
 ZStream
      .repeat(Foo(1L))
      .bufferSliding(1)
      .runDrain
      .exitCode
```
<img width="1244" alt="image" src="https://user-images.githubusercontent.com/7480075/85884081-b4192b00-b7ea-11ea-9f1e-fe7958d2a939.png">
<img width="1242" alt="image" src="https://user-images.githubusercontent.com/7480075/85884219-f2164f00-b7ea-11ea-963d-a7d0499abc55.png">

iravid:
Thank you for reporting (and minimizing!) @jsfwa. We'll look into this.

jsfwa:
Just checked, the problem is in this recursive call https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L434
changed it to 
```
Take.fromPull(as).tap(take => queue.offer(take)).doWhile(_ != Take.end).unit
```
And looks like it solved the problem, but not sure about the correctness of this approach

iravid:
@jsfwa Interesting. Would you like to open a PR with the fix and we can verify it there?

narma:
Hey! I'm @jsfwa colleague, and he asked me to create PR because he away from the computer for a weekend.

PR #3892, opened 2020-06-27
Fix OOM at bufferSliding

narma:
Fix https://github.com/zio/zio/issues/3882


narma:
I think to make separate PR with memory leaks test, @iravid what do you think? Also question, they should be part of all tests or placed separately like benchmarks?

narma:
With that fix
jvm args: `-Xmx2048M -Xms512M -XX:+UseG1GC `
program:
```scala
 ZStream
      .repeat(Foo(1L))
      .bufferSliding(1)
      .runDrain
      .exitCode
```

изображение

Issue #3880, opened 2020-06-26
ZManaged integration improvements


jdegoes:
These could be added in the JVM (platform-specific) trait that is mixed into the `ZManaged` companion object:

- [x] `ZManaged.readFile(path: Path): ZManaged[Blocking, IOException, FileInputStream]`
- [x] `ZManaged.readFile(path: String): ZManaged[Blocking, IOException, FileInputStream]`
- [x] `ZManaged.writeFile(path: Path): ZManaged[Blocking, IOException, FileOutputStream]`
- [x] `ZManaged.writeFile(path: String): ZManaged[Blocking, IOException, FileOutputStream]`
- [x] `ZManaged.readURL(url: URL): ZManaged[Blocking, IOException, InputStream]`
- [x] `ZManaged.readURL(url: String): ZManaged[Blocking, IOException, InputStream]`
- [x] `ZManaged.readURI(uri: URI): ZManaged[Blocking, IOException, InputStream]`

```scala
final case class FileInputStream private ( private ... ) {
  def readN(n: Int): ZIO[Blocking, IOException, Option[Chunk[Byte]]]
  def readAll: ZIO[Blocking, IOException, Option[Chunk[Byte]]]
  def seekTo(n: Long): ZIO[Blocking, IOException, Unit] // look / copy FIS
}
final case class InputStream private ( private ... ) {
  def readN(n: Int): ZIO[Blocking, IOException, Option[Chunk[Byte]]]
  def readAll: ZIO[Blocking, IOException, Option[Chunk[Byte]]]
}
final case class OutputStream private ( private ... ) {
  def write(chunk: Chunk[Byte]): ZIO[Blocking, IOException, Unit]
}
```

/cc @adrianfilip

adrianfilip:
Thanks! I'll implement this!

iravid:
Resolved in #3901.

PR #3901, opened 2020-06-28
3880 z managed integration improvements

adrianfilip:
Hi @jdegoes,

This is for #3880.
Let me know if I anything should be changed!


iravid:
Hi @adrianfilip, this looks pretty cool and a good addition. I have a few suggestions:
1. The case classes can be placed outside of the platform-specific trait. Only the constructors need to be there.
2. It'd be nice if these were not case classes but rather traits or abstract classes, such that we can later implement them using a backing `ZStream` in `ZStream#toInputStream` or a future `ZSink#toOutputStream`.

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  def readAll: ZIO[Blocking, IOException, Option[Chunk[Byte]]] =
+    effectBlocking(is.available())
+      .flatMap(available =>
+        effectBlocking {
+          available match {
+            case 0 => None
+            case _ => {
+              val buffer = new java.io.ByteArrayOutputStream();
+              val data   = new Array[Byte](4096);
```
iravid, line 65: The chunk size here should be configurable

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
```diff
@@ -0,0 +1,132 @@
...
+private[zio] trait ZOutputStream {
+  def write(chunk: Chunk[Byte]): ZIO[Blocking, IOException, Unit]
+  def close(): ZIO[Blocking, Nothing, Unit]
+}
+
+/**
+ * A functional wrapper over a java.io.InputStream.
+ */
+private[zio] case class InputStream private (private val is: java.io.InputStream) extends ZInputStream {
+  def readN(n: Int): ZIO[Blocking, IOException, Option[Chunk[Byte]]] =
```
iravid, line 42: With this signature, there's no way for the caller to infer if the input stream is exhausted. How about encoding this as `ZIO[Blocking, Option[IOException], Chunk[Byte]]`, where `Some(IOException)` is an actual error, `None` is an end-of-stream signal and a value is a chunk?

I would also suggest to make the non-blocking read optional (using a flag). If the flag is false, don't check `available` before reading and just let the underlying input stream manage that.

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
```diff
@@ -0,0 +1,132 @@
...
+          val b: Array[Byte] = new Array[Byte](n)
+          is.read(b)
+          Some(Chunk.fromArray(b))
+      }
+    }.refineToOrDie[IOException]
+
+  def skip(n: Long): ZIO[Blocking, IOException, Long] =
+    effectBlocking(is.skip(n)).refineToOrDie[IOException]
+
+  def readAll: ZIO[Blocking, IOException, Option[Chunk[Byte]]] =
```
iravid, line 57: This can be `ZIO[Blocking, IOException, Chunk[Byte]]` and the blocking can be configured through a flag.

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/**
+ * A functional wrapper over a java.io.OutputStream.
+ */
+private[zio] case class OutputStream private (private val os: java.io.OutputStream) extends ZOutputStream {
+  def write(chunk: Chunk[Byte]): ZIO[Blocking, IOException, Unit] =
+    effectBlocking {
+      os.write(chunk.toArray)
+      os.flush()
```
iravid, line 90: Should we always flush?

adrianfilip:
@iravid Sounds great. I'll make the changes.

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
adrianfilip: Great suggestion. I think we should remove the available check alltogether if we use the encoding you proposed. And have each implementation decide how it wants to deal with readN on an exhausted stream. I think it makes things simpler.

 What do you think?

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
adrianfilip: I understand. Looking at ZIO[Blocking, Option[IOException], Chunk[Byte]] from readN I think this place would be a good one to use the same encoding and also drop the available check alltogether (we already get the info from the read return code).

core/jvm/src/main/scala/zio/ZManagedPlatformSpecific.scala
adrianfilip: Good point. No, we don't.  I was thinking about treating write + flush as "atomic" in a write, but it doesn't have to be and it would actually reduce performance (because of all that IO). I will remove the flush.

core-tests/jvm/src/test/scala/zio/ZManagedPlatformSpecificSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+import zio.test._
+import java.io.IOException
+import java.io.File
+import java.{ util => ju }
+import java.nio.file.Files
+import zio.test.Assertion._
+
+object ZManagedPlatformSpecificSpec extends ZIOBaseSpec {
+
+  val target = new ZManagedPlatformSpecific {}
```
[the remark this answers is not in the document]
adrianfilip: I can't access it like you mention.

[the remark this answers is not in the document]
iravid: Oh right. We need to modify the ZManaged companion object to extend the ZManagedPlatformSpecific trait.

adrianfilip:
@iravid I implemented the changes based on your suggestions. I updated my approach as a reply to your comments.
Let me know if this is ok and also if the new ZInputStream and ZOutputStream files should be moved elsewhere.

(The failing tests are unrelated to the PR (i raised a ticket for that https://github.com/zio/zio/issues/3908).)

adrianfilip:
@jdegoes @iravid I think it's good to go now. Let me know if you think I should change anything.

PR #3902, opened 2020-06-28
Add ZStream.retry

### LINKED ISSUE #3785: Add ZStream#retry to retry the stream
There is `ZIO#retry`, we should have an equivalent for `ZStream`. 

TBD: semantics, retry the entire stream or just the last pull?

svroonland:
Closes #3785


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2290,6 +2290,44 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+          self
+            .catchAll(e =>
+              ZStream.unwrap {
+                (for {
+                  s        <- state.get
+                  newState <- schedule.update(e, s)
+                } yield newState).fold(
+                  _ => ZStream.fail(e), // Failure of the schedule indicates it doesn't accept the input
+                  newState =>
+                    ZStream.fromEffect(state.set(newState)) *> go.mapChunksM { chunk =>
```
iravid, line 2319: This will unfortunately leak memory on streams that repeat indefinitely (try `ZStream.fail("Boom").retry(Schedule.forever)`).

To be heap-safe, this needs to be rewritten like the formulation of `ZStream#forever` - the "retried" stream needs to replace the "currently opened" stream with something like `ZManaged.switchable`.

svroonland: D'oh! Thanks, I will look into that

svroonland:
Something like this @iravid ?

iravid:
@svroonland Yes, this looks great. Could you add some tests?

svroonland:
The implementation unfortunately became  obsolete with the new Schedule in ZIO 1.0.0

iravid:
Would you like to revamp it @svroonland? Schedule's `Driver` abstraction makes it pretty easy to write these operators now.

svroonland:
Yup, already on it, was less hard than I thought

svroonland:
I'm hoping to get around to writing some tests soon, but if anybody else wants to pick that up I won't object :)

jdegoes:
@svroonland Would love to have this land! Please re-open when you get a chance to write a couple tests.

PR #3904, opened 2020-06-28
Explore only previously matched branch when visiting partially satisfied OR expectation

### LINKED ISSUE #3769: ZIO Test Mocks - `or` with `andThen` apparently not working as expected
Shouldn't the following test pass? However, when combining two expectations with `andThen` and applying the `or` method to the different combinations it doesn't.

```scala
@accessible
object AModule {
  trait Service {
    def fa(a: Int): UIO[Int]
    def fb(b: Int): UIO[String]
  }
}

@mockable[AModule.Service]
object AMock

object TestSpec extends DefaultRunnableSpec {
  override def spec: ZSpec[zio.test.environment.TestEnvironment, Any] =
    suite("TestMockSpec")(
      testM("") {
        val aMock = {
          val first = AMock.Fa(
            equalTo(1),
            value(1)
          )
          val second = AMock.Fb(
            equalTo(2),
            value("2")
          )

          (first andThen second) or (second andThen first)
        }

        for {
          _ <- (AModule.fb(2) *> AModule.fa(1)).provideLayer(aMock)
        } yield assertCompletes
      }
    )
}
```

The output error is

```shell
[info]    - unsatisfied expectations
[info]       one of
[info]         in sequential order
[info]           AMock.Fb with arguments equalTo(2)
[info]         in sequential order
[info]           AMock.Fa with arguments equalTo(1)
```

Changing the expectations composition to `(second andThen first) or (first andThen second)` makes the test pass.

Is this expected behavior?

ioleo:
Fixes #3769

Also adds debugging that can be enabled by `ZIO_TEST_MOCK_DEBUG=true test`. It will come in handy when analysing potential future issues or new features.

/cc @adamgfraser @renancaju


test/shared/src/main/scala/zio/test/mock/DebugProperties.scala
```diff
@@ -0,0 +1,34 @@
...
+object DebugProperties {
+
+  /**
+   * To see mock debug output during test execution, set `-Dzio.test.mock.debug=true` system property!
+   *
+   * {{{
+   *   sbt -Dzio.test.mock.debug=true test
+   * }}}
+   */
+  final val `zio.test.mock.debug` = "zio.test.mock.debug"
```
iravid, line 33: If possible, can we configure this through a TestAspect? We don't normally use system properties for anything in the library.

ioleo: It would require threading the TestAspect information to mock execution, not sure if possible without major refactoring.
It apprears System.properties fails to link on ScalaJS, so I switched to env variable instead.

This debug information is not really intended for the end user, but to developers when extending/debugging the Mocking framework itself.

adamgfraser: Is this supposed to be a user facing feature at all? Maybe it should just be a `private val` that contributors can set to `true` for debugging purposes when developing?

ioleo: Sure, that will work too. Updated code.

PR #3909, opened 2020-06-28
WIP: try Alex encoding of ZSink

simpadjo:
Based on: https://scalafiddle.io/sf/mY7Svca/0
Very early stage.

Problems faced so far: 
1) Can't declare a sink from lambda: no more `ZSink.apply` and `ZSink.fromPush`. Too much polymorphism.
2) Seems that effectful `ZSink.contramapM` will not work as well (not sure)
3) Should `invert` operate on chunks or on single values?
4) in vast majority of cases `invert` will be `identity`. Can it be optimized?

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/SinkUtils.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/SinkUtils.scala
@@ -6,7 +6,7 @@ import zio.{ IO, UIO }
 
 object SinkUtils {
 
-  def findSink[A](a: A): ZSink[Any, Unit, A, A, A] =
+  def findSink[A](a: A): ZSink[Any, Unit, A, A] =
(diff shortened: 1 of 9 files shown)


simpadjo:
Can't implement `ZTransducer >>> ZSink` :thinking:

iravid:
@simpadjo Sorry, just now got to playing with this to see the problem with `>>>`. You're right - it's not possible to implement `transducer >>> sink` in this encoding, because the leftovers have to be "un-transduced" back to the transducer's input type.

However, I think losing this combinator is a fair tradeoff for improved type inference. What do you think?

simpadjo:
@iravid I'd rather postpone it. Arrow operations should be the preferred style of how users write pipelines.
Also I'm afraid that inability to implement `>>>` is a symptom of deeper problems w/ composition.

Issue #3913, opened 2020-06-29
ZTransducer should provide a way to split on Chunk[A]


TobiasPfeifer:
Currently ZTransducer provides a way to split on a String, however there is no generalized way to split on Chunk[A].

use case video stream processing:
I've got a ZStream of Chunk[Byte]. Each Chunk[Byte] contains 0 to n video frames in PNG-format (possibly with some leftovers).
I want to split the the stream on the PNGs magic number (=delimiter).

PR #3914, opened 2020-06-29
ZTransducer should provide a way to split on Chunk[A] #3913

TobiasPfeifer:
#3913

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -1,6 +1,7 @@
 package zio.stream
 
 import scala.collection.mutable
+import scala.reflect.ClassTag
 
 import zio._
 
@@ -513,7 +514,7 @@ object ZTransducer {
   /**
    * Creates a transducer that purely transforms incoming values.
    */
-  def fromFunction[I, O](f: I => O): ZTransducer[Any, Unit, I, O] =
+  def fromFunction[I, O](f: I => O): ZTransducer[Any, Nothing, I, O] =
     identity.map(f)
 
   /**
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -609,40 +610,50 @@ object ZTransducer {
...
+  def splitOn(delimiter: String): ZTransducer[Any, Nothing, String, String] = {
+    val chars = ZTransducer.fromFunction[String, Chunk[Char]](s => Chunk.fromArray(s.toArray)).mapChunks(_.flatten)
+    val split = splitOnChunk[Char, String](Chunk.fromArray(delimiter.toArray), _.mkString(""))
+    chars >>> split
+  }
+
+  /**
+   * Splits elements on a delimiter and transforms the splits into desired output.
+   */
+  def splitOnChunk[A, O: ClassTag](delimiter: Chunk[A], fn: Chunk[A] => O): ZTransducer[Any, Nothing, A, O] =
```
iravid, line 622: Not sure we need `fn` here? That's easy enough for the user to apply with `ZTransducer#map`, no?

TobiasPfeifer: Type signature would then be `ZTransducer[Any, Nothing, A, Chunk[A]]` that results in a `ZStream[R, E, Chunk[Chunk[A]]]`
Flattening the `Chunk[Chunk[A]]` would just reverse the split. So mapping the nested `Chunk[A]` is the only useful continuation I can think of.

Is there another useful thing to do with `Chunk[Chunk[A]]` than mapping it to some `Chunk[O]`?

iravid: The resulting stream would be `ZStream[R, E, Chunk[A]]`, not `Chunk[Chunk[A]]` - applying `flattenChunks` would not reverse the split (because the delimiter is discarded) but rather provide a stream of all the individual elements.

Does that make sense?

TobiasPfeifer: you are right. I was confused when working on a ZStream[R, E, Chunk[A]], but need to call `flattenChunks` in this case before splitting anyways.

changes applied already

iravid:
Thank you @TobiasPfeifer!

PR #3915, opened 2020-06-29
Implement ZIO#filterPar & ZIO#filterNotPar

cipriansofronia:
My attempt for implementing ZIO#filterPar and ZIO#filterNotPar.

This is my first contribution to zio/zio, so if something is not right please let me know. Thanks!


adamgfraser:
@cipriansofronia Thanks for contributing! Will take a look later today!

Issue #3865, opened 2020-06-24
ZTransducer: Add more useful transducers


jdegoes:
- [ ] `def splitLine(delimiter: String, escapeStart: String = "\"", escapeEscape: String = "\"\"", escapeEnd: String = "\""): ZTransducer[Any, Nothing, String, Chunk[String]]`
- [x] `val iso8859_1: ZTransducer[Any, Nothing, Byte, String]`
- [x] `val usASCII: ZTransducer[Any, Nothing, Byte, String]`
- [x] `val utf16: ZTransducer[Any, Nothing, Byte, String]`
- [x] `val utf16BE: ZTransducer[Any, Nothing, Byte, String]`
- [x] `val utf16LE: ZTransducer[Any, Nothing, Byte, String]`
- [ ] `def splitOnPattern(regex: String): ZTransducer[Any, Nothing, String, String]`

mschuwalow:
I'll take these unless somebody wants to split up the work

jdegoes:
I believe that @vinhhv would like to reserve one or two. 😄

vinhhv:
@mschuwalow I can give `splitLine` and `usASCII` a go. Feel free to grab the others!

er1c:
I added utf32Decode/utf32BEDecode/utf32LEDecode and a "utfDecode" in https://github.com/zio/zio/pull/3980

I primarily want the "utfDecode" that doesn't require manually knowing the input BOM

PR #3920, opened 2020-06-30
add ZTransducer#iso8859_1

mschuwalow:
Part of #3865

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -1,5 +1,7 @@
 package zio.stream
 
+import java.nio.charset.StandardCharsets
+
 import scala.collection.mutable
 
 import zio._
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -540,6 +540,18 @@ object ZTransducer {
         }
     }
 
+  /**
+   * Decodes chunks of ISO/IEC 8859-1 bytes into strings.
+   *
+   * This transducer uses the String constructor's behavior when handling malformed byte
+   * sequences.
+   */
+  val iso8859_1: ZTransducer[Any, Nothing, Byte, String] =
```
iravid, line 549: shall we call it iso8859_1Decode to be aligned with utf8Decode? And allow for a potential encoding transducer in the future.

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -540,6 +540,18 @@ object ZTransducer {
...
+    ZTransducer.fromPush {
+      case Some(is) => ZIO.succeedNow(Chunk.single(new String(is.toArray, "ISO-8859-1")))
```
fsvehla, line 551: Using `java.nio.charset.StandardCharsets.ISO_8859_1` might save an allocation here

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
fsvehla, line 549: In the JVM the charset is referred to as 'ISO_8859_1', maybe we should add a `_` after iso

iravid:
Nice work Maxim, thank you!

PR #3923, opened 2020-07-01
Handle Partial Ordering in TPriorityQueue

adamgfraser:
The current encoding of `TPriorityQueue` requires the `Ordering` on elements to be a total ordering. While potentially desirable, sometimes intentionally or unintentionally users can define a partial ordering (e.g. ordering events by timestamp but not defining a further way of ordering events with the same timestamp). This PR updates `TPriorityQueue` to correctly handle these situations.

Copying @phderome.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/TPriorityQueueSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TPriorityQueueSpec.scala
@@ -1,54 +1,72 @@
 package zio.stm
 
-import zio.ZIOBaseSpec
+import zio.random.Random
 import zio.test.Assertion._
 import zio.test._
+import zio.{ Chunk, ZIOBaseSpec }
 
 object TPriorityQueueSpec extends ZIOBaseSpec {
 
+  final case class Event(time: Int, description: String)
+
+  implicit val eventOrdering: Ordering[Event] =
+    Ordering.by(_.time)
+
+  val genEvent: Gen[Random with Sized, Event] =
+    for {
+      time        <- Gen.int(-10, 10)
+      description <- Gen.alphaNumericString
+    } yield Event(time, description)
+
+  val genEvents: Gen[Random with Sized, Chunk[Event]] =
+    Gen.chunkOf(genEvent)
+
+  val genPredicate: Gen[Random, Event => Boolean] =
+    Gen.function(Gen.boolean)
+
   def spec = suite("TPriorityQueueSpec")(
     testM("offerAll and takeAll") {
-      checkM(Gen.chunkOf(Gen.anyInt)) { as =>
+      checkM(genEvents) { as =>
         val transaction = for {
-          queue  <- TPriorityQueue.empty[Int]
+          queue  <- TPriorityQueue.empty[Event]
           _      <- queue.offerAll(as)
           values <- queue.takeAll
         } yield values
-        assertM(transaction.commit)(equalTo(as.sorted))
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/stm/TPriorityQueueSpec.scala
```diff
@@ -1,54 +1,72 @@
...
+  val genEvent: Gen[Random with Sized, Event] =
+    for {
+      time        <- Gen.int(-10, 10)
+      description <- Gen.alphaNumericString
+    } yield Event(time, description)
+
+  val genEvents: Gen[Random with Sized, Chunk[Event]] =
+    Gen.chunkOf(genEvent)
+
+  val genPredicate: Gen[Random, Event => Boolean] =
```
phderome, line 24: ❤️

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -22,26 +22,25 @@ import zio.stm.ZSTM.internal._
...
  */
-final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, Int]]) extends AnyVal {
+final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, ::[A]]]) extends AnyVal {
 
   /**
    * Offers the specified value to the queue.
    */
   def offer(a: A): USTM[Unit] =
-    ref.update(map => map + (a -> map.get(a).fold(1)(_ + 1)))
+    ref.update(map => map + (a -> map.get(a).fold(::(a, Nil))(::(a, _))))
```
phderome, line 37: we specify all the values at same partial ordering in a non-empty list ( a cons), as opposed to counting how many elements of same partial ordering we have.

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -22,26 +22,25 @@ import zio.stm.ZSTM.internal._
...
- * that elements with the same priority will be taken from the queue is not
- * guaranteed.
+ * A `TPriorityQueue` contains values of type `A` that an `Ordering` is defined
+ * on. Unlike a `TQueue`, `take` returns the highest priority value (the value
+ * that is first in the specified ordering) as opposed to the first value
+ * offered to the queue. The ordering that elements with the same priority will
+ * be taken from the queue is not guaranteed.
  */
-final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, Int]]) extends AnyVal {
+final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, ::[A]]]) extends AnyVal {
```
phderome, line 31: so it seems like `TRef[SortedMap[Unit, ::[A]]` would work just as well?

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -123,14 +124,14 @@ final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, Int]])
...
-        while (i < n && k < j) {
-          builder += a
-          i += 1
-          k += 1
+        val (a, as) = iterator.next()
+        val (l, r)  = as.splitAt(n - i)
+        builder ++= l
+        r match {
+          case h :: t => updated += (a -> ::(h, t))
+          case Nil    => updated -= a
```
phderome, line 132: this is where we make use of the `a`.

PR #3925, opened 2020-07-01
Add Console support for printing to standard error

BardurArantsson:
Since I had a need for printing to the standard error in a little console application, I thought I'd try adding support upstream :). I might be missing something, but it seems like a small little win to add it.

I'm not exactly in love with the naming here, but OTOH I couldn't come up with anything much better. I considered using simply 'E' as the suffix, but that might be a little bit too opaque. Prefixing seems a bit odd because you then get different cases for the 'p' in errPutStrLn and putStrLn which might be annoying.

EDIT: I also tried to see if there was precedent in Haskell, but they only have putStr/putStrLn and hPutStr/hPutStrLn (which is the generalized handle version which already exists in the "live" scala implementation, but isn't exposed -- presumably because of a lack of safety for arbitrary PrintStreams).

### DIFF
--- a/test/shared/src/main/scala/zio/test/environment/package.scala
+++ b/test/shared/src/main/scala/zio/test/environment/package.scala
@@ -642,6 +642,7 @@ package object environment extends PlatformSpecific {
       def debug[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A]
       def feedLines(lines: String*): UIO[Unit]
       def output: UIO[Vector[String]]
+      def outputErr: UIO[Vector[String]]
       def silent[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A]
     }
 
@@ -691,7 +692,7 @@ package object environment extends PlatformSpecific {
                     IO.fromOption(d.input.headOption)
                       .orElseFail(new EOFException("There is no more input left to read"))
                   )
-          _ <- consoleState.update(data => Data(data.input.tail, data.output))
+          _ <- consoleState.update(data => Data(data.input.tail, data.output, data.errOutput))
         } yield input
       }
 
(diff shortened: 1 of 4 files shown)


adamgfraser:
@BardurArantsson Sorry for the delay, will review this weekend.

BardurArantsson:
No worries at all! I was just confused about what the exact process was :)

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -681,6 +682,13 @@ package object environment extends PlatformSpecific {
       def feedLines(lines: String*): UIO[Unit] =
         consoleState.update(data => data.copy(input = lines.toList ::: data.input))
 
+      /**
+       * Returns the contents of the output buffer. The first value written to
+       * the output buffer will be the first in the sequence.
+       */
+      val errOutput: UIO[Vector[String]] =
```
adamgfraser, line 689: Maybe `outputErr` for IDE discoverability? Also update Scaladoc.

BardurArantsson: Sure.

I wasn't quite sure what if you'd like to have the member in the Data class changed too, so I've left it for now. (It seems reasonable to me since it isn't really an issue for IDE discoverability and it is a "thing"... "the error output".)

Let me know if you'd like me to change that too -- I have no strong opinion on it :).

BardurArantsson:
New version uploaded. Hopefully it'll get through CI without issue :)

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -783,6 +808,13 @@ package object environment extends PlatformSpecific {
     def debug[R <: TestConsole, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
       ZIO.accessM(_.get.debug(zio))
 
+    /**
+     * Accesses a `TestConsole` instance in the environment and returns the
+     * contents of the error buffer.
+     */
+    val errOutput: ZIO[TestConsole, Nothing, Vector[String]] =
```
adamgfraser, line 815: Update name to conform to change above.

BardurArantsson: Sorry about missing that. Can you tell I'm a bit of a noob with this pattern? ;)

I've rebased on latest master and pushed a new version.

BardurArantsson:
Hm, seems to be a test failure, but I'm pure sure it's unrelated to anything I did. Not sure what the procedure here is...

(I see there's a Rerun button in the CircleCI user interface, but it's grayed out for me.)

test/shared/src/main/scala/zio/test/environment/package.scala
adamgfraser: No worries! You're contributing so you're already one step ahead! 😃

adamgfraser:
I reran that test.

BardurArantsson:
Cool, thanks! :)

PR #3928, opened 2020-07-01
Integrate ZStream with java.io.Reader

### LINKED ISSUE #3872: Add ZStream.fromReader
Interop with java.io.Reader

mijicd:
Closes #3872

### DIFF
--- a/streams/jvm/src/main/scala/zio/stream/platform.scala
+++ b/streams/jvm/src/main/scala/zio/stream/platform.scala
@@ -289,27 +289,51 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
   /**
    * Creates a stream from [[java.io.Reader]].
    */
-  def fromReader(reader: => Reader): ZStream[Blocking, IOException, Char] =
-    ZStream.repeatEffectOption {
-      blocking
-        .effectBlockingIO(reader.read())
-        .foldM(
-          e => ZIO.fail(Some(e)),
-          r => if (r == -1) ZIO.fail(None) else ZIO.succeed(r.toChar)
-        )
+  def fromReader(reader: => Reader, chunkSize: Int = ZStream.DefaultChunkSize): ZStream[Blocking, IOException, Char] =
+    ZStream {
+      for {
+        done           <- Ref.make(false).toManaged_
+        capturedReader <- Managed.effectTotal(reader)
+        pull = {
+          def go: ZIO[Blocking, Option[IOException], Chunk[Char]] = done.get.flatMap {
+            if (_) Pull.end
+            else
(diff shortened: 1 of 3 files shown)


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -286,6 +286,31 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
       .managed(is)
       .flatMap(fromInputStream(_, chunkSize))
 
+  /**
+   * Creates a stream from [[java.io.Reader]].
+   */
+  def fromReader(reader: => Reader): ZStream[Blocking, IOException, Char] =
```
iravid, line 292: Ahoy! Let's make this similar to `fromInputStream`. Important points:
- Chunking: there's a `read` variant that reads into an array buffer
- The by-name reader needs to be captured in value (otherwise it gets re-initialized every time)

mijicd: Ah, good points. I was overeager to make this super simple :). Updated.

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -289,27 +286,51 @@ trait ZStreamPlatformSpecificConstructors { self: ZStream.type =>
...
    */
-  def fromReader(reader: => Reader): ZStream[Blocking, IOException, Char] =
-    ZStream.repeatEffectOption {
-      blocking
-        .effectBlockingIO(reader.read())
-        .foldM(
-          e => ZIO.fail(Some(e)),
-          r => if (r == -1) ZIO.fail(None) else ZIO.succeed(r.toChar)
-        )
+  def fromReader(reader: => Reader, chunkSize: Int = ZStream.DefaultChunkSize): ZStream[Blocking, IOException, Char] =
```
iravid, line 289: I liked your previous approach with `repeatEffectOption` better :-) How about:
`ZStream.fromEffect(reader).flatMap { capturedReader => ZStream.repeatEffectOptionChunk(...)`?

mijicd: Haha, merged it, but dropped the exceptions used in first try.

mijicd: Btw. should I do the same for `fromInputStream`?

iravid: Yeah why not. Can you add a few tests for `fromReader`?

mijicd: Done.

PR #3934, opened 2020-07-04
add ZTransducer#utf16decode

mschuwalow:
part of #3865

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -1,5 +1,6 @@
 package zio.stream
 
+import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets
 
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -560,6 +561,52 @@ object ZTransducer {
   def last[O]: ZTransducer[Any, Nothing, O, Option[O]] =
     foldLeft[O, Option[O]](Option.empty[O])((_, a) => Some(a))
 
+  def magicSequence[R, E, I, O](n: Int)(f: Chunk[I] => ZTransducer[R, E, I, O]): ZTransducer[R, E, I, O] =
```
mschuwalow, line 564: I think this can be more general in that we could have a combinator that "decides" on a transducer after the first element has been emitted and is reset on a push(None).
Should be add something like this? Signature would be:

```
ZTransducer[R, E, I, O] {
  def replace[R1 <: R, E1 >: E, I1 <: I, O1](f: O => ZTransducer[R1, E1, I1, O1]): ZTransducer[R1, E1, I1, O1]
}
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -731,6 +778,64 @@ object ZTransducer {
...
+   */
+  val utf16LEDecode: ZTransducer[Any, Nothing, Byte, String] =
+    utf16Decode(StandardCharsets.UTF_16LE)
+
+  private def utf16Decode(charset: Charset): ZTransducer[Any, Nothing, Byte, String] =
+    ZTransducer {
+      ZRef.makeManaged[Option[Byte]](None).map { stateRef =>
+        {
+          case None =>
+            // should we raise an error here if we have leftovers?
```
mschuwalow, line 822: What should we do here? If we do not fail the transducer we will lose data as there is no way to decode the orphaned byte

iravid: With the UTF-8 decoding, if the stream ends (which is what `None` signifies) in the middle of 2/3/4 byte sequence, we just defer to the underlying charset decoder (which if I recall correctly just drops the malformed bytes). I'm fine with adopting the same approach here.

In the future, the `None` case will really signal the end of stream and we will stop allowing transducers to be restarted during stream processing. This will be done by separating transducers for which restarts do make sense (all the folds basically) to a separate abstraction (`ZAggregator` or something like that).

mschuwalow: Ah, I see. Thanks for the explanation 👍

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
iravid: This is kind of similar to `ZStream#peel` but in a transducer form. The infix form would be `self >>> ZStream.magicSequence` no?

Either way would be great if we can just find a better name before merging. Something around branching, headers?

mschuwalow: True, so we probably don't need an additional combinator.

What about `branchAfter` for the name? As in "after reading n values, branch"

iravid:
branchAfter sounds great to me

iravid:
Thank you @mschuwalow!

PR #3947, opened 2020-07-08
Add ZTranducer#US-ASCII

vinhhv:
Original issue: https://github.com/zio/zio/issues/3865

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -908,6 +908,18 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
       }
     }
 
+  /**
+   * Decodes chunks of US-ASCII bytes into strings.
+   *
+   * This transducer uses the String constructor's behavior when handling malformed byte
+   * sequences.
+   */
+  val usASCIIDecode: ZTransducer[Any, Nothing, Byte, String] =
+    ZTransducer.fromPush {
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -731,6 +731,18 @@ object ZTransducer {
...
 
+  /**
+   * Decodes chunks of US-ASCII bytes into strings.
+   *
+   * This transducer uses the String constructor's behavior when handling malformed byte
+   * sequences.
+   */
+  val usASCIIDecode: ZTransducer[Any, Nothing, Byte, String] =
+    ZTransducer.fromPush {
+      case Some(chunk) => ZIO.succeedNow(Chunk.single(new String(chunk.toArray, StandardCharsets.US_ASCII)))
```
iravid, line 742: Will this throw an exception when non-ASCII bytes are encountered? If so, we should use `ZIO(...)` rather than `succeedNow`.

vinhhv: @iravid it doesn't seem to be the case. From the docs:
```
This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string. 
```
In `ASCII`'s case, it returns `���`. No exception thrown

iravid:
@vinhhv Could you resolve the conflicts? Will merge afterwards

vinhhv:
@iravid done!

Issue #3948, opened 2020-07-08
./sbt issue - java.util.concurrent.RejectedExecutionException


myFree:
I am trying to follow the steps https://github.com/myFree/zio/blob/master/docs/about/contributing.md

After cloning the repo; I'm hitting

```
./sbt 
[info] welcome to sbt 1.3.13 (Ubuntu Java 11.0.7)
[info] loading global plugins from /home/brent/.sbt/1.0/plugins
[info] loading settings for project zio-build from plugins.sbt ...
[info] loading project definition from /home/brent/repos/zio/project
[info] loading settings for project root from build.sbt ...
[info] resolving key references (31663 settings) ...
[error] java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.CompletableFuture$AsyncSupply@3ce53abe rejected from java.util.concurrent.ThreadPoolExecutor@66ed726a[Running, pool size = 2, active threads = 2, queued tasks = 0, completed tasks = 0]
```
and the output of last is huge; and ends with

```
[info] resolving key references (31663 settings) ...
[error] java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.CompletableFuture$AsyncSupply@3ce53abe rejected from java.util.concurrent.ThreadPoolExecutor@66ed726a[Running, pool size = 2, active threads = 2, queued tasks = 0, completed tasks = 0]
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2055)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:825)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1355)
[error] 	at java.base/java.util.concurrent.CompletableFuture.asyncSupplyStage(CompletableFuture.java:1714)
[error] 	at java.base/java.util.concurrent.CompletableFuture.supplyAsync(CompletableFuture.java:1931)
[error] 	at org.eclipse.jgit.util.FS$FileStoreAttributes.getFileStoreAttributes(FS.java:329)
[error] 	at org.eclipse.jgit.util.FS$FileStoreAttributes.get(FS.java:296)
[error] 	at org.eclipse.jgit.util.FS.getFileStoreAttributes(FS.java:767)
[...]

PR #3952, opened 2020-07-10
Added LocalDateTime, LocalDate and BigDecimal to DeriveGen instances

### LINKED ISSUE #3951: Add missing common types to DeriveGen
There are types with no dependencies that could be added to `DeriveGen` such as `BigDecimal`

Adriani-Furtado:
closes #3951

### DIFF
--- a/test-magnolia/shared/src/main/scala/zio/test/magnolia/DeriveGen.scala
+++ b/test-magnolia/shared/src/main/scala/zio/test/magnolia/DeriveGen.scala
@@ -16,6 +16,7 @@
 
 package zio.test.magnolia
 
+import java.time.{ LocalDate, LocalDateTime }
 import java.util.UUID
 
 import magnolia._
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Hmm there is a weird CI failure which is unrelated:
```
 scalafix.interfaces.ScalafixException: Failed to fetch [Dependency(Module(ch.epfl.scala, scalafix-cli_2.12.11), 0.9.18)]from [IvyRepository(file:/home/circleci/.ivy2/local/[organisation]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext], dropInfoAttributes = true), MavenRepository(https://repo1.maven.org/maven2), MavenRepository(https://oss.sonatype.org/content/repositories/public), MavenRepository(https://oss.sonatype.org/content/repositories/snapshots)]
```

Will try again later.

ghostdogpr:
Seems same as https://github.com/zio/zio/issues/3948

Adriani-Furtado:
Yeah I noticed it was having issues downloading dependencies

test-magnolia/shared/src/main/scala/zio/test/magnolia/DeriveGen.scala
```diff
@@ -66,17 +67,22 @@ object DeriveGen {
...
+  implicit val genInt: DeriveGen[Int]                     = instance(Gen.anyInt)
+  implicit val genLong: DeriveGen[Long]                   = instance(Gen.anyLong)
+  implicit val genShort: DeriveGen[Short]                 = instance(Gen.anyShort)
+  implicit val genString: DeriveGen[String]               = instance(Gen.anyString)
+  implicit val genUnit: DeriveGen[Unit]                   = instance(Gen.unit)
+  implicit val genUUID: DeriveGen[UUID]                   = instance(Gen.anyUUID)
+  implicit val genLocalDateTime: DeriveGen[LocalDateTime] = instance(Gen.anyLocalDateTime)
+  implicit val genLocalDate: DeriveGen[LocalDate]         = instance(Gen.anyLocalDateTime.map(_.toLocalDate()))
+  implicit val genBigDecimal: DeriveGen[BigDecimal] = instance(
+    Gen.bigDecimal(BigDecimal(Double.MinValue), BigDecimal(Double.MaxValue))
```
adamgfraser, line 84: There is an issue here that `BigDecimal` is supposed to represent arbitrary precision floating point values that could be smaller than `Double.MinValue` or larger than `Double.MaxValue`. I'm not quite sure what the right answer is here since the domain is theoretically unlimited but it seems like we should be generating broader set of values than `Double` values.

Adriani-Furtado: I was trying to think of what the best approach would be but couldn't really find anything that stood out.
Any suggestions?

adamgfraser: Why don't we do 128 bit precision (twice that of `Double`) values?

Adriani-Furtado: @adamgfraser Have changed it. Please let me know if it is correct

PR #3958, opened 2020-07-11
Java Time Dependency for ScalaJS

### LINKED ISSUE #3139: Scala.js scala-java-time dependency
I brought this up on Discord, but seeing as this might need long-form discussion, thought I'd create an issue.

Basically, it seems like ZIO Scala.js projects require a dependency on `scala-java-time` starting with RC18. When I upgraded one of my Scala projects from RC17 to RC18-2, I started getting errors like:

```
Referring to non-existent class java.time.DateTimeException
  called from zio.clock.package$Clock$Service$$anon$1.currentDateTime()zio.ZIO
...
Referring to non-existent class java.time.ZoneId
  called from zio.clock.package$Clock$Service$$anon$1.$$anonfun$currentDateTime$1(scala.Long)zio.ZIO
...
Referring to non-existent method java.time.Instant$.ofEpochMilli(scala.Long)java.time.Instant
  called from zio.clock.package$Clock$Service$$anon$1.$$anonfun$currentDateTime$4(scala.Long)java.time.Instant
...
```

It seems like the moment you refer to `Runtime.default` you start getting these errors. This wasn't the case with RC17 for reasons I'm not yet sure about...

The easy fix was to add a dependency on `"io.github.cquiroz" %%% "scala-java-time" % "2.0.0-RC5"`, but is there a nicer way of handling this?

These are 2 that I can see:
1. If `scala-java-time` is a hard dependency from here on out, then maybe it should be a transitive dependency from the get-go? Rather than requiring people to manually add the `scala-java-time` dependency themselves after running into these kind of errors.
2. Make the Clock service more lazy somehow, where `scala-java-time` is only needed if you use the parts that require `java.time`. Not sure how doable this is though.

Option 1 is straightforward, but is there a concern about making the generated js files bigger? Or actually maybe that's not an issue if the unused parts get pruned out during optimization anyway.

mgenereu:
Made the java-time shim for ScalaJS always active since the default runtime includes Clock.

Resolves #3139

### DIFF
--- a/docs/interop/javascript.md
+++ b/docs/interop/javascript.md
@@ -13,10 +13,6 @@ println(s"""libraryDependencies += "dev.zio" %%% "zio" % "${zio.BuildInfo.versio
 println(s"""```""")
 ```
 
-For Java Time API which is not a part of ScalaJS, you might have to add dependency that provides `java.time` package to
-avoid linker errors when using `Clock`. ZIO uses [scala-java-time](https://github.com/cquiroz/scala-java-time) for
-running it's test.
-
 ## Example
 
 Your main function can extend `App` as follows.
@@ -31,11 +27,11 @@ object MyApp extends App {
 
   def run(args: List[String]): IO[Nothing, Int] =
     for {
-      p <- IO.effectSuspendTotal(document.createElement("p"))
-      t <- IO.effectSuspendTotal(document.createTextNode("Hello World"))
-      _ <- IO.effectSuspendTotal(p.appendChild(t))
-      _ <- IO.effectSuspendTotal(document.body.appendChild(p))
-    } yield 0
+      p <- IO.effectTotal(document.createElement("p"))
+      t <- IO.effectTotal(document.createTextNode("Hello World"))
+      _ <- IO.effectTotal(p.appendChild(t))
+      _ <- IO.effectTotal(document.body.appendChild(p))
+    } yield ExitCode.success
 }
 
 ```
--- a/docs/interop/javascript.md
+++ b/docs/interop/javascript.md
@@ -19,7 +19,7 @@ Your main function can extend `App` as follows.
 This example uses [scala-js-dom](https://github.com/scala-js/scala-js-dom) to access the DOM; to run the example you
 will need to add that library as a dependency to your `build.sbt`.
 
-```scala mdoc:js
+```scala
 import org.scalajs.dom.{document, raw}
 import zio._
 import zio.duration._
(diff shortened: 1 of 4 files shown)


mgenereu:
Opened this to do the work but I'm still not clear on the "zero dependency" concept and what falls in or out of that requirement.

docs/interop/javascript.md
```diff
@@ -31,11 +27,11 @@ object MyApp extends App {
 
   def run(args: List[String]): IO[Nothing, Int] =
```
adamgfraser, line 28: Can we use `mdoc` here to make sure this compiles?

adamgfraser:
Can you fix the compilation issue with the documentation?

mgenereu:
Sorry I didn't comment that I'm going to change the JS example to actually run in the browser per the mdoc example.

mgenereu:
I created an example that uses ZIO and ZIO Clock.  It runs locally in a demo app but mdoc:js is using a different version of ScalaJS than the ZIO build.  I think that's causing the build fail issues. Do you have any SBT experts that can get the code to compile on mdoc?

mgenereu:
Started a PR to update mdoc to ScalaJS 1.1.1 but the Linker API has changed significantly and I will need additional help from someone familiar with ScalaJS to complete that Pull Request. 
https://github.com/scalameta/mdoc/pull/365

adamgfraser:
Do we think this will be fixed in the short term? Otherwise we could merge without mdoc and do that in a follow up.

mgenereu:
@adamgfraser Since this is my first ZIO program, could you review that you like the actual code in the JavaScript example?  I wanted to:
1. Give a good example of Scala.JS code.
2. Use a Clock example since that's what we're enabling.
3. Use Console to show that it goes to the browser log.
3. Show it in the website.
    1. This is the magic `node` reference that mdoc provides.  
    2. My local copy had a getElementById assigning `node`.
    3. This is disabled until PR #3974 comes through.

mgenereu:
> Do we think this will be fixed in the short term? Otherwise we could merge without mdoc and do that in a follow up.

I broke it out.

docs/interop/javascript.md
```diff
@@ -13,29 +13,43 @@ println(s"""libraryDependencies += "dev.zio" %%% "zio" % "${zio.BuildInfo.versio
...
-import zio.{App, IO}
+import org.scalajs.dom.{document, raw}
+import zio._
+import zio.duration._
+import zio.clock._
 
-object MyApp extends App {
+object Main extends App {
 
-  def run(args: List[String]): IO[Nothing, Int] =
```
adamgfraser, line 32: Looks great! I would just specify explicit return types for all these methods so it is easier for someone to follow along.

mgenereu:
Sorry @adamgfraser. Ran short on time to finish the clean up.  I'm still looking at the mdoc upgrade stuff.  Hopefully I'll have more soon.

docs/interop/javascript.md
mgenereu: Will do this in the next version when mdoc is working.

PR #3959, opened 2020-07-11
Fix: ZStream.repeatEffectWith delays its outputs

### LINKED ISSUE #3863: ZStream.tick does not run the action first before waiting the interval

svroonland:
Resolves #3863 

In the current implementation every element is unnecessarily delayed by the updating of the schedule. The new implementation updates the schedule as part of the next pull. 

One consequence when using `Schedule.never` is that in the new version, the stream emits at least 1 element, while in the old version it would not, although the effect was evaluated.

The new version has the same semantics as `ZIO(..).repeat(schedule)`, which is to evaluate the effect once and then use the schedule to repeat.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -3523,7 +3523,7 @@ object ZStreamSpec extends ZIOBaseSpec {
             for {
               ref      <- Ref.make(0)
               effect   = ref.getAndUpdate(_ + 1).filterOrFail(_ <= length + 1)(())
-              schedule = Schedule.identity[Int].whileOutput(_ <= length)
+              schedule = Schedule.identity[Int].whileOutput(_ < length)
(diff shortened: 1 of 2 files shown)


iravid:
@svroonland Your solution looks more correct to me! Thanks for opening this. Can you grab the tests from #3891 and add them here?

PR #3960, opened 2020-07-11
Add polymorphic service support in accessible macro

### LINKED ISSUE #3950: zio-macros: support polymorphic service
Hi, it seems the macro doesn't accept polymorphic services

```scala
@accessible
object Module {
    trait Service[T <: X] {
        ....
      }
  }
```

should work

timaliberdov:
Closes #3950

### DIFF
--- a/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
+++ b/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
@@ -16,19 +16,20 @@
 
 package zio.macros
 
-import scala.reflect.macros.whitebox.Context
+import scala.reflect.macros.whitebox
 
 import com.github.ghik.silencer.silent
 
 /**
  * Generates method accessors for a service into annotated object.
  */
-private[macros] class AccessibleMacro(val c: Context) {
+private[macros] class AccessibleMacro(val c: whitebox.Context) {
   import c.universe._
 
   protected case class ModuleInfo(
     module: ModuleDef,
-    service: ClassDef
+    service: ClassDef,
+    serviceTypeParams: List[TypeDef]
   )
 
   def abort(msg: String) = c.abort(c.enclosingPosition, msg)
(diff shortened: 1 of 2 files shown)


macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
```diff
@@ -153,40 +157,50 @@ private[macros] class AccessibleMacro(val c: Context) {
...
 
-      if (isVal) q"$mods val $name: $returnType = $returnValue"
-      else
+      if (isVal && serviceTypeParams.isEmpty) q"$mods val $name: $returnType = $returnValue"
+      else {
+        val allTypeParams = serviceTypeParams ::: typeParams
+        val tagsForServiceTypeArgs = serviceTypeArgs.zipWithIndex.map {
+          case (name, idx) =>
+            val tagName = TermName(s"tag$idx")
+            q"$tagName: _root_.izumi.reflect.Tag[$name]"
```
neko-kai, line 184: It is generally more performant to request the entire service tag. e.g.
instead of:
```scala
def v[T <: Foo: Tag, U >: Bar: Tag]: ZIO[Has[Module.Service[T, U]], Throwable, T]
```
->
```scala
def v[T <: Foo, U >: Bar](implicit ev: Tag[Module.Service[T, U]]): ZIO[Has[Module.Service[T, U]], Throwable, T]
```

In the latter case, if both T & U are statically known (their tags are not provided at runtime with Tag evidences), `Tag[Module.Service[T, U]]` will be a compile-time constant. In the former case there will always be a runtime cost associated with merging the runtime tags for `Module.Service`, `T` & `U` together.

macros/shared/src/test/scala-2.x/zio/macros/AccessibleSpec.scala
```diff
@@ -117,6 +117,79 @@ object AccessibleSpec extends DefaultRunnableSpec {
...
+      },
+      testM("generates accessor for service with two type params and type bounds") {
+        assertM(typeCheck {
+          """
+             trait Foo
+             trait Bar
+
+             @accessible
+             object Module {
+               trait Service[T <: Foo, U >: Bar] {
```
neko-kai, line 163: Could you also add a test for a type parameter in contravariant position, e.g.
```scala
trait Service[-R] {
  def r: RIO[R, Unit]
}
```
?

macros/shared/src/main/scala-2.x/zio/macros/AccessibleMacro.scala
timaliberdov: Thank you! Didn't think about that.

macros/shared/src/test/scala-2.x/zio/macros/AccessibleSpec.scala
timaliberdov: Added a test. Had to update the TypeDef modifiers for Service type params to avoid the generation of def foo[-R], etc.
I'm not sure if it is efficient enough.

PR #3974, opened 2020-07-16
Activate mdoc on JavaScript.

mgenereu:
Once https://github.com/scalameta/mdoc/pull/365 is complete, we can enable mdoc and demonstrate Scala.JS and ZIO compiled right into the ZIO microsite.


jdegoes:
@mgenereu Just to avoid PR clutter, I will close this for now, but this feature enhancement would be awesome. Please re-open when it's ready!

mgenereu:
10-4

PR #3976, opened 2020-07-17
Preview new schedule

jdegoes:
A new design for Schedule based on the model that each schedule is a set of intervals, in which recurrence is allowed to occur. Schedule union, intersection, and complement all have meanings given by geometric union, intersection, and complement on the set of intervals.

This pull request is not complete nor working, some details must be figured out. But it looks quite promising and is posted for early feedback and outside help.


core/shared/src/main/scala/zio/Schedule2.scala
```diff
@@ -0,0 +1,576 @@
...
+ * object for `Schedule` contains all common types of schedules, both for performing retrying, as
+ * well as performing repetition.
+ */
+final case class Schedule2[-Env, -In, +Out](
+  private val step0: Schedule2.StepFunction[Env, In, Out]
+) { self =>
+  import Schedule2._
+  import Schedule2.Decision._
+
+  def ||[Env1 <: Env, In1 <: In, Out2](that: Schedule2[Env1, In1, Out2]): Schedule2[Env1, In1, (Out, Out2)] =
```
er1c, line 54: This isn't really a comment about the code, just a question about code naming convention.  Why refer to it as `Env1`/`In1` instead of `Env2`/`In2`.  But then use `Out2` instead of `Out1`?

core/shared/src/main/scala/zio/Schedule2.scala
```diff
@@ -0,0 +1,576 @@
...
+   *
+   * @param start
+   * @param end
+   */
+  final case class Interval(start: Instant, end: Instant) { self =>
+    def <(that: Interval): Boolean = (self min that) == self
+
+    def <=(that: Interval): Boolean = (self < that) || (self == that)
+
+    def >(that: Interval): Boolean = that < self
```
er1c, line 434: `>`?

core/shared/src/main/scala/zio/Schedule2.scala
```diff
@@ -0,0 +1,1009 @@
...
+        case Decision.Done(v)           => Decision.Done(v)
+        case Decision.Continue(v, _, _) => Decision.Done(v)
+      }
+  }
+  object Decision {
+    final case class Done[-Env, +Out](out: Out) extends Decision[Env, Any, Out]
+    final case class Continue[-Env, -In, +Out](
+      out: Out,
+      interval: Interval,
+      next: StepFunction[Env, In, Out]
```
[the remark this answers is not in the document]
regiskuckaertz: or perhaps
```scala
type StepFunction[-Env, -I, +O] = UIO[I => ZIO[Env, Nothing, Decision[Env, I, O]]]
```

core/shared/src/main/scala/zio/Schedule2.scala
jdegoes: I use `Env1` when it's a 'variation' on `Env` (I'd really prefer to use `Env'`, but it's not possible!). I use `Env2` when it is a different environment with no relation to `Env`.

core/shared/src/main/scala/zio/Schedule2.scala
```diff
@@ -0,0 +1,1009 @@
...
+ *
+ * Schedules are defined as a possibly infinite set of intervals spread out over time. Each
+ * interval defines a window in which recurrance is possible.
+ *
+ * When schedules are used to repeat or retry effects, the starting boundary of each interval
+ * produced by a schedule is used as the moment when the effect will be executed again.
+ *
+ * Schedules compose in the following primary ways:
+ *
+ *  * Union. This performs the union of the intervals of two schedules.
```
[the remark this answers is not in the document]
jdegoes: It's trying to take the "earlier" interval when there is no union. That's because if there is no overlap, then the "earlier" interval will determine when the soonest moment the effect can repeat is. Then later, we have a chance for the "later" interval to contribute to the union. Because after the effect is executed due to the first interval, then the schedule will be stepped, and the next interval up may be the "later" interval ("may be" because the schedule with the earlier interval might have another interval before the other one, etc.).

Does that make sense and is the reasoning sound?

core/shared/src/main/scala/zio/Schedule2.scala
er1c: Thanks!

core/shared/src/main/scala/zio/Schedule2.scala
```diff
@@ -0,0 +1,1009 @@
...
+      (now: Instant, in: In) =>
+        self(now, in).flatMap {
+          case Done(out) => that(now, out).map(_.toDone)
+          case Continue(out, interval, next1) =>
+            that(now, out).map {
+              case Done(out2) => Done(out2)
+              case Continue(out2, interval2, next2) =>
+                val combined = (interval union interval2).getOrElse(interval min interval2)
+
+                Continue(out2, combined, loop(next1, next2))
```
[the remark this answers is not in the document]
jdegoes: Loop returns right away, it returns a function, which isn't evaluated till later. So it's lazy recursion happening in the future.

Now every time we wrap a loop function, we add another level of indirection, and increase the stack frame size. So in essence the number of combinators you can apply to schedule is finite--too many, and you blow the stack. But we're already in that situation now since we already implement operators in existing schedule through "decoration".

core/shared/src/main/scala/zio/Schedule2.scala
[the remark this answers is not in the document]
jdegoes: The problem is if you call `unfold` twice, with the same params, you should get back the same schedule. But in this case, you will get back two different schedules, because the creation of the mutable variable is done as part of the (supposedly pure) function invocation. The two purely functional encodings are:

1. Explicit state, like the current `Schedule`, or maybe stored in a `Ref`.
2. Implicit state, which involves using "successor" step functions like `next`. 

After working with both, I think the second one is cleaner. I think I can make combinators to minimize duplication, but never having to worry about explicit state is quite nice.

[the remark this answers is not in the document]
regiskuckaertz: Fair enough 👌

core/shared/src/main/scala/zio/Schedule2.scala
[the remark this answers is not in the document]
regiskuckaertz: Yes I think so. It is still union then, in the sense that the overall behaviour of the schedule will likely range over the both intervals in order.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2262,31 +2275,39 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                          state =>
-                            switchPull((self.map(f) ++ ZStream.succeed(g(schedule.extract((), state)))).process)
-                              .tap(currPull.set(_)) *> schedStateRef.set(state) *> go
-                        )
+                      schedStateRef.get.flatMap {
+                        case (start, step) =>
+                          clock.instant.flatMap(now =>
+                            start.fold[URIO[Clock, Any]](ZIO.unit)(start => ZIO.sleep(Interval(now, start).size)) *>
+                              step(now, ()).flatMap {
+                                case Done(_) => doneRef.set(true) *> Pull.end // TODO: `out` not used???
```
iravid, line 2304: Yes that looks like an oversight. We should emit the last `out` and then end.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2262,31 +2275,39 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                        .foldM(
-                          _ => doneRef.set(true) *> Pull.end,
-                          state =>
-                            switchPull((self.map(f) ++ ZStream.succeed(g(schedule.extract((), state)))).process)
-                              .tap(currPull.set(_)) *> schedStateRef.set(state) *> go
-                        )
+                      schedStateRef.get.flatMap {
+                        case (start, step) =>
+                          clock.instant.flatMap(now =>
+                            start.fold[URIO[Clock, Any]](ZIO.unit)(start => ZIO.sleep(Interval(now, start).size)) *>
```
iravid, line 2302: Here we're sleeping and then stepping. Should it be the other way around? E.g. only sleep when `Continue` is returned?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2417,27 +2438,34 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-            case Some((a, scheduleState)) =>
-              schedule
-                .update(a, scheduleState)
-                .foldM(
-                  _ => state.set(None).as(Chunk.single(g(schedule.extract(a, scheduleState)))),
-                  s => state.set(Some(a -> s)).as(Chunk.single(f(a)))
-                )
+            case Some((start, a, step)) =>
+              clock.instant.flatMap(now =>
+                start.fold(ZIO.unit: URIO[Clock, Any])(start => clock.sleep(Interval(now, start).size)) *>
```
iravid, line 2462: Same comment: Should we only sleep when `Continue` is returned?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2471,13 +2504,14 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                          schedule
-                            .update(a, sched0)
-                            .foldM(
-                              _ =>
-                                schedule.initial
-                                  .flatMap(s1 => state.set((s1, Some(() => schedule.extract(a, sched0))))),
-                              s => state.set((s, None))
+                          clock.instant
+                            .flatMap(now =>
+                              start.fold(ZIO.unit: URIO[Clock, Any])(start => clock.sleep(Interval(now, start).size)) *>
```
iravid, line 2509: Same comment: should we only sleep when `Continue` is returned?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3538,11 +3572,21 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
-      ZStream.succeed(schedule.extract((), s)) ++
-        ZStream.unfoldM(s)(s => schedule.update((), s).map(s => (schedule.extract((), s), s)).option)
+  def fromSchedule[R, A](schedule: Schedule[R, Any, A]): ZStream[R with Clock, Nothing, A] = {
+    import Schedule.Interval
+
+    ZStream.unfoldM(Option(Option.empty[Instant] -> schedule.step)) {
+      case None => ZIO.succeed(None)
+      case Some((start, step)) =>
+        clock.instant.flatMap(now =>
+          start.fold[URIO[Clock, Any]](ZIO.unit)(start => clock.sleep(Interval(now, start).size)) *>
```
iravid, line 3582: Same: sleep only on Continue?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2417,27 +2438,34 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                  _ => state.set(None).as(Chunk.single(g(schedule.extract(a, scheduleState)))),
-                  s => state.set(Some(a -> s)).as(Chunk.single(f(a)))
-                )
+            case Some((start, a, step)) =>
+              clock.instant.flatMap(now =>
+                start.fold(ZIO.unit: URIO[Clock, Any])(start => clock.sleep(Interval(now, start).size)) *>
+                  step(now, a).flatMap {
+                    case Done(b) => state.set(None).as(Chunk.single(g(b)))
+                    case Continue(_, interval, next) =>
+                      state.set(Some((Some(interval.start), a, next))).as(Chunk.single(f(a))) // TODO: 'b' not used?
```
iravid, line 2466: The schedule's output should only be emitted when it's done, no?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2471,13 +2504,14 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                                schedule.initial
-                                  .flatMap(s1 => state.set((s1, Some(() => schedule.extract(a, sched0))))),
-                              s => state.set((s, None))
+                          clock.instant
+                            .flatMap(now =>
+                              start.fold(ZIO.unit: URIO[Clock, Any])(start => clock.sleep(Interval(now, start).size)) *>
+                                step(now, a).flatMap {
+                                  case Done(b) => state.set((None, schedule.step, Some(() => b)))
+                                  case Continue(_, interval, next) =>
+                                    state.set((Some(interval.start), next, None)) // TODO: 'b' not used???
```
iravid, line 2513: I think that's fine: `b` should only be emitted when the schedule is done

benchmarks/src/main/scala/zio/stm/TMapContentionBenchmarks.scala
```diff
@@ -29,8 +30,8 @@ class TMapContentionBenchmarks {
     val map          = unsafeRun(TMap.fromIterable(data).commit)
     val ref          = ZTRef.unsafeMake(data.toMap)
 
-    mapUpdates = UIO.foreachPar_(keysToUpdate)(i => map.put(i, i).commit.repeat(schedule))
-    refUpdates = UIO.foreachPar_(keysToUpdate)(i => ref.update(_.updated(i, i)).commit.repeat(schedule))
+    mapUpdates = ZIO.foreachPar_(keysToUpdate)(i => map.put(i, i).commit.repeat(schedule))
```
adamgfraser, line 33: This feels like a situation where we could potentially use a few repeat variants that don't use a Schedule to do very simple things like repeat an effect a specified number of times.

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -190,24 +190,25 @@ sealed trait ZLayer[-RIn, +E, +ROut] { self =>
...
-                .foldM(
-                  _ => ZIO.fail(e),
-                  s => ZIO.succeed((r, s))
-                )
+              clock.instant.flatMap(now => s(now, e).flatMap {
+                case Done(_) => ZIO.fail(e)
+                case Continue(_, interval, next) => clock.sleep(Interval(now, interval.start).size) as ((r, next))
+              }).provide(r)
           }
         update >>> loop.fresh
```
adamgfraser, line 209: I think you will need `ZLayer.suspend` around `loop.fresh` here for this to terminate once you merge the upstream changes that make parameters no longer by name.

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -79,6 +81,9 @@ package object clock {
   val currentDateTime: ZIO[Clock, DateTimeException, OffsetDateTime] =
     ZIO.accessM(_.get.currentDateTime)
 
+  val instant: ZIO[Clock, Nothing, java.time.Instant] =
```
adamgfraser, line 84: Probably should add `localDateTime` as an accessor as well. I can also pick up in a follow up.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -335,15 +335,15 @@ object TestAspect extends TimeoutVariants {
...
     flaky(100)
 
   /**
    * An aspect that retries a test until success, with the specified limit, for
    * use with flaky tests.
    */
-  def flaky(n: Int): TestAspectAtLeastR[ZTestEnv with Annotations] =
-    retry(Schedule.recurs(n))
+  def flaky(n: Int): TestAspectAtLeastR[ZTestEnv with Annotations with Live] =
+    retry(Schedule.recurs(n.toLong))
```
adamgfraser, line 346: Maybe we should overload for `Int` and `Long` so users don't have to do this everywhere?

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -16,724 +16,702 @@
...
 /**
- * Defines a stateful, possibly effectful, recurring schedule of actions.
+ * A `Schedule[Env, In, Out]` defines a recurring schedule, which consumes values of type `In`, and
+ * which returns values of type `Out`.
  *
- * A `Schedule[R, A, B]` consumes `A` values, and based on the inputs and the
- * internal state, decides whether to continue or halt. Every decision is
- * accompanied by a (possibly zero) delay, and an output value of type `B`.
+ * Schedules are defined as a possibly infinite set of intervals spread out over time. Each
+ * interval defines a window in which recurrance is possible.
```
adamgfraser, line 30:
```suggestion
 * interval defines a window in which recurrence is possible.
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2244,49 +2241,101 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
    * the end of each repetition.
    */
-  final def repeatEither[R1 <: R, B](schedule: Schedule[R1, Any, B]): ZStream[R1, E, Either[B, O]] =
+  final def repeatEither[R1 <: R, B](schedule: Schedule[R1, Any, B]): ZStream[R1 with Clock, E, Either[B, O]] =
     repeatWith(schedule)(Right(_), Left(_))
 
+  /**
+   * Repeats each element of the stream using the provided `schedule`, additionally emitting the schedule's output
+   * each time a schedule is completed.
+   * Repeats are done in addition to the first execution, so that `scheduleElements(Schedule.once)` means "emit element
```
iravid, line 2258:
```suggestion
   * Repeats are done in addition to the first execution, so that `repeatElements(Schedule.once)` means "emit element
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2244,49 +2241,101 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   * and if not short circuited, repeat element once".
+   */
+  final def repeatElements[R1 <: R](schedule: Schedule[R1, O, Any]): ZStream[R1 with Clock, E, O] =
+    repeatElementsEither(schedule).collect { case Right(a) => a }
+
+  /**
+   * Repeats each element of the stream using the provided `schedule`, additionally emitting the schedule's output
+   * each time a schedule is completed.
+   * Repeats are done in addition to the first execution, so that `scheduleElements(Schedule.once)` means "emit element
```
iravid, line 2267:
```suggestion
   * Repeats are done in addition to the first execution, so that `repeatElements(Schedule.once)` means "emit element
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2244,49 +2241,101 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+        state  <- Ref.make[Option[O]](None).toManaged_
+        pull = {
+          def go: ZIO[R1 with Clock, Option[E1], Chunk[C]] =
+            state.get.flatMap {
+              case None =>
+                as.pullElement.flatMap(o => state.set(Some(o))).as(Chunk.empty)
+
+              case Some(o) =>
+                (driver.next(o) as Chunk(f(o))) orElse (driver.last.orDie.map(b => Chunk(f(o), g(b))) <* driver.reset <* state
+                  .set(None))
```
iravid, line 2298: Two changes required here I think:
1. In `case None`, emit the element after storing it in the ref. This means the element is emitted immediately after being pulled, and the sleep will only be applied on the next pull.
2. Due to 1, if the schedule ends in `case Some(o)`, we can assume that we've already emitted all the `O` elements required. So we can emit just `Chunk(g(b))` in the `orElse` case.

iravid:
```suggestion
              case None =>
                as.pullElement.tap(o => state.set(Some(o))).map(Chunk.single(_))

              case Some(o) =>
                (driver.next(o) as Chunk(f(o))) orElse (driver.last.orDie.map(b => Chunk(g(b))) <* driver.reset <* state
                  .set(None))
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3696,17 +3656,15 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
    * Creates a stream from an effect producing a value of type `A` which repeats using the specified schedule
    */
-  def repeatEffectWith[R, E, A](effect: ZIO[R, E, A], schedule: Schedule[R, A, _]): ZStream[R, E, A] =
-    fromEffect(schedule.initial).flatMap { initial =>
-      unfoldM(initial) { state =>
-        effect.flatMap(value => schedule.update(value, state).fold(_ => None, state => Some((value, state))))
-      }
+  def repeatEffectWith[R, E, A](effect: ZIO[R, E, A], schedule: Schedule[R, A, Any]): ZStream[R with Clock, E, A] =
+    ZStream {
+      schedule.driver.toManaged_.map(driver => effect.mapError(Some(_)).tap(driver.next).map(Chunk(_)))
```
iravid, line 3661: This was updated in master. The current implementation is:
```scala
    ZStream.fromEffect(schedule.initial zip effect).flatMap {
      case (initialScheduleState, value) =>
        ZStream.succeed(value) ++ ZStream.unfoldM((initialScheduleState, value)) {
          case (state, lastValue) =>
            schedule
              .update(lastValue, state)
              .foldM(_ => ZIO.succeed(Option.empty), newState => effect.map(value => Some((value, (newState, value)))))
        }
    }
```

The main difference is that it executes the effect once without any delays, and then starts stepping the schedule. I think translated to the driver formulation that would be (untested):
```scala
ZStream.fromEffect(effect zip schedule.driver).flatMap { case (a, driver) =>
  ZStream(a) ++ 
  ZStream.repeatEffect(effect.mapError(Some(_)).tap(driver.next))
}
```

test/shared/src/main/scala/zio/test/TestAspect.scala
jdegoes: Agreed!

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Fixed

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Fixed

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: Fixed

benchmarks/src/main/scala/zio/stm/TMapContentionBenchmarks.scala
jdegoes: We could also define a `timeless` combinator that deletes `Clock` from a requirement, so you could do: `timeless(effect.retry(foo))`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1504,33 +1507,77 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
+      self.foldM(
+        e => orElse(e, None).map(Left(_)),
+        a => loop(a)
+      )
+    }
+
+  /**
+   * Repeats this effect until its error satisfies the specified predicate.
+   */
+  final def repeatUntil(f: A => Boolean): ZIO[R, E, A] =
```
adamgfraser, line 1533: I realized we already have these named `doXXX` and so on. I kind of like the symmetry of `repeatXXX` and `retryXXX` but maybe it is too late for name changes. Either way we should only have one or the other.

PR #3978, opened 2020-07-17
Add MiMa to the project

iravid:


iravid:
Taking the liberty of not waiting for the Scala.js builds

PR #3981, opened 2020-07-17
fixes for ZTransducer#utfXX

mschuwalow:
* add ZTransducer#prepend
* adds support for optional byte order mark to ZTransducer#utf8Decode
* ZTransducer#utf16Decode will now default to big endianness if no byte order mark is detected

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZTransducer.scala
@@ -119,6 +119,7 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
 
   /**
    * Reads the first n values from the stream and uses them to choose the transducer that will be used for the remainder of the stream.
+   * If the stream ends before it has collected n values the partial chunk will be provided to f.
    */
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -812,21 +828,33 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
...
+          prepend(bytes) >>> transducer
+      }
+    }
+  }
+
   /**
    * Decodes chunks of UTF-16 bytes into strings.
+   * If no byte order mark is found big-endianness is assumed.
    *
    * This will fail with an `IllegalArgumentException` if no byte order mark was found and will
```
er1c, line 846: Comment is outdated

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -812,21 +828,32 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
       }
     }
 
+    // handle optional byte order mark
+    branchAfter(3) { bytes =>
```
iravid, line 832: Very elegant!

PR #3983, opened 2020-07-18
refineToOrDie for ZStream.

quelgar:
Implement `ZStream#refineToOrDie`.

Also implement `ZManaged#refineToOrDie` in the same way as for ZIO, for type safety.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -740,12 +740,6 @@ final class ZManaged[-R, +E, +A] private (val zio: ZIO[(R, ZManaged.ReleaseMap),
     refineOrDieWith(pf)(ev1)
 
   /**
-   * Keeps some of the errors, and terminates the fiber with the rest.
-   */
-  def refineToOrDie[E1: ClassTag](implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZManaged[R, E1, A] =
-    refineOrDieWith { case e: E1 => e }(ev1)
-
-  /**
    * Keeps some of the errors, and terminates the fiber with the rest, using
    * the specified function to convert the `E` into a `Throwable`.
    */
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -2291,4 +2285,14 @@ object ZManaged extends ZManagedPlatformSpecific {
 
   private[zio] def succeedNow[A](r: A): ZManaged[Any, Nothing, A] =
     ZManaged(IO.succeedNow((Finalizer.noop, r)))
+
+  implicit final class RefineToOrDieOps[R, E <: Throwable, A](private val self: ZManaged[R, E, A]) extends AnyVal {
+
+    /**
+     * Keeps some of the errors, and terminates the fiber with the rest.
+     */
+    def refineToOrDie[E1 <: E: ClassTag](implicit ev: CanFail[E]): ZManaged[R, E1, A] =
```
iravid, line 2294: Can we keep it on the class with implicit evidence? (`implicit ev: E <:< Throwable`). Better discoverability that way.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4046,4 +4048,14 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
     case object Both   extends TerminationStrategy
     case object Either extends TerminationStrategy
   }
+
+  implicit final class RefineToOrDieOps[R, E <: Throwable, A](private val self: ZStream[R, E, A]) extends AnyVal {
+
+    /**
+     * Keeps some of the errors, and terminates the fiber with the rest.
+     */
+    def refineToOrDie[E1 <: E: ClassTag](implicit ev: CanFail[E]): ZStream[R, E1, A] =
```
iravid, line 4057: Same - can we keep it on the class?

core/shared/src/main/scala/zio/ZManaged.scala
quelgar: Yep, but the cost is that we can no longer constrain `E1 <: E` and the test cases won't pass (#2355). This PR is just a copy of what was done for `zio.ZIO` (#2360) to fix that.

Issue #3926, opened 2020-07-01
Possible scope issue after upgrading to RC21-2 (Performance)


sebver:
I still have some issues after upgrading to RC21-2 with the same code sample as in #3910. After hitting the endpoint a few times the server doesn't respond, maybe because of issues wrt the scope fix of yesterday.

https://github.com/sebver/zio-http4s-scope-issue

The odd thing is that this becomes very noticeable as soon as you run it in a container. We didn't notice it yesterday because we weren't doing that, so I've updated the example/readme. You should notice it really quickly now.

I've tested it with RC21-1 and there it works (but with the defect being logged). On RC20 it behaves normally.

fsvehla:
I can also confirm that I still see it.

adamgfraser:
Okay, on it. Thanks for reporting and for the reproducer!

adamgfraser:
Is everyone who is seeing this issue using `interop-cats`? My current hypothesis is that the issue is related to the interaction between that and the fiber supervision changes.

In particular, my understanding is that in `cats-interop` we want to fork everything on the global scope (because applications that are built using Cats Effect don't know anything about ZIO's fiber supervision model and can't be relied on to properly handle fibers per the supervision model / don't have the combinators to do so). If this is correct then the parent scope should never be closed in `FiberContext#fork` for a Cats Effect application (because the global scope is always open). But adding debug statements it appears that fibers are being forked on scopes other than the global scope, which then leads to them immediately being interrupted when the scope they are forked on is already closed.

Copying @ghostdogpr and @neko-kai.

ghostdogpr:
Hmm in interop-cats we always use `forkDaemon` and never `fork`. But maybe this could come from inside another operator. Do you have any idea where that fork operation come from?

ghostdogpr:
The only non-global fork done in this example comes from a call to `raceTimeout` in http4s. Our implementation for `race` is here https://github.com/zio/interop-cats/blob/332c7ea4debeeb70caae66850e785af86f655dad/interop-cats/shared/src/main/scala/zio/interop/cats.scala#L231 
See anything wrong with it?

adamgfraser:
Yes I think the issue is that we are doing internal forks in those combinators that are on local scopes. For example in the snippet below both raced fibers are forked on the scope of the parent fiber instead of the global scope:

```scala
package zio

object Example extends App {

  def run(args: List[String]): URIO[ZEnv,ExitCode] =
    race(showScope, showScope).exitCode

[...]

PR #3987, opened 2020-07-20
Add benchmark for race

Fristi:
Complementary benchmark for #3926

### DIFF
--- a/benchmarks/src/main/scala/zio/IORaceBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/IORaceBenchmark.scala
@@ -1,0 +1,57 @@
+package zio
+
+import java.util.concurrent.TimeUnit
+
+import cats.effect.{IO => CIO}
+import monix.eval.{Task => MTask}
+import org.openjdk.jmh.annotations._
+import zio.IOBenchmarks._
+
+import scala.annotation.tailrec
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Measurement(iterations = 5, timeUnit = TimeUnit.SECONDS, time = 3)
+@Warmup(iterations = 3, timeUnit = TimeUnit.SECONDS, time = 3)
+@Fork(3)
+class IORaceBenchmark {
+  @Param(Array("20"))
+  var depth: Int = _
+
+  @Benchmark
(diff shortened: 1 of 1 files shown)


jdegoes:
In order for this benchmark to be meaningful, it would need to perform racing repeatedly inside the same effect. Otherwise it would just measure unsafePerformSync overhead. In addition, ZIO race semantics are "first winner", not "first to finish", which means the other ones need modification to match ZIO semantics.

adamgfraser:
I'm also not sure this is the underlying cause of the issue. It seems the problem in #3926 is a performance regression between the latest two ZIO release candidates. ZIO's race may or may not be faster than the race implementation in other effect systems but I think we would need to compare the performance of race between this version and the last version to get some indication of whether it could be the cause of the performance issue. Maybe the best next step here is a minimal reproducible example showing the performance degradation between the two versions. Then as we're experimenting with different changes we can evaluate whether they move the need on that.

Fristi:
> In order for this benchmark to be meaningful, it would need to perform racing repeatedly inside the same effect. Otherwise it would just measure unsafePerformSync overhead. In addition, ZIO race semantics are "first winner", not "first to finish", which means the other ones need modification to match ZIO semantics.

Thanks for the feedback! Just reaching out to help :) I can try to have shot to implementing this benchmark correctly later this week, but if anyone else is up for to pin point the problem go ahead!

Fristi:
I tested most RC versions with JMH with this test code to see any differences:
```
def zio(): Unit = {
    @tailrec
    def sumTo(t: Task[BigInt], n: Int): Task[BigInt] =
      if (n <= 1) t
      else sumTo(t.map(_ + n), n - 1)

    def racing =
      sumTo(IO.effectTotal(0), depth).disconnect raceEither sumTo(IO.effectTotal(0), depth).disconnect

    runtime.unsafeRun(racing.repeat(Schedule.recurs(1000)))
  }
```

As John mentioned, to make the test meaningful repeating the racing behavior should give a better idea of the performance of `raceEither`.

All versions performed equally except RC17 which had twice the performance (~ 80,000 ops/s versus ~ 35,000 ops/s)

Anyway I guess my assumption there is a performance issue with `raceEither` was wrong.

jdegoes:
Will close for now although improvements to benchmarks are _always_ appreciated! 

Thanks @Fristi for your work on this!

PR #3991, opened 2020-07-21
Fixed #3854 - added Either.cond like constructor

### LINKED ISSUE #3854: Cond like constructor
Hello I was looking for a method like cond on Either. It would work like so:

```
object ZIO {
 def cond[E, A](predicate: Boolean, error: => E, result: => A): IO[E, A] = if (predicate) ZIO.succeed(result) else ZIO.fail(error)
}
```

is there another way of doing this? if not would you consider a pr with a method like this?

gerdreiss:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2401,6 +2401,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def collectParN[R, E, A, B](n: Int)(in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
     foreachParN(n)(in)(a => f(a).optional).map(_.flatten)
 
+  /**
+   * If the condition is satisfied, return the given A as success, otherwise, return the given E as error.
+   */
+  def cond[E, A](predicate: Boolean, error: => E, result: => A): IO[E, A] =
```
adamgfraser, line 2407: Should the order of `result` and `error` be reversed? I believe that would be more consistent with `Either.cond` where the `Right` ("success") case is first and then the `Left` ("failure") case. This would also be more consistent with `if else` syntax where we would write `if (predicate) success else failure`. What do you think?

gerdreiss: I agree with changing the order - it feels indeed more logical and also familiar. 
Regarding the CLA: I'm going to create a new PR with exactly one contributor, and close this one. I'll also add the the aliases you mention in the new PR. That way there'll be one commit, one contributor, one email address :)

adamgfraser: Sounds good!

adamgfraser: You could also potentially add constructors for `ZManaged` and `ZSTM` if you are interested.

PR #3994, opened 2020-07-21
Fix Potential Nontermination When Layer Fails With Error

### LINKED ISSUE #3956: Program stuck when Layer return error
I found really weird behavior in `ZLayer` with `>+>`

ZIO version: 1.0.0-R21-2

```
import zio._
import zio.console._
import zio.clock._

object LayerStuck extends zio.App {

  val env =
    ZLayer.fromEffect(ZIO.effect(1/0)) ++ ((Console.live ++ Clock.live) >+> ZLayer.fromManaged(
      ZManaged
        .makeEffect {
          //This can be anything that take time to initialize for me it was ActorSystem
          Thread.sleep(5000)
          println("open")
        }{_ => Thread.sleep(5000); println("closed")}
    ))

    def run(args: List[String]) =
      ZIO.unit.as(ExitCode.success)
        .provideCustomLayer(env)
        .catchAll(cause => putStrLn("error: " + cause).as(ExitCode.failure)) <* putStrLn("the end")

}
```

for given code I'm getting output below and program never ends.
```
error: java.lang.ArithmeticException: / by zero
the end
open
closed
```

if I change `>+>` to `>>>` program works as expected and exit with code 1
```
open
closed
error: java.lang.ArithmeticException: / by zero
the end
```

one more observation is that if I changed `putStrLn("the end")` to `putStrLn("the end").delay(10.seconds)` program complete as expected.

adamgfraser:
Resolves #3956.

The issue is that if the scope is already closed in `ZIO#foreachPar_` the forked effects will never run and will never complete the promise, resulting in non-termination.

Also restores the changes to remove the by name parameters from ZLayer to address early initialization issues.

Note that this PR is not binary compatible.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2812,11 +2812,10 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     else {
       val size = as.size
       for {
-        parentId       <- ZIO.fiberId
-        causes         <- Ref.make[Cause[E]](Cause.empty)
-        result         <- Promise.make[Nothing, Boolean]
-        failureTrigger <- Promise.make[Unit, Unit]
-        status         <- Ref.make((0, 0, false))
+        parentId <- ZIO.fiberId
+        causes   <- Ref.make[Cause[E]](Cause.empty)
+        result   <- Promise.make[Unit, Unit]
+        status   <- Ref.make((0, 0, false))
 
         startTask = status.modify {
           case (started, done, failing) =>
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2858,8 +2858,8 @@ object ZIO extends ZIOCompanionPlatformSpecific {
           .forkManaged
         _ <- interrupter.use_ {
               ZIO
-                .whenM(result.await.map(!_)) {
-                  causes.get.flatMap(ZIO.halt(_))
+                .whenM(ZIO.foreach(fibers)(_.await).map(_.exists(!_.succeeded))) {
```
iravid, line 2861: Nothing is awaiting the `result` with this change. Is it still required?

adamgfraser: You're right. Deleted the old `result` and renamed `failureTrigger` to `result`.

adamgfraser:
Yes. I think that may require additional combinators because child fibers may never run at all if the scope is already closed so they may not be able to signal that the task is completed.

iravid:
Makes sense!

PR #3999, opened 2020-07-23
Fix #3854: Add Either.cond like constructor to IO, RIO, UIO, URIO, ZIO

### LINKED ISSUE #3854: Cond like constructor

gerdreiss:
I'm not sure whether it makes sense to add the constructor to UIO and URIO, since these types do not presume errors.

### DIFF
--- a/core/shared/src/main/scala/zio/RIO.scala
+++ b/core/shared/src/main/scala/zio/RIO.scala
@@ -209,7 +209,7 @@ object RIO {
   /**
    * @see See [[zio.ZIO.cond]]
    */
-  def cond[E, A](predicate: Boolean, result: => A, error: => E): IO[E, A] =
+  def cond[R, A](predicate: Boolean, result: => A, error: => Throwable): RIO[R, A] =
     ZIO.cond(predicate, result, error)
 
   /**
--- a/core/shared/src/main/scala/zio/RIO.scala
+++ b/core/shared/src/main/scala/zio/RIO.scala
@@ -209,7 +209,7 @@ object RIO {
   /**
    * @see See [[zio.ZIO.cond]]
    */
-  def cond[R, A](predicate: Boolean, result: => A, error: => Throwable): RIO[R, A] =
+  def cond[A](predicate: Boolean, result: => A, error: => Throwable): Task[A] =
     ZIO.cond(predicate, result, error)
 
   /**
--- a/core/shared/src/main/scala/zio/RIO.scala
+++ b/core/shared/src/main/scala/zio/RIO.scala
@@ -207,6 +207,12 @@ object RIO {
     ZIO.collectParN(n)(in)(f)
 
   /**
+   * @see See [[zio.ZIO.cond]]
+   */
+  def cond[E, A](predicate: Boolean, result: => A, error: => E): IO[E, A] =
+    ZIO.cond(predicate, result, error)
+
+  /**
    * @see See [[zio.ZIO.descriptor]]
    */
   def descriptor: UIO[Fiber.Descriptor] = ZIO.descriptor
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2402,6 +2402,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     foreachParN(n)(in)(a => f(a).optional).map(_.flatten)
 
   /**
+   * Similar to Either.cond, evaluate the predicate,
+   * return the given A as success if predicate returns true, and the given E as error otherwise
+   */
+  def cond[E, A](predicate: Boolean, result: => A, error: => E): ZIO[Any, E, A] =
+    if (predicate) ZIO.succeed(result) else ZIO.fail(error)
+
+  /**
    * Returns information about the current fiber, such as its identity.
    */
   def descriptor: UIO[Fiber.Descriptor] = descriptorWith(succeedNow)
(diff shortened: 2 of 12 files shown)


adamgfraser:
Yes you're 100% right. It should only be on types that can fail and for types where the error type is fixed to a particular type such as `Throwable` the constructor should be fixed to that error type as well.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2401,6 +2401,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def collectParN[R, E, A, B](n: Int)(in: Iterable[A])(f: A => ZIO[R, Option[E], B]): ZIO[R, E, List[B]] =
     foreachParN(n)(in)(a => f(a).optional).map(_.flatten)
 
+  /**
+   * Similar to Either.cond, evaluate the predicate,
+   * return the given A as success if predicate returns true, and the given E as error otherwise
+   */
+  def cond[R, E, A](predicate: Boolean, result: => A, error: => E): ZIO[R, E, A] =
```
adamgfraser, line 2408:
```suggestion
  def cond[E, A](predicate: Boolean, result: => A, error: => E): IO[E, A] =
```

core/shared/src/main/scala/zio/RIO.scala
```diff
@@ -206,6 +206,12 @@ object RIO {
   def collectParN[R, E, A, B](n: Int)(in: Iterable[A])(f: A => ZIO[R, Option[Throwable], B]): RIO[R, List[B]] =
     ZIO.collectParN(n)(in)(f)
 
+  /**
+   * @see See [[zio.ZIO.cond]]
+   */
+  def cond[R, A](predicate: Boolean, result: => A, error: => Throwable): RIO[R, A] =
```
adamgfraser, line 212:
```suggestion
  def cond[A](predicate: Boolean, result: => A, error: => Throwable): Task[A] =
```

core/shared/src/main/scala/zio/ZIO.scala
gerdreiss: True, the 'R' is totally useless in my version

core/shared/src/main/scala/zio/RIO.scala
gerdreiss: Agreed

gerdreiss:
I've added constructors to ZManaged and ZSTM you've mentioned in my previous PR. Not 100% sure, all is correct, though it does seem to make sense...

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1463,6 +1463,13 @@ object ZManaged extends ZManagedPlatformSpecific {
   def collectParN[R, E, A, B](n: Int)(in: Iterable[A])(f: A => ZManaged[R, Option[E], B]): ZManaged[R, E, List[B]] =
     foreachParN(n)(in)(a => f(a).optional).map(_.flatten)
 
+  /**
+   * Similar to Either.cond, evaluate the predicate,
+   * return the given A as success if predicate returns true, and the given E as error otherwise
+   */
+  def cond[R, E, A](predicate: Boolean, result: => A, error: => E): ZManaged[R, E, A] =
+    if (predicate) effectTotal(result) else fail(error)
```
adamgfraser, line 1471: Can use `succeed` here.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1034,6 +1034,13 @@ object ZSTM {
   def collectAll_[R, E, A](in: Chunk[ZSTM[R, E, A]]): ZSTM[R, E, Unit] =
     foreach_(in)(ZIO.identityFn)
 
+  /**
+   * Similar to Either.cond, evaluate the predicate,
+   * return the given A as success if predicate returns true, and the given E as error otherwise
+   */
+  def cond[R, E, A](predicate: Boolean, result: => A, error: => E): ZSTM[R, E, A] =
```
adamgfraser, line 1041: Don't need `R` parameter. Can also add alias to `STM` object.

gerdreiss: Should I add one to TaskSTM, similar to RIO?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1463,6 +1463,13 @@ object ZManaged extends ZManagedPlatformSpecific {
   def collectParN[R, E, A, B](n: Int)(in: Iterable[A])(f: A => ZManaged[R, Option[E], B]): ZManaged[R, E, List[B]] =
     foreachParN(n)(in)(a => f(a).optional).map(_.flatten)
 
+  /**
+   * Similar to Either.cond, evaluate the predicate,
+   * return the given A as success if predicate returns true, and the given E as error otherwise
+   */
+  def cond[R, E, A](predicate: Boolean, result: => A, error: => E): ZManaged[R, E, A] =
```
adamgfraser, line 1470: No `R` parameter here.

gerdreiss: Oh, of course.

gerdreiss:
Well, with so much help it could only get perfect! ;-))

PR #4000, opened 2020-07-24
Add Schedule#resetAfter

ghostdogpr:
Adds a new method `Schedule#resetAfter` that resets the schedule to its initial state after some time of inactivity.
This is particularly useful when using `retry` on a never-ending task: it allows the retry policy to start fresh again.


jdegoes:
Looks fantastic! 🙏

PR #4003, opened 2020-07-24
Update scala 2.12.12

ohze:


ohze:
CI fail because silencer 1.4.4 is not available for scala 2.12.12
Pls merge #4004 instead - That PR also update silencer 1.7.0

Issue #3360, opened 2020-04-14
ClassTag on Chunk is having issue in ScalaJS


ahoy-jon:
while working on #3358, found a bug on the master.

To reproduce : 
```scala
zio.test.test("0 to 128") {
      (1 to 128).foldLeft(Chunk(1))(_ + _).materialize
      assertCompletes
    }
```
The error:
```scala
[info] ZIO Test
[info] - ChunkSpec
[info]   - 0 to 128
[info]     Fiber failed.
[info]     An unchecked error was produced.
[info]     scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: 128 is not an instance of java.lang.Byte
[info]       at scala.scalajs.runtime.UndefinedBehaviorError.fillInStackTrace(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:49052:14)
[info]       at scala.scalajs.runtime.UndefinedBehaviorError.fillInStackTrace(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:197175:67)
[info]       at scala.scalajs.runtime.UndefinedBehaviorError.<init>(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:49176:10)
[info]       at scala.scalajs.runtime.UndefinedBehaviorError.<init>(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:197182:58)
[info]       at scala.scalajs.runtime.UndefinedBehaviorError.<init>(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:197178:67)
[info]       at $throwClassCastException(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:217:46)
[info]       at $asByte(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:650:5)
[info]       at $uB(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:689:10)
[info]       at scala.runtime.ScalaRunTime$.array$undupdate(/Users/jon/Projects/zio/core-tests/js/target/scala-2.12/core-tests-test-fastopt.js:33744:17)
[...]

PR #4006, opened 2020-07-24
Optimize ZStream.range to emit chunks

### LINKED ISSUE #3877: Optimize ZStream.range to emit chunks
Instead of singleton chunks.

TobiasPfeifer:
fixes #3877 

according to the javadoc `ZStream.range` should not include the max value but it did so. I fixed this to behave the same as `Chunk.range` and `scala.collection.immutable.Range` with this change. Notice it might break existing code.

`bufferDropping - fast producer progress independently` and `bufferSliding - fast producer progress independently` Tests in `ZStreamSpec` are failing with timeout. However I've added a bunch of tests for `ZStream.range` to make sure they are not failing because of incorrect ranges. These might timeout because of performance/deadlock issues when the ranges are chunked. I'll open an Issue for that


TobiasPfeifer:
opened #4008 for the failing tests to investigate and applied workaround `ZStream.range(min, max).chunkN(1)` to make the tests pass

iravid:
Hey Tobias, thank you for implementing this! Would be great if you could investigate why these tests are failing rather than deferring to a separate ticket.

TobiasPfeifer:
@iravid I actually tried and I'm willing to fix the problem myself. I tried debugging with IntelliJ but wasn't able to follow the internals of underlying fiber executions in the test. I think I need some help or at least some hint for this.
As this was the first time debugging into the internals of ZIO and ZStream I felt kinda lost there. How would you debug this?

iravid:
Hey @TobiasPfeifer - I would usually just use `println` to figure out what's going wrong with the test (along with running it in an isolated fashion with `testOnly *ZStreamSpec* -- -t "fast producer progress independently"`.

I took a look at the tests and the problem there is that the `buffer*` combinators currently operate on whole chunks. SO those tests actually rely on range emitting singleton chunks. I think it's fair to just add a chunkN(1) on those tests.

jdegoes:
@TobiasPfeifer Any interest in continuing this? I think you're super close and we could also work on it in our next 1-on-1. Will close for now but feel free to re-open any time!

TobiasPfeifer:
after putting chunkN(1) to the failing tests I faced problems with type inference in ScalaJS. I fixed it locally with help from @robmwalsh  before realizing that it has already been fixed with #3360 

@jdegoes  I've merged zio/master and cleaned up the changes. However I cannot reopen the PR myself.

iravid:
Hi @TobiasPfeifer, can you run `sbt prepare` in your branch?

iravid (approved):
Thank you @TobiasPfeifer, and apologies for this PR being dragged out for so long.

TobiasPfeifer:
@iravid Thanks for merging. There were 2 build steps failing (2.12 jdk and dotty) bc of test failure and timeout. I'm curious if this can happen occasionally or those will be fixed elsewere? I planned to investigate how my changes break tests that seem to be unrelated (foreachParN "Maintains finalizer ordering in inner ZManaged values").

Can you clarify why you've merged the PR with 2 failing checks?

PR #4013, opened 2020-07-27
Generalize Has.AreHas to Has.Union

### LINKED ISSUE #3997: Union of Has[_] and Any
In zio-grpc, users can define traits that take two type parameters:

```
trait MyService[R,C] { 
  def method1(req: Request): ZIO[R with C, Status, Response]
}
```

Where `R` represents an environment needed by all the service methods, and `C` represents metadata that the service is expecting (value changes on each request). Both `R` and `C` should be be either a subclass of `zio.Has[_]` or `Any`. 

The library provides a few operations that allow doing all sort of transformations over `R` and `C` separately. For example,
```
trasformContext[R,C,R1](s: MyService[R,C], f: R1=>R): MyService[R1, C] 
``` 
 
To implement those transformations 4 cases need to be considered : R and C are (Any, Any), (Any, Has[_]), (Has[_], Any), (Has[_], Has[_]). This leads to some repetition. To avoid the repetition I wrote the following typeclass that let me handle the 4 cases uniformly:

```
sealed trait Combinable[R, C] {
  def union(r: R, c: C): R with C
}

object Combinable {
  implicit def anyAny: Combinable[Any, Any]                            =
    new Combinable[Any, Any] {
      def union(r: Any, c: Any): Any = r
    }
  implicit def anyHas[C <: Has[_]]: Combinable[Any, C]                 =
    new Combinable[Any, C] {
      def union(r: Any, c: C): C = c
    }
  implicit def hasAny[R <: Has[_]]: Combinable[R, Any]                 =
    new Combinable[R, Any] {
      def union(r: R, c: Any): R = r
    }
  implicit def hasHas[R <: Has[_], C <: Has[_]: Tag]: Combinable[R, C] =
    new Combinable[R, C] {
      def union(r: R, c: C): R with C = r.union[C](c)
    }
}
```

And using it whenever I need to `union` over R and C without any constraints on them. It's somewhat similar to the already existing are `AreHas` that addresses a similar problem.

It would be nice to consider a way to have something in ZIO to address situations like this. That `Combinable` would do, or possibly a superclass of `Has[_]` that can replace Any and would act as zero for union.

adamgfraser:
Resolves #3997.

There has been a desire by library authors to generalize over the types `_ <: Has[_]` and `Any` to describe conceptually "the set of all environmental requirements that can be combined generically".

This PR changes the encoding of `Has.AreHas`, which describes the capability to combine two types to form their intersection, from:

```scala
  trait AreHas[-R, -R1] {
    def union[R0 <: R, R00 <: R1: Tag](r: R0, r1: R00): R0 with R00
```

to:

```scala
  trait Union[R, R1] {
    def union(r: R, r1: R1): R with R1
  }
```

Instances are then provided for combinations of `Has` with `Has`, `Has` with `Any`, `Any` with `Has`, and `Any` with `Any`.

The biggest disadvantage of this proposal I can see is we lose the contravariance on the type parameters of `R` and `R1`. Previously, we could say that if `R` and `R1` are subtypes of `Has`, then `R0 <: R` and `R00 <: R1` are also subtypes of `Has` and so can be combined generically. But now if `R` is `Any` it is not the case that we can combine subtypes of `R` generically. This requires type annotations in the implementation of a couple of additional `ZLayer` combinators, but it only required one change to the API (adding an additional `ROut1 >: ROut` parameter in `+!+` for variance) and no additional type annotations in any of the tests, so I think the impact will be limited to library authors who are writing their own combinators to combine environment types generically.

We don't gain a lot from this capability in our own library but it avoids the need for libraries like `zio-grpc` to implement it themselves and potentially provides them with a common way to do so.

Copying @thesamet.

### DIFF
--- a/core/shared/src/main/scala/zio/Has.scala
+++ b/core/shared/src/main/scala/zio/Has.scala
@@ -81,17 +81,78 @@ object Has {
       "operator, you must ensure the service produced by your layer is " +
       "wrapped in Has."
   )
-  trait AreHas[-R, -R1] {
-    def union[R0 <: R, R00 <: R1: Tag](r: R0, r1: R00): R0 with R00
-    def unionAll[R0 <: R, R00 <: R1](r: R0, r1: R00): R0 with R00
+  trait Union[R, R1] {
+    def union(r: R, r1: R1): R with R1
   }
-  object AreHas {
-    implicit def ImplicitAre[R <: Has[_], R1 <: Has[_]]: AreHas[R, R1] =
-      new AreHas[R, R1] {
-        def union[R0 <: R, R00 <: R1: Tag](r: R0, r1: R00): R0 with R00 =
-          r.union[R00](r1)
-        def unionAll[R0 <: R, R00 <: R1](r: R0, r1: R00): R0 with R00 =
-          r.unionAll[R00](r1)
+  object Union extends LowPriorityUnionImplicits {
+    implicit def HasHasUnion[R <: Has[_], R1 <: Has[_]: Tag]: Union[R, R1] =
+      new Union[R, R1] {
+        def union(r: R, r1: R1): R with R1 =
+          r.union[R1](r1)
+      }
+  }
+  trait LowPriorityUnionImplicits {
+    implicit def HasAnyUnion[R <: Has[_]]: Union[R, Any] =
+      new Union[R, Any] {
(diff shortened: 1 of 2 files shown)


jdegoes:
Good to merge when passing! 👍

Issue #4007, opened 2020-07-24
[zio-test] Control checkM behaviour via aspects


darl:
It will be very helpful if this things could be controlled via aspects:
1. Disabling or limiting shrinking. Current default is 1000 and it cannot be controlled. 
2. Running some effect before or after checkM's `test`. E.g. cleaning up db state.

lemastero:
I am using zio-test and I ❤️  it. I would like to support it - by contributing.

adamgfraser:
This would be great. There is a related issue in #2851. We already have `Sized` in the environment to allow configuring the size of generated values but would be great to expand this to something like `TestConfig` that includes additional parameters like number of times to run property based tests, whether to shrink, how many times to shrink, etc... Then those could also be modified using test aspects.

lemastero:
Thank you @adamgfraser for letting me know. There is good chance problem 1 will be solved by #2851 I will wait for this.

Problem 2 could be explored.
Making a common setup for every test in a suite sounds like nice thing to have - set up DB before all tests, clean table contents after every test. Similar to: `beforeAll`, `afterAll`. `before`, `after` from ScalaTest JUnit.

Possibly

```scala
final def suite[R, E, T](
    label: String,
    specs: ZIO[R, E, Vector[Spec[R, E, T]]],
    exec: Option[ExecutionStrategy]
  )
```

could be extended by:

```scala
final def suite[R, E, T](
    label: String,
    specs: ZIO[R, E, Vector[Spec[R, E, T]]],
    actions: Vector[ZTestAction]
    exec: Option[ExecutionStrategy]
  )

ZTestAction = Before(...) | BeforeAll(...) | After(...) | AfterAll(...)
```
I will poke around in the code and propose something.

Not sure if that should happen before 1.0 - sounds like a change that could be not binary compatible.

adamgfraser:
@lemastero For the before and after functionality take a look at the existing `before` and `after` test aspects. `provideLayer` with a layer that includes the before or after actions can also be used.

darl:
@lemastero will you work on this?

adamgfraser:
I can also work on this once 1.0 is released.

PR #4023, opened 2020-07-30
Tune maxShrinks for small checks count

darl:
Currently `maxShrinks` always equal to 1000, even for `checkN(1)`.
With this change `maxShrinks` will correlate with checks count. 
So with long failing test it will not timeout on shrinking phase.
Default values not changed: `check(...)` == `checkN(200)(...)`

Better control over `maxShrinks` covered in #4007

### DIFF
--- a/test/shared/src/main/scala/zio/test/package.scala
+++ b/test/shared/src/main/scala/zio/test/package.scala
@@ -616,7 +616,7 @@ package object test extends CompileVariants {
 
     final class CheckNM(private val n: Int) extends AnyVal {
       def apply[R, R1 <: R, E, A](rv: Gen[R, A])(test: A => ZIO[R1, E, TestResult]): ZIO[R1, E, TestResult] =
-        checkStream(rv.sample.forever.take(n.toLong))(test)
+        checkStream(rv.sample.forever.take(n.toLong), maxShrinks = n * 5)(test)
       def apply[R, R1 <: R, E, A, B](rv1: Gen[R, A], rv2: Gen[R, B])(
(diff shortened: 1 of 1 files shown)


adamgfraser:
I think It would definitely be good to have more ways to control the number of shrinks, but this feels a bit ad hoc. We have two concepts: (1) how much samples to test and (2) how many shrinks to do. And they are not necessarily related. We might want to test only a few samples but do a lot of shrinking, or test a lot of sample but do only a little shrinking.

I think the best way to solve this problem is to create a `TestConfig` service in the environment that would contain a bunch of different variables such as default number of samples for property based tests, number of shrinks, etc... Then we could provide aspects to easily modify these. I think that should be relatively straightforward and I can probably do it in a week or so after 1.0 is released.

If we need something in the shorter term there are a couple of other options we could pursue:

- We could add an additional parameter to each of the check methods with a default value for `maxShrinks`. I think this would allow you to specify it if you wanted with no change to ergonomics for other users.
- I think we could implement a combinator on `Gen` and `Sample` to limit the size of the shrink tree

Today you can also use `Gen#noShink` if shrinking is taking too long, though that is obviously not ideal since you don't get any shrinking versus being able to control the amount of shrinking.

I think my preference would be to just do what we think the "right" thing is since we have a clear line of sight on it and I think can implement it relatively quickly.

jdegoes:
I agree we have to have more control here and the idea suggested by @adamgfraser will pave the way for future tweaks.

@darl Any chance you have the bandwidth to push this through? Feel free to re-open if you are able!

adamgfraser:
I've been working on this. Will have something today.

PR #4026, opened 2020-07-30
3685: Remove zio's Duration and replace it with java.time.Duration

adrianfilip:

PR #4029, opened 2020-07-31
3685: Replace Duration with java.time.Duration

adrianfilip:
Using the suggestions from https://github.com/zio/zio/pull/4026


adamgfraser:
@adrianfilip Looks great! Will review in more detail but can you fix the compilation failures on the documentation?

adrianfilip:
@adamgfraser Thanks. Yes, I am on that now.

core-tests/jvm/src/test/scala/zio/SerializableSpec.scala
```diff
@@ -74,8 +74,8 @@ object SerializableSpec extends ZIOBaseSpec {
       }
     },
     testM("Duration is serializable") {
-      import zio.duration.Duration
-      val duration = Duration.fromNanos(1)
+      import zio.duration._
+      val duration = 1.nanos
```
jdegoes, line 78: We should be able to keep `Duration.fromNanos` just for backward compatibility reasons.

jdegoes: Plus other ones that existed, like `fromMillis`.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -823,20 +823,20 @@ object Schedule {
    * </pre>
    */
   def fixed(interval: Duration): Schedule[Clock, Any, Int] = interval match {
-    case Duration.Infinity                    => once >>> never.as(1)
-    case Duration.Finite(nanos) if nanos == 0 => forever
-    case Duration.Finite(nanos) =>
+    case Duration.Infinity => once >>> never.as(1)
+    case Duration.Zero     => forever
+    case duration =>
```
jdegoes, line 828: Through `object Finite` apply & unapply, we can preserve backward compatibility here and reduce changes.

core-tests/jvm/src/test/scala/zio/SerializableSpec.scala
adrianfilip: It's still there but there are a few places where I used the new .nanos instead. Should I revert it in those places?

core/shared/src/main/scala/zio/Schedule.scala
adrianfilip: Oh I see. We still want the notion of Finite. I'll make the changes.

core-tests/jvm/src/test/scala/zio/SerializableSpec.scala
adrianfilip: I will revert the change to .nanos to make it easier to track changes.

core/shared/src/main/scala/zio/Schedule.scala
jdegoes: I think if we can make this change backward compatible, it would go a long way to easing migration for 1.0. Thank you!

core-tests/jvm/src/test/scala/zio/SerializableSpec.scala
jdegoes: Yes, if you can do it without any changes outside `zio.duration._`, then it's a great sign users won't have any pains, either!

adrianfilip:
The new PR is: https://github.com/zio/zio/pull/4034

PR #4031, opened 2020-07-31
Implement ZSink.take

### LINKED ISSUE #4028: Add ZSink.take(Int)
Hi guys,

Is there a function on ZStream[R,E,O] such as:

`  def runTake(n: Int): ZIO[R,E,(Chunk[O], ZStream[R,E,O])] `

Not sure I got it all right, but looking for a function that will return an effect that can collect n items and return the remainder stream.
I can also think of a number of other functions that could operate similarly, such as using runTakeUntil, runfoldUntil etc.

For example I want to do:

```
for
  (xs,rem) <- stream.runTake(100)
  _            <-  rem.map(some_func(xs)).run(ZSink....)
  yield ()
```

adamgfraser:
Resolves #4028.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
@@ -236,6 +236,21 @@ object ZSinkSpec extends ZIOBaseSpec {
           }
         }
       ),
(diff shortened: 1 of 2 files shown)


adamgfraser:
@iravid I think this is ready to go.

PR #4033, opened 2020-08-01
Fix Bugs In Chunk#indexWhere and Chunk#mapAccumM

### LINKED ISSUE #4032: Possible bug with NonEmptyChunk's `++`
This code 

```scala
object ZUUID {
  def randomUUID = UIO(ju.UUID.randomUUID())
}

case class Foo(id: ju.UUID, name: String)

object Hello extends zio.App {

  override def run(args: List[String]): zio.URIO[zio.ZEnv, ExitCode] = {
    val buildA = NonEmptyChunk(ZUUID.randomUUID.map(id => Foo(id, "AAAAA")))
    val buildB = NonEmptyChunk(ZUUID.randomUUID.map(id => Foo(id, "BBBBB")))

    (for {
      result <- ZIO.collectAll(buildA ++ buildB)
      _ <- putStrLn(result.toString())
    } yield ()) *>
      (for {
        aa <- ZIO.collectAll(buildA)
        bb <- ZIO.collectAll(buildB)
        _ <- putStrLn((aa ++ bb).toString())
      } yield ()).exitCode
  }

}
```

Produces the following result:

```
NonEmptyChunk(Foo(015366ce-64de-4993-a934-707471922065,AAAAA), Foo(c32cf3bc-9b80-4a0a-96e8-71822d6d4506,AAAAA))
NonEmptyChunk(Foo(bfb6ac09-a342-4898-8d0a-78456afa6d48,AAAAA), Foo(3ebb9b82-8b94-464d-9b83-4cf1cc316d42,BBBBB))
```

However I would expect the two result lines to be similar (disregarding the randomness of `UUID.randomUUID`)

Converting to vector before performing concatenation produces the expected result (as @reibitto pointed out on Discord)

adamgfraser:
Resolves #4032. Fixes a bug in `Chunk#mapM` that could lead to incorrect results with concatenated chunks. Hardens tests.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ChunkSpec.scala
@@ -16,21 +16,59 @@ object ChunkSpec extends ZIOBaseSpec {
   def toBoolFn[R <: Random, A]: Gen[R, A => Boolean] =
     Gen.function(Gen.boolean)
 
+  def chunks[R <: Random, A](n: Int)(a: Gen[R, A]): Gen[R, Chunk[A]] = {
+
+    val empty: Gen[R, Chunk[Nothing]] =
+      Gen.const(Chunk.empty)
+
+    val single: Gen[R, Chunk[A]] =
+      a.map(Chunk.single)
+
+    def append(n: Int): Gen[R, Chunk[A]] =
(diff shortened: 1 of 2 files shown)


adamgfraser:
@iravid Minimized this to a couple of bug fixes.

PR #4034, opened 2020-08-01
3685: zio duration uses java.time.Duration

adrianfilip:
Used the suggestions from:
- https://github.com/zio/zio/pull/4026
- https://github.com/zio/zio/pull/4029

There were minor test changes where tests relied on types.
Also because duration is now a java.time.Duration Duration.fromNanos(Long.MaxValue) will equal Duration.Infinity.


adrianfilip:
cc @adamgfraser @jdegoes

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
+      if (duration.isNegative) Zero
+      else if (duration.compareTo(JavaDuration.ofNanos(Long.MaxValue)) >= 0) Infinity
+      else fromNanos(duration.toNanos)
+  }
+
   implicit def durationInt(n: Int): DurationSyntax = new DurationSyntax(n.toLong)
 
   implicit def durationLong(n: Long): DurationSyntax = new DurationSyntax(n)
 
+  implicit class DurationOps(duration: Duration) {
```
adamgfraser, line 68:
```suggestion
  implicit final class DurationOps(private val duration: Duration) extends AnyVal {
```

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
+    final val Infinity = java.time.Duration.ofNanos(Long.MaxValue)
+    final val Zero     = java.time.Duration.ZERO
+
+    object Finite {
+
+      def apply(nanos: Long): Duration =
+        if (nanos >= 0) nanos.nanos
+        else Duration.Zero
+
+      def unapply(arg: Duration): Option[Long] = if (arg == Duration.Infinity) None else Some(arg.toNanos)
```
adamgfraser, line 42:
```suggestion
      def unapply(arg: Duration): Option[Long] = if (arg >= Duration.Infinity) None else Some(arg.toNanos)
```

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
+      }
+    }
+
+    def asScala: ScalaDuration = duration match {
+      case Duration.Infinity => ScalaDuration.Inf
+      case _                 => ScalaFiniteDuration(duration.toNanos, TimeUnit.NANOSECONDS)
+    }
+
+    def asJava: JavaDuration = duration match {
+      case Duration.Infinity => JavaDuration.ofMillis(Long.MaxValue)
```
adamgfraser, line 120:
```suggestion
      case Duration.Infinity => JavaDuration.ofNanos(Long.MaxValue)
```

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
+    final def min(other: Duration): Duration = if (duration < other) duration else other
+
+    def compare(other: Duration): Int = duration.toNanos compare other.toNanos
+
+    final def fold[Z](infinity: => Z, finite: Duration => Z): Z = duration match {
+      case Duration.Infinity => infinity
+      case f: Duration       => finite(f)
+    }
+
+    def copy(nanos: Long = duration.toNanos): Duration = Finite(nanos)
```
adamgfraser, line 135:
```suggestion
```

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
 import scala.language.implicitConversions
 
+import zio.duration.Duration
+import zio.duration.Duration.Finite
+
 package object duration {
 
+  type Duration = java.time.Duration
+  object Duration {
+    final val Infinity = java.time.Duration.ofNanos(Long.MaxValue)
```
adamgfraser, line 33:
```suggestion
   val Infinity = java.time.Duration.ofNanos(Long.MaxValue)
```

core/shared/src/main/scala/zio/duration/package.scala
```diff
@@ -16,12 +16,157 @@
...
+        if (nanos >= 0) nanos.nanos
+        else Duration.Zero
+
+      def unapply(arg: Duration): Option[Long] = if (arg == Duration.Infinity) None else Some(arg.toNanos)
+
+    }
+
+    def apply(amount: Long, unit: TimeUnit): Duration = fromNanos(unit.toNanos(amount))
+
+    def fromNanos(nanos: Long): Duration = nanos.nanos
```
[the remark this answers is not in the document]
adrianfilip: I did not use Finite because if we pass a Long.MaxValue it will result in an Infinite.

core/shared/src/main/scala/zio/duration/package.scala
adrianfilip: good point

core/shared/src/main/scala/zio/duration/package.scala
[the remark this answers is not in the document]
adrianfilip: Also all constructors (.nanos, .millis, ...) build Duration.Zero if given negative values.

jdegoes:
We could consider using a super lightweight "newtype" encoding to present our `Duration` as a subtype of Java `Duration`, if we didn't want the user to have to import syntax.

Albeit, I think it's common to do `import zio.duration._`. At least, that's what I always do.

jdegoes:
@adrianfilip Can you resolve conflicts so we can merge?

adrianfilip:
Yes.

adrianfilip:
Done

jdegoes:
💪

PR #4036, opened 2020-08-02
Implement ZTransducer#filterInput and ZTransducer#filterInputM

unclebob418:
@iravid ran into the need for this today and figured I'd have a crack at implementing it. Can you please review?


adamgfraser:
`filterInput` and `filterInputM`?

iravid:
Looks good! Like @adamgfraser mentioned, `filterInput/M` would be more suitable names as we're already using them elsewhere (ZQueue for example).

unclebob418:
thanks @adamgfraser & @iravid, I've renamed/reshuffled

iravid:
Thanks @unclebob418!

PR #4037, opened 2020-08-02
Generalize Combinators for Effectually Transforming Collections

### LINKED ISSUE #2781: How Should We Support Chunk In Effect Combinators?
We just merged #2380 providing overloaded versions of `foreach` for `Option` and `Chunk`.

For `Option` this could just be regarded as a convenience method for making working with optional values slightly more ergonomic in some cases. `Option` has at most one value so a lot of variants of "collection" operations aren't meaningful.

But `Chunk` is our own collection type so theoretically I would think we would want it to be first class, which would mean all operations that were supported on Scala collections in ZIO would also be supported on chunks. Unfortunately this is quite a few operations, not only the `foreach` and `collectAll` variants but `foldLeft`, `foldRight`, `filter`, `validate`, `partition`, etc... and their various aliases on all the companion objects.

The alternatives I see are:

1. Just implement the methods. It would be a great source of first issues but I worry a bit about increasing the maintenance burden of adding new variants where essentially when we add a new method we often need `ZIO`, `ZManaged`, and `ZSTM` variants as well as `Iterable` and `Chunk` variants and forwarders for various companion objects.
2. Implement some kind of minimal collection type class.
3. Say that because of its lazy nature only some subset of operations make sense on chunks and delineate what those are.

Right now we seem to be going down the first route but I want to be sure we really think that is the road to go down before we start adding a bunch of additional methods for `Chunk`.

### LINKED ISSUE #3369: Explore Unifying ZIO Collection Combinators
Right now we have a lot of collection combinators like `foreach` and `collectAll` that have `Iterable`, `Chunk`, and in some case `Option` variants. We also have a variety of other combinators that could probably use `Chunk` variants (see #2781 for a relatively comprehensive list).

It would be nice if we could unify these different variants. I think there are a couple of options that we could pursue:

**Use Existing Mechanics from Scala Collection Librabry**

Now that `Chunk` extends `Iterable` we could do something like:

```scala
  def foreach0[CC[x] <: Iterable[x], R, E, A, B, That](
    in: CC[A]
  )(f: A => ZIO[R, E, B])(implicit bf: ZIOBuildFrom[CC[A], B, That]): ZIO[R, E, That] =
    ???
```

Here `ZIOBuildFrom` would just be a wrapper around `CanBuildFrom` on Scala 2.12 and `BuildFrom` on Scala 2.12 to avoid having version specific files anywhere else.

Advantages:
-  All collection operations would return the same type of collection (a Chunk would return a Chunk, a List would return a List, etc...) which would improve the user experience and let us get rid of a bunch of variants
- Minimal existing abstractions beyond those that already exist elsewhere

Disadvantages:
- Signatures of these methods get more complicated and less readable
- Internal combinators now need to work with an `Iterable` instead of a `List` which means internal code will often need to be written in a less functional style
- I haven't figured out a way to write the signature above in a way that abstracts over the kind of the collection (e.g. a `Map` versus a `List`). It seems that we want to say `CC <: Iterable[A]` instead of `CC[x] <: Iterable[x]` but so far I have not been able to make this work.

**Implement some kind of `ChunkLike` trait**

`ChunkLike` witnesses an isomorphism between some other collection type and a Chunk. All collection combinators internally translate things other than Chunks to Chunks, perform operations on them, and convert back.

Advantages:
- Type signatures somewhat simpler than in the case above

Disadvantages:
- Need new `ChunkLike` abstraction
- Need to define instances for all common collection types. New user defined collection types don't work out of the box (maybe we can widen to Iterable?).
- Need to rewrite combinators in terms of `Chunk`
- Less performant for collection types other than `Chunk` because of need to convert to and from

### LINKED ISSUE #3942: Introduce ZIO.foreach and ZIO.collect that takes BuildFrom/CanBuildFrom implicit
It can be useful for ZIO foreach/collect API to accept implicit CanBuildFrom/BuildFrom implicit.

It'll provide a way to collect elements into your own collection type. The biggest value I see in when you try to collect to Map. Or even just aggregate values into a single value (count, min, max).

### LINKED ISSUE #4032: Possible bug with NonEmptyChunk's `++`

adamgfraser:
Currently combinators for effectually transforming collections generally take an `Iterable[A]` and return a `List[B]`, resulting in loss of information and requiring separate versions of combinators to preserve the specific collection type. This results in code duplication (e.g. `Chunk#mapM` versus `ZIO.foreach`) and lack of support for some collection types (e.g. there is not currently a combinator to effectually transform a `Vector` and preserve its specific collection type).

This PR uses the `BuildFrom` machinery in the Scala standard library to support transformation methods that return the same collection type `ZIO`, `ZManaged`, and `ZSTM`. Variants of `foreach` that are specialized to `Chunk` are removed since `Chunk` is a subtype of `Iterable` and `Chunk#mapM` and variants are reimplemented in terms of `ZIO.foreach`.

Resolves #2781. Resolves #3369. Resolves #3942. Resolves #4032.


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2873,34 +2853,27 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
-   *
-   * For a sequential version of this method, see `foreach_`.
-   */
-  final def foreachPar_[R, E, A](as: Chunk[A])(f: A => ZIO[R, E, Any]): ZIO[R, E, Unit] =
-    as.mapMPar_(f)
-
-  /**
-   * Applies the functionw `f` to each element of the `Iterable[A]` in parallel,
-   * and returns the results in a new `List[B]`.
+   * Applies the functionw `f` to each element of the `Collection[A]` in parallel,
```
ghostdogpr, line 2856: Not from your PR but I just noticed a small typo here.

adamgfraser: Thanks! Fixed now!

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -827,18 +827,19 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(ZIO.forkAll(list.map(a => ZIO.effectTotal(a))).flatMap(_.join))(equalTo(list))
       },
       testM("empty input") {
-        assertM(ZIO.forkAll(List.empty).flatMap(_.join))(equalTo(List.empty))
+        assertM(ZIO.forkAll(List.empty[ZIO[Any, Nothing, Unit]]).flatMap(_.join))(equalTo(List.empty))
```
jdegoes, line 830: This is the price we pay, eh?

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
```diff
@@ -292,7 +292,7 @@ object TArraySpec extends ZIOBaseSpec {
         for {
           tArray    <- makeStair(N).commit
           findFiber <- tArray.findLast(_ % largePrime == 0).commit.fork
-          _         <- STM.foreach(0 until N)(i => tArray.update(i, _ => 1)).commit
+          _         <- STM.foreach(List.range(0, N))(i => tArray.update(i, _ => 1)).commit
```
jdegoes, line 295: The old `0 until N` no longer works?

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -615,79 +615,25 @@ sealed trait Chunk[+A] extends ChunkLike[A] { self =>
...
-        var i      = 0
-        while (i < length) {
-          val j = i
-          dest = dest.zipWith(f(self(j)))(_ += _)
-          i += 1
-        }
-      }
-      dest.map(_.result())
-    }
+    ZIO.foreach(self)(f)
```
jdegoes, line 618: 😍

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -542,25 +542,29 @@ object Fiber extends FiberPlatformSpecific {
...
         UIO
-          .foreach(fibers)(_.interruptAs(fiberId))
+          .foreach[Fiber[E, A], Exit[E, A], Iterable](fibers)(_.interruptAs(fiberId))
           .map(_.foldRight[Exit[E, List[A]]](Exit.succeed(Nil))(_.zipWith(_)(_ :: _, _ && _)))
-      def poll: UIO[Option[Exit[E, List[A]]]] =
+          .map(_.map(bf.fromSpecific(fibers)))
+      def poll: UIO[Option[Exit[E, Collection[A]]]] =
         UIO
-          .foreach(fibers)(_.poll)
+          .foreach[Fiber[E, A], Option[Exit[E, A]], Iterable](fibers)(_.poll)
```
jdegoes, line 562: Yikes! These are some pretty monstrous type annotations. Could partial application trick help here?

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -1665,7 +1683,7 @@ object ZManaged extends ZManagedPlatformSpecific {
   def foreach_[R, E, A](as: Iterable[A])(f: A => ZManaged[R, E, Any]): ZManaged[R, E, Unit] =
     ZManaged(ZIO.foreach(as)(f(_).zio).map { result =>
       val (fins, _) = result.unzip
-      (e => ZIO.foreach(fins.reverse)(_.apply(e)), ())
+      (e => ZIO.foreach(fins.toList.reverse)(_.apply(e)), ())
```
jdegoes, line 1686: Seems like we can no longer work with `Iterable`. Could a low priority implicit build from compat work to restore that?

test/shared/src/main/scala/zio/test/environment/package.scala
```diff
@@ -382,7 +382,7 @@ package object environment extends PlatformSpecific {
       private lazy val freeze: IO[Unit, Set[Fiber.Status]] =
         supervisedFibers.flatMap { fibers =>
           ZIO
-            .foreach(fibers)(_.status.filterOrFail {
+            .foreach[Any, Unit, Fiber.Runtime[Any, Any], Fiber.Status, Iterable](fibers)(_.status.filterOrFail {
```
jdegoes, line 385: 😱  If there's a way to fix this, we really should explore that.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
adamgfraser: It works fine on Scala 2.12. We don't have to pay it if https://github.com/scala/bug/issues/12104 is resolved.

core-tests/shared/src/test/scala/zio/stm/TArraySpec.scala
adamgfraser: Ranges work if it is an inclusive range but not an exclusive range. The same issue applies to `Future.traverse`. This seems like a bug. Opened an issue for it at https://github.com/scala/bug/issues/12105.

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: I think this actually isn't an issue with our implementation at all. `reverse` isn't defined on `Iterable` so since we no longer automatically return a `List` we don't have a `Seq` that can be reverse.

test/shared/src/main/scala/zio/test/environment/package.scala
adamgfraser: This one is particular bad because `Set` is invariant. All we really need here is a way to widen `Set` to `Iterable`, which unfortunately the compiler won't do for us automatically here. On Scala 2.12 we can just do `foreach(fibers.toIterable)(???)` but that doesn't work on Scala 2.13 because now `toIterable` returns `this.type` so it doesn't actually widen. Let me see if we can use a type ascription on `fibers`.

adamgfraser: This one is fixed now with the new `foreach` variant for `Set`.

core/shared/src/main/scala/zio/Fiber.scala
adamgfraser: This can be rewritten as `UIO.foreach(fibers: Iterable[Fiber[E, A]])(_.poll)`.

PR #4047, opened 2020-08-03
Make binary compatibility fatal

iravid:


iravid:
Probably need to wait until 1.0.0 is resolvable :-)

PR #4048, opened 2020-08-03
Report root fibers to the supervisor

jczuchnowski:
First PR after 1.0.0 release :mechanical_arm: 

As the title says, adds reporting root fibers to the Supervisor. Based on a similar code in `FiberContext.fork` https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/internal/FiberContext.scala#L720

Needed for ZIO-ZMX.

### DIFF
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -109,6 +109,8 @@ trait Runtime[+R] {
 
     val scope = ZScope.unsafeMake[Exit[E, A]]()
 
+    val supervisor = platform.supervisor
+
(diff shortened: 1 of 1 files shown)


adamgfraser:
Can't you already do this with `Runtime.mapPlatform(_.withSupervisor(_ => zmxSupervisor))`?

jczuchnowski:
I can. But it's a different problem. Right now fibers are reported only when forked. Supervisor doesn't get notified about the root fiber.
In practice I'm ending up with a list of fibers with the parentId = 0, but I have no fiber 0.

adamgfraser:
Got it! Looks great!

PR #4056, opened 2020-08-04
Implement ZManaged#reserve

adamgfraser:
Implements `ZManaged#reserve` to allow separately accessing effects describing resource acquisition and release without having to work with the finalizer map directly.


adamgfraser:
Is there something wrong with the binary compatibility checks? Shouldn't we be able to add a new method?

iravid:
That’s weird. Must be a backwards/forwards compatibility misconfiguration. Will check soon

iravid:
@adamgfraser Looks like a MiMa bug (opened https://github.com/lightbend/mima/issues/532). Can you add the suggested filter `exclude[DirectMissingMethodProblem]("zio.ZManaged.reserve")` to `mimaBinaryIssuesFilter` in `MimaSettings.scala`?

jdegoes:
@adamgfraser Will let you take this one

PR #4059, opened 2020-08-06
intGen and longGen without retryUntil

runtologist:
The previous implementation runs into a slow path when the range is wider than `Int/Long.MaxValue`. Now both paths are fast.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -483,10 +483,10 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
     Gen.fromEffectSample {
       if (min > max) UIO.die(new IllegalArgumentException("invalid bounds"))
       else {
-        val difference = max - min + 1
         val effect =
-          if (difference > 0) nextIntBounded(difference).map(min + _)
-          else nextInt.repeatUntil(n => min <= n && n <= max)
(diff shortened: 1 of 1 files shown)


runtologist:
The test failures seem to  be unrelated flakiness. Maybe someone with permissions could restart the CI pipeline?

adamgfraser:
@runtologist Done! Will review in the morning.

adamgfraser:
Is this really faster? We're calling `nextIntBetween`, which itself is using `repeatUntil` in the same way [here](https://github.com/zio/zio/blob/60c6ba66bc5527b6bb5ad961131d8bca5d626632/core/shared/src/main/scala/zio/random/package.scala#L106). Maybe there is value in just consolidating that logic in one place but I don't think we are doing less work. I think some repetition may be necessary when going from a generator of a superset to a generator of a subset when we want the values to be uniformly distributed. The same thing is done in the implementation of `Random.between` in the Scala standard library [here](https://github.com/scala/scala/blob/d23424cd3aa17f7ad95b81018c70ceed2566962b/src/library/scala/util/Random.scala#L102).

runtologist:
Interesting, I wasn't aware of that. Thanks for the explanation. Then this change indeed boils down to consolidation, as you said. Please accept or close this MR as you find appropriate.

jdegoes:
@runtologist Thank you!

PR #4064, opened 2020-08-07
Fix Schedule.unfold causing stack overflow

tuleism:
This test

```scala
testM("free from stack overflow") {
  assertM(run(Schedule.forever *> Schedule.recurs(10000000))(List.fill(1000000)(())))(
    equalTo(Chunk.range(0L, 1000000L))
  )
}
```

causes

```
java.lang.StackOverflowError
at zio.Schedule$.$anonfun$unfold$3(Schedule.scala:1041)
```

After the fix, it takes around 5-7 seconds to execute on my machine. Not sure if it's good to leave it in.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -128,6 +128,11 @@ object ScheduleSpec extends ZIOBaseSpec {
         val scheduled = clock.currentDateTime.orDie.flatMap(schedule.run(_, 1 to 10))
         val expected  = Chunk((0L, 1.minute), (1L, 2.minute), (2L, 4.minute))
         assertM(scheduled)(equalTo(expected))
+      },
+      testM("free from stack overflow") {
+        assertM(run(Schedule.forever *> Schedule.recurs(10000000))(List.fill(1000000)(())))(
+          equalTo(Chunk.range(0L, 1000000L))
+        )
       }
     ),
     suite("Retry on failure according to a provided strategy")(
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -1,9 +1,9 @@
 package zio
 
 import scala.concurrent.Future
-
 import zio.clock.Clock
 import zio.duration._
+import zio.stream.ZStream
 import zio.test.Assertion._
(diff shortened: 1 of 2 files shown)


tuleism:
In the 1st commit, I used `Chunk.range`. But it's not available for dotty/2.13. I tried to change it to `Chunk.fromIterable`, but then got Stack Overflow from Chunk append. Issue created at #4065

ghostdogpr:
I also ran into this issue at work. The stack points at line 1041 which is `unfold`, but I am not sure from which operator it came from.

ghostdogpr:
It seems to crash after consuming events from Kafka for about 20 minutes (using zio-kafka cc @iravid ). I suspect it's the `Schedule.forever` used in `ZStream#aggregateAsync` which is itself used by zio-kafka.

iravid:
Ouch!

@ghostdogpr I think we're only using `aggregateAsync` in `Consumer.consumeWith`, so potentially you could copy `consumeWith` to your project and replace the use of `aggregateAsync` with `aggregateAsyncWithin(... Schedule.fixed(1.millis))`. It's used for batching commits so that should probably be fine.

@jdegoes @adamgfraser Can you take a look at the fix?

ghostdogpr:
Ah, actually I'm not using `consumeWith` but `subscribeAnd` so it must be somewhere else. It's hard to pinpoint, I just know that it happens only on my branch introducing zio-kafka 😄  Will investigate more tomorrow if there is no fix in between.

jdegoes:
Fix looks fine assuming it works (it would have led to infinite thunks before).

jdegoes:
@tuleism Thank you for the contribution and the test!

adamgfraser:
One issue here is we're evaluating the by name parameter immediately. What about something like?

```
  def unfold[A](a: => A)(f: A => A): Schedule[Any, Any, A] = {
    def loop(a: => A): StepFunction[Any, Any, A] = {
      lazy val a0 = a
      (now, _) => ZIO.succeed(Decision.Continue(a0, now, loop(f(a0))))
    }

    Schedule((now, _) => ZIO.succeed(Decision.Continue(a, now, loop(f(a)))))
  }
```

This resolves the stack overflow issue and preserves the laziness of the parameter.

adamgfraser:
Also I think we may want to scale back the size of the test. I was able to reproduce with 10,000 elements instead of 1,000,000.

hmemcpy:
Whoa, just hit this in production! Will upgrade asap. Thanks!

PR #4075, opened 2020-08-09
gzip transducer

LGLO:
gzip transducer. Has common parts with deflate (because it uses deflate as compression), so I pulled out some common part.


jdegoes:
I think the JS failure is unrelated, been seeing that on a lot of pull requests.

@iravid WIll let you take this one.

PR #4093, opened 2020-08-13
Docs: adjust schedule.onDecision example for ZIO 1.x

leszekgruchala:


leszekgruchala:
@adamgfraser sure, but what's that?

adamgfraser:
@leszekgruchala It is a tool that makes sure that the code example compiles. You write it by putting "scala mdoc" after the triple quotes in code blocks and then when you run the `mdoc` step in SBT or in CI it makes sure it compiles. There is more information available [here](https://scalameta.org/mdoc/) or you can just check out some of the code examples in the documentation. Ideally we should be doing this for everything, but some things may have slipped through. Let me know if you have any specific questions as you're working on it!

adamgfraser:
@leszekgruchala Thank you! Having these examples work is really important for new users.

PR #4097, opened 2020-08-14
Add ZStream.onError

joroKr21:
Use case: forward a stream from storage to HTTP server, but log errors.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2063,6 +2063,14 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     }
 
   /**
+   * Runs the specified effect if this stream fails, providing the error to the effect if it exists.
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2062,6 +2062,12 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
       }
     }
 
+  /**
+   * Runs the specified effect if this stream fails, providing the error to the effect if it exists.
```
iravid, line 2066: The corresponding method on `ZIO` guarantees the execution of `cleanup` even in the face of interruption, and also executes it uninterruptibly.

I don't think it's critical to replicate those same semantics here, but let's add a note on the scaladoc about this difference.

joroKr21: Yeah I'm not sure if it's possible to provide the same semantics.
I added a note to the scaladoc.

PR #4098, opened 2020-08-14
Implement Summaries for the ZIO Streams module

iravid:


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -498,6 +498,26 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
     exists
   }
 
+  /**
+   * Applies the side-effecting function `f` to the elements of this chunk for as long
+   * as it returns `true`.
+   */
+  def foreachWhile(f: A => Boolean): Unit = {
```
jdegoes, line 505: By convention we should call this `unsafeForeachWhile` or make it package private to `zio`.

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,333 @@
...
+  /**
+   * A summary that produces a mutable builder that is used for
+   * summarizing the inputs.
+   */
+  abstract class Mutable[-I, +O] extends Summary[I, O] {
+
+    /**
+     * Creates a new mutable builder.
+     */
+    def create: UIO[Builder[I, O]]
```
jdegoes, line 214: Maybe we should keep builder methods package private, and for users, expose a driver-like concept?

jdegoes: Of course, internally, we'd use the "unsafe" methods, not the "pure" driver methodl

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -620,6 +596,55 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
   def fromPush[R, E, I, O](push: Option[Chunk[I]] => ZIO[R, E, Chunk[O]]): ZTransducer[R, E, I, O] =
     ZTransducer(Managed.succeed(push))
 
+  /**
+   * Creates a summary from a transducer and a predicate function.
```
jdegoes, line 600:
```suggestion
   * Creates a transducer from a summary and a predicate function.
```

streams/shared/src/main/scala/zio/stream/ZTransducer.scala
```diff
@@ -620,6 +596,55 @@ object ZTransducer extends ZTransducerPlatformSpecificConstructors {
...
+   */
+  def fromSummary[I, O](summary: Summary[I, O])(emitWhen: O => Boolean): ZTransducer[Any, Nothing, I, O] =
```
jdegoes, line 602: Consider also adding `toTransducer` on `Summary`.

iravid: It's there and uses the `Summary[I, Boolean]` scheme we came up with, but I'm still experimenting with different ways of signaling when to emit.

We may still want a way to abstract over these emission conditions.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2444,6 +2464,30 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   * Feeds this stream to the specified summary, resulting in a stream of the summary's outputs.
+   */
+  final def scanWith[O1 >: O, O3](summary: Summary[O1, O3]): ZStream[R, E, O3] =
+    zipWithSummary(summary).map(_._2)
+
+  /**
+   * Applies an accumulating function to the stream, resulting in a stream of all the intermediate
+   * results of the function.
+   */
+  def scan[O1 >: O, S](s: S)(f: (S, O1) => S): ZStream[R, E, S] =
```
jdegoes, line 2477: This looks really nice.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2444,6 +2464,30 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+    scanWith(Summary.foldLeft(s)(f))
+
+  /**
+   * Applies an effectual accumulating function to the stream, resulting in a stream of all the
+   * intermediate results of the function.
+   */
+  def scanM[R1 <: R, E1 >: E, O1 >: O, S](s: S)(f: (S, O1) => ZIO[R1, E1, S]): ZStream[R1, E1, S] =
```
jdegoes, line 2484: This presumably could also be done with `ZSummary`.

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,368 @@
...
+        }
+    }
+}
+
+object Summary {
+  // Summary -> Transducer: process until defined, emit
+  // Summary -> AggregateAsync: process until defined then emit; emit when extracted
+  // Summary -> Sink: process until defined then emit; emit when stream ends
+
+  sealed trait Step[+O] { self =>
```
jdegoes, line 164: `EmitConditions` returns! Why did this come back?

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,368 @@
...
+  object Mutable {
+    def apply[I, O](makeBuilder: UIO[Builder[I, O]]): Mutable[I, O] =
+      new Mutable[I, O] {
+        def create = makeBuilder
+      }
+  }
+
+  // Delete conditions
+  // Boolean operators on summaries
+  // Transducer <-> Summary? One a newtype over the other?
```
jdegoes, line 233: Leftover todos.

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,368 @@
...
+      cb.result()
+    }
+
+    /**
+     * Extracts the summary value maintained by the builder.
+     */
+    def extract(): O
+  }
+
+  object Builder {
```
jdegoes, line 276: Should we make builder package-private?

iravid:
Ready for review! @zio/core

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,813 @@
+package zio.stream
+
+import scala.collection.mutable
+
+import zio._
+import zio.stream.internal.Builder
+
+/**
```
mschuwalow, line 8: These docstrings are amazing. Especially comparing the different options :+1:

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,813 @@
...
+  def toTransducer: ZTransducer[R, E, I, O] =
+    ZTransducer.fromSummary(self)
+
+  /**
+   * Creates a sink from this summary.
+   */
+  def toSink[I1 <: I]: ZSink[R, E, I1, I1, O] =
+    ZSink.fromSummary(self)
+
+  private def zipMutableEffectual[R0, E0, I0, A, B, Z](
```
mschuwalow, line 279: Do we have a convention of putting private methods after public ones?

streams/shared/src/main/scala/zio/stream/ZSummary.scala
```diff
@@ -0,0 +1,813 @@
...
+ *
+ * ZSummaries can be zipped together (using [[ZSummary#zipWith]]) to create composite
+ * aggregations and can be transformed using familiar combinators like [[ZSummary#map]],
+ * [[ZSummary#contramap]] and more.
+ *
+ * The majority of ZSummaries are pure: they do not use environmental effects and
+ * cannot fail. However, a select few can be effectful. It is recommended to stick
+ * to the pure ones for superior performance.
+ */
+sealed trait ZSummary[-R, +E, -I, +O] { self =>
```
mschuwalow, line 30: Is it possible/useful to allow arrow composition of summaries?

streams/shared/src/main/scala/zio/stream/internal/Builder.scala
```diff
@@ -0,0 +1,114 @@
...
+     * before stepping the builder.
+     */
+    def copy(): Mutable[I, O]
+
+    /**
+     * Tests whether the builder will accept the proposed element. The caller
+     * must always verify this predicate before calling `step`, and must
+     * only call `step` if this method returns `true`.
+     */
+    def accepts(i: I): Boolean
```
mschuwalow, line 57: Can you clarify the relationship between accepts and reset as you are using in for example
`ZTransducer#fromSummary?`

mschuwalow:
Do we want to add a dedicated testsuite for ZSummary? It's being excercised a bit as part of the ZStreamSpec, but it would be nice to have more exhaustive tests.

streams/shared/src/main/scala/zio/stream/ZSummary.scala
iravid: Definitely useful. I gave it a try and ran into some trouble; I might have given up too fast but I thought it'd be fine to follow up with later.

streams/shared/src/main/scala/zio/stream/ZSummary.scala
iravid: :shrug: happy to adjust

iravid:
> Do we want to add a dedicated testsuite for ZSummary? It's being excercised a bit as part of the ZStreamSpec, but it would be nice to have more exhaustive tests.

Yes definitely. On my todo list.

streams/shared/src/main/scala/zio/stream/internal/Builder.scala
```diff
@@ -0,0 +1,114 @@
+package zio.stream.internal
+
+import zio._
+
+/**
+ * A mutable builder used for aggregating values of type `I` into values of type `O`
+ * through a [[ZSummary]].
+ */
+sealed abstract class Builder[-R, +E, -I, +O] { self =>
```
jdegoes, line 9: I think we should probably hide all this to the ZIO package.

streams/shared/src/main/scala/zio/stream/internal/Builder.scala
jdegoes: I too feel like we've kind of gone backward a bit, in that we're incorporating a notion of "I'm done" into a summary that is never really done, in order to support aggregate async use case.

PR #4113, opened 2020-08-19
Implement Schedule#unionWith

### LINKED ISSUE #4101: scheduleA || scheduleB completes unexpectedly when scheduleA is done
Given
```scala
scheduleA = Schedule.recurWhile[Boolean](_ == true)
scheduleB = Schedule.fixed(1.second)
schedule = scheduleA || scheduleB
```

In RC21-2 this would create a schedule that continued at intervals of 1 seconds whenever the input was `false`. But in ZIO 1.0 the schedule stops as soon as an input of `false` is seen. 

The definition of this behavior seems to be in `Schedule#combineWith` which translates the `Done` of either side to a `Done` of the combined Schedule. 

See also https://scastie.scala-lang.org/pV3odrT4TwqI7XQFYWnBgA 

For context, my use case here is AWS Kinesis polling, where a value in the response indicates whether more data is available to poll, in which case I want to repeat the poll action immediately, otherwise wait for a second.

adamgfraser:
Resolves #4101.

`combineWith` takes a function to specify how to merge the intervals of two schedules, but it is always done if either schedule is done so to a certain extent it already embodies intersection semantics in that one schedule can no longer continue if the other one doesn't want to.

This PR renames `combineWith` to `unionWith` and implements a new combinator `intersectWith` that still takes a merge function but now continues as long as either schedule wants to continue. `||` can then be implements in terms of `unionWith` to make use cases like `Schedule.recurWhile[Boolean](_ == true) || Schedule.fixed(1.second)` work.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -384,6 +384,12 @@ object ScheduleSpec extends ZIOBaseSpec {
           retries        <- retriesCounter.get
         } yield retries
       }(equalTo(10))
+    },
+    testM("union of two schedules should continue as long as either wants to continue") {
+      val schedule = Schedule.recurWhile[Boolean](_ == true) || Schedule.fixed(1.second)
+      assertM(run(schedule >>> Schedule.elapsed)(List(true, false, false, false, false)))(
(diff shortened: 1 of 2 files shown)


jdegoes:
Nice spot!

PR #4115, opened 2020-08-19
Improve Support For Single Threaded Environments

adamgfraser:
Ideally we would like the programs to produce the same results whether they are run in a single threaded or multithreaded environment. This PR pushes this by trying to get the entire test suite working with a single thread and no automatic yielding. We are not that far off. I added `yieldNow` in some combinators like `eventually` and `forever` to make sure that other fibers have an opportunity to make progress between repetitions. There are a couple of tests that aren't working right now. Will comment on them in more detail below.


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -2284,7 +2284,7 @@ object ZIOSpec extends ZIOBaseSpec {
       testM("deep fork/join identity") {
```
adamgfraser, line 2284: Not sure why this test isn't working on Scala.js. It completes normally if I run the tests in parallel but fails with out of memory if I run them sequentially. Maybe something about starting with a fresh fiber?

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -1103,7 +1103,7 @@ object ZSTMSpec extends ZIOBaseSpec {
           updater = ref.update(_ + 10).commit.forever
           res    <- (left <|> right).commit.race(updater)
```
adamgfraser, line 1104: I think we need a way to yield to the runtime in the middle of evaluating `commit`. Right now we are racing an STM effect that retries as long as the reference is less than 500 with an effect that increments the reference. But it looks like since we are single threaded and we don't yield we are just caught on the left hand side and never complete. @mijicd Not sure what the best way to do this is. Any thoughts?

core-tests/shared/src/test/scala/zio/CancelableFutureSpec.scala
```diff
@@ -43,15 +43,15 @@ object CancelableFutureSpec extends ZIOBaseSpec {
         val result = roundtrip.orDie.forever
 
         assertM(Live.live(result.timeout(1.seconds)))(isNone)
-      } @@ zioTag(supervision, regression),
+      } @@ zioTag(supervision, regression) @@ ignore,
       testM("unsafeRunToFuture interruptibility") {
         for {
           runtime <- ZIO.runtime[Any]
           f        = runtime.unsafeRunToFuture(UIO.never)
           _       <- UIO(f.cancel())
```
adamgfraser, line 51: It looks like we are hanging on attempting to interrupt, I think because our thread is waiting for the future to complete so it can return from the interruption rather than completing execution of the future.

jdegoes: We could make sure that `unsafeRunToFuture` implementation does not block, but rather, uses Scala `Promise`.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
jdegoes: Could test hypothesis with `@@ forked` to further narrow down. Good job locating the problem to this test though.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
jdegoes: I think `repeatN` and any other "spin loops" should yield in the middle. We can't really make `commit` yield without killing performance, and it shouldn't be necessary anyway if we don't have spin loops elsewhere.

mijicd: Imho, "early yield" approach you mentioned seems to be the only reasonable solution here.

adamgfraser: Agreed on the spin loops yielding in the middle but I don't think that helps here. In this example the right hand side is using `forever` and is yielding back to the runtime on every evaluation. But the right hand side is never getting evaluated at all because the left hand side is never completing or yielding because the transactional condition is never satisfied. In fact if I put a debug statement in `left` it looks like `left` is getting evaluated repeatedly in a busy loop, which I think is undesirable no matter what. So I think maybe if we fix that then if there are no more changes to the transactional variables the left hand side should be able to suspend which would then allow the runtime to use the thread to evaluate the right hand side and make progress?

core-tests/shared/src/test/scala/zio/CancelableFutureSpec.scala
adamgfraser: It actually does use `Promise` already, but apparently the current implementation is not sufficient.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
adamgfraser: Still seeing the same issue with forking both that test and all tests in the suite so still not sure what is going on here.

core-tests/shared/src/test/scala/zio/CancelableFutureSpec.scala
adamgfraser: The issue was that the cancelation action returned by `unsafeRunAsyncCancelable` blocks for the cancelation to complete, which we can't afford to do when we only have a single thread. Added a new variant that takes a callback we can use to complete a Scala `Promise` and implemented the blocking version in terms of it.

core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
adamgfraser: @mijicd Tried an alternative implementation of `orTry` that seems to avoid the issue. Would you take a look when you have the chance?

adamgfraser:
I think we have just about everything taken care of at this point. The issue with the out of memory error on the concurrent Fibonacci test on Scala.js with sequential tests is still outstanding but I think we can handle that in a follow up. If we can get agreement on the STM change I think I revert the runtime to the original settings, we get this in since it has a bunch of miscellaneous fixes, and then we can open issues to add single threaded runs to the CI matrix and to come back to using `setImmediate` on Scala.js.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -623,32 +623,11 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
-              throw res
-          }
-
-        continued.exec(journal, fiberId, stackSize, r)
-      }
-    })
+    continueWithM {
+      case TExit.Fail(e)    => ZSTM.fail(e)
+      case TExit.Succeed(a) => ZSTM.succeedNow(a)
+      case TExit.Retry      => that
```
mijicd, line 629: Hmm, this behavior is the one we tried to remove with `orTry`, i.e. being "stuck" in retrying the first transaction.

jdegoes:
Looks good to me. Great improvement!

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -683,7 +683,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Repeats this effect forever (until the first error). For more sophisticated
    * schedules, see the `repeat` method.
    */
-  final def forever: ZIO[R, E, Nothing] = self *> self.forever
+  final def forever: ZIO[R, E, Nothing] =
+    self *> ZIO.yieldNow *> forever
```
jdegoes, line 687: These look awesome. Making ZIO "single-thread safe" by default is going to be amazing.

PR #4120, opened 2020-08-20
Immediately Shift to ZIO Thread Pool In UnsafeRunAsync

### LINKED ISSUE #4117: `unsafeExecute` evaluates on the current thread pool
Currently, there is an undocumented behavior around `unsafeExecute` where typically when you call it inside your program it would always evaluate the given ZIO instance on the current thread unless an async boundary is reached.

```scala
object Main extends App {
  // Print the current thread name
  def threadName = UIO(println(Thread.currentThread.getName))

  // Yield to trigger async boundary and then print the current thread name
  def threadNameY = ZIO.yieldNow *> threadName

  // Test program that "tries" to run the provided ZIO in a thread
  def program(uio: UIO[Unit]) =
    ZIO.runtime[Any].flatMap { runtime =>
      UIO {
        new Thread("my-thread") {
          override def run = runtime.unsafeRunAsync_(uio)
        }.start()
      }
    }

  def run(args: List[String]) =
    (for {
      _ <- program(threadName)  // Prints "my-thread"
      _ <- program(threadNameY) // Prints zio-async-thread
    } yield ()).exitCode
}

```
In the above example using `threadName` actually prints "my-thread" when in fact it should have printed something that's prefixed with "zio-async-thread...". The computation moves only to the `zio-async-thread` when an async boundary is seen for the first time. Which is what is being synthetically inside `threadNameY`.

It seems like `.lock` doesn't help either —

```diff
   def program(uio: UIO[Unit]) =
    ZIO.runtime[Any].flatMap { runtime =>
       UIO {
         new Thread("my-thread") {
-         override def run = runtime.unsafeRunAsync_(uio)
+         override def run = runtime.unsafeRunAsync_(uio.lock(runtime.platform.executor))
         }.start()
       }
     }
```

This IMHO is a bit counter-intuitive and can potentially be missed, thus causing some thread designed for handling only events to choke by spending time in evaluating ZIO fibers.

adamgfraser:
Resolves #4117.

### DIFF
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -130,7 +130,9 @@ trait Runtime[+R] {
       context.onDone(exit => supervisor.unsafeOnEnd(exit, context))
     }
 
-    context.evaluateNow(ZIOFn.recordStackTrace(() => zio)(zio.asInstanceOf[IO[E, A]]))
+    lazy val curZio = ZIO.yieldNow *> zio
+
+    context.evaluateNow(ZIOFn.recordStackTrace(() => curZio)(curZio.asInstanceOf[IO[E, A]]))
     context.runAsync(k)
 
     fiberId => unsafeRun(context.interruptAs(fiberId))
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -79,7 +79,7 @@ trait Runtime[+R] {
   final def unsafeRunSync[E, A](zio: => ZIO[R, E, A]): Exit[E, A] = {
     val result = internal.OneShot.make[Exit[E, A]]
 
-    unsafeRunAsync(zio)((x: Exit[E, A]) => result.set(x))
+    unsafeRunWith(zio)(result.set)
 
     result.get()
   }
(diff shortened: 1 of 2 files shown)


jdegoes:
I think we have to avoid this behavior on Scala.js, or else it becomes impossible to run anything asynchronously, right? At least, assuming we end up with `setImmediate` for our execution context.

Some callback code in JS has to be run synchronously: e.g. when you handle a mouse or keyboard event.

adamgfraser:
Yes, I think that is a fairly straightforward change on Scala.js. The fact that all the test suites fail with this change on the JVM is somewhat concerning.

adamgfraser:
The issue was that `unsafeRunSync` was defined in terms of `unsafeRunAsync`. `unsafeRunSync` is relatively safe for small effects if the effect is run on the same thread because then the `OneShot` will be set by the time `get` is called so there will be no blocking. But if we introduce an asynchronous step at the beginning every time the current thread immediately proceeds to evaluate `OneShot#get` blocking and potentially starving the ZIO thread pool if the thread that was evaluating `unsafeRunSync` was one of the ones from the ZIO thread pool.

tusharmath:
Just a thought: Why not use `Promise.then()` in Scala.js to schedule things instead of `setImmediate`.

That will actually guarantee that the callback scheduled will execute as soon as the current call stack is completed.

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -103,37 +103,8 @@ trait Runtime[+R] {
...
-      supervisor.unsafeOnStart(environment, zio, None, context)
-
-      context.onDone(exit => supervisor.unsafeOnEnd(exit, context))
-    }
-
-    context.evaluateNow(ZIOFn.recordStackTrace(() => zio)(zio.asInstanceOf[IO[E, A]]))
-    context.runAsync(k)
-
-    fiberId => unsafeRun(context.interruptAs(fiberId))
+    lazy val curZIO = if (Platform.isJS) zio else ZIO.yieldNow *> zio
```
sideeffffect, line 106: isn't the situation of JavaScript here similar to that of Native?

adamgfraser: I'm not sure what the situation is on Native here. At this point we have no tests for Native and there is an ongoing effort to get those working with associated changes so my overall point of view is we address Native specific issues as part of that but could definitely flip this to only yield on the JVM.

PR #4121, opened 2020-08-20
Prevent Interruption While Closing Scopes

### LINKED ISSUE #4116: Auto-supervision doesn't wait for nested fibers
Consider the following snippet: 
```scala
for {
  fiber <- (latch.succeed(()) *> ZIO.sleep(10 seconds)).uninterruptible.fork.fork
  _     <- latch.await
  _     <- fiber.interrupt
} yield ExitCode.success
```
Despite entering the uninterruptible region (which is ensured by using the latch), when we interrupt the fiber it stops immediately without waiting. If you remove one of the `.fork`, it waits 10 seconds as expected. It seems isn't waiting on the interruption of the child fiber for the interruption of the parent to return.

adamgfraser:
Resolves #4116.

It looks like the issue here is that we are getting interrupted when attempting to close the scope of the child fiber. The child fiber tries to close its scope after it finishes forking the grandchild but then is externally interrupted. As a result, the finalizers in its scope are never run and so we do not wait for the interruption of the grandchild to return.

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -944,7 +944,7 @@ private[zio] final class FiberContext[E, A](
          * We are not done yet, because we have to close the scope of the fiber.
          */
         if (!state.compareAndSet(oldState, Executing(oldStatus.toFinishing, observers, interrupted))) done(v)
-        else openScope.close(v) *> ZIO.done(v)
+        else openScope.close(v).uninterruptible *> ZIO.done(v)
 
       case Done(_) => null // Already done
     }
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -944,7 +944,10 @@ private[zio] final class FiberContext[E, A](
          * We are not done yet, because we have to close the scope of the fiber.
          */
         if (!state.compareAndSet(oldState, Executing(oldStatus.toFinishing, observers, interrupted))) done(v)
-        else openScope.close(v).uninterruptible *> ZIO.done(v)
+        else {
+          setInterrupting(true)
+          openScope.close(v) *> ZIO.done(v)
+        }
 
       case Done(_) => null // Already done
     }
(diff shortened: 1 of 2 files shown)


ghostdogpr (approved):

jdegoes:
@adamgfraser Let's pair on this since I think it will be quite tricky.

adamgfraser:
@jdegoes That would be great!

jdegoes (approved):

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -963,7 +966,7 @@ private[zio] final class FiberContext[E, A](
       val oldState = state.get
 
       oldState match {
-        case Executing(Status.Suspended(oldStatus, true, _, _, _), observers, interrupted) =>
+        case Executing(Status.Suspended(oldStatus, true, _, _, _), observers, interrupted) if !oldState.interrupting =>
```
jdegoes, line 969: There is a race condition in `if !oldState.interrupting`. I think it's ok. 

We may want to add a test for this if one doesn't exist, basically that async finalizers during winddown can't be interrupted. Could be tested with a latch.

jdegoes: `kill` is called from other fibers, so what happens if the fiber itself changes the state of `setInterrupting` to false at the same time this check is happening. In this case, this code here (in `kill`) will make a different decision about what to do based on the changing boolean value.

adamgfraser: Right. Then `state` would not be equal to `oldState` and we would loop and then on the next loop `interrupting` would be `false` and we would interrupt. But isn't that the behavior we want? You're definitely right there can be a race between setting the interrupting status and interrupting but I think with this whatever the interrupting status is when the `compareAndSet` gets evaluated will control which seems correct.

adamgfraser: Added a test for this.

jdegoes: Ah, you're right, because you are looking at `oldState.interrupting`, which will be consistent with the state. 

Good with me. 👍

jdegoes (approved):

PR #4131, opened 2020-08-25
unsafeRunAsyncCancelable takes Scala Native into account

sideeffffect:
context: https://github.com/zio/zio/pull/4120/files#r476702764

### DIFF
--- a/core/shared/src/main/scala/zio/Runtime.scala
+++ b/core/shared/src/main/scala/zio/Runtime.scala
@@ -104,10 +104,7 @@ trait Runtime[+R] {
    */
   final def unsafeRunAsyncCancelable[E, A](zio: => ZIO[R, E, A])(k: Exit[E, A] => Any): Fiber.Id => Exit[E, A] = {
     lazy val curZio = if (Platform.isJVM) ZIO.yieldNow *> zio else zio
-      if (Platform.isJS) zio
-      else if (Platform.isNative) zio
-      else ZIO.yieldNow *> zio
-    val canceler = unsafeRunWith(curZio)(k)
+    val canceler    = unsafeRunWith(curZio)(k)
     fiberId => {
       val result = internal.OneShot.make[Exit[E, A]]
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -103,8 +103,11 @@ trait Runtime[+R] {
    * This method is effectful and should only be invoked at the edges of your program.
    */
   final def unsafeRunAsyncCancelable[E, A](zio: => ZIO[R, E, A])(k: Exit[E, A] => Any): Fiber.Id => Exit[E, A] = {
-    lazy val curZio = if (Platform.isJS) zio else ZIO.yieldNow *> zio
-    val canceler    = unsafeRunWith(curZio)(k)
+    lazy val curZio =
```
adamgfraser, line 106:
```suggestion
    lazy val curZio = if (Platform.isJVM) ZIO.yieldNow *> zio else zio
```

core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -103,8 +103,11 @@ trait Runtime[+R] {
    * This method is effectful and should only be invoked at the edges of your program.
    */
   final def unsafeRunAsyncCancelable[E, A](zio: => ZIO[R, E, A])(k: Exit[E, A] => Any): Fiber.Id => Exit[E, A] = {
-    lazy val curZio = if (Platform.isJS) zio else ZIO.yieldNow *> zio
-    val canceler    = unsafeRunWith(curZio)(k)
+    lazy val curZio = if (Platform.isJVM) ZIO.yieldNow *> zio else zio
+      if (Platform.isJS) zio
```
adamgfraser, line 107: I think we need to delete these lines.

sideeffffect: :heavy_check_mark:

PR #4141, opened 2020-08-28
Make ZStream#aggregateAsyncWithin fail fast

### LINKED ISSUE #4039: AggregateAsyncWithin does not fail fast - it goes to next element before failing
Hi I have a question RE ZStream's `AggregateAsyncWithin` combinator.
```scala
object AggregateAsyncWithinTest extends DefaultRunnableSpec {

  override def spec =
    suite("AggregateAsyncWithinTest")(
      testM("test Foo") {
        val s       = ZStream(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
        val program = s
          .tap(i => ZIO.when(i == 6)(ZIO.fail("BOOM!")) *> putStrLn(s"processed $i"))
          .aggregateAsyncWithin(ZTransducer.foldUntil((), 5)((_, _) => ()), Schedule.fixed(5.seconds))
          .tap(_ => putStrLn("YYYYYYYYYYYYYYYYY"))
        for {
          _ <- program.runDrain
        } yield assertCompletes
      }
    )

}
/* OUTPUT
processed 1
processed 2
processed 3
processed 4
processed 5
YYYYYYYYYYYYYYYYY
processed 7
*/
```
In the above i would not expect to see `processed 7` after failure processing element 6 - ie I would expect fail fast behaviour. Is this expected behaviour?
Version of ZIO used is ` 1.0.0-RC21-2+37-be08a9e6-SNAPSHOT`

luis3m:
Closes #4039

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -203,6 +203,19 @@ object ZStreamSpec extends ZIOBaseSpec {
               equalTo(Chunk(Right(List(2, 1, 1, 1, 1)), Right(List(2))))
             )
           },
+          testM("fails fast") {
+            for {
+              queue <- Queue.unbounded[Int]
+              _ <- ZStream
+                     .range(1, 10)
+                     .tap(i => ZIO.fail("BOOM!").when(i == 6) *> queue.offer(i))
+                     .aggregateAsyncWithin(ZTransducer.foldUntil((), 5)((_, _) => ()), Schedule.forever)
+                     .runDrain
+                     .catchAll(_ => ZIO.succeedNow(()))
+              value <- queue.takeAll
+              _     <- queue.shutdown
+            } yield assert(value)(equalTo(List(1, 2, 3, 4, 5)))
+          } @@ zioTag(errors),
           testM("error propagation 1") {
             val e = new RuntimeException("Boom")
             assertM(
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -230,7 +230,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
                           .makeManaged[Option[Fiber[Nothing, Take[E1, O]]]](None)
         sdriver   <- schedule.driver.toManaged_
         lastChunk <- ZRef.makeManaged[Chunk[P]](Chunk.empty)
-        producer   = Take.fromPull(pull).repeatWhileM(take => handoff.offer(take).as(!take.isDone))
+        producer   = Take.fromPull(pull).repeatWhileM(take => handoff.offer(take).as(take.isSuccess))
         consumer = {
           // Advances the state of the schedule, which may or may not terminate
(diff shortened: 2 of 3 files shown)


streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -203,6 +203,19 @@ object ZStreamSpec extends ZIOBaseSpec {
               equalTo(Chunk(Right(List(2, 1, 1, 1, 1)), Right(List(2))))
             )
           },
+          testM("fails fast") {
+            for {
+              queue <- Queue.unbounded[Int]
+              _ <- ZStream
+                     .range(1, 10)
+                     .tap(i => ZIO.fail("BOOM!").when(i == 6) *> queue.offer(i))
+                     .aggregateAsyncWithin(ZTransducer.foldUntil((), 5)((_, _) => ()), Schedule.fixed(5.seconds))
```
iravid, line 212: This could very well use `Schedule.forever` as we're not advancing the test clock anywhere, right?

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -203,6 +203,19 @@ object ZStreamSpec extends ZIOBaseSpec {
...
+                     .runDrain
+                     .catchAll(_ => ZIO.succeedNow(()))
+              value <- queue.takeAll
+              _     <- queue.shutdown
+            } yield assert(value)(equalTo(List(1, 2, 3, 4, 5)))
+          } @@ zioTag(errors),
```
iravid, line 218: Can you do a few runs with `nonFlaky(50000)` locally to verify the test is not flaky? Some parts of it (usage of TestClock with no advancing, use of `takeAll` which is not deterministic) make me a bit wary.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -230,7 +230,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
                           .makeManaged[Option[Fiber[Nothing, Take[E1, O]]]](None)
         sdriver   <- schedule.driver.toManaged_
         lastChunk <- ZRef.makeManaged[Chunk[P]](Chunk.empty)
-        producer   = Take.fromPull(pull).repeatWhileM(take => handoff.offer(take).as(!take.isDone))
+        producer   = Take.fromPull(pull).repeatWhileM(take => handoff.offer(take).as(take.isSuccess))
```
iravid, line 233: Great fix 👍

iravid (commented):
Great fix. I have a small concern about potential flakiness in the test. We can merge afterwards.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: I just ran it locally with `nonFlaky(50000)` ten times. No failures.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
luis3m: True that, changed.

iravid (approved):
Thank you @luis3m!

luis3m:
@iravid ready for review again

PR #4142, opened 2020-08-28
Implement ZIO#forkInternal

### LINKED ISSUE #4136: zipPar fiber failures not handled correctly
When using `zipPar`, if one of the `ZIO` contains a fiber failure, said failure will be seen as an "unchecked error" even if the error has been handled.

```scala
val a = ZIO.succeed(1) zipPar ZIO.dieMessage("Error!") // zipPar with a fiber failure
val res = a.sandbox.catchAll(_ => UIO(1)) // Failure is handled however we get a log message stating
/*
Fiber failed.
*/
```

[Reproduced issue](https://scastie.scala-lang.org/GAQgcoemTYqn3Sl3nCvlHQ)

adamgfraser:
Resolves #4136.

We had actually resolved this issue previously in #1949 but removed it when we stopped logging pure interruptions, which inadvertently caused a regression because fiber failures other than interruption can subsequently be handled at a higher level of the application.

The basic problem here is that a fiber doesn't know who will join it or what they will do once joining it, so a fiber has to either report the error or not. Thus, when we fork a fiber if we intend to handle its errors ourselves we need to signal to the fiber that it does not need to log fiber failures. We can do this with calling `run` on the effect we are forking and then using `flatMap(IO.done(_))` to flatten the exit value back into an effect. The implementation in `raceWith` is slightly more complicated now that it has been moved into `FiberContext` but conceptually the same.

I explored an alternative solution where we pushed more of this logic into `FiberContext` and each `FiberContext` had an error handler that could be overridden that we could explore if we wanted.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -735,6 +735,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def forkDaemon: URIO[R, Fiber.Runtime[E, A]] = forkIn(ZScope.global)
 
   /**
+   * Forks an effect that will be executed without unhandled failures being
+   * reported. This is useful for implementing combinators that handle failures
+   * themselves.
+   */
+  final def forkInternal: ZIO[R, Nothing, Fiber[E, A]] =
+    run.fork.map(_.mapM(IO.done(_)))
+
+  /**
    * Forks the fiber in a [[ZManaged]]. Using the [[ZManaged]] value will
    * execute the effect in the fiber, while ensuring its interruption when
    * the effect supplied to [[ZManaged#use]] completes.
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -578,9 +578,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
     filterOrElse_[R, E1, A](p)(ZIO.fail(e))
 
   /**
-   * Returns an effect that races this effect with all the specified effects,
-   * yielding the value of the first effect to succeed with a value.
-   * Losers of the race will be interrupted immediately
+   * Returns an effect that runs this effect and in case of failure,
+   * runs each of the specified effects in order until one of them succeeds.
    */
   final def firstSuccessOf[R1 <: R, E1 >: E, A1 >: A](rest: Iterable[ZIO[R1, E1, A1]]): ZIO[R1, E1, A1] =
     ZIO.firstSuccessOf(self, rest)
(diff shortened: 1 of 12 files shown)


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -234,29 +234,79 @@ private[zio] final class FiberContext[E, A](
 
     val raceIndicator = new AtomicBoolean(true)
 
-    val left  = fork[EL, A](race.left.asInstanceOf[IO[EL, A]], race.scope)
-    val right = fork[ER, B](race.right.asInstanceOf[IO[ER, B]], race.scope)
+    val left  = fork(race.left.asInstanceOf[IO[EL, A]].traced.run.untraced, race.scope)
```
jdegoes, line 237: This will add a lot of performance overhead to such a low level operation. Any way to cheat?

adamgfraser: The tracing or the running? The tracing probably yes. The running will I think require some changes in `FiberContext` but I think they can be relatively straightforward.

adamgfraser: Take a look at this version. Should be more efficient.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4169,7 +4177,8 @@ object ZIO extends ZIOCompanionPlatformSpecific {
 
   private[zio] final class Fork[R, E, A](
     val value: ZIO[R, E, A],
-    val scope: Option[ZScope[Exit[Any, Any]]]
+    val scope: Option[ZScope[Exit[Any, Any]]],
+    val reportFailure: Option[Cause[Any] => Unit]
```
adamgfraser, line 4181: Adding an additional field here is a binary incompatible change even though this is package private. We could potentially revert and use the previous implementation of `forkInternal` for everything except the `raceWith` implementation.

iravid: Safe to mark as an exclusion I think

adamgfraser: I think so too. Will do.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -234,8 +235,8 @@ private[zio] final class FiberContext[E, A](
 
     val raceIndicator = new AtomicBoolean(true)
 
-    val left  = fork[EL, A](race.left.asInstanceOf[IO[EL, A]], race.scope)
-    val right = fork[ER, B](race.right.asInstanceOf[IO[ER, B]], race.scope)
+    val left  = fork[EL, A](race.left.asInstanceOf[IO[EL, A]], race.scope, Some(_ => ()))
+    val right = fork[ER, B](race.right.asInstanceOf[IO[ER, B]], race.scope, Some(_ => ()))
```
jdegoes, line 239: Factor out `Some(_ => ())` into a constant of `FiberContext` companion object.

PR #4147, opened 2020-08-29
Fix delay computation in Schedule.fixed

### LINKED ISSUE #4144: ZStream#repeat doesn't repeat a stream evenly. ZStream#schedule is probably broken either.
This code is supposed to repeat the entire stream once per 10 sec, but in fact after 10 sec it repeats the stream unpredictable number of times/delays and only after that starts working as expected :

     ZStream.fromIterable(1 to 10)
            .repeat(Schedule.fixed(10.seconds))
            .foreachChunk(e => putStrLn(s"E $e"))

It seems that `ZStream#schedule(Schedule.fixed(10.seconds))` is working not very smoothly either. At the beginning of the stream It produces elements of the stream somehow strange, not evenly. Seems to be the same source of the issue as for `repeat`.

Tested with ZIO 1.0.1

iravid:
This commit fixes two issues:

1. The state maintained by `fixed` keeps the schedule's start time and  the last run. The last run was wrongly initialized to the start time  instead of `start + interval`. This caused a redundant immediate repetition when starting.
2. The delay computation was wrongly computed as `(now - start) %  interval`. For example, given `start = 0, now = 5000,    intervalMillis = 3000`, the delay would be `2000` instead of `1000`.

   This is fixed by computing the delay as `interval - ((now - start) % interval)`.

Resolves #4144.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -1,6 +1,7 @@
 package zio
 
-import scala.concurrent.Future
+import java.time.{ Instant, OffsetDateTime, ZoneId }
+
 import zio.clock.Clock
 import zio.duration._
 import zio.stream.ZStream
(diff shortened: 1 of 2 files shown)


jdegoes:
Awesome fixes!

Issue #3841, opened 2020-06-20
`ZStream#fromIterator(Managed)` should have an optional `chunkSize` parameter


simpadjo:
[Good first issue]
`ZStream#fromIterator` creates a task for pulling each individual element. This makes it infeasible for low-level tasks.

It can be worked around by calling `ZStream.fromIterator(it.grouped(1000))`.
Explicitly adding `chunkSize` parameter would save users from accidentally hitting this problem.

simpadjo:
How to undo an accidental thumbsup on my own issue?  :rofl:

lvitaly:
@simpadjo @iravid hey, can I take it?

simpadjo:
Sure!

ogresto:
'This job has been blocked because resource-class:large is not available on your plan. Please upgrade to continue building.' 

 could you please upgrade ci plan ?

emilhotkowski:
@adamgfraser Looks fixed

PR #4152, opened 2020-08-29
3841 `ZStream#fromIterator(Managed)` should have an optional `chunkSize`

ogresto:
…arameter #3841

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3434,7 +3434,7 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
       .fromEffect(Task(iterator.grouped(chunkSize)) <*> ZIO.runtime[Any])
       .flatMap {
         case (it, rt) =>
-          ZStream.repeatEffectOption {
+          ZStream.repeatEffectChunkOption {
             Task {
               val hasNext: Boolean =
                 try it.hasNext
@@ -3444,7 +3444,7 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
                 }
 
               if (hasNext) {
-                try it.next()
+                try Chunk.fromIterable(it.next())
                 catch {
                   case e: Throwable if !rt.platform.fatal(e) =>
                     throw e
@@ -3456,7 +3456,6 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
             }
           }
       }
-      .mapConcat(identity)
   }
 
   /**
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3438,7 +3438,6 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
               catch {
                 case e: Throwable if !rt.platform.fatal(e) => throw e
               }
-
             if (hasNext) {
               try Chunk.fromIterable(it.next())
               catch {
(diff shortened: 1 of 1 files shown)


ogresto:
https://github.com/zio/zio/issues/3841

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3424,33 +3424,39 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
-        ZStream.repeatEffectOption {
-          Task {
-            val hasNext: Boolean =
-              try it.hasNext
-              catch {
-                case e: Throwable if !rt.platform.fatal(e) =>
-                  throw e
-              }
+    ZStream
+      .fromEffect(Task(iterator.grouped(chunkSize)) <*> ZIO.runtime[Any])
```
iravid, line 3434: Great use of `grouped` 👍

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3424,33 +3424,39 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+      .flatMap {
+        case (it, rt) =>
+          ZStream.repeatEffectOption {
```
iravid, line 3437: Let's use `repeatEffectChunkOption` - should be slightly more efficient

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3424,33 +3424,39 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
-              catch {
-                case e: Throwable if !rt.platform.fatal(e) =>
-                  throw e
-              }
-            } else throw StreamEnd
-          }.mapError {
-            case StreamEnd => None
-            case e         => Some(e)
+              if (hasNext) {
+                try it.next()
```
iravid, line 3447: You can do `Chunk.fromIterable(it.next())` here

iravid (commented):
Looks great! Some minor efficiency comments.

ogresto:
@iravid Why do we need an optional `chunkSize` parameter if ZStream offers inherent laziness (as the documentation says) ?

iravid:
> @iravid Why do we need an optional `chunkSize` parameter if Iterator is already lazy?

If I understand the question, Chunks are used in ZStream for performance reasons. Most pure operations benefit from being performed on chunk in a tight loop.

ogresto:
'This job has been blocked because resource-class:large is not available on your plan. Please upgrade to continue building.' @iravid @simpadjo could you please upgrade ci plan ?

ogresto:
> > @iravid Why do we need an optional `chunkSize` parameter if Iterator is already lazy?
> 
> If I understand the question, Chunks are used in ZStream for performance reasons. Most pure operations benefit from being performed on chunk in a tight loop.

i am not sure this is a problem, but 'grouped' iterator is not lazy anymore (under the hood it initiate values batch-by-batch as an array). And it breaks some tests (for example 
`testM("toIterator") {
          (for {
            counter  <- Ref.make(0).toManaged_ //Increment and get the value
            effect    = counter.updateAndGet(_ + 1)
            iterator <- ZStream.repeatEffect(effect).toIterator
            n         = 2000
            out <- ZStream
                     .fromIterator(iterator.map(_.merge))
                     .mapConcatM(element => effect.map(newElement => List(element, newElement)))
                     .take(n.toLong)
                     .runCollect
                     .toManaged_
          } yield assert(out)(equalTo(Chunk.fromIterable(1 to n)))).use(ZIO.succeed(_))
        }`
)

adamgfraser:
@ogresto That warning is being generated because it looks like the CI is being run on your Circle CI account instead of the one for the ZIO organization. We've seen this a couple times before. @softinio Do you remember how we resolved this?

iravid:
@ogresto Could you try the following:
- Close this PR and open a new one
- Rename your fork of ZIO from `fzio` to `zio`

ogresto:
@iravid it seems it doesn't work

PR #4155, opened 2020-08-30
fromOptionValue operator

politrons:
I know with the migration of Scalaz to dev.zio the fromOption operator was refactor to allow use composition and other goodies to use Option instead the wrapped value.

Still in my team and other teams in my organization they claim they still want to have an operator to extract the value from the option once we transform into IO. And they will like to have still an operator IO[Unit, A] 

That's why I bring back the feature we had in scalaz here with the tmp name **fromOptionValue** that return **IO[Unit, A]**

Again I understand the reasons why the operator was refactor, and I'm just bringing another alternative for those that dont want use the goodies of the new operator.

This MR is still work in progress, and only in case it's consider I will continue working with documentation and all the necessary stuff


adamgfraser:
@politrons I think it definitely makes sense to have a variant like this. Not sure about the name. Right now the diff is pretty huge. I think maybe there is an inconsistency with the `scalafmt` settings in the project. Can you check that and try to get it down to just the differences related to this change?

politrons:
@adamgfraser sure I used scalafmt format instead my default google one, but it's true that there're format change in the code.
Let me tackle that. And regarding the name, not sure what it could fit for this operator **fromOptionValue** was the first thing that cross my mind 🤣

adamgfraser:
You should just be able to run `sbt fmt` to get everything formatted the right way. Maybe `getOrFailUnit`?

politrons:
Since it's an operator to transform the option into IO it's not better keep the standard name "fromOption" as part of the name of the operator?. What about **fromOptionOrFailUnit**?

I change the format into scalafmt and it seems some files that were formatted has been reverted but for some reason in ZIO file it keeps the new line limit different. Using as you suggest `sbt fmt` in the whole project it did not make any change, it'w weird

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3104,6 +3104,9 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def fromOption[A](v: => Option[A]): IO[Option[Nothing], A] =
     effectTotal(v).flatMap(_.fold[IO[Option[Nothing], A]](fail(None))(succeedNow))
 
+  def fromOptionOrFailUnit[A](v: => Option[A]): IO[Unit, A] =
```
adamgfraser, line 3107: Should we call `getOrFailUnit` for symmetry with existing `getOrFail` constructor?

adamgfraser:
I just thought it either got the optional value or failed with the unit value so `getOrFailUnit`. It is similar to the `getOrFail` constructor we have that fails with `NoSuchElementException` if the option is empty, which is another variant like this that just uses another value to denote the option not existing that is more useful in some situation.

adamgfraser (dismissed):
Looks great!

politrons:
@adamgfraser thanks mate, it seems I still have issues with the format of ZIO file, since the pipeline it fail.  I just change in my Intellij in `CodeStyle -> Scala -> Formatter` and I just set the `scalafmt`  then I just reformat the `ZIO` class.
Any idea if I have to do something else?

adamgfraser:
@politrons I would suggest just using the `fmt` command from `sbt` which is the same one that ZIO uses and the formatting check in CI is based on. Is that working for you?

politrons:
yes I did that, and by command line it works fine. I think I solved, It might be because I had marked in my IntelliJ the option `Use IntelliJ formatter for code range formatting`

let see if the pipeline pass and marked as approved if you consider is good enough again

politrons:
@adamgfraser hi again, just one question. I made the first two commits with a wrong credentials in my `git config` file so instead being marked as commits of `politrons`, which is my GitHub user, it was done as `Pablo Perez Garcia` which is not a GitHub user. So now I have the problem  that I cannot sign the `CLA` for that username, I only was able to do it for the rest of commits of politrons. I tried to revert those commits but since the code it was not altered in the last commits GitHub told me it's not necessary revert anything

Any idea how to solve this?.

Thanks for all the support!

adamgfraser:
No problem at all! Thanks for contributing!

There are two things you can do. First is add that e-mail address to your Github profile. If you click on the link on the post by the CLA not there is some more detail on how to do that. Second if that gets to be too much of a hassle is close this, open a fresh PR, and we can merge that.

PR #4158, opened 2020-08-30
Explore Alternatives for ExecutionContext on Scala.js

adamgfraser:


jdegoes:
@adamgfraser This was no faster than the other one, and still gives us timeouts??

regiskuckaertz:
Have you tried `process.nextTick` on Node? or in the browser [using this technique](http://blog.millermedeiros.com/promise-nexttick/).

adamgfraser:
@jdegoes I think the issue here is now how it is interacting with tasks that are scheduled using `Clock`. Right now if we still have pending tasks in the queue we keep executing them so tasks scheduled with the `Clock` that are ready to be run don't have a chance to be executed until all the tasks in the queue are done. I think we need one `ExecutionContext` on Scala.js for both immediate and scheduled tasks that maintains two separate queues for immediate and scheduled tasks and then if there is a task in the scheduled tasks queue we run that otherwise we keep running tasks from the normal queue.

adamgfraser:
Okay, so to summarize the status here. The basic problem we are dealing with is that on Scala.js we currently don't yield to other pending tasks when we submit a new task to the executor. Exacerbating this is the fact that yielding seems to be significantly more expensive on Scala.js relative to the JVM.

As a result, the approach I have been pursuing in this PR is to continue executing in the current frame unless a yield is explicitly called using `ZIO.yieldNow`. We also previously added `ZIO.yieldNow` in repetition operators to prevent busy loops.

This generally works and at this point I have all tests passing with these changes but it raises two issues.

First, at times this creates significant tension between fairness and performance. For example, consider the `ZStream.forever` operator. It seems like we want to do a yield here to allow other fibers to make progress between pulling from the stream. But `ZStream.forever` is used in all of the `check` methods in ZIO Test, so yielding results in a yield between every generated value, which is not really what we want in most cases where generators are just generating random numbers and not doing real interaction with the outside world, and is fairly catastrophic for performance in this case.

The second issue is more particular to our test suite. Consider a test like:

```scala
testM("shard example") {
  def shard[R, E, A](queue: Queue[A], n: Int, worker: A => ZIO[R, E, Unit]): ZIO[R, E, Nothing] = {
    val worker1: ZIO[R, E, Unit] = queue.take.flatMap(a => worker(a).uninterruptible).forever
    ZIO.forkAll(List.fill(n)(worker1)).flatMap(_.join) *> ZIO.never
  }
  for {
    queue <- Queue.unbounded[Int]
    _     <- queue.offerAll(1 to 100)
    worker = (n: Int) => if (n == 100) ZIO.fail("fail") else queue.offer(n).unit
    exit  <- shard(queue, 4, worker).run
    _     <- queue.shutdown
  } yield assert(exit)(fails(equalTo("fail")))
}
```

This needs to yield after pulling every value from the queue because of the `forever` operator, so this test will yield 100 times. This is conceptually fine, but it has the effect that with many other tests executing between each yield, this test may not be able to complete execution within the one minute timeout. We could increase the timeout or I think it may make sense to just run tests sequentially on Scala.js because everything we are doing is CPU bound and the underlying platform is single threaded.

PR #4160, opened 2020-08-30
 parameter #3841 patch

alek-w:


ogresto:
`@ogresto Could you try the following:

Close this PR and open a new one
Rename your fork of ZIO from fzio to zio`

@iravid it seems it doesn't work

iravid:
Very strange. Pulling at straws here.. can you try deleting the remote branch and re-pushing it?

`git push origin :patch-3841`

adamgfraser:
Seen this a couple of times before when contributors have their own Circle CI accounts. We may need @softinio to help resolve.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3444,25 +3444,24 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
 
-  def fromIterator[A](iterator: => Iterator[A]): ZStream[Any, Throwable, A] = {
+  def fromIterator[A](
+    iterator: => Iterator[A],
+    chunkSize: Int = ZStream.DefaultChunkSize
+  ): ZStream[Any, Throwable, A] = {
     object StreamEnd extends Throwable
-
-    ZStream.fromEffect(Task(iterator) <*> ZIO.runtime[Any]).flatMap {
+    ZStream.fromEffect(Task(iterator.grouped(chunkSize)) <*> ZIO.runtime[Any]).flatMap {
```
CptnKirk, line 3453: Wouldn't it be more efficient to avoid constructing the intermediate `List` from Iterator.grouped() and instead manually `take()` one by one, appending to a `Chunk` until `hasNext == false` or `chunkSize` is reached?

CptnKirk: Although this is essentially running the initial implementation through a `ZTransducer.collectAllN`. isn't it?  Should the new impl simply be `original >>> ZTransducer.collectAllN(chunkSize)`?

ogresto:
@adamgfraser @softinio it seems ci doesnt work (

ogresto:
@iravid could you pls suggest how to solve the problem with the broken laziness ZStream#fromIterator in case no creation of tasks per each item in the iterator ?

PR #4161, opened 2020-08-30
`ZStream#fromIterator(Managed)` should have an optional `chunkSize` parameter #3841 patch

ogresto:


iravid:
Still running on your account apparently

PR #4165, opened 2020-08-31
ZIO.fromPromise operator

politrons:
I don't know if this operator it has any sense in the ZIO ecosystem, I know we can achieve the same goal of lazy evaluation of futures using fibers. But maybe is not a bad alternative, to provide an operator just like `fromFuture`, but `fromPromise` as inter-opt operator for those scala developers, that want to use `promises` for lazy evaluation of futures.

#### Description:

Imports a [[scala.concurrent.Promise[A]] an input value of B and a function to apply in [complete] function of promise.
We complete the promise in another thread avoiding blocking the function execution, and in the main thread we generate a future from promise, and we pass to [fromFuture] to transform into ZIO[Any, Throwable, A]. 

With this operator we allow use Scala promise to have a lazy evaluation of the logic that it will be executed in another thread only when the whole program is evaluated.


politrons:
In this PR there're also some code from a previous PR (https://github.com/zio/zio/pull/4163) that is pending to being merged once @adamgfraser approve it

adamgfraser:
@politrons Can't the use case of lazy evaluation be addressed by `ZIO.effect(func(input))`? `ZIO.effect` and `ZIO.effectTotal` are both lazy so you can wrap a computation in them and that computation won't be evaluated until the program is executed by the ZIO runtime.

If we want something just for interop I think we would want it to just take a `Promise` and it to be the responsibility of the caller to complete that Promise themselves separately. Take a look at `fromPromiseJS` for an example of that. Maybe we could have something similar like `fromPromiseScala` since we also want to make clear that this is referring to a Scala promise and not ZIO's own promise.

politrons:
@adamgfraser Hi again,

 If I understand your comment, I just extract the logic of execution and completing the promise out of the operator. So basically we just extract the future from the promise. 

Not sure to be honest if now this is a useful operator.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3059,6 +3058,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
       }
     }
 
+  /**
+   * Imports a [[scala.concurrent.Promise]] we generate a future from promise,
+   * and we pass to [fromFuture] to transform into ZIO[Any, Throwable, A]
+   */
+  def fromPromiseScala[A, B](promise: scala.concurrent.Promise[A]): ZIO[Any, Throwable, A] =
```
adamgfraser, line 3065: You can drop the `B` type parameter here and return a `Task[A]`.

adamgfraser:
@politrons Hey! 😃

Yes I think so. That gives the caller the flexibility to complete the promise where they want to. If this is being used for interoperability the code that completes the promise may be impure and in another part of the application where it may not be possible to just import it into the ZIO runtime.

adamgfraser:
Can you merge upstream changes?

core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
```diff
@@ -293,6 +295,73 @@ object StackTracesSpec extends DefaultRunnableSpec {
         assert(cause.traces.head.stackTrace.size)(equalTo(3)) &&
         assert(cause.traces.head.stackTrace.exists(_.prettyPrint.contains("selectHumans")))(isTrue)
       }
+    },
+    testM("basic option test") {
```
adamgfraser, line 299: Can we move these from `StackTracesSpec` to `ZIOSpec`? This file should only be for tests of the tracing functionality which shouldn't be impacted here or by your prior PR.

politrons: sure, I was not sure where to add those scenarios, but indeed the file name of the test was not matching pretty well. I will change it in this PR

core-tests/jvm/src/test/scala-2.12/zio/StacktracesSpec.scala
```diff
@@ -5,7 +5,7 @@ import zio.duration._
 import zio.internal.stacktracer.ZTraceElement
 import zio.internal.stacktracer.ZTraceElement.{ NoLocation, SourceLocation }
 import zio.test.Assertion._
-import zio.test._
+import zio.test.{ testM, _ }
```
adamgfraser, line 8: I think we can revert this change now?

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -11,7 +10,7 @@ import zio.internal.Platform
 import zio.random.Random
 import zio.test.Assertion._
 import zio.test.TestAspect.{ flaky, forked, ignore, jvm, jvmOnly, nonFlaky, scala2Only }
-import zio.test._
+import zio.test.{ suite, _ }
```
adamgfraser, line 13: No need to import one member and then do wildcard import.

politrons: 👍

politrons:
@adamgfraser I already delete those test when I moved to ZIOSpec

adamgfraser:
@politrons Looks like the ones from when we added getOrFailOption are still there unless I am looking at the wrong version?

politrons:
@adamgfraser I made in this commit https://github.com/zio/zio/pull/4165/commits/ee7d8f4f34bb275cc4ec249fcb02c41204ccd2f7

politrons:
you welcome, thanks again for the support

PR #4166, opened 2020-08-31
Add a promise poll to ZStream#interruptWhen

### LINKED ISSUE #4118: ZStream.interruptWhen sometimes does not emit the failure from the promise
In the following example:

```scala
for {
  promise <- Promise.make[Throwable, Unit]
  _ <- promise.fail(new RuntimeException("x"))
  stream = 
    ZStream
      .fromIterable(0 to 100)
      .chunkN(8)
      .interruptWhen(promise)          
  result <- stream.runCollect.either    
  _ <- console.putStrLn(result.toString)
} yield ()
```
sometimes the stream interrupts without failure (but finishing before going through all the elements), in other cases it fails with the failure from the promise as expected. 

I was able to reproduce both behaviors by changing the chunk size.

### LINKED ISSUE #4119: ZStream.interruptWhen could do an initial poll on the promise
The first chunk of elements gets processed even if the promise passed to `interruptWhen` is already completed when constructing the stream. A possible improvement would be to do an initial poll on the promise to avoid this.

luis3m:
Closes #4118
Closes #4119

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -1739,7 +1739,7 @@ object ZStreamSpec extends ZIOBaseSpec {
                 halt <- Promise.make[String, Nothing]
                 _    <- halt.fail("Fail")
                 result <- ZStream(1)
-                            .haltWhen(halt)
+                            .interruptWhen(halt)
                             .runDrain
                             .either
               } yield assert(result)(isLeft(equalTo("Fail")))
@@ -1768,8 +1768,9 @@ object ZStreamSpec extends ZIOBaseSpec {
               for {
                 halt <- Promise.make[String, Nothing]
                 _    <- halt.fail("Fail")
-                result <- ZStream(1).forever
-                            .haltWhen(halt.await)
+                result <- ZStream
+                            .fromEffect(ZIO.never)
+                            .interruptWhen(halt.await)
                             .runDrain
                             .either
               } yield assert(result)(isLeft(equalTo("Fail")))
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1761,19 +1761,13 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   final def interruptWhen[E1 >: E](p: Promise[E1, _]): ZStream[R, E1, O] =
     ZStream {
       for {
-        as   <- self.process
-        done <- Ref.make(false).toManaged_
-        pull = done.get flatMap {
-                 if (_) Pull.end
-                 else
-                   as.raceFirst(
-                     p.await
-                       .mapError(Some(_))
-                       .foldCauseM(
-                         c => done.set(true) *> ZIO.halt(c),
-                         _ => done.set(true) *> Pull.end
-                       )
-                   )
+        as    <- self.process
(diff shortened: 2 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1761,19 +1761,13 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                     p.await
-                       .mapError(Some(_))
-                       .foldCauseM(
-                         c => done.set(true) *> ZIO.halt(c),
-                         _ => done.set(true) *> Pull.end
-                       )
-                   )
+        as    <- self.process
+        done  <- Ref.make(false).toManaged_
+        asPull = (p: Promise[E1, _]) => p.await.asSomeError.foldCauseM(ZIO.halt(_), _ => Pull.end) <* done.set(true)
```
iravid, line 1766: This can just be a value, right?
```suggestion
        asPull = p.await.asSomeError.foldCauseM(ZIO.halt(_), _ => Pull.end) <* done.set(true)
```

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -1739,7 +1739,7 @@ object ZStreamSpec extends ZIOBaseSpec {
                 halt <- Promise.make[String, Nothing]
                 _    <- halt.fail("Fail")
                 result <- ZStream(1)
-                            .haltWhen(halt)
+                            .interruptWhen(halt)
```
iravid, line 1742: 🤦

iravid:
This appears to break on Scala.js. Could you look into that?

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Lol, right.

luis3m:
@iravid done, `ZStream(1).forever.interruptWhen(halt.await)` was failing to fail on Scala.js apparently because of the initial overhead of `.disconnect`.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1761,19 +1761,13 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-                     p.await
-                       .mapError(Some(_))
-                       .foldCauseM(
-                         c => done.set(true) *> ZIO.halt(c),
-                         _ => done.set(true) *> Pull.end
-                       )
-                   )
+        as    <- self.process
+        done  <- Ref.makeManaged(false)
+        asPull = p.await.asSomeError *> done.set(true) *> Pull.end
```
svroonland, line 1766: Interesting. I compared with my [suggestion here](https://github.com/zio/zio/pull/2442/files#r355213169) and that one got #4119 right, but would also experience #4118

luis3m: @svroonland are you sure? it checks on every pull whether the promise is done and there's a unit test for that case.

svroonland: @luis3m Ah yes on closer look, you are right. So they would appear to be equivalent then..? (just for learning purposes, you can disregard this comment thread when it comes to this PR)

luis3m: @svroonland no worries. Yes, they are.

iravid:
Thank you @luis3m!

PR #4167, opened 2020-08-31
Make ZStream#buffer buffer single values

### LINKED ISSUE #3766: ZStream.buffer buffers chunks instead of single values

luis3m:
Closes #3766

A follow-up of #3803, `bufferSignal` was slightly modified so `bufferSliding` and `bufferDropping` can use it.


iravid (dismissed):
@luis3m Looks great. Sorry for the delay here. One request- we just merged an improvement to `ZStream.range` that makes it use chunks; can you rebase on master for this branch and just verify that your tests don't break because of that?

luis3m:
@iravid sorry, I somehow missed this comment. I just rebased this branch and made sure tests are passing.

luis3m:
@iravid this one as well

PR #4169, opened 2020-09-01
Don't transform to Chunks when using foreach in ZManaged

ghostdogpr:
I had a case with a large map and was seeing unnecessary time spent in `Chunk.fromIterable` in the profiler.


iravid (approved):
Good catch

PR #4171, opened 2020-09-01
Fix Memory Leak in FiberContext#await

### LINKED ISSUE #4135: Memory leak when racing join
When racing a join there is a memory leak. The source of the leak seems to be that the ZIO internal FiberState.observers list keeps on growing, which could be because the .join is never being completed/interrupted successfully. 

Heap when running

Simple code to reproduce can be found at https://github.com/ollyw/zio-memleak-reproducer

Tested with ZIO 1.0.0-RC21-2, 1.0.0 and 1.0.1

CC @adamgfraser

adamgfraser:
Resolves #4135.

The issue is that if fiber A awaits Fiber B, A would add itself as an observer to B with a callback to be invoked when B completes. However, if A is then subsequently interrupted that callback would never be removed from B's list of observers. This could result in a memory leak in situations where effects were waiting on a fiber and then being interrupted many times. We address by adding logic so that when fiber A awaits fiber B, interruption logic is also added so that if A is interrupted the callback is also removed from B's observer list.

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -783,10 +783,27 @@ private[zio] final class FiberContext[E, A](
   final def interruptAs(fiberId: Fiber.Id): UIO[Exit[E, A]] = kill0(fiberId)
 
   def await: UIO[Exit[E, A]] =
-    ZIO.effectAsyncMaybe[Any, Nothing, Exit[E, A]](
-      k => observe0(x => k(ZIO.done(x))),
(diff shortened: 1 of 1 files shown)


adamgfraser:
@jdegoes I added the test to #4121 for preventing interruption when closing scopes. Not sure how to test this one since I don't think we expose the list of observers of a fiber anywhere outside of `FiberContext`. I added a debug statement with the number of observers to verify that it was no longer growing in the original example. We could include `observers` as a method on runtime fibers but not sure if that is exposing too much of an implementation detail.

jdegoes:
Ok, sounds good!

ollyw:
Thanks for the fix and following up so fast @adamgfraser.

Issue #4043, opened 2020-08-03
Add some Schedule constructors for cron-like functionality


jdegoes:
These should utilize java.time where possible:

 - [ ] `Schedule.minuteOfHour(...)`
 - [ ] `Schedule.hourOfDay(...)`
 - [ ] `Schedule.dayOfWeek(....)`
 - [ ] `Schedule.dayOfMonth(...)`

adrianfilip:
I am taking this one.

jdegoes:
@adrianfilip 🙏

adrianfilip:
PR open: https://github.com/zio/zio/pull/4066

heaven-born:
I'm wondering how to deal with invalid input parameters here. Just throw an IllegalStateException exception?
Or create and pass something like "DayOfWeek" parameter objects instead.

heaven-born:
PR #4174 . Only `Schedule.minuteOfHour(...)`  in order to check that this approach is acceptable. There will be another PR for the rest of the methods.

heaven-born:
PR #4195 for `dayOfWeek`. 

`minuteOfHour`, `hourOfDay` and `dayOfWeek` were implemented in a scope of previous PR.

The last one `dayOfMonth` is tricky so it may take a while. If someone take take it before I do, I don't mind.

jdegoes:
@heaven-born Thank you for all your work on this ticket. It's a long-awaited addition to ZIO Schedule.

heaven-born:
PR #4212 . Failed checks are not related to functionality introduced in scope of this ticket.

Issue #3850, opened 2020-06-21
Schedule lags between occurs 1 and 2


kubukoz:
I'm sorry if the description is too specific, but I haven't dived into why this happens yet.

Given such a program, which prints the amounts of milliseconds since it started, on a schedule:

```scala

import zio._
import zio.console._
import zio.clock._
import zio.duration._
import java.util.concurrent.TimeUnit

object ZIOMain extends zio.App {
  def run(args: List[String]): URIO[ZEnv, ExitCode] = program.as(ExitCode.success)

  val program = clock.nanoTime.flatMap { startTime =>
    val newTime = clock.nanoTime.map(_ - startTime).map(_ / 1000000).map(_.toString)

    newTime.flatMap(a => ZIO.effectTotal(println(a))).repeat(Schedule.spaced(100.millis))
  }
}
```

On my machine it results in output like:

```
191
481
583
684
785
886
987
1089
1191
1293
1394
1495
...
```

While most numbers are greater than the previous ones by roughly 100, the difference between 191 and 481 is 290, so about 190ms of extra delay. I've been getting consistent results on this, and the difference is even more visible if the spacing (100ms) is decreased, or removed altogether (`Schedule.recurs`):

```
83
164
164
164
165
165
```

It doesn't happen when I repeat with manual recursion, or in `forever` (which doesn't use `repeat` but recurses as well).

Could it be just some overhead from initially constructing the schedule, which benefits from caching on later attempts? It feels a bit too repeatable, but maybe there are some optimizations there.

I'm running RC21 on graalvm CE 20.0.0.

jdegoes:
Thanks for reporting, will look into this soon!

Did you try `Schedule.fixed`?

kubukoz:
I just did, same thing:

```
80
236
236
242
251
261
272
282
292
301
311
322
332
341
351
361
371
381
392
401
412
421
432
```

(`Schedule.fixed(10.millis)`)

kubukoz:
Also, `Schedule.forever.addDelay(_ => 10.millis)` (which I guess is fewer combinators on top of the primitive implementation).

heaven-born:
From what I could find, the issue is most likely inside `ZIO#repeatOrElseEither`. It produces the original effect without any delay, and only after that starts using `schedule.driver` for scheduling. 

    self.foldM(   // <- problem is here
            e => ...
            a => loop(a) // <-- here scheduling begins
    )

[...]

PR #4174, opened 2020-09-02
Add some Schedule constructors for cron-like functionality

heaven-born:
`minuteOfHour`,  `secondOfMinute` and `hourOfDay`  from #4043

### DIFF
--- a/core/shared/src/main/scala/zio/Schedule.scala
+++ b/core/shared/src/main/scala/zio/Schedule.scala
@@ -1129,19 +1129,18 @@ object Schedule {
 
   /**
    * Cron-like schedule that recurs every specified `minute` of each hour.
-   * Exact second of the minute can be provided as optional `second` param.
+   * It triggers at zero second of the minute.
    * Producing a count of repeats: 0, 1, 2.
    */
-  def minuteOfHour(minute: Int, second: Int = 0): Schedule[Any, Any, Long] = {
+  def minuteOfHour(minute: Int): Schedule[Any, Any, Long] = {
     assert(minute < 60 && minute >= 0, s"Invalid minute parameter. Must be in range 0 ... 59")
-    assert(second < 60 && second >= 0, s"Invalid second parameter. Must be in range 0 ... 59")
 
     def loop(n: Long): StepFunction[Any, Any, Long] =
       (now: OffsetDateTime, _: Any) =>
         ZIO.succeed(
           Decision.Continue(
             n + 1,
-            nextFixedOffset(now, minute, ChronoField.MINUTE_OF_HOUR).withSecond(second),
+            nextFixedOffset(now, minute, ChronoField.MINUTE_OF_HOUR).withSecond(0),
             loop(n + 1L)
           )
         )
@@ -1149,7 +1148,7 @@ object Schedule {
     Schedule(loop(0L))
   }
 
-  def nextFixedOffset(currentOffset: OffsetDateTime, fixedTimeUnitValue: Int, timeUnit: ChronoField) = {
+  private[this] def nextFixedOffset(currentOffset: OffsetDateTime, fixedTimeUnitValue: Int, timeUnit: ChronoField) = {
     val fixedSec = currentOffset.`with`(timeUnit, fixedTimeUnitValue.toLong).withNano(0)
     if (currentOffset.get(timeUnit) <= fixedTimeUnitValue.toLong) fixedSec
     else fixedSec.plus(1, timeUnit.getRangeUnit)
--- a/core/shared/src/main/scala/zio/Schedule.scala
+++ b/core/shared/src/main/scala/zio/Schedule.scala
@@ -17,6 +17,7 @@
 package zio
 
 import java.time.OffsetDateTime
+import java.time.temporal.ChronoField
 import java.util.concurrent.TimeUnit
 
 import zio.duration._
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -1126,6 +1127,34 @@ object Schedule {
     Schedule(loop(None, 0L))
   }
 
+  /**
+   * Cron-like schedule that recurs every specified `minute` of each hour.
+   * Exact second of the minute can be provided as optional `second` param.
+   * Producing a count of repeats: 0, 1, 2.
+   */
+  def minuteOfHour(minute: Int, second: Int = 0): Schedule[Any, Any, Long] = {
```
jdegoes, line 1135: We should delete `second` and instead provide `secondOfMinute` as a separate constructor.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -1126,6 +1127,34 @@ object Schedule {
...
+    assert(minute < 60 && minute >= 0, s"Invalid minute parameter. Must be in range 0 ... 59")
```
jdegoes, line 1136: If you are going to throw here, you need to wrap the whole function in `ZIO.effectSuspend`, so they will be translated into `ZIO.die`.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -1126,6 +1127,34 @@ object Schedule {
...
+            n + 1,
+            nextFixedOffset(now, minute, ChronoField.MINUTE_OF_HOUR).withSecond(second),
+            loop(n + 1L)
+          )
+        )
+
+    Schedule(loop(0L))
+  }
+
+  def nextFixedOffset(currentOffset: OffsetDateTime, fixedTimeUnitValue: Int, timeUnit: ChronoField) = {
```
jdegoes, line 1152: I'd hide this, make it package private, as it shouldn't be used by end-users.

core/shared/src/main/scala/zio/Schedule.scala
heaven-born: that's kind of tricky. We can't wrap it with `ZIO.effectSuspend`, because the constructor is supposed to return `Schedule`. It doesn't seem that constructors are able to fail at all.

heaven-born: Probably solution for this and also for #3850 would be warm-up run of `next` function inside Schedule#driver with `ref` change and `sleep` logic disabled. In this case placing `assert` inside `StepFunction` would be natural.

jdegoes: Ah, you're right, we can't use effect suspend here. Maybe you should just clamp the values to `[0, 59]` or use wraparound? The alternative is to delay the failure until the schedule is used, which of course is possible but doesn't seem obviously better than clamping/wrapping.

heaven-born: Could not think of anything better than use wraparound. We can clamp it in dotty latter using union and literal types.
The wrapping is still unsafe, so user will be responsible for catching errors.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -1126,6 +1127,40 @@ object Schedule {
...
+        )
+
+    Schedule(loop(0L))
+  }
+
+  /**
+   *  Value must be in range [0 ... 59]
+   *  Throws [[java.lang.IllegalArgumentException]] on invalid values.
+   */
+  case class Minute(value: Int) {
```
jdegoes, line 1154: I would not introduce this. It doesn't really belong in ZIO and it has the same problem, which is that if you call it "outside ZIO", you can't deal with the error inside ZIO.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -1126,6 +1127,34 @@ object Schedule {
     Schedule(loop(None, 0L))
   }
 
+  /**
+   * Cron-like schedule that recurs every specified `minute` of each hour.
+   * It triggers at zero second of the minute.
+   * Producing a count of repeats: 0, 1, 2.
+   */
+  def minuteOfHour(minute: Int): UIO[Schedule[Any, Any, Long]] = ZIO.effectTotal {
```
jdegoes, line 1135: This is not acceptable in my view. All the constructors of schedule should return Schedule values. Otherwise composition is too painful. If you really want to check the range, then please make that lazy inside the `loop` call. It will defer the error until the schedule is run, but that is the best we can do if we want to guarantee the errors can be managed from within ZIO.

jdegoes (dismissed):
Looks great! You want to get this merged in or add the other ones following this same pattern?

Btw I didn't check the logic too closely, but I assume this recurs always on the the next specified minute boundary, e.g. if it's minute 15 and it's supposed to recur at minute 15, then that's what it does; but if it's minute 15 and it's supposed to recur on minute 16, then it will wait till that minute. Right?

heaven-born:
@jdegoes 
> You want to get this merged in or add the other ones following this same pattern?

I prefer to merge this one. I'll make one more PR for the rest of the methods. 

> if it's minute 15 and it's supposed to recur at minute 15, then that's what it does; but if it's minute 15 and it's supposed to recur on minute 16, then it will wait till that minute. Right?

Right. Works like this:

| Minute param  | Current time | Recur time  |
| ------------- | ------------- | ------------- |
| 14                   | 00:15:00  0000 | 1:14:00 |
| 15                   | 00:15:01  0000  | immediately |
| 15                   | 00:15:00  0001| immediately|
| 16                   | 00:15:00  0000 | 00:16:00 |


 I've just added one more test for millisecond case, so now it ready to merge.

heaven-born:
just one more non-functional fix. Now it's ready.

heaven-born:
since it is not merged yet, I also added `secondOfMinute` and `hourOfDay` because it follows the same pattern. It may require some extra effort for the rest of the methods. So separate PR will be better.

jdegoes (approved):

jdegoes:
@heaven-born Excellent work. Thank you for your contribution!

fireworks

PR #4177, opened 2020-09-02
ZQueue reactive runner

politrons:
This is an implementation that I normally do in each ZIO project, to use Queue as runner of each ZIO program that I create per each Request that I receive, to be processed in a reactive way, applying NIO, Back-pressure and Bulkhead pattern.

I decide that it might be useful not only for me, to avoid have to create same patter per each new project, but for everybody else that use ZQueue in ZIO ecosystem.

If it has no sense at all in the library is totally fine. I also I will appreciate if there's a better way to implement NIO, Back-pressure and Bulkhead pattern as I do here, to improve the way that I do it my projects.

Also the place where I put the implementation was the logical place, but in case it's consider useful it can be in whatever place it fit better.


jdegoes:
@politrons I think this would make a great addition for ZIO Concurrent. I'd suggest opening up this pull request there.

PR #4179, opened 2020-09-06
[WIP] Cause: enable custom comparators and hashers

sideeffffect:


core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -69,6 +69,12 @@ sealed abstract class Cause[+E] extends Product with Serializable { self =>
   final def dieOption: Option[Throwable] =
     find { case Die(t) => t }
 
+  def equalsWith[E1 >: E](
```
adamgfraser, line 72: I think to determine whether two causes are equal we are going to need a way to either hash or order `E` values consistent with that custom notion of equality so we can determine efficiency whether two unordered sets of `E` values are equal.

sideeffffect: I'm not sure what you mean... There is [`def hashWith(hashCause: Cause.HashCause[E]): Int`](https://github.com/zio/zio/pull/4179/files#diff-1135e3fcd6165ca11a68e2706ae3450dR138) and we do compare sets of Es using the custom hashing instance in [`private def hashCode[E](hash: Cause[E] => Int)(c: Cause[E]): Int`](https://github.com/zio/zio/pull/4179/files#diff-1135e3fcd6165ca11a68e2706ae3450dR937).
Is that what you had in mind?

adamgfraser: So right now we are implementing this `equalsWith` method that basically says if we have a way to determine whether two `E` values are equal we can determine whether two cause values are equal. But I don't think that is actually the case. Let's say we have a left hand side that is a bunch of nested `Both` failures then we have a right hand side that is also a bunch of nested `Both` failures.

Equality is supposed to be independent of the order of these failures because they are conceptually so in parallel. So we want to determine whether these two sets of `E` values are equal. But we don't have an ability to do that in an efficient way if all we can do is compare two `E` values for equality because then we have to do a quadratic time operation of comparing each element in one set to each element of the other set. If we had a way of hashing them or a way of ordering them then we could do it more efficiency.

Basically, using set equality inherently uses the default notion of equality and hash code instead of the custom one we are trying to provide.

sideeffffect: Right, I think I see the issue, there is `flatten` which uses `Set[Cause[E]]`, which makes this effort futile :(
Oh well.

PR #4188, opened 2020-09-10
Enable effectual filtering of Sets

kelvin-chappell:
This PR extends #4037 to enable effectual filter of Sets as well as other Collections.

After the introduction of #4037 any pre-existing client code that used a Set as an argument to `filter` had to be converted to use a covariant collection type instead.  [Here is an example](https://github.com/guardian/price-migration-engine/pull/164/files#diff-2181d3ae8e6b272b5ea04ebe034beed3R19). Now client code need not distinguish between Sets and any other type of Collection.

The reason Sets have to be treated differently has been discussed in the implementation of `foreach` [here](https://github.com/zio/zio/pull/4037#discussion_r464073374).

Co-authored-by: @mario-galic


kelvin-chappell:
Tests are failing for reasons unrelated to this PR.

adamgfraser (approved):

PR #4195, opened 2020-09-11
Add some Schedule constructors for cron-like functionality: dayOfWeek

heaven-born:
adds `dayOfWeek` from #4043 .

Follows exactly the same pattern as the rest of the `xxxOfxxx` methods.


jdegoes (approved):
Great work! Appreciate the positive and negative test.

PR #4201, opened 2020-09-13
Implement ZIO.not

### LINKED ISSUE #4197: Implement ZIO.not
In my code base, I found useful having such combinator:

```scala
def not[R, E](effect: ZIO[R, E, Boolean]) =
  effect.map(result => !result)
```

e.g., if you have method `activated(foo)`, it's quite concise to write `not(activated(foo))`.

evis:
Resolves #4197

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -678,7 +678,7 @@ object IO {
   /**
    *  @see See [[zio.ZIO.not]]
    */
-  def not[R, E](effect: ZIO[R, E, Boolean]): ZIO[R, E, Boolean] =
+  def not[E](effect: IO[E, Boolean]): IO[E, Boolean] =
     ZIO.not(effect)
 
   /**
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -676,6 +676,12 @@ object IO {
   val none: UIO[Option[Nothing]] = ZIO.none
 
   /**
+   *  @see See [[zio.ZIO.not]]
+   */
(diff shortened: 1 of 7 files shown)


evis:
@adamgfraser could you restart builds please? They failed because of reasons not related to this PR.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3291,6 +3297,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     if (cont(initial)) body(initial) *> loop_(inc(initial))(cont, inc)(body)
     else ZIO.unit
 
+  /**
+   * Returns a new effect where boolean value of this effect is negated.
+   */
+  def not[R, E](effect: ZIO[R, E, Boolean]): ZIO[R, E, Boolean] =
```
adamgfraser, line 3303: Would be good to have variants of this for the type alias companion objects.

evis:
Again, tests failed due to some random timeouts not related to this PR :(

core/shared/src/main/scala/zio/IO.scala
```diff
@@ -675,6 +675,12 @@ object IO {
    */
   val none: UIO[Option[Nothing]] = ZIO.none
 
+  /**
+   *  @see See [[zio.ZIO.not]]
+   */
+  def not[R, E](effect: ZIO[R, E, Boolean]): ZIO[R, E, Boolean] =
```
adamgfraser, line 681: No `R` type parameter here and just accept and return an `IO[E, A]`. Similar thing for each of the other companion object methods so we return the more specific type.

adamgfraser (approved):
Thank you!

Issue #4189, opened 2020-09-10
ZStream.fromQueue & large chunks lead to larger stream latency


svroonland:
`ZStream.fromQueue` and variants will create Chunks with everything there is currently in the queue. This can lead to large chunks being processed in the stream, which is probably good for throughput but not for latency.

We could add a `maxChunkSize` parameter to the method or maybe there is a good fixed maximum value.

adamgfraser:
Closed by #4205.

PR #4205, opened 2020-09-15
Add maxChunkSize to ZStream.fromQueue

svroonland:
Implements #4189

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3583,11 +3583,13 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
 
   /**
    * Creates a stream from a [[zio.ZQueue]] of values
+   * 
+   * @param maxChunkSize Maximum number of queued elements to put in one chunk in the stream
    */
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3583,11 +3583,13 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
 
   /**
    * Creates a stream from a [[zio.ZQueue]] of values
+   *
+   * @param maxChunkSize Maximum number of queued elements to put in one chunk in the stream
    */
-  def fromQueue[R, E, O](queue: ZQueue[Nothing, R, Any, E, Nothing, O]): ZStream[R, E, O] =
+  def fromQueue[R, E, O](queue: ZQueue[Nothing, R, Any, E, Nothing, O], maxChunkSize: Int = 1024): ZStream[R, E, O] =
```
iravid, line 3589: Can we use `ZStream.DefaultChunkSize` here and in `fromQueueWithShutdown`?

iravid (commented):
Small comment and this should be good to go.

iravid (approved):
Thank you @svroonland!

PR #4210, opened 2020-09-16
Start using Mergify

sideeffffect:


sideeffffect:
Spurious errors in some builds, can somebody please restart them? :pray:

adamgfraser:
Yep, will do. Got to wait for them to finish first.

adamgfraser (approved):

sideeffffect:
@adamgfraser gentle reminder :smile_cat:

adamgfraser:
Was actually waiting to get a little more feedback on this one until we flip the switch for ZIO as opposed to just ZIO Prelude and give @mijicd and @softinio time to weight in. 😃

mijicd:
I'm generally in favor of it. Let's see how it works for everyone and (re-)decide :).

ghostdogpr:
It's a bit pointless as long as there are so many flaky test issues. If you look at [commits to master](https://github.com/zio/zio/commits/master), 85% of the commits on the first page had the build fail. Maybe we mark them flaky and make issues for it?

adamgfraser:
Yes I think that is a good idea.

sideeffffect:
> pointless

Yes and no.
Flaky builds, while hindering this to its fully automated luxury potential, are an orthogonal issue.
Even if we had flaky CI forever, which I hope won't be the case, Mergify would still help, because it would merge PRs immediately after the builds are eventually green, without human having to keep that in her mind that she has to merge that stuff.

sideeffffect:
Btw, most of the flakiness that I've witnessed is either because of 1 min timeout (which I suspect CircleCI is the root cause of -- it's slow) or outright CircleCI failures where the test suite finishes, but the CircleCI job keeps hanging there and it eventually killed because of "not producing and std output in the last 10 mins".

So short term: raise 1 min to a higher value (2 mins? 3 mins?)
Long term: reconsider CircleCI as the CI solution.

adamgfraser:
I actually have a different analysis. I think that when the 1 minute timeout is exceeded it is generally a case where a flaky test is resulting in nontermination, so extending the timeout would not help. In particular, I believe there is some flakiness related to the `TestClock` that I tried to resolve in #4087 but has been hampered by some Javascript compatibility issues addressed in #4115 and #4158. So my hope is we can resolve those issues and then make those refinements to the `TestClock`, along with potentially fixing or marking any other flaky tests, we can get back to a more consistent status with the CI infrastructure as we had in the past.

I also think the cases where there is no response in 10 minutes generally indicate some kind of error on our end that we need to address and understand, potentially solved by the steps above.

adamgfraser:
Going to merge this as it seems like automating dependency updates is itself a good thing even if we aren't able to fully take advantage of it right now due to flakiness in the build. We can always revisit if there are issues.

sideeffffect:
Oh, interesting! Thanks for the analysis and explanation @adamgfraser . Seems like I was too quick to judge CircleCI...

PR #4213, opened 2020-09-17
Update Dotty to 0.27.0-RC1

jczuchnowski:


jczuchnowski:
Ok, turns out the latest Metals doesn't support this version of Dotty. I don't think it's a problem, but still something to consider.

sideeffffect:
Wouldn't it be better to wait for 0.27.0 final? :thinking:

adamgfraser:
@sideeffffect At least historically we've always done the RC. That is the one that comes out along with the blog post with each new Dotty version. I believe the final comes out along with the next release candidate so I think if we did the final we would always be six weeks behind, which might mean a little more stability but also would mean that we might not see issues until later. My sense is that we want to stick with the release candidates so if there are problems we can immediately open issues for them.

jczuchnowski:
Wow, I never thought about it this way. Just wanted to use ZIO with the latest version :laughing:

jczuchnowski:
The failing tests look kind of random....

adamgfraser:
I will rerun.

neko-kai:
The final is always the same commit as RC just with the incremented version number. Intellij also supports RCs not finals so in general I think there's no real reason for non-RC releases in dotty.

neko-kai:
@jczuchnowski  @adamgfraser 
Hey, the only test currently failing is
```
[info] - SerializableSpec
[info]   - ZManaged is serializable
[info]     Fiber failed.
[info]     An unchecked error was produced.
[info]     java.io.NotSerializableException: zio.ZManaged$Running$1$
[info]     	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
[info]     	at java.io.ObjectOutputStream.defaultWriteFields(ObjectOutputStream.java:1548)
[info]     	at java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1509)
[info]     	at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1432)
```
in `test_dotty_jdk8_jvm`.

Maybe there was a bug/change in dotty that makes `ZManaged.Running` not serializable. if there's no workaround we could probably just ignore serializability on dotty for 0.27 version?

adamgfraser:
@jczuchnowski The issue is that `State`, `Exited`, and `Running` are defined in the `make` method of `ZManaged`. If you move their definitions to the `ZManaged` companion object and make them `private` the test should pass.

jczuchnowski:
@adamgfraser thanks! I was trying to pinpoint the problem, but had problem with stabilizing local tests. Need to work on this repo more ;)

jczuchnowski:
@adamgfraser I think we need to restart a build again

adamgfraser:
@jczuchnowski On it.

jczuchnowski:
Looks like we just need a couple of iterations more :smile:

adamgfraser (approved):

PR #4219, opened 2020-09-19
Add example of testing layered effects with Clock

leszekgruchala:
This is a more complex test example using layers and TestClock. I had to spend some time to have it working, hopefully this will help other people too 👍 

Unfortunately, I cannot make it compilable under mdoc even though it compiles in [scastie](https://scastie.scala-lang.org/d1gkSoQzSSSwev3j2aISMA). 

```
error: /Users/lgr/projects/opensource/zio/docs/howto/test_effects.md:342:63: type mismatch;
 found   : repl.MdocSession.App1.ZIO[Any,Exception,Unit]
 required: zio.ZIO[?,?,Any]
        (ZIO.sleep(10.seconds) *> promise.succeed(1).tap(b => logger.log(b.toString)))
                                                              ^^^^^^^^^^^^^^^^^^^^^^
error: /Users/lgr/projects/opensource/zio/docs/howto/test_effects.md:349:64: type mismatch;
 found   : zio.IO[Exception,Nothing]
    (which expands to)  zio.ZIO[Any,Exception,Nothing]
 required: repl.MdocSession.App1.ZIO[Any,Exception,Unit]
    override def log(msg: String): ZIO[Any, Exception, Unit] = ZIO.fail(new Exception("BOOM"))
                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
error: /Users/lgr/projects/opensource/zio/docs/howto/test_effects.md:357:57: type mismatch;
 found   : repl.MdocSession.App1.ZIO[Any,Exception,Boolean]
 required: zio.ZIO[?,?,?]
      result  <- ZIO.service[SchedulingService].flatMap(_.schedule(promise)).run.fork
                                                        ^^^^^^^^^^^^^^^^^^^
error: /Users/lgr/projects/opensource/zio/docs/howto/test_effects.md:362:45: This operation assumes that your effect requires an environment. However, your effect has Any for the environment type, which means it has no requirement, so there is no need to provide the environment.
  testCase.provideSomeLayer[TestEnvironment](partialLayer)
                                            ^
```

Any idea how to solve it?

### DIFF
--- a/docs/howto/test_effects.md
+++ b/docs/howto/test_effects.md
@@ -311,6 +311,65 @@ testM("One can control time as he see fit") {
 
 The above code creates a write once cell that will be set to "1" after 10 seconds asynchronously from a different thread thanks to call to `fork`. At the end we wait on the promise until it is set. With call to `TestClock.adjust(10.seconds)` we simulate passing of 10 seconds of time. Because of it we don't need to wait for the real 10 seconds to pass and thus our unit test can run faster This is a pattern that will very often be used when `sleep` and `TestClock` are being used for testing of effects that are based on time. The fiber that needs to sleep will be forked and `TestClock` will used to adjust the time so that all expected effects are run in the forked fiber.
 
+**Example 3**
+
+A more complex example leveraging layers and multiple services is shown below. 
+
+```scala mdoc
+import zio.clock.Clock
+import zio.duration._
+import zio.test.Assertion._
+import zio.test._
+import zio.test.environment.{ TestClock, TestEnvironment }
+import zio._
+
+trait SchedulingService {
+  def schedule(promise: Promise[Unit, Int]): ZIO[Any, Exception, Boolean]
+}
+
+trait LoggingService {
+  def log(msg: String): ZIO[Any, Exception, Unit]
+}
+
+val schedulingLayer: ZLayer[Clock with Has[LoggingService], Nothing, Has[SchedulingService]] =
+  ZLayer.fromFunction { env =>
+    new SchedulingService {
+      private val clock                     = env.get[Clock.Service]
+      private val logger                    = env.get[LoggingService]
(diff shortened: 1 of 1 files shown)


leszekgruchala:
I decided to remove `mdoc` from my example, I am unable to make it work. It works well in scastie.

adamgfraser:
Will take a look. Thanks for working on this!

docs/howto/test_effects.md
```diff
@@ -311,6 +311,62 @@ testM("One can control time as he see fit") {
 
 The above code creates a write once cell that will be set to "1" after 10 seconds asynchronously from a different thread thanks to call to `fork`. At the end we wait on the promise until it is set. With call to `TestClock.adjust(10.seconds)` we simulate passing of 10 seconds of time. Because of it we don't need to wait for the real 10 seconds to pass and thus our unit test can run faster This is a pattern that will very often be used when `sleep` and `TestClock` are being used for testing of effects that are based on time. The fiber that needs to sleep will be forked and `TestClock` will used to adjust the time so that all expected effects are run in the forked fiber.
 
+**Example 3**
+
+A more complex example leveraging layers and multiple services is shown below. 
+
+```
```
adamgfraser, line 318:
```suggestion
```scala mdoc:reset
```

adamgfraser (changes requested):
@leszekgruchala You're getting the error because the definition of `ZIO` is being shadowed by the definition on line 199. You can use the `reset` command in `mdoc` to get a fresh scope. With this change your code compiles correctly when I run `mdoc` so then we should be good to go.

leszekgruchala:
@adamgfraser awesome, thank you, fixed!

adamgfraser (approved):
Thank you for adding this! 🙏

PR #4221, opened 2020-09-19
fix-3688

### LINKED ISSUE #3688: zio-test: Parallel checkM
Would be useful to have some kind of parallel  `checkM`, given some generators would execute in parallel instead of sequential.

lightning95:
fixed #3688 - added parallel checkM (checkMPar)


adamgfraser (approved):
Great addition! Thank you! 🙏

PR #4222, opened 2020-09-20
Add rezilience library to ecosystem

svroonland:
A zio-native library I made with a circuit breaker, rate limiter and bulkhead (ala resilience4j).


adamgfraser (approved):
Thanks for adding!

PR #4228, opened 2020-09-22
ZStream.fromResource

fsvehla:
I’ve found this handy in https://github.com/zio/zio-json/pull/12 and think it is general enough to make it into ZIO proper.


jdegoes (approved):

iravid (approved):
Looks great @fsvehla. Thanks for contributing this!

PR #4230, opened 2020-09-23
[WIP] STM support for `Cause`

robmwalsh:
Initial work from a 1:1 session with @jdegoes 

The general idea is to add support for `Cause` to `STM` allowing a transaction to fail for more than one reason.
Instead of failing early, the transaction would attempt to continue, collecting any other errors, before eventually failing with all failures. This wouldn't work with `flatMap` but `zip` & friends should be easy enough. 

This would allow reporting all errors preventing a transaction to the user, so they can fully understand all the problems before retrying (rather than fixing one issue at a time).

Mima currently has 5 complaints all stemming from changing Fail from `Fail[+A](value: A)` to `Fail[+A](cause: Cause[A])`

`TExit` is private by virtue of being within the `ZSTM.internal` object, but it does appear in 
```scala
def done[R, E, A](exit: => TExit[E, A]): ZSTM[R, E, A]
```
which is the only non-private method I can find that references it

Is it safe to filter these errors in mima?

If not, options from here:
1. Do nothing
1. Add support for `Cause` using new combinators with evidence that `E` is a subtype of `Cause` e.g. `zipRightCause`. This may be a better plan than the original because it would allow failing fast with the standard `zip` & friends or optionally fail slow with better error info if needed.
1. Wait for 2.x development to start so I can break bincompat

any thoughts? cc @adamgfraser @mijicd


adamgfraser:
@robmwalsh I think we might want to define a `zipWithPar` combinator on STM to support this behavior. Generally, the expectation is that `zipWith` should behave like `flatMap` and implement "fail fast" semantics so I tend to think we would not want it to accumulate errors. If we defined `zipWithPar` it could have this behavior and while it wouldn't be parallel in the sense of forking fibers it would be parallel in the sense of performing both transactions and accumulating their errors. In this respect it would be similar to the `zipWithPar` operator on `ZValidation`.

Regarding the binary compatibility issue, I think it is okay to exclude this from the MiMa filter. I think the fact that `ZSTM.done` is public at all is an oversight and as I understand it there is no way a user could actually instantiate a `TExit` value without being another ZIO library since it is package private. So I don't think there is a way that a user could actually be calling the method in a way that would break their code. Therefore, I think we should just make `ZSTM.done` private and then we can have the nice semantics we want.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -563,9 +578,15 @@ final class ZSTM[-R, +E, +A] private[stm] (
       val reset = prepareResetJournal(journal)
 
       val continueM: TExit[E, A] => STM[E1, A1] = {
-        case TExit.Fail(_)    => { reset(); that.provide(r) }
+        case TExit.Fail(_) => {
+          reset();
+          that.provide(r)
+        }
```
mijicd, line 584:
```suggestion
        case TExit.Fail(_) =>
          reset()
          that.provide(r)
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -563,9 +578,15 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
         case TExit.Succeed(a) => ZSTM.succeedNow(a)
-        case TExit.Retry      => { reset(); that.provide(r) }
+        case TExit.Retry => {
+          reset();
+          that.provide(r)
+        }
```
mijicd, line 589:
```suggestion
        case TExit.Retry =>
          reset()
          that.provide(r)
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1517,7 +1546,11 @@ object ZSTM {
         saved.put(entry.getKey, entry.getValue.copy())
       }
 
-      () => { journal.clear(); journal.putAll(saved); () }
+      () => {
+        journal.clear();
+        journal.putAll(saved);
```
mijicd, line 1551:
```suggestion
        journal.clear()
        journal.putAll(saved)
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1691,7 +1724,11 @@ object ZSTM {
     )(
       k: ZIO[R, E, A] => Any
     ): Unit = {
-      def complete(io: IO[E, A]): Unit = { done.set(true); k(io); () }
+      def complete(io: IO[E, A]): Unit = {
+        done.set(true);
+        k(io);
```
mijicd, line 1729:
```suggestion
        done.set(true)
        k(io)
```

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1782,13 +1819,16 @@ object ZSTM {
     sealed abstract class TExit[+A, +B] extends Serializable with Product
     object TExit {
       val unit: TExit[Nothing, Unit] = Succeed(())
-
-      final case class Fail[+A](value: A)    extends TExit[A, Nothing]
-      final case class Succeed[+B](value: B) extends TExit[Nothing, B]
-      case object Retry                      extends TExit[Nothing, Nothing]
+      final case class Fail[+A](cause: Cause[A]) extends TExit[A, Nothing] /*{
+        def apply[A1 >: A](value: A1): Fail[Cause[A1]]        = apply(Cause.fail(value))
+        def apply[A1 >: A](cause: Cause[A1]): Fail[Cause[A1]] = new Fail(cause)
```
mijicd, line 1824: Make sure to remove this once it's ready to go.

mijicd (commented):
Just a few minor remarks, other than that, it looks good. Regarding binary compatibility, I think it is acceptable to filter it out and make `done` private, as it is an obvious mistake to be public when `TExit` is invisible.

Last, but not least, please don't forget to propagate all additions to `STM` :).

robmwalsh:
@adamgfraser agree about the expectations RE zip right, but to me, `zipWithPar` implies something is happening in parallel when there's no concept of parallelism in `STM`. How would you feel about calling it `zipWithCause` instead?

adamgfraser:
@robmwalsh I kind of feel like we have crossed that bridge with like the implementation of `zipWithPar` in `Validation`, which also doesn't do any actual parallelism even though it has the `Par` suffix. We could always revisit that as well but I think there is a lot of value in consistency between different ZIO libraries and I think right now the principle we are following at least implicitly is I think that `zipWithPar` can describe operators that are parallel in the sense that they do both sides in an unbiased way even if they don't fork fibers.

Maybe one way to think about it if you take as a given that there is no actual forking of fibers within STM then doing both sides and keeping both errors is the only thing that `Par` could mean.

robmwalsh:
Ok, zipWithPar it is. Now with the naming out of the way and no cache invalidation to deal with, that just leaves me with the easy bit of implementation :)

robmwalsh:
Looks like this is now working (CI failures are unrelated; side note, there are frequent CI timeout failures now - is there an issue for that? I'm not sure how to see what's actually failing since the suite isn't rendered until all the tests are complete), 

what other combinators should I add?
sandbox/unsandbox 
???

@mijicd I haven't added anything to `STM` yet - it seems there's a lot in `ZSTM` that's missing from `STM` (e.g. all zip operators). should I add all missing combinators or is there some logic to what stays in `ZSTM` only (other than things that use `R`)this? the changes you suggested were the unfortunate result of IJ deciding to fall back to its formatter instead of scalafmt. I've reverted those changes.

mijicd:
@robmwalsh let's add the ones you introduced here to STM, and have a separate PR for others.

robmwalsh:
@mijicd I've added the `STM` variants and ended up removing `ZSTM.done` & `STM.done` because they're not actually used anywhere. I think this should be good to merge now - can you please review?

mijicd (approved):
LGTM as long as all tests are passing. One minor remark: please add stack safety test(s) to `ZSTMSpec`.

adamgfraser:
@robmwalsh Can you merge upstream changes to pick up the improvements we made to the stability of the CI environment?

robmwalsh:
@mijicd Mima doesn't like the symbolic operators on STM - shall I remove them?

```
[error] /home/circleci/project/core/shared/src/main/scala/zio/stm/STM.scala:28:7: multiarg infix syntax looks like a tuple and will be deprecated [error] def <&[E, A, B](left: => STM[E, A], right: => STM[E, B]): STM[E, A] = [error] ^ [error] /home/circleci/project/core/shared/src/main/scala/zio/stm/STM.scala:34:7: multiarg infix syntax looks like a tuple and will be deprecated [error] def <&>[E, A, B, C](left: => STM[E, A], right: => STM[E, B]): STM[E, (A, B)] = [error] ^ [error] /home/circleci/project/core/shared/src/main/scala/zio/stm/STM.scala:40:7: multiarg infix syntax looks like a tuple and will be deprecated [error] def &>[E, A, B](left: => STM[E, A], right: => STM[E, B]): STM[E, B] = [error] ^
```

mijicd:
@robmwalsh Ah, darn. Let's remove them unless someone knows how to fix the incompatibility (disabling the check doesn't count as fix :) ).

robmwalsh:
I love the "will be" deprecated. Why not just deprecate?

Perhaps this is why none of the symbolic operators are on the zio variants either.

I'm playing farmer and helping shear alpacas today but will try to fix it up this evening then we should be able to merge.

robmwalsh:
@mijicd lgtm :)

adamgfraser:
@robmwalsh I'm not sure I understand the infix syntax. Wouldn't we want this on the trait like we already have versus on the companion object? We may be able to find a workaround if we really want this but not sure it makes sense.

robmwalsh:
@adamgfraser there isn't an STM trait. I could create one if that's preferred but assumed it didn't exist for a reason. I think that would fix the Mima issue with the symbolic operators though.

adamgfraser:
@robmwalsh Right but there is a `ZSTM` trait and `STM` is a type alias for `ZSTM` so any operators defined on the `ZSTM` trait are also available on instances of `STM`.

robmwalsh:
@adamgfraser, should I just remove the operators from STM then? I only added them because @mijicd asked me to https://github.com/zio/zio/pull/4230#pullrequestreview-495250582

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -111,6 +105,20 @@ object STM {
   def flatten[E, A](task: STM[E, STM[E, A]]): STM[E, A] =
     ZSTM.flatten(task)
 
+  /**
+   * @see See [[zio.stm.ZSTM.foldCause]]
+   */
+  def foldCause[E, A, B](v: => STM[E, A])(f: Cause[E] => B, g: A => B)(implicit ev: CanFail[E]): USTM[B] =
```
adamgfraser, line 111: I think I would delete these as they already exist on the `ZSTM` trait.

core/shared/src/main/scala/zio/stm/STM.scala
```diff
@@ -358,4 +366,28 @@ object STM {
 
   private[zio] def succeedNow[A](a: A): USTM[A] =
     ZSTM.succeedNow(a)
+
+  /**
+   * @see See [[zio.stm.ZSTM.zipLeftPar]]
+   */
+  def zipLeftPar[E, A, B](left: => STM[E, A], right: => STM[E, B]): STM[E, A] =
```
adamgfraser, line 373: I would delete these as well.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -359,11 +380,38 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
       case TExit.Succeed(a) => ZSTM.succeedNow(g(a))
       case TExit.Retry      => ZSTM.retry
     }
 
+  /**
+   * A more powerful version of `foldM` that allows recovering from any kind of failure.
+   *
+   * todo: is this actually useful?
+   */
+  def foldCauseM[R1 <: R, E2, B](
```
adamgfraser, line 403: Seems perfectly useful. Might be worth adding a note that since these are transactions versus effects the `Cause` could contain multiple parallel failures but not interruption or die failures. And I believe at least as of today there are not situations in which the cause would have multiple sequential failures.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -359,11 +380,38 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
+    failure: Cause[E] => ZSTM[R1, E2, B],
+    success: A => ZSTM[R1, E2, B]
+  ): ZSTM[R1, E2, B] =
+    self.continueWithM {
+      case TExit.Fail(c)    => failure(c)
+      case TExit.Succeed(a) => success(a)
+      case TExit.Retry      => ZSTM.retry
+    }
+
+  //new ZIO.Fold(self, failure, success)
```
adamgfraser, line 413: Delete commented out code?

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -857,25 +905,64 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
   /**
    * Named alias for `<*`.
    */
   def zipLeft[R1 <: R, E1 >: E, B](that: => ZSTM[R1, E1, B]): ZSTM[R1, E1, A] =
     (self zip that) map (_._1)
 
+  /**
+   * Named alias for `<&`.
+   */
+  def zipLeftPar[R1 <: R, E1 >: E, B](that: => ZSTM[R1, E1, B]): ZSTM[R1, E1, A] =
```
adamgfraser, line 923: `zipParLeft`

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -857,25 +905,64 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
   /**
    * Named alias for `*>`.
    */
   def zipRight[R1 <: R, E1 >: E, B](that: => ZSTM[R1, E1, B]): ZSTM[R1, E1, B] =
     (self zip that) map (_._2)
 
+  /**
+   * Named alias for `&>`.
+   */
+  def zipRightPar[R1 <: R, E1 >: E, B](that: => ZSTM[R1, E1, B]): ZSTM[R1, E1, B] =
```
adamgfraser, line 935: `zipParRight`

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -857,25 +905,64 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
+    self.continueWithM {
+      case TExit.Succeed(a) =>
+        that.continueWithM {
+          case TExit.Succeed(b) => ZSTM.succeed(f(a, b))
+          case TExit.Fail(c2)   => ZSTM.halt(c2)
+          case TExit.Retry      => ZSTM.retry
+        }
+      case TExit.Fail(c1) =>
+        that.continueWithM {
+          case TExit.Fail(c2) => ZSTM.halt(c1 ++ c2)
```
adamgfraser, line 959: Shouldn't it be `&&` here to indicate that these errors occurred "in parallel"?

adamgfraser (changes requested):
@robmwalsh Left some comments. I think pretty straightforward changes and then this should be ready to go! Thanks for working with this and wrangling the MiMa issues!

core/shared/src/main/scala/zio/stm/STM.scala
mijicd: These are "my fault" in a way. I proposed adding these operators for symmetry with ZSTM companion. However, if they aren't here, then they should not be on STM either.

Sorry for causing the confusion :pray:.

core/shared/src/main/scala/zio/stm/ZSTM.scala
robmwalsh: @adamgfraser The transactions occur sequentially so I thought `++` would be better suited in the usual case, leaving the user to use `&&` as part of an explicit return if they wanted to say something failed for two reasons at the same time. My motivation for these changes was that I wanted to fail adding an edge to a graph if the in vertex or the out vertex was missing, and wanted to be able to use `&&` to say that both the in vertex and out vertex were missing at the same time (i.e. it failed for both `Cause`s).

robmwalsh:
@adamgfraser I think this is good to go if you're happy to leave `++` as is (see https://github.com/zio/zio/pull/4230#discussion_r504461149 )

robmwalsh:
@adamgfraser a gentle reminder on this one

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: Hey sorry for the delay on this. I think `&&` makes more sense here as we are doing them "in parallel" in the sense that we are doing both of them regardless of whether one fails and in particular I think that if both transactions fails we want  the cause from `l.zipPar(r)` to be the same as `r.zipPar(l)`, which is only true if the errors are "parallel". @mijicd do you have any thoughts on this?

adamgfraser:
Sorry, just responded on the `++` issue.  Other than that good to go!

core/shared/src/main/scala/zio/stm/ZSTM.scala
robmwalsh: Order is important to my use case, which is the reason for raising this PR. 
```scala
//succeeds
graph.addVertex(sydney) &> graph.addVertex(melbourne) &> graph.addEdge(sydney, route1, melbourne) 

//fails at step 2 because melbourne isn't in the graph
graph.addVertex(sydney) &> graph.addEdge(sydney, route1, melbourne) &> graph.addVertex(melbourne) 

//fails at step 1 because melbourne isn't in the graph and (at the same time) because sydney isn't in the graph, and then later fails at step 3 because melbourne isn't in the graph
graph.addEdge(sydney, route1, melbourne) &> graph.addVertex(sydney) &> graph.addEdge(sydney, route2, melbourne) &>graph.addVertex(melbourne) 
```
I think this comes back to my previous concerns using the term `Par` for what are fundamentally sequential operations and aren't necessarily independent (`l` is always interpreted before `r`).

How about I swap to `&&` for `zipPar` and friends because that make sense if we're using the concept of parallel and will likely be useful for other use cases, then I can add some extra combinators that meet my needs for sequential collection of errors using `++`, say `zipSequential` & friends using variants of the symbolic operator `<#>` which isn't used anywhere else in the code base? `zipSeq` would probably be preferable but may cause confusion with the standard scala `Seq` trait. `<+>` would have nice parallels in the way the errors compose (`++` for `<+>` and `&&` for `<&>`) but that's already used for `orElseEither`.

adamgfraser: Got it. So then I think maybe what we need is the STM version of the `validateWith` operator on ZIO, which performs both effects in sequence and combines the causes with `++`. Then maybe we don't actually need the parallel versions. Sorry for the delay on this, just trying to think about how we fit it in with the tools we already have instead of doing something completely new.

mijicd: I think Par is misleading here. `validateWith` sounds reasonable.

robmwalsh: Yep, `validateWith` better reflects what I'm trying to do. I actually think we should keep the parallel versions because I do think they're useful for when people do want to run two unrelated transactions e.g. maintaining some sort of index of some other STM structure.

I'm actually using the `new ZSTM((journal, _, _, _) => {...}` trick in my graph library, but conceptually addEdge could look something like:
```scala
graph.getVertex(sydney) <&> graph.getVertex(melbourne) validateWith graph.underlying.addE(_.1, route1, _.2)
```
which I think would be pretty sweet!

PR #4238, opened 2020-09-24
ZIO Test: Evaluate Effects Sequentially In Gen#zipWith

adamgfraser:
`Gen#zipWith` describes combining two generators in a pairwise fashion. We ran into an issue where doing so was leading to generated values not being deterministic because in `ZStream#zipAllWith` we pull from the two streams in parallel so we could potentially evaluate the random effect in the second stream before the first stream. It would be nice to be able to describe combining streams point wise while evaluating the effects of pulling from the two streams sequentially versus in parallel.

To support this I added a new `zipAllWithExec` operator that takes an execution strategy and implemented `zipAllWith` in terms of that. Then we can call `zipAllWithExec` from `Gen#zipWith` to regain deterministic generated values.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3008,7 +3008,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   def zipAllWith[R1 <: R, E1 >: E, O2, O3](
     that: ZStream[R1, E1, O2]
   )(left: O => O3, right: O2 => O3)(both: (O, O2) => O3): ZStream[R1, E1, O3] =
-    zipAllWithExec(that)(left, right)(both)(ExecutionStrategy.Parallel)
+    zipAllWithExec(that)(ExecutionStrategy.Parallel)(left, right)(both)
 
   /**
    * Zips this stream with another point-wise. The provided functions will be used to create elements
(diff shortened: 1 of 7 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3007,7 +3007,22 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
+   *
+   * The functions `left` and `right` will be used if the streams have different lengths
+   * and one of the streams has ended before the other.
+   *
+   * The execution strategy `exec` will be used to determine whether to pull
+   * from the streams sequentially or in parallel.
+   */
+  def zipAllWithExec[R1 <: R, E1 >: E, O2, O3](
+    that: ZStream[R1, E1, O2]
+  )(left: O => O3, right: O2 => O3)(both: (O, O2) => O3)(exec: ExecutionStrategy): ZStream[R1, E1, O3] = {
```
iravid, line 3025: Should we place the strategy’s parameter list first so that the functions are still specified last?

iravid (approved):
Great addition 👍🏻

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Good idea! Will change now.

iravid (approved):

PR #4243, opened 2020-09-25
Trigger auto-approve-action on pull_request_target

sideeffffect:
I would actually like to experiment with `pull_request_target` once more.
This time, we will have to patently wait for Scala Steward to open a PR, so that `github.actor == 'scala-steward'` holds and the action actually runs.
:crossed_fingers: 
/cc @adamgfraser 

---
We can safely conclude, that `auto-approve-action` on `pull_request` doesn't work, it fails with `Resource not accessible by integration`.
Just to make sure, we don't have any _cusom_ token in Secrets with the name `GITHUB_TOKEN` that would alias with the token GitHub automatically creates for each repo, right?

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -1,7 +1,7 @@
 name: Auto approve
 
 on:
-  pull_request
+  pull_request_target
 
 jobs:
   auto-approve:


adamgfraser:
Should we do these experiments with ZIO Prelude before moving over to ZIO itself?

sideeffffect:
@adamgfraser We're actually running 2 sub-experiments:

1. ZIO Prelude with `automatic-pull-request-review` on `pull_request` and
1. ZIO with `auto-approve-action` on `pull_request_target`

I would like to see how they both behave to decide which one (if any) to use in the future. Given that Scala Steward's PRs don't happen very often, having this in 2 repositories will shorten the time needed to evaluate this experiment.

Since this experimentation doesn't cause any problems to ZIO developers, I figured it's OK to perform in even on `zio/zio` even though it is a high stakes repo (compared to `zio/zio-prelude`)

adamgfraser:
@sideeffffect Okay.

adamgfraser (approved):

PR #4248, opened 2020-09-27
Harden TestClock

adamgfraser:


adamgfraser:
This PR incorporates the `TestClock` improvements from my previous work and runs tests sequentially on Scala.js to get all tests working pending a more comprehensive solution to the issues with the Executor on Scala.js. In my testing this resolved all flakiness with the `TestClock` itself. The only flakiness I saw is the test regarding ZManaged inner finalizers, which was already identified as flaky and does not involve the clock, and the issues where there is just no response after ten minutes in the CI builds, which I believe is unrelated.

I think it may make sense to merge a version of this while we finalize the Scala.js changes and investigate the CI to start to improve the stability of our builds.

jdegoes (approved):

PR #4249, opened 2020-09-27
Try To Identify Source Of CI Flakiness

adamgfraser:
We are seeing some builds fail on CI with a message that no output has been received for 10 minutes. This indicates that something is going wrong but it is not simply a test timing out since that should be caught by the test framework and reported as a test failure. Initial work to identify the underlying cause.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1343,10 +1343,10 @@ object ZManaged extends ZManagedPlatformSpecific {
                       Exited(nextKey, exit)
                     )
 
-                  case ExecutionStrategy.ParallelN(n) =>
+                  case ExecutionStrategy.ParallelN(_) =>
                     (
                       ZIO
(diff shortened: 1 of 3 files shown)


fsvehla:
Might be worth trying to double the Runtime’s executor size and see if it makes a difference - maybe it’s getting starved?

adamgfraser:
@fsvehla Good idea! I will try that after this run. 🙏

adamgfraser:
It looks like the CI failures we are seeing are being driven by an issue in `ZIO.foreachParN` where it can leak fibers. `ZIO.foreachParN` is called in the implementation of `releaseAll` in `ZManaged` and tests in ZIO Test are represented as managed effects. So some of the time when a test suite has terminated the test suite will try to clean everything up but will suspend indefinitely waiting for one of these fibers forked in an uninterruptible region to terminate. You can see the same thing in the `preserves ordering of nested finalizers` test for `ZManaged#foreachParN`, which will sometimes fail to terminate.

adamgfraser:
Superseded by #4253.

PR #4253, opened 2020-09-28
Don't Leak Fibers From Within ZIO#foreachParN

adamgfraser:
The current implementation calls `forever` on the worker fibers. If `foreachParN` is called in an uninterruptible region this can result in creating fibers that continue forever and are not interruptible, potentially causing nontermination in higher levels of the application if the application waits for the interruption of these fibers through auto supervision. This is particularly problematic because `foreachParN` is called in an uninterruptible region in the release action of `ZManaged` when the execution strategy is `ParallelN`, which is the default in ZIO Test.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2972,13 +2972,10 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   )(fn: A => ZIO[R, E, B])(implicit bf: BuildFrom[Collection[A], B, Collection[B]]): ZIO[R, E, Collection[B]] = {
 
     def worker(q: Queue[(Promise[E, B], A)], pairs: Iterable[(Promise[E, B], A)], ref: Ref[Int]): URIO[R, Unit] =
-      ref.getAndUpdate(_ - 1).flatMap { remaining =>
-        if (remaining >= 0)
-          q.take.flatMap { case (p, a) =>
-            fn(a).foldCauseM(c => ZIO.foreach(pairs)(_._1.halt(c)), b => p.succeed(b))
-          } *> worker(q, pairs, ref)
-        else
-          ZIO.unit
+      ZIO.whenM(ref.modify(n => (n > 0, n - 1))) {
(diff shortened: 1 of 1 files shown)


adamgfraser:
Seeing a dramatic improvement to stability of CI with this change.

ghostdogpr:
Wow 🤩 👏

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2969,22 +2969,33 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    */
   def foreachParN[R, E, A, B, Collection[+Element] <: Iterable[Element]](n: Int)(
     as: Collection[A]
-  )(fn: A => ZIO[R, E, B])(implicit bf: BuildFrom[Collection[A], B, Collection[B]]): ZIO[R, E, Collection[B]] =
+  )(fn: A => ZIO[R, E, B])(implicit bf: BuildFrom[Collection[A], B, Collection[B]]): ZIO[R, E, Collection[B]] = {
+
+    def worker(q: Queue[(Promise[E, B], A)], pairs: Iterable[(Promise[E, B], A)], ref: Ref[Int]): URIO[R, Unit] =
+      ref.getAndUpdate(_ - 1).flatMap { remaining =>
+        if (remaining >= 0)
```
ghostdogpr, line 2976: Minor: you could use `ZIO.when`

ghostdogpr (dismissed):

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Good idea! Updated now.

ghostdogpr (approved):

sideeffffect:
These are splendid news :fireworks: 
Thank you @adamgfraser :clap: , would it be possible to release ZIO with this improvements, so that ZIO Prelude tests/CI can also benefit?

adamgfraser:
@sideeffffect Yes I think we should definitely do another release very shortly. Ideally would like to get some version of the `TestClock` changes in as well, so I think in the next day or two.

PR #4262, opened 2020-09-29
[Experiment] Use GitHub's merge branch from PRs

sideeffffect:
So this is what we to ideally happen: We don't want to test against the `pull/$CIRCLE_PR_NUMBER/head` branch. We want the `pull/$CIRCLE_PR_NUMBER/merge`.
What do you think?
Do you know of any easier way to achieve this in CircleCI?

/cc @softinio @mijicd


sideeffffect:
We're moving to GitHub Actions

PR #4265, opened 2020-09-30
ZIO Test: Preserve Original Failure When Shrinks Is Set To Zero

adamgfraser:
In `checkStream` we need to take `maxShrinks + 1` instead of `maxShrinks` values because we need to take the original failure in addition to the shrunk values.


jdegoes (approved):

jdegoes:
👍

PR #4268, opened 2020-10-01
Explore New Encoding Of Test Aspects

### LINKED ISSUE #4257: Investigate new encoding for ZIO Test Aspects
Ideally, ZIO Test Aspects would have the following properties:

 - [ ] They allow us to introduce and eliminate errors
 - [ ] They allow us to introduce and eliminate environments, so we can make `provideXYZ` aspects instead of hard-coded methods
 - [x] They infer fully as today
 - [x] They compose as today

[Here](https://scastie.scala-lang.org/m1gZIXsKSGO8Jfy5bdN65Q) is a sketch for a new possible encoding that might meet the first two goals, in addition to the second.

Note this encoding requires separate source files for Scala 2.x and Scala 3.

adamgfraser:
Initial work to resolve #4257.

### DIFF
--- a/test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
+++ b/test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
@@ -98,7 +98,7 @@ object TestAspect {
     }
 
   /**
-   * An aspect that marks tests as ignored.
+   * Constructs an aspect that runs the specified effect after every test.
    */
   def after[R, E](effect: ZIO[R, E, Any]): TestAspect.Aux[
     Nothing,
@@ -125,7 +125,7 @@ object TestAspect {
     }
 
   /**
-   * An aspect that marks tests as ignored.
+   * Annotates tests with the specified test annotation.
    */
   def annotate[V](key: TestAnnotation[V], value: V): TestAspectPoly =
     new TestAspect[Nothing, Any, Nothing, Any] {
@@ -137,7 +137,9 @@ object TestAspect {
     }
 
   /**
-   * An aspect that marks tests as ignored.
+   * Constructs an aspect that evaluates every test between two effects,
+   * `before` and `after`,  where the result of `before` can be used in
+   * `after`.
    */
   def around[R, E, A](before: ZIO[R, E, A])(after: A => ZIO[R, Nothing, Any]): TestAspect.Aux[
     Nothing,
@@ -176,7 +178,8 @@ object TestAspect {
     around(before)(_ => after)
 
   /**
-   * An aspect that marks tests as ignored.
+   * Constructs an aspect that evaluates every test inside the context of the
+   * managed function.
    */
   def aroundTest[R, E](
     managed: ZManaged[R, TestFailure[E], TestSuccess => ZIO[R, TestFailure[E], TestSuccess]]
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
```diff
@@ -0,0 +1,378 @@
...
+        }
+    }
+
+  /**
+   * An aspect that applies the specified aspect on Dotty.
+   */
+  def dotty[LowerR, UpperR, LowerE, UpperE](
+    that: TestAspect[LowerR, UpperR, LowerE, UpperE]
+  ): TestAspect[LowerR, UpperR, LowerE, UpperE] =
+    if (TestVersion.isDotty) that else ???
```
adamgfraser, line 302: We can't implement test aspects like this in the same way because if the test aspect provides the environment or handles errors then we can't return the original spec unchanged because it doesn't satisfy the constraints of the target type. We may need to have a less polymorphic subtype of aspects that only add environment requirements or failure modes to work with operators like this.

test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
```diff
@@ -0,0 +1,378 @@
...
+
+  /**
+   * An aspect that marks tests as ignored.
+   */
+  val dottyOnly: TestAspectAtLeastR[Annotations] =
+    if (TestVersion.isDotty) ??? else ignore
```
adamgfraser, line 308: The compiler seems to have difficulty unifying type level functions even if the output types of the functions can be unified. Here `ignore` adds `Annotations` to the environment and we want to call `identity` if we are on another platform which would return the environment unchanged. So we conceptually should be able to unify that to adding `Annotations`, but the compiler doesn't seem to be able to do that.

test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
```diff
@@ -0,0 +1,378 @@
...
+    restore[TestSystem](_.get)
+
+  /**
+   * An aspect that restores the
+   * [[zio.test.environment.TestClock TestClock]]'s state to its starting
+   * state after the test is run. Note that this is only useful when repeating
+   * tests.
+   */
+  def restoreTestEnvironment: TestAspectAtLeastR[ZTestEnv] =
+    restoreTestClock >>>[Nothing, Any] restoreTestConsole >>>[Nothing, Any] restoreTestRandom >>>[Nothing, Any] restoreTestSystem
```
adamgfraser, line 377: We need to manually specify type parameters here.

test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
```diff
@@ -0,0 +1,384 @@
...
+  /**
+   * Returns a new aspect that represents the sequential composition of this
+   * aspect with the specified one.
+   */
+  def >>>[EnvOut1, ErrOut1](that: TestAspect[EnvOut, ErrOut, EnvOut1, ErrOut1]): TestAspect.Aux[
+    EnvIn,
+    ErrIn,
+    EnvOut1,
+    ErrOut1,
+    ({ type Apply[+R >: EnvIn @uncheckedVariance] = that.ModifyEnv[self.ModifyEnv[R]] })#Apply,
```
adamgfraser, line 41: This creates a warning that `EnvIn` and `ErrIn` appear in invariant position on Scala 2.12 but not Scala 2.13.

test/shared/src/main/scala/zio/test/experimental/TestAspect.scala
jdegoes: This one looks like the worst, although the `if` / `else` is pretty disappointing too.

PR #4274, opened 2020-10-03
Optimize Chunk.Tags.fromValue

iravid:
Been looking at some profiles of the stream benchmarks and noticed
that ClassTag equality showed up pretty noticeably on profiles from
`Tags.fromValue`. Replacing the comparisons by `java.lang.Class`
comparisons speeds that up considerably (15-20%).


adamgfraser (dismissed):

jdegoes:
Awesome.

jdegoes (approved):

PR #4279, opened 2020-10-05
Add a bunch of collect combinators

iravid:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -679,41 +711,20 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
-               }
-      } yield pull
-    }
+  final def collectWhileRight[L1, O1](implicit ev: O <:< Either[L1, O1]): ZStream[R, E, O1] = {
+    val _ = ev
+    self.asInstanceOf[ZStream[R, E, Either[L1, O1]]].collectWhile { case Right(a) => a }
+  }
+
+  /**
+   * Terminates the stream when encountering the first [[Exit.Success]].
```
adamgfraser, line 722: Doesn't it terminate the stream when encountering the first `Failure`?

iravid: Oops, yes.

adamgfraser (approved):

PR #4289, opened 2020-10-06
add STM warnings

robmwalsh:
doc update because ZSTM repeatWhile/repeatUntil and forever are dangerous

### DIFF
--- a/core/shared/src/main/scala/zio/stm/ZSTM.scala
+++ b/core/shared/src/main/scala/zio/stm/ZSTM.scala
@@ -249,7 +249,8 @@ final class ZSTM[-R, +E, +A] private[stm] (
   /**
    * Repeats this `STM` effect until its result satisfies the specified predicate.
    * '''WARNING''': `repeatUntil` uses a busy loop to repeat the effect and will consume a
-   * thread until it completes (it cannot yield).
+   * thread until it completes (it cannot yield). This is because STM describes a single atomic transaction
+   * which must either complete, retry or fail a transaction before yielding back to the ZIO Runtime.
    * - Use [[retryUntil]] instead if you don't need to maintain transaction state for repeats.
    * - Ensure repeating the STM effect will eventually satisfy the predicate.
    * - Consider using the Blocking thread pool for execution of the transaction.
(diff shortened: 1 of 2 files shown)


robmwalsh:
@adamgfraser can you please review? Would you like me to add a deprecation for `forever` as well?

adamgfraser (commented):
@robmwalsh Looks good!

I might say a little more about why this is the case so it doesn't just seem like an arbitrary gotcha. Something like. `Warning: because STM describes a single atomic transaction, repeatUntil will continuously repeat the transaction until the specified condition is true without yielding back to the ZIO Runtime...`

As far as `forever` I think it is okay to deprecate it. @mijicd What do you think?

mijicd:
If there's no compatibility constraint I'd say remove it. So yes, let's deprecate it.

robmwalsh:
@adamgfraser updated wording and deprecated forever

mijicd (approved):

robmwalsh:
Can this be merged now?

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -379,7 +391,16 @@ final class ZSTM[-R, +E, +A] private[stm] (
...
-   */
+   * '''WARNING''': `forever` uses a busy loop to repeat the effect and will consume a
+   * thread until it fails. This is because STM describes a single atomic transaction, and so must either complete,
+   * retry or fail a transaction before yielding back to the ZIO Runtime.
+   * - Ensure repeating the STM effect will eventually fail.
+   * - Consider using the Blocking thread pool for execution of the transaction.
+   */
+  @deprecated(
+    "Repeating until failure doesn't make sense in the context of STM because it will always roll back the transaction",
+    "1.0.2"
```
adamgfraser, line 402: This should be the version that the deprecated method will be removed, so `2.0.0`.

adamgfraser (changes requested):
One very minor comment then good to go!

core/shared/src/main/scala/zio/stm/ZSTM.scala
robmwalsh: According to the scaladoc for `@deprecated` it's when it was first deprecated. If zio does things differently that's fine but this should probably be documented somewhere.  I can't find any mention of this in the repo. Do you still want me to change it?

```scala
/**
*...
*@param  since   a string identifying the first version in which the definition was deprecated
*...
*/
class deprecated(message: String = "", since: String = "") extends scala.annotation.StaticAnnotation
```

adamgfraser: You're right.

adamgfraser (approved):

robmwalsh:
Thanks @adamgfraser :)

PR #4296, opened 2020-10-08
Implement ZIO#schedule

### LINKED ISSUE #4218: Replace ZIO#repeat(Schedule) with ZIO#schedule(Schedule)
Current implementation of `ZIO#repeat(Schedule)` combines both scheduling and repeating concepts which leads to behaviour that is not easy to predict from method signature and also hard to customise. 

Here is an example: let's say I have `Schedule` that recur on Mondays. Assume that today is `Monday` and I want my effect to repeat starting from today. Now I can do `putString("Hello Monday").repeat(mondaySchedule)` and expect it to repeat one time today . But this will immediately print `Hello Monday` two times - first time initiated by `repeat`-specific logic and the second time by `mondaySchedule` because today is Monday, which is not what I would expect from this method. Even if today wasn't Monday it would still print `Hello Monday` one time, which is also pretty unexpected.

Another example of this confusion is issue #3850. Having `schedule` method instead of `repeat` in that case would have prevented problem of the extra delay by combining two schedules: 

```scala
putString("foo").schedule(Schedule.once ++ Schedule.spaced(100.millis))
```

New `schedule` method must not emit original effect outside flow managed by `Schedule` delays.

adamgfraser:
Resolves #4218.

The main problem with this is that since we have to run the schedule before the effect we don't have an `A` value to provide to the `Schedule` initially so we need a `Schedule` with an input type of `Any`, which works for time based recurrence but doesn't allow the schedule to continue to depend on the prior result of running the effect. Potentially we could have a variant that accepts an initial `A` value to kick off the schedule though that is not ideal because we would like the schedule to completely describe the pattern of recurrence.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -1668,17 +1668,17 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Runs this effect according to the specified schedule.
    *
-   * See [[scheduleWith]] for a variant that allows the schedule's decision to
+   * See [[scheduleFrom]] for a variant that allows the schedule's decision to
    * depend on the rsult of this effect.
    */
   final def schedule[R1 <: R, B](schedule: Schedule[R1, Any, B]): ZIO[R1 with Clock, E, B] =
-    scheduleWith(())(schedule)
+    scheduleFrom(())(schedule)
 
   /**
-   * Runs this effect according to the specified schedule with the specified
-   * initial input to the schedule.
+   * Runs this effect according to the specified schedule starting from the
+   * specified input value.
    */
-  final def scheduleWith[R1 <: R, A1 >: A, B](a: A1)(schedule: Schedule[R1, A1, B]): ZIO[R1 with Clock, E, B] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1665,6 +1665,17 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    */
   final def sandbox: ZIO[R, Cause[E], A] = foldCauseM(ZIO.fail(_), ZIO.succeedNow)
 
+  /**
+   * Runs this effect according to the specified schedule.
+   */
+  final def schedule[R1 <: R, B](schedule: Schedule[R1, Any, B]): ZIO[R1 with Clock, E, B] =
```
jdegoes, line 1671: How about this:

```scala
scheduleWith(a)(schedule)
```

Then:

```scala
def schedule(s) = scheduleWith(())(s)
```

adamgfraser: Exactly.

jdegoes (approved):
Looks great!

I might consider the names `scheduleFrom`, i.e. schedule starting from this value.

But the signatures look correct. Feel free to merge when ready!

adamgfraser:
Sounds good to me!

jdegoes (approved):

PR #4297, opened 2020-10-08
Implement a ZIO#foreach Variant for Array

### LINKED ISSUE #4198: ZIO.foreachParN when using Array does not compile due to invariance
Collections with invariant elements (like Array) need to have type ascription to use them:

```scala
ZIO.foreachParN(10)(Array(1,2,3):Iterable[Int])(i => UIO(println(i))) 
res2: ZIO[Any, Nothing, Iterable[Unit]] = <function1>
```

This means you have to add type info back in later to call other methods e.g. length. The issue is that there should be consideration for an invariant version of the methods, like for Set.

This is per a discord conversation with adamfraser.

adamgfraser:
Resolves #4198.

### DIFF
--- a/core/shared/src/main/scala/zio/IO.scala
+++ b/core/shared/src/main/scala/zio/IO.scala
@@ -1,6 +1,7 @@
 package zio
 
 import scala.concurrent.ExecutionContext
+import scala.reflect.ClassTag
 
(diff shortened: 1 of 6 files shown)


jdegoes (approved):
Looks great!

PR #4301, opened 2020-10-08
ZIO Test: Implement TestAspect.sized

adamgfraser:
An aspect that allows controlling the size parameter for generated values.


jdegoes (commented):
Might be worth changing this to `sized` to make name clashes slightly less likely (and it's no less accurate, I think?).

adamgfraser:
Good idea!

jdegoes (approved):

PR #4303, opened 2020-10-08
Add ZStream#cursor

mschuwalow:
Something I played around with.

I don't know how widely useful it is, but I think the option to traverse a stream twice / have lookahead can come in handy.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/StreamCursor.scala
+++ b/streams/shared/src/main/scala/zio/stream/StreamCursor.scala
@@ -1,0 +1,30 @@
+package zio.stream
+
+import zio._
+
+/**
+ * A cursor that allows effectful iteration over a stream. Usage is similar to
+ * [[ZStream#Pull]], but additionally offers the option to traverse a stream multiple
+ * times.
+ */
+trait StreamCursor[-R, +E, +A] {
+
+  /**
+   * Pull the next value of the stream. As with [[ZStream#Pull]], `fail(None)` signals end of stream.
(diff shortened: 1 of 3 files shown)


dpovey:
This would be really nice, I have a few use cases where I want to synchronize two different streams and advance them at different rates and none of the existing functions do this well, so you end up having to get your hands in the guts of ZStream.

streams/shared/src/main/scala/zio/stream/StreamCursor.scala
```diff
@@ -0,0 +1,30 @@
...
+ * [[ZStream#Pull]], but additionally offers the option to traverse a stream multiple
+ * times.
+ */
+trait StreamCursor[-R, +E, +A] {
+
+  /**
+   * Pull the next value of the stream. As with [[ZStream#Pull]], `fail(None)` signals end of stream.
+   * If any cursor has advanced past this point already, the effect will not be reevaluated.
+   */
+  val next: ZIO[R, Option[E], Chunk[A]]
```
dpovey, line 16: Would be useful to have a variant that you can return: `ZIO[R, Option[E], A]` like the private `BufferedPull` as this is sometimes more convenient than processing things in chunks, you need a second cursor into the current chunk.

mschuwalow: I agree that it can be convenient to have, but I feel this is better solved by working with the stream representation.
We could include a `next: ZIO[R, Option[E], A]` method, but that would mean keeping at least one additional ref around per cursor.

That makes me a bit reluctant to include it. Any other opinions?

PR #4304, opened 2020-10-09
Fix MatchError in `zio.Has.equals`

myazinn:
Odds are no one has ever compared Has[_] to other types, but I guess it should be fixed anyway.


jdegoes (approved):

jdegoes:
@myazinn Thank you, good catch!

PR #4317, opened 2020-10-14
Increase the number of generators to 10 in zipN* methods

aartigao:
Recently, as a ZIO user, I came across the need to use a `zipN` method for 6 generators.

I've created `zipN` methods that accept up to 10 different generators. The choose of 10 is somewhat arbitrary 😆 but I expect to use more than 6 generators at once in the near future.

I've also added the derivation for Magnolia to balance between `genTuple*` and `zipN*` methods.


adamgfraser (approved):

Issue #3971, opened 2020-07-14
zio-macros: support higher-kinded polymorphic services


neko-kai:
follow up to https://github.com/zio/zio/issues/3950

It would be great if `@accessible` macro supported higher-kinded polymorphic services e.g.

```scala
@accessible
trait Service[F[_, _]] {
   def x: F[Throwable, Unit]
}

// generated object:
object service extends Service[ZIO[Has[Service[IO]], ?, ?]] {
  def x = ZIO.accessM[Service[IO]](_.x)
}
```

Fristi:
Typically these tagless final algebras live in modules which don't have a zio dependency. Would be great if you could derive 'em like so

```scala
trait Service[F[_]] {
   def x: F[Unit]
}

// generated value:
val service = Derive.accessors[Service, Task] // first param (* -> *) -> *, second param effect type * -> *
```

neko-kai:
@Fristi yep, there's another ticket for that - https://github.com/zio/zio/issues/3261

PR #4332, opened 2020-10-19
Add accessibleM and accessibleMM macro annotations to support HKT

myazinn:
Closes https://github.com/zio/zio/issues/3971

The purpose of this PR is to add `@accessibleM` macro that will support HKT and should work as follows
```
@accessibleM[Task]
object MacroTest {
  trait Service[F[_]] {
    def foo: F[Int]
  }
  
  // generated by @accessibleM macro
  def foo: ZIO[Has[Service[Task]], Throwable, Int] = ZIO.accessM(_.get[Service[UIO]].foo)
}
```
Same for the `@accessibleMM` macro
```
@accessibleMM[URIO]
object MacroTest {
  trait Service[F[_, _], R] {
    def foo: F[R, Int]
  }
  
  // generated by @accessibleMM macro
  def foo[R]: ZIO[Has[Service[URIO]] with R, Nothing, Int] = ZIO.accessM(_.get[Service[URIO, R]].foo)
}
```
Both `@accessibleM` and `@accessibleMM` support polymorphic functions and traits as good as existing `@accessible` does.
For simplicity, for now it is only possible to use type aliases defined in ZIO (i.e. you can't use custom type alias nor type lambda).

### DIFF
--- a/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMMacroBase.scala
+++ b/macros/shared/src/main/scala-2.x/zio/macros/AccessibleMMacroBase.scala
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2019-2020 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.macros
+
+import scala.reflect.macros.whitebox
+
+import com.github.ghik.silencer.silent
+
+private[macros] abstract class AccessibleMMacroBase(override val c: whitebox.Context) extends AccessibleMacroBase(c) {
+
+  import c.universe._
+
+  protected def expectedTypeParams: Long
+
+  protected def aliases: Seq[Tree]
+
+  @silent("pattern var [^\\s]+ in method unapply is never used")
+  protected val tp: Tree = c.prefix.tree match {
+    case q"new ${macroNm: Ident}[$typeParam]" if macroNm.name == TypeName(macroName) =>
+      typeParam
+    case _ =>
+      abort("could not unquote annotation. Make sure you have specified the type parameter.")
+  }
+
+  protected val tpTpe: Type = {
+    val res = c.typecheck(tq"$tp", c.TYPEmode).tpe
+    if (types.keySet.contains(res))
+      res
+    else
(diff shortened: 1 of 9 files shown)


macros/shared/src/main/scala-2.x/zio/macros/AccessibleMMacroBase.scala
```diff
@@ -0,0 +1,92 @@
...
+  private lazy val types: Map[Type, Tree] =
+    aliases.map(a => c.typecheck(tq"$a", c.TYPEmode).tpe -> a).toMap
+
+  @silent("pattern var [^\\s]+ in method unapply is never used")
+  protected def macroApply(annottees: Seq[c.Tree]): MacroApply = new MacroApply(annottees) {
+
+    private val typeParamToInject = {
+      val candidates =
+        moduleInfo.serviceTypeParams.filter { tp =>
+          tp.tparams.forall(_.name == `_`) && tp.tparams.size == expectedTypeParams
```
neko-kai, line 58: > tp.tparams.forall(_.name == `_`)

That's an odd check. It's legal to name type parameters in Scala, as in `trait Service[F[e, a]]`

macros/shared/src/test/scala-2.x/zio/macros/AccessibleMMSpec.scala
```diff
@@ -0,0 +1,385 @@
...
+             }
+          """
+        })(isRight(anything))
+      },
+      testM("generates accessor for service with contravariant type param") {
+        assertM(typeCheck {
+          """
+             @accessibleMM[RIO]
+             object Module {
+               trait Service[-A, F[_, _]] {
```
neko-kai, line 231: It would be good to add tests also for variance annotations inside the F parameter as in:

```scala
trait Service[F[+_, _]] // should work and variance should be preserved in output code (e.g. Service[UIO] <: Service[Task])
trait Service[F[_, +_]] // should work and variance should be preserved in output code (e.g. Service[UIO] <: Service[Task])
trait Service[F[+_, +_]] // should work and variance should be preserved in output code (e.g. Service[UIO] <: Service[Task])

trait Service[F[-_, _]] // should fail because IO is not contravariant
trait Service[F[_, -_]] // should fail because IO is not contravariant
```

neko-kai (approved):
Aside from the 2 comments I left, I think you've done a fantastic job and it would be great to see this merged soon! 👍

macros/shared/src/main/scala-2.x/zio/macros/AccessibleMMacroBase.scala
myazinn: Yeah, it was really too strict a requirement. Fixed.

macros/shared/src/test/scala-2.x/zio/macros/AccessibleMMSpec.scala
myazinn: I've added 4 more test at the end of `AccessibleMMSpec` and `AccessibleMSpec`. They don't cover all cases due to limitations of typechecking in tests, but I hope this is better than nothing.
Thank you for your help!

neko-kai (approved):
LGTM

myazinn:
I had to push a small fix, as I remembered that ZIO streams are not included in ZIO.

neko-kai (approved):

PR #4342, opened 2020-10-21
Sink.take should also emit at end of chunk.

quelgar:
Currently if `Sink.take(n)` collects the nth element at the end of a chunk, it will not emit its output until the stream ends or another element is received. In this situation it should just emit immediately, with an empty left-over.

It's a one character fix, ~but I lack the knowledge and/or imagination to construct a good test for it~. I've been manually testing like this:

```scala
val s1 = Stream(1, 2) ++ Stream(3, 4)
runtime.unsafeRun(s1.run(Sink.take(2))) // emits Chunk(1, 2) immediately

val s2 = Stream(1, 2) ++ Stream(3, 4).mapChunksM(c => ZIO.sleep(15.seconds).as(c))
runtime.unsafeRun(s2.run(Sink.take(2)))  // master takes 15 seconds to emit Chunk(1, 2)
```

~Any suggestions on how to write a good test for this?~ Never mind, I had a brainwave that `Stream.never` is what I needed.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
@@ -1,7 +1,6 @@
 package zio.stream
 
 import scala.util.Random
-
 import zio.duration._
 import zio.stream.SinkUtils.{ findSink, sinkRaceLaw }
 import zio.stream.ZStreamGen._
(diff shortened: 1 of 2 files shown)


narma:
Good catch!

Couple of thoughts:
`idx` is not correct naming there because actually,  this value means how much left we want elements left to grab in. `n - take.length where take is an accumulator. And this, I think, was leads to the issue, because `idx` always should be less than length.

My proposal here is rename `idx` to `remain` or something like that.
Also, I think `Stream.die` is a more convenient way to test behavior like this instead of `Stream.never`, because of in bad case it will be faster.

quelgar:
@narma yes, `Stream.die` is better, done 👍

narma:
@adamgfraser @iravid Guys, any suggestions?

adamgfraser (approved):
Good catch!

PR #4349, opened 2020-10-24
ZManagedSpec: Tag: timeout, remove extraneous call

virdis:
Motivation: 

While go through [ZManagedSpec](https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala) I noticed there was an extraneous call `reserveLatch.succeed(())` at [L1211](https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala#L1211)
When [Reservation](https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala#L1209) is defined, `release` is already set to an expression where `releaseLatch` will be completed with ().  This PR addresses this issue.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
@@ -1208,7 +1208,6 @@ object ZManagedSpec extends ZIOBaseSpec {
           releaseLatch <- Promise.make[Nothing, Unit]
           managed       = ZManaged.reserve(Reservation(reserveLatch.await, _ => releaseLatch.succeed(())))
           res          <- managed.timeout(Duration.Zero).use(ZIO.succeed(_))
-          _            <- reserveLatch.succeed(())
           _            <- releaseLatch.await
         } yield assert(res)(isNone)
       },


core-tests/shared/src/test/scala/zio/ZManagedSpec.scala
```diff
@@ -1208,7 +1208,6 @@ object ZManagedSpec extends ZIOBaseSpec {
           releaseLatch <- Promise.make[Nothing, Unit]
           managed       = ZManaged.reserve(Reservation(reserveLatch.await, _ => releaseLatch.succeed(())))
           res          <- managed.timeout(Duration.Zero).use(ZIO.succeed(_))
-          _            <- reserveLatch.succeed(())
```
adamgfraser, line 1211: Don't we need this? The contract for `timeout` is that the finalizer will be run if the acquire action completes successfully, even after the timeout. But the acquire can't complete until the latch is completed. If anything it seems problematic that the test sometimes passes without this.

virdis: I see, make sense. I will go ahead and close this PR.

PR #4352, opened 2020-10-26
ZIOFn fixes

### LINKED ISSUE #2358: ZIO Execution Traces: Remove "guts" of ZIO from execution traces

robmwalsh:
will (eventually) fix #2358 

At this stage this is just a POC for tests which should pass once the functions have been properly wrapped (if I'm understanding the originally issue correctly). @jdegoes can you please advise if I'm on the right track with this?

@kitlangton & I discussed a few options for getting rid of ZIOFn but I think we need to fix this regardless so we have an accurate benchmark for performance. Wrapping more is going to cost more in overhead (more allocations + more unwrapping).

The good news is that the failing tests will show exactly where the problem is :)


core-tests/jvm/src/test/scala/zio/ZIOFnSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+import zio.test.Assertion.isEmpty
+import zio.test.{ assert, suite, testM, BoolAlgebra, FailureDetails, ZSpec }
+
+object ZIOFnSpec extends ZIOBaseSpec {
+
+  val x: UIO[Int] = UIO.succeed(1)
+  val y: UIO[Int] = UIO.succeed(2)
+
+  def findInternal(trace: List[ZTraceElement]): UIO[List[SourceLocation]] =
+    ZIO.succeed(trace.collect { case sl: SourceLocation =>
```
luis3m, line 14: `ZIO.succeed(trace.collect { case sl: SourceLocation if sl.file != "ZIOFnSpec.scala") => sl })`?

robmwalsh: well... yes :)

robmwalsh:
@jdegoes @neko-kai I think this is... very unfortunate. `flatMap` doesn't have a by name parameter, which means we can't trace the closure to the call site using `ZIOFn`. No problem, we can change it like this:
```scala
def flatMap[R1 <: R, E1 >: E, B](k: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] =
    new ZIO.FlatMap(self, k)
```
to
```scala
def flatMap[R1 <: R, E1 >: E, B](k: => A => ZIO[R1, E1, B]): ZIO[R1, E1, B] =
    new ZIO.FlatMap(self, ZIOFn(() => k)(k))
```
... except of course mima said no :'(

This is a problem across all of the combinators which don't use by name parameters, which at a quick glance is unfortunately most of them. To not do this severely cripples the usefulness of execution tracing; to do this breaks bincompat, and on the most used part of the API. This will be an issue even if we do find a way to avoid `ZIOFn` (at least as long as we're still using closures to trace call sites). I guess that means we'll need to wait for a 2.x release for this?

```
[error] zio: Failed binary compatibility check against dev.zio:zio_2.12:1.0.1! Found 27 potential problems (filtered 3)
[error]  * method flatMap(scala.Function1)zio.ZIO in class zio.ZIO#FiberRefModify's type is different in current version, where it is (scala.Function0)zio.ZIO instead of (scala.Function1)zio.ZIO
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("zio.ZIO#FiberRefModify.flatMap")
[error]  * method flatMap(scala.Function1)zio.ZIO in class zio.ZIO#CheckInterrupt's type is different in current version, where it is (scala.Function0)zio.ZIO instead of (scala.Function1)zio.ZIO
...
[error]  * method flatMap(scala.Function1)zio.ZIO in class zio.ZIO#Succeed's type is different in current version, where it is (scala.Function0)zio.ZIO instead of (scala.Function1)zio.ZIO
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("zio.ZIO#Succeed.flatMap")
[error] stack trace is suppressed; run last coreJVM / mimaReportBinaryIssues for the full output
[error] (coreJVM / mimaReportBinaryIssues) Failed binary compatibility check against dev.zio:zio_2.12:1.0.1! Found 27 potential problems (filtered 3)
[error] Total time: 2 s, completed Oct 30, 2020, 2:27:22 PM
```

iravid:
@robmwalsh No those are fine to exclude. We don't guarantee bincompat on the ZIO constructors I think.

iravid:
Ah sorry, you wanted to change flatMap. Yeah that's problematic unfortunately.

robmwalsh:
@iravid that seems very surprising to me - I would have thought nothing is more core than these combinators (to be clear, I'm talking about the combinators on the zio trait here, inherited by the classes that mima is complaining about). Is there anywhere that documents what is meant by bincompat and what is/isn't guaranteed to be compatible in zio?

robmwalsh:
Oh, sorry didn't see your follow-up there. What sort of timeline are we looking at for 2.x development to start?

robmwalsh:
@neko-kai I think your assertion is correct, but I wasn't seeing user land locations show up in the execution trace for flatmaps until I did that. One of my tests is that the call site of checkTrace should be included in the execution trace. I'll review the logic and see if I can figure out where the issue is (or if I'm just doing something wrong, which isn't unlikely since I'm learning how this works as I go along!). Great to hear about the possibilities of making changes to this in a bincompat way! I've got exciting plans that rely on it :)

jdegoes:
The overall goal is clear: reduce maintenance burden and improve quality of traces without affecting performance.

We should discuss approach:

1. Embrace intermediate classes, and multiply them.
2. Pass through the original lambda / thunk, which can cut down on the number of intermediate classes.
3. Collect stack traces at the point where methods are called.

Additionally, I see that [sourceline](https://github.com/lihaoyi/sourcecode) is now supported on Dotty. This potentially raises a new approach that uses implicits to propagate the tracing information through intermediate call sites, to eliminate "ZIO-internal junk".

We should maybe have a session to brainstorm / discuss!

robmwalsh:
@jdegoes I'm meeting with @kitlangton tomorrow to work on this but that could easily turn into a strategy/brainstorming session - I've sent you a message on discord

PR #4364, opened 2020-10-30
Updating link to migration blog

atooni:


softinio (commented):
This PR isn't just related to updating link to a migration blog only. Thanks for the additional changes but I personally prefer them to be separated in general.

If there is a problem and we need to revert due to the other changes you don't want your link update to be reverted too.

atooni:
Fair enough - I will withdraw the PR for now as there is a project reference within the resources section as well.

PR #3641, opened 2020-05-19
WIP: add empty mocks.

ioleo:
Fixes #3115. This PR introduces new branch `NoCalls` used internally to implement `Mock#empty`.

The PR is still missing wider test coverage, examples and documentation update.

Issue #3115, opened 2020-03-12
ZIO Mock: expect no calls


ioleo:
Previous ZIO Mock encoding had the `Expectation.nothing` helper which allowed to create mocked services which should not be called. As a follow-up to discussion on discord, there are a few ideas how to support that in the new encoding:

* create empty layers which don't call a Proxy, but immidietly die with MockException
```scala
val empty: ULayer[Has[Service]] = ZLayer.succeed(new Service {
    override def repositoryExists(name: Name): IO[PagerError, Name] = ZIO.die(UnexpectedCallExpection(RepositoryExists, name))
    override def releases(name: Name): IO[PagerError, List[GitHubRelease]] = ZIO.die(UnexpectedCallExpection(Releases, name))
  })
```
* intercept all calls to `A` service at `Has` level, dieing with MockException at `_.get[A]` invocation
* extend the `Expectation` with another case allowing to disallow certain calls (or all calls in extreme case), an equivalent of logical `Not` and use that to handle this case

Ideally the solution should require as little typeing from the developer as possible.

ragb:
any updates on this or any suggested workaround?

ex0ns:
Hello,
I would love to continue the work done on #3641. I re-implemented the changes on the latest version of master and I think it's more or less working but I'm no expert with zio-mocks and I would love to have some feedbacks and pointers about the implementations.

First of all I'm not sure what I should do in the `prettify` method of the Debug object.
I started by creating an object (like it was done before) that looks like:

```scala
  private[test] case class NoCalls[R <: Has[_]: Tag](mock: Mock[R]) extends Expectation[R] {
    override private[test] val invocations: List[Int] = Nil
    override private[test] val state: ExpectationState = Satisfied
  }
```

Should I open a pull requests with the changes I adapted from #3641 so we can discuss about them ?

adamgfraser:
@ex0ns That would be fantastic!

pettermachado:
@ragb I worked around this by creating an optional "fake" expectation without parameters.

```scala
object MockedService extends Mock[Api] {
  private object empty extends Effect[Unit, Nothing, Unit]

  val Empty = empty().optional

  // object SayHello extends Effect[String, Throwable, String]
  // val compose: ...
```

Setting up my testing environment can then be done like
```scala
object ServiceTest extends DefaultRunnableSpec {

  def spec = {
    suite("ServiceTest")(
      testM("get returns a playlist") {
        val testEnv = MockedService.Empty ++ live
[...]

PR #4371, opened 2020-11-02
[WIP] zio-mock: support no calls

ex0ns:
As discussed on #3115 this is updated implementation for https://github.com/zio/zio/pull/3641/files

The goal here is to be able to check that no calls are made to a specific method inside a service, this is currently a missing feature and is pretty hard/ugly to emulate with the current tools available in the lib.

The decision that was taken in the issue was to extend the Expectation with another case allowing to disallow certain calls (or all calls in extreme case).

This is my first try to implement this, I would love to have some pointers and remarks on this. For instance, I'm not sure that i got the purpose of the `invocations` list in the `Expectation` class, could someone clarify its usage ?

Please note that one of the two tests added in the `exampleJVM` project is expected to fail, but from what I understood, it should not be an issue as this is a project containing only examples, and not actual tests.

I'm also not sure at all on the behavior we should implement in the `Debug` class and how to use it.
When everything else is fine, I'll take care of the documentation for this feature as well.

Thanks

### DIFF
--- a/examples/shared/src/test/scala/zio/examples/test/MockExampleSpec.scala
+++ b/examples/shared/src/test/scala/zio/examples/test/MockExampleSpec.scala
@@ -9,6 +9,22 @@ import zio.{ clock, console, random }
 object MockExampleSpec extends DefaultRunnableSpec {
 
   def spec = suite("suite with mocks")(
+    testM("expect no call for overloaded method") {
+      def maybeConsole(invokeConsole: Boolean) =
+        zio.ZIO.when(invokeConsole)(console.putStrLn("foo"))
+
+      val maybeTest1 = maybeConsole(false).provideLayer(MockConsole.empty)
+      val maybeTest2 = maybeConsole(true).provideLayer(MockConsole.PutStrLn(equalTo("foo")))
+      assertM(maybeTest1)(isUnit) *> assertM(maybeTest2)(isUnit)
+    },
+    testM("should fail if call for unexpected method") {
+      def maybeConsole(invokeConsole: Boolean) =
+        zio.ZIO.when(invokeConsole)(console.putStrLn("foo"))
+
+      val maybeTest1 = maybeConsole(true).provideLayer(MockConsole.empty)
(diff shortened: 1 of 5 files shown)


adamgfraser:
@ex0ns Thanks for working on this! This is definitely a feature that would be great to have and has been requested multiple times.

@ioleo Would you have time to review this? If not I can dive in but you are probably the best person to do it as you were the original author of this?

ioleo (commented):
@ex0ns Hi, sorry to keep you waiting. The general model for mocks is we're building a tree of expectations, which is later mutated at runtime, to keep track of every call. If at any point the called method does not match our any of our exectations, we throw a MockException, which serves the purpose of short-circuting our program, failing the test and contains information for the renderer.

The `prettify` method recurses on `Expectation` tree and pretty-prints it. It is used only for develoment puproses. To enable it you have to flip the `Debug.enabled` flag (in ZIO codebase). If you do it and run `sbt examplesJVM/testOnly *.MockExampleSpec -- -t "should fail if call for unexpected method"` you will see the following output:

```
::: invoked zio.test.mock.MockConsole.PutStrLn
[Head]:zio.examples.test.MockExampleSpec 2s
>>>
Invocation ID: 1
NoCalls(
    state = Satisfied,
    invocations = [],,

)
::: invoked zio.test.mock.MockConsole.PutStrLn

[info] - suite with mocks
[info]   - should fail if call for unexpected method
[info]     - unexpected call to zio.test.mock.MockConsole.PutStrLn with arguments
[info]       foo
[info] Ran 1 test in 1 s 887 ms: 0 succeeded, 0 ignored, 1 failed
```

And yes, the tests in `examplesJVM` are meant to fail. These are to demonstrate output for various failure scenarios.

In general, where I left the PR the feature was implemented and covered with basic test, but I felt it needs more test coverage (for example, a complex test with 3 environments, one of which is expected no calls) and good documentation.

ioleo:
@adamgfraser This can be closed. Fixed by #4673

PR #4381, opened 2020-11-06
Add Variants of STM Collection Operators for Set

adamgfraser:


core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -1125,6 +1145,13 @@ object ZSTM {
       tx.zipWith(f(a))(_ += _)
     }.map(_.result())
 
+  /**
+   * Applies the function `f` to each element of the `Set[A]` and returns a
+   * transactional effect that produces a new `Set[B]`.
+   */
+  def foreach[R, E, A, B](in: Set[A])(f: A => ZSTM[R, E, B]): ZSTM[R, E, Set[B]] =
```
mijicd, line 1152: Should we have the one for `foreach_` as well?

mijicd (approved):
Looks good to go as soon as build is passing.

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: We don't need it because that one just takes an `Iterable` since it doesn't need to honor the "returns the same collection type" guarantee.

mijicd: Good point.

Issue #4384, opened 2020-11-07
NullPointerException in ZTrace 


atooni:
Running an application with ZMX diagnostics enabled and a panopticon instance connected, panopticon regularly executes the dump command on the ZMX server. Non-deterministically, the request fails with:

```
Fiber failed.
An unchecked error was produced.
java.lang.NullPointerException
	at zio.ZTrace.$anonfun$prettyPrint$1(ZTrace.scala:36)
	at scala.collection.immutable.List.map(List.scala:246)
	at zio.ZTrace.prettyPrint(ZTrace.scala:36)
	at zio.internal.FiberRenderer$.$anonfun$prettyPrint$5(FiberRenderer.scala:52)
	at scala.Option.fold(Option.scala:263)
	at zio.internal.FiberRenderer$.prettyPrint(FiberRenderer.scala:52)
	at zio.internal.FiberRenderer$.$anonfun$prettyPrintM$1(FiberRenderer.scala:20)
	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:344)
	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:778)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

hearnadam:
Is this still an issue with 2.x? Reopen if so.

PR #4385, opened 2020-11-07
Investigate NullPointerExceptions when pretty printing Fiber dumps #4384

atooni:
This avoids running into non deterministic null pointer exceptions when the application is monitored with zio-zmx.


core/shared/src/main/scala/zio/ZTrace.scala
```diff
@@ -33,7 +33,9 @@ final case class ZTrace(
     val stackPrint =
       if (stackTrace)
         s"Fiber:$fiberId was supposed to continue to:" ::
-          this.stackTrace.map(loc => s"  a future continuation at " + loc.prettyPrint)
+          this.stackTrace.map(loc =>
+            s"  a future continuation at " + Option(loc).map(_.prettyPrint).getOrElse("unknown")
```
jczuchnowski, line 37: This seems like it would fix the error. But I'm not sure this fix addresses the underlying problem.

`stacktrace` is a `List[ZTraceElement]`. It should not contain `null` elements. I wonder how they get there.

atooni: I agree and have thought the same thing. We might as well reject this particular change - I might be able to spend some more time getting to the bottom of it. I was kind of after a quick fix to keep ZMX happy and working

atooni:
Hi again, 

I have looked into this a bit more - definitely the list of ZTraceElements contains null values. I am now trying to get further to the root cause of that.

atooni:
In my build I have now augmented the retrieval of the reversed list in the SingleThreadedRingBuffer with some debug code. 

```
  @inline private[this] def arrayToReversedList(array: Array[AnyRef]): List[AnyRef] = {
    var i                    = 0
    var result: List[AnyRef] = Nil
    while (i < array.length) {
      result ::= array(i)
      i += 1
    }

    val nullIdx = result.indexWhere(e => Option(e).isEmpty)
    if (nullIdx >= 0) {
      println(s"Encountered null value in content of SingleThreadedRingBuffer, capacity=$capacity, size=$size, current=$current, nullAt=$nullIdx")
    }

    result
  }
```


The cases that run into the error:

```
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=23, current=23, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=14, current=15, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=15, current=15, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=13, current=13, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=27, current=27, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=12, current=12, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=14, current=14, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=15, current=15, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=25, current=25, nullAt=21
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=14, current=14, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=15, current=15, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=22, current=23, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=22, current=23, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=17, current=17, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=18, current=17, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=24, current=24, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=23, current=23, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=15, current=15, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=14, current=14, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=15, current=14, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=16, current=17, nullAt=0
Encountered null value in content of SingleThreadedRingBuffer, capacity=100, size=14, current=14, nullAt=0
```
Either Traces with null Elements are put into the RingBuffer or the RingBuffer as such causes the error

atooni:
Ok ...

I can confirm that the underlying FiberContexxt does not try to put elements into the RingBuffer that are null. 

If the underlying RingBuffer is used Concurrently, another implementation than SingleThreadedRingBuffer should be used.

atooni:
I have now reverted the changes I made earlier and have in my branch shielded the access to the ringbuffer methods with a Seamphore. I believe that kind of defeats the purpose of the SingleThreadedRingBuffer in the first place, bu it demonstrates that using the SingleThreadedRingBuffer without modification is the root cause of the NullPointerExxception.

I need to look into the RingBuffer implementation, but so far it seems that this implementation does not give me the list of elements currently in the buffer.

adamgfraser:
@atooni and I spent some time on this and narrowed down the problem. The `SingleThreadedRingBuffer` is being accessed concurrently because it is possible that the thread in `FiberContext` is writing to it at the same time as a client is reading from it as part of generating a fiber dump. This is expected behavior in some sense because `FiberContext` can ensure that only one thread is executing the main run loop but the request for a fiber dump is coming from another fiber that may be being executed by a different thread.

Specifically, the issue appears to be caused by the `dropLast` method being called from the running fiber in `popStackTrace` at the same time that another fiber requesting a fiber dump is calling `toReversedList` in `captureTrace`.

atooni:
I have now updated the test case removing the debug print statements and make sure it terminates when no error is encountered by properly timing it. 

Also, I think in the retrieval of the reversed list in SingleThreadedRingBuffer we can filter out the null values. As far as I can reconstruct it, without the filtering, the method might return a list that contains an element that is currently being dropped. This happens if the Thread dropping the last element is suspended before it can properly adjust the indices. 

The test cases incl. mine work with the filtering and without introducing semaphores.

toxicafunk (dismissed):

core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
```diff
@@ -51,7 +52,8 @@ private[zio] final class SingleThreadedRingBuffer[A <: AnyRef](capacity: Int) {
       result ::= array(i)
       i += 1
     }
-    result
+
+    result.filter(e => Option(e).isDefined)
```
adamgfraser, line 56: We can do this more efficiently if we just get each element in the `while` loop, check if it is `null`, and if not append it to the list. That way we only do a single traversal and don't allocate any additional data structures.

atooni:
I am closing this PR for now and will open a new one based on the latest changes in zio master

atooni:
ReOpened as https://github.com/zio/zio/pull/4473

PR #4386, opened 2020-11-09
Add `NeedsEnv` requirement to the `provideCustomLayer` method

myazinn:
At the moment, `provideCustomLayer` doesn't require `NeedsEnv` evidence, which leads to confusing cases.
```
val effect: UIO[Int] = UIO(1)
val layer: ULayer[Has[String]] = ZLayer.succeed("")

// won't compile as expected
val provided: UIO[Int] = effect.provideLayer(layer)

// won't compile as expected
val someProvided: URIO[ZEnv, Int] = effect.provideSomeLayer[ZEnv](layer)

// compiles without an error, even though it should behave the same way as a previous line
val customProvided: URIO[ZEnv, Int] = effect.provideCustomLayer(layer)
```
I found that `provideSomeLayer` and `NeedsEnv` requirement were introduced almost simultaneously (zio#2897 and zio#2909). I guess this is why there is a little inconsistency now.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -375,7 +375,7 @@ object ZLayerSpec extends ZIOBaseSpec {
         val layer3 = ZLayer.succeed("baz")
         val layer4 = ZLayer.fromAcquireRelease(sleep)(_ => sleep)
         val env    = layer1 ++ ((layer2 ++ layer3) >+> layer4)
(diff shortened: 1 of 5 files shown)


adamgfraser:
@myazinn I agree that this would be nice to add. However, I don't think it can be done in a binary compatible way and I don't think it is worth adding a separate operator just for this so I would suggest we put this on the list of things to do for the next ZIO version.

myazinn:
@adamgfraser I miss the time when ZIO changed with every RC :)

PR #4403, opened 2020-11-14
Prevent Fibers From Being Terminated Prematurely In ZIO#foreachPar

adamgfraser:
Currently code like this will not work:

```scala
for {
  ref    <- Ref.make(0)
  fibers <- ZIO.foreachPar(1 to 100)(_ => ref.update(_ + 1).fork)
  _      <- ZIO.foreach(fibers)(_.await)
  value  <- ref.get
} yield assert(value)(equalTo(100))
```

The issue is that `foreachPar` forks a fiber to perform each of the effects and under the supervision model when that fiber terminates it terminates the fibers in its scope, including the fiber that is updating the `Ref` in this case.

The fix is to use the `transplant` operator similar to how we did with `zipWithPar` to make sure that when fibers are forked within one of the effects in `foreachPar` they are forked on the original scope and not the scope of the fiber internally created by `foreachPar`. Conceptually this is the same issue as with `zipWithPar`, it looks like we just didn't push the change through to `foreachPar`.


mijicd (approved):

jdegoes:
Excellent catch! 🎉 💪

PR #4409, opened 2020-11-16
ZStream#retry (with tests this time)

svroonland:
Implements #3785, followup of #3902


iravid (commented):
Looks great. Not sure about the default parameter to schedule - what's your thinking on this?

svroonland:
I figured that 'retry immediately' would make sense as a default semantic for stream retrial. Not particularly attached to it though.

iravid:
Ok. I tend to agree but the need for consistency with ZIO#retry ranks higher. So i would suggest that we drop the default argument.

iravid (approved):
Thank you @svroonland!

PR #4416, opened 2020-11-20
From future with execution context operator 

politrons:
Hi all,

The idea behind this new operator in `ZIO` is to allow the transformation from scala Future into ZIO specifying  the execution context in the operator.

This need is coming since we detect that when we use this `ZIO.fromFuture` inside an `Akka actor`, after the use, we lose the ActorContext of the actor because the context change of the threads. 

That's why we need to force the operator to continue working in the same Execution context where the actor is running.

I read the use of `ZIO.lock` to specify the Execution context where the program must be executed, but is not working for some reason, but adapting the existing code and passing the implicit of the execution context where my actor is running it works.

I can continue working with this code adapted in my program, but it would be great to have as part of ZIO ecosystem to avoid me to port this code for all my projects.

 I just pick up the name of the operator but it could be completely called different. 

Regards.


politrons:
I dont know why continue failing the format. I did a `sbt fmt` There's anything else that I have to do?

politrons:
Pipeline failing all the time trying to download a library

```

[error] sbt.librarymanagement.ResolveException: Error downloading qa.hedgehog:hedgehog-core_2.12:0.6.0
[error]   Not found
[error]   Not found
[error]   not found: /home/circleci/.ivy2/local/qa.hedgehog/hedgehog-core_2.12/0.6.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/qa/hedgehog/hedgehog-core_2.12/0.6.0/hedgehog-core_2.12-0.6.0.pom
[error]   not found: https://dl.bintray.com/hedgehogqa/scala-hedgehog/qa.hedgehog/hedgehog-core_2.12/0.6.0/ivys/ivy.xml
[error] 	at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:258)
[error] 	at lmcoursier.CoursierDependencyResolution.$anonfun$update$38(CoursierDependencyResolution.scala:227)
[error] 	at scala.util.Either$LeftProjection.map(Either.scala:573)
[error] 	at lmcoursier.CoursierDependencyResolution.update(CoursierDependencyResolution.scala:227)
[error] 	at sbt.librarymanagement.DependencyResolution.update(DependencyResolution.scala:60)
[error] 	at sbt.internal.LibraryManagement$.resolve$1(LibraryManagement.scala:53)
[error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$12(LibraryManagement.scala:103)
[error] 	at sbt.util.Tracked$.$anonfun$lastOutput$1(Tracked.scala:73)
[error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$20(LibraryManagement.scala:116)
[error] 	at scala.util.control.Exception$Catch.apply(Exception.scala:228)
[error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11(LibraryManagement.scala:116)
[error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11$adapted(LibraryManagement.scala:97)
[error] 	at sbt.util.Tracked$.$anonfun$inputChangedW$1(Tracked.scala:219)
[error] 	at sbt.internal.LibraryManagement$.cachedUpdate(LibraryManagement.scala:130)
[error] 	at sbt.Classpaths$.$anonfun$updateTask0$5(Defaults.scala:3484)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
[error] 	at sbt.Execute.work(Execute.scala:291)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[error] (benchmarks / update) sbt.librarymanagement.ResolveException: Error downloading qa.hedgehog:hedgehog-core_2.12:0.6.0
```

Any idea how to solve this issue?

adamgfraser:
@politrons Thanks for reporting this. I think we need to understand the cause of the underlying issue before we implement a new operator. As you say I think `lock` should be ensuring this behavior so if it is not we need to either fix that or if there is a more grained distinction understand it and develop an appropriate resolution.

At least in my testing with a simple program like the one below `lock` does correctly cause the `Future` to be run on the specified `Executor`. Do you have a small example that would show the issue you are seeing?

```scala
package zio

import zio.console._

import scala.concurrent.{ ExecutionContext, Future }

object Example extends App {

  def run(args: List[String]): ZIO[ZEnv, Nothing, ExitCode] =
    myProgram.exitCode

  def makeFuture(ec: ExecutionContext): Future[String] = {
    val syncThread = Thread.currentThread()
    Future(Thread.currentThread.getName())(ec).map { asyncThread =>
      s"sync thread: $syncThread, asyncThread: $asyncThread"
    }(ec)
  }

  val global: ExecutionContext =
    ExecutionContext.global 

  val myProgram: ZIO[Console, Throwable, Unit] =
    ZIO.fromFuture(makeFuture).flatMap(putStrLn(_)).on(global)
}
```

politrons:
@adamgfraser Hi Adam, let me reproduce the issue with a test and I will share with you. With this test it will require to have Akka Typed 2.5.X installed since the issue happens inside an Actor.

adamgfraser:
@politrons Okay. If possible it would be great to have a reproducer without other dependencies. Akka shouldn't be able to do anything with `Future` that we can't do with the interface so we should be able to reproduce directly. That will also make it much easier to diagnose and fix.

politrons:
Finally, it cost me a lot to reproduce the issue. It's seems a problem between AkkaTestKit and ZIO when they change context of Threads.

Here the code:


```
package com.ing.mb


import akka.pattern.ask
import akka.actor.{Actor, ActorRef, ActorSystem, Props}
import akka.stream.ActorMaterializer
import akka.testkit.{DefaultTimeout, ImplicitSender, TestActorRef, TestKit}
import akka.util.Timeout
import com.ing.mb.ZIOFutureError.MyActor
import com.typesafe.config.{ConfigFactory, ConfigValueFactory}
import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpecLike}
import zio.Runtime.default
import zio.ZIO

import scala.collection.JavaConverters._
import scala.concurrent.duration._
import scala.concurrent.{Await, ExecutionContextExecutor, Future}

class ZIOFutureError extends TestKit(ActorSystem("HeartbeatSpecActorSystem",
  ConfigFactory.load("kafka/consumer.conf")
    .withValue("akka.loggers", ConfigValueFactory.fromIterable(List("akka.testkit.TestEventListener").asJava))))
  with ImplicitSender
  with DefaultTimeout
  with WordSpecLike
  with Matchers
  with BeforeAndAfterAll {

  "This test prove that using Akka TestActorRef to create the actor and run" should {
    " break when we change Thread context between Akka Test kit and ZIO" in {

      //Working
      //---------
//      val system: ActorSystem = ActorSystem("ActorSystem")
//      val myActorRef: ActorRef = system.actorOf(Props(classOf[ZIOFutureError.MyActor]))
//      val futureSuccess = myActorRef ? "Bla"
//      Await.result(futureSuccess, 5 seconds)

      //Not working
      //-------------
      val testKitActorRef: TestActorRef[MyActor] = TestActorRef(new MyActor())
      val futureBlocked = testKitActorRef ? "Bla"
      Await.result(futureBlocked, 5 seconds)
    }
  }
}

object ZIOFutureError {

  class MyActor() extends Actor {

    implicit val timeout: Timeout = Timeout(10 seconds)
    implicit val executionContext: ExecutionContextExecutor = context.system.dispatcher

    implicit val materializer: ActorMaterializer = ActorMaterializer()

    override def receive: Receive = {
      case "Bla" =>
        val program: ZIO[Any, Throwable, Unit] = for {
          _ <- ZIO.effect(println(Thread.currentThread().getName))
          value <- ZIO.fromFuture(_ => Future("hello world"))
          _ <- ZIO.effect(println(Thread.currentThread().getName))
          _ <- ZIO.effect {
            context.actorOf(Props(classOf[MyChildActor]))
          }
        } yield value
        sender() ! default.unsafeRun(program)
    }
  }

  class MyChildActor() extends Actor {

    override def preStart(): Unit = {
      println("Actor created")
    }

    override def receive: Receive = {
      case "Foo" => println("I'm working")
    }

  }

}
```

PR #4421, opened 2020-11-21
FiberRef.link - mirror FiberRef contents to ThreadLocals

dkarlinsky:
This attempts to add an ability for a `FiberRef` contents to be mirrored to non-zio infra (outside of our control) backed by `ThreadLocals`.
For example, when using NewRelic and it's [Tokens](https://docs.newrelic.com/docs/agents/java-agent/async-instrumentation/java-agent-api-asynchronous-applications#tokens) to connect async flows to what they call a transaction (think an incoming request).
In this case, we would store the new relic token in a `FiberRef` like this:
```scala
    FiberRef.make(nrToken, link = _.link())
```
The provided `link` function will be called by zio after every async boundary so any effect running in this fiber (or a fiber inheriting this ref) will run on a thread "linked" to the token

(`MDC` is a simpler example, but maybe easier to work around with libraries like `zio-logging` or `FiberRef.unsafeAsThreadLocal`)

**Note**: this is a very rough draft:
1. Not sure if running `link` in at the start of `FiberContext.evaluateNow()` covers all cases where fibers "jump" threads.
1. Need to make sure `FiberContext` additions are optimized and no overhead is incurred when `link` function is not provided  (need benchmarks) 
2. Maybe also need to support `unlink` for cleanup at the end of the boundary.

@adamgfraser this is what we discussed on discord.
@jdegoes WDYT?

### DIFF
--- a/core-tests/shared/src/test/scala/zio/FiberRefSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/FiberRefSpec.scala
@@ -1,6 +1,7 @@
 package zio
 
 import zio.FiberRefSpecUtil._
+import zio.ZRefSpec.current
 import zio.duration._
 import zio.test.Assertion._
 import zio.test.TestAspect._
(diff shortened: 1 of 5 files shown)


dkarlinsky:
@jdegoes I'm ready to work on this. 
Maybe we can add the associated issue #4528 to the Hackathon, or create a new one

PR #4427, opened 2020-11-21
zio-test: make `assert` capture expression and source location using macro

dkarlinsky:
image

Most tests verifying assert output are failing and need to be fixed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -341,6 +341,11 @@ lazy val testRunner = crossProject(JVMPlatform, JSPlatform)
   .in(file("test-sbt"))
   .settings(stdSettings("zio-test-sbt"))
   .settings(crossProjectSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "com.lihaoyi" %%% "sourcecode" % "0.2.1" % Test
+    )
+  )
   .settings(mainClass in (Test, run) := Some("zio.test.sbt.TestMain"))
   .jsSettings(libraryDependencies ++= Seq("org.scala-js" %% "scalajs-test-interface" % scalaJSVersion))
   .jvmSettings(libraryDependencies ++= Seq("org.scala-sbt" % "test-interface" % "1.0"))
--- a/build.sbt
+++ b/build.sbt
@@ -278,11 +278,6 @@ lazy val testTests = crossProject(JSPlatform, JVMPlatform)
   .settings(buildInfoSettings("zio.test"))
   .settings(skip in publish := true)
   .settings(macroExpansionSettings)
-  .settings(
-    libraryDependencies ++= Seq(
-      "com.lihaoyi" %%% "sourcecode" % "0.2.1" % Test
-    )
-  )
   .enablePlugins(BuildInfoPlugin)
 
 lazy val testTestsJVM = testTests.jvm.settings(dottySettings)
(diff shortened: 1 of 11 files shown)


adamgfraser:
@dkarlinsky Great! How does this interrelate with #4242 by @hmemcpy?

dkarlinsky:
It doesn't yet. We talked about it and agreed that I'd rebase on top of @hmemcpy 's PR, once it's merged.

adamgfraser:
@dkarlinsky @hmemcpy Can we circle up on the two open PRs we have on this? It seems like this PR has a way of capturing the source information that results in fewer changes to the public API. @hmemcpy Does the way the information is being captured here give you the ability to access it and do what you want with it?

hmemcpy:
I need to take a better look at this, but I *think* this could work for my case too! It seems to be extracting the location from the tree rather than capturing it in a separate object... Might work! This would mean I'd have to have to change the implementation of `test` and `testM` to point to the private macro-based implementation, but their public signature wouldn't change!

I'll experiment with it. Thanks!

dkarlinsky:
> It doesn't yet. We talked about it and agreed that I'd rebase on top of @hmemcpy 's PR, once it's merged.

Looking at @hmemcpy 's [PR](https://github.com/zio/zio/pull/4242) again, there doesn't seem to be an actual overlap in code.
So maybe they can proceed in parallel

hmemcpy:
I want to give it a try your way, capture the `Tree` of the `label` element of `test`/`testM`, so I don't have to introduce the implicit parameter. I can piggyback off your changes, so if you can finish this PR, I'll rebase mine on top instead!

adamgfraser:
@dkarlinsky You've definitely tackled it slightly different ways but are trying to address similar goals. Anyway, I think the approach suggested by @hmemcpy makes sense. If you can get this PR ready to go will merge it and then @hmemcpy can build on top of it.

hmemcpy:
So good news - it indeed works! I was able to capture the location info of the `test`/`testM` inside a macro without needing to change the public API!

This means that similarly to `assert`, the test methods will move into the `CompileVariant` interface, implemented by a macro that will call into the original implementation, but augmenting it with the location.

I'd still like to put my changes in next, are you close to finishing the PR, @dkarlinsky?

dkarlinsky:
> So good news - it indeed works! I was able to capture the location info of the `test`/`testM` inside a macro without needing to change the public API!
> 
> This means that similarly to `assert`, the test methods will move into the `CompileVariant` interface, implemented by a macro that will call into the original implementation, but augmenting it with the location.
> 
> I'd still like to put my changes in next, are you close to finishing the PR, @dkarlinsky?

Just need to fix all the broken DefaultTestReporterSpec` and `SummaryBuilderSpec` that verify assert output. 
Might be a little tricky, as now the output includes the absolute path to the source file.
(I would really prefer to make the path relative to the source root, but can't find anything in macro reflect api that helps figure out what the root is)

hmemcpy:
I *think* if you keep just the filename, IntelliJ will be smart enough to hyperlink it anyway. Not sure this would work for the console output. But I agree, relative path to the source would be better.

dkarlinsky:
> I _think_ if you keep just the filename, IntelliJ will be smart enough to hyperlink it anyway. Not sure this would work for the console output. But I agree, relative path to the source would be better.

Didn't work for me in console output - if it's just the file name, no hyperlink :(

hmemcpy:
Hallelujah 🤣

dkarlinsky:
> Hallelujah 🤣

Yay! Finally! 😅

hmemcpy:
@adamgfraser would love if you could take a look.

adamgfraser (approved):
I think this looks really good!

For the rendering of the results, right now we're also labeling the assert with the left hand side of the assertion. Do you think that is necessary or maybe we can just render the line of the failing test as a new line? Just wondering if it could be confusing if users can add their own labeling and now we are adding this labeling? Also happy to address the rendering in a follow up PR.

image

dkarlinsky:
> I think this looks really good!
> 
> For the rendering of the results, right now we're also labeling the assert with the left hand side of the assertion. Do you think that is necessary or maybe we can just render the line of the failing test as a new line? Just wondering if it could be confusing if users can add their own labeling and now we are adding this labeling? Also happy to address the rendering in a follow up PR.
> 
> image

@adamgfraser 
I'm not crazy about the way it is rendered either. Adding a label to the assert was just the easiest way to do this.
Ideally I think the label (expression + source location) should go into `AssertionValue` as a new field, not on the `Assertion` that it contains.
And yes, I would prefer to address this as a separate PR.

adamgfraser:
@dkarlinsky Sounds good to me!

Issue #417, opened 2018-12-03
Improve microsite


jdegoes:
# Documentation

- [x] Home — The "why"
- [x] Getting Started — Hello World
- [x] Overview — On-boards developers to the effect train
  - [x] Creating Effects
  - [x] Basic Operations
  - [x] Handling Errors
  - [x] Handling Resources
  - [x] Basic Concurrency
  - [x] Testing Effects
  - [x] Running Effects
  - [x] Background
- [ ] Data Types — Provides an introduction to every data type in ZIO
     - [ ] Concurrent
       - [ ] Promise
       - [ ] Queue
       - [ ] Ref
       - [ ] RefM
       - [ ] Semaphore
       - [x] STM
       - [x] TPromise
       - [x] TQueue
       - [x] TRef
       - [x] TSemaphore
   - [ ] Effects
     - [ ] IO
     - [ ] Task
     - [ ] TaskR
     - [ ] UIO
     - [ ] ZIO
   - [ ] Fibers
     - [ ] Cause
     - [ ] Exit
     - [ ] Fiber
     - [ ] FiberLocal
   - [ ] Miscellaneous
     - [ ] Chunk
     - [ ] Schedule
   - [ ] Resources
     - [ ] Managed
     - [ ] Reservation
   - [ ] Streams
     - [ ] Sink
     - [ ] Stream
 - [ ] Interop — Provides information on interop with the broader ecosystem
   - [ ] Cats
   - [ ] Cats Effect
   - [ ] Future
   - [ ] Java
   - [ ] Javascript
   - [ ] Monix
   - [ ] Scalaz 7.x
   - [ ] Scalaz 8
 - [ ] Use Cases — Provides information targeted at specific use cases for ZIO
   - [ ] Asynchronicity
   - [ ] Blocking IO
   - [ ] Concurrency
   - [ ] Error Management
   - [ ] Parallelism
   - [ ] Queuing
   - [ ] Retrying
   - [ ] Scheduling
   - [ ] Streaming
   - [ ] Thread Management
 - [ ] Advanced - In-depth, technical nitty-gritty for power users
   - [ ] Concurrency Model
   - [ ] Error Model
   - [ ] Interruption Model
   - [ ] Performance
   - [ ] Runtime System

sidnt:
[...]

PR #4431, opened 2020-11-22
#417 :: Improve documentation about ZLayer

timrourke:
This PR makes some improvements to the documentation around `ZLayer`.
- Cleans up a few minor grammatical errors and awkward bits
- Attempts to make some concepts slightly more clear to newcomers
- Adds links to code to facilitate deep dives on technical concepts

### DIFF
--- a/docs/datatypes/zlayer.md
+++ b/docs/datatypes/zlayer.md
@@ -5,41 +5,58 @@ title:  "ZLayer"
 
 `ZLayer[A, E, B]` describes a layer of an application: every layer in an
  application requires some services (the input) and produces some services
- (the output). Layers can be thought of as recipes for producing bundles of services, given
- their dependencies (other services).
+ (the output). Layers can be thought of as recipes for producing bundles of
+ services, given their dependencies (other services).
 
  Construction of layers can be effectful and utilize resources that must be
  acquired and safely released when the services are done being utilized.
 
- By default layers are shared, meaning that if the same layer is used twice
- the layer will only be allocated a single time. Because of their excellent composition properties, layers are the idiomatic
- way in ZIO to create services that depend on other services.
+ Layers are shared by default, meaning that if the same layer is used twice,
+ the layer will only be allocated a single time. Because of their excellent
+ composition properties, layers are the idiomatic way in ZIO to create services
+ that depend on other services.
 
 ### The simplest ZLayer application
+
+ This application demonstrates a ZIO program with a single dependency on a
+ simple string value.
+
 ```scala mdoc:silent
 import zio._
 
 object Example extends zio.App {
 
-  def run(args: List[String]): URIO[ZEnv, ExitCode] =
-    zio.provideLayer(nameLayer).as(ExitCode.success)
-
+  // Define our simple ZIO program
   val zio = for {
     name <- ZIO.access[Has[String]](_.get)
     _    <- UIO(println(s"Hello, $name!"))
   } yield ()
 
-  val nameLayer = ZLayer.succeed("Adam")}
+  // Create a ZLayer that exposes a string as a dependency to a ZIO program
+  val nameLayer = ZLayer.succeed("Adam")
+
+  // Run the program, providing the `nameLayer`
+  def run(args: List[String]): URIO[ZEnv, ExitCode] =
+    zio.provideLayer(nameLayer).as(ExitCode.success)
+}
 
 ```
 
 ### ZLayer application with dependencies 
+
+ In the following example, our ZIO application has several dependencies:
+ - `zio.clock.Clock`
+ - `zio.console.Console`
+ - `ModuleB`
+
+ `ModuleB` in turn depends upon `ModuleA`.
+
 ```scala mdoc:silent
 import zio._
-import zio.console._
 import zio.clock._
-import java.io.IOException
+import zio.console._
 import zio.duration.Duration._
+import java.io.IOException
 
 object moduleA {
   type ModuleA = Has[ModuleA.Service]
@@ -113,6 +130,13 @@ object ZLayerApp0 extends zio.App {
 ```
 
 ### ZLayer example with complex dependencies
+
+ In this example, we can see that `ModuleC` depends upon `ModuleA`, `ModuleB`,
+ and `Clock`. The layer provided to the runnable application shows how
+ dependency layers that are siblings can be combined using `++` into a single
+ combined layer, and then can be provided together as the dependency for another
+ layer.
+
 ```scala mdoc:silent
 import zio._
 import zio.clock._
--- a/docs/datatypes/zlayer.md
+++ b/docs/datatypes/zlayer.md
@@ -18,8 +18,7 @@ title:  "ZLayer"
 
 ### The simplest ZLayer application
 
- This application demonstrates a ZIO program with a single dependency on a
- simple string value.
+ This application demonstrates a ZIO program with a single dependency on a simple string value.
 
 ```scala mdoc:silent
 import zio._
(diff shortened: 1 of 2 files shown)


timrourke:
The tests that failed here are unrelated to my diff, are they potentially flaky tests?
- https://app.circleci.com/pipelines/github/zio/zio/9454/workflows/cabeca3b-2843-4126-a60d-d4e7cec562fd/jobs/96677/steps
- https://app.circleci.com/pipelines/github/zio/zio/9454/workflows/cabeca3b-2843-4126-a60d-d4e7cec562fd/jobs/96680

In the second case, it looks like a failure came from a timeout during test execution:

> `Too long with no output (exceeded 10m0s): context deadline exceeded`

docs/datatypes/zlayer.md
```diff
@@ -113,6 +130,13 @@ object ZLayerApp0 extends zio.App {
 ```
 
 ### ZLayer example with complex dependencies
+
+ In this example, we can see that `ModuleC` depends upon `ModuleA`, `ModuleB`,
+ and `Clock`. The layer provided to the runnable application shows how
+ dependency layers that are siblings can be combined using `++` into a single
+ combined layer, and then can be provided together as the dependency for another
+ layer.
+
```
adamgfraser, line 139: Minor comment and I realize it was already this way but shouldn't each of these paragraphs be a single line of markdown and then they will be rendered based on the screen versus us manually adding all these line breaks? I feel like this is going to make it look weird and bad on some screens.

docs/howto/use_modules_and_layers.md
```diff
@@ -196,7 +200,7 @@ object UserRepo {
 }
 ```
 
-Then, we define another module to perform some basic logging. We provide also a `consoleLogger` implementation that relies on zio `Console`
+Then, we define another module to perform some basic logging. We provide also a `consoleLogger` implementation that relies on Zio's `Console`:
```
adamgfraser, line 203:
```suggestion
Then, we define another module to perform some basic logging. We provide also a `consoleLogger` implementation that relies on ZIO's `Console`:
```

docs/howto/use_modules_and_layers.md
```diff
@@ -269,7 +276,8 @@ Now the requirements of our program are richer, and we can satisfy them partiall
 Notice that `provideCustomLayer` is just a special case of `provideSomeLayer`.
 
 ## Updating local dependencies
-Given a layer, it is possible to update one or more components it provides. One way is through a function to replace an old service with new service
+Given a layer, it is possible to update one or more components it provides. The `.update` method allows us to replace one
```
adamgfraser, line 279:
```suggestion
Given a layer, it is possible to update one or more components it provides. The `update` method allows us to replace one
```

docs/howto/use_modules_and_layers.md
```diff
@@ -313,24 +323,24 @@ val fullRepo: Layer[Nothing, UserRepo] = connectionLayer >>> postgresLayer
 ```
 
 ## Layers are shared in the dependency graph
-One important feature of `ZIO` layers is that they are acquired in parallel wherever possible, and they are shared. For every layer in our dependency graph, there is only one instance of it that is shared between all the layers that depend on it. If you don't want to share a module, create a fresh, non-shared version of it through `ZLayer.fresh`.
+One important feature of `ZIO` layers is that they are acquired in parallel wherever possible, and they are shared. For every layer in our dependency graph, there is only one instance of it that is shared between all the layers that depend on it. If you don't want to share a module, create a fresh, non-shared version of it through [`ZLayer.fresh`][ZLayer.fresh].
```
adamgfraser, line 326: Normally we would use `#` for methods on a trait and `.` for methods on the companion object.

adamgfraser (changes requested):
Thanks for your work on this! Very cool adding the links to the source code like this. Left some more specific comments but always great to improve documentation!

docs/datatypes/zlayer.md
timrourke: @adamgfraser do you mean in an editor, or when the markdown is rendered to HTML? FWIW, "weird and bad" is subjective, but I'd like to know what you mean.

It looks perfectly fine in my editor, and markdown renderers will usually ignore line breaks that are not explicitly followed by a space character and combine them into a single `<p>` tag.

I tend to prefer keeping my code to <= 80 chars wide because it makes the code easier to see in a small screen context, or when I have a split-screen view of some code. I don't own the coding styles for ZIO, of course, so I have no real iron in this fire.

I will update my PR to make every paragraph a single line of markdown, let me know if that is what you would prefer.

docs/howto/use_modules_and_layers.md
timrourke: Just so I'm clear, what is the semantic distinction? Not disagreeing, curious. Is it about making it clear that one thing is a method call on a concrete implementation, and another is a call on something that may not be concrete? (In this case ZLayer is a sealed abstract class, not a trait, but I assume this is what you mean?)

docs/datatypes/zlayer.md
timrourke: @adamgfraser I've only updated the whitespace of paragraphs that I added as entirely new paragraphs to avoid code churn. I think it's worthwhile to adopt a consistent code formatting convention for markdown if possible. I know little about the options but they exist:
- https://github.com/markdownlint/markdownlint
- https://github.com/DavidAnson/markdownlint

If that seems reasonable, let me know. If we can identify a consensus on formatting markdown I can try to tackle this. Would probably be ideal to unify the formatting of all markdown in a single commit to avoid whitespace commit noise.

docs/datatypes/zlayer.md
```diff
@@ -5,41 +5,58 @@ title:  "ZLayer"
...
-    zio.provideLayer(nameLayer).as(ExitCode.success)
-
+  // Define our simple ZIO program
   val zio = for {
     name <- ZIO.access[Has[String]](_.get)
     _    <- UIO(println(s"Hello, $name!"))
   } yield ()
 
-  val nameLayer = ZLayer.succeed("Adam")}
+  // Create a ZLayer that exposes a string as a dependency to a ZIO program
```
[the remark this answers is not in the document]
timrourke: This is an important distinction! Thanks for explaining that

docs/howto/use_modules_and_layers.md
adamgfraser: It is more that there are cases where the same method name is used on both a type and its companion object. For example, there is a `unit` method on the `ZIO` trait that transforms the result type to `Unit` so you could do `promise.succeed(()).unit` (ignoring the `Boolean` value returned by succeed indicating whether the promise had already been completed). There is also a `unit` method on the `ZIO` companion object that just constructs an effect that succeeds with the unit value. So it is useful to be able to distinguish between the two when talking about them.

docs/datatypes/zlayer.md
adamgfraser: Maybe it doesn't make that much of a difference. I was worried that manually adding those breaks could for example result in line breaks being inserted too early on wide screens but maybe that is dealt with separately. Looking through the rest of the documentation it looks like one line per paragraph is more common but it is not completely consistent. Anyway, thanks for all your work on this. Will merge when it passes CI.

adamgfraser (approved):
Thanks again for your work on this! Great improvements!

docs/datatypes/zlayer.md
timrourke: Thanks! I'm all for one-line-per-paragraph. I've seen folks using dedicated markdown editors (as opposed to something IDE-like) and in that context especially that is a much more natural way to format the text.

PR #4440, opened 2020-11-23
Make go and loop recursive functions naming consistent

asakaev:
Sometimes zio codebase uses "go" name for recursive functions and sometimes "loop". This PR makes naming consistent.


asakaev:
@jdegoes @iravid

adamgfraser:
I'm not sure it makes sense to have this level of standardization on method names that are purely internal implementation details and not part of the public facing API. This implies that this should be in the contributors guide somewhere so people know to do it and that reviewers should know about it and maintain it in reviews of future PRs. I don't know if the cost benefit is there.

PR #4442, opened 2020-11-24
Update reactivestreams.md

mingyang91:
It has been renamed, but the document has not been updated.


adamgfraser (approved):
Thanks for the fix!

PR #4443, opened 2020-11-24
add TRandom

mschuwalow:
Adds TRandom to the stm package.
This allows accessing random numbers inside of a transaction.

\cc @mijicd


core/shared/src/main/scala/zio/stm/TRandom.scala
```diff
@@ -0,0 +1,368 @@
...
+    if (minInclusive >= maxExclusive)
+      STM.die(new IllegalArgumentException("invalid bounds"))
+    else
+      nextDouble.map { n =>
+        val result = n * (maxExclusive - minInclusive) + minInclusive
+        if (result < maxExclusive) result
+        else Math.nextAfter(maxExclusive, Float.NegativeInfinity)
+      }
+
+  private[zio] def nextFloatBetweenWith(minInclusive: Float, maxExclusive: Float)(
```
mschuwalow, line 114: This is currently duplicated with `zio.random.Random`, but I don't think there is a good way to share this code sadly.

core/shared/src/main/scala/zio/stm/TRandom.scala
```diff
@@ -0,0 +1,368 @@
+package zio.stm
+
+import zio.random.Random
+import zio.{ BuildFrom, BuildFromOps, Chunk, ZLayer, random }
+
+object TRandom extends Serializable {
+
+  trait Service {
+    def nextBoolean: STM[Nothing, Boolean]
```
mijicd, line 9: Just a super-nitpick, I'd use `USTM` instead of `STM[Nothing, A]` wherever possible.

core/shared/src/main/scala/zio/stm/TRandom.scala
```diff
@@ -0,0 +1,368 @@
...
+      _         <- bufferRef.update(_ ++= collection)
+      swap = (i1: Int, i2: Int) =>
+               bufferRef.update { case buffer =>
+                 val tmp = buffer(i1)
+                 buffer(i1) = buffer(i2)
+                 buffer(i2) = tmp
+                 buffer
+               }
+      _ <-
+        STM.foreach((collection.size to 2 by -1).toList)((n: Int) => nextIntBounded(n).flatMap(k => swap(n - 1, k)))
```
mijicd, line 182: You can replace with `foreach_`.

core/shared/src/main/scala/zio/stm/TRandom.scala
```diff
@@ -0,0 +1,368 @@
...
+        }
+      }
+    }
+
+  private[zio] def shuffleWith[A, Collection[+Element] <: Iterable[Element]](
+    nextIntBounded: Int => STM[Nothing, Int],
+    collection: Collection[A]
+  )(implicit bf: BuildFrom[Collection[A], A, Collection[A]]): STM[Nothing, Collection[A]] =
+    for {
+      bufferRef <- TRef.make(new scala.collection.mutable.ArrayBuffer[A])
```
mijicd, line 172: Refs shouldn't hold mutables.

core/shared/src/main/scala/zio/stm/TRandom.scala
adamgfraser: Aren't they just in a static object so we can reference them?

adamgfraser: Ah but they return ZIO effects.

adamgfraser:
I wonder if it would make sense to have an implementation more like we have for `TestRandom` where we just carry the seed around and use it ourselves to generate random numbers instead of creating these `scala.util.Random` instances all the time and then calling those.

mschuwalow:
I would definitely prefer not to have our own random number generator here

core/shared/src/main/scala/zio/stm/TRandom.scala
mschuwalow: good catch! Blindly copied that from the random one

adamgfraser:
Yeah, I just wonder about performance. Because we're generating a new `scala.util.Random` every time we generate a random value and we have to generate an additional value each time because we need to both generate the value we are trying to generate and a new seed even though `scala.util.Random` has already generated a new seed internally. We already have all the code for this in `TestRandom` and it generates the same values as `scala.util.Random` for any seed. But maybe it doesn't matter in the context of a transaction. We could always pursue it later as an optimization.

core/shared/src/main/scala/zio/stm/TRandom.scala
adamgfraser: Definitely a separate issue but we shouldn't be doing that in the original version either.

mschuwalow:
Ah, I wasn't aware that we have tests for generating the same values as scala Random. Nice!

In that case it does make sense to use that one. Currently it seems to be hardcoded for normal Refs. I would prefer not to have duplicated code, so what do you think about changing it to use `TRef` internally and just commit when used for `TestRandom`? Of course this is going to negatively affect test performance slightly.

adamgfraser:
Yes, agree with you about duplicated code. Have you seen benchmarks for `TRef` versus `Ref`? Random number generation is on the hot path for property based testing so I would not want to do something that would cause a significant regression there.

Maybe we merge as is or duplicate it for now based on your preference and then pursue optimization / consolidation in a follow up PR? Unless we know that the difference really is small.

mschuwalow:
I think we can go with the unoptimized version for now and update later.
BTW, do you have an idea for the test failure on 2.11? It seems to be unhappy with the type alias in the package object.

adamgfraser:
@mschuwalow Let me take a look. We had an issue with where those got defined on Scala 2.11 when we were originally implementing `Has` and `ZLayer`. I think it may be the same problem.

adamgfraser:
Ah looks like you got it. Should we move it into a `zio.stm.random` package object? I think that would make the naming more consistent. Like `zio.random.nextInt` and `zio.stm.random.nextInt`.

mijicd:
@adamgfraser Good point. I think we can play the familiarity card and have zio. stm.random.

mschuwalow:
done. The only problem I see is that it might be a bit annoying when users want to have functions of both `zio.random` and `zio.stm.random` in scope in a module. Currently we will have name clashes.

adamgfraser (approved):

mijicd (approved):

mschuwalow:
@mijicd  can you kick the failed one?

PR #4445, opened 2020-11-24
Don't Use Mutable Data Structures Inside Ref

adamgfraser:
We were using a mutable `ArrayBuffer` inside a `Ref`. It wasn't actually causing an issue because the `Ref` wasn't being concurrently updated but that also means we don't really need a `Ref` and I think better not to do unsafe things in our code base unless we have to.

### DIFF
--- a/core/shared/src/main/scala/zio/random/package.scala
+++ b/core/shared/src/main/scala/zio/random/package.scala
@@ -149,8 +149,10 @@ package object random {
       collection: Collection[A]
     )(implicit bf: BuildFrom[Collection[A], A, Collection[A]]): UIO[Collection[A]] =
       for {
-        buffer <- ZIO.effectTotal(new scala.collection.mutable.ArrayBuffer[A])
-        _      <- ZIO.effectTotal(buffer ++= collection)
+        buffer <- ZIO.effectTotal {
+                    val buffer = new scala.collection.mutable.ArrayBuffer[A]
+                    buffer += collection
+                  }
         swap = (i1: Int, i2: Int) =>
                  ZIO.effectTotal {
                    val tmp = buffer(i1)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/random/package.scala
```diff
@@ -149,18 +149,17 @@ package object random {
       collection: Collection[A]
     )(implicit bf: BuildFrom[Collection[A], A, Collection[A]]): UIO[Collection[A]] =
       for {
-        bufferRef <- Ref.make(new scala.collection.mutable.ArrayBuffer[A])
-        _         <- bufferRef.update(_ ++= collection)
+        buffer <- ZIO.effectTotal(new scala.collection.mutable.ArrayBuffer[A])
+        _      <- ZIO.effectTotal(buffer ++= collection)
```
mijicd, line 153: I would merge these two.

core/shared/src/main/scala/zio/random/package.scala
```diff
@@ -149,18 +149,17 @@ package object random {
...
         swap = (i1: Int, i2: Int) =>
-                 bufferRef.update { case buffer =>
+                 ZIO.effectTotal {
                    val tmp = buffer(i1)
                    buffer(i1) = buffer(i2)
                    buffer(i2) = tmp
                    buffer
                  }
         _ <-
           ZIO.foreach((collection.size to 2 by -1).toList)((n: Int) => nextIntBounded(n).flatMap(k => swap(n - 1, k)))
```
mijicd, line 162: Maybe foreach_?

mijicd (dismissed):
With minor remarks.

core/shared/src/main/scala/zio/random/package.scala
adamgfraser: Yes.

core/shared/src/main/scala/zio/random/package.scala
adamgfraser: Good idea.

adamgfraser:
@mijicd Incorporated your comments.

mijicd (dismissed):

mijicd (approved):

PR #4448, opened 2020-11-25
Chunk.Arr.mkString optimization.

alexknvl:
This optimization is primarily motivated by `zio-codec`'s use of `Chunk[Char]`. Building up a `Chunk[Char]` and then converting it to `String` using `mkString` was slow due to all of the bookkeeping that `ChunkLike.mkString` was doing 
and it showed up on the profiler.


adamgfraser:
What was causing the slowdown? It looks like the only thing we are doing differently is calling `arrayIterator`, which for an `Arr` is just a single `Iterator.single(array)` call for the entire `Chunk` and then iterating over the array in a `while` loop, which in the case of an `Arr` would just execute a single time with the same logic as the optimized method.

alexknvl:
The main slowdown is in the case of `Arr.mkString` (no arguments), the version in `ChunkLike` calls `Character.toString` for each character and concatenates them all together. Of course there is no asymptotic slowdown, but the constant factor is very large.

adamgfraser:
Ah, nice! Do we need to override the other variants then or can we just delegate to `ChunkLike` if it is not a no arguments `mkString` call on an `Arr[Char]`?

alexknvl:
@adamgfraser Well, that certainly makes the change smaller :) Btw I noticed that there is no `mkString` override in `2.13` version, but that's a separate issue.

adamgfraser:
Do you want to mode the new definition of `mkString` to the Scala 2.12 version of `ChunkLike` to resolve the compilation issues so we can get this merged? Should we implement a new `render` method to get around the Scala 2.13 issue that could be available on both versions and called from ZIO Codec?

PR #4449, opened 2020-11-25
Use IntelliJ's default order for Organize Imports

sideeffffect:
Many of our contributors use IntelliJ and it's then causing friction, when they bum up against unhappy linter step in CI (even though they believe they've made everything orderly).

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -20,11 +20,11 @@ Disable {
 OrganizeImports {
   expandRelative = true
   groupedImports = Merge
+  # IntelliJ IDEA's order so that they don't fight each other
   groups = [
-    "re:javax?\\.",
-    "scala.",
-    "*",
-    "zio."
+    "java."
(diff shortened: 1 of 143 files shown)


sideeffffect:
@adamgfraser this is ready to be merged

adamgfraser:
@sideeffffect Is there a risk that these non-sensical imports are going to be created again when contributors organize imports?

sideeffffect:
@adamgfraser these non-sensical imports are very unlikely to happen, they occurred only when Scalafix was making changes to a huge number of files (like in this PR), they shouldn't happen to contributors making usual changes.
More at https://github.com/liancheng/scalafix-organize-imports/issues/137

sideeffffect:
@adamgfraser just one spurious failure that needs restarting and then we're good to go :smiley:

adamgfraser (approved):

PR #4450, opened 2020-11-25
Make Dotty happy about type tests

sideeffffect:
fix
```
the type test for ... cannot be checked at runtime
```


jdegoes:
Awesome. Can you run through lint?

adamgfraser:
@sideeffffect It seems like this is causing compilation errors because we are losing type information here.

sideeffffect:
worked around in https://github.com/zio/zio/pull/4447
The underlying problem of Dotty not being able to validate the types is still present, but it's only a warning and we keep on not having _fatal warnings_ for Dotty.

PR #4455, opened 2020-11-26
Fix Bug in ZQueue#takeBetween

### LINKED ISSUE #4436: Flaky test for ZQueue?
```
[info] Ran 64 tests in 43 s 229 ms: 64 succeeded, 0 ignored, 0 failed
[info] - ZQueueSpec
[info]   - takeBetween
[info]     - blocks until a required minimum of elements is collected
[info]       false did not satisfy isTrue()
```

https://app.circleci.com/pipelines/github/zio/zio/9454/workflows/cabeca3b-2843-4126-a60d-d4e7cec562fd/jobs/96677/steps

This test failed for the CircleCI build for JVM tests for Scala 2.12 on JDK8. Not sure what the issue is, haven't explored.

adamgfraser:
Resolves #4436.

It looks like there were two issues here.

First, there is a race condition in the test where we update the `Ref` after offering a value to the queue, so it is possible that the `getter` fiber takes a value, completes, and interrupts the `updater` fiber before it has a chance to update the `Ref`. I addressed that by simplifying the test to just assert that the list of taken values has the requisite size.

Second, it looks like there was an off by one error in `takeBetween` itself. At the point we call `takeRemainder` we need to pass it the total number of remaining values and not the number of remaining values minus one.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZQueueSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZQueueSpec.scala
@@ -264,13 +264,11 @@ object ZQueueSpec extends ZIOBaseSpec {
       },
       testM("blocks until a required minimum of elements is collected") {
         for {
-          queue   <- Queue.bounded[Int](100)
-          counter <- Ref.make(0)
-          updater  = (queue.offer(10) *> counter.update(_ + 1)).forever
-          getter   = queue.takeBetween(5, 10)
-          _       <- getter.race(updater)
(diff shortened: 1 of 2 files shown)


adamgfraser:
Tested locally with `nonFlaky(100000)`.

ghostdogpr (approved):
LGTM! cc @mijicd as I think he added this one

mijicd (approved):

PR #4456, opened 2020-11-26
Fix Flaky Transplant Test

adamgfraser:
There is a race condition in the current version where after the first latch gets completed the fiber can be interrupted in the continuation before beginning to execute `ZIO.infinity`. In this case the `onInterrupt` effect would never be triggered and the second latch would not be completed, causing the test to time out. Tested locally with `nonFlaky(100000)`.


jdegoes (approved):
Subtle! Those damn parens.

PR #4458, opened 2020-11-26
Fix refineOrDie implementation

mijicd:
As mentioned in Discord, the current `refineOrDie` implementation contains a non-exhaustive match.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2285,11 +2285,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   final def refineOrDieWith[E1](
     pf: PartialFunction[E, E1]
   )(f: E => Throwable)(implicit ev: CanFail[E]): ZStream[R, E1, O] =
-    ZStream(self.process.map(_.mapError {
-      case None                         => None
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2276,10 +2276,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   final def refineOrDie[E1](
     pf: PartialFunction[E, E1]
   )(implicit ev1: E <:< Throwable, ev2: CanFail[E]): ZStream[R, E1, O] =
-    ZStream(self.process.map(_.mapError {
-      case None                         => None
-      case Some(e) if pf.isDefinedAt(e) => Some(pf.apply(e))
-    }))
+    refineOrDieWith(pf)(ev1)
```
adamgfraser, line 2279: In `refineOrDieWith` wouldn't it be better to use `catchAll` and `die` instead of throwing?

mijicd: Actually, yes, one sec.

mijicd: Done in ae53e0504e6fc06fc6407b9aa6b5f0d498496c92

adamgfraser (approved):

PR #4459, opened 2020-11-26
New Sequence operator for collections of ZIO elements.

bFollon:
Working with ZIO in our project, we noticed that there is not a Sequence operator like in Scala's Future.
I've implemented it, based on the one from Future.

It is very useful for when you end up with a Collection of ZIO elements.


adamgfraser:
@bFollon Isn't this `collectAll`?

adamgfraser:
@bFollon Going to close this as it is getting a little stale and I believe this operator already exists in `collectAll`. If you have further questions or comments please feel free to reach out here or on Discord!

PR #4469, opened 2020-11-30
Provide random UUIDs.

erikvanoosten:
UUIDs are quite fundamental in many systems. This PR adds a UUID generator to the zio random facilities.


adamgfraser:
I'm a little worried that we are making `Random` quite a broad interface here. I tend to think it would be better if people built functionality on top of `Random` to use the base level random number generation to generate other data types, though I do see the utility here.

kitlangton:
This has been implemented 🥳  #4828

PR #4470, opened 2020-12-01
Implement Unfold Constructors For Chunk

### LINKED ISSUE #4465: Add `unfold` to `zio.Chunk`

adamgfraser:
Resolves #4465.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -1234,16 +1234,17 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
    * Constructs a `Chunk` by repeatedly applying the effectual function `f` as
    * long as it returns `Some`.
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1213,6 +1214,37 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
...
+   */
+  def unfoldM[R, E, A, S](s: S)(f: S => ZIO[R, E, Option[(A, S)]]): ZIO[R, E, Chunk[A]] = {
+
+    def go(s: S, builder: ChunkBuilder[A]): ZIO[R, E, Chunk[A]] =
+      f(s).flatMap {
+        case Some((a, s)) => go(s, builder += a)
+        case None         => ZIO.succeedNow(builder.result())
+      }
+
+    go(s, ChunkBuilder.make[A]())
```
iravid, line 1245: I think you need to use `effectSuspend` here, because the ChunkBuilder will be shared between all invocations of the ZIO returned from `unfoldM`.

adamgfraser: Good point!

jdegoes (approved):

PR #4471, opened 2020-12-01
Support Providing Layers To Schedules

adamgfraser:
Right now we have `provide` on `Schedule` but not `provideLayer`. This is a little bit awkward given how much other stuff we have that works with layers (e.g. standard ZIO environmental services are normally available as layers). This adds `provideLayer`, `provideSomeLayer`, and `provideCustomLayer` variants for `Schedule`. The layers are required not to fail since a `Schedule` does not have the structure to support the `Schedule` itself failing.


adamgfraser:
@jdegoes Can I get a review on this?

jdegoes (approved):

jdegoes:
Looks great!

PR #4473, opened 2020-12-01
Fix #4384 - Avoid null elements returned in result of SingleThreadedRingBuffer arrayToReversedList

### LINKED ISSUE #4384: NullPointerException in ZTrace

atooni:
Reopening PR based on comments from @adamgfraser and resyncing with latest master

### DIFF
--- a/core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
+++ b/core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
@@ -48,7 +48,10 @@ private[zio] final class SingleThreadedRingBuffer[A <: AnyRef](capacity: Int) {
     var i                    = 0
     var result: List[AnyRef] = Nil
     while (i < array.length) {
-      result ::= array(i)
+      // https://github.com/zio/zio/issues/4384: We might concurrently drop the last
+      // element of the Ringbuffer and create the reversed list. Therefore we omit
+      // Null elements from the result.
+      if (Option(array(i)).isDefined) result ::= array(i)
       i += 1
     }
     result
--- a/core-tests/jvm/src/test/scala/zio/RepeatedFiberDumpSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RepeatedFiberDumpSpec.scala
@@ -1,0 +1,153 @@
+package zio
+
+import java.util.concurrent.atomic.AtomicReference
+import java.util.function.UnaryOperator
+
+import scala.collection.immutable.SortedSet
+
+import zio.Supervisor.Propagation
+import zio.clock._
+import zio.console._
+import zio.duration._
+import zio.test.TestAspect._
+import zio.test._
+import zio.test.environment.{ Live, TestClock }
+
+// This test reproduces the behavior reported in https://github.com/zio/zio/issues/4384
+// The general idea is to have a supervisor that can be used to regularly pull the complete
+// Fiber.Dumps. Then we spin up some work to be done in the background, thereby causing
+// changes to one or more stack traces.
+// Eventually the dump poll will fail with a NullPointerException like in the sample output below.
+// The original issue was found while monitoring a sample app with ZMX.
+
+// Will produce something like
+// sbt:core-tests> test:testOnly *FiberDumpSpec
+// Starting test
+// DEAHGIFBKJC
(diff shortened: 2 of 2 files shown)


atooni:
Based on PR comments and findings of #4385

core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
```diff
@@ -48,7 +48,10 @@ private[zio] final class SingleThreadedRingBuffer[A <: AnyRef](capacity: Int) {
     var i                    = 0
     var result: List[AnyRef] = Nil
     while (i < array.length) {
-      result ::= array(i)
+      // https://github.com/zio/zio/issues/4384: We might concurrently drop the last
+      // element of the Ringbuffer and create the reversed list. Therefore we omit
+      // Null elements from the result.
+      if (Option(array(i)).isDefined) result ::= array(i)
```
sideeffffect, line 54: Please excuse an observation from a mere passer-by, but `Option(_)` allocates.
Wouldn't it be more frugal to just check `array(i) != null`? We're already using `var` and a `while` loop, so it wouldn't taint any pure FP here :D

atooni: No need to excuse - well spotted. I have to get used to thinking in allocations on one hand and pure FP on the other. I have just updated my PR according to your suggestion.

atooni: I have just see, the ci job failed. According to the test logs I don't think it is related to my change.

phderome: other PRs are failing mysteriously in CI as well. Let me bring that up on Discord.

core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
```diff
@@ -48,7 +48,10 @@ private[zio] final class SingleThreadedRingBuffer[A <: AnyRef](capacity: Int) {
     var i                    = 0
     var result: List[AnyRef] = Nil
     while (i < array.length) {
-      result ::= array(i)
+      // https://github.com/zio/zio/issues/4384: We might concurrently drop the last
+      // element of the Ringbuffer and create the reversed list. Therefore we omit
+      // Null elements from the result.
+      if (array(i) != null) result ::= array(i)
```
sideeffffect, line 54:
```suggestion
      val elem = array(i)
      if (elem != null) result ::= elem
```
Sorry for yet another nit pick :smile_cat: 
This indexes the array only once -- whether it has a meaningful performance impact, I can't really tell :shrug: 
Regardless, maybe it could express the idea -- that this is always about the same `array(i)` -- even clearer.

atooni: No objections. Might be inetresting to compare the differences for such a case - just out of curiosity.

Issue #4395, opened 2020-11-10
High cpu utilization of Promise.await and app degradation over time


jsfwa:
Hello, this issue is a result of investigation of related problem in [https://github.com/zio/zio-kafka/issues/238](https://github.com/zio/zio-kafka/issues/238)

For some reason during continuous recreation of Promises inner state of each new Promise(after old one was already completed) slowly growing and start to consumes too much cpu time as a results significantly affects performance of entire application.

The actual problem is in the next function and List of joiners, its starts from 0.1-2% of cpu time and goes up until you reset parent effect or app stops to respond(usually about up to 30%+ of cpu):
<img width="1226" alt="image" src="https://user-images.githubusercontent.com/7480075/98675634-93054100-236b-11eb-8449-2dce78a0b981.png">

<img width="1235" alt="image" src="https://user-images.githubusercontent.com/7480075/98676107-37878300-236c-11eb-8dec-e69b2063a9fa.png">

How it looks in real application
<img width="1711" alt="image" src="https://user-images.githubusercontent.com/7480075/98676195-57b74200-236c-11eb-9ba0-8ccb0dffe72c.png">

Reproducible scenario(takes at least 1-3 hours to see the effect), but the growing time it spends on inner joiners list is visible in cpu sampler from the beginning:
```
testM("stream should not go brrr") {

      def newJob(queue: Queue[Promise[Option[Throwable], Chunk[Int]]]) = ZStream {
        ZManaged.succeed {
          for {
            p <- Promise.make[Option[Throwable], Chunk[Int]]
            _ <- queue.offer(p)
            r <- p.await
          } yield r
        }
      }

      (for {
        jobQueue  <- Queue.unbounded[ZStream[Any, Throwable, Int]]
        feedQueue <- Queue.unbounded[Promise[Option[Throwable], Chunk[Int]]]
        
        //Actual stream of substreams
        stream = ZStream
          .fromQueue(jobQueue)
          .mapMParUnordered(50) { s =>
            //Just to affect performance faster
            s.mapMPar(10)(Task(_))
              .groupedWithin(1000, 1.millis)
              .buffer(2)
              .bufferSliding(1)
              .runDrain
          }
          .runDrain
        
        //Feed to fullfill promises
        feed = ZStream
          .fromQueue(feedQueue)
          .mapMPar(10) { x =>
            x.succeed(Chunk(1))
          }
          .runDrain
        
        fb1 <- stream.fork
        fb2 <- feed.fork
        
[...]

PR #4476, opened 2020-12-02
Fix #4457 race condition during FiberContext interruption

### LINKED ISSUE #4457: Promise.await not always interrupting properly
During research of the previous problem https://github.com/zio/zio/issues/4395 find out that we have a probability to not properly cancel `await` in https://github.com/zio/zio/blob/master/streams/shared/src/main/scala/zio/stream/ZStream.scala#L1981
```
errorSignal.await raceFirst f(a)
```
Here is simple reproducible scenario, the faster you create & interrupt fibers, the more chance that Canceler will not get fired
```
testM("joiners should properly clean up") {
      (for {
        p <- Promise.make[Throwable, Unit]

        spawner = ZStream.repeatEffect(ZIO.unit raceFirst p.await).runDrain

        _ <- spawner.fork

        joinersCount <- Task
          .succeed(p.state.get.asInstanceOf[Pending[_, _]].joiners.size)
          .tap(_ => p.succeed(()))
          .delay(60.seconds)

      } yield assert(joinersCount)(isLessThan(2)))
        .provideCustomLayer(zio.clock.Clock.live)
}
```

After debug this piece of code we are able to see such picture, some of fibers got interrupted, but not cleaned up:
image

Upd. the problem appears when we delegate interruption to another fiber:
```
for {
        p <- Promise.make[Throwable, Unit]
        _ <- ( for {
          fb <- p.await.fork
          _ <- fb.interrupt.fork //<<< If remove this fork and keep just `fb.intterupt`, everything works fine
        } yield ()).forever.fork
        _ <- p.succeed(()).delay(30.seconds)
      } yield ()
```

jsfwa:
Lock-free fix for 
https://github.com/zio/zio/issues/4457
https://github.com/zio/zio/issues/4395
https://github.com/zio/zio-kafka/issues/238
Just ensure that executions of `evaluateNow` in single FiberContext is ordered

Also run some benchmarks, no visible impact on performance obviously, would be surprised otherwise

**IOForkInterruptBenchmark**
```
New version with fix
[info] Benchmark                                        (size)   Mode  Cnt     Score    Error  Units
[info] IOForkInterruptBenchmark.catsForkInterrupt          100  thrpt   10  1886.918 ± 54.849  ops/s
[info] IOForkInterruptBenchmark.monixForkInterrupt         100  thrpt   10  1889.506 ± 54.990  ops/s
[info] IOForkInterruptBenchmark.zioForkInterrupt           100  thrpt   10   981.964 ± 18.538  ops/s
[info] IOForkInterruptBenchmark.zioTracedForkInterrupt     100  thrpt   10   839.156 ± 13.164  ops/s

Old version without fix
[info] Benchmark                                        (size)   Mode  Cnt     Score    Error  Units
[info] IOForkInterruptBenchmark.catsForkInterrupt          100  thrpt   10  1962.195 ± 51.303  ops/s
[info] IOForkInterruptBenchmark.monixForkInterrupt         100  thrpt   10  1803.737 ± 67.609  ops/s
[info] IOForkInterruptBenchmark.zioForkInterrupt           100  thrpt   10   943.982 ± 24.360  ops/s
[info] IOForkInterruptBenchmark.zioTracedForkInterrupt     100  thrpt   10   801.554 ± 10.535  ops/s
```

**ReduceAllParBenchmark**
```
New version with fix
[info] Benchmark                                 Mode  Cnt   Score   Error  Units
[info] ReduceAllParBenchmark.naiveReduceAllPar  thrpt    5  13.098 ± 1.667  ops/s
[info] ReduceAllParBenchmark.reduceAllPar       thrpt    5  28.519 ± 6.495  ops/s

Old version without fix
[info] Benchmark                                 Mode  Cnt   Score   Error  Units
[info] ReduceAllParBenchmark.naiveReduceAllPar  thrpt    5  12.868 ± 0.705  ops/s
[info] ReduceAllParBenchmark.reduceAllPar       thrpt    5  28.716 ± 6.213  ops/s
```

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/internal/FiberInterruptSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/internal/FiberInterruptSpec.scala
@@ -1,12 +1,12 @@
 package zio.internal
 
 import zio.Promise.internal.Pending
-import zio.duration._
 import zio.test.Assertion.equalTo
-import zio.test.TestAspect.flaky
-import zio.test._
 import zio.test.environment.Live
+import zio.test.{ ZSpec, assertM, suite, testM }
 import zio.{ Promise, ZIOBaseSpec }
+import zio.duration._
+import zio.test.TestAspect._
 
 object FiberInterruptSpec extends ZIOBaseSpec {
 
--- a/core-tests/jvm/src/test/scala/zio/internal/FiberInterruptSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/internal/FiberInterruptSpec.scala
@@ -1,12 +1,12 @@
 package zio.internal
 
 import zio.Promise.internal.Pending
+import zio.duration._
 import zio.test.Assertion.equalTo
+import zio.test.TestAspect._
 import zio.test.environment.Live
 import zio.test.{ ZSpec, assertM, suite, testM }
 import zio.{ Promise, ZIOBaseSpec }
-import zio.duration._
-import zio.test.TestAspect._
 
 object FiberInterruptSpec extends ZIOBaseSpec {
 
--- a/core-tests/shared/src/test/scala/zio/internal/FiberInterruptSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/internal/FiberInterruptSpec.scala
@@ -3,13 +3,13 @@ package zio.internal
 import zio.Promise.internal.Pending
 import zio.duration._
 import zio.test.Assertion.equalTo
+import zio.test.TestAspect.flaky
 import zio.test._
 import zio.test.environment.Live
 import zio.{ Promise, ZIOBaseSpec }
 
 object FiberInterruptSpec extends ZIOBaseSpec {
 
-  //Flaky
   override def spec: ZSpec[Environment, Failure] =
     suite("FiberInterruptSpec")(
       testM("must interrupt all Promise joiners") {
@@ -33,6 +33,6 @@ object FiberInterruptSpec extends ZIOBaseSpec {
         } yield joinerSize
 
         assertM(Live.live(io))(equalTo(0))
-      }
+      } @@ flaky
     )
 }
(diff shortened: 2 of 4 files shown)


jsfwa:
@adamgfraser Could you review it please?

jdegoes:
@jsfwa `evaluateNow` should never be called concurrently. The fact it is being called concurrently means we have a bug somewhere, which should be fixed directly.

I will have a chance to look at this in a week.

iravid:
@jsfwa @narma just wanted to chime in and thank you guys for investigating and narrowing this down. I know it took a while!

aartigao:
Hey @jdegoes, we are seeing a similar pattern in one of our services that use ZIO Kafka. Did you have a chance to look into this?

I tried to understand @jsfwa code in this PR but I'm afraid it's too low level for me. But if you need some heap dumps or something that can help you, just ask, I'll be willing to help in anything that help to solve this issue 👍🏽

core-tests/jvm/src/test/scala/zio/internal/FiberInterruptSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import zio.duration._
+import zio.test.Assertion.equalTo
+import zio.test.TestAspect._
+import zio.test.environment.Live
+import zio.test.{ ZSpec, assertM, suite, testM }
+import zio.{ Promise, ZIOBaseSpec }
+
+object FiberInterruptSpec extends ZIOBaseSpec {
+
+  override def spec: ZSpec[Environment, Failure] =
```
adamgfraser, line 13: Was this testing failing for you before? It is consistently passing for me on `master` even without the `nonFlaky`.

jdegoes:
`evaluateNow` cannot be called concurrently (except as noted above), and the solution to any problem in that logic cannot be to synchronize on `evaluateNow` (that's just working around a bug in a way that negatively impacts performance).

I believe @adamgfraser did work to fix one bug in this area. We should investigate to see if that fixes this, and if not, then base work off a minimal reproducer that can demonstrate concurrent calls to `evaluateNow`.

PR #4485, opened 2020-12-05
Add note for ifM to ZIO.cond

blast-hardcheese:
Useful in the case where you have a `Ref[Boolean]` and need to perform some effects, for example

```scala
ZIO.condM(isEnabled)(consumeNext, log.info("Disabled").as(StopProcessing))
```

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2477,6 +2477,8 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   /**
    * Similar to Either.cond, evaluate the predicate,
    * return the given A as success if predicate returns true, and the given E as error otherwise
+   *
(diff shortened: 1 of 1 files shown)


adamgfraser:
@blast-hardcheese This seems pretty similar to `ZIO.ifM`. Would that work for your use case?

blast-hardcheese:
@adamgfraser Hmm, it does indeed, and the implementation is identical. Let me update this PR in order to point users of `cond` over to `ifM`, since there's no natural symmetry there.

blast-hardcheese:
as for my initial example,
```scala
ZIO.ifM(isEnabled)(consumeNext, log.info("Disabled").as(StopProcessing).flip)
```
would do what I want quite nicely.

blast-hardcheese:
Done, thank you!

adamgfraser (approved):

PR #4487, opened 2020-12-06
Temporary magnolia-free implementation of zio-test-magnolia on Scala 3

vigoo:
This is an alternative Scala 3 implementation of `DeriveGen` that allows us to have snapshot releases published for `zio-test-magnolia` for Scala 3 until magnolia support is ready. Once magnolia supports Scala 3 this could be deleted to have a single implementation again.


adamgfraser (approved):
This looks great! Can you resolve conflicts and then we can get this merged?

PR #4490, opened 2020-12-07
Fix Race Condition in ZIO#effectAsyncInterrupt

### LINKED ISSUE #4457: Promise.await not always interrupting properly

adamgfraser:
Resolves #4457.

I think the issue in #4457 can be traced to a race condition in `effectAsyncInterrupt`, which is used in the implementation of `Promise`. Specifically, if `effectAsyncInterrupt` is interrupted before the asynchronous effect starts executing then no canceler will be run, but the state is not updated to ensure that the register function is not invoked. With this change the original test that joiners are properly removed passes.

Copying @jsfwa and @narma.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2926,6 +2926,17 @@ object ZIOSpec extends ZIOBaseSpec {
           _       <- fiber.interrupt
           value   <- ref.get
         } yield assert(value)(isTrue)
+      },
+      testM("effectAsyncInterrupt cancelation") {
+        for {
+          ref <- ZIO.effectTotal(new java.util.concurrent.atomic.AtomicInteger(0))
+          effect = ZIO.effectAsyncInterrupt[Any, Nothing, Any] { _ =>
+                     ref.incrementAndGet()
+                     Left(ZIO.effectTotal(ref.decrementAndGet()))
+                   }
+          _     <- ZIO.unit.race(effect).repeatN(100000)
+          value <- ZIO.effectTotal(ref.get)
+        } yield assert(value)(equalTo(0))
       }
     ) @@ zioTag(interruption),
     suite("RTS environment")(
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -2934,10 +2934,10 @@ object ZIOSpec extends ZIOBaseSpec {
                      ref.incrementAndGet()
                      Left(ZIO.effectTotal(ref.decrementAndGet()))
                    }
-          _     <- ZIO.unit.race(effect).repeatN(100000)
-          value <- ZIO.effectTotal(ref.get)
+          _     <- ZIO.unit.race(effect)
+          value <- ZIO.effectTotal(ref.get())
         } yield assert(value)(equalTo(0))
-      }
+      } @@ jvm(nonFlaky(100000))
     ) @@ zioTag(interruption),
     suite("RTS environment")(
       testM("provide is modular") {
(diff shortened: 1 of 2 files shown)


jdegoes:
@adamgfraser I think one could argue this is ultimately a problem in the implementation of `EffectAsync`, such that the race is even possible. Essentially the logic we do "after" calling register must be carefully inspected because it cannot change any fiber state detail, or else if the callback is called concurrently, then we could have 2 running `evaluateNow`.

adamgfraser:
@jdegoes I'm not sure we have enough information in `FiberContext` to handle this particular issue. My understanding is that this issue happens in the following situation:

1. On the main fiber we are interpreting `EffectAsync` and check for interruption in `enterAsync`. At that point we have not been interrupted so we continue and call `register`.
2. Then another fiber interrupts the main fiber and the logic in `onInterrupt` in `effectAsyncInterrupt` begins executing. It checks the status of `started` and finds that `started` is still `false` so it doesn't execute any cancellation logic.
3. Then within `register` we set `started` to `true` and do whatever logic is in the `register`, such as adding a joiner.  At this point we have a leak because cancellation logic will no longer be run (since it already was run in (2) above).

It seems like we are already checking for interruption immediately before invoking `register` and we need interruption to proceed concurrently with the evaluation of `register` because we potentially need the cancelation logic to prevent `register` from continuing to do something. So I don't see this as an issue where we have multiple evaluations of `evaluateNow` causing inconsistent state but where we have already implemented our own interruption logic in `effectAsyncInterrupt` through the use of the `AtomicBoolean` and `onInterrupt` and have just done it in a faulty way.

jsfwa:
Do we really have guarantees that other instructions wont be affected by such race?
The first thing that comes to my mind is Fail of uninterruptible effect right after interruption, something like this:
image
1. We have effect with mix of flatMap/uninterruptible/foldCause
2. Main `evaluationNow` executes `ZIO.Tags.InterruptStatus` [this line](https://github.com/zio/zio/blob/5fad6ba4959ef8ae387a46476f47448dc9e0c25b/core/shared/src/main/scala/zio/internal/FiberContext.scala#L468) 
2.1 We call interrupt and before (2) set `uninterruptible` we change state and executes `evaluateLater` in `kill0` [this line](https://github.com/zio/zio/blob/5fad6ba4959ef8ae387a46476f47448dc9e0c25b/core/shared/src/main/scala/zio/internal/FiberContext.scala#L1019)
3. (2) set `uninterruptible` and throws exeception
4. (2) & (3) executes `foldCause` twice [this line](https://github.com/zio/zio/blob/5fad6ba4959ef8ae387a46476f47448dc9e0c25b/core/shared/src/main/scala/zio/internal/FiberContext.scala#L453)

Its sort of conspiracy theory and looks very flaky, but I dont see any barriers, that will protect us from described or similar situation

narma:
I'd like merge this PR and address further investigation to new issue, shall we?

jdegoes (approved):

jdegoes:
We do have a guarantee that at most one thread will be inside `evaluateNow`, with one small exception: the async effect can invoke a callback before the last evaluteNow finishes. That logic needs to be inspected to ensure it does not alter any state (it was inspected at one point, of course).

PR #4498, opened 2020-12-11
Changed the actual api key with "API_KEY"

faith:
Seems like someone have exposed their api key. I replaced it with "API_KEY". Such sensitive data as api keys shouldn't be inside of the stuff hardcoded, instead one file in root directory of repository or such that contains all of them then take them from there.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -108,7 +108,7 @@ const siteConfig = {
 
   customDocsPath: 'zio-docs/target/mdoc',
   algolia: {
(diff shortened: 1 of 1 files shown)


faith:
Please notify the api key owner to reset it.

faith:
Nevermind did some research turns it out it can be a public one too or smth? But i guess a admin should review anyways

mijicd:
@softinio set this one up back in the day. I think intentional, but it's up to him to verify.

PR #4504, opened 2020-12-16
Implement Chunk#split

adamgfraser:
Splits a `Chunk` into `n` equally sized chunks. I had to implement this for dividing a collection into groups that could be processed in parallel and thought it could be useful for others.

Note that there is an existing `grouped` operator on Scala collections that does something similar but I think is slightly less useful for these types of operations. `grouped` divides a collection into groups where each group has the specified size, which means the actual number of groups may be more or less than the desired level of parallelism. This guarantees that the resulting number of groups will be equal to `n` (unless the chunk size is smaller than `n`) and that the chunks will be as close to equally sized as possible.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -701,7 +701,7 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
     val chunks    = ChunkBuilder.make[Chunk[A]]()
     var i         = 0
     while (i < remainder) {
-      val chunk = ChunkBuilder.make[A]
+      val chunk = ChunkBuilder.make[A]()
       var j     = 0
       while (j <= quotient) {
         chunk += iterator.next()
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -690,6 +690,41 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
...
 
+  /**
+   * Splits this chunk into `n` equally sized chunks.
+   */
+  final def split(n: Int): Chunk[Chunk[A]] = {
+    val length    = self.length
+    val quotient  = length / n
+    val remainder = length % n
+    val iterator  = self.iterator
+    val chunks    = ChunkBuilder.make[Chunk[A]]()
```
jdegoes, line 701: May be faster to materialize into an array, and then copy portions of it into sub arrays. But this looks good for now!

jdegoes (approved):

PR #4516, opened 2020-12-22
update to scala 3.0.0-M3

scalavision:
This is dependent on izumi-reflect:

https://github.com/zio/izumi-reflect/pull/118

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -4071,7 +4071,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
      * foreground.
      */
     def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
-      if (flag == InterruptStatus.Uninterruptible) zio.uninterruptible.disconnect.interruptible
+      if (flag == _root_.zio.InterruptStatus.Uninterruptible) zio.uninterruptible.disconnect.interruptible
       else zio.interruptStatus(flag)
   }
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ scala_213: &scala_213
   SCALA_VERSION: 2.13.3
 
 scala_dotty: &scala_dotty
-  SCALA_VERSION: 3.0.0-M2
+  SCALA_VERSION: 3.0.0-M3
 
 jdk_8: &jdk_8
   JDK_VERSION: 8
--- a/core/shared/src/main/scala-dotty/zio/=!=.scala
+++ b/core/shared/src/main/scala-dotty/zio/=!=.scala
@@ -17,7 +17,7 @@
 package zio
 
 import scala.annotation.implicitNotFound
-import scala.util.Not
+import scala.util.NotGiven
 
 /**
  * Evidence type `A` is not equal to type `B`.
@@ -27,5 +27,5 @@ import scala.util.Not
 abstract class =!=[A, B] extends Serializable
 
 object =!= {
-  implicit def neq[A, B](implicit ev: Not[A =:= B]): A =!= B = new =!=[A, B] {}
+  implicit def neq[A, B](implicit ev: NotGiven[A =:= B]): A =!= B = new =!=[A, B] {}
 }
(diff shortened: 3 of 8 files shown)


adamgfraser:
@scalavision Are you able to update this now that a new version of `izumi-reflect` has been published so we can get this in and do another ZIO release?

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4070,9 +4070,9 @@ object ZIO extends ZIOCompanionPlatformSpecific {
      * is interruptible, then the effect can be interrupted normally, in the
      * foreground.
      */
-    def force[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
-      if (flag == InterruptStatus.Uninterruptible) zio.uninterruptible.disconnect.interruptible
-      else zio.interruptStatus(flag)
+    def force[R, E, A](zioF: ZIO[R, E, A]): ZIO[R, E, A] =
```
scalavision, line 4073: The dotty compiler seemed not able to differentiate between the package  `zio` and the argument name `zio` here. Maybe we should call the `zioF` something else, or is it ok?

adamgfraser: We could rename it to `effect` to resolve the issue for now but that seems like a pretty serious problem so we should probably open an issue for it with the Dotty team.

scalavision:
I think this should work, the izumi-reflect version is correctly set at version `M12`. Do you want me to rebase against master, or are there other changes just let me know, I can work on it tomorrow, as it is getting a bit late here now :-) 

When trying to test:compile at version 3.0.0-M3, I also get this strange error:

```
error] /home/yoda/Projects/zio/zio/benchmarks/src/main/scala/zio/IOBenchmarks.scala:10:29: not found: type BootstrapRuntime
[error] object IOBenchmarks extends BootstrapRuntime {
````

I wonder if this is the dotty build not being able to propagate dependencies correctly. I have used both zio and zio-streams a bit locally without any problems, but there are a lot of bugs in the Scala 3 compiler still, so I don't think one can rely on everything here yet.

adamgfraser:
Doesn't look like there are any merge conflicts currently so don't think you need to rebase so couldn't hurt. Don't really know what to make of that error.

scalavision:
I've renamed to `effect` @adamgfraser , hehe the obvious choice of course! I am working on minimizing the issue with the scala 3 compiler. I think it is related to another problem I have encountered before. Will file an issue there when it gets ready.

scalavision:
I have added a bug report with a minimal example here in Dotty:

https://github.com/lampepfl/dotty/issues/11044

scalavision:
@adamgfraser , could you take a look again, is there anything else that needs change? It seemed to be a red flag on the ci test for dotty_jdk8, but I wasn't able to see what was wrong.

The problem with shadowing of package name turned out to be well defined behavior according to the scala specification. I renamed the arg in `def force` function of `ZIO.scala` back from `effect` to `zio`, using `_root_` as a prefix on the package name:

* https://scala-lang.org/files/archive/spec/2.13/09-top-level-definitions.html#package-references

adamgfraser (approved):
@scalavision I think this is fine for now but I do still think there is a potential issue with `force`.

You're right that the definition of `zio` in the argument to `force` shadows the definition of the `zio` package object, so we shouldn't be able to refer to `zio.InterruptStatus` without qualifying it using `_root_`. However, we're not doing that but just trying to refer to `InterruptStatus` directly, which should already be in scope. This worked on all versions of Scala 2 as well as all versions of Scala 3 before this so unless there has been some discussion about how this is an intentional change I think it is worth at least raising the issue.

Do you want to do so? I am happy to as well.

scalavision:
just go ahead @adamgfraser :-) I think I understand what you mean, and it surprised me that the compiler didn't automatically search across all available namespaces in scope (as I would describe it). I also agree, it was surprising that this suddenly should break. However, I feel I don't have the deep knowledge into the compiler semantics that I can raise this issue in a convincing way.

Anyways, thanks for all the help getting this merged! I am really looking forward to the next release of this amazing library :-)

adamgfraser:
@scalavision Okay, will do and copy you. Thanks for your work on this! I know there are a lot of people who are eager to get a release for the latest version of Scala 3.

Issue #4518, opened 2020-12-23
ZIO Test: Add new spec type


jdegoes:
To provide a more "flat" style and without any commas, it would be nice to define a mutable spec, supporting the following syntax:

```scala
object MySpec extends MutableRunnableSpec {
  suite("foo") {
    testM("name") {
    } @@ ignore

    test("name 2") {
      suite("another suite") {
      }
    }
  }
}
```

This would be accomplished by defining `suite` and `test` and `testM` methods directly inside `MutableRunnableSpec`, which would mutably append to a data structure, which would be materialized when the `def spec` of the parent runnable spec is called.

In order to support annotations, the `test` and `suite` methods would have to return some object with `@@` side-effect method, which modifies in-place the last definition by applying the test aspect.

The goal would be to preserve the existing ZIO Test syntax and feature set almost exactly, while at the same time getting away from the "comma" syntax which many find annoying.

fokot:
I will take this

hmemcpy:
Thanks so much for this! I'll add support for it in the IntelliJ plugin and it will make my teammates happy 👍 

One question: can top-level tests be added without suite?

jdegoes:
Yes, no need for top-level suites. By default there is a top-level suite that has the same name as the spec object.

hmemcpy:
Excellent! Thank you!

hmemcpy:
Also I think this issue can be closed?

PR #4523, opened 2020-12-27
MutableRunnableSpec

fokot:
Fixing issue #4518 

I was trying to do it without so many `instanceOf` but couldn't find a way.

If we have the same parameter types on the methods in `SpecBuilder` as in `Spec` will have type safety in subsequent calls on the `SpecBuilder`. Is it enough?

Should I add also proving of environments? And do we want to have typesafety there (that all tests have environment which is supertype of the suite's environment)?


fokot:
I removed type parameters from the `suite` and `test` methods itself and `SpecBuilder`. Before it was: 
```
private[test] final class SpecBuilder[R, E, T](_spec: Spec[R, E, T]) {...}

var allSuites: Chunk[SpecBuilder[_, _, _]] = Chunk.empty
var tests: Chunk[SpecBuilder[_, _, _]]     = Chunk.empty
```
Because it needed me to do casting and existential types. And even if someone would like to provide some R he would need to do casting himself, so no typesafety.

To have `test` methods with R and E maybe I can add phantom type if it is convertible to the same type as is in `allSuites` and `tests`  but it would probably require some additional method, which would be there only if criteria is met e.g.
```
  suite("first") {

    test("simple") {
      assert(1)(equalTo(1))
    }.addTest
}
```
If you know of any better approach let me know.

Another safe option would be to have add type params to `MutableRunnableSpec[R, E] extends RunnableSpec[R, E]` and then have also `DefaultMutableRunnableSpec extends MutableRunnableSpec[TestEnvironment, Any]`. This would require all test methods in `MutableRunnableSpec` to have common R and E. If you think it is useful I can add it.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+  def testM(label: String)(assertion: => ZIO[Environment, Failure, TestResult]): SpecBuilder = {
+    val test = zio.test.testM(label)(assertion)
+    val sb   = new SpecBuilder(test)
+    tests = tests :+ sb
+    sb
+  }
+
+  override def spec: ZSpec[Environment, Failure] =
+    zio.test.suite(self.getClass.getSimpleName)(
+      allSuites.map(_.spec.asInstanceOf[Spec[Any, TestFailure[Any], TestSuccess]]): _*
```
adamgfraser, line 81: You can get rid of the cast here.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+ * }
+ * }}}
+ */
+trait MutableRunnableSpec extends DefaultRunnableSpec { self =>
+
+  type ZS = ZSpec[Environment, Any]
+
+  final class SpecBuilder(_spec: ZS) {
+
+    var spec: ZS = _spec
```
adamgfraser, line 27: `private`?

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+    final def @@(
+      aspect: TestAspect[Environment, Environment, Failure, Failure]
+    ): SpecBuilder = {
+      spec = spec @@ aspect
+      this
+    }
+
+  }
+
+  var allSuites: Chunk[SpecBuilder] = Chunk.empty
```
adamgfraser, line 44: `private`?

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+  var tests: Chunk[SpecBuilder]     = Chunk.empty
```
adamgfraser, line 45: `private`?

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+   * Builds a spec with a single effectful test.
+   */
+  def testM(label: String)(assertion: => ZIO[Environment, Failure, TestResult]): SpecBuilder = {
+    val test = zio.test.testM(label)(assertion)
+    val sb   = new SpecBuilder(test)
+    tests = tests :+ sb
+    sb
+  }
+
+  override def spec: ZSpec[Environment, Failure] =
```
adamgfraser, line 79: `final`?

adamgfraser (commented):
Looks good. As you say, the biggest issue is providing the environment. Since `suite`, `test`, and `testM` now perform the side effect of adding to the spec we need whatever we are adding to have the right environment type at the time we add it, which means it is difficult for us to implement operators that provide the environment.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixed

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixes

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+ *     test("name 3")
+ *   }
+ * }
+ * }}}
+ */
+class MutableRunnableSpec[R <: Has[_]](layer: ZLayer[TestEnvironment, Throwable, R])
+    extends RunnableSpec[TestEnvironment, Any] {
+  self =>
+
+  type ZS = ZSpec[R, Any]
```
jdegoes, line 30: If the user doesn't need to see this type alias, then I would make it private, so it doesn't clutter up the Scaladoc.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  class InAnotherTestException(`type`: String, label: String)
```
jdegoes, line 32: If the user doesn't need to see this class, then I would make it private, so it doesn't clutter up the Scaladoc.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+      extends Exception(s"${`type`} `${label}` is in another test")
+      with NoStackTrace
+
+  sealed trait SpecBuilder {
+    def toSpec: ZS
+    def label: String
+  }
+
+  case class SuiteBuilder(label: String) extends SpecBuilder {
```
jdegoes, line 41: `final` or at least `sealed`

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+ *
+ *     test("name 2")
+ *   }
+ *   suite("another suite") {
+ *     test("name 3")
+ *   }
+ * }
+ * }}}
+ */
+class MutableRunnableSpec[R <: Has[_]](layer: ZLayer[TestEnvironment, Throwable, R])
```
jdegoes, line 26: There is some overlap here with what Balasz is doing.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+    final def @@(
+      aspect: TestAspect[R, R, Failure, Failure]
+    ): TestBuilder = {
+      toSpec = toSpec @@ aspect
+      this
+    }
+  }
+
+  // init SpecBuilder for this test class
+  var stack: List[SuiteBuilder] = SuiteBuilder(self.getClass.getSimpleName.stripSuffix("$")) :: Nil
```
jdegoes, line 84: These vars should definitely be hidden from view.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+  var testRunning               = false
+
+  /**
+   * Builds a suite containing a number of other specs.
+   */
+  def suite(label: String)(specs: => SpecBuilder): SuiteBuilder = {
```
jdegoes, line 90: I'd make all the methods in here `final` unless there is a good reason not to do so.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+    if (testRunning)
+      throw new InAnotherTestException("Test", label)
+    val test    = zio.test.testM(label)(assertion)
+    val builder = TestBuilder(label, test)
+    stack.head.nested = stack.head.nested :+ builder
+    builder
+  }
+
+  final override def spec: ZSpec[Environment, Failure] = {
+    testRunning = true
```
jdegoes, line 127: This is not quite correct. When `spec` is called, the effect is constructed, but it may not be run then (or ever). What you could do is attach a `@@ before` annotation to the spec which sets `testRunning = true` inside there.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: I used it only on 3 places so I removed it. And I also need different `ZSpec[Environment, Failure]`

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixed

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixed

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixed

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Fixed

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: You are right :+1: The effect is not run but is already constructed. For me it is enough so I just renamed the variable to `specBuilt`. After that point no test method should be called to prevent `test` in `test` errors. I can not test it with stact because assertion to test is call by name `final def test(label: String)(assertion: => TestResult): TestBuilder`.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Yes, I will speak with him. Maybe renaming would be good to call this 
DefaultMutableRunnableSpec -> MutableRunnableSpec
DefaultMutableRunnableSpec -> CustomMutableRunnableSpec
to have it consistent. What do you think? @jdegoes

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: I made them just `sealed`. It is fine to have them inside `MutableRunnableSpec` not to pollute `zio.test` package and I can also use type parameters from `MutableRunnableSpec` and do not need to add them to `*Builder` classes. If I make them final, I get this compilation error `The outer reference in this type test cannot be checked at run time.`

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
jdegoes: Maybe so.

jdegoes (commented):
I am good to merge if @adamgfraser is good. Thanks for your work on this!

adamgfraser (approved):
Looks good!

PR #4527, opened 2020-12-30
adding .indexOf for Chunk.scala

jyoo980:
Following conversation in #2271, adds `.indexOf` on `Chunk`


jyoo980:
Looks like I'll have to override 
```Scala
def indexOf[B >: A](elem: B, from: Int): Int
```
instead, as overriding the default `indexOf` function is deprecated in Scala 2.13

adamgfraser (changes requested):
It looks like the `indexOf` method that takes a single argument with an element is a more specialized version of the overloaded `indexOf` method that takes an element and a starting index. Can you update your PR to implement that? Should resolve the Scala 2.13 compilation issue as well.

Issue #4520, opened 2020-12-23
ZIO Test: Solve the `test` conflict with `zio.test`


jdegoes:
Currently it's a source of frustration that if one has imported `zio._`, one automatically gets `test` as a top-level package. Then if one wishes to write a pure test using the `test` function, it becomes impossible without qualification because of the name clash between the `test` package and the `test` function.

There are several ways to solve this:

1. Rename `test` function so that it has another name. However, `test` is a rather [common convention](https://scalameta.org/munit/docs/getting-started.html) across many testing libraries.
2. Rename `test` package, e.g. `ziotest`. The redundancy would be annoying `import zio.ziotest`. Perhaps the whole package could live in `ziotest`, although that would be non-idiomatic because it uses 2 words in the package name.
3. Turn `test` package into an object, which has an `apply` method, which delegates to `zio.test.test(...)`. In this case, importing `zio._` would import the `test` object located in the `zio` package. This same object would contain everything that is currently inside the package `zio.test`, but also it would have an apply function so you could call it as a function. The drawback is that importing `zio.test._` would bring into scope a method called `apply`, which could yield clashes.
4. Experiment, e.g. try to define `test` as a method in `RunnableSpec` (etc.) that delegates to the `zio.test.test` method, and see that's enough information to allow Scala to disambiguate (I don't know).
5. Other ideas?

dkarlinsky:
Adding test method to AbstractRunnableSpec seems to do the trick. 
class/object member `test` shadows `test` package (and `test.test`)

adamgfraser:
Closed by #4532.

PR #4532, opened 2021-01-03
ZIO-Test: Solve the `test` conflict with `zio.test`.

dkarlinsky:
Adding `test` method to `AbstractRunnableSpec` seems to do the trick.
#4520

### DIFF
--- a/examples/shared/src/test/scala/zio/examples/test/ExampleSpec.scala
+++ b/examples/shared/src/test/scala/zio/examples/test/ExampleSpec.scala
@@ -1,6 +1,5 @@
 package zio.examples.test
 
-import zio._
 import zio.test._
 
 object ExampleSpec extends DefaultRunnableSpec {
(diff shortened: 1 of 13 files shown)


examples/shared/src/test/scala/zio/examples/test/ExampleSpec.scala
```diff
@@ -1,6 +1,7 @@
 package zio.examples.test
 
-import zio.test.{ assert, suite, test, Assertion, DefaultRunnableSpec }
+import zio._
+import test._
```
adamgfraser, line 4:
```suggestion
import zio.test._
```

adamgfraser (changes requested):
Looks great! As proof that this solves the conflict and prevent regressions it would be good to replace usages of `zio.test.test` in our own internal test suites with `test`.

jdegoes:
Awesome!!!

adamgfraser (approved):

PR #4533, opened 2021-01-04
zio-test: improvements for assert capturing expression and location

dkarlinsky:
This is a refinement of #4427
Moves the captured expression into the description of the value being asserted on.
Moves the source location to a separate line.
image

Also made rendered expressions much cleaner - closer to how you would normally write the code.
Under dotty the actual source code is shown.
Example (for Scala 2.x)
``` scala
//before
Tuple4.apply[String, Int, String, Boolean]("foo", 123, Test.this.fooVal, false)
// now 
("foo", 123, fooVal, false)
```

### DIFF
--- a/core/shared/src/main/scala-2.13+/zio/ChunkBuilder.scala
+++ b/core/shared/src/main/scala-2.13+/zio/ChunkBuilder.scala
@@ -16,7 +16,7 @@
 
 package zio
 
-import scala.collection.mutable.{ ArrayBuilder, Builder }
+import scala.collection.mutable.{ArrayBuilder, Builder}
 import scala.{
   Boolean => SBoolean,
   Byte => SByte,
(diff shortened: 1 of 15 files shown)


adamgfraser (approved):
Nice!

jdegoes (approved):

hmemcpy:
@dkarlinsky nice! Just noticed this when rebasing my changes. Did you make this adjustment for the dotty variant as well?

dkarlinsky:
> @dkarlinsky nice! Just noticed this when rebasing my changes. Did you make this adjustment for the dotty variant as well?

Yes, and in dotty you can just get the actual source:
```scala
Term.of(expr).pos.sourceCode
```

hmemcpy:
Ah cool! I saw that specific line was rewritten by @sideeffffect (https://github.com/zio/zio/blob/master/test/shared/src/main/scala-dotty/zio/test/CompileVariants.scala#L105), but I guess it's for compat with the latest dotty.

Thanks!

sideeffffect:
Exactly, Scala 3.0.0-M3's method `sourceCode` changed from returning `String` to `Option[String]`.
I added a little `.get` to make it all work out and unblock the CI. But if it turns out that in some cases it indeed is `None`, then we would have to deal with that -- using `.getOrElse("")` or something like that I presume?

PR #4534, opened 2021-01-04
Removed second argument group on `Chunk.nonEmptyOrElse`

ahoy-jon:
When using `Chunk.nonEmptyOrElse`, the inference cannot select the expected type.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -664,7 +664,7 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
   /**
    * Runs `fn` if a `chunk` is not empty or returns default value
    */
-  def nonEmptyOrElse[B](ifEmpty: => B)(fn: NonEmptyChunk[A] => B): B =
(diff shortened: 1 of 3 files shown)


ahoy-jon:
@adamgfraser Caught this kind of behavior using the method, especially when using variant types such as `ZIO[-R, +E, +A]` as a return type.

Maybe I don't understand some design choices on this one!

Cheers,
Jonathan

adamgfraser:
@ahoy-jon I think the goal was consistency with methods like `Option#fold` that allow specifying a default value if a value does not exist and otherwise doing something with the value if it exists. Admittedly that method does have the same issue typically requiring you to specify type parameters or add a type ascription to the first argument. Not sure if that is worth the worse type inference.

The other issue here is I don't think this change can be binary compatible. Not sure why the MiMa checks aren't picking it up but this method has been in the API for the last eight months so don't think we can change the method signature without potentially breaking existing code, though we could definitely consider for 2.0.

ahoy-jon:
👍 for ZIO 2.0, or a ZIO 1.1 (it could be binary compatible, but definitely not source compatible). Closing it for now

PR #4539, opened 2021-01-07
Fix #4472 - make findMatching pure

### LINKED ISSUE #4472: ProxyFactory is not thread safe
Proxy factory can miss successful calls to mocked methods if methods are called in parallel.

This test succeeds (so both methods are called successfully), note `atMost`:
```scala
def method = Srv.mPar("b").zipPar(Srv.mPar("c"))
...
val mockEnv =
  SrvMock.MPar(equalTo("b"), value("bRes")).atMost(1) &&
    SrvMock.MPar(equalTo("c"), value("cRes"))

assertM(method)(equalTo(("bRes", "cRes"))).provideLayer(mockEnv)
```

This one (without `atMost`) fails on expectations check, so one successful call is missed:
```scala
val mockEnv =
  SrvMock.MPar(equalTo("b"), value("bRes")) &&
    SrvMock.MPar(equalTo("c"), value("cRes"))

assertM(method)(equalTo(("bRes", "cRes"))).provideLayer(mockEnv)
```

```
[info]     - unsatisfied expectations
[info]       in any order
[info]         .IssueSpec.SrvMock.MPar with arguments equalTo(b)
```

See [zio-test-issue-4472](https://github.com/senia-psm/zio-test-issue-4472) for reproducible example.

I believe this issue might be caused by [ProxyFactory implementation](https://github.com/zio/zio/blob/master/test/shared/src/main/scala/zio/test/mock/internal/ProxyFactory.scala#L273-L277), namely:
```scala
root    <- state.expectationRef.get
...
_       <- state.expectationRef.set(matched.expectation)
```

senia-psm:
Fix #4472
Make `findMatching` pure and use atomic `Ref::modify` instead of `get`/`set` pair.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/mock/ParallelMockSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/mock/ParallelMockSpec.scala
@@ -1,0 +1,33 @@
+package zio.test.mock
+
+import zio.ZIO
+import zio.test._
+import zio.test.mock.module.{ImpureModule, ImpureModuleMock}
+
+object ParallelMockSpec extends ZIOBaseSpec {
+
+  import Assertion._
+  import Expectation._
+
+  def spec: Spec[Any, TestFailure[Any], TestSuccess] =
+    suite("ParallelMockSpec")(
+      testM("Count calls for the same expectation") {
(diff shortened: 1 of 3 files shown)


jdegoes (approved):

jdegoes:
Thank you!

PR #4544, opened 2021-01-09
spelling fix in a comment (Schedule.scala)

phderome:
recurrance to recurrence


adamgfraser:
@softinio Looks like we're having the issue again where CircleCI is trying to run on the contributor's account instead of ZIO's account. Sorry I don't remember exactly how we resolved this in the past. Could you help here?

PR #4545, opened 2021-01-10
primitive evaluation of delay

phderome:
Update Schedule.scala for a primitive evaluation of delay. The proposed change seems to be more direct. We just have to make sure tests pass and that I am not getting order wrong between interval and now.


jdegoes:
@phderome Please re-open if you get a chance to fix the issues!

Issue #4207, opened 2020-09-16
Add basic combinators (map,flatMap...) to Promise


dkarlinsky:
I've recently needed a to `map` over a promise and discovered there was no `map` or `flatMap` defined.
Is there a reason for this? 
I guess `map` on `zio.Promise` is not exactly the same as a map on `ZIO` (or Scala `Future`) because a promise also has the `complete()` side.
So if we have:
```scala
 promiseA <- Promise.make[E, A]
 promiseB <- promiseA.map(f)
```
`promiseB` can be completed directly, independently of `promiseA`, while completing `promiseA` will always complete `promiseB`
Which might be confusing behaviour...

Anyway I implemented `map` [here](https://github.com/wix/greyhound/blob/master/core/src/main/scala/com/wixpress/dst/greyhound/core/zioutils/PromiseSyntax.scala)

If it makes sense to add it to `Promise`, I'd be happy to submit a PR.

natansil:
@iravid WDYT does `map` makes sense for `Promise`?

iravid:
Sure why not. You can probably do `zipWith` too.

I admit that most of the time when I was thinking of these combinators for Promise, I found that it made more sense to hand out the `p.await` value to subscribers rather than the entire promise. Then you can use the entire range of `ZIO` combinators.

natansil:
Interesting. So if `ZIO` is a generalized form of `Promise` when do you use `Promise` specifically?

iravid:
Sorry that's not what I meant. Let me clarify.

A promise `Promise[E, A]` is usually subscribed to by one fiber and resolved by another. Those two fibers don't necessarily have to hold on to  the entire promise; the subscriber can, equivalently, just use the ZIO value of `p.await` (`UIO[Exit[E, A]]`). 

So if the subscriber is holding on to just the `UIO[Exit[E, A]]`, it can use the entire range of ZIO combinators to transform it before evaluating it. Which subsequently makes implementing those combinators directly on `Promise` redundant in some cases.

natansil:
Oh I see. In our case we don't want to await on the first promise (we want to batch the effects first), but we do want to map the result. so seems like `p.await` is not an option

hearnadam:
I don't think these will be pursued.

PR #4546, opened 2021-01-10
Added 'Promise' methods 'map', 'flatMap', 'zip', 'zipWith' (#4207))

arnoldlacko:
Based on the discussion in #4207 I've implemented a few combinators for `Promise` (`map`, `flatMap`, `zip`, `zipWith`).
A concern I have with this is that the type signatures are quite unconventional (due to the usage of `UIO`), which is why I'm wondering if renaming `map` and `flatMap` to something else might be a good idea. I'm thinking perhaps `flatMap` could be something like `andThen`, while for `map` I can't think of a good alternative. Let me know please about your thoughts, I'm happy to adjust (or remove if needed).
@iravid

### DIFF
--- a/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/PromiseSpec.scala
@@ -74,12 +74,144 @@ object PromiseSpec extends ZIOBaseSpec {
         v <- p.await
       } yield assert(s)(isFalse) && assert(v)(equalTo(1))
     },
+    testM("flatMap a promise that completes") {
+      def f(v: Int) =
+        for {
+          p <- Promise.make[Exception, Int]
+          _ <- p.succeed(v * 2)
+        } yield p
+      for {
+        a      <- Promise.make[Exception, Int]
+        b      <- a.flatMap(f)
+        _      <- a.succeed(1)
+        result <- b.await.run
+      } yield assert(result)(succeeds(equalTo(2)))
+    },
+    testM("flatMap a promise that fails") {
+      val exception = new Exception
+      for {
+        a      <- Promise.make[Exception, Int]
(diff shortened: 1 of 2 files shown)


arnoldlacko:
Closing temporarily to do some further checks (I suspect using `ZIO.to` this way might not be correct here)

arnoldlacko:
> Closing temporarily to do some further checks (I suspect using `ZIO.to` this way might not be correct here)

Resolved my concern, it seems fine. Reopening PR. :)

jdegoes:
There are some drawbacks to these operations:

1. The `map` / `flatMap` / etc. methods have well-defined meanings and laws around them, which can't be satisfied by these (identically-named) operators.
2. Reasoning about these operators is somewhat difficult. For example, take `map`. Ordinarily, `map` will return a new value, in a way that is completely disconnected from the prior value. But in this case, `Promise#map` returns a new promise, but if someone completes the old promise, it will complete the new promise too (so there are 2 separate ways to complete the promise).

Do you have use cases we could look at, which could help us determine if these operators would be useful in a very broad context? If they are targeted toward niche use cases, it may be more useful to have them in a third-party library, rather than adding them to Promise directly.

arnoldlacko:
@jdegoes thanks for the feedback. I don't actually have use cases where these operators would be strictly necessary. I've implemented them mostly as an exercise based on the discussion in #4207 and I was curious to see if it would be useful to have them added to `Promise`. I see your point about these operators having well defined meanings, which is why I was wondering if the proposed ones would make sense under different names (struggling to come up with appropriate ones though), but I am happy to close this PR in case it seems better not to have these on `Promise` at all.

arnoldlacko:
Closing this one for now due to the concerns discussed above.

arnoldlacko:
Thanks for the feedback @jdegoes . It was a nice exercise anyway, so no effort is wasted. :)

PR #4551, opened 2021-01-12
Fix master (Dotty)

sideeffffect:


adamgfraser (approved):

sideeffffect:
It seems like we indeed have a red master. That is a bit painful when it happens, because it blocks any PRs until the master is made green again.

This can we prevented by a properly set-up CI. Unfortunately CircleCI doesn't make that easy for us, so we would have to do it ourselves, something like this https://github.com/zio/zio/pull/4262
Or we would have to mark the checkbox in GitHub `master` branch protection rule _Require branches to be up to date_.

JFrog/Bintray repository also hasn't proven itself to be reliable, so switching to Ubuntu as the source of Java.

sideeffffect:
Ready to be merged.

adamgfraser (approved):

PR #4554, opened 2021-01-13
microsecond precision for Clock on JVM

### LINKED ISSUE #4550: ZIO Clock should provide microsecond precision
Since JDK9, JVM supports microsecond precision timestamps in `java.time`. `zio.clock` uses `System.currentTimeMillis` and hence is limited to millisecond precision. If ZIO `Clock` was implemented on top of `java.time.Instant`, it would automatically use the precision available on the current platform. As we require at least JDK8, `java.time` is always available on the JVM.

Motivation: The current lower precision for example leads to zio-telemetry reporting negative span lengths for short running effects.

runtologist:
/fixes #4550


runtologist:
I'll need to fix the test to only run um JDK > 8.

adamgfraser:
@runtologist Looks like there is a compilation error on Dotty. I think merging upstream changes should fix.

core-tests/jvm/src/test/scala/zio/clock/ClockSpecJVM.scala
```diff
@@ -0,0 +1,26 @@
...
+      for {
+        jreVersion <- property("java.version")
+        a          <- clock.currentTime(unit)
+        _          <- ZIO.foreach(1 to 1000)(_ => UIO.unit) // just pass some time
+        b          <- clock.currentTime(unit)
+      } yield
+      // This test should only run on JRE >= 9, which is when microsecond precision was introduced.
+      // Versions of JREs < 9 started with s"1.${majorVersion}", then with JEP 223 they switched to semantic versioning.
+      assert(jreVersion)(isSome(startsWithString("1."))) || assert((b - a) % 1000)(not(equalTo(0L)))
+    }.provideLayer(Clock.live ++ System.live) @@ TestAspect.flaky
```
adamgfraser, line 24: You can use `TestAspect.ifProp` to clean this up slightly. Also if you use `Live.live(clock.currentTime(unit)` you can avoid having to provide any layers. Why is this test flaky?

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -27,18 +27,41 @@ package object clock {
 
   object Clock extends PlatformSpecific with Serializable {
     trait Service extends Serializable {
+
       def currentTime(unit: TimeUnit): UIO[Long]
+
+      // Could be UIO. We keep IO to preserve binary compatibility.
```
adamgfraser, line 33: Can this really be `UIO`? I know we changed this at what point because we observed that it really did throw `DateTimeException` when the system time zone wasn't in the list that was supported by the `java-time` implementation.

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -27,18 +27,41 @@ package object clock {
...
       def currentDateTime: IO[DateTimeException, OffsetDateTime]
-      def instant: UIO[java.time.Instant]                               = currentTime(TimeUnit.MILLISECONDS).map(java.time.Instant.ofEpochMilli(_))
+
+      // The implementation is only here to preserve binary compatibility.
+      def instant: UIO[java.time.Instant] = currentTime(TimeUnit.MILLISECONDS).map(java.time.Instant.ofEpochMilli(_))
+
+      // This could be a UIO. We keep IO to preserve binary compatibility.
+      // The implementation is only here to preserve binary compatibility.
       def localDateTime: IO[DateTimeException, java.time.LocalDateTime] = currentDateTime.map(_.toLocalDateTime())
```
adamgfraser, line 41: Same comment as above.

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -27,18 +27,41 @@ package object clock {
...
-          IO.effectTotal(System.currentTimeMillis).map(l => unit.convert(l, TimeUnit.MILLISECONDS))
+          instant.map { inst =>
+            // A nicer solution without loss of precision or range would be
+            // unit.toChronoUnit.between(Instant.EPOCH, inst)
+            // However, ChronoUnit is not available on all platforms
+            unit match {
+              case TimeUnit.NANOSECONDS =>
+                val micros = inst.toEpochMilli() * 1000000 + inst.getNano()
+                unit.convert(micros, TimeUnit.NANOSECONDS)
+              case TimeUnit.NANOSECONDS | TimeUnit.MICROSECONDS =>
```
adamgfraser, line 59: Not sure I understand this line. If we matched on `NANOSECONDS` above shouldn't we just be matching on `MICROSECONDS` here?

core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -27,18 +27,41 @@ package object clock {
...
     trait Service extends Serializable {
+
       def currentTime(unit: TimeUnit): UIO[Long]
+
+      // Could be UIO. We keep IO to preserve binary compatibility.
       def currentDateTime: IO[DateTimeException, OffsetDateTime]
-      def instant: UIO[java.time.Instant]                               = currentTime(TimeUnit.MILLISECONDS).map(java.time.Instant.ofEpochMilli(_))
+
+      // The implementation is only here to preserve binary compatibility.
+      def instant: UIO[java.time.Instant] = currentTime(TimeUnit.MILLISECONDS).map(java.time.Instant.ofEpochMilli(_))
```
adamgfraser, line 37: Maybe keep these abstract and provide helper methods for default implementations like we do in other services? I think providing default implementations potentially results in counterintuitive behavior with like mock implementations.

adamgfraser (changes requested):
Looks good. Couple of specific comments but great idea!

core/shared/src/main/scala/zio/clock/package.scala
runtologist: The previous implementation used `ZoneId.systemDefault` in `currentDateTime`, which may throw a `DateTimeException`. The new implementation uses `Instant.now`, which does not throw according to the docs.

core/shared/src/main/scala/zio/clock/package.scala
runtologist: The previous implementation was based on `currentDateTime`, which could a `DateTimeException`. The new implementation directly uses `LocalDateTime.now`, which does not throw according to the docs.

core/shared/src/main/scala/zio/clock/package.scala
runtologist: Good catch. That's a leftover. Initially I only had the second case, but the decided to add the first one in order to keep the loss in range and precision as low as possible.

core/shared/src/main/scala/zio/clock/package.scala
runtologist: I'd also prefer to do that, but it breaks binary compatibility. If we decide that's OK here, I'll remove the default implementation.

core-tests/jvm/src/test/scala/zio/clock/ClockSpecJVM.scala
runtologist: `TestAspect.ifProp` is great, thanks. One of these small tools you rarely need, but that make life easier when you do.

This test tests the live clock. I think that is more obvious when providing a Layer once than when wrapping in `Live.live` everywhere clock is used. Do you prefer wrapping as more idiomatic?

runtologist: the test is flaky, because we might actually have measured exactly one millisecond. In that case we can simply retry.

runtologist:
> @runtologist Looks like there is a compilation error on Dotty. I think merging upstream changes should fix.

Yes, I was waiting for #4551

runtologist:
@adamgfraser I think all review comments have been addressed.

adamgfraser (approved):
Looks good! We can update the type signatures and remove the default implementation when we move to 2.0.

Issue #4512, opened 2020-12-19
Higher arity versions for ZIO.mapN, ZIO.tupled, etc.


reibitto:
There are a few of these tuple functions that stop around an arity of 4 or 6. It's time consuming to write all the variations up to the 22-limit by hand, but maybe we could use a generator like [sbt-boilerplate](https://github.com/sbt/sbt-boilerplate)?

adamgfraser:
This is addressed through compositional zips now.

PR #4561, opened 2021-01-16
ZIO.mapN, ZIO.mapParN, ZIO.tupled, and ZIO.tupledPar with arity up to 22

lvitaly:
alphabetical exercise for #4512

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1127,6 +1127,186 @@ object ZIOSpec extends ZIOBaseSpec {
             val expected = f(int, str1, str2, str3)
             assertM(actual)(equalTo(expected))
         }
+      },
+      testM("with Tuple5") {
+        checkM(Gen.anyInt, Gen.alphaNumericString, Gen.alphaNumericString, Gen.alphaNumericString, Gen.alphaNumericString) {
+          (int: Int, str1: String, str2: String, str3: String, str4: String) =>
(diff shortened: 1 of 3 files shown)


jdegoes:
Thank you for your work on this! I think at the very least, we should pull these out into separate traits, in separate files, and "mix them in" to the package objects. The reason is this is a lot of boilerplate and clutter and will make navigating the source code much more difficult.

The other open question is how far to go up to? I think 22 is probably too high. I am not sure what the right number is.

/cc @adamgfraser

adamgfraser:
@lvitaly Thanks for working on this!

Yes, I think maybe if we have files `ZIOArities`, `ZLayerArities`, and so on and mix those into the companion objects for the appropriate data types that is a good way to provide this functionality. We had all the arities for `ZLayer` and while I think people wanted more of them having every possible arity, and all in the main file, I think hurt the ability of people to just read and understand the code.

I'm not sure what the right limit is either and 22 is going to be less important since Scala 3 is supporting tuples of arbitrary size, but maybe if we already have them implemented we just include them as long as they are in a separate file.

I would also put the tests for these methods in a separate file and we may not need them, or be able to use very minimal tests, since the methods are so polymorphic that it is really hard to implement them incorrectly unless you try.

For the `check` methods in particular, my recollection is that for that specific file there was a reason we couldn't mix it in because it led to runtime errors in some situations. I will have to go back and see if I can find the original issue. Would definitely be good to get those out of the package object if we are going to have all of them.

jdegoes:
Since this PR, we have introduced composable zips, which is now the preferred and more scalable path yo handling arities. Thanks for your work on this!

PR #4582, opened 2021-01-22
Support Collection Transformations For Effects Returning Nothing

adamgfraser:
We have had an issue with collection transformation where an implicit `BuildFrom` cannot be found on Scala 2.13 when the return type of the effect is `Nothing`. See scala/bug#12104. This PR resolves that by implementing a new `canBuildFromAny` instance which provides proof that given an original collection `Collection[Any]` and an element type `Element` we can build a `Collection[Element]`.


jdegoes (approved):

joroKr21:
Why `Any` - shouldn't it be `Nothing` instead?

PR #4583, opened 2021-01-22
Add zio-test-refined module

afsalthaj:
The module currently makes use of `DeriveGen`; allowing users to call `DeriveGen[Refined[String, NonEmpty]]` to get a `Gen[Refined[String, NonEmpty]]`


test-refined/shared/src/main/scala/zio/test/refined/CharInstances.scala
```diff
@@ -0,0 +1,47 @@
...
+
+trait CharInstances {
+  private val alphaCharGen: Gen[Random, Char] =
+    Gen.weighted(Gen.char(65, 90) -> 26, Gen.char(97, 122) -> 26)
+
+  private val numericCharGen: Gen[Random, Char] =
+    Gen.weighted(Gen.char(48, 57) -> 10)
+
+  private val whitespaceChars: Seq[Char] =
+    (Char.MinValue to Char.MaxValue).filter(_.isWhitespace)
```
afsalthaj, line 19: Oh. May be move these to `Gen` companion object.

jdegoes: Good idea. 👍

adamgfraser (approved):
This looks great!

In addition to just more instances, the biggest thing I see here is we may want to split out the `Refined` support from the automatic derivation. Right now these are bundles together so it is great if you want to automatically derive generators for refined types, but if you want to manually put more complex ones together from the simpler ones to build in some custom logic it is a little clunky because you have to unpack the `DeriveGen` instances.

I think we could have the same functionality here and make it more flexible if we first defined a set of constructors for `Gen` values for refined types and then used these to construct `DeriveGen` instances for those types.

I will merge this now and open separate issues to follow up on this.

afsalthaj:
Thanks @adamgfraser

PR #4585, opened 2021-01-23
Support Bulk Queue Operations Directly in the RingBuffer

adamgfraser:
Operations on the `RingBuffer` take place in three phases:

1. Determination of whether a space is available based on its `seq` value
2. Reservation of the space by compare and set on the `head` or `tail`
3. Publishing the changes by setting the `seq`

Unfortunately given the concurrent nature of the `RingBuffer` there is no way to determine whether a group of spaces are available other than checking each of their `seq` values. However, we can take advantage of bulk operations by doing compare and swap once to reserve a whole block of spaces after we determine whether they are available. This reduces the number of compare and swap operations from `n` to `1`. The disadvantage is that there is more time between getting the initial state and doing the compare and set since we have to check the `seq` values in between.

As a baseline, here is the performance of the existing `ZQueue` benchmarks on my local machine for 1,000 elements with no chunking on `master`. Note that this is with my latest optimizations to `takeBetween`.

```
[info] Benchmark                           Mode  Cnt     Score    Error  Units
[info] QueueParallelBenchmark.zioQueue    thrpt    5  5043.535 ± 173.277  ops/s
[info] QueueSequentialBenchmark.zioQueue  thrpt    5  9357.960 ± 61.826  ops/s
```

Here is the performance, also on `master` of a new benchmark that is exactly the same except it uses `offerAll` and `takeBetween` with a chunk size of 10.

```
[info] Benchmark                                Mode  Cnt      Score     Error  Units
[info] QueueChunkBenchmark.zioQueueParallel    thrpt    5   6288.556 ± 255.551  ops/s
[info] QueueChunkBenchmark.zioQueueSequential  thrpt    5  20674.860 ±  86.972  ops/s
```

Here is the performance of the same benchmark on this branch.

```
[info] Benchmark                                Mode  Cnt      Score     Error  Units
[info] QueueChunkBenchmark.zioQueueParallel    thrpt   45  13610.646 ± 116.241  ops/s
[info] QueueChunkBenchmark.zioQueueSequential  thrpt   45  28048.018 ± 186.584  ops/s
```

One risk with this strategy is that under very high contention we could be caught repeatedly retrying. It could be interesting to explore tracking how many times we have retried a bulk operation and either reverting to the single element case or doing exponential backoff on the chunk size we attempt to offer.

### DIFF
--- a/benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
@@ -16,8 +16,8 @@ import java.util.concurrent.TimeUnit
  */
 class QueueChunkBenchmark {
 
-  val totalSize   = 100000
-  val chunkSize   = 1000
+  val totalSize   = 1000
+  val chunkSize   = 10
   val parallelism = 5
 
   var zioQ: Queue[Int] = _
--- a/benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
@@ -8,9 +8,9 @@ import java.util.concurrent.TimeUnit
 @State(Scope.Thread)
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
-@Measurement(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 3)
-@Warmup(iterations = 15, timeUnit = TimeUnit.SECONDS, time = 3)
-@Fork(3)
+@Measurement(iterations = 5, timeUnit = TimeUnit.SECONDS, time = 3)
+@Warmup(iterations = 5, timeUnit = TimeUnit.SECONDS, time = 3)
+@Fork(1)
 /**
  * This benchmark offers and takes chunks of items.
  */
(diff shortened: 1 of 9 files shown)


benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Setup(Level.Trial)
+  def createQueues(): Unit =
+    zioQ = unsafeRun(Queue.bounded[Int](totalSize))
+
+  @Benchmark
+  def zioQueueParallel(): Int = {
+
+    val io = for {
+      offers <- IO.forkAll {
```
jdegoes, line 35: `forkAll_` since you don't want to measure the overhead of the join concatenation.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+                  List.fill(parallelism) {
```
jdegoes, line 36: Probably best to fill this list outside of the benchmark.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+                    repeat(totalSize * 1 / chunkSize * 1 / parallelism)(zioQ.offerAll(chunk).unit)
+                  }
+                }
+      takes <- IO.forkAll {
```
jdegoes, line 40: Ditto.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+                 List.fill(parallelism) {
```
jdegoes, line 41: Ditto.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  @Benchmark
+  def zioQueueSequential(): Int = {
+
+    def repeat(task: UIO[Unit], max: Int): UIO[Unit] =
+      if (max < 1) IO.unit
+      else task.flatMap(_ => repeat(task, max - 1))
+
+    val io = for {
+      _ <- repeat(zioQ.offerAll(chunk).unit, totalSize / chunkSize)
```
jdegoes, line 60: Consider `ZIO#repeatN`.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
jdegoes, line 41: Ditto.

core/js/src/main/scala/zio/internal/RingBuffer.scala
```diff
@@ -109,6 +111,80 @@ final class RingBuffer[A](override final val capacity: Int) extends MutableConcu
...
+
+    if (state == STATE_RESERVED) {
+      // We have successfully resserved space in the queue and have exclusive
+      // ownership of each space until we publish our changes. Enqueue the
+      // elements sequentially and publish our changes as we go.
+      val iterator = as.iterator
+      while (enqHead < enqTail) {
+        val a = iterator.next()
+        curIdx = posToIdx(enqHead, capacity)
+        buf(curIdx) = a.asInstanceOf[AnyRef]
```
jdegoes, line 176: This is tricky logic to manually verify. I suggest we make a stress test, which can bombard a ring buffer with concurrent operations (of the newly-added variety) and verify invariants. This can be a pretty robust way to catch race conditions in concurrent code.

jdegoes (changes requested):
Overall, looks great! Love the fact it does not involve changes to existing code.

I had a few suggestions for improving the benchmark. My main suggestion is that I think we need at least one stress test that can verify invariants hold under concurrent load to the newly-added operations. Then will be good to merge!

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
adamgfraser: I don't think we can use `forkAll` here because we need a fiber that we can await to ensure that the work was completed.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
adamgfraser: Done.

benchmarks/src/main/scala/zio/QueueChunkBenchmark.scala
adamgfraser: This skews the benchmark pretty severely because `repeatN` inserts a yield between every repetition.

core/js/src/main/scala/zio/internal/RingBuffer.scala
adamgfraser: Added.

adamgfraser:
@jdegoes I added concurrency stress tests and refined the benchmarks. Results are similar.

Here is current performance. Chunking is marginally faster for parallel operations and significantly faster for sequential operations.

```
[info] Benchmark                               Mode  Cnt      Score     Error  Units
[info] QueueChunkBenchmark.parallel           thrpt    5   5267.461 ± 479.507  ops/s
[info] QueueChunkBenchmark.parallelChunked    thrpt    5   6381.639 ± 184.475  ops/s
[info] QueueChunkBenchmark.sequential         thrpt    5  10983.624 ± 112.716  ops/s
[info] QueueChunkBenchmark.sequentialChunked  thrpt    5  18385.563 ± 104.438  ops/s
```

Here are the same benchmarks on this branch. The benchmarks without chunking are the same because there is no change for that code. With chunking performance dramatically increases in both the parallel and sequential cases.

```
[info] Benchmark                               Mode  Cnt      Score     Error  Units
[info] QueueChunkBenchmark.parallel           thrpt    5   5264.719 ±  58.599  ops/s
[info] QueueChunkBenchmark.parallelChunked    thrpt    5  13599.361 ± 161.081  ops/s
[info] QueueChunkBenchmark.sequential         thrpt    5  10832.904 ± 145.367  ops/s
[info] QueueChunkBenchmark.sequentialChunked  thrpt    5  32397.516 ± 545.071  ops/s
```

jdegoes (approved):
Excellent work! 💪

PR #4586, opened 2021-01-24
Improve Implementation of ZIO.foreachParN

### LINKED ISSUE #4146: ZManaged#foreachParN test flakiness
```
[info] - ZManaged
[info]   - foreachParN
[info]     - Maintains finalizer ordering in inner ZManaged values
[info] Timeout of 1 m exceeded.
```

adamgfraser:
Resolves #4146.


iravid (approved):
Nice. Probably much more efficient too compared to the use of `n` promises.

adamgfraser:
Yes. And the queue was unnecessarily being created with a size equal to the specified parallelism so it was applying back pressure on each value being offered to the queue.

PR #4587, opened 2021-01-24
Parameterized tests

hmemcpy:
This PR adds a method called `testAllM` which accepts a generator and produces multiple tests from that generator. It simulates a concept called Parameterized tests where the same test operates on a different input value.

We've been using this function at work and it's proven to be handy when need to test against a fixed input parameter list, but does not necessarily need to be a property test.

Usage:

```scala
val gen = Gen.fromIterable(List("file1.bin", "file2.bin", "file3.bin"))

testAllM("Can parse contents for file")(gen) { file =>
   ... // do something with file
  assertM(...)(...)
}
```
When ran, this produces three tests:

```
+ Can parse contents for file
  + Can parse contents for file file1.bin
  + Can parse contents for file file2.bin
  + Can parse contents for file file3.bin
```

Would be nice to have this in ZIO.

### DIFF
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -41,14 +41,14 @@ lazy val tests = project
   .settings(
     skip in publish := true,
     libraryDependencies += "ch.epfl.scala" % "scalafix-testkit" % V.scalafixVersion % Test cross CrossVersion.full,
-    compile.in(Compile) := 
+    compile.in(Compile) :=
       compile.in(Compile).dependsOn(compile.in(input, Compile)).value,
     scalafixTestkitOutputSourceDirectories :=
       sourceDirectories.in(output, Compile).value,
     scalafixTestkitInputSourceDirectories :=
       sourceDirectories.in(input, Compile).value,
     scalafixTestkitInputClasspath :=
-      fullClasspath.in(input, Compile).value,
+      fullClasspath.in(input, Compile).value
   )
   .dependsOn(rules)
   .enablePlugins(ScalafixTestkitPlugin)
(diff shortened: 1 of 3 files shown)


hmemcpy:
@adamgfraser WDYT?

adamgfraser:
@hmemcpy Interesting!

The way I have typically implemented that is in terms of `checkM`, so your example above would look something like:

```scala
val gen = Gen.fromIterable(List("file1.bin", "file2.bin", "file3.bin"))

testM("Can parse contents for files") {
  checkM(gen) { file =>
     ... // do something with file
    assertM(...)(...)
  }
}
```

That has the advantage of being a little more orthogonal I think and using existing operators and does basically the same thing. Your version is a little shorter and has slightly different semantics in that it checks all the files even if one fails and renders all the successes versus this would just render the name of the one that fails.

Another approach to this that might be a little more general would be to implement `suiteM`.

```scala
def suiteM[R, E, T](label: String)(specs: ZIO[R, E, Iterable[Spec[R, E, T]]]): Spec[R, E, T] =
  Spec.suite(label, specs.toManaged_, None)
```

Then your example could just be something like:

```scala
val files: List[String] = List("file1.bin", "file2.bin", "file3.bin")

suiteM("Can parse contents for file") {
  ZIO.foreach(files) { file =>
    testM(file) {
      assertM(...)(...)
    }
  }
}
```

That isn't quite as terse as your original example but is a lot more general in that you can perform whatever effects you want and can also potentially construct a suite that has a nested structure instead of just a flat one.

jdegoes:
I like the idea of `suiteM` 👍

hmemcpy:
This looks great! Let me rework my PR and I'll add this! Thanks!

jdegoes:
@hmemcpy Looks like @adamgfraser beat you to it. But please check out the new `suiteM` and let us know if it works for you.

Thanks for bringing this issue up!

PR #4590, opened 2021-01-24
Have Scala Steward's PRs autoupdated

sideeffffect:
If we don't have this, PRs like https://github.com/zio/zio/pull/4548 won't automatically retried and won't get merged.
It's a shame, that GitHub's PR mechanism needs a workaround like this, but such is life :crying_cat_face:

### DIFF
--- a/.github/workflows/autoupdate.yaml
+++ b/.github/workflows/autoupdate.yaml
@@ -1,0 +1,14 @@
+name: autoupdate
+on:
+  push: {}
+jobs:
+  autoupdate:
+    name: autoupdate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          PR_FILTER: "labelled"
+          PR_LABELS: "dependency-update,autoupdate"
+          EXCLUDED_LABELS: "do-not-autoupdate"


adamgfraser:
@sideeffffect What are the semantics of autoupdate? They get updated when there is a new dependency or they get autoupdated whenever there is a new commit to the main branch?

sideeffffect:
> when there is a new dependency

This scenario is already handled by Scala Steward. If Steward opens a PR updating a dependency to `1.0.0` and before the PR is merged a version `1.0.1` is released, Steward will take care of this. Either update the current PR open a new one and close the old (I don't remember which).

> whenever there is a new commit to the main branch

This is the scenario we need Autoupdate for. It merges the target (`master` in our case) to the PR if a new commit lands to the target branch.
More info: https://github.com/marketplace/actions/auto-update

adamgfraser:
Isn't that going to result in potentially a lot more CI usage? I see some of these updates where they fail because some other dependency is not available. If we end up merging a bunch of unrelated PRs for new features or bug fixes before we resolve whatever was causing the issue for that dependency update it is going to keep rerunning and failing.

sideeffffect:
Yes, you're right that would be happening. But I'm hopeful that it won't be as much CI usage and that it will result in less grunt work for us.

There are not as many outstanding Scala Steward's PRs
https://github.com/zio/zio/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Adependency-update
And some of them should be fixed by just closing them :wink:

jdegoes:
I am not sure what to do with this. I'm not really a fan of Scala Steward in ZIO to begin with (so much noise for very little practical gain). Will close for now but let's have a discussion here and we can re-open if necessary.

PR #4598, opened 2021-01-25
make ZStream#peel more specific

### LINKED ISSUE #4597: ZStream#peel should return more specific ZStream
`zio.stream.ZStream#peel` returns a stream with the same environment `R1` and error `E1` as the returned ZManaged. This is overly wide and could be more specific , namely the original `R` and `E`.

SrTobi:
fixes #4597

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2197,7 +2197,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * stream is valid only within the scope of [[ZManaged]].
    */
   def peel[R1 <: R, E1 >: E, O1 >: O, Z](
-    sink: ZSink[R1, E1, O1, O1, Z]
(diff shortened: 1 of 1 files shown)


adamgfraser (approved):

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2198,7 +2198,7 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    */
   def peel[R1 <: R, E1 >: E, O1 >: O, Z](
     sink: ZSink[R1, E1, O1, O1, Z]
```
adamgfraser, line 2200:
```suggestion
    sink: ZSink[R1, E1, O, O1, Z]
```

adamgfraser: The input type to the `ZSink` can also be `O` instead of `O1`.

adamgfraser (approved):

adamgfraser:
@SrTobi Thank you! 🙏

PR #4603, opened 2021-01-26
Fixed Clock.currentTime implementation

darl:
`Instant.getEpochMilli` and `Instant.getNano` both contains millisecond part.
so `inst.toEpochMilli() * 1000000 + inst.getNano()` resulted to milliseconds being counted twice.

First commit has test which fails with old implementation.


adamgfraser (approved):
Thanks for the fix!

PR #4605, opened 2021-01-26
ZIO Test: Implement SuiteM

adamgfraser:
Analogous to `testM`, `suiteM` allows building effectual suites. For example, you could open a collection of files and create a test based on the contents of each file.


jdegoes (changes requested):
Looks great! I'd also add this to the spec so no imports are required to use it.

adamgfraser:
@jdegoes Good idea!

jdegoes (approved):

Issue #4388, opened 2020-11-09
Add ZStream.execute


jdegoes:
Something like this:

```scala
object ZStream {
  def execute[R, E, A](zio: ZIO[R, E, A]): ZStream[R, E, Nothing] = ZStream.fromEffect(zio).drain
}
```

meckesl:
Hi, I'd like to take a first issue, is it available ?

adamgfraser:
@meckesl Go for it!

meckesl:
I am struggling to run sbt build and tests with decent performance. Any recommendation for the dev environment - like Java version, memory settings, and the computer specs ? Regards,

adamgfraser:
@meckesl Try navigating to a specific project, like within `sbt` to `project streamsJVM` or `project streamsTestsJVM`.

meckesl:
@adamgfraser thanks, that helps !

DomWos:
@meckesl are You working on that ?

jeroenr:
I think this one can be closed now since it's added by https://github.com/zio/zio/pull/6091

PR #4608, opened 2021-01-27
WIP Add Zstream.execute #4388

meckesl:
Hi. I added the function in ZStream,
I am still confused about how to test it, could you help me ?


adamgfraser (changes requested):
@meckesl You can run `prepare` within SBT to format the code so that it passes CI. I think you are on the right track with the test here. I might test combining the stream from `Execute` with another stream to make sure you can still get the elements back. So something like:

```scala
ZStream.execute(ref.update(_ + 2)) ++ ZStream(1, 2, 3) ++ ZStream.execute(ref.update(_ * 3))
```

Then you could run the stream and test that you got the right elements back and that the `Ref` has the correct value.

PR #4609, opened 2021-01-27
Add Demyst to adopter list

fsvehla:
- Adds Demyst to adopter list
- Also sorts (`sort -f`) the adopter list


iravid (approved):
Thanks @fsvehla :-)

PR #4591, opened 2021-01-24
Scala Native 0.4.0

sideeffffect:
closes https://github.com/zio/zio/pull/4567
closes https://github.com/zio/zio/pull/4562
closes https://github.com/zio/zio/pull/4604
closes https://github.com/zio/zio/pull/4548

Blocked on
https://github.com/zio/izumi-reflect/pull/121

PR #4621, opened 2021-01-28
Update izumi-reflect to 1.0.0-M13

sideeffffect:


sideeffffect:
`izumi-reflect` 1.0.0-M13 published only for Scala Native 0.4.0, so has to be done as part of https://github.com/zio/zio/pull/4591

PR #4622, opened 2021-01-28
Change Default Scala Version To 2.13

adamgfraser:
There is a regression in the latest version of Scala 2.12 that causes compiler crashes when there is a compilation issue instead of providing a helpful warning. In addition, with Scala 3 around the corner I think it makes sense that we are on Scala 2.13 by default. We will still include earlier versions in the CI matrix to make sure we support them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -287,14 +287,14 @@ jobs:
     <<: *lint
     <<: *machine_ubuntu
     environment:
-      - <<: *scala_212
+      - <<: *scala_213
       - <<: *jdk_8
 
   mdoc:
     <<: *mdoc
     <<: *machine_ubuntu
     environment:
-      - <<: *scala_212
+      - <<: *scala_213
       - <<: *jdk_8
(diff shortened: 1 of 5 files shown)


sideeffffect (changes requested):
Since we're changing the default Scala version for "local development", we might as well change some CircleCI tasks to use 2.13 too, namely:
 * `lint`
 * `mdoc`
 * change `test_212_jdk11_jvm` to `test_213_jdk11_jvm`

mijicd (approved):

sideeffffect (approved):
Great! :+1:

sideeffffect:
Although I like the code, Scalafix does not. But nothing a little `fix` couldn't fix :wink:

sideeffffect (approved):

adamgfraser:
@mijicd Can I get a review on this?

mijicd:
@adamgfraser looks good to me, but press the button :)

PR #4623, opened 2021-01-28
Implement TPriorityQueue#isEmpty

### LINKED ISSUE #4618: TPriorityQueue could introduce a empty/isEmpty predicate

adamgfraser:
Resolves #4618.

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
+++ b/core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
@@ -37,6 +37,12 @@ final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, ::[A]]
     ref.get.map(_.isEmpty)
 
   /**
+   * Checks whether the queue is not empty..
+   */
+  def nonEmpty: USTM[Boolean] =
+    ref.get.map(_.nonEmpty)
+
+  /**
    * Offers the specified value to the queue.
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
```diff
@@ -30,6 +30,12 @@ import scala.collection.immutable.SortedMap
  */
 final class TPriorityQueue[A] private (private val ref: TRef[SortedMap[A, ::[A]]]) extends AnyVal {
 
+  /**
+   * Checks whether the queue is empty.
+   */
+  def isEmpty: USTM[Boolean] =
```
phderome, line 36: do you normally implement `nonEmpty` in similar collections?

mijicd: We don't yet, but we could provide them separately.

mijicd (approved):

core/shared/src/main/scala/zio/stm/TPriorityQueue.scala
adamgfraser: We haven't on a lot of these historically but I don't think there is any reason not to. I just added it to this PR for `TPriorityQueue`.

mijicd (approved):

Issue #4593, opened 2021-01-25
Ensure All Examples From Documentation Are Compiled


adamgfraser:
We should ensure that all code snippets in the documentation are checked with `mdoc` to ensure that they compile as written.

Can I also ask that for all examples we NOT use hidden or invisible code? I know sometimes people do this because they want to omit things they see as boilerplate like imports or type definitions. However, a lot of people are going to copy and paste and compile these examples as written so if there are things missing it makes it harder for them to get started.

etandon-coatue:
Picking up this issue

emilhotkowski:
I've checked and all snippets work correctly. 

I could pick up this task and remove all invisible / hidden code so it is easier to copy/paste working snippets. Should I do this? @adamgfraser

PR #4630, opened 2021-01-30
GH-4593: Ensure all docs compiled, remove invisible tags

etandon:
Original issue:
```
We should ensure that all code snippets in the documentation are checked with mdoc to ensure that they compile as written.

Can I also ask that for all examples we NOT use hidden or invisible code? I know sometimes people do this because they want to omit things they see as boilerplate like imports or type definitions. However, a lot of people are going to copy and paste and compile these examples as written so if there are things missing it makes it harder for them to get started.
```


etandon:
Closes: #4593

PR #4631, opened 2021-01-30
Added ZSink.toQueue and ZSink.toQueueWithShutdown (#3735)

### LINKED ISSUE #3735: Add ZSink.toQueue

arnoldlacko:
Closes #3735

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZSinkSpec.scala
@@ -3,7 +3,7 @@ package zio.stream
 import zio.duration._
 import zio.stream.SinkUtils.{findSink, sinkRaceLaw}
 import zio.stream.ZStreamGen._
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):
Looks great!

arnoldlacko:
Thank you @adamgfraser  🙂

adamgfraser (approved):

iravid:
@adamgfraser @arnoldlacko Sorry to nitpick - I think this should be named `fromQueue`, as the queue is used to create a sink. `toQueue` would be used for converting an existing sink to a queue.

What do you think?

adamgfraser:
@iravid Makes sense to me.

PR #4632, opened 2021-01-30
Added ZQueue.takeN

arnoldlacko:
Added `ZQueue.takeN` relying on `takeBetween(n, n)`


adamgfraser (approved):
This looks good! For 2.0 we should change all these methods to return `Chunk` but for now makes sense to maintain consistency with `takeBetween`.

PR #4635, opened 2021-01-31
Check That Both Queues Are Shutdown In Implementation of isShutdown in Queue#bothWithM

adamgfraser:
I believe we should be checking if both queues have been shutdown here.

### DIFF
--- a/core/shared/src/main/scala/zio/ZQueue.scala
+++ b/core/shared/src/main/scala/zio/ZQueue.scala
@@ -186,7 +186,7 @@ sealed abstract class ZQueue[-RA, -RB, +EA, +EB, -A, +B] extends Serializable { 
       def awaitShutdown: UIO[Unit] = self.awaitShutdown *> that.awaitShutdown
       def size: UIO[Int]           = self.size.zipWithPar(that.size)(math.max)
       def shutdown: UIO[Unit]      = self.shutdown.zipWithPar(that.shutdown)((_, _) => ())
-      def isShutdown: UIO[Boolean] = self.isShutdown.zipWithPar(that.isShutdown)(_ && _)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZQueue.scala
```diff
@@ -186,7 +186,7 @@ sealed abstract class ZQueue[-RA, -RB, +EA, +EB, -A, +B] extends Serializable {
       def awaitShutdown: UIO[Unit] = self.awaitShutdown *> that.awaitShutdown
       def size: UIO[Int]           = self.size.zipWithPar(that.size)(math.max)
       def shutdown: UIO[Unit]      = self.shutdown.zipWithPar(that.shutdown)((_, _) => ())
-      def isShutdown: UIO[Boolean] = self.isShutdown
+      def isShutdown: UIO[Boolean] = self.isShutdown.zipWithPar(that.isShutdown)(_ && _)
```
adamgfraser, line 189: Actually should this be `&&` or `||`? You're going to get interrupted if you try to pull from a composite queue if either queue has been shutdown, so in that sense `||` seems like the wrong semantics. But on the other hand shutdown of the composite queue hasn't been completed until both queues have been shut down.

adamgfraser:
@jdegoes Do you have a sense of what the right semantics are here or should we just close this since these operators are deprecated anyway?

jdegoes:
@adamgfraser `isShutdown` should reflect the following: if a putter or taker will be interrupted when attempting to put or take, then the queue should be regarded as shut down. Otherwise, not.

adamgfraser:
Okay, then I think reporting that the combined queue is shutdown if either of the original queues is shut down is the correct semantic. Updated to reflect that now.

regiskuckaertz (approved):

PR #4639, opened 2021-02-01
Implement ZQueue#filterOutput

adamgfraser:
Implements `filterOutput` and `filterOutputM` operators on `ZQueue` analogous to the existing `filterInput` and `filterInputM` operators. Particularly useful if the queue represents values that are being broadcast from an upstream producer to allow a consumer to receive only a subset of the values.


adamgfraser:
@jdegoes Could I get a review on this?

iravid (approved):

Issue #4643, opened 2021-02-01
Trying to run tests with ZIO 1.0.4.-1 fails with NoClassDefFoundError


sideeffffect:
When I try running tests on a project that uses ZIO 1.0.4.-1, it fails with
```
java.lang.NoClassDefFoundError: scala/reflect/api/TypeCreator
	at zio.test.sbt.BaseTestTask.specInstance$lzycompute(BaseTestTask.scala:18)
	at zio.test.sbt.BaseTestTask.specInstance(BaseTestTask.scala:15)
	at zio.test.sbt.BaseTestTask.execute(BaseTestTask.scala:42)
	at sbt.TestRunner.runTest$1(TestFramework.scala:139)
	at sbt.TestRunner.run(TestFramework.scala:154)
	at sbt.TestFramework$$anon$3$$anonfun$$lessinit$greater$1.$anonfun$apply$1(TestFramework.scala:317)
	at sbt.TestFramework$.sbt$TestFramework$$withContextLoader(TestFramework.scala:277)
	at sbt.TestFramework$$anon$3$$anonfun$$lessinit$greater$1.apply(TestFramework.scala:317)
	at sbt.TestFramework$$anon$3$$anonfun$$lessinit$greater$1.apply(TestFramework.scala:317)
	at sbt.TestFunction.apply(TestFramework.scala:329)
	at sbt.Tests$.$anonfun$toTask$1(Tests.scala:422)
	at sbt.std.Transform$$anon$3.$anonfun$apply$2(Transform.scala:46)
	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
	at sbt.Execute.work(Execute.scala:291)
	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #4619, opened 2021-01-28
Update portable-scala-reflect to 1.1.0

sideeffffect:
closes https://github.com/zio/zio/pull/4604

---
EDIT: fixed

Unfortunately this fails with
```
[error] /home/ondra/Projects/zio/core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala:8:22: object io is not a member of package reflect
[error] import scala.reflect.io.File
[error]                      ^
[error] /home/ondra/Projects/zio/core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala:15:65: not found: value File
[error]         assertM(live(system.env("PATH")))(isSome(containsString(File.separator + "bin")))
[error]                                                                 ^
[error] /home/ondra/Projects/zio/core-tests/jvm/src/test/scala/zio/system/SystemSpec.scala:23:77: not found: value File
[error]         assertM(live(system.envs.map(_.get("PATH"))))(isSome(containsString(File.separator + "bin")))
[error]                                                                             ^
...
[error] (coreTestsJVM / Test / compileIncremental) Compilation failed
```

Any idea what could be going wrong?

PR #4645, opened 2021-02-01
Use scala-reflect as a dependency of JVM test-sbt

sideeffffect:
Only JVM test-sbt needs scala-reflect, but it needs to be a normal dependency -- `Test` nor `Provided` work
(tested by `publishLocal` and using such artifacts on a project)

fixes https://github.com/zio/zio/issues/4643

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -166,7 +166,6 @@ lazy val coreTestsJVM = coreTests.jvm
   .settings(dottySettings)
   .configure(_.enablePlugins(JCStressPlugin))
   .settings(replSettings)
-  .settings(scalaReflectTestSettings)
 
 lazy val coreTestsJS = coreTests.js
   .settings(jsSettings)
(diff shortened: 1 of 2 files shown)


adamgfraser:
Is there a way that we can automatically verify that this fixes the breakage and that similar issues will not occur? The fact that the previous version of the build passed all checks and still failed when we released it is not great.

sideeffffect:
Yeah, it's bummer that such a thing could have slipped through. The only 100% and fool proof solution would be to perform a `publishLocal` and then used those artifacts on some standardized, intentionally unchanging project, as the last step of our CI.

adamgfraser:
Is there any module or configuration we could have compiled short of publishing locally that would ha e revealed it?

sideeffffect:
You could say that `coreTestsJVM` is enough to spot the missing `scala-reflect`. But such a "test" was neutered by adding ad-hoc `scala-reflect` (and only as a `Test`) dependency: https://github.com/zio/zio/pull/4619/files
All of this is my fault -- I hadn't realized that's a wrong thing to do :disappointed:

adamgfraser (approved):

adamgfraser:
No worries. Will do another release later today.

PR #4651, opened 2021-02-02
ZIO 2.0: Delete Deprecated Operators

adamgfraser:


jdegoes (approved):

jdegoes:
@adamgfraser Can you disable MIMA on this branch?

adamgfraser:
@jdegoes I disabled it in the branch protection rules but I think we need update the CI to not run it at all for PRs submitted against this branch or the failure will still be reported.

@mijicd or @softinio Would you be able to help with that?

iravid:
A better approach would be to keep the branch protection rules enabled, but set `failOnProblem` to false in the call to `mimaSettings` in `build.sbt`.

adamgfraser:
@iravid Good idea. I will make this change now.

PR #4653, opened 2021-02-03
Improve Scala Native integration

lolgab:
- Copy-paste JS version to align latest improvements
- Use scala-native-loop Timer instead of Scala.js compatibility layer


adamgfraser (approved):
Thank you! 🙏

PR #4660, opened 2021-02-06
Retrieve Current Executor in ZIO#runtime and ZIO#executor

adamgfraser:
Currently the `runtime` and `executor` operators on `ZIO` retrieve ZIO's overall runtime system and the executor associated with it. This is somewhat non-compositional because it means that if the user accesses the `Runtime` or the `Executor` in an inner scope they can end up running effects on the default thread pool instead of the thread pool that the region is locked on. This PR changes that behavior to access the current `Executor` for the fiber and use that.


jdegoes (approved):
Great catch!

PR #4662, opened 2021-02-06
Implement ZManaged.lock

### LINKED ISSUE #4661: Create ZManaged.lock(...) for locking on custom Executor.
It would be nice to have `lock` method on `ZManaged` that can do the lock on `Executor` for effects enclosed in `ZManaged`.

adamgfraser:
Resolves #4661.

To support this we need to change the algebra of `ZIO` slightly. `lock` describes a static scope of shifting onto and off of a thread pool around a given effect. However, that doesn't give us the power to describe shifting for dynamic scopes. To support this we remove `lock` as a primitive and replace it with `shift`, which shifts to a specified `Executor`. `lock` can then be implemented using a combination of `shift` and `bracket`. With this change, `lock` just becomes a `ZManaged` effect that describes shifting onto a thread pool as its `acquire` effect and shifting back as its `release` effect.


jdegoes (approved):
Looks really nice!

PR #4669, opened 2021-02-09
ZIO 2.0: Clean up Clock implementation

runtologist:
Follow up of #4554


adamgfraser (changes requested):
Just need to remove an `orDie` call in the documentation and then ready to go!

adamgfraser (approved):

PR #4672, opened 2021-02-10
Update scheduling for 70% change to success

majidhosseini:


docs/usecases/scheduling.md
```diff
@@ -15,7 +15,7 @@ import java.util.Random
 
 object API {
   def makeRequest = Task.effect {
-    if (new Random().nextInt(10) > 7) "some value" else throw new Exception("hi")
+    if (new Random().nextInt(10) <= 7) "some value" else throw new Exception("hi")
```
adamgfraser, line 18:
```suggestion
    if (new Random().nextInt(10) < 7) "some value" else throw new Exception("hi")
```

adamgfraser: I think technically we want this to be `<`. `nextInt` returns integers up to the specified number, exclusive. So `nextInt(10)` generates integers between `0` and `9`. So if we want to have a 70% change of success then we should fail on values equal to `7`, `8`, or `9`. While we're here I would also suggest replacing this with `scala.util.Random` since that is more idiomatic Scala and there is no need need to work directly with the Java interface here.

adamgfraser (changes requested):
Thanks for the fix! See comments below.

adamgfraser:
@majidhosseini Can you sign the CLA?

majidhosseini:
> @majidhosseini Can you sign the CLA?

Yes, done.

adamgfraser (approved):

adamgfraser (approved):

Issue #3563, opened 2020-05-09
Mocking framework doesn't provide the means to test the absence of method calls


nartamonov:
I want to test that `ServiceB` does not call some method of `ServiceA` in specific circumstances. In order to do that I made a mock implementation for `ServiceA` and wrote following test:

```scala
object SampleSpec extends DefaultRunnableSpec {

  val spec = testM("service B does not calls service A if it's not required") {
    val mock = MockServiceA.MethodA.returns(unit).atMost(0)

    assertM(ServiceB.methodB(mustCallA = false))(anything)
      .provideLayer(mock >>> ServiceB.live)
  }

  trait ServiceA {
    def methodA: UIO[Unit]
  }

  trait ServiceB {
    def methodB(mustCallA: Boolean): UIO[Unit]
  }

  object ServiceB {
    def methodB(mustCallA: Boolean): URIO[Has[ServiceB],Unit] =
      ZIO.accessM(_.get.methodB(mustCallA))

    def live: URLayer[Has[ServiceA],Has[ServiceB]] =
      ZLayer.fromService(new ServiceBImpl(_))
  }

  class ServiceBImpl(serviceA: ServiceA) extends ServiceB {
    def methodB(mustCallA: Boolean): UIO[Unit] =
      if (mustCallA)
        serviceA.methodA
      else UIO.unit
  }

  object MockServiceA {
    sealed trait Tag[I,A] extends Method[Has[ServiceA],I,A] {
      def envBuilder: URLayer[Has[mock.Proxy], Has[ServiceA]] = EnvBuilder
    }

    object MethodA extends Tag[Unit,Unit]

    private val EnvBuilder: URLayer[Has[mock.Proxy], Has[ServiceA]] =
      ZLayer.fromService { invoke =>
        new ServiceA {
          def methodA: UIO[Unit] =
            invoke(MethodA)
        }
      }
  }
}
```

zio-test failed to execute test:

```
Fiber failed.
An unchecked error was produced.
java.util.NoSuchElementException: head on empty Range
        at scala.collection.immutable.Range$.scala$collection$immutable$Range$$emptyRangeError(Range.scala:631)
        at scala.collection.immutable.Range.head(Range.scala:126)
        at scala.collection.immutable.Range.min(Range.scala:413)
        at zio.test.FailureRenderer$.loop$4(DefaultTestReporter.scala:449)
        at zio.test.FailureRenderer$.renderUnsatisfiedExpectations(DefaultTestReporter.scala:459)
[...]

PR #4673, opened 2021-02-10
ZIO Mock: Empty mock services asserting no calls are made to them

### LINKED ISSUE #3115: ZIO Mock: expect no calls

ioleo:
Rebased an old draft to current master and added missing documentation and tests.

Closes #3115 #3563 #3641 #4371 

/cc @adamgfraser

### DIFF
--- a/docs/howto/mock_services.md
+++ b/docs/howto/mock_services.md
@@ -359,6 +359,25 @@ object AccountObserverSpec extends DefaultRunnableSpec {
 }
 ```
 
+## Mocking unused collaborators
+
+Often the dependency on a collaborator is only in some branches of the code. To test the correct behaviour of branches without depedencies, we still have to provide it to the environment, but we would like to assert it was never called. With the `Mock.empty` method you can obtain a `ZLayer` with an empty service (no calls expected).
+
+```scala mdoc:silent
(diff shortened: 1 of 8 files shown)


adamgfraser (approved):
Fantastic!

PR #4697, opened 2021-02-15
ZIO Stream: Implement ZStream#changes

adamgfraser:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -554,6 +554,36 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   ): ZStream[R1, E1, O1] =
     catchAllCause(pf.applyOrElse[Cause[E], ZStream[R1, E1, O1]](_, ZStream.halt(_)))
 
+  /**
+   * Returns a new stream that only emits elements that are not equal to the
+   * previous element emitted, using natural equality to determine whether two
+   * elements are equal.
+   */
+  def changes: ZStream[R, E, O] =
```
jdegoes, line 562: Would `dedup` be a better name? (Not sure about that.)

adamgfraser: Yes I wondered about that too. This is `distinctUntilChanged` in reactive streams and `removeDuplicates` in some other frameworks so `deduplicate` is definitely reasonable. The slightly tricky thing is it doesn't necessary remove all duplicates in the sense that if the stream emits `A, A, B, A` the resulting stream will be `A, B, A` so you will still have two `A` values just not too consecutive ones.

jdegoes: I'm ok with `changes` especially if there is no clear pattern.

jdegoes (approved):

PR #4703, opened 2021-02-17
Allow chaining multiple aspects to MutuableRunnableSpec suite

hmemcpy:
A small fix, allowing chained aspects on the `suite` of MutableRunnableSpec, otherwise it's a compilation error.

Thanks to @somdoron for the fix!


adamgfraser (approved):
Thank you! 🙏

PR #4709, opened 2021-02-18
More tests for Chunk.zipWithIndex

vigoo:
The issue I wanted to fix turned out to be already fixed by https://github.com/zio/zio/pull/4675 but as I already wrote some extra tests, I thought we can keep them.


adamgfraser (approved):
This is definitely an improvement over the test coverage before!

One thing we used to have that I actually think was useful was a generator that specifically created chunks built different ways (e.g. from concatenating, slicing, appending, etc...). That isn't the way the default generator in ZIO Test works because external users shouldn't care about the internal representation of `Chunk`, but it is very helpful for our own tests.

It looks like it got deleted in [this](https://github.com/zio/zio/pull/3332/files) PR so may be useful to bring it back and use it more often as well as make sure it reflects the latest `Chunk` constructors (I think we added append and prepend after this).

Additional discussion [here](https://github.com/zio/zio/pull/3321).

adamgfraser (approved):

PR #4710, opened 2021-02-18
Implement ZIO#cachedWith

adamgfraser:
We've gotten several requests recently for a variant of `cached` that allows the caller to invalidate the cache before the `timeToLive` has expired.  This is potentially getting a little specialized but given user demand I think it makes sense to add and we can implement the current `cached` operator in terms of this one so it is not adding much new code.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -148,7 +148,7 @@ object ZIOSpec extends ZIOBaseSpec {
           d     <- cache
         } yield assert(a)(equalTo(b)) && assert(b)(not(equalTo(c))) && assert(c)(equalTo(d))
       },
-      testM("correctly handled an infinite duration time to live") {
+      testM("correctly handles an infinite duration time to live") {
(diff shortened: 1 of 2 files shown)


remiguittaut:
That's really great and a much awaited feature! Just a comment though, I'm not sure about the naming, it doesn't tell me if I don't know that this combinator gives me the possibility to invalidate. `cachedInvalidate` maybe?

adamgfraser:
@remiguittaut Yes, I wasn't sure about the name. Changed it based on your suggestion. Open to other ideas as well.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -159,6 +159,28 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield assert((a, b, c))(equalTo((0, 0, 0)))
       }
     ),
+    suite("cachedWith")(
```
regiskuckaertz, line 162:
```suggestion
    suite("cachedInvalidate")(
```

jdegoes (approved):

Issue #4715, opened 2021-02-18
ZIO Mock: Mockable macro doesn't retain mock object body


ollyw:
When manually creating mocks it can be useful to put helper methods in the mock object which combine multiple expectations. When using the @mockable macro, the body is not retained. For example when compiling the following snippet, the body is lost, so someFooHelper can never be used:
```
          @mockable[Example.Service]
          object ExampleMock {
            val someFooHelper: Expectation[...] = ???
          }
         ExampleMock.someFooHelper // compiler error
```
It may be some explicit decision made to not include the body when using mockable, in which case fine. However, when using `@accessible`, the body is retained, so there is a certain amount of inconsistency. If it is not desirable to keep the body, perhaps the macro should abort to save other devs from some head scratching.

I'll put in an MR in case you think retaining the body is a good idea

ollyw:
This can be closed because #4716 is merged

adamgfraser:
@ollyw Thank you! 🙏

PR #4716, opened 2021-02-18
Retain object body when using @mockable (#4715)

ollyw:
Implementation of #4715


ollyw:
@jdegoes  The build on this MR was also stalled for a week for so, but is now successful

adamgfraser (approved):
Thank you! 🙏

Issue #4717, opened 2021-02-18
ZIO Mock: Services taking type parameters can't use the type in methods


ollyw:
When the service being mocked takes a type parameter and uses it in a method, the macro fails
```
object Example {
  trait Service[R] {
    def get: RIO[R, Unit]
  }
}

@mockable[Example.Service[Int]]
object ExampleMock
```
This would be pretty useful behaviour. I have some code that works for this and will try and prepare an MR if this behaviour is desired by others in ZIO

PR #4718, opened 2021-02-18
Add support for polymorphic service types to mockable macro (#4717)

ollyw:
Implementation for #4718 
This implementation seems to work, but am keen for advice on how to improve it. Perhaps there are some corner cases missed?

### DIFF
--- a/test-tests/shared/src/main/scala/zio/test/mock/modules.scala
+++ b/test-tests/shared/src/main/scala/zio/test/mock/modules.scala
@@ -124,7 +124,12 @@ object modules {
   type PolyModulePureDefsModule[R, E, A] = Has[PolyModulePureDefsModule.Service[R, E, A]]
   object PolyModulePureDefsModule {
     trait Service[R, E, A] {
+      val static: ZIO[R, E, A]
+      def zeroParams: ZIO[R, E, A]
       def zeroParamsWithParens(): ZIO[R, E, A]
+      def singleParam(a: Int): ZIO[R, E, A]
+      def manyParams(a: Int, b: String, c: Long): ZIO[R, E, A]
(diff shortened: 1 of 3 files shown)


jdegoes:
@ollyw Thanks for your work on this! Looks like the build has stalled. Can you force a rebuild by committing something benign?

ollyw:
@jdegoes built has past now. I think it had to do with the fact that CircleCI didn't have permissions on my github account

adamgfraser (approved):
Thank you!

aalbul:
Hi all. I know that it's already resolved, but I have some issues with the current implementation. 
It doesn't allow to use of generic types in method parameters definitions.
For example, i tweaked the test case from ZIO codebase a bit to demonstrate it:

```scala
  trait Service[R, E, A] {
    val static: ZIO[R, E, A]
    def zeroParams: ZIO[R, E, A]
    def zeroParamsWithParens(): ZIO[R, E, A]
    def singleParam(a: Int): ZIO[R, E, A]
    def manyParams(a: Int, b: String, c: Long): ZIO[R, E, A]
    def manyParamLists(a: Int)(b: R)(c: Long): ZIO[R, E, A]
  }
```

As you see, `manyParamLists` accepts `b` as `R` now and this causes the following problem during compilation:

```scala
not found: type R
  case object ManyParamLists extends Effect[(Int, R, Long), Throwable, Double] 
```

Is it a known limitation or simply was not taken into account?

ollyw:
@aalbul the mockable macro has lots of more advanced cases that aren't currently supported. I have experimental version of the macro that I haven't upstreamed that supports a wider range of use cases. If I get some time this week I'll see if it supports your use case, and what effort it will take to upstream

aalbul:
Cool! It would be great!

Issue #4600, opened 2021-01-25
ZIO Test: Upstream Generators From ZIO Test Refined Module


adamgfraser:
We just merged support for generating values of types from the `refined` library through the `zio-test-refined` module thanks to work from @afsalthaj and @pierangeloc .

In [this](https://github.com/zio/zio/blob/2769933b193e47a21a205a9fcecd626876333f02/test-refined/shared/src/main/scala/zio/test/refined/CharInstances.scala#L12) file there are a couple of generators for alphabetical, numeric, and whitespace characters that could be more generally applicable. We should move these instances to the `Gen` companion object in ZIO Test and just use them here. A completed ticket will also add tests verifying that the generators generate values in the appropriate range and shrink to the expected values.

maxim092001:
I think I can handle this one. But this is my very first experience with open source contribution, is there any guideline for newbies?

987Nabil:
@adamgfraser can this be closed?

adamgfraser:
@987Nabil Yes!

PR #4723, opened 2021-02-22
Upstream Generators From ZIO Test Refined Module #4600

maxim092001:
Implementation for #4600. Same to #4721, but commits from personal github.

Moved three generators from [CharInstances](https://github.com/zio/zio/blob/2769933b193e47a21a205a9fcecd626876333f02/test-refined/shared/src/main/scala/zio/test/refined/CharInstances.scala#L11) to [Gen](https://github.com/zio/zio/blob/2769933b193e47a21a205a9fcecd626876333f02/test/shared/src/main/scala/zio/test/Gen.scala#L167)

However, I'm not sure about javadoc description and shrinks. Please, check them out.

### DIFF
--- a/test/shared/src/main/scala/zio/test/Gen.scala
+++ b/test/shared/src/main/scala/zio/test/Gen.scala
@@ -182,6 +182,24 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
     weighted(char(48, 57) -> 10, char(65, 90) -> 26, char(97, 122) -> 26)
 
   /**
+   * A generator of alpha characters.
+   */
+  val alphaChar: Gen[Random, Char] =
+    weighted(char(65, 90) -> 26, char(97, 122) -> 26)
+
+  /**
+   * A generator of numeric characters. Shrinks toward '0'.
+   */
+  val numericChar: Gen[Random, Char] =
+    weighted(char(48, 57) -> 10)
+
+  /**
+   * A generator of whitespace characters.
+   */
+  val whitespaceChars: Seq[Char] =
+    (Char.MinValue to Char.MaxValue).filter(_.isWhitespace)
+
+  /**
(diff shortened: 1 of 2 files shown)


test/shared/src/main/scala/zio/test/Gen.scala
```diff
@@ -181,6 +181,24 @@ object Gen extends GenZIO with FunctionVariants with TimeVariants {
   val alphaNumericChar: Gen[Random, Char] =
     weighted(char(48, 57) -> 10, char(65, 90) -> 26, char(97, 122) -> 26)
 
+  /**
+   * A generator of alpha characters.
+   */
+  val alphaChar: Gen[Random, Char] =
```
adamgfraser, line 187: Can we alphabetize please? Otherwise looks great!

adamgfraser (changes requested):
Just one small change and then good to go.

test/shared/src/main/scala/zio/test/Gen.scala
maxim092001: Sorry, didn't get what you mean.
Alphabetize random char generator?

adamgfraser: Oh sorry. Just put the constructors in alphabetical order with all the existing constructors. So like right now `whiteSpaceChars` is near the top of the file above the existing `alphaNumericString` generator but it should be towards the end of the file after the `weighted` constructor. It doesn't actually change the logic of your implementation but just makes it easier for users and contributors to find things in the file, especially when it is a large one like this one.

maxim092001: Ok, I think now they are alphabetized:)

adamgfraser (approved):

PR #4731, opened 2021-02-23
Fix possible lose of error with mapMPar

### LINKED ISSUE #4514: With two `mapMPar`s in a row, possible to lose an error
Code to reproduce:
```
"it should not show interrupted exception" in {
      def maybeError(i: Int) = {
        if (Math.random() < 0.3) {
          throw new AssertionError("Bad luck")
        }
        i
      }
      for (i <- 1 to 100) {
        try {
          unsafeRun(
            ZStream
              .fromIterable(1 to 100)
              .mapMPar(20)(i => Task(maybeError(i)))
              .mapMPar(20)(i => Task(maybeError(i)))
              .runCollect)
        } catch {
          case e =>
            if (!e.getMessage.contains("Bad luck")) {
              throw e
            } else {
              println("OK")
            }
        }
      }
    }
```

If you run above code, you see `OK` printed a couple of times but at some point you get an InterruptedException instead. 
It seems like the parent thread catches the exception from the child (the assertion error in this case) and decides to stop other children. As a result some children throw InterruptedException and sometimes that is bubbled up instead of the root cause, i.e., the assertion error in this case.

luis3m:
Close #4514


luis3m:
I'm going to update this PR, I just realized the order of the error is not preserved.

iravid:
Thanks @luis3m super useful!

luis3m:
@iravid PR updated.

luis3m:
Looking why JVM hangs

luis3m:
@iravid this is finally ready for review.

jdegoes:
@iravid Ping!

PR #4736, opened 2021-02-25
Implement Default Supervisor

adamgfraser:
Includes a supervisor in the default runtime that tracks all fibers forked in the application. The one thing that is not great here is the type of the `Supervisor` in `Runtime` is `Any` so the user has to call `Supervisor#value` to get the result of a supervision event and then pattern match on it to recover the information that it is a chunk of fibers. Possibly in 2.0 we could add a type parameter to `Runtime` for the result of a supervision event to improve this.

### DIFF
--- a/core/js/src/main/scala/zio/internal/PlatformSpecific.scala
+++ b/core/js/src/main/scala/zio/internal/PlatformSpecific.scala
@@ -116,6 +116,7 @@ private[internal] trait PlatformSpecific {
    */
   final def makeDefault(yieldOpCount: Int = defaultYieldOpCount): Platform =
     fromExecutor(Executor.fromExecutionContext(yieldOpCount)(ExecutionContext.global))
+      .withSupervisor(Supervisor.unsafeTrack(true))
 
   final def newWeakSet[A](): JSet[A] = new HashSet[A]()
 
--- a/core/jvm/src/main/scala/zio/internal/PlatformSpecific.scala
+++ b/core/jvm/src/main/scala/zio/internal/PlatformSpecific.scala
@@ -126,7 +126,7 @@ private[internal] trait PlatformSpecific {
    * Makes a new default platform. This is a side-effecting method.
    */
   def makeDefault(yieldOpCount: Int = defaultYieldOpCount): Platform =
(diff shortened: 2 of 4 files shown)


iravid:
Just wondering - can this be disabled like tracing?

adamgfraser:
That's a good question.

It can be disabled at the global level by doing something like `Runtime.default.mapPlatform(platform => platform.withSupervisor(Supervisor.none))`.

It can't be disabled locally now because `Supervise` in `FiberContext` is always interpreted as `supervisor && lastSupervisor`, which propagates supervision events to higher supervisors. We could potentially have a variant that took a parameter for whether to propagate so you could do something like `zio.supervised(Supervisor.none)` to remove supervision inside a region, though that does to some extent reduce the guarantees that as someone installing a supervisor at a higher level of an application you will really see all activity in that application.

jdegoes:
I think this will have significant overhead (a gigantic constantly changing set of millions of fibers), and at the very least, we need to make this opt-in rather than opt-out. 

A longer term goal is to ensure that if you are using ZMX, you are opting in to the tracking without additional effort.

jdegoes:
Maybe what would satisfy this is a global supervisor that is package private to `zio` and can be appended to (`unsafeInstallGlobalSupervisor`). Then ZMX can append to this supervisor the tracking one in its initialization logic, so that all fibers created after that point in time are tracked. Alternately, if we want to get really hacky, ZIO can detect if ZMX is on the classpath and install something right away, so even the first fiber would be tracked.

It's tricky. We want a combination of no overhead by default, but also a way for people to get at the information they want.

adamgfraser:
Maybe we can just have a `supervisedRuntime` in addition to the `defaultRuntime` and then people can swap that in if they want lightweight tracking without going all the way to ZIO XMX. That way we don't impact existing users at all but give users who want this functionality an easy way to get it "out of the box".

adamgfraser:
Updated to be opt in.

jdegoes:
Maybe we should merge as a PlatformAspect? Or discuss (in person or at the hackathon) using ZScheduler to do some of the work for us. Will close for now but we should keep in mind the work done here in case we need to revive it.

PR #4738, opened 2021-02-25
Cleanup useless type parameter

regiskuckaertz:
*never gets tired of studying the zio code :laughing:*


jdegoes (approved):
I think I fixed this in a private branch but good to have it here too! 🙏

PR #4742, opened 2021-02-26
Extract Gen implement

jxnu-liguobin:


jdegoes (commented):
Looks good to me, @adamgfraser may want to take a look!

test-refined/shared/src/main/scala/zio/test/refined/BooleanInstances.scala
```diff
@@ -16,8 +16,11 @@ trait BooleanInstances {
   ): DeriveGen[Refined[T, A Or B]] = {
     val genA: Gen[random.Random with Sized, T] = raGen.derive.map(_.value)
     val genB: Gen[random.Random with Sized, T] = rbGen.derive.map(_.value)
-    DeriveGen.instance(
-      Gen.oneOf[Random with Sized, T](genA, genB).map(Refined.unsafeApply)
-    )
+    DeriveGen.instance(orGen(genA, genB))
   }
+
+  def orGen[T, A, B](implicit
```
adamgfraser, line 22: When we compose generators like here we want to be polymorphic in the environment type. We can combine any two generators this way and the new generator will require all the services required by both generators, plus the `Random` service since we need to use random number generation to choose which generator to pull from each time. So I would rewrite this as:

```scala
  def orGen[R <: Random, T, A, B](implicit
    genA: Gen[R, T],
     genB: Gen[R, T]
   ): Gen[R, Refined[T, A Or B]] = Gen.oneOf[R, T](genA, genB).map(Refined.unsafeApply)
```

test-refined/shared/src/main/scala/zio/test/refined/BooleanInstances.scala
```diff
@@ -16,8 +16,11 @@ trait BooleanInstances {
...
+    genA: Gen[random.Random with Sized, T],
+    genB: Gen[random.Random with Sized, T]
+  ): Gen[Random with Sized, Refined[T, A Or B]] = Gen.oneOf[Random with Sized, T](genA, genB).map(Refined.unsafeApply)
```
adamgfraser, line 25: Do we need to specify the type parameters to `oneOf` here?

test-refined/shared/src/main/scala/zio/test/refined/CollectionInstances.scala
```diff
@@ -2,73 +2,108 @@ package zio.test.refined
...
 import zio.{Chunk, NonEmptyChunk}
+import zio.random.Random
+import zio.test.{Gen, Sized}
+import zio.test.magnolia.DeriveGen
 
 object collection extends CollectionInstances
 
 trait CollectionInstances {
+
+  def nonEmptyChunkRefinedGen[C, T](implicit
```
adamgfraser, line 14: Similarly I think you can do `R <: Random with Sized` here.

test-refined/shared/src/main/scala/zio/test/refined/CharInstances.scala
```diff
@@ -18,30 +18,28 @@ trait CharInstances {
   private val whitespaceChars: Seq[Char] =
     (Char.MinValue to Char.MaxValue).filter(_.isWhitespace)
 
+  val digitGen: Gen[Random, Refined[Char, Digit]]   = numericCharGen.map(value => Refined.unsafeApply(value))
```
adamgfraser, line 21: Some Scaladoc would be great here as well!

test-refined/shared/src/main/scala/zio/test/refined/CollectionInstances.scala
```diff
@@ -2,73 +2,108 @@ package zio.test.refined
...
+    genT: Gen[Random with Sized, T]
+  ): Gen[Random with Sized, Refined[NonEmptyChunk[T], NonEmpty]] =
+    nonEmptyChunkGen(genT).map(Refined.unsafeApply)
+
+  def nonEmptyListRefinedGen[T](implicit
+    genT: Gen[Random with Sized, T]
+  ): Gen[Random with Sized, Refined[List[T], NonEmpty]] =
+    nonEmptyChunkGen(genT)
```
adamgfraser, line 22: Maybe construct these directly instead of going through `Chunk`. `Gen.listOf1(genT).map(Refined.unsafeApply)`,

test-refined/shared/src/main/scala/zio/test/refined/CollectionInstances.scala
```diff
@@ -2,73 +2,108 @@ package zio.test.refined
...
     )
 
-  private def nonEmptyChunkDeriveGen[T](implicit
-    arbT: DeriveGen[T]
-  ): DeriveGen[NonEmptyChunk[T]] =
-    DeriveGen.instance(Gen.chunkOf1(arbT.derive))
+  private[refined] def nonEmptyChunkGen[T](
+    arbT: Gen[Random with Sized, T]
+  ): Gen[Random with Sized, NonEmptyChunk[T]] =
+    Gen.chunkOf1(arbT)
```
adamgfraser, line 99: Not sure it Is worth implementing separate helper methods here since you are just called a single method or at most two in the sized variant below.

adamgfraser (changes requested):
Looks good. A couple of comments on the implementation but definitely going in the right direction!

test-refined/shared/src/main/scala/zio/test/refined/CharInstances.scala
jxnu-liguobin: This place has been moved to another place.

jxnu-liguobin: This has been moved to `object Gen`

test-refined/shared/src/main/scala/zio/test/refined/CollectionInstances.scala
```diff
@@ -2,73 +2,99 @@ package zio.test.refined
...
+    sizeGen: Gen[Random, Int Refined P]
+  ): Gen[R, Refined[Chunk[T], Size[P]]] =
+    sizedChunkGen(genT, sizeGen)
+      .map(r => Refined.unsafeApply(r))
+
+  def listSizeRefinedGen[R <: Random with Sized, T, P](implicit
+    genT: Gen[R, T],
+    sizeGen: Gen[Random, Int Refined P]
+  ): Gen[R, Refined[List[T], Size[P]]] =
+    sizedChunkGen(genT, sizeGen)
```
adamgfraser, line 38: I would do `sizeGen.flatMap(n => listOfN(n.value)(genT))` and similarly below to avoid the unnecessary conversion between collection types.

test-refined/shared/src/main/scala/zio/test/refined/StringInstances.scala
```diff
@@ -3,28 +3,37 @@ package zio.test.refined
...
-import zio.test.Gen
+import zio.random.Random
 import zio.test.magnolia.DeriveGen
+import zio.test.{Gen, Sized}
 
 object string extends StringInstances
 
 trait StringInstances {
+
+  def endsWithStringGen[S <: String](implicit
```
adamgfraser, line 14: I think we probably want to parameterize these generators on a `Char` generator and a size similar to the existing `string` and `stringN` generators and the collection variants we implemented above so the user can potentially generate different types of strings.

adamgfraser (approved):
Looking really good! A few more changes and this will be ready to go.

test-refined/shared/src/main/scala/zio/test/refined/CollectionInstances.scala
jxnu-liguobin: But this does not extract their logic into the new method.

adamgfraser: Maybe I'm not understanding. It seems like you want to construct three things in this and the methods below:
1. Chunks with size that confirms to a refinement type
2. Lists with size that conforms to a refinement type
3. Vectors with size that confirms to a refinement type

Right now you're implementing `sizedChunkGen` which generates (1) and then you are using the `toList` and `toVector` operators to generate (2) and (3). From a code re-use perspective that uses slightly less code since you are using `sizedChunkGen` everywhere.

However, it means that the generators (2) and (3) above are unnecessarily inefficient because they first build a `Chunk` and then tear it down and rebuild it into a `List` or `Vector`. Furthermore, the logic in `sizedChunkGen` is fairly trivial, since it basically just combines two existing operators in the most straightforward way possible. So the gains to this versus just implementing (2) and (3) directly in terms of `listOfN` and `vectorOfN` is minimal.

Does that make sense?

jxnu-liguobin: Oh, I know how to do it.

test-refined/shared/src/main/scala/zio/test/refined/StringInstances.scala
jxnu-liguobin: Please see if this is reasonable.Thanks
```scala
package zio.test.refined

import eu.timepit.refined.api.Refined
import eu.timepit.refined.string.{EndsWith, StartsWith, Uuid}
import shapeless.Witness
import zio.random.Random
import zio.test.magnolia.DeriveGen
import zio.test.{Gen, Sized}

object string extends StringInstances

trait StringInstances {

  def endsWithStringGen[R <: Random with Sized, S<:String](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char]
  ): Gen[R, Refined[S, EndsWith[S]]] = Gen.string(charGen).map(v => Refined.unsafeApply(v + ws.value))

  def endsWithString1Gen[R <: Random with Sized, S<:String](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char]
  ): Gen[R, Refined[S, EndsWith[S]]] = Gen.string1(charGen).map(v => Refined.unsafeApply(v + ws.value)) //Only String can call `+` since 2.13.0.

  def endsWithStringNGen[R <: Random with Sized, S<:String, P](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char],
    sizeGen: Gen[R, Int Refined P]
  ): Gen[R, Refined[S, EndsWith[S]]] =
    sizeGen.flatMap(s => Gen.stringN(s.value)(charGen).map(v => Refined.unsafeApply(v + ws.value)))

  def startsWithStringGen[R <: Random with Sized, S<:String](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char]
  ): Gen[R, Refined[S, StartsWith[S]]] = Gen.string(charGen).map(v => Refined.unsafeApply(ws.value + v))

  def startsWithString1Gen[R <: Random with Sized, S<:String](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char]
  ): Gen[R, Refined[S, StartsWith[S]]] = Gen.string1(charGen).map(v => Refined.unsafeApply(ws.value + v))

  def startsWithStringNGen[R <: Random with Sized, S<:String, P](implicit
    ws: Witness.Aux[S],
    charGen: Gen[R, Char],
    sizeGen: Gen[R, Int Refined P]
  ): Gen[R, Refined[S, StartsWith[S]]] =
    sizeGen.flatMap(s => Gen.stringN(s.value)(charGen).map(v => Refined.unsafeApply(ws.value + v)))

  val uuidStringGen: Gen[Random, Refined[String, Uuid]] = Gen.anyUUID.map(value => Refined.unsafeApply(value.toString))

  implicit val uuidStringDeriveGen: DeriveGen[Refined[String, Uuid]] =
    DeriveGen.instance(uuidStringGen)

  implicit def endsWithStringDeriveGen[S<:String](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char]
  ): DeriveGen[Refined[S, EndsWith[S]]] =
    DeriveGen.instance(endsWithStringGen(ws, deriveGenT.derive))

  implicit def startsWithStringDeriveGen[S<:String](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char]
  ): DeriveGen[Refined[S, StartsWith[S]]] =
    DeriveGen.instance(startsWithStringGen(ws, deriveGenT.derive))

  implicit def endsWithString1DeriveGen[S<:String](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char]
  ): DeriveGen[Refined[S, EndsWith[S]]] =
    DeriveGen.instance(endsWithString1Gen(ws, deriveGenT.derive))

  implicit def startsWithString1DeriveGen[S<:String](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char]
  ): DeriveGen[Refined[S, StartsWith[S]]] =
    DeriveGen.instance(startsWithString1Gen(ws, deriveGenT.derive))

  implicit def endsWithStringNDeriveGen[S<:String, P](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char],
    deriveGenSize: DeriveGen[Int Refined P]
  ): DeriveGen[Refined[S, EndsWith[S]]] =
    DeriveGen.instance(endsWithStringNGen(ws, deriveGenT.derive, deriveGenSize.derive))

  implicit def startsWithStringNDeriveGen[S<:String, P](implicit
    ws: Witness.Aux[S],
    deriveGenT: DeriveGen[Char],
    deriveGenSize: DeriveGen[Int Refined P]
  ): DeriveGen[Refined[S, StartsWith[S]]] =
    DeriveGen.instance(startsWithStringNGen(ws, deriveGenT.derive, deriveGenSize.derive))

}

```

adamgfraser: Yes exactly!

adamgfraser (approved):
Looks fantastic! Thanks again for working on this!

PR #4751, opened 2021-02-28
Add TSemaphore To Index

adamgfraser:
We have existing documentation on `TSemaphore` but it was not being included in the index [here](https://zio.dev/docs/datatypes/datatypes_index).


adamgfraser:
Updated to also include `TSemaphore` and a couple of other data types on left hand navigation bar.

jdegoes (approved):

PR #4752, opened 2021-02-28
Flatten exit in supervisor unsafeOnEnd

mikearnaldi:
As per discussion in Discord `unsafeOnEnd` should get the actual exit `Exit[E,A]` and not `Exit[Nothing, Exit[E,A]]` that is provided by `FiberContext#onDone` this PR uses `flatten` to re-enstablish the correct exit value


adamgfraser (approved):
Thank you!

PR #4760, opened 2021-03-03
organize-imports-0.5.0

sideeffffect:
organize-imports 0.5.0 has the same import order as IntelliJ
https://github.com/liancheng/scalafix-organize-imports/releases/tag/v0.5.0


adamgfraser (approved):
Thank you for being on top of this stuff!

PR #4761, opened 2021-03-04
Add NotNull and ensure refineToOrDie has type parameter

sideeffffect:
MiMa fails with
```scala
[error] zio: Failed binary compatibility check against dev.zio:zio_2.12:1.0.1! Found 2 potential problems (filtered 181)
[error]  * extension method refineToOrDie$extension(zio.ZIO,scala.reflect.ClassTag,zio.CanFail)zio.ZIO in object zio.ZIO#ZioRefineToOrDieOps does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("zio.ZIO#ZioRefineToOrDieOps.refineToOrDie$extension")
[error]  * method refineToOrDie(scala.reflect.ClassTag,zio.CanFail)zio.ZIO in class zio.ZIO#ZioRefineToOrDieOps does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("zio.ZIO#ZioRefineToOrDieOps.refineToOrDie")
```
@adamgfraser Do we want to ignore this error or do we want to put this change to ZIO 2.0?


adamgfraser:
I'm not sure this makes sense. Doing `zio.refineToOrDie` without specifying a type results in `E1` being inferred as `Nothing`, which is the equivalent of `orDie`. It terminates the fiber with any of the errors, which I don't think I can even say is erroneous behavior. Furthermore, I don't think I have seen anyone unintentionally using `refineToOrDie` without specifying a type parameter and this actually prevents users from specifying type`Nothing` as the type parameter if they want to. While `orDie` would be a more specialized operator for that, it seems like you should also be able to do `refineToOrDie[Nothing]` if you want to and that has a very logical meaning.

This also isn't really related to `null`. If we really want to prevent users from using `Nothing` as the error type for `E1` we can add a `CanFail` constraint to `E1` without any other changes to the code base, but I'm not sure that really makes sense for the reasons outlined above. I don't think there is actually a problem we have observed here and this actually prevents some reasonable behavior.

sideeffffect:
`refineToOrDie` without type parameter is almost always a mistake, not what the user intends. And as you say, there's `orDie` when the user actually wants for every error to go to the die channel.

But you're right, that we don't need `NonNull` for this at all -- that was just me being too eager cargo cultist :smile_cat:

sideeffffect:
@adamgfraser I suppose you're not a big fan of this, so closing

PR #4772, opened 2021-03-07
Explore Improving Ergonomics Of Providing Partial Layers

adamgfraser:
@deusaquilus had the really interesting idea that if we provide the partial environment first then the Scala compiler is able to infer the remaining environment even though it cannot normally do so without the type being explicitly provided. This allows syntax such as:

```scala
// eliminates `Clock` requirement, leaving other requirements outstanding
ZIO.provideSomeLayer(clockLayer)(zio)
```


jdegoes (approved):

kitlangton:
Here's a version for ZLayer that similarly correctly infers the set difference between layer requirements.
So `URLayer[LIn, LOut] >%> URLayer[LOut with Remainder, ROut]` will return `URLayer[LIn with Remainder, ROut]`.

```scala
  /**
   * Feeds the output services of this layer into the some of the input of the
   * specified layer, resulting in a new layer with the combined inputs of this
   * layer and the remainder, and the outputs of the specified layer.
   */
  final def >%>[RIn2 <: Has[_], E1 >: E, ROut1 >: ROut, ROut2](
    that: ZLayer[ROut1 with RIn2, E1, ROut2]
  )(implicit ev: Has.Union[ROut1, RIn2], tag: Tag[RIn2]): ZLayer[RIn with RIn2, E1, ROut2] =
    self.zipWithPar(requires[RIn2])(ev.union) >>> that
```

Same thing really, but this could work nicely with the inject macro, as one could provide a specific dependency to a specific layer.

kitlangton:
Okay, I was wrong :P. But I figured out how to get it to compile:

```scala
object LayerOps {
  implicit final class ProvideSomeInputPartiallyApplied[R0, E, R <: Has[_]](val self: ZLayer[R0, E, R]) extends AnyVal {
    def >%>[R1 <: Has[_] : Tag, E1 >: E, A](that: ZLayer[R with R1, E1, A])(implicit ev: Has.Union[R, R1], tag: Tag[R]): ZLayer[R0 with R1, E1, A] =
      (self ++ ZLayer.requires[R1]) >>> that
  }
}
```

Combined with a `toLayer` extension method on functions, the following compiles (as well as infers in IntelliJ):

```scala
  trait Foo {
    def bar: UIO[Int]
  }

  object Foo {
    val live: ZLayer[Has[Boolean] with Has[String], Nothing, Has[Foo]] =
      Ref.make(10).toLayer >%> FooLive.toLayer
  }

  case class FooLive(int: Ref[Int], string: String, boolean: Boolean) extends Foo {
    override def bar: UIO[Int] = int.get
  }
```

With this it's possible to write a `case class` for a Live implementation, and merely override the effect produced layers (like Refs or Queues) by using `>%>`.

vigoo:
What about having it on `ZLayer` like

```scala
clockLayer.providedFor(zio)
```

jdegoes:
@vigoo That's an idea worth exploring.

adamgfraser:
I think the primary issue here is around cross version compatibility. The trick above works on Scala 2 but not Scala 3. So I think our next step has to be to try to develop a version of this that works on Scala 3 and then see if we can unify the two.

PR #4773, opened 2021-03-07
Fix deprecated empty argument list auto-application in macros

### LINKED ISSUE #4735: zio-macros: @accessible fails for an empty argument list
For the given service declaration 

```
@accessible
object FooService {
  trait Service {
    def foo(): Task[Unit]
  }
}
```

The compiler fails with the following error
```
Auto-application to `()` is deprecated. Supply the empty argument list `()` explicitly to invoke method foo,
or remove the empty argument list from its definition (Java-defined methods are exempt).
In Scala 3, an unapplied method like this will be eta-expanded into a function.
@accessible
```

But for 
```
@accessible
object FooService {
  trait Service {
    def foo: Task[Unit]
  }
}
```
It works fine. Is it an expected behaviour?

ioleo:
Fixes #4735

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -428,6 +428,7 @@ lazy val examples = crossProject(JVMPlatform, JSPlatform)
   .settings(stdSettings("examples"))
   .settings(crossProjectSettings)
   .settings(macroExpansionSettings)
+  .settings(scalacOptions += "-Xfatal-warnings")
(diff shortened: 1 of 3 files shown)


ioleo:
The error is in `ZStream JVM` -> `fromOutputStreamWriter` test, unrelated to this PR. I guess this must be flaky or failing on master.

adamgfraser (approved):

PR #4776, opened 2021-03-08
Fix TMap.toChunk ArrayIndexOutOfBoundsException

### LINKED ISSUE #4648: STM: NullPointerException and ArrayIndexOutOfBoundsException in TMap
This might be an issue with TMap (haven't tried with other STM data structures yet)

```
import zio.stm.{TMap, ZSTM}
import zio.Runtime
import zio.{ZIO, _}

Runtime.default.unsafeRun {
  for {
      map <- TMap.make[Int, Int]().commit
      _ <- ZIO.foreach_(0 to 10) { i =>
        for {
        _ <- map.put(i, i).commit
        _ <- map.delete(i).commit.fork
        _ <- map.foreach((_, _) => ZSTM.unit).commit.fork
      } yield ()
    }
  } yield map
}
```
https://scastie.scala-lang.org/DaD8pAXySjSitvvOCCcBfA

- NullPointerException happens intermittently in case of a race between delete  and foreach
- ArrayIndexOutOfBoundsException happens intermittently while running foreach (it is independent of deleting entry from map, try removing `_ <- map.delete(i).commit.fork` in above code

alextoth:
Resolves #4648

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
@@ -17,8 +17,9 @@
 package zio.stm
 
 import zio.test.Assertion._
+import zio.test.TestAspect.nonFlaky
 import zio.test._
-import zio.{URIO, ZIOBaseSpec}
+import zio.{URIO, ZIO, ZIOBaseSpec}
 
 object TMapSpec extends ZIOBaseSpec {
 
@@ -336,6 +337,22 @@ object TMapSpec extends ZIOBaseSpec {
 
         assertM(tx.commit)(equalTo(0))
       }
+    ),
+    suite("folds")(
+      testM("avoid NullPointerException caused by race condition") {
+        for {
+          keys <- ZIO.succeed((0 to 10).toList)
+          map  <- TMap.fromIterable(keys.zipWithIndex).commit
+          exit <- ZIO
+                    .foreach_(keys) { k =>
+                      for {
+                        _ <- map.delete(k).commit.fork
+                        _ <- map.toChunk.commit
+                      } yield ()
+                    }
+                    .run
+        } yield assert(exit)(succeeds(isUnit))
+      } @@ nonFlaky
     )
   )
 
--- a/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
@@ -338,7 +338,7 @@ object TMapSpec extends ZIOBaseSpec {
         assertM(tx.commit)(equalTo(0))
       }
     ),
-    suite("folds")(
+    suite("bug #4648")(
       testM("avoid NullPointerException caused by race condition") {
         for {
           keys <- ZIO.succeed((0 to 10).toList)
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -292,9 +292,8 @@ final class TMap[K, V] private (
 
       while (i < capacity) {
         val bucket = buckets.array(i)
-        val pairs  = bucket.unsafeGet(journal)
 
-        builder.addAll(pairs)
+        builder ++= bucket.unsafeGet(journal)
 
         i += 1
       }
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -19,6 +19,8 @@ package zio.stm
 import zio.Chunk
 import zio.stm.ZSTM.internal._
 
+import scala.collection.mutable.ArrayBuffer
+
 /**
  * Transactional map implemented on top of [[TRef]] and [[TArray]]. Resolves
  * conflicts via chaining.
@@ -85,7 +87,7 @@ final class TMap[K, V] private (
    * Atomically folds using a transactional function.
    */
   def foldM[A, E](zero: A)(op: (A, (K, V)) => STM[E, A]): STM[E, A] =
-    toChunk.flatMap(STM.foldLeft(_)(zero)(op))
+    toChunk.flatMap(ZSTM.foldLeft(_)(zero)(op))
 
   /**
    * Atomically performs transactional-effect for each binding present in map.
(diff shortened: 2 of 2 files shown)


mijicd:
@alextoth Could you please describe the issue a bit more? `tSize` should be the exact number of elements in the map, if that's not true, it might indicate a bigger problem.

alextoth:
@mijicd Well I may be mistaken. But I didn't find anything that guaranties that when modifying the pair (tBuckets, tSize) the modification is atomic. Moreover I don't find evidence of isolation, as this could have prevented the problem.
So concurrency can do what it wants and there are times where both would not be consistent. Be it when modifying or when reading.
Hence, yes it would mean a bigger problem.

as a simple illustration (one example in many) : 
two threads/fibers in concurrency : 
```
tBuckets write
                                   tBuckets read
                                   tSize read
tSize write

```

Once again I'm new to ZIO's STM but not new to concurrent code.

mijicd:
True, but these transactional variables (`tSize` and `tBuckets`) are always edited in the same transaction, meaning that their intermediate results won't be visible until committed.

alextoth:
True.
Atomicity at the transaction level is ensured by one `Journal` per transaction and `Sync(globalLock)`.
Isolation between transactions is given by `Journal / Entry` tandem + `TRef` delegates to `Journal/Entry`

So this leaves us with concurrent Thread/Fibers sharing the same transaction. Doesn't look obvious how this could happen in the problematic example ( #4648 ).
The problem remains since we can see that running the problematic example ( #4648 ) gives us out sync 'tSize' and 'tBuckets'

So basically you're saying my patch solves the issue but for the wrong reason. And I agree. Is the problem that there would be parallelism within the transaction?

mijicd:
Ah, ok, I finally understand what you are referring to, sorry. I've already digged at the referenced issue. Long story short, let me try to wrap it up later today, and then I'll get back to this PR.

As i mentioned in the initial comment, I'd like to solve the broader problem here, as it obviously exists.

mijicd:
Small update to my previous comment - the proposed patch doesn't fix the test I created based on #4648 discussion:

```scala
    suite("bug #4648")(
      testM("avoid NullPointerException caused by race condition") {
        for {
          keys <- ZIO.succeed((0 to 10).toList)
          map  <- TMap.fromIterable(keys.zipWithIndex).commit
          exit <- ZIO
                    .foreach_(keys) { k =>
                      for {
                        _ <- map.delete(k).commit.fork
                        _ <- map.toChunk.commit
                      } yield ()
                    }
                    .run
        } yield assert(exit)(succeeds(isUnit))
      } @@ nonFlaky
    )
```

alextoth:
Thanks @mijicd for the clarification and the test code.
This is interesting, when I run your test without the patch it fails.
When I run your test with the patch the **test passes**.

I tried the default 100 and 5k repetitions.
Any idea what I'm doing wrong?

```scala
package zio

import zio.stm._
import zio.test.Assertion._
import zio.test._
import zio.test.TestAspect.nonFlaky

object _Non_Regression extends ZIOBaseSpec {

  def spec: ZSpec[Environment, Failure] =
    suite("bug #4648") {
      testM("avoid NullPointerException caused by race condition") {
        for {
          keys <- ZIO.succeed((0 to 100).toList)
          map <- TMap.fromIterable(keys.zipWithIndex).commit
          exit <- ZIO
            .foreach_(keys) { k =>
              for {
                _ <- map.delete(k).commit.fork
                _ <- map.toChunk.commit
              } yield ()
            }
            .run
        } yield assert(exit)(succeeds(isUnit))
      } @@ nonFlaky(5000)
    }
}
```

mijicd:
Hah, failures were caused by the testing helper (see the following snippet), when removed, the patch indeed solves the issue.

```scala
 private def checkInvariance[R, E, A](label: String)(stm: ZSTM[R, E, A]): ZSTM[R, E, A] =
  invariant(s"$label before") *> stm <* invariant(s"$label after")

  private def invariant(label: String): USTM[Unit] =
    new ZSTM((journal, _, _, _) => {
      val buckets = tBuckets.unsafeGet(journal)
      val array   = buckets.array

      val actual   = array.map(_.unsafeGet(journal).length).sum
      val expected = tSize.unsafeGet(journal)

      require(actual == expected, s"$label $actual isn't $expected")

      TExit.unit
    })
```

mijicd:
To push it even further, the following implementation passes the test:

```scala
def toChunk: USTM[Chunk[(K, V)]] =
  new ZSTM((journal, _, _, _) => {
        val capacity = buckets.array.length
        val size     = tSize.unsafeGet(journal)
        var i        = 0

        val builder = ChunkBuilder.make[(K, V)](size)

        while (i < capacity) {
          val bucket = buckets.array(i)
          val pairs  = bucket.unsafeGet(journal)

          builder.addAll(pairs)

          i += 1
        }

        TExit.Succeed(builder.result())
      })
```

alextoth:
`buckets` seems to be missing but I get the gist of your code.
Well done.

mijicd:
Yeah, that's just a copy-pasta typo. Fixing it.

mijicd:
@alextoth I propose the following changes:
- append the suite to `TMapSpec`
- use `ChunkBuilder`

We can wait for a green build then, but other than that, it looks good to go.

core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
```diff
@@ -336,6 +337,22 @@ object TMapSpec extends ZIOBaseSpec {
 
         assertM(tx.commit)(equalTo(0))
       }
+    ),
+    suite("folds")(
```
mijicd, line 341:
```suggestion
    suite("bug #4648")(
```

I'd leave a reference for the future, plus there are folds among operations :).

mijicd (approved):
Thank you for digging into this, and congrats on the first contribution! :rocket:

alextoth:
Thanks a lot for your help, support and comments.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -287,24 +287,19 @@ final class TMap[K, V] private (
...
       while (i < capacity) {
         val bucket = buckets.array(i)
         val pairs  = bucket.unsafeGet(journal)
 
-        val it = pairs.iterator
-        while (it.hasNext) {
-          data(j) = it.next()
-          j += 1
-        }
+        builder.addAll(pairs)
```
mijicd, line 297: One last thing, since cross-build is "yelling", this can be replaced with `++=`. So:

```scala
while (i < capacity) {
  val bucket = buckets.array(i)

  builder ++= bucket.unsafeGet(journal)

  i += 1
}
```

adamgfraser:
Sorry to jump in late, but do we feel like we understand the underlying cause of the issue here?

mijicd:
@adamgfraser Basically it's a dirty read, as described by @alextoth above. `toChunk` was making an assumption of size, while the size was modified later on (hence array failed with an exception).

adamgfraser:
@mijicd Got it! Thanks!

Issue #4786, opened 2021-03-10
Testing STM for atomicity fails


alextoth:
I designed a test to verify that the atomicity of transactions is preserved inside the STM. This is a followup on #4648.
If atomicity is not preserved, It can not be compensated later on and lead to a false negative. Said differently succeeding failures can not cancel each other. Such situations would lead to think the test passed when it did not.

The reason why this test fails has to do with `zio.internal.Sync` (a wrapper to native synchronised) and the interaction with either the compiler or the JVM.

As evidence of something is wrong (aside of this test):
A/ you can in the code of ZSTM replace `globalLock` value `new AnyRef` with `null` this should trigger a NPE each time the lock is acquired => It never triggers a NPE.   
```scala
val globalLock: AnyRef = null
```
B/ When placing breakpoints with the debugger inside the `Sync(globalLock){` sections, you will see after a few iterations that multiple threads are at the breakpoints inside the critical sections, which is impossible if the sections where locked by `globalLock`.

To solve this the only thing that worked for me  is : 
A/ replace `Sync(globalLock)` by its less magical  counterpart `globalLock.synchronized` 

I guess this is not an issue inside ZIO but inside the compile/exec chain. 

My environment is: scala-2.13.5 && ( java8 || java11.0.2 || java15.0.2) on intel platform macOS

I hope I'm doing something wrong.

```scala
object STM_issue_globalLock_tref extends zio.App {

  import zio.stm.{TRef, ZSTM}
  import zio.{ZIO, _}

  val max = 99

  def run(args: List[String]) = {
    /**
     * If the transaction is executed atomically  it is :
     * sum = sum + ref + i
     * ref = i
     */
    def transaction(i: Int, ref: TRef[Int], sum: TRef[Int]) =
      for {
        ref_prev <- ref.get
        sum_prev <- sum.get
        _ <- sum.set(sum_prev + ref_prev + i)
        _ <- ref.set(i)
      } yield i

    (for {
      (sum, ref) <- ZSTM.atomically(TRef.make[Int](0) zip TRef.make[Int](0))

      fibers <- ZIO.forkAll {
        scala.util.Random.shuffle(1 to max).map { i => ZSTM.atomically(transaction(i, ref, sum)).fork }
      }
      _ <- fibers.join
      (sum, ref) <- ZSTM.atomically {
        sum.get zip ref.get
      }
      passed = sum + ref == (max + 1) * max
[...]

PR #4788, opened 2021-03-11
putting STM atomicity to the test #4786

alextoth:
* Add to the test SPEC suite

### DIFF
--- a/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
@@ -1392,6 +1392,33 @@ object ZSTMSpec extends ZIOBaseSpec {
 
         assertM(tx.commit)(equalTo((Left("error"), "error")))
       } @@ zioTag(errors)
+    ),
+    suite("issue #4786")(
+      testM("check atomicity ") {
+        val max = 99
+        def transaction(i: Int, ref: TRef[Int], sum: TRef[Int]) =
+          for {
+            ref_prev <- ref.get
+            sum_prev <- sum.get
+            _        <- sum.set(sum_prev + ref_prev + i)
+            _        <- ref.set(i)
+          } yield i
+
+        for {
+          (sum, ref) <- ZSTM.atomically(TRef.make[Int](0) zip TRef.make[Int](0))
+
+          fibers <- ZIO.forkAll {
+                      scala.util.Random.shuffle(1 to max).map { i =>
+                        ZSTM.atomically(transaction(i, ref, sum)).fork
+                      }
+                    }
+          _ <- fibers.join
+          (sum, ref) <- ZSTM.atomically {
+                          sum.get zip ref.get
+                        }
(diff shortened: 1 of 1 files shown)


core-tests/shared/src/test/scala/zio/stm/ZSTMSpec.scala
```diff
@@ -1392,6 +1392,33 @@ object ZSTMSpec extends ZIOBaseSpec {
...
+          } yield i
+
+        for {
+          (sum, ref) <- ZSTM.atomically(TRef.make[Int](0) zip TRef.make[Int](0))
+
+          fibers <- ZIO.forkAll {
+                      scala.util.Random.shuffle(1 to max).map { i =>
+                        ZSTM.atomically(transaction(i, ref, sum)).fork
+                      }
+                    }
```
mijicd, line 1414: I don't think this is necessary an issue with STM. This part involves nested forking, which can be treated as an antipattern. Dropping it or replacing with forkDaemon would cause the test to pass.

cc @adamgfraser

adamgfraser: @FrancisToth There is an issue with your test.

Notice that you are actually forking two levels of fibers here. `forkAll` forks one set of fibers and then each of those does as its only working forking another set of fibers which are supposed to perform the actual STM transaction.

This creates a number of issues unrelated to STM. First, due to fiber supervision when the child fiber completes execution it attempts to interrupt its children, so the grandchild fibers that are actually supposed to do the STM transaction may be interrupted during or even before they begin execution, since the child fibers do no work other than forking the child.

Second, this means that when the fibers are joined on the line below we do not have any guarantee that the work has actually been completed.

The easiest way to fix this test is to remove the extraneous `fork` entirely like this:

```scala
          fibers <- ZIO.forkAll {
                      scala.util.Random.shuffle(1 to max).map { i =>
                        ZSTM.atomically(transaction(i, ref, sum))
                      }
                    }
          _ <- fibers.join
```

In this case the test passes.

If we do want to fork the second level of fibers we would have to fork them as daemon fibers to make sure they are not interrupted due to auto supervision and we would also have to join back both the children and grandchildren to ensure that they have completed execution like this:

```scala
           fibers <- ZIO.forkAll {
                       scala.util.Random.shuffle(1 to max).map { i =>
                         ZSTM.atomically(transaction(i, ref, sum)).forkDaemon
                       }
                     }
           _ <- fibers.join.flatMap(fibers => ZIO.foreach(fibers)(_.join))
```

In this case the test passes as well.

So I think we can close this PR and the associated issue.

Copying @mijicd.

alextoth: Thanks a lot @mijicd and @adamgfraser  for the details and your time.
I'm sorry this is a clear mistake on my part.

alextoth:
Thanks a lot @mijicd and @adamgfraser for the details and your time.
I'm sorry this is a clear mistake on my part.

PR #4795, opened 2021-03-12
Remove Test Dependency From ZIO Macros

### LINKED ISSUE #4787: weird dependency between projects
`zio-macros` has compile (not test) dependencies on `zio-test-sbt` which itself has compile (not test) dependencies on `zio-test`.

This inhibits clean use in build systems like `bazel`, which would prohibit testonly dependencies (like `zio-test`) from being transitive compile dependencies of non-test code.  This basically makes zio-macros unusable (at least, without some possibly truly ugly hacks).

Is this dependency correct?  can it be made a test-only dependency instead?  If not, what's the rationale for relying on a testing library for non-test code?

adamgfraser:
Resolves #4787.


adamgfraser:
@mijicd Could you take a look at this when you have a chance?

mijicd (approved):

PR #4808, opened 2021-03-16
Update README.md

ftrianakast:
In eBay Germany, we use actively ZIO in one of our Global Projects. We are really happy with it!


adamgfraser (approved):
@ftrianakast That's fantastic! So happy to hear it!

PR #4811, opened 2021-03-17
Add type-inferrable toLayer extension on functions

kitlangton:
Unlike `ZLayer.fromServices`, this can completely infer the input types. It can save a lot of boilerplate in the common cases. I can also be used in interesting ways if we add a variant of `>>>` that is not required to fully satisfy the dependencies of the righthand side (e.g., https://github.com/zio/zio/pull/4772#issuecomment-799222154).

P.S. I'll go up to the magic number 22 and beyond. I just first want to make sure this is alright with everyone before giving myself unnecessary RSI 😜

### DIFF
--- a/core/shared/src/main/scala/zio/FunctionToLayerSyntax.scala
+++ b/core/shared/src/main/scala/zio/FunctionToLayerSyntax.scala
@@ -1,0 +1,930 @@
+package zio
+
+trait FunctionToLayerOps {
+  implicit final class Function0ToLayerSyntax[A: Tag](self: () => A) {
+
+    /**
+     * Converts this function to a Layer.
+     *
+     * {{{
+     * case class FooLive() extends Foo
+     *
+     * val live: ULayer[Has[Foo]] =
+     *   FooLive.toLayer
+     * }}}
+     */
+    def toLayer: URLayer[Any, Has[A]] =
+      UIO(self()).toLayer
+  }
+
+  implicit final class Function1ToLayerSyntax[A: Tag, B: Tag](self: A => B) {
+
+    /**
+     * Converts this function to a Layer that depends upon its inputs.
+     *
+     * {{{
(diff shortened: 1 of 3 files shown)


adamgfraser (approved):
Very interesting use of case classes!

kitlangton:
Thanks @adamgfraser!

Just realized, I didn't make these `AnyVal`, I think I could as long as I move the `Tag` requirement onto `apply`. The only issue with that is that one cannot define `AnyVal`s on traits, so they'd all have to hang out on the bottom of the `zio` package object. Am I mistaken about that? If not, should I just move them there and `AnyVal` them up or leave them in a trait?

adamgfraser:
@kitlangton I think I would leave it as you have it. The overhead of one additional allocation is going to be trivial compared to the cost of constructing even a simple layer.

kitlangton:
Nice! I'll add the overloads then :)

kitlangton:
Overloaded 🔌 !

adamgfraser (dismissed):

neko-kai:
I thought case classes extending Function goes away in Scala 3+ (replaced by implicit `new`)

kitlangton:
@neko-kai The example is compiling on both at the moment, luckily. Do you mean that it's likely to get changed post RC-1?

kitlangton:
Whew, okay, changing it to `(FooLive.apply _).toLayer` will work on both. Now `dotty_jdk8` is failing for the _good_ reason.

However, maybe I should just open this against ZIO 1.0.0, as it's completely orthogonal and would be nice to have earlier.

adamgfraser:
If it can be binary compatible yes it is better to submit it against the 1.0 branch.

kitlangton:
Cherry-picked onto master 👍

adamgfraser (approved):

PR #4813, opened 2021-03-17
Change `>>>` to allow (inferable) remainder

kitlangton:
As described here:https://github.com/zio/zio/pull/4772#issuecomment-799222154

Unfortunately, I can't fully replace the old version, as the new one has more implicit constraints that don't work internally, but I've moved the old implementation of `>>>` to `to`, using that internally. I could also move `to` to the implicit class as a proper named alias of `>>>` as it was before, I'll just need a new name for the original implementation.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -75,15 +75,8 @@ sealed abstract class ZLayer[-RIn, +E, +ROut] { self =>
   final def >+>[E1 >: E, RIn2 >: ROut, ROut1 >: ROut, ROut2](
     that: ZLayer[RIn2, E1, ROut2]
   )(implicit ev: Has.Union[ROut1, ROut2], tagged: Tag[ROut2]): ZLayer[RIn, E1, ROut1 with ROut2] =
-    ZLayer.ZipWith(self, self >>> that, ev.union)
+    ZLayer.ZipWith(self, self andThen that, ev.union)
 
-  /**
-   * Feeds the output services of this layer into the input of the specified
-   * layer, resulting in a new layer with the inputs of this layer, and the
-   * outputs of the specified layer.
-   */
-  final def >>>[E1 >: E, ROut2](that: ZLayer[ROut, E1, ROut2]): ZLayer[RIn, E1, ROut2] =
-    fold(ZLayer.fromFunctionManyM { case (_, cause) => ZIO.halt(cause) }, that)
 
   /**
    * A named alias for `++`.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2382,4 +2377,34 @@ object ZLayer {
...
+     */
+    def >>>[RIn2, E1 >: E, ROut2](that: ZLayer[ROut with RIn2, E1, ROut2])(implicit ev: Has.Union[RIn2, ROut], tag: Tag[ROut]): ZLayer[RIn with RIn2, E1, ROut2] =
+      ZLayer.requires[RIn2] ++ self >>> that
+
+    /**
+     * Feeds the output services of this layer into the input of the specified
+     * layer, resulting in a new layer with the inputs of this layer as well as
+     * any leftover inputs, and the outputs of the specified layer.
+     */
+    def >>>[E1 >: E, ROut2](that: ZLayer[ROut, E1, ROut2]): ZLayer[RIn, E1, ROut2] =
```
adamgfraser, line 2395: Can this version go on the trait or does it shadow the other definition then?

core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2382,4 +2377,34 @@ object ZLayer {
 
   private def andThen[A0, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21, B, C](f: (A0, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21) => B)(g: B => C): (A0, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21) => C =
     (a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21) => g(f(a0, a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16, a17, a18, a19, a20, a21))
+
+  implicit final class ZLayerVerticalComposeSyntax[-RIn, E, ROut](private val self: ZLayer[RIn, E, ROut]) extends AnyVal {
```
adamgfraser, line 2381: Do we need the contravariance here?

kitlangton: Ah, no. Forgot to delete that.

PR #4816, opened 2021-03-18
ZIO Test: Implement assertEqual

adamgfraser:
A dramatically simpler variant of `assert` that doesn't take an `Assertion` but just asserts that two values are equal. Uses our macro functionality to report the original code of the left hand side, so a sample output would look like this:

image

You don't get the super details highlighting and recursive structure of the existing assertions but you also don't have to remember any assertions.


jdegoes:
How about we go "all the way" and just have `assert` so you can do:

```scala
assert(expected == actual)
```

In Scala 3:

```scala
inline def assert(inline b: Boolean) = ${assertImpl('b)}
```

jdegoes:
This is the approach taken by [utest](https://github.com/com-lihaoyi/utest)

adamgfraser:
We definitely can. We potentially lose a little but of the concept of the "actual versus expected result" in that case. For example in:

```scala
val x = 2
assert(someComplicatedExpression(x) == 125)
```

utest would report that the test failed with `x` equal to `2` since it reports variables captured within the assertion. However, if the actual result of evaluating `someComplicatedExpression` is `126` that would not get reported. But I agree it is simpler.

jdegoes:
> utest would report that the test failed with x equal to 2 since it reports variables captured within the assertion. However, if the actual result of evaluating someComplicatedExpression is 126 that would not get reported. But I agree it is simpler.

I saw a Scala assert function that handled this case by evaluating the code to find the value, but I can't find it at the moment.

Basically, it split apart the expression into a tree and rendered every evaluated intermediate, something like:

```
assert(x.foo.bar == 3)
       |  |  |
       |  |  4
       |  |
       |  Baz@287283
       |
       Bar@23292
```

adamgfraser:
@jdegoes Yes I think we could do that. Not sure how it would scale if you had a significant logic in the assertions but assertions themselves do tend to be small.

adamgfraser:
Closing in favor of #5036.

PR #4819, opened 2021-03-20
Implement replicateParM

kitlangton:
We have `replicateM`, may as well have `replicateParM`?


adamgfraser:
The question is what happens when we want to do this with bounded parallelism. 😃

adamgfraser:
Actually there is precedent in the check variants in ZIO Test, though I don't know if that is great either.

kitlangton:
> The question is what happens when we want to do this with bounded parallelism. 😃

Hmmm. `.replicateParMinnesota`??? 😆

adamgfraser:
Yes I like that.

PR #4822, opened 2021-03-20
Add migrating from Monix guide

kitlangton:


docs/resources/migrating_from_monix.md
```diff
@@ -0,0 +1,93 @@
...
+| `sleep` | `sleep` |
+| `suspend` | `effectSuspend` |
+| `traverse` | `foreach` |
+| `unit` | `unit` |
+
+### Data Structures
+
+| Monix / Cats Effect | ZIO |
+|-------|-----|
+| `Deferred[Task, A]` | `Promise[Throwable, A]` |
```
adamgfraser, line 86: Maybe we should include the type parameters for the other data types as well?

joroKr21: + `TaskLocal` -> `FiberRef` I believe

adamgfraser: That's right.

adamgfraser (approved):

PR #4823, opened 2021-03-21
Add helpful implicit errors

kitlangton:
ZIO uses many implicit evidence constraints to implement methods on the ZIO trait that only work with particular subtypes. For instance, `ZIO#some`, which requires that the `A` type is optional, is defined as:

```scala
final def some[B](implicit ev: A <:< Option[B]): ZIO[R, Option[E], B] = ???
```

This is a very cool technique. The only problem is that the error message requires the user to understand how this technique works, particularly the meaning of the `<:<` symbol:

```scala
[error] /Users/kit/code/example.scala:17:25: Cannot prove that Int <:< Option[B].
[error]   val example = UIO(10).some
```

This could be clearer. Basically, by making our own versions, we can give more ZIO specific compile errors:

image

We could make this more specific for particularly difficult to use messages, potentially linking to documentation. Overall, this is basically an extension of the pattern used by the `CanFail` constraint.

### DIFF
--- a/core/shared/src/main/scala/zio/IsSubtype.scala
+++ b/core/shared/src/main/scala/zio/IsSubtype.scala
@@ -1,0 +1,27 @@
+package zio
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("\nOutput Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
+sealed abstract class HasOutput[-A, +B] extends (A => B)
+object HasOutput {
+  implicit def impl[A, B](implicit subtype: A <:< B): HasOutput[A, B] = new HasOutput[A, B] {
+    override def apply(a: A): B = subtype(a)
+  }
+
+  implicit def implNothing[B]: HasOutput[Nothing, B] = new HasOutput[Nothing, B] {
+    override def apply(a: Nothing): B = a
+  }
+}
+
+@implicitNotFound("\nError Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
+sealed abstract class HasError[-A, +B] extends (A => B)
+object HasError {
+  implicit def impl[A, B](implicit subtype: A <:< B): HasError[A, B] = new HasError[A, B] {
+    override def apply(a: A): B = subtype(a)
+  }
+
+  implicit def implNothing[B]: HasError[Nothing, B] = new HasError[Nothing, B] {
+    override def apply(a: Nothing): B = a
+  }
+}
--- a/core/shared/src/main/scala/zio/IsSubtype.scala
+++ b/core/shared/src/main/scala/zio/IsSubtype.scala
@@ -2,26 +2,26 @@ package zio
 
 import scala.annotation.implicitNotFound
 
-@implicitNotFound("\nOutput Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
-sealed abstract class HasOutput[-A, +B] extends (A => B)
-object HasOutput {
-  implicit def impl[A, B](implicit subtype: A <:< B): HasOutput[A, B] = new HasOutput[A, B] {
+@implicitNotFound("This operator requires that the output type be a subtype of ${B} but the actual type was ${A}.")
+sealed abstract class IsSubtypeOfOutput[-A, +B] extends (A => B) with Serializable
+object IsSubtypeOfOutput {
+  implicit def impl[A, B](implicit subtype: A <:< B): IsSubtypeOfOutput[A, B] = new IsSubtypeOfOutput[A, B] {
     override def apply(a: A): B = subtype(a)
   }
 
-  implicit def implNothing[B]: HasOutput[Nothing, B] = new HasOutput[Nothing, B] {
+  implicit def implNothing[B]: IsSubtypeOfOutput[Nothing, B] = new IsSubtypeOfOutput[Nothing, B] {
     override def apply(a: Nothing): B = a
   }
 }
 
-@implicitNotFound("\nError Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
-sealed abstract class HasError[-A, +B] extends (A => B)
-object HasError {
-  implicit def impl[A, B](implicit subtype: A <:< B): HasError[A, B] = new HasError[A, B] {
+@implicitNotFound("This operator requires that the error type be a subtype of ${B} but the actual type was ${A}.")
+sealed abstract class IsSubtypeOfError[-A, +B] extends (A => B) with Serializable
+object IsSubtypeOfError {
+  implicit def impl[A, B](implicit subtype: A <:< B): IsSubtypeOfError[A, B] = new IsSubtypeOfError[A, B] {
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/zio/IsSubtype.scala
```diff
@@ -0,0 +1,27 @@
+package zio
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("\nOutput Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
```
adamgfraser, line 5: What about something like:

`This operator requires that the output type be a subtype of Option[Int] but the actual type String was not a subtype of Option[Int]`

core/shared/src/main/scala/zio/IsSubtype.scala
```diff
@@ -0,0 +1,27 @@
...
+sealed abstract class HasOutput[-A, +B] extends (A => B)
+object HasOutput {
+  implicit def impl[A, B](implicit subtype: A <:< B): HasOutput[A, B] = new HasOutput[A, B] {
+    override def apply(a: A): B = subtype(a)
+  }
+
+  implicit def implNothing[B]: HasOutput[Nothing, B] = new HasOutput[Nothing, B] {
```
adamgfraser, line 12: Do you need these second instances?

kitlangton: I know right? Scala weirdness. I did! It couldn't just find one for `Nothing`, which means it didn't trigger the `CanFail` warning, but instead said gave the implicit error for `HasError`.

core/shared/src/main/scala/zio/IsSubtype.scala
```diff
@@ -0,0 +1,27 @@
+package zio
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("\nOutput Type Mismatch\n  expected: ${B}\n    actual: ${A}\n\n")
+sealed abstract class HasOutput[-A, +B] extends (A => B)
```
adamgfraser, line 6: I wonder if we want to avoid using the term `Has` here given its existing definition as another data type? Something like `IsSubtypeOutput` or `IsSubtypeOfOutput` might be more descriptive and reads relatively well in infix position. Like `implicit ev: A IsSubtypeOfOutput Option[B]`. We could simplify if we didn't care about distinguishing the channels but I think that is helpful to indicate to the user where something went wrong.

core/shared/src/main/scala/zio/IsSubtype.scala
kitlangton: I agree 👍 Wasn't sure about the alignment / terseness. How about just the follow, so as not to repeat the expected type?
`This operator requires that the output type be a subtype of Option[Int] but the actual type was String.`

adamgfraser: Yes.

core/shared/src/main/scala/zio/IsSubtype.scala
kitlangton: That crossed my mind—Has is accounted for. I like `implicit ev: A IsSubtypeOfOutput Option[B]` a lot.

kitlangton:
Updated @adamgfraser, but I left STM and ZStream alone for now, as I'm not sure how completely those files are going to change. Though I suppose most of ZIO will as well :P

adamgfraser:
@kitlangton Looks good! Yes I think we're going to have changes to everything to some extent so can't let that stop us. I think this is probably a good point to get feedback from others though and if we have alignment it will be easy to push to the streams and STM as well.

kitlangton:
Sounds good! I'll poke some people on Monday for feedback :)

iravid (approved):
Big like 👍

joroKr21:
Have you considered using a type alias? `@implicitNotFound` should work on type aliases

adamgfraser:
@joroKr21 That's interesting! I didn't realize you could do that! Unfortunately it doesn't seem to work on Scala 3.

```scala
import scala.annotation.implicitNotFound

@implicitNotFound("${A} is not a subtype of ${B}")
type IsSubtypeOf[-A, +B] = A <:< B

object Example {
  implicitly[String IsSubtypeOf Int]
}

// Cannot prove that String <:< Int.
```

Maybe the Scala 3 team can resolve that. It would certainly be nice to make the implementation of this lighter weight. I will open an issue.

adamgfraser:
Ah, I think this may be resolved by lampepfl/dotty#11823. I commented to confirm.

kitlangton:
> Have you considered using a type alias? `@implicitNotFound` should work on type aliases

Oh cool. Also had no idea that worked :P 
But yeah, it looks like they're leaning towards not supporting this in Scala 3. Maybe we could use zio-prelude's Subtype trick? However, as currently there are only two new types, I don't think its too terrifying a sum of boilerplate 😉.

adamgfraser:
I totally agree. Hopefully they will address in Scala 3 but otherwise this is already pretty minimal and any new type machinery is going to be more heavy weight than what you have here.

kitlangton:
Him, though it looks like there may be some other issues on dotty right now:
```scala
[info] compiling 106 Scala sources and 1 Java source to /home/circleci/project/core/jvm/target/scala-3.0.0-RC1/classes ...
[error] -- [E008] Not Found Error: /home/circleci/project/core/shared/src/main/scala/zio/ZIO.scala:934:18 
[error] 934 |    map(result => !result)
[error]     |                  ^^^^^^^
[error]     |                  value unary_! is not a member of A
[error] -- [E008] Not Found Error: /home/circleci/project/core/shared/src/main/scala/zio/ZIO.scala:942:13 
[error] 942 |      a => a.fold[ZIO[R, Option[E], Unit]](ZIO.succeedNow(()))(_ => ZIO.fail(None))
[error]     |           ^^^^^^
[error]     |           value fold is not a member of A
```
Not sure if that's because of my changes. I'll have to take a look.

joroKr21:
> But yeah, it looks like they're leaning towards not supporting this in Scala 3.

That's a shame. Having a different class means you can't convert back and forth. You are converting from `<:<` to your custom type but not the other way around so if I have your type I can't use a function that expects `<:<`.

kitlangton:
> That's a shame. Having a different class means you can't convert back and forth. You are converting from `<:<` to your custom type but not the other way around so if I have your type I can't use a function that expects `<:<`

Oh, what if I make these type extend `<:<`?

joroKr21:
> Oh, good point! What if I make these type extend <:<?

You could try, but that sounds like it would cause a circular dependency

kitlangton:
> > But yeah, it looks like they're leaning towards not supporting this in Scala 3.
> 
> That's a shame. Having a different class means you can't convert back and forth. You are converting from `<:<` to your custom type but not the other way around so if I have your type I can't use a function that expects `<:<`

I actually don't think this will be a problem, as this will only be used internally, and they can be derived from `<:<`, so as a user, you should never run into this.

adamgfraser (approved):
Let's get this merged!

Issue #4804, opened 2021-03-15
ZIO Macros: the error channel could be better controlled when we have to deal with non zio vals/defs


octavz:
Please see:

https://scastie.scala-lang.org/AAHvt5uRQqC4kdygcEPNQQ

@ioleo  as discussed on Discord:

**ioleo**: _regular def and vals (as opposed to ZIO effects) can potentially hold side-effects, so current behaviour is assuming no discipline from the programmer (but then forces orDie to ignore the error channel when using accessor)_
**ioleo**: _now perhaps that is too defensive, perhaps we should assume discipline from the programmer, or at least give a tool to control macro behavior.

Thank You

PR #4825, opened 2021-03-21
Zio macros no throw

octavz:
Resolves  https://github.com/zio/zio/issues/4804


macros/shared/src/main/scala-2.x/zio/macros/noThrow.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.macros
+
+import scala.annotation.{StaticAnnotation, compileTimeOnly}
+
+@compileTimeOnly("marks that a field or method will not throw")
+class noThrow extends StaticAnnotation
```
adamgfraser, line 22: `throwsNothing`?

octavz: done

adamgfraser (approved):
Looks good to me!

You could update the internal names to match the new name of the annotation.

One question is with this change should the default behavior be that we assume these operators throw and you annotate to say they don't or we assume they don't throw and you annotate to say they do?

Right now we are being very defensive by assuming they throw but I think we are "wrong" most of the time because best practice would be all operators that can throw return effects that model those failures. So I would be inclined to change this but open to thoughts from others.

@octavz What do you think?

@ioleo Any comments from your end?

octavz:
> Looks good to me!
> 
> You could update the internal names to match the new name of the annotation.
> 
> One question is with this change should the default behavior be that we assume these operators throw and you annotate to say they don't or we assume they don't throw and you annotate to say they do?
> 
> Right now we are being very defensive by assuming they throw but I think we are "wrong" most of the time because best practice would be all operators that can throw return effects that model those failures. So I would be inclined to change this but open to thoughts from others.
> 

I explained on Discord (#zio-macros) as well that the defensive behavior is not according to ZIO best practices, is for bad code, but I am happy with any of these since I this way I won't have to do `.orDie` which is quite bad practice in production code.

adamgfraser:
I think setting the error type to `Nothing` when the operator doesn't return an effect makes a lot of sense. No error type is specified so by definition any error that occurs is unanticipated which is a perfect use for the `Die` channel.

octavz:
Let me know if you want me to change I'll update my PR accordingly. Thanks.

adamgfraser:
@ioleo What do you think?

I understand wanting to manage possible side effects from operators that aren't suspended? But isn't this what the `Die` channel is for? As long as we suspend them in the code we generate so that thrown exceptions are still translated into fiber failures I think we are safely managing them.

This should make it a lot more ergonomic for people who are doing the "right thing" where they will get the desired signature and not have to add an annotation.

ioleo (approved):
@octavz Thanks for contibuting. I like this approach. It makes it explicit that the implementation should not throw.

adamgfraser:
@ioleo What do you think about just always typing these as `UIO`? This seems like a good use for the `Die` channel where the user hasn't indicated than an effect can fail but we still want to manage errors that do occur.

ioleo:
@adamgfraser I'm fine with both. The `UIO` way assumes some discipline though, so if we go that way we should add some section to macro docs.

adamgfraser:
Okay, I think that is the better way to go.

@octavz Would you like to merge as is and we can open an issue for that or would you like to revise this PR along those lines?

octavz:
I will redo the PR, and use the `@throws` or `@throwing` if is impure code. I will close this and make another one next days.

PR #4826, opened 2021-03-21
Unify Ref and RefM

adamgfraser:
Unifies `Ref` and `RefM`. `RefM` becomes a subtype of `Ref` that has additional capabilities (i.e. the ability to perform updates within the modify operation) at some cost to performance. This allows a `RefM` to be used anywhere a `Ref` is currently being used. In particular, this allows the `RefM` contained in a `SubscriptionRef` to be a drop in replacement for an ordinary `Ref`.


jdegoes:
@adamgfraser What do you mean, 'at some cost to performance'? From what I can see, Atomic is used as the default and there are no synchronized operations in `Ref`, only `RefM`.

This is a nice change. If there is no cost to performance for `Ref`, then it would be nice to unify the two.

adamgfraser:
Using a 'ZRefM' instead of a 'ZRef' comes at "some coat to performance" because we have to synchronize to support performing effects within the update operation. There is no cost to performance if you are using an ordinary 'ZRef'.

jdegoes (approved):

jdegoes:
@adamgfraser Looks great!

Issue #4519, opened 2020-12-23
ZIO Test: Share layers across specs


jdegoes:
Currently layers can only be shared within a spec. There is no ability to share layers across specs. To support this, a spec should be able to have a public ZLayer that indicates its dependencies. Then the SBT spec runner needs to be able to detect duplicate layers (i.e. layers with the same identity) and share them across executions of the specs.

carlos-verdes:
I think this issue can be close right?

PR #4827, opened 2021-03-21
(wip) Share layers across Specs

zagyi:
This PR addresses #4519.


zagyi:
The first commit is a POC that defines `CustomRunnableSpec` that gets a custom layer in its constructor. 
The test framework will provide that layer to the spec, when executing it and will make sure that other `CustomRunnableSpec`s that use the same custom layer (in terms of reference equality) will access the same shared instance without re-initialization.

Currently there is only a half-baked test that can be used to verify this functionality. There are two specs that use the same simple layer wrapping a `Ref[Int]`. Both increment it 4 times, so the expectation is to have `8` in the `Ref` at the end of the execution:
```
testOnly zio.test.ShareLayersAcrossSpecsSpec1 zio.test.ShareLayersAcrossSpecsSpec2
```

The actual assertions in these specs will always fail, so it only serves as a feedback on the console.

adamgfraser:
Awesome! Why don't we move the shared layer to `AbstractRunnableSpec` and then in `DefaultRunnableSpec` we can have a new abstract method `sharedLayer` with a default implementation `none: ZLayer[Any, Nothing, Any]`. The users can just override that to create a shared layer versus extending a different kind of spec.

hmemcpy:
This is great, thank you very much! Immediate use case for this - shared testcontainer instances across suites!

zagyi:
Hi @adamgfraser,

finally I had some time to make progress on this PR.
Just as a reminder, last time when we discussed things, there was still a lot of test failures. The problem turned out to be that my modified `TestExecutor` didn’t provide the `TestEnvironment` to `Spec`s, when running them. Instead, it delegated the provision of the entire environment to external code by simply returning `URIO[R, E]` from its `run` method. The end result was that `TestEnvironment` was shared by all tests within a spec, which is obviously wrong.

I followed the long tradition of fixing things by introducing a new type parameter. :) Let me refer to the [docs of the current `TestExecutor`](https://github.com/zagyi/zio/blob/feature/custom_runnable_spec/test/shared/src/main/scala/zio/test/TestExecutor.scala#L21-L31), because that’s the essence of my recent changes.

Carrying out this change throughout the test framework has fixed the failures, so at least we have a baseline. (For JVM/sbt, other platforms/test runners are not yet covered.) However, I’m not sure if you like the resulting structure. Also, I haven’t succeeded yet with the unification of `DefaultRunnableSpec` and the new `CustomRunnableSpec`.

Could you please take a look at my branch and give some feedback if my approach is valid?

adamgfraser:
@zagyi What do we need to do to get this to a point where it can be merged? Will be easier for other people to help make further improvements once we have it in.

jdegoes:
@zagyi Please re-open when you get a chance to update with the new compositional specs and series/2.x branch!

PR #4830, opened 2021-03-22
Zio to option operator

politrons:
Not sure if this it will fit as part of the ZIO core as new operator. 

We ave several situation where we have to pass to some legacy system some Option types where in order to obtain that value we have to ensure there's no side-effects, so we end up doing this.

```
  for {
    maybeValue <- ZIO.effect(Option(args(8))).catchAll(_ => ZIO.succeed(None))
    _ <- ZIO.effect(legacySystem(maybeValue))
} yield ()

```

instead it would be nice just to this

```
  for {
    maybeValue <-  ZIO.toOption(args(8))
    _ <- ZIO.effect(legacySystem(maybeValue))
} yield ()

```

And forget about side-effect. This operator it would take care of that.

If there's a better approach it would be great to know. 

Regards


adamgfraser:
@politrons This can be implemented as `ZIO.effect(args(8)).option`.

politrons:
Wow I completely miss that, thanks @adamgfraser

adamgfraser:
@politrons No problem! Yes we have a lot of operators which is great because most of the time we do have one for your use case but It can make them more challenging to find.

PR #4834, opened 2021-03-23
Supporting Chunking in ZStream.fromIterator

adamgfraser:
Sometimes we want to pull values from an `Iterator` in chunks for greater efficiency.


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3616,33 +3616,33 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
-          } else throw StreamEnd
-        }.mapError {
-          case StreamEnd => None
-          case e         => Some(e)
-        }
-      }
+  /**
+   * Creates a stream from an iterator that may throw exceptions.
+   */
+  def fromIterator[A](iterator: => Iterator[A], maxChunkSize: Int = 1): ZStream[Any, Throwable, A] =
```
svroonland, line 3622: Just a heads up that adding a default parameter breaks backwards compatibility, like [here](https://github.com/zio/zio/pull/4205).

adamgfraser: Yes. I think since we are not guaranteeing binary compatibility at this point that is okay. I think this is the right API and it should be a very easy change, but open to other perspectives.

adamgfraser:
@iravid Can I get a review on this?

streams/shared/src/main/scala/zio/stream/ZStream.scala
jdegoes: You could also add another method with 2 parameters, and the 1 parameter one can delegate to this one.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3616,33 +3616,33 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+    ZStream {
+      ZManaged
+        .effect(iterator)
+        .fold(
+          Pull.fail,
+          iterator =>
+            ZIO.effect {
```
jdegoes, line 3629: Since you know the chunk size, it's probably more efficient to pre-allocate an array, fill it, convert to `Chunk`, and `take` the number of elements actually read (which is `<=` the chunk size).

iravid: Hinting the chunk builder with the desired size will have the same effect and avoid dealing with the array directly

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3659,16 +3659,25 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
-          if (it.hasNext)
-            Pull.emit(it.next())
-          else
-            Pull.end
+      Managed.effectTotal(iterator).map { iterator =>
+        ZIO.effectTotal {
+          if (maxChunkSize <= 1) {
+            if (iterator.isEmpty) Pull.end else Pull.emit(iterator.next())
+          } else {
+            val builder = ChunkBuilder.make[A]()
```
iravid, line 3671:
```suggestion
            val builder = ChunkBuilder.make[A](maxChunkSize)
```

iravid (approved):
Looks good with the sizeHint changes!

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: So the one watch out here is if the user uses an extremely large `maxChunkSize`. For example, if I say "I'm not sure how many elements I have, but just read them all as one Chunk" and do `fromIterator(iterator, Int.MaxValue)` I am going to blow up with an `OutOfMemoryError` because it is going to try to actually allocate an `Array` with `Int.MaxValue` size.

iravid: My POV on this is that if it says "chunk" on the label, it's eager and must fit in memory. So I kinda think it's a fair tradeoff.

adamgfraser: I think that is right but just to push on this that means that `ZStream.fromIterator(Iterator(1, 2, 3, 4, 5), Int.MaxValue)` will blow up. I think definitely if the `Iterator` is actually too large to fit in memory it has to fail, but failing when the `Iterator` is small is a little counterintuitive.

jdegoes: True. Maybe we could call it a `chunkHint` and that way we would be free to limit it to some reasonable value (our default chunk size, e.g.).

jdegoes (approved):

deusaquilus:
Wow, you guys really care about getting the naming right. I’m impressed!

PR #4836, opened 2021-03-23
add alias for CompletableFuture

sken77:
`CompletionStage` is a bit confusing for scala users with no java background. a lot of people that deal with java libraries dont have much experience in java. `CompletableFuture` matches the most comon implementation of `CompletitionStage` that is used in java libraries that scala users might use. so matching the type with the call is helpful for them


adamgfraser (approved):
Thank you! 🙏

PR #4837, opened 2021-03-23
Annotation for throwing

octavz:
Resolves https://github.com/zio/zio/issues/4804


adamgfraser (dismissed):
Looks great! Good to go when compilation issues are resolved.

octavz:
I will get back to this these days, it seems is more time consuming than I thought since apparently there is a bug causing that warning  in 2.11.
I have to find another way to avoid the warning.

octavz:
@ioleo @adamgfraser 

It appears that on scala 211 with java8 it will throw an warning if it finds an annotation on a `val` or `lazy val`. 

Now, the `lazy val` worths the treatement because we can still suspend the error however for `vals` it makes no sense, since the runtime will crash anyway. In order to avoid the warning and keep things simple I would suggest to have `@throwing` annotations only for `defs` and not `vals` lazy or otherwise, and the values by default will have `Nothing` on the error channel. 

This still adds values since: 
  1. It fixes my problem of not needing `.orDie` 
  2. for `def` one can still mark it as `@throwing` getting the "proper ZIO behavior"

I don't have the resources to go more into macros and fix that warning for scala 211. I suppose that I have to expand that annotation and make it `compileTimeonly` in order for the compiler to erase it, but I don't know how to do it nor if it's worth the effort IMO. 

Please let me know what you think.

Thanks

adamgfraser:
@octavz I think your proposal makes sense.

octavz:
I removed the `@throwing` annotation support for `vals`, and things seem to be in order, if the build passes please have another look on it and we can merge it.

Thank you.

docs/howto/howto_macros.md
```diff
@@ -53,6 +53,46 @@ object AccountObserver {
...
+If the code is throwing exceptions see `@throwing` below.
+
+```scala
+import zio.{ Has, ZIO }
+import zio.macros.accessible
+
+@accessible
+object Module {
+  trait Service {
+    def pureMethog(v: Something): SomethingElse
```
regiskuckaertz, line 66:
```suggestion
    def pureMethod(v: Something): SomethingElse
```

adamgfraser:
@octavz Can you resolve conflicts?

octavz:
@adamgfraser it is done

adamgfraser (approved):

PR #4838, opened 2021-03-24
Support adding aspects to top level suite in MutableRunnableSpec

vladimirkl:
This PR adds support for adding aspects to top level suite in `MutableRunnableSpec`, which is always created automatically.
Alternate approach could be supplying aspects via additional constructor parameter, but this prevents us for using same object methods in aspects like `before`, `after` or others.  And  `topSuite @ aspect1 @ aspect2...`  reads better.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/MutableRunnableSpecSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/MutableRunnableSpecSpec.scala
@@ -1,0 +1,28 @@
+package zio.test
+
+import zio.test.Assertion._
+import zio.test.TestAspect._
+import zio.test.environment.TestEnvironment
+import zio.{Ref, ZIO}
+
+object MutableRunnableSpecSpec
+    extends MutableRunnableSpec(
+      TestEnvironment.any ++ Ref.make(0).toLayer,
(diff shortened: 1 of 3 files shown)


fokot:
Interesting approach.
Also another approach would be to add it after the Spec
```
object MutableRunnableSpecSpec extends MutableRunnableSpec(Ref.make(0).toLayer) {
  ...
} @@ sequential @@ before(ZIO.service[Ref[Int]].flatMap(_.update(_ + 1)))
```
`topSuite` looks better to me than this or aspects in constructor.

jdegoes:
I think the most discoverable option might be a constructor parameter, e.g.:

```scala
... extends MutableRunnableSpec(myLayer, myAspects) { 
}
```

Ping @adamgfraser

vladimirkl:
Constructor parameters have limitation - cannot use methods  from same spec object. They are useful for complex aspects like `before`  or `after`

adamgfraser:
I'm inclined to think the constructor argument is the way to go as well. Seems more discoverable and you could always factor out the definition of the aspects and any common functionality you wanted to reference into another object.

adamgfraser (approved):
Looks good to me!

adamgfraser:
Can you merge upstream changes so the new CI triggers?

vladimirkl:
@adamgfraser Done

adamgfraser (approved):

PR #4845, opened 2021-03-26
Add Type Alias For Intersection Types On Scala 2

### LINKED ISSUE #4844: with is too long
I suggest defining a type alias `&`, as in `Console & Database`. Works great:

```scala
import zio.console._
import zio.clock._
import zio.random._
import zio.{ App, ExitCode, URIO, ZIO }

object HelloWorld extends App {
  type &[A, B] = A with B

  override def run(args: List[String]): URIO[Console & Random & Clock, ExitCode] =
    myAppLogic.exitCode

  val myAppLogic: ZIO[Console & Random & Clock, IOException, Unit] =
    for {
      _    <- putStrLn("Hello! What is your name?")
      name <- getStrLn
      time <- currentDateTime.orDie
      _    <- putStrLn(s"Hello, $name, $time welcome to ZIO!")
    } yield ()
}
```

adamgfraser:
Resolves #4844. Adds a type alias to allow `&` to be used on both Scala 2 and Scala 3.

### DIFF
--- a/core/shared/src/main/scala-2.11/zio/IntersectionTypeCompat.scala
+++ b/core/shared/src/main/scala-2.11/zio/IntersectionTypeCompat.scala
@@ -16,6 +16,6 @@
 
  package zio
 
- private[zio] trait VersionSpecific {
+ private[zio] trait IntersectionTypeCompat {
   type &[+A, +B] = A with B
  }
--- a/core/shared/src/main/scala-2.11/zio/IntersectionTypeCompat.scala
+++ b/core/shared/src/main/scala-2.11/zio/IntersectionTypeCompat.scala
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2020-2021 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ package zio
+
+ private[zio] trait VersionSpecific {
+  type &[+A, +B] = A with B
+ }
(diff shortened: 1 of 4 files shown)


regiskuckaertz:
Bloody hell that was fast 🕺 I refrained from going one step further because of the dotty compat but this would be nicer:
```scala
type &[A, B] = Has[A] & Has[B]
```
What do you think?

I thought about this after @kitlangton 's presentation at zio world. I think it's a great idea to get rid of the type alias boilerplate but `Has[_]` is 5 characters long and life is too short 😄

jdegoes (approved):

kitlangton:
> Bloody hell that was fast 🕺 I refrained from going one step further because of the dotty compat but this would be nicer:
> 
> ```scala
> type &[A, B] = Has[A] with Has[B]
> ```
> 
> What do you think?
> 
> I thought about this after @kitlangton 's presentation at zio world. I think it's a great idea to get rid of the type alias boilerplate but `Has[_]` is 5 characters long and life is too short 😄

The following works in Dotty, but there's not much we can do in Scala 2.

```scala
type ++[A,B] <: Has[_] = (A,B) match {
  case (Has[_], Has[_]) => A & B & Has[_]
  case (_, Has[_]) => Has[A] & B
  case (Has[_], _) => A & Has[B]
  case (_,_) => Has[A] & Has[B] 
}

val layer = ZLayer.fromAuto[Console ++ String ++ Has[Int] ++ Boolean](layer, boolLayer, Console.live) 
```

adamgfraser:
@regiskuckaertz I think one of our lessons from ZIO 1.0 has been that hiding the `Has` data type doesn't pay off in the long term. It seems really nice to avoid writing `Has` everywhere, which is why we created the type aliases for the services. However, it creates user confusion about which date types are `Has` and which ones are not and leads to errors when users accidentally created nested `Has` data types so I think it is best to stick to the simple definition of intersection types that is the same across Scala versions.

adamgfraser:
That is really interesting regarding the match types.

kitlangton:
While I agree it could be nice to have something like:
```scala
type ++[A,B] = Has[A] with Has[B]
```

The problem is when you use it to join more than two types, the left hand side is now a Has and will become doubly wrapped, so you'd actually need:

```scala
type ++[A,B] = A with Has[B]
```

But then it's a bit weird to use, as you have to explicitly wrap the first type—or add some second operator that also wraps the lhs. Either way, not too great. Probably best for individuals to implement it themselves in their own libraries if they really want that, until we can figure out something more elegant (like that match type approach).

regiskuckaertz:
> ```scala
> type ++[A,B] <: Has[_] = (A,B) match {
>   case (Has[_], Has[_]) => A & B & Has[_]
>   case (_, Has[_]) => Has[A] & B
>   case (Has[_], _) => A & Has[B]
>   case (_,_) => Has[A] & Has[B] 
> }
> ```

@kitlangton hmm, I see, at that point `++` is not a type but a type constructor, like `*:` for inductive tuples. Then the problem disappears.

@adamgfraser good point, I think this is wise.

regiskuckaertz (approved):

PR #4869, opened 2021-04-02
Require mdoc for mergify

mijicd:
@khajavi I've added the mdoc to mergify and set it as required check.


khajavi:
@mijicd 
Thank you.

Issue #4886, opened 2021-04-06
Channels: follow-up work


iravid:
After #4885 is merged, we have the following open issues that need to be resolved before it can replace the implementation in `zio.stream`:
* Channels
	* Decide what to do with the Bridge node
	* Add a suspend operator so we can implement combinators without `Ref`
	* Implement n-ary merge
* Sinks
	* Improve inference problems with InErr, In
	* Combinators
		* collectAllWhileWith
		* foldM - preserve leftovers on failure
		* raceBoth
		* timed
		* repeat
		* zipWithPar
		* untilOutputM
	* Constructors
		* managed
	* Port from ZTransducer:
		* filterInput/M
		* branchAfter
		* collectAllN
		* collectAllToMapN
		* collectAllToSetN
		* dropWhile/M
		* iso_8859_1Decode
		* splitLines
		* splitOn
		* splitOnChunk
		* utfDecode
		* utf8Decode
		* utf16Decode
		* utf16BEDecode
		* utf16LEDecode
		* utf32Decode
		* utf32BEDecode
		* utf32LEDecode
		* usASCIIDecode
		* deflate
		* inflate
		* gzip
		* gunzip
* ZStream
	* Combinators
		* Rewrite broadcast* distributed* and groupBy on top of ZHub (where possible)
		* bufferDropping
		* bufferSliding
		* combine without pull
		* combineChunks without pull
		* drop swallows errors
		* dropWhile swallows errors
		* decide what to do with ensuringFirst
		* use sinks in all the run* methods and do not implement functionality on ZStream
		* Implement flatMapParSwitch somehow (maybe push this functionality into Channel’s n-ary merge)
		* haltWhen
		* grouped
		* groupedWithin
		* interleaveWith
		* intersperse
		* mapAccumM (use loopOnPartial)
		* mapError (through channel)
		* mapErrorCause (through channel)
		* mapMPar (probably through channel)
		* mergeWith (through channel)
		* peel without Pull
		* refineOrDie
		* refineOrDieWith
		* repeatElementsWith
		* repeatWith
		* runLast
		* runCount
		* runSum
		* scanReduceM (loopOnPartial)
		* scheduleWith
		* takeRight
		* throttleEnforceM (package functionality into a sink)
		* throttleShapeM (package functionality into a sink)
		* debounce (sink)
		* toInputStream
		* toIterator
		* toReader
		* zipWith and zipAllWithExec without pull, with Adam’s specialized zip RingBuffer
		* zipWithLatest
[...]

PR #4901, opened 2021-04-07
Implement mapError, mapErrorCause, mapAccumM, scanReduceM

paulpdaniels:
Part of #4886

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
@@ -457,6 +457,12 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
     catchAllCause((cause: Cause[OutErr]) => cause.failureOrCause.fold(err => ZChannel.fail(f(err)), ZChannel.halt(_)))
 
   /**
+   * A more powerful version of [[mapError]] which also surfaces the [[Cause]] of the channel failure
+   */
+  final def mapErrorCause[OutErr2](f: Cause[OutErr] => Cause[OutErr2]): ZChannel[Env, InErr, InElem, InDone, OutErr2, OutElem, OutDone] =
+    catchAllCause((cause: Cause[OutErr]) => ZChannel.halt(f(cause)))
+
+  /**
    * Returns a new channel, which is the same as this one, except the terminal value of the
    * returned channel is created by applying the specified effectful function to the terminal value
    * of this channel.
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
```diff
@@ -456,6 +456,12 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
   final def mapError[OutErr2](f: OutErr => OutErr2): ZChannel[Env, InErr, InElem, InDone, OutErr2, OutElem, OutDone] =
     catchAllCause((cause: Cause[OutErr]) => cause.failureOrCause.fold(err => ZChannel.fail(f(err)), ZChannel.halt(_)))
```
iravid, line 457: Let's delegate to mapErrorCause while you're here :-)

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1417,7 +1417,17 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
    * new elements.
    */
   final def mapAccumM[R1 <: R, E1 >: E, S, A1](s: S)(f: (S, A) => ZIO[R1, E1, (S, A1)]): ZStream[R1, E1, A1] =
-    ???
+    ZStream.unwrap(
+      for {
+        state <- Ref.make(s)
```
iravid, line 1422: We can now implement combinators using recursion on channels to avoid the Ref overhead. Check out how the other combinators are implemented. Let me know if you need a pointer and I will write one up when I'm on my computer.

iravid (commented):
Great start 🚀

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
paulpdaniels: Hmm could you point to the one you mean? I tried refactoring based on the `mapAccum` above it, but hit a snag because it loses laziness to naively change to `ZChannel.fromEffect(chunk.mapAccumM(s)(f))`. Do you think I should still try to incorporate the `loopPartial...`?

streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
paulpdaniels: Done

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1887,7 +1912,10 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
    * See also [[ZStream#scanM]].
    */
   def scanReduceM[R1 <: R, E1 >: E, A1 >: A](f: (A1, A) => ZIO[R1, E1, A1]): ZStream[R1, E1, A1] =
-    ???
+    mapAccumM[R1, E1, Option[A1], A1](Option.empty[A1]) {
+      case (Some(a1), a) => f(a1, a).map(a2 => Some(a2) -> a2)
+      case (None, a) => ZIO.succeedNow(Some(a) -> a)
+    }
```
paulpdaniels, line 1918: Added `scanReduceM` because it could be implemented in terms of `mapAccumM`

iravid (approved):
Awesome, thanks! It's a bit unfortunate we had to inline loopOnChunks but we can enhance it later with state tracking.

iravid:
Merging despite failures which are unrelated; will fix them in another PR.

PR #4908, opened 2021-04-07
Compile-time ZLayer construction

kitlangton:
A few more things need to happen here, changes I've made to `zio-magic` (thanks to reports from @hmemcpy's production usage). But it'd be great to get some overall feedback.


hmemcpy:
Many thanks for this!

Did you have a solution for providing additional layers that are not strictly required?

kitlangton:
> Many thanks for this!
> 
> Did you have a solution for providing additional layers that are not strictly required?

Not quite yet. Going to work on that next 😄

adamgfraser (approved):
We can play with naming a little but let's get this in once conflicts are resolved.

adamgfraser (approved):

adamgfraser (approved):

PR #4911, opened 2021-04-08
Renovate: rebase PR only on conflict

sideeffffect:
https://docs.renovatebot.com/configuration-options/#rebasewhen


sideeffffect:
Btw there's no point waiting for the CI to finish -- it just changes Renovate config, which isn't being checked in any way.

PR #4916, opened 2021-04-09
Reimplement SubscriptionRef in Terms of ZHub

adamgfraser:
One issue with the current implementation of `SubscriptionRef` is that since it is backed by a `Queue`, if more than one subscriber is observing changes to the `Ref` each subscriber will miss changes, since each time a value is taken from the queue by one subscriber it is no longer available to other subscribers. We can address this by backing `SubscriptionRef` with a `Hub` . This way each subscriber can receive the current value of the `Ref` and all changes after they have subscribed. 

This allows `SubscriptionRef ` to function much like a `Var` in functional reactive programming frameworks, where the `SubscriptionRef` represents some state that can be changes and multiple processes can subscribe to take actions in response to any changes in that state.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZRefMSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZRefMSpec.scala
@@ -1,5 +1,6 @@
 package zio
 
+import com.github.ghik.silencer.silent
 import zio.test.Assertion._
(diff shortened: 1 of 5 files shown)


iravid (approved):
This is awesome.

PR #4919, opened 2021-04-09
Implement ZStream#repeatElementsWith

paulpdaniels:
Related to #4886

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
@@ -2377,60 +2377,60 @@ object ZStreamSpec extends ZIOBaseSpec {
             )(equalTo(Chunk(Right("A"), Right("B"), Right("C"), Left("!"))))
           )
         ),
-        //     suite("repeatElements")(
-        //       testM("repeatElementsWith")(
-        //         assertM(
-        //           ZStream("A", "B", "C")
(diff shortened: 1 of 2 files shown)


streams-tests/shared/src/test/scala/zio/stream/experimental/ZStreamSpec.scala
```diff
@@ -2367,60 +2367,60 @@ object ZStreamSpec extends ZIOBaseSpec {
...
+               testM("short circuits after schedule")(
+                 assertM(
+                   ZStream("A", "B", "C")
+                     .repeatElements(Schedule.once)
+                     .take(3)
+                     .runCollect
+                 )(equalTo(Chunk("A", "A", "B")))
+               )
+             ),
+             testM("some") {
```
iravid, line 2412: Something looks off in the suite structure here?

paulpdaniels: I'm guessing this is because these individual operators only have a single test, so they aren't being grouped under a suite. I just uncommented them because they appear to have been implemented already. I checked and the structure is the same in the previous version as well.

iravid:
@paulpdaniels can you rebase on series/2.x?

paulpdaniels:
@iravid rebased, but looks like lint is and one of the tests is 💥

iravid (approved):
Super elegant implementation @paulpdaniels!

PR #4923, opened 2021-04-09
Implement ZSink.collectAllN

mijicd:
Part of #4886

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
@@ -26,27 +26,21 @@ object ZSinkSpec extends ZIOBaseSpec {
               .fromChunk(Chunk(1, 2, 3, 4))
               .transduce(ZSink.collectAllN[Nothing, Int](3))
               .runCollect
-              .map { chunks =>
-                assert(chunks)(equalTo(Chunk(Chunk(1, 2, 3), Chunk(4))))
-              }
+              .map(assert(_)(equalTo(Chunk(Chunk(1, 2, 3), Chunk(4)))))
           },
           testM("produces empty trailing chunks") {
             ZStream
               .fromChunk(Chunk(1, 2, 3, 4))
               .transduce(ZSink.collectAllN[Nothing, Int](4))
               .runCollect
-              .map { chunks =>
-                assert(chunks)(equalTo(Chunk(Chunk(1, 2, 3, 4), Chunk())))
-              }
+              .map(assert(_)(equalTo(Chunk(Chunk(1, 2, 3, 4), Chunk()))))
           },
           testM("handles empty input") {
             ZStream
               .fromChunk(Chunk.empty: Chunk[Int])
               .transduce(ZSink.collectAllN[Nothing, Int](3))
               .runCollect
-              .map { chunks =>
-                assert(chunks)(equalTo(Chunk(Chunk())))
-              }
+              .map(assert(_)(equalTo(Chunk(Chunk()))))
           }
         ),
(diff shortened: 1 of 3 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -346,7 +346,7 @@ object ZSink {
    * A sink that collects all of its inputs into chunks of maximum size `n`.
    */
   def collectAllN[Err, In](n: Int): ZSink[Any, Err, In, Err, In, Chunk[In]] =
-    foldUntil(Chunk.empty: Chunk[In], n.toLong)(_ appended _)
+    foldUntil(Chunk.empty: Chunk[In], n.toLong)(_ :+ _)
```
adamgfraser, line 349: Should we be using a `ChunkBuilder` here?

mijicd: Would be neat, but it's one extra step to materialize it.

iravid: Ah a ChunkBuilder is a good point. We can also hint it with the size so should be significantly more efficient.

mijicd: Tackled in 0427c094c5af0a5bbe66f122582c65363bede6e3.

mijicd: Here's the updated version: 375647a70a2149f6950e378174a04d390e47fdd9

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -342,6 +342,14 @@ object ZSink {
     new ZSink(loop(Chunk.empty))
   }
 
+  /**
+   * A sink that collects all of its inputs into chunks of maximum size `n`.
```
iravid, line 346: Can you also add a note that every chunk is preallocated and must fit in memory? Or @adamgfraser should we use the same approach you arrived at with regards to the size hint? (Which I don't remember what it was 😅)

streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
```diff
@@ -20,6 +20,14 @@ object ZSinkSpec extends ZIOBaseSpec {
   def spec: ZSpec[Environment, Failure] = {
     suite("ZSinkSpec")(
       suite("Constructors")(
+        suite("collectAllN")(
```
iravid, line 23: Same re tests from ZTransducerSpec :-)

mijicd: Done in 239c55418cd31a51a0416d96f8037cda6119f4b7. As discussed over chat, we have to leave with trailing chunks, therefore the nonapplicable tests are dropped.

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
adamgfraser: I think I've come to the point of view that we should just hint the specified size if an operator allows specifying a particular chunk size. I think we have to do that for maximum efficiency and it is just user error if you specify a chunk size like `Int.MaxValue`.

I was a little hesitant on this because of an issue we had with someone blowing themselves up using `queue.takeUpTo(Int.MaxValue)` but I think that is different in that there the `n` is how many elements to take and chunking is an implementation detail, whereas here the whole point of the parameter is to specify the size of the chunk to take.

PR #4928, opened 2021-04-09
fix typeCheck macro expansion should use fully qualified names

### LINKED ISSUE #4927: zio-test: `typeCheck` macro expansion should use fully qualified names to avoid conflicts
When implementing a test for zio-sql using `typeCheck(String)` I encountered an unexpected compile error. After some digging into the issue I found the cause and a workaround.

The `typeCheck` macro expansion in this example should compile. However it's not for `postgresql/FunctionDefSpec.scala` in zio-sql since there is already a `Left` and `Right` in scope.
```scala
    testM("cannot compile a select without from clause if a table source is required") {
      //the following execute only compiles with 'from customers' clause
      execute((select(CharLength(Customers.fName)) from customers).to[Int, Int](identity))
      val result = typeCheck("execute((select(CharLength(Customers.fName))).to[Int, Int](identity))")
      assertM(result)(isLeft)
    }
// https://github.com/TobiasPfeifer/zio-sql/commit/784fdd47e438ce38541b3fcd2168260c0f7d743b#r49148360
```

The macro expansion `zio.UIO.succeed(Right(()))` (or Left variant) will pick up the `Right` in scope which is not `scala.util.Right` and therefore fails to compile. The compile error is very misleading in this case. The `typeCheck` macro expansion should use fully qualified names to avoid such conflicts.

A workaround is to import `Right` and `Left` from `scala.util` before calling `typeCheck` (which is quite annoying with dead code detection):

```scala
    testM("cannot compile a select without from clause if a table source is required") {
      //the following execute only compiles with 'from customers' clause
      execute((select(CharLength(Customers.fName)) from customers).to[Int, Int](identity))

      // imports for Left and Right are necessary to make the typeCheck macro expansion compile
      import scala.util.Right
      import scala.util.Left
      val dummyUsage = zio.ZIO.succeed((Left(()), Right(())))


      val result = typeCheck("execute((select(CharLength(Customers.fName))).to[Int, Int](identity))")
      assertM(dummyUsage *> result)(isLeft)
    }
```

TobiasPfeifer:
fixes #4927'

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/CompileSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/CompileSpec.scala
@@ -12,6 +12,14 @@ object CompileSpec extends ZIOBaseSpec {
       val expected = "value ++ is not a member of Int"
       if (TestVersion.isScala2) assertM(typeCheck("1 ++ 1"))(isLeft(equalTo(expected)))
       else assertM(typeCheck("1 ++ 1"))(isLeft(anything))
+    },
+    testM("typeCheck expansion must use Left and Right from scala.util") {
+      // if this does not compile the macro expansion probably picks up the wrong Left or Right
+      case class Left(s: Nothing)
+      case class Right(s: Nothing)
+      val assertRight = assertM(typeCheck("1 + 1"))(isRight(anything))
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):
Thank you!

PR #4931, opened 2021-04-09
Implement ZSink.collectAllToSetN

mijicd:
Part of #4886


streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
```diff
@@ -26,6 +26,14 @@ object ZSinkSpec extends ZIOBaseSpec {
               .run(ZSink.collectAllToSet[Nothing, Int])
           )(equalTo(Set(1, 2, 3, 4)))
         ),
+        suite("collectAllToSetN")(
```
iravid, line 29: Could you also port the tests from `ZTransducerSpec`?

mijicd: They were already there, I adjusted their structure in 79553d91d7669b8aacbec9ebd302309bcef97073

PR #4934, opened 2021-04-10
introduction page for concurrency primitives.

khajavi:
@jdegoes 
In the past few days, I was working on concurrency and stm pages. 
In this merge request, I'm going to send the concurrency work It's ready to be reviewed. I will do the stm introduction in the separate merge request.


docs/datatypes/concurrency/index.md
```diff
@@ -1,13 +1,100 @@
...
+        loop = !value.compareAndSet(current, tuple._2)
+      }
+      b
+    }
+ }
+```
+
+The idea behind the `modify` is that a variable is only updated if it still has the same value as the time we had read the value from the original memory location. If the value has changed, it retries in the while loop until it succeeds. 
+
+## Advantage of Using ZIO Concurrency
```
jdegoes, line 65: Another key point: ZIO concurrency structures, including those built with STM, are _safe for interruption_. This means that you can place them into timeouts, race them, and so forth, without leaking any resources.

jdegoes (approved):

jdegoes:
🙏

PR #4937, opened 2021-04-10
Implement serviceWith

kitlangton:
`serviceWith` is merely a terser variant of `accessM`, that takes care of the `_.get` behind the scenes.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -3868,7 +3868,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   /**
    * Effectfully accesses the specified service in the environment of the effect.
    *
-   * Especially, useful for creating "accessor" methods on Services' companion objects.
+   * Especially useful for creating "accessor" methods on Services' companion objects.
    *
    * {{{
    * def foo(int: Int) = ZIO.serviceWith[Foo](_.foo(int))
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -3866,6 +3866,18 @@ object ZIO extends ZIOCompanionPlatformSpecific {
     ZIO.access(r => (r.get[A], r.get[B], r.get[C], r.get[D]))
 
   /**
+   * Effectfully accesses the specified service in the environment of the effect.
+   *
+   * Especially, useful for creating "accessor" methods on Services' companion objects.
+   *
+   * {{{
+   * def foo(int: Int) = ZIO.serviceWith[Foo](_.foo(int))
+   * }}}
+   */
+  def serviceWith[Service]: ServiceWithPartiallyApplied[Service] =
+    new ServiceWithPartiallyApplied[Service]
+
+  /**
    * Returns an effect that shifts execution to the specified executor. This
    * is useful to specify a default executor that effects sequenced after this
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3865,6 +3865,18 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def services[A: Tag, B: Tag, C: Tag, D: Tag]: URIO[Has[A] with Has[B] with Has[C] with Has[D], (A, B, C, D)] =
     ZIO.access(r => (r.get[A], r.get[B], r.get[C], r.get[D]))
 
+  /**
+   * Effectfully accesses the specified service in the environment of the effect.
+   *
+   * Especially, useful for creating "accessor" methods on Services' companion objects.
```
zagyi, line 3871:
```suggestion
   * Especially useful for creating "accessor" methods on Services' companion objects.
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3865,6 +3865,18 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+   *
+   * {{{
+   * def foo(int: Int) = ZIO.serviceWith[Foo](_.foo(int))
+   * }}}
+   */
+  def serviceWith[Service]: ServiceWithPartiallyApplied[Service] =
```
AdamPaynter, line 3877: Apologies for the naive question, but why is the "partially applied" pattern used in this scenario? I was under the impression that it helps with type inference if the method is supposed to walk, talk and quack as though it had multiple parameter lists.

core/shared/src/main/scala/zio/ZIO.scala
kitlangton: Good eye, @zagyi!

core/shared/src/main/scala/zio/ZIO.scala
kitlangton: Unfortunately, one cannot _partially apply_ type parameters. We need to use this pattern if we want to be able to call this method as such:
```scala
ZIO.serviceWith[Foo](_.foo(int))
```
If it didn't use this awkward PartiallyApplied class trick, we'd have to specify the error and return types:

```scala
ZIO.serviceWith[Foo, Nothing, Int](_.foo(int))
```

AdamPaynter: Gotcha. In an alternative universe, would an alternative Scala allow something like this?
```
def serviceWith[Service][E, A](f: Service => ZIO[Has[Service], E, A])
```
If so, has such syntax been proposed with the keepers of Scala?

adamgfraser (approved):

core/shared/src/main/scala/zio/ZIO.scala
ivanopagano: > Gotcha. In an alternative universe, would an alternative Scala allow something like this?
> 
> ```
> def serviceWith[Service][E, A](f: Service => ZIO[Has[Service], E, A])
> ```
> 
> If so, has such syntax been proposed with the keepers of Scala?

In scala3 you have [named type arguments](https://dotty.epfl.ch/docs/reference/other-new-features/named-typeargs)

If I'm not mistaken, it should allow you to do something like:
```scala
ZIO.serviceWith[Service = Foo](_.foo(int))
```

PR #4939, opened 2021-04-11
Fix drop and dropWhile

paulpdaniels:
Relates to #4886 

Makes sure that `drop` and `dropWhile` don't swallow errors and fixes a bug in `drop` that broke the units when I uncommented them.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -807,19 +807,24 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
         self.interruptWhen(bgDied)
     }
 
+  /**
+   * Drops the specified number of elements from this stream.
+   */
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -807,19 +807,24 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
-          val dropped  = chunk.drop(n)
-          val leftover = (n - dropped.length).min(0)
-          val more     = notEnded && leftover > 0
-
-          if (more) loop(leftover) else ZChannel.write(dropped) *> ZChannel.identity[E, Chunk[A], Any]
-        }
+        .readWith(
+          (in: Chunk[A]) => {
+            val dropped  = in.drop(r)
+            val leftover = (r - in.length).max(0)
```
paulpdaniels, line 819: Glad to know that I'm not the only one who has done this 😺

iravid (approved):

PR #4949, opened 2021-04-12
Delete Semaphore

adamgfraser:
With the extremely high performance of our implementation of software transactional memory we can delete `Semaphore` and just use `TSemaphore`. This both minimizes the number of different data structures and allows users to seamlessly move from `withPermit` for simple cases to using `acquire` and `release` in a transactional context to solve more complex concurrency problems.


core/shared/src/main/scala/zio/stm/TSemaphore.scala
```diff
@@ -43,14 +91,54 @@ final class TSemaphore private (val permits: TRef[Long]) extends AnyVal {
...
 
-  def withPermit[E, B](stm: STM[E, B]): STM[E, B] =
-    acquire *> stm <* release
+  /**
+   * Executes the specified effect, acquiring a permit immediately before the
+   * effect begins execution and releasing it immediately after the effect
+   * completes execution, whether by success, failure, or interruption.
+   */
+  def withPermit[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
+    withPermits(1L)(zio)
```
mijicd, line 100: Does it make sesne to keep the old `withPermit`?

mijicd (approved):

jdegoes (approved):

core/shared/src/main/scala/zio/stm/TSemaphore.scala
adamgfraser: One of the challenges of this is creating an API that feels natural both when you are using it as just a `Semaphore` and when you are using it as a `TSemaphore`.

I thought making the `withPermit` variants take a `ZIO` effect and having everything else return an `STM` effect struck a nice balance because `withPermit` variants were the only ones used with `Semaphore` before and created a clear communication message of `withPermit` for basic stuff, other operators for fancy stuff.

Also I think for STM there is less need for it because you don't have to worry about interruption within STM and I think most of the time if you are using STM you want to compose it with other things.

So I would be inclined to keep it as is but maybe there is a use case I am missing.

mijicd: Agreed, let's keep it this way.

mijicd (dismissed):

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -933,17 +933,26 @@ object ZSTM {
...
-        tryCommit(platform, fiberId, stm, r) match {
-          case TryCommit.Done(io) => io // TODO: Interruptible in Suspend
-          case TryCommit.Suspend(journal) =>
-            val txnId     = makeTxnId()
-            val done      = new AtomicBoolean(false)
-            val interrupt = UIO(Sync(done)(done.set(true)))
-            val async     = ZIO.effectAsync(tryCommitAsync(journal, platform, fiberId, stm, txnId, done, r))
-
-            async ensuring interrupt
+    ZIO.uninterruptibleMask { restore =>
```
jdegoes, line 936: Unfortunately this will decrease performance significantly. 

Could it be the issue is caused by the TODO on line 941? In which case a simpler fix might suffice.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -933,17 +933,26 @@ object ZSTM {
...
+        ZIO.effectSuspendTotalWith { (platform, fiberId) =>
+          val state = new AtomicReference[State[E, A]](State.Running)
+          tryCommit(platform, fiberId, stm, state, r) match {
+            case TryCommit.Done(io) => io // TODO: Interruptible in Suspend
+            case TryCommit.Suspend(journal) =>
+              val txnId     = makeTxnId()
+              val interrupt = ZIO.effectTotal(state.compareAndSet(State.Running, State.Interrupted))
+              val async     = ZIO.effectAsync(tryCommitAsync(journal, platform, fiberId, stm, txnId, state, r))
+
+              restore(async).onInterrupt(interrupt).run.flatMap {
```
jdegoes, line 947: `onInterrupt` is way more costly than `ensuring`, and `onInterrupt(...).run` ends up calling `run` two times. Overall, this section of code will have significant performance implications.

core/shared/src/main/scala/zio/stm/ZSTM.scala
```diff
@@ -933,17 +933,26 @@ object ZSTM {
...
-            val txnId     = makeTxnId()
-            val done      = new AtomicBoolean(false)
-            val interrupt = UIO(Sync(done)(done.set(true)))
-            val async     = ZIO.effectAsync(tryCommitAsync(journal, platform, fiberId, stm, txnId, done, r))
-
-            async ensuring interrupt
+    ZIO.uninterruptibleMask { restore =>
+      ZIO.accessM[R] { r =>
+        ZIO.effectSuspendTotalWith { (platform, fiberId) =>
+          val state = new AtomicReference[State[E, A]](State.Running)
```
jdegoes, line 939: The state is always created even in the "happy path", which degrades performance of all code.

Perhaps `TryCommit.Suspend` could return the atomic reference, i.e. push its allocation into the non-happy path (async STM).

jdegoes:
The issues with `atomically` surely have to be fixed, although the changes here are of such a nature, they will have a significant impact on all benchmarks.

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: I simplified this substantially to use a single fold.

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: I created a `tryCommitSync` variant that doesn't require the state so that we could push its creation into the asynchronous path.

core/shared/src/main/scala/zio/stm/ZSTM.scala
adamgfraser: I think that is right but I'm not sure how we can avoid it. To avoid using `uninterruptibleMask` in the synchronous case I think we need to avoid performing any continuation after we call `tryCommit` in the synchronous case. But we already need to return an effect in the asynchronous branch and even in the synchronous branch if we change the result of `TryCommit` to return an `Exit` value we are still going to need a continuation to flatten the result back to a `ZIO` value.

adamgfraser:
Completely agree. If we can solve the issue with the interruptibility in suspend we can push `uninterruptibleMask` into the asynchronous case and then the happy path will be unaffected by these changes.

adamgfraser:
@jdegoes I implemented `effectSuspendMaybeWith` with the following signature:

```scala
  /**
   * A variant of `effectiveSuspendTotalWith` that allows optionally returning
   * an `Exit` value immediately instead of constructing an effect.
   */
  def effectSuspendMaybeWith[R, E, A](f: (Platform, Fiber.Id) => Either[Exit[E, A], ZIO[R, E, A]]): ZIO[R, E, A] =
    new ZIO.EffectSuspendMaybeWith(f)
```

It is a bit ad hoc but with this change the happy path is now faster than it was originally. Let me know what you think.

adamgfraser:
This could actually be useful in some other situations where we want to perform some side effect and based on that potentially immediately return a result or else do something else. For example in the implementation of `ZQueue#take` immediately returning the element if there is one or else performing an effect to create a promise and wait for it. I believe we had an issue with potentially getting interrupted immediately after taking a value from the queue as well though that was not as significant as it is here.

jdegoes (approved):

PR #4955, opened 2021-04-12
Implement ZSink#managed

jupposessho:
Naive implementation which is failing on the "sad path", I need some hint here.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/ZChannelSimulatedChecks.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/ZChannelSimulatedChecks.scala
@@ -7,7 +7,7 @@ import zio.test._
 import zio.{Chunk, IO, ZIO, ZIOBaseSpec}
 
 object ZChannelSimulatedChecks extends ZIOBaseSpec {
-  override def spec =
+  override def spec: ZSpec[Environment, Failure] =
     suite("ZChannel simulated checks")(
       testM("done channel")(
         checkM(gen) { sim =>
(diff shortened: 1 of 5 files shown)


adamgfraser:
@jupposessho What failure are you getting?

jupposessho:
@adamgfraser I got 
```
`r` = Right(ok) did not satisfy equalTo(Left(fail))
```

adamgfraser:
I think your implementation is correct but reflects a difference in the semantics between the semantics of the previous implementation and this one. In this implementation since the sink is just a channel, the upstream is only evaluated when the downstream issues a read instruction. In this case `ZSInk.succeed` immediately succeeds without issuing a read instruction so the upstream is never evaluated, hence returning `Right(ok)`.

adamgfraser:
For example consider the following test:

```scala
testM("ZSink failure handling") {
  assertM(ZStream.fail("fail").run(ZSink.succeed("ok")))(
    equalTo("ok")
  )
}
```

With the new encoding this test will pass but with the old encoding it will fail.

So I think you just need to update the test and then this is good to go.

jupposessho:
Thanks @adamgfraser for the explanation! Just fixed the test.

adamgfraser (approved):

adamgfraser:
@iravid Is this test flaky? Doesn't seem related to this PR.

```
[info] - ZChannelSpec
[info]   - interpreter
[info]     - reads
[info]       - concurrent reads
[info]         - simple concurrent reads
[info]         - nested concurrent reads
```

adamgfraser:
Observed this test being flaky on another PR as well.

iravid:
@adamgfraser Yeah that's possible. I saw that before but thought the nonFlaky `n` was too high for CI, but might be a real bug.

adamgfraser:
@iravid Ah, okay, maybe we just need to reduce the number of repetitions in `nonFlaky`.

Issue #4796, opened 2021-03-12
FR: Add a means to `zio.clock` to retrieve a `ScheduledExecutorService` for interop with legacy Java code


virusdave:
`zio.clock.PlatformSpecific` has one internally, and sometimes java libraries want one.  

It would be nice to be able to connect those two together, or alternatively have one created that's backed by this module.

adamgfraser:
I will look at adding this.

virusdave:
Thanks.  I notice that the existing on in the module is using a thread pool of size 1, so it's perhaps better to create a new object that is backed by ZIO fibers and `Clock` rather than reuse the existing one.  But you're the expert :)  Meanwhile, I'm doing this latter idea right now to unblock.  Not sure it'll be all that optimal, but could contribute the impl if interested.

adamgfraser:
Yes. The principle with the `ScheduledExecutorService` is that it is not supposed to be used for anything other than scheduling tasks and the actual execution of the task is to be immediately shifted to another executor so that the `ScheduledExecutorService` is free to keep triggering other effects scheduled for the future. So agree the tricky part is giving the user an implementation that is safe and does not pose a risk to the integrity of the ZIO runtime.

PR #4956, opened 2021-04-13
Implement Clock#asScheduledExecutorService

adamgfraser:
Second take at resolving #4796.

Implements a new `asScheduledExecutorService` operator on `Clock` that supports viewing a `Clock` as a `ScheduledExecutorService` for interoperability. For binary compatibility there is a default implementation that is not very efficient because it has to effectually access the current time, but the implementation for the `live` service should be quite efficient as it just delegates to the underlying `ScheduledExecutorService` backing the `Clock`.

Logic is included for ensuring that all work is immediately shifted back to the core asynchronous thread pool for execution so that submission of tasks does not impair the ability to schedule other work.

The default implementation can be moved to the `TestClock` in ZIO 2.0. The downside of this is we have more platform specific logic for the `Clock` service and are committing that whatever we do in the future for scheduling tasks with a delay will ultimately be able to support the `ScheduledExecutorService` interface, which is kind of annoying because it requires implementing a lot of low level logic that we would like to handle at a higher level like `Schedule`. But it is nice to support interoperability with other code that expects a `ScheduledExecutorService`.


virusdave:
Hi!  Anything I can do to help review this?

PR #4961, opened 2021-04-15
Handle careless use in bracket

### LINKED ISSUE #4960: Release in bracket does not get eval'd with careless use action
I was preparing a ZIO workshop at Pirum (YAY 🎉) when this broke my expectations:
```scala
ZIO.bracket(putStrLn("Start"))(_ => putStrLn("End"))(_ => throw new RuntimeException("oh noes")) 
```
... did not print `End`. I think it should, the runtime loop has logic to handle that case.

regiskuckaertz:
Closes #4960

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2331,7 +2331,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   ): ZIO[R, E, B] =
     ZIO.uninterruptibleMask[R, E, B](restore =>
       acquire.flatMap(ZIOFn(traceAs = use) { a =>
-        restore(ZIO.effectSuspendTotal(use(a))).run.flatMap(ZIOFn(traceAs = release) { e =>
+        ZIO.effectSuspendTotal(restore(use(a))).run.flatMap(ZIOFn(traceAs = release) { e =>
           ZIO
             .effectSuspendTotal(release(a, e))
             .foldCauseM(
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -2331,7 +2331,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   ): ZIO[R, E, B] =
     ZIO.uninterruptibleMask[R, E, B](restore =>
       acquire.flatMap(ZIOFn(traceAs = use) { a =>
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2331,7 +2331,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   ): ZIO[R, E, B] =
     ZIO.uninterruptibleMask[R, E, B](restore =>
       acquire.flatMap(ZIOFn(traceAs = use) { a =>
-        restore(use(a)).run.flatMap(ZIOFn(traceAs = release) { e =>
+        restore(ZIO.effectSuspendTotal(use(a))).run.flatMap(ZIOFn(traceAs = release) { e =>
           release(a, e).foldCauseM(
```
regiskuckaertz, line 2335: Should this also be wrapped?

regiskuckaertz (commented):
Just left one question. Cannot hurt right?

regiskuckaertz:
Looks like spurious failures

regiskuckaertz:
Not spurious after all. @adamgfraser  looks like I plugged a leak in the stack tracking stuff. Ready for review sir.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2331,11 +2331,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   ): ZIO[R, E, B] =
     ZIO.uninterruptibleMask[R, E, B](restore =>
       acquire.flatMap(ZIOFn(traceAs = use) { a =>
-        restore(use(a)).run.flatMap(ZIOFn(traceAs = release) { e =>
-          release(a, e).foldCauseM(
-            cause2 => ZIO.halt(e.fold(_ ++ cause2, _ => cause2)),
-            _ => ZIO.done(e)
-          )
+        restore(ZIO.effectSuspendTotal(use(a))).run.flatMap(ZIOFn(traceAs = release) { e =>
```
adamgfraser, line 2334: I think we want `restore` around `use` instead of `effectSuspendTotal`.

regiskuckaertz, line 2334:
```suggestion
        ZIO.effectSuspendTotal(restore(use(a))).run.flatMap(ZIOFn(traceAs = release) { e =>
```
Yes, thank you.

adamgfraser (approved):

PR #4963, opened 2021-04-15
Fix Flaky Test

adamgfraser:
This test appears to be flaky because the number of repetitions is too high and is causing the test to time out. I tested locally and was not able to observe any flakiness in the code being tested.


iravid (approved):
Thanks 🙏

PR #4967, opened 2021-04-17
Rename to bufferChunks and add other buffer variants

luis3m:
Part of #4886
Relates to #3766 and #4167

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -303,11 +303,39 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
 
   /**
    * Allows a faster producer to progress independently of a slower consumer by buffering
+   * up to `capacity` elements in a queue.
+   *
+   * @note This combinator destroys the chunking structure. It's recommended to use chunkN afterwards.
+   * @note Prefer capacities that are powers of 2 for better performance.
+   */
+  final def buffer(capacity: Int): ZStream[R, E, A] = {
+    val queue = self.toQueueOfElements(capacity)
+    new ZStream(
+      ZChannel.managed(queue) { queue =>
+        lazy val process: ZChannel[Any, Any, Any, Any, E, Chunk[A], Unit] =
+          ZChannel.fromEffect {
+            queue.take
+          }.flatMap { (exit: Exit[Option[E], A]) =>
+            exit.fold(
+              Cause
+                .flipCauseOption(_)
+                .fold[ZChannel[Any, Any, Any, Any, E, Chunk[A], Unit]](ZChannel.end(()))(ZChannel.halt(_)),
+              value => ZChannel.write(Chunk.single(value)) *> process
+            )
+          }
+
+        process
+      }
+    )
+  }
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
    * up to `capacity` chunks in a queue.
    *
    * @note Prefer capacities that are powers of 2 for better performance.
    */
-  final def buffer(capacity: Int): ZStream[R, E, A] = {
+  final def bufferChunks(capacity: Int): ZStream[R, E, A] = {
     val queue = self.toQueue(capacity)
     new ZStream(
       ZChannel.managed(queue) { queue =>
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1368,6 +1498,37 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
 
+  /**
+   * Like [[ZStream#into]], but provides the result as a [[ZManaged]] to allow for scope
+   * composition.
+   */
+  final def runIntoElementsManaged[R1 <: R, E1 >: E](
+    queue: ZQueue[R1, Nothing, Nothing, Any, Exit[Option[E1], A], Any]
+  ): ZManaged[R1, E1, Unit] = {
+    val writer = ZChannel
+      .managed(self.toPull.mapM(BufferedPull.make[R, E, A](_))) { pull =>
```
iravid, line 1509: I think `toPull` is not required here; you could just use `readWith`, no?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -361,6 +392,105 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+
+    ZChannel.managed(managed.tap(queue => runProducer(queue).fork))(runConsumer)
+  }
+
+  /**
+   * Allows a faster producer to progress independently of a slower consumer by buffering
+   * elements into an unbounded queue.
+   */
+  final def bufferUnbounded: ZStream[R, E, A] = {
+    val queue = self.toQueueUnboundedOfElements
```
iravid, line 474: We should use the chunking variant here so we don't lose the chunking structure after the operator.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -321,27 +348,31 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+    new ZStream(bufferSignal[R, E, A](queue, _.upstream))
+  }
 
   /**
    * Allows a faster producer to progress independently of a slower consumer by buffering
-   * elements into an unbounded queue.
+   * chunks into an unbounded queue.
    */
-  final def bufferUnbounded: ZStream[R, E, A] = {
+  final def bufferChunksUnbounded: ZStream[R, E, A] = {
```
iravid, line 375: Hmm. We're buffering with an unbounded size - so there's no real difference between chunked/unchunked no? This could be the only operator we have.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -295,11 +295,38 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
 
   /**
    * Allows a faster producer to progress independently of a slower consumer by buffering
-   * up to `capacity` chunks in a queue.
+   * up to `capacity` elements in a queue.
    *
    * @note Prefer capacities that are powers of 2 for better performance.
```
iravid, line 300: Can you add another note that this combinator destroys the chunking structure and it is recommended to add a call to `chunkN` afterwards?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -361,6 +392,105 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+        )
+
+      for {
+        as    <- self.toPull.mapM(BufferedPull.make[R1, E1, A1](_))
+        start <- Promise.make[Nothing, Unit].toManaged_
+        _     <- start.succeed(()).toManaged_
+        ref   <- Ref.makeManaged(start)
+        _     <- Take.fromPull(toPull(as)).tap(offer(ref)).repeatWhile(_.isSuccess).unit.toManaged_.fork
+      } yield ()
+    }
```
iravid, line 448: It seems that we can write this without `toPull` as we are not relying on reading from it in another channel. Is that possible?

iravid: Another interesting note is that @adamgfraser has shown that we can write things that use sliding buffers using the ring buffer directly. Would you like to look into that here?

iravid (commented):
Great work! I think we can simplify in some places by not using `toPull`. I am also happy to do this in a follow up if you'd like to merge first. Let me know.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: Makes sense, keeping just one of these then.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: I will try now to implement this part without `toPull`. I just rewrote `runIntoElementsManaged` without it. Is there any place where you can point me at to look at the ring buffer?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: You're right, done.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: As suggested just kept the chunking variant for unbounded buffer.

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: Done

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: Done

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
luis3m: Re implemented without Pull. Maybe explore the ring buffer in a follow-up @iravid?

luis3m: I just found Adam's implementation of the ring buffer. I will update this PR with its usage.

luis3m:
Rebased. Solved conflicts. (@jdegoes @iravid)

iravid (approved):

PR #4971, opened 2021-04-18
Implement haltWhen

luis3m:
Part of #4886

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -1255,8 +1255,8 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
       }
 
     new ZStream(
-      ZChannel.unwrap {
-        io.fork.map { fiber =>
+      ZChannel.unwrapManaged {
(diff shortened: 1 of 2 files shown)


luis3m:
I'm going to rewrite this one without `.toPull`

luis3m:
Done

luis3m:
Actually, I can simplify. Changing.

luis3m:
Done, @iravid

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1238,8 +1238,30 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+            )
+
+          case Some(exit) =>
+            exit.fold(ZChannel.halt(_), _ => ZChannel.unit)
+        }
+      }
+
+    new ZStream(
+      ZChannel.unwrap {
+        io.fork.map { fiber =>
```
iravid, line 1259: We should use forkManaged here

luis3m: Done, @iravid

iravid (approved):

Issue #4803, opened 2021-03-15
ZIO Test: Include duration and failure message in junit test report


xkrillex:
Right now duration and failure message in the junit test reports is as following;

`<testcase classname="MyTest" name="Should assert something" time="0.0">
  <failure message="No Exception or message provided"/>
</testcase>`

Would be nice to get these implemented. For us this would make it a great addition so that we can get this data into our CI tool.

PR #4972, opened 2021-04-18
feat(zio-test): include duration in JUnit test reports

MichalPawlicki:
Include durations of individual tests in JUnit reports generated by sbt. Addresses https://github.com/zio/zio/issues/4803.

This is still WIP and I'd like to get some feedback before I move forward. Basically, my idea is to:
- Add a `duration` field to `TestSuccess` and `TestFailure`.
- Time an assertion in `ZTest#apply` and pass the duration to `TestSuccess`/`TestFailure`.
- In `ZTestEvent#from`, pass the duration to `ZTestEvent`.

A couple of questions about this approach:
- Is it ok to always time tests, or should we give users the option to disable this?
- Are `TestSuccess` and `TestFailure` the right places to store duration?
- Timing the `assertion` in `ZTest#apply` produces expected results, but maybe there's a better place to do it?

### DIFF
--- a/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
+++ b/test-sbt/jvm/src/test/scala/zio/test/sbt/ZTestFrameworkSpec.scala
@@ -1,7 +1,8 @@
 package zio.test.sbt
 
 import sbt.testing._
-import zio.UIO
+import zio.duration._
+import zio.test.environment.Live
 import zio.test.sbt.TestingSupport._
 import zio.test.{
   Annotations,
@@ -15,6 +16,7 @@ import zio.test.{
   TestSuccess,
   ZSpec
 }
+import zio.{UIO, ZIO}
 
 import java.util.regex.Pattern
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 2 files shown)


adamgfraser:
I think this use case is already addressed by the `timed` test aspect.

MichalPawlicki:
Thanks, `timed` does the job, I changed the implementation so that duration is extracted from the `timing` annotation.

I still have a problem though, the duration of a failing test is reported as 0. (I rebased on master so my branch already includes this fix: https://github.com/zio/zio/pull/4988). The `timing` annotation seems to disappear before the event is reported to sbt. Could it be caused by suite transformations, e.g. the one here: https://github.com/zio/zio/blob/eb6b543ac533c66d2300865f0cb5c93bfacc28fb/test/shared/src/main/scala/zio/test/Spec.scala#L186 ?

adamgfraser:
@MichalPawlicki No that line of code only runs if the suite can't be created at all (e.g. the suite of tests was supposed to be generated based on contents from a file and couldn't be loaded). The time of failing tests is being reported correctly in tests with SBT so I would look in the implementation of the JUnit runner itself.

MichalPawlicki:
So basically, this seems to work. When I break a random test, annotate it with `timed`, and run it with the sbt `testOnly` task,
sbt generates a report in `target/test-reports` which includes the failure with a duration:
```xml
<testcase classname="zio.test.TestAspectSpec" name="around evaluates tests inside context of Managed" time="0.208">
  <failure message="No Exception or message provided"/>
</testcase>
```
I'm just having problems with making it work in `ZTestFrameworkSpec`. I'll look further into this.

adamgfraser:
@MichalPawlicki That's fine.

MichalPawlicki:
@adamgfraser this is ready for review. 🙂 In the end, I added a test, but just for a passing spec.

adamgfraser (approved):
Thank you! 🙏

PR #4974, opened 2021-04-18
Implement interleave

luis3m:
Part of #4886

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
@@ -1324,8 +1324,86 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
    */
   final def interleaveWith[R1 <: R, E1 >: E, A1 >: A](
     that: ZStream[R1, E1, A1]
-  )(b: ZStream[R1, E1, Boolean]): ZStream[R1, E1, A1] =
-    ???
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1279,7 +1279,59 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
   final def interleaveWith[R1 <: R, E1 >: E, A1 >: A](
     that: ZStream[R1, E1, A1]
   )(b: ZStream[R1, E1, Boolean]): ZStream[R1, E1, A1] =
-    ???
+    new ZStream(
+      ZChannel.managed {
```
iravid, line 1283: Hmm this was previously implemented using `combine`. If we still need to use `toPull` here, could we keep the previous implementation? That means that when (if) we find a way to implement combine/combineChunks without toPull, this implementation would be solved as well.

luis3m: Makes sense, @iravid. I will update.

luis3m:
Actually, I can simplify a bit. Will update soon.

luis3m:
Done

jdegoes:
@luis3m Thanks for your work on this. Can you check into build failures?

luis3m:
@jdegoes rebased.

iravid (approved):

PR #4979, opened 2021-04-19
Implement ZSink#filterInput and filterInputM

jupposessho:
Part of #4886

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
@@ -154,6 +154,14 @@ class ZSink[-R, -InErr, -In, +OutErr, +L, +Z](val channel: ZChannel[R, InErr, Ch
   ): ZSink[R1, InErr1, In1, OutErr1, L, Z1] =
     contramapChunksM(f).mapM(g)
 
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -154,6 +154,28 @@ class ZSink[-R, -InErr, -In, +OutErr, +L, +Z](val channel: ZChannel[R, InErr, Ch
   ): ZSink[R1, InErr1, In1, OutErr1, L, Z1] =
     contramapChunksM(f).mapM(g)
 
+  def filterInput[In1 <: In](p: In1 => Boolean): ZSink[R, InErr, In1, OutErr, L, Z] = {
+    lazy val loop: ZChannel[R, InErr, Chunk[In1], Any, InErr, Chunk[In], Any] =
+      ZChannel.readWith[R, InErr, Chunk[In1], Any, InErr, Chunk[In], Any](
```
iravid, line 159: These can be simplified to `contramap(_.filter(p ))` :-) (same for filterInputM)

jupposessho: Almost had it but found foldChunks first and tried with that but it was not good for this :)
Thanks!

PR #4983, opened 2021-04-19
Implement toReader

luis3m:
Part of #4886


luis3m:
Initially I implemented this one with Handoff/ZQueue but got JS complaining about blocking code.

luis3m:
I just realized this one was already implemented here #4980

PR #4984, opened 2021-04-19
Implement refineOrDie

luis3m:
Part of #4886


luis3m:
Already implemented in #4982, closing.

PR #4991, opened 2021-04-20
Implement mapMPar

luis3m:
Part of #4886

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
@@ -619,6 +619,65 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
     self >>> reader
   }
 
+  def mapOutMPar[Env1 <: Env, OutErr1 >: OutErr, OutElem2](n: Int)(
+    f: OutElem => ZIO[Env1, OutErr1, OutElem2]
+  ): ZChannel[Env1, InErr, InElem, InDone, OutErr1, OutElem2, OutDone] = {
+
+    def producer(
+      queue: Queue[ZIO[Env1, Either[OutErr1, OutDone], OutElem2]],
+      pull: ZIO[Env1, Either[OutErr, OutDone], OutElem]
+    ): ZManaged[Env1, Nothing, Unit] =
+      for {
+        errorSignal <- Promise.makeManaged[OutErr1, Nothing]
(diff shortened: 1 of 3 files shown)


luis3m:
Initially tried to implement without `toPull` but I couldn't use `(self >>> producer).runManaged.fork` due to the type parameter constraints runManaged has.

luis3m:
Trying to figure out how to resolve that failing test.

streams/shared/src/main/scala/zio/stream/experimental/ZChannel.scala
```diff
@@ -621,6 +621,59 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
...
+    ZChannel.managed {
+      ZManaged.withChildren { getChildren =>
+        for {
+          _           <- ZManaged.finalizer(getChildren.flatMap(Fiber.interruptAll(_)))
+          queue       <- Queue.bounded[ZIO[Env1, Either[OutErr1, OutDone], OutElem2]](n).toManaged(_.shutdown)
+          errorSignal <- Promise.makeManaged[OutErr1, Nothing]
+          permits     <- Semaphore.make(n.toLong).toManaged_
+          pull        <- self.toPull
+          _ <- pull
+                 .foldM(
```
iravid, line 636: A `Die` will make this deadlock. So we should use `foldCauseM` here.

luis3m: Good point, thanks.

iravid:
@luis3m Also I think we could avoid the `chunkN(1)` by doing `self.channel.concatMap(ZChannel.fromChunk(_))` (and implementing `ZChannel.fromChunk` :-))

iravid:
Ah I have `ZChannel.fromChunk` in my branch:
```scala
  def writeAll[Out](outs: Chunk[Out]): ZChannel[Any, Any, Any, Any, Nothing, Out, Unit] = {
    def writer(idx: Int, len: Int): ZChannel[Any, Any, Any, Any, Nothing, Out, Unit] =
      if (idx == len) ZChannel.unit
      else ZChannel.write(outs(idx)) *> writer(idx + 1, len)

    writer(0, outs.size)
  }

```

iravid:
@luis3m could you run `sbt prepare` and look into the failing test?

luis3m:
@iravid all passing now (cc @jdegoes)

iravid (approved):
Fantastic!!

Issue #4992, opened 2021-04-20
ZIO 1.0.6 fails to compile implicit


juliano:
I have the following code running on ZIO 1.0.5:

```scala
@accessible
object Example {
  trait Service {
    def foo[A](value: String)(implicit decoder: Decoder[A]): IO[Throwable, A]
  }
}

object AnotherService {
  case class Bar(value: Int)

  object Bar {
    implicit val decoder: Decoder[Bar] = deriveDecoder
  }

  def bar(implicit decoder: Decoder[Bar]): ZIO[Has[Example.Service], Throwable, Bar] =
    Example.foo[Bar]("hey")(decoder)
}
```

After migrating to ZIO 1.0.6, it fails to compile:

```sbt
[error] /Users/juliano/dev/proj/src/main/scala/example/AnotherService.scala:14:28: not enough arguments for method foo: (implicit decoder: io.circe.Decoder[example.AnotherService.Bar], ev: izumi.reflect.Tag[example.Example.Service]): zio.ZIO[zio.Has[example.Example.Service] with Any,Throwable,example.AnotherService.Bar].
[error] Unspecified value parameter ev.
[error]     Example.foo[Bar]("hey")(Bar.decoder)
```

I've noticed it only happens when `Bar.decoder` is explicitly there, it compiles otherwise.

adamgfraser:
@nadenf You want to take a look at this?

nadenf:
@adamgfraser .. Looking into it now.

adamgfraser:
@nadenf Thank you! 🙏

nadenf:
I am actually confused how this compiles in 1.0.5 because @accessible expands:
```
def foo[A](value: String)(implicit decoder: Decoder[A]): IO[Throwable, A]
```

to:
```
def foo[A](value: String)(implicit decoder: Decoder[A])(implicit ev: implicit ev: _root_.izumi.reflect.Tag[Service])
```

And according to the Scala spec multiple implicit blocks are not allowed. @ioleo any ideas ?

Otherwise it's best just to revert #4466 and instead workaround it in the IntelliJ Plugin.

ioleo:
It should've been combined into a single implicit parameter list, I'm suprised to see there are multiple.

adamgfraser:
Yes I think that is right it should just be one implicit parameter list. Does someone have time to work on this?

nadenf:
In #4466 it combines multiple implicit into a single list which is what is causing this bug.

[...]

PR #4467, opened 2020-11-28
Support use of @accessible macro with implicits (#4466)

nadenf:
Existing automated tests pass but not sure how useful they are given that never picked up the bug described in #4466.

However manual testing with various combinations of implicits and normal parameters appear to work fine.

Issue #4466, opened 2020-11-28
@accessible generates incorrect code when implicits present


nadenf:
So I have a Logger module which looks like this:

```scala
@accessible
object Logger {
  type Logger = Has[Logger.Service]

  trait Service {
    def error(message: => String)(implicit file: File): UIO[Unit]
  }
```

Which gets translated to:
```scala
object Logger extends scala.AnyRef {
  type Logger = zio.Has[com.harana.sdk.backend.modules.logger.Logger.Service]
  trait Service extends scala.AnyRef {
    def error(message : => _root_.scala.Predef.String)(implicit file : sourcecode.File) : zio.UIO[scala.Unit]
  }
  def error(message : => _root_.scala.Predef.String)(implicit file : sourcecode.File)(implicit ev : izumi.reflect.Tag[com.harana.sdk.backend.modules.logger.Logger.Service]) : zio.ZIO[zio.Has[com.harana.sdk.backend.modules.logger.Logger.Service] with scala.Any, scala.Nothing, scala.Unit] = { /* compiled code */ }
}
```
When I try to use this module it is rendered in IntelliJ as follows:

<img width="398" alt="Screen Shot 2020-11-28 at 3 27 21 pm" src="https://user-images.githubusercontent.com/657041/100494122-abaa8f00-3192-11eb-846f-a7469c06c926.png">

The issue seems to be that when an existing implicit group is present it needs to be merged with the one the macro adds i.e. the Izumi one.

nadenf:
Closed with #4467

PR #5001, opened 2021-04-22
Revert "Support use of @accessible macro with implicits (#4466)"

nadenf:
Reverts #4467 in order to fix a regression: #4992


adamgfraser:
Is it possible to modify this to fix the original issue as well as the regression? My understanding was the regression was caused because we were generating two different implicit parameter lists so can we modify the logic here to generate a single implicit parameter list in this case instead?

adamgfraser (approved):

PR #5006, opened 2021-04-22
Implement Generalized Aspects

adamgfraser:


adamgfraser:
@jdegoes Generalized this to include lower and upper bounds. I think we need both for error handling aspects like `retry` since we may need an upper bound on the error type we can handle. I also found an interesting use for bounds on the value type with `MetricAspect` where certain metric aspects will place a constraint on the return type of the effect they can be applied to (e.g. it only makes sense to add a `Double` value to a histogram).

core/shared/src/main/scala/zio/ZIOAspect.scala
```diff
@@ -0,0 +1,78 @@
...
+import zio.duration._
+import zio.internal.Executor
+
+import scala.concurrent.ExecutionContext
+
+trait ZIOAspect[+LowerR, -UpperR, +LowerE, -UpperE, +LowerA, -UpperA] {
+  def apply[R >: LowerR <: UpperR, E >: LowerE <: UpperE, A >: LowerA <: UpperA](zio: ZIO[R, E, A]): ZIO[R, E, A]
+}
+
+object ZAspect {
```
jdegoes, line 13: Companion object should be the same name as the trait.

adamgfraser: Fixed now!

jdegoes (approved):

jdegoes:
The stream aspect wholly subsumes `ZPipeline`, I think, so we can deprecate / remove that. ❤️

PR #5007, opened 2021-04-22
ZSink.utf8Decode and more

iravid:


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2316,12 +2317,14 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
-            ZChannel.fromEffect(leftovers.set(leftover.flatten)) *>
-              ZChannel.write(Chunk.single(z)) *>
-              ZChannel.fromEffect(upstreamDone.get).flatMap { done =>
-                if (done) ZChannel.end(())
-                else transducer
-              }
+            ZChannel.fromEffect(upstreamDone.get <*> leftovers.updateAndGet(_ ++ leftover.flatten)).flatMap {
+              case (done, newLeftovers) =>
+                val nextChannel =
+                  if (done && newLeftovers.isEmpty) ZChannel.end(())
```
iravid, line 2323: There's a potential non-termination condition here. If a sink keeps returning leftovers, transduce will never terminate. Interested in hearing other opinions. The options I thought of are:
* This behavior
* If the stream is done, feed the leftovers to the sink one more time and check if the returned leftovers are identical, and if so, terminate (while dropping the leftovers)

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -160,22 +160,36 @@ class ZSink[-R, -InErr, -In, +OutErr, +L, +Z](val channel: ZChannel[R, InErr, Ch
...
   )(implicit ev: L <:< In1): ZSink[R1, InErr1, In1, OutErr2, L1, Z1] =
     new ZSink(
       channel.doneCollect.foldM(
         failure(_).channel,
         { case (leftovers, z) =>
-          (ZChannel.write(leftovers.flatMap(_.map(ev))) *> ZChannel.identity[InErr1, Chunk[In1], Any]) >>> success(
-            z
-          ).channel
+          ZChannel.fromEffect(Ref.make(leftovers.flatten)).flatMap { leftoversRef =>
+            val refReader = ZChannel.fromEffect(leftoversRef.getAndSet(Chunk.empty)).flatMap { chunk =>
```
iravid, line 177: It looks like we keep needing Refs for storing leftovers. This is unfortunate. Maybe we should accommodate this in the encoding?

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -2387,31 +2388,58 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
   def transduce[R1 <: R, E1, A1 >: A, Z](sink: ZSink[R1, E, A1, E1, A1, Z]): ZStream[R1, E1, Z] =
-    new ZStream(ZChannel.fromEffect(Ref.make[Chunk[A1]](Chunk.empty).zip(Ref.make(false))).flatMap {
-      case (leftovers, upstreamDone) =>
-        val buffer = ZChannel.bufferChunk[E, A1, Any](leftovers)
+    new ZStream(
+      ZChannel.effectSuspendTotal {
+        val leftovers: AtomicReference[Chunk[Chunk[A1]]] = new AtomicReference(Chunk.empty)
+        val upstreamDone: AtomicBoolean                  = new AtomicBoolean(false)
+
+        lazy val buffer: ZChannel[Any, E, Chunk[A1], Any, E, Chunk[A1], Any] =
```
jdegoes, line 2396: Doing this with laziness is really nice and will avoid repeated allocations. 👍

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
jdegoes: I think if the sink returns the leftovers, it is a sign that the sink needs more to continue. So feeding them to the sink again should not change the result. In fact that almost seems like a law of a well-behaved sink: if it returns leftovers, and then is fed those leftovers again, it must again return them.

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
jdegoes: It looks like you found a way around this???

jdegoes:
Looks good to me!

Issue #4997, opened 2021-04-21
MutableSpec: the TestEnvironment layer is shared between tests


hmemcpy:
I ran into an issue where I believe the `TestEnvironment` is provided as a shared layer to the `MutableRunnableSpec`, causing the services (`TestClock` in particular) to be shared between the tests.

My simplest repro for this (printing the hashcode of the `TestClock` service) is:
```scala
object MutableSpec extends DefaultMutableRunnableSpec {
  suite("test") {
    testM("a") {
      for {
        t <- ZIO.service[TestClock.Service]
        _ <- putStrLn(t.##.toString)
      } yield assertCompletes
    }

    testM("b") {
      for {
        t <- ZIO.service[TestClock.Service]
        _ <- putStrLn(t.##.toString)
      } yield assertCompletes
    }
  }
}
```

prints:
```
942338196
942338196
```

While doing this with the `DefaultRunnableSpec` prints two different values every time.

I believe this is a bug, or was this behavior intended? /cc @fokot

fokot:
@hmemcpy It looks like a bug `DefaultMutableRunnableSpec` should behave as `DefaultRunnableSpec`. I will take a look on it this week

hmemcpy:
Cheers, appreciate it!

adamgfraser:
@fokot Thanks for working on this!

fokot:
The problem lies here
https://github.com/zio/zio/blob/master/test/shared/src/main/scala/zio/test/DefaultMutableRunnableSpec.scala#L38
and here
https://github.com/zio/zio/blob/master/test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala#L148

I can change `provideLayerShared` to `provideLayer` and it would work the same as in `DefaultRunnableSpec`.

But problem is what if user wants to use  `MutableSpec extends MutableRunnableSpec(myOwnLayer)` and have custom `R`. Then his layer won't be shared.

I think there are two solutions.

1) Splitting it to two type parameters. The resulting `R` for test would be `TestEnvironment with CustomR` and `CustomR` would be shared.

2) Or ability to provide layers directly on the `tests` and `spec`
```
object MutableSpec extends DefaultMutableRunnableSpec {
  suite("test") {
    testM("a") {
      for {
        t <- ZIO.service[TestClock.Service]
        _ <- putStrLn(t.##.toString)
      } yield assertCompletes
    }.provideLayer(aaa)

    testM("b") {
      for {
[...]

Issue #4809, opened 2021-03-16
Defect in zio.Has: Set(SttpBackend[=λ %2:0 → ZIO[-Any,+Throwable,+2:0],+{ZioStreams & package::WebSockets}]) statically known to be contained within the environment are missing


dpennell:
This is similar to https://github.com/zio/zio/issues/4340.

sttp zio - v3.1.7
zio -        v1.0.4-2 and 1.0.5

I haven't been able to minimize this yet and won't be able to for a couple of days, but thought I'd see if anybody has any ideas.

TIA,
-david

dpennell:
@neko-kai tagging you due to similarity with #3558

virusdave:
I had something similar, and crafted a local ZLayer that looked like:
```
/** Local impl because of seeming macro bug in LightTypeTag macro impl */
object LocalSlackClient {
  lazy val live: ZLayer[Has[SttpClient.Service], Nothing, SlackClient] = ???  // locally duplicated code here
```

and this got me past the immediate failure.  I also haven't had time to minimize yet.

dpennell:
Thanks for the workaround.  I wasn't able to make that work in the time I had available.  I'll work on minimization this weekend.

neko-kai:
@dpennell Can't reproduce locally and with scastie (https://scastie.scala-lang.org/rqC3PUk9QSeg45Zxae6GHA). For these issues reproduction more or less equals solution (since they're deterministic) but I can't guess at the problem without a reproduction either...

hamsterready:
Try code from https://sttp.softwaremill.com/en/latest/examples.html#get-and-parse-json-using-the-zio-async-http-client-backend-and-circe

```
package sttp.client3.examples

import sttp.client3._
import sttp.client3.circe._
import sttp.client3.asynchttpclient.zio._
import io.circe.generic.auto._
import zio._
import zio.console.Console

object GetAndParseJsonZioCirce extends App {

  override def run(args: List[String]): ZIO[ZEnv, Nothing, ExitCode] = {

    case class HttpBinResponse(origin: String, headers: Map[String, String])

    val request = basicRequest
      .get(uri"https://httpbin.org/get")
      .response(asJson[HttpBinResponse])

    // create a description of a program, which requires two dependencies in the environment:
    // the SttpClient, and the Console
    val sendAndPrint: ZIO[Console with SttpClient, Throwable, Unit] = for {
      response <- send(request)
      _ <- console.putStrLn(s"Got response code: ${response.code}")
[...]

PR #5016, opened 2021-04-23
Test layer is not shared between tests anymore

fokot:
Fixes https://github.com/zio/zio/issues/4997 @adamgfraser

### DIFF
--- a/test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
+++ b/test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
@@ -57,8 +57,9 @@ class MutableRunnableSpec[R <: Has[_]: Tag](
 
   sealed case class SuiteBuilder(label: String) extends SpecBuilder {
 
-    private[test] var nested: Chunk[SpecBuilder]                     = Chunk.empty
-    private var aspects: Chunk[TestAspect[R with TestEnvironment, R with TestEnvironment, Failure, Failure]] = Chunk.empty
+    private[test] var nested: Chunk[SpecBuilder] = Chunk.empty
+    private var aspects: Chunk[TestAspect[R with TestEnvironment, R with TestEnvironment, Failure, Failure]] =
+      Chunk.empty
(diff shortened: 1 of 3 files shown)


adamgfraser (approved):
Good catch!

adamgfraser:
Looks like we're getting a test failure on Scala, possibly related to #4809. Can we also add a test for this?

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
```diff
@@ -39,25 +40,27 @@ import scala.util.control.NoStackTrace
...
  */
-class MutableRunnableSpec[R <: Has[_]](
+class MutableRunnableSpec[R <: Has[_]: Tag](
   layer: ZLayer[TestEnvironment, Throwable, R],
-  aspect: TestAspect[R, R, Any, Any]
+  aspect: TestAspect[R with TestEnvironment, R with TestEnvironment, Any, Any] = TestAspect.identity
 ) extends RunnableSpec[TestEnvironment, Any] {
   self =>
 
+  type RT = R with TestEnvironment
```
adamgfraser, line 49: Can we just use `R with TestEnvironment` instead of creating this new type alias?

adamgfraser (approved):
Left one comment but otherwise looks good. The Scala 3 issue is not great but I think that needs to be solved at the level of the tag machinery rather than here.

test/shared/src/main/scala/zio/test/MutableRunnableSpec.scala
fokot: Removed

fokot:
@adamgfraser ready to merge

adamgfraser (approved):
Awesome!

adamgfraser:
@fokot Can you run `prepare`?

adamgfraser (approved):

fokot:
@adamgfraser done

Issue #5013, opened 2021-04-23
groupedWithin ignoring duration parameter


AndyMoreland:
Hi! I asked about this in discord and Adam Fraser asked me to open an issue here. I've written this simple code sample:

```scala
      ZStream.repeatWith("hi", Schedule.spaced(Duration.fromMillis(100)))
           .groupedWithin(1000, Duration.fromMillis(300))
           .tap(x => UIO(println(s"Got chunk: ${x.size} at time ${Instant.now}")))
           .run(ZSink.take(50))
```

I'd expect to see a log line every 300ms because it will take 100 seconds to fill up the 1000 element buffer. However, I see a log line every 100 seconds. This seems counter to the documentation:

> Partitions the stream with the specified chunkSize or until the specified duration has passed, whichever is satisfied first.

PR #5018, opened 2021-04-24
Port test for groupedWithin

luis3m:
Part of #5013 
Relevant #4886


iravid (approved):
Always great to have more coverage 😍

PR #5029, opened 2021-04-28
Fix race condition in fromOutputStreamWriter

joroKr21:
Impose order on the closing of piped streams.
Manage the input stream automatically.


joroKr21:
Ok it's not that but it works locally 😭

PR #5033, opened 2021-05-01
Refactoring the test renderer: separating reporting from rendering

### LINKED ISSUE #3133: ZIO Test: make test reporting more extensible
Following up a discussion on Discord:

To make ZIO Test more pluggable we need to address two problems:

### Seperate report generating from rendering

We need to have an intermediate format which captures all of the information. We can base it of current **Line**, **Message** and **Fragment** in the **FailureRenderer**, but we need to abstract over the target.

For example: in [Fragment.apply](https://github.com/zio/zio/blob/28bc047fe15758f8de9dc63dd3703cb81659eed9/test/shared/src/main/scala/zio/test/DefaultTestReporter.scala#L263) we'd introduce a sealed trait `Style` with possible values:
* `Info` (cyan in console)
* `Error` (red in console)
* `Warning` (yellow in console)
* `Default` (white in console)
* `Primary` (blue in console)

Then each target output it might be interpreted diffrently:
* the **ConsoleReporter** would then translate that into ANSI color codes
* an **HtmlReporter** would instead add CSS classes to the output _<div>_ elements

We could also have a `FanOutReporter` which basically writes the output to multiple reporters (use case: Write to console + generate JUnit report).

### Pluggable design

We need to make it possible for external components to "plug in" at various points and transform the output. Example use case: we need **ZIO Mock** to be able to catch `Cause.Die` of **MockException** and handle them accordingly (by transforming them into **Messages**).

For that we could use traits as [demonstrated here](https://scastie.scala-lang.org/v0usS3SHSIe2Cv82CBWGIw).

From the user perspective all the user he needs to do is to add mixin another trait (`extends MockSpec`) which can hook into the process at various points and add its custom handling logic or fallback to `super`.

> For the "pluggable" aspect I wonder if we could handle more simply by just providing some methods that modify the test reporter. Something akin to withReportDefect that is just a method on the TestReporter that returns a new test reporter with that logic. Similar to what we do with Platform and Runtime in ZIO Core.

### Related work

This would allow us to seperate ZIO Mock from ZIO Test and move it to seperate subproject, it would also make it much easier to write custom reporters for other formats such as JUnit, HTML, etc.

@adamgfraser @jdegoes @hmemcpy @dkarlinsky

hmemcpy:
This PR separates producing the report of a test run and its rendering into two separate steps.
The test reporter prepares a data structure (`ExecutionResult`), containing output + some metadata in an intermediate format. In addition, a `ConsoleRenderer` takes this structure and writes it to the console.

This is the first step in a series of refactorings, aiming making reporting to different outputs (such as, html and IntelliJ) easier. Based on the ideas of, and closes #3133.

The output is identical to the current ZIO test output, however I'm sending the PR to the 2.0 branch, as it somewhat significantly modifies the current flow.

### DIFF
--- a/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
+++ b/test-junit/jvm/src/main/scala/zio/test/junit/ZTestJUnitRunner.scala
@@ -21,13 +21,14 @@ import org.junit.runner.notification.{Failure, RunNotifier}
 import org.junit.runner.{Description, RunWith, Runner}
 import zio.ZIO.effectTotal
 import zio._
-import zio.test.FailureRenderer.FailureMessage.Message
-import zio.test.RenderedResult.CaseType.Test
-import zio.test.RenderedResult.Status.Failed
+import zio.test.DefaultTestReporter.rendered
 import zio.test.Spec.{SpecCase, SuiteCase, TestCase}
 import zio.test.TestFailure.{Assertion, Runtime}
 import zio.test.TestSuccess.{Ignored, Succeeded}
 import zio.test._
+import zio.test.render.ExecutionResult.ResultType.Test
+import zio.test.render.ExecutionResult.Status.Failed
+import zio.test.render.LogLine.Message
 
 /**
  * Custom JUnit 4 runner for ZIO Test Specs.<br/>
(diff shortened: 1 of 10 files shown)


hmemcpy:
Not sure why mdoc is failing, but otherwise this is ready for review!

adamgfraser (approved):
@hmemcpy Thank you for this! Can you fix the `mdoc` issue and we can get this merged?

hmemcpy:
Sure! Let me look at it, I will make it a separate PR and rebase mine on top, once it's merged!

adamgfraser:
Sounds good!

adamgfraser (approved):

PR #5036, opened 2021-05-02
Smart assertions

kitlangton:
@adamgfraser and I have put together a first pass at **smart assertions**.

> Basically we use a macro to automatically construct the lens like assertions without the user having to do anything other than write an expression returning a Boolean, similar to how the the automatic layer construction functionality automatically wires together layers. — @adamgfraser 

Basically a series of rewrites (*I mean, all macros are rewrites I guess*:stuck_out_tongue_winking_eye: ):

```scala
val result: Int = ???
assert(result == 1)
//becomes
assert(result)(equalTo(1))

val result: Option[Int] = ???
assert(result.get == 1)
//becomes
assert(result)(isSome(equalTo(1))

val result: List[Int] = ???
assert(result.contains(10))
//becomes
assert(result)(contains(10))

case class Person(nameOption: Option[String])
val result: Person = ???
assert(result.nameOption.get.contains("howdy"))
//becomes
assert(result)(hasField("nameOption", _.nameOption, isSome(containsString("howdy"))))
```

EDIT: The above is no longer true 😜 We now translate into a specialized arrow-based dsl.


kitlangton:
CleanShot 2021-05-27 at 01 38 26@2x
CleanShot 2021-05-27 at 01 38 21@2x
CleanShot 2021-05-27 at 01 38 14@2x
CleanShot 2021-05-27 at 01 38 08@2x

test/shared/src/main/scala/zio/test/AssertionResult.scala
```diff
@@ -0,0 +1,32 @@
+package zio.test
+
+sealed trait AssertionResult { self =>
+  import AssertionResult._
+
+  def genFailureDetails: Option[GenFailureDetails]
+
+  def label(label: String): AssertionResult = {
+    println(s"label ${label}")
```
adamgfraser, line 9: Stray debug statement?

adamgfraser (approved):
Let's get this merged!

PR #5039, opened 2021-05-03
Add Support For State To ZIO

adamgfraser:


core/shared/src/main/scala/zio/State.scala
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Because `State` is typically used as part of the environment, it is
+ * recommended to define your own state type `S` such as `MyState` above
+ * rather than using a type such as `Int` to avoid the risk of ambiguity.
+ *
+ * To run an effect that depends on some state, create the initial state with
+ * the `make` constructor and then use `toLayer` to convert it into a layer
+ * that you can provide along with your application's other dependencies.
+ */
+sealed trait State[S] {
```
adamgfraser, line 42: This could be generalized to `ZState[E1, E2, S1, S2]` and then `type State[S] = ZState[Nothing, Nothing, S, S]`. I'm not sure how useful this would be and I think for ergonomics it would require another set of operators on `ZIO` (e.g. `getZState`, `setZState`) to avoid having to specify additional type parameters in the common case.

benchmarks/src/main/scala/zio/chunks/ChunkArrayBenchmarks.scala
```diff
@@ -1,7 +1,7 @@
 package zio.chunks
 
 import org.openjdk.jmh.annotations._
-import zio._
+import zio.{State => _, _}
```
adamgfraser, line 4: `State` is a relatively common name. We could potentially rename to `ZState` if we did not want to use `ZState` for a more polymorphic version.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4017,6 +4036,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def untraced[R, E, A](zio: ZIO[R, E, A]): ZIO[R, E, A] =
     zio.untraced
 
+  /**
+   * Updates a state in the environment with the specified function.
+   */
+  def updateState[S: Tag](f: S => S): ZIO[Has[State[S]], Nothing, Unit] =
```
adamgfraser, line 4042: Could potentially add more variants here consistent with the ones on `Ref` but not sure how useful they would be given that changes are occurring on single fiber and would also like to keep the number of ways to do things down.

core/shared/src/main/scala/zio/State.scala
regiskuckaertz: I could see this proliferating very fast in the `R` parameter, it would be great if either:
- `State` was renamed to `ST` ([homage](https://wiki.haskell.org/Monad/ST)), or
- the state was composable, though you'd have to think of how this would interfere with the semantics of `FiberRef`

regiskuckaertz (dismissed):
This is exactly what I was looking for 😃 Do you need to update `RIO`, `URIO`, etc. as well?

benchmarks/src/main/scala/zio/chunks/ChunkArrayBenchmarks.scala
jdegoes: I think we have to move to `ZState` given how many clashes there will be with `State`. Alternatively, it could be something like `EnvState`.

core/shared/src/main/scala/zio/State.scala
jdegoes: The other consideration is that it would place additional strain on overloading (the highly polymorphic ZState).

Still if there's a good argument, now is the time to consider it.

Neutral on rename to `ST` but I do think it needs to be renamed. 👍

core/shared/src/main/scala/zio/State.scala
```diff
@@ -0,0 +1,76 @@
...
+   */
+  def update(f: S => S): UIO[Unit]
+}
+
+object State {
+
+  /**
+   * Creates an initial state with the specified value.
+   */
+  def make[S](s: S): UIO[State[S]] =
```
jdegoes, line 65: Probably worthwhile to make a layer version of this.

core/shared/src/main/scala/zio/ZIO.scala
jdegoes: Exactly. Best to keep it minimal.

jdegoes (changes requested):
Mainly the name change but looks great! 👍 ❤️

core/shared/src/main/scala/zio/State.scala
adamgfraser: Renamed to `ZState`. I think it is nice to have names that are slightly more descriptive, especially since it is easy to use type aliases for your state type or types.

core/shared/src/main/scala/zio/State.scala
adamgfraser: Done.

adamgfraser:
This is ready to go.

adamgfraser:
> This is exactly what I was looking for 😃 Do you need to update `RIO`, `URIO`, etc. as well?

adamgfraser:
> This is exactly what I was looking for 😃 Do you need to update `RIO`, `URIO`, etc. as well?

jdegoes (approved):

PR #5040, opened 2021-05-04
Implement Scheduler#asScheduledExecutorService

### LINKED ISSUE #4796: FR: Add a means to `zio.clock` to retrieve a `ScheduledExecutorService` for interop with legacy Java code

adamgfraser:
Resolves #4796. I think this version is pretty clean. We add a `scheduler` method on `Clock` that exposes the `Scheduler` backing the `Clock`. This lets us push all the platform specific code into the `Scheduler`. The live implementation just returns the `ScheduledExecutorService` the `Scheduler` was created with. Reimplementation of the `ScheduledExecutorService` interface occurs purely in the `TestClock`.


jdegoes:
Excellent work. I think this is a direction we want to go. Please re-open if you have a chance to update!

PR #5046, opened 2021-05-06
ZLayer from constructor

notxcain:


jdegoes:
@notxcain Thank you for this!

@kitlangton Can you take a look?

kitlangton:
Would this feature be sufficient @notxcain? 👉 #4811 
Obviously, this macro would also work with classes and not just case classes. However, I think—as this is a white box macro—that IntelliJ would not be able to know the type of the generated layer without the help of a plugin.

jdegoes:
@notxcain Due to the lack of annotation macros in Scala 3, I think we should prefer options that are forward-migration compatible. I think a lot of the work we have done simplifying layer construction will address this issue. 

Please discuss further or re-open if you would like to push this through for 2.x only!

PR #5048, opened 2021-05-07
Implement ZStream.fromHubManaged

### LINKED ISSUE #5047: Enrich ZHub's documentation
I wanted to familiarise myself with the ZHub data type. While there are nice explanations of what the ZHub is and how it works, I couldn't find any usage examples. The only resources I found were the tests in ZHubSpec, which alone do not provide much information.

In specifically I was thinking I could do something like this:

```
for
    hub <- Hub.bounded[String](2)
    stream = ZStream.fromHub(hub)
     _ <- hub.publish("Hello")
     _ <- hub.publish("World")
     collected <- stream.runCollect
     _ <- ZIO.foreach(collected) { msg => console.putStrLn(msg) }
yield
     ()
```

But this program does not terminate.

adamgfraser:
Resolves #5047.

Currently `ZStream.fromHub` returns a stream directly, where the stream handles subscribing to the hub. While this creates a simple type signature, it can make it somewhat awkward in situations where we need to wait for a subscription to be acquired before performing further actions because we can't signal that the subscription has been completed. This PR changes the signature to return a managed effect producing a stream, where the outer managed effect describes subscribing to the hub and the inner stream describes taking elements from the hub.

Example before:

```scala
import zio._
import zio.stream._

object Example extends App {

  def run(args: List[String]): URIO[ZEnv, ExitCode] =
    for {
      promise <- Promise.make[Nothing, Unit]
      hub     <- Hub.bounded[String](2)
      stream = ZStream.managed(hub.subscribe).flatMap { queue =>
                 ZStream.fromEffect(promise.succeed(())) *>
                   ZStream.fromQueue(queue)
               }
      fiber     <- stream.take(2).runCollect.fork
      _         <- promise.await
      _         <- hub.publish("Hello")
      _         <- hub.publish("World")
      collected <- fiber.join
      _         <- ZIO.foreach(collected)(console.putStrLn(_)).orDie
    } yield ExitCode.success
}
```

And after:

```scala
import zio._
import zio.stream._

object Example extends App {

  def run(args: List[String]): URIO[ZEnv, ExitCode] =
    for {
      promise   <- Promise.make[Nothing, Unit]
      hub       <- Hub.bounded[String](2)
      stream     = ZStream.unwrapManaged(ZStream.fromHub(hub).tapM(_ => promise.succeed(())))
      fiber     <- stream.take(2).runCollect.fork
      _         <- promise.await
      _         <- hub.publish("Hello")
      _         <- hub.publish("World")
      collected <- fiber.join
      _         <- ZIO.foreach(collected)(console.putStrLn(_)).orDie
    } yield ExitCode.success
}
```

### DIFF
--- a/docs/datatypes/concurrency/hub.md
+++ b/docs/datatypes/concurrency/hub.md
@@ -341,7 +341,9 @@ object ZStream {
 }
 ```
 
-The managed effect here describes subscribing to receive messages from the hub while the stream describes taking messages from the hub. This can be useful when we need to ensure that a consumer has subscribed before a producer begins publishing values. Here is an example of using it:
+The managed effect here describes subscribing to receive messages from the hub while the stream describes taking messages from the hub. This can be useful when we need to ensure that a consumer has subscribed before a producer begins publishing values.
+
+Here is an example of using it:
 
 ```scala mdoc:reset:invisible
 import zio._
--- a/docs/datatypes/concurrency/hub.md
+++ b/docs/datatypes/concurrency/hub.md
@@ -325,14 +325,25 @@ import zio.stream._
 object ZStream {
   def fromHub[R, E, O](
     hub: ZHub[Nothing, R, Any, E, Nothing, O]
+  ): ZStream[R, E, O] =
+    ???
+}
+```
+
+This will return a stream that subscribes to receive values from a hub and then emits every value published to the hub while the subscription is active. When the stream ends the subscriber will automatically be unsubscribed from the hub.
+
+There is also a `fromHubManaged` operator that returns the stream in the context of a managed effect.
+
+```scala mdoc:nest
+object ZStream {
+  def fromHubManaged[R, E, O](
+    hub: ZHub[Nothing, R, Any, E, Nothing, O]
   ): ZManaged[Any, Nothing, ZStream[R, E, O]] =
     ???
 }
 ```
 
-The outer managed effect here describes subscribing to receive messages from the hub, while the inner stream describes taking messages from the hub. When the stream ends the subscriber will automatically be unsubscribed from the hub.
-
-Here is an example of using it:
+The managed effect here describes subscribing to receive messages from the hub while the stream describes taking messages from the hub. This can be useful when we need to ensure that a consumer has subscribed before a producer begins publishing values. Here is an example of using it:
 
 ```scala mdoc:reset:invisible
 import zio._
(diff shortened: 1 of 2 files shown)


adamgfraser:
@jdegoes Could you take a look at this when you get the chance? If we are going to make a change here I think it would be good to do it sooner rather than later.

iravid:
Just wanted to chime in - I think it's important to have this variant, but it's unfortunate to make it the shape of `fromHub`. 

It seems to me that in most cases, subscribing to a hub would be a runtime thing where it's not important to maintain ordering between consumers subscribing and producers starting to write. For example, when a consumer would subscribe when a websocket connection is opened; that would mean there's no inherent ordering between whoever is writing to the hub and whoever is reading from it.

So it would be great to keep `fromHub` as-is, with its straightforward type signature, and find another name for this variant.

Unfortunately I have no suggestions for such a name 😆

adamgfraser:
@iravid Yes I agree in a lot of real world situations the ordering doesn't matter and the current signature is really nice for dealing with something like a web socket connection. What about `fromHubManaged`? It creates the stream from a subscription to the hub in the context of a `ZManaged` to facilitate scope composition. Perhaps we could do the same thing with `broadcastDynamicManaged` which is the other place this comes up.

iravid:
@adamgfraser `fromHubManaged` sounds good to me!

docs/datatypes/concurrency/hub.md
```diff
@@ -297,12 +323,39 @@ We can create a `ZStream` from a subscription to a hub using the `fromHub` opera
 import zio.stream._
 
 object ZStream {
-  def fromHub[R, E, O](hub: ZHub[Nothing, R, Any, E, Nothing, O]): ZStream[R, E, O] =
+  def fromHub[R, E, O](
+    hub: ZHub[Nothing, R, Any, E, Nothing, O]
+  ): ZManaged[Any, Nothing, ZStream[R, E, O]] =
```
jdegoes, line 328: This should be updated.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3679,16 +3681,28 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
 
   /**
-   * Creates a stream from a subscription to a hub.
+   * Creates a stream from a subscription to a [[zio.ZHub]]. The outer managed
+   * effect describes subscribing to receive messages from the hub while the
+   * inner stream describes taking messages from the hub.
    */
-  def fromHub[R, E, A](hub: ZHub[Nothing, R, Any, E, Nothing, A]): ZStream[R, E, A] =
-    managed(hub.subscribe).flatMap(queue => fromQueue(queue))
+  def fromHub[R, E, A](
```
jdegoes, line 3688: I am onboard with keeping this one the same, and introducing `fromHubManaged` for what is (now) this one here (the one that returns the stream in a managed).

docs/datatypes/concurrency/hub.md
adamgfraser: Done.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Done.

jdegoes (approved):

PR #5049, opened 2021-05-08
Add test assertions for infinite/nan values

fsarradin:
I have added some assertions on types Float and Double in the test module.

I'm not that satisfied with assertion names, but they are the best that I've found.


adamgfraser (dismissed):
Agree with your feelings about the names but think they are the best there is.

fsarradin:
@adamgfraser I've brought a fix to this PR because `isFinite` is not defined in Scala 2.11. I've directly used the Java definition instead. Can you perform review again?

adamgfraser:
@fsarradin Will do today!

adamgfraser (approved):
Thank you! 🙏

PR #5054, opened 2021-05-09
add support for validatePar_ and validate_

phderome:
A possible purpose or use case is to execute jobs that are side-effecting (returning `Unit`) whose overall success value is not interesting but knowing about all failures remains useful. Proposed solution: call `.unit`

Solution: now leverages `partitionPar` as per @adamgfraser comment to avoid superfluous data construction of `BuildFrom`

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3355,6 +3355,13 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(res)(equalTo(in))
       }
     ),
+    suite("validatePar_")(
+      testM("returns all errors if never valid") {
+        val in                      = List.fill(10)(0)
+        def fail[A](a: A): IO[A, A] = IO.fail(a)
+        val res                     = IO.validatePar_(in)(fail).flip
+        assertM(res)(equalTo(in))
+      } @@ zioTag(errors)),
     suite("validateFirst")(
       testM("returns all errors if never valid") {
         val in  = List.fill(10)(0)
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3355,13 +3355,12 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(res)(equalTo(in))
       }
     ),
-    suite("validatePar_")(
-      testM("returns all errors if never valid") {
-        val in                      = List.fill(10)(0)
-        def fail[A](a: A): IO[A, A] = IO.fail(a)
-        val res                     = IO.validatePar_(in)(fail).flip
-        assertM(res)(equalTo(in))
-      } @@ zioTag(errors)),
+    suite("validatePar_")(testM("returns all errors if never valid") {
+      val in                      = List.fill(10)(0)
+      def fail[A](a: A): IO[A, A] = IO.fail(a)
+      val res                     = IO.validatePar_(in)(fail).flip
+      assertM(res)(equalTo(in))
+    } @@ zioTag(errors)),
     suite("validateFirst")(
       testM("returns all errors if never valid") {
         val in  = List.fill(10)(0)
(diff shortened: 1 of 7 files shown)


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -3355,6 +3355,13 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(res)(equalTo(in))
       }
     ),
+    suite("validatePar_")(
+      testM("returns all errors if never valid") {
+        val in                      = List.fill(10)(0)
+        def fail[A](a: A): IO[A, A] = IO.fail(a)
```
adamgfraser:
```suggestion
        def fail[A](a: A): IO[A, Nothing] = IO.fail(a)
```

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4012,6 +4012,14 @@ object ZIO extends ZIOCompanionPlatformSpecific {
       else ZIO.fail(::(es.head, es.tail.toList))
     }
 
+  /**
+   * @see See [[zio.ZIO.validatePar]]
+   */
+  def validatePar_[R, E, A, Collection[+Element] <: Iterable[Element]](in: Collection[A])(
+    f: A => ZIO[R, E, Any]
+  )(implicit bf: BuildFrom[Collection[A], Any, Collection[Any]], ev: CanFail[E]): ZIO[R, ::[E], Unit] =
+    validatePar(in)(f).unit
```
adamgfraser, line 4021: In the success case here there is no need to build the collection if we are going to immediately discard it. I think this will also let us to remove the `BuildFrom` constraint and accept any `Iterable` instead of `Collection[+Element] <: Iterable[Element]`, which will allow this operator to work with some other collection types like arrays and sets.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -3355,6 +3355,12 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(res)(equalTo(in))
       }
     ),
+    suite("validatePar_")(testM("returns all errors if never valid") {
+      val in                      = List.fill(10)(0)
+      def fail[A](a: A): IO[A, A] = IO.fail(a)
```
adamgfraser, line 3360:
```suggestion
      def fail[A](a: A): IO[A, Nothing] = IO.fail(a)
```

adamgfraser (changes requested):
Looks good! Do you want to also add versions for the other `validate` variants as well?

phderome:
> Looks good! Do you want to also add versions for the other `validate` variants as well?

Phil says: I may as well, yes, I didn't think of it.

core/shared/src/main/scala/zio/ZIO.scala
phderome: good point.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
phderome: addressed.

phderome:
all checks passed.

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Great improvement on the signature! We can still avoid unnecessarily creating the collection by implementing in terms of `partitionPar` directly and returning the unit value if there are no errors.

phderome: Right, addressed in latest commit.

phderome:
Should I squash commits on this PR?

adamgfraser (approved):
Thank you! 🙏

phderome:
thank you for the review.

adamgfraser:
Definitely! By the way I realized I didn't answer your question about the `_` naming. I think we would like to move away from the `_` naming for ZIO 2.0 but I can't say we have a better alternative at this point and we definitely have comparable methods with similar naming in ZIO 1.0 such as `foreachPar_` and `collectAllPar_` so I think the naming is correct for now, though could possibly be improved in the future.

Issue #5057, opened 2021-05-09
ZStream - Add the collectType operator to collect elements of a given subtype


remiguittaut:
Add the collectType operator to collect elements of a given subtype in ZStreams.

PR #5058, opened 2021-05-09
Fixing issue #5057 - ZStream collectType operator

remiguittaut:
Add the collectType operator to collect elements of a given subtype in ZStreams. The implementation must be defined as an extension method in an implicit class because of variance issues. The way to define the implicit class is up for discussion.


streams/shared/src/main/scala/zio/stream/ZStreamOps.scala
```diff
@@ -0,0 +1,13 @@
+package zio.stream
+
+import scala.reflect.ClassTag
+
+trait ZStreamOps {
+  implicit class Syntax[R, E, O](self: ZStream[R, E, O]) {
```
adamgfraser, line 6: You can just make this an implicit class inside the `ZStream` companion object.

remiguittaut: Ok. Wasn't sure I should touch the companion object. Will do

remiguittaut: Moved it to the companion object

iravid (approved):

remiguittaut:
> Thank you @remiguittaut! Would you like to add this to the new ZStream encoding as well? You can PR the implementation to series/2.x.

Sorry, long vacations spent moving to a new house, I'm completely off the OSS hook :). I'll make a new PR when I have time.

iravid:
No worries! Took care of that in #5190

PR #5062, opened 2021-05-10
Deflater

jkobejs:


streams/jvm/src/main/scala/zio/stream/experimental/Deflate.scala
```diff
@@ -0,0 +1,66 @@
...
+        }) { case (deflater, _) =>
+          ZIO.effectTotal(deflater.end())
+        }
+    } {
+      case (deflater, buffer) => {
+
+        def loop(): ZChannel[Any, Err, Chunk[Byte], Done, Err, Chunk[Byte], Done] =
+          ZChannel.readWithCause(
+            chunk =>
+              ZChannel.fromEffect {
```
iravid, line 33: Let's use `ZChannel.effectTotal` instead of `ZChannel.fromEffect(ZIO.effectTotal`. Should be a bit faster. Also below.

iravid (approved):
Awesome!

PR #5068, opened 2021-05-11
Implement ZStream#lock

adamgfraser:
I think we could use more direct support for controlling where a stream is run without having to drop back down to the level of `ZIO`. This PR implements a `lock` operator on `ZStream` to address that. The semantics are that it is called on a stream and shifts execution of that stream to the specified executor, but shifts back to the default executor for any continuations of that stream and after the stream is run.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -19,7 +19,7 @@ package zio.stream
 import zio._
 import zio.clock.Clock
 import zio.duration._
-import zio.internal.UniqueKey
+import zio.internal.{Executor, UniqueKey}
 import zio.stm.TQueue
 import zio.stream.internal.Utils.zipChunks
 import zio.stream.internal.{ZInputStream, ZReader}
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1946,6 +1946,17 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
 
+  /**
+   * Locks the execution of this stream to the specified executor. Any streams
+   * that are composed after this one will automatically be shifted back to the
+   * previous executor.
+   */
+  def lock(executor: Executor): ZStream[R, E, O] =
+    ZStream.fromEffect(ZIO.descriptor).flatMap { descriptor =>
+      ZStream.managed(ZManaged.lock(executor)) *>
+        self <* ZStream.fromEffect(ZIO.shift(descriptor.executor))
```
jdegoes, line 1957: Shouldn't we use some type of bracket here so in case there are failures, we shift back?

adamgfraser: The `ZManaged.lock` is already going to shift back if there is a failure, since that is a managed resource that describes shifting onto the specified executor and shifting back. What we need the `ZIO.shift` for is to make sure that if this stream is successful any continuation is shifted back to the original executor.

jdegoes: All right, makes sense!

jdegoes (approved):

PR #5079, opened 2021-05-13
Added find and findM for ZStream

Guisanpea:
Find is a really useful operation for all kind of collections. Its common in codebases to attempt to get the first element matching a predicate.

This pull request adds two methods. The fist is find returning the first element in a stream matching a predicate and the second findM which does it effectfully.

If I should add something else please tell me!

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1126,6 +1126,18 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     ZStream(self.process.ensuringFirst(fin))
 
   /**
+   * Yields the first element matching the provided predicate or `None` if no element satisfies the predicate
+   */
+  final def find(pred: O => Boolean): ZIO[R, E, Option[O]] =
+    self.filter(pred).runHead
+
+  /**
+   * Effecfully yields the first element matching the provided predicate or `None` if no element satisfies the predicate
+   */
+  final def findM[R1 <: R, E1 >: E](pred: O => ZIO[R1, E1, Boolean]): ZIO[R1, E1, Option[O]] =
+    self.filterM(pred).runHead
+
+  /**
    * Executes a pure fold over the stream of values - reduces all elements in the stream to a value of type `S`.
    */
   final def fold[S](s: S)(f: (S, O) => S): ZIO[R, E, S] =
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1125,6 +1125,18 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
   final def ensuringFirst[R1 <: R](fin: ZIO[R1, Nothing, Any]): ZStream[R1, E, O] =
     ZStream(self.process.ensuringFirst(fin))
 
+  /**
+   * Yields the first element matching the provided predicate or `None` if no element satisfies the predicate
+   */
+  final def find(pred: O => Boolean): ZIO[R, E, Option[O]] =
```
adamgfraser, line 1131: I think we want to make a strong distinction between operators that `run` the stream and operators that transform stream. Ideally we would like to have as many operators that transform streams as possible. In this case, I think that means the return type here should be `ZStream[R, E, O]` where the contract is that the stream will either emit a single element with the first element from the original stream satisfying the filter predicate or no elements otherwise. That way you could compose that element with other streaming operators or run it to a `ZIO` effect if you wanted.

Guisanpea: I agree with what you mention. It is important to differenciate methods that run the stream and methods that transform it.
However as far as I see it this method is mostly used as a terminating operation. Once you have terminated the operations you just want to find the first element that matches a predicate and just return it.
Although it is true that you could further transform it I don't think that it would be the most common scenario, and probably most of the transformation could be done as a ZIO.

So what about changing the signatures as `runFind` and `runFindM`? What do you think?

Of course this is just my opinion and I would appreciate any kind of response with yours 😄

adamgfraser: My personal opinion is that running the stream as part of the operator is not very compositional. Transforming and running the stream are conceptually different and ideally we would want to separate these concepts. That way you could always find an element satisfying a predicate and apply further transformations to that value or run it if you really wanted to terminate the stream with a final result.

iravid:
We're aiming to prefix all methods that translate a stream to a ZIO value with `run*`.

In this case, I think we could just have `find(p: O => Boolean): ZStream[R, E, O]` like @adamgfraser suggested, and users could use `runHead` on that.

iravid:
Superseded by #5116. Thank you @Guisanpea!

PR #5080, opened 2021-05-13
Remove Blocking From The Environment

### LINKED ISSUE #5076: Remove Blocking from ZEnv / environment / standard services
There is almost never a need to use a separate blocking pool. We should just bake one global blocking pool into the application, move all the blocking operators to `ZIO.*` (and companion objects), and delete the `Blocking` service and package.

adamgfraser:
Resolves #5076.

Removes the `Blocking` service from the environment. Blocking is now managed by a collection of operators that lock the effect to an `Executor` backed by a blocking thread pool. Right now this is just a global blocking thread pool but we could potentially allow each fiber to have a blocking executor as well as a normal executor associated with it. Then we could provide a variant of `shift` that allowed setting that blocking executor to a new value. This would allow setting the blocking executor to a new value within a scope.


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -2719,6 +2732,111 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+   *
+   * Note that this adds significant overhead. For performance sensitive
+   * applications consider using `effectBlocking` or
+   * `effectBlockingCancel`.
+   */
+  def effectBlockingInterrupt[A](effect: => A): Task[A] =
+    // Reference user's lambda for the tracer
+    ZIOFn.recordTrace(() => effect) {
+      ZIO.effectSuspendTotal {
+        import java.util.concurrent.atomic.AtomicReference
```
regiskuckaertz, line 2772: Is this fine in `shared`?

core/jvm/src/main/scala/zio/internal/Blocking.scala
```diff
@@ -0,0 +1,27 @@
+package zio.internal
+
+import java.util.concurrent.{SynchronousQueue, ThreadPoolExecutor, TimeUnit}
+
+object Blocking {
+
+  val blockingExecutor: Executor =
```
regiskuckaertz, line 7: lazy val perhaps

core/shared/src/main/scala/zio/ZIO.scala
regiskuckaertz: ah yeah looks like it https://github.com/scala-js/scala-js/blob/8a0655b2ddf3047c13d49814c4debb7f10b88931/javalanglib/src/main/scala/java/lang/Thread.scala

regiskuckaertz (approved):

jdegoes (approved):
Excellent work! Going to clean things up for a lot of use cases.

PR #5081, opened 2021-05-13
Gzip channel

jkobejs:
Related to https://github.com/zio/zio/issues/4886

I haven't ported this test https://github.com/zio/zio/blob/master/streams-tests/jvm/src/test/scala/zio/stream/compression/CompressionSpec.scala#L266

I'm not sure what would reusable mean in terms of a channel. 

I'm using Gzipper implementation from non experimental compression package https://github.com/zio/zio/blob/master/streams/jvm/src/main/scala/zio/stream/compression/Gzipper.scala

Should I move that implementation in experimental package?

### DIFF
--- a/streams/jvm/src/main/scala/zio/stream/experimental/Gzip.scala
+++ b/streams/jvm/src/main/scala/zio/stream/experimental/Gzip.scala
@@ -1,0 +1,39 @@
+package zio.stream.experimental
+
+import zio.stream.compression.{CompressionLevel, CompressionStrategy, FlushMode, Gzipper}
+import zio.{Chunk, ZIO, ZManaged}
+
+object Gzip {
+  def makeGzipper[Err, Done](
+    bufferSize: Int = 64 * 1024,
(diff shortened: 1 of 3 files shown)


jdegoes:
@iravid Ping!

iravid (approved):
Great stuff. Good to see that we could just reuse `Gzipper`.

streams/jvm/src/main/scala/zio/stream/experimental/Gzip.scala
```diff
@@ -0,0 +1,39 @@
...
+      ZManaged
+        .make(
+          Gzipper.make(bufferSize, level, strategy, flushMode)
+        ) { gzipper =>
+          ZIO.effectTotal(gzipper.close())
+        }
+    } {
+      case gzipper => {
+
+        def loop(): ZChannel[Any, Err, Chunk[Byte], Done, Err, Chunk[Byte], Done] =
```
jdegoes, line 23: This will be much more efficient as a `lazy val`.

PR #5082, opened 2021-05-13
Javatime generators

igrebenik:
Add  `ZIO Test` generators for the following java.time types:
- DayOfWeek
- LocalDate
- LocalTime
- Month
- MonthDay
- OffsetTime
- Period
- Year
- YearMonth
- ZonedDateTime
- ZoneId
- ZoneOffset

### DIFF
--- a/test/shared/src/main/scala/zio/test/TimeVariants.scala
+++ b/test/shared/src/main/scala/zio/test/TimeVariants.scala
@@ -67,7 +67,7 @@ trait TimeVariants {
       hour   <- Gen.int(0, 23)
       minute <- Gen.int(0, 59)
       second <- Gen.int(0, 59)
-      nanos  <- Gen.int(0, 999_999_999)
+      nanos  <- Gen.int(0, 999999999)
     } yield LocalTime.of(hour, minute, second, nanos)
 
   /**
--- a/test/shared/src/main/scala/zio/test/TimeVariants.scala
+++ b/test/shared/src/main/scala/zio/test/TimeVariants.scala
@@ -39,15 +39,6 @@ trait TimeVariants {
     )
 
   /**
-   * A generator of `java.time.Duration` values. Shrinks toward `Duration.of(Long.MinValue, 0)`
-   */
-  final def anyDuration: Gen[Random, Duration] =
-    for {
-      seconds <- Gen.anyLong
-      nanos   <- Gen.long(0, 999999999L)
-    } yield java.time.Duration.ofSeconds(seconds, nanos)
-
-  /**
    * A generator of finite `zio.duration.Duration` values. Shrinks toward `Duration.Zero`.
    */
   final def anyFiniteDuration: Gen[Random, Duration] = Gen.long(0L, Long.MaxValue).map(Duration.Finite(_))
@@ -58,17 +49,18 @@ trait TimeVariants {
   final def anyInstant: Gen[Random, Instant] = instant(Instant.MIN, Instant.MAX)
 
   /**
-   * A generator of `java.time.LocalDate` values. Shrinks toward `LocaDate.MIN`
+   * A generator of `java.time.LocalDate` values. Shrinks toward `LocalDate.MIN`.
    */
   final def anyLocalDate: Gen[Random, LocalDate] =
     for {
       year  <- anyYear
       month <- Gen.int(1, 12)
-      day   <- Gen.int(1, Month.of(month).maxLength())
+      maxLen = if (!year.isLeap && month == 2) 28 else Month.of(month).maxLength
+      day   <- Gen.int(1, maxLen)
     } yield LocalDate.of(year.getValue, month, day)
 
   /**
-   * A generator of `java.time.LocalTime` values. Shrinks toward `LocalTime.MIN`
+   * A generator of `java.time.LocalTime` values. Shrinks toward `LocalTime.MIN`.
    */
   final def anyLocalTime: Gen[Random, LocalTime] =
     for {
@@ -103,7 +95,7 @@ trait TimeVariants {
     )
 
   /**
-   * A generator of `java.time.MonthDay` values. Shrinks toward `MonthDay.of(Month.JANUARY, 1)`
+   * A generator of `java.time.MonthDay` values. Shrinks toward `MonthDay.of(Month.JANUARY, 1)`.
    */
   final def anyMonthDay: Gen[Random, MonthDay] =
     for {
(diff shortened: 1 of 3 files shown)


adamgfraser:
@igrebenik Let me know when you are ready for a review on this!

jdegoes:
@igrebenik 

Looks like some Scala-specific version failures, e.g.:

```
[error] /home/runner/work/zio/zio/test/shared/src/main/scala/zio/test/TimeVariants.scala:23:14: object jdk is not a member of package scala
[error] import scala.jdk.CollectionConverters._
[error]              ^
[error] /home/runner/work/zio/zio/test/shared/src/main/scala/zio/test/TimeVariants.scala:158:48: value asScala is not a member of java.util.Set[String]
[error]     val zoneIds = ZoneId.getAvailableZoneIds().asScala
[error]                                                ^
[error] /home/runner/work/zio/zio/test/shared/src/main/scala/zio/test/TimeVariants.scala:159:23: value range is not a member of object scala.collection.immutable.Set
[error]     val idMap   = Set.range(0, zoneIds.size).zip(zoneIds).toMap
```

igrebenik:
@jdegoes @adamgfraser  Right! Actually,  30 checks have failed. Well, I've never had a chance to support multiple versions of Scala let alone different platforms!!! gotta figure out how do it properly)

igrebenik:
@adamgfraser ready for the review)

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -154,12 +155,18 @@ trait TimeVariants {
...
-    Gen.int(0, zoneIds.size).noShrink.map(idx => ZoneId.of(idMap(idx)))
-  }
+  final def anyZoneId: Gen[Random, ZoneId] =
+    Gen(
+      ZStream
+        .fromJavaIteratorTotal(ZoneId.getAvailableZoneIds.iterator)
+        .mapChunksM { ids =>
+          for {
+            random   <- ZIO.service[Random.Service]
+            shuffled <- random.shuffle(ids)
```
igrebenik, line 165: Hmm, not sure how fast it is

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -16,13 +16,29 @@
...
-import java.time.{Instant, LocalDateTime, OffsetDateTime, ZoneOffset}
+import java.time._
 
 trait TimeVariants {
 
+  /**
+   * A generator of `java.time.DayOfWeek` values. Doesn't have any shrinking.
+   */
+  final def anyDayOfWeek: Gen[Random, DayOfWeek] =
+    Gen.oneOf(
```
adamgfraser, line 32: You can use `Gen.elements` here to avoid constructing another layer of generators.

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -33,16 +49,133 @@ trait TimeVariants {
...
   /**
    * A generator of `java.time.LocalDateTime` values. Shrinks toward `LocalDateTime.MIN`.
    */
   final def anyLocalDateTime: Gen[Random, LocalDateTime] = localDateTime(LocalDateTime.MIN, LocalDateTime.MAX)
 
+  /**
+   * A generator of `java.time.Month` values. Doesn't have any shrinking.
+   */
+  final def anyMonth: Gen[Random, Month] =
+    Gen.oneOf(
```
adamgfraser, line 83: Same comment as above regarding using `Gen.elements`.

test/shared/src/main/scala/zio/test/TimeVariants.scala
adamgfraser: I think you can just use `Gen.elements` here as well once you get the available `ZoneId` values.

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -16,13 +16,29 @@
...
 import zio.random.Random
+import zio.stream.ZStream
 
-import java.time.{Instant, LocalDateTime, OffsetDateTime, ZoneOffset}
+import java.time._
 
 trait TimeVariants {
 
+  /**
+   * A generator of `java.time.DayOfWeek` values. Doesn't have any shrinking.
```
adamgfraser, line 29: This will shrink towards the first value, in this case `MONDAY`, which I think is fine.

adamgfraser (changes requested):
Left a few comments on the implementation but otherwise looks good. Thanks for adding!

test/shared/src/main/scala/zio/test/TimeVariants.scala
igrebenik: Yeah, but I would need to convert those `ZoneId` values into Scala collection types such as `List` for instance and Scala `CollectionConverters` differ from version to version (at least, they are in different packages if we compare Scala 2.11 and Scala 2.13). 
That's why I decided to take this approach to avoid conversion between Java collections and Scala collections entirely. Otherwise that would make sense to use `Gen.elements`)

test/shared/src/main/scala/zio/test/TimeVariants.scala
igrebenik: Right! I've taken a look at `DayOfWeek` values and indeed `MONDAY` goes first. That's why I put those in the same order.

igrebenik: So it makes sense to shrink toward `MONDAY`!

test/shared/src/main/scala/zio/test/TimeVariants.scala
igrebenik: Do I need to implement anyZoneId for each Scala version (in order to use appropriate CollectionConverters) and remove it from share?

igrebenik:
@adamgfraser could you check the changes I made?

test/shared/src/main/scala/zio/test/TimeVariants.scala
```diff
@@ -33,16 +49,133 @@ trait TimeVariants {
...
+  final def anyZonedDateTime: Gen[Random, ZonedDateTime] =
+    for {
+      dateTime <- anyLocalDateTime
+      zoneId   <- anyZoneId
+    } yield ZonedDateTime.of(dateTime, zoneId)
+
+  /**
+   * A generator of `java.time.ZoneId` values. Doesn't have any shrinking.
+   */
+  final def anyZoneId: Gen[Random, ZoneId] =
```
adamgfraser, line 158: It looks like we have separate versions specific implementations of this that seem to be identical and then this version. I think we should be able to just have one implementation of this that is like the version specific ones currently and just uses `import scala.collection.JavaConverters._` and `@silent("JavaConverters")`.

adamgfraser (changes requested):
Something still seems off about the `anyZoneId` implementation but otherwise looks good!

test/shared/src/main/scala/zio/test/TimeVariants.scala
igrebenik: Oh, didn't know about `@silent` annotation! Thank you Adam! 
Will do as you suggested)

igrebenik: Oh my... I forgot to remove `anyZoneId` from `TimeVariants` when it was implemented using `ZStream` This looks somewhat ugly(

igrebenik: @adamgfraser done!

adamgfraser (approved):

PR #5100, opened 2021-05-15
Support Combining Specs

### LINKED ISSUE #5092: ZIO Test: Add binary composition operator to Spec
Currently, specs cannot compose directly, although if you have a sweet, you can combine all the children into a spec via the suite itself.

As a consequence of this design choice, it is not possible to have a binary composition operator on `Spec`, which composes two specs into one. 

Yet, if two specs could compose, then it would be simpler to break up mega-suites into smaller subsuites (e.g. `val bigSuite = littleSweet + otherSuiite + yetAnotherSuite`), without having to unnecessarily nest them inside another suite just for purposes of composition. 

In addition, if specs could compose together, then it would be possible to improve upon the syntax of test composition without `MutableRunnableSpec`:

```scala
suite("LayerSpec") {
  test("horizontal compose") {
  } + 
  test ("vertical compose") {
  } + 
  ...
}
```

(The `+` operator is a strawman proposal and anything could be chosen instead.)

This ticket is to add a binary composition operator to `Spec`:

```scala
final case class Spec[-R, +E, +T](caseValue: SpecCase[R, E, T, Spec[R, E, T]]) { self =>
  def + [R1 <: R, E1 >: E, T1 >: T](that: Spec[R1, E1, T1]): Spec[R1, E1, T1] = ...
}
```

This must be implemented by adding a new subtype to `SpecCase`, which can store both left and right hand specs, and then adding support for this new subtype in all cases where `SpecCase` is pattern-matched.

Possibly at the same time, we could refactor the `SpecCase` type to be a bit more modular, maybe something like:

```scala
final case class LabeledCase[+Spec](label: String, spec: Spec) extends SpecCase[Any, Nothing, Nothing, Spec]
final case class MultipleCase[+Spec](specs: Vector[Spec]) extends SpecCase[Any, Nothing, Nothing, Spec]
final case class ManagedCase[-R, +E, +Self](managed: ZManaged[R, E, Nothing, Self) 
  extends SpecCase[R, E, Nothing, Spec]
final case class TestCase[-R, +E, +T](test: ZIO[R, E, T], annotations: TestAnnotationMap)
  extends SpecCase[R, E, T, Nothing]
```

This way you pay for just the features you need.

adamgfraser:
Initial work to resolve #5092. New subtypes are implemented and working on pushing the changes through the various operators that use the structure of `Spec`.


hmemcpy:
I feel this is going to be fun fixing for series/2.0 :P

adamgfraser:
Spent some time on this and I think there are a couple of potential issues here.

First, with the more modular encoding we no longer statically know that if we have a test there is at least one label for it. That is an issue if we have a test failure or need to report something else like a test that is not terminating with `TestAspect.diagnose` we may not have any meaningful information to report about where the error is occurring. In practice all tests we construct do have a label but the fact that we no longer know this at the type level is not great.

Second, we no longer have the ability to control the execution strategy granularly. To address this I think we may need to do something like:

```scala
final case class ExecCase(exec: ExecutionStrategy) extends SpecCase[Any, Nothing, Nothing, Spec]
final case class LabeledCase[+Spec](label: String, spec: Spec) extends SpecCase[Any, Nothing, Nothing, Spec]
final case class MultipleCase[+Spec](specs: Vector[Spec]) extends SpecCase[Any, Nothing, Nothing, Spec]
final case class ManagedCase[-R, +E, +Self](label: String, managed: ZManaged[R, E, Nothing, Self) 
  extends SpecCase[R, E, Nothing, Spec]
final case class TestCase[-R, +E, +T](label: String, test: ZIO[R, E, T], annotations: TestAnnotationMap)
  extends SpecCase[R, E, T, Nothing]
```

There is a little bit of duplication here because the label can now exist in both the test and managed case as well as the labeled case but this way we know that if we ever have a failure in a test or a managed resource we have at least one label to associate with it.

@jdegoes What do you think?

jdegoes (approved):

PR #5102, opened 2021-05-16
ZIO Streams documentation

khajavi:
- [X] Introduction
- [X] ZStream
- [X] Stream
- [x] ZSink
- [X] Sink
- [x] Transducer
- [x] Type Aliases

### DIFF
--- a/docs/datatypes/stream/zstream.md
+++ b/docs/datatypes/stream/zstream.md
@@ -1418,6 +1418,36 @@ val stream: ZIO[Console with Random with Clock, IOException, Unit] =
       case _ => ZIO.dieMessage("unhandled case")
     }
 ```
+### Distribution
+
+The `ZStream#distributedWith` operator is a more powerful version of `ZStream#broadcast`. It takes a `decide` function, and based on that decide how to distribute incoming elements into the downstream streams:
+
+```scala
+abstract class ZStream[-R, +E, +O] {
+  final def distributedWith[E1 >: E](
+    n: Int,
+    maximumLag: Int,
+    decide: O => UIO[Int => Boolean]
+  ): ZManaged[R, Nothing, List[Dequeue[Exit[Option[E1], O]]]] = ???
+}
+```
+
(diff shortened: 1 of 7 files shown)


docs/datatypes/stream/zstream.md
```diff
@@ -0,0 +1,1865 @@
...
+```scala mdoc:silent:nest
+val schedule: Schedule[Clock with Random, Chunk[Chunk[Record]], Long] =
+  // Start off with 30-second timeouts as long as the batch size is < 1000
+  Schedule.fixed(30.seconds).whileInput[Chunk[Chunk[Record]]](_.flatten.length < 100) andThen
+    // and then, switch to a shorter jittered schedule for as long as batches remain over 1000
+    Schedule.fixed(5.seconds).jittered.whileInput[Chunk[Chunk[Record]]](_.flatten.length >= 1000)
+    
+dataStream
+  .aggregateAsyncWithin(ZTransducer.collectAllN(2000), schedule)
+```
```
khajavi, line 1660: Hello @iravid 
Yesterday all the time, I was working on the adaptive batching pattern which you have presented in one of your talk about ZIO streams.
I tried to rewrite and test that and and then include this nice use-case in ZIO stream documentation. But I'm not sure that this example works as what you presented.
When I debug that program, I found that in the `_.flatten.length < 100` predicate, the value of `_.flatten.length` does not always represent the batch size. Most of the time, its value is zero and sometime it is `2000`. I didn't find a situation when the size of the batch is is something between 0 and 2000. (I also jittered the upstream to make sure I have different throughout)
Could you please look at that and check if I'm right or not?

khajavi:
Hi @jdegoes 
It is ready to be reviewed.

docs/datatypes/stream/zstream.md
iravid: Hey! Will try to take a look later today

jdegoes (approved):

jdegoes:
@khajavi Thank you! It's a huge PR and the quality of the material looks good. I think we can do followup PRs to address improvements and fixes.

PR #5105, opened 2021-05-17
Add ZIO#tapEither and tests

2ne1ugly:
I came across of a use case where I need to report the result of effect to slack.
`tapBoth` was the first option but it simply required me to write a function or duplicate codes.
Through `tapEither`, we can execute an effect with the execution result information in one lambda.

Similar function exists in [typelevel/cats/MonadError](https://github.com/typelevel/cats/blob/adf6680362224509d0b2a50c40bb6e59a32dc45b/core/src/main/scala/cats/MonadError.scala#L104)

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3175,8 +3175,8 @@ object ZIOSpec extends ZIOBaseSpec {
         for {
           ref <- Ref.make(0)
           _ <- Task(42).tapEither {
-                 case Left(value) => ref.set(-1)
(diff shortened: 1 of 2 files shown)


jdegoes:
@2ne1ugly Looks great! If we merge here, can you cherry-pick to the `series/2.x` branch? 🙏

mingyuchi:
> @2ne1ugly Looks great! If we merge here, can you cherry-pick to the `series/2.x` branch? 🙏

Sure, I will do that tomorrow (about 14 hours later from now)

jdegoes:
@mingyang91 Thank you!

jdegoes (approved):

PR #5128, opened 2021-05-19
Implement ! As Symbolic Operator For orDie

adamgfraser:


neko-kai:
^ disregard that, misclicked while reading notification

notxcain:
What are the use cases for that?

jdegoes (approved):

PR #5130, opened 2021-05-20
Implement Accessor trait for simple macro-less accessor methods

kitlangton:
A simple, macro-less means of creating accessors from Services. Extend the companion object with `Accessor[ServiceName]`, then simply call`Companion(_.someMethod)`, to return a ZIO effect that requires the Service in its environment.

Example: 
```scala
trait FooService {
  def magicNumber: UIO[Int]
  def castSpell(chant: String): UIO[Boolean]
}

object FooService extends Accessor[FooService]

object Example {
  val program: ZIO[Has[FooService], Nothing, String] = for {
    int  <- FooService(_.magicNumber)
    bool <- FooService(_.castSpell("Oogabooga!"))
  } yield s"$int and $bool"
}
```

### DIFF
--- a/core/shared/src/main/scala/zio/Accessor.scala
+++ b/core/shared/src/main/scala/zio/Accessor.scala
@@ -1,0 +1,28 @@
+package zio
+
+/**
+ * A simple, macro-less means of creating accessors from Services. Extend
+ * the companion object with `Accessor[ServiceName]`, then simply call
+ * `Companion(_.someMethod)`, to return a ZIO effect that requires the
+ * Service in its environment.
+ *
+ * Example:
+ * {{{
+ *   trait FooService {
+ *     def magicNumber: UIO[Int]
+ *     def castSpell(chant: String): UIO[Boolean]
+ *   }
+ *
+ *   object FooService extends Accessor[FooService]
+ *
+ *   val example: ZIO[Has[FooService], Nothing, String] =
+ *     for {
+ *       int  <- FooService(_.magicNumber)
+ *       bool <- FooService(_.castSpell("Oogabooga!"))
+ *     } yield s"$int and $bool"
+ * }}}
+ */
+trait Accessor[R] {
+  def apply[E, A](f: R => ZIO[Any, E, A])(implicit tag: Tag[R]): ZIO[Has[R], E, A] =
+    ZIO.serviceWith[R](f)
+}
--- a/core/shared/src/main/scala/zio/Accessor.scala
+++ b/core/shared/src/main/scala/zio/Accessor.scala
@@ -15,11 +15,11 @@ package zio
  *
  *   object FooService extends Accessor[FooService]
  *
- *   val example: ZIO[Has[FooService], Nothing, String] =
+ *   val example: ZIO[Has[FooService], Nothing, Unit] =
  *     for {
  *       int  <- FooService(_.magicNumber)
  *       bool <- FooService(_.castSpell("Oogabooga!"))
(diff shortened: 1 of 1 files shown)


adamgfraser (dismissed):

adamgfraser (approved):

core/shared/src/main/scala/zio/Accessor.scala
```diff
@@ -0,0 +1,28 @@
...
+ *   object FooService extends Accessor[FooService]
+ *
+ *   val example: ZIO[Has[FooService], Nothing, Unit] =
+ *     for {
+ *       int  <- FooService(_.magicNumber)
+ *       bool <- FooService(_.castSpell("Oogabooga!"))
+ *     } yield ()
+ * }}}
+ */
+trait Accessor[R] {
```
jdegoes, line 25: @adamgfraser @kitlangton 

My only potential concern with this was the name `Accessor`, which might be pretty common.

The other potential improvement we could make: warn if the user tries to use this on an effect that requires an environment.

```scala
trait Accessor[R] {
   def apply[R2, E, A](f: R => ZIO[R2, E, A])(implicit tag: Tag[R], isAny: Accessor.IsAny[R2]): ZIO[Has[R], E, A] = ...
}
object Accessor {
  @implicitNotFound("The methods of your service definition should not use the environment, because this leaks implementation details to clients of the service, and these implementation details should be hidden and free to change based on the specific nature of the implementation. In order to use this accessor, please consider refactoring your service methods so they no longer use ZIO environment.")
  sealed trait IsAny[R]
  implicit val anyIsAny: IsAny[Any] = new IsAny{}
```

Finally, we should probably extend all companion objects of our own services by this trait.

jdegoes: On trait name:

 - `ServiceCompanion`
 - `ServiceAccessor`
  - ???

It could also be useful to think of other utilities which could be placed into the service companion object.

For example:

```scala
def makeLayer[R2, E](zio: ZIO[R2, E, R])(implicit tag: Tag[R]): ZLayer[R2, E, Has[R]] = zio.toLayer
```

I am not sure if this utility pays for itself, but right now, using `toLayer` sometimes fails because of the invariance of `Has`, requiring you to widen to the service in the final `yield` of the `for` comprehension that builds the effect, e.g.:

```scala
(for {
 a <- b 
 c <- d 
} yield ServiceLive(a, c): Service).toLayer
```

In the companion object, you have access to the type `R` of the service so you can add utilities that fix it.

This could be changed to:

```scala
Service.makeLayer {
  for {
    a <- b 
    c <- d 
  } yield ServiceLive(a, c)
}
```

If not this one, we should look at opportunities to introduce other utilities that require the service type to be known.

/cc @kitlangton @adamgfraser

adamgfraser: @jdegoes Yes. If we just want to use it for these accessors then perhaps we could do `Accessible` which I imagine would be less common but if we want to have other generic utilities related to services then something like `ServiceCompanion` seems more appropriate.

There is a little risk that we are creating two ways to do things here (e.g. `Clock.nanotime` versus `Clock(_.nanotime)`) so could be good to think about our messaging around this. The `Clock.nanotime` reads really nicely like it is just a method on a static object but does require a little more work to implement the accessors. Is this the "not quite as nice" shortcut for when you don't want to do that? Is it a replacement for the accessors?

kitlangton: Made changes: #5133

PR #5131, opened 2021-05-20
Implement ZIO.debug

adamgfraser:
Static version of `debug` for when you just want to print something to the console for debugging purposes without introducing additional environmental requirements or error types.


jdegoes (approved):

hmemcpy:
RIP `UIO(println("message")` :)

PR #5133, opened 2021-05-20
Accessible trait rename and implicit hints

kitlangton:
Updated regarding the feedback on #5130


adamgfraser (approved):
Looks good to me!

kitlangton:
Good to merge?

PR #5134, opened 2021-05-20
Move Retrying and Repetition Logic to Clock

### LINKED ISSUE #5132: Move logic of retrying and repetition to Clock service
Without changing the current API, we should move the retrying, repetition, and scheduling logic (embodied by such methods as repeat, repeatOrElse, etc.) into the `Clock` service.

This will solve a common anti-pattern, whereby middleware in application that uses `Clock` via this retrying, repetition, or scheduling logic (but must, itself, provide `R = Any` in its service definition) must `provideLayer` clock on every method invocation.

After this change, the API will not change but the `Clock` trait will become bigger, having more methods that implement the raw scheduling logic.

adamgfraser:
Resolves #5132.


jdegoes (approved):
So great to get this logic out of ZIO and into `Clock`!

PR #5143, opened 2021-05-22
clarify ZLayer.identity (in relation to >>> as opposed to ++)

phderome:
title says it.

### DIFF
--- a/core/shared/src/main/scala/zio/ZLayer.scala
+++ b/core/shared/src/main/scala/zio/ZLayer.scala
@@ -2166,7 +2166,8 @@ object ZLayer {
     ZLayer(m)
 
   /**
-   * An identity layer that passes along its inputs.
+   * An identity layer that passes along its inputs. Represents an identity with respect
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/ZLayer.scala
```diff
@@ -2166,7 +2166,8 @@ object ZLayer {
     ZLayer(m)
 
   /**
-   * An identity layer that passes along its inputs.
+   * An identity layer that passes along its inputs. Represents an identity with respect
+   * to operator >>>. User may choose to use requires in conjunction with operator ++.
```
adamgfraser, line 2170:
```suggestion
   * An identity layer that passes along its inputs. Note that this represents
   * an identity with respect to the `>>>` operator. It represents an identity
   * with respect to the `++` operator when the environment type is `Any`.
```

adamgfraser (changes requested):
Looks good. Suggested slightly changing the wording to try to make it a little more clear.

adamgfraser (approved):

PR #5145, opened 2021-05-24
Fix IO operations returning Tasks

jrglee:
Some of the `IO` operations such as `apply` and `effect` were returning `Task`, effectively setting the error as `Throwable` instead of `Any`. This caused some confusion when working with error hierarchies that does not descend from `Throwable` or `Exception`. This fix corrects the return types in `IO.scala` and follows up in all compiler related issues.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1978,7 +1978,7 @@ object ZIOSpec extends ZIOBaseSpec {
       },
       testM("effect, bind, map") {
         def fibIo(n: Int): Task[BigInt] =
-          if (n <= 1) IO.effect(n)
+          if (n <= 1) Task.effect(n)
           else
             for {
               a <- fibIo(n - 1)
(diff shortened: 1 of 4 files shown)


adamgfraser:
@jrglee I don't think this change is correct. Operators on the companion objects for type aliases should always have the same return type as the operator on `ZIO` or a more specific type if possible. Having `IO.effect` return `Any` for the error type violates this property. Returning `Any` as the error type also throws away all type information on the error type type and is not very useful. The error type can be automatically widened to `Any` if unrelated errors are unified anyway.

Perhaps if you could talk more about the issue you are dealing with we could discuss other ways to use existing functionality or add new functionality to address them but I do not think this is the right approach.

This is also a binary incompatible change that would have to be made in the `series/2.x` if we were to do anything at all.

jrglee:
@adamgfraser Thanks for replying. I was not totally sure that `IO.apply` had to behave like `ZIO.apply`. It was confusing to  to use `IO.apply` and have it calling `Task.apply`, but I think I should have been using `IO.succeed` or `IO.fail` instead. The `ZIO.apply` and `Task.apply` semantics makes sense because they can catch exceptions.

I think I can close this PR.

PR #5147, opened 2021-05-26
Improve Support For Layers Run For Their Effects

adamgfraser:
We are seeing more uses for layers provided purely for their effects (e.g starting some background process). Ideally we would like these layers to produce a value of type `Any` but this does not compose well with many existing operators that expect the environment to be a subtype of `Has`. We already have the tool to deal with this in the form of the `Has.Union` abstraction but we are not using this as broadly as we could be.

This PR addresses that by replacing constraints that the environment is a subtype of `Has` with constraints that an instance of `Has.Union` exists. It also lifts the requirement that layers cannot be provided to an effect with an environment type of `Any` since it can still make sense to provide a layer for its effects.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/NeedsEnvSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/NeedsEnvSpec.scala
@@ -23,7 +23,7 @@ object NeedsEnvSpec extends ZIOBaseSpec {
             import zio._
             import zio.console._
             val uio = UIO.succeed("Hello, World!")
-            uio.provideLayer(Console.live)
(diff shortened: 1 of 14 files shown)


jdegoes (approved):
Big improvement. There may be some other places we missed. We should do a review before 2.0.

adamgfraser:
I looked for all occurrences of `<: Has[_]` and `NeedsEnv`. I think the only area that is not covered by these changes is the new functionality for automatically building layers which I didn't want to mess with here.

@kitlangton Could you take a look at that when you get the chance and how we can incorporate similar changes into the automatic layer functionality.

PR #5154, opened 2021-05-28
Add Additional Operators to NonEmptyChunk

adamgfraser:
Adds a few additional operators that preserve the nonempty property.


kitlangton:
Would this also cover prepended / appended

adamgfraser:
It does now. :smiley:

jdegoes (approved):

PR #4910, opened 2021-04-07
Remove type aliases, hoist traits out of package objects, and miscellaneous naming changes

kitlangton:
- Remove `Service` type aliases
- Lift `clock.Clock` (and other ZEnv types) directly to the `zio` package (e.g., `zio.Clock`).
- Lift `duration._` directly to the `zio` package
- Rename `Console.putStrLn` -> `Console.printLine`

PR #5171, opened 2021-06-03
Update Documentation Regarding Module Pattern

adamgfraser:
I think it is most idiomatic to define layers for a service in the service interface companion object. The user should really be interacting with the service exclusively through the service interface. The fact that the live implementation is defined as a case class is really an implementation detail and arguably the live implementation should not even be visible to the user of the interface.

### DIFF
--- a/docs/datatypes/contextual/index.md
+++ b/docs/datatypes/contextual/index.md
@@ -339,7 +339,7 @@ case class LoggingLive(console: Console.Service, clock: Clock.Service) extends L
 }
 ```
 
-4. **Defining ZLayer** — Now, we create a companion object for `LoggingLive` data type and lift the service implementation into the `ZLayer`:
+4. **Defining ZLayer** — Now, we create a companion object for the `Logging` service interface and lift the service implementation into a `ZLayer`:
 
 ```scala mdoc:silent
 object Logging {
--- a/docs/datatypes/contextual/index.md
+++ b/docs/datatypes/contextual/index.md
@@ -329,7 +329,6 @@ trait Logging {
 ```scala mdoc:silent
 import zio.console.Console
 import zio.clock.Clock
-
 case class LoggingLive(console: Console.Service, clock: Clock.Service) extends Logging {
   override def log(line: String): UIO[Unit] = 
     for {
@@ -356,8 +355,7 @@ object Logging {
   val live: URLayer[Has[Console.Service] with Has[Clock.Service], Has[Logging]] =
     (LoggingLive(_, _)).toLayer
 
-  def log(line: String): URIO[Has[Logging], Unit] =
-    ZIO.serviceWith[Logging](_.log(line))
+  def log(line: String): URIO[Has[Logging], Unit] = ZIO.serviceWith(_.log(line))
 }
 ```
 
@@ -520,15 +518,14 @@ trait Logging {
 }
 
 object Logging {
-  def log(line: String) = ZIO.serviceWith[Logging](_.log(line))
-}
 
-object LoggingLive {
-  val layer: ULayer[Has[Logging]] = ZLayer.succeed {
+  val live: ULayer[Has[Logging]] = ZLayer.succeed {
     new Logging {
       override def log(str: String): UIO[Unit] = ???
     }
   }
+
+  def log(line: String) = ZIO.serviceWith[Logging](_.log(line))
 }
 
 val myApp: ZIO[Has[Logging] with Console with Clock, Nothing, Unit] = for {
@@ -543,5 +540,5 @@ This program uses two ZIO built-in services, `Console` and `Clock`. We don't nee
 By using `ZIO#provideCustomLayer` we only provide the `Logging` layer, and it returns a `ZIO` effect which only requires `ZEnv`:
 
 ```scala mdoc:silent
-val mainEffect: ZIO[ZEnv, Nothing, Unit] = myApp.provideCustomLayer(LoggingLive.layer)
+val mainEffect: ZIO[ZEnv, Nothing, Unit] = myApp.provideCustomLayer(Logging.live)
 ```
--- a/docs/datatypes/contextual/index.md
+++ b/docs/datatypes/contextual/index.md
@@ -329,6 +329,7 @@ trait Logging {
 ```scala mdoc:silent
 import zio.console.Console
 import zio.clock.Clock
+
 case class LoggingLive(console: Console.Service, clock: Clock.Service) extends Logging {
   override def log(line: String): UIO[Unit] = 
     for {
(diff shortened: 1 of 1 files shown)


docs/datatypes/contextual/index.md
```diff
@@ -341,17 +342,22 @@ case class LoggingLive(console: Console.Service, clock: Clock.Service) extends L
 4. **Defining ZLayer** — Now, we create a companion object for `LoggingLive` data type and lift the service implementation into the `ZLayer`:
```
khajavi, line 342: We should also change this instruction to something like this:

Now, we create a companion object for `Logging` service interface and lift the service implementation into the `ZLayer`:

adamgfraser: Great point! Updated now.

khajavi: One other change remains under [_Using provideCustomLayer Method_](https://zio.dev/docs/datatypes/contextual/#dependency-propagation) section.

adamgfraser: Thanks!

mschuwalow:
Nice cleanup!

What is the current recommended structure between `Has[Logging]` and `Has[Logging.Service]`? Currently, both are mixed in the document because of Console and Clock.

adamgfraser:
@mschuwalow Thanks! The recommendation for ZIO 2.0 is to do `Has[Logging]`. We are limited in the ability to do that in ZIO 1.0 for the default services because of binary compatibility but we recently merged a PR to the 2.0 branch to do that for all the standard services in the ZIO environment.

mschuwalow:
> @mschuwalow Thanks! The recommendation for ZIO 2.0 is to do `Has[Logging]`. We are limited in the ability to do that in ZIO 1.0 for the default services because of binary compatibility but we recently merged a PR to the 2.0 branch to do that for all the standard services in the ZIO environment.

Ah cool :+1: Maybe it makes sense to mention this in the doc. Then people can be more confident about picking this structure for their own services.

adamgfraser:
That's a good idea.

@khajavi Do you have an idea for where would be the best place to put that?

khajavi:
> @mschuwalow Thanks! The recommendation for ZIO 2.0 is to do `Has[Logging]`. We are limited in the ability to do that in ZIO 1.0 for the default services because of binary compatibility but we recently merged a PR to the 2.0 branch to do that for all the standard services in the ZIO environment.

Could you please reference that PR? I'm not sure I'm in sync with ongoing changes in 2.0 according to ZIO built-in service definitions. Are we going to rewrite all built-in ZIO services using _Module Pattern 2.0_?

adamgfraser:
Sure. It is #4910. The change I am referred to specifically is in the service definitions. So instead of having doing `Has[Console.Service]` as in the current documentation it will be `Has[Console]`. So then in examples like this it will all be consistent.

khajavi:
> Sure. It is #4910. The change I am referred to specifically is in the service definitions. So instead of having doing `Has[Console.Service]` as in the current documentation it will be `Has[Console]`. So then in examples like this it will all be consistent.

Remarkable changes! The current documentation would be finally consistent with these changes.

There is also a note at ZIO documentation that we encourage not using type aliases for Has wrappers:

> Note:
>
> In Module Pattern 2.0 we don't use type aliases for Has wrappers, like type Logging = Has[Logging.Service]. So unlike the previous pattern, we encourage using explicitly the Has wrappers whenever we want to specify the dependency on a service.
>
> So instead of writing ZLayer[Console with Clock, Nothing, Logging], we write ZLayer[Has[Logging] with Has[Clock], Nothing, Has[Logging]].

I think it makes sense and suffices. What do you think? Do we need to provide more guidance to the user?

/cc @mschuwalow

adamgfraser:
@khajavi Maybe we could just add there that the existing type aliases will be removed in ZIO 2.0 so they know that it is "official" and something we are doing too.

jdegoes:
@adamgfraser 

I think putting the live layers into the companion object of the service trait makes a lot of sense for ZIO services, because it reduces namespace pollution and provides a consistent user experience.

But there are a few drawbacks to recommending this approach:

1. **Unfamiliarity**. Most developers do not nest implementation classes inside companion objects for traits. They may be located in different packages entirely, for example.
2. **Architecture**. A service trait does not depend on any other services; it is independent, and stand alone, merely defining a set of capabilities that service implementations may rely on. Therefore, a service trait does not contribute to cycles in packages. However, implementation of a service trait will depend on other service traits. So by grouping the implementation together with the service trait, any code which depends on the service trait will now depend on the services required by a specific (live) implementation, which has the potential to introduce cycles in the dependency graph.

The first argument is minor, but the second one is more major. In a big code base I would not expect the live implementations to live alongside the service traits, because then all client code of the service traits will depend on all transitive dependencies of the live implementation(s) (and some service traits will have multiple implementations with different dependencies, e.g. netty-async-client versus Java HttpConnection implementations for some HttpClient trait).

Overall, I feel like our practices should be geared for what will help the developer of a large code base organize the code well.

adamgfraser:
@jdegoes That's fair. My concern is less with putting it in the companion object of the service definition than with not putting it in the companion object of the live implementation. I think ideally you don't want to expose the live implementation at all. But you could address that by making the case class defining the live implementation private and then only exposing the layer in the companion object.

```scala
trait Logging {
  def logLine(line: String): UIO[Unit]
}

private final case class LiveLogging(console: Console) extends Logging {
  def logLine(line: String: UIO[Unit] =
    console.putStrLn(line)!
}

object LiveLogging {
  val layer: ZLayer[Has[Console], Nothing, Has[Logging]] =
    LiveLogging.toLayer
}
```

jdegoes:
@adamgfraser There is 1 potential upside to encouraging top-level public case classes for service implementations:

 * Users who do not wish to use layers or ZIO environment can use these as ordinary classes. In fact, maybe they start out using them as ordinary classes, and eventually and gradually introduce layers and ZIO environment.

It is convenient to put initialize / destroy methods in the case class, which I often make public in the case class, even though they do not appear as part of the service interface. Then the layer wraps constructor and automatically performs setup / teardown, which means that you can have the manual and somewhat painful Java experience using the constructor directly, but you get the nice ZIO experience using layers.

Making the case classes public and top-level can be regarded as meeting the Java developer halfway, with a few carrots to incentivize them to make progress down the path toward ZLayer.

Arguing against myself here: on the other hand, it's two ways to do the same thing instead of one. I'm not sure how it balances out in the end.

adamgfraser:
@jdegoes Hmmm. I see a strong advantage to making the live implementation private to force clients to code to the interface versus the implementation and create modularity.

For example, in the snippet below I added an additional operator `logDouble` to the live implementation that was not defined on the `Logging` service. This creates the risk that clients of the service will interact directly with the live implementation and call this operator. This then means that I can't freely refactor the live logging implementation because clients may be relying on implementation details beyond those exposed by the interface.

So I see only exposing publicly the service interface and the live implementation as just a layer constructing that interface to be important for being clear on what the "contract" is and allowing both the service and the clients of the service to evolve independently as long as they adhere to that contract.

```scala
trait Logging {
  def logLine(line: String): UIO[Unit]
}

final case class LiveLogging(console: Console) extends Logging {
  def logLine(line: String: UIO[Unit] =
    console.putStrLn(line)!
  def logDouble(double: Double): UIO[Unit] =
    ???
}

object LiveLogging {
  val layer: ZLayer[Has[Console], Nothing, Has[Logging]] =
    LiveLogging.toLayer
}
```

PR #5067, opened 2021-05-11
Implement ZStream.fromBlockingIterator

luis3m:

PR #5174, opened 2021-06-04
Implement ZStream.fromBlockingIterator

luis3m:
Relevant:

- #5067 
- #4886


adamgfraser:
Now that we have the `blocking` operator on `ZStream` can we get rid of all these blocking variants or implement them in terms of that?

iravid:
Did we port blocking to channels? It seems that we would need to implement support for that on the interpreter for it to work reliably.

luis3m:
@adamgfraser is there any reason why the `maxChunkSize` parameter was dropped for regular `fromIterator`? That's the main difference at the moment between this `fromBlockingIterator` and `ZStream.blocking(ZStream.fromIterator(iterator))`

adamgfraser:
@iravid Yes we did port that to channels though I think you are right that we need to look more closely at how it interacts with the channel executor.

adamgfraser:
@luis3m I think we had some discussion about the proper interpretation of the `maxChunkSize`. For efficiency we would like to allocate the full chunk size up front but that potentially creates an issue where the user can shoot themselves in the foot doing `maxChunkSize = Int.MaxValue` thinking "I just want to get everything in one chunk") and then blowing up with an out of memory error when we allocate this giant array even if the iterator only produces a few elements. Regardless, would be good to make these signatures consistent.

vigoo:
I suggest to implement these with `ZStream.blocking` without the chunk size parameter to be consistent with the non-blocking versions and mark them as deprecated in favor of using `blocking` directly. What do you think?

adamgfraser:
@vigoo Yes.

PR #5175, opened 2021-06-04
Add ZStream.fromFile and others

luis3m:
Part of #4886


streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+                         Chunk.fromByteBuffer(reusableBuffer)
+                       }
+            } yield chunk
+          )
+        }
+      }
+
+  /**
+   * Creates a stream from a `java.io.InputStream`.
+   * Note: the input stream will not be explicitly closed after it is exhausted.
```
searler, line 151: What is the rationale for not closing InputStream?

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+    is: => InputStream,
+    chunkSize: Int = ZStream.DefaultChunkSize
+  ): ZStream[Any, IOException, Byte] =
+    ZStream.fromEffect(UIO(is)).flatMap { capturedIs =>
+      ZStream.repeatEffectChunkOption {
+        for {
+          bufArray  <- UIO(Array.ofDim[Byte](chunkSize))
+          bytesRead <- ZIO.effectBlockingIO(capturedIs.read(bufArray)).asSomeError
+          bytes <- if (bytesRead < 0)
+                     ZIO.fail(None)
```
searler, line 163: Per JavaDoc
_the total number of bytes read into the buffer, or -1 if there is no more data because the end of the stream has been reached._
`bytesRead <0 `is thus expected (and not a failure) and `bytesRead==0` would never occur

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
vigoo: I think it is correct this way, closing the `InputStream` should be the responsibility of the creator (for example it could be coming from a `ZManaged`)

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+   */
+  def fromReaderManaged[R](
+    reader: => ZManaged[R, IOException, Reader],
+    chunkSize: Int = ZStream.DefaultChunkSize
+  ): ZStream[R, IOException, Char] =
+    ZStream.managed(reader).flatMap(fromReader(_, chunkSize))
+
+  /**
+   * Creates a stream from a callback that writes to `java.io.OutputStream`.
+   * Note: the input stream will be closed after the `write` is done.
```
vigoo, line 252: This note is a bit misleading here, as the input stream is an implementation detail which is not visible on the API

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+
+    /**
+     * Read the entire `AsynchronousSocketChannel` by emitting a `Chunk[Byte]`
+     * The caller of this function is NOT responsible for closing the `AsynchronousSocketChannel`.
+     */
+    def read: ZStream[Any, Throwable, Byte] =
+      ZStream.unfoldChunkM(0) {
+        case -1 => ZIO.succeed(Option.empty)
+        case _ =>
+          val buff = ByteBuffer.allocate(ZStream.DefaultChunkSize)
```
vigoo, line 358: We could use a single `ByteBuffer` repeatedly as in the other stream readers instead of allocating one for each `socket.read`

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+
+                override def failed(error: Throwable, attachment: Void): Unit = callback(ZIO.fail(error))
+              }
+            )
+          }
+      }
+
+    /**
+     * Write the entire Chuck[Byte] to the socket channel.
+     * The caller of this function is NOT responsible for closing the `AsynchronousSocketChannel`.
```
vigoo, line 378: This again is a bit misleading for me, the caller already acquired the `Connection`, hopefully via using the `ZManaged` in the companion object, and will close it when the resource is released. This comment suggests for me that `write` itself closes the channel.

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -119,4 +125,288 @@ trait ZStreamPlatformSpecificConstructors {
...
+     */
+    def close(): UIO[Unit] = ZIO.effectTotal(socket.close())
+  }
+
+  object Connection {
+
+    /**
+     * Create a `Managed` connection
+     */
+    def make(socket: AsynchronousSocketChannel): UManaged[Connection] =
```
vigoo, line 409: I would probably make the constructor and the `close` of `Connection` private to force the users to use this method

vigoo (changes requested):
Nice work @luis3m! Could you update it to the latest `series/2.x` version and check the few comments I added?

luis3m:
@vigoo sure, I'll take a look over the weekend

PR #5176, opened 2021-06-06
Add fromSocketClient

searler:
Provides stream based read/write for client sockets.

Symmetrical with fromSocketServer


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -484,6 +484,34 @@ trait ZStreamPlatformSpecificConstructors {
               }
     } yield conn
 
+  /**
+   * Create a Connection from client socket
+   */
+  def fromSocketClient(
+    port: Int,
+    host: String
+  ): ZIO[Blocking, Throwable, UManaged[Connection]] =
```
iravid, line 493: This is a great addition @searler, thank you for contributing.

Since this is not an actual streaming operation (but rather the operations on `Connection` itself are), I am hesitant to place this here.

This is probably because we should create a separate module for these I/O operations, as we do with the `compression` operations.

What do you think @jdegoes @adamgfraser?

iravid:
I think it should be in this repo, yes. Its focus should be ease of use and ergonomics (without compromising on performance ofc) rather than completeness. There’s zio-nio for folks looking for complete control.

searler:
Obviously fromSocketServer would move together with from fromClientSocket.

jdegoes:
@searler Would you be willing to make a new module in this project for zio-streams-io?

searler:
I will make a new module, modeled on zio.stream.compression under streamsJVM

searler:
Hopefully this matches your expections for an io module

searler:
See [zio-tcp](https://github.com/searler/zio-tcp) which is freestanding implementation of this functionality

jdegoes:
Since zio-tcp exists, it may be best to close this PR and put additional effort into expanding adoption around the existing library. However, if we still want to include this in ZIO proper, then let's bring this up to date and re-open.

@searler Thanks for your work on this, Richard, and let us know how to proceed!

PR #5179, opened 2021-06-08
Display Warning If Test Clock Is Being Advanced But Fiber Is Not Suspending

### LINKED ISSUE #5178: TestClock hangs with background thread
The example below is a simplification, in our real code the background worker does more important things than this of course, things that have a direct effect on the test succeeding.

```scala

import zio._
import zio.duration._
import zio.test.environment.TestClock
import zio.test.{DefaultRunnableSpec, ZSpec, assertCompletes}

import java.time.{OffsetDateTime, ZoneOffset}

object TestClockHangup extends DefaultRunnableSpec {

  val backgroundWorker = (
    for {
      _ <- ZIO(1 + 1).forever.fork
    } yield ()
  ).toLayer

  def spec: ZSpec[Environment, Failure] = suite("TestClockHangup")(
    testM("test clock hangup") {
      println("Before clock hangup")
      for {
        testClock <- ZIO.service[TestClock.Service]
        _         <- testClock.adjust(2.seconds)
        _         <- testClock.setDateTime(OffsetDateTime.of(2012, 12, 12, 12, 12, 12, 12, ZoneOffset.UTC))
        _          = println("This will never occur")
      } yield assertCompletes
    },
  ).provideCustomLayer(backgroundWorker)
}
```

We are now restructuring our tests to not require background workers anymore, but that's not the point.

The real problem is that `TestClock`, without explanation or good reason, waits that all threads in the test are either suspended or completed, instead of just the threads it has an effect on.

Maybe that is a technical limitation as one might not know which threads that would be, but the problem is that there's no information on what is happening.

It might make sense to introduce a warning that `TestClock` is waiting on a thread after a certain amount of time, and to maybe return a stack trace of the thread(s) in question, as my colleagues and I lost a full work day on this.

adamgfraser:
Resolves #5178.

### DIFF
--- a/test/shared/src/main/scala/zio/test/environment/package.scala
+++ b/test/shared/src/main/scala/zio/test/environment/package.scala
@@ -656,7 +656,7 @@ package object environment extends PlatformSpecific {
      * The warning message that will be displayed if a test is advancing the
      * clock but a fiber is still running.
      */
-    val suspendedWarning =
+    private val suspendedWarning =
       "Warning: A test is advancing the test clock, but a fiber is not " +
         "suspending, which may result in the test hanging. Use " +
         "TestAspect.diagnose to identity the fiber that is not suspending."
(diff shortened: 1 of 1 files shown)


jdegoes (approved):
Great usability improvement. 👍

PR #5182, opened 2021-06-08
fix link in docs/usecases/testing

domdorn:


docs/usecases/testing.md
```diff
@@ -103,6 +103,6 @@ To get the runner, add the equivalent of following dependency definition under y
 
 To make your spec appear as a JUnit test to build tools and IDEs, convert it to a `class` (JUnit won't run scala objects) and 
 annotate it with `@RunWith(classOf[zio.test.junit.ZTestJUnitRunner])` or simply extend `zio.test.junit.JUnitRunnableSpec`.
-See [MockingExampleSpecWithJUnit](https://github.com/zio/zio/blob/master/examples/jvm/src/test/scala/zio/examples/test/ExampleSpecWithJUnit.scala)
+See [MockExampleSpecWithJUnit](https://github.com/zio/zio/blob/master/examples/jvm/src/test/scala/zio/examples/test/MockExampleSpecWithJUnit.scala)
```
adamgfraser, line 106:
```suggestion
See [ExampleSpecWithJUnit](https://github.com/zio/zio/blob/master/examples/jvm/src/test/scala/zio/examples/test/ExampleSpecWithJUnit.scala)
```

adamgfraser: Shouldn't this be the other way around? Both of these files exist but it seems like the general example of testing with Unit rather than the specific one of mocking is more relevant.

adamgfraser (changes requested):
@domdorn Thank you! Left one comment below.

mijicd:
Closed in #5213

PR #5184, opened 2021-06-09
Add ZStream companion object examples

kazchimo:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3430,24 +3430,67 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
   /**
    * Accesses the environment of the stream in the context of an effect.
+   *
+   * {{{
+   * trait A {
+   *  def a: UIO[String]
+   * }
+   *
+   * ZStream
+   *  .accessM[Has[A]](_.get.a)
```
jdegoes, line 3469: Consider `.serviceWith[A](_.a)` for brevity, here and elsewhere.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3467,19 +3518,36 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+   * }}}
    */
   def apply[A](as: A*): ZStream[Any, Nothing, A] = fromIterable(as)
 
   /**
    * Creates a stream from a single value that will get cleaned up after the
    * stream is consumed
+   *
+   * {{{
+   * ZStream.bracket(ZIO(1))(a => ZIO.succeed(println(s"finish: ${a.toString}")))
```
iravid, line 3534:
```suggestion
   * ZStream.bracket(ZIO(1))(a => UIO(println(s"finish: ${a.toString}")))
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3467,19 +3518,36 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
    */
   def bracket[R, E, A](acquire: ZIO[R, E, A])(release: A => URIO[R, Any]): ZStream[R, E, A] =
     managed(ZManaged.make(acquire)(release))
 
   /**
    * Creates a stream from a single value that will get cleaned up after the
    * stream is consumed
+   *
+   * {{{
+   * ZStream.bracketExit(ZIO(1))((a, e) => URIO(println(s"finish: $a $e")))
```
iravid, line 3547:
```suggestion
   * ZStream.bracketExit(ZIO(1))((a, e) => UIO(println(s"finish: $a $e")))
```

iravid (commented):
Thank you @kazchimo, this is super helpful. Could you address the comments @jdegoes and I left? We can merge right afterwards.

streams/shared/src/main/scala/zio/stream/ZStream.scala
kazchimo: How about this?

```scala
ZStream
  .accessM[A](_.a)
  .provideSome[ZEnv](_ =>
    new A {
      override def a: UIO[String] = UIO("str")
    }
  )
```

I want to use `accessM` here to describe that method.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3630,6 +3768,20 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+   *  stream = ZStream.fromChunkHub(hub)
+   *  fiber <- stream.foreach(i => putStrLn(i.toString)).fork
+   *  _     <- clock.sleep(1.seconds)
+   *  _     <- hub.publish(Chunk(1, 2, 3))
+   *  _     <- fiber.join
+   * } yield ()
+   * // 1
+   * // 2
+   * // 3
+   * }}}
```
khajavi, line 3784: I think it would be better to write this example using a latch:

```scala
for {
  promise <- Promise.make[Nothing, Unit]
  hub <- ZHub.unbounded[Chunk[Int]]
  stream = ZStream
    .fromChunkHub(hub)
    .ensuring(
      promise.succeed(())
    )
  fiber <- stream.foreach(i => putStrLn(i.toString)).fork
  _ <- promise.await
  _ <- hub.publish(Chunk(1, 2, 3))
  _ <- fiber.join
} yield ()
```

Issue #5163, opened 2021-05-31
ZStream.Connection#close assumes no exceptions are thrown from close


searler:
`def close(): UIO[Unit] = ZIO.effectTotal(socket.close())`

Exceptions are admittedly rarely thrown and would not be seen by the tests.
I managed to generate a exception by closing in the middle of a read operation.

This appears to be only instance of this problem in the code base

searler:
See [zio-tcp](https://github.com/searler/zio-tcp) which resolves this issue

Issue #5027, opened 2021-04-28
Flaky ZStream Tests


adamgfraser:
If you run `ZStreamPlatformSpecificSpec` for a while with `nonFlaky` you can get it to suspend indefinitely. Replacing `ZIOBaseSpec` with `DefaultRunnableSpec` and adding debug statements to the `fromSocketServer` tests helps to replicate. Replacing `retry(Schedule.forever)` with `retryN(Int.MaxValue)` may also help replicate by automatically introducing a yield between each repetition, which `retry` should probably be doing anyway. I think the issue is in `read data` and `write data` and my hypothesis would be we are doing something that can potentially block without shifting it onto the blocking thread pool, which could also explain other test failures we have seen in that spec if we are seeing thread starvation.

I've also seen a couple of failures of `write data` like the below. Not sure if these issues relate to the underlying implementation or just the test at this point.

```scala
[info] - ZStream JVM
[info]   - Constructors
[info]     - fromSocketServer
[info]       - write data
[info]         Test failed after 11 iterations with input: 벛
[info]         `receive` =  did not satisfy equalTo(벛)
[info]           at /Users/adamfraser/zio-private/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala:303

[info] - ZStream JVM
[info]   - Constructors
[info]     - fromSocketServer
[info]       - write data
[info]         Test failed after 1 iteration with input: 
[info]         Original input before shrinking was: ᱪ都땾
[info]         `receive` = � did not satisfy equalTo()
[info]           at /Users/adamfraser/zio/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala:303
```

searler:
This test is creating and destroying ~ 40_000 sockets (given the default values for the Gen and nonFlaky )!
That is a significant stress on the OS, which might explain some of the issues.

I do not see the above failures on receive; using Fedora 29 and Zulu 11 .
Only timeout failures are seen.

Your failures could occur if the OS mixes the different TCP connections; which is theoretically possible if reuse_address is enabled.

searler:
```        IO.effectAsync[Throwable, Int] { callback =>
          socket.write(
            ByteBuffer.wrap(c.toArray),
            null,
            new CompletionHandler[Integer, Void] {
              override def completed(result: Integer, attachment: Void): Unit =
                callback(ZIO.succeed(nbBytesWritten + result.toInt))

[...]

PR #5186, opened 2021-06-09
Handle partial writes to NIO Socket and NIO close exceptions

searler:
Ignore exception when closing channel #5163
fix partial write #5027
Handle expected close exceptions during read or write


streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -537,28 +546,37 @@ trait ZStreamPlatformSpecificConstructors {
...
+        case (_: Boolean, c: Chunk[Byte]) => {
+          val buffer = ByteBuffer.wrap(c.toArray)
+          IO.effectAsync[Throwable, Boolean] { callback =>
+            socket.write(
+              buffer,
+              buffer,
+              new CompletionHandler[Integer, ByteBuffer] {
+                override def completed(result: Integer, attachment: ByteBuffer): Unit =
+                  if (attachment.hasRemaining)
+                    socket.write(attachment, attachment, this)
```
iravid, line 562: Great catch!

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -537,28 +546,37 @@ trait ZStreamPlatformSpecificConstructors {
...
+            )
+          }
         }
       }
 
     /**
      * Close the underlying socket
      */
-    def close(): UIO[Unit] = ZIO.effectTotal(socket.close())
+    def close(): UIO[Unit] = ZIO.effect(socket.close()).ignore
```
iravid, line 579: This is a bit problematic as we'd want exceptions thrown here to be turned into defects.

searler: My rationale:
1. Network connection is impacted by remote systems and other events completely out of our control.
There is nothing we can do to resolve and an execution trace referencing them has no value.
2. A close() indicates we have no further interest in this connection. 
That would include whether it had trouble in closing. 

I did see that `effectTotal(close())` is the standard approach in the code base.
That approach assumes that exceptions will be very rarely thrown and are not of interest to the programmer.

Seems that is not appropriate when writing to Files.
A failure on close is quite likely (disk quote exceeded, NAS offline, etc).
The client code can sensibly recover in many cases, e.g. by recreating the file in a different location. 
That would require the file close use an error rather than a defect.

I am happy to change the code back to effectTotal.

searler:
Please review this change @iravid

streams/jvm/src/main/scala/zio/stream/platform.scala
```diff
@@ -533,21 +542,30 @@ trait ZStreamPlatformSpecificConstructors {
...
-            ByteBuffer.wrap(c.toArray),
-            null,
-            new CompletionHandler[Integer, Void] {
-              override def completed(result: Integer, attachment: Void): Unit =
-                callback(ZIO.succeed(nbBytesWritten + result.toInt))
-
-              override def failed(error: Throwable, attachment: Void): Unit = callback(ZIO.fail(error))
-            }
-          )
+    def write: Sink[Throwable, Byte, Byte, Boolean] =
```
iravid, line 547: We can keep this returning the number of written bytes by keeping a var with the count on the CompletionHandler and incrementing it after every write.

What do you think?

iravid (commented):
Looks good @searler. One last question - as this is a change in functionality - and we are good to go.

streams/jvm/src/main/scala/zio/stream/platform.scala
searler: I had thought the byte count was not useful for network connections since they are not a fixed interaction (like a file).
The boolean does provide information that is not otherwise available and could have value for a caller.
However, avoiding unnecessary breakage of existing code is a concern.
I will adjust the code to restore the original functionality

searler:
Restored byte count @iravid

searler:
See [zio-tcp](https://github.com/searler/zio-tcp) which is freestanding implementation of this functionality

PR #5188, opened 2021-06-10
Fix broken link in docs/howto/mock-services.md

axtsnee:
Changed link target from nonexistent .md file to the file that seems most likely to be the intended target.

### DIFF
--- a/docs/howto/mock-services.md
+++ b/docs/howto/mock-services.md
@@ -474,7 +474,7 @@ You can find more examples in the `examples` and `test-tests` subproject:
 - [ComposedEmptyMockSpec][link-gh-composed-empty-mock-spec]
 - [PolyMockSpec][link-gh-poly-mock-spec]
 
-[doc-use-modules-and-layers]: use-modules-and-layers.md
+[doc-use-modules-and-layers]: datatypes/contextual/index.md
 [doc-macros]: howto-macros.md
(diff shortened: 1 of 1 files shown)


docs/howto/mock-services.md
```diff
@@ -474,7 +474,7 @@ You can find more examples in the `examples` and `test-tests` subproject:
 - [ComposedEmptyMockSpec][link-gh-composed-empty-mock-spec]
 - [PolyMockSpec][link-gh-poly-mock-spec]
 
-[doc-use-modules-and-layers]: use-modules-and-layers.md
+[doc-use-modules-and-layers]: ../datatypes/contextual/index.md
```
khajavi, line 477: We also need to change the title of the page we are referring to that. The following sentence doesn't reflect that:

> We'll be assuming you've read How to use modules and layers guide.

axtsnee: That makes sense. Does this sound better?
"We'll be assuming you read about modules and layers in the contextual types guide."

khajavi: Yes, it sounds better.

khajavi (approved):

docs/howto/mock-services.md
axtsnee: Done.

adamgfraser (approved):

Issue #5187, opened 2021-06-10
Mocks hide layer construction failures


vigoo:
In the following example:

```scala
import zio.test.Assertion.{anything, fails}
import zio.test.environment.TestEnvironment
import zio.test.mock.{Expectation, mockable}
import zio.{Has, Task, UIO, ZIO, ZLayer}
import zio.test.{DefaultRunnableSpec, ZSpec, assertM}

trait Service1

trait Service2 {
  def f(): UIO[Unit]
}

@mockable[Service2] object Service2

object Reproducer extends DefaultRunnableSpec {

  val failingLayer: ZLayer[Any, Throwable, Has[Service1]] = Task(throw new RuntimeException("fail!")).toLayer

  override def spec: ZSpec[TestEnvironment, Any] =
    suite("reproducer")(
      testM("test") {
        val f = ZIO.service[Service1] *> ZIO.serviceWith[Service2](_.f())

        val s2 = Service2.F(Expectation.value(()))
        assertM(f.run)(fails(anything)).provideCustomLayer(failingLayer ++ s2)
      }
    )
}
```

the failure produced by `failingLayer` is not getting reported only the failed expectations:

```
- reproducer
  - test
    - unsatisfied expectations
      com.coralogix.archivequery.services.Service2.F with arguments isUnit()
Ran 1 test in 753 ms: 0 succeeded, 0 ignored, 1 failed
```

This can be very misleading.

vigoo:
- the non-mock layer fails with a regular ZIO failure (Fail)
- the mock finalizer fails with a defect (Die)
- then the cause in the `DefaultTestReporter` will be a `Both` of these
- and then this ignores the fail via `dieOption`:

https://github.com/zio/zio/blob/master/test/shared/src/main/scala/zio/test/DefaultTestReporter.scala#L397-L402

PR #5193, opened 2021-06-12
Show all failures in the presence of a mock exception in the report

vigoo:
Fixes https://github.com/zio/zio/issues/5187

- In case of mock exceptions it renders all of them, and also the remaining failures in the cause if there is any
- In case of a test timeout exception it does not render the mock exceptions (because unsatisfied expectations are very likely) but still renders the cause dump if it contains any other information


hmemcpy:
Commenting to keep track, will need to port this to series/2.0's new rendering

adamgfraser (approved):
Looks great!

adamgfraser:
@vigoo Good to go when tests are passing.

adamgfraser (approved):

adamgfraser (approved):

Issue #5137, opened 2021-05-21
Flaky Stream Test On ZIO 2.0 Branch


adamgfraser:
The `mapMPar - awaits children fibers properly` test is showing significant flakiness on the `series/2.x` branch. At first I thought it could be related to the reimplementation of `Semaphore` in terms of `TSemaphore` since `Semaphore` is used internally in `mapMPar`. However, I just tried reverting that change in #5136 and continued to observe the flakiness. Unfortunately I have not been able to replicate locally.

luis3m:
In my implementation of `mapMPar` for series/2.x I have the exact test failing for some specific JVM version + Scala version. When I run that test specifically or all for ZStreamSpec I usually don't replicate the flakiness. However, I do almost every time when I run all tests.

adamgfraser:
Any idea what could be causing it? It seems like it is only occurring on the `series/2.x` branch and is not related to the semaphore change.

luis3m:
Unfortunately not, I spent some time a few weeks ago trying to figure out what's causing this issue but haven't found anything yet. I'll continue exploring this issue over the weekend.

luis3m:
@adamgfraser I updated 8 branches today and I have now all PRs passing. It looks like this test isn't failing anymore. Any idea what changed?

adamgfraser:
@luis3m I'm not sure. We definitely merged changes into the `series/2.x` branch but I can't think of any of them that would have changed this. I have also been pulling changes from `master` into `series/2.x`, but not over the last few days. So maybe we were working off a version of `master` that had this issue and it was subsequently resolved there? Let's keep watching this and make sure it doesn't recur.

adamgfraser:
This test is flaky in the `experimental` package as well now. I think previously the test was not enabled.

PR #5194, opened 2021-06-12
Smart assert + Myers diff 2.0

hmemcpy:
This is a cherry-pick of #5036 and #5158 , porting it to `series/2.0` with the new test rendering format.


adamgfraser (approved):
@hmemcpy Can you remove the `nonFlaky` call in the failing test for the new stream encoding? This test is flaky as reported in #5137 and that aspect is causing a high rate of CI failure which we're seeing in your PR and others. Otherwise good to go!

hmemcpy:
Sure thing! Thanks!

adamgfraser (approved):

PR #5200, opened 2021-06-14
ZIO Runtime System Documentation

khajavi:


khajavi:
@jdegoes  Could you please review this?

jdegoes (approved):
I will approve and merge this. However, I suggest to add a section on constructing `Runtime` from a user-defined environment and user-specified `Platform`. Many real applications should not use `Runtime.default`, but should make their own `Runtime`, which configures the `Platform` and environment accordingly.

PR #5202, opened 2021-06-14
Add ZTransducer.groupAdjacentBy

davidlar:
I needed something similar to fs2.groupAdjacentBy but didn't manage to figure out how to do it without creating this low level transducer. Not sure about the naming and if there should be a convenience function in ZStream as well.

The use case for me is that I get a stream of left joined database rows from skunk (Postgres driver) and I prefer to use zio instead of fs2 / cats effect. I need to aggregate the rows by the columns from the "left" table.

A left join with 3 outer join tables now looks like:

```
def leftJoin3[A, B, C, D](
      s: ZStream[Any, Throwable, (A ~ Option[B] ~ Option[C] ~ Option[D])]
  ): ZStream[Any, Throwable, (A, Chunk[B], Chunk[C], Chunk[D])] = {
    s.aggregate(groupAdjacentBy { case a ~ _ ~ _ ~ _ => a }).map { case (a, s) =>
      val b = s.collect { case _ ~ Some(b) ~ _ ~ _ => b }
      val c = s.collect { case _ ~ _ ~ Some(c) ~ _ => c }
      val d = s.collect { case _ ~ _ ~ _ ~ Some(d) => d }
      (a, b, c, d)
    }
  }```

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZTransducerSpec.scala
@@ -341,6 +341,21 @@ object ZTransducerSpec extends ZIOBaseSpec {
             .runCollect
         )(equalTo(Chunk(3, 4, 5, 1, 2, 3, 4, 5)))
       ),
+      testM("groupAdjacentBy")(
+        assertM(
+          ZStream((1, 1), (1, 2), (1, 3), (2, 1), (2, 2), (1, 4))
+            .aggregate(ZTransducer.groupAdjacentBy(_._1))
+            .runCollect
+        )(
+          equalTo(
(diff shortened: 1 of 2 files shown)


guymers:
What do you think about having it return `(K, NonEmptyChunk[I])` instead of `(K, Chunk[I])`? When using the `fs2` version I quite often end up needing a non-empty collection.

davidlar:
> What do you think about having it return `(K, NonEmptyChunk[I])` instead of `(K, Chunk[I])`? When using the `fs2` version I quite often end up needing a non-empty collection.

That sounds good to me!

davidlar:
I managed to screw this PR up. Created #5767 instead

PR #5222, opened 2021-06-18
Gunzip channel

kmatasflp:
Related to #4886

I'm using Gunzipper implementation from non experimental compression package
https://github.com/zio/zio/blob/master/streams/jvm/src/main/scala/zio/stream/compression/Gunzipper.scala

Had to move some of test fixture code from `compression.TestData` to my `GunzipSpec` because they were using old `ZStream`


kmatasflp:
Bump

vigoo:
Nice work! Can you rebase it to the latest `series/2.x` branch? `halt` was renamed to `failCause`, there may be other small differences.

kmatasflp:
@vigoo Will take a look this weekend

PR #5228, opened 2021-06-20
Improve Connection API - address and shutdownWrite

searler:
Addresses cannot be null in this context.
The Option[SocketAddress] thus unnecessarily complicates the client code.
The original code was inspired by nio-zio, which (presumably) has the goal of exposing the entire NIO API.
This code is part of ZStream and should not unnecessarily expose platform details.

A common TCP interaction pattern uses EOF when writing to indicate the end of the request.
That avoids the need to specify size or parse the request looking for an end marker.
shutdownWrite provides the required functionality.


iravid:
@searler Thank you for the numerous improvements you've contributed, and apologies for the delay :-) I will have time tomorrow to review all of them.

iravid (approved):

PR #5231, opened 2021-06-21
Harden ZHub Implementation

adamgfraser:
If we tentatively remove a subscriber and have to add it back we need to check for pollers again since another fiber may not have done so.


jdegoes (approved):
This looks good to me, although it'd be nice if we could add a stress test to reproduce the old race condition.

PR #5238, opened 2021-06-23
Fix #5152: include .sjsir files when publishing for scalajs on scala 3

### LINKED ISSUE #5152: Missing sjsir files in Scala 3 jars for Scala.js
From what I can see the sjsir files are missing. For example, the file zio_sjs1_3-1.0.8.jar contains class files and tasty files, but not sjsir. The result of this is that, while you can compile code using ZIO on Scala 3/Scala.js, it will not link. This is also the case for zio-test (and interop-cats which is a separate repository).

roald-di:
Fixes #5152

The problem was that the .sjsir files are only generated when the compiler gets passed the `-scalajs` option, this option is correctly added by the scalajs plugin but was then lost when setting scalac options.

I also had to add a type signature in ZScope.scala as it would not compile with -scalajs, showing this error:

```
[error] -- Error: \zio\core\shared\src\main\scala\zio\ZScope.scala:309:57
[error] 309 |              (l: OrderedFinalizer, r: OrderedFinalizer) => if (l eq null) -1 else if (r eq null) 1 else l.order - r.order
[error]     |                                                         ^
[error]     |object creation impossible, since def compare(x$0: T, x$1: T): Int in trait Comparator in package java.util is not defined
[error]     |(The class implements a member with a different type: final def compare
[error]     |  (l: zio.ZScope.OrderedFinalizer, r: zio.ZScope.OrderedFinalizer): Int in anonymous class Object with
[error]     |  java.util.Comparator[? >: zio.ZScope.OrderedFinalizer]
[error]     | {...})
[error] one error found
```

this is probably a compiler bug? I didn't want to invest a lot of time figuring this out since the workaround is pretty easy.

To validate my change I did a publishLocal and checked that the files are included.
I also verified with println that the original scalac options are always empty except for when they contain the single `-scalajs` switch, this means that the rest of the build should not be impacted by this change.

### DIFF
--- a/core/shared/src/main/scala/zio/ZScope.scala
+++ b/core/shared/src/main/scala/zio/ZScope.scala
@@ -18,8 +18,8 @@ package zio
 
 import zio.internal.Sync
 
-import java.util.Map
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
+import java.util.{Comparator, Map}
 
 /**
  * A `ZScope[A]` is a value that allows adding finalizers identified by a key.
@@ -304,11 +304,10 @@ object ZScope {
             weakFinalizers.clear()
             strongFinalizers.clear()
 
-            java.util.Arrays.sort(
-              array,
-              (l: OrderedFinalizer, r: OrderedFinalizer) =>
-                if (l eq null) -1 else if (r eq null) 1 else l.order - r.order
-            )
+            val comparator: Comparator[OrderedFinalizer] = (l: OrderedFinalizer, r: OrderedFinalizer) =>
+              if (l eq null) -1 else if (r eq null) 1 else l.order - r.order
+
+            java.util.Arrays.sort(array, comparator)
 
             val a = exitValue.get()
 
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -227,7 +227,7 @@ object BuildHelper {
     name := s"$prjName",
     crossScalaVersions := Seq(Scala211, Scala212, Scala213),
     ThisBuild / scalaVersion := Scala213,
-    scalacOptions := stdOptions ++ extraOptions(scalaVersion.value, optimize = !isSnapshot.value),
+    scalacOptions ++= stdOptions ++ extraOptions(scalaVersion.value, optimize = !isSnapshot.value),
     libraryDependencies ++= {
       if (scalaVersion.value == ScalaDotty)
         Seq(


ghostdogpr:
Great news! Glad someone finally figured it out 👏

adamgfraser (dismissed):

roald-di:
Ok, I think that I'm going to need some help fixing that `assertTrue` macro..

I have taken a look but I don't quite know how to approach this.

The test that fails to compile under scalajs is this one:

```scala
test("Object constructor") {
    assertTrue(zio.duration.Duration.fromNanos(1000) == zio.duration.Duration.Zero)
}
```

The output of the macro is approximately this

```scala
zio.test.Arrow.succeed[zio](zio)
    .>>>[zio.duration](zio.test.Arrow.fromFunction[zio, zio.duration](((from: zio) => from.duration)))
    .>>>[zio.duration.package.type](zio.test.Arrow.fromFunction[zio.duration, zio.duration.package.type]((from: zio.duration) => from))
    .>>>[zio.duration.Duration.type](zio.test.Arrow.fromFunction[zio.duration.package.type, zio.duration.Duration.type]((from: zio.duration.package.type) => from.Duration))
    .>>>[java.time.Duration](zio.test.Arrow.fromFunction[zio.duration.Duration.type, java.time.Duration]((from: zio.duration.Duration.type) => from.fromNanos(1000L)))
    .>>>[scala.Boolean](zio.test.Assertions.equalTo[java.time.Duration](zio.duration.Duration.Zero)(zio.test.OptionalImplicit.none[zio.test.Diff[java.time.Duration]]))
```

the first line already fails to translate to js with the [error](https://github.com/lampepfl/dotty/blob/7d26a96db6e63aba8db1362b09f4adf521bc8327/compiler/src/dotty/tools/backend/sjs/JSCodeGen.scala#L1411)

```
java.lang.AssertionError: assertion failed: Cannot use package as value: Select(This(Ident(<root>)),zio)
```

which makes sense? trying to compile that line with scalac will fail with the similar error `package zio is not a value`

why is this even working on the jvm?

so I'm not shure what to do, should this be supported by scalajs, should the macro output something different, like

```scala
zio.test.Arrow.succeed[zio.duration.Duration.type](zio.duration.Duration)
    .>>>[java.time.Duration](zio.test.Arrow.fromFunction[zio.duration.Duration.type, java.time.Duration]((from: zio.duration.Duration.type) => from.fromNanos(1000L)))
    .>>>[scala.Boolean](zio.test.Assertions.equalTo[java.time.Duration](zio.duration.Duration.Zero)(zio.test.OptionalImplicit.none[zio.test.Diff[java.time.Duration]]))
```

or something else?

roald-di:
Turns out this was not working on the jvm either, I opened a separate issue for that: #5244

roald-di:
@adamgfraser can we try running ci again to see if this works now?

ghostdogpr:
@roald-di can you rebase to the latest master, so that your other PR is included? That should re-trigger the CI.

adamgfraser (approved):

ghostdogpr:
I just checked the JAR from the latest snapshot and .sjsir files are included 🎉

PR #5242, opened 2021-06-24
Cause: Then distributes over Both even in the presence of Empty

sideeffffect:
Thanks @jdegoes for the very thought-provoking talk on FP without Type Classes. There you promoted using free structures, like `Cause`, which should be a semiring.
So I've tried to implement a semiring instance over at our playground in ZIO Prelude: https://github.com/zio/zio-prelude/pull/351 (has been stale for some time, but I would like to get back to it sooner or later).
And something interesting happened -- the automated test uncovered an edge case in distributivity.

This PR fixes `equals` in `Then` and `Both` so that `Then` truly distributes over `Both`, even in the presence of `Empty` (a neutral element).

/cc @adamgfraser

### DIFF
--- a/core-tests/shared/src/test/scala/zio/CauseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/CauseSpec.scala
@@ -1,6 +1,6 @@
 package zio
 
-import zio.Cause.{Both, Then}
+import zio.Cause.{Both, Then, empty}
 import zio.random.Random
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):

jdegoes:
@sideeffffect Thank you! 🙏 

Now it's actually true. 😅 😆

PR #5246, opened 2021-06-25
Smart Constructors

adamgfraser:


adamgfraser:
@jdegoes This is quite interesting. Type inference seems to be excellent. A couple of observations:

* Some care must be taken to appropriately prioritize instances. For example `fromFuture` and `fromFunction` can conflict if not prioritized because `fromFuture` takes a function from an `ExecutionContext` to a `Future` that is itself a function.
* We may need different variants of this for different ZIO types. For example `UIO.apply` should have different behavior than `ZIO.apply`. We might be able to abstract over that but could also just make sense to have `ZIOConstructor`, `UIOConstructor`, and so on in the companion objects for each type.
* It pushes us a bit to pick the "canonical" variant of each constructor (e.g. `fromFuture` versus `fromFutureInterrupt`).
* Could potentially be used even more broadly (e..g `ZIO.suspend`).

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -5322,6 +5323,158 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+      new ZIOConstructor[Fiber[E, A]] {
+        type Out = IO[E, A]
+        def make(input: => Fiber[E, A]): IO[E, A] =
+          ZIO.fromFiber(input)
+      }
+
+    /**
+     * Constructs a `ZIO[R, E, A]` from a `ZIO[R, E, Fiber[E, A]]`.
+     */
+    implicit def FiberZIOConstructor[R, E, A]: WithOut[ZIO[R, E, Fiber[E, A]], ZIO[R, E, A]] =
```
adamgfraser, line 5369: Not sure if unwrapping the `ZIO` effect like this is a little too much magic.

luis3m:
Perhaps support for `R => Either[E, A]` as well?

adamgfraser:
@luis3m Yes I think we could add a `fromFunctionEither`.

jdegoes:
This is going to be really amazing! Probably the different aliases need their own specific constructors.

I think it could be useful to do it with `ZIO.suspend` too.

adamgfraser:
@jdegoes Yes I agree. Will do.

iravid:
I have to say I'm not a fan of the apply overloading on ZStream, particularly for the numerous variants of from* constructors. I would personally discourage teams from using these liberally as they make the code harder to read through without an IDE.

The ZIO ones are fine, I think, as the amount of types that get overloaded is relatively small. Although I would still encourage using the explicit constructors.

For ZManaged it's a bit trickier, particularly the AutoCloseable one which relies on the reader knowing the implemented interfaces of the type being passed to apply.

jdegoes:
@adamgfraser For things like `Either[E, A]`, we should have a higher priority one for `Either[Throwable, A]`, which would essentially do `ZIO.suspend(ZIO.fromEither(either))`.

i.e. in any case we know the error is Throwable, we can wrap the whole thing to capture accidental `throw`s instead.

adamgfraser:
@jdegoes Will do.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -5322,6 +5329,313 @@ object ZIO extends ZIOCompanionPlatformSpecific {
...
+      new ZIOConstructor[Option[A]] {
+        type Out = ZIO[Any, Option[Nothing], A]
+        def make(input: => Option[A]): ZIO[Any, Option[Nothing], A] =
+          ZIO.fromOption(input)
+      }
+
+    /**
+     * Constructs a `ZIO[Any, Option[Nothing], A]` from a `None`.
+     */
+    implicit def OptionNoneConstructor[A]: WithOut[None.type, ZIO[Any, Option[Nothing], A]] =
```
iravid, line 5507: Should this not be:
```suggestion
    implicit val OptionNoneConstructor: WithOut[None.type, ZIO[Any, Option[Nothing], Nothing]] =
```

core/shared/src/main/scala/zio/ZManaged.scala
```diff
@@ -2986,6 +2989,198 @@ object ZManaged extends ZManagedPlatformSpecific {
...
+      new ZManagedConstructor[Option[A]] {
+        type Out = ZManaged[Any, Option[Nothing], A]
+        def make(input: => Option[A]): ZManaged[Any, Option[Nothing], A] =
+          ZManaged.fromOption(input)
+      }
+
+    /**
+     * Constructs a `ZManaged[Any, Option[Nothing], A]` from a `None`.
+     */
+    implicit def OptionNoneConstructor[A]: WithOut[None.type, ZManaged[Any, Option[Nothing], A]] =
```
iravid, line 3075: Same as with the ZIO instance:
```suggestion
    implicit val OptionNoneConstructor: WithOut[None.type, ZManaged[Any, Option[Nothing], Nothing]] =
```

core/shared/src/main/scala/zio/ZIO.scala
adamgfraser: Yes. Fixed in the latest version.

core/shared/src/main/scala/zio/ZManaged.scala
adamgfraser: Done.

adamgfraser:
@jdegoes One question about the `suspend` point is what to do if the error is a more specific subtype of `Throwable`. Suspending would imply widening the error from the more specific subtype to `Throwable` itself. That seems like not what you would want. We could potentially only suspend if the error type was exactly `Throwable` but that seems to contradict the use of variance and I think would be very confusing.

adamgfraser:
@jdegoes @iravid This is ready for review.

iravid (dismissed):
Really pleased about the outcome here. Awesome work 💪🏼

jdegoes:
@adamgfraser Looking really good. How about `Cause[E]`? This would always translate into `failCause`.

I think it's ok to widen to `Throwable` in the event the existing error is a subtype of `Throwable`. It's not correct, just less precise, but then again, you can always `refineOrDie` your way back to a specific type.

See also: the behavior of existing `suspend`.

jdegoes:
One more comment:

What if you defined polymorphic `ZIOConstructor` for ZIO, ZManaged, Either, etc. where the success value was itself anything that can be converted to ZIO?

Just thinking that if you had, e.g. `ZIO[Option[Right[ZIO]]]` and you could somehow benefit from induction, you could smash that all the way to `ZIO`.

adamgfraser:
Can definitely add `Cause`.

I wonder about widening the error type in all of these cases. Like consider this:

```scala
trait MyDomainError extends Throwable

val myEither: Either[MyDomainError, Int] = ???
val zio = ZIO.from(myEither) // Task[Int]
```

I agree we do this with `suspend` but that operator is explicitly designed for that. I worry when we are doing something automatically where a data type already models the failure type that is not going to be what the user wants.

jdegoes:
I agree we should not widen the error unless we need to. For example, `ZIO.from(either)` should not widen the error type. But `ZIO.suspend(...)` always widens, so it's not the end of the world to allow `ZIO.from(zio))` to widen.

An alternative is that whenever we have `E <: Throwable`, we can require a tag, so that we can `ZIO.suspend().refineToOrDie[type]`, in effect. Thus keeping the more narrow error.

jdegoes:
Maybe not on `Cause`. I am not sure. So far, `ZIO.from` is all about the successful case so far. `Cause` would be the first case of producing a ZIO that always fails. If you add `Cause`, you might think about adding `Throwable`, but I am not sure if it makes sense to go down that path.

adamgfraser:
Yeah, I think that is fair regarding `Cause`. We actually don't currently have `suspend` implemented with a smart constructor though we certainly can and I agree widening the error like with `suspend` is the most logical interpretation there. I thought you were suggesting above suspending all of these other constructors like the `Either` one which I think would require widening the error type in many more cases.

jdegoes:
Yes, I'd just widen on the `suspend` one, or do the tag trick there so that, e.g. `ZIO.from(effect: IO[IOException, A])` would end up as `IO[IOException, A]`.

adamgfraser:
So I guess we're saying if `E` is a subtype of `Throwable` that gets translated as `suspend` and if it is not it gets translated as `suspendSucceed`, is that right?

adamgfraser:
I do wonder a bit whether it makes sense to include this at all. Suspending effects is relatively advanced. If you have a `ZIO` and want a `ZIO` I would think the most basic thing you would do if you were getting started is just use it and not even look to a constructor, smart or otherwise.

adamgfraser:
Also leads to somewhat unintuitive conclusion I think that `E =:= Nothing` translates to `suspendSucceed`, `E <:< Throwable` goes to `suspend`, and all other `E` goes to `suspendSucceed`.

adamgfraser:
The recursive idea is interesting. I think we could do something there.

jdegoes:
> I do wonder a bit whether it makes sense to include this at all

I am fine omitting suspend. Do you want to merge this as-is or do more work on it?

adamgfraser:
I think at this point we have everything resolved except for the potential support for recursive constructors. There is a lot here so I think we should merge this and then I can open a separate PR to explore the recursive constructors. That way we will. have a baseline of at least this functionality working as we explore that.

adamgfraser:
@jdegoes How about this?

```scala
implicit def OptionConstructor[R, E, A, B](implicit constructor: ZManagedConstructor.WithOut[A, ZManaged[R, E, B]]): WithOut[Option[A], ZManaged[R, Option[E], B]] =
  new ZManagedConstructor[Option[A]] {
    type Out = ZManaged[R, Option[E], B]
    def make(input: => Option[A]): ZManaged[R, Option[E], B] = {
      input.fold[ZManaged[R, Option[E], B]](ZManaged.fail(None))(a => constructor.make(a).asSomeError)
    }
  }

lazy val x: Option[Either[String, Int]] = ???
lazy val y = ZManaged.from(x)
lazy val z: ZManaged[Any, Option[String], Int] = y
```

jdegoes:
@adamgfraser That's amazing. How robust / generalized is it?

Good to merge this one as soon as conflicts are fixed!

jdegoes (approved):

PR #5248, opened 2021-06-26
Introduce ZApp, an App with custom R

ghostdogpr:
The current `App` is not usable if you want your `Runtime` to have an environment different from `ZEnv` which is quite annoying as you have to duplicate the code of `App` into your own project. I introduced `ZApp` which lets you use any environment.

Unfortunately this breaks mima checks, so I submitted this to the 2.x series branch. It could be made compatible by duplicating the code from `App` and not changing the trait inheritance, but it's probably not worth it.

### DIFF
--- a/core/js/src/main/scala/zio/App.scala
+++ b/core/js/src/main/scala/zio/App.scala
@@ -16,17 +16,4 @@
 
 package zio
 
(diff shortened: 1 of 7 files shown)


adamgfraser (approved):
Makes sense to me.

PR #5259, opened 2021-06-29
ZIO 2.0: Implement left, right, unleft, and unright

adamgfraser:
Changes signatures of `left` and` right` to preserve the value in the other side of the `Either` and implements `unleft` and `unright` operators that are the inverse of these. This allows you to ergonomically work with values inside an `Either` by calling `left` or `right`, performing some operations, and then calling `unleft` or `unright`.


jdegoes (approved):

jdegoes:
Excellent work!

PR #5260, opened 2021-06-29
Bug fix: `ZStream#timeoutError` has a by-name error parameter that is eagerly evaluated

### LINKED ISSUE #4725: `ZStream#timeoutError` has by-name error parameter, but evaluates it in any case
```scala
  final def timeoutError[E1 >: E](e: => E1)(d: Duration): ZStream[R with Clock, E1, O] =
    timeoutErrorCause(Cause.fail(e))(d)

  final def timeoutErrorCause[E1 >: E](cause: Cause[E1])(d: Duration): ZStream[R with Clock, E1, O] =
    ..

  //in Cause
  def fail[E](error: E): Cause[E] = Internal.Fail(error)
```

Probably the most reasonable solution would be to make `timeoutErrorCause`'s `cause` a by-name too, but that breaks binary compatibility. Possibly a workaround can be done by making the existing method `private` and shadowing it with a public one with the right signature.

harveywi:
Resolves #4725.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/StreamLazinessSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/StreamLazinessSpec.scala
@@ -2,6 +2,7 @@ package zio.stream
 
 import zio.test._
 import zio.{UIO, ZIOBaseSpec}
+import zio.duration._
 
 object StreamLazinessSpec extends ZIOBaseSpec {
 
@@ -25,7 +26,12 @@ object StreamLazinessSpec extends ZIOBaseSpec {
       testM("fromChunk")(assertLazy(ZStream.fromChunk)),
       testM("fromIterable")(assertLazy(ZStream.fromIterable)),
       testM("halt")(assertLazy(ZStream.halt)),
-      testM("succeed")(assertLazy(ZStream.succeed))
+      testM("succeed")(assertLazy(ZStream.succeed)),
+      testM("timeoutError")(
+        assertLazy(
+          ZStream.succeed(1).timeoutError(_)(Duration.Infinity)
+        )
+      )
     )
   )
 }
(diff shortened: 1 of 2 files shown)


iravid:
Nice work @harveywi! It's surprising to me that this works because I would assume that calling a function that requires a strict parameter with a by-name parameter would cause the by-name parameter to be evaluated.

This is what I'm seeing in the REPL:
```scala
scala> :paste
// Entering paste mode (ctrl-D to finish)
def strict(s: String) = "not s"

def byName(s: => String) = strict(s)

// Exiting paste mode, now interpreting.

def strict(s: String): String
def byName(s: => String): String

scala> byName(s())
evaluating
val res0: String = not s
```

iravid:
Sorry I forgot to paste the definition of `s` in the snippet above; fixed now.

iravid:
So we might have a problem with `assertLazy`? 🤔

harveywi:
> Nice work @harveywi!

🥳 Thank you!

> It's surprising to me that this works because I would assume that calling a function that requires a strict parameter with a by-name parameter would cause the by-name parameter to be evaluated.

In this particular case, the by-name parameter `e` is only evaluated if the right-hand side of the argument here is evaluated: `.mapErrorCause(_ => Cause.fail(e))`. So we don't actually pass `e` in as a strict parameter: It is only evaluated if it must be wrapped in a `Cause`, which happens lazily by virtue of the mechanics of function evaluation. Conceptually, it looks okay to me, but I may have overlooked something.

To put your mind at ease a bit about `assertLazy`: The new test that was added to the `zio.stream.StreamLazinessSpec` suite fails if we use the original implementation of `timeoutError`. So it appears that `assertLazy` is working properly.

iravid (approved):
Oh you're right @harveywi. I had the example backwards. This is good to merge then; could you just check why the lint step is failing?

iravid (approved):
Great work @harveywi!

PR #5262, opened 2021-06-29
feat: ZStream.*Decode

runtologist:
Implements `utfDecode`, `utf8Decode`, `utf16Decode`, `utf16BEDecode`, `utf16LEDecode`, `utf32Decode`, `utf32BEDecode`, `utf32LEDecode`, `usASCIIDecode` on `ZStream`


streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -1848,15 +1849,28 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+    mapAccumChunk(s)((currS, in) => in.mapAccum(currS)(f))
+
+  /**
+   * Statefully maps over the chunks of this stream to produce new chunks.
+   */
+  def mapAccumChunk[S, A1](s: S)(f: (S, Chunk[A]) => (S, Chunk[A1])): ZStream[R, E, A1] =
+    mapAccumChunkEnd(s)(f)(_ => Chunk.empty)
+
+  /**
+   * Statefully maps over the chuinks of this stream to produce new chunks,
```
iravid, line 1862:
```suggestion
   * Statefully maps over the chunks of this stream to produce new chunks,
```

streams/shared/src/main/scala/zio/stream/experimental/ZStream.scala
```diff
@@ -3164,6 +3178,193 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+      }
+    )
+
+  /**
+   * Decodes chunks of UTF-8 bytes into strings.
+   *
+   * This transducer uses the String constructor's behavior when handling malformed byte
+   * sequences.
+   */
+  def utf8Decode[R1 <: R, E1 >: E, A1 >: A <: Byte](implicit ev: A <:< Byte): ZStream[R1, E1, String] = {
```
iravid, line 3201: This was already ported to `ZSink.utf8Decode` :-)

iravid (commented):
The additional `mapAccum` variants are super useful.

For the `decode` methods, it'd be great if we port them as ZSinks; that would make them more composable compared to functions that operate on streams.

iravid:
Oh and we should also port the tests :-)

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
```diff
@@ -0,0 +1,229 @@
...
+    def utf32Decode: ZStream[R, E, String]   = new ZStream[R, E, String](stream.channel >>> StringDecoder.utf32Decode)
+    def utf32BEDecode: ZStream[R, E, String] = new ZStream[R, E, String](stream.channel >>> StringDecoder.utf32BEDecode)
+    def utf32LEDecode: ZStream[R, E, String] = new ZStream[R, E, String](stream.channel >>> StringDecoder.utf32LEDecode)
+    def usASCIIDecode: ZStream[R, E, String] = new ZStream[R, E, String](stream.channel >>> StringDecoder.usASCIIDecode)
+
+  }
+
+  def utfDecode[Err, Done]: ZChannel[Any, Err, Chunk[Byte], Done, Err, Chunk[String], Done] =
+    ZChannel.readOrIdentity { chunk =>
+      if (chunk.startsWith(0 :: 0 :: -2 :: -1 :: Nil) && Charset.isSupported("UTF-32BE"))
```
iravid, line 25: Checking the bytes in this form will box them. You can use `Chunk#byte(Int)` to perform indexed access that won't  box.

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
```diff
@@ -0,0 +1,229 @@
+package zio.stream.experimental
+
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
+import zio.Chunk
+
+object StringDecoder {
```
iravid, line 7: We can perhaps name this `StringCodecs` and add encoders here as well.

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  implicit class ByteStreamDecode[R, E, A](val stream: ZStream[R, E, Byte]) extends AnyVal {
```
iravid, line 9: We should probably just add these methods on ZStream itself if we want infix versions of these. Implicit classes are unfortunately not very discoverable.

Pinging @adamgfraser for thoughts around API.

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Decodes chunks of UTF-8 bytes into strings.
+   *
+   * This channel uses the String constructor's behavior when handling malformed byte
+   * sequences.
+   */
+  def utf8Decode[Err, Done]: ZChannel[Any, Err, Chunk[Byte], Done, Err, Chunk[String], Done] =
+    utf8Decode(Chunk.empty)
+
+  def utf8Decode[Err, Done](initial: Chunk[Byte]): ZChannel[Any, Err, Chunk[Byte], Done, Err, Chunk[String], Done] = {
```
iravid, line 47: Does this overload need to be public? Could be an inner method on `utf8Decode` perhaps.

iravid: Also, now that this is here, we can delete `ZSink.utf8Decode`.

iravid (commented):
Great work @runtologist. A few more finishing touches and we're good to go.

@mijicd Please have a look. After some internal discussions we realized it is far more efficient to formulate these codec operations as channels. So perhaps you could do the same for the split combinators?

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
runtologist: Directly on ZStream these methods will have very fancy type signatures, while this might be a bit less discoverable but as simple as it gets.

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
runtologist: Right, it should be as private as the other ones.

runtologist:
All review comments should be addressed now. I still need to add tests. I'll open a separate PR for `ZStream.mapAccumChunk*`, as it is no longer needed here.

streams/jvm/src/main/scala/zio/stream/experimental/StringDecoder.scala
iravid: You mean in terms of the implicit evidence? That's fine- many of the ZStream methods already require implicit evidence. It's preferred in the project because of the much better discoverability.

paulpdaniels:
Oh I guess I should have done a search before doing #5437

vigoo:
I think this can be closed as these were added to `ZPipeline` in a different PR

vigoo:
I think these are now already implemented (as a mix of this and another PR)

PR #5266, opened 2021-06-30
Make More Constructors Lazy

### LINKED ISSUE #5261: ZIO 2.0: By-name parameter check
This ticket is to review all parameters for all methods that return a ZIO effect, and ensure they are by-name; and to, if a by-name parameter is referenced more than once, to put it into a local `lazy val` to ensure it is not evaluated more than once.

This will improve the ability of ZIO to deal with "accidentally embedded side-effects", which are quite common in practice.

adamgfraser:
Resolves #5261.

This tackles the low hanging fruit of constructors that take a value that is not a `ZIO` and construct a `ZIO` value. I think the question is how broadly to take this. We could potentially expand to methods that accept a `ZIO` value such as:

```scala
  final def absolve[E1 >: E, B](implicit ev: A <:< Either[E1, B]): ZIO[R, E1, B] =
    ???
```

One other issue to think through here is consistency with the version of the operator on the `ZIO` trait that I don't think could be by name.

There are also operators that don't directly construct `ZIO` values but perform various transformations:

```scala
  def foreachParN[R, E, A, B, Collection[+Element] <: Iterable[Element]](n: Int)(
    as: Collection[A]
  )(fn: A => ZIO[R, E, B])(implicit bf: BuildFrom[Collection[A], B, Collection[B]]): ZIO[R, E, Collection[B]] =
    ???
```

Here the desired parallelism as well as the original collection could potentially be by name.

I think the right answer here is probably "make all the things by name" but that does have some cost in terms of having to suspend at points where we otherwise wouldn't so wanted to get alignment before I made broader changes.


mijicd (approved):

jdegoes:
I think this is an easy-approve. 👍 

For more complicated scenarios, we probably need some guiding principles, and embed that into the contributor guidelines (indeed, also the naming guidelines we adopted should make it into the guidelines doc).

I think that if an operator is available as a method and as a constructor, e.g. `ZIO.absolve` vs `ZIO#absolve`, then that's a good reason to make it eager rather than by-name, just so we have the guarantee that `ZIO.absolve(foo)` behaves the same as `foo.absolve` (that is a weak guarantee but it seems slightly useful and therefore a reason to bias).

As for things like `ZIO.foreachParN`, you can see a lot of value in making that lazy, e.g.:

```scala
ZIO.foreachParN(System.getRuntime().availableProcessors())(Source.fromFile("data.txt").getLines) { line =>
}
```

This code would be quite broken now, but would be much less broken if the parameters were by-name.

The overall pattern in implementing such methods will be:

```scala
def foreachParN[A](n0: => Int)(a0: => Iterable[A]) = 
  ZIO.suspend {
    val n = n0 
    val a = a0
    ...
  }
```

This will produce code that is robust to double evaluation, as well as robust to side-effects embedded in parameters.

adamgfraser:
Great. Will start working on this.

adamgfraser:
@jdegoes A couple of other observations as I have spent a bit more time on this:

1. The principle that operators that are available as methods in addition to constructors should be by value ends up being fairly awkward. It required "non-local" knowledge of whether the constructor has a method version that often cannot be determined by analyzing the constructor signature itself and can depend on arbitrary factors like whether we have happened to implement a method version. If we implement a method version in the future are we supposed to go back and change the constructor signature? I think perhaps a better principle is something like "all parameters that are not functions should be by name". Then we can say that of course if you call `foo.bar` it assumes that `foo` exists and if you want that to be lazy you should use the constructor version or suspend yourself.
2. By name parameters play quite poorly with overloaded methods like `foreach` and `collectAll` because the by name parameters erase to `Function0`. Perhaps we should consider suffixes like `foreachArray` for those variants, thought it does feel a little like an implementation leak. We could also explore using path dependent types to unify those methods.

adamgfraser:

jdegoes (dismissed):

adamgfraser:
@jdegoes This is ready to go.

kitlangton (approved):

PR #5268, opened 2021-06-30
Implement Compositional Zips

### LINKED ISSUE #5267: ZIO 2.0: Composable zips
The goal of this ticket is to ensure that when zipping together different effects:

1. Tuples are not nested.
2. Units do not contribute to output.

For example:

```scala
(x1: Task[Int]) <*> (x2: Task[Unit]) <*> (x3: Task[String]) <*> (x4: Task[Boolean]) : Task[(Int, String, Boolean)]
```

adamgfraser:
Resolves #5267.


iravid (approved):
Great stuff. Improves usability significantly.

We have the `<*>` extractor in one of the package objects; I think we can remove it now.

jdegoes:
@adamgfraser I would also remove the `<*>` extractor. Otherwise, looks awesome!

jdegoes (approved):

Issue #4351, opened 2020-10-25
Consider adding Schedule.deadline?


reibitto:
This one seems to come up quite a bit. Right now I write this as:

```scala
Schedule.elapsed.whileOutput(_ < 10.seconds)
```

which might short enough that a specific method isn't needed. I'm not sure. But a helper like `Schedule.deadline(10.seconds)` might be better for discoverability. (I'm not sure about the name `deadline` though)

Prior to 1.0.0, `Schedule.duration` seemed to do this:
https://javadoc.io/static/dev.zio/zio_2.12/1.0.0-RC20/zio/Schedule$.html#duration(duration:zio.duration.Duration):zio.Schedule[zio.clock.Clock,Any,zio.duration.Duration]

But the meaning of that function has changed since then. I don't think the old behavior was ever replaced with anything.

reibitto:
There's also one other point to consider. Rather than doing:

```scala
Schedule.spaced(1.second) && Schedule.deadline(10.seconds)
// for: Schedule.spaced(1.second) && Schedule.elapsed.whileOutput(_ < 10.seconds)
```

It could go on the class itself to make this possible:

```scala
Schedule.spaced(1.second).deadline(10.seconds)
```

dohrisalim:
Hi, I think that I solved this issue with the help of Openhands. I will create a pull request with the solution.

kyri-petrou:
@dohrisalim this issue was opened before `upTo` was implemented. It should have been closed but it wasn't. I don't think adding an alias to it is needed

PR #5274, opened 2021-07-02
Introduce Schedule.upTo operator

agarella:
This introduces an operator that allows a Schedule to recur for a given duration. As suggested here https://github.com/zio/zio/issues/4351


jdegoes (changes requested):
Looks great!

How about we call it `upTo`, e.g. `Schedule.spaced(1.second).upTo(5.seconds)`.

agarella:
@jdegoes I've updated it according to your suggestion

adamgfraser:
@agarella Can you run `prepare` to format?

adamgfraser (approved):

adamgfraser (approved):

PR #5279, opened 2021-07-04
Start documenting migration from cats effect to ZIO

zhrebicek:
While migrating something from Monix to ZIO I have found that there was `flatTap` from Cats which were nowhere to be found in Monix migration guide. So I have started with documentation of it at least like this.


zhrebicek:
Wanted to explore some more `cats.*` but it would be better to merge at least something.

jdegoes (approved):

jdegoes:
Thank you!

PR #5280, opened 2021-07-05
Implement Unzippable

adamgfraser:
The dual of `Zippable`. Supports combining contravariant types in the same way as `Zippable` for covariant types. This allows us to implement compositional zips for `ZRef.Synchronized`.


jdegoes (dismissed):

jdegoes:
Clever idea. I like it!

core-tests/shared/src/test/scala/zio/ZRefSynchronizedSpec.scala
```diff
@@ -490,7 +490,7 @@ object ZRefMSpec extends ZIOBaseSpec {
           tuple   <- composed.get
           (a, b)   = tuple
         } yield assert(a)(equalTo(6765)) && assert(b)(equalTo(10946))
-      },
+      } @@ nonFlaky,
```
jdegoes, line 493: Is this strictly necessary?

jdegoes (approved):

core-tests/shared/src/test/scala/zio/ZRefSynchronizedSpec.scala
adamgfraser: Wanted to confirm that the updating of the two `Ref` values is performed atomically and we didn't just get "lucky" that two of the fibers updating the `Ref` didn't make conflicting updates, but happy to remove if you want.

nartamonov:
> The dual of `Zippable`. Supports combining contravariant types in the same way as `Zippable` for covariant types. This allows us to implement compositional zips for `ZRef.Synchronized`.

Just discovered `Unzippable`, it's awesome! 👍

adamgfraser:
🎉

PR #5291, opened 2021-07-09
Fixed #5288 - Adding serviceWithManaged to ZManaged constructors

### LINKED ISSUE #5288: Adding serviceWithManaged to ZManaged contructors
`ZIO` and `ZManaged` companion objects have very convenient constructors `serviceWith` which more or less replace the previous `accessM` to account for the new module recommendation, i.e. dropping the type aliases `type ServiceName = Has[ServiceName.Service]` and using a more flattened convention.

`serviceWith` allows to access the service from the environment to execute an effectual service member, effectively duplicating the behavior of `accessM` without dealing with `_.get`, in the function argument, because we're now dealing directly with a service instance, unwrapped from its `Has[_]`. There is an `accessManaged` overload, which has the same behavior as `accessM`, but this time to deal with service members not only returning an effectual computation, but a managed one.

The same overload of `serviceWith` seems to be missing, when dealing with managed services members in the module pattern "2.0".

remiguittaut:
The naming "serviceWithManaged" is open for discussion.

There might be a flaky test problem which needs to be investigated.

### DIFF
--- a/core/shared/src/main/scala/zio/ZManaged.scala
+++ b/core/shared/src/main/scala/zio/ZManaged.scala
@@ -1218,6 +1218,13 @@ object ZManaged extends ZManagedPlatformSpecific {
       ZManaged.fromEffect(ZIO.serviceWith[Service](f))
   }
 
+  final class ServiceWithManagedPartiallyApplied[Service](private val dummy: Boolean = true) extends AnyVal {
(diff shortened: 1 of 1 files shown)


adamgfraser (approved):
Thank you! 🙏

PR #5297, opened 2021-07-11
Adding foreachFlatten: ZIO.foreachFlatten(Col[A])(A => Col[B]): Col[B]

blast-hardcheese:
Mirrors expected functionality from `flatTraverse` in cats. Currently, the only way to do this is:

```scala
ZIO.foreach(List.empty[Int])(a => UIO(List(a.toString, a.toString))).map(_.flatten)
```

with this change, that can be simplified to
```scala
ZIO.flatForeach(List.empty[Int])(a => UIO(List(a.toString, a.toString)))
```

This has been asked about a few times on the ZIO Discord, @iravid gave a reason why it wouldn't be practical to do, though this was a while back so perhaps this can be revisited.

Initially, I was just going to add this change to `ZIO`/`UIO`/`IO`, but for completeness I branched out to the other objects that had `foreach` defined, though I don't have a use for that right now and can revert those if desired.

---

I have another change, but I don't know how to make the ergonomics correct without doing the type pivot trick used commonly in more functional codebases, so I'll leave it here as a comment. `ZIO.flatForeach` would be a welcome change without the instance function:

Example (note the required `(a: Int)`, which should be inferable considering `UIO[List[Int]]` has a concrete type):
```scala
scala> runSync(UIO(List(1,2,3,4)).flatForeach((a: Int) => putStrLn(s"Processing: ${a}").as(List(a.toString, a.toString))))
Processing: 1
Processing: 2
Processing: 3
Processing: 4
val res1: zio.Exit[java.io.IOException,List[String]] = Success(List(1, 1, 2, 2, 3, 3, 4, 4))
```

<details><summary>expand diff</summary>

### diff

```diff
diff --git a/core/shared/src/main/scala/zio/ZIO.scala b/core/shared/src/main/scala/zio/ZIO.scala
index a44d084bd..94a3e5a0f 100644
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -633,6 +633,11 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   def flatMap[R1 <: R, E1 >: E, B](k: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] =
     new ZIO.FlatMap(self, k)
 
+  def flatForeach[R1 <: R, E1 >: E, Collection[+Element] <: Iterable[Element], AA, B](
+    k: AA => ZIO[R1, E1, Collection[B]]
+  )(implicit ev1: A <:< Collection[AA], bf: BuildFrom[Collection[AA], B, Collection[B]]): ZIO[R1, E1, Collection[B]] =
+   new ZIO.FlatMap(self.asInstanceOf[ZIO[R, E, Collection[AA]]], (as: Collection[AA]) => ZIO.flatForeach(as)(k))
+
   /**
    * Creates a composite effect that represents this effect followed by another
    * one that may depend on the error produced by this one.
```

</details>

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -701,6 +701,27 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(res.run)(dies(isSubtype[NumberFormatException](anything)))
       } @@ zioTag(errors)
     ),
+    suite("foreachFlatten")(
+      testM("returns the list of results") {
+        checkAllM(functionListIOGen, listGen) { (f, list) =>
+          val res = IO.foreachFlatten(list)(f)
+          assertM(res)(isSubtype[List[Int]](anything) && hasSize(equalTo(1000)))
+        }
+      },
+      testM("both evaluates effects and returns the list of results in the same order") {
+        val list = List("1", "2", "3")
+        for {
+          ref     <- Ref.make(List.empty[String])
+          res     <- IO.foreachFlatten(list)(x => ref.update(_ :+ x) *> IO.effectTotal[List[Int]](List(x.toInt, x.toInt)))
+          effects <- ref.get
+        } yield assert(effects)(equalTo(list)) && assert(res)(equalTo(List(1, 1, 2, 2, 3, 3)))
+      },
+      testM("fails if one of the effects fails") {
+        val list = List("1", "h", "3")
+        val res  = IO.foreachFlatten(list)(x => IO.effectTotal[List[Int]](List(x.toInt, x.toInt)))
+        assertM(res.run)(dies(isSubtype[NumberFormatException](anything)))
+      } @@ zioTag(errors)
+    ),
     suite("foreach_")(
       testM("runs effects in order") {
         val as = List(1, 2, 3, 4, 5)
@@ -3731,6 +3752,9 @@ object ZIOSpec extends ZIOBaseSpec {
 
   def functionIOGen: Gen[Random with Sized, String => Task[Int]] =
     Gen.function[Random with Sized, String, Task[Int]](Gen.successes(Gen.anyInt))
+
+  def functionListIOGen: Gen[Random with Sized, String => Task[List[Int]]] =
+    Gen.function[Random with Sized, String, Task[List[Int]]](Gen.successes(Gen.listOfN(10)(Gen.anyInt)))
 
   def listGen: Gen[Random with Sized, List[String]] =
     Gen.listOfN(100)(Gen.alphaNumericString)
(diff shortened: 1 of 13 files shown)


jdegoes:
This would probably be called `foreachFlatten` or `foreachFlat`, if it existed. I have no opinion on whether to add it or not, I have needed it a couple times before. The drawback is that as `foreach` itself comes in many variants, it would be tempting to expand this to include those variants too over time (and that, I think, would definitely be too much). But good to merge with the name change and if others are interested in the functionality. 👍 

Thank you for your work on this! 🙏

adamgfraser:
@blast-hardcheese Thanks for your work on this!

I agree with @jdegoes about the naming.

I'm also concerned with the method proliferation here. I've run into this occasionally and it is slightly shorter than `foreach(f).map(_.flatten)` but not by much. It is nice to have single method variants for common things but there is also a cost at some point to the sheer number of methods on ZIO.

I don't think adding this and not adding other variants (e.g. `foreachFlattenPar`, `foeachFlattenParN`, all the overloads of those for various collection types like `Set`, `Array`, and `NonEmptyChunk`) is really a tenable position because there isn't a logical basis for turning down a PR to add any of those individually.

At the same time, I share the instinct that those are too many methods and feel like that is an symptom of us mixing two different concerns here. We have one issue of how we want to perform effects for a bunch of different elements in a collection, described by `foreachPar`, `foreachParN`, and friends.. We have a second issue of how we want to flatten nested effects, which is described by `map(_.flatten)`. These two concerns are independent of each other (doing an effect for a bunch of things doesn't care what we do with it next and we might want to flatten a collection of collections regardless of how it was produced). So by combining them we are just taking one thing that had N methods and another thing that had 1 method and creating another N * 1 methods.

adamgfraser:
@blast-hardcheese I would do `foreachFlatten` if we are going to do anything here.

blast-hardcheese:
@adamgfraser Yeah, the proliferation problem is quite a challenging one to tackle. Even when adding these methods, it didn't feel "right" to leave off the `Par` (though fortunately the `_`-suffixed variants weren't really relevant here).

As much as I would use this functionality, being a big proponent of `flatTraverse` and preferring somewhat more of a "fluent"-style data flow, I can understand if this is an overall 👎 on merging this into ZIO core.

That being said, these methods really do feel somewhat homeless, in the sense that if they don't belong in core, and they aren't "functional", so they really don't belong in prelude, where could they go? I would almost want a `zio-extras` or similar, but then you end up muddying the water of which variant of zio are you using, doing damage to the userbase by fracturing it into successively smaller groups.

Is there a solution here?

blast-hardcheese:
Closing for now, this is a nice to have not a must have, and the concerns about method proliferation are valid.

PR #5298, opened 2021-07-12
Implement ZStream#serviceWithStream

adamgfraser:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4093,36 +4093,44 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
   /**
-   * Accesses the specified service in the environment of the effect.
+   * Accesses the specified service in the environment of the stream.
    */
   def service[A: Tag]: ZStream[Has[A], Nothing, A] =
     ZStream.access(_.get[A])
 
   /**
-   * Accesses the specified services in the environment of the effect.
+   * Accesses the specified services in the environment of the stream.
```
jdegoes, line 4102: This is a longer conversation but we need to delete all these arity versions and require users expand multiple service dependencies in a for comprehension.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4278,6 +4286,13 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
       ZStream.fromEffect(ZIO.serviceWith(f))
   }
 
+  final class ServiceWithStreamPartiallyApplied[Service](private val dummy: Boolean = true) extends AnyVal {
+    def apply[E, A](f: Service => ZStream[Has[Service], E, A])(implicit
```
jdegoes, line 4290: I am not sure about this. If the user is using our pattern, why would they need this?

```scala
for {
  serviceA <- ZStream.service[ServiceA]
  serviceB <- f(serviceA)
} yield serviceB
```

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Yes.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: It the counterpart of the `serviceWith` operator. So I think it comes up in two cases.

First, implementing service accessors:

```scala
trait Persistence {
  def loadOrders(customerID: CustomerId, from: OrderId, to: OrderId): ZStream[Any, Nothing, Customer]
}

object Persistence {
  def loadOrders(customerID: CustomerId, from: OrderId, to: OrderId): ZStream[Has[Persistence], Nothing, Customer] =
    ZStream.serviceWithStream(_.loadOrders(customerId, from, to))
}
```

Second, if you are working with a service in normal application code and not in a service implementation, for example in test code, it can be useful to access a service in the environment do something with it that returns a stream.

We have `serviceWith` and `serviceWithManaged` so I think either we should have this for symmetry or we should revisit those operators. I agree all of these accessors are becoming less important as we access services more within other services but still seems like they are useful in some cases.

jdegoes (approved):

PR #5300, opened 2021-07-12
Add ZIO.ifF

nryanov:
To be able to do something like this for pure values:
```
ZIO.ifF(ZIO.effect(...))(1, 2)
// instead of:
ZIO.ifM(ZIO.effect(...))(ZIO.succeed(1), ZIO.succeed(2))
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1127,6 +1127,16 @@ object ZIOSpec extends ZIOBaseSpec {
         ZIO.succeed(assertCompletes)
       }
     ),
+    suite("ifF")(
+      testM("returns `onTrue` if result of `b` is `true`") {
+        val zio = ZIO.ifF(ZIO.succeed(true))(true, false)
+        assertM(zio)(isTrue)
+      },
+      testM("returns `onFalse` if result of `b` is `false`") {
+        val zio = ZIO.ifF(ZIO.succeed(false))(true, false)
+        assertM(zio)(isFalse)
+      }
+    ),
     suite("ignore")(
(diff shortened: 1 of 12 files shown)


adamgfraser:
@nryanov Thanks for your work on this!

This operator would logically be called `ZIO.if` except that `if` is a reserved keyword in Scala. I think `ifF` is not very accessible (we're actually renaming everything with the `M` suffix in ZIO 2.0 so `ifM` is going to `ifZIO`). Perhaps this could be called `ZIO.ifThen` or `ZIO.conditional` but I wonder somewhat if having another variant of these methods is worthwhile

There is definitely an advantage to not having to do the `ZIO.succeed` in situations where the condition is effectual but the branches are not. At the same time I'm not sure how often that occurs and how great the benefit is when this operator is going to have another different name people have to find relative to doing `zio.map(b => if (b) 1 else 2)`, especially if the name is going to be less obvious given the keyword issue above.

joroKr21:
Even `zio.map(if (_) 1 else 2)` works. But I find that most of the time one branch is an effect and the other might not be (but then which one?)

adamgfraser:
Yes which I think is an argument for sticking with the existing `ifM` operator since you can always lift things into a `ZIO` but you can't go the other way around and I think having variants for different combinations if effectual and non-effectual branches is way too far.

nryanov:
Thank you for reviews!
The situation when it's needed to use pure values with effectful condition is not very often and definitely could be solved using `zio.map(b => if (b) 1 else 2)` and in this case something like `IfF` would be an overkill and just syntax-sugar.

One of the reason of this proposal was to make it possible to fully stick to zio combinators. But i agree that such new combinators much like some special cases and probably not for general usage

PR #5301, opened 2021-07-12
Add ZIO.ifElse

nryanov:
Currently, to achieve logic like: `if (b) onTrue else onFalse` i'll do:
```
ZIO.ifM(ZIO.succeed(b))(onTrue, onFalse)
```
and that is ok, but there is already `when` combinator which operates on a pure boolean and i think that something like `ifElse` also could be useful:

```
ZIO.ifElse(b)(onTrue, onFalse)
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -1105,6 +1105,16 @@ object ZIOSpec extends ZIOBaseSpec {
         assertM(ZIO.fail("Fail").head.either)(isLeft(isSome(equalTo("Fail"))))
       } @@ zioTag(errors)
     ),
+    suite("ifElse")(
+      testM("runs `onTrue` if result of `b` is `true`") {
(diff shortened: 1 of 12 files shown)


adamgfraser:
Hi @nryanov! Thanks for your contribution. It looks like this is the same as `if (b) onTrue else onFalse`. If so I tend to think it doesn't make sense to implement our operator that is the same as a built in language construct.

nryanov:
@adamgfraser Yes, this is the same as `if (b) ... else ...`. Initially, my point was to add syntactic sugar for this logic to be able to write it in a shorter way (especially in for-comprehension). But maybe it's on overkill for this

adamgfraser:
@nryanov Yeah the comparison would be:

```scala
if (b) onTrue else onFalse // before
ZIO.ifElse(b) onTrue else OnFalse // after
```

So it seems like it would actually be longer and I think there is also a significant cost to having a second way to do something that is already built into the language.

nryanov:
Re-thought it again with a fresh eye and your example -- it's definitely an overkill

PR #5306, opened 2021-07-14
Snapshot tests

fokot:


test/js/src/main/scala/zio/test/FileIOPlatformSpecific.scala
```diff
@@ -0,0 +1,91 @@
+package zio.test
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSImport
+import zio.{Task, ZIO}
+
+
+trait FileIOPlatformSpecific {
```
adamgfraser, line 8: We already have operators for working with files in ZIO core for the JVM. Can we create analogous versions of these for Scala.js so we can make this available to users more generally in a platform agnostic way and get this logic out of ZIO Test?

adamgfraser:
@fokot My main question is the best way to fit this in. Right now this is available directly in any spec which makes it easily accessible but pushes a good amount of complexity into the implementation of every runnable spec for what is a fairly specific feature. What do you think the trade offs are of doing this versus having a separate `SnapshotSpec` that extends `AbstractRunnableSpec` and then more of the logic goes in there.

test/js/src/main/scala/zio/test/FileIOPlatformSpecific.scala
fokot: Yes, I will move it there.

fokot:
@adamgfraser I can move it to to `SnapshotSpec`. It has very specific use I think it will be ok. I put the code into `AbstractRunnableSpec` because it is the type of spec in the sbt task. But I can add pattern match on type here https://github.com/zio/zio/pull/5306/files#diff-713d1e1dd3624b2d85413039827121367b50285c56af5e528e6a742a53b05c94R27

fokot:
@adamgfraser I created `AbstractRunnableSpec` and also fixed the test summary. So now you get the message that now tests were run when fixing the snapshots which is correct.

Question: I wanted to move file IO to the core, but the return type for reading is `def readFile(path: String): ZManaged[Blocking, IOException, ZInputStream]` and similar for writing. Node file api works with `Byte arrays` or with `Strings`. Can I create new simplified methods `def readFileToString(path: String): ZIO[Blocking, IOException, String]` for JVM and JS or should I rather implement JS with streams?

TODO: I want to have better error message when running snapshots, but it is then harder to get the right value.

adamgfraser:
@fokot I think ideally we would have the same interface on the JVM and Scala.js. If you think that is relatively straightforward to do we can do it here or otherwise we can open an issue for it and merge this once everything else is ready.

PR #5309, opened 2021-07-17
Update Runtime.scala

tusharmath:
The idea behind this change is to optimize evaluation for very small ZIO programs eg: `UIO(1)`.

By default, the current thread should be used (better performance). In case the developer want's to ensure the execution happens in the ZIO thread pool, they are free to use `yieldNow` in their program.

This optimization also works much better with single-threaded frameworks such as netty and nio.


adamgfraser (commented):
@tusharmath This is going to be the behavior in ZIO 2.0 but I think this is a semantic change that we probably shouldn't just push through for all users in a minor release. We could potentially make this a setting in the platform that is configurable. Do you think it makes sense to do that or just wait for ZIO 2.0?

PR #5226, opened 2021-06-20
Rename Operators For ZIO 2.0

adamgfraser:
Resolves #5221. Covers the items listed there. Still need to look at other effect constructors and eliminate more `_` suffix variants but may make sense to merge given the scope of the changes and do those as a follow up.

Issue #5221, opened 2021-06-18
ZIO 2.0 Name Changes


jdegoes:
We should rename the old methods to the new names, but also, create the old methods that forward to the new methods, and use the `@deprecated` annotation, marking 2.0.0 as the deprecation version.

 - `ZIO.effect` => `ZIO.attempt`
 - `ZIO.effectTotal` => `ZIO.succeed`
 - `ZIO.effectSuspend` => `ZIO.suspend`
 - `ZIO.effectSuspendTotal` => `ZIO.suspendSucceed`
 - `ZIO.effectSuspendWith` => `ZIO.suspendWith`
 - `ZIO.effectSuspendTotalWith` => `ZIO.suspendSucceedWith`
 - `ZIO#run` => `ZIO#exit`
 - `ZManaged.make` => `ZManaged.bracket`
 - `ZIO#ensuring_` => `ZIO.ensuring`
 - `ZIO#ensuring` => `ZIO.ensuringWith`
 - `ZIO#toManaged_` => `ZIO#toManaged`
 - `ZIO#toManaged` => `ZIO#toManagedWith` / `ZIO#bracketManaged`
 - `putStrLn` => `printLine`
 - `getStrLn` => `readLine`
 
More generally, there is an intention to replace all `_` methods in ZIO 2.0 (`bracket_`, etc.). This naming convention stems from Haskell and has no analogues in the JVM.

adamgfraser:
I will do this.

jdegoes:
Awesome, thank you! There will be variations of some of these methods (e.g. ZIO#run).

jdegoes:
For posterity, @kitlangton also suggested `ZManaged` => `ZResource` (with pros & cons).

kitlangton:
Also, possibly:

`environment`
`access` -> `environmentWith`
`accessM` -> `environmentWithM`

to be in line with

`service`
`serviceWith` (Though there is no `serviceWithM` variant, to encourage the recommended module pattern. `serviceWith` expects a ZIO to be returned)

adamgfraser:
Closed by #5229.

PR #5330, opened 2021-07-27
2.0 Renaming & Deprecation Bike-Shed Party 💃🕺 

kitlangton:
Starting a PR to discuss further `2.0` name changes and deprecations. 

The goal would be to help with API discovery by deprecating duplicated and/or non-idiomatic methods as well as rarely used symbolic aliases. Arrow combinator methods such as `andThen` show up in autocomplete, and will not (99% of the time) be what the user is actually seeking. Furthermore, we can continue to solidify naming patterns, ensuring each common prefix or suffix has a consistent, shared meaning across the codebase.

ZIO's API is already one of the best I've ever used, with a incredible focus on consistency and discoverability. As a ZIO user, I've often had the experience of hoping that some potential method existed, and after typing a few letters, having autocomplete suggest precisely what I was looking for. There's no better _DX_ than sensing an API is pre-anticipating your needs. I believe this is achieved through consistent and thoughtful API design, where the user subconsciously observes patterns of naming/meaning as they use the library, allowing them to anticipate and _reach out_ for similar methods they've never seen before.

So yeah, ZIO 2.0 presents the opportunity to achieve even more consistency, to step back and observe new patterns, and to codify these new patterns with shared pre/postfixes and APIs. Much work has already been done on this front (#5226 and #5221), but there's always more.

What I've done in this PR:

1. Remove the Arrow combinators (`&&&`, `+++`, `|||`, `onSecond`, `onFirst`, `second`, `first`, `andThen`, `>>>`, `compose`, `<<<`).
    - While lawful and interesting, these methods do not reflect how the environment is idiomatically used in the vast majority of applications. Users exploring the codebase (a great way to learn) might be misled by their existence, they also consume precious auto-complete real estate.
  
2. Remove duplicate methods.
    - `get` was essentially a more constrained version of `some`
    - `require` was essentially an awkward variant of `someOrFail`

3. Nix `>>=`.
    - We should probably cut ties with the rest of the Haskellisms, now that `putStrLn` and `getStr` have gone away. Using `flatMap` is only a few extra characters, and won't frighten non-Haskellers.

4. Some v. Partial
    - I haven't made this change yet, but I'd like to discuss it. `Some` is mostly used for combinators involving `Option` (`someOrFail`, `some`, `someOrFailException`, 'asSome', 'asSomeError'), yet methods `tapSome`, `catchSomeCause`, 'catchSome` also use the same modifier and deal with PartialFunctions.
    - It would be great to give the pattern of PartialFunction-accepting variants its own discrete suffix. I think `catchPartial`, `tapPartial`, whilst slightly more verbose, would remove the ambiguity.
    
5. `as`, `to`, `into` prefixes
    - `as` is used when transforming the `A` inside of a ZIO, generally as shortcuts for `map(aToFoo(_))`. `asSome`, `asRight`, etcetera.
    - `to` is when the ZIO is transformed into something else, `.toLayer`, `.toManaged`. `.toFuture` is slightly more interesting, as you get back a `URIO[R, Future]`, but that's mostly to maintain purity.
    - Currently there's a `ZIO.to` method that doesn't follow either of the above patterns. It instead accepts a `Promise`, completing that promise with the result of the ZIO. This pattern of accepting some secondary data-type, modifying it with the result of the current effect, already exists in ZStream as `intoHub`, `intoQueue`. So I've renamed `to` -> `intoPromise`. 
    
6. `collect` v. `collectAll`
    - `collect` is a very common Scala collections method, accepting a partial function. A variant exists on the `ZIO` trait that performs the expected combination of map/filter using a partial function. However, we also have `collectAll`, which is the `ZIO` equivalent of Haskell's `sequence`. `sequence` is truly a horrible name, as there's often a `sequencePar` variant (and even `sequenceParUnordered` in some libraries), which is oxymoronic. So, while `collectAll` is a much better name than `sequence`, it has this unfortunate ambiguity with Scala's famous `collect` operator. Can we think of a better name here and vanquish this ambiguity? 😆 
  
Okay, that was a lot. Hoping to iterate a lot on this PR and consolidate all the naming bike-sheddery.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -16,6 +16,7 @@
 
 package zio
 
+import zio.ZIO.intoPromise
 import zio.internal.tracing.{ZIOFn, ZIOFn1, ZIOFn2}
 import zio.internal.{Executor, Platform}
 import zio.{TracingStatus => TracingS}
@@ -79,10 +80,11 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Sequentially zips this effect with the specified effect, combining the
    * results into a tuple.
    */
+  @deprecated("use zip", "2.0.0")
   final def &&&[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B])(implicit
     zippable: Zippable[A, B]
   ): ZIO[R1, E1, zippable.Out] =
-    self.zipWith(that)((a, b) => zippable.zip(a, b))
+    self zip that
 
   /**
    * Returns an effect that executes both this effect and the specified effect,
@@ -96,6 +98,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Splits the environment, providing the first part to this effect and the
    * second part to that effect.
    */
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def ***[R1, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[(R, R1), E1, (A, B)] =
     (ZIO.first >>> self) &&& (ZIO.second >>> that)
 
@@ -109,6 +112,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Depending on provided environment, returns either this one or the other
    * effect lifted in `Left` or `Right`, respectively.
    */
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def +++[R1, B, E1 >: E](that: ZIO[R1, E1, B]): ZIO[Either[R, R1], E1, Either[A, B]] =
     ZIO.accessZIO[Either[R, R1]](_.fold(self.provide(_).map(Left(_)), that.provide(_).map(Right(_))))
 
@@ -138,12 +142,12 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
     self.flatMap(new ZIO.ZipLeftFn(() => that))
 
   /**
-   * Alias for `&&&`.
+   * Symbolic alias for `zip`.
    */
   final def <*>[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B])(implicit
     zippable: Zippable[A, B]
   ): ZIO[R1, E1, zippable.Out] =
-    self &&& that
+    self zip that
 
   /**
    * A symbolic alias for `orElseEither`.
@@ -160,12 +164,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Operator alias for `orElse`.
    */
+  // TODO: This usually the monoid operator
   final def <>[R1 <: R, E2, A1 >: A](that: => ZIO[R1, E2, A1])(implicit ev: CanFail[E]): ZIO[R1, E2, A1] =
     orElse(that)
 
   /**
    * A symbolic alias for `raceEither`.
    */
+  @deprecated("use raceEither", "2.0.0")
   final def <|>[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[R1, E1, Either[A, B]] =
     self.raceEither(that)
 
@@ -176,11 +182,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * val parsed = readFile("foo.txt") >>= parseFile
    * }}}
    */
+  @deprecated("use flatMap", "2.0.0")
   final def >>=[R1 <: R, E1 >: E, B](k: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] = flatMap(k)
 
   /**
    * Operator alias for `andThen`.
    */
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def >>>[E1 >: E, B](that: ZIO[A, E1, B]): ZIO[R, E1, B] =
     self.flatMap(that.provide)
 
@@ -203,6 +211,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Depending on provided environment returns either this one or the other effect.
    */
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def |||[R1, E1 >: E, A1 >: A](that: ZIO[R1, E1, A1]): ZIO[Either[R, R1], E1, A1] =
     ZIO.accessZIO(_.fold(self.provide, that.provide))
 
@@ -292,6 +301,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
       }
     )(use)
 
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def andThen[E1 >: E, B](that: ZIO[A, E1, B]): ZIO[R, E1, B] =
     self >>> that
 
@@ -599,7 +609,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def collectZIO[R1 <: R, E1 >: E, B](e: => E1)(pf: PartialFunction[A, ZIO[R1, E1, B]]): ZIO[R1, E1, B] =
     self.flatMap(v => pf.applyOrElse[A, ZIO[R1, E1, B]](v, _ => ZIO.fail(e)))
 
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def compose[R1, E1 >: E](that: ZIO[R1, E1, R]): ZIO[R1, E1, A] = self <<< that
+
+  /**
+   * Transforms the environment.
+   */
+  final def contramap[R0](f: R0 => R)(implicit ev: NeedsEnv[R]): ZIO[R0, E, A] =
+    ZIO.accessZIO(r0 => self.provide(f(r0)))
 
   /**
    * Taps the effect, printing the result of calling `.toString` on the value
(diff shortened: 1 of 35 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -176,11 +182,13 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * val parsed = readFile("foo.txt") >>= parseFile
    * }}}
    */
+  @deprecated("use flatMap", "2.0.0")
   final def >>=[R1 <: R, E1 >: E, B](k: A => ZIO[R1, E1, B]): ZIO[R1, E1, B] = flatMap(k)
```
adamgfraser, line 186: 👍 for removing this. This one is a little unusual in that it combines values of two different types unlike other binary operators. It is also something that you actually see used sometime which can make code bases less accessible.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -160,12 +164,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
...
    */
+  // TODO: This usually the monoid operator
   final def <>[R1 <: R, E2, A1 >: A](that: => ZIO[R1, E2, A1])(implicit ev: CanFail[E]): ZIO[R1, E2, A1] =
     orElse(that)
 
   /**
    * A symbolic alias for `raceEither`.
    */
+  @deprecated("use raceEither", "2.0.0")
   final def <|>[R1 <: R, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[R1, E1, Either[A, B]] =
```
adamgfraser, line 175: I would be inclined to keep this since `race` is actually a relatively commonly used operator on `ZIO` and it has a nice symmetry with `<&>`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -160,12 +164,14 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Operator alias for `orElse`.
    */
+  // TODO: This usually the monoid operator
   final def <>[R1 <: R, E2, A1 >: A](that: => ZIO[R1, E2, A1])(implicit ev: CanFail[E]): ZIO[R1, E2, A1] =
```
adamgfraser, line 168: I think this one is also sometimes actually used. It also actually is an associative binary operator and the most information preserving one you can define.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -96,6 +98,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Splits the environment, providing the first part to this effect and the
    * second part to that effect.
    */
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def ***[R1, E1 >: E, B](that: ZIO[R1, E1, B]): ZIO[(R, R1), E1, (A, B)] =
```
adamgfraser, line 102: I think fine to get rid of the arrow operators. Maybe just delete becauseI don't think these are being used and there isn't really a direct alternative.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -599,8 +609,15 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   final def collectZIO[R1 <: R, E1 >: E, B](e: => E1)(pf: PartialFunction[A, ZIO[R1, E1, B]]): ZIO[R1, E1, B] =
     self.flatMap(v => pf.applyOrElse[A, ZIO[R1, E1, B]](v, _ => ZIO.fail(e)))
 
+  @deprecated("the Arrow combinators are deprecated", "2.0.0")
   final def compose[R1, E1 >: E](that: ZIO[R1, E1, R]): ZIO[R1, E1, A] = self <<< that
 
+  /**
+   * Transforms the environment.
+   */
+  final def contramap[R0](f: R0 => R)(implicit ev: NeedsEnv[R]): ZIO[R0, E, A] =
```
adamgfraser, line 618: Not sure there is a lot of value in changing this if we are going to leave `provide`. If we want to take back the `provide` name and use it for new layer operators I think that is interesting but I think otherwise we're just fragmenting things splitting up the old layer operators between `provide` and `contramap` versus `provide` and `provideSome`.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -733,6 +750,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Supplies `zio` if the predicate fails.
    */
+  // TODO: Should we have a `filterOrElseZIO` and `filterOrElse`?
   final def filterOrElse[R1 <: R, E1 >: E, A1 >: A](p: A => Boolean)(zio: => ZIO[R1, E1, A1]): ZIO[R1, E1, A1] =
```
adamgfraser, line 754: Methods like this already have so many variants I would be inclined to not add more. It is already quite easy to do this but evaluating an effectual condition and then doing something else based on it.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -993,6 +1011,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
   /**
    * Unwraps the optional success of this effect, but can fail with an None value.
    */
+  @deprecated("use some", "2.0.0")
   final def get[B](implicit
```
adamgfraser, line 1015: Agreed.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1003,6 +1022,7 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Returns a successful effect with the head of the list if the list is
    * non-empty or fails with the error `None` if the list is empty.
    */
+  // TODO: Should we change this to work with any Iterable?
   final def head[B](implicit ev: A IsSubtypeOfOutput List[B]): ZIO[R, Option[E], B] =
```
adamgfraser, line 1026: I don't know if we want to encourage potentially inefficient data access patterns. The promise of an `Iterable` is really only that you can iterate over it. If this was a frequently used method would be happy to put that aside for ergonomics but don't know if that is really the case.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -1180,6 +1201,8 @@ sealed trait ZIO[-R, +E, +A] extends Serializable with ZIOPlatformSpecific[R, E,
    * Executes the effect on the specified `ExecutionContext` and then shifts back
    * to the default one.
    */
+  // TODO: Could we use the same `lock` name?
+  //  Instead of two different names for the same (somewhat advanced) method.
   final def on(ec: ExecutionContext): ZIO[R, E, A] =
```
adamgfraser, line 1206: I think we have found overloading itself created issues, it is nice to say `for X to do A, for Y do B` rather than explain that there are actually two different variants of `A` that do different things.  We're moving away from that a little with `from` but I think that is quite explicitly saying "give me everything and I will turn it into a ZIO".

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -1947,10 +1947,11 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * Enqueues elements of this stream into a queue. Stream failure and ending will also be
    * signalled.
    */
+  @deprecated("use intoQueue", "2.0.0")
   final def into[R1 <: R, E1 >: E](
```
adamgfraser, line 1951: Probably good clarification here.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4676,6 +4721,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * Returns an effectful function that extracts out the second element of a
    * tuple.
    */
+  @deprecated("use fromFunction(_._2)", "2.0.0")
```
adamgfraser, line 4724: Even `fromFunction` would be a fairly unusual operator to use.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4301,6 +4341,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   /**
    * Sequentially zips the specified effects. Specialized version of mapN.
    */
+  // TODO: Rename these all `ZIO.zip`
   def tupled[R, E, A, B](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B]): ZIO[R, E, (A, B)] =
```
adamgfraser, line 4345: I think we can delete all of these now that with have compositional zips.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4348,6 +4389,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
    * Sequentially zips the specified effects using the specified combiner
    * function.
    */
+  // TODO: Rename these all `ZIO.zipWith`
   def mapN[R, E, A, B, C](zio1: ZIO[R, E, A], zio2: ZIO[R, E, B])(f: (A, B) => C): ZIO[R, E, C] =
```
adamgfraser, line 4393: Same here.

jdegoes:
> Remove the Arrow combinators (&&&, +++, |||, onSecond, onFirst, second, first, andThen, >>>, compose, <<<).

Neutral to 👍 .

> get was essentially a more constrained version of some
> require was essentially an awkward variant of someOrFail
> Nix >>=.

👍 

> Some v. Partial

Neutral to 👎 on this. We know what `Partial` means because we are functional programmers and know the difference between _partial_ and _total_ functions. But `updateSome`, for example, directly says that it will update some cases but not others (versus performing a _partial update_, which would have a different meaning).

That said, I do recognize the argument for consistency with `ZIO#some` and others, but I wonder if there are any cases of ambiguity where the meaning is not clear or the same named method could mean two useful things.

> Currently there's a ZIO.to method that doesn't follow either of the above patterns. It instead accepts a Promise, completing 
> that promise with the result of the ZIO. This pattern of accepting some secondary data-type, modifying it with the result of the 
> current effect, already exists in ZStream as intoHub, intoQueue. So I've renamed to -> intoPromise.

Overall 👍 , but I think in cases like this, overloading can be used to avoid creating too many variants. `mapZIO` and `mapStream` have  `Function` as their argument so they all erase the same, but `into` would accept different (unerased) types, so we could use an overloaded `into` for all of them.

> So, while collectAll is a much better name than sequence, it has this unfortunate ambiguity with Scala's famous collect 
> operator. Can we think of a better name here and vanquish this ambiguity? 😆

I don't mind `collectAll` although the overlap with Scala's `collect` is unfortunate. `executeAll` might work, or `doAll` or `awaitAll`.

adamgfraser (approved):

PR #5340, opened 2021-07-29
Support Serializing Schedule State

### LINKED ISSUE #5337: ZIO 2.0: Make Schedule serializable
The current encoding of Schedule makes state implicit, which cleans up the implementation for operators and constructors, but means there is not a clear path to making schedule state serializable.

By reverting to a more "unfold" like definition, e.g.:

```scala
trait Schedule[-Env, -In, +Out] {
  type State
  def initial: State
  def step(OffsetDateTime, In, State): ZIO[Env, Nothing, (State, Out, Schedule.Decision)]
}
...
sealed trait Decision { ... }
object Decision {
  case object Done extends Decision
  final case class Continue(
    interval: Interval
  ) extends Decision
}
```

We make it possible to think about serialization. Now, updating the logic for all the schedule operators and constructors is not trivial, but it will always be possible since the two ways to encode schedules are equivalent (implicit vs explicit state).

In order to handle serialization "from the outside", we would need to type every constructor for schedule to return a more specific type. The first step in this process would be introducing a type alias for the more specific type, inside the `Schedule` companion object:

```scala
object Schedule {
  type WithState[State0, -In0, +Out0] = Schedule[In0, Out0] { type State = State0 }
}
```

The second step would be changing any return type in any method inside the `Schedule` trait or inside the `Schedule` companion object from `Schedule[A, B]` to `Schedule.WithState[S, A, B]` for some state type `S`.

Then, with all methods and constructors precisely typed, serialization can happen in user-land code according to whatever method the user desires (a `ZIO Schema` integration would be ideal).

Probably we would modify the `Driver` to expose the state of the schedule to facilitate serialization in-progress.

adamgfraser:
Resolves #5337. Still need to maintain type information on the state type but the encoding has been changed to use explicit state and all tests are passing.


adamgfraser:
@jdegoes Okay, I still want to clean this up a little but this preserves type information now and I implemented an operator on the driver to expose the current state.

Some thoughts on the new encoding:

**Positives**
- It supports serialization of the state, which was the goal
- I think it is actually more straightforward to implement operators since you just have to think of one step of the stat
- Composition of the state type works flawlessly

**Negatives**
- By necessity this exposes what was previously an internal implementation detail of the state type, which puts some limits on ability to refactor in the future
- Related to this, it exposes the "provenance" of composed types in a a way that didn't exist before, so I think there are several operators that I would want to reimplement directly to avoid introducing unnecessary state
- Actual state types are not particularly meaningful, though not sure if this is a problem if we just want to be able to serialize and deserialize them
- In a couple of cases this triggers spurious warnings that a type is being inferred as `Any`, even though the type is being inferred flawlessly and `Any` is the accurate type for that parameter

jdegoes (approved):

PR #5357, opened 2021-08-07
[RFC] Test on Java 17 and adjust the build/test matrix

sideeffffect:
This restructures the build/test matrix to be more sparse, so that when we add yet another Java version, the number of builds doesn't explode on us.

Don't test on all the Cartesian product of Scala versions, Target platforms and JVM versions.
 * Test all Scala versions on Scala/JVM and Java 8.
 * Test Scala 2.13 on Scala/JVM, Scala.js and Scala Native and Java 8.
 * Test Scala 2.13 on Scala/JVM and Java 8, 11 and 17.


sideeffffect:
Ha! This seems to fail on Java 17 (LTS, to be released in September).
```
[info]     - converts Cause to original exception with ZTraces in root cause
[info]       Fiber failed.
[info]       An unchecked error was produced.
[info]       java.lang.NullPointerException: Cannot invoke "java.lang.Throwable.getMessage()" because the return value of "scala.Function0.apply()" is null
[info]       	at zio.test.Assertion$.$anonfun$hasMessage$1(Assertion.scala:211)
```
/cc @adamgfraser @jdegoes

sideeffffect:
Good, https://github.com/zio/zio/pull/5364 has indeed fixed the issue on Java 17 :tada:

adamgfraser:
Nice!

sideeffffect:
@mijicd What do you think about this setup?

mijicd:
I like it.

mijicd (approved):

PR #5362, opened 2021-08-07
ZIO 2.0: Allow type annotation for empty Chunk & Cause

kitlangton:
Most `empty` methods allow you to specify the type. If you leave this off, it is inferred as `Nothing` as before. 

This is mostly useful in `foldLeft`, so you can write `foldLeft(Chunk.empty[String])`.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -540,7 +540,7 @@ sealed abstract class Cause[+E] extends Product with Serializable { self =>
 }
 
 object Cause extends Serializable {
-  val empty: Cause[Nothing]                               = Internal.Empty
+  def empty[A]: Cause[A]                                  = Internal.Empty
   def die(defect: Throwable): Cause[Nothing]              = Internal.Die(defect)
   def fail[E](error: E): Cause[E]                         = Internal.Fail(error)
   def interrupt(fiberId: Fiber.Id): Cause[Nothing]        = Internal.Interrupt(fiberId)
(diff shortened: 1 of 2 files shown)


adamgfraser:
Not sure about this one.

You can do the same thing by specifying the type parameter to `foldLeft`, though this can be nicer if the state type you are returning in `foldLeft` is a larger structure to avoid having to specify the full type.

At the same time, there is only one empty chunk or empty cause, which the current signature reflects and adding additional type parameters usually creates worse type inference in at least some situations.

It also seems to create some lack of consistency. `Chunk` and `Cause` constructors now take this additional type parameter which really only has use for this specific situation but other constructors don't. However, there are a variety of other cases where you can run into the same thing. For example, `ZIO.succeed` sets the environment type to `Any` and the error type to `Nothing` which is typically too narrow if you are folding over a collection of `ZIO` effects with a summary value.

So while the need to specify type parameters to `foldLeft` is definitely not ideal for the Scala language I'm not sure it is a problem we can solve at our level in a principled way.

kitlangton:
I think my main argument for it would be that this is the signature of most other standard Scala `.empty` methods (`List`, `Option`, etc., which also have singleton empty values which could technically be vals). So, I suppose it's a "POLA" (that fun acronym I just learned about) thing. But also, happy either way 😸

adamgfraser:
Yeah, I would be inclined to say that isn't best practice and to keep the existing type signature but I hear you.

kitlangton:
I'm cool with that 😄

PR #5363, opened 2021-08-07
Use SourceLocation for assertTrue

kitlangton:
Besides a simplification, this will allow users to use `SourceLocation` in custom assertions, and get the location of each invocation of the custom assertion, rather than the source location of where the custom assertion uses `assertTrue`.


adamgfraser (dismissed):
Nice!

adamgfraser (approved):

PR #5364, opened 2021-08-07
Remove Illegal Reflective Access

adamgfraser:
We are currently using reflective access when we squash a `Cause` down to a `Throwable` to attach the rendered `Cause` as the cause of the original `Throwable`. We do our best to do this in a principled way by finding the root cause of the `Throwable` and then adding to that. However, the expectation in Java is that the cause will be set at most once and never changed and even if the root cause is `null`, typical `Throwable` constructors set the cause to `null` permanently rather than leaving it unset so we can set it later. We work around this by using reflective access to set the field directly.

However, this reflective access has triggered warnings for a long time and is not going to be possible on JDK 17, which is being released later this year.

Given this, and that I think we really do want to return one of the original `Throwable` values rather than creating an entirely new cause, this PR changes the implementation to instead attach the rendered `Cause` as a suppressed exception. This is not as ideal as the "caused by" chain indicates a linear relationship that exists here where the `Cause` is the literal cause of the exception, but I think it is the best we can do within the constraints of the `Throwable` API.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/CauseSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/CauseSpec.scala
@@ -219,12 +219,8 @@ object CauseSpec extends ZIOBaseSpec {
           assert(squashed)(
             equalTo(e) &&
               hasMessage(equalTo(throwableMessage)) &&
-              hasThrowableCause(
-                isSubtype[IllegalArgumentException](
-                  hasMessage(equalTo(causeMessage)) &&
-                    hasThrowableCause(hasMessage(equalTo(renderedCause)))
-                )
-              )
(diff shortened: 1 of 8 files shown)


jdegoes (approved):

dkarlinsky:
This indeed seems to be the the best that can be done within the constraints of the Throwable API, without resorting to hacks. 
Unfortunately for us, we use NewRelic for reporting errors, and their stack trace reporting ignores suppressed exceptions completely :(
Fortunately, the NewRelic java agent is open source, so I created a PR to fix this: https://github.com/newrelic/newrelic-java-agent/pull/405  :)
So hopefully soon, we (and perhaps some other zio users) will be able to upgrade to 1.0.11 without loosing fiber traces in New Relic reported errors.

adamgfraser:
@dkarlinsky Great! Thank you! 🙏

PR #5369, opened 2021-08-09
Fix Flaky Stream Tests

adamgfraser:
In `flatMapPar`, the managed permit is currently constructed in the context of the stream, which means then when we run the stream the permit can be released and the outer can interrupt the inner fibers before we can offer the error from running the inner stream. We run the stream in the `use` action of the managed permit to ensure that the inner error is signaled.


iravid (approved):
Great catch!

PR #5381, opened 2021-08-11
ZIO 2.0 Migration Guide

khajavi:
In this pull request, I'm going to write the ZIO 2.0 migration guide to make the transition easy and smooth.


adamgfraser:
This looks really good!

jdegoes:
Among items I'd love to see in the migration guide:

1. The _Why_: What are the big focus areas in ZIO 2.0, and why are they changing? Why not just leave everything the way it was before? How will the changes help companies scale ZIO into new domains and across larger teams and more complex scenarios?
2. Scalafix: How can companies leverage Scalafix migrations to do the upgrade work for them? (TBD)
3. Name changes: What are the top name changes that have been made to ZIO 2.0? What is a more comprehensive listing of all of them?
4. Semantical changes: What sorts of things have changed, which won't stop code from compiling, but which require updating anyway?
5. What are replacements for deprecated methods?
6. How has the module pattern changed? What is the new recommended pattern for interfaces, implementations, and accessors?


More coming...

khajavi:
Thank you, John!
I will work on these topics.

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,591 @@
...
+  override def log(line: String): UIO[Unit] =
+    for {
+      current <- clock.currentDateTime.orDie
+      _       <- console.putStrLn(current.toString + "--" + line).orDie
+    } yield ()
+}
+
+object LoggingLive {
+  val layer: URLayer[Has[Console] with Has[Clock], Has[Logging]] =
+    (LoggingLive(_, _)).toLayer
```
kitlangton, line 72: In practice, I've found it can help to annotate this, such that you can have the IDE [re]generate the type without narrowing it to `LoggingLive`.

```suggestion
    (LoggingLive(_, _)).toLayer[Logging]
```
```

File: docs/howto/migrate/migration-guide.md [hunk 2]
```diff
@@ -0,0 +1,591 @@
...
+
+To create the dependency graph in ZIO 1.x, we should compose the required layers manually. As the ordering of layer compositions matters, and also we should care about composing layers in both vertical and horizontal manner, it would be a cumbersome job to create a dependency graph with a lot of boilerplates.
+
+Assume we have the following dependency graph with two top-level dependencies:
+
+```
+           DocRepo                ++          UserRepo
+      ____/   |   \____                       /     \
+     /        |        \                     /       \
+ Logging  Database  BlobStorage          Logging   Database
```
kitlangton, line 122: Pretty :)

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,591 @@
...
+    |                    |                  |
+ Console              Logging            Console
+                         |       
+                      Console    
+```
```
kitlangton, line 127: Pretty 😍

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,591 @@
...
+## ZIO Services
+
+There are two significant changes in ZIO Services:
+
+| ZIO 1.x                  | ZIO 2.x                 |
+|--------------------------|-------------------------|
+| zio.Clock.Service.live   | zio.Clock.ClockLive     |
+| zio.Console.Service.live | zio.Console.ConsoleLive |
+| zio.System.Service.live  | zio.System.SystemLive   |
+| zio.Random.Service.live  | zio.Random.RandomLive   |
```
kitlangton, line 348: I think you can get rid of this, as it's listed below on line 361.

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,591 @@
...
+    |--------------------------|-------------------------|
+    | zio.Clock.Service.live   | zio.Clock.ClockLive     |
+    | zio.Console.Service.live | zio.Console.ConsoleLive |
+    | zio.System.Service.live  | zio.System.SystemLive   |
+    | zio.Random.Service.live  | zio.Random.RandomLive   |
+
+
+2. In ZIO 2.0 all type aliases like `type Logging = Has[Logging.Service]` removed. So we should explicitly use `Has` wrappers when we want to specify dependencies on ZIO services.
+
+So instead of writing `ZLayer[Console with Clock, Nothing, ConsoleLogger]`, we should write `ZLayer[Has[Console] with Has[Clock], Nothing, Has[ConsoleLogger]]`.
```
kitlangton, line 372: Might also be worth mentioning that instead of things like `ZIO.service[Console.Service]`, we also now do `ZIO.service[Console]`.

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,591 @@
...
+val x3: Task[String]  = Task.succeed(???)
+val x4: Task[Boolean] = Task.succeed(???)
+```
+
+In ZIO 1.x, the output of zipping together these effects are nested:
+
+```scala
+val zipped = x1 <*> x2 <*> x3 <*> x4
+// zipped: ZIO[Any, Throwable, (((Int, Unit), String), Boolean)] = zio.ZIO$FlatMap@3ed3c202
+```
```
kitlangton, line 420: Task.succeed actually returns a `UIO`. Perhaps we can do:

```scala
val x1: UIO[Int]     = ZIO.succeed(???)
val x2: UIO[Unit]    = ZIO.succeed(???)
val x3: UIO[String]  = ZIO.succeed(???)
val x4: UIO[Boolean] = ZIO.succeed(???)
```

In ZIO 1.x, the output of zipping together these effects are nested:

```scala
val zipped = x1 <*> x2 <*> x3 <*> x4
// zipped: ZIO[Any, Throwable, (((Int, Unit), String), Boolean)] = zio.ZIO$FlatMap@3ed3c202
```
```

File: docs/howto/migrate/migration-guide.md (hunk 5 above)
Comment (khajavi) [reply docs/howto/migrate/migration-guide.md]: +1

File: docs/howto/migrate/migration-guide.md [hunk 7]
```diff
@@ -0,0 +1,782 @@
...
+
+## Automated Scalafix Rules
+TODO
+
+## ZLayer
+
+### Functions to Layers
+
+In ZIO 1.x, when we want to write a service that depends on other services, we need to use `ZLayer.fromSrevice*` variants with a lot of boilerplate:
```
guersam, line 26:
```suggestion
In ZIO 1.x, when we want to write a service that depends on other services, we need to use `ZLayer.fromService*` variants with a lot of boilerplate:
```

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1163 @@
...
+```scala mdoc:silent:nest
+def log(line: String): URIO[Has[Logging], Unit] = ZIO.serviceWith(_.log(line))
+```
+### Accessing Multiple Services in the Environment
+
+In ZIO 1.x, we could access multiple services using higher arity service accessors like `ZIO.services` and `ZManaged.services`:
+
+```scala mdoc:silent:nest
+for {
+  (console, random) <- ZIO.services[Console, Random]
```
domdorn, line 117: hmm.. I think this isn't working as expected cause of the `withFilter` thingy.. e.g. it was always 

```scala
for { 
  services <- ZIO.services[Console, Random]
  (console, random) = services
  ...
  }
  ```
  
  not sure if it matters in the migration guide, though.

File: docs/howto/migrate/migration-guide.md [hunk 9]
```diff
@@ -0,0 +1,1163 @@
...
+4. **Accessor Methods** — The new pattern reduced one level of indirection on writing accessor methods. So instead of accessing the environment (`ZIO.access/ZIO.accessM`) and then retrieving the service from the environment (`Has#get`) and then calling the service method, the _Module Pattern 2.0_ introduced the `ZIO.serviceWith` that is a more concise way of writing accessor methods. For example, instead of `ZIO.accessM(_.get.log(line))` we write `ZIO.serviceWith(_.log(line))`.
+
+   We also have accessor methods on the fly, by extending the companion object of the service interface with `Accessible`, e.g. `object Logging extends Accessible[Logging]`. So then we can simply access the `log` method by calling the `Logging(_.log(line))` method:
+   
+    ```scala mdoc:silent:nest
+    trait Logging {
+      def log(line: String): UIO[Unit]
+    }
+    
+    object Logging extends Accessible[Logging]
```
domdorn, line 552: we should mention if this replaces the `@accessible` macro that people know from `zio-macros`

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1163 @@
...
+    | `zio.Random.Service.live`  | `zio.Random.RandomLive`   |
+
+
+2. In ZIO 2.0 all type aliases like `type Logging = Has[Logging.Service]` removed. So we should explicitly use `Has` wrappers when we want to specify dependencies on ZIO services.
+
+So instead of writing `ZLayer[Console with Clock, Nothing, ConsoleLogger]`, we should write `ZLayer[Has[Console] with Has[Clock], Nothing, Has[ConsoleLogger]]`, or when accessing services instead of `ZIO.service[Console.Service]` we also now do `ZIO.service[Console]`.
+
+### Blocking Service
+
+Since there is rarely a need to use a separate blocking thread pool, ZIO 2.0 created _one global blocking pool_ removed the Blocking service from `ZEnv` and the built-in services.
```
domdorn, line 697: I think there's a word missing before "removed" .. I guess its `and`

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1163 @@
...
+
+All blocking operations were moved to the `ZIO` data type:
+
+| ZIO 1.x                   | ZIO 2.x |
+|---------------------------|---------|
+| `zio.blocking.Blocking.*` | `ZIO.*` |
+
+We can also provide a user-defined blocking executor in ZIO 2.x with the `Runtime#withBlockingExecutor` operator that constructs a new `Runtime` with the specified blocking executor.
```
domdorn, line 705: we should mention that we now have `ZIO.attemptBlocking`, `ZIO.attemptBlockingIO` etc. instead

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1163 @@
...
+// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+```
+
+## Removed Methods
+
+- **Arrow Combinators** — (`+++`, `|||`, `onSecond`, `onFirst`, `second`, `first`, `onRight`, `onLeft`, `andThen`, `>>>`, `compose`, `<<<`, `identity`, `swap`, `join`)
+
+As the module pattern in ZIO 2.0 encourages users to use `Has` with the environment `R` (`Has[R]`), it doesn't make sense to have arrow combinators. An arrow makes the `R` parameter as the _input_ of the arrow function, and it doesn't match properly with environments with the `Has` data type. So In ZIO 2.0, all arrow combinators are removed, and we need to use alternatives like doing monadic for-comprehension style or `flatMap`, `provide`, `zip`, and so on.
+
+## Deprecated Methods
```
domdorn, line 950: I would pull this section right to the top, as it makes understanding other things in the migration guide easier

docs/howto/migrate/migration-guide.md
khajavi: Thank you! Fixed.

docs/howto/migrate/migration-guide.md
khajavi: > hmm.. I think this isn't working as expected cause of the withFilter thingy.. e.g. it was always

Does the `withFilter` method of for-comprehension cause any problem in this case?

> not sure if it matters in the migration guide, though.

The `ZIO.services` is deprecated in ZIO 2.x so we should cover that in the migration guide.

docs/howto/migrate/migration-guide.md
khajavi: I'm not sure, it is going to replace the `@accessible` macro or not.

@adamgfraser  probably knows

docs/howto/migrate/migration-guide.md
khajavi: Thank you!

docs/howto/migrate/migration-guide.md
adamgfraser: Scala 3 doesn't support macro annotations so users can continue to use but it won't be future proof for migration to Scala 3.

adamgfraser:
@khajavi We've been seeing some issues with nondeterministic `mdoc` failures on the `series/2.x` branch. Could this be related?

khajavi:
> @khajavi We've been seeing some issues with nondeterministic `mdoc` failures on the `series/2.x` branch. Could this be related?

When I rebased my branch (`migration-guide`) to the latest works on `series/2.x`, I found out the mdoc was failing, although it was being passed successfully in my local machine! 
After investigating a lot, I found a workaround to fix that but I'm not sure how does it relate to the `series/2.x` branch.

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1542 @@
...
+
+So the error channel of the output of `left` and `right` operators is changed from `Option` to `Either`.
+
+### Descriptive Errors
+
+ZIO's type system uses implicit evidence to ensure type safety, and some level of correctness at compile time. In ZIO 2.x, the _subtype evidence_, `<:<` replaced by these two descriptive implicit evidences:
+
+1. **`IsSubtypeOfError`** — The `O1 IsSubtypeOfError O2` ensures that the output type `O1` is subtype of `O2`
+
+2. **`IsSubtypeOfOutput`** — The `E1 IsSubtypeOfOutput E2` ensures that the error type `E1` is a subtype of `E2`
```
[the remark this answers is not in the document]
khajavi: Thank you! fixed.

docs/howto/migrate/migration-guide.md
```diff
@@ -0,0 +1,1542 @@
...
+[info] 
+[info] Mermaid Live Editor Link
+[info] https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZ3JhcGhcbiAgICBDb25zb2xlLmxpdmVcbiAgICBCbG9iU3RvcmFnZS5saXZlIC0tPiBMb2dnaW5nLmxpdmVcbiAgICBMb2dnaW5nLmxpdmUgLS0+IENvbnNvbGUubGl2ZVxuICAgIFVzZXJSZXBvLmxpdmUgLS0+IExvZ2dpbmcubGl2ZVxuICAgIFVzZXJSZXBvLmxpdmUgLS0+IERhdGFiYXNlLmxpdmVcbiAgICBEb2NSZXBvLmxpdmUgLS0+IERhdGFiYXNlLmxpdmVcbiAgICBEb2NSZXBvLmxpdmUgLS0+IEJsb2JTdG9yYWdlLmxpdmVcbiAgICBEYXRhYmFzZS5saXZlXG4gICAgIiwibWVybWFpZCI6ICJ7XG4gIFwidGhlbWVcIjogXCJkZWZhdWx0XCJcbn0iLCAidXBkYXRlRWRpdG9yIjogdHJ1ZSwgImF1dG9TeW5jIjogdHJ1ZSwgInVwZGF0ZURpYWdyYW0iOiB0cnVlfQ==
+```
+
+### Module Pattern
+
+_Module Patternـ is one of the most significant changes in ZIO 2.x. Before discussing changes, let see how did we define services in ZIO 1.x.
+
+Let's try to write a Logging service using _Module Pattern 2.0_:
```
gchudnov, line 785: > Let's try to write a Logging service using _Module Pattern 2.0_:

one might think that the code that immediately follows is from 2.0, but it looks 1.0 related

khajavi: Thanks! I've fixed that.

jdegoes (approved):

PR #5384, opened 2021-08-12
Performance: Do not "over synchronize". The `SynchronizedMap::isEmpty` is already synchronized.

guizmaii:
More info: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Collections.java#L2662-L2664

### DIFF
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -813,18 +813,19 @@ private[zio] final class FiberContext[E, A](
   def id: Fiber.Id = fiberId
 
   def inheritRefs: UIO[Unit] = UIO.effectSuspendTotal {
-    // The docs for `Collections.synchronizedMap` say that we must synchronize on the map itself when
-    // using collection views of the map; `asScala` uses an iterator on the map, so we must
-    // synchronize on the map when using it. We have to run some IO actions afterwards, so we
-    // make a copy for use in the `foreach_` below.
-    val locals = Sync(fiberRefLocals)(fiberRefLocals.asScala.toMap): @silent("JavaConverters")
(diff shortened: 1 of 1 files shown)


guizmaii:
Also, I wonder if there's not a mistake here: https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/internal/FiberContext.scala#L687-L690

Nor the `.asScala` and the `.foreach` are synchronized while, the `.foreach` is using an iterator. 🤔

Shouldn't it be:

```scala
    Sync(fiberRefLocals)(
      fiberRefLocals.asScala.foreach { case (fiberRef, value) =>
        childFiberRefLocals.put(fiberRef, fiberRef.fork(value))
      }
    ): @silent("JavaConverters")
```

??

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -813,18 +813,19 @@ private[zio] final class FiberContext[E, A](
   def id: Fiber.Id = fiberId
 
   def inheritRefs: UIO[Unit] = UIO.effectSuspendTotal {
```
adamgfraser, line 815: I'm not sure this is buying us anything. `locals` is our own snapshot of the `fiberRefLocals` and we are free to work with `locals` as much as we want without any synchronization. We just need to synchronize when we create our snapshot to make sure we have a consistent view.

So in the current implementation we are always synchronizing exactly once when we do `Sync` on line 820. In contrast, in the proposed implementation we synchronize either once or twice. We always synchronize once when we do `fiberRefLocals.isEmpty` since as you say this synchronizes internally and then whenever the result is not empty we synchronize a second time to actually create the snapshot of the values.

So I tend to think this is not an improvement over the current implementation.

PR #5397, opened 2021-08-14
Implement Specialized Version Of Register Function For ZStream.effectAsync

adamgfraser:
Implements a specialized version of the register function in `ZStream.effectAsync`. This is compatible with all existing code since `Register` is just a more specific version of the previous register function.

This lets us implement some additional convenience methods on `Register` for emitting a single value or a chunk of values, failing, ending, and so on.

It does add an additional layer of indirection to the type signature of `effectAsync`, but the existing one is already relatively complicated so not sure this is a step back.

Copying @kitlangton.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -4499,14 +4499,14 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
   }
 
   /**
-   * A `Register[R, E, A, B]` represents an asynchronous callback that can be
+   * An `Emit[R, E, A, B]` represents an asynchronous callback that can be
    * called multiple times. The callback can be called with a value of type
    * `ZIO[R, Option[E], Chunk[A]]`, where succeeding with a `Chunk[A]`
    * indicates to emit those elements, failing with `Some[E]` indicates to
    * terminate with that error, and failing with `None` indicates to terminate
    * with an end of stream signal.
    */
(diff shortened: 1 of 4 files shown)


kitlangton:
Nice! Thanks, @adamgfraser! I think a benefit of this, besides clarifying the somewhat confusing signature (`(ZIO[R, Option[E], Chunk[A]] => Future[Boolean]) => Unit`), is that it allows us to document and suggest, via autocomplete, its intended usage.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4498,4 +4498,73 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
...
+     */
+    def fail(e: E): B =
+      apply(ZIO.fail(Some(e)))
+
+    /**
+     * Either emits the success value of this effect or terminates the stream
+     * with the failure value of this effect.
+     */
+    def fromEffect(zio: ZIO[R, E, A]): B =
+      apply(zio.mapBoth(e => Some(e), a => Chunk(a)))
```
kitlangton, line 4556: Perhaps also a `fromEffectChunk`?

kitlangton: I actually didn't realize this took an `Option[E]` as the error. This is probably even more useful than I thought 😄, now users don't have to know about how the Option is used internally to `end` a stream.

iravid (approved):
Love it 👌

PR #5400, opened 2021-08-16
Pass argument by name in `ZIO.debug`

tusharmath:
Otherwise bad things happen like this — 

https://scastie.scala-lang.org/Y005JdE3RqOKkH2tBpjidw


adamgfraser (approved):

adamgfraser (changes requested):
@tusharmath Looking at this again I don't think this is a binary compatible change. It passes right now because the binary compatibility version was not updated which I just submitted #5402 to do. So I think it is probably better to just pick this up in the `series/2.x` branch where #5266 makes this lazy along with other operators to avoid exactly this issue.

tusharmath:
Got it! Thanks for pointing that out @adamgfraser. Please feel free to close the PR :)

PR #5407, opened 2021-08-16
Implement ZStream#dropRight

luis3m:


iravid (dismissed):
Great addition. Thanks @luis3m!

core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
```diff
@@ -28,6 +28,12 @@ private[zio] final class SingleThreadedRingBuffer[A](capacity: Int) {
     increment()
   }
 
+  def putAndGet(value: A): Option[A] = {
```
adamgfraser, line 31: This feels a little bit uncompositional but can address in a follow up.

adamgfraser (approved):

core/shared/src/main/scala/zio/internal/SingleThreadedRingBuffer.scala
luis3m: @adamgfraser any suggestions? I can address this in this PR if you like.

adamgfraser: @luis3m Basically just separate putting and getting into their own operators instead of having one that does both.

adamgfraser:
@luis3m Can you format?

luis3m:
@adamgfraser done, also removed `putAndGet`.

adamgfraser (approved):
Nice!

PR #5410, opened 2021-08-17
Fix Documentation On ZIO 2 Branch

adamgfraser:


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -3414,7 +3416,8 @@ object ZIO extends ZIOCompanionPlatformSpecific {
 
   /**
    * Filters the Set[A] in parallel using the specified effectual predicate.
-   * See [[filter[R,E,A,Collection*]] for a sequential version of it.
+   *
+   * See [[[zio.ZIO.filter[R,E,A,Collection*]]] for a sequential version.
```
ghostdogpr, line 3420: Is it ok that there's one less closing bracket?

adamgfraser: Yes. Since you only need to supply enough of the signature to disambiguate the brackets don't need to balance. This is the same link we use in the companion object variants.

ghostdogpr: Nice, I didn't know that

ghostdogpr (approved):

PR #5427, opened 2021-08-19
Rename ZStream#repeatZIOWith to ZStream#repeatZIOWithSchedule

adamgfraser:
The current name feels a little odd in that you would normally expect a `With` operator to be take a function or possibly a value but not necessarily a schedule.


kitlangton:
Hm. Very odd that these layer tests failed here.

adamgfraser:
I think the timeout on those tests is too short and is causing them to nondeterministically fail if other tests are being executed concurrently after these tests start if we don't get back to them in time.

jdegoes (approved):
The consistency of names within the ZIO ecosystem is a highlight. Happy to merge this change in, which improves upon this consistency!

PR #5429, opened 2021-08-20
Implement ZIO#collect Variants For Map

adamgfraser:


jdegoes (approved):
Happy to see these arrive!

PR #5433, opened 2021-08-21
Standardize Names Of Operators Returning Boolean Values

adamgfraser:
I went through all the operators in ZIO that return `Boolean` values based on the principle that they should have the form `isXYZ`. I excluded methods that take an argument (e.g. `exists`) or change state in some way and return a value based on that (e.g. `offer`).

We are generally already following this rule pretty well. The main exception is in some of the operators on `Cause` and `Exit` where we use past tense terms like `died` or `interrupted` instead of `isDie` or `isInterrupted`. There were also some places in `ZScope` we were not following this, though those were mostly internally facing. This standardizes all of those.


kitlangton (approved):
Sounds great to me :) 
I think this `isMuchClearer`.

jdegoes (approved):

jdegoes:
Happy to see this!

PR #5437, opened 2021-08-21
Implement iso/utf decoders

paulpdaniels:
Relates to #4886 

Ports the following operators from ZTransducer:
- ZSink.utfDecode
- ZSink.utf16Decode/ZSink.utf16LEDecode/ZSink.utf16BEDecode
- ZSink.utf32Decode/ZSink.utf32LEDecode/ZSink.utf32BEDecode
- ZSink.iso_8859_1Decode

Adds the following operator:
- ZStream#mkString a specialization of `intersperse`


streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
```diff
@@ -770,6 +772,101 @@ object ZSinkSpec extends ZIOBaseSpec {
...
+            ZStream
+              .fromChunk(Chunk.fromArray(s.getBytes(StandardCharsets.ISO_8859_1)))
+              .transduce(ZSink.iso_8859_1Decode.map(_.getOrElse("")))
+              .mkString
+              .map(result => assertTrue(result == s))
+          })
+        ),
+        suite("utf16BEDecode")(
+          test("regular strings") {
+            checkM(Gen.const("").map(s => Chunk.fromArray(s.getBytes(StandardCharsets.UTF_16BE))).map(Chunk.single)) {
```
vigoo, line 787: This should be `Gen.anyString`

streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
```diff
@@ -770,6 +772,101 @@ object ZSinkSpec extends ZIOBaseSpec {
...
+                  }
+            }
+          }
+        ),
+        suite("utf16FEDecode")(
+          test("regular strings") {
+            checkM(Gen.anyString.map(s => Chunk.fromArray(s.getBytes(StandardCharsets.UTF_16LE))).map(Chunk.single)) {
+              byteChunks =>
+                ZStream
+                  .fromChunks(byteChunks.toList: _*)
```
vigoo, line 804: Instead of wrapping the `Chunk` in another `Chunk` and using `fromChunks` you could just use `ZIO.fromChunk` on the chunk you get from the array.

streams/shared/src/main/scala/zio/stream/experimental/ZSink.scala
```diff
@@ -1197,6 +1203,22 @@ object ZSink {
       new ZSink(ZChannel.unwrap(ZIO.access[R](f(_).channel)))
   }
 
+  def utfDecode[Err]: ZSink[Any, Err, Byte, Err, Byte, Option[String]] = {
+    def prepend(bytes: Chunk[Byte]): ZChannel[Any, Err, Chunk[Byte], Any, Err, Chunk[Byte], Any] =
+      ZChannel.write(bytes) *> ZChannel.identity[Err, Chunk[Byte], Any]
+
+    ZSink.take[Err, Byte](4).flatMap { bytes =>
+      bytes.toList match {
```
vigoo, line 1211: We could skip the conversion to list and back to chunk and work directly on chunks (I know the 1.0 implementation  is like this too, so consider it as an optional optimization - we can also do it later)

vigoo:
Ready to merge if the test is fixed

vigoo (changes requested):
Ready to merge if the test is fixed

paulpdaniels:
Woah thanks @vigoo! I did not register the notifications.

PR #5447, opened 2021-08-23
Add ZIOApp

kitlangton:
A simple change, but very nice for examples/teaching/onboarding—and, most usually, a fine default for non-CLI applications.
Not sure about the name/location though (if this seems like a good addition).


jdegoes:
@kitlangton Implemented in #5482. Thanks for the inspiration! 🙏

PR #5451, opened 2021-08-24
Add >%> for providing partial envs to layers

kitlangton:
Another attempt at #4772


examples/shared/src/main/scala/zio/examples/LayerDefinitionExample.scala
```diff
@@ -17,12 +17,12 @@ object LayerDefinitionExample extends App {
   }
 
   override def run(args: List[String]): URIO[ZEnv, ExitCode] = {
-    val liveLayer: ULayer[Has[Foo]] =
-      (Console.live ++ ZLayer.succeed("Hello") ++ ZLayer.succeed(3)) >>> Foo.live
+    val liveLayer: URLayer[Has[String], Has[Foo]] =
```
adamgfraser, line 20: Maybe better to leave the example as is and add a test to verify that this operator is working correctly?

kitlangton: That's a much better idea :)

kitlangton: So yeah. Very strangely, it only compiles for Scala 3 if given a type signature. Perhaps I can turn this into a bug report. 😞

core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
```diff
@@ -390,6 +390,18 @@ object ZLayerSpec extends ZIOBaseSpec {
           _     <- layer.build.useNow
           value <- ref.get
         } yield assert(value)(equalTo("bar"))
+      },
+      testM(">%>") {
+        val layer: ZLayer[Has[Int] with Has[String], Nothing, Has[Int] with Has[String]] =
+          ZLayer.service[Int] ++ ZLayer.service[String]
+        val justInt: ULayer[Has[Int]]                                         = ZLayer.succeed(10)
+        val provided: ZLayer[Has[String], Nothing, Has[Int] with Has[String]] = justInt >%> layer
```
adamgfraser, line 398: Is this type annotation required?

jdegoes:
@kitlangton @adamgfraser Since we don't know what to do with this, I will close for now, but we can discuss more at the hackathon and re-open if appropriate. Thank you!

PR #5469, opened 2021-08-27
Align Return Values Of Schedule Constructors

### LINKED ISSUE #5462: Unexpected `Schedule.fromDurations` output
It's tough to say whether this is a bug or intentional, but `Schedule.fromDurations(1.seconds, 2.seconds, 3.seconds).run(OffsetDateTime.now, 1 to 4)` currently returns `Chunk(PT0S,PT0S,PT0S,PT3S)` while I would have expected it to return `Chunk(PT0S,PT1S,PT2S,PT3S)`. The docs states "Returns the length of the current duration between recurrences." so I assume the latter? I mean, returning 0 for every element besides the last isn't conveying too much information in the first place, so that's another reason why I think the latter makes more sense.

The 1st element being 0 makes sense if you consider the singular version does that also:

```scala
Schedule.fromDuration(1.second).run(OffsetDateTime.now, 1 to 2) // Chunk(PT0S,PT1S)
```

That's because the 1st step is `Decision.Continue(Duration.Zero)` and the 2nd step is `Decision.Done(duration)`

### LINKED ISSUE #5463: Schedule: Method for outputting delays
We often simulate complex Schedules to calculate what the delay would be at repetition `n`. The thing is `Schedule` has a delay _and_ and output. And the output is often something unrelated to the delay (for example `Schedule.fixed` returns the number of repetitions of the Schedule so far).

I think it would be helpful to have a method to output the delay. My coworker wrote the following as a temporary workaround for the time being:

```scala
/** Returns a new schedule that emits the current tick's delay. Returns Duration.Zero when the schedule finishes */
def outputDelay: Schedule[R, In, Duration] = {
  def loop(self: StepFunction[R, In, Out]): StepFunction[R, In, Duration] =
    (now: OffsetDateTime, in: In) =>
      self(now, in).map {
        case Done(_) => Done(Duration.Zero)
        case Continue(_, interval, next) =>
          val delay = Duration(interval.toInstant.toEpochMilli - now.toInstant.toEpochMilli, TimeUnit.MILLISECONDS)
          Continue(delay, interval, loop(next))
      }

  Schedule(loop(schedule.step))
}
```

(note that `step` is private so we had create a `zio` package to implement this)

I'm not sure if the above method is ideal. Maybe a method like `schedule.zipWithDelay` (similar to `zipWithIndex`) instead would be better, but I don't have too strong of a preference. As long as there is some way to achieve this within ZIO itself somehow.

adamgfraser:
Resolves #5462. Resolves #5463.

This PR makes a couple of changes to improve the usefulness of the return values of schedules.

First, it implements a new `delays` operator on `Schedule` analogous to the existing `repetitions` operator that returns a version of the schedule that outputs the delay between each step.

Second, it uses this to align the return values of the schedule constructors. Specifically, we implement the principle that constructors that return the number of repetitions must always return a value that is consistent with `Schedule#repetitions` and constructors that return the delay must return a value that is consistent with `Schedule#delays`.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -567,8 +567,42 @@ object ScheduleSpec extends ZIOBaseSpec {
       val scheduleIntervals = runManually(Schedule.fixed(5.seconds), inputs).map(_._1.map(_._1))
 
       assertM(scheduleIntervals)(equalTo(List(offsetDateTime(5000), offsetDateTime(10000))))
-    }
+    },
+    suite("return values")(
+      suite("delays")(
+        test("duration")(checkDelays(Schedule.duration(1.second))),
+        test("exponential")(checkDelays(Schedule.exponential(1.second))),
+        test("fibonacci")(checkDelays(Schedule.fibonacci(1.second))),
+        test("fromDuration")(checkDelays(Schedule.fromDuration(1.second))),
+        test("fromDurations")(checkDelays(Schedule.fromDurations(1.second, 2.seconds, 3.seconds, 4.seconds))),
+        test("linear")(checkDelays(Schedule.linear(1.second)))
+      ),
+      suite("repetitions")(
+        test("count")(checkRepetitions(Schedule.count)),
+        test("fixed")(checkRepetitions(Schedule.fixed(1.second))),
+        test("forever")(checkRepetitions(Schedule.forever)),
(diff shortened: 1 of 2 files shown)


jdegoes (approved):

jdegoes:
> Specifically, we implement the principle that constructors that return the number of repetitions must always return a value that is consistent with Schedule#repetitions and constructors that return the delay must return a value that is consistent with Schedule#delays.

❤️

PR #5477, opened 2021-08-28
Implement Compositional Platform

adamgfraser:


core-tests/shared/src/test/scala/zio/ZIOSpec.scala
```diff
@@ -3948,6 +3948,14 @@ object ZIOSpec extends ZIOBaseSpec {
         } yield {
           assert(value)(equalTo("Controlling side-effect of function passed to promise"))
         }
+      },
+      test("onPlatform") {
+        for {
+          platform <- ZIO.suspendSucceedWith((platform, _) => ZIO.succeed(platform))
```
jdegoes, line 3954: Looking at these tests, probably we should introduce `ZIO.platform` to do the whole `suspendSucceedWith` thing of extracting and returning the platform, in a similar vein to `ZIO.runtime`.

core/shared/src/main/scala/zio/IO.scala
```diff
@@ -1030,6 +1030,12 @@ object IO {
   def not[E](effect: IO[E, Boolean]): IO[E, Boolean] =
     ZIO.not(effect)
 
+  /**
+   *  @see See [[zio.ZIO.onPlatform]]
+   */
+  def onPlatform[E, A](platform: => Platform)(io: => IO[E, A]): IO[E, A] =
```
jdegoes, line 1036: This raises the question of whether to rename `lock` to `onExecutor`.

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -30,15 +30,14 @@ import scala.annotation.{switch, tailrec}
...
   startEnv: AnyRef,
   startExec: Executor,
   startLocked: Boolean,
   startIStatus: InterruptStatus,
   parentTrace: Option[ZTrace],
   initialTracingStatus: Boolean,
   var fiberRefLocals: FiberRefLocals,
-  supervisor0: Supervisor[Any],
   openScope: ZScope.Open[Exit[E, A]],
   reportFailure: Cause[Any] => Unit
```
jdegoes, line 42: In a similar vein, I might get rid of `startExec` (use the one in platform, modifying as appropriate), and `reportFailure` (use the one in platform, modifying as appropriate).

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -710,6 +708,17 @@ private[zio] final class FiberContext[E, A](
...
                     curZio = nextInstr(())
+
+                  case ZIO.Tags.OnPlatform =>
+                    val zio = curZio.asInstanceOf[ZIO.OnPlatform[Any, Any, Any]]
+
+                    val oldPlatform = platform
+
+                    platform = zio.platform()
+
+                    ensure(ZIO.succeed { platform = oldPlatform })
```
jdegoes, line 719: If you want to be slower about this, but also simplify `FiberContext`, you could have `SetPlatform`, and do the ensuring bit with acquireRelease. It will be slower, for sure, albeit changing the platform is not something that we would expect to occur with any frequency.

jdegoes (approved):
Looks fantastic! A few small comments if we want to push forward with this direction a little more.

core/shared/src/main/scala/zio/IO.scala
adamgfraser: Yes. We already renamed `on` to `onExecutionContext` so it would definitely be consistent.

core-tests/shared/src/test/scala/zio/ZIOSpec.scala
adamgfraser: Done now. So much nicer!

core/shared/src/main/scala/zio/internal/FiberContext.scala
adamgfraser: This would also be more compositional because it would open the door to implementing an operator to just set the platform which would then allow higher level operators to be defined for `ZManaged` and `ZStream` that allowed changing the platform for a scope.

jdegoes: Exactly

adamgfraser: Done.

core/shared/src/main/scala/zio/internal/FiberContext.scala
adamgfraser: I'm a little worried about semantics here. `startExec` means "the executor we should start running on" which in `fork` I don't think is necessarily the same as "the default executor to run on if none is specified". In particular I think that means if we did `ZIO.platform.fork.flatMap(_.join)` we might not get the same value as `ZIO.platform` which seems strange. There is a little bit of the same issue in `reportFailure`. For example `race` uses `noop` for `reportFailure` so the platform in the branches of a race would be different than the platform in the parent.

core/shared/src/main/scala/zio/IO.scala
adamgfraser: Will do this in a separate PR.

adamgfraser:

core/shared/src/main/scala/zio/internal/FiberContext.scala
jdegoes: You are right about this 👍

jdegoes (approved):

jdegoes:
Beautiful work! ❤️

PR #5480, opened 2021-08-28
Back Port Zippable and Unzippable To ZIO 1.0

adamgfraser:
Some ZIO ecosystem libraries would like to start using `Zippable` to construct tuples without nesting. We can't actually change the signatures of operators in ZIO 1.0 to use `Zippable` but we can make it available so that other ecosystem libraries can depend on it.


kitlangton (approved):
LGTM

PR #5481, opened 2021-08-29
Implement HasMany

adamgfraser:
This approach seems relatively promising.

### DIFF
--- a/core/shared/src/main/scala/zio/Has.scala
+++ b/core/shared/src/main/scala/zio/Has.scala
@@ -60,7 +60,7 @@ object Has {
     def add[R0 <: R, M: Tag](r: R0, m: M): R0 with Has[M]
     def union[R0 <: R, R1 <: Has[_]: Tag](r: R0, r1: R1): R0 with R1
     def update[R0 <: R, M: Tag](r: R0, f: M => M)(implicit ev: R0 <:< Has[M]): R0
-    def updateAt[R0 <: R, K: Tag, A: Tag](r: R0, k: K, f: A => A)(implicit ev: R0 <:< Has[Map[K, A]]): R0
+    def updateAt[R0 <: R, K: Tag, A: Tag](r: R0, k: K, f: A => A)(implicit ev: R0 <:< HasMany[K, A]): R0
   }
   object IsHas {
     implicit def ImplicitIs[R <: Has[_]]: IsHas[R] =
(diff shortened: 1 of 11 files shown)


core/shared/src/main/scala/zio/package.scala
```diff
@@ -80,4 +80,6 @@ package object zio
   val Hub: ZHub.type = ZHub
 
   type Semaphore = stm.TSemaphore
+
+  type HasMany[K, A] = Has[_ <: Map[K, A]]
```
jdegoes, line 84: We have to find a way to test that use of a type wildcard here doesn't cause unsoundness issues with type tags.

jdegoes: In fact with a type alias approach, maybe just `Has[Map[K, A]]` is better, since it is in fact a map.

adamgfraser: Yes I think that would help.

core/shared/src/main/scala/zio/Has.scala
```diff
@@ -217,8 +220,59 @@ object Has {
...
+        .asInstanceOf[Map[K, A]]
+        .updated(k, a)
+
+      new Has(self.map + (tag -> map)).asInstanceOf[Self]
+    }
+
+    /**
+     * Retrieves a service at the specified key from the environment.
+     */
+    def getAt[K, A](k: K)(implicit ev: Self <:< HasMany[K, A], tagged: Tag[Map[K, A]]): Option[A] = {
```
jdegoes, line 247: Any thoughts on `Tagged[Map[K, A]]` versus `Tagged[K], Tagged[A]`.

adamgfraser: Yes. One should be derivable from the other but probably better to require the most primitive components.

jdegoes (dismissed):
💯 🚀

jdegoes:
One `mdoc` error.

core/shared/src/main/scala/zio/Has.scala
adamgfraser: Scratch that. We actually need to require the `Tagged[Map[K, A]]` to work around an issue with the tag machinery in Scala 3. Opened zio/izumi-reflect#208 for it.

adamgfraser:

jdegoes (approved):

PR #5488, opened 2021-08-30
Rename ZIO#tapCause to ZIO#tapErrorCause

### LINKED ISSUE #5483: ZIO 2.0: Ensure consistent use of *Cause variants
Currently we have some variation in the *Cause variants, including, on the one hand, a convention of adding the `Cause` postfix immediately following the `Error` postfix, e.g.:

 - `ZIO.failCause`
 - `ZIO.mapErrorCause`

And on the other hand, dropping `Error` entirely and just using the `Cause` suffix directly:

 - `ZIO.tapCause` (`ZIO.tapError`)

We should standardize on which form we want to adopt and drop the other one, leaving deprecations until the 2.0 release.

adamgfraser:
Resolves #5483. In my review it seemed like the most common pattern was for the `Cause` variant to be identical to the normal one except have `Cause` at the end (with slight variations if there is another postfix (e.g. `foldCauseZIO`). This seems to be logical and many operators don't read well otherwise (e.g. `failCause` doesn't really make sense as `cause`). Renames `tapCause` to `tapErrorCause` to conform to this.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOSpec.scala
@@ -3353,11 +3353,11 @@ object ZIOSpec extends ZIOBaseSpec {
         }
       }
     ),
-    suite("tapCause")(
+    suite("tapErrorCause")(
       test("effectually peeks at the cause of the failure of this effect") {
         for {
           ref    <- Ref.make(false)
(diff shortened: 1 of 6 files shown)


kitlangton:
👍 for the naming decision. Autocomplete can help reveal the alternatives this way.

kitlangton (approved):

PR #5491, opened 2021-08-30
Log Warnings In ZIO Test

adamgfraser:
Now that we have logging we can log warnings instead of just rendering them to the console.


jdegoes (approved):

jdegoes:
Nice to see this! Maybe we can make developer-mode experience for logging use colors and stuff to make this even nicer. /cc @kitlangton 😄

Issue #5466, opened 2021-08-27
zio-test: Introduce Equality typeclass for Smart Assertions


vladimirkl:
It looks to be necessary to introduce `Equality` typeclass similar to `Diff` for `zio.test.internal.SmartAssertions#equalTo`.  Currently equality check is hardcoded:
```scala
val result = (a, that) match {
  case (a: Array[_], that: Array[_]) => a.sameElements[Any](that)
  case _                             => a == that
}
```

This doesn't work well for floating point numbers where we need customizable fuzzy equality. Currently I have to use something like this in tests:

```scala
implicit class DoubleSyntax(self: Double) {
  def =~(other: Double): Boolean = math.abs(self - other) <= 1e-10
}
// ...
assertTrue(a =~ b )
```

This is rendered bad in smart assertions output and cannot be used with `Diff`. Same for case classes - I can introduce typeclass for fuzzy equality and derive it, but it is not useful much because of missing `Diff` support.

kitlangton:
It's certainly a pain at the moment. I have an idea that could support fully custom smart assertions; and I think that could solve this case. It would allow for user-defined error messages. I've got to toy around a bit with it. I'll do that this weekend. 😄 Hopefully, I can update you soon.

kitlangton:
I'd love your thoughts on #5492 @vladimirkl 😄

PR #5492, opened 2021-08-31
Custom Smart Assertions

kitlangton:
Addresses #5466


test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
```diff
@@ -19,389 +20,451 @@ object SmartAssertionSpec extends ZIOBaseSpec {
...
+
+  implicit final class ColorOps(private val self: Color) extends AnyVal {
+    def asRed: SmartAssertion[Red] =
+      SmartAssertion.fromEither {
+        self match {
+          case Red(int) => Right(Red(int))
+          case _        => Left(s"$self IS NOT NICE!")
+        }
+      }
+  }
```
kitlangton, line 458: How to write custom assertions.

test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
```diff
@@ -19,389 +20,451 @@ object SmartAssertionSpec extends ZIOBaseSpec {
...
   private final case class MyClass(name: String)
   private final case class OtherClass(name: String)
+
+  implicit final class DoubleOps(private val self: Double) extends AnyVal {
+    def ~=(that: Double): SmartAssertion[Boolean] =
+      SmartAssertion.cond(self == that, s"THIS AIN'T RIGHT $self !~= $that")
+
+    def asEven: SmartAssertion[Double] =
+      SmartAssertion.fromEither(Either.cond(self % 2 == 0, self, s"$self must be even."))
+  }
```
kitlangton, line 469: Another example.

vladimirkl: Looks great!  If we use this with case classes, should we provide `DiffResult.render` output as parameter to  `SmartAssertion.cond`?  I'm wondering how it will look in final output.

adamgfraser (dismissed):
Looks great!

PR #5498, opened 2021-08-31
Metrics

adamgfraser:


core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -0,0 +1,239 @@
+package zio
+
+import zio.internal.metrics._
```
jdegoes, line 3: As this, I believe, is a public data type, we should probably move it to `zio.metrics`. Ditto for other public data types associated with metrics.

core/shared/src/main/scala/zio/internal/Platform.scala
```diff
@@ -31,7 +31,8 @@ final case class Platform(
   reportFailure: Cause[Any] => Unit,
   supervisor: Supervisor[Any],
   enableCurrentFiber: Boolean,
-  logger: ZLogger[Unit]
+  logger: ZLogger[Unit],
+  metrics: ZMetrics
```
jdegoes, line 35: Good way to hide the "global singleton". 👍

core/shared/src/main/scala/zio/internal/ZMetrics.scala
```diff
@@ -0,0 +1,32 @@
+package zio.internal
+
+import zio.internal.metrics._
+
+trait ZMetrics {
```
jdegoes, line 5: All this definitely belongs in the internal package as hopefully users won't see or touch it. 👍

core/shared/src/main/scala/zio/internal/metrics/Gauge.scala
```diff
@@ -0,0 +1,52 @@
+package zio.internal.metrics
```
jdegoes, line 1: Probably this and other such data types need pulling up into `zio.metrics` because they have safe public APIs that we expect users to use directly.

core/shared/src/main/scala/zio/internal/metrics/MetricState.scala
```diff
@@ -0,0 +1,70 @@
...
+    MetricState(key.name, help, Chunk(key.tags: _*), MetricType.Counter(value))
+
+  // --------- Methods creating and using Prometheus Gauges
+
+  def gauge(
+    key: MetricKey.Gauge,
+    help: String,
+    startAt: Double
+  ): MetricState =
+    MetricState(key.name, help, Chunk(key.tags: _*), MetricType.Gauge(startAt))
```
jdegoes, line 31: Performance note but probably `key.tags` should already be a chunk so we don't have to wrap it every time.

core/shared/src/main/scala/zio/internal/metrics/package.scala
```diff
@@ -0,0 +1,10 @@
+package zio.internal
+
+package object metrics {
+
+  private[zio] lazy val metricState: ConcurrentState =
```
jdegoes, line 5: Ah this can go in platform too, perhaps.

core/shared/src/main/scala/zio/internal/metrics/package.scala
```diff
@@ -0,0 +1,10 @@
...
+    new ConcurrentState
+
+  val dblOrdering: Ordering[Double] =
```
jdegoes, line 8: ???

core/shared/src/main/scala/zio/ZIOMetric.scala
adamgfraser: My question here is how many data types we actually need to expose to users. My hypothesis is that the only data type we need to expose to users of metrics is `ZIOMetric` so we could certainly put it in a separate metrics package but if we only have a single data type could also make sense to expose at the top level. Then there could be a couple of other data types necessary for implementers of metrics backend interfaces that could be provided in the `internal` package.

core/shared/src/main/scala/zio/internal/metrics/Gauge.scala
adamgfraser: I think we actually may be able to delete this. At this point users interact with metrics through the `ZIOMetric` interface so I'm not sure we need this but perhaps it could be useful to implement variants of `ZIOMetric` for other data types (e.g. `ZStreamMetric`).

core/shared/src/main/scala/zio/internal/metrics/MetricListener.scala
```diff
@@ -0,0 +1,33 @@
+package zio.internal.metrics
+
+/**
+ * A `MetricListener` is capable of taking some action in response to a metric
+ * being recorded, such as sending that metric to a third party service.
+ */
+trait MetricListener {
```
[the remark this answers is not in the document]
adamgfraser: 💯

core/shared/src/main/scala/zio/internal/metrics/MetricState.scala
adamgfraser: 👍

core/shared/src/main/scala/zio/internal/metrics/package.scala
adamgfraser: Yes. There is a little thinking to be done here about how the ability to potentially change the platform interacts with this.

core/shared/src/main/scala/zio/internal/metrics/package.scala
adamgfraser: Scala 2.13 and later have incompatible implementations of this so we need to provide our own.

core/shared/src/main/scala/zio/internal/metrics/Gauge.scala
jdegoes: That would be nice to delete it!

core/shared/src/main/scala/zio/internal/metrics/package.scala
jdegoes: Right now it is `???` implementation.

adamgfraser: Yes, was just trying to get it to compile. Will update shortly.

adamgfraser: Implemented now.

jdegoes:
@adamgfraser Let me know when you want another look.

core/shared/src/main/scala/zio/internal/metrics/MetricListener.scala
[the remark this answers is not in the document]
adamgfraser: I don't know if we can make this one in particular package private as backend implementers are going to need access to this and we have already had cases where users wanted to implement their own backends for integration into their systems.

adamgfraser:
@jdegoes So I think we have a couple of levels of the API.

At the highest level there is `ZIOMetric`, which can be either applied as an aspect or called directly. If you are just using existing metrics defined for you this is all you should ever need to use. You could also imagine variants of this like `ZStreamMetric` being exposed by other libraries. This is defined in the `zio.metrics` package right now though perhaps it should be pulled up to the top level.

At the next level there are the specific types `Counter`, `Gauge`, and so on. These hide the underlying implementation of metric creation and updating and describe a type of metric independent of how it is being used (e.g. a counter independent of what it is counting). You need these if you want to define your own metrics or if you wanted to define your own type like `ZStreamMetric`. These are located in the `zio.metrics` package`.

At the next level there are the data types you would need as an implementer of a backend to integrate with this functionality. This includes the highest level `ZMetrics` as well as the data types it uses `MetricKey`, `MetricState`, `MetricType`, and `MetricListener`. Right now `ZMetrics` is in the `zio.internal` package and the others are in the `zio.internal.metrics` package.

At the lowest level are the concurrent data structures we use to track metrics. These are all private or package private and located in the `zio.internal.metrics` package

adamgfraser:

core/shared/src/main/scala/zio/metrics/package.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+package object metrics {
```
jdegoes, line 19: For another time, but I think this should be replaced by a newtype.

jdegoes (approved):

jdegoes:
Loving how clean and well-integrated this is, while not opening up a giant surface area to existing ZIO users. ❤️

core/shared/src/main/scala/zio/metrics/package.scala
adamgfraser: Implemented in #5538.

PR #5506, opened 2021-09-01
Platform Improvements

### LINKED ISSUE #5501: ZIO 2.0: Platform improvements
`zio.internal.Platform` could stand a number of improvements in ZIO 2.0:

1. Rename `Platform` => `RuntimeConfig`.
2. Rename `PlatformAspect` => `RuntimeConfigAspect`.
3. Move `RuntimeConfig` **and** `RuntimeConfigAspect` to `zio`, rather than `zio.internal`.
4. Move `Platform.OS` / `Platform.os` into `System` companion object, because it is logically more related to `System` than `RuntimeConfig`.
5. Introduce several `RuntimeConfigAspect` for modifying `RuntimeConfig`, inside the companion object of `RuntimeConfigAspect` (currently there is only `identity`). These could include:
    - `RuntimeConfigAspect.addSupervisor`
    - `RuntimeConfigAspect.addReportFailure`
    - `RuntimeConfigAspect.addReportFatal`
    - `RuntimeConfigAspect.addLogger`
    - `RuntimeConfigAspect.setTracing`
    - `RuntimeConfigAspect.setExecutor`
    - `RuntimeConfigAspect.setBlockingExecutor`
    The `add` methods would be additive: they would not replace, but rather, compose the existing fields inside the `RuntimeConfig`. The `set` methods would replace the existing fields.

Inside the `zio.internal` package object, deprecated `type` and `val` aliases (`type Platform = _root_.zio.RuntimeConfig; val Platform = _root_.zio.RuntimeConfig`) can be added to ease migration from ZIO 1.x to ZIO 2.x.

An open question is whether to push the `unsafe` prefix into `RuntimeConfig` since this will be top-level visible package that users would be expected to see and interact with as part of configuring ZIO applications.

adamgfraser:
Resolves #5501.

I wonder if we should retain a version of `Platform` in the `internal` package for operators like `newConcurrentSet`. Also, I wonder if we should move `Executor` out of the `internal` package.


jdegoes:
> I wonder if we should retain a version of Platform in the internal package for operators like newConcurrentSet. 

I think this is correct. Indeed, it fits with the meaning of the word `Platform` as describing whether we are running on native, JS, or JVM. Existing `Platform` started out with this connotation (e.g. thread pool, which is a platform-specific configuration) but has now become almost entirely generic runtime config.

The parallel between Runtime needing a RuntimeConfig reads very well in your PR and convinces me this is the right change to make. 👍 

The one thing I wonder about is whether `onRuntimeConfig` should instead be `withRuntimeConfig`, but I think you can make an argument for `on` based on consistency. 

> Also, I wonder if we should move Executor out of the internal package.

I agree. We could put it into `zio.exec` (just to avoid name clashes) or just leave it top-level.

adamgfraser:
Yes I was thinking the same thing regarding `onRuntimeConfig` reading awkwardly but left it for consistency. I think I will change it.

jdegoes:
@adamgfraser Maybe we should mark the tests as `flaky` so they at least run? Otherwise we run the risk of them going from "somewhat flaky" to "always broken"?

jdegoes (approved):

adamgfraser:
@jdegoes These tests were timing out so I don't think `nonFlaky` would have helped but completely agree we need to not lose track of these and I opened an issue for them.

PR #5509, opened 2021-09-01
Use Logger in Platform#reportFailure

### LINKED ISSUE #5503: ZIO 2.0: Leverage logging for reportFailure
Currently, default `reportFailure` in Platform (/RuntimeConfig) logs to console. This is not desirable. Instead, anytime we have a failure, we should use the logger inside Platform (/RuntimeConfig), using the same parameters as a user would trigger, if the user called `ZIO.logCause`.

This will allow us to delete `reportFailure` and give the user a uniform way of handling errors.

Ideally, at the same time, we should rename `ZIO.logCause` to `ZIO.logErrorCause` to be consistent with the new naming convention.

adamgfraser:
Resolves #5503.

I'm not sure we can delete `reportFailure` from `Platform`. For example in ZIO Test we do `.withReportFailure(_ => ())` and otherwise test output is extremely noisy. We could just modify the logger to not log errors but that seems too heavyweight, preventing users from logging errors as well. It seems like we do want some way of controlling what the runtime does to log errors separate from what users do to log errors.

This PR just uses the the logger as the default way of reporting failures. That way we can still customize how the runtime handles failures separately from how logging works but we get the benefit of using the logger to report failures.


jdegoes:
The thing I don't like about this is that it makes `reportFailure` pure, whereas previously, it was side-effecting. `RuntimeConfig` is such a low-level interface, it would be nice if all hooks inside it were side-effecting.

On the other hand, you're right about the noise level and the need for the ability to squelch it. 

Maybe we could somehow unify them: log at a low-level that is off by default. Indeed, probably the "developer-mode" experience for logging should only log info and higher. We could place fiber failures at FINE or even invent a new level specifically for them, which would not be logged.

adamgfraser:
I don't think it necessarily has to be. There is significant logic in the interpretation of `ZIO.Logged` that I wanted to reuse but maybe I can pull that out into a separate impure operator in `FiberContext` and then call that directly.

adamgfraser:

core/shared/src/main/scala/zio/internal/ZLogger.scala
```diff
@@ -33,6 +33,26 @@ trait ZLogger[+A] { self =>
         )
     }
 
+  /**
+   * Returns a version of this logger that only logs messages when the log
+   * level satisfies the specified predicate.
+   */
+  final def filterLogLevel(f: LogLevel => Boolean): ZLogger[Unit] =
```
jdegoes, line 40: If performance is not critical, this one could return `ZLogger[Option[A]]`.

jdegoes (dismissed):
Tiny comment on return type of `filterLogLevel`, but feel free to merge now. Looks great!

adamgfraser:
Updated now.

jdegoes (approved):

jdegoes:
Beautiful work! Now only a logging hook is required to receive complete diagnostic information about abnormal fiber end-of-life.

jdegoes (approved):

PR #5510, opened 2021-09-01
Deprecate ZIO#forkOn

adamgfraser:
While we have been adding new operators over time this seems like one we can maybe cut out. It can already be implemented extremely compositionally as `zio.onExecutionContext(ec).fork` which still provides a fluent API. There are a variety of ways one might want to customize a forked effect and I don't think this one is sufficiently common to warrant its own variant. It also seems odd that this takes an `ExecutionContext` and not an `Executor` when `Executor` is supposed to be our first class representation of a thread pool.


kitlangton:
Sounds great. I don't think this is common enough to necessitate a special pre-composed operator :)

jdegoes:
I agree with this. There's some code that can be removed in FiberContext too.

PR #5514, opened 2021-09-02
Run Console Operations On Blocking Executor

adamgfraser:


jdegoes (approved):
Easy win until we can figure out a way to do this in non-blocking style!

PR #5523, opened 2021-09-04
Refactor Schedule To Support Bounded Intervals

### LINKED ISSUE #5497: Intersected Schedules created with `hourOfDay`/`minuteOfHour`/`secondOfMinute` don't work as expected.
When intersecting `Schedules` constructed with `hourOfDay`/`minuteOfHour`/`secondOfMinute` etc. one would expect to get cron-like behavior. However, this seems to not be the case at the moment.
Here's a spec that demonstrates the problem:

```scala
import zio._
import zio.clock.Clock
import zio.duration._
import zio.test.Assertion._
import zio.test._
import zio.test.environment._

import java.time.LocalTime

object CronLikeScheduleSpec extends DefaultRunnableSpec {
  private val AtFourTwenty      = Schedule.hourOfDay(4) && Schedule.minuteOfHour(20)
  private val ExpectedLocalTime = LocalTime.of(4, 20)

  private val PassingTest =
    testM("schedule with TestClock adjusted just prior to the first iteration works as expected") {
      for {
        _       <- TestClock.setTime(4.hours + 19.minutes)
        ref     <- Ref.make(List.empty[LocalTime])
        _       <- recordLocalTime(ref).schedule(AtFourTwenty && Schedule.once).zipPar(TestClock.adjust(1.day))
        results <- ref.get
      } yield assert(results.headOption)(isSome(equalTo(ExpectedLocalTime)))
    }

  private val FailingTests =
    suite("failing tests suite")(
      testM("Schedule with unadjusted TestClock recurs as expected") {
        for {
          ref     <- Ref.make(List.empty[LocalTime])
          _       <- recordLocalTime(ref).schedule(AtFourTwenty && Schedule.once).zipPar(TestClock.adjust(1.day))
          results <- ref.get
        } yield assert(results.headOption)(isSome(equalTo(ExpectedLocalTime)))
      },
      testM("Schedule with TestClock adjusted just past the first iteration works as expected") {
        for {
          _       <- TestClock.setTime(4.hours + 21.minutes)
          ref     <- Ref.make(List.empty[LocalTime])
          _       <- recordLocalTime(ref).schedule(AtFourTwenty && Schedule.once).zipPar(TestClock.adjust(1.day))
          results <- ref.get
        } yield assert(results.headOption)(isSome(equalTo(ExpectedLocalTime)))
      },
      testM("Multiple recurrences of the Schedule occur at the same expected time") {
        for {
          ref     <- Ref.make(List.empty[LocalTime])
          _       <- recordLocalTime(ref).schedule(AtFourTwenty && Schedule.recurs(10)).zipPar(TestClock.adjust(10.days))
          results <- ref.get
        } yield assertTrue(results.size == 10) && assert(results)(forall(equalTo(ExpectedLocalTime)))
      },
      testM("schedule with TestClock adjusted just prior to the first iteration works as expected multiple times") {
        for {
          _       <- TestClock.setTime(4.hours + 19.minutes)
          ref     <- Ref.make(List.empty[LocalTime])
          _       <- recordLocalTime(ref).schedule(AtFourTwenty && Schedule.recurs(10)).zipPar(TestClock.adjust(10.day))
          results <- ref.get
        } yield assertTrue(results.size == 10) && assert(results)(forall(equalTo(ExpectedLocalTime)))
      }
    )

  override def spec: ZSpec[TestEnvironment, Any] =
    suite("testing intersected cron-like schedules")(PassingTest, FailingTests)

  private def recordLocalTime(ref: Ref[List[LocalTime]]): URIO[Clock, Unit] =
    for {
      time <- clock.currentDateTime.map(_.toLocalTime).orDie
      _    <- ref.update(time :: _)
    } yield ()
}
```

adamgfraser:
Resolves #5497.

Currently `Schedule` is based on a concept of a step function that is either done or wants to recur at a specific time. This works for most schedules but not ones like `minuteOfHour` or `dayOfWeek` that want to recur only in a bounded interval (e.g. between 4 PM and 4:59 PM) rather than just after a certain time. To support this we need to change the definition of the step function to return an interval in which the schedule is willing to recur and then change schedule composition operators accordingly.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
@@ -308,7 +308,7 @@ object ScheduleSpec extends ZIOBaseSpec {
     suite("cron-like scheduling. Repeats at point of time (minute of hour, day of week, ...)")(
       test("recur at 01 second of each minute") {
         def toOffsetDateTime[T](in: (List[(OffsetDateTime, T)], Option[T])): List[OffsetDateTime] =
-          in._1.map(t => t._1.withNano(0))
+          in._1.map(t => t._1)
 
         val originOffset = OffsetDateTime.now().withMinute(0).withSecond(0).withNano(0)
 
@@ -322,7 +322,7 @@ object ScheduleSpec extends ZIOBaseSpec {
         assertM(runManually(Schedule.secondOfMinute(1), input).map(toOffsetDateTime)) {
           val expected          = originOffset.withSecond(1)
           val afterTimeExpected = expected.withMinute(expected.getMinute + 1)
-          equalTo(List(expected, afterTimeExpected, expected, afterTimeExpected))
+          equalTo(List(inTimeSecondNanosec, inTimeSecond, inTimeSecond, afterTimeExpected))
         }
       },
       test("throw IllegalArgumentException on invalid `second` argument of `secondOfMinute`") {
@@ -333,7 +333,7 @@ object ScheduleSpec extends ZIOBaseSpec {
       } @@ failing(diesWith(isSubtype[IllegalArgumentException](anything))),
       test("recur at 01 minute of each hour") {
         def toOffsetDateTime[T](in: (List[(OffsetDateTime, T)], Option[T])): List[OffsetDateTime] =
-          in._1.map(t => t._1.withNano(0))
+          in._1.map(t => t._1)
 
         val originOffset = OffsetDateTime.now().withHour(0).withSecond(0).withNano(0)
 
@@ -347,7 +347,7 @@ object ScheduleSpec extends ZIOBaseSpec {
         assertM(runManually(Schedule.minuteOfHour(1), input).map(toOffsetDateTime)) {
           val expected          = originOffset.withMinute(1)
           val afterTimeExpected = expected.withHour(expected.getHour + 1)
-          equalTo(List(expected, afterTimeExpected, expected, afterTimeExpected))
+          equalTo(List(inTimeMinuteNanosec, inTimeMinute, inTimeMinute, afterTimeExpected))
         }
       },
       test("throw IllegalArgumentException on invalid `minute` argument of `minuteOfHour`") {
@@ -374,7 +374,7 @@ object ScheduleSpec extends ZIOBaseSpec {
         assertM(runManually(Schedule.hourOfDay(1), input).map(toOffsetDateTime)) {
           val expected          = originOffset.withHour(1)
           val afterTimeExpected = expected.withDayOfYear(expected.getDayOfYear + 1)
-          equalTo(List(expected, afterTimeExpected, expected, afterTimeExpected))
+          equalTo(List(inTimeHourSecond, inTimeHour, inTimeHour, afterTimeExpected))
         }
       },
       test("throw IllegalArgumentException on invalid `hour` argument of `hourOfDay`") {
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -174,7 +196,7 @@ trait Schedule[-Env, -In, +Out] extends Serializable { self =>
   def ||[Env1 <: Env, In1 <: In, Out2](that: Schedule[Env1, In1, Out2])(implicit
     zippable: Zippable[Out, Out2]
   ): Schedule.WithState[(self.State, that.State), Env1, In1, zippable.Out] =
-    (self unionWith that)((l, r) => Schedule.minOffsetDateTime(l, r))
+    (self unionWith that)((l, r) => (l union r).getOrElse(l min r))
```
adamgfraser, line 199: I think we probably need a richer representation of `Interval` that allows us to represent disjoint unions. For example, if we have `Schedule.dayOfWeek(2) || Schedule.dayOfWeek(4)` we want to preserve the full information that the schedule is willing to recur on either day because we may later intersect it within a schedule that is only willing to recur on the later day.

core/shared/src/main/scala/zio/Schedule.scala
```diff
@@ -493,23 +516,48 @@ trait Schedule[-Env, -In, +Out] extends Serializable { self =>
...
+  )(f: (Interval, Interval) => Interval)(implicit
+    zippable: Zippable[Out, Out2]
+  ): Schedule.WithState[(self.State, that.State), Env1, In1, zippable.Out] =
     new Schedule[Env1, In1, zippable.Out] {
       type State = (self.State, that.State)
-      val initial: State =
-        (self.initial, that.initial)
+      val initial = (self.initial, that.initial)
+
+      def loop(
```
adamgfraser, line 526: We recurse here until either schedule is done or we find a nonempty interval when both schedules are willing to recur. There is some question here of whether this should be the responsibility of the schedule or the driver but I think this makes the most sense because there are some ways of running the schedule other that the driver and otherwise we need to keep the last output in the state which will create variance issues.

core/shared/src/main/scala/zio/Schedule.scala
jdegoes: Another approach is to keep track of something like "current interval", which would be the interval that either overlaps with the current moment in time (if there is such an interval), or the first interval after the current moment in time (if there is no such interval). If you are keeping track of the "current interval" on both sides for an intersection or union, then you can make a decision about the union or intersection without first performing an interval union / intersection. Basically, the current intervals become the "system of record" and decisions about the next schedule interval occur based on those.

core/shared/src/main/scala/zio/Schedule.scala
jdegoes: Interestingly there is a possibility of non-termination.

core-tests/shared/src/test/scala/zio/ScheduleSpec.scala
```diff
@@ -579,13 +577,28 @@ object ScheduleSpec extends ZIOBaseSpec {
...
+        test("hourOfDay")(checkRepetitions(Schedule.hourOfDay(1))),
+        test("minuteOfHour")(checkRepetitions(Schedule.minuteOfHour(1))),
         test("recurs")(checkRepetitions(Schedule.forever)),
+        test("secondOfMinute")(checkRepetitions(Schedule.secondOfMinute(1))),
         test("spaced")(checkRepetitions(Schedule.spaced(1.second))),
         test("windowed")(checkRepetitions(Schedule.windowed(1.second)))
       )
-    )
+    ),
+    test("intersection of schedules recurring in bounded intervals") {
```
jdegoes, line 592: This is good. It would be a good hackathon ticket to add more tests for intersection and union with the bounded intervals case.

jdegoes (approved):

PR #5529, opened 2021-09-06
Documentation

atooni:
This is a refactoring for the docusaurus v2 website using only Infima CSS or customized Infima stylesheets. It does not use Tailwind to generate Custon CSS.

- [x] Generate basic versioned site 
- [x] Check all links 
- [ ] Regenerate Algolia index (search is currently broken)
- [ ] Review analytics setup if required 
- [ ] Optional: Generate versioned 1.x sidebar from 1.x sidebar.json
- [ ] Update contributors guide to describe specifics for working with versioned docs
- [ ] Optional: Prepare for I18n ?


atooni:
Working on a versioned ZIO page. 

[ ] Generate versioned site 
[ ] Check links (automatically ?)
[ ] Optional: Generate version 1 sidebar from 1.x sidebar.json

khajavi:
@atooni 
Fantastic work, although I haven't checked your works here, would like to know if you can fix sidebar alignment on wide screens:
image

atooni:
@khajavi Hi there, I can see why you are asking that and I have asked myself the same question. At the moment I have done very minmal changes to the components of the Docusaurus Default Theme components. AFAIK we would have to change the Layout component. 

The default theme ises Infima as a CSS framework and from the documentation I have not seen how I could change the layout - and even more to do so responsively. 

My first of the webpage used Tailwind CSS on top of Infima. That was a bit of an overkill, so I decided to go with Infima entirely. This having said - creating a responsive layout with Tailwind is very straight forward.

From my perspective we could do :

* Either we stick with Infima. That would mean studying the CSS code to see what has to be changed and how to achieve the desired layout. 
* Try to understand whether we could create a tailwind based theme and leverage the config options in there. 

At the moment it is a bit hard to decide as Infima is developed alongside Docusaurus 2 and that is in beta-6 right now. It is fairly stable for building sites, but the Theme API is not yet final. 

I would suggest to finalize the work on the docs for now and keep the layout change request in a separate ticket afterwards.

khajavi:
@atooni
Hi Andreas,
Thank you for your detailed reply. Until now, since the dacusaurus has not been released yet, I decided not to start the migration. However, after seeing your work, I find it worthwhile. I agree with you to postpone layout issues. I will create another issue to collect layout problems.
Looking forward to the results of your work, especially the versioned documentation. 💯

adamgfraser (approved):

PR #5533, opened 2021-09-07
Fast Chunk Updates

### LINKED ISSUE #5525: Support Efficient Updates Of Chunks

adamgfraser:
Resolves #5525.

We accumulate updates in a buffer, allowing us to amortize the cost of rebuilding the chunk over multiple updates at the cost of slower lookups until we materialize it.

```
[info] Benchmark                           (size)   Mode  Cnt     Score    Error  Units
[info] ChunkUpdateBenchmarks.chunkUpdate    10000  thrpt   15  5214.138 ± 55.786  ops/s
[info] ChunkUpdateBenchmarks.vectorUpdate   10000  thrpt   15  3836.942 ± 48.807  ops/s
```


jdegoes (approved):

jdegoes:
Similar idea to `:+`. Love it!

PR #5537, opened 2021-09-07
Return Option from ZIO.When

### LINKED ISSUE #5534: Create .whenOption operator
It would be great if we'd have a `.whenOption` operator, that returns a `Option[A]` when applied to a  `ZIO[_,_,A]` that is `Some(A)` when the effect was executed and `None` when it wasn't executed. 

Code that I have that could use it: 
```scala
    val e: ZIO[System, Throwable, Option[File]] = for {
    sys <- ZIO.service[zio.system.System.Service]
    shouldLog <- sys.envOrElse("TRACE_REQUESTS", "false").map(_.toLowerCase == "true").orDie
    logLocation <- sys.envOrElse("TRACE_REQUESTS_STORAGE_PATH", ".").orDie
    logDir <- createLogDirectory(correlationId, jobId, calculationId, logLocation).whenOption(shouldLog)
  } yield (logDir)
```

I think the code could like this: 
```scala
  def whenOption[R, E,A](b: => Boolean)(zio: => ZIO[R, E, A]): ZIO[R, E, Option[A]] =
    effectSuspendTotal(if (b) zio.map(Some(_)) else None)
 ```
 
 WDYT?

adamgfraser:
Resolves #5534.

One issues this raises a bit is whether we should be using `Any` more pervasively when an effect doesn't return a meaningful result.

### DIFF
--- a/benchmarks/src/main/scala/zio/BenchmarkUtil.scala
+++ b/benchmarks/src/main/scala/zio/BenchmarkUtil.scala
@@ -20,7 +20,7 @@ object BenchmarkUtil extends Runtime[ZEnv] {
     else zio *> repeat(n - 1)(zio)
 
   def verify(cond: Boolean)(message: => String): IO[AssertionError, Unit] =
-    ZIO.when(!cond)(IO.fail(new AssertionError(message)))
+    ZIO.when(!cond)(IO.fail(new AssertionError(message))).unit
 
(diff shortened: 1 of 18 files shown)


jdegoes (approved):

jdegoes:
> One issues this raises a bit is whether we should be using Any more pervasively when an effect doesn't return a meaningful result.


I think that we should, and this will motivate that direction further.

adamgfraser:
Okay, I will open an issue for that and start working on it.

PR #5549, opened 2021-09-07
Implement ZStream#changesWithZIO

adamgfraser:


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -606,18 +606,29 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
    * whether two elements are equal.
    */
   def changesWith(f: (O, O) => Boolean): ZStream[R, E, O] =
+    changesWithZIO((o, o1) => ZIO.succeedNow(f(o, o1)))
```
jdegoes, line 609: This specialized variant may be higher enough performance that it's worth keeping around. /cc @iravid

jdegoes (dismissed):

streams/shared/src/main/scala/zio/stream/ZStream.scala
iravid: Yeah, tend to agree. That's also important in the channels version where avoiding ZIO effects skips ZIO entirely.

iravid (approved):

PR #5552, opened 2021-09-08
Deprecate ZIO#forkInternal

adamgfraser:
This operator no longer makes sense based on the recent changes to `reportFailure`.


jdegoes (approved):
Love the simplification. ❤️

PR #5554, opened 2021-09-08
Check Bounds In Chunk Operators

adamgfraser:


core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1472,7 +1474,9 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
       }
 
     def apply(n: Int): A =
-      if (n < start.length) start(n) else buffer(n - start.length).asInstanceOf[A]
+      if (n < 0 || n >= length) throw new IndexOutOfBoundsException(s"Append chunk access to $n")
```
jdegoes, line 1477: In any case where `length` is used more than once, it is best to cache it into a local variable, as a general rule in the `Chunk`.* methods.

jdegoes (approved):

jdegoes:
Looks great, will merge now! Would be good to add a ticket to test these boundary conditions. 👍

PR #5556, opened 2021-09-08
Update scalafmt to 3.x

mijicd:
Closes #5521 

This pull requests "helps out" Scala Steward to upgrade scalafmt :). Note that I'm not sure if we should even pursue this path due to the diff size. In case we decide not to, I'd modify this PR to ignore further scalafmt updates.


core/jvm/src/main/scala/zio/internal/OneElementConcurrentQueue.scala
```diff
@@ -20,26 +20,18 @@ import java.io.Serializable
...
- *      12     4                                           int OneElementConcurrentQueue.capacity
- *      16     4   java.util.concurrent.atomic.AtomicReference OneElementConcurrentQueue.ref
- *      20     4         java.util.concurrent.atomic.LongAdder OneElementConcurrentQueue.deqAdder
- * Instance size: 24 bytes
- * Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
+ * zio.internal.OneElementConcurrentQueue object internals: OFFSET SIZE TYPE DESCRIPTION 0 4 (object header) 4 4 (object
+ * header) 8 4 (object header) 12 4 int OneElementConcurrentQueue.capacity 16 4
+ * java.util.concurrent.atomic.AtomicReference OneElementConcurrentQueue.ref 20 4 java.util.concurrent.atomic.LongAdder
+ * OneElementConcurrentQueue.deqAdder Instance size: 24 bytes Space losses: 0 bytes internal + 0 bytes external = 0
+ * bytes total
```
mijicd, line 34: Stuff like this obviously must not be formatted.

adamgfraser:
I'm not sure the new formatting is any better. I think I might be inclined to leave it for now and maybe at some point we could look at settings that produce more minimal / positive changes when someone has time.

mijicd:
In that case I'll stop the updates and we can experiment with it in the background.

core/jvm/src/main/scala/zio/internal/OneElementConcurrentQueue.scala
sideeffffect: We could perhaps use `{{{`/`}}}` for this?

sideeffffect:
I'm a bit late to the party, but I'm of the opinion that it's better to bite the bullet and go with the development of Scalafmt. Sure, this makes many changes, but it's all automated. The more we lag behind, the more the eventual inevitable change will hurt.

Apart from exceptions like custom formatted tables (which are solvable), the changes are fine.

adamgfraser:
I agree we need to update, I just think we need to look at the setting to minimize the changes. Making all the Scaladoc use even more columns, for example, is not helping. I will take a look at this.

sideeffffect:
Oh, I see what you mean.

adamgfraser:
I think #5577 resolves this in a pretty satisfactory way.

PR #5558, opened 2021-09-08
Generalize Type Signature of ServiceWith

### LINKED ISSUE #5555: ZIO.serviceWith is too restrictive
Currently it's implemented as:

```scala
  final class ServiceWithPartiallyApplied[Service](private val dummy: Boolean = true) extends AnyVal {
    def apply[E, A](f: Service => ZIO[Has[Service], E, A])(implicit tag: Tag[Service]): ZIO[Has[Service], E, A] =
      new ZIO.Read(r => f(r.get))
  }
```

which requires that `f` return a `ZIO[Has[Service], ...]`. Why do we need that?

Let's say `Service.doSomething` is a function with type `ZIO[Has[AnotherService], ...]`, then this works:

```scala
ZIO.service[Service].flatMap(_.doSomething) :: ZIO[Has[Service] with Has[AnotherService], ...]
```

while this doesn't:

```scala
ZIO.serviceWith[Service](_.doSomething)
```

adamgfraser:
Fixes #5555.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -5415,7 +5415,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   }
 
   final class AccessZIOPartiallyApplied[R](private val dummy: Boolean = true) extends AnyVal {
-    def apply[E, A](f: R => ZIO[R, E, A]): ZIO[R, E, A] =
+    def apply[R1 <: R, E, A](f: R => ZIO[R1, E, A]): ZIO[R with R1, E, A] =
       new ZIO.Read(f)
   }
 
--- a/core/shared/src/main/scala/zio/ZIO.scala
+++ b/core/shared/src/main/scala/zio/ZIO.scala
@@ -5427,7 +5427,9 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   }
 
   final class ServiceWithPartiallyApplied[Service](private val dummy: Boolean = true) extends AnyVal {
-    def apply[R <: Has[Service], E, A](f: Service => ZIO[R, E, A])(implicit tag: Tag[Service]): ZIO[R with Has[Service], E, A] =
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -5427,7 +5427,7 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   }
 
   final class ServiceWithPartiallyApplied[Service](private val dummy: Boolean = true) extends AnyVal {
-    def apply[E, A](f: Service => ZIO[Has[Service], E, A])(implicit tag: Tag[Service]): ZIO[Has[Service], E, A] =
+    def apply[R <: Has[Service], E, A](f: Service => ZIO[R, E, A])(implicit tag: Tag[Service]): ZIO[R with Has[Service], E, A] =
```
jdegoes: Will this work:

```suggestion
    def apply[R, E, A](f: Service => ZIO[R, E, A])(implicit tag: Tag[Service]): ZIO[R, E, A] =
```

adamgfraser: That was my original formulation. 😃

The issue there is that it leads to a lot of spurious warnings about `R` being inferred as `Any`.

adamgfraser: Or to be more precise there are a couple of permutations:

```scala
    def apply[R, E, A](f: Service => ZIO[R, E, A])(implicit
      tag: Tag[Service]
    ): ZIO[R with Has[Service], E, A] =
```

This one leads to spurious warnings about a type being inferred as `Any`.

```scala
    def apply[R <: Has[Service], E, A](f: Service => ZIO[R, E, A])(implicit
      tag: Tag[Service]
    ): ZIO[R, E, A]
```

This one leads to numerous warnings of missing parameter type for the effectual function.

jdegoes (approved):

PR #5563, opened 2021-09-09
Shared Layers

### LINKED ISSUE #4519: ZIO Test: Share layers across specs

### LINKED ISSUE #5093: ZIO Test: Make specs more compositional
Depends on #5092 .

The goal of this ticket is to support:

```scala
object MySpec1 extends DefaultRunnableSpec ...

object MySpec2 extends DefaultRunnableSpec...

object MySpec3 extends DefaultRunnableSpec {
  def spec = MySpec1 + MySpec2
}
```

In order to support this, we would probably have to change `Spec` from a case class to a `trait` with a `fold` method that encodes the subtypes of `SpecCase`. 

Then `DefaultRunnableSpec` and friends could implement the abstract methods of the `Spec` trait and get operators like `+` for free.

### LINKED ISSUE #5536: ZIO 2.0: Simplify runnable spec (ZIO Test)
Requires discussion:

In an ideal world, there would be only 1 spec type, which would itself extend `zio.ZIOApp`:

```scala
object MySpec extends ZIOSpec {
  override def layer = ZEnv.live ++ myLayer

  def spec = ...
}
```

adamgfraser:
Resolves #4519. Resolves #5093. Resolves #5536.

Implements `ZIOSpec`, a spec that includes a layer that will be globally shared. Combining specs results in a new spec where running the spec will build the layers from all of the specs and then provide them to all the tests, so memoization of layers will automatically take care of making sure we don't build the same layer twice.

The next step, still to be done here, is to update the SBT test runner to reflectively instantiate all specs and then to combine them into a single spec to achieve global sharing of layers across tests.

### DIFF
--- a/test/shared/src/main/scala/zio/test/ZIOSpec.scala
+++ b/test/shared/src/main/scala/zio/test/ZIOSpec.scala
@@ -16,80 +16,10 @@
 
 package zio.test
 
-import org.portablescala.reflect.annotation.EnableReflectiveInstantiation
 import zio._
-import zio.test.environment.TestEnvironment
-import zio.test.render._
 
-@EnableReflectiveInstantiation
 abstract class ZIOSpec[R <: Has[_]: Tag] extends ZIOApp { self =>
+  type Environment = R
 
-  def layer: ZLayer[TestEnvironment, Any, R]
-
-  def spec: ZSpec[R with TestEnvironment with Has[ZIOAppArgs], Any]
-
-  def aspects: Chunk[TestAspect[Nothing, R with TestEnvironment with Has[ZIOAppArgs], Nothing, Any]] =
-    Chunk.empty
-
-  final def run: ZIO[ZEnv with Has[ZIOAppArgs], Any, Any] =
-    runSpec.provideSomeLayer[ZEnv with Has[ZIOAppArgs]](TestEnvironment.live ++ (TestEnvironment.live >>> layer))
-
-  final def <>[R1 <: R: Tag](that: ZIOSpec[R1]): ZIOSpec[R with R1] =
-    new ZIOSpec[R with R1] {
-      def layer: ZLayer[TestEnvironment, Any, R with R1] =
-        self.layer ++ that.layer
-      override def runSpec: ZIO[R with R1 with TestEnvironment with Has[ZIOAppArgs], Any, Any] =
-        self.runSpec.zipPar(that.runSpec)
-      def spec: ZSpec[R with R1 with TestEnvironment with Has[ZIOAppArgs], Any] =
-        self.spec + that.spec
-    }
-
-  protected def runSpec: ZIO[R with TestEnvironment with Has[ZIOAppArgs], Any, Any] =
-    for {
-      args     <- ZIO.service[ZIOAppArgs]
-      testArgs  = TestArgs.parse(args.args.toArray)
-      exitCode <- runSpec(spec, testArgs)
-      _        <- doExit(exitCode)
-    } yield ()
-
-  private def createTestReporter(rendererName: String): TestReporter[Any] = {
-    val renderer = rendererName match {
-      case "intellij" => IntelliJRenderer
-      case _          => TestRenderer.default
-    }
-    DefaultTestReporter(renderer, TestAnnotationRenderer.default)
-  }
-
-  private def doExit(exitCode: Int): UIO[Unit] =
-    exit(ExitCode(exitCode)).when(!isAmmonite).unit
-
-  private def isAmmonite: Boolean =
-    sys.env.exists { case (k, v) =>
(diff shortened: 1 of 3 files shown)


test/shared/src/main/scala/zio/test/ZIOSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+import zio._
+import zio.test.environment.TestEnvironment
+import zio.test.render._
+
+@EnableReflectiveInstantiation
+abstract class ZIOSpec[R <: Has[_]: Tag](val layer: ZLayer[TestEnvironment, Any, R]) extends ZIOApp { self =>
+
+  def spec: ZSpec[R with TestEnvironment with Has[ZIOAppArgs], Any]
+
+  def aspects: Chunk[TestAspect[Nothing, R with TestEnvironment with Has[ZIOAppArgs], Nothing, Any]] =
```
adamgfraser, line 29: Combination of aspects is somewhat ill defined here. It doesn't really matter because a combined spec will run each individual spec with its own aspects, which is the right thing to do I think, but not clear what the interpretation of calling this on a combined spec should be. Perhaps it should be empty in the case of combined specs?

test/shared/src/main/scala/zio/test/ZIOSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+package zio.test
+
+import org.portablescala.reflect.annotation.EnableReflectiveInstantiation
+import zio._
+import zio.test.environment.TestEnvironment
+import zio.test.render._
+
+@EnableReflectiveInstantiation
+abstract class ZIOSpec[R <: Has[_]: Tag](val layer: ZLayer[TestEnvironment, Any, R]) extends ZIOApp { self =>
```
adamgfraser, line 25: Ideally would like to unify over the case where the environment is `Any`. May be able to use `Has.Union` to do this.

test/shared/src/main/scala/zio/test/ZIOSpec.scala
adamgfraser: If you think of this as a tree like structure and `aspects` as representing aspects at that level of the tree then empty actually makes sense here. When we combine two specs each spec may have its own aspects but there are no aspects that apply to the entire tree at that point. We could potentially add an operator allowing adding additional aspects to the new root.

test/shared/src/main/scala/zio/test/ZIOSpec.scala
adamgfraser: May just be easier to have `DefaultZIOSpec` and `ZIOSpec` here.

jdegoes: I think this `val layer` could as well be a `def`, since we are using `def` style for `spec` and it's odd to mix the two styles.

I agree on `ZIOSpec` and `ZIOSpecDefault`. I'd even suggest that the latter _extend_ the former, so everything can be treated as a `ZIOSpec`.

test/shared/src/main/scala/zio/test/ZIOSpec.scala
jdegoes: I agree that in the composition, this should be empty since the aspects have been pushed into the tests. Really, `aspects` is just a sort of "lazy application" of aspecs to those tests defined in `spec`, so this is the only sensible composition.

test/shared/src/main/scala/zio/test/ZIOSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+package zio.test
+
+import org.portablescala.reflect.annotation.EnableReflectiveInstantiation
+import zio._
+import zio.test.environment.TestEnvironment
+import zio.test.render._
+
+@EnableReflectiveInstantiation
```
jdegoes, line 24: This will only work with a non-polymorphic trait. Hence the need to have a dummy parent without type parameters (only type members).

test/shared/src/main/scala/zio/test/ZIOSpec.scala
adamgfraser: Done.

test/shared/src/main/scala/zio/test/ZIOSpec.scala
adamgfraser: Done.

jdegoes (approved):
Although this may be tweaked a bit in the future, it's a great step forward, will merge now. Thank you for your work on this!

PR #5571, opened 2021-09-10
Implement ZStream#zipAllSortedByKeyWith

adamgfraser:


adamgfraser:
Perhaps this should be called `zipSortedByKeyWith`? I can imagine another version that just emits elements from both streams instead of combining them, using the ordering on the keys to determine which stream to pull from next.

iravid:
Yes you're right.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4574,4 +4574,153 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
     def single(a: A): B =
       apply(ZIO.succeedNow(Chunk(a)))
   }
+
+  implicit final class SortedByKey[R, E, K, A](private val self: ZStream[R, E, (K, A)]) {
```
iravid, line 4578: Is it possible to put this on the class itself with an implicit subtype evidence for the tuple? Should be more inline with the other operators.

iravid (approved):
Looks great!

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: No unfortunately it leads to significantly worse type inference because then the input type to the combining function needs to be specified.

iravid: Ah alright.

iravid (approved):
Great additions with the variants. 💪🏼

PR #5573, opened 2021-09-10
Add tests and better null handling for isThrowableCause

stephendavidmarsh:
This PR changes the behavior of `isThrowableCause` to better handle exceptions with `null` as their cause (all exceptions not created with an explicit cause). Instead of producing a null pointer exception, it will produce an appropriate "does not match" message.

Tests  are also added for `isThrowableCause`.

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
@@ -570,17 +570,17 @@ object AssertionSpec extends ZIOBaseSpec {
       assert(assertion.equals(assertionM))(isFalse ?? "assertion != assertionM") &&
       assert(assertionM.equals(assertion))(isFalse ?? "assertionM != assertion")
     },
-    test("isThrowableCause must succeed when supplied value has matching cause") {
+    test("hasThrowableCause must succeed when supplied value has matching cause") {
       val cause = new Exception("cause")
       val t = new Exception("result", cause)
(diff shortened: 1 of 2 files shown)


test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
```diff
@@ -569,7 +569,21 @@ object AssertionSpec extends ZIOBaseSpec {
       val assertionM = AssertionM.assertionDirect[Unit]("sameName")()(_ => ???)
       assert(assertion.equals(assertionM))(isFalse ?? "assertion != assertionM") &&
       assert(assertionM.equals(assertion))(isFalse ?? "assertionM != assertion")
-    }
+    },
+    test("isThrowableCause must succeed when supplied value has matching cause") {
```
adamgfraser, line 573: "hasThrowableCause" here?

adamgfraser (changes requested):
One very minor comment but otherwise looks great! Thank you for this!

test-tests/shared/src/test/scala/zio/test/AssertionSpec.scala
stephendavidmarsh: :facepalm: Yep, hasThrowableCause, fixed in the commit below.

adamgfraser: Awesome!

adamgfraser (approved):

adamgfraser (approved):

Issue #5540, opened 2021-09-07
Migrate JVM Specific Metrics from ZIO ZMX


adamgfraser:
Previously a bunch of nice JVM specific metrics were added to ZIO ZMX by @vigoo in zio/zio-zmx#282. These need to be migrated to `zio` using the new `ZIOMetric` data type.

adamgfraser:
Closed by #5582.

PR #5575, opened 2021-09-10
Prometheus-style JVM metrics

vigoo:
Resolves https://github.com/zio/zio/issues/5540


vigoo:
See https://github.com/zio/zio/pull/5582

PR #5579, opened 2021-09-10
Enable fatal-warnings

sideeffffect:
Scalafix, when run on its own, suppresses `fatal-warnings`, so that it can operate smoothly (and not get shot down because there is a warning in the codebase).

But in our case, we want to verify both that Scalafix is happy _and_ that the codebase doesn't contain any warnings.

related https://github.com/zio/zio-prelude/pull/727


adamgfraser (approved):

sideeffffect:
Green :tada:

PR #5581, opened 2021-09-11
Rename ZIO#unoption to ZIO#unsome

adamgfraser:
`unoption` is the inverse operation of `some` which is a little confusing and not consistent with our `left` and `unfelt` naming convention. I think if we are going to go with the `unXYZ` naming convention it needs to be `unsome`.


jdegoes (approved):
❤️

PR #5582, opened 2021-09-11
JVM metrics + DefaultJvmMetricsExporter ZIOApp

vigoo:
This is my proof of concept solution for specifying "daemons" to be executed as background processes through the whole lifecycle of a (potentially composed) application.

I based it on top of https://github.com/zio/zio/pull/5575 to show the motivating use case.

I think "Enabling JVM metrics" conceptually belongs to aspects like tracing and logging. More than that, if multiple `ZIOApps` would enable collecting these metrics with some of the available methods by providing a layer around their main program, or by expressing it as `ZIOApp` "sidecar" composed together with the main app, further composition of them would lead to multiple instances of the metrics reporter.

This solution eliminates this problem by maintaining a set of "daemons" to be executed on the edge of the composed application.


adamgfraser:
@vigoo Can you rebase on top of #5643  and implement this functionality in terms of layers?

core/jvm/src/main/scala-2.11-2.12/zio/metrics/jvm/JvmMetricsVersionSpecific.scala
```diff
@@ -0,0 +1,11 @@
+package zio.metrics.jvm
+
+import scala.collection.JavaConverters._
+
+trait JvmMetricsVersionSpecific {
```
adamgfraser, line 5: Minor thing but in other places we have used silencer to silence the deprecation warnings and avoid having version specific files for this.

core/jvm/src/main/scala/zio/metrics/jvm/DefaultJvmMetrics.scala
```diff
@@ -0,0 +1,37 @@
...
+        VersionInfo.collectMetrics
+    ).unit
+
+  /** Layer that starts collecting the same JVM metrics as the Prometheus Java client's default exporters */
+  val live: ZLayer[Has[Clock] with Has[System], Throwable, Has[DefaultJvmMetrics]] =
+    collectDefaultJvmMetrics.as(new DefaultJvmMetrics {}).toLayer
+}
+
+/** A ZIO application that collects the same JVM metrics as the Prometheus Java client's default exporters. */
+object DefaultJvmMetricsExporter extends ZIOApp {
```
adamgfraser, line 27: Could potentially have a `JvmMetricsExporter` that leaves the `Schedule` for metrics collection abstract and then extend that.

core/jvm/src/main/scala/zio/metrics/jvm/DefaultJvmMetrics.scala
```diff
@@ -0,0 +1,37 @@
+package zio.metrics.jvm
+
+import zio._
+
+trait DefaultJvmMetrics
+
+/** JVM metrics, compatible with the prometheus-hotspot library */
+object DefaultJvmMetrics {
+  val collectDefaultJvmMetrics: ZManaged[Has[Clock] with Has[System], Throwable, Unit] =
```
adamgfraser, line 9: This is going to be faster if you use `foreachParDiscard`. Might also change if you end up composing at the `App` level instead of the `ZManaged` level.

core/jvm/src/main/scala/zio/metrics/jvm/BufferPools.scala
```diff
@@ -0,0 +1,45 @@
...
+  private def bufferPoolUsedBuffers(pool: String): ZIOMetric.Gauge[Long] =
+    ZIOMetric.setGaugeWith("jvm_buffer_pool_used_buffers", MetricLabel("pool", pool))(_.toDouble)
+
+  private def reportBufferPoolMetrics(
+    bufferPoolMXBeans: List[BufferPoolMXBean]
+  ): ZIO[Any, Throwable, Unit] =
+    ZIO.foreachParDiscard(bufferPoolMXBeans) { bufferPoolMXBean =>
+      for {
+        name <- Task(bufferPoolMXBean.getName)
+        _    <- Task(bufferPoolMXBean.getMemoryUsed) @@ bufferPoolUsedBytes(name)
```
adamgfraser, line 27: Not related to this PR but you could imagine there being an `unsafe` method to update a metric so this whole thing could take place within one `ZIO.succeed` block.

core/jvm/src/main/scala/zio/metrics/jvm/BufferPools.scala
```diff
@@ -0,0 +1,45 @@
...
+        _    <- Task(bufferPoolMXBean.getTotalCapacity) @@ bufferPoolCapacityBytes(name)
+        _    <- Task(bufferPoolMXBean.getCount) @@ bufferPoolUsedBuffers(name)
+      } yield ()
+    }
+
+  val collectMetrics: ZManaged[Has[Clock], Throwable, Unit] =
```
adamgfraser, line 33: Perhaps this should be a layer so it is more idiomatic if users want to collect just one type of JVM metric versus all of them? You can even imagine each of these being its own `ZIOApp` and then the `ZIOApp` that starts all of them is just the composition of those applications.

core/jvm/src/main/scala/zio/metrics/jvm/BufferPools.scala
```diff
@@ -0,0 +1,45 @@
...
+    ZManaged.acquireReleaseWith {
+      for {
+        bufferPoolMXBeans <-
+          Task(fromJavaList(ManagementFactory.getPlatformMXBeans(classOf[BufferPoolMXBean])).toList)
+        fiber <-
```
adamgfraser, line 38: I think `forkManaged` will take care of a lot of this logic for you.

adamgfraser (dismissed):
Left some comments but looks great!

vigoo:
Good comments, thanks, will address them soon

core/jvm/src/main/scala/zio/metrics/jvm/BufferPools.scala
vigoo: That would probably also help in some migration situations allowing you to (temporarily) not convert something to an effect that deep inside called some non-ZIO metric API to report a metric.

adamgfraser (approved):

PR #5584, opened 2021-09-11
Expose State In Additional Schedule Operators

adamgfraser:
I missed preserving type information on the `State` in a couple of the `Schedule` operators


jdegoes (approved):
👍

PR #5587, opened 2021-09-12
Update README.md (Add Sanjagh.pro)

majidhosseini:
Add Sanjagh company, we use scala language and use ZIO in our Microservices specific ZIO and ZLayer


adamgfraser (approved):
Awesome!

PR #5591, opened 2021-09-13
Add "tapError" operator to ZStream

jypma:
This commits adds a tapError operator to ZStream, with the same
signature as the one on plain ZIO. The use of a "None" error to signal
end-of-stream is kept internal, as it's not a semantic error as such.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -2837,6 +2837,15 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
     mapM(o => f0(o).as(o))
 
   /**
+   * Returns a stream that effectfully "peeks" at the failure of the stream.
+   */
+  final def tapError[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, Any])(implicit ev: CanFail[E]): ZStream[R1, E1, O] =
+    ZStream(self.process.map(_.tapError {
+      case None      => ZIO.fail(None)
+      case Some(err) => f(err).mapError(Some(_))
+    }))
+
+  /**
    * Throttles the chunks of this stream according to the given bandwidth parameters using the token bucket
    * algorithm. Allows for burst in the processing of elements by allowing the token bucket to accumulate
    * tokens up to a `units + burst` threshold. Chunks that do not meet the bandwidth constraints are dropped.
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2836,6 +2836,16 @@ abstract class ZStream[-R, +E, +O](val process: ZManaged[R, Nothing, ZIO[R, Opti
...
 
+  /**
+   * Returns a stream that effectfully "peeks" at the failure of the stream.
+   */
+  final def tapError[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, Any])(implicit ev: CanFail[E]): ZStream[R1, E1, O] = {
+    ZStream(self.process.map(_.tapError {
+      case None => ZIO.fail(None)
+      case Some(err) => f(err).mapError(Some(_))
+    }))
+  }
```
adamgfraser, line 2847:
```suggestion
  final def tapError[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, Any])(implicit ev: CanFail[E]): ZStream[R1, E1, O] =
    catchAll(e => ZStream.fromEffect(f(e) *> ZIO.fail(e)))
```

adamgfraser: I think this can be implemented more compositionally using `catchAll`. Good to minimize the number of operators working with the low level stream encoding when possible.

adamgfraser (changes requested):
Great addition! Left one suggestion about the implementation.

streams/shared/src/main/scala/zio/stream/ZStream.scala
jypma: That's indeed more compositional and minimal. Comparing the two, I'm wondering:
- `ZIO.fail` always invokes `trace()` internally. Could that overwrite an earlier trace, and/or have a performance impact?
- `ZStream.catchAll` invokes `ZStream.catchAllCause`, which adds a fair bit of `ZManaged` and `Ref` wizardry; do we need that for a "simple" `tapError` side-effect operation?

adamgfraser: `ZIO.fail` wouldn't overwrite the old trace but would just add a new trace which is very low cost relative to everything else we are doing in streams.

I think you are right about some of the overhead of `catchAllCause` though that really needs to be addressed at the level of `catchAllCause`, which is occurring in the new channel based stream encoding. So one advantage of the more compositional encoding is when we move to the new encoding everything just works.

But I am fine with sticking with this for now.

adamgfraser (approved):

jypma:
thanks! :)

PR #5601, opened 2021-09-13
Remove Assertions

### LINKED ISSUE #5598: ZIO 2.0: Remove Assertion in ZIO Test API
We have places like `ifEnv` (test aspects) that use `Assertion` and which should probably be changed to use Scala boolean functions.

adamgfraser:
Resolves #5598.


test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -413,34 +412,34 @@ object TestAspect extends TimeoutVariants {
    * An aspect that only runs a test if the specified environment variable
    * satisfies the specified assertion.
    */
-  def ifEnv(env: String, assertion: Assertion[String]): TestAspectAtLeastR[Has[Live] with Has[Annotations]] =
+  def ifEnv(env: String, assertion: String => Boolean): TestAspectAtLeastR[Has[Live] with Has[Annotations]] =
```
jdegoes, line 415: I think now that this is no longer an assertion, it would be more idiomatic to take two parameter lists, e.g. `ifEnv("foo")(_.contains("bar"))`.

test/shared/src/main/scala/zio/test/TestAspect.scala
```diff
@@ -413,34 +412,34 @@ object TestAspect extends TimeoutVariants {
...
   def ifEnvSet(env: String): TestAspectAtLeastR[Has[Live] with Has[Annotations]] =
-    ifEnv(env, Assertion.anything)
+    ifEnv(env, _ => true)
 
   /**
    * An aspect that only runs a test if the specified Java property satisfies
    * the specified assertion.
    */
-  def ifProp(prop: String, assertion: Assertion[String]): TestAspectAtLeastR[Has[Live] with Has[Annotations]] =
+  def ifProp(prop: String, assertion: String => Boolean): TestAspectAtLeastR[Has[Live] with Has[Annotations]] =
```
jdegoes, line 432: Ditto here.

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Done!

test/shared/src/main/scala/zio/test/TestAspect.scala
adamgfraser: Done!

jdegoes (approved):

PR #5604, opened 2021-09-14
Collect Results Of Repeat Operators

### LINKED ISSUE #5599: ZIO 2.0: Change ZIO#repeatXYZ combinators to be lossless
Currently a lot of things like `zio.repeatN(2)` just return the last result. It is more useful to return a chunk of results, which allow you to pull out the last yourself with a `map` operation (`zio.repeatN(2).map(_.last)`).

adamgfraser:
Resolves #5599.

I'm not sure we have this figured out yet. A couple of observations:

1. At this point there is a lot of overlap between `repeatN` and `replicateZIO`. They are identical with the exception of being off by one and whether they yield between operations, which is important if you are waiting for some condition to change that may determine whether you repeat but not if you just want to do things a certain number of times.
2. For some of the other variants of `repeat` such as `repeatUntil` I think most of the time you actually don't want the intermediate results. You can always throw them away with `last` but if you are worried about performance or are going to repeat many times you don't want this and kind of want another variant for this which is unfortunate considering how many other closely related operators we have in this space.
3. The "off by one" issue in `repeat` is perhaps more apparent. We could now having `repeat(0)` not run the effect at all, but that would potentially break existing code and also makes it more difficult to get the `last` value if there may not even be a `last` value.


jdegoes:
I agree there are problems and we should defer merging before resolving them.

PR #5627, opened 2021-09-16
Implement Zstream.execute

### LINKED ISSUE #5623: ZIO 2.0: Add ZStream.execute again
```scala
object ZStream {
  def execute[R, E](zio: ZIO[R, E, Any]): ZStream[R, E, Nothing] = ZStream.fromEffect(zio).drain
}
```

adamgfraser:
Resolves #5623.

I wonder if the fact that it doesn't emit any elements is going to be confusing to people if they try to compose other streams after it with operators derived from `flatMap`.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3923,6 +3923,12 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
     fromZIO(ZIO.environment[R])
 
   /**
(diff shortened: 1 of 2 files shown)


jdegoes:
I think that's true. Essentially this stream (and any stream whose success is `Nothing`) should only be composed with `++`.

However, this operator is so important when you want to insert an effect between two streams.

Possibly if we had "prepend effect" and "append affect", we could do without this operator directly.

jdegoes (approved):

PR #5636, opened 2021-09-18
Compositional Chunking

### LINKED ISSUE #5634: ZIO Streams 2.0: Compositional chunking
Using an unsafely created `FiberRef` left into `object ZStream` (or package object `stream`), we should be able to grab the chunk size from the fiber ref, allowing all chunk sizes to be specified via combinator:

```scala
ZStream.chunkSize(1024) {
  ZStream.fromFile("foo.txt")
}
```

adamgfraser:
Resolves #5634.

### DIFF
--- a/core/shared/src/main/scala/zio/ZFiberRef.scala
+++ b/core/shared/src/main/scala/zio/ZFiberRef.scala
@@ -97,6 +97,13 @@ sealed abstract class ZFiberRef[+EA, +EB, -A, +B] extends Serializable { self =>
   def locally[R, EC >: EA, C](value: A)(use: ZIO[R, EC, C]): ZIO[R, EC, C]
 
   /**
+   * Returns a managed effect that sets the value associated with the curent
+   * fiber to the specified value as its `acquire` action and restores it to
+   * its original value as its `release` action.
+   */
+  def locallyManaged(value: A): ZManaged[Any, EA, Unit]
+
+  /**
    * Sets the value associated with the current fiber.
    */
   def set(value: A): IO[EA, Unit]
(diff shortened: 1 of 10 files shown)


adamgfraser:
We could also potentially apply this approach to parallelism and buffer size.

```scala
stream.flatMapPar(f)
stream.flatMapPar(f).withParallelism(8).withBuffer(16)
```

We could potentially push this even further and make the execution strategy itself compositional.

```scala
ZIO.foreach(as)(f)
ZIO.foreach(as)(f).parallel
```

This could allow us to dramatically cut down on the number of variants we have for collection operators, or at least reduce them to convenience methods we provided for the most common cases while making all parallelism versions possible to achieve even for more specialized versions we currently don't provide all the variants for.

vigoo:
One disadvantage I feel is that if you wrap a complex stream into `.parallel` you have to think about each part of the stream to know if it modifies its behavior or not (supports parallelism or not). Probably in many cases the user don't care but still we would loose some readability. 
For the other mentioned ones this argument is not that strong as for example if we set parallelism globally but still have the `Par` suffix then we can immediately see where that applies.

adamgfraser:
One other thing to think about here is how regional settings interact with the fact that stream scopes are dynamic. For example consider:

```scala
stream.withChunkSize(1) *> ZStream.chunkSize
```

Currently the return value of `chunkSize` here will be `1` because the stream after `*>` is being evaluated within the scope of the original stream. This is logical when you think about it but different from `ZIO` where the strong expectation is that the original regional setting has been restored in the continuation.

It also creates some issues with composition because you don't necessarily know if a constructor has changed a regional setting in a way that will impact you. For example, I might have:

```scala
someStreamConstructorFromAnotherLibrary.flatMap(a => generateMyStream(a))
```

If `someStreamConstructor` set the chunk size to `1` that is also going to be the chunk size for my operations which I probably don't want. I could of course reset the chunk size but then I am forced to program somewhat defensively and set my own chunk size instead of relying on the chunk size set at an even high level of the application.

An alternative is to `flatMap` the stream in the implementation of the region operator and to locally set the chunk size back to the original value after emitting each stream element, which is actually what I did in my implementation of `onExecutor` to avoid the issue discussed above.

However, this comes at some cost to performance because you are resetting the regional setting for each stream element and introduces a new concept of scope of "the scope of this stream emitting its elements but not anything that is done with those elements" which is different from the existing concept of stream scopes. Perhaps we could minimize the performance impact but resetting the regional setting for each chunk emitted instead of each value.

Either way, we should probably be consistent between these different operators.

PR #5643, opened 2021-09-18
Upstream Shared Layers To ZIOApp

adamgfraser:
Upstreams the shared layers functionality previously implemented for `ZIOSpec` to `ZIOApp`. Given that this will be the first interaction users have with ZIO we need to spend some time to make sure the API is as nice as possible but I thought it would be helpful to have something to work off of.

Once we have this we can describe background processes as applications that can be composed with other applications and in particular reimplement #5582 in terms of layers.

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZIOAppSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOAppSpec.scala
@@ -47,7 +47,7 @@ object ZIOAppSpec extends ZIOBaseSpec {
         }
       }
 
-      val app1 = ZIOApp(ZIO.fail("Uh oh!"), RuntimeConfigAspect.addLogger(logger1))
+      val app1 = ZIOAppDefault(ZIO.fail("Uh oh!"), RuntimeConfigAspect.addLogger(logger1))
 
       for {
         c <- app1.invoke(Chunk.empty)
--- a/core-tests/shared/src/test/scala/zio/ZIOAppSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZIOAppSpec.scala
@@ -4,28 +4,28 @@ import zio.test._
 
 object ZIOAppSpec extends ZIOBaseSpec {
   def spec: ZSpec[Environment, Failure] = suite("ZIOAppSpec")(
-    test("fromEffect") {
+    test("fromZIO") {
       for {
         ref <- Ref.make(0)
-        _   <- ZIOApp.fromEffect(ref.update(_ + 1)).invoke(Chunk.empty)
+        _   <- ZIOApp.fromZIO(ref.update(_ + 1)).invoke(Chunk.empty)
         v   <- ref.get
       } yield assertTrue(v == 1)
     },
     test("failure translates into ExitCode.failure") {
       for {
-        code <- ZIOApp.fromEffect(ZIO.fail("Uh oh!")).invoke(Chunk.empty)
+        code <- ZIOApp.fromZIO(ZIO.fail("Uh oh!")).invoke(Chunk.empty)
       } yield assertTrue(code == ExitCode.failure)
     },
     test("success translates into ExitCode.success") {
       for {
-        code <- ZIOApp.fromEffect(ZIO.succeed("Hurray!")).invoke(Chunk.empty)
+        code <- ZIOApp.fromZIO(ZIO.succeed("Hurray!")).invoke(Chunk.empty)
       } yield assertTrue(code == ExitCode.success)
     },
     test("composed app logic runs component logic") {
       for {
         ref <- Ref.make(2)
-        app1 = ZIOApp.fromEffect(ref.update(_ + 3))
-        app2 = ZIOApp.fromEffect(ref.update(_ - 5))
+        app1 = ZIOApp.fromZIO(ref.update(_ + 3))
+        app2 = ZIOApp.fromZIO(ref.update(_ - 5))
         _   <- (app1 <> app2).invoke(Chunk.empty)
         v   <- ref.get
       } yield assertTrue(v == 0)
--- a/core/shared/src/main/scala/zio/ZIOApp.scala
+++ b/core/shared/src/main/scala/zio/ZIOApp.scala
@@ -62,7 +62,7 @@ trait ZIOApp { self =>
   final def exit(code: ExitCode): UIO[Unit] =
     UIO {
       if (!shuttingDown) {
-        shuttingDownå = true
+        shuttingDown = true
         try Platform.exit(code.code)
         catch { case _: SecurityException => }
       }
--- a/core/shared/src/main/scala/zio/ZIOApp.scala
+++ b/core/shared/src/main/scala/zio/ZIOApp.scala
@@ -33,7 +33,7 @@ trait ZIOApp { self =>
    * A layer that manages the acquisition and release of services necessary for
    * the application to run.
    */
-  def layer: ZLayer[ZEnv with Has[ZIOAppArgs], Any, Environment]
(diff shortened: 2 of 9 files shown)


core-tests/shared/src/test/scala/zio/ZIOAppSpec.scala
```diff
@@ -47,7 +47,7 @@ object ZIOAppSpec extends ZIOBaseSpec {
         }
       }
 
-      val app1 = ZIOApp(ZIO.fail("Uh oh!"), RuntimeConfigAspect.addLogger(logger1))
+      val app1 = ZIOAppDefault(ZIO.fail("Uh oh!"), RuntimeConfigAspect.addLogger(logger1))
```
vigoo, line 50: Maybe the default could be `ZIOApp` and the customizable something longer like `CustomZIOApp`?

adamgfraser: Yeah I was wondering about that too. I think for most "real" cases you are going to want to be able to provide some layers but it is definitely nice to have something simple for demos and prototyping.

adamgfraser: May also be some value in being consistent with `ZIOSpec` / `ZIOSpecDefault`, though we could change that too.

core/shared/src/main/scala/zio/ZIOApp.scala
```diff
@@ -18,31 +18,37 @@ package zio
...
 
+  implicit def tag: Tag[Environment]
+
+  type Environment <: Has[_]
+
+  /**
+   * A layer that manages the acquisition and release of services necessary for
+   * the application to run.
+   */
+  def layer: ZLayer[ZEnv with Has[ZIOAppArgs], Any, Environment]
```
jdegoes, line 36: The `Has[ZIOAppArgs]` is necessary since those come from the `main` function, implemented by `ZIOApp`. One could argue either way on supplying `ZEnv` to `layer`. If we supply it, it has to be the real one, if it has to be the real one, then we can't provide a test helper for ZIO Test (although we could work around this in other ways).

adamgfraser: I changed the implementation in `ZIOSpec` to override `layer` with a new `final` definition and expose a new `testLayer` method that uses the live environment to build a `TestEnvironment`. We could definitely remove and people could provide `ZEnv` themselves if they needed it.

core/shared/src/main/scala/zio/ZIOApp.scala
```diff
@@ -18,31 +18,37 @@ package zio
...
+   */
+  def run: ZIO[Environment with ZEnv with Has[ZIOAppArgs], Any, Any]
+
   /**
    * Composes this [[ZIOApp]] with another [[ZIOApp]], to yield an application that
    * executes the logic of both applications.
    */
-  final def <>(that: ZIOApp): ZIOApp = ZIOApp(self.run.zipPar(that.run), self.hook >>> that.hook)
+  final def <>(that: ZIOApp): ZIOApp =
+    ZIOApp(self.run.zipPar(that.run), self.layer +!+ that.layer, self.hook >>> that.hook)
```
jdegoes, line 51: I know why the "unsafe" version is being used here `+|+`. It's probably ok in this case although it means the right hand side may clobber some implementations in the left, without the user being aware of this.

adamgfraser: Yes I think it is an unavoidable consequence of using the type member. Perhaps we could log a warning in the implementation of `Has` if we ever have implementations on the left or right and they are not the same object?

core/shared/src/main/scala/zio/ZIOAppDefault.scala
```diff
@@ -0,0 +1,68 @@
...
+    new ZIOAppDefault {
+      override def hook = hook0
+      def run           = run0
+    }
+
+  /**
+   * Creates a [[ZIOAppDefault]] from an effect, using the unmodified default
+   * runtime's configuration.
+   */
+  def fromEffect(run0: ZIO[ZEnv with Has[ZIOAppArgs], Any, Any]): ZIOAppDefault =
```
jdegoes, line 66: We should probably be calling these `fromZIO` this point.

adamgfraser: Yes I noticed that too! Will change!

test/shared/src/main/scala/zio/test/RunnableSpec.scala
```diff
@@ -44,16 +44,23 @@ abstract class RunnableSpec[R, E] extends AbstractRunnableSpec {
...
    * A simple main function that can be used to run the spec.
    */
-  override final def run: ZIO[ZEnv with Has[ZIOAppArgs], Any, Any] =
-    for {
-      args     <- ZIO.service[ZIOAppArgs]
-      testArgs  = TestArgs.parse(args.args.toArray)
-      exitCode <- run(spec, testArgs).provideLayer(runner.bootstrap)
-      _        <- doExit(exitCode)
-    } yield ()
+  final def main(args: Array[String]): Unit = {
```
jdegoes, line 47: This seems a regression of sorts, since you now have to implement `main` rather than `ZIOApp#run`?

test/shared/src/main/scala/zio/test/ZIOSpecDefault.scala
```diff
@@ -5,7 +5,7 @@ import zio.test.environment._
 
 abstract class ZIOSpecDefault extends ZIOSpec[TestEnvironment] {
 
-  final val layer: ZLayer[TestEnvironment, Any, TestEnvironment] = ZLayer.environment
+  final val testLayer: ZLayer[TestEnvironment, Any, TestEnvironment] = ZLayer.environment
```
jdegoes, line 8: Why both `layer` and `testLayer`?

test/shared/src/main/scala/zio/test/RunnableSpec.scala
adamgfraser: This is the old `RunnableSpec` which isn't constrained to the environment being a subtype of `Has` so I reverted to where this does not extend `ZIOApp`. I think this is okay because going forward we want to use `ZIOSpec` for everything, which does extend `ZIOApp`, though we have to finish the word there of implementing support for that in the SBT test runner.

test/shared/src/main/scala/zio/test/ZIOSpecDefault.scala
adamgfraser: At least my thought was that we wanted users to be able to use the `TestEnvironment` in implementing their own layer, and so if the signature has a dependency on `ZEnv` in `ZIOApp` it would be inconsistent to override it with `TestEnvironment` here. But to your point above maybe we just make the input type `Any` and allow the user to provide any needed environment themselves.

jdegoes: I think this is another reason to remove `ZEnv` from `ZIOApp`. Because we can see from this example that different types of applications will have their own types of requirements, so while the default apps / specs can provide default environments, if you want the God mode you should wire everything together yourself (or use helper functions to add things like ZEnv, test environment, etc.). 

(Also from an app composition perspective, polymorphism in layer output is fine while polymorphism in layer input is not fine; so since polymorphism cannot solve this problem it may be best to push it onto the user.)

adamgfraser: Will change.

jdegoes (approved):

PR #5655, opened 2021-09-21
unify assertTrue with assertions

senia-psm:
`assertTrue` should produce and run a regular `Assertion`.
Add `is` assertion macros to use with mocks and `AssertionM`:
```scala
val _: Assertion[User] = is(_.company.get.name == "abc")
```

TODO:

- [ ] fix assertion highlighting. Currently `is` assertion always highlights entire expression
- [ ] add functions support to dotty macros implementation

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
@@ -1,6 +1,7 @@
 package zio.test
 
 import zio.duration.durationInt
+import zio.test.Assertion.{isNonEmpty, isSome, not}
 import zio.test.SmartTestTypes._
 import zio.test.environment.TestClock
 import zio.{Chunk, NonEmptyChunk}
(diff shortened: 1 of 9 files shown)


senia-psm:
At the moment we are getting `() did not satisfy ...` messages.
Can we keep it or should we use something else? Unfortunately there are no failed tests for assertion rendering.
Multi-line values are printed without offset. I'm not sure if this is as intendend since there are no failed tests.
> [info]     () did not satisfy company.users.head.posts.exists(_.title == "hii")
> [info]     • 0 elements satisfied the predicate
> [info]     company.users.head.posts.exists(_.title == "hii")
> [info]     .posts = List(Post(
> [info]   title = "Post #0",
> [info]   publishDate = None
> [info] ), Post(
> [info]   title = "Post #1",
> [info]   publishDate = None
> [info] ))
> [info]     .head = User(
> [info]   name = "Bobo",
> [info]   posts = List(Post(
> [info]     title = "Post #0",
> [info]     publishDate = None
> [info]   ), Post(
> [info]     title = "Post #1",
> [info]     publishDate = None
> [info]   ))
> [info] )
```

**Comment by adamgfraser** [comment]
@senia-psm This is a really interesting effort to unify the classic and smart assertions.

I think the big issue here is that I think our plan has been to deprecate the classic assertions in ZIO 2.0, probably providing them in a separate module so that users don't have to change existing tests. There is definitely some more work to be done figuring out how the smart assertions replace the classic assertions in all cases and actually extracting the classic assertions but I think that is where we have been going.

If so, need to figure out how this PR fits into that. Can everything in the mocking framework be an `assertTrue`?

Alternatively, potentially this PR points the way to unify the existing assertions and smart assertions much like how we now have specific constructors and smart constructors for ZIO values in ZIO 2.0. I worry somewhat that we have two different ways to do the same thing there but perhaps there are ways to address and further unify.

Copying @kitlangton.

**Comment by senia-psm** [comment]
@adamgfraser with unified assertions we can have a generic approach to mocks (not only in the current mocks approach - we need some approach to assertions for any mocks).
I'm not against rewriting assertions from scratch (current assertion-related code is quite cryptic), but I'd argue that we need 2 things:
1. We should be able to work with `Arrow[T, Boolean]`, not only with `Arrow[Any, Boolean]` at top level. For mocks and complicated assertions.
2. We have to be able to compose `Arrow` not only on the same level, but on nested levels:
```scala
val inner: Arrow[D, Boolean] = libraryProvidedAssertion
assertTrue(a.b.c.d should inner)
```

`Arrow[A, Boolean]` and `Assertion[A]` are more or less the same (`T => TestResult`).

PS I'd like to keep `AssertionM` or any other way to work with effectful assertions since I'm using it, but it's not the most impotent point.

jdegoes:
>We should be able to work with Arrow[T, Boolean], not only with Arrow[Any, Boolean] at top level. For mocks and complicated assertions.
> We have to be able to compose Arrow not only on the same level, but on nested levels:

These sound like good ideas. We have to be careful as @adamgfraser mentions, because our intention is indeed to move `Assertion` away, but changes we can make to ZIO Test to make its core more foundational and useful for higher-level layers are much welcome.

jdegoes:
I think this is superceded by work done since the PR. Please re-open with updates if you'd like to do more work around unification. Thank you!

PR #5675, opened 2021-09-26
izumi-reflect 2.0.0

sideeffffect:
A step towards 2.0.1: https://github.com/zio/zio/pull/5590


sideeffffect:
This is ready to be merged :tada: @adamgfraser @mijicd

adamgfraser:
Don't we have an issue that this breaks binary compatible because the hashing algorithm for the tags was changed? Do we need to retarget this to the `series/2.x` branch?

sideeffffect:
Yes, I think you're right Adam. And ZIO 2.x is already on izumi-reflect 2.x.

PR #5677, opened 2021-09-26
Breadth First Search Generators

adamgfraser:
Refactors `Gen#flatMap` to use breadth first search. This will not impact random generators but will significantly increase the utility of deterministic generators.

### DIFF
--- a/test/shared/src/main/scala/zio/test/package.scala
+++ b/test/shared/src/main/scala/zio/test/package.scala
@@ -20,7 +20,7 @@ import zio.stream.{ZSink, ZStream}
 import zio.test.AssertionResult.FailureDetailsResult
 import zio.test.environment._
 
-import scala.collection.immutable.Queue
+import scala.collection.immutable.{Queue => ScalaQueue}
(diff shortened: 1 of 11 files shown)


test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -757,4 +758,159 @@ package object test extends CompileVariants {
           .catchAll(ZStream.succeed(_))
       }
     }
+
+  /**
+   * An implementation of `ZStream#flatMap` that supports breadth first search.
+   */
+  private[test] def flatMapStream[R, R1 <: R, A, B](
```
jdegoes, line 765: Wow! I think tests are the best way to ensure correctness (it's not very easy to audit this code for correctness, at least not quickly).

jdegoes (approved):

test/shared/src/main/scala/zio/test/package.scala
adamgfraser: Added more tests.

jdegoes (approved):

PR #5683, opened 2021-09-27
Managed and Transactional Tests

adamgfraser:
Just like we can have pure and effectual tests, we can support tests that return test results in the context of a managed or transactional effect simply by using the `ZManaged` or committing the transaction.

```scala
test("managed effects can be tested") {
  for {
    ref   <- Ref.make(false).toManaged
    _     <- ZManaged.acquireRelease(ref.set(true))(ref.set(false))
    value <- ref.get.toManaged
  } yield assert(value)(isTrue)
},
test("transactional effects can be tested") {
  for {
    message <- STM.succeed("Hello from an STM transaction!")
  } yield assert(message)(anything)
}
```

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/TestSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/TestSpec.scala
@@ -2,6 +2,7 @@ package zio.test
 
 import zio._
 import zio.Clock._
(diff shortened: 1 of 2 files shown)


jdegoes (approved):
This is really great!

PR #5693, opened 2021-09-30
Preserve Partial Ordering Of Stream Elements In ZStream#zipWithLatest

### LINKED ISSUE #5679: Investigate Issue With ZStream#zipWithLatest
It appears that in the example below values can be emitted out of order.

```scala
import zio._
import zio.console._
import zio.duration._

import java.util.concurrent.Executors
import scala.concurrent.{ExecutionContext, Future}

object MyApp extends zio.App {
  val e1 = Executors.newSingleThreadExecutor()
  val c1 = ExecutionContext.fromExecutor(e1)
  val source1 = ZStream.effectAsync[Any, Throwable, String] {
    k => Future {  
      Thread.sleep(10)
      k.single("A1")
      Thread.sleep(20)
      k.single("A2")
      // k.end // Not ending the stream on purpose
    }(c1).onComplete(_ => e1.shutdown())
  }

  val e2 = Executors.newSingleThreadExecutor()
  val c2 = ExecutionContext.fromExecutor(e2)
  val source2 = ZStream.effectAsync[Any, Throwable, String] {
    k => Future {
      Thread.sleep(20)
      k.single("B1")
      Thread.sleep(130)
      k.single("B2")
      // k.end // Not ending the stream on purpose
    }(c2).onComplete(_ => e2.shutdown())
  }

  val zipped = source1.zipWithLatest(source2) { (c, e) => c + "-" + e }

  val result = zipped
    .interruptWhen(ZIO.sleep(Duration.fromMillis(10000)))
    .runCollect
    .tap(result => putStrLn(result.toString()))

  def run(args: List[String]) = { result.exitCode }
}
```

adamgfraser:
Resolves #5679.

As observed in that issue, `zipWithLatest` can currently emit stream elements in an order that is inconsistent with the order of the original stream elements. For example, if we have a left stream with `A` and `B` and a right stream with `1` and `2`, we can observe outputs such as `A1, B2, B1` and `A1, B2, B2`.

This occurs because of two issues. First, two different `Ref` values are used so updated are not performed atomically, resulting in situations such as where both upstreams update the "latest" values and then get them. Second, we currently perform the effect of reading and updating the `Ref` before merging so we have no guarantee that the order of effects matches the order of stream elements.

This PR fixes these issues so that we have the guarantee that a combination of values that appear at the same point or later in the input streams will appear in an order that is consistent with the order of the original stream elements.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -3401,6 +3401,17 @@ object ZStreamSpec extends ZIOBaseSpec {
                 .take(3)
                 .runCollect
             )(equalTo(Chunk(1, 1, 1)))
+          },
+          testM("preserves partial ordering of stream elements") {
+            val genSortedStream = for {
+              chunk  <- Gen.chunkOf(Gen.int(1, 100)).map(_.sorted)
+              chunks <- splitChunks(Chunk(chunk))
+            } yield ZStream.fromChunks(chunks: _*)
+            checkM(genSortedStream, genSortedStream) { (left, right) =>
+              for {
+                out <- left.zipWithLatest(right)(_ + _).runCollect
+              } yield assert(out)(isSorted)
+            }
           }
         ),
         suite("zipWithNext")(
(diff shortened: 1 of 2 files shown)


iravid (approved):
Great catch. Unifying the refs seems obvious in hindsight.

PR #5711, opened 2021-10-04
Document some missing sections on the STM data type

khajavi:
In this PR I want to introduce both ZTHub and TRandom along with STM type aliases.

### DIFF
--- a/docs/datatypes/stm/zthub.md
+++ b/docs/datatypes/stm/zthub.md
@@ -1,0 +1,20 @@
+---
+id: zthub
+title: "ZTHub"
+---
+
+A `ZTHub[RA, RB, EA, EB, A, B]` is a transactional message hub. Publishers can publish messages of type `A` to the hub and subscribers can subscribe to take messages of type `B` from the hub. Publishing messages can require an environment of type `RA` and fail with an error of type `EA`. Taking messages can require an environment of type `RB` and fail with an error of
+type `EB`.
+
+A `ZTHub` is an asynchronous message hub like `ZHub` but it can participate in STM transactions. APIs are almost identical, but they are in the `STM` world rather than the `ZIO` world.
+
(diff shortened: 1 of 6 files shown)


khajavi:
@jdegoes It's ready for review.

docs/datatypes/stm/zthub.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+A `ZTHub[RA, RB, EA, EB, A, B]` is a transactional message hub. Publishers can publish messages of type `A` to the hub and subscribers can subscribe to take messages of type `B` from the hub. Publishing messages can require an environment of type `RA` and fail with an error of type `EA`. Taking messages can require an environment of type `RB` and fail with an error of
+type `EB`.
+
+A `ZTHub` is an asynchronous message hub like `ZHub` but it can participate in STM transactions. APIs are almost identical, but they are in the `STM` world rather than the `ZIO` world.
+
+The fundamental operators on a `ZTHub` are `publish` and `subscribe`:
+
+```scala
```
adamgfraser, line 13: Could be good to talk about the difference between `subscribe` on `ZHhub` versus THub`.

On `ZHub` `subscribe` returns a `ZManaged` value that takes care of automatically subscribing and unsubscribing for us.

On `THub` to allow us to potentially compose multiple subscriptions transactionally `subscribe` gives us a `USTM[TDequeue[A]]` that we can then `unsubscribe` from by shutting down the queue. If we don't need to compose the subscription itself with other transactions we can use the `subscribeManaged` operator which will give us a transactional subscription in the context of a managed effect that will automatically take care of subscribing and unsubscribing for us.

This is similar to other STM data structures such as `TSemaphore` and `TReentrantLock` where there is a transactional operator and then a managed variant for common use cases.

khajavi: Thank you, Adam! will work on that.

jdegoes (approved):

PR #5717, opened 2021-10-04
Removes ensuringFirst from the new ZStream

vigoo:
Relates to https://github.com/zio/zio/issues/4886 and https://github.com/zio/zio/pull/5669


iravid (approved):
ciao

PR #5720, opened 2021-10-05
Preserve Scope of Forked Fibers in ZStream#interruptWhen

adamgfraser:
When we use `raceFirst` in the implementation of `interruptWhen` the original stream may fork fibers in its implementation, either as part of user code or internal stream operators. We need to transplant these fibers to the current scope to make sure they don't get interrupted prematurely.


iravid (dismissed):
Ah... tricky. I think we would need the same fix in 2.x.

adamgfraser:
I will do in `series/2.x` branch once we merge this.

iravid (approved):

PR #5730, opened 2021-10-06
Update Documentation of Running a ZIO Effect

khajavi:


docs/datatypes/core/runtime.md
```diff
@@ -73,14 +64,30 @@ def myAppLogic =
...
+```
+
+ZIO has a service that contains command-line arguments of an application called `ZIOAppArgs`. We can access command-line arguments using `ZIO.service[ZIOAppArgs]` :
+
+```scala mdoc:compile-only
+
+import zio._
+object HelloApp extends ZIOAppDefault {
+  def run = for {
+    args <- ZIO.service[ZIOAppArgs].map(_.args)
```
khajavi, line 80: As a feedback, an optional helper method for obtaining arguments would be useful.
cc @adamgfraser

jdegoes: It already exists inside `ZIOppDefault` and is called `args`, but maybe it should be called `getArgs`. Probably we should replicated it into the companion object of `ZIOAppArgs` for consistency.

PR #5744, opened 2021-10-09
Execution Tracing

adamgfraser:
Updates work by @robmwalsh presented at ZIO World.


adamgfraser:
@jdegoes Resolved by adding a marker trait similar to what we used to support implicit resolution of instances defined in new type companion objects in ZIO Prelude.

```scala
trait Traced extends Any
type ZTraceElement = Tracer.instance.Type with Traced
```

PR #5778, opened 2021-10-15
Readd helper methods to create histogram buckets with linear / exponential growing boundaries 

atooni:


atooni:
> @atooni I was suggesting in `ZIOMetric.Histogram`.

Yes, figured that out after I asked the question - just updated my PR accordingly

PR #5782, opened 2021-10-16
Added encoders for ZIO Streams.

### LINKED ISSUE #5630: ZIO Streams 2.0: Encoders to match decoders
We have utf8Decode but not utf8Encode, etc.

ithinkicancode:
Resolves #5630


ithinkicancode:
First attempt to add encoders that corresponds to existing decoders. Fixed a bug in `utfDecode` along the way.

ithinkicancode:
TODO's
- [ ] Add scaladoc for the new methods
- [ ] Implement the same for `ZSink`
- [ ] Implement all these as Aspects

jdegoes:
@ithinkicancode Awesome work! Let us know when you are ready for a review. 👍

adamgfraser (changes requested):
Do you want to move this over to the channel encoding?

ithinkicancode:
Yes, the v2 encoders will be implemented as pipelines. 

What has been implemented in this PR are the encoders for the v1 ZStream, which might be removed when the `experimental` ZStream takes over.

adamgfraser:
Yeah I think we are close enough to merging the experimental stream encoding that I would just go straight to that.

adamgfraser (approved):

PR #5783, opened 2021-10-16
Implement ChunkIterator

### LINKED ISSUE #5780: ZIO 2.0: Create ChunkIterator
```scala
trait Chunk[+A] {
  ...
  def chunkIterator(): ChunkIterator[A]
  ...
}

trait ChunkIterator[+A] {
  def hasNextAt(index: Int): Boolean
  def nextAt(index: Int): A
}
```

where the contract is that you call the iterator as normal but you must pass it increasing indexes starting from 0; and if you don't, that's a violation of the contract and the results of iteration are undefined.

This way an array-backed chunk does not need to create or store the index on the heap, which is quite slow compared to the access of a local integer. Yet the lazy chunks can operate incrementally as is in their interest.

adamgfraser:
Resolves #5780.

### DIFF
--- a/benchmarks/src/main/scala/zio/chunks/ChunkArrayBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/chunks/ChunkArrayBenchmarks.scala
@@ -38,5 +38,5 @@ class ChunkArrayBenchmarks {
 
   @Benchmark
   def foldZIO(): Int =
-    BenchmarkUtil.unsafeRun(chunk.foldZIO(0)((s, a) => ZIO.succeed(s + a)))
+    BenchmarkUtil.unsafeRun(chunk.foldZIO[Any, Nothing, Int](0)((s, a) => ZIO.succeed(s + a)))
 }


jdegoes:
Looks great to merge! Would be nice to do a sanity-test benchmark to make sure performance is no worse than before.

adamgfraser:
No appreciable difference in benchmarks for array backed chunks. I think these are already so fast that not keeping the current index on the stack doesn't move the needle relative to other things we are doing in these benchmarks.

`series/2.x`

```
[info] Benchmark                     (size)  Mode  Cnt      Score     Error  Units
[info] ChunkArrayBenchmarks.find       1000  avgt    5      7.796 ±   0.027  ns/op
[info] ChunkArrayBenchmarks.flatMap    1000  avgt    5  32491.004 ± 394.033  ns/op
[info] ChunkArrayBenchmarks.fold       1000  avgt    5   2475.732 ±  24.744  ns/op
[info] ChunkArrayBenchmarks.map        1000  avgt    5   3750.498 ±  25.021  ns/op
[info] ChunkArrayBenchmarks.mapZIO     1000  avgt    5  13725.142 ± 134.258  ns/op
```

`5780`

```
[info] Benchmark                     (size)  Mode  Cnt      Score     Error  Units
[info] ChunkArrayBenchmarks.find       1000  avgt    5      8.615 ±   0.026  ns/op
[info] ChunkArrayBenchmarks.flatMap    1000  avgt    5  32275.816 ± 153.981  ns/op
[info] ChunkArrayBenchmarks.fold       1000  avgt    5   2603.498 ±   6.762  ns/op
[info] ChunkArrayBenchmarks.map        1000  avgt    5   3818.170 ±  12.564  ns/op
[info] ChunkArrayBenchmarks.mapZIO     1000  avgt    5  13713.579 ±  69.950  ns/op
```

Very significant improvement on most operators on mixed chunks. `find` slightly slower. In that benchmark we are constructing a very complex nested chunk and then just finding the second element so I think slightly more work there to create the `ChunkIterator` but still very fast. Most other operators much faster.

`series/2.x`

```
[info] Benchmark                       (size)  Mode  Cnt      Score      Error  Units
[info] MixedChunkBenchmarks.filterZIO    1000  avgt    5  86611.433 ± 4340.034  ns/op
[info] MixedChunkBenchmarks.find         1000  avgt    5     39.952 ±    2.053  ns/op
[info] MixedChunkBenchmarks.flatMap      1000  avgt    5  68252.212 ±  601.825  ns/op
[info] MixedChunkBenchmarks.fold         1000  avgt    5  48187.238 ±  189.955  ns/op
[info] MixedChunkBenchmarks.foldZIO      1000  avgt    5  59545.077 ±   54.357  ns/op
[info] MixedChunkBenchmarks.map          1000  avgt    5  38570.640 ±   62.860  ns/op
[info] MixedChunkBenchmarks.mapZIO       1000  avgt    5  58258.618 ±  371.438  ns/op
```

`5780`

```
[info] Benchmark                       (size)  Mode  Cnt      Score     Error  Units
[info] MixedChunkBenchmarks.filterZIO    1000  avgt    5  58854.532 ± 104.431  ns/op
[info] MixedChunkBenchmarks.find         1000  avgt    5    239.573 ±   1.208  ns/op
[info] MixedChunkBenchmarks.flatMap      1000  avgt    5  40644.183 ± 221.212  ns/op
[info] MixedChunkBenchmarks.fold         1000  avgt    5  11924.802 ±   7.722  ns/op
[info] MixedChunkBenchmarks.foldZIO      1000  avgt    5  30448.681 ± 125.890  ns/op
[info] MixedChunkBenchmarks.map          1000  avgt    5  14633.681 ±  92.732  ns/op
[info] MixedChunkBenchmarks.mapZIO       1000  avgt    5  61151.532 ± 165.656  ns/op
```

PR #5787, opened 2021-10-17
Support ZPool Max Pool Size and Improve Shutdown Robustness

### LINKED ISSUE #5762: ZIO 2.0: Support ZPool max pool size
Currently only min size is supported. In the event that all items in the pool are being used, then we could allocate more items, assuming the maximum has not yet been reached.

Eventually, if left alone for a sufficient amount of time, then a pool whose excess items are not being used should shrink down, but under no circumstances may it shrink to less than the minimum size.

### LINKED ISSUE #5763: ZIO 2.0: Improve shutdown robustness in ZPool
For performance reasons, ZPool uses Queue, which means the size parameter may be inconsistent with the state of the pool. This might lead in rare situations to `shutdown` waiting for a state that will never be reached. The logic should be inspected and modified appropriately to ensure this can't happen.

adamgfraser:
Resolves #5762. Resolves #5763.

### DIFF
--- a/core/shared/src/main/scala/zio/ZPool.scala
+++ b/core/shared/src/main/scala/zio/ZPool.scala
@@ -16,6 +16,8 @@ package zio
  * limitations under the License.
  */
 
+import zio.stacktracer.TracingImplicits.disableAutoTrace
+
 /**
  * A `ZPool[E, A]` is a pool of items of type `A`, each of which may be
  * associated with the acquisition and release of resources.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/ZPool.scala
```diff
@@ -100,94 +117,251 @@ object ZPool {
...
+      ZIO.uninterruptible {
+        state.modify { case State(size, free) =>
+          if (size > range.start && free > 0)
+            (
+              items.take.flatMap { attempted =>
+                attempted.forEach(a => invalidated.update(_ - a)) *>
+                  attempted.finalizer *>
+                  state.update(state => state.copy(size = state.size - 1))
+              },
+              State(size, free - 1)
```
guizmaii, line 229: @adamgfraser Can I ask you why you're decrementing the state `size` this way: 

```scala
... >* state.update(state => state.copy(size = state.size - 1))
```

and not this way:

```scala
State(size - 1, free - 1)
```

So that if `attempted.forEach(a => invalidated.update(_ - a)) *> attempted.finalizer` fails the `size` is not decremented? 
`attempted.forEach(a => invalidated.update(_ - a)) *> attempted.finalizer` seems to be a `UIO` so I'm not sure to understand 🤔

core/shared/src/main/scala/zio/ZPool.scala
```diff
@@ -100,94 +117,251 @@ object ZPool {
...
-            _           <- allocating.update(_ + 1)
-            reservation <- creator.reserve
-            exit        <- restore(reservation.acquire).exit
-            attempted   <- ZIO.succeed(Attempted(exit, reservation.release(Exit.succeed(()))))
-            _           <- free.offer(attempted) *> size.update(_ + 1)
-          } yield attempted).ensuring(allocating.update(_ - 1))
-        }
-      }
+    def excess(implicit trace: ZTraceElement): UIO[Int] =
+      state.get.map { case State(free, size) => size - range.start min free }
```
guizmaii, line 138: @adamgfraser Why do we need to `min` between `range.start` and `free`? 🤔

It seems to me that `excess` could just be: `size - range.start`, no? 🤔

core/shared/src/main/scala/zio/ZPool.scala
adamgfraser: Yes we reduce `free` immediately and then `size` only after the resource has been finalized.

core/shared/src/main/scala/zio/ZPool.scala
adamgfraser: If `free` is smaller, for example because we are in the process of shrinking, then we don't really have that excess.

PR #5812, opened 2021-10-21
Scala Native 0.4.1

sideeffffect:


sideeffffect:
Btw, I've noticed that [Scala Steward doesn't manage zio/zio](https://github.com/zio/zio/pull/5765) anymore. A PR like this one here could have been done automatically instead of requiring actions of us 2 humans.
So I'm wondering, can we have our cake and eat it too? The issues that Steward caused -- could we somehow solve or at least mitigate them, so that we can have automation around dependency versions?

For example, one of the steps that we have already undergone is the rationalization of jobs in CI -- it is now faster and less resource demanding (while even more reliable than before), which was a good thing not only for Steward but also ZIO development in general.

I don't want to make the impression that I'm super insistent on Steward (I'm not the most active zio/zio contributor after all, so I don' bear the cons of having it turned on), but if there's something I can do, I want to offer my help (I'm a great believer in automation) :wink:

adamgfraser:
I guess my perspective is that there generally isn't a need for such frequent updates. Our updates should be binary compatible, most of them are making relatively minor changes, and most of them are not even used in production code but just in documentation or benchmarks. So I would tend to think that going through and updating all the dependencies approximately once a quarter would keep us up to date, be very little work at that frequency, and would keep the noise of dependency updates to a minimum.

There may be cases like Scala Native where understand it is under active development and very happy to accept PRs but seems like that should be the exception rather than the rule.

PR #5814, opened 2021-10-21
Fix zio-test-scalacheck publishing

mijicd:


bbarker:
@mijicd @adamgfraser am I looking in the wrong place for ZIO 1 artifacts?: it seems like this [wasn't published](https://repo1.maven.org/maven2/dev/zio/zio-test-scalacheck_2.13/) for 1.0.12, but if I understand right, [it should be](https://github.com/zio/zio/blob/v1.0.12/test-scalacheck/shared/src/main/scala/zio/test/scalacheck/package.scala)

mijicd:
@bbarker This pull request landed after 1.0.12 was released. While the functionality you pointed out was there, publishing was accidentally disabled, and enabled in this pull request later on.

PR #5815, opened 2021-10-22
Rename ZLayer to ZServiceBuilder

### LINKED ISSUE #5805: [Suggestion] rename ZLayer to ZDependency
When I introduce ZIO to my colleagues, who are new to both Scala and ZIO, they find it hard to understand the `ZLayer` concept and the `R` parameter in ZIO, even after they understand `IO[E, A]`.

One feedback I get from them is that it may be better to rename `ZLayer` to `ZDependency`, as "dependency" is a more familiar concept. They can recall their earlier experience using Spring and other dependency injection frameworks. And ZDependency would be better aligned with the `inject` methods(as in "inject dependencies").

An alternative name to ZDependency, as suggested by @jdegoes is `ZConstructor`.

This issue is to discuss what name we can rename ZLayer to, and hope we can make it for the ZIO 2.0.
Any other thoughts?

adamgfraser:
Resolves #5805.

The `ZLayer` name really only makes sense in terms of the "onion architecture" and representing different "layers" of the onion (e.g. the `Github` service might translate from the lower level domain of HTTP to the higher level domain of the Github API). While this is a very insightful way to view program architecture, it requires you to know this particular concept which is an obstacle to usability and teachability because many people may not.

This PR explores instead naming it `ZDeps` for dependencies, a bundle of dependencies that you provide to your application.


alphaho:
@guersam , how about a name like `ZBuilder`? A builder requires some setup/configuration, and can build an object out of it. That may capture the both input and output side of ZLayer.

regiskuckaertz:
Throwing my hat in the game with `ZRequire` for the duality with `provide` 😁

ghostdogpr:
> Throwing my hat in the game with `ZRequire` for the duality with `provide` 😁

But `provide` doesn't actually require a `ZLayer`, `provideLayer` does 🤪

regiskuckaertz:
True 😂 I think the profunctor property is not exploited anymore,  iirc @adamgfraser told me this wasn't a pattern that was encouraged so much anymore, and I find it has been a challenge to teach layers and the `R` parameter as two separate things. So perhaps an opportunity to make layers or whatever their name more natural to express.

guersam:
Frankly, I often get confused of when to use `R` or `ZLayer` too.

I think we need to gather the best practices for this first, then upon a deeper understanding we may find better name for `ZLayer` or reach a consensus to keep it as it is.

jdegoes:
After a lot of thought, the I've grown to like `ZServiceBuilder`.

Why? Because the terminology seems to be consistent:

 - You access services from the environment with `ZIO.service`
 - You create service definitions with interfaces / traits
 - You implement services with classes
 - You build services from their dependencies with service builders
 - Service builders describe how to build services from their dependencies

As part of this, we would change "Module Pattern" to "Service Pattern", thus consolidating a lot of different terms under the "service" umbrella. /cc @khajavi

jdegoes:
@guersam I agree there is confusion here. @khajavi Let's meet to discuss best practices around R versus Layer so we can thoroughly document it on zio.dev.

khajavi:
@jdegoes 
I definitely agree with `ZServiceBuilder` and `Service Pattern` as we have discussed before. 

In addition, regarding this term consolidation, I propose renaming the `Has` data type to the `Service` or `Srv`. Therefore, the `ZServiceBuilder[Srv[A] with Srv[B], Throwable, Srv[C]]` makes more sense than the _has_ terminology for newcomers.

This would be consistent with the `provideService` API which indicates that we should provide it an `Srv[A]` instead of raw `A` data type.

adamgfraser:
I like it too. Will update accordingly.

adamgfraser:
@khajavi I like the direction you are going in renaming `Has`. How would you propose describing the type inside the `Has`. For example if we have `Has[Console]` I would normally describe `Console` as a service and `Has` as a tool that we use to combine services. If we use `Service[Console]` that seems a bit more unclear. Potentially we could say that `Service` denotes that `Console` is a service?

adamgfraser:

jdegoes (approved):

jdegoes:
@adamgfraser One minor question:

Should it be `provideService` (singular), or `provideServices` (plural)? I would have slightly leaned toward the latter, but I saw you went with the former. What are your thoughts?

jdegoes:
The `Has` has existing usage inside scala code bases, e.g. `HasFoo` is usually a trait that exposes a `foo`, for purposes of `with`ing all the containers together. I agree with @adamgfraser that `Service[Foo]` would be a bit confusing because it equivocates on what is the service.

adamgfraser:
@jdegoes I wondered about that too. I went with `provideService` because you can potentially provide one or more services, `ZServiceBuilder` is itself singular, and it provides a consistent namespace for everything "service related".

However, thinking about it more I like making it plural because it creates a distinction between operators like `ZIO#updateService` (update a single service) and `ZIO#provideServices` (provide zero or more services if you count the fact that a `ServiceBuilder` can be built purely for its effects).

Will change.

jdegoes (approved):

octavz:
Late to the party, I didn't know why I hated so much reading about this change. Thinking about it I figured out that apart from sounding like a a "Java something" that I ran from and came to `ZIO`, it is very long. Given the invariance of `Has`, `ZLayer` is the most typed thing when using `ZIO`, apart from `ZIO`. 

You took something not great to type and made it worse, `ZServiceBuilder[Any, Throwable,  Service`] , that's a mouthful, also is nothing like a builder so this will not help anyone coming from Java expecting a `Builder` to be a builder.

I don't know if is better for new people coming to `ZIO`, I helped many to adopt `ZIO` and the name `Layer` was the least of their problems, but for me and my codebase will be a hassle to use the new name and adopt a new vocabulary.

adamgfraser:
@octavz Sorry to hear that!

We're always trying to strike the right balance for all users. There are definitely some people who found the concept of layers confusing and it really did only have a meaning within the concept of an "onion pattern" architecture which is great but not everyone knows.

As is often the case with naming I don't think there was a perfect solution here. As you can see we went through a variety of alternatives that people proposed and on balance this seemed to be the best alternative. It does exactly what it says, which is build the services your application needs. You are right it is longer but we think conceptual clarity wins out over terseness, especially with features like autocomplete in modern IDEs.

One thing that did jump out at me is your statement that this is the second most frequent data type you type. That is very surprising to me. Use cases are definitely going to vary but I find it is typically one of the ones I write least frequently.

Anyway, definitely appreciate your perspective and hope that over time with other features in ZIO 2.0 like automatic construction of the dependency graph and a more structured way of defining services we can win you over and make building your application's dependencies as easy as possible.

octavz:
@adamgfraser 

I am not complaining I will just adapt, but I am also surprised by  your "surprise" :), 

Unless we are using this completely wrong:

You have to type `ZLayer` every time you make a layer (twice every time: once for the type since it will always infer with the sub type `LiveXXX` and once for `ZLayer.succeed` or `fromfunction` or friends). 

At first sight it might not be a lot but think about the tests, every test will stub and every stub has to become a layer, every test has at least one stub which makes "ZLayer" or variants basically the most used thing in your code.

Also if we follow conventions any `vals `that now contain layer would have to contain `serviceBuilder`.


For a small module:

```
>  ag --stats --scala 'ULayer|ZLayer|URLayer'

550 matches
44 files contained matches
```
compare to ZIO usages

```
ag --stats --scala -c 'RIO|Task|ZIO'
677 matches
50 files contained matches
```
Also IDEs for Scala are slower than typing most of the time so they are not big help there, IntelliJ hardly works with ZIO, Metals "almost" works but more times it doesn't than it does.

joroKr21:
Happy this got reverted but for posterity I wanted to share another opinion. `ZLayer` is not a simple builder to me because it caches services with the same type. Maybe `ZDependencyGraph` would make more sense because it really represents a graph of your dependencies - every service type is its own node with edges between services representing a dependency. But still I like `ZLayer` more anyway - two words are too much for such a core concept.

Issue #5785, opened 2021-10-16
Include the JVM stack up to `Runtime.unsafeRun*` in the fiber trace


dkarlinsky:
**The problem:**
Currently the fiber traces collected and attached to errors only cover the execution inside ZIO universe.
Which is fine if your whole app (or most of it) is ZIO. 
But when interoperating with non-ZIO code we often have flows where non-ZIO code is invoking ZIO effects via `Runtime.unsafeRun*` and if the effect fails, or an error is logged by it, the trace doesn't contain any info on where `Runtime.unsafeRun*` was called from in the non-ZIO part of the application.
(For example a future based GRPC request handler calling a library that is implemented in ZIO).

**The proposal:**
When `Runtime.unsafeRunWith` is creating the main `FiberContext` it could capture the current thread's stack, convert it into a `Ztrace` and pass it to the `FiberContext` as `parentTrace`.
Then in the fiber trace it will appear that the main fiber was spawned by this other fiber, whose trace is actually the captured stack.

Example (parent trace appears as `Fiber:Id(0,0) execution trace:`)

```
Exception in thread "main" java.lang.RuntimeException: kaboom
	at zio.Somewhere$.$anonfun$kaboom$5(RuntimeSpec.scala:13)
	at zio.ZIO$.$anonfun$fail$1(ZIO.scala:2864)
	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:407)
	at zio.internal.FiberContext.$anonfun$evaluateLater$1(FiberContext.scala:776)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
	Suppressed: zio.Cause$FiberTrace: Fiber failed.
A checked error was not handled.
java.lang.RuntimeException: kaboom

Fiber:Id(1634380884543,0) was supposed to continue to:
  a future continuation at zio.Somewhere$.kaboom(RuntimeSpec.scala:11)

Fiber:Id(1634380884543,0) execution trace:
  at zio.Somewhere$.kaboom(RuntimeSpec.scala:13)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:160)
  at zio.internal.FiberContext$InterruptExit$.apply(FiberContext.scala:153)
[...]

PR #5818, opened 2021-10-22
Include the JVM stack up to Runtime.unsafeRun* in the fiber trace

dkarlinsky:
Issue #5785

### DIFF
--- a/core-tests/jvm/src/test/scala/zio/RuntimeSpec.scala
+++ b/core-tests/jvm/src/test/scala/zio/RuntimeSpec.scala
@@ -8,12 +8,12 @@ import zio.test._
 import java.io.{PrintWriter, StringWriter}
 
 object RuntimeSpec extends ZIOBaseSpec {
-  private val runtime = Runtime.default
+  private val captureUnsafeRunStackRuntime = Runtime.default.withTracingConfig(_.withCaptureUnsafeRunStack(true))
   override def spec: ZSpec[Environment, Failure] = suite("RuntimeSpec")(
-    suite("By default include stack trace up to unsafeRun* in Fiber Trace")(
+    suite("When captureUnsafeRunStack enabled, trace up to unsafeRun* in Fiber Trace")(
       test("in unsafeRunTask") {
         for {
-          res <- attemptBlocking(CallSite.failingUnsafeRunTask(runtime)).exit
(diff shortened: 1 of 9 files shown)


adamgfraser (changes requested):
So I think we want to get this out of the runtime configuration.

If we implement:

```scala
object ZIO {
  /** Captures the current thread's stack trace as a `ZTrace`. */
  val stackTrace: UIO[ZTrace] =
    ???
}
```

Then we can add the stack trace to an existing `Cause` used `Cause.traced`. This will allow us to implement `unsafeRunWithStackTrace` as:

```scala
trait Runtime[+R] {
  def unsafeRunWithStackTrace[E, A](zio: => ZIO[R, E, A]): A =
    unsafeRun(ZIO.stackTrace.flatMap(trace => zio.mapErrorCause(_.traced(trace))))
}
```

dkarlinsky:
@adamgfraser I think this is basically what @jdegoes [suggested](https://github.com/zio/zio/issues/5785#issuecomment-951435614) on the issue. 
I responded to his comment [here](https://github.com/zio/zio/issues/5785#issuecomment-951618071), but 
the main issue I see is that, if we want to log exceptions inside our effect (which in code I work on we do alot), the captured stacktrace won't be in the `Cause`.
This as opposed to passing the `ZTace` of the current stack to `new FiberContext` as the parent trace. 

I  could have  the separate `unsafeRun*WithStackTrace` methods capture the stack and propagate it to the private `unsafeRunWith` that creates the initial `FiberContext`. 
But I count seven public `unsafeRun*` methods. Do we want to add seven `*withStackTrace` variants?

dkarlinsky:
I now see that `parentTrace` is gone from the `FiberContext` constructor...
Maybe we could add `runtimeStackTrace: Chunk[ZTraceElement]` parameter, which would be appended to the trace generated in `unsafeCaptureTrace`?

jdegoes:
@dkarlinsky Can you open up a ticket where we can discuss how to do this in ZIO 2.0, in a way that doesn't degrade performance for those who aren't interested in the feature? Thank you!

dkarlinsky:
@jdegoes we have https://github.com/zio/zio/issues/5785 
Or do you prefer a fresh issue, specific to ZIO 2.0?

PR #5838, opened 2021-10-27
Add failsWithA and diesWithA to test assertions

droptheplot:
Like [throwsA](https://github.com/zio/zio/blob/7434c5bd51b058bfeea78ef09dfdc363eb2839c5/test/shared/src/main/scala/zio/test/Assertion.scala#L821) but for [fails](https://github.com/zio/zio/blob/7434c5bd51b058bfeea78ef09dfdc363eb2839c5/test/shared/src/main/scala/zio/test/Assertion.scala#L255) and [dies](https://github.com/zio/zio/blob/7434c5bd51b058bfeea78ef09dfdc363eb2839c5/test/shared/src/main/scala/zio/test/Assertion.scala#L201)

https://github.com/search?l=Scala&q=fails+isSubtype+anything&type=Code


adamgfraser (changes requested):
I would stick with the `A` syntax here since `With` normally indicates a variant that takes an additional function as a parameter.

adamgfraser (dismissed):

adamgfraser (approved):

PR #5872, opened 2021-10-29
Fix race condition in fromOutputStreamWriter

joroKr21:
Because the input and output streams are piped together,
when the output stream is closed, the input stream completes.
Due to the semantics of `drainFork` that would sometimes cause
the output stream writer to be interrupted before completing
the error channel promise. But in fact we don't need a separate
error channel, we just need to make sure the input stream reader
waits for the output stream writer to complete.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/ZStreamPlatformSpecificSpec.scala
@@ -2,9 +2,7 @@ package zio.stream
 
 import zio._
 import zio.blocking.{Blocking, effectBlockingIO}
-import zio.duration._
 import zio.test.Assertion._
-import zio.test.TestAspect._
 import zio.test._
 
 import java.io.{FileNotFoundException, FileReader, IOException, OutputStream, Reader}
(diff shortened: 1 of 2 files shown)


adamgfraser:
@joroKr21 Yes! Great to get to the bottom of this. Looks like there is still some flakiness on one of these tests but excited to merge when ready!

joroKr21:
Hmm I really thought I had it this time. I ran it a lot locally and this didn't happen. Does the timeout always mean there is a deadlock or could it be that it's just too slow? I.e. is there a separate timeout per iteration of `nonFlaky` or is it for all iterations combined?

adamgfraser:
It is a combined timeout for all the iterations of `nonFlaky` so that is a possibility, though the timeout is 2 minutes right now.

joroKr21:
It's 1 minute in this test which means 60ms per iteration for `nonFlaky(1000)`

joroKr21:
I think on CI especially it can happen to get a slow executor. Should we go ahead and merge this? If you agree I will remove the `nonFlaky` annotations.

joroKr21:
> If you agree I will remove the nonFlaky annotations.

@adamgfraser I've done this now - we stress tested that implementation in our application and it works well.

PR #5876, opened 2021-10-29
#5823 - Add ConcurrentSet

### LINKED ISSUE #5823: Add ConcurrentSet
A wrapper over `java.util.concurrent.ConcurrentHashMap.KeySetView[K,​ Boolean]`.

```scala
trait ConcurrentSet[A]:
  def add(x: A): UIO[Unit]
  def addAll(xs: Iterable[A]): UIO[Unit]
  def clear: UIO[Unit]
  def contains(x: A): UIO[Boolean]
  def containsAll(xs: Iterable[A]): UIO[Boolean]
  ...

object ConcurrentSet:
  def make[A](xs: A*): UIO[ConcurrentSet[A]]
  def make[A](size: Int): UIO[ConcurrentSet[A]]
```

aleksandarskrbic:
Resolves #5823


core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
```diff
@@ -0,0 +1,42 @@
+package zio.concurrent
+
+import zio.UIO
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.jdk.CollectionConverters._
+
+final class ConcurrentSet[A] private (underlying: ConcurrentHashMap.KeySetView[A, java.lang.Boolean]) {
```
regiskuckaertz, line 8: Do the data structures exist in scala-js and native?

regiskuckaertz (commented):
A small test spec would be most welcome

core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
regiskuckaertz: Yes. https://github.com/scala-js/scala-js/blob/master/javalib/src/main/scala/java/util/concurrent/ConcurrentHashMap.scala

regiskuckaertz (approved):
Great work!

regiskuckaertz:
aargh, 2.11 comes biting us in the 🍑

regiskuckaertz (approved):
🎉  Brilliant

regiskuckaertz:
I think we should review `addAll` and `containsAll` in a follow-up, they currently are not atomic. The type of the underlying method is not great, but we may go by with creating our own version of `Collection` that can only read data.

PR #5881, opened 2021-10-30
Refactor FiberId

### LINKED ISSUE #5871: ZIO 2.0: Redesign FiberId to be more efficient and more accurate
Currently, FiberID is a `case class` that contains a timestamp and a sequence number. There is no way to represent composite fiber ids, or the absence of a fiber id, and various hacks are used to work around both issues.

We should instead refactor it to the following:

```scala
sealed trait FiberId
object FiberId {
  case object None extends FiberId 
  final case class Runtime(id: Int, startTime: Int) extends FiberId 
  final case class Both(left: FiberId, right: FiberId) extends FiberId
}
```

where `Runtime#id` is a monotonically increasing sequence number, and `startTime` is a measure of the starting time of the fiber that will fit into a 32 bit integer.

The reason for making `id` and `startTime` both `Int` instead of `Long` is that it will allow us to pack both of these values into a single `Long`, which can be stored in the `Fiber`, with the `FiberId` computed only when necessary (e.g. `def fiberId = ...`).

**Note**: _This ticket is only to redesign `FiberId` and update the code, not to actually utilize a more efficient way of storing the fiber id inside the fiber._

adamgfraser:
Resolves #5871.

### DIFF
--- a/core/shared/src/main/scala/zio/Fiber.scala
+++ b/core/shared/src/main/scala/zio/Fiber.scala
@@ -422,7 +422,7 @@ object Fiber extends FiberPlatformSpecific {
     /**
      * The identity of the fiber.
      */
-    def id: FiberId
+    def id: FiberId.Runtime
 
     def scope: ZScope[Exit[E, A]]
 
@@ -440,7 +440,7 @@ object Fiber extends FiberPlatformSpecific {
   private[zio] object Runtime {
 
     implicit def fiberOrdering[E, A]: Ordering[Fiber.Runtime[E, A]] =
-      Ordering.by[Fiber.Runtime[E, A], (Long, Long)](fiber => (fiber.id.startTimeMillis, fiber.id.seqNumber))
+      Ordering.by[Fiber.Runtime[E, A], (Long, Long)](fiber => (fiber.id.startTime, fiber.id.id))
 
     abstract class Internal[+E, +A] extends Runtime[E, A]
   }
@@ -497,7 +497,7 @@ object Fiber extends FiberPlatformSpecific {
 
   sealed abstract class Dump extends Serializable { self =>
 
-    def fiberId: FiberId
+    def fiberId: FiberId.Runtime
 
     def fiberName: Option[String]
 
@@ -521,13 +521,13 @@ object Fiber extends FiberPlatformSpecific {
 
   object Dump {
     def apply(
-      fiberId0: FiberId,
+      fiberId0: FiberId.Runtime,
       fiberName0: Option[String],
       status0: Status,
       trace0: Option[ZTrace]
     ): Dump =
       new Dump {
-        def fiberId: FiberId          = fiberId0
+        def fiberId: FiberId.Runtime  = fiberId0
         def fiberName: Option[String] = fiberName0
         def status: Status            = status0
         def trace: Option[ZTrace]     = trace0
@@ -824,11 +824,11 @@ object Fiber extends FiberPlatformSpecific {
   def unsafeCurrentFiber(): Option[Fiber[Any, Any]] =
     Option(_currentFiber.get)
 
-  private[zio] def newFiberId(): FiberId =
-    FiberId(java.lang.System.currentTimeMillis(), _fiberCounter.getAndIncrement())
+  private[zio] def newFiberId(): FiberId.Runtime =
+    FiberId.Runtime((java.lang.System.currentTimeMillis / 1000).toInt, _fiberCounter.getAndIncrement())
 
   private[zio] val _currentFiber: ThreadLocal[internal.FiberContext[_, _]] =
     new ThreadLocal[internal.FiberContext[_, _]]()
 
-  private[zio] val _fiberCounter = new java.util.concurrent.atomic.AtomicLong(0L)
+  private[zio] val _fiberCounter = new java.util.concurrent.atomic.AtomicInteger(0)
 }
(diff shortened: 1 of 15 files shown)


core/shared/src/main/scala/zio/FiberId.scala
```diff
@@ -17,17 +17,30 @@
...
  * The identity of a Fiber, described by the time it began life, and a
  * monotonically increasing sequence number generated from an atomic counter.
  */
-final case class FiberId(startTimeMillis: Long, seqNumber: Long) extends Serializable
+sealed trait FiberId extends Serializable { self =>
 
-object FiberId {
+  def ids: List[Int] = {
+
+    def loop(stack: List[FiberId], result: List[Int]): List[Int] =
```
regiskuckaertz, line 30: tailrec

core/shared/src/main/scala/zio/Fiber.scala
```diff
@@ -824,11 +824,11 @@ object Fiber extends FiberPlatformSpecific {
   def unsafeCurrentFiber(): Option[Fiber[Any, Any]] =
     Option(_currentFiber.get)
 
-  private[zio] def newFiberId(): FiberId =
-    FiberId(java.lang.System.currentTimeMillis(), _fiberCounter.getAndIncrement())
+  private[zio] def newFiberId(): FiberId.Runtime =
+    FiberId.Runtime((java.lang.System.currentTimeMillis / 1000).toInt, _fiberCounter.getAndIncrement())
```
regiskuckaertz, line 828: Don't we need more precision here?

core/shared/src/main/scala/zio/FiberId.scala
adamgfraser: Thanks!

core/shared/src/main/scala/zio/Fiber.scala
regiskuckaertz: BTW wouldn't it make more sense to use jvm time instead of wall-clock time here?

regiskuckaertz (approved):

core/shared/src/main/scala/zio/Fiber.scala
jdegoes: Interesting post [here](http://pzemtsov.github.io/2017/07/23/the-slow-currenttimemillis.html) on the performance of `currentTimeMillis`. 

We'd like to use the fastest method of measuring some notion of time, rather than the most accurate one. 

This data will only be used by metrics (as well as to make an `AtomicInteger`'s wraparound issues irrelevant).

jdegoes: From what I gather:

1. `nanoTime` is extremely slow (expected)
2. `currentTimeMillis` is fast on Windows, and slow (but not extremely slow) on Linux

An OpenJDK issue was closed as "Won't fix" regarding (2).

I think we can use `currentTimeMillis` here and fix it later.

"second time", which is achieved by `/ 1000` is more than sufficient because in order for that to lead to uniqueness issues via `AtomicInteger` sequence number, we would have to be generating 2,147,483,647 fibers per second, which of course can't happen.

core/shared/src/main/scala/zio/FiberId.scala
```diff
@@ -18,16 +18,31 @@ package zio
...
 
-  /**
-   * A sentinel value to indicate a fiber without identity.
-   */
-  final val None: FiberId = FiberId(0L, 0L)
+    loop(List(self), List.empty)
+  }
+}
+
+object FiberId {
```
jdegoes, line 44: I would add:

```suggestion
object FiberId {
  def apply(id: Int, startTime: Int): FiberId.Runtime = Runtime(id, startTime)
```

core/shared/src/main/scala/zio/FiberId.scala
```diff
@@ -18,16 +18,31 @@ package zio
...
+  case object None                                          extends FiberId
+  final case class Runtime(id: Int, startTime: Int)         extends FiberId
```
jdegoes, line 46:
```suggestion
  final case class Runtime(id: Int, startTimeSeconds: Int)         extends FiberId
```

jdegoes:
@adamgfraser Looks really good! One more suggestion:

1. Go through `FiberContext.scala` and search for the 2 instances of `interruptors.headOption`.
2. Replace both of these by `FiberId.combineAll`, where `combineAll` will just combine all the fibers.

(Currently the `interruptAs` is being fed an arbitrary interruptor from among the set, which is lossy.)

core/shared/src/main/scala/zio/FiberId.scala
jdegoes: With this constructor, you can eliminate some changes to the code.

adamgfraser:
@jdegoes Will do.

jdegoes:
On precision: If stored in an `Int`, `currentTimeMillis()` wraps around every ~24 days. That's an argument to reduce precision. Second precision will mean fiber lifetime metrics would only be useful for outliers (fibers that live > 1 second), but millisecond precision is too much since that will screw up fiber lifetime metrics due to wraparound.

jdegoes (approved):

core/shared/src/main/scala/zio/Fiber.scala
987Nabil: Interesting discussion. Thought neither of you realised, that `FiberId.Runtime` changed the order of id and timestamp and this impl. is therefore a bug.
I already fixed it in [this pr](https://github.com/zio/zio/pull/5916) so don't worry 😃

PR #5885, opened 2021-10-30
Implemented `ZSink#fromFile`

### LINKED ISSUE #5868: ZIO Streams 2.0: JVM-specific ZSink#fromFile
The JVM-specific `ZSink` constructor, `fromFile` is not yet defined for the new ZIO Streams 2.0 `ZSink`, in `zio.stream.experimental.ZSinkPlatformSpecificConstructors`.

Define it in the same way as it was for ZIO Stream 1.0 and provide some tests.

ithinkicancode:
Resolve #5868


streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -6,8 +6,54 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
...
+    val writer = ZSink.unwrapManaged {
+      managedChannel.map { chan =>
+        ZSink.foreachChunk[Any, Throwable, Byte] { byteChunk =>
+          ZIO.attemptBlockingInterrupt {
+            chan.write(ByteBuffer.wrap(byteChunk.toArray))
+          }
+        }
+      }
+    }
+    writer &> ZSink.count
```
adamgfraser, line 54: Do we really need to run two channels in parallel here? All the second channel is doing is keeping track of the count of bytes written. Can't we keep track of that in the `writer` channel and write it as our done value when we are done? That will be much more performant because it will only use one channel instead of two and avoid the overhead of managing the concurrency between them.

ithinkicancode: Done.

adamgfraser (approved):

PR #5886, opened 2021-10-30
Add ZSink.fromQueue, ZSink.fromHub,  ZSink.fromQueueWithShutdown and ZSink.fromHubWithShutdown to Streams 2.0

kmatasfp:


kmatasfp:
Probably want to revisit this sometime later because the type inference is not the best, for example following would not compile
```
test("should enqueues all elements") {
  for {
    queue          <- ZQueue.unbounded[Int]
    stream          = ZStream.fromZIO(ZIO(List(1, 2, 3))).flattenIterables
    sink            = ZSink.fromQueue(queue)
    _              <- stream.run(sink)
    enqueuedValues <- queue.takeAll
  } yield {
    assert(enqueuedValues)(hasSameElementsDistinct(List(1, 2, 3)))
  }
}
```

streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
```diff
@@ -530,6 +530,92 @@ object ZSinkSpec extends ZIOBaseSpec {
...
+          } yield {
+            assert(enqueuedValues)(hasSameElementsDistinct(List(1, 2, 3)))
+          }
+
+        }
+      ),
+      suite("fromQueueWithShutdown")(
+        test("should enqueue all elements and shutsdown queue") {
+
+          def createQueueSpy[A](q: Queue[A]) = new ZQueue[Any, Any, Nothing, Nothing, A, A] {
```
vigoo, line 549: Is this necessary? Can't you just create a queue and check if it was shut down with it's `isShutdown` method?

vigoo:
If https://github.com/zio/zio/pull/5913 is merged you can change the signatures

streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
kmatasfp: I generally like to test the whole behaviour from users perspective and thing they care about it that all elements were enqueued and then the queue was shut down. Sadly was not able to find a more elegant/reliable way to get all the enqueued elements, after shutdown occurs one cannot dequeue anymore. Would have wanted to write a similar test for Hub as well. If you think it is a overkill in this case I can scrap it and convert it to similar to  `fromHubWithShutdown`, but personally I think more testing the better :)

kmatasfp:
Also I think even after #5913 is merged this wouldn't compile:
```
test("should enqueues all elements") {
  for {
    queue          <- ZQueue.unbounded[Int]
    stream          = ZStream.fromZIO(ZIO(List(1, 2, 3))).flattenIterables
    sink            = ZSink.fromQueue(queue)
    _              <- stream.run(sink)
    enqueuedValues <- queue.takeAll
  } yield {
    assert(enqueuedValues)(hasSameElementsDistinct(List(1, 2, 3)))
  }
}
```
I. need to do something like this, which imho is not the best from dev ergonomics standpoint
```
for {
              queue                                                 <- ZQueue.unbounded[Int]
              stream                                                 = ZStream.fromZIO(ZIO(List(1, 2, 3))).flattenIterables
              sink: ZSink[Any, Throwable, Int, Throwable, Int, Unit] = ZSink.fromQueue(queue)
              _                                                     <- stream.run(sink)
              enqueuedValues                                        <- queue.takeAll
            } yield {
              assert(enqueuedValues)(hasSameElementsDistinct(List(1, 2, 3)))
            }
```

streams-tests/shared/src/test/scala/zio/stream/experimental/ZSinkSpec.scala
vigoo: Ok I understand now. Let's keep it :)

vigoo (approved):

PR #5893, opened 2021-10-30
Remove Scalafix artifacts

mijicd:
Follow-up to #5766


adamgfraser (commented):
So I think there is a distinction between using Scalafix as a tool in ZIO to run Scalafix rules on our own code and using it to create Scalafix rules that we are going to provide to users for migrations. I think we want to keep this because it is what enables the migrations. In ZIO 1.0 the only one we had was this one for the curried assert which is definitely a little old at this point but in the ZIO 2.0 branch we have a ton of content here for the ZIO 2.0 migration Scalafix rule that is going to be really valuable for users.

mijicd:
@adamgfraser Reintroduced `scalafix`. At the same time, I reintroduced `prepare` alias, and removed  scalafix config file.

build.sbt
```diff
@@ -24,6 +24,7 @@ inThisBuild(
 )
 
 addCommandAlias("build", "; prepare; testJVM")
+addCommandAlias("prepare", "fix")
```
adamgfraser, line 27: Why do we need this alias?

mijicd: Immediate need is to fix the `build` alias as it's broken at the moment. Personally, I wouldn't mind fixing it by just using `fmt` directly, but @sideeffffect suggested that we should keep this for extensibility.

P.S. Now that I'm responding to you I realize there's an error in it (used `fix` instead of `fmt`).

build.sbt
```diff
@@ -24,6 +24,7 @@ inThisBuild(
 )
 
 addCommandAlias("build", "; prepare; testJVM")
+addCommandAlias("prepare", "fmt")
```
adamgfraser, line 27: Should we just have `fmt` so there is only one way to do the same thing?

mijicd: cc @sideeffffect :). As I mentioned before, I'm fine with either as long as all aliases are working.

build.sbt
adamgfraser: I would suggest we just use `fmt` for everything.

build.sbt
sideeffffect: Personally, I got used to just using `fix` to "fix" all the issues that can be automatically fixed, regardless of whether we use Scalafix, Scalafmt, or both.
Thus I've never used `prepare`.
But I don't have a strong opinion on this.

regiskuckaertz:
Would it hurt to keep the configuration? I usually write code in anger and let scalafix clean up my mess (usually spread over several files). The cli is way faster.

build.sbt
mijicd: Ok, in that case I'm removing the `prepare` as well, and I'll replace it directly in `build`.

adamgfraser:
@regiskuckaertz Well we don't use Scalafix in the main project. So you use `fmt` to format your code but the only thing Scalafix was doing for you was cleaning up your imports and it just added a lot of time to run it.

adamgfraser (approved):

Issue #5889, opened 2021-10-30
Add scala collection-like methods to ConcurrentMap


regiskuckaertz:
After #5808 is done, we can start adding operators that will bring the data structure closer to what a collection type looks like in Scala-land. Things like:

- ++, +
- collect, collectFirst, filter
- map, flatMap, fold, groupBy, grouped
- exists, forall
- etc

oliveiosello:
I'll give this a shot!

mijicd:
@oliveiosello Base implementation is in #5898. Hopefully the CI gods will allow me to merge it soon, but anyhow, you should be able to use it :).

oliveiosello:
For similar reasons listed in the comments left on #5890, I've omitted collect, map, and flatMap. Additionally, I've broken the logic of filter into two methods, retainIf and removeIf, to disambiguate from the immutable methods and to be more consistent with `TMap` and `TSet`. Lastly, instead of ++, I'll add putAll shortly.

regiskuckaertz:
@oliveiosello Thanks a lot for picking this up 😍 I think your arguments are valid, let me check your code.

PR #5898, opened 2021-10-30
Implement ConcurrentMap

### LINKED ISSUE #5808: Add ConcurrentMap
> A Map providing thread safety and atomicity guarantees.

A wapper over `java.util.concurrent.ConcurrentHashMap`:

```
package zio.concurrent

trait ConcurrentMap[K, V]:
  def compute​(key: K, remap: (K, V) => V): UIO[Option[V]]
  def computeIfAbsent​(key: K, map: K => V): UIO[V]
  def computeIfPresent​(key: K, remap: (K, V) => V): UIO[Option[V]]
  def replace​(key: K, value: => V): UIO[Option[V]]
  def replace​(key: K, oldValue: V, newValue: => V): UIO[Boolean]
  // ... and all other underlying methods

object ConcurrentMap:
  def make[K, V](kvs: (K, V)*): UIO[ConcurrentMap[K, V]]
  def make[K, V](size: Int): UIO[ConcurrentMap[K, V]]
```

### What about `TConcurrentMap`?

There is a question of whether a wrapper over the STM flavour would compete with the above in terms of performance... unlikely, but perhaps there are some optimisations that could be made to the code. One thing though, is that Java `Function`s and `Consumer`s are not functionals; they can and often have side-effects. This is really discouraged in ZIO, for obvious reasons, but what if internally we exposed the runtime? Then we could provide methods accepting at least transactional side-effects:

```scala
  def computeSTM​[R, E](key: K, remap: (K, V) => ZSTM[R, E, V]): ZIO[R, E, Option[V]]
  def computeIfAbsentSTM​[R, E](key: K, map: K => ZSTM[R, E, V]): ZIO[R, E, V]
  def computeIfPresentSTM​[R, E](key: K, remap: (K, V) => ZSTM[R, E, V]): ZIO[R, E, Option[V]]
  def replaceSTM[R, E]​(key: K, value: ZSTM[R, E, V]): ZIO[R, E, Option[V]]
  def replaceSTM[R, E]​(key: K, oldValue: V, newValue: ZSTM[R, E, V]): ZIO[R, E, Boolean]
```

mijicd:
Resolves #5808.

### Summary

- Verified that `java.util.concurrent.ConcurrentHashMap` is the most suited for the "backend" job.
- Exposed the "core" functionality from `scala.collection.concurrent.Map` "lifted" in ZIO. Other combinators will be added as part of #5889.

### Issue

What to do with `compute*` combinators? They're coming from Java world, but I can't think of a way to provide them in a "pure" way, i.e. even if we change their signatures to accept ZIO values, we'd have to `unsafeRun` at one point, which is unfortunate.

### DIFF
--- a/core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
+++ b/core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
@@ -1,7 +1,8 @@
 package zio.concurrent
 
-import java.util.concurrent.ConcurrentHashMap
 import zio.{Chunk, ChunkBuilder, UIO}
+
+import java.util.concurrent.ConcurrentHashMap
 
 /**
  * Wrapper over [[java.util.concurrent.ConcurrentHashMap]].
--- a/core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
+++ b/core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
@@ -4,7 +4,12 @@ import java.util.concurrent.ConcurrentHashMap
 import zio.UIO
 
 final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashMap[K, V]) extends AnyVal {
-  def get(key: K): UIO[Option[V]]                             = ???
+  /**
+   * Retrieves the value associated with the given key.
+   */
+  def get(key: K): UIO[Option[V]] =
+    UIO(Option(underlying.get(key)))
+
   def putIfAbsent(key: K, value: V): UIO[Option[V]]           = ???
   def remove(key: K): UIO[Option[V]]                          = ???
   def replace(key: K, value: V): UIO[Option[V]]               = ???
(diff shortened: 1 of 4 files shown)


adamgfraser:
I think the `compute` operators can only be exposed for pure functions, similar to how `update` is only supported on `Ref` for pure functions. If you want to effectually update the map you need a more powerful data structure.

mijicd:
After discussing with @adamgfraser, we come to the conclusion that `compute*` should be dealt with in the same way how we're doing it for all other combinators that accept functions, i.e. "please use pure functions otherwise you own your fiber deaths" :). I'll update the PR.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -0,0 +1,111 @@
...
+    UIO {
+      val builder = ChunkBuilder.make[(K, V)]()
+
+      val it = underlying.entrySet().iterator()
+      while (it.hasNext()) {
+        val entry = it.next()
+        builder += entry.getKey() -> entry.getValue()
+      }
+
+      builder.result()
```
regiskuckaertz, line 69:
```suggestion
      Chunk.fromIterator(underlying.entrySet().iterator())
```

mijicd: It's java iterator here :).

regiskuckaertz: Ha yeah, it looks like sth that would be useful as a package-private `Chunk` constructor.

regiskuckaertz (approved):

regiskuckaertz (dismissed):

regiskuckaertz (approved):

PR #5899, opened 2021-10-30
Add JVM-specific ZSink.outputFromStream and ZSink.outputFromStreamManaged in experimental

### LINKED ISSUE #5867: ZIO Streams 2.0: JVM-specific ZSink#fromOutputStream and fromOutputStreamManaged
The new ZIO Streams 2.0 sink does not have `fromOutputStream` and `fromOutputStreamManaged` yet. 
These are JVM-specific constructors and should be defined in `zio.stream.experimental.ZSinkPlatformSpecificConstructors`.

Some tests should also be added for them in the JVM-specific test suite.

grouzen:
Resolves #5867 

**Summary** 

- It is a copy&paste of non-experimental implementation
- In the first place, I added one successful test for the original `ZSink.fromOutputStream`, because there was no any.
- Added the same test for experimental implementation.

### DIFF
--- a/streams-tests/jvm/src/test/scala/zio/stream/ZSinkPlatformSpecificSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/ZSinkPlatformSpecificSpec.scala
@@ -5,6 +5,7 @@ import zio.test.Assertion._
 import zio.test._
 
 import java.nio.file.Files
+import java.io.ByteArrayOutputStream
 
 object ZSinkPlatformSpecificSpec extends ZIOBaseSpec {
   override def spec: Spec[Any, TestFailure[Throwable], TestSuccess] = suite("ZSink JVM")(
@@ -21,6 +22,18 @@ object ZSinkPlatformSpecificSpec extends ZIOBaseSpec {
             } yield assert(data)(equalTo(str)) && assert(bytes.length.toLong)(equalTo(length))
           }
 
+      }
+    ),
+    suite("fromOutputStream")(
+      test("writes to byte array ouput stream") {
+        val data = (0 to 100).mkString
+
+        for {
+          bytes  <- Task(data.getBytes("UTF-8"))
+          os      = new ByteArrayOutputStream(data.length)
+          length <- ZStream.fromIterable(bytes).run(ZSink.fromOutputStream(os))
+          str    <- Task(os.toString())
+        } yield assert(data)(equalTo(str)) && assert(bytes.length.toLong)(equalTo(length))
       }
     )
   )
--- a/streams-tests/jvm/src/test/scala/zio/stream/ZSinkPlatformSpecificSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/ZSinkPlatformSpecificSpec.scala
@@ -32,7 +32,7 @@ object ZSinkPlatformSpecificSpec extends ZIOBaseSpec {
           bytes  <- Task(data.getBytes("UTF-8"))
           os      = new ByteArrayOutputStream(data.length)
           length <- ZStream.fromIterable(bytes).run(ZSink.fromOutputStream(os))
-          str    <- Task(os.toString())
+          str    <- Task(os.toString("UTF-8"))
         } yield assert(data)(equalTo(str)) && assert(bytes.length.toLong)(equalTo(length))
       }
     )
--- a/streams-tests/jvm/src/test/scala/zio/stream/experimental/ZSinkPlatformSpecificSpec.scala
+++ b/streams-tests/jvm/src/test/scala/zio/stream/experimental/ZSinkPlatformSpecificSpec.scala
@@ -1,0 +1,26 @@
+package zio.stream.experimental
+
+import zio._
+import zio.test._
+import zio.test.Assertion._
+import java.io.ByteArrayOutputStream
+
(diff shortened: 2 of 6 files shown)


streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -503,3 +503,43 @@ trait ZStreamPlatformSpecificConstructors {
...
+
+  /**
+   * Uses the provided `OutputStream` to create a [[ZSink]] that consumes byte chunks
+   * and writes them to the `OutputStream`. The sink will yield the count of bytes written.
+   *
+   * The caller of this function is responsible for closing the `OutputStream`.
+   */
+  final def fromOutputStream(
+    os: OutputStream
+  )(implicit trace: ZTraceElement): ZSink[Any, IOException, Byte, IOException, Byte, Long] = fromOutputStreamManaged(
```
grouzen, line 518: I hope it's correct to have IOException for both `InErr` and `OutErr` types.

streams-tests/jvm/src/test/scala/zio/stream/ZSinkPlatformSpecificSpec.scala
```diff
@@ -22,6 +23,18 @@ object ZSinkPlatformSpecificSpec extends ZIOBaseSpec {
...
+    ),
+    suite("fromOutputStream")(
+      test("writes to byte array ouput stream") {
+        val data = (0 to 100).mkString
+
+        for {
+          bytes  <- Task(data.getBytes("UTF-8"))
+          os      = new ByteArrayOutputStream(data.length)
+          length <- ZStream.fromIterable(bytes).run(ZSink.fromOutputStream(os))
+          str    <- Task(os.toString())
```
vigoo, line 35: To avoid system-dependent issues, let's explicitly set `UTF-8` charset for the `toString` here because this way it uses the default charset which may not be the same as the one you passed for `getBytes`

streams/jvm/src/main/scala/zio/stream/experimental/platform.scala
vigoo: I think we need a review of the whole `ZSink` regarding this because what we probably want here is `InErr >: IOException` but `foldLeftChunksZIO` forces you to have the same in both type params now.

So let's merge it like this and we will do that review separately

vigoo:
You need to add `ZSinkPlatformSpecificConstructors` to the other platforms too (even if they are empty) and please run a `sbt fmt` to pass the lint check

streams/js/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -200,3 +200,7 @@ trait ZStreamPlatformSpecificConstructors {
 
   trait ZStreamConstructorPlatformSpecific extends ZStreamConstructorLowPriority1
 }
+
+trait ZSinkPlatformSpecificConstructors {
+  self: ZSink.type =>
```
ithinkicancode, line 205: Can be simply `trait ZSinkPlatformSpecificConstructors`

streams/native/src/main/scala/zio/stream/experimental/platform.scala
```diff
@@ -200,3 +200,7 @@ trait ZStreamPlatformSpecificConstructors {
 
   trait ZStreamConstructorPlatformSpecific extends ZStreamConstructorLowPriority1
 }
+
+trait ZSinkPlatformSpecificConstructors {
```
ithinkicancode, line 204: Same - can be just `trait ZSinkPlatformSpecificConstructors`

streams-tests/jvm/src/test/scala/zio/stream/experimental/ZSinkPlatformSpecificSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+  override def spec: ZSpec[Any, Throwable] = suite("ZSink JVM experimental")(
+    suite("fromOutputStream")(
+      test("writes to byte array output stream") {
+        val data = (0 to 100).mkString
+
+        for {
+          bytes  <- Task(data.getBytes("UTF-8"))
+          os      = new ByteArrayOutputStream(data.length)
+          length <- ZStream.fromIterable(bytes).run(ZSink.fromOutputStream(os))
+          str    <- Task(os.toString("UTF-8"))
```
ithinkicancode, line 19: I would use the constant defined in java.nio.charset.StandardCharset

grouzen: I tried it, but I got a compilation error on 19 line:
```
ZSinkPlatformSpecificSpec.scala:20:29: overloaded method toString with alternatives:
[error]   (x$1: Int)String <and>
[error]   (x$1: String)String
[error]  cannot be applied to (java.nio.charset.Charset)
[error]           str    <- Task(os.toString(UTF_8))
[error]                             ^
[error] one error found
```
It's strange because in source there is overloaded method toString parameterized by java.nio.charset.Charset.
That's why I decided to use String literal instead.

vigoo (approved):

adamgfraser:
@grouzen Thank you and congratulations on your first contribution to ZIO!

Issue #5849, opened 2021-10-29
ZIO 2.0: Create test suite for Cause.prettyPrint


jdegoes:
A simple test suite can verify that useful traces are captured in multi-method scenarios.

An example:

```scala
object TraceExample extends ZIOAppDefault {
  def otherFailure =
    for {
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.fail("Uh oh!").ensuring(ZIO.dieMessage("other failure"))
    } yield ()

  def deepFailure =
    for {
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- ZIO.succeed(123)
      _ <- otherFailure
    } yield ()

  def run = deepFailure.catchAllCause(cause => Console.printLine(cause.prettyPrint))
}
```

We want to see the `run`, `otherFailure`(inside `deepFailure`), and `ZIO.fail` (inside `otherFailure`) in the stack trace, together with the finalizer failure (inside `ensuring`) as a `Suppressed` cause.

A few simple examples like this which test tracing and pretty printing together can constitute our test suite.

baldram:
Hi #zio-hackathon! I would like to work on this issue 😸

adamgfraser:
Awesome! It is all yours!

baldram:
Here is a draft implementation: https://github.com/zio/zio/pull/5900 
Would you please have a look whether I'm on the good track?

I might try moving a code to the `StackTracesSpec.scala`, however this test seems to be a bit different from others, as it's kind of alone in the `scala-2.12` tree, and doesn't run locally from my IDE (no issues with others).

PS: The PR got stuck, as I'm the "first time contributor" requiring an approval for running workflows.

baldram:
I created a new clean PR for this. It is blocked by a #5931 bug fix.
Once a bugfix is merged, time to rebase the MR of this ticket, and the all done.

baldram:
In general tests are implemented. I might implement one more scenario to verify embedded stack traces (2nd and 3rd level); however, a possible bug requires fixing first. More details here: https://discord.com/channels/724305612166660202/776907921761370112/908299720160600124

PR #5900, opened 2021-10-30
#5849 Create test suite for Cause.prettyPrint

baldram:
Here is an implementation of 3 test cases out of the ticket description. If there is any additional thing we could check, please let me know.

I check failures existence in the stack trace, including pretty markers.

---
Blocked by: https://github.com/zio/zio/pull/5931
A target branch to be switched then, and changes adjusted.

---
A new, "clean" MR created https://github.com/zio/zio/pull/5949.
Closing this one.


jdegoes:
@baldram Can you base your work of `series/2.x`? The `master` branch is for ZIO 1.x, and it already has tests.

Thank you!

baldram:
I will base my work of series/2.x. Actually I tried, and then I faced an issue when calling `cause.prettyPrint`:
```
Exception in thread "zio-fiber-0" java.lang.NumberFormatException: java.lang.NumberFormatException: For input string: "\data\git\zio-hackaton\zio\core-tests\shared\src\test\scala\zio\StackTracesSpec.scala"
```
It seems like some bug has appeared. Once I find a solution, I will deliver a result of work adjusted for 2.x.

baldram:
A ticket is blocked by a bug I found when executing `cause.prettyPrint`. 
The problem is with casting `.toInt` on some strings.

Here is a code causing an issue:
```scala
case regex(location, file, line, column) => Some((location, file, line.toInt, column.toInt))
``` 
It gives:
```
Exception in thread "zio-fiber-0" java.lang.NumberFormatException: java.lang.NumberFormatException: For input string: "81:26"
```
Tried to look for the root cause. Below, what is happening in the pattern matching mentioned above. 
In my case (Windows unfortunately 😏) I get the following:
```
location=[zio.StackTracesSpec.spec.failure], 
file=[C], 
line=[\data\git\zio-hackaton\zio\core-tests\shared\src\test\scala\zio\StackTracesSpec.scala], 
column=[81:26]
```
Now, the line, and column can't be casted in this way.
Not sure whether a bug https://github.com/zio/zio/issues/5914 is about the same problem, or we need a new one.

Another possibility is to fix it within this ticket. 
However I'm not sure whether we should extend the mentioned pattern matching in Tracer, or it's a problem with strings generated by macro?

baldram:
Ok, here is a bug connected to the problem.
https://github.com/zio/zio/issues/5917

PR #5901, opened 2021-10-30
Add CyclicBarrier

### LINKED ISSUE #5801: Add CyclicBarrier
>A synchronization aid that allows a set of fibers to all wait for each other to reach a common barrier point. CyclicBarriers are useful in programs involving a fixed sized party of fibers that must occasionally wait for each other. The barrier is called cyclic because it can be re-used after the waiting fibers are released.

```scala
package zio.concurrent

trait CyclicBarrier:
  def parties: Int
  def waiting: UIO[Int]
  def await: IO[Unit, Unit]
  def reset: UIO[Unit]
  def isBroken: UIO[Boolean]

object CyclicBarrier:
  def make(n: Int): UIO[CyclicBarrier]
```

regiskuckaertz:
Closes #5801 

@adamgfraser @mijicd The action is quite restrictive, what are your thoughts on this?

### DIFF
--- a/core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
+++ b/core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
@@ -1,0 +1,61 @@
+package zio.concurrent
+
+import zio._
+
+final class CyclicBarrier(
+  private val _parties: Int,
+  private val _waiting: Ref[Int],
+  private val _lock: Ref[Promise[Unit, Unit]],
+  private val _action: UIO[Any],
+  private val _broken: Ref[Boolean]
+) {
+
+  /** The number of parties required to trip this barrier. */
+  def parties: Int = _parties
+
+  /** The number of parties currently waiting at the barrier. */
+  val waiting: UIO[Int] = _waiting.get
+
+  /** Waits until all parties have invoked await on this barrier. Fails if the barrier is broken. */
+  val await: IO[Unit, Int] =
+    _broken.get.flatMap(if (_) IO.fail(()) else UIO.unit) *>
+      _waiting.modify {
+        case n if n + 1 == parties => _action *> succeed.as(_parties - n - 1)                            -> (n + 1)
+        case n                     => _lock.get.flatMap(_.await.onInterrupt(break)).as(_parties - n - 1) -> (n + 1)
+      }.flatten
+
+  /** Resets the barrier to its initial state. Breaks any waiting party. */
+  val reset: UIO[Unit] =
+    for {
+      w <- waiting
+      _ <- if (w > 0)
+             (fail *>
+               Promise.make[Unit, Unit].flatMap(_lock.set) *>
+               _waiting.set(0) *>
+               _broken.set(false)).uninterruptible
+           else UIO.unit
+    } yield ()
+
+  val isBroken: UIO[Boolean] = _broken.get
+
+  private val break: UIO[Unit] =
+    _broken.set(true) *> fail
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  /** The number of parties required to trip this barrier. */
+  def parties: Int = _parties
+
+  /** The number of parties currently waiting at the barrier. */
+  val waiting: UIO[Int] = _waiting.get
+
+  /** Waits until all parties have invoked await on this barrier. Fails if the barrier is broken. */
+  val await: IO[Unit, Int] =
+    _broken.get.flatMap(if (_) IO.fail(()) else UIO.unit) *>
```
regiskuckaertz, line 39: Should the whole action be uninterruptible? (except for `action` probably)

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
```diff
@@ -0,0 +1,66 @@
...
+ * A synchronization aid that allows a set of fibers to all wait for each other to reach a common barrier point.
+ *
+ * CyclicBarriers are useful in programs involving a fixed sized party of fibers that must occasionally wait for each
+ * other. The barrier is called cyclic because it can be re-used after the waiting fibers are released.
+ *
+ * A CyclicBarrier supports an optional action command that is run once per barrier point, after the last fiber in the
+ * party arrives, but before any fibers are released. This barrier action is useful for updating shared-state before
+ * any of the parties continue.
+ */
+final class CyclicBarrier(
```
mijicd, line 15: Does it makes sense to make the constructor private?

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
```diff
@@ -0,0 +1,66 @@
...
+  private val _broken: Ref[Boolean]
+) {
+  private val break: UIO[Unit] =
+    _broken.set(true) *> fail
+
+  private val fail: UIO[Unit] =
+    _lock.get.flatMap(_.fail(()).unit)
+
+  private val succeed: UIO[Unit] =
+    _lock.get.flatMap(_.succeed(()).unit)
```
mijicd, line 29: I'd define these as `UIO[Any]`.

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
regiskuckaertz: Yes, good catch

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
regiskuckaertz: Done

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
regiskuckaertz: Done

core/shared/src/main/scala/zio/concurrent/CyclicBarrier.scala
regiskuckaertz: Done

mijicd (approved):

mijicd (dismissed):

mijicd (approved):

PR #5904, opened 2021-10-30
Make Recursive Functions Stack Safe

### LINKED ISSUE #5852: ZIO 2.0: Make Cause#prettyPrint stack safe
Looking at Cause#prettyPrint, one can identify a number of recursive functions that are not tail-recursive. These functions should be refactored to be tail-recursive so that the Cause rendering is stack safe, even for very deep renderings.

adamgfraser:
Resolves #5852.

### DIFF
--- a/core/shared/src/main/scala/zio/Cause.scala
+++ b/core/shared/src/main/scala/zio/Cause.scala
@@ -233,26 +233,37 @@ sealed abstract class Cause[+E] extends Product with Serializable { self =>
     }
 
   def linearize[E1 >: E]: Set[Cause[E1]] = {
-    def loop[E](cause: Cause[E]): Set[Cause[E]] =
-      cause match {
-        case Empty => Set()
 
-        case Both(left, right) => loop(left) ++ loop(right)
+    sealed trait Step
 
-        case Stackless(cause, stackless) => loop(cause).map(Stackless(_, stackless))
+    case object Both                               extends Step
+    case object Then                               extends Step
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -233,26 +233,37 @@ sealed abstract class Cause[+E] extends Product with Serializable { self =>
...
+          out.foldLeft[List[Set[Cause[E1]]]](List.empty) {
+            case (acc, Right(causes)) => causes :: acc
+            case (acc, Left(Both)) =>
+              val left :: right :: causes = (acc: @unchecked)
+              (left ++ right) :: causes
+            case (acc, Left(Then)) =>
+              val left :: right :: causes = (acc: @unchecked)
+              left.flatMap(l => right.map(r => l ++ r)) :: causes
+            case (acc, Left(Stackless(stackless))) =>
+              acc.map(_.map(Cause.stackless))
```
regiskuckaertz, line 263: Shouldn't this be 
```scala
val cause :: causes = acc
cause.map(Cause.stackless) :: causes
```
?

regiskuckaertz (commented):
Wow, did you just read "clowns to the left, jokers to the right"? Beautiful implementation 😍

core/shared/src/main/scala/zio/Cause.scala
adamgfraser: Yes! Fixed now!

regiskuckaertz (approved):

Issue #4528, opened 2020-12-30
Support mirroring FiberRef contents to ThreadLocal


dkarlinsky:
It would be really useful for `FiberRef` contents to be mirrored to non-zio infra (outside of our control) backed by `ThreadLocal`s.
For example, when using NewRelic and it's [Tokens](https://docs.newrelic.com/docs/agents/java-agent/async-instrumentation/java-agent-api-asynchronous-applications#tokens) to connect async flows to what they call a transaction (think an incoming request).
In this case, we would store the new relic token in a `FiberRef` like this:
```scala
    FiberRef.make(nrToken, link = _.link())
```
The provided `link` function will be called by zio after every async boundary so any effect running in this fiber (or a fiber inheriting this ref) will run on a thread "linked" to the token.
There might also be a need for `unlink` function to clear the the thread at the end.

(`MDC` is a simpler example, but maybe easier to work around with libraries like `zio-logging` or `FiberRef.unsafeAsThreadLocal`)

I've created this [Draft PR](https://github.com/zio/zio/pull/4421) a while back with a naive implementation for this.

@adamgfraser, @jdegoes WDYT?

dkarlinsky:
Gentle Ping :) 
@adamgfraser @jdegoes

jdegoes:
I think some generic RTS work could be done that would satisfy this use case, although it would not be specifically connected to thread local. Some sort of callback that can be invoked when a fiber starts running on a thread. In fact, the existing code that sets the active fiber (in a thread local) could leverage that mechanism

dkarlinsky:
Yes that is exactly my idea - provide a callback when creating `FiberRef` that will be invoked after every async boundary.
I mentioned `ThreadLocal` since, in most cases the callback will be setting some state backed by a `ThreadLocal`.
My [Draft PR](https://github.com/zio/zio/pull/4421) already implements this, but I did want some guidance on where the correct spot was in `FiberRef`, to insert the call to the `link` callback. 
And I expect there will have to be some benchmarks, to make sure there was no performance hit, especially when not providing the `link` callback.

jdegoes:
I wouldn't do it when creating fiber ref, but do it separately. There's nothing necessarily connected to fiber ref in such a feature. It may be more connected with `Supervisor`. Let me think about this for a week, but I do absolutely think we can get this into ZIO (thanks for the suggestion).

dkarlinsky:
Thanks @jdegoes !

d12frosted:
[...]

PR #5907, opened 2021-10-30
Issue #4528 -  Support mirroring FiberRef contents to ThreadLocal

dkarlinsky:
Defineds a new layer `ThreadLocalBridge`.
The layer installs a supervisor that handles syncing on async boundaries (`unsafeOnResume`, `unsafeOnSuspend`).
`ThreadLocalBridge` provides `makeFiberRef` method to create a special FiberRef that syncs the changes

issue #4528

### DIFF
--- a/core-tests/shared/src/test/scala/zio/SupervisorSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/SupervisorSpec.scala
@@ -4,28 +4,17 @@ import zio.test._
 
 object SupervisorSpec extends ZIOBaseSpec {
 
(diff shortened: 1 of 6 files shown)


dkarlinsky:
@jdegoes need your input on possible ways to reset to the fiber ref's previous value at the end of `FiberRef.locally`
Currently I intercept `FiberRefLocally` (in `unsafeOnEffect`) effects and call `link(locally.localValue)` but I don't see a way reset to the fiber refs previous value after

adamgfraser (approved):
Long time coming! 😃

PR #5908, opened 2021-10-30
Eagerly Capture Trace

adamgfraser:


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -517,8 +518,7 @@ private[zio] final class FiberContext[E, A](
 
                     setFiberRefValue(fiberRef, zio.localValue)
 
-                    curZio =
-                      zio.zio.ensuring(ZIO.succeed(setFiberRefValue(fiberRef, oldValue))(curZio.trace))(curZio.trace)
+                    curZio = zio.zio.ensuring(ZIO.succeed(setFiberRefValue(fiberRef, oldValue))(zio.trace))(zio.trace)
```
adamgfraser, line 521: This change causes the ZIO Stream test suite to not terminate on Scala 3.

Issue #5853, opened 2021-10-29
ZIO 2.0: Delete ZTrace#prettyPrint


jdegoes:
Instead, pretty printing should directly utilize methods similar to `Cause#prettyPrint` (in fact, maybe `Cause#prettyPrint` can always be used).

As far as I know, the only place that uses this is fiber dumps.

zhrebicek:
I'll take this (hope kids would let me finish it)

zhrebicek:
@jdegoes There was really only one place it was used. 

Is it ok like in the PR?
```
-       |${dump.trace.fold("")(_.prettyPrint)}
+      |${dump.trace.fold("")(Cause.traced(Cause.empty, _).prettyPrint)}
```

zhrebicek:
@jdegoes 
I did it, it complies locally just fine but it does not see `Cause#traced` in the CI, what am I missing here?

```
[error] /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/FiberRenderer.scala:65:5: value traced is not a member of object zio.Cause; incompatible interpolation method s
[error]     s"""
[error]     ^
```

zhrebicek:
Did not give it enough time to see I just did not pulled latest changes that remove `Cause#traced` it should be ok now.

PR #5909, opened 2021-10-30
Delete ZTrace#prettyPrint, use Cause#prettyPrint instead (#5853)

zhrebicek:
Hope it is intended to use the printer from `Cause` like this `Cause.traced(Cause.empty, _).prettyPrint`


zhrebicek:
It compiles locally just fine but fails in CI because it supposedly can't find `Cause#traced`  what am I missing here?

adamgfraser:
It looks like it is trying to find Scaladoc for that which doesn't really make sense. Does it help if you extract the variable out of the string interpolation?

zhrebicek:
@adamgfraser tried that with no luck

zhrebicek:
The reason was me force pushing stuff over new implementation of `Cause` that was really missing `traced`

adamgfraser (approved):
Thank you! 🙏

PR #5910, opened 2021-10-30
Consolidating all source location macros into `ZTraceElement`

### LINKED ISSUE #5903: ZIO Test 2.0: Use ZTraceElement instead of SourceLocation
ZIO Test independently evolved the ability to capture source locations. Now that we have ZTraceElement which directly captures source locations, we can replace usages of SourceLocation inside ZIO Test with ZTraceElement.

hmemcpy:
Fixes #5903


adamgfraser (changes requested):
Great cleanup here! Can you resolve conflicts and then we can get this merged assuming tests are passing?

adamgfraser (approved):
Thank you! 🙏

adamgfraser:
@hmemcpy Can you resolve conflicts?

hmemcpy:
Again!? 😆

adamgfraser (approved):
Too much good stuff going on! 😃

hmemcpy:
What did I miss, by the way? I noticed you did some work around the test runner?

adamgfraser:
We merged the PR on shared layers in specs so we are going to be replacing `DefaultRunnableSpec` with `ZIOSpec` which will let you define a layer and then if that layer is referenced in multiple tests it will automatically be shared. We also moved all the test services out of the `zio.test.environment` package so you don't need to do as many imports.

hmemcpy:
Ooh very cool! Looking forward to the refactoring :)

adamgfraser (approved):

Issue #5050, opened 2021-05-08
ZIO 2.0: Improve Chunk[Boolean] for bit-oriented operations


jdegoes:
Recently a project tried to use `Chunk[Boolean]` for bit-oriented operations (low-level codecs), and encountered some difficulties. This ticket documents everything currently missing that is required to make `Chunk[Boolean]` suited for these use cases:

1. New constructors for `Chunk[Boolean]` that can convert from `Chunk[Int]` and `Chunk[Long]` into `Chunk[Byte]` (currently, only `Chunk[Byte]` can be "interpreted" as a `Chunk[Boolean]`). These constructors must accept `Endianness` (`BigEndian` or `LittleEndian`) in order to know how bytes are packed into an int or long.  (**DIFFICULTY: HARD**)
2. Make all constructors from `Byte`, `Int`, and `Long` accept padding, because not all of the chunk may represent valid data (e.g. it's possibly 7 bits in the last byte do not correspond to valid data in the bit array and they are there only because of padding).  (**DIFFICULTY: MODERATE**)
3. For performance reasons, it may be desirable to generalize the constructors to accept `minBit` and `maxBit`, i.e. allow the user to define a range on the raw Chunk with the byte/int/long data. With this range, we could further specialize `take`, `drop`, and `slice` on `Chunk[Boolean]` in order to accelerate these operations by using the range parameters.  (**DIFFICULTY: MODERATE**)
4. A "destructor" (which can go from `Chunk[Boolean]` to `Chunk[Byte]`, `Chunk[Int]`, or `Chunk[Long]`). When going to either `Int` or `Long` (but not when going to `Chunk[Byte]`), the destructor must accept `Endianness`, because it affects how the bytes are packed into the resulting int or long values.   (**DIFFICULTY: MODERATE**)
5. Bitwise operations, including `&`, `|`, and `^`. These should not operate inefficiently on booleans (e.g. `left.zipWith(right)(_ && _)`), but rather, should operate directly on the underlying chunks of byte/int/long that back the `Chunk[Boolean]`.  (**DIFFICULTY: EASY**)
6. Shift left and shift right operations. If one has, `Chunk(true, false, true)` and does a shift left, then one should end up with `Chunk(false, true, false)`. If one does shift right, then one should end up with `Chunk(false, true, false)`. These correspond to how the bitwise shift left/right operators work on a data type with a fixed size.  (**DIFFICULTY: EASY**)
[...]

PR #5911, opened 2021-10-30
construct Chunk[Boolean] from Chunk[Int] and Chunk[Long] based on endianness

helanto:
Deals with first requirement from #5050


jdegoes:
@helanto I think this was since done independently. Thank you for your work on this PR!

PR #5916, opened 2021-10-30
Save FiberId as long in FiberContext

### LINKED ISSUE #5874: ZIO 2.0: Improve efficiency of FiberId storage inside FiberContext
After #5871, we can change `FiberContext` to store only a `Long` for the `FiberId`, by packing both sequence number and start time into a single `Long`. Then the `def fiberId: FiberId.Runtime` can construct a new `FiberId.Runtime` dynamically from the `Long`.

987Nabil:
Resolves #5874


987Nabil:
@jdegoes Do you have any idea from where this test failure might come? Only happening on Scala 2.11 feels like I need some inside knowledge I might not have.

adamgfraser:
@987Nabil That is flakiness in one of the stream tests that is not related to your PR. Will run again.

987Nabil:
Thx @adamgfraser

PR #5921, opened 2021-10-30
Add concurrent set methods

### LINKED ISSUE #5890: Add scala set-like methods to ConcurrentSet
After #5823 is done, we can start adding operators that will bring the data structure closer to what a collection type looks like in Scala-land. Things like:

- |, &, &~
- collect, collectFirst, filter
- map, flatMap, fold, groupBy, grouped
- exists, forall
- etc

kitlangton:
Attempts to close #5890


kitlangton:
@regiskuckaertz Many methods lack atomicity (as you mentioned on #5823). It would be interesting to explore what we can do here.

core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
```diff
@@ -11,15 +12,47 @@ final class ConcurrentSet[A] private (private val underlying: ConcurrentHashMap.
     UIO(underlying.add(x))
 
   def addAll(xs: Iterable[A]): UIO[Boolean] =
-    UIO {
-      var added = false
-      xs.foreach(x => added = underlying.add(x))
-      added
-    }
+    UIO(underlying.addAll(xs.asJavaCollection))
```
regiskuckaertz, line 15: Nice! And I didn't notice the subtle bug there (should have been `underlying.add(...) || added`. You may have add an annotation if compiler complains here.
```suggestion
    UIO(underlying.addAll(xs.asJavaCollection): @silent("JavaConverters"))
```

core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
```diff
@@ -11,15 +12,47 @@ final class ConcurrentSet[A] private (private val underlying: ConcurrentHashMap.
...
+
+  def collectFirst[B](pf: PartialFunction[A, B]): UIO[Option[B]] =
+    UIO(underlying.asScala.toSet.collectFirst(pf))
+
+  def exists(p: A => Boolean): UIO[Boolean] =
+    UIO(underlying.asScala.toSet.exists(p))
```
regiskuckaertz, line 21: To regain atomicity, I don't see another solution than to pay the cost of `forEach`. It would be possible to shortcut via `stream` I suppose but very unlikely this would be atomic?

core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
```diff
@@ -11,15 +12,47 @@ final class ConcurrentSet[A] private (private val underlying: ConcurrentHashMap.
...
+
+  def fold[R, E, S](zero: S)(f: (S, A) => S): UIO[S] =
+    UIO(underlying.asScala.toSet.foldLeft(zero)(f))
+
+  def foldZIO[R, E, S](zero: S)(f: (S, A) => ZIO[R, E, S]): ZIO[R, E, S] =
+    ZIO.foldLeft(underlying.asScala.toSet)(zero)(f)
+
+  def foreach[R, E, B](f: A => ZIO[R, E, B]): ZIO[R, E, Set[B]] =
+    ZIO.foreach(underlying.asScala.toSet)(f)
```
regiskuckaertz, line 30: Here if we want to provide this we don't have other choice than to pass use the RTS. I'd be inclined to follow Adam's suggestion that we don't add effectual functions.

regiskuckaertz (commented):
Awesome, thanks for all your work! I feel like I added more questions than answers 😆

core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
kitlangton: Ah yes! `forEach` would work better here.

regiskuckaertz (approved):
YAY! 💪

PR #5922, opened 2021-10-30
Implement `ZSink#raceBoth`

### LINKED ISSUE #5864: ZIO Streams 2.0: Implement ZSink#raceBoth
In ZIO Streams 2.0 `ZSink`'s `raceBoth` is not implemented yet. It's signature is the following:

```scala
  /**
   * Runs both sinks in parallel on the input, returning the result or the error from the
   * one that finishes first.
   */
  final def raceBoth[R1 <: R, InErr1 <: InErr, OutErr1 >: OutErr, A0, In1 <: In, L1 >: L, Z1 >: Z](
    that: ZSink[R1, InErr1, In1, OutErr1, L1, Z1]
  )(implicit trace: ZTraceElement): ZSink[R1, InErr1, In1, OutErr1, L1, Either[Z, Z1]]
```

The `ZSink#race` also uses this so right now it does not work either.

Please add tests for it as well, as we currently don't have any.

eriktim:
This PR adds an implementation for `ZSink#raceBoth`.

Pretty new to `ZStream`, but to my understanding `ZChannel#mergeWith` does exactly what we need here.
For some reason the test is flaky in cases where both `success1` and `success2` are `true`.
Still trying to wrap my head around that, so any insight is welcome :slightly_smiling_face: 

Fixes #5864

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/experimental/SinkUtils.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/experimental/SinkUtils.scala
@@ -1,0 +1,37 @@
+package zio.stream.experimental
+
+import zio.test.Assertion.equalTo
+import zio.test.{Assertion, TestResult, assert}
+import zio.{IO, UIO}
+
+object SinkUtils {
+
+  def findSink[A](a: A): ZSink[Any, Nothing, A, Unit, A, A] =
+    ZSink
+      .fold[Nothing, A, Option[A]](None)(_.isEmpty)((_, v) => if (a == v) Some(a) else None)
+      .mapZIO {
+        case Some(v) => IO.succeedNow(v)
+        case None    => IO.fail(())
+      }
+
+  def sinkRaceLaw[E, A, L](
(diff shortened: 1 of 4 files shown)


vigoo:
Thanks! I will try to understand where the flakiness is coming from and get back to you.

adamgfraser:
I tried running this with changing the input stream to `ZStream.fromIterable(List(20, 40))` and got the following error. So I think something is not working in our implementation of `mergeWith`.

```
[info] - ZSinkSpec
[info]   - Combinators
[info]     - raceBoth
[info]       Exception in thread "zio-fiber-" java.lang.InterruptedException: Interrupted by thread "zio-fiber-"
```

eriktim:
Cleanup the debugging stuff and also updated `ZChannel#toHub` and `ZChannel#toQueue` to accept a `Hub[Either[Exit[Err, Done], Elem]]` instead of `Hub[Exit[Either[Err, Done], Elem]]` as was done with the `ZChannel#from*` counterparts.

adamgfraser (approved):

PR #5923, opened 2021-10-30
#5855 - Add "Message" as type parameter to ZLogger

### LINKED ISSUE #5855: ZIO 2.0: Generalize ZLogger to accept messages of type `-Message`
ZLogger should get a new type parameter:

```scala
trait ZLogger[-Message, +Output] { ... }
```

This will allow generalizing the `message: () => String` parameter of `ZLogger#apply` from `String` to `Message`. 

Existing functionality can be captured by requiring `ZLogger[String, A]` wherever the old `ZLogger[A]` was required.

This will pave the way for later tickets that generalize the primitive ZIO logging operator to accept messages of `Any` type, which will later allow higher-performance structured logging.

/cc @vigoo

xsistens:
Hi all,
this is my first approach to tackle ticket #5855 (Fixes #5855). Can someone confirm that what i did is going into the right direction? Something to check are maybe the adjustments to the `++` combinator I hope that my solution to the variance is correct. I also stopped quite early at the `RuntimeConfig` and added `String` there as the fix type parameter for the `ZLogger` as suggested in the ticket.

I would like to add some more tests but I'm not really sure in what to test since the ZIO runner is fixed to the `String` type which is already tested. 

Would be greatful for every advice and happy to adjust or incorperate any kind of feedback!

Thanks in advance

### DIFF
--- a/core-tests/shared/src/test/scala/zio/LoggingSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/LoggingSpec.scala
@@ -15,7 +15,7 @@ object LoggingSpec extends ZIOBaseSpec {
     context: Map[FiberRef.Runtime[_], AnyRef],
     spans: List[LogSpan]
   ) {
-    def call[A](zlogger: ZLogger[A]): A =
+    def call[A](zlogger: ZLogger[String, A]): A =
       zlogger(trace, fiberId, logLevel, message, context, spans)
   }
 
@@ -25,8 +25,8 @@ object LoggingSpec extends ZIOBaseSpec {
 
   val clearOutput: UIO[Unit] = UIO(_logOutput.set(Vector.empty))
 
-  val testLogger: ZLogger[Unit] =
-    new ZLogger[Unit] {
+  val testLogger: ZLogger[String, Unit] =
+    new ZLogger[String, Unit] {
       @tailrec
       def apply(
         trace: ZTraceElement,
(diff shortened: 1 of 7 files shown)


regiskuckaertz:
@xsistens I think this is precisely what the ticket asked for and you did a great work.

Your adjustments to `++` look correct to me:
- if `self` can log any animal, it can also log any specific animal, like if `that` could only log elephants, it would also work for `self` because an elephant is an animal
- OTOH, if `self` can log only elephants, and `that` can log any animal, then you're in a pickle as `message` must be able to produce any animal (otherwise you would not be able to pass it to `that`), like a giraffe, and in that case `self` would be none the wiser

🦒 🐘 

Testing-wise, I think we will only be able to check if this animal logic makes sense once the follow-up ticket is done 😄

xsistens:
@regiskuckaertz Thanks for the feedback. Do you have any hints for me what I can do in regards to the failing ci pipeline? To be honest I cannot find out what exactly the problem is. It seems to me that the failing tests are somehow flaky and not connected to the changes I made. But I'm not 100% sure about that.

regiskuckaertz:
@xsistens these errors are unrelated to your changes. I will let @adamgfraser approve if he's happy with your changes, great work!

jdegoes:
Hi @xsistens, this looks great to me! Can you fix the conflicts? The build should pass now and we can merge swiftly.

xsistens:
Hi John. I'm currently not available but I will try to finish it in the next 12 hours.

xsistens:
Hi @jdegoes, I think one of the maintainers have to kick the CI job to run. Locally I still have some compile issues but they again are not related to any of my changes :/ If I can do anything else please let me know!

adamgfraser:
@xsistens Running!

xsistens:
It feels like there is randomly failing a different test suite :(

xsistens:
@adamgfraser Hey Adam, seems like all tests are passing now. Maybe it's time to merge it. In two days I won't have access to the internet for a while, so let me know if I can do something before that.

adamgfraser (approved):

adamgfraser:
@xsistens Merged!

Issue #5875, opened 2021-10-29
ZIO 2.0: Implement useful `toString()` method on `FiberContext`


jdegoes:
The `toString()` method on `FiberContext` is not very useful. We should override this method to dump the fiber, by using the fiber dump rendering. The goal is that if someone `println(myFiber)`, they should see lots of useful information that will help them debug the problem they are trying to solve.

ASRagab:
Working on this.

PR #5924, opened 2021-10-30
Fixes - zio/zio#5875

ASRagab:
- Adds `.toString()` for a `FiberContext`,`ExecutionMetrics`, `Executor`, `Platform`

```
FiberContext(identityHashCode=521960438)
  fiberId:         Id(1635628369111,0)
  platform:        
    executor:
      yieldOpCount:   2048
      metrics:        ExecutionMetrics(concurrency=16, capacity=2147483647, size=0, enqueuedCount=0, dequeuedCount=0, workersCount=0)
    tracing:       Tracing(zio.internal.stacktracer.Tracer$$anon$1@212bf671,TracingConfig(true,true,true,100,100,10,10,10))
  state:           Executing(Running(false),List(),Empty)
  interruptStatus: Interruptible
  scopeKey:        null
```

Fixes - zio/zio#5875

### DIFF
--- a/core/shared/src/main/scala/zio/internal/Executor.scala
+++ b/core/shared/src/main/scala/zio/internal/Executor.scala
@@ -66,6 +66,8 @@ abstract class Executor extends ExecutorPlatformSpecific { self =>
       if (submit(command)) ()
       else throw new java.util.concurrent.RejectedExecutionException
 
+  override def toString: String =
+    s"Executor(yieldOpCount=$yieldOpCount, metrics=${metrics.mkString})"
 }
 
 object Executor extends DefaultExecutors with Serializable {
--- a/core/shared/src/main/scala/zio/internal/FiberContext.scala
+++ b/core/shared/src/main/scala/zio/internal/FiberContext.scala
@@ -1077,6 +1077,19 @@ private[zio] final class FiberContext[E, A](
     // For improved fairness, we resume in order of submission:
     observers.reverse.foreach(k => k(result))
   }
+
+  override def toString(): String =
+    s"""${super.toString()}
+       |  fiberId:         $fiberId
+       |  platform:        
+       |    executor:
+       |      yieldOpCount:   ${platform.executor.yieldOpCount}
+       |      metrics:        ${platform.executor.metrics.getOrElse("not tracked")}
+       |    tracing:       ${platform.tracing}
+       |  state:           $state
+       |  interruptStatus: ${InterruptStatus.fromBoolean(interruptStatus.peekOrElse(true))}
+       |  scopeKey:        ${scopeKey}
+    """.stripMargin
 }
 private[zio] object FiberContext {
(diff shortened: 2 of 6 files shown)


core/shared/src/main/scala/zio/internal/Executor.scala
```diff
@@ -66,6 +66,8 @@ abstract class Executor extends ExecutorPlatformSpecific { self =>
       if (submit(command)) ()
       else throw new java.util.concurrent.RejectedExecutionException
 
+  override def toString: String =
+    s"Executor(yieldOpCount=$yieldOpCount, metrics=${metrics.map(_.toString).getOrElse("")})"
```
regiskuckaertz, line 70:
```suggestion
    s"Executor(yieldOpCount=$yieldOpCount, metrics=${metrics.mkString})"
```

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -1067,6 +1067,20 @@ private[zio] final class FiberContext[E, A](
     // For improved fairness, we resume in order of submission:
     observers.reverse.foreach(k => k(result))
   }
+
+  override def toString: String =
+    s"""
```
regiskuckaertz, line 1072: Is the newline here going to be stripped off?

core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -1067,6 +1067,20 @@ private[zio] final class FiberContext[E, A](
...
+       |FiberContext(identityHashCode=${System.identityHashCode(this)})
+       |  fiberId:         $fiberId
+       |  platform:
```
regiskuckaertz, line 1075: Shouldn't you delegate to the underlying `toString` here? This version is going to be more readable and I haven't looked at the existing pretty-printer, so feel free to ignore.

regiskuckaertz (commented):
😍 superb, I just had two small questions.

core/shared/src/main/scala/zio/internal/FiberContext.scala
ASRagab: Actually, no, I'll update!

core/shared/src/main/scala/zio/internal/FiberContext.scala
ASRagab: Updated

ASRagab:
@regiskuckaertz Thank you, I have updated the PR based on your comments!

regiskuckaertz:
Great work @ASRagab ! Let's get this merged, I think @jdegoes or @adamgfraser will want to give the final 👍

jdegoes:
@ASRagab This has bit-rotted since 2.0's release. I think we could definitely improve the rendering. Please re-open if you find the time to continue your work with main!

PR #5928, opened 2021-10-30
Catch Logging Failures

### LINKED ISSUE #5925: When ZLogger implementation throws exception program hangs
snippet to reproduce

```
object App extends ZIOAppDefault {

  override def hook: RuntimeConfigAspect = RuntimeConfigAspect(_.copy(logger = new ZLogger[Unit] {
    override def apply(
      trace: ZTraceElement,
      fiberId: FiberId,
      logLevel: zio.LogLevel,
      message: () => String,
      context: Map[ZFiberRef.Runtime[_], AnyRef],
      spans: List[LogSpan]
    ): Unit = ???
  }))
 

  override def run = ZIO.log("bla")
}
```

adamgfraser:
Resolves #5925.

We log fiber failures but at this point we can't allow exceptions to be thrown and we can't log the error because the logger just failed so we print it to the console instead.


core/shared/src/main/scala/zio/internal/FiberContext.scala
```diff
@@ -1017,8 +1017,15 @@ private[zio] final class FiberContext[E, A](
   }
 
   private[this] def reportUnhandled(v: Exit[E, A], trace: ZTraceElement): Unit = v match {
-    case Exit.Failure(cause) => log(() => cause.prettyPrint, ZIO.someDebug, trace = trace)
-    case _                   =>
+    case Exit.Failure(cause) =>
+      try {
+        log(() => cause.prettyPrint, ZIO.someDebug, trace = trace)
+      } catch {
+        case t: Throwable =>
```
jdegoes, line 1024: Should probably only catch "non-fatal" exceptions as defined by `runtimeConfig`.

adamgfraser: Yes, as long as that doesn't itself throw an exception.

jdegoes (approved):

PR #5931, opened 2021-10-30
#5917 Improve regex deconstructing trace

### LINKED ISSUE #5914: ZTraceElement.SourceLocation does not work
to reproduce

```
object App extends ZIODefaultApp {
   def getClassNameFromTrace(implicit trace: ZTraceElement) = trace match {
     case ZTraceElement.SourceLocation(_, clazz, _, _) => ZIO.succeed(clazz)
     case _ => ZIO.fail("blah")
  }
  def run = getClassNameFromTrace
 }
```

it seems that format of the trace in my case `zio.logging.slf4j.Slf4jLoggerTest.spec(/Users/rzbikson/projects/zio-logging/slf4j/src/test/scala/zio/logging/slf4j/Slf4jLoggerTest.scala:31:28)` doesn't match regex in `zio.internal.stacktracer.Tracer#regex`

### LINKED ISSUE #5917: ZIO 2.0: Create more robust regular expression for deconstructing ZTraceElement
Inside `Tracer.scala` (there are **two** of these, for Scala 2.x and Scala 3.x), there is a regular expression that tries to extract out the location, file name, row, and column of a trace element.

However, it does not do a good job on Windows because Windows uses the colon operator (:) for drives.

We need to improve this regular expression so it does not match against `:` inside the path name. 

One possibility is as follows:

```
(.*?)\((.*?):([^:]*?):([^:]*?)\)
```

(See [here](https://regex101.com/r/ddGSAU/1))

We need to update these regular expressions so they are more robust to paths that contain colons.

baldram:
Here is a fixed regular expression for deconstructing ZTraceElement. It's tested on Windows platform. 
A test runs fine from IntelliJ. Additional tests introduced here: https://github.com/zio/zio/pull/5949.

Fixes #5917
Fixes #5914

### DIFF
--- a/core-tests/shared/src/test/scala/zio/StackTracesSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/StackTracesSpec.scala
@@ -1,31 +1,35 @@
 package zio
 
-import zio.test.{ZSpec, assertTrue}
+import zio.test.{AssertionValue, BoolAlgebra, ZSpec, assertTrue}
 
 object StackTracesSpec extends ZIOBaseSpec {
 
   def spec: ZSpec[Environment, Failure] = suite("StackTracesSpec")(
     suite("captureSimpleCause")(
       test("captures a simple failure") {
-        val failure =
-          for {
-            _ <- ZIO.succeed(15)
-            _ <- ZIO.fail("Oh no!")
-          } yield ()
-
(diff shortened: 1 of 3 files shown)


baldram:
Fixes #5917
Fixes #5914

adamgfraser (approved):

PR #5933, opened 2021-10-30
Added ConcurrentMap methods

### LINKED ISSUE #5889: Add scala collection-like methods to ConcurrentMap

oliveiosello:
Resolves #5889

### DIFF
--- a/core-tests/shared/src/test/scala/zio/concurrent/ConcurrentMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/concurrent/ConcurrentMapSpec.scala
@@ -176,11 +176,11 @@ object ConcurrentMapSpec extends ZIOBaseSpec {
       suite("putAll")(
         testM("associates all non-existent keys with given value") {
           for {
-            map    <- ConcurrentMap.empty[Int, String]
-            putRes <- map.putAll((1, "A"), (2, "B"), (3, "C"))
-            resA   <- map.get(1)
-            resB   <- map.get(2)
-            resC   <- map.get(3)
+            map  <- ConcurrentMap.empty[Int, String]
+            _    <- map.putAll((1, "A"), (2, "B"), (3, "C"))
+            resA <- map.get(1)
+            resB <- map.get(2)
+            resC <- map.get(3)
           } yield assertTrue(
             resA.get == "A",
             resB.get == "B",
--- a/core-tests/shared/src/test/scala/zio/concurrent/ConcurrentMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/concurrent/ConcurrentMapSpec.scala
@@ -1,13 +1,26 @@
 package zio.concurrent
 
-import zio.ZIOBaseSpec
+import zio._
 import zio.test._
 import zio.test.Assertion._
-import zio.{Chunk, UIO}
 
 object ConcurrentMapSpec extends ZIOBaseSpec {
   def spec: ZSpec[Environment, Failure] =
     suite("ConcurrentMap")(
+      suite("collectFirst")(
+        testM("returns Some of first result for which the partial function is defined") {
+          for {
+            map <- ConcurrentMap.make((1, "A"), (2, "B"), (3, "C"))
+            res <- map.collectFirst { case (3, _) => "Three" }
+          } yield assertTrue(res.get == "Three")
+        },
+        testM("returns None when the partial function is defined for no values") {
+          for {
+            map <- ConcurrentMap.make((1, "A"), (2, "B"), (3, "C"))
+            res <- map.collectFirst { case (4, _) => "Four" }
+          } yield assertTrue(res.isEmpty)
+        }
+      ),
       suite("compute")(
         testM("computes a new value") {
           for {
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
@oliveiosello may I please ask you to sign the CLA?

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -4,11 +4,19 @@ import zio.{Chunk, ChunkBuilder, UIO}
 
 import java.util.concurrent.ConcurrentHashMap
 
+import scala.collection.JavaConverters._
```
mijicd, line 7: I would avoid this.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -27,6 +35,28 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
   def computeIfPresent(key: K, remap: (K, V) => V): UIO[Option[V]] =
     UIO(Option(underlying.computeIfPresent(key, remapWith(remap))))
 
+  /**
+   * Tests whether a given predicate holds true for at least one element in a map.
+   */
+  def exists(p: ((K, V)) => Boolean): UIO[Boolean] =
+    UIO(
+      underlying.asScala.exists(p)
+    )
```
mijicd, line 44:
```suggestion
    UIO(underlying.asScala.exists(p))
```

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -27,6 +35,28 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
...
+  def fold[S](zero: S)(f: (S, (K, V)) => S): UIO[S] =
+    UIO(underlying.asScala.foldLeft(zero)(f))
+
+  /**
+   * Tests whether a predicate is satisfied by all elements of a map.
+   */
+  def forall(p: ((K, V)) => Boolean): UIO[Boolean] =
+    UIO(
+      underlying.asScala.forall(p)
+    )
```
mijicd, line 58:
```suggestion
    UIO(underlying.asScala.forall(p))
```

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -58,6 +88,26 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
   def remove(key: K, value: V): UIO[Boolean] =
     UIO(underlying.remove(key, value))
 
+  /**
+   * Removes all elements which do not satisfy the given predicate.
+   */
+  def removeIf(p: ((K, V)) => Boolean): UIO[Unit] =
```
mijicd, line 94: I'd define the predicate as `(K, V) => Boolean`.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -58,6 +88,26 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
...
+    UIO(
+      underlying.forEach { (k: K, v: V) =>
+        if (p((k, v))) underlying.remove(k)
+      }
+    )
+
+  /**
+   * Removes all elements which do not satisfy the given predicate.
+   */
+  def retainIf(p: ((K, V)) => Boolean): UIO[Unit] =
```
mijicd, line 104: Same here.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -27,6 +35,28 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
   def computeIfPresent(key: K, remap: (K, V) => V): UIO[Option[V]] =
     UIO(Option(underlying.computeIfPresent(key, remapWith(remap))))
 
+  /**
+   * Tests whether a given predicate holds true for at least one element in a map.
+   */
+  def exists(p: ((K, V)) => Boolean): UIO[Boolean] =
```
mijicd, line 41: I'd define predicate as `(K, V) => Boolean`, even though it's not aligned with Scala's specification.

core-tests/shared/src/test/scala/zio/concurrent/ConcurrentMapSpec.scala
```diff
@@ -183,6 +247,19 @@ object ConcurrentMapSpec extends ZIOBaseSpec {
             getRes <- map.get(1)
           } yield assert(repRes)(isFalse) && assert(getRes)(isSome(equalTo("a")))
         }
-      )
+      ),
+      testM("retainIf") {
```
mijicd, line 251: I'd wrap it in a suite just to be consistent with the remaining methods.

mijicd (commented):
It's getting there, a few tweaks and it should be ready to go!

P.S. Please run `sbt fmt` to make sure the sources are properly formatted.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -39,6 +92,14 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
   def put(key: K, value: V): UIO[Option[V]] =
     UIO(Option(underlying.put(key, value)))
 
+  /**
+   * Adds all new key-value pairs
+   */
+  def putAll(keyValues: (K, V)*): UIO[Unit] =
+    UIO(keyValues.foreach { case (k: K, v: V) =>
```
regiskuckaertz, line 99: It would be great if we could use `putAll` here. Would this work: `underlying.putAll(keyValues.toMap.asJava)`?

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
```diff
@@ -58,6 +119,30 @@ final class ConcurrentMap[K, V] private (private val underlying: ConcurrentHashM
...
     UIO(underlying.remove(key, value))
 
+  /**
+   * Removes all elements which do not satisfy the given predicate.
+   */
+  def removeIf(p: (K, V) => Boolean): UIO[Unit] =
+    UIO(
+      underlying.forEach { (k: K, v: V) =>
+        if (p(k, v)) {
+          val _ = underlying.remove(k)
```
regiskuckaertz, line 129: The only thing that I'm not sure of: can we update the collection while we are iterating over its elements?

regiskuckaertz (commented):
Beautiful @oliveiosello ! I just have two questions but otherwise it looks good to me.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
oliveiosello: It seems to pass the tests. Do you have any ideas as to how we can rewrite it to make it more atomic?

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
oliveiosello: I was trying to avoid using JavaConverters (https://github.com/zio/zio/pull/5933#discussion_r739778002). I see we used it in the methods on ConcurrentSet, so I switched it over and just added a silencer.

core/shared/src/main/scala/zio/concurrent/ConcurrentMap.scala
regiskuckaertz: I don't think it is possible, so we'll have to live with it. Worst case we remove those methods, if they prove to create more issues than they solve.

oliveiosello: Okay, we'll see how it goes. Happy to remove it if needed.

regiskuckaertz (approved):

oliveiosello:
@regiskuckaertz Sorry, there was just one minor fix it looked like. All is good to go now!

mijicd (approved):
Thanks for taking care of this!

PR #5934, opened 2021-10-31
[wip] concurrent primitives

fwbrasil:


core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.ConcurrentLinkedQueue
+import scala.annotation.tailrec
+import zio.Synchronizer._
+
+final class CountDownLatch(count: Int, maxWaiters: Option[Int]) {
+  private[this] val sync = Synchronizer(count, maxWaiters)
+
+  // TODO is it possible to evaluate once and reuse the Promise?
+  private[this] val awaitZero = sync.acquire(count)
```
fwbrasil, line 12: `awaitZero` is an `UIO` so I imagine this might not work because the promise will be created on every evaluation?

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  def await: UIO[Unit] = awaitZero.asInstanceOf[UIO[Unit]]
```
fwbrasil, line 14: I think we can suppress the `Full` error here but I'm not sure how to do that

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+      } else {
+        sem(ZIO.effectSuspendTotal {
+          owner = fiberId
+          zio.ensuring(ZIO.effectTotal(owner = null))
+        })
+      }
+    }
+}
+
+final class Synchronizer(initialPermits: Int, maxWaiters: Int) {
```
fwbrasil, line 52: `acquire` and `release` are missing validations in case the caller misuses the API. I think this class should be `private[zio]`, though

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+  private[this] val sync = Synchronizer(initialPermits, maxWaiters)
+
+  def apply[R, E, A](zio: ZIO[R, E, A]): ZIO[R, Either[E, Full], A] =
+    sync
+      .acquire(1)
+      .mapError(Right(_))
+      .flatMap { _ =>
+        zio
+          .mapError(Left(_))
+          .ensuring(sync.release(1))
```
fwbrasil, line 30: is it possible that the computation is cancelled after `acquire` but before it reaches this point?

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+  private[this] val sem = new AsyncSemaphore(1, maxWaiters)
+
+  def apply[R, E <: Throwable, A](zio: ZIO[R, E, A]): ZIO[R, Either[E, Full], A] =
+    ZIO.effectSuspendTotalWith { (_, fiberId) =>
+      if (owner == fiberId) {
+        zio.mapError(Left(_))
+      } else {
+        sem(ZIO.effectSuspendTotal {
+          owner = fiberId
+          zio.ensuring(ZIO.effectTotal(owner = null))
```
fwbrasil, line 46: I'm not sure about cancellation behavior here as well

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+}
+
+final class Synchronizer(initialPermits: Int, maxWaiters: Int) {
+
+  // if >= 0, # of permits
+  // if < 0, # of waiters
+  private[this] val state = new AtomicInteger(initialPermits)
+
+  // TODO requires multiple allocations per waiter
+  private[this] val waiters = new ConcurrentLinkedQueue[(AtomicInteger, Promise[Full, Unit])]
```
regiskuckaertz, line 59: Do you need to wrap the number of permits in an `AtomicInteger` here?

core/shared/src/main/scala/zio/ZIOConcurrent.scala
regiskuckaertz, line 52: How about keeping track of state as follows:

```scala
type Permits = Int
type Waiter = (Permits, Promise[Nothing, Unit])
type State = (Permits, ConcurrentLinkedQueue[Waiter])

class Synchronizer(state: Ref[State], maxWaiters: Int)
```

The [`Ref` datatype](https://zio.dev/datatypes/concurrency/ref) encapsulates an atomic reference and I _think_ is still extremely fast.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  def acquire(permits: Int): IO[Full, Unit] =
```
regiskuckaertz, line 61: If you like the idea of using `Ref`, your logic can stay the same, except for the loop that is taken care of by `Ref` itself. Something like this:
```scala
def acquire(permits: Permits): IO[Full, Unit] =
  IO.effectSuspendTotalWith { (_, fiberId) =>
    state.modify {
      case (available, waiters) if available >= permits =>
        IO.unit -> (available - permits, waiters)
      case (available, waiters) if waiters.size == maxWaiters =>
        Synchronizer.full -> (available, waiters)
      case (available, waiters) => 
        val p = Promise.unsafeMake[Nothing, Unit](fiberId)
        val rem = permits - available
        p.await -> (0, waiters.add((rem, p)))
    }.flatten
  }
```

`modify` is a method that allows you to atomically update the state and produce a result. Here the result is an IO action that must follow the update of the state. Technically this is not 100% atomic because the result of `modify` is `UIO[IO[Full, Unit]]` and so there is the flatmapping that needs to happen after that, so this can definitely be improved.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+            p.await
+          } else {
+            loop()
+          }
+        }
+      }
+      loop()
+    }
+
+  def release(permits: Int): UIO[Unit] =
```
regiskuckaertz, line 90: Here too, if using the `Ref` with the `modify` trick, although you will still need a loop to dequeue as many waiters as possible.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
```diff
@@ -0,0 +1,153 @@
...
+      loop(permits)
+    }
+}
+
+object Synchronizer {
+  case class Full() extends Exception
+  private val full = IO.fail(Full())
+
+  def apply(permits: Int, maxWaiters: Option[Int]) =
+    new Synchronizer(permits, maxWaiters.getOrElse(Int.MaxValue))
```
regiskuckaertz, line 152: Finally this would need a small change to
```scala
for {
  waiters <- UIO(new ConcurrentLinkedQueue[Waiter])
  state   <- Ref.make[State]((permits, waiters))
} yield new Synchronizer(state, maxWaiters.getOrElse(Int.MaxValue))
```

regiskuckaertz (commented):
I hope my comments are helpful. I think once you get `Synchronizer` right then the downstream structures will be trivial to implement.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
regiskuckaertz: Also if wrapped in a `Ref`, I'm not sure if `ConcurrentLinkedQueue` is still necessary.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
adamgfraser: I think we would want the creation of the `CountDownLatch` to be an effect (e..g `def make: UIO[CountDownLatch]`) because it allocates mutable state and then we could create the `Promise` as part of creating the `CountDownLatch`.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
adamgfraser: Well we need to signal the error somehow so the caller can know whether the latch has been completed and they can proceed or the latch rejected them even trying to wait. I think changing the return type to `IO[Unit, Unit]` might be the most idiomatic way to do this. We have generally shied away form putting bounds on the number of waiters just because it does push a degree of complexity onto users that can be hard to reason about but I hear your point about unbounded data structures being potential leaks.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
adamgfraser: Yes interruption can in general be observed between every continuation. You can prevent interruption from being observed in a particular region by using `uninterruptibleMask` though that does have some performance cost. If you do a bunch of things within a single effect like `effectSuspendTotal` below interruption cannot be observed within that so that can be another alternative.

core/shared/src/main/scala/zio/ZIOConcurrent.scala
adamgfraser: The guarantee of `ensuring` is only that the finalizer will be run if `zio` begins executing but interruption can potentially be observed between `apply` and `effectSuspendTotal` and between `effectSuspendTotal` and `zio`.

adamgfraser (commented):
@fwbrasil There is a lot of good stuff here!

This is definitely tricky stuff so would be great to add some tests for this as well as benchmarks eventually to confirm that the performance benefit is worth the additional complexity versus a simpler structure like a `Ref` and a `Promise`.

Would be happy to work with you more on this if you would like to push this forward.

jdegoes:
These all moved over to _ZIO Concurrent_. Please re-open there if you have the time and interest. Thanks for your work on this! 🙏

PR #5936, opened 2021-10-31
Implement the fromInputStream constructors

asakaev:


vigoo:
Nice work! Could you also add the other `fromInputStream` variants I mentioned in the issue, and the rename of the `fromReaderEffect`?

asakaev:
Thanks! I'll do this, but I suggest making each function a separate PR. What do you think, @vigoo?

vigoo:
I think that would be an overkill because the rest of them are just simple variants calling this implementation. Let's just keep adding them here!

vigoo (approved):

PR #5941, opened 2021-10-31
add padToMod in Chunk of bits

helanto:
Deals with requirement 7 of #5050

### DIFF
--- a/core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
@@ -11,9 +11,7 @@ object BitChunkSpec extends ZIOBaseSpec {
       bytes <- Gen.listOf(Gen.anyByte)
     } yield Chunk.fromIterable(bytes)
 
-  val genBitChunk: Gen[Random with Sized, Chunk[Boolean]] = for {
-    bits <- Gen.listOf(Gen.boolean)
-  } yield Chunk.fromIterable(bits)
+  val genBitChunk: Gen[Random with Sized, Chunk[Boolean]] = genByteChunk.map(_.asBits)
 
   val genInt: Gen[Random with Sized, Int] =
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
```diff
@@ -11,6 +11,10 @@ object BitChunkSpec extends ZIOBaseSpec {
       bytes <- Gen.listOf(Gen.anyByte)
     } yield Chunk.fromIterable(bytes)
 
+  val genBitChunk: Gen[Random with Sized, Chunk[Boolean]] = for {
```
adamgfraser, line 14: Need to use `asBits` here to construct a `BitChunk`.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -118,6 +118,23 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
...
+   * Pads to the specified number of bits.
+   *  - if n > size, chunk is 0-padded to length n.
+   *  - if n <= size, chunk is left intact.
+   *
+   * @example
+   * {{{
+   *  Chunk(true, false).padToMod(8) => Chunk(true, false, false, false, false, false, false, false)
+   * }}}
+   */
+  def padToMod(n: Int)(implicit ev: A <:< Boolean): Chunk[Boolean] =
```
adamgfraser, line 131: This needs to have a lower level implementation on the `BitChunk` data types to have the performance we are looking for.

adamgfraser (changes requested):
See comments below about using the `BitChunk` data type, which provides a more optimized representation of a chunk of bits than just a `Chunk[Boolean]`.

jdegoes:
@helanto Will close for now. Please re-open if you get a chance to push this through!

PR #5943, opened 2021-10-31
BitChunk binary operations

loathwine:
Deals with requirement 5 of [#5050](https://github.com/zio/zio/issues/5050).
I see some problems with this solution that I would like help with.

1. Chunk.BitChunk is now exposed in `Chunk#asBits`, I'm not even sure if that will work since `BitChunk` is private.
**Update**: I added the bitwise operations to `Chunk[Boolean]` and override them in `Chunk.BitChunk`. `Chunk.BitChunk` is now hidden again.
2. I added a `Chunk.fromBits(bits: Boolean*): Chunk.BitChunk` constructor that assumes big endian bit-endianness. This constructor was mostly for tests, maybe it should be moved to `BitChunkSpec`?
**Update**:  I moved the `fromBits` constructor to `BitChunkSpec`.

### DIFF
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -97,7 +97,7 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
   /**
    * Converts a chunk of bytes to a chunk of bits.
    */
-  final def asBits(implicit ev: A <:< Byte): Chunk[Boolean] =
+  final def asBits(implicit ev: A <:< Byte): Chunk.BitChunk =
     Chunk.BitChunk(self.map(ev), 0, length << 3)
 
   /**
@@ -1112,6 +1112,9 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
   override def apply[A](as: A*): Chunk[A] =
     fromIterable(as)
 
+  def fromBits(bits: Boolean*): Chunk.BitChunk =
+    fromIterator(bits.reverse.grouped(8).map(_.foldRight(0)((b, i) => (i << 1) | (if (b) 1 else 0)).toByte)).asBits
+
   /**
    * Returns a chunk backed by an array.
    *
@@ -1763,6 +1766,17 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
     override private[zio] def reverseArrayIterator[A1 >: Boolean]: Iterator[Array[A1]] =
       arrayIterator
 
+    def &(that: BitChunk): BitChunk = binaryOp(that)(_ => 0, _ => 0)((a: Byte, b: Byte) => (a & b).toByte)
+
+    def |(that: BitChunk): BitChunk = binaryOp(that)(identity, identity)((a: Byte, b: Byte) => (a | b).toByte)
+
+    def ^(that: BitChunk): BitChunk = binaryOp(that)(identity, identity)((a: Byte, b: Byte) => (a ^ b).toByte)
+
+    def unary_~ : BitChunk =
+      BitChunk(bytes.mapChunk(b => (~b).toByte), minBitIndex, maxBitIndex)
+
+    private def binaryOp(that: BitChunk)(left: Byte => Byte, right: Byte => Byte)(both: (Byte, Byte) => Byte): BitChunk =
+      BitChunk(this.bytes.zipAllWith(that.bytes)(left, right)(both), this.minBitIndex.min(that.minBitIndex), this.maxBitIndex.max(that.maxBitIndex))
   }
 
   private case object Empty extends Chunk[Nothing] { self =>
--- a/core/shared/src/main/scala/zio/Chunk.scala
+++ b/core/shared/src/main/scala/zio/Chunk.scala
@@ -1711,15 +1711,22 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
     override val length: Int =
       maxBitIndex - minBitIndex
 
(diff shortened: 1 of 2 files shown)


jdegoes:
> Chunk.BitChunk is now exposed in Chunk#asBits, I'm not even sure if that will work since BitChunk is private.
> Update: I added the bitwise operations to Chunk[Boolean] and override them in Chunk.BitChunk. Chunk.BitChunk is 
> now hidden again.

This is the correct way to handle it: add it as a method on `Chunk` by requiring implicit evidence `A <:< Boolean`, with a slow default implementation, and then override it in `BitChunk` to have the high-efficiency behavior.

> I added a Chunk.fromBits(bits: Boolean*): Chunk.BitChunk constructor that assumes big endian bit-endianness. This 
> constructor was mostly for tests, maybe it should be moved to BitChunkSpec?

I believe you will get this from `Chunk(true, false, true, true)`. If not, we should fix that.

However, an explicit method could be useful if it also supported:

```scala
object Chunk {
  sealed trait Endianness
  object Endianness {
    case object Big extends Endianness 
    case object Little extends Endianness 
  }

  ...
  def fromBits(endianness: Endianness)(bits: Boolean*): Chunk[Boolean] = ???
}
```

Edvin-san:
> > I added a Chunk.fromBits(bits: Boolean*): Chunk.BitChunk constructor that assumes big endian bit-endianness. This
> > constructor was mostly for tests, maybe it should be moved to BitChunkSpec?
> 
> I believe you will get this from `Chunk(true, false, true, true)`. If not, we should fix that.

The issue is more about when converting an inefficient `Chunk[Boolean]` or `Iterator[Boolean]` to a `Chunk.BitChunk` that is backed by a `Chunk[Byte]`. For example, `Chunk(true, true, true, true, true, true, true, true, false)` (9 elements) should be backed by two bytes, and depending on how we implement that conversion it will be `Chunk(0xFF, 0x00)` or `Chunk(0x01, 0xFE)`. Maybe endianness isn't the correct term here, and instead whether we pad a sequence of bits with zeroes on the right or on the left. E.g. is the underlying byte of `Chunk(true, true, false, true)` equal to `0b11010000` or `0b00001101`?

However, I feel like this is more related to the 4th requirement:
> A "destructor" (which can go from Chunk[Boolean] to Chunk[Byte], Chunk[Int], or Chunk[Long]). When going to either Int or Long (but not when going to Chunk[Byte]), the destructor must accept Endianness, because it affects how the bytes are packed into the resulting int or long values.

I wouldn't mind working on that one, but probably it is out of scope for this one. For now, I've explicitly set the conversions in `BitChunkSpec`.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -1763,6 +1775,32 @@ object Chunk extends ChunkFactory with ChunkPlatformSpecific {
...
+
+    override def ^(that: Chunk[Boolean])(implicit ev: Boolean <:< Boolean): Chunk[Boolean] = that match {
+      case thatBitChunk: BitChunk => byteWiseOp(thatBitChunk)(identity, identity)((a: Byte, b: Byte) => (a ^ b).toByte)
+      case that                   => super.^(that)
+    }
+
+    override def unary_~(implicit ev: Boolean <:< Boolean): BitChunk =
+      BitChunk(bytes.mapChunk(b => (~b).toByte), minBitIndex, maxBitIndex)
+
+    private def byteWiseOp(
```
adamgfraser, line 1796: I don't think this is going to correctly handle cases where one of the chunk has been sliced so the bit indices don't line up. Can we add some tests for that?

adamgfraser (changes requested):
Looks like you are going in the right direction here! I don't think this will correctly handle cases where the bit indices don't line up but like how you are implementing things in terms of low level operators and keeping implementation details private.

core/shared/src/main/scala/zio/Chunk.scala
Edvin-san: You're right, I've added tests.

Edvin-san:
When investigating slicing issues I think I found an issue with the apply-method. It didn't consider `minBitIndex` so the following test failed:
```scala
testM("drop then apply") {
  check(genByteChunk.filterNot(_.isEmpty)) { bytes =>
    val bitChunk = bytes.asBits
    val dropped1 = bitChunk.drop(1)
    assert(dropped1(0))(equalTo(bitChunk(1)))
  }
}
```

I added the test and a fix of the implementation.

I'm not completely happy with my fix to the slicing issues, but the idea is that I try to keep two invariants that allow a simple  byte zipping.

1. `0 <= minBitIndex < 8`, and `maxBitIndex` refers to a bit in the final byte (+1). I think this invariant already existed.
2. For all i: `i < minBitIndex || i >= maxBitIndex => accessBit(i) == 0`

This shifts some work to `take` and `drop` to maintain the second invariant, but it simplifies the bitwise operations. However, I put in a fallback so we don't even try to do an intelligent bitwise join when `minBitIndex` are not equal (byte chunks do not align). I wonder if we can use the [shift operators (#5945)](https://github.com/zio/zio/pull/5945) to align chunks?

Edvin-san:
Any advice how to progress with this one?

adamgfraser:
@Edvin-san Will take a look later today!

jdegoes:
@Edvin-san This has bit-rotted, so will close for now. Note that ZIO 2 has some bitwise operations already, although they could be improved. Please re-open if you want to extend the existing functionality or tests in any way. Thank you!

PR #5945, opened 2021-10-31
Implement shift left and right in Chunk[Boolean]

koukoustath:
number 6 of issue: #5050

### DIFF
--- a/core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
@@ -11,6 +11,10 @@ object BitChunkSpec extends ZIOBaseSpec {
       bytes <- Gen.listOf(Gen.anyByte)
     } yield Chunk.fromIterable(bytes)
 
+  val genBitChunk: Gen[Random with Sized, Chunk[Boolean]] = for {
+    bits <- Gen.listOf(Gen.boolean)
+  } yield Chunk.fromIterable(bits)
+
   val genInt: Gen[Random with Sized, Int] =
     Gen.small(Gen.const(_))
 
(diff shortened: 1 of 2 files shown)


core-tests/shared/src/test/scala/zio/BitChunkSpec.scala
```diff
@@ -11,6 +11,10 @@ object BitChunkSpec extends ZIOBaseSpec {
       bytes <- Gen.listOf(Gen.anyByte)
     } yield Chunk.fromIterable(bytes)
 
+  val genBitChunk: Gen[Random with Sized, Chunk[Boolean]] = for {
```
adamgfraser, line 14: To construct a `BitChunk` you need to call `asBits` on a `Chunk[Byte]`.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -118,6 +118,26 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
   def char(index: Int)(implicit ev: A <:< Char): Char =
     ev(apply(index))
 
+  /**
+   * Shift left the specified number of bits.
+   */
+  def <<(n: Int)(implicit ev: A <:< Boolean): Chunk[Boolean] =
```
adamgfraser, line 124: We want to provide optimized versions of these operators on `BitChunk`.

core/shared/src/main/scala/zio/Chunk.scala
```diff
@@ -118,6 +118,26 @@ sealed abstract class Chunk[+A] extends ChunkLike[A] { self =>
...
+    if (n >= length) Chunk.fill(length)(false)
+    else {
+      val paddedBits = Chunk.fill(n)(false)
+      self.map(ev).drop(n) ++ paddedBits
+    }
+
+  /**
+   * Shift right the specified number of bits.
+   */
+  def >>(n: Int)(implicit ev: A <:< Boolean): Chunk[Boolean] =
```
adamgfraser, line 134: Would be nice to provide a version of `>>>` as well.

adamgfraser (changes requested):
Looks like you are off to a good start here but there is more work you are going to need to use the `BitChunk` data type and provide an optimized implementation of these operators.

koukoustath:
@adamgfraser 
Has this functionality been implemented? 

I see that the related [issue](https://github.com/zio/zio/issues/5050) has been closed.

Just asking to close the PR if it is already implemented. Thanks

jdegoes:
@koukoustath This does not look completed. We need definitions of left/right shift in `Chunk[Boolean]`, which do not exist, and then we need overrides on them for the packed classes, which also do not exist.

Please re-open if you want to continue your work here, and thanks for the PR!

PR #5950, opened 2021-10-31
Add possibility to feed UUIDs to TestRandom

Edvin-san:
I tried to use `TestRandom.feedUUIDs` but it was missing!

image


adamgfraser (approved):
Thank you! 🙏

PR #5953, opened 2021-10-31
Custom Smart Assertions

kitlangton:
Addresses https://github.com/zio/zio/issues/5466

Example:

``` scala
// Definition
implicit final class DoubleOps(private val self: Double) extends AnyVal {
  def ~=(that: Double): SmartAssertion[Boolean] =
    SmartAssertion.cond(self == that, s"THIS AIN'T RIGHT $self !~= $that")

  def asEven: SmartAssertion[Double] =
    SmartAssertion.fromEither(Either.cond(self % 2 == 0, self, s"$self must be even."))
}

// Usage
    test("custom assertions") {
      val double = 12.5
      assertTrue(double ~= 12.8)
    },
    test("custom assertions") {
      val double = 12.3
      assertTrue(double.asEven * 3 == 13.5)
    }
```

### DIFF
--- a/test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
+++ b/test-tests/shared/src/test/scala/zio/test/SmartAssertionSpec.scala
@@ -7,7 +7,6 @@ import zio.{Chunk, NonEmptyChunk}
 
 import java.time.LocalDateTime
 import scala.collection.immutable.SortedSet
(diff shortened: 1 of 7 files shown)


fsvehla:
I’m very much looking forward to this, @kitlangton!

kitlangton:
Closing in favor of #5996

PR #5963, opened 2021-11-01
Replace UIO[Unit] with UIO[Any]

mijicd:
### Summary

- Removes Unit from ConcurrentSet responses.
- Fixes the scala 3 compilation issue introduced in #5921 (no implicit conversions for predicates).

### DIFF
--- a/core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
+++ b/core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
@@ -91,7 +91,7 @@ final class ConcurrentSet[A] private (private val underlying: ConcurrentHashMap.
   def retainIf(p: A => Boolean): UIO[Boolean] =
     UIO(underlying.removeIf((t: A) => p(t)))
 
-  def clear: UIO[Unit] =
+  def clear: UIO[Any] =
     UIO(underlying.clear())
 
   def contains(x: A): UIO[Boolean] =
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/concurrent/ConcurrentSet.scala
```diff
@@ -110,10 +110,10 @@ final class ConcurrentSet[A] private (private val underlying: ConcurrentHashMap.
     UIO(underlying.asScala.toSet: @silent("JavaConverters"))
 
   @silent("JavaConverters")
-  def transform(f: A => A): UIO[Unit] = UIO {
+  def transform(f: A => A): UIO[Any] = UIO {
```
mijicd, line 113: I didn't touch the existing implementation as it's out of this PR scope, but I think we'll either have to make it atomic, or state the non-atomicity in its scaladoc.

adamgfraser:
I think our latest convention is that we should accept `Any` but return `Unit`. Basically accept the most general type possible and return the most specific type possible. There has definitely been some discussion about this as it can have some performance implication but I think there is some ongoing work that is going to special case `ZIO.unit` in the runtime to make that operation extremely fast.

mijicd:
@adamgfraser Then this whole package needs restructuring. I'd still merge this one and restructure in the follow-up, as the current implementation breaks compilation.

adamgfraser (approved):
Sounds good.

Issue #5918, opened 2021-10-30
ZIO Streams 2.0: splitLines, splitOn, splitOnChunk


vigoo:
Implement the `splitLines`, `splitOn` and `splitOnChunk` transducers from the ZIO Streams 1.0 as ZIO Streams 2.0 `ZPipeline`s

NavidJalali:
I'd like to pick this up if no one is doing it

NavidJalali:
This can be resolved now! 🥳

adamgfraser:
@NavidJalali Great work on this!

PR #5969, opened 2021-11-02
Add splitLines on ZPipeline

NavidJalali:
Part 1/3 for fixing https://github.com/zio/zio/issues/5918


adamgfraser (approved):
Great! This kind of "pipeline first" implementation is exactly what we want.

PR #5972, opened 2021-11-02
Implement RuntimeConfigFlags

### LINKED ISSUE #5880: ZIO 2.0: Deal with boolean blindness in RuntimeConfig
`RuntimeConfig` has a number of boolean fields. These can be tricky to get right and are not very stable in the presence of refactorings. Rather than continue this pattern of introducing more boolean flags, we can refactor to an alternate pattern: sets of case objects.

For example, instead of:

```scala
def foo(x: Boolean, y: Boolean, z: Boolean)
```

We can have:

```scala
final case class FooFlags(set: Set[FooFlag]) {
  def <> (that: FooFlags): FooFlags = ???
}
object FooFlags {
  val x: FooFlags = FooFlags(Set(FooFlag.X))
  val y: FooFlags = FooFlags(Set(FooFlag.X))
  val z: FooFlags = FooFlags(Set(FooFlag.X))
}
sealed trait FooFlag
object FooFlag {
  case object X extends FooFlag
  case object Y extends FooFlag
  case object Z extends FooFlag
}
```

Then the API becomes:

```scala
def foo(flags: FooFlags) = ???

foo(FooFlags.x <> FooFlags.y <> FooFlags.z)
```

In this case, we can have:

```scala
final case class RuntimeConfigFlags(...)
object RuntimeConfigFlags {
  val logRuntime: RuntimeConfigFlags = ...
}

sealed trait RuntimeConfigFlag
object RuntimeConfigFlag {
  case object ...
}
```

adamgfraser:
Resolves #5880.

### DIFF
--- a/core/shared/src/main/scala/zio/RuntimeConfigFlags.scala
+++ b/core/shared/src/main/scala/zio/RuntimeConfigFlags.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2021 John A. De Goes and the ZIO Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+final case class RuntimeConfigFlags(flags: Set[RuntimeConfigFlag]) {
+
+  def +(flag: RuntimeConfigFlag): RuntimeConfigFlags =
+    RuntimeConfigFlags(flags + flag)
+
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/zio/RuntimeConfigFlags.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package zio
+
+final case class RuntimeConfigFlags(flags: Set[RuntimeConfigFlag]) {
+
+  def +(flag: RuntimeConfigFlag): RuntimeConfigFlags =
+    RuntimeConfigFlags(flags + flag)
+
+  def apply(runtimeConfigFlag: RuntimeConfigFlag): Boolean =
```
jdegoes, line 24:
```suggestion
  def isEnabled(runtimeConfigFlag: RuntimeConfigFlag): Boolean =
```

jdegoes (commented):
Slight preference for a rename, but otherwise, looks fantastic!

core/shared/src/main/scala/zio/RuntimeConfigFlags.scala
adamgfraser: I like it!

adamgfraser:
Done!

jdegoes (approved):

PR #5973, opened 2021-11-02
Make Cause#traced Additive

adamgfraser:


kitlangton (approved):
👍

core/shared/src/main/scala/zio/Cause.scala
```diff
@@ -528,10 +528,10 @@ sealed abstract class Cause[+E] extends Product with Serializable { self =>
       .reverse
 
   /**
-   * Replaces traces with the specified trace.
+   * Adds the specified execution trace to traces.
    */
   final def traced(trace: ZTrace): Cause[E] =
-    mapTrace(_ => trace)
+    mapTrace(_ ++ trace)
```
jdegoes, line 534: I like this, but now `.untraced` needs to be rewritten as `mapTrace(_ => ZTrace.none)` (and maybe good to add a test for `untraced` to make sure it clears the trace so this won't break in the future).

adamgfraser: Done now!

jdegoes (approved):

Issue #5878, opened 2021-10-29
ZIO 1.x: Backport RuntimeConfigAspect.trackFiberRef


jdegoes:
See [here](https://github.com/zio/zio/issues/4528#issuecomment-955061100) for context.

To backport to ZIO 1.x, we will need to add this as a `Supervisor` inside the companion object of `Supervisor`:

```scala
object Supervisor {
  def trackFiberRef[A](fiberRef: FiberRef[A])(onResume: A => Unit): Supervisor[Unit] = ???
}
```

The usage of this will be slightly different than the usage of an ordinary supervisor. But still, it will allow users to do the same things. 

/cc @dkarlinsky

PR #5980, opened 2021-11-02
issue #5878 - Backport ThreadLocalBridge

dkarlinsky:
Back-porting `ThreadLocalBridge` from ZIO 2
ZIO 2 PR: https://github.com/zio/zio/pull/5907

### DIFF
--- a/core-tests/shared/src/test/scala/zio/SupervisorSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/SupervisorSpec.scala
@@ -22,12 +22,10 @@ object SupervisorSpec extends ZIOBaseSpec {
             assertTrue(
               recorded.toSet == (Set(
                 s"unsafeOnResume($fiber1)",
-                s"unsafeOnResume($fiber2)"
-              ).filter(_ => superviseOperations) ++
-                Set(
-                  s"unsafeOnSuspend($fiber1)",
-                  s"unsafeOnSuspend($fiber2)"
-                ))
+                s"unsafeOnResume($fiber2)",
+                s"unsafeOnSuspend($fiber1)",
+                s"unsafeOnSuspend($fiber2)"
+              ))
             )
           }
         }
--- a/core-tests/shared/src/test/scala/zio/ThreadLocalBridgeSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ThreadLocalBridgeSpec.scala
@@ -1,0 +1,87 @@
+package zio
+
+import zio.ThreadLocalBridge.TrackingFiberRef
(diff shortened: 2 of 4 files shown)


core/shared/src/main/scala/zio/Supervisor.scala
```diff
@@ -216,6 +216,28 @@ object Supervisor {
...
+   *
+   * *IMPORTANT*: Requires `Platform.superviseOperations` enabled to work:
+   *
+   * {{{
+   *    val runtime = ???
+   *    val (supervisor, fiberRef) = Supervisor.trackFiberRefUnsafeMake(initialValue)
+   *    runtime.mapPlatform(_.withSuperviseOperations(true).withSupervisor(supervisor)).
+   * }}}
+   */
+  def trackFiberRef[A](initialValue: A)(link: A => Unit): UIO[(Supervisor[Unit], TrackingFiberRef[A])] =
```
dmytr, line 233: Maybe this is not really related to this PR, but I just wanted to get some opinions regarding the following:

I'm wondering if it would be possible to enable users to track all events from `Fiber`'s lifecycle: start, resume, suspend, end? Ideally, providing `Fiber` identifier to callbacks.

This would allow users to add interoperability with libraries that utilize `ThreadLocal`s. One prominent example is automatic instrumentation by OpenTelemetry agent. It provides a lot of value out of the box, but there is no way to pick up current context when `Fiber`s are started by HTTP server, for example.

I made a proof of concept supervisor that enables such interoperability: https://github.com/dmytr/zio-otel-instrumentation/blob/main/src/main/scala/zio/OpenTelemetrySupervisor.scala. It tracks OpenTelemetry context on per-fiber base, so it needs to have a hold on some kind of `Fiber` identifier. Unfortunately, it relies on ZIO's package-private APIs making it usable by very brave people only :joy: 

Maybe, if `initialValue`'s type is changed to `=> A` and it gets evaluated in `Supervisor.unsafeOnStart` we could achieve this context capturing capability? And `A` could be a user-defined unique identifier for that particular `Fiber`.

I guess, ideally, API would look like this:
```scala
def trackFiberLifecycle(
  onStart: => A, 
  onEnd: A => Unit,
  onSuspend: A => Unit,
  onResume: A => Unit
): Supervisor[Unit]
```

adamgfraser: @dmytr I would take a look at the functionality for this in ZIO 2.0. I think it provides basically what you are describing, though I do think we need to implement a constructor to use it from third party code.

adamgfraser (approved):

PR #5988, opened 2021-11-02
WIP: .is SmartAssertions

kitlangton:


test/shared/src/main/scala/zio/test/package.scala
```diff
@@ -29,25 +29,23 @@ import scala.util.Try
 /**
  * _ZIO Test_ is a featherweight testing library for effectful programs.
  *
- * The library imagines every spec as an ordinary immutable value, providing
```
adamgfraser, line 32: Stray formatting changes?

kitlangton: Hrm. Yeah. Weird. Scalaformat seems to have changed all the comments. Will undo.

adamgfraser: Yeah I think it tries to make them longer but I think it is nice to keep them shorter if possible for viewing on mobile devices.

adamgfraser (approved):

PR #5990, opened 2021-11-03
Reimplemented text decoders as ZPipelines.

### LINKED ISSUE #5869: ZIO Streams 2.0: Implement character decoding as pipelines
Currently in ZIO Streams 2.0 we have a family of character decoding _sinks_, such as `ZSink#utfDecode`. We also have a new kind of streaming primitive called `ZPipeline` which is essentially a stream transformer.

We would like to have character decoding defined as a `ZPipeline` so it can be applied to any byte stream and convert it to a stream of characters.

Note that this is partially possible by using `transduce` and the existing sinks, but it is not correct for cases requiring to maintain state over the whole stream, such as `utfDecode`. Transducing that sink repeats the `take(4)` that determines the BOM for each chunk which is incorrect.

ithinkicancode:
Resolves #5869


ithinkicancode:
TODO's:

- [ ] Implement utf8 decoders
- [ ] Implement usASCIIDecode
- [ ] Implement iso_8859_1Decode
- [ ] Migrate unit tests from ZSinkSpec to ZPipelineSpec
- [ ] Clean up/delete decoders from ZSink
- [ ] Add scaladoc

adamgfraser:
@ithinkicancode Looks like there is a failing test.

streams/shared/src/main/scala/zio/stream/experimental/ZPipeline.scala
```diff
@@ -75,7 +78,7 @@ object ZPipeline extends ZPipelineCompanionVersionSpecific {
 
   type Identity[A] = A
 
-  def branchAfter[LowerEnv, UpperEnv, LowerErr, UpperErr, LowerElem, UpperElem, OutElem0[Elem]](n: Int)(
+  def branchAfter[LowerEnv, UpperEnv, LowerErr, UpperErr, LowerElem, UpperElem, OutElem[Elem]](n: Int)(
```
adamgfraser, line 81: You can delete this type parameter entirely.

streams/shared/src/main/scala/zio/stream/experimental/ZPipeline.scala
```diff
@@ -273,6 +276,9 @@ object ZPipeline extends ZPipelineCompanionVersionSpecific {
         stream
     }
 
+  def iso_8859_1Decode: UtfDecodingPipeline =
```
adamgfraser, line 279: I understand why you did it but I think this type alias may ultimately make things less accessible when we have a lot of different pipelines with different types. I would suggest just going back to the normal type parameters.

streams/shared/src/main/scala/zio/stream/experimental/ZPipeline.scala
```diff
@@ -731,4 +826,241 @@ object ZPipeline extends ZPipelineCompanionVersionSpecific {
         type Out[In] = RightOut
       }
   }
+
+  private def textDecodeUsing(charset: Charset): UtfDecodingPipeline =
```
adamgfraser, line 830: Love the low level implementations here!

adamgfraser (changes requested):
A couple of minor suggestions. Otherwise good to go when tests are passing!

streams/shared/src/main/scala/zio/stream/experimental/ZPipeline.scala
ithinkicancode: Will remove this type alia.

adamgfraser (approved):

PR #5995, opened 2021-11-03
Simplify ZIO Test Package Structure

adamgfraser:
Deletes the `zio.test.environment` package and moves its contents into the `zio.test` package. This way if users are already importing `zio.test._`, which they typically will for all its test related functionality, they won't have to do any more important to work with the test environment.


kitlangton (approved):
This has been on my secret wish/todo list for so long! Thanks Adam!

PR #5954, opened 2021-10-31
ZIO Streams 2.0: Implement ZSink#zipWithPar #5865

domartynov:
WIP not working

PR #6001, opened 2021-11-04
Do Not Read Ahead In ZChannel#mergeWith

adamgfraser:
Addresses the remaining issue in the implementation of `ZChannel.mergeWith` to allow #5922 and #5954 to be merged.

The key insight here is that it is not safe in general for a channel operator to read from its upstream if it does not have a request to do so from its downstream. The reason is that the downstream may never want to read at all and if the channel operator reads from the upstream it may "steal" an element that someone else wants to consume.

This comes up in `mergeWith` because `mergeWith` is currently implemented in terms of `Bridge` and `SingleProducerAsyncInput`, which basically creates a single element queue and forks a fiber to read from the upstream while the merged channels read from the queue. This results in a situation where the `mergeWith` operator "reads ahead" by one element.

This becomes particularly problematic in cases where you have nested merges. For example, consider:

```scala
readFromUpstreamAndOfferToHub.mergeWith(
  readFromLeftHubSubscriptionAndWriteToDownstream.mergeWith(
    readFromRightHubSubscriptionAndOfferToDownStream
  )
)
```

This seems correct because `readFromUpstreamAndOfferToHub` is the only channel that reads from upstream, so we would expect it to read all the upstream elements, publish them to the hub, and then the two other merged channels to read from the hub subscriptions, process, and write to downstream.

However, what actually happens is that the inner `mergeWith` "reads ahead" by a single element and thus can take one of the elements from the upstream that is supposed to be read by the outer channel. The inner merge doesn't do anything with this element so it just gets lost.

With this change the fiber that is forked in the execution of `Bridge` waits for the downstream to read before executing the upstream channel.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/experimental/internal/ChannelExecutor.scala
+++ b/streams/shared/src/main/scala/zio/stream/experimental/internal/ChannelExecutor.scala
@@ -770,12 +770,12 @@ private[zio] object SingleProducerAsyncInput {
       .flatMap(p => Ref.make[State[Err, Elem, Done]](State.Empty(p)))
       .map(new SingleProducerAsyncInput(_))
 
-  sealed trait State[+Err, +Elem, +Done]
+  sealed trait State[Err, Elem, Done]
   object State {
-    case class Empty(notifyProducer: Promise[Nothing, Unit]) extends State[Nothing, Nothing, Nothing]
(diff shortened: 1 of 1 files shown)


vigoo (dismissed):
Looks good

vigoo (approved):

vigoo (approved):

PR #6002, opened 2021-11-05
add splitOnChunk on ZStream

NavidJalali:
Part 2/3 of fixing https://github.com/zio/zio/issues/5918
Please check the logic, it is NOT THE SAME as the one in ZTransducer.


NavidJalali:
Rare instance of no CI failures 😮

adamgfraser (approved):
Great!

PR #6004, opened 2021-11-05
Port Fix To ZStream#interruptWhen To Channel Encoding

adamgfraser:
In `mergeWith` we are either joining or interrupting the fibers we are forking or explicitly giving the user the choice of how to handle the fibers through the `MergeDecision` returned by the continuation so we should not be using automatic supervision.


regiskuckaertz (approved):
First time reading this code, it's super clear.

PR #6011, opened 2021-11-07
ZIO Streams 2

vigoo:
Removes ZIO Streams 1 and moves out the new implementation from the experimental package

## NOTES
Compression `ZPipeline` specs seems to time out but not sure why - as they were already passing before this PR, my only guess is that now zio-test uses the new stream encoding too and somehow that made it slower.


vigoo:
Not completely ready for merge, see the notes

adamgfraser (approved):

PR #6026, opened 2021-11-10
Experiment with simple tag for Scala 3

kitlangton:


jdegoes:
👀 👀 👀

pshirshov:
Mmmhm. I'm not sure what's going on here. I think it would be lot cheaper to finish missing parts in `izumi-reflect` than to reimplement it (trust me, in the end this implementation would be very close to ours), though I won't mind it in general.

PR #6035, opened 2021-11-13
Backport ZIO 2 ClockJava

ThijsBroersen:
* add java-clock to zio-clock as timekeeper

* extend ClockSpec

### DIFF
--- a/core/shared/src/main/scala/zio/clock/package.scala
+++ b/core/shared/src/main/scala/zio/clock/package.scala
@@ -46,39 +46,45 @@ package object clock {
     }
 
     object Service {
-      val live: Service = new Service {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/clock/package.scala
```diff
@@ -46,46 +46,58 @@ package object clock {
     }
 
     object Service {
-      val live: Service = new Service {
```
adamgfraser, line 49: Let's please not change the implementation of the live service but just provide this as an additional layer.

ThijsBroersen: restored

adamgfraser (approved):
Thank you! 🙏

Issue #6040, opened 2021-11-15
Add scalafix project validation to CI process


swoogles:
Currently, nothing compiles/executes our `scalafix` code.
We should probably make it a submodule in our root `build.sbt`, and execute `scalafmt/test` as one of our CI stages.

adamgfraser:
The disadvantage of this is that it requires any contributor making a change to the ZIO 2.0 API to make the corresponding changes to the Scalafix rule, which may be a somewhat different knowledge base and potentially create a barrier to new contributions.

swoogles:
If we can set things up such that the only command that's executed is `scalafix/test`, then we shouldn't have that problem. It only confirms that transformations we've defined are correctly applied to source code; it's unaffected by broken code in the output directory. 

I know this because we've got all sorts of Stream breakages in there now, but the command still succeeds 😄

If the structure of the scalafix project is too unusual to be a proper submodule in the main project, we could also fall back to just executing that one SBT command as a shell script, since it returns a success code, despite the compilation errors in those other directories.

swoogles:
More simply-
If your API change PR doesn't change anything in the `scalafix` directory, it won't break the ScalaFix part of our build.

adamgfraser:
Sounds great!

PR #6041, opened 2021-11-15
Fix scalafix after ZLayer->ZServiceBuilder PR

swoogles:
Corrections to some errant refactoring for `ZServiceBuilder`.
Also added some migrations around this type.

Created a ticket here to prevent this in the future:
https://github.com/zio/zio/issues/6040

### DIFF
--- a/scalafix/input/src/main/scala/fix/ZIOSpec.scala
+++ b/scalafix/input/src/main/scala/fix/ZIOSpec.scala
@@ -1638,19 +1638,19 @@ object ZIOSpec extends DefaultRunnableSpec {
         } yield assert(res._1)(equalTo(List(0, 2, 4, 6, 8))) && assert(res._2)(equalTo(List(1, 3, 5, 7, 9)))
       }
     ),
-    suite("provideCustomServices")(
+    suite("provideCustomLayer")(
       testM("provides the part of the environment that is not part of the `ZEnv`") {
-        val loggingServiceBuilder: ZServiceBuilder[Any, Nothing, Logging] = Logging.live
+        val loggingLayer: ZLayer[Any, Nothing, Logging] = Logging.live
         val zio: ZIO[ZEnv with Logging, Nothing, Unit]  = ZIO.unit
-        val zio2: URIO[ZEnv, Unit]                      = zio.provideCustomServices(loggingServiceBuilder)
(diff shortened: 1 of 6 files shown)


scalafix/input/src/main/scala/fix/ZIOSpec.scala
```diff
@@ -1638,19 +1638,19 @@ object ZIOSpec extends DefaultRunnableSpec {
         } yield assert(res._1)(equalTo(List(0, 2, 4, 6, 8))) && assert(res._2)(equalTo(List(1, 3, 5, 7, 9)))
       }
     ),
-    suite("provideCustomServices")(
```
swoogles, line 1641: I think in the general case, it would be a mistake to try and rename user test names, so it's more accurate for our in Specs to have an old name and _keep_ the old name in the output.

scalafix/input/src/main/scala/fix/ZIOSpec.scala
```diff
@@ -1638,19 +1638,19 @@ object ZIOSpec extends DefaultRunnableSpec {
...
+    suite("provideCustomLayer")(
       testM("provides the part of the environment that is not part of the `ZEnv`") {
-        val loggingServiceBuilder: ZServiceBuilder[Any, Nothing, Logging] = Logging.live
```
swoogles, line 1643: Our `input` files should still have the old types, since they compile with the old version of ZIO

scalafix/rules/src/main/scala/fix/Zio2Upgrade.scala
```diff
@@ -162,7 +162,11 @@ class Zio2Upgrade extends SemanticRule("Zio2Upgrade") {
 
   val ZIORenames = Renames(
     List("zio.ZIO"),
-    Map("run" -> "exit")
```
swoogles, line 165: This class has the only new migration logic added under this PR

scalafix/input/src/main/scala/fix/ZIOSpec.scala
adamgfraser: 💯

scalafix/input/src/main/scala/fix/ZIOSpec.scala
adamgfraser: I may have been a little aggressive in excising any references to `ZLayer`. 😅

adamgfraser (approved):

PR #6045, opened 2021-11-15
Bring scalafix into root project/process

swoogles:
In addition to Scalafix running in the CI process, now we can edit files in the Scalafix project with navigation, integrated test running, etc. from the top-level ZIO project. Previously, you needed to open a separate editor in the `scalafix` directory.

ScalaFixTestsInIDE


scalafix/rules/src/main/scala/fix/CurriedAssert.scala
```diff
@@ -13,7 +13,7 @@ class CurriedAssert extends SemanticRule("CurriedAssert") {
   override def fix(implicit doc: SemanticDocument): Patch =
     doc.tree.collect {
       case t @ assert(Term.Apply(name, List(value, assertion))) =>
-        Patch.replaceTree(t, name + "(" + value + ")(" + assertion + ")")
```
swoogles, line 16: This has been a warning for a long time, but now that it's part of the full build, we have to fix it to keep things passing.

adamgfraser (approved):

PR #6046, opened 2021-11-15
Delete Has

adamgfraser:


ghostdogpr:
Does that mean that we can't mix traits that have the same methods, generic traits or simple case classes? It was a huge pain before the introduction of Has. Or is there an alternative?

adamgfraser:
You can still do all those things. Essentially the type level map is just "baked in" to ZIO. So you can still access services using the `service` constructor and combine services using the `++` operator just like you do today. You just don't need to wrap them in `Has`.

ghostdogpr:
Sounds fantastic!

core/shared/src/main/scala/zio/Clock.scala
```diff
@@ -24,6 +24,9 @@ import zio.Schedule.Decision._
 import java.lang.{System => JSystem}
 import java.time.{Instant, LocalDateTime, OffsetDateTime}
 import java.util.concurrent.TimeUnit
+import zio.ZTraceElement
+import zio.ZTraceElement
+import zio.ZTraceElement
```
regiskuckaertz, line 29: giphy

regiskuckaertz:
Hands down the best ZIO 2.0 improvement for me 😍 Thanks for doing that!

Would it be too far fetched add this to the 2.x branch:
```scala
type A | B = A with B
```
😁

adamgfraser:
@regiskuckaertz I already did in #4845! 😃

regiskuckaertz:
I had forgotten 😅 great!

AdamPaynter:
> @regiskuckaertz I already did in #4845! 😃

Minor question: If ZIO has an `A & B` type, why isn't it used more prolifically?

kitlangton:
> > @regiskuckaertz I already did in #4845! 😃
> 
> Minor question: If ZIO has an `A & B` type, why isn't it used more prolifically?

Not a bad question. Should we do that, other Adam (@adamgfraser)? Change all `with` to use the type alias within the library?

core/shared/src/main/scala/zio/ZEnvironment.scala
```diff
@@ -18,10 +18,78 @@ package zio
...
+    self.map
+      .getOrElse(
+        tag,
+        self.cache.getOrElse(
+          tag,
+          self.map.collectFirst {
+            case (curTag, value) if taggedIsSubtype(curTag, tag) =>
+              self.cache = self.cache + (curTag -> value)
+              value
+          }.getOrElse(throw new Error(s"Defect in zio.ZEnvironment: Could not find ${tag} inside ${self}"))
```
AdamPaynter, line 89: I know it's a minor edge case, but what happens if `map` contains multiple `curTag`s that are subtypes of `tag`?

adamgfraser:
I think there is a little trade off here of not deviating too far away from the normal Scala language.

In Scala 2 `&` doesn't exist. So in Scala 2 it does make things potentially a little more confusing. You're trying to understand this code that already has all of these other new things for you to learn and now you see this new symbol that you've never seen anywhere else and you are like "what is this?". And later you find out it isn't even because it expresses some new idea but just to save a few characters.

So I think in Scala 2 it made more sense to use the standard syntax of the language if there wasn't a compelling justification.

Clearly in the Scala 3 world we are going to use `&` everywhere eventually.

The question is where we are between those two worlds. I tend to think that as much as we are enthusiastic about Scala 3 most organizations are not there yet and we are better off waiting a bit and doing it as part of taking more advantage of Scala 3 features in the next ZIO version. But that is somewhat subjective.

core/shared/src/main/scala/zio/ZEnvironment.scala
adamgfraser: I will be adding logic for that shortly but we will return an arbitrary but deterministic service.

jdegoes (approved):
Incredible work pushing this through so quickly! Let's get it in now and fixup any minor issues later!

notxcain:
Not sure that it's a common thing but how to encode `Has[A with C] with Has[B with C]` with this encoding?

jdegoes:
@notxcain You need a nominal type that extends `A with C` and `B with C`, e.g.:

```scala
trait AWithC extends A with C 

trait BWithC extends B with C
```

`ZIO[AWithC & BWithC, ...]`

notxcain:
Right, thanks. So it's less powerful encoding and we're fine with it because ergonomics is 🚀 . Just wanted to be sure 😄

kitlangton:
@notxcain You can also make some `case class Box[A](value: A)` and do `Box[A with C] & Box[B with C]`
So, essentially, you can re-`Has`-ify in those situations.

jdegoes:
In fact, maybe for backward compatibility:

```scala
final case class Has[A: Tag](get: A)
```

🤔

PR #6050, opened 2021-11-16
Convert all specs to ZIOSpec

swoogles:


test/shared/src/main/scala/zio/test/ZIOSpecDefault.scala
```diff
@@ -2,12 +2,20 @@ package zio.test
 
 import zio._
 import zio.internal.stacktracer.Tracer
+import zio.internal.stacktracer.Tracer.newTrace
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
-abstract class ZIOSpecDefault extends ZIOSpec[TestEnvironment] {
+// TODO Is it sensible for this to be a class-level implicit param?
```
swoogles, line 8: @adamgfraser I think this is the only thing that I need clarity on for this specific PR chunk. The rest will be ironed out as the rest of the Specs get migrated.

adamgfraser (approved):

PR #6054, opened 2021-11-17
Fix foldWeighted emitting too big outputs

rtimush:
When `Transducer.foldWeighted` was used in constructions like `aggregateAsyncWithin` it sometimes produced outputs costing more than the defined max cost.


adamgfraser (approved):
Thank you! 🙏

PR #6055, opened 2021-11-17
Updating Naming Of Environment Operators

adamgfraser:
Updates naming of environment operators to reflect recent changes to delete `Has`.

For operators to provide the environment, we rename `provideServices` to `provide` to reflect that providing services through service builders is the idiomatic way of providing services. The variant that provides the entire `Environment` outside of a builder goes to `provideAll`. The current `provideSome` variant that transforms the environment type with a function goes to `contramap` to make it less prominent since that is not an idiomatic way to work with the environment in most cases.

- [x] `provide` to `provideAll`
- [x] `provideSome` to `contramap`
- [x] `provideServices` to `provide`
- [x] `provideSomeServices` to `provideSome`
- [x] `provideCustomServices` to `provideCustom`

For operators to access the environment `service` and `serviceWith` are the most idiomatic operators as we have previously emphasized. `environment` is available for cases in which you need to access the entire environment and `environmentWith` added to get it and do something with it. `access` and `accessZIO` will be removed.

- [x] `accessZIO` to `environmentWith`
- [x] `access` deprecated

`ZServiceBuilder` apply method modified to take a single service so you can do `ZServiceBuilder(zio)` to `ZServiceBuilder(managed)`. `Many` variants that allow returning multiple services renamed to `All` to mirror `provideAll` described above.


adamgfraser:
Renamed `contramap` to `contramapEnvironment` to avoid conflicts with data types that have a similar operator defined with respect to another type parameter. Added `environmentWith` and `serviceWith` variants that take a pure function.

kitlangton:
It passes!

adamgfraser:
Huzzah!

regiskuckaertz (dismissed):
giphy

benchmarks/src/main/scala/zio/RegionBenchmark.scala
```diff
@@ -54,10 +54,10 @@ class RegionBenchmark {
   }
 
   @Benchmark
-  def zioProvide(): Unit = {
+  def zioprovideAll(): Unit = {
```
kitlangton, line 57: OH NOO!

adamgfraser: Fixed now!

kitlangton (approved):

adamgfraser:
I think the old `provideSome` variant can be useful for partial provision now that the environment is baked in so renamed that to `provideSomeEnvironment` and updated `All` suffices to `Environment` suffixes accordingly.

kitlangton (approved):

PR #6062, opened 2021-11-19
The Provider Provides

kitlangton:
istockphoto-659661462-612x612


hmemcpy:
I'm sorry, this is literally the Abstract Factory pattern, and IMO has no place in ZIO. With ZLayer you had a *value* of a layer which you *provide* to whoever requires it. This creates another entity (*er suffix, already a javaism) just to create whatever's being provided. Overall, 👎, sorry.

regiskuckaertz:
@hmemcpy this does not change anything but the name, this is still a value describing dependency allocation, disposal and wiring, right? Now one SNOT like me may find that providing a provider is a language and logical horror but again that's my OCD 😅

hmemcpy:
Ah, I didn't read the entire thing to see this was just a "naming things" PR :) Still... I feel that ZLayers are/were a killer feature for ZIO 1.0, and if they are simplified for 2.0 that'd be great. I don't, however, think they need to be renamed.

kitlangton:
The people have spoken. Long love ZLayer!

PR #6065, opened 2021-11-19
Assert -> AssertTrue Scalafix

kitlangton:


adamgfraser (dismissed):

hmemcpy:
I've added a similar quickfix to the intellij plugin https://github.com/zio/zio-intellij/pull/328
It's by no means complete, and I'm waiting on reports where it might not do the right things. Also, it currently ignores things that are "false" or "not" tests since it doesn't read fluently with `assertTrue`, as well as `Either`s

I haven't tried it with the new `.is`, however...

PR #6066, opened 2021-11-20
Warn for unused layers instead of aborting

kitlangton:
CleanShot 2021-11-19 at 21 06 46@2x


adamgfraser (commented):
Do you think we should have a more composable way of doing this? Like maybe we should have a layer that you add like the mermaid ones that would trigger this warning? Not failing for unused layers is definitely a step in the right direction but having a warning every time you compile can also be annoying if it is not something you are immediately going to resolve.

ghostdogpr:
When I build my layer stack at the application launch, there are several layers that are not needed anywhere else. There are just components that need to be started in some specific order. I think it would be nice to not have warnings (this is one reason I don't use zio-magic today).

kitlangton:
> When I build my layer stack at the application launch, there are several layers that are not needed anywhere else. There are just components that need to be started in some specific order. I think it would be nice to not have warnings (this is one reason I don't use zio-magic today).

That's interesting. You want to start some layers up solely for their side-effects, correct?
I don't think I can default to that behavior, because if one removes the need for a particular layer, it will still get started in the background and could be confusing. A user could easily miss this. However I do think it's important to be able to do this. 

What about `sideEffect` method on ZLayer which, when used inside of `.inject` or `.wire`, will include it in the output layer just for its effects, without emitting a warning.

As far as configuring the warnings. I think we could use a setting/property/whatever they're called in `.sbt` to disable them, similar to Quill's configuration for squelching the SQL output messages. I think the default should be to warn, because otherwise, with all of the magic involved, people won't know what's actually being used or not.

adamgfraser:
Hmmm, I think we are seeing more cases where layers are being used to start background processes and I don't think we would want to make them "second class" by triggering this warning about them that you have to go into your SBT settings to disable.

Regarding the ordering of layers, I think we could preserve their ordering relative to the layers they are originally being provided to / created from in the original layers.

adamgfraser:
One thing I think you could do is look at the return type of the layer and see if it is `Unit` or `Any`. If so the layer is indicating that ti is not trying to return any value that is used by anyone else, so we should definitely not warn about that.

adamgfraser:
@kitlangton Do we still need this or is it incorporated into your other PR? With unifying provide it will be more important than ever to not fail the compilation for unused layers and not to warn about them at least if their return type is `Unit` or `Any`.

kitlangton:
Ah, yup. It was included in #6073. They'll only be warnings now. 

The `Any` / `Unit` trick is clever, but that feels a bit weird to have special behavior based on the return type. What do you think of being slightly more explicit with a `.sideEffect` method? That would also allow any layer of any type to be included.

adamgfraser:
It is the same behavior as the Scala compiler with regard to warning about discarded values, but otherwise I think I would turn off the warnings entirely.

PR #6078, opened 2021-11-21
Improvements to ZTraceElement

### LINKED ISSUE #6077: ZIO 2.0: Improvements to ZTraceElement
ZTraceElement has a few issues we should fix before 2.0 release:

 - [x] **Security**. `ZTraceElement` macro tracks the full source code location to the file on the machine that did the packaging. Under ordinary circumstances, a path will not leak any private information, but we don't have any guarantee what user names (in a home directory) or hashes or other pieces of information might be embedded in the full path leading to a file. We should change the macro so it only captures the file name, not the full path to the file on the user's local machine. 
 - [x] **Performance**. `ZTraceElement` macro tracks the column number. This means when the traces are converted to Java-style traces, before rendering, the traces have to be de-duplicated by column number, which imposes a performance impact. Column number is not likely to be useful enough to justify this performance cost, so we should omit it. 
 - [x] **Interop**. Related to the column number, there is no way to convert a StackTraceElement from Java into a ZTraceElement. That's because they are not information-equivalent. We should make them information equivalent so we can go both ways: from ZTraceElement, to a StackTraceElement, and from a StackTraceElement, to a ZTraceElement. We should add to/from helper functions on the `ZTraceElement` companion object to do just this, and then leverage these in `ZTrace`.

adamgfraser:
Resolves #6077.

### DIFF
--- a/core/shared/src/main/scala/zio/ZTrace.scala
+++ b/core/shared/src/main/scala/zio/ZTrace.scala
@@ -32,29 +32,8 @@ final case class ZTrace(
    * Converts the ZIO trace into a Java stack trace, by converting each trace
    * element into a Java stack trace element.
    */
-  def toJava: Chunk[StackTraceElement] = {
-    val chunkBuilder  = ChunkBuilder.make[StackTraceElement](stackTrace.size)
-    val chunkIterator = stackTrace.chunkIterator
-    var index         = 0
-    var previous      = -1
-    while (chunkIterator.hasNextAt(index)) {
-      chunkIterator.nextAt(index) match {
-        case ZTraceElement.SourceLocation(location, file, line, _) if line != previous =>
-          previous = line
-          val last            = location.lastIndexOf(".")
-          val (before, after) = if (last < 0) ("", "." + location) else location.splitAt(last)
-          def stripSlash(file: String): String = {
-            val last = file.lastIndexOf("/")
-            if (last < 0) file else file.drop(last + 1)
-          }
-          val stackTraceElement = new StackTraceElement(before, after.drop(1), stripSlash(file), line)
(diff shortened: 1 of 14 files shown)


core/shared/src/main/scala/zio/ZTraceElement.scala
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+  def toJava(trace: ZTraceElement): Option[StackTraceElement] =
+    trace match {
+      case ZTraceElement(location, file, line) =>
+        val last = location.lastIndexOf(".")
+
+        val (before, after) = if (last < 0) ("", "." + location) else location.splitAt(last)
+
+        def stripSlash(file: String): String = {
```
jdegoes, line 44: I am not sure if this is still necessary, because if we are only capturing the file name, then there won't be any path separators anymore (previously, this converted, e.g. `/Users/John/.../ZIO.scala` into `ZIO.scala`).

jdegoes (commented):
Looks great! One quick comment on code cleanup, then good to merge.

core/shared/src/main/scala/zio/ZTrace.scala
```diff
@@ -32,29 +32,8 @@ final case class ZTrace(
...
-          }
-          val stackTraceElement = new StackTraceElement(before, after.drop(1), stripSlash(file), line)
-          chunkBuilder += stackTraceElement
-        case _ =>
-      }
-      index += 1
-    }
-    chunkBuilder.result()
-  }
+  def toJava: Chunk[StackTraceElement] =
```
jdegoes, line 35: So much cleaner (& faster)! ❤️

jdegoes (approved):

PR #6089, opened 2021-11-24
Simplify Sinks

adamgfraser:
Simplifies sinks by removing their input error type. Implements a new `pipeToOrFail` operator that pipes the output of one channel to another, where the second channel does not handle errors. The errors from the original channel are passed through unchanged and unified with the error type of the second channel.

### DIFF
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -2632,7 +2632,7 @@ object ZStreamSpec extends ZIOBaseSpec {
           }
         ),
         test("peel") {
-          val sink: ZSink[Any, Int, Nothing, Int, Any] = ZSink.take(3)
+          val sink: ZSink[Any, Nothing, Int, Int, Any] = ZSink.take(3)
 
           ZStream.fromChunks(Chunk(1, 2, 3), Chunk(4, 5, 6)).peel(sink).use { case (chunk, rest) =>
             rest.runCollect.map { rest =>
--- a/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
+++ b/streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
@@ -172,7 +172,7 @@ object ZStreamSpec extends ZIOBaseSpec {
                      .range(1, 10)
                      .tap(i => ZIO.fail("BOOM!").when(i == 6) *> queue.offer(i))
                      .aggregateAsyncWithin(
-                       ZSink.foldUntil[String, Int, Unit]((), 5)((_, _: Int) => ()),
+                       ZSink.foldUntil[Int, Unit]((), 5)((_, _: Int) => ()),
                        Schedule.forever
                      )
                      .runDrain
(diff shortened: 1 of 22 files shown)


streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -781,6 +781,16 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
   )(implicit trace: ZTraceElement): ZChannel[Env1, InErr, InElem, InDone, OutErr2, OutElem2, OutDone2] =
     ZChannel.PipeTo(() => self, () => that)
 
+  def pipeToOrFail[Env1 <: Env, OutErr1 >: OutErr, OutElem2, OutDone2](
+    that: => ZChannel[Env1, Nothing, OutElem, OutDone, OutErr1, OutElem2, OutDone2]
+  )(implicit trace: ZTraceElement): ZChannel[Env1, InErr, InElem, InDone, OutErr1, OutElem2, OutDone2] = {
+    final case class ChannelFailure(err: OutErr) extends Throwable
```
adamgfraser, line 787: There is a missing primitive here that combines two channels to produce a new channel that can accept either of their inputs and produce either of their outputs.

streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -84,32 +84,35 @@ class ZSink[-R, -InErr, -In, +OutErr, +L, +Z](val channel: ZChannel[R, InErr, Ch
...
+  ): ZSink[R, In, E, L, S] =
     new ZSink(
       ZChannel
         .fromZIO(Ref.make(Chunk[In]()).zip(Ref.make(false)))
-        .flatMap { case (leftoversRef, upstreamDoneRef) =>
-          lazy val upstreamMarker: ZChannel[Any, InErr, Chunk[In], Any, InErr, Chunk[In], Any] =
-            ZChannel.readWith(
-              (in: Chunk[In]) => ZChannel.write(in) *> upstreamMarker,
-              ZChannel.fail(_: InErr),
+        .flatMap[R, Nothing, Chunk[In], Any, E, Chunk[L], S] { case (leftoversRef, upstreamDoneRef) =>
```
adamgfraser, line 91: Unfortunately type inference is terrible here with `Nothing` for the `InErr` type. The Scala compiler really does not want to infer `Nothing` here. It is only a problem if you are implementing your own sink operators in terms of channels but it is definitely painful.

streams-tests/shared/src/test/scala/zio/stream/ZStreamSpec.scala
```diff
@@ -2798,7 +2798,7 @@ object ZStreamSpec extends ZIOBaseSpec {
               assert(result)(equalTo(Chunk(1, 1, 1))) && assert(state)(isFalse) && assert(finalState)(isTrue)
             }
           )
-        ),
+        ) @@ ignore,
```
adamgfraser, line 2801: This test fails when `left >>> right` is replaced with `left.catchAll(???) >>> right`. I think this is a bug in the channel executor.

vigoo: I will take a look

vigoo (approved):
Very nice simplification, let's merge and fix the finalization error separately

PR #6092, opened 2021-11-24
Added `removeIf` as alternative to `removeIfDiscard` for `TSet` and `TMap`

### LINKED ISSUE #6083: TMap: removeIf should return keys of removed elements
currently I have to use code like this to figure out which elements have been removed:
```scala
...
 jobMap = TMap[JobId, Timestamp]
...
      jobKeys <- jobMap.keys
      _ <- jobMap.removeIf{ case(_, ts) => before.value >= ts.value }
      newJobKeys <- jobMap.keys
      removedJobs = jobKeys.diff(newJobKeys)
```
      
would be good if `removeIf` would return me a `List`/`Seq`/`Chunk` of the *keys* or elements that got removed

jeroenr:
Closes #6083

### DIFF
--- a/core/shared/src/main/scala/zio/stm/TMap.scala
+++ b/core/shared/src/main/scala/zio/stm/TMap.scala
@@ -274,9 +274,46 @@ final class TMap[K, V] private (
     get(k).flatMap(_.fold(put(k, v))(_ => STM.unit))
 
   /**
+   * Removes bindings matching predicate and returns the removed entries.
+   */
+  def removeIf(p: (K, V) => Boolean): USTM[Chunk[(K, V)]] =
+    ZSTM.Effect { (journal, _, _) =>
+      val f        = p.tupled
+      val buckets  = tBuckets.unsafeGet(journal)
+      val capacity = buckets.array.length
+
+      var i       = 0
+      var newSize = 0
+      val removed = ChunkBuilder.make[(K, V)]()
+
+      while (i < capacity) {
+        val bucket    = buckets.array(i).unsafeGet(journal)
+        var newBucket = List.empty[(K, V)]
+
+        val it = bucket.iterator
+        while (it.hasNext) {
+          val pair = it.next()
+          if (!f(pair)) {
+            newBucket = pair :: newBucket
+            newSize += 1
+          } else {
+            removed += pair
+          }
+        }
+
+        buckets.array(i).unsafeSet(journal, newBucket)
+        i += 1
+      }
+
+      tSize.unsafeSet(journal, newSize)
+
+      removed.result()
+    }
+
+  /**
    * Removes bindings matching predicate.
    */
-  def removeIf(p: (K, V) => Boolean): USTM[Unit] =
+  def removeIfDiscard(p: (K, V) => Boolean): USTM[Unit] =
     ZSTM.Effect { (journal, _, _) =>
       val f        = p.tupled
       val buckets  = tBuckets.unsafeGet(journal)
@@ -308,38 +345,16 @@ final class TMap[K, V] private (
     }
 
   /**
+   * Retains bindings matching predicate and returns removed bindings.
+   */
+  def retainIf(p: (K, V) => Boolean): USTM[Chunk[(K, V)]] =
+    removeIf(!p(_, _))
+
+  /**
    * Retains bindings matching predicate.
    */
-  def retainIf(p: (K, V) => Boolean): USTM[Unit] =
-    ZSTM.Effect { (journal, _, _) =>
-      val f        = p.tupled
-      val buckets  = tBuckets.unsafeGet(journal)
-      val capacity = buckets.array.length
-
-      var i       = 0
-      var newSize = 0
-
-      while (i < capacity) {
-        val bucket    = buckets.array(i).unsafeGet(journal)
-        var newBucket = List.empty[(K, V)]
-
-        val it = bucket.iterator
-        while (it.hasNext) {
-          val pair = it.next()
-          if (f(pair)) {
-            newBucket = pair :: newBucket
-            newSize += 1
-          }
-        }
-
-        buckets.array(i).unsafeSet(journal, newBucket)
-        i += 1
-      }
-
-      tSize.unsafeSet(journal, newSize)
-
-      ()
-    }
+  def retainIfDiscard(p: (K, V) => Boolean): USTM[Unit] =
+    removeIfDiscard(!p(_, _))
 
   /**
    * Returns the number of bindings.
--- a/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/stm/TMapSpec.scala
@@ -3,7 +3,7 @@ package zio.stm
 import zio.test.Assertion._
 import zio.test.TestAspect.nonFlaky
 import zio.test._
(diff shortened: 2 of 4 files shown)


jeroenr:
> `removeIfAndGet`?

@adamgfraser also a good suggestion. I used the name suggested by the author of the issue

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -233,6 +233,43 @@ final class TMap[K, V] private (
       TExit.unit
     })
 
+  /**
+   * Removes bindings matching predicate.
+   */
+  def removeIfReturning(p: (K, V) => Boolean): USTM[List[(K, V)]] =
```
adamgfraser, line 239: This should return a `Chunk`.

adamgfraser:
@mijicd What do you think?

core/shared/src/main/scala/zio/stm/TMap.scala
jeroenr: technically, multiple entries could be removed. For instance when you pass the following predicate:

```
for { 
   tmap <- TMap.make("a" -> 1, "aa" -> 2, "aaa" -> 3)
   removed <- tmap.removeIfReturning((_, v) => v > 1)
} yield (removed)
```

adamgfraser: Sure. A `Chunk` is a collection so if you can return a `List` you can also return a `Chunk`.

jeroenr: I changed to use `zio.Chunk` now

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -233,6 +233,43 @@ final class TMap[K, V] private (
...
+   */
+  def removeIfReturning(p: (K, V) => Boolean): USTM[Chunk[(K, V)]] =
+    new ZSTM((journal, _, _, _) => {
+      val f        = p.tupled
+      val buckets  = tBuckets.unsafeGet(journal)
+      val capacity = buckets.array.length
+
+      var i       = 0
+      var newSize = 0
+      var removed = Chunk[(K, V)]()
```
mijicd, line 247: Perhaps it makes sense to use `ChunkBuilder` instead.

adamgfraser, line 247: You can use a `ChunkBuilder` here.

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -233,6 +233,43 @@ final class TMap[K, V] private (
       TExit.unit
     })
 
+  /**
+   * Removes bindings matching predicate and returns the removed entries.
+   */
+  def removeIfReturning(p: (K, V) => Boolean): USTM[Chunk[(K, V)]] =
```
mijicd, line 239: My cup of tea:
- If we go with two combinators we should probably follow the approach from `foreach` case, i.e. `foreach_` was renamed to `foreachDiscard`. Unfortunately, using it here would mean we have to postpone this change for 2.x.
- Regardless of the approach, please provide the same alternative for `retainIf` :).

mijicd (commented):
Please make sure to reference the issue in the initial PR comment. Thank you!

core/shared/src/main/scala/zio/stm/TMap.scala
jeroenr: I also wrote the `retainIfReturning` (which uses `removeIfReturning` under the hood). I don't really have a preference with regards to the naming so I'm happy to apply any suggestion from your side

core/shared/src/main/scala/zio/stm/TMap.scala
jeroenr: I used `ChunkBuilder` now. Note that it does change the order of the output (because the builder is appending and not prepending)

core/shared/src/main/scala/zio/stm/TMap.scala
```diff
@@ -267,6 +276,12 @@ final class TMap[K, V] private (
       TExit.unit
     })
 
+  /**
+   * Retains bindings matching predicate and returns removed bindings.
+   */
+  def retainIfReturning(p: (K, V) => Boolean): USTM[Chunk[(K, V)]] =
+    removeIfReturning(!p(_, _))
```
mijicd, line 283: Sorry for being nitpicky, but this is a bit odd as we have a bit convoluted situation:
- `retainIfReturning` is implemented in terms of `removeIfReturning`
- `removeIf` is implemented in terms of `retainIf`

core/shared/src/main/scala/zio/stm/TMap.scala
mijicd: Order isn't important here, so this should be fine.

adamgfraser: Well I don't think there is a deterministic order because the map is unordered but to the extent we did care about order isn't this more "correct" because we are returning the removed elements in the same order we originally traversed them?

core/shared/src/main/scala/zio/stm/TMap.scala
adamgfraser: Yeah I think the `Returning` suffix is confusing and it is not used anywhere in our code base. The `Discard` suffix is not ideal in a way because it makes the more common use case where you don't care about the values longer but it is at least consistent. Let's retarget to the `series/2.x` branch and do that.

mijicd: @jeroenr please rename the combinators as follows:

old name | new name
-|-
`removeIf` | `removeIfDiscard`
`removeIfReturning` | `removeIf`
`retainIf` | `retainIfDiscard`
`retainIfReturning` | `retainIf`

As Adam pointed out, these are unfortunate, but if we find better names in the meantime, the change is "one `sed` away" :).

mijicd (approved):

core/shared/src/main/scala/zio/stm/TMap.scala
jeroenr: I applied the suggested renaming and also introduced the same in `TSet` to keep APIs in sync. PTAL :)

adamgfraser:
How is this binary compatible?

mijicd:
@adamgfraser It is not, that's why it's pointed to series/2.x

adamgfraser:
Looks like it was changed to target `series/2.x` and then changed back to target `master`?

mijicd:
There seem to be "a few" conflicts to fix though.

mijicd:
@jeroenr As suggested [here](https://github.com/zio/zio/pull/6092#discussion_r756087943), this should be a part of 2.x. Not sure why did you change the branch again, but I pointed it back to `series/2.x`. Please resolve the conflicts and we can proceed with merging it.

jeroenr:
> @jeroenr As suggested [here](https://github.com/zio/zio/pull/6092#discussion_r756087943), this should be a part of 2.x. Not sure why did you change the branch again, but I pointed it back to `series/2.x`. Please resolve the conflicts and we can proceed with merging it.

Ok, I think I got it now. PTAL :)

adamgfraser (approved):
Thank you! 🙏

PR #6093, opened 2021-11-24
Extract ZIO Concurrent Into Separate Module

adamgfraser:
Extracts `concurrent` package into separate module to allow it to evolve independently of ZIO.

I wonder if we should also move into separate repository. That was not successful before but I think that may be because we did not have a critical mass of initial code. It would be nice to move things out if possible as there is already a lot of stuff in this repository.


mijicd (approved):
A separate module makes sense and it's less "operational" overhead for now. Depending on the outcome of Flavio's work (i.e. whether or not we'll implement some of the internals in terms of them), we might consider revamping zio/zio-concurrent.

PR #6098, opened 2021-11-25
Fixed #4593 - Removed "invisible" tag from docs

### LINKED ISSUE #4593: Ensure All Examples From Documentation Are Compiled

emilhotkowski:
Original issue:

> We should ensure that all code snippets in the documentation are checked with mdoc to ensure that they compile as written.
> 
> Can I also ask that for all examples we NOT use hidden or invisible code? I know sometimes people do this because they want to omit things they see as boilerplate like imports or type definitions. However, a lot of people are going to copy and paste and compile these examples as written so if there are things missing it makes it harder for them to get started.

### DIFF
--- a/docs/datatypes/concurrency/hub.md
+++ b/docs/datatypes/concurrency/hub.md
@@ -153,7 +153,7 @@ trait ZStream[-R, +E, +O] {
 
 We would now like to have multiple downstream consumers process each of these transactions, for example to persist them and log them in addition to applying our business logic to them. With `Hub` this is easy because we can just use the `toQueue` operator to view any `Hub` as a `Queue` that can only be written to.
 
-```scala mdoc:invisible
+```scala mdoc
 type ??? = Nothing
 ```
 
@@ -203,7 +203,7 @@ The `map` operator allows us to transform the type of messages received from the
 
 The `mapM` operator works the same way except it allows us to perform an effect each time a value is taken from the hub. We could use this for example to log each time a message is taken from the hub.
 
-```scala mdoc:reset:invisible
+```scala mdoc:reset
 import zio._
 import zio.console._
 ```
@@ -244,7 +244,7 @@ The `contramapM` operator works the same way except it allows us to perform an e
 
 Using these operators, we could describe a hub that validates its inputs, allowing publishers to publish raw data and subscribers to receive validated data while signaling to publishers when data they attempt to publish is not valid.
 
-```scala mdoc:reset:invisible
+```scala mdoc:reset
 import zio._
 import zio.console._
 ```
@@ -345,7 +345,7 @@ The managed effect here describes subscribing to receive messages from the hub w
 
 Here is an example of using it:
 
-```scala mdoc:reset:invisible
+```scala mdoc:reset
 import zio._
 import zio.stream._
 ```
--- a/docs/datatypes/concurrency/index.md
+++ b/docs/datatypes/concurrency/index.md
@@ -39,7 +39,7 @@ As the lock-oriented programming does not compose and has lots of drawbacks, ZIO
 
 Let's see how the `modify` function of `Ref` is implemented without any locking mechanism:
 
-```scala mdoc:invisible
+```scala mdoc
 import java.util.concurrent.atomic.AtomicReference
 import zio.UIO
 ```
--- a/docs/datatypes/contextual/index.md
+++ b/docs/datatypes/contextual/index.md
@@ -12,7 +12,7 @@ The input type is also known as _environment type_. This type-parameter indicate
 `R` represents dependencies; whatever services, config, or wiring a part of a ZIO program depends upon to work. We will explore what we can do with `R`, as it plays a crucial role in `ZIO`.
 
 For example, when we have `ZIO[Console, Nothing, Unit]`, this shows that to run this effect we need to provide an implementation of the `Console` service:
-```scala mdoc:invisible
+```scala mdoc
 import zio.ZIO
 import zio.console._
 ```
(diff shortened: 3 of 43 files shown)


docs/datatypes/concurrency/hub.md
```diff
@@ -153,7 +153,7 @@ trait ZStream[-R, +E, +O] {
 
 We would now like to have multiple downstream consumers process each of these transactions, for example to persist them and log them in addition to applying our business logic to them. With `Hub` this is easy because we can just use the `toQueue` operator to view any `Hub` as a `Queue` that can only be written to.
 
-```scala mdoc:invisible
+```scala mdoc
 type ??? = Nothing
 ```
```
khajavi, line 158: There are two blocks of code, one after the other. It's better to merge them.

docs/datatypes/concurrency/hub.md
```diff
@@ -203,7 +203,7 @@ The `map` operator allows us to transform the type of messages received from the
 
 The `mapM` operator works the same way except it allows us to perform an effect each time a value is taken from the hub. We could use this for example to log each time a message is taken from the hub.
 
-```scala mdoc:reset:invisible
+```scala mdoc:reset
```
khajavi, line 206: Instead of using the `reset` modifier, it is better not to share imports between blocks of codes. To do so, it's better to use `compile-only` modifier for those blocks that are not supposed to be evaluated.

docs/datatypes/concurrency/index.md
```diff
@@ -39,7 +39,7 @@ As the lock-oriented programming does not compose and has lots of drawbacks, ZIO
 
 Let's see how the `modify` function of `Ref` is implemented without any locking mechanism:
 
-```scala mdoc:invisible
+```scala mdoc
```
khajavi, line 42: It would be better to merge this block of code into the next block.

khajavi: Whenever possible, I recommend merging invisible imports into their next block.

docs/datatypes/contextual/index.md
```diff
@@ -203,12 +203,12 @@ Let's start learning this pattern by writing a `Logging` service:
...
-```scala mdoc:invisible:reset
+```scala mdoc:reset
 import zio._
 import zio.console._
 ```
 
-```scala mdoc:invisible
+```scala mdoc
 import zio.{Has, UIO, Layer, ZLayer, ZIO, URIO}
 ```
```
khajavi, line 213: same here

docs/datatypes/contextual/index.md
```diff
@@ -203,12 +203,12 @@ Let's start learning this pattern by writing a `Logging` service:
 
 Accessor methods allow us to utilize all the features inside the service through the ZIO Environment. That means, if we call `log`, we don't need to pull out the `log` function from the ZIO Environment. The `accessM` method helps us to access the environment of effect and reduce the redundant operation, every time.
 
-```scala mdoc:invisible:reset
+```scala mdoc:reset
 import zio._
 import zio.console._
 ```
```
khajavi, line 209: same here

docs/datatypes/concurrency/hub.md
khajavi: Another note would like to mention:
Removing invisible imports isn't sufficient. We should make sure that each code blocks have fresh imports.

This addresses the following issue, which @adamgfraser raised in the https://github.com/zio/zio/issues/4593 issue:
> However, a lot of people are going to copy and paste and compile these examples as written so if there are things missing it makes it harder for them to get started.

emilhotkowski: I want to confirm before I put too much work into wrong path. 

we would like to be EVERY snippet of code be self sufficient when it comes to import? So we can just copy it and run? Should I do anything in a case when snippet is split into two parts with some comment in between when one having both parts compile it? Should both have sufficient imports or only first?

khajavi:
Hello @emilhotkowski,
I appreciate your contribution and thanks for this merge request.
I added some notes, which might help to improve your work.

docs/datatypes/concurrency/hub.md
khajavi: We have these types of snippets:
1. Independent snippets: snippets that are not dependent on previous works. They should have all the required imports. 
 - Whenever we do not require evaluating an expression and we don't require showing the result of an expression to the user, I suggest using `compile-only` modifier, which doesn't pollute the scope. The good practice is to use `compile-only` by default.
 - If we need to evaluate the expression, we cannot use the `compile-only` modifier, so after that to clear the previous scopes, we need to use the 'reset' modifier to ensure that the scope is cleared and the next blocks are independent of previous scopes. 
2. Dependent snippets: snippets that depend on previous blocks of code. When a user copies and pastes all those separate blocks, the final code should be compiled, requiring no additional imports. After finishing these groups of related snippets, we should clear the scope using the `reset` modifier.

emilhotkowski:
Hi @khajavi,

I changed the code, but also I used my new laptop with new personal access token and probably because of that you can see two different accounts. This creates a problem with CLA, any idea how to resolve this?

khajavi:
> Hi @khajavi,
> 
> I changed the code, but also I used my new laptop with new personal access token and probably because of that you can see two different accounts. This creates a problem with CLA, any idea how to resolve this?

HI Emil
What is the exact problem? Can't you sign the CLA again?

PR #6099, opened 2021-11-25
Track Fiber Fork Locations

adamgfraser:


jdegoes (changes requested):
Looks fantastic! I'd do two more things:

1. Expose the fiber location with something like a `ZTraceElement` method on `Fiber.Runtime`. We should think about putting that in `UIO` since it's not stable to copy/paste.
2. Pass the `fiberLocation` to the `ZLogger` so logging backends have a chance to report on the location of the fiber.

adamgfraser:
Made those changes. It seems like the `location` can just be a pure value since forking a fiber is side effecting but once a fiber is forked the location it is forked from is just a constant value. This would be consistent with the treatment of the fiber identifier.

jdegoes (approved):

PR #6100, opened 2021-11-26
Implement MockRandom#nextUUID

adamgfraser:


stevebakh:
Thanks for introducing this change @adamgfraser. For the record, it appears nextuuid has also been missed out of `TestRandom` too, i.e. there's no way to feed the buffer with a set of UUID's to use in tests. 

For context, I was attempting to write a test that verifies my code uses ZIO random to generate a UUID and invoke another service with that value. I'm using the test generators to generate a random UUID each time the test is executed. This is then used in the test as part of a mock expectation to verify that the code being tested uses ZIO Random to generate a UUID and pass it to the mocked injected service. e.g. tell MockRandom to return the generated value when calling `nextUUID`, and then check that this same value is passed along. 

For now, I've created my own Mock Random (essentially a copy pasta of the file in your commit) until this is backported into ZIO.

adamgfraser:
@stevebakh It already exists and will be available in the next release.

https://github.com/zio/zio/blob/18a17b03fdebcc1f5832b2ac3627c4899618ab60/test/shared/src/main/scala/zio/test/environment/package.scala#L1598

jdegoes (approved):

PR #6103, opened 2021-11-26
Optimize Metrics

adamgfraser:
When notifying listeners that a metric has changed we always need to only send the delta and never the entire metric state.


jdegoes (approved):

jdegoes:
Good to merge when tests are passing!

PR #6107, opened 2021-11-27
Restoring having `NeedsEnv[R]` on the provide overloads

hmemcpy:
This PR restores having the `NeedsEnv[R]` implicit on the macro-calling overloads of `provide*`, to allow statically detecting whether this effect indeed requires some R. This helps tooling (mainly, IntelliJ) detect these erroneous situations and offer a fix.

I've asked @kitlangton for this fix, but couldn't resist trying it myself :D

- [x] Ensure all tests pass
- [x] Ensure Scala 3 macros still work

### DIFF
--- a/core-tests/shared/src/test/scala/zio/TagCorrectnessSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/TagCorrectnessSpec.scala
@@ -65,8 +65,9 @@ object TagCorrectnessSpec extends ZIOSpecDefault {
             environment.get
           }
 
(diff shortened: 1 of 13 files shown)


hmemcpy:
@adamgfraser in `TagCorrectnessSpec.scala:69`:
https://github.com/zio/zio/blob/b1688d0e3569b319dfc332df97ed7f096cce6059/core-tests/shared/src/test/scala/zio/TagCorrectnessSpec.scala#L68-L69

This test should not compile. Is there another way to test it?

PR #6109, opened 2021-11-28
Tests for pipeToOrFail + experimenting with union

vigoo:


vigoo:
Needs https://github.com/zio/zio/pull/6127

PR #6112, opened 2021-11-29
ZSink#raceBoth: relax the type bounds of the return type

sideeffffect:


core/shared/src/main/scala/zio/FiberId.scala
```diff
@@ -76,7 +76,7 @@ object FiberId {
 
   private[zio] val _fiberCounter = new java.util.concurrent.atomic.AtomicInteger(0)
 
-  case object None                                           extends FiberId
+  final case object None                                     extends FiberId
```
adamgfraser, line 79: I think this will result in an error on Scala 3.

adamgfraser (approved):
Good to go when CI is passing.

core/shared/src/main/scala/zio/FiberId.scala
sideeffffect: Seems like it passes :crossed_fingers:

sideeffffect:
All green :tada:

PR #6118, opened 2021-12-01
Make ZChannel.end strict again

vigoo:
A few weeks ago while trying to make all the stream/sink tests pass I changed `ZChannel.end` to lazy, backed by `effectTotal`, just to support the lazyness of `ZSink.succeed`.

This is very bad for performance and channels are low level enough to provide both a strict `end` and a lazy `effectTotal` for its users, so in this I change it back and only use `effectTotal` for `ZSink.succeed`.

(I have no idea why didn't I do this the first time)

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZSink.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZSink.scala
@@ -1445,7 +1445,7 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
    * A sink that immediately ends with the specified value.
    */
   def succeed[Z](z: => Z)(implicit trace: ZTraceElement): ZSink[Any, Nothing, Any, Nothing, Z] = new ZSink(
-    ZChannel.effectTotal(z)
+    ZChannel.succeed(z)
   )
 
   /**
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZSink.scala
```diff
@@ -1445,7 +1445,7 @@ object ZSink extends ZSinkPlatformSpecificConstructors {
    * A sink that immediately ends with the specified value.
    */
   def succeed[Z](z: => Z)(implicit trace: ZTraceElement): ZSink[Any, Nothing, Any, Nothing, Z] = new ZSink(
-    ZChannel.succeed(z)
+    ZChannel.effectTotal(z)
```
adamgfraser, line 1448: This can still be `succeed`, right?

adamgfraser (dismissed):
Would be good to make the strict variant package private.

vigoo:
> Would be good to make the strict variant package private.

I'm not sure about this. If you look at how channels are used to implement higher level constructs like pipelines or sinks, the strict `end` is something that you would use many times in them. Maybe we can say channels are low level enough to get the user the freedom to choose it? I think not having to get out of the channel executor to ZIO is a good enough reason (for using the strict one)

streams/shared/src/main/scala/zio/stream/ZSink.scala
vigoo: fixed

adamgfraser:
Is that telling us we need a variant that is lazy but doesn't go through ZIO?

adamgfraser (approved):

vigoo:
> Is that telling us we need a variant that is lazy but doesn't go through ZIO?

Maybe, I think we should start benchmarking these things

adamgfraser:
💯

adamgfraser (approved):

PR #6120, opened 2021-12-01
Add ZStream#sliding

luis3m:
Discussed on Discord: https://discord.com/channels/629491597070827530/630498701860929559/915623139013181521

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -1917,7 +1917,7 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
    *   size of the chunk
    */
   def grouped(chunkSize: Int)(implicit trace: ZTraceElement): ZStream[R, E, Chunk[A]] =
-    transduce(ZSink.collectAllN[A](chunkSize))
+    rechunk(chunkSize).chunks
 
   /**
    * Partitions the stream with the specified chunkSize or until the specified
@@ -2919,6 +2919,7 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
     ZStream.unwrap {
       ZIO.succeed {
         val rechunker = new ZStream.Rechunker[A](n)
+
         lazy val process: ZChannel[R, E, Chunk[A], Any, E, Chunk[A], Unit] =
           ZChannel.readWithCause(
             (chunk: Chunk[A]) =>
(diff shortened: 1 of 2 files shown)


luis3m:
Suggestions for a better name welcome :)

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4017,6 +4017,60 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
   def whenZIO[R1 <: R, E1 >: E](b: ZIO[R1, E1, Boolean])(implicit trace: ZTraceElement): ZStream[R1, E1, A] =
     ZStream.whenZIO(b)(self)
 
+  /**
+   * Emits a sliding window of n elements.
+   */
+  def windowed(chunkSize: Int)(implicit trace: ZTraceElement): ZStream[R, E, Chunk[A]] =
```
adamgfraser, line 4023: I think I might call this `grouped`. I believe that is the name for the Scala collection operation this corresponds to and I think normally we have found it has worked well to give stream versions of collection operators the same name. There is also potentially a more general version of this `sliding` that allows the step size to be different than the window size.

A further question for at least this version is how it related to chunking. Can this just be rewritten as `chunkN(chunkSize).chunks`. That is at least in its simple form where the step size and window size are the same I think it just amounts to rechunking the stream into the window size and then exposing those chunks.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4017,6 +4017,60 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
...
+    if (chunkSize <= 0) ZStream.empty
```
adamgfraser, line 4024: I think I would die in this case and treat it as an illegal argument exception instead of returning an empty stream. That would be consistent with the behavior of `grouped(0)` on a collection and the implicit contract that this should return the same stream, just with elements grouped differently.

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Ah I see this uses a step size of one instead of the window size. Yes I think `sliding` is the right name for this though would be good to be able to control the step size in addition to the window size.

streams/shared/src/main/scala/zio/stream/ZStream.scala

streams/shared/src/main/scala/zio/stream/ZStream.scala
luis3m: Renamed to `sliding`. As suggested, now it includes a stepSize parameter.

luis3m:
`ZStream#sliding(x, y) == Vector#sliding(x, y)`
`ZStream#sliding(x, x) == ZStream#grouped(x).filterNot(_.isEmpty)` (grouped emits Chunk(Chunk()), sliding does not)

adamgfraser:
Maybe `grouped` should not emit empty chunks?

luis3m:
@adamgfraser changing `grouped` implementation to `rechunk(chunkSize).chunks` works. Should I update this PR with this change?

adamgfraser:
@luis3m Up to you whether you want to do it here or separately.

luis3m:
@adamgfraser if this PR is good to get merged then I'll submit a separate PR.

adamgfraser (approved):

adamgfraser (approved):

PR #6123, opened 2021-12-02
Implement ZStream#tapSink

adamgfraser:
Allows sending all elements emitted from a stream to a sink in addition to emitting them.

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZStream.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZStream.scala
@@ -3492,6 +3492,18 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
     catchAll(e => ZStream.fromZIO(f(e) *> ZIO.fail(e)))
 
   /**
+   * Sends all elements emitted by this stream to the specified sink in
+   * addition to emitting them.
+   */
+  final def tapSink[R1 <: R, E1 >: E](
+    sink: ZSink[R1, E1, A, Any, Any],
+    maximumLag: Int
+  )(implicit trace: ZTraceElement): ZStream[R1, E1, A] =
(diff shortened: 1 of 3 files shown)


vigoo (approved):

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3491,6 +3491,18 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
   )(implicit ev: CanFail[E], trace: ZTraceElement): ZStream[R1, E1, A] =
     catchAll(e => ZStream.fromZIO(f(e) *> ZIO.fail(e)))
 
+  /**
+   * Sends all elements emitted by this stream to the specified sink in addition
+   * to emitting them.
+   */
+  final def tapSink[R1 <: R, E1 >: E](
+    sink: ZSink[R1, E1, A, Any, Any],
+    maximumLag: Int
```
adamgfraser, line 3500: It could be nice to have a default value for this and similar operators.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -257,7 +257,7 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
    */
   final def broadcast(n: Int, maximumLag: Int)(implicit
     trace: ZTraceElement
-  ): ZManaged[R, Nothing, List[ZStream[Any, E, A]]] =
+  ): ZManaged[R, Nothing, Chunk[ZStream[Any, E, A]]] =
```
adamgfraser, line 260: I changed the signature here to return a `Chunk`. I think this is more idiomatic and consistent with our philosophy of using `Chunk` throughout. It is also somewhat more ergonomic because you can index into the streams directly without having to pattern match and deal with the impossible case where the list does not have the expected number of elements.

vigoo (approved):

PR #6134, opened 2021-12-07
Implement ZStream#rechunkMin

luis3m:


adamgfraser (commented):
I worry bit about having too many ways to do things here (e.g. when should a user use `rechunk` versus this). I understand the idea of not splitting up chunks that are a little larger than the target size but having chunks that are too large can have costs too.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -2923,6 +2930,49 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
       }
     }
 
+  /**
+   * Re-chunks the elements of the stream into at least chunks of `n` elements
+   * each. The last chunk might contain less than `n` elements unless specified
+   * otherwise.
+   */
+  def rechunkMin(n: Int, smallerLastChunk: Boolean = true)(implicit trace: ZTraceElement): ZStream[R, E, A] =
```
adamgfraser, line 2938: `rechunkAtLeast`?

adamgfraser: Dropping elements if `smallerLastChunk` is `false`  seems pretty unexpected. I think the contract with chunking should be that it may change the chunk structure but shouldn't change the elements of the stream.

jdegoes:
@luis3m Since there was no clear resolution on these issues, will close for now. Thanks for your PR!

PR #6135, opened 2021-12-07
Implement ZStream#forall & ZStream#exists

luis3m:


adamgfraser (changes requested):
The issue with these operators is that they really "run" the stream. Right now they return a `ZStream` but they will always emit exactly one element so the more accurate return type would be `ZIO`. In this way these operators are analogous to `runSum` and `runCount`.

We have debated whether they should exist at all but I think enough people have been interested in them that we should just add them with the `run` prefix, probably implementing them in terms of sinks so they can potentially be composed.

Separately, we could consider having a more "streaming" variant of these operators that would basically do a scan, for example emitting a stream of the running total instead of just the total.

borissmidt:
In monix there are seperate calls for a lazy variant which returns an IO, `forallL`, and an Observable,`forall` , 
this makes it easier to keep on composing as a Stream or as an IO.
It is especially handy if you are flatmapping your stream.

adamgfraser:
Yes we could definitely do something like `sum` and `runSum` to describe the versions that return a stream versus a ZIO value. I think there is some danger in confusing operators which effectively run the stream with operators that transform the stream, but it is nice to be able to stay in the stream context.

jdegoes:
I think we could add `ZSink.exists`, `ZSink.forall` etc., then you can just do `stream.run(forall)`.

Given the fact that streams are not collection-like in the sense you cannot, in general, repeat their elements, it's good to make the interface in this respect deliberately different than collections.

Please re-open if you have a chance to continue your work here. Thank you!

PR #6138, opened 2021-12-08
Add ZSink.digest to calculate MessageDigest of incoming bytes

jypma:
This adds a ZSink that can calculate SHA-1 or SHA-256 hashes of an
incoming byte stream, emitting the value when the stream completes.
- [x] Move to `platform.scala` and related spec
- [x] Take a `=> MessageDigest` as argument


jypma:
Hmm, `MessageDigest` (i guess obviously) doesn't exist on ScalaJS. Should we make `ZSink.digest` just throw an exception there? if so, how do we do that?

adamgfraser (approved):
Thank you! :pray:

jypma:
You're welcome, thanks for merging!

PR #6140, opened 2021-12-08
Make All Stream Operators Lazy

adamgfraser:


adamgfraser:
There is an issue with the website that @atooni is looking into. I think we aren't locking our dependencies and something got updated so it is pulling in an incompatible version. But otherwise this is ready for review.

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -4479,10 +4538,10 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
    * variant.
    */
   @deprecated("use cross", "2.0.0")
-  def crossN[R, E, A, B, C](zStream1: ZStream[R, E, A], zStream2: ZStream[R, E, B])(
+  def crossN[R, E, A, B, C](zStream1: => ZStream[R, E, A], zStream2: => ZStream[R, E, B])(
     f: (A, B) => C
   )(implicit trace: ZTraceElement): ZStream[R, E, C] =
-    zStream1.crossWith(zStream2)(f)
+    ZStream.suspend(zStream1.crossWith(zStream2)(f))
```
vigoo, line 4544: I think this suspend does nothing as the `f` itself is not lazy?

vigoo (approved):
Looks good!

PR #6144, opened 2021-12-10
Specialize Pipelines

### LINKED ISSUE #6122: ZPipeline has a really impenetrable type-signature and doesn't infer well.
I recently started exploring the ZIO Streams encoding for ZIO 2. I like the idea of ZPipeline, but I find that working with this encoding to be more cumbersome that just using the ZStreams directly.

1) The types don't infer as well as with streams

The following sliding method composes two ZPipelines. I needed to specify the types everywhere to do this

```scala 
def sliding[R,E,A](n:Int) =
    scan[A, Chunk[A]](Chunk.empty) {
      case (acc, a) => if(acc.size < n) acc.appended(a) else acc.drop(1).appended(a)
    } @@ dropWhile[Chunk[A]](_.size < n)
```

I couldn't get either VS Code nor Intellij to print out the correct method signature (it was abbreviated in VS Code, and intellij shows red squiggles). Contrast this to the implementation directly on a stream

```scala
  def sliding[R,E,A](stream:ZStream[R,E,A], n:Int):ZStream[R,E,Chunk[A]] =
    stream.scan[Chunk[A]](Chunk.empty) {
      case (acc, a) => if(acc.size < n) acc.appended(a) else acc.drop(1).appended(a)
    }.dropWhile(_.size < n)
```

Here I only needed to specify one type for scan (I think that would probably infer correctly in Scala 3). Both intellij and vs code were able to infer the correct method return type.

Currently I'm forced to use ZPipeline, because certain operations are only available as ZPipelines (utf8decode and splitLines for example). 

What I like about this encoding is that I can separate the processing of streams from the source of the stream, but the typesignature of ZPipeline and additional overhead of having to add type ascriptions everywhere negate the benefit for me at least.

adamgfraser:
Resolves #6122.


vigoo:
This looks good! Much simpler

vigoo (approved):

PR #6148, opened 2021-12-11
Log Annotations

### LINKED ISSUE #6146: zio 2.0: provide equivalent of `_.locally(SomeLogAnnotation, value)` in new logging mechanism
`zio-logging` provides a way to store a key-value pair (e.g. "correlationId" -> "abcd", "userId" -> "domdorn") in the logger which is then passed down to MDC and thus will be transfered to third party systems like kibana. 
The new logging functionality in zio 2.0-M6-2 provides `LogSpan`s, however they are there for measuring how long a span was used. 
A functionality like `_.locally(..)` is missing and should be provided for upgrading 1.x apps.

adamgfraser:
Resolves #6146.

Allows adding arbitrary annotations to logging. The default logger just converts the values to strings but other loggers could potentially check if the value associated with a particular key was of an expected type for structured logging.


adamgfraser:
Updated the annotations to be more strongly typed. This is basically incorporating the annotations functionality from ZIO Logging at this point but does provide a more powerful and type safe interface.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4604,6 +4604,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def log(message: => String)(implicit trace: ZTraceElement): UIO[Unit] =
     new Logged(ZLogger.stringTag, () => message, trace = trace)
 
+  /**
+   * Logs the specified annotation.
+   */
+  def logAnnotations[V](key: => LogAnnotation[V], value: => V): LogAnnotations[V] =
```
domdorn, line 4610: I would call this logAnnotate as its just adding one annotation

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -5792,6 +5798,13 @@ object ZIO extends ZIOCompanionPlatformSpecific {
       }
   }
 
+  final class LogAnnotations[V](val key: () => LogAnnotation[V], val value: () => V) {
```
domdorn, line 5801: also here, I'd rename this to `LogAnnotation` as its just one key-value pair

domdorn: ok, as we already have that class name in the same package, maybe `LogAnnotationEntry` ( from Map->Entries )?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -5333,6 +5333,12 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
   def logInfo(message: => String)(implicit trace: ZTraceElement): ZStream[Any, Nothing, Unit] =
     ZStream.fromZIO(ZIO.logInfo(message))
 
+  /**
+   * Logs the specified message at the informational log level.
```
domdorn, line 5337: here its "informational", in ZManaged its "runtime"

adamgfraser: Fixed in my latest commit.

core/shared/src/main/scala/zio/ZIO.scala
```diff
@@ -4652,6 +4652,12 @@ object ZIO extends ZIOCompanionPlatformSpecific {
   def logInfo(message: => String)(implicit trace: ZTraceElement): UIO[Unit] =
     new Logged(ZLogger.stringTag, () => message, someInfo, trace = trace)
 
+  /**
+   * Sets the log level for this effect.
+   */
+  def logLevel(level: LogLevel): LogLevel =
```
domdorn, line 4658: hmm.. this looks wrong

adamgfraser: No `logLevel` has an `apply` method so you use it like `logLevel(LogLevel.Fine)(zio)`.

domdorn: could you please add this example to the scaladoc above? I wouldn't have guessed that from the signature alone

adamgfraser: It is already in the documentation.

domdorn: I meant like 
```scala

/**
   * Sets the log level for this effect.
   *
   * Use this to wrap your effect like 
   * <code>
   * ZIO.logLevel(LogLevel.Fine)(myEffect)
   * </code>
   */
  def logLevel(level: LogLevel): LogLevel =
   level
   ```

Comment (adamgfraser) [reply core/shared/src/main/scala/zio/ZIO.scala:4663]: Oh sure. Added now.

File: core/shared/src/main/scala/zio/LogAnnotation.scala [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio
+
+final class LogAnnotation[V] private (
```
vigoo, line 19: I know this is the same as in `zio-logging` and that `initial` and `combine` makes it possible to generalize for example the name stack and custom annotations but in practice I found that all the annotations I'm creating are such that `V` is an option type and it feels a bit annoying having to write `Option` and `Some` everywhere - would be nice if I want to add a log annotation of a request ID then I create a `LogAnnotation[RequestId]` without having to deal with specifying that there is by default no request ID, and that there can be only one request ID. 

Probably we could have this as an advanced log annotation and have something called `LogAnnotation` for the more simpler use case built on top of it?

streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -5284,6 +5284,81 @@ object ZStream extends ZStreamPlatformSpecificConstructors {
   def iterate[A](a: => A)(f: A => A)(implicit trace: ZTraceElement): ZStream[Any, Nothing, A] =
     unfold(a)(a => Some((a, f(a))))
 
+  /**
+   * Logs the specified message at the current log level.
+   */
+  def log(message: => String)(implicit trace: ZTraceElement): ZStream[Any, Nothing, Unit] =
```
vigoo, line 5290: Maybe we want these for sinks too?

core/shared/src/main/scala/zio/LogAnnotation.scala
adamgfraser: Yeah I tried to go in that direct by specifying default arguments for the `combine` and `render` parameters. It seems like the most idiomatic thing to do is just to specify some default value for `V` like `LogAnnotation("key", "default")`. That depends a little on your data type whether there is some default value you can use.

We could do something like:

```scala
def optional[V](identifier: String)(implicit tag: Tag[V]): LogAnnotation[V] =
  LogAnnotation(identifier, null.asInstanceOf[V])
```

That requires the implementers of `ZLogger` to deal with the possibility of null values, though it is a relatively low level interface. I'm also not sure the ergonomics are that much better because then you have to specify a type parameter like `LogAnnotation.optional[String]("key")`.

vigoo: One alternative idea:

- Let's have a different input and storage type (parameter) of log annotations. So we can have a `LogAnnotation[V, Option[V]]` that accepts a `V` when you annotate, but produces an `Option[V]` for the renderer.
- Maybe rename this type from `LogAnnotation` to something like `LogAnnotationBase` or `RawLogAnnotation`, I don't know; in order to make `LogAnnotation` a simplified version
- Make a simplified version that the users can find by default (so I would call this one `LogAnnotation`, probably) that defaults to the option-backed version. While more advanced annotations can still directly use the full interface to provide all types and combine/initial/render

domdorn: hmm.. why we need optional log annotations? Either I annotate a certain effect or I don't. Whatever is there gets logged, what is not, does not get logged?

domdorn: Or are we actually now mixing LogAnnotations with something like a `RequestContext`  ?

vigoo: If you look at the "built-in" annotations, `Name` for example is a stack of names, with default value being an empty list and `combine` adding the name to the stack, while `Throwable` is something that is either there or not, represented by an `Option[Throwable]` annotation.

Personally I'm using zio-logging's similar capability to annotate logs with contextual info like request id, user id, etc., I believe that is the primary reason for having these annotations user-defineable.

domdorn: yeah, I know the annotations in `zio-logging`.. I also annotate my logs with request-id, user-id, correlation-id etc.. but never had the need for a value anything else than Some(v), so the only reason for me to have this as a Option would be to say "by default this key should be there, but the value may or may not be there.. if its missing, something is wrong, please search for the error" .. I must admit, I only ever used the feature to work together with the `Slf4jLogger.makeWithAnnotationsAsMdc(...)` thingy which basically is a Map[String, String] behind the scenes. 

However, I see the beauty in having special types like `Option`s, `List`s, `Stack`s, etc., I'm just not sure we need to have "special" support for `Option`?

vigoo: I think we are on the same page, my idea was exactly to make this most common case more convenient. Maybe one difference that for me the optionality of these contextual annotations is quite common - for example I may log things about a request that I could not authenticate and assign with a user id

adamgfraser: I think the issue is if you use something like a `UUID` for your annotation. So in `zio-logging` as well as in this PR you would do something like `val correlationID = LogAnnotation("user-id", UUID.fromString("0-0-0-0-0")` (this is basically taken from the `zio-logging` documentation). 

This is fine but I think what @vigoo is saying is the `default` value here is somewhat arbitrary and in some cases there may not be a logical default value so it would be nice to not have to specify it at all, and write something like `val correlationId = LogAnnotation[UUID]("user-id")`

domdorn: > This is fine but I think what @vigoo is saying is the `default` value here is somewhat arbitrary and in some cases there may not be a logical default value so it would be nice to not have to specify it at all, and write something like `val correlationId = LogAnnotation[UUID]("user-id")`

Exactly. I would have something like 
```scala
object Annotations { 
 val UserId = LogAnnotation[String]("userId")
 val CorrelationId = LogAnnotation[CorrelationId]("correlationId")
}
```

and then in my controller/filter/whatever would do 

```scala
 ...
 import {Annotations => l}
 
 myEffect.logAnnotate(l.UserId, jwt.userId).logAnnotate(l.CorrelationId, c)
 ```
 
 So I guess the question is if we need to preserve the information that a annotation should be there, but was not filled (which we could model with optional values). If we skip optionals / empty list anyway when rendering, there's no need for that.

Comment (domdorn) [reply core/shared/src/main/scala/zio/LogAnnotation.scala:19]: on the other hand, if we have complex data types like 
```scala
val Actions = LogAnnotation[List]("actions", empty = List.empty, combine = (a,b) => a ++ b, render => v => Json.toObj(v) } 
```
we can't necessarly fix the render type to string.. e.g. in the above case, the list could be rendered as Json Object or Array or `v.mkString(",")` 

of course, a `ZLogger` could match on the return value of the `render` call and then e.g. decide what to do with a `Seq` ..

adamgfraser: At least from my perspective we want to keep the typed logging but just make it more convenient for the common case. Let me play with this.

adamgfraser:
@jdegoes @pshemass This is ready for review.

pshemass (approved):

core/shared/src/main/scala/zio/ZLogger.scala
```diff
@@ -221,6 +236,27 @@ object ZLogger {
...
+          first = false
+        } else {
+          sb.append(" ")
+        }
+
+        val (key, value) = it.next()
+
+        sb.append(key)
+        sb.append("=")
+        sb.append(value)
```
jdegoes, line 256: Maybe `appendQuoted` here??

jdegoes (commented):
Looks great! Seems like the "right level" for ZIO: not too little, not too much. Other libraries like ZIO Logging could take it to a fancier and / or more structured level.

It would also be nice to get a `ZIOAspect` into place:

```scala
handleRequest(request) @@ annotated("userId", request.userId)
```

Maybe with a vararg overload for tuples.

adamgfraser:
@jdegoes Done now!

kitlangton (approved):

PR #6152, opened 2021-12-11
Integrated fiber dumps

jdegoes:
Fiber dumps have been integrated into Core. As part of these changes:

 - `Fiber#children` now returns the children of the specified fiber, using the new `evalOn` machinery to safely copy the contents of the weak set backing the children fibers
 - `Fiber.roots` now returns the root fibers, which are those created implicitly from `Runtime#unsafeRun*` family of methods
 - `RuntimeConfigFlag.EnableFiberRoots` can now be used to turn on or off root fiber tracking. If off, then fiber dumps won't work
 - `Fiber.dump*` family of methods has been pruned and cleaned up; now only two methods exist, `Fiber.dumpAllWith`, and `Fiber.dumpAll`.
 - Stack traces in fiber dumps (`Fiber.Dump`) are no longer optional; they always appear
 - Dead code has been deleted
 - A `WeakConcurrentBag` has been introduced to aid tracking root fibers. This achieves zero allocations in the happy path and allows (weakly consistent) concurrent iteration
 - Fiber dumps can be triggered locally by SIGINT (Windows) or SIGINFO / SIGUSR1 (elsewhere); signals are emulated on Scala.js by URL hash changes (e.g. zio-signal=INFO in hash).


core/shared/src/main/scala/zio/Runtime.scala
```diff
@@ -362,6 +363,8 @@ trait Runtime[+R] {
   private final def unsafeRunWith[E, A](
     zio: ZIO[R, E, A]
   )(k: Exit[E, A] => Any)(implicit trace: ZTraceElement): FiberId => (Exit[E, A] => Any) => Unit = {
+    val enableRoots = runtimeConfig.flags.isEnabled(RuntimeConfigFlag.EnableFiberRoots)
```
vigoo, line 366: This is not used for anything

adamgfraser (dismissed):
Looks great!

core/shared/src/main/scala/zio/Runtime.scala
jdegoes: Fixed

vigoo (dismissed):

vigoo (approved):

PR #6153, opened 2021-12-11
the test shows that ZLayer could ignore actual values and use cached "image" of dependency 

alek-w:
the test shows that ZLayer could ignore actual values and use cached "image" of dependency

### DIFF
--- a/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/ZLayerSpec.scala
@@ -390,6 +390,26 @@ object ZLayerSpec extends ZIOBaseSpec {
           _     <- layer.build.useNow
           value <- ref.get
         } yield assert(value)(equalTo("bar"))
+      },
+      testM("caching values in dependencies") {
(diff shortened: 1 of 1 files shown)


adamgfraser:
I think this is expected behavior. Layers are memorized so the A layer is only built once and made available wherever it is needed in the dependency graph. If you want to build it again you can use the `fresh` operator.

guersam:
I'm actually relying on this behavior. For example: 

```scala
val userRepositoryLayer: URLayer[Database, UserRepository] = ???
val archivedUserRepositoryLayer: URLayer[Database, ArchivedUserRepository] = ???

val mainDB: ULayer[Database] = ???
val archiveDB: ULayer[Database] = ???

val main: RIO[UserRepository & ArchivedUserRepository, Unit] = ???

main.provideCustomLayer(
  (mainDB >>> userRepositoryLayer) ++
  (archiveDB >>> archivedUserRepository)
)
```

alek-w:
@adamgfraser is this test needed in the main branch?

PR #6169, opened 2021-12-15
Streams scalafix

swoogles:
Doesn't include ZSink changes or the more complex Stream ones, but knocks out all of the easier renames in ZStream.


adamgfraser (approved):
Awesome!

PR #6181, opened 2021-12-17
Fix Off By One Error In Chunk Update

### LINKED ISSUE #6180: ZIO 2.0-RC1: Chunk throws mysterious ArrayOutOfBoundsException
Hi!

I'm getting a rather mysterious ArrayOutOfBoundsException from Chunk.

Here's the smallest reproducer that I have so far:

```scala
import zio.Chunk

case class Grid[T](
  w: Int,
  h: Int,
  private val init: T,
  private val data: Chunk[Chunk[T]]) {

  private def get(x: Int, y: Int): T =
    data(y)(x)

  private def updated(x: Int, y: Int, t: T): Grid[T] =
    copy(data =
      data.updated(y,
        data(y).updated(
          x, t)))

  private def points: Iterable[(Int, Int)] = for {
    y <- 0 until h
    x <- 0 until w
  } yield (x, y)

  def map(f: T => T): Grid[T] = {
    points.foldLeft(this) { case (acc, (x, y)) =>
      Console.println(s"p = {$x, $y}")
      acc.updated(x, y, f(get(x, y)))
    }
  }

}

def fill[T](w: Int, h: Int)(init: T): Grid[T] =
  Grid(
    w = w,
    h = h,
    init = init,
    data = Chunk.fill(h)(Chunk.fill(w)(init)))

val g = fill(w = 66, h = 256)(20)

// Throws:
Console.println(g.map(_ + 1))
```

Running this shows

```txt
(snip)
p = {54, 3}
p = {55, 3}
p = {56, 3}
p = {57, 3}
p = {58, 3}
java.lang.ArrayIndexOutOfBoundsException: Index 256 out of bounds for length 256
  at zio.Chunk$Update.apply(Chunk.scala:1362)
  at Grid.updated(<console>:13)
  at Grid.$anonfun$map$1(<console>:24)
  at scala.collection.IterableOnceOps.foldLeft(IterableOnce.scala:646)
  at scala.collection.IterableOnceOps.foldLeft$(IterableOnce.scala:642)
  at scala.collection.AbstractIterable.foldLeft(Iterable.scala:926)
  at Grid.map(<console>:22)
  ... 39 elided
  ```
  
  and I'm a bit mystified where the 256 in "Index 256 out of bounds" is coming from. Obviously, the "length 256" bit comes from the length I've specified.

As you can see from the output we're querying indexes 58 and 3.

The code works fine on ZIO 1.0.12.

I'll try to minimize the reproducer further.

### PR DESCRIPTION
**Pull request by adamgfraser**
Resolves #6180.

### CODE REVIEW

**Review by regiskuckaertz — DISMISSED** [review]

File: core/shared/src/main/scala/zio/Example.scala [hunk 1]
```diff
@@ -0,0 +1,27 @@
+import zio.Chunk
+
+object Example extends scala.App {
```
vigoo, line 3: example

adamgfraser: Fixed now!

vigoo (approved):

PR #6191, opened 2021-12-21
Implement Pipelines as Channels

adamgfraser:
Implements pipelines as channels, allowing pipelines to be composed with both streams and sinks.


vigoo:
I'm wondering if the duplicated code of streams and these new pipelines could be somehow unified in something like a `ChunkedChannel`? (and `ZChannel` itself if there is anything not depending on the input/output being chunked)

vigoo:
Ok sorry I realized that you refactored those stream combinators to just use the pipelines which is probably the right thing to do.

vigoo (approved):

PR #6194, opened 2021-12-21
Docs: Purpose of `Z` prefix on data types across ZIO ecosystem

khajavi:
addresses https://github.com/zio/zio-prelude/issues/825

### DIFF
--- a/docs/about/faq.md
+++ b/docs/about/faq.md
@@ -1,0 +1,17 @@
+---
+id: faq
+title: "Frequently Answered Questions (FAQ)"
+sidebar_label: "FAQ"
+---
+
+In this page we are going to answer general questions related to the ZIO project.
+
+## In ZIO ecosystem, there are lots of data types which they have `Z` prefix in their names. What this prefix stands for? Does it mean, that data type is effectual?
+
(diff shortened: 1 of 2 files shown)


khajavi:
@adamgfraser 
This is ready for your review.

adamgfraser (approved):
Good idea adding a FAQ. There is probably a lot of good stuff that can go here. How do you think this should relate, if at all, to the Wiki on the ZIO Github repository?

khajavi:
What I know is that the majority of the information we will discuss on this page is geared toward developers who have yet to adopt or are new to ZIO. So I think it would be better to put them in a place which is more accessible to newcomers, maybe not on the microsite. But I'm not sure how this relates to the Wiki.
What are your thoughts on this?

adamgfraser:
Both the website and the Wiki seem like places where we could have this kind of frequently asked questions material. The Wiki is I think a little easier to update, which could be nice to add more content from other users in the future. However, I think you are right that the website is the first point of contact. I'm fine with adding here and if more content does end up getting added to the Wiki we can always link to it or pull in particularly relevant content.

PR #6196, opened 2021-12-22
Release Outer Resource if Inner Acquire Fails in ZStream.unwrapManaged

adamgfraser:
In the `ChannelExecutor` interpretation of `BracketOut` if `acquire` fails then `release` is not run. This makes sense in general but can create problems when managed resources are implemented in terms of this because `acquire` can fail for an inner resource but an out resource that was successfully acquired still needs to be released.


jdegoes:
This argues that something closer to `ensuring` is the atom, rather than acquire/release.

jdegoes (approved):

adamgfraser:
Yes. I think that would require the ability to control interruptibility at the channel level.

PR #6199, opened 2021-12-22
Update ZIO Environment Documentation

khajavi:
- [x] Introduction to Contextual Data Types
- [x] Remove deprecated Module Pattern 1.x, to encourage only one method to create ZIO services.
- [x] [The 3 Laws of ZIO Environment](https://twitter.com/jdegoes/status/1463261876150849547)
- [x] ZEnvironment
- [x] Rewrite how to use macros article to reflect recent changes 
- [x] Document accessible macro and Accessible trait
- [x] Move mockable macro to its page 
- [x] Global and Local Environment (scopes)
- [x] Built-in and Manual Layer Memoization
- [x] Rename "Module Pattern" to "Service Pattern" 
- [x] Macros


docs/datatypes/contextual/index.md
```diff
@@ -323,24 +225,102 @@ object LoggingLive {
...
-Finally, we provide required layers to our `app` effect:
+## The Three Laws of ZIO Environment
 
-```scala
- val app = Logging.log("Application Started")
+When we are working with the ZIO environment, one question might arise: "When should we use environment and when do we need to use constructors?".
+
+Using ZIO environment follows three laws:
+
+1. **Service Interface (Trait)** — When we are defining service interfaces, we shouldn't care about implementation details like what dependencies we have. So at this level, we should _never_ use the environment. For example, the following service definition is wrong:
```
jdegoes, line 268: Let's clarify a bit here that we should _never_ use the environment for *dependencies of the service itself*.

For example, if implementation of service X depends on services Y and Z, then these should *never* be reflected in the trait that defines Service X. It's leaking implementation details.

In rare occasions, all of which involve _local context_ that is _independent of implementation_, it's acceptable to use the environment in the definition of a service.

Two examples come to mind:

1. In a web application, a service may be defined only to operate in the context of an HTTP request. In such a case, the request itself could be stored in the environment: `ZIO[HttpRequest, ...]`. This is acceptable because this use of the environment is part of the semantics of the trait itself, rather than leaking an implementation detail of some particular class that implements the service trait.
2. In a database application, a service may be defined only to operate in the context of a larger database transaction. In such a case, the transaction could be stored in the environment: `ZIO[DatabaseTransaction, ...]`. As in the previous example, because this is part of the semantics of the trait itself (whose functionality all operates within a transaction), this is not leaking implementation details and therefore it is valid.

While it's better to err on the side of "don't put it into the environment", there are cases where it's acceptable, and we should provide a few examples so more expert users have guidance on best practices.

khajavi: Thank you for your explanations. I addressed them with two examples.

jdegoes (approved):

PR #6200, opened 2021-12-22
Pull Only As Needed in ZStream#zipAllWith

adamgfraser:
Prevents a potential memory leak by only pulling from both sides if we need additional input from both sides. I also deleted the `Exec` variant. It seemed strange that we had this variant for `zipAllWith` but not `zipWith`, it didn't appear to be being used anywhere, and since we only use it in the case when we need input from both sides at the same time it seemed less important.


adamgfraser:
Alternatively we could add an `Exec` variant of `zipWith`. Not sure.

vigoo (approved):

vigoo (approved):

PR #6202, opened 2021-12-22
Improve Reporting of Test Timing

### LINKED ISSUE #6197: zio-test: including parallelization info in test output
I think it would be nice if both wall clock time and CPU time (is this the right term for it?) were outputted when the tests complete. My reasoning is:
- It would show you how much time you're saving by having parallelized the tests.
- It can show you how parallelizable your code currently is in relation to the number of cores available. If the number is low it would be a hint that there is an opportunity for speedup gains.
- Having output separated this way makes it obvious that your tests are running in parallel. I often see people not realize that zio-test is parallel by default, so this would likely clear up that confusion as well.

For example, in Haskell there is sydtest and it outputs like this:

image

I don't have strong opinions on the formatting/presentation of this, but just thought I'd include an example of what I mean.

adamgfraser:
Resolves #6197.

I've always regarded the way the duration of a spec is reported as the sum of the durations of its children, even though the children are typically run in parallel, as a bit of a wart. For example, if you had two tests that ran in parallel and each took four seconds the time for the suite would be reported as eight seconds.

This PR implements a more sophisticated data type for representing test durations that keeps track of the start and end times for each test. Composition is then defined based on the earliest start time and latest end time. This way in the example above the total time for the suite would be correctly reported as four seconds.

### DIFF
--- a/test/shared/src/main/scala/zio/test/TestDuration.scala
+++ b/test/shared/src/main/scala/zio/test/TestDuration.scala
@@ -1,0 +1,50 @@
+package zio.test
+
+import zio._
+
+import java.time.Instant
+
+sealed trait TestDuration { self =>
+  import TestDuration._
+
+  final def <>(that: TestDuration): TestDuration =
+    (self, that) match {
+      case (Zero, Zero)  => Zero
+      case (Zero, right) => right
+      case (left, Zero)  => left
+      case (Finite(leftStart, leftEnd), Finite(rightStart, rightEnd)) =>
+        val start = if (leftStart.isBefore(rightStart)) leftStart else rightStart
+        val end   = if (leftEnd.isAfter(rightEnd)) leftEnd else rightEnd
+        Finite(start, end)
+    }
+
+  final def isZero: Boolean =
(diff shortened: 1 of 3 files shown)


test/shared/src/main/scala/zio/test/TestDuration.scala
```diff
@@ -0,0 +1,50 @@
...
+import zio._
+
+import java.time.Instant
+
+sealed trait TestDuration { self =>
+  import TestDuration._
+
+  final def <>(that: TestDuration): TestDuration =
+    (self, that) match {
+      case (Zero, Zero)  => Zero
```
regiskuckaertz, line 12: This case seems redundant.

regiskuckaertz (dismissed):

test/shared/src/main/scala/zio/test/TestDuration.scala
adamgfraser: Yes. I was thinking of enumerating all four possible cases but you're right.

jdegoes (approved):

PR #6205, opened 2021-12-23
Make Console.readLine Interruptible

### LINKED ISSUE #780: `getStrLn` uninterruptible
In case this is unknown, the following never terminates (at least on Windows/JVM) `getStrLn.fork.flatMap(_.interrupt)`.

`blocking(getStrLn).fork.flatMap(f => ZIO.sleep(1.second) *> f.interrupt)` does not terminate either.

I've wanted this to work in order to do `longRunningComputation race getStrLn` to allow cancellation from the user through the command line.

Perhaps it's just my terminal, but I've gotten around this by polling with `scala.Console.in.ready()` while looping `scala.Console.in.read(charBuffer)` to read the input. This seems to interrupt as expected.

P.S. Thanks for an awesome library! It provides so many abstractions I've longed for.

adamgfraser:
Resolves #780.

This PR takes a different approach than that discussed in the issue by interrupting the thread that is waiting on the input using `attemptBlockingInterrupt` rather that the underlying source. With this change, the following example works.

```scala
object Example extends ZIOAppDefault {

  val run =
    for {
      _     <- Console.printLine("What's your name?")
      fiber <- Console.readLine.fork
      _     <- ZIO.sleep(5.seconds)
      _     <- fiber.interrupt
      _     <- Console.printLine("You didn't give me any input!")
      name  <- Console.readLine
      _     <- Console.printLine(s"Hello, $name!")
    } yield ()
}

// What's your name?
// You didn't give me any input!
// Adam
// Hello, Adam!
```


jdegoes:
[Seems to work](https://scastie.scala-lang.org/qxj8lgjpTaOJvvwfpjBomw) although I wonder if this is Java-version-specific, as it is widely believed to not work, and I have tested it before (years ago) and could not make it work.

jdegoes (approved):

jdegoes:
Indeed, [closing the socket](https://stackoverflow.com/questions/3595926/how-to-interrupt-bufferedreaders-readline) seems to be the only way to interrupt this according to common wisdom.

PR #6209, opened 2021-12-24
Make ZScope.Local Private

adamgfraser:
It exposes internal implementation regarding `FiberContext` and the `WeakReference` and there does not appear to be any reason that others need access to it.


adamgfraser:
Also implemented a convenience method to access the current scope without having to go through `descriptor`.

jdegoes (approved):

PR #6212, opened 2021-12-26
Add ZChannel.map

sideeffffect:
* add `ZChannel.map`
* implement `mapOut` and `identity` in terms of it
* use in `ZPipeline.map`

### DIFF
--- a/streams/shared/src/main/scala/zio/stream/ZChannel.scala
+++ b/streams/shared/src/main/scala/zio/stream/ZChannel.scala
@@ -683,16 +683,8 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
 
   def mapOut[OutElem2](
     f: OutElem => OutElem2
-  )(implicit trace: ZTraceElement): ZChannel[Env, InErr, InElem, InDone, OutErr, OutElem2, OutDone] = {
-    lazy val reader: ZChannel[Env, OutErr, OutElem, OutDone, OutErr, OutElem2, OutDone] =
-      ZChannel.readWith(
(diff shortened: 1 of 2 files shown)


streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -1345,6 +1325,15 @@ object ZChannel {
       releaseMap.releaseAll(exit, ExecutionStrategy.Sequential)
     }.mapOut(_._1)
 
+  def map[Err, InElem, OutElem, Done](
```
adamgfraser, line 1328: `mapOut`

streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -1345,6 +1325,15 @@ object ZChannel {
...
+    f: InElem => OutElem
+  )(implicit trace: ZTraceElement): ZChannel[Any, Err, InElem, Done, Err, OutElem, Done] =
+    readWith(
+      (in: InElem) => write(f(in)) *> map[Err, InElem, OutElem, Done](f),
```
adamgfraser, line 1332: Type inference is pretty terrible here.

adamgfraser (commented):
There are a lot of `ZChannel` operators that could be written as channel constructors. Can definitely see them being useful in some situations but does create additional ways to do things and type inference is generally going to be worse. You can also always construct a channel like this by doing `ZChannel.identity.mapOut(f)`.

sideeffffect:
Ah, yes, I see. In that case this PR doesn't bring much to the table.
If you see anything worth salvaging, please let me know, otherwise I'll just close it :+1:

PR #6213, opened 2021-12-26
Add ZChannel convertors to ZPipeline, ZSink and ZStream

sideeffffect:


streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -933,6 +933,88 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
...
+//        InDone,
+//        Predef.identity,
+//        Out,
+//        Predef.identity
+//      )
+      self.asInstanceOf[ZChannel[Env, Any, Any, Any, OutErr, Chunk[Out], Any]]
+    )
+
+  def transform[InErr2, InElem2, InDone2, OutErr2, OutElem2, OutDone2](
+//    env: Env2 => Env,
```
sideeffffect, line 992: How can I transform the Environment? :thinking:

streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -933,6 +933,88 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
...
+    new ZPipeline[Env, OutErr, In, Out](
+//      transform[Nothing, Chunk[In], Any, OutErr, Chunk[Out], Any](
+//        Predef.identity,
+//        In,
+//        InDone,
+//        Predef.identity,
+//        Out,
+//        Predef.identity
+//      )
+      self.asInstanceOf[ZChannel[Env, Nothing, Chunk[In], Any, OutErr, Chunk[Out], Any]]
```
sideeffffect, line 951: I cheat by using `asInstanceOf`, but get a faster code, because the transformation is only busywork to make type checker happy. What do you think about it?

adamgfraser: I think it is fine to use `asInstanceOf` here.

streams/shared/src/main/scala/zio/stream/ZChannel.scala
adamgfraser: This needs to be a function from `ZEnvironment[Env2]` to `ZEnvironment[Env]` and then you can use `provideSomeEnvironment` which I just added in #6214.

streams/shared/src/main/scala/zio/stream/ZChannel.scala
```diff
@@ -933,6 +933,88 @@ sealed trait ZChannel[-Env, -InErr, -InElem, -InDone, +OutErr, +OutElem, +OutDon
         ZIO.suspendSucceed(interpret(exec.run().asInstanceOf[ChannelState[Env, OutErr]]))
       }
 
+  def toZPipeline[In, Out](implicit
```
adamgfraser, line 936: `toPipeline`, `toSink`, and `toStream`.

streams/shared/src/main/scala/zio/stream/ZChannel.scala
adamgfraser: Not sure about the ergonomics of having this "super" method versus just having more specific operators like `map`, `mapOut`, `contramapIn`, `provideSomeEnvironment`, and so on that transform individual type parameters and you can compose to do any transformation you want.

adamgfraser (commented):
I wonder how much benefit there is to having these operators versus just doing `new ZStream(channel)`. There is potentially a lack of clarity that we're not really turning a channel into a stream, a stream *is* a channel. There is also some potential for confusion with other `to` methods that generally send the outputs of a channel somewhere else, versus this is just wrapping the channel in a more specialized type.

streams/shared/src/main/scala/zio/stream/ZChannel.scala
sideeffffect: I removed `transform`.

sideeffffect:
I renamed the methods to `asSink`, `asPipeline`, `asStream`, I hope they're clearer now

adamgfraser:
Let's stick with `to` and we can revisit the other operator names if needed.

sideeffffect:
OK, renamed :+1:

adamgfraser (approved):

adamgfraser (approved):

PR #6214, opened 2021-12-26
Implement Environment Operators for Channels

adamgfraser:
If we are going to have channels we might as well have the same operators for working with the environment on them as we do for other ZIO data types.


sideeffffect (approved):
:heart:

jdegoes (approved):

PR #6233, opened 2022-01-01
Update docs about zio-aws

vigoo:
Updates the link and the example to the ZIO 2 version of zio-aws. My intention is to keep the old link and example for the ZIO 1.x documentation. I was not sure how the dual 1.x/2.x website is built so if this is not the way let me know ;)

I also removed the zio-aws dependencies from the `docs` project because the rest of the deps are not yet updated to ZIO 2 versions and it would make versions collide. I think the example was not in an `mdoc` block anyway.


khajavi:
@vigoo 
Temporarily, we exclude all snippets from the mdoc compilation in the `series/2.x` branch, so it does not matter at this time. Finally, we should update all examples and include them in the mdoc block.

khajavi (dismissed):

adamgfraser (approved):

PR #6234, opened 2022-01-01
Adds withDynamicTags to metrics

vigoo:
While migrating `zio-aws` to the new metrics API I ran into the problem of defining a metric aspect in which the tags depend on the result value of the wrapped effect. In this particular example, each AWS call returns a `Described[A]` value where the `Described` wrapper contains metadata about the call such as AWS service and operation name, which could be used as tags of the metric.

I tried to find a way in which required changes to the new the metric API are minimal while all the predefined and user defined metric aspects can be somehow used with this more dynamic tagging approach.

The result is a `def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A]` on each metric type which converts them to a metric which no longer exposes any metric-type specific API (as the tags are unknown) but are still valid metric aspects.

### DIFF
--- a/core/shared/src/main/scala/zio/ZIOMetric.scala
+++ b/core/shared/src/main/scala/zio/ZIOMetric.scala
@@ -292,20 +292,20 @@ object ZIOMetric {
      * effect's result value
      */
     def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] = {
-      if (self.counterRef eq null) {
-        self.counterRef = ZFiberRef.unsafeMake(self.counter)
-        self.counter = null
-      }
+      val cloned = copy()
+      cloned.counterRef = ZFiberRef.unsafeMake(cloned.counter)
+      cloned.counter = null
+
       new ZIOMetric[A] {
         override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
-          appliedAspect(zio.tap(changeCounter))
+          cloned.appliedAspect(zio.tap(changeCounter))
 
         private def changeCounter(value: A)(implicit trace: ZTraceElement): UIO[Unit] =
-          self.counterRef.update { counter =>
+          cloned.counterRef.update { counter =>
             val extraTags = f(value)
-            val allTags   = self.tags ++ extraTags
+            val allTags   = cloned.tags ++ extraTags
             if (counter.metricKey.tags != allTags) {
-              internal.metrics.Counter(self.name, allTags)
+              internal.metrics.Counter(cloned.name, allTags)
             } else {
               counter
             }
--- a/core/shared/src/main/scala/zio/ZIOMetric.scala
+++ b/core/shared/src/main/scala/zio/ZIOMetric.scala
@@ -268,7 +268,7 @@ object ZIOMetric {
      * Converts this counter metric to one where the tags depend on the measured
      * effect's result value
      */
-    def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
+    def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
       new ZIOMetric[A] {
         override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
           self.apply(zio.tap(changeCounter))
@@ -352,7 +352,7 @@ object ZIOMetric {
      * Converts this gauge metric to one where the tags depend on the measured
      * effect's result value
      */
-    def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
+    def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
       new ZIOMetric[A] {
         override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
           self.apply(zio.tap(changeGauge))
@@ -456,7 +456,7 @@ object ZIOMetric {
      * Converts this histogram metric to one where the tags depend on the
      * measured effect's result value
      */
-    def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
+    def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
       new ZIOMetric[A] {
         override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
           self.apply(zio.tap(changeHistogram))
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -263,6 +264,25 @@ object ZIOMetric {
...
+     */
+    def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
+      new ZIOMetric[A] {
+        override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
+          self.apply(zio.tap(changeCounter))
+
+        private def changeCounter(value: A)(implicit trace: ZTraceElement): UIO[Unit] =
+          ZIO.succeed {
+            val extraTags = f(value)
+            val allTags   = self.tags ++ extraTags
```
vigoo, line 279: it could also replace the tags instead of combining the static ones with the dynamic ones

core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -263,6 +264,25 @@ object ZIOMetric {
...
+            if (self.counter.metricKey.tags != allTags) {
```
vigoo, line 280: this chunk comparison can be dropped but then each invocation allocates a new `internal.metrics.Counter`

vigoo:
This is not thread safe as is, but I think can be fixed by storing the internal metric objects in fiber refs. I will fix it if in general the idea is accepted

core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -255,13 +256,34 @@ object ZIOMetric {
...
      */
     final def increment(implicit trace: ZTraceElement): UIO[Any] =
-      counter.increment(1.0)
+      counter.getWith(_.increment(1.0))
+
+    /**
+     * Converts this counter metric to one where the tags depend on the measured
+     * effect's result value
+     */
+    def withDynamicTags(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
```
jdegoes, line 271:
```suggestion
    def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] =
```

core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -278,15 +300,15 @@ object ZIOMetric {
    * values over time.
    */
   abstract class Gauge[A](final val name: String, final val tags: Chunk[MetricLabel]) extends ZIOMetric[A] { self =>
-    private[this] val gauge = internal.metrics.Gauge(name, tags)
+    private[this] val gauge = ZFiberRef.unsafeMake(internal.metrics.Gauge(name, tags))
```
jdegoes, line 303: The addition of a `FiberRef` wrapper here will impact performance. It may be better to make tags mutable if we can guard access sufficiently as to not violate referential transparency.

core/shared/src/main/scala/zio/ZIOMetric.scala
jdegoes: The happy path should be free of allocations, to the extent possible.

core/shared/src/main/scala/zio/ZIOMetric.scala
vigoo: One idea that would only make the construction non-threadsafe which I think is OK: 
- keep the non-fiber ref internal metric fields as they were but change them to `var`
- add a fiber-ref version too which is also a `var` and initialized to `null`
- the `taggedWith` sets the direct reference to `null` and initializes the fiber ref
- using a helper function to get the gauge from either the fiber ref or the direct ref to work with the metric

I was thinking a lot on how else could I achieve this - to be able to use the existing (and arbitray) aspects with dynamic tags but could not come up with anything that is better or more performant so if you have a better idea please let me know

vigoo: (by the way I did not do what I described here first because I assumed that `FiberRef` itself does something like this - that it's cheap until its never used on multiple fibers)

vigoo:
I added the above described optimization

core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -582,14 +706,40 @@ object ZIOMetric {
...
     def occurrences(implicit trace: ZTraceElement): UIO[Chunk[(String, Long)]] =
-      setCount.occurrences
+      withSetCount(_.occurrences)
+
+    /**
+     * Converts this set count metric to one where the tags depend on the
+     * measured effect's result value
+     */
+    def taggedWith(f: A => Chunk[MetricLabel]): ZIOMetric[A] = {
+      if (self.setCountRef eq null) {
```
jdegoes, line 723: I think this may violate referential transparency, because it has a side-effect which is not described in the return type. 

Will the behaviour of the original `self` differ in any way after this method is invoked?

core/shared/src/main/scala/zio/ZIOMetric.scala
```diff
@@ -582,14 +706,40 @@ object ZIOMetric {
...
+        self.setCountRef = ZFiberRef.unsafeMake(self.setCount)
+        self.setCount = null
+      }
+      new ZIOMetric[A] {
+        override def apply[R, E, A1 <: A](zio: ZIO[R, E, A1])(implicit trace: ZTraceElement): ZIO[R, E, A1] =
+          self.apply(zio.tap(changeSetCount))
+
+        private def changeSetCount(value: A)(implicit trace: ZTraceElement): UIO[Unit] =
+          self.setCountRef.update { setCount =>
```
jdegoes, line 732: You can improve performance by factoring this lambda out into a constant lambda inside the class. That way, it won't be created each time `changeSetCount` is invoked.

core/shared/src/main/scala/zio/ZIOMetric.scala
vigoo: Thanks, this made me realize that there is such a problem: if `c1` is a counter and `c2` is the modified counter, applying `c2` first modifies the `FiberRef` and does not restore it, so a next call to `c1` on the same fiber will have wrong tags.

`taggedWith` should call `self.copy`

vigoo: Actually just checked and `copy` does not work correctly currently either (without my change):

```scala
  val count = ZIOMetric
    .count("test", MetricLabel("x", "1"))
    .copy("test2", Chunk(MetricLabel("x", "2")))

  val run =
    for {
      _      <- ZIO.succeed("hello") @@ count
      states <- UIO(MetricClient.unsafeStates)
      _      <- ZIO.debug(states.toList.mkString("\n"))
    } yield ()
```

prints

```
(Counter(test2,Chunk(MetricLabel(x,2))),MetricState(test2{x->2}, Counter(0.0)))
(Counter(test,Chunk(MetricLabel(x,1))),MetricState(test{x->1}, Counter(1.0)))
```

(because the copied aspect just calls the original `self`'s `apply` which refers to the original metric).

So in addition to fix the referential transparency issue, should also:
- fix `copy`
- write some metric tests :)

adamgfraser (approved):

PR #6246, opened 2022-01-13
Implement ZChannel.fromSink

adamgfraser:
Lets you convert a sink into a pipeline by repeatedly sending elements through the sink.


streams/shared/src/main/scala/zio/stream/ZStream.scala
```diff
@@ -3950,60 +3950,7 @@ class ZStream[-R, +E, +A](val channel: ZChannel[R, Any, Any, Any, E, Chunk[A], A
   def transduce[R1 <: R, E1 >: E, A1 >: A, Z](
```
vigoo, line 3950: Maybe we should remove the whole `transduce`? It is not the same as in ZIO 1 anyway

vigoo (approved):

streams/shared/src/main/scala/zio/stream/ZStream.scala
adamgfraser: Yes, totally agree. We could use `through` here. Or perhaps `viaSink`?

PR #6256, opened 2022-01-17
Fix some deprecation warnings.

thesamet:
In some instances, the deprecation message and version are swapped.


adamgfraser (approved):
Thank you! 🙏

